PR #31, opened 2025-07-10
internal/typeutil: move clone to this internal package

hyangah:
Address comments during review
  - Limit the usage of this to our specific use cases.
  - Handle interface, array, pointer of pointer.
  - Explicitly panic if unsupported types are found.

### DIFF
--- a/internal/typeutil/clone.go
+++ b/internal/typeutil/clone.go
@@ -65,7 +65,7 @@ func clone[M any](src M) M {
 	newVal := reflect.New(val.Type()).Elem()
 
 	// Recursively copy the base value.
-	deepCopy(val, newVal)
+	deepCopy(newVal, val)
 
 	// Re-wrap the copied value with the original number of pointers.
 	finalVal := newVal
@@ -82,7 +82,7 @@ func clone[M any](src M) M {
 }
 
 // deepCopy copies src to dst using reflect.
-func deepCopy(src, dst reflect.Value) {
+func deepCopy(dst, src reflect.Value) {
 	switch src.Kind() {
 	case reflect.Struct:
 		t := src.Type()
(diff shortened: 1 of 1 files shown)


internal/typeutil/clone.go
```diff
@@ -0,0 +1,153 @@
...
+	"google.golang.org/genai"
+)
+
+// Clone returns a deep copy of the src.
+// This is not generally usable. Only selected types typeutil.cloneable
+// are allowed to use.
+// If the type is a struct type with unexported fields, or contains
+// unhandled types (function, array, channel, or embedded type),
+// this function will panic.
+func Clone[M cloneable](src M) M {
```
dpasiukevich, line 30: Would it make sense to use a library for deep copy (e.g. github.com/mohae/deepcopy)?
This should reduce maintenance toil and simplify adk-go codebase.

Or are there reasons for the in-house implementation?

jba: That particular implementation is buggy:
- Drops unexported fields instead of failing.
- Shallow-copies arrays instead of recursing through their elements.
- Shallow-copies channels and functions instead of failing.

It's extremely hard to get this right, and although I admit it's a bit of not-invented-here syndrome, I trust us to get it closer to correct than arbitrary third-party code.

I would consider copying an unexported version of this code from the standard library or an x repo.

internal/typeutil/clone.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// clonable are the types explicitly allowed to use the Clone.
+type cloneable interface {
+	*genai.GenerateContentConfig | []*genai.Content | *genai.Content
+}
+
+// clone returns a deep copy of the src.
+// NOTE: this does not work for types with unexported fields.
+func clone[M any](src M) M {
```
jba, line 41: Couldn't you get rid of `clone` and put this logic into `Clone`?

rakyll:
What's the status of this change?

hyangah:
Things were changed a lot. Will close this.

PR #34, opened 2025-07-17
agent: introduce BaseAgent

hyangah:
This implements functionalities that should be shared by all adk.Agent.
See the documentation of `BaseAgent` for how a custom agent can be
constructed. The "self" field is a workaround to reference the containing
agent type correctly. That means subclasses must use a constructor
function to create the object and correctly set the Self field.
    
This PR also introduces agent.Option type that can be used by
the constructors.

### DIFF
--- a/agent/base_agent.go
+++ b/agent/base_agent.go
@@ -15,9 +15,7 @@
 package agent
 
 import (
-	"context"
 	"fmt"
-	"iter"
 
 	"github.com/google/adk-go"
 )
@@ -69,7 +67,6 @@ func WithSubAgents(agents ...adk.Agent) Option {
 	   base, err := agent.NewBaseAgent(name, agent, opts...)}
 	   if err != nil { return nil, err }
 	   agent.BaseAgent = base
-	   agent.Self = a  // This allows methods implemented in BaseAgent can access *MyCustomAgent.
 	    ...
 	   return agent, nil
 	}
@@ -86,7 +83,7 @@ type BaseAgent struct {
 // NewBaseAgent returns a BaseAgent that can be the base of the implementation agent.
 func NewBaseAgent(name string, implementation adk.Agent, opts ...Option) (*BaseAgent, error) {
 	if implementation == nil {
-		panic("implementation is nil")
+		return nil, fmt.Errorf("implementation is nil")
 	}
 	b := &BaseAgent{name: name, self: implementation}
 	for _, opt := range opts {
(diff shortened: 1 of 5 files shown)


agent/base_agent.go
```diff
@@ -0,0 +1,101 @@
...
+//		   ....
+//		}
+//
+//		func NewMyCustomAgent(name string, opts ...agent.Option) *MyCustomAgent {
+//		   a := &MyCustomAgent{AgentBase: agent.NewAgentBase(name, opts...)}
+//		   a.Self = a  // This allows methods implemented in AgentBase can access *MyCustomAgent.
+//	        ...
+//		   return a
+//		}
+type AgentBase struct {
```
dpasiukevich: Let's rename to `BaseAgent` to be consistent with all other frameworks?

I understand the semantics here for golang are different (embedding vs inheritance). But IMO public API consistency including naming has more priority than to reflect the implementation via naming.

agent/llm_agent.go
```diff
@@ -17,15 +17,45 @@ package agent
...
 	"context"
 	"fmt"
+	"iter"
 
 	"github.com/google/adk-go"
 	"google.golang.org/genai"
 )
 
+// NewLLMAgent returns a new LLMAgent configured with the provided options.
+func NewLLMAgent(name string, opts ...Option) *LLMAgent {
```
dpasiukevich: `model` is a required parameter for LLMAgent: https://google.github.io/adk-docs/agents/llm-agents/#defining-the-agents-identity-and-purpose

WDYT if we make the constructor like `NewLLMAgent(name string, model adk.Model, opts ...Option)` to specifically ensure the required params are set?

agent/base_agent.go
```diff
@@ -0,0 +1,101 @@
...
+func (a *AgentBase) Parent() adk.Agent      { return a.parentAgent }
+func (a *AgentBase) SubAgents() []adk.Agent { return a.subAgents }
+func (a *AgentBase) Run(ctx context.Context, parentCtx *adk.InvocationContext) iter.Seq2[*adk.Event, error] {
+	panic("unimplemented")
+}
+
+func (a *AgentBase) _base_() *AgentBase { return a }
+
+// AddSubAgents adds the agents to the subagent list.
+func (a *AgentBase) AddSubAgents(agents ...adk.Agent) {
```
dpasiukevich: WDYT if we make an Option -- `WithSubAgents`? This will be then consistent with the rest of the setup, like `WithDescription` etc.

agent/base_agent.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/google/adk-go"
+)
+
+// Option configures the Base agent.
+type Option interface {
+	apply2Base(*AgentBase)
+}
+
+type optionFunc func(*AgentBase)
```
dpasiukevich: I'd suggest from the start for option to return error and check those in the constructor call. This way we'd get early validation at the constructor level already.

`type optionFunc func(*AgentBase) error`

agent/base_agent.go
```diff
@@ -0,0 +1,101 @@
...
+func (a *AgentBase) _base_() *AgentBase { return a }
+
+// AddSubAgents adds the agents to the subagent list.
+func (a *AgentBase) AddSubAgents(agents ...adk.Agent) {
+	for _, subagent := range agents {
+		a.subAgents = append(a.subAgents, subagent)
+
+		if s, ok := subagent.(interface{ _base_() *AgentBase }); ok {
+			if base := s._base_(); base != nil {
+				base.parentAgent = a.Self
```
dpasiukevich: What about the case if the parent is already set? e.g. some agent was added as a subAgent to 2 different ones?

agent/base_agent.go
```diff
@@ -0,0 +1,101 @@
...
+// For example,
+//
+//		type MyCustomAgent struct {
+//		   *AgentBase
+//		   ....
+//		}
+//
+//		func NewMyCustomAgent(name string, opts ...agent.Option) *MyCustomAgent {
+//		   a := &MyCustomAgent{AgentBase: agent.NewAgentBase(name, opts...)}
+//		   a.Self = a  // This allows methods implemented in AgentBase can access *MyCustomAgent.
```
dpasiukevich: I thought about it a bit longer than I'd expected, but still didn't find a better solution than this one (keeping this reference in the base agent). 😄 

At least we can improve one thing now -- we can also hide `Self` and make it unexported?

```
NewBaseAgent(name string, implementation adk.Agent, opts ...Option)
if implementation == nil { return err } // here we can even validate it.
```

It's slightly better IMO:
1. We can validate the implementation is set by the user. And with current `a.Self = a` user may forget to set it.
2. By moving it to the constructor we actually make the user code shorter (e.g. when they set up subAgents):
```
a := &MyCustomAgent{}
a.BaseAgent = NewBaseAgent(name, a,
      WithSubAgents(a1, a2))
```

vs current
```
a := &MyCustomAgent{AgentBase: agent.NewAgentBase(name, opts...)}
a.Self = a // this MUST be set before AddSubAgents call.
a.AgentBase.AddSubAgents(a1, a2)
```

3. hide implementation detail `Self` field. For example if it's public/exported, user code may alter it at runtime (due to bug for example), leading to nasty bugs.

agent/llm_agent.go
hyangah: Done.

agent/base_agent.go
hyangah: Done. Currently the constructor panics. Should we make the constructor return error? 

`func NewLLMAgent(...) error`

agent/base_agent.go
hyangah: I considered the option but I am afraid if that limits the order of how multi-agents are constructed too much; i.e. all subagents should be created before the parent is created.  Let's continue discussing.

agent/base_agent.go
hyangah: Good point. Added checking and made the function return error.

agent/base_agent.go
hyangah: Nice. Done.

agent/base_agent.go
hyangah: based on offline discussion, changed constructors to return an error.

Users can also define their own `Must` utility function for convenience. (as we do in our test code)

agent/base_agent.go
```diff
@@ -0,0 +1,143 @@
...
+	   *BaseAgent
+	   ....
+	}
+
+	func NewMyCustomAgent(name string, opts ...agent.Option) (*MyCustomAgent, error) {
+	   agent := &MyCustomAgent{}
+	   base, err := agent.NewBaseAgent(name, agent, opts...)}
+	   if err != nil { return nil, err }
+	   agent.BaseAgent = base
+	   agent.Self = a  // This allows methods implemented in BaseAgent can access *MyCustomAgent.
```
dpasiukevich: this line is not needed anymore

agent/base_agent.go
```diff
@@ -0,0 +1,143 @@
...
+			return a, true
+		}
+	}
+	return nil, false
+}
+
+// AddSubAgents adds the agents to the subagent list.
+func (a *BaseAgent) AddSubAgents(agents ...adk.Agent) error {
+	for _, subagent := range agents {
+		// O(n^2) search, but n is small enough.
```
dpasiukevich: If you'd like you can unwrap it for O(n) check:

1. preprocess and get a map/set of a.subAgents (e.g. map[string]struct{})
2. check if a new agent is in this map already -> if yes return error

hyangah:
Merging. We will reiterate on the API.

PR #36, opened 2025-07-23
Add basic runner implementation.

dpasiukevich:


runner/runner.go
```diff
@@ -0,0 +1,121 @@
...
+
+type Runner struct {
+	AppName        string
+	RootAgent      adk.Agent
+	SessionService adk.SessionService
+}
+
+// Run runs the agent.
+func (r *Runner) RunAsync(ctx context.Context, userID, sessionID string, msg *genai.Content, cfg *adk.AgentRunConfig) iter.Seq2[*adk.Event, error] {
+	// TODO(hakim): we need to validate whether cfg is compatible with the Agent.
```
hyangah, line 45: do we still have extra validation to do, given that r.setupCFC now exists?
(for a longer term, I wonder if we need a method in Model that does the check)

runner/runner.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if llmAgent.Model == nil {
+		return fmt.Errorf("LLMAgent has no model")
+	}
+
+	if !strings.HasPrefix(llmAgent.Model.Name(), "gemini-2") {
+		return fmt.Errorf("CFC is not supported for model: %v", llmAgent.Model.Name())
+	}
+
+	// TODO handle CFC setup for LLMAgent, e.g. setting code_executor
```
hyangah: nit: `TODO:`

runner/runner.go
dpasiukevich: I will still keep this TODO for a moment to take an extra glance later (soon to add tests and more logic there).

runner/runner.go
dpasiukevich: done

PR #38, opened 2025-07-24
agent: replace BaseAgent with AgentSpec and simplify Agent interface

hyangah:
This PR proposes to simplify the Agent interface to be

```
type Agent interface {
  Spec() *AgentSpec
  Run(context.Context, *InvocationContext) iter.Seq2[*Event, error]
  // TODO: RunLive
}
```

And consolidate common properties of ADK Agents into `AgentSpec`.
In the offline discussion, AgentSpec was suggested as a type in a sub package and used by LLMAgent and other custom agents. However, circular dependency makes it challenging and harder to maintain.
I hope it's not too bad to have AgentSpec along with the Agent.

The AgentSpec needs some initialization so it can be associated with the Agent that is containing the AgentSpec. 
This is done by `Init`.

For example, if a logic in the AgentSpec needs to call the logic implemented in the containing Agent, it needs to know which Agent implementation it should use. See the usage of `AgentSpec.self` in this draft.

Instead of making the interface include methods like `Name`, `Parent`, `SubAgents`, `Description`, ...
I propose to have just `Spec()` that returns the Agent's AgentSpec. That will help us avoid expanding the interface.
Adding methods or fields to AgentSpec won't break custom agents in the future.

For LLMAgent, this PR experiments with the variadic Option
that can configure both LLMAgent's AgentSpec and LLMAgent's own properties.


agent.go
```diff
@@ -150,3 +150,69 @@ type AgentRunConfig struct {
...
+		agentSpec *adk.AgentSpec
+		...
+	}
+	func NewMyAgent(name string) *MyAgent {
+		a := &MyAgent{agentSpec: &adk.AgentSpec{Name: name}}
+		_ = a.agentSpec.Init(a)
+		return a
+	}
+*/
+type AgentSpec struct {
```
dpasiukevich, line 170: WDYT if the constuctor is better here + unexported fields to control creation and run validation of the spec when created.

1. We then can do validation and ensure the correct internal state. For example `self` is set.
2. With the current flow with `Init(a)` users may forget to call Init, e.g. just creating agent with a spec. This will potentially uncover itself only in runtime if the spec is invalid.

With the constructor the setup flow looks like this:
```
	type MyAgent struct {
		agentSpec *adk.AgentSpec
		...
	}
	func NewMyAgent(name string) *MyAgent {
		a := &MyAgent{ ... }
        a.agentSpec = adk.NewAgentSpec(name, a, ...)
		return a
	}
```

With this we ensure users add agent spec AND the spec has a valid internal state.

agent/llm_agent.go
```diff
@@ -23,39 +23,165 @@ import (
...
+
+func WithOutputSchema(s *genai.Schema) Option {
+	return llmOptionFunc(func(a *LLMAgent) error {
+		a.OutputSchema = s
+		return nil
+	})
 }
 
-type llmAgentOptionFunc func(*LLMAgent) error
+type LLMAgentSpec struct {
```
dpasiukevich: In ADK runtime we need the access to LLMAgentSpec, example: https://github.com/google/adk-python/blob/c25911c91255e75c5d0cd5035fbe39c4187bf536/src/google/adk/runners.py#L489

Right now this struct is not used anywhere in the codebase.

WDYT should we make it a part of the AgentSpec?

For me it seems nice if agentSpec has LLMAgentSpec as a field. Coupled with the constructors ensuring valid state, we may actually understand in the if the agent is LLM-based agent or not, simply by checking if `.llmSpec` is set.

agent.go
hyangah: Good point. Constructor will be easier than requiring Init call.

But I still want to figure out what is a better pattern for Agent construction. (option struct or variadic options).

Even though currently I am experimenting the variadic options for LLMAgent, as warned in 
go/go-style/best-practices#variadic-options it adds a lot of boilerplate to the LLMAgent implementation and also to the users' code. Adding `AgentSpecOption` type to allow to configure a couple of fields also seems like an overkill now. I want to minimize the number of types in the adk package. Especially this is only useful for custom agent implementer which is rarer than LLMAgent users, I want to take a look if I can move the options and the AgentSpec construction to the agent or agentspec package. Let's keep experimenting it.

agent/llm_agent.go
hyangah: Heh, deleted. I was trying to test variadic option vs option struct, and left this unused. 

> For me it seems nice if agentSpec has LLMAgentSpec as a field. ...  simply by checking if `.llmSpec` is set.

Agree (except unexported `.llmSpec` can't be accessed from external packages). I think, if users need to build an Agent extending an LLMAgent, we will consider to define an interface that has `LLMAgentSpec() *LLMAgentSpec` and if so, treat it like an `LLMAgent` even though the Agent is not `*LLMAgent`. 
But let's see if users actually need to extend a custom LLM agent as a different type instead of utilizing the various before/after X callbacks.

agent.go
dpasiukevich: 1. I agree that we don't need the `AgentSpecOption` (at least for now).
2. I've experimented locally with `LLMAgentSpec` moved to `agent.go` as part of the API.

```
type AgentOption func(*AgentSpec) error

func WithDescription(desc string) AgentOption {
	return func(s *AgentSpec) error {
		s.Description = desc
		return nil
	}
}

// setup of the LLMAgentSpecFields
func WithInstruction(inst string) AgentOption {
	return func(a *AgentSpec) error {
		if a.llmAgentSpec == nil {
			a.llmAgentSpec = &LLMAgentSpec{}
		}
		a.llmAgentSpec.instruction = inst
		return nil
	}
}
```

I looks fine. But right now I'm actually thinking we should separate LLMAgentSpec options from AgentSpec options in the API
So it's clear what user sets. For example in the code above, user may by mistake add WithInstruction to the LoopAgent, making it considered as LLMAgent in ADK runtime.

PR #39, opened 2025-07-24
artifact: in-memory artifact service

hyangah:
This is similar to the in-memory session service, and built on top of the omap.

### DIFF
--- a/artifact.go
+++ b/artifact.go
@@ -20,28 +20,79 @@ import (
 	"google.golang.org/genai"
 )
 
-// ArtifactLoadOption is the option for [ArtifactService.Load].
-type ArtifactLoadOption struct {
+// ArtifactSaveRequest is the parameter for [ArtifactService.Save].
+type ArtifactSaveRequest struct {
+	AppName, UserID, SessionID, FileName string
+	// Part is the artifact to store.
+	Part *genai.Part
+
+	// Belows are optional fields.
+
+	// If set, the artifact will be saved with this version.
+	// If unset, a new version will be created.
 	Version int64
 }
 
-// ArtifactDeleteOption is the option for [ArtifactService.Delete].
-type ArtifactDeleteOption struct {
+// ArtifactSaveResponse is the return type of [ArtifactService.Save].
+type ArtifactSaveResponse struct {
 	Version int64
 }
 
+// ArtifactLoadRequest is the parameter for [ArtifactService.Load].
+type ArtifactLoadRequest struct {
+	AppName, UserID, SessionID, FileName string
+
+	// Belows are optional fields.
+	Version int64
+}
+
+// ArtifactLoadResponse is the return type of [ArtifactService.Load].
+type ArtifactLoadResponse struct {
+	// Part is the artifact stored.
+	Part *genai.Part
+}
+
+// ArtifactDeleteRequest is the parameter for [ArtifactService.Delete].
+type ArtifactDeleteRequest struct {
+	AppName, UserID, SessionID, FileName string
+
+	// Belows are optional fields.
+	Version int64
+}
+
+// ArtifactListRequest is the parameter for [ArtifactService.List].
+type ArtifactListRequest struct {
+	AppName, UserID, SessionID string
+}
+
+// ArtifactListResponse is the return type of [ArtifactService.List].
+type ArtifactListResponse struct {
+	FileNames []string
+}
+
+// ArtifactVersionsRequest is the parameter for [ArtifactService.Versions].
+type ArtifactVersionsRequest struct {
+	AppName, UserID, SessionID, FileName string
+}
+
+// ArtifactVersionsResponse is the parameter for [ArtifactService.Versions].
+type ArtifactVersionsResponse struct {
+	Versions []int64
+}
+
+// ArtifactService is the artifact storage service.
 type ArtifactService interface {
 	// Save saves an artifact to the artifact service storage.
 	// The artifact is a file identified by the app name, user ID, session ID, and fileName.
 	// After saving the artifact, a revision ID is returned to identify the artifact version.
-	Save(ctx context.Context, appName, userID, sessionID, fileName string, artifact *genai.Part) (int64, error)
+	Save(ctx context.Context, req *ArtifactSaveRequest) (*ArtifactSaveResponse, error)
 	// Load loads an artifact from the storage.
 	// The artifact is a file indentified by the appName, userID, sessionID and fileName.
-	Load(ctx context.Context, appName, userID, sessionID, fileName string, opts *ArtifactLoadOption) (*genai.Part, error)
+	Load(ctx context.Context, req *ArtifactLoadRequest) (*ArtifactLoadResponse, error)
 	// Delete deletes an artifact. Deleting a non-existing entry is not an error.
-	Delete(ctx context.Context, appName, userID, sessionID, fileName string, opts *ArtifactDeleteOption) error
+	Delete(ctx context.Context, req *ArtifactDeleteRequest) error
 	// List lists all the artifact filenames within a session.
-	List(ctx context.Context, appName, userID, sessionID string) ([]string, error)
+	List(ctx context.Context, req *ArtifactListRequest) (*ArtifactListResponse, error)
 	// Versions lists all versions of an artifact.
-	Versions(ctx context.Context, appName, userID, sessionID, fileName string) ([]int64, error)
+	Versions(ctx context.Context, req *ArtifactVersionsRequest) (*ArtifactVersionsResponse, error)
 }
--- a/artifact/artifact.go
+++ b/artifact/artifact.go
@@ -121,7 +121,13 @@ func (s *InMemoryArtifactService) delete(appName, userID, sessionID, fileName st
 	s.artifacts.Delete(key)
 }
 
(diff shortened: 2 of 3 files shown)


artifact.go
```diff
@@ -0,0 +1,41 @@
...
+type ArtifactService interface {
+	// Save saves an artifact to the artifact service storage.
+	// The artifact is a file identified by the app name, user ID, session ID, and fileName.
+	// After saving the artifact, a revision ID is returned to identify the artifact version.
+	Save(ctx context.Context, appName, userID, sessionID, fileName string, artifact *genai.Part) (int64, error)
+	// Load loads an artifact from the storage.
+	// The artifact is a file indentified by the appName, userID, sessionID and fileName.
+	Load(ctx context.Context, appName, userID, sessionID, fileName string, opts *ArtifactLoadOption) (*genai.Part, error)
+	// Delete deletes an artifact.
+	Delete(ctx context.Context, appName, userID, sessionID, fileName string) error
```
dpasiukevich: I find it interesting the API (as in adk-python as well) allows to have multiple versions of one artifact. But when it comes to deletion, there's only one option -- deleete the artifact and all its versions.

IMO we should proceed with the current API for starters, but I'm curious if there be any need to allow to delete specific version of the API?

artifact/artifact.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func (s *InMemoryArtifactService) Delete(ctx context.Context, appName, userID, sessionID, fileName string) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	lo := artifactKey{AppName: appName, UserID: userID, SessionID: sessionID, FileName: fileName, Version: math.MaxInt64}.Encode()
+	hi := artifactKey{AppName: appName, UserID: userID, SessionID: sessionID, FileName: fileName}.Encode()
+	n := 0
+	for key := range s.scan(lo, hi) {
```
dpasiukevich: optional: should we use artifact.DeleteRange(lo, hi)?  It should be faster + we lose the ability to return `NotFound`, but do we want to return `NotFound` if the user already wants to delete it?

artifact/artifact.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+}
+
+func (s *InMemoryArtifactService) find(appName, userID, sessionID, fileName string) (int64, *genai.Part, bool) {
+	lo := artifactKey{AppName: appName, UserID: userID, SessionID: sessionID, FileName: fileName, Version: math.MaxInt64}.Encode()
+	hi := artifactKey{AppName: appName, UserID: userID, SessionID: sessionID, FileName: fileName, Version: 0}.Encode()
+	var latestKey *artifactKey
+	var latestVal *genai.Part
+	for key, val := range s.scan(lo, hi) {
+		if key.FileName != fileName {
```
dpasiukevich: Curious how can this happen if `FileName` is a part of the key?

artifact/artifact.go
```diff
@@ -0,0 +1,216 @@
...
+			break // no matching key.
+		}
+		latestKey = &key // first key is the latest one.
+		latestVal = val
+		break
+	}
+	if latestKey == nil {
+		return 0, nil, false
+	}
+	return latestKey.Version, latestVal, true
```
dpasiukevich:
```suggestion
		return latestKey.Version, latestVal, true
	}
	return 0, nil, false
```

artifact/artifact.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/google/adk-go"
+	"google.golang.org/genai"
+	"rsc.io/omap"
+	"rsc.io/ordered"
+)
+
+// InMemorySessionService is an in-memory implementation of adk.SessionService.
+// It is primarily for testing and demonstration purposes.
+type InMemoryArtifactService struct {
+	mu sync.Mutex
```
dpasiukevich: Does it make sense to use `sync.RWMutex` and do RLock on reads (Load,List,Versions)?

artifact/artifact.go
```diff
@@ -0,0 +1,216 @@
...
+
+	versions := map[int64]bool{}
+	lo := artifactKey{AppName: appName, UserID: userID, SessionID: sessionID, FileName: fileName, Version: math.MaxInt64}.Encode()
+	hi := artifactKey{AppName: appName, UserID: userID, SessionID: sessionID, FileName: fileName}.Encode()
+	// TODO(hyangah): extend omap to search key only and skip value decoding.
+	for key := range s.scan(lo, hi) {
+		versions[key.Version] = true
+	}
+	if len(versions) == 0 {
+		return nil, fmt.Errorf("artifact not found: %w", os.ErrNotExist)
```
dpasiukevich: optional: For me at least `os.ErrNotExist` feels like the error for the OS filesystem. Would it be better to have our custom`artifact.NotFound` error?

artifact/artifact.go
```diff
@@ -0,0 +1,216 @@
...
+// Versions implements adk.ArtifactService.
+func (s *InMemoryArtifactService) Versions(ctx context.Context, appName string, userID string, sessionID string, fileName string) ([]int64, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	versions := map[int64]bool{}
+	lo := artifactKey{AppName: appName, UserID: userID, SessionID: sessionID, FileName: fileName, Version: math.MaxInt64}.Encode()
+	hi := artifactKey{AppName: appName, UserID: userID, SessionID: sessionID, FileName: fileName}.Encode()
+	// TODO(hyangah): extend omap to search key only and skip value decoding.
+	for key := range s.scan(lo, hi) {
```
dpasiukevich, line 222: Do we need a `versions` set here? I assume for appName+userID+sessionID+fileName there will be only unique versions already, so we can just append to a slice immediately? Or do we expect duplicates?

artifact.go
hyangah: Good point. For forwards compatibility, added ArtifactDeleteOption.

artifact/artifact.go
hyangah: Didn't occur to me. Added DeleteRange. I don't know. 
I made it return `NotFound` just because the Session service reports an error if the entry doesn't exist. But I am fine with just treating the case as non-error though. The python in-memory service doesn't treat it as an error either.

artifact.go
```diff
@@ -0,0 +1,47 @@
...
+	// The artifact is a file identified by the app name, user ID, session ID, and fileName.
+	// After saving the artifact, a revision ID is returned to identify the artifact version.
+	Save(ctx context.Context, appName, userID, sessionID, fileName string, artifact *genai.Part) (int64, error)
+	// Load loads an artifact from the storage.
+	// The artifact is a file indentified by the appName, userID, sessionID and fileName.
+	Load(ctx context.Context, appName, userID, sessionID, fileName string, opts *ArtifactLoadOption) (*genai.Part, error)
+	// Delete deletes an artifact. Deleting a non-existing entry is not an error.
+	Delete(ctx context.Context, appName, userID, sessionID, fileName string, opts *ArtifactDeleteOption) error
+	// List lists all the artifact filenames within a session.
+	List(ctx context.Context, appName, userID, sessionID string) ([]string, error)
```
dpasiukevich: WDYT if we have dedicated requests/responses like in https://github.com/google/adk-go/pull/42?

dpasiukevich: This way we can extend the API in future without worrying for backward compatibility:

1. Adding a field to a e.g. `ListArtifactRequest` is backwards compatible
2. Extending `List(ctx context.Context, appName, userID, sessionID string)` with a new field is not.

hyangah:
I assume you meant to include FileName in the artifact key.

`map[(appName, userID, sessionID, filename)]artifact` is problematic for `List` (i.e. "list all file names of (appName, userID, sessionID)"). It will force scanning all maps. And, the num of versions will be small, so I think omap will be an overkill if we take nested map approach.

Something like 

`map[(appname, userid, sessionid)]map[filename]map[version]artifact` may do the job, but the nested map isn't great.

Or 

`map[(appName, userID, sessionID)]omap.Map[(fileName, version), artifact]` maybe another approach. 

But this type of optimization can be done without api change, and it is "In-memory" service that doesn't persist data. So I'd wait until we understand the access pattern in the real application.

dpasiukevich:
Right I thought about something like `map[(appName, userID, sessionID)]omap.Map[(fileName, version), artifact]`. But I definitely agree, the current implementation is already what's needed.

@hyangah one more nit -- about the req/response naming: `SaveArtifactRequest` vs `ArtifactSaveRequest`. For me somehow first feels more natural: verb + noun it operates on.
I don't want to bike-shed too much, so if you'd prefer `ArtifactSaveRequest`, we can leave it. And if you think it's may be better to have `SaveArtifactRequest` I can send the PR.

hyangah:
> @hyangah one more nit -- about the req/response naming: `SaveArtifactRequest` vs `ArtifactSaveRequest`. For me somehow first feels more natural: verb + noun it operates on. I don't want to bike-shed too much, so if you'd prefer `ArtifactSaveRequest`, we can leave it. And if you think it's may be better to have `SaveArtifactRequest` I can send the PR.

Good point. What about `Versions` request/response?  And placing "Artifact" prefix makes godoc cluster these types.
Screenshot 2025-08-01 at 11 23 38 AM

PR #41, opened 2025-07-25
Update code for AgentSpec

dpasiukevich:
This removes need to embed/cast between agent types. And a `self` back-reference is not needed as well.

Currently all fields are exported but if we decide it would be easy to hide them behind constructor and accessors.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -16,7 +16,6 @@ package adk
 
 import (
 	"context"
-	"fmt"
 	"iter"
 
 	"github.com/google/uuid"
@@ -32,6 +31,74 @@ type Agent interface {
 	Run(ctx context.Context, invocationCtx *InvocationContext) iter.Seq2[*Event, error]
 
 	// TODO: RunLive.
+}
+
+type AgentSpec struct {
+	// Agent name.
+	// It must be unique within the agent tree.
+	// Agent name cannot be "user" since it is reserved for user input.
+	Name string
+
+	// Description about the agent's capability.
+	// The model uses this to determine whether to delegate control to the agent.
+	// One-line description is enough and preferred.
+	Description string
+
+	// The parent agent of this agent.
+	// Note that an agent can ONLY be added as sub-agent once.
+	// If you want to add one agent twice as sub-agent, consider to create two agent
+	// instances with identical config, but with different name and add them to the
+	// agent tree.
+	Parent Agent
+
+	// The sub-agents of this agent.
+	SubAgents []Agent
+
+	// Configuration for the agent backed by LLM.
+	LLMAgent *LLMAgentSpec
+
+	// TODO:
+	//  BeforeAgentCallback
+	//  AfterAgentCallback
+}
+
+type LLMAgentSpec struct {
+	Model Model
+
+	Instruction       string
+	GlobalInstruction string
+
+	Tools []Tool
+
+	GenerateContentConfig *genai.GenerateContentConfig
+
+	// LLM-based agent transfer configs.
+	DisallowTransferToParent bool
+	DisallowTransferToPeers  bool
+
+	// Whether to include contents in the model request.
+	// When set to 'none', the model request will not include any contents, such as
+	// user messages, tool requests, etc.
+	IncludeContents string
+
+	// The input schema when agent is used as a tool.
+	InputSchema *genai.Schema
+
+	// The output schema when agent replies.
+	//
+	// NOTE: when this is set, agent can only reply and cannot use any tools,
+	// such as function tools, RAGs, agent transfer, etc.
+	OutputSchema *genai.Schema
+
+	// OutputKey
+	// Planner
+	// CodeExecutor
+	// Examples
+
+	// BeforeModelCallback
+	// AfterModelCallback
+	// BeforeToolCallback
+	// AfterToolCallback
 }
 
 // An InvocationContext represents the data of a single invocation of an agent.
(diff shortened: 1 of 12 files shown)


agent/llm_agent.go
```diff
@@ -233,15 +63,15 @@ func (a *LLMAgent) Run(ctx context.Context, parentCtx *adk.InvocationContext) it
 	// TODO: Select model (LlmAgent.canonical_model)
```
hyangah, line 63: (sorry that GH UI doesn't allow to comment on unchanged line. I meant to add comment on either LLMAgent.Run (line 62) or baseFlow.Run (line 107) or baseFlow.runOnestep (line 195))

It's unclear to me yet whether it is sufficient to let ADK (e.g. runner or other Agent.Run during agent transfer)  directly call `Agent.Run`. In Python, BaseAgent's `run_async` is a final method (https://github.com/google/adk-python/blob/main/src/google/adk/agents/base_agent.py#L184) and custom agents are supposed to implement its custom logic by overriding  `_run_async_impl` (https://github.com/google/adk-python/blob/f1889ae440386668bee9283527db6d0631e35aa3/src/google/adk/agents/base_agent.py#L243). In our API, the custom agent implements `Run` which I expect to serve like python's `_run_async_impl`.  Even though I am hopeful that we will be able to let runner or a free function implement what python's `run_async` does, without all the functionalities ready (callbacks, tracing, ...) I was afraid we would end up departing from python/java too far. That's why I left the `self` pointer (in case we end up needing the `run_async`).

As I replied in my PR when you proposed this, can we wait to see the followings?

1) finish before/after agent callbacks support
2) implement a couple of custom agents that are not extending LLMAgent, and agents that are extending LLMAgent .
3) see if we can refactor the main adk package and move most of "Agent" and "AgentSpec" types to "agent" package while breaking the circular dependencies.

agent.go
```diff
@@ -34,6 +33,74 @@ type Agent interface {
...
+	// If you want to add one agent twice as sub-agent, consider to create two agent
+	// instances with identical config, but with different name and add them to the
+	// agent tree.
+	Parent Agent
+
+	// The sub-agents of this agent.
+	SubAgents []Agent
+
+	// Configuration for the agent backed by LLM.
+	LLMAgent *LLMAgentSpec
```
hyangah, line 58: This can be confusing. Many fields copied over to LLMAgentSpec in this PR are very  specific to this particular LLMAgent type. (not all agents backed by LLM). And, does setting this field make the agent LLM? Or will Agent that isn't backed by LLM err if this is set?

For example, in python, LangchainAgent (agent backed by LLM) is extending BaseAgent, not LlmAgent.

If this is to support a custom agent type that wants to extend LLMAgent type, can we let the agent implement certain methods unique to LLMAgent and check with type assertion?

BTW nothing wrong with checking underlying types with type assertion.

agent.go
```diff
@@ -34,6 +33,74 @@ type Agent interface {
...
+
+	// The parent agent of this agent.
+	// Note that an agent can ONLY be added as sub-agent once.
+	// If you want to add one agent twice as sub-agent, consider to create two agent
+	// instances with identical config, but with different name and add them to the
+	// agent tree.
+	Parent Agent
+
+	// The sub-agents of this agent.
+	SubAgents []Agent
```
anFatum, line 55: I wonder if that shouldn't be protected under sync.lock - can we e.g. add/delete agents to the subagents during the executions? if yes we could run into the race condition then

dpasiukevich: Currently all fields are exported, for faster iteration. We will definitely later make the fields unexported and hide behind constructor + accessors.

The main goal for the PR is to switch the codebase to use `AgentSpec`. Before it required a `BaseAgent` and some type casts + `self` field for backreference. With the `AgentSpec` that is not needed anymore and a PR serves as a PoC.

agent/agent_transfer_test.go
```diff
@@ -130,90 +138,94 @@ func TestAgentTransferRequestProcessor(t *testing.T) {
...
 
 	t.Run("SoloAgent", func(t *testing.T) {
-		agent := must(NewLLMAgent("Current", model))
+		agent := newTestAgent("Current", model, nil)
 		check(t, agent, "", nil, []string{"Current"})
 	})
 	t.Run("NotLLMAgent", func(t *testing.T) {
 		check(t, newMockAgent("mockAgent"), "", nil, nil)
 	})
 	t.Run("LLMAgentParent", func(t *testing.T) {
```
anFatum, line 147: nit: we could potentially modify to use test structs here, but we can refactor that later if needed

agent.go
```diff
@@ -34,6 +33,74 @@ type Agent interface {
...
+
+	// Configuration for the agent backed by LLM.
+	LLMAgent *LLMAgentSpec
+
+	// TODO:
+	//  BeforeAgentCallback
+	//  AfterAgentCallback
+}
+
+type LLMAgentSpec struct {
```
hyangah, line 65: Is it possible to move this to agent package? As we extend the functionalities, I am afraid we will end up needing to define all types in the top-level adk package.

agent.go
anFatum: LGTM, let's change that later

agent/contents_processor_test.go
```diff
@@ -479,10 +491,29 @@ func TestConvertForeignEvent(t *testing.T) {
...
+}
+
+func (c *customAgent) Spec() *adk.AgentSpec {
+	return &adk.AgentSpec{
+		Name:        "customAgent",
+		Description: "A custom agent for testing.",
 	}
+}
+
+func (c *customAgent) Run(ctx context.Context, ictx *adk.InvocationContext) iter.Seq2[*adk.Event, error] {
```
hyangah, line 504: Is this being used anywhere? (sorry, it's hard to see in GH UI)

agent/llm_agent.go
dpasiukevich: Definitely, let me follow up on this. I have a local `LoopAgent` implementation. Let me also check and do the before/after agent callbacks.

> see if we can refactor the main adk package and move most of "Agent" and "AgentSpec" types to "agent" package while breaking the circular dependencies.

Regarding this I thought about it. But right now it looks impossible (by definition we have `Agent` interface with `Spec() *AgentSpec`. And AgentSpec needs to import adk package for example for parent.

agent/contents_processor_test.go
dpasiukevich: yes, it replaces the previous 
```
	type customAgent struct {
		adk.Agent
	}
```

as the `contents_processor.go` requires `Spec()` now, so I've added a stub test agent

agent.go
dpasiukevich: As mentioned, Regarding this I thought about it. But right now it looks impossible (by definition we have Agent interface with Spec() *AgentSpec. And AgentSpec needs to import adk package for example for parent.

We can try to see if it's possible to achieve this.

But on the other hand, why is it bad to define public types in the top-level adk package? But all the implementation logic is inside specific packages.

agent.go
dpasiukevich: Hmm, I'm thinking along the lines -- it's up to an Agent to decide.

LLMAgentSpec is the configuration for the Agents that would be backed by the LLM.

For example for the `LoopAgent` it can take `NewLoopAgent(maxIterations int, spec *adk.AgentSpec)`. And it takes only fields it needs. If we'd like, the non-LLM agent constructors can additionally make sure LLMAgentSpec is not set.

As another option we can revert it so that LLMAgentSpec contains the AgentSpec similar more/less to the inheritance. This way NewLLMAgent will just get LLMAgentSpec.
But then again we'd need to juggle across types within the code.

My assumption is that we only have two agent types in ADK: LLM and non-LLM based. For non-LLM we just have the `AgentSpec` without caring about `LLMAgentField`.

agent/llm_agent.go
hyangah: * Custom agent type: I meant a custom agent type developed outside this repo (so no access to unexported functions, or internal types). It looks like this PR removed exported APIs to compute/validate parent/subagent setup (`validateSubagent` in llm_agent.go is still unexported). Or do you want every custom agent type implementer to reimplement it? 

* When creating an LLMAgent, it's unclear what field needs to be set or not. For example, if I call NewLLMAgent with Parent set, it is an error currently. I
 
* Can we make NewLLMAgent take two specs (one AgentSpec, one LLMAgentSpec), and define LLMAgentSpec inside the agent package?

dpasiukevich: > Custom agent type: I meant a custom agent type developed outside this repo (so no access to unexported functions, or internal types). It looks like this PR removed exported APIs to compute/validate parent/subagent setup (validateSubagent in llm_agent.go is still unexported). Or do you want every custom agent type implementer to reimplement it?

It's a good idea. WDYT if we expose some `ValidateSpec`? We'd call it in all adk provided agents. And it will also allow custom implementation to call to ensure a valid spec. 
Just as idea -- we can even call it ourselves for example in the runner before invoking the agent.

> When creating an LLMAgent, it's unclear what field needs to be set or not. For example, if I call NewLLMAgent with Parent set, it is an error currently. I

For me it looks like the users can just set whatever fields/feature they'd like to have. And the NewLLMAgent can validate if there's anything invalid from its point of view.

> Can we make NewLLMAgent take two specs (one AgentSpec, one LLMAgentSpec), and define LLMAgentSpec inside the agent package?

But the question then, how to access the .LLMAgentSpec in the adk code? e.g. we have `Spec()` part of the interface, should add another method to a `Agent` interface?

PR #42, opened 2025-07-28
Add the ArtifactService interface

dpasiukevich:


artifact/artifact.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"context"
+
+	"google.golang.org/genai"
+)
+
+type ArtifactService interface {
+	// Stores the artifact data and returns its assigned version number.
+	SaveArtifact(context.Context, *SaveArtifactRequest) (*SaveArtifactResponse, error)
```
hyangah, line 25: Save, Load, List, Delete, ListVersions since this is about Artifact already. 

Why not #39?
Isn't the use of struct type making it unclear about the required fields?
(In fact I was about to propose a change to the session service interface)

dpasiukevich: > Why not #39?

Oopsy, forgot about it. I will close mine.

PR #44, opened 2025-07-29
Add findAgentToRun to the runner implementation.

dpasiukevich:


runner/runner.go
```diff
@@ -119,3 +124,58 @@ func (r *Runner) appendMessageToSession(ctx context.Context, ictx *adk.Invocatio
 	}
 	return nil
 }
+
+// returns the agent that should handle the next request based on session history.
```
hyangah: nit:
// findAgentToRun returns the agent that should handle ...

Let's follow the doc style even though this is an unexported func.

PR #45, opened 2025-07-29
Add before/after agent callbacks

dpasiukevich:
Commit on top of https://github.com/google/adk-go/pull/44

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -98,7 +98,7 @@ type InvocationContext struct {
 
 // NewInvocationContext creates a new invocation context for the given agent
 // and returns context.Context that is bound to the invocation context.
-func NewInvocationContext(ctx context.Context, agent Agent, sessionService SessionService, session *Session) (context.Context, *InvocationContext) {
+func NewInvocationContext(ctx context.Context, agent Agent, sessionService SessionService, session *Session, runCfg *AgentRunConfig) (context.Context, *InvocationContext) {
 	ctx, cancel := context.WithCancelCause(ctx)
 	return ctx, &InvocationContext{
 		InvocationID:   "e-" + uuid.NewString(),
@@ -106,6 +106,7 @@ func NewInvocationContext(ctx context.Context, agent Agent, sessionService Sessi
 		cancel:         cancel,
 		SessionService: sessionService,
 		Session:        session,
+		RunConfig:      runCfg,
 	}
 }
 
@@ -174,9 +175,8 @@ type AgentSpec struct {
 	Description string
 	SubAgents   []Agent
 
-	// TODO:
-	//  BeforeAgentCallback
-	//  AfterAgentCallback
+	BeforeAgentCallbacks []BeforeAgentCallback
+	AfterAgentCallbacks  []AfterAgentCallback
 
 	// the followings are set by Init.
 	parentAgent Agent
--- a/agent/llm_agent.go
+++ b/agent/llm_agent.go
@@ -59,6 +59,20 @@ func WithSubAgents(agents ...adk.Agent) AgentOption {
 	})
 }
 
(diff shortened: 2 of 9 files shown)


runner/runner.go
```diff
@@ -67,29 +73,94 @@ func (r *Runner) RunAsync(ctx context.Context, userID, sessionID string, msg *ge
...
-		// TODO: run _find_agent_to_run(session, root_agent). For now just run root.
+		for event, err := range RunAgent(ctx, ictx, agentToRun) {
+			if !yield(event, err) {
+				return
+			}
+		}
+	}
+}
+
+// RunAgent is called by adk internally wrapping extra logic on top of agent's Run.
```
hyangah: does it make sense to move this in a package under internal/ ?

callback.go
```diff
@@ -0,0 +1,35 @@
...
+
+package adk
+
+import (
+	"context"
+
+	"google.golang.org/genai"
+)
+
+// AgentCallback
```
hyangah: // BeforeAgentCallback is a callback function called before agent's run.

callback.go
```diff
@@ -0,0 +1,35 @@
...
+type BeforeAgentCallback func(ctx context.Context, callbackCtx *CallbackContext) *genai.Content
```
hyangah, line 26: In python, how does it handle if callback encounters an error. Should we consider returning (*genai.Content, error) or require callbacks should not return but panic?

agent.go
```diff
@@ -174,9 +175,8 @@ type AgentSpec struct {
 	Description string
 	SubAgents   []Agent
 
-	// TODO:
-	//  BeforeAgentCallback
-	//  AfterAgentCallback
+	BeforeAgentCallbacks []BeforeAgentCallback
```
hyangah, line 178: in llm_agent.go, add WithBeforeAgentCallbacks and WithAfterAgentCallbacks

callback.go
dpasiukevich: Done. We'd also need later to go through all public api entities (funcs/methods/structs/fields) to ensure all are documented properly.

agent.go
dpasiukevich: Done.

callback.go
dpasiukevich: Yep, I've originally wanted to implement with `(*genai.Content, error)` and if callback fails, return its error from agent's Run.

But then I've checked python & java:
1. https://github.com/google/adk-python/blob/0c6086cb158612ef05b210f77aa7914a79ba0939/src/google/adk/agents/base_agent.py#L378
2. a. https://github.com/google/adk-java/blob/a3746ed46cf8a616c26f440678d854d0500135fb/core/src/main/java/com/google/adk/agents/BaseAgent.java#L276
2. b. https://github.com/google/adk-java/blob/a3746ed46cf8a616c26f440678d854d0500135fb/core/src/main/java/com/google/adk/agents/BaseAgent.java#L201

My interpretation, that both do not care about the callback errors and just check if the returned content is non-empty.

runner/runner.go
dpasiukevich: I agree, this definitely should not be a part of public api. (I originally wanted to later see where better it fits), but it's a good idea to move this piece already. For now I've put it to `internal/runner`

agent/llm_agent.go
```diff
@@ -343,7 +357,7 @@ func (f *baseFlow) runOneStep(ctx context.Context, parentCtx *adk.InvocationCont
 			}
 			if ev == nil {
 				// nothing to yield/process.
-				return
+				continue
```
dpasiukevich, line 360: @hyangah jfyi -- I've updated here. This is needed for streaming api to work correctly, otherwise with adk.StreamingModeSSE on the 2nd chunk we get `Error from agent: TODO: last event is not final`, because due to this return it doesn't wait for all chunks.

PR #46, opened 2025-07-30
tool: add Google Search Tool

ngeorgy:
Java example: https://github.com/google/adk-java/blob/main/core/src/main/java/com/google/adk/tools/GoogleSearchTool.java
Python example: 
Python example: https://github.com/google/adk-python/blob/main/src/google/adk/tools/google_search_tool.py

As you can see in tests we only check the modified requests, this is not an e2e test, once I figure out how to run whole flow, I'll add integration tests as well.

### DIFF
--- a/tool/google_search_tool.go
+++ b/tool/google_search_tool.go
@@ -1,0 +1,87 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tool
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"google.golang.org/adk"
+	"google.golang.org/genai"
+)
+
+// GoogleSearchTool is a tool that adds google search configuration to the LLM request.
+type GoogleSearchTool struct {
+	name        string
+	description string
+}
+
+// Assert that GoogleSearchTool implements adk.Tool
+var _ adk.Tool = (*GoogleSearchTool)(nil)
+
+// NewGoogleSearchTool creates a new GoogleSearchTool.
+func NewGoogleSearchTool() *GoogleSearchTool {
+	return &GoogleSearchTool{
+		name:        "google_search",
+		description: "google_search",
+	}
+}
+
+// Name implements adk.Tool.
+func (t *GoogleSearchTool) Name() string {
+	return t.name
+}
+
+// Description implements adk.Tool.
+func (t *GoogleSearchTool) Description() string {
+	return t.description
+}
+
+// ProcessRequest modifies the LLM request to include the google search tool configuration.
+func (t *GoogleSearchTool) ProcessRequest(ctx context.Context, tc *adk.ToolContext, req *adk.LLMRequest) error {
+	if req == nil {
+		return fmt.Errorf("llm request is nil")
+	}
+
+	if req.GenerateConfig == nil {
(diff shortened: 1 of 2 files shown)


ngeorgy:
Okay, I was able to verify that it works
Screenshot 2025-07-30 at 17 34 24

tool/google_search_tool.go
```diff
@@ -0,0 +1,90 @@
...
+func (t *GoogleSearchTool) ProcessRequest(ctx context.Context, tc *adk.ToolContext, req *adk.LLMRequest) error {
+	if req == nil {
+		return fmt.Errorf("llm request is nil")
+	}
+
+	if req.GenerateConfig == nil {
+		req.GenerateConfig = &genai.GenerateContentConfig{}
+	}
+
+	if req.GenerateConfig.Tools == nil {
```
hyangah: `append` and `len` work with nil slice, so this isn't necessary.

tool/google_search_tool_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"github.com/google/adk-go"
+	"github.com/google/adk-go/model"
+	"github.com/google/adk-go/tool"
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/genai"
+)
+
+func TestGoogleSearchTool_ProcessRequest(t *testing.T) {
+	ctx := context.Background()
```
hyangah: ctx := t.Context()

tool/google_search_tool_test.go
```diff
@@ -0,0 +1,155 @@
...
+	gsTool := tool.NewGoogleSearchTool()
+
+	_, err := gsTool.Run(ctx, &adk.ToolContext{}, nil)
+	if err == nil {
+		t.Fatalf("Run expected error, but got nil")
+	}
+
+	wantError := "google search tool runs internally on the model, it can not be run directly"
+	if !cmp.Equal(err.Error(), wantError) {
```
hyangah: go/go-style/decisions#test-error-semantics

I personally think just checking Run returns a non-nil error is sufficient in this case.

tool/google_search_tool_test.go
```diff
@@ -0,0 +1,155 @@
...
+					}
+					return
+				}
+			}
+			if err != nil {
+				t.Fatalf("ProcessRequest failed: %v", err)
+			}
+
+			if req.GenerateConfig == nil {
+				t.Fatalf("GenerateConfig should not be nil")
```
hyangah: `t.Fatal`

tool/google_search_tool.go
ngeorgy: Indeed, thanks!

tool/google_search_tool_test.go
ngeorgy: Makes sense, done

tool/google_search_tool_test.go
ngeorgy: Done

tool/google_search_tool_test.go
ngeorgy: Done

ngeorgy:
@hyangah Thanks for the review!

ngeorgy:
@hyangah  Hm, I've just noticed that when I do force push on the commit it completely removes previous version. 

When addressing comments, what approach should we take, do we make a new commit and squash at the end before merging? To imitate Critique behavior...

tool/google_search_tool.go
```diff
@@ -0,0 +1,86 @@
...
+	"strings"
+
+	"github.com/google/adk-go"
+	"google.golang.org/genai"
+)
+
+// GoogleSearchTool is a tool that adds google search configuration to the LLM request.
+type GoogleSearchTool struct {
+	// Embed the adk.Tool struct to satisfy the Tool interface
+	adk.Tool
```
hyangah: (optional) I prefer to have the following in the code, instead of embedding

```
var _ adk.Tool = (*GoogleSearchTool)(nil)
```

because that will allow to detect issues statically when `adk.Tool` interface changes.
(And gopls also provides the auto-completion option as a code action). 
Embedding the interface hides issues until the newly added methods are called in the runtime.

ngeorgy: Uh, that's nice, thanks, I've changed it!

PR #49, opened 2025-07-31
Add agent events to session service in runner

dpasiukevich:


agent/llm_agent_test.go
```diff
@@ -531,100 +534,26 @@ func (r *testAgentRunner) session(t *testing.T, sessionID string) (*adk.Session,
...
+		t.Fatalf("failed to get/create session: %v", err)
 	}
-}
 
-func findAgent(agent adk.Agent, name string) adk.Agent {
-	if agent.Spec().Name == name {
-		return agent
-	}
-	return findSubAgent(agent, name)
+	return r.runner.Run(ctx, userID, session.ID, genai.NewContentFromText(newMessage, genai.RoleUser), &adk.AgentRunConfig{})
```
hyangah: (not in the scope of this PR) this makes me wonder if userID/sessionID/Contents, etc should be a option struct. and wonder why they don't belong to adk.AgentRunConfig.

dpasiukevich: Right now we have IMO borderline size of distinct parameters in the method to consider an option struct or similar.

But python and java have the same prototype. And it looks fine actually: userID + sessionID + content fully describe who sends what. And any new feature parameters will either be added to Runner or AgentRunConfig.

PR #50, opened 2025-07-31
Add examples dir

dpasiukevich:
It hosts basic simple examples with the intention for new users to easily try out ADK Go.

### DIFF
--- a/examples/quickstart/main.go
+++ b/examples/quickstart/main.go
@@ -1,0 +1,47 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"os"
+
+	"github.com/google/adk-go/agent"
+	"github.com/google/adk-go/examples"
+	"github.com/google/adk-go/model"
+	"google.golang.org/genai"
+)
+
+func main() {
+	ctx := context.Background()
+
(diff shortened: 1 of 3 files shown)


examples/userloop.go
```diff
@@ -0,0 +1,74 @@
...
+	runner := runner.Runner{
+		AppName:        appName,
+		RootAgent:      rootAgent,
+		SessionService: sessionService,
+	}
+
+	reader := bufio.NewReader(os.Stdin)
+
+	for {
+		fmt.Print("\nUser -> ")
```
ngeorgy, line 52: In the java and python examples, this is usually referred as 'You' instead of 'User', we can pick 'You' for consistency, although 'User' is fine too

examples/quickstart/main.go
```diff
@@ -0,0 +1,46 @@
...
+		APIKey: os.Getenv("GEMINI_API_KEY"),
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	llmAgent, err := agent.NewLLMAgent("weather_time_agent",
+		m,
+		agent.WithDescription("Agent to answer questions about the time and weather in a city."),
+		agent.WithInstruction("I can answer your questions about the time and weather in a city."))
```
ngeorgy, line 40: I've seen both first-person and second-person perspective. But from the most of the examples, usually it's second-person, so should we say "You can answer..."? Don't have a strong preference really so up to you

Also, since this is LLMAgent without tools, I think it cannot really answer questions about time and the weather, maybe we should add a TODO to add google search tool (WithTools) once the other pr is merged?

dpasiukevich: > I've seen both first-person and second-person perspective. But from the most of the examples, usually it's second-person, so should we say "You can answer..."? Don't have a strong preference really so up to you

I don't have a preference here either. It's a quickstart example so we def don't need to optimize the instructions IMO.
I've copied from https://github.com/google/adk-python/blob/314d6a4f95c6d37c7da3afbc7253570564623322/contributing/samples/quickstart/agent.py#L79. Let's keep as is.

> Also, since this is LLMAgent without tools, I think it cannot really answer questions about time and the weather, maybe we should add a TODO to add google search tool (WithTools) once the other pr is merged?

Yep, added the TODO.

examples/userloop.go
dpasiukevich: In https://google.github.io/adk-docs/get-started/quickstart/#run-your-agent it's `user`.

image

Let's keep the `user`. If anything, this can be updated later.

PR #51, opened 2025-07-31
Add loop and sequential agents.

dpasiukevich:
In addition also update runner to set event.Author if it's not set (e.g. by a custom agent).
Logic to determine author is same as in adk-python: if event.contents.author == "user" -> "user" else agent_name.

### DIFF
--- a/agent/loop_agent.go
+++ b/agent/loop_agent.go
@@ -19,7 +19,7 @@ import (
 	"fmt"
 	"iter"
 
-	"github.com/google/adk-go"
+	"google.golang.org/adk/types"
 )
 
 // LoopAgent repeatedly runs a sequence of agents for a specified number of
@@ -28,7 +28,7 @@ import (
 // Use the LoopAgent when your workflow involves repetition or iterative
 // refinement, such as like revising code.
 type LoopAgent struct {
-	agentSpec *adk.AgentSpec
+	agentSpec *types.AgentSpec
 
 	maxIterations uint
 }
@@ -36,7 +36,7 @@ type LoopAgent struct {
 // NewLoopAgent creates a LoopAgent.
 // If maxIterations == 0, it runs indefinitely or until any subagent escalates.
 func NewLoopAgent(name string, maxIterations uint, opts ...AgentOption) (*LoopAgent, error) {
-	agentSpec := &adk.AgentSpec{Name: name}
+	agentSpec := &types.AgentSpec{Name: name}
 
 	a := &LoopAgent{
 		maxIterations: maxIterations,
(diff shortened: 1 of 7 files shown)


agent/loop_agent.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	return a, nil
+}
+
+func (a *LoopAgent) Spec() *adk.AgentSpec {
+	return a.agentSpec
+}
+
+func (a *LoopAgent) Run(ctx context.Context, ictx *adk.InvocationContext) iter.Seq2[*adk.Event, error] {
```
mazas-google: Should we check for context cancelation somewhere, or is it responsibility of a subagent to check for it?

agent/loop_agent.go
```diff
@@ -0,0 +1,93 @@
...
+						return
+					}
+
+					if event.Actions != nil && event.Actions.Escalate {
+						return
+					}
+				}
+			}
+
+			if count > 0 {
```
mazas-google, line 83: nit: maybe check count before iterating through subagents?

agent/loop_agent_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+				if tt.args.maxIterations == 0 && len(gotEvents) == len(tt.wantEvents) {
+					break
+				}
+
+				gotEvents = append(gotEvents, event)
+			}
+
+			if len(tt.wantEvents) != len(gotEvents) {
+				t.Fatalf("Unexpected event length, want: %v, got: %v", len(tt.wantEvents), len(gotEvents))
```
mazas-google: Let's try to keep Google - internal style standards, i.e. "Test outputs should include the actual value that the function returned before printing the value that was expected. A standard format for printing test outputs is YourFunc(%v) = %v, want %v. Where you would write "actual" and "expected", prefer using the words "got" and "want", respectively."

agent/sequential_agent.go
mazas-google, line 1: nit: prefer small PRs, this can easily be done as a separate one.

agent/sequential_agent.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// NewSequentialAgent creates a new SequentialAgent.
+func NewSequentialAgent(name string, opts ...AgentOption) (*SequentialAgent, error) {
+	a, err := NewLoopAgent(name, 0, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	a.maxIterations = 1
```
mazas-google: Why not setting maxIterations in line 32?

agent/loop_agent.go
dpasiukevich: Workflow agents are just blunt things. All execution actions are on the sub-agents.

agent/loop_agent.go
dpasiukevich: This has no effect or am I missing anything?

agent/sequential_agent.go
dpasiukevich: Good catch! Forgot to update this after some code updates. Updated.

agent/loop_agent.go
mazas-google: How many elements will be returned with maxIterations == 0?

agent/loop_agent_test.go
mazas-google: You need to change the order of printing values to comply to the Go style guide. Got before want, go/go-style/decisions#got-before-want.

agent/loop_agent.go
dpasiukevich: maxIterations == 0 means infinite loop for adk LoopAgents until either user stops or agent escalates

agent/loop_agent_test.go
dpasiukevich: Seems like you're still watching first commit only. This was addressed in the subsequent [review comments](https://github.com/google/adk-go/pull/51/commits/cb2dfe73ac976ffc15f72da6ef0fd295cbf648fe) commit. Or am I missing anything?

PR #56, opened 2025-08-01
all: move core types to the types package

hyangah:
The root of the module will host the documentation for users and contributors.

cc @rakyll @rsc


rsc:
I'd like to talk soon about a restructuring of the module into different package boundaries. "Package types" is usually a sign of a problem. I hope to have something to talk about on Monday.

PR #57, opened 2025-08-05
chore: extend checks on CI

yarolegovich:
### Details

* Pinned action dependencies to commit hashes
* Parallel lint and test jobs in PR check workflow
* Enforce tidy `go.mod`
* Enforce `goimports` (and transitively `gofmt`)
* Use `golangci-lint` with default linters and `goheader` for license header enforcement
* Add a nightly workflow that runs shuffled test with race detector and cache ignore
* Run `govulncheck` in the nightly workflow
* Make code changes for `golangci-lint` to pass
* Add concurrency control block for cancelling CI runs if multiple commits modifying a PR are pushed in rapid succession

With [golangcli-lint](https://golangci-lint.run/usage/linters/#enabled-by-default) some checks from the [statickcheck](https://golangci-lint.run/usage/linters/#staticcheck) linter seem somewhat opinionated.

The suggestion is to start using it with the default configuration on and disable checks that don't seem right as we go.

One check I did this for is "Apply De Morgan's law" which encourages `!p1 || !p2` over `!(p1 && p2)` which I think is very condition specific and not always objectively more readable.


.github/workflows/go.yml
```diff
@@ -1,28 +1,50 @@
...
-      with:
-        go-version: '1.24'
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+    
+    - name: Setup
+      uses: ./.github/actions/setup
 
+    - name: Tidy
+      run: |
+        go mod tidy
```
hyangah: https://pkg.go.dev/cmd/go#hdr-Add_missing_and_remove_unused_modules

Consider go mod tidy -diff

.github/workflows/go.yml
```diff
@@ -1,28 +1,50 @@
...
+        git diff --exit-code go.mod
+        git diff --exit-code go.sum
+      
     - name: Build
-      run: go build -v ./...
+      run:  go build -mod=readonly -v -trimpath ./...
```
hyangah: extra space

(do we need -trimpath?)

.golangci.yml
```diff
@@ -0,0 +1,53 @@
...
+    - goheader
+  
+  settings:
+    
+    goheader:
+      values:
+        const:
+          COMPANY: Google LLC
+      template: |-
+        Copyright {{ YEAR }} {{ COMPANY }}
```
hyangah, line 20: Does the goheader require to change YEAR over time? Or is it a place holder excluded from comparison?

https://g3doc.corp.google.com/company/teams/opensource/copyright.md?cl=head#the-year

"When making changes to code with an existing copyright notice:

 - Retain earliest copyright year.
...
"

yarolegovich: I thought this was by default, but checked and [this rule](https://github.com/google/adk-go/pull/57/commits/ef92f298a84f8b607df245487bc545c7f203cf6c) was also needed

.github/workflows/go.yml
yarolegovich: > (do we need -trimpath?)

some CI platforms (eg. gitlab) use random mount points for different runner invocations. this would potentially make cached build artifacts not usable. 
checked the docs and previous CI run log, github actions is providing a stable `home/runner/work/adk-go/adk-go` path. so removed the flag

PR #59, opened 2025-08-05
Basic LLMAgent implementation based on the restruct proposal.

dpasiukevich:
I've also added a temporary `examples/restruct/main.go` to run the agent locally.

I've added 2 diffs from the doc. Let me know if that should be updated. Once we submit the PR, I will proceed and wire up the existing code to the new entities.

The diffs:
1. I've added the internal method to `Agent` interface.
```
setParent(parent Agent)
```

As we'd need to set a parent field for the provided subAgents when they are added.

Another approach is create a wrapper Agent each time (as @rakyll have:
```
type hierarchicalAgent struct {
	agent  agent.Agent
	subs   []agent.Agent
	parent agent.Agent
}
func (a *hierarchicalAgent) Name() string {
	return a.agent.Name()
}
func (a *hierarchicalAgent) Description() string {
	return a.agent.Description()
}
func (a *hierarchicalAgent) Run(ctx agent.Context) iter.Seq2[session.Event, error] {
	return a.agent.Run(ctx)
}
func (a *hierarchicalAgent) Subs() []agent.Agent {
	return a.subs
}
func (a *hierarchicalAgent) Parent() agent.Agent {
	return a.parent
}
// NewAgent provides an agent with the provided parent and sub agents.
// It doesn't change the behavior of the underlying agent otherwise.
func NewAgent(agent agent.Agent, parent agent.Agent, subs ...agent.Agent) agent.Agent {
	return &hierarchicalAgent{
		agent:  agent,
		subs:   subs,
		parent: parent,
	}
}
```

I suspect this will not enforce the agent **tree.** Because semantically the agent may have multiple parents then. 
e.g. `a1 with subAgent(a) and a2 with subAgent(a)`. Ideally we need to error when a2 wants to add subAgent `a`. But with `hierarchicalAgent` it won't error and both `a1` and `a2` will have a subAgent `a`.
Moreover if you run just `a` it won't have a parent.

If I'm wrong here, I can briefly add `hierarchicalAgent` and remove `setParent(parent Agent)` from the interface.

2. Renamed `subs` to `subAgents`.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -26,11 +26,13 @@ type Agent interface {
 	Name() string
 	Description() string
 	// TODO: verify if the interface would have "Run(Context) error" and agent will call agent.Context.Report(Event)
-	Run(Context) iter.Seq2[session.Event, error]
+	Run(Context) iter.Seq2[*session.Event, error]
 	Parent() Agent
 	SubAgents() []Agent
 	// TODO: verify if we should add unexported methods to ensure only this package can implement this interface.
 	// TODO: maybe opact struct?
+
+	setParent(parent Agent)
 }
 
 type Builder struct {
@@ -40,12 +42,12 @@ type Builder struct {
 
 	BeforeAgent []Callback
 	// TODO: verify if the interface would have "Run(Context) error" and agent will call agent.Context.Report(Event)
-	Run        func(Context) iter.Seq2[session.Event, error]
+	Run        func(Context) iter.Seq2[*session.Event, error]
 	AfterAgent []Callback
 }
 
 func (b Builder) Agent() Agent {
-	return &agent{
+	a := &agent{
 		name:        b.Name,
 		description: b.Description,
 		subAgents:   b.SubAgents,
@@ -53,6 +55,12 @@ func (b Builder) Agent() Agent {
 		run:         b.Run,
 		afterAgent:  b.AfterAgent,
 	}
+
+	for _, sub := range b.SubAgents {
+		sub.setParent(a)
+	}
+
+	return a
 }
 
 type Context interface {
@@ -66,7 +74,7 @@ type Context interface {
 	Session() session.Session
 	Artifacts() Artifacts
 
-	Report(session.Event)
+	Report(*session.Event)
 
 	End()
 	Ended() bool
@@ -87,7 +95,7 @@ type agent struct {
 	parent Agent
 
 	beforeAgent []Callback
-	run         func(Context) iter.Seq2[session.Event, error]
+	run         func(Context) iter.Seq2[*session.Event, error]
 	afterAgent  []Callback
 }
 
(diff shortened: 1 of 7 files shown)


rsc:
We should probably sync with jbd; she had a slightly different approach that doesn't have the agent.Builder at all and won't need unexported methods in Agent.

agent/agent.go
```diff
@@ -0,0 +1,126 @@
...
+	Name() string
+	Description() string
+	// TODO: verify if the interface would have "Run(Context) error" and agent will call agent.Context.Report(Event)
+	Run(Context) iter.Seq2[*session.Event, error]
+	Parent() Agent
+	SubAgents() []Agent
+	// TODO: verify if we should add unexported methods to ensure only this package can implement this interface.
+	// TODO: maybe opact struct?
+
+	setParent(parent Agent)
```
rakyll, line 35: We should be able get rid of setParent, see how we wrapped the original agent instances at the builder. https://user.git.corp.google.com/jbd/adk-go/+/refs/heads/version3/agent/llmagent/llmagent.go#39 Not a big deal as an unexported method.

agent/agent.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+type Builder struct {
```
rakyll, line 38: We may not need this generic builder, see the doc for discussion.

agent/agent.go
dpasiukevich: I suspect, this implementation won't be able to enforce the agent **tree**.

e.g. a1 with subAgent(a) and a2 with subAgent(a). Ideally we need to error when a2 wants to add subAgent `a`. But with hierarchicalAgent it won't error and both a1 and a2 will have a subAgent a.

Moreover if you run just `a` (e.g. feed it to a Runner) it won't have a parent.

agent/agent.go
dpasiukevich: @rakyll could you point me to the comment section? I cannot find the relevant discussion.

rakyll:
We chatted offline about wrapping agent to modify their hierarchy vs `setParent`. 

The pros of wrapping is the fact that it allows reusing the same instance:

``` go
greeter := NewGreeterAgent()
agent1 := llmagent.Builder{
   // ...
   Subs: []agent.Agent{greeter}
}
agent2 := llmagent.Builder{
     // ...
   Subs: []agent.Agent{greeter}
}
```

`greeter`, `agent1` and `agent2` are valid objects.

The con is the inconsistency with other languages, e.g. Python implementation mutates the agent instances and sets the parent field: https://github.com/google/adk-python/blob/main/src/google/adk/agents/base_agent.py#L109. `setParent` isn't awful if agent.Builder returns an error if the parent is already set but doesn't allow the reused of the agent instances. I personally prefer the wrapper but I cannot foresee how things may break in the long term -- agent objects don't hold too much state specific to their hierarchy right now but we need to verify that that soft promise is not broken in the long term. We need to discuss this more broadly in ADK preferably in a mini RFC. 

@dpasiukevich would you be interested in a mini RFC?

rakyll:
To clarify my previous comment, we decided to go with `setParent` for consistency for now but can revisit this decision after the RFC discussion.

dpasiukevich:
I will add one-pager for this tomorrow.

PR #63, opened 2025-08-06
Add sessionservice and inmemory implementation

dpasiukevich:
1. I've made the in-memory implementation thread safe (i think).
If you'd consider we should better provide non-thread safe and remove mutex/copy overhead, I can update the PR.

2. stored session has `updated_at` field. Both python&java set it to `event.time` during `append_event` call. So theoretically updated_at may jump forward and backward.
We may make it non-decreasing as least: `if event.Time.After(session.UpdatedAt`. Or use `time.Now()` to reflect when the stored session was actually updated.
For now I've kept as python/java implemented this.

### DIFF
--- a/sessionservice/inmemory.go
+++ b/sessionservice/inmemory.go
@@ -1,0 +1,259 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sessionservice
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"iter"
+	"sync"
+	"time"
+
+	"github.com/google/uuid"
+	"google.golang.org/adk/session"
+	"rsc.io/omap"
+	"rsc.io/ordered"
+)
+
+// inMemoryService is an in-memory implementation of types.SessionService.
+// It is primarily for testing and demonstration purposes.
+// Thread-safe.
+type inMemoryService struct {
+	mu       sync.RWMutex
+	sessions omap.Map[string, *storedSession] // session.ID) -> storedSession
+}
+
+func newInMemoryService() *inMemoryService {
+	return &inMemoryService{}
+}
+
+func (s *inMemoryService) Create(ctx context.Context, req *CreateRequest) (StoredSession, error) {
+	sessionID := req.SessionID
+	if sessionID == "" {
+		sessionID = uuid.NewString()
+	}
+
+	key := sessionKey{
+		AppName:   req.AppName,
+		UserID:    req.UserID,
+		SessionID: sessionID,
+	}
+
+	encodedKey := key.Encode()
+
+	val := &storedSession{
+		id:        session.ID(key),
+		state:     req.State,
+		updatedAt: time.Now(),
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	s.sessions.Set(encodedKey, val)
+
+	return val, nil
+}
+
+func (s *inMemoryService) Get(ctx context.Context, req *GetRequest) (StoredSession, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	res, ok := s.sessions.Get(sessionKey(req.ID).Encode())
+	if !ok {
+		return nil, fmt.Errorf("session %+v not found", req.ID)
+	}
+
+	// TODO: handle req.NumRecentEvents and req.After
+	return res, nil
+}
+
+// List returns an iterator over current snapshot.
+func (s *inMemoryService) List(ctx context.Context, req *ListRequest) ([]StoredSession, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	lo := sessionKey{AppName: req.AppName, UserID: req.UserID}.Encode()
+	hi := sessionKey{AppName: req.AppName, UserID: req.UserID + "\x00"}.Encode()
+
+	var res []StoredSession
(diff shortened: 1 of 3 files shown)


sessionservice/inmemory.go
```diff
@@ -0,0 +1,259 @@
...
+	"sync"
+	"time"
+
+	"github.com/google/uuid"
+	"google.golang.org/adk/session"
+	"rsc.io/omap"
+	"rsc.io/ordered"
+)
+
+// inMemoryService is an in-memory implementation of types.SessionService.
```
hyangah, line 31: [Service]?

sessionservice/inmemory.go
```diff
@@ -0,0 +1,259 @@
...
+// It is primarily for testing and demonstration purposes.
```
hyangah, line 32: you can remove this comment.

sessionservice/inmemory.go
```diff
@@ -0,0 +1,259 @@
...
+// Thread-safe.
+type inMemoryService struct {
+	mu       sync.RWMutex
+	sessions omap.Map[string, *storedSession] // session.ID) -> storedSession
+}
+
+func newInMemoryService() *inMemoryService {
```
hyangah, line 39: why not just use &inMemoryService{} or new(inMemoryService)?

sessionservice/inmemory.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func (s *inMemoryService) Create(ctx context.Context, req *CreateRequest) (StoredSession, error) {
+	sessionID := req.SessionID
+	if sessionID == "" {
+		sessionID = uuid.NewString()
+	}
+
+	key := sessionKey{
+		AppName:   req.AppName,
```
hyangah, line 50: how about checking if required fields (AppName, UserID) are provided?
same applies to other methods.

sessionservice/inmemory.go
```diff
@@ -0,0 +1,259 @@
...
+	res, ok := s.sessions.Get(sessionKey(req.ID).Encode())
+	if !ok {
+		return nil, fmt.Errorf("session %+v not found", req.ID)
+	}
+
+	// TODO: handle req.NumRecentEvents and req.After
+	return res, nil
+}
+
+// List returns an iterator over current snapshot.
```
hyangah, line 84: comment doesn't match the implementation.

sessionservice/inmemory.go
```diff
@@ -0,0 +1,259 @@
...
+	return res, nil
+}
+
+func (s *inMemoryService) Delete(ctx context.Context, req *DeleteRequest) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	_, ok := s.sessions.Get(sessionKey(req.ID).Encode())
+	if !ok {
+		return fmt.Errorf("session %+v not found", req.ID)
```
hyangah, line 114: (optional) I think as we discussed in ArtifactService, we can consider not to return an error.

sessionservice/inmemory.go
```diff
@@ -0,0 +1,259 @@
...
+	return nil
+}
+
+func (s *inMemoryService) AppendEvent(ctx context.Context, session StoredSession, event *session.Event) error {
+	// TODO: no-op if event is partial.
+	// TODO: process event actions and state delta.
+
+	storedSession, ok := session.(*storedSession)
+	if !ok {
+		return errors.New("unexpected session type")
```
hyangah, line 127: fmt.Errorf("unexected session type %T", session)

I think AppendEvent should just take the session ID, not StoredSession. But I also see that the pointer to the object (storedSession) allows to update the event slice directly so the `InvocationContext.Session` can have up-to-date info without fetching the latest snapshot from the storage every time. (the complexity was what stopped me from implementing the readonly session initially :-P)

sessionservice/inmemory.go
```diff
@@ -0,0 +1,259 @@
...
+	return ordered.Decode([]byte(key), &sk.AppName, &sk.UserID, &sk.SessionID)
+}
+
+type storedSession struct {
+	id        session.ID
+	events    []*session.Event
+	state     map[string]any
+	updatedAt time.Time
+
+	// guards all mutable fields
```
hyangah, line 156: move this above `events`

https://go.dev/talks/2014/readability.slide#21

sessionservice/inmemory.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+type events struct {
+	mu     *sync.RWMutex
+	events []*session.Event
+}
+
+func (e *events) All() iter.Seq[*session.Event] {
+	return func(yield func(*session.Event) bool) {
+		e.mu.RLock()
```
hyangah, line 200: Do we need this lock?

`storedSession.events` is append-only and only grows.

`storedSession.Events` return `&events` whose `events` field (slice) is set when constructed. Basically, the length of `events.events` is fixed.  `storedSession.appendEvent` will append to s.events but the newly added events won't be visible from `events` created before anyway. 

If we want to reflect the up-to-date changes, I think it needs to point to storedSession.events and then with locking.

sessionservice/inmemory_test.go
```diff
@@ -0,0 +1,400 @@
...
+				SessionID: "session1",
+				State: map[string]any{
+					"k": 10,
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.Background()
```
hyangah, line 73: t.Context

sessionservice/inmemory_test.go
```diff
@@ -0,0 +1,400 @@
...
+			state: map[string]any{
+				"k2": "v2",
+			},
+		},
+	} {
+		service.sessions.Set(sessionKey(storedSession.id).Encode(), storedSession)
+	}
+
+	return service
+}
```
hyangah, line 400: TODO: test concurrency

sessionservice/inmemory.go
dpasiukevich: I had exactly the same train of thoughts 🙃

1. when first was, I thought of session_id
2.> pointer to the object (storedSession) allows to update the event slice directly so the InvocationContext.Session can have up-to-date info without fetching the latest snapshot from the storage every time.
And for this reason decided the storedSession is better.

sessionservice/inmemory.go
dpasiukevich: right, let me remove this lock

PR #67, opened 2025-08-12
Update runner to new packages

dpasiukevich:
* [Add agent_context implementation.](https://github.com/google/adk-go/commit/90dbc33659e74b65ae2b003cab78f40a0aa310ea)
* [Update runner from types package.](https://github.com/google/adk-go/commit/324de0f392ab08a475b47017dce6227591bfa6be)
* [Update workflow agents to use new packages.](https://github.com/google/adk-go/commit/90dbcfe33ebea275c9020e03b52ef15ad948878a)

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -37,7 +37,7 @@ type Agent interface {
 	internal() *agent
 }
 
-func New(cfg Config) (*agent, error) {
+func New(cfg Config) (Agent, error) {
 	a := &agent{
 		name:        cfg.Name,
 		description: cfg.Description,
--- a/agent/workflowagents/loopagent/loop_agent_test.go
+++ b/agent/workflowagents/loopagent/loop_agent_test.go
@@ -15,7 +15,6 @@
 package loopagent_test
 
 import (
-	"context"
 	"fmt"
 	"iter"
 	"testing"
@@ -89,7 +88,7 @@ func TestNewLoopAgent(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			ctx := context.Background()
+			ctx := t.Context()
 
 			agent, err := loopagent.New(loopagent.Config{
 				MaxIterations: tt.args.maxIterations,
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -15,7 +15,6 @@
 package agent
 
 import (
-	"context"
 	"iter"
 	"testing"
 
@@ -30,7 +29,7 @@ import (
 func TestAgentCallbacks(t *testing.T) {
 	t.Parallel()
 
-	ctx := context.Background()
+	ctx := t.Context()
 
 	tests := []struct {
 		name                 string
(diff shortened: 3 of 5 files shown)


agent/agent.go
```diff
@@ -36,7 +37,7 @@ type Agent interface {
 	internal() *agent
 }
 
-func New(cfg Config) (Agent, error) {
+func New(cfg Config) (*agent, error) {
```
hyangah: It is strange to return an unexported type from New. Can we use `Agent`?
(we can document that returned Agent with non-nil error is invalid...)

We didn't have time to officially codify this in the style guide yet, but  
exported functions should have exported types for parameters and results
in most cases (there are some cases that justify exception. This isn't the case I think)

https://groups.google.com/a/google.com/g/go-style/c/7GpvJGWpPkQ/m/tpHzQyFJBgAJ

----



Unexported types do not appear in godoc documentation. Using an unexported type as part of an exported function signature produces incomplete documentation.

It is clumsy for callers to work with functions with a signature that cannot be expressed in the calling package. For example, it is difficult to create a wrapper for a function when you cannot reference the types of that function's parameters or results.

agent/agent.go
```diff
@@ -123,11 +125,89 @@ func (a *agent) SubAgents() []Agent {
...
+		}
+
+		event := session.NewEvent(ctx.InvocationID())
+		event.LLMResponse = &llm.Response{
+			Content: content,
+		}
+		event.Author = agent.Name()
+		event.Branch = ctx.Branch()
+		// TODO: how to set it. Should it be a part of Context?
+		// event.Actions = callbackContext.EventActions
```
hyangah, line 184: Looks like EventActions has different lifespan from other fields in agent.Context. I wonder if this suggests revival of CallbackContext or change of the callback function signature to accept EventActions or similar.

agent/workflowagents/loopagent/loop_agent_test.go
```diff
@@ -84,15 +87,38 @@ func TestNewLoopAgent(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			agent, err := agent.NewLoopAgent("test_agent", tt.args.maxIterations, agent.WithSubAgents(tt.args.subAgents...))
+			t.Parallel()
+
+			ctx := context.Background()
```
hyangah: t.Context

agent/agent.go
dpasiukevich: Definitely. Thanks for catching this!

agent/agent.go
dpasiukevich: Yes, this is a good example of identifying the gaps on how we'd define context types and callbacks.

agent/workflowagents/loopagent/loop_agent_test.go
dpasiukevich: Done for the changes in this PR. I've made a note to fix this in the other code places with the next PRs.

Issue #55, opened 2025-08-01
Remove LLMResponseStream


rakyll:
`LLMResponseStream` seems to be redundant. Given, we haven't introduced a type for `iter.Seq2[*Event, error]` either, consider removing `LLMResponseStream`.

PR #70, opened 2025-08-18
Remove old model package in favor of llm.

dpasiukevich:


dpasiukevich:
Fixes https://github.com/google/adk-go/issues/55

llm/gemini/gemini_test.go
```diff
@@ -0,0 +1,146 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			httpRecordFilename := filepath.Join("testdata", strings.ReplaceAll(t.Name(), "/", "_")+".httprr")
+
+			model, err := NewModel(t.Context(), tt.modelName, newGeminiTestClientConfig(t, httpRecordFilename))
+			if err != nil {
+				t.Fatal(err)
+			}
```
anFatum, line 62: nit: It looks like it's the same as below, maybe we can extract this to be in `setup` method? 

e.g. `model := setup(t, modelName)`

dpasiukevich: Right now we only have 2 tests, and moving common to `setup` would save 1 line at the cost of reading/introducing one extra function.

Let's keep as is. Once/if there'd appear more tests and setup becomes more difficult, we'd move it to a separate func. WDYT?

PR #71, opened 2025-08-18
Introduce the internal parentmap.

dpasiukevich:
When an agent is created, adk automatically sets a parent for subagents. This is done in the "agent" package.

For LLMAgent creation this means, it's needed to propagate "self" to this code to ensure each subagent.Parent is set with a proper concrete type of llmagent (and not just a base agent).

To get rid of the "self" field in the public API, we can move the agent tree setup to an internal package, used by runner on creation.

Also, this shows an example of propagating data from runner to the agent.
Such propagation will be done only for 2 cases:
* agent hierarchy (created by runner, needed by LLMAgent for transfer logic).
* agent run config (set by users to control agents).


agent/agent.go
```diff
@@ -29,7 +29,6 @@ type Agent interface {
 	Description() string
 	// TODO: verify if the interface would have "Run(Context) error" and agent will call agent.Context.Report(Event)
 	Run(Context) iter.Seq2[*session.Event, error]
-	Parent() Agent
```
rakyll, line 32: Nite: Wouldn't some custom implementations of runners need to know about the hierarchy?

dpasiukevich: That's a good point.

* adk runners can reuse the package. And with hierarchy encapsulated in its own internal package gives us nice flexibility here.
* for user defined runners we'd also have a great flexibility here to decide based on the needs, simply because the current approach is not a part of public API.

PR #73, opened 2025-08-19
Updated licensing, including httprr

kdroste-google:
Moving the original httprr files to 'rr' folder with a proper LICENCE file

### DIFF
--- a/llm/gemini/gemini_test.go
+++ b/llm/gemini/gemini_test.go
@@ -118,7 +118,7 @@ func TestModel_GenerateStream(t *testing.T) {
 // newGeminiTestClientConfig returns the genai.ClientConfig configured for record and replay.
 func newGeminiTestClientConfig(t *testing.T, rrfile string) *genai.ClientConfig {
 	t.Helper()
-	rr, err := testutil.NewGeminiTransportForTesting(rrfile)
+	rr, err := testutil.NewGeminiTransport(rrfile)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/agent/llmagent/llmagent_test.go
+++ b/agent/llmagent/llmagent_test.go
@@ -658,7 +658,7 @@ func collectTextParts(stream iter.Seq2[*session.Event, error]) ([]string, error)
 
 func newGeminiTestClientConfig(t *testing.T, rrfile string) (http.RoundTripper, bool) {
 	t.Helper()
-	rr, err := testutil.NewGeminiTransportForTesting(rrfile)
+	rr, err := testutil.NewGeminiTransport(rrfile)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


internal/httprr/rr/LICENCE
```diff
@@ -0,0 +1,24 @@
+Copyright (c) 2009 The Go Authors. All rights reserved.
```
dpasiukevich: LICENCE -> LICENSE

internal/httprr/rr/LICENCE
```diff
@@ -0,0 +1,24 @@
...
+Redistribution and use in source and binary forms, with or without
```
dpasiukevich: As idea -- WDYT if we move `internal/httprr/genai.go `to a new package `internal/testutil/genai.go`?

As a result we'd have two separate packages:
1. `internal/httprr` -- keeps imported 3p unmodified httprr code.
2. `internal/testutil` contains our helpers for the test setup. Right now it would be `genai.go`

main
dpasiukevich, line 1: Remove this file. Seems like it's accidentally commited.

kdroste-google: true, removed

internal/httprr/rr/LICENCE
kdroste-google: done

llm/gemini/gemini_test.go
```diff
@@ -117,7 +118,7 @@ func TestModel_GenerateStream(t *testing.T) {
 // newGeminiTestClientConfig returns the genai.ClientConfig configured for record and replay.
 func newGeminiTestClientConfig(t *testing.T, rrfile string) *genai.ClientConfig {
 	t.Helper()
-	rr, err := httprr.NewGeminiTransportForTesting(rrfile)
+	rr, err := testutil.NewGeminiTransportForTesting(rrfile)
```
dpasiukevich: As nit: NewGeminiTransportForTesting -> NewGeminiTransport

WDYT, `testutil.NewGeminiHTTPTransport(...)`

README.md
```diff
@@ -0,0 +1,13 @@
+# Agent Development Kit (ADK) for GO
+
+[![License](https://img.shields.io/badge/License-Apache_2.0-blue.svg)](LICENSE)
+
+
+## 📄 License
+
+
+This project is licensed under the Apache 2.0 License - see the
+[LICENSE.Apache](LICENSE.Apache) file for details.
```
dpasiukevich, line 10: Can we keep just `LICENSE`? I assume `LICENSE.postfix` is used when there are multiple licenses at root to avoid collision.

kdroste-google: I would say no, I expect some problems with collisions between initial LICENSE and merged version.

PR #74, opened 2025-08-20
Remove dead code

rakyll:
As a part of the overall reconsuction effort, the contents of the types package is living in their corresponding package. Removing the dead/duplicate code that is not used for the rest of the packages.


artifact/artifact.go
```diff
@@ -24,20 +24,40 @@ import (
...
-type InMemoryArtifactService struct {
+type InMemoryService struct {
 	mu sync.RWMutex
 	// ordered(appName, userID, sessionID) -> session
 	artifacts omap.Map[string, *genai.Part]
 }
 
+func Mem() *InMemoryService {
+	return &InMemoryService{}
+}
```
dpasiukevich:
```suggestion
func Mem() Service {
	return &inMemoryService{}
}
```

WDYT if we do not add `InMemoryService` to the public API (at least for now)?

artifact/artifact.go
```diff
@@ -24,20 +24,40 @@ import (
...
 
-	"google.golang.org/adk/types"
 	"google.golang.org/genai"
 	"rsc.io/omap"
 	"rsc.io/ordered"
 )
 
-// InMemorySessionService is an in-memory implementation of types.SessionService.
+// Service is the artifact storage service.
+type Service interface {
```
dpasiukevich: Two nits, WDYT:

1. Separate Service interface + req/resp types in one file and the inmemory implementation in another? e.g. (`artifact/service.go` and `artifact/inmemory.go`). Should improve codebase navigation.
2. renaming package name `artifact` -> `artifactservice`.  To be aligned with `sessionservice` package.

types/event.go
dpasiukevich, line 1: Why cannot we remove this file?

artifact/artifact.go
rakyll: Both done.

artifact/artifact.go
rakyll: Done.

types/event.go
rakyll: Right, this is removable. Done.

PR #75, opened 2025-08-20
Migrate agent_run_config.

dpasiukevich:
Added as runner.AgentRunConfig for users to specify runtime configuration.

### DIFF
--- a/examples/userloop.go
+++ b/examples/userloop.go
@@ -40,7 +40,7 @@ func Run(ctx context.Context, rootAgent agent.Agent) {
 		log.Fatalf("Failed to create the session service: %v", err)
 	}
 
-	agentRunner := runner.Runner{
+	r := runner.Runner{
 		AppName:        appName,
 		RootAgent:      rootAgent,
 		SessionService: sessionService,
@@ -59,7 +59,7 @@ func Run(ctx context.Context, rootAgent agent.Agent) {
 		userMsg := genai.NewContentFromText(userInput, genai.RoleUser)
 
 		fmt.Print("\nAgent -> ")
-		for event, err := range agentRunner.Run(ctx, userID, session.ID().SessionID, userMsg, &runner.AgentRunConfig{
+		for event, err := range r.Run(ctx, userID, session.ID().SessionID, userMsg, &runner.RunConfig{
 			StreamingMode: runner.StreamingModeSSE,
 		}) {
 			if err != nil {
--- a/agent/llmagent/llmagent_test.go
+++ b/agent/llmagent/llmagent_test.go
@@ -553,7 +553,7 @@ func (r *testAgentRunner) Run(t *testing.T, sessionID, newMessage string) iter.S
 		content = genai.NewContentFromText(newMessage, genai.RoleUser)
 	}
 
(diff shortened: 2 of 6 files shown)


runner/agent_run_config.go
```diff
@@ -0,0 +1,56 @@
...
+type StreamingMode string
+
+const (
+	StreamingModeNone StreamingMode = "none"
+	StreamingModeSSE  StreamingMode = "sse"
+	StreamingModeBidi StreamingMode = "bidi"
+)
+
+// AgentRunConfig controls runtime behavior of agents.
+type AgentRunConfig struct {
```
rakyll: Should we name it to RunConfig? It accidentally got named to AgentRunConfig in the Go implementation. RunConfig is for runners, it really doesn't have any strong ties to an agent. https://google.github.io/adk-docs/api-reference/python/google-adk.html#google.adk.agents.RunConfig

internal/agent/runconfig/run_config.go
```diff
@@ -0,0 +1,45 @@
...
+
+type StreamingMode string
+
+const (
+	StreamingModeNone StreamingMode = "none"
+	StreamingModeSSE  StreamingMode = "sse"
+	StreamingModeBidi StreamingMode = "bidi"
+)
+
+type RunConfig struct {
```
rakyll, line 27: Why do we need this duplicate?

examples/userloop.go
```diff
@@ -41,7 +40,7 @@ func Run(ctx context.Context, rootAgent agent.Agent) {
 		panic(err)
 	}
 
-	runner := runner.Runner{
+	agentRunner := runner.Runner{
```
rakyll: Maybe call this `r`. The runner is the runtime, it's narrow to represent them as agent runners. Agent is just a component in the runtime like a tool or a model.

internal/agent/runconfig/run_config.go
dpasiukevich: Mostly to avoid dependency cycle and keeping public api clean.

`RunConfig` is defined in the `runner` package as it seems the best place for the struct. And it is the last package in dependency chain, it can import everything to run it.

Still we need to propagate RunConfig from Runner to LLMAgent. So `llmagent` needs to be able to import&read `RunConfig`.

As we're propagating RunConfig via context, **the context key (or funcs To/FromContext) must be public then** so that both runner and llmagent implementation can use them.

With the internal struct:
1. There's no cycle.
2. Separate public entity from internal.

I've duplicated the internal entity (and only the streaming field) mostly for convenience right now. But in future we can reshape it as needed without affecting API.

runner/agent_run_config.go
dpasiukevich: Renamed.

Although, @rakyll WDYT, maybe we should put it to `agent.RunConfig`? Looking at python it's in agents.RunConfig. Plus if our Runner becomes more advanced, its constructor may become `runner.New(runner.Config)` creating confusion.

dpasiukevich: Thinking again, I will keep `runner.RunConfig`, because in ADK RunConfig is global and not per-agent.

PR #77, opened 2025-08-20
Use log.Fatal in examples instead of panic

rakyll:


examples/workflowagents/sequential/main.go
```diff
@@ -65,7 +66,7 @@ func main() {
 		Run:         myAgent{id: 2}.Run,
 	})
 	if err != nil {
-		panic(err)
+		log.Fatalf("Failed to create agent: %v", err)
 	}
 
 	loopAgent, err := sequentialagent.New(sequentialagent.Config{
```
dpasiukevich, line 72: If it's ok, could you update loopAgent -> sequentialAgent?

copy-paste artifact 🙃

dpasiukevich: Ah, nvm, I'd update later on.

PR #83, opened 2025-08-22
Add parallel agent.

dpasiukevich:


agent/workflowagents/parallelagent/agent.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+
+		ctx.End()
+	}
+}
+
+func runSubAgent(ctx agent.Context, agent agent.Agent, results chan<- result) {
+	for event, err := range agent.Run(ctx) {
+		select {
+		case <-ctx.Done():
```
hyangah, line 96: in case of ctx being cancelled or timedout, don't we want to report the error?

agent/workflowagents/parallelagent/agent.go
```diff
@@ -0,0 +1,102 @@
...
+
+	go func() {
+		wg.Wait()
+		close(results)
+	}()
+
+	return func(yield func(*session.Event, error) bool) {
+		for res := range results {
+			if !yield(res.event, res.err) {
+				break
```
hyangah, line 85: I feel like we need to drain the results channel if there are goroutines that are blocked on sending to `results`.

agent/workflowagents/parallelagent/agent.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+type Config struct {
+	// Basic agent setup.
+	AgentConfig agent.Config
+}
+
+// New creates a ParallelAgent.
+//
+// It is a shell agent that run its sub-agents in parallel in isolated manner.
```
hyangah: nit: Parallel agent runs its sub-agents in parallel in isolated manner.

dpasiukevich: Updated to suggestion.

Original text is from [adk-python](https://github.com/google/adk-python/blob/ad81aa54de1f38df580915b7f47834ea8e5f1004/src/google/adk/agents/parallel_agent.py#L162). Both are ok.

agent/workflowagents/parallelagent/agent.go
dpasiukevich: Definitely, updated to handle both agent error and context cancel/timeout. Thanks!

agent/workflowagents/parallelagent/agent.go
dpasiukevich: Yep, good catch! Instead of draining the channel on user stopping iteration (which may take time until all agents finish or context is cancellec), I've updated to immediately send a termination signal to subagent.

Thanks!

agent/workflowagents/parallelagent/agent.go
hyangah: Should we have another select around `results <- result { ...} ` below in case the reader of results (for loop in line 81) is already finished? (Or make results a buffered channel?)

agent/workflowagents/parallelagent/agent_test.go
```diff
@@ -0,0 +1,218 @@
...
+				if diff := cmp.Diff(tt.wantEvents, gotEvents); diff != "" {
+					t.Errorf("events mismatch (-want +got):\n%s", diff)
+				}
+			}
+		})
+	}
+}
+
+// newParallelAgent creates parallel agent with 2 subagents emitting maxIterations events or infinitely if maxIterations==0.
+func newParallelAgent(t *testing.T, maxIterations uint, agentErr error) agent.Agent {
```
hyangah: is it possible to have a test case where one of the agents terminates early with error while others keep running (to test the other agents stops). And I wonder if # agents >> 2, that will increase the chance of race the race detector can catch.

agent/workflowagents/parallelagent/agent.go
dpasiukevich: Made it `defer close(doneChan)` to ensure the done channel is closed and subAgent goroutines are not blocked.

agent/workflowagents/parallelagent/agent_test.go
dpasiukevich: Done.

agent/workflowagents/parallelagent/agent.go
hyangah: Consider this case:

 - ctx is cancelled (timeout, etc).
 - goroutine SubAgent-A running runSubAgent detects ctx.Done() and returns non-nil error.
 - The main goroutine M reading from the `results` detected it, got out of the loop, but not yet handling deferred close.
 - Meanwhile, the goroutine SubAgent-B running runSubAgent reaches to line 97, and starts to send to `results`.
 - The goroutine M finally handles `defer close(doneChan)`. But SubAgent-B goroutine is already past `<-done` case check.

dpasiukevich: Right, I've added an extra read from done to ensure the goroutine won't stuck. This one should cover all cases.

Thanks!

PR #80, opened 2025-08-20
[DRAFT] Update tool interface and add genai wrapper.

dpasiukevich:
types in the tool package

```
type Tool interface {
	Name() string
	Description() string
	ProcessRequest(ctx Context, req *llm.Request) error
}

type FunctionTool interface {
	Tool
	Declaration() *genai.FunctionDeclaration
	Run(ctx Context, args any) (result any, err error)
}
```

PR #87, opened 2025-08-26
Minimize tool interface.

dpasiukevich:
public Tool interface

```
type Tool interface {
	Name() string
	Description() string
}
```

Tool specific methods are moved to the internal package `internal/tool`

```
type FunctionTool interface {
	Run(ctx tool.Context, args any) (result any, err error)
}

type RequestProcessor interface {
	ProcessRequest(ctx tool.Context, req *llm.Request) error
}
```

Problem -- we cannot ensure at compile-time that any new tool implementation satisfies `internal/tool` types due to cycle. (can be verified in a hacky way via init() or in unit-tests).

### DIFF
--- a/tool/google_search_tool_test.go
+++ b/tool/google_search_tool_test.go
@@ -115,7 +115,7 @@ func TestGoogleSearchTool_ProcessRequest(t *testing.T) {
 
 			requestProcessor, ok := gsTool.(toolinternal.RequestProcessor)
 			if !ok {
-				t.Fatalf("googleSearchTool does not implement RequestProcessor")
+				t.Fatal("googleSearchTool does not implement RequestProcessor")
 			}
 
 			if err := requestProcessor.ProcessRequest(nil, req); err != nil {


hyangah:
> Problem -- we cannot ensure at compile-time that any new tool implementation satisfies internal/tool types due to cycle. (can be verified in a hacky way via init() or in unit-tests).

Since all tool object creation is done by ADK, IMHO unit test in ADK (either in the tool package too_test, or in a new unit test in the itype) is good enough. WDYT?

BTW, I propose to rename `itype` package to something like internal/tool or internal/toolinternal to match internal/agent or internal/llminternal.

dpasiukevich:
@hyangah I've closed https://github.com/google/adk-go/pull/80 and updated this PR with tool/tool_test.go to verify types.

internal/llminternal/base_flow.go
```diff
@@ -188,9 +189,13 @@ func (f *Flow) preprocess(ctx agent.Context, req *llm.Request) error {
 	// run processors for tools.
 	// TODO: check need/feasibility of running this concurrently.
 	for _, t := range Reveal(llmAgent).Tools {
+		requestProcessor, ok := t.(toolinternal.RequestProcessor)
```
hyangah, line 192: Consider TODO here to relax the requirement.
If t has the other methods (e.g. Descriptor, etc...) we can also handle the cases.

tool/google_search_tool_test.go
```diff
@@ -112,7 +113,12 @@ func TestGoogleSearchTool_ProcessRequest(t *testing.T) {
 
 			gsTool := tool.NewGoogleSearchTool(m)
 
-			if err := gsTool.ProcessRequest(nil, req); err != nil {
+			requestProcessor, ok := gsTool.(toolinternal.RequestProcessor)
+			if !ok {
+				t.Fatalf("googleSearchTool does not implement RequestProcessor")
```
hyangah: nit: t.Fatal

internal/llminternal/base_flow.go
dpasiukevich: Could you clarify the suggestion? 

Right now we type cast the tool to extract the methods we need. In this particular place (preprocess func) we need only `ProcessRequest()` func.

hyangah: I thought we can allow to implement a function tool by implementing the following hypothetical interface instead of implementing `ProcessRequest`.

```
package toolinternal

type Function interface {
	Run(ctx Context, args map[string]any) (result any, err error)
}
```

Then, 

```
switch i := t.(type) {
case toolinternal.Function:
     // use i.Declaration
case toolinternal.RequestProcessor:
     // use i.ProcessRequest
...
}
``` 

But if we think any function tool should implement `ProcessRequest`, I am fine with it too. (it's implementation detail so we don't need to address it now)

dpasiukevich: SG!

My thinking was -- we type cast to only extract what we need. For example in this particular place (`preprocess` function) we only need to call tool's `ProcessRequest()`. We don't need switch type `toolinternal.Function` here, because we don't need it.

PR #88, opened 2025-08-28
Update after agent callback to receive  resp and error.

dpasiukevich:
This would allow the callback implementation to inspect and take action based on the actual agent response and errors.

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -31,15 +31,15 @@ func TestAgentCallbacks(t *testing.T) {
 	ctx := t.Context()
 
 	tests := []struct {
-		name                 string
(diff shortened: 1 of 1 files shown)


agent/agent_test.go
```diff
@@ -32,14 +32,14 @@ func TestAgentCallbacks(t *testing.T) {
 
 	tests := []struct {
 		name                 string
-		beforeAgentCallbacks []Callback
-		afterAgentCallbacks  []Callback
+		beforeAgentCallbacks []BeforeAgentCallback
```
hyangah: nit: rename to beforeAgent and afterAgent

PR #89, opened 2025-08-29
Add memoryservice with in-memory implementation.

dpasiukevich:
Reflects
https://github.com/google/adk-python/tree/e45c3be23895b5ec68908ad9ee19bd622dcbd003/src/google/adk/memory

It currently operates with session.Session. I'd follow up regarding the use-cases for Session vs StoredSession and update if needed in future PR.

Inmemory implementation is based on the 2-level map: {appName, userID} ->
sessionID -> {event + eventWords}
eventWords is a precomputation to fetch based on the read request words.

### DIFF
--- a/memoryservice/inmemory_test.go
+++ b/memoryservice/inmemory_test.go
@@ -147,8 +147,7 @@ func Test_inMemoryService_SearchMemory(t *testing.T) {
 
 			got, err := s.Search(t.Context(), tt.req)
 			if (err != nil) != tt.wantErr {
-				t.Errorf("inMemoryService.SearchMemory() error = %v, wantErr %v", err, tt.wantErr)
-				return
+				t.Fatalf("inMemoryService.SearchMemory() error = %v, wantErr %v", err, tt.wantErr)
 			}
 			if diff := cmp.Diff(tt.wantResp, got, sortMemories); diff != "" {
 				t.Errorf("inMemoryiService.SearchMemory() mismatch (-want +got):\n%s", diff)
--- a/memoryservice/service.go
+++ b/memoryservice/service.go
@@ -32,6 +32,7 @@ type Service interface {
 	// A session can be added multiple times during its lifetime.
 	AddSession(ctx context.Context, s session.Session) error
 	// Search returns memory entries relevant to the given query.
+	// Empty slice is returned if there are no matches.
 	Search(ctx context.Context, req *SearchRequest) (*SearchResponse, error)
 }
 


memoryservice/inmemory.go
```diff
@@ -0,0 +1,162 @@
...
+	return &inMemoryService{
+		store: make(map[key]map[sessionID][]value),
+	}
+}
+
+type key struct {
+	appName, userID string
+}
+
+type sessionID = string
```
hyangah, line 39: alias vs type discussion

Is there a reason to prefer aliasing?

memoryservice/inmemory.go
```diff
@@ -0,0 +1,162 @@
...
+	v, ok := s.store[k]
+	if !ok {
+		s.store[k] = map[sessionID][]value{
+			curSession.ID().SessionID: values,
+		}
+		return nil
+	}
+
+	v[curSession.ID().SessionID] = values
+	return nil
```
hyangah: nit: Bikeshedding. WDYT to keep the value setting in one place?


v, ok := s.store[k]
if !ok {
  v = map[sessionID][]value{}
  s.store[k] = v
}
v[curSession.ID().SessionID] = values
return nil

memoryservice/inmemory.go
```diff
@@ -0,0 +1,162 @@
...
+
+			maps.Copy(words, extractWords(part.Text))
+		}
+
+		if len(words) == 0 {
+			continue
+		}
+
+		values = append(values, value{
+			event: event,
```
hyangah: optional: Makes me wonder if we want to keep the entire event if the SearchMemory API needs only the LLMResponse.Content.

memoryservice/service.go
```diff
@@ -0,0 +1,58 @@
...
+// Service is a definition of the memory service.
+//
+// The service provides functionalities to ingest sessions into memory so that
+// the memory can be used for user queries.
+type Service interface {
+	// AddSession adds a session to the memory service.
+	//
+	// A session can be added multiple times during its lifetime.
+	AddSession(ctx context.Context, s session.Session) error
+	// SearchMemory searches the memory for entries relevant to the given query.
```
hyangah: Why not just `Search` and SearchRequest/SearchResponse?

memoryservice/inmemory.go
```diff
@@ -0,0 +1,162 @@
...
+	for k := range m1 {
+		if _, ok := m2[k]; ok {
+			return true
+		}
+	}
+
+	return false
+}
+
+var wordRegex = regexp.MustCompile(`[A-Za-z]+`)
```
hyangah: Does the in-memory memory service allows only english alphabet text search?
I wonder if we just split (strings.SplitSeq) based on space and index everything.

memoryservice/service.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// MemoryEntry represents a single memory entry.
+type MemoryEntry struct {
+	// Content contains the main content of the memory.
+	Content *genai.Content
+	// Author of the memory.
+	Author string
+	// Timestamp shows when the original content of this memory happened.
+	// This string will be forwarded to LLM. Preferred format is ISO 8601 format.
```
hyangah, line 58: Can't we just use time.Time here? Is it the tool's responsibility that how time.Time is presented (format, timezone, ..) in LLM request?

memoryservice/inmemory.go
dpasiukevich: I wanted to make store readable `map[sessionID][]value` so it's not a `map[string][]value`

If we make a `type sessionID string`, then we'd need to type convert sessionID <-> string.

Having type alias in this small implementation looked nice to me, but I don't have any preference, and can update to any other alternative.

memoryservice/service.go
dpasiukevich: Yes, updated to time.Time.

memoryservice/inmemory.go
dpasiukevich: Content + Author + Time, where Content will be taking most of memory.

But it's a good idea to store only what we need. Updated the `value` to do that. Thanks!

memoryservice/inmemory.go
dpasiukevich: definitely, updated.

memoryservice/inmemory.go
dpasiukevich: I blindly followed the python implementation. But I think your idea is better. Updated, thanks!

memoryservice/service.go
dpasiukevich: sg

memoryservice/inmemory_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+			for _, session := range tt.initSessions {
+				if err := s.AddSession(t.Context(), session); err != nil {
+					t.Fatalf("inMemoryService.AddSession() error = %v", err)
+				}
+			}
+
+			got, err := s.Search(t.Context(), tt.req)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("inMemoryService.SearchMemory() error = %v, wantErr %v", err, tt.wantErr)
```
hyangah: t.Fatalf instead of t.Errorf+return

memoryservice/service.go
```diff
@@ -0,0 +1,59 @@
...
+// Service is a definition of the memory service.
+//
+// The service provides functionalities to ingest sessions into memory so that
+// the memory can be used for user queries.
+type Service interface {
+	// AddSession adds a session to the memory service.
+	//
+	// A session can be added multiple times during its lifetime.
+	AddSession(ctx context.Context, s session.Session) error
+	// Search returns memory entries relevant to the given query.
```
hyangah, line 34: Should we clarify that no matching result is not an error?

PR #94, opened 2025-09-09
Implement Artifacts for agent context

ngeorgy:
In the following PRs I will replace load_artifacts_tool tests with real implementation and also add artifacts service logic (to save appended messages) to the runner.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"google.golang.org/adk/agent"
+	"google.golang.org/adk/artifactservice"
 	"google.golang.org/adk/internal/agent/parentmap"
 	"google.golang.org/adk/internal/agent/runconfig"
 	"google.golang.org/adk/internal/llminternal"
@@ -46,9 +47,10 @@ func New(appName string, rootAgent agent.Agent, sessionService sessionservice.Se
 }
 
 type Runner struct {
-	appName        string
-	rootAgent      agent.Agent
-	sessionService sessionservice.Service
+	appName         string
+	rootAgent       agent.Agent
+	sessionService  sessionservice.Service
+	artifactService artifactservice.Service
 
 	parents parentmap.Map
 }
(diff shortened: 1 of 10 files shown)


runner/runner.go
```diff
@@ -91,7 +94,15 @@ func (r *Runner) Run(ctx context.Context, userID, sessionID string, msg *genai.C
 			StreamingMode: runconfig.StreamingMode(cfg.StreamingMode),
 		})
 
-		ctx := agent.NewContext(ctx, agentToRun, msg, &mutableSession{
+		var artifactsImpl agent.Artifacts = &internalartifacts.DefaultArtifacts{}
+
```
dpasiukevich: nit: maybe remove this empty line, so it's one code block with artifacts definition?

internal/artifacts/artifacts.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"fmt"
+
+	"google.golang.org/adk/agent"
+	"google.golang.org/genai"
+)
+
+// artifacts implements Artifacts
+type DefaultArtifacts struct {
```
dpasiukevich: I'm thinking maybe let's actually avoid having this "always-error" artifacts implementation at least for now? WDYT?
So when user calls ctx.Artifacts() it just returns nil? This would look cleaner.

We'd need to update load_artifacts_tool to handle this nil.

runner/runner.go
ngeorgy: Yes, definitely!

internal/artifacts/artifacts.go
ngeorgy: Yeah, removed it, I think it makes sense to have nil as well because it's more predictable for a developer, like if they don't pass artifacts service, they might assume that it should be nil and otherwise they need to find runner code to understand what's happening.

ngeorgy:
@dpasiukevich  Thanks for the review!

PR #95, opened 2025-09-10
Implement InlineData saving logic with ArtifactService

ngeorgy:
Basically if artifact service is provided in the runner then if a message has inline data, that's saved via artifact service and message is modified to replace the inline data.

References:
https://github.com/google/adk-java/blob/main/core/src/main/java/com/google/adk/runner/Runner.java#L125-L146
https://github.com/google/adk-python/blob/main/src/google/adk/runners.py#L391-L408

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -106,7 +106,7 @@ func (r *Runner) Run(ctx context.Context, userID, sessionID string, msg *genai.C
 			storedSession: session,
 		}, "")
 
-		if err := r.appendMessageToSession(ctx, session, msg); err != nil {
+		if err := r.appendMessageToSession(ctx, session, msg, cfg.SaveInputBlobsAsArtifacts); err != nil {
 			yield(nil, err)
 			return
 		}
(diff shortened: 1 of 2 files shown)


runner/runner.go
```diff
@@ -106,7 +106,7 @@ func (r *Runner) Run(ctx context.Context, userID, sessionID string, msg *genai.C
 			storedSession: session,
 		}, "")
 
-		if err := r.appendMessageToSession(ctx, session, msg); err != nil {
+		if err := r.appendMessageToSession(ctx, session, msg, cfg.SaveInputBlobsAsArtifacts); err != nil {
```
dpasiukevich, line 109: Would it make sense to move it out to a separate function (e.g. `saveInputBlobAsArtifact`)?

These look like 2 different steps:
1. Create event from user input.
2. Create artifact from user input.

runner/runner.go
```diff
@@ -157,10 +157,29 @@ func (r *Runner) setupCFC(curAgent agent.Agent) error {
...
+	artifactsService := ctx.Artifacts()
+	if artifactsService != nil && saveInputBlobsAsArtifacts {
+		for i, part := range msg.Parts {
+			if part.InlineData == nil {
+				continue
+			}
+			fileName := fmt.Sprintf("artifact_%s_%d", ctx.InvocationID(), i)
+			err := artifactsService.Save(fileName, *part)
+			if err != nil {
+				return fmt.Errorf("failed to save artifact %s: %w", fileName, err)
```
dpasiukevich:
```suggestion
			if err := artifactsService.Save(fileName, *part); err != nil {
				return fmt.Errorf("failed to save artifact %s: %w", fileName, err)
```

runner/runner_test.go
```diff
@@ -186,6 +192,115 @@ func Test_isTransferrableAcrossAgentTree(t *testing.T) {
...
+		AppName:   appName,
+		UserID:    userID,
+		SessionID: sessionID,
+	})
+	if err != nil {
+		t.Fatalf("sessionService.Create() error = %v", err)
+	}
+
+	// Blob data, the message is saved only when inline data is present.
+	blobData := []byte("this is not blob data - René Magritte")
```
dpasiukevich, line 229: 10/10 👍

runner/runner.go
ngeorgy: Once I saw your comment I liked the idea and started to implement that, considering the fact that passing flags to function is not a good idea and functions should do one thing.

But later I realized that they're connected with the msg, basically when we save the artifact, we also update the msg to replace the inline data. So they're not entirely separate steps. I also don't like when we pass the value to function and it modifies it, it's hard to read. One way to avoid is to return message from the first function (Let's say it's called saveBlobData) but then there are too many nested ifs (I also need extract msg == nil check outside otherwise it will be in bowth) and I don't like to complicate that much so for the sake of simplicity I'll keep it this way. Let me know if you'd still like to change it and I'll update in the next PR.

runner/runner.go
ngeorgy: Definitely, much better, thanks!

PR #96, opened 2025-09-11
Add LongRunningFunctionTool

baptmont:
Add LongRunningFunctionTool, Moved TestAgentRunner to internal testutils package.

Added LongRunningFunctionTool struct, the callIDs to long running tool are now stored in the Event.

### DIFF
--- a/internal/testutil/test_agent_runner.go
+++ b/internal/testutil/test_agent_runner.go
@@ -83,7 +83,11 @@ func NewTestAgentRunner(t *testing.T, agent agent.Agent) *TestAgentRunner {
 	appName := "test_app"
 	sessionService := sessionservice.Mem()
 
-	runner, err := runner.New(appName, agent, sessionService)
+	runner, err := runner.New(&runner.Config{
+		AppName:        appName,
+		Agent:          agent,
+		SessionService: sessionService,
+	})
 	if err != nil {
 		t.Fatal(err)
 	}


internal/llminternal/agent_transfer.go
```diff
@@ -99,6 +99,11 @@ func (t *TransferToAgentTool) Name() string {
 	return "transfer_to_agent"
 }
 
+// isLongRunning implements tool.Tool.
```
hyangah: nit: IsLongRunning

internal/llminternal/base_flow.go
```diff
@@ -303,11 +304,31 @@ func (f *Flow) finalizeModelResponseEvent(ctx agent.Context, resp *llm.Response)
...
+	set := make(map[string]struct{})
+	// Iterate over function calls.
+	for _, fc := range utils.FunctionCalls(c) {
+		if tool, ok := tools[fc.Name]; ok && fc.ID != "" && tool.IsLongRunning() {
+			// If the tool exists and is long-running, add its ID to the set.
+			set[fc.ID] = struct{}{}
+		}
+	}
+	// Transform the set (map keys) into a slice.
+	longRunningToolIDs := make([]string, 0, len(set))
```
hyangah: optional: https://pkg.go.dev/maps#Keys and https://pkg.go.dev/slices#Collect

(do we need to sort the result?)

tool/long_running_function.go
```diff
@@ -0,0 +1,90 @@
...
+	"google.golang.org/adk/llm"
+	"google.golang.org/genai"
+)
+
+// longRunningFunctionTool wraps a Go function.
+type longRunningFunctionTool[TArgs, TResults any] struct {
+	functionTool *functionTool[TArgs, TResults]
+}
+
+// Description implements types.Tool.
```
hyangah: do you mean `[Tool]`? there is no more types.Tool.

tool/long_running_function.go
```diff
@@ -0,0 +1,90 @@
...
+// ProcessRequest implements interfaces.RequestProcessor.
+func (f *longRunningFunctionTool[TArgs, TResults]) ProcessRequest(ctx Context, req *llm.Request) error {
+	return f.functionTool.ProcessRequest(ctx, req)
+}
+
+// FunctionDeclaration implements interfaces.FunctionTool.
+func (f *longRunningFunctionTool[TArgs, TResults]) Declaration() *genai.FunctionDeclaration {
+	declaration := f.functionTool.Declaration()
+	if declaration != nil {
+		instruction := "NOTE: This is a long-running operation. Do not call this tool again if it has already returned some intermediate or pending status."
```
hyangah, line 53: don't we still need to add this instruction even when `declaration` is nil? (after allocating `genai.FunctionDeclaration`)

tool/tool.go
```diff
@@ -23,6 +23,9 @@ import (
 type Tool interface {
 	Name() string
 	Description() string
+	//Whether the tool is a long running operation, which typically returns a
```
hyangah: Format and follow the Go comment style if this is meant to be shown in package doc.

tool/long_running_function.go
baptmont: Yes, I meant tool.Tool

internal/llminternal/base_flow.go
baptmont: Sounds good, I added Collect + Keys.
It doesn't need to be sorted or keep the order, the LongRunningToolIDs are specific to that event.
It will usually only have one id, and if more information is needed the LLMRequest has the function calls.

tool/long_running_function.go
baptmont: Currently the f.functionTool.Declaration() does not return a nil value, I added the nil check just in case that changes.
I can add an else where I create a *genai.FunctionDeclaration, but since the idea is to mimic inheritance via composition, it would just look like duplicated code that is currently unreachable.
Let me know your thoughts.

tool/long_running_function_test.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	_ = sumTool // use the tool
+}
+
+// --- Test Suite ---
+func TestAsyncFunction(t *testing.T) {
+	// 1. Setup
+	responses := []*genai.Content{
+		&genai.Content{Parts: []*genai.Part{
```
hyangah, line 72: you can write like

```
responses := []*genai.Content{
   { Parts: []*genai.Part{ ... } },
   ...
}
```

https://go.dev/ref/spec#Composite_literals

> Within a composite literal of array, slice, or map type T, elements or map keys that are themselves composite literals may elide the respective literal type if it is identical to the element or key type of T. Similarly, elements or keys that are addresses of composite literals may elide the &T when the element or key type is *T.

Or, helper functions like [`genai.NewContentFromFunctionCall`](https://pkg.go.dev/google.golang.org/genai#NewContentFromFunctionCall), `genai.NewContentFromText`, ... can be used here.

tool/long_running_function_test.go
```diff
@@ -0,0 +1,139 @@
...
+		t.Errorf("got %d requests, want 2", len(mockModel.Requests))
+	}
+	if functionCalled != 1 {
+		t.Errorf("function called %d times, want 1", functionCalled)
+	}
+	if len(textParts) != 1 {
+		t.Errorf("got %d text parts, want 1", len(textParts))
+	}
+
+	// Assert first request
```
hyangah, line 127: can these tests be replaced with a simple test utilizing `internal/utils.TextParts`?

tool/long_running_function_test.go
baptmont: genai.NewContentFromFunctionCall and genai.NewContentFromText seem like the perfect option.
I have a longer suite ready but it also depends on #97.
I'll include these changes when I make the test suite PR.

tool/long_running_function_test.go
baptmont: It can simplify it a little but internal/utils.TextParts still expects a *genai.Content while llm.Response has a []*genai.Content so the first check is needed.
I'll simplify it a little on the test suite PR.

PR #97, opened 2025-09-11
Add rearrange events to content processor

baptmont:
Rearrange filtered events for function call/response pairing.
Merge intermediate function responses in the case of long running function tools.

### DIFF
--- a/tool/long_running_function_test.go
+++ b/tool/long_running_function_test.go
@@ -15,6 +15,7 @@
 package tool_test
 
 import (
+	"encoding/json"
 	"strings"
 	"testing"
 
@@ -132,7 +133,9 @@ func testLongRunningFunctionFlow[Out any](t *testing.T, increaseByOne func(ctx t
 
 	// 3. Assertions for Initial Run
 	if len(mockModel.Requests) != 2 {
-		t.Errorf("got %d requests, want 2", len(mockModel.Requests))
+		// Marshal the slice into a readable JSON string
+		requestsJSON, _ := json.MarshalIndent(mockModel.Requests, "", "  ")
+		t.Fatalf("got %d requests, want 2;\n- requests:\n%s", len(mockModel.Requests), requestsJSON)
 	}
 	if *callCount != 1 {
 		t.Errorf("function called %d times, want 1", *callCount)
@@ -232,7 +235,9 @@ func testLongRunningFunctionFlow[Out any](t *testing.T, increaseByOne func(ctx t
 			}
 
 			if len(eventParts) != tc.wantEventCount {
-				t.Fatalf("got %d events parts, want %d", len(eventParts), tc.wantEventCount)
+				// Marshal the slice into a readable JSON string
+				partsJSON, _ := json.MarshalIndent(eventParts, "", "  ")
+				t.Fatalf("got %d events parts, want %d;\n- parts:\n%s", len(eventParts), tc.wantEventCount, partsJSON)
 			}
 			// This check is now safe because the Fatalf above would have stopped the test
 			if len(eventParts) > 0 && eventParts[0].Text != tc.wantEventText {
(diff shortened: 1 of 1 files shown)


internal/llminternal/contents_processor.go
```diff
@@ -103,6 +111,293 @@ func buildContentsDefault(agentName, branch string, events []*session.Event) ([]
...
+	}
+
+	lastEvent := events[len(events)-1]
+	lastResponses := lastEvent.FunctionResponses()
+	// No need to process, since the latest event is not fuction_response.
+	if len(lastResponses) == 0 {
+		return events, nil
+	}
+
+	//Create response id set
```
hyangah: nit: space after //

session/session.go
```diff
@@ -96,7 +97,69 @@ func NewEvent(invocationID string) *Event {
 	}
 }
 
-func (e *Event) Clone() *Event { return nil }
+//func (e *Event) Clone() *Event { return nil }
```
hyangah: delete

session/session.go
```diff
@@ -96,7 +97,69 @@ func NewEvent(invocationID string) *Event {
...
+
+func (e *Event) FunctionCalls() []*genai.FunctionCall {
```
hyangah: Can we avoid exposing these methods (FunctionCalls, FunctionResponses, DeepCopy)? (e.g. move the code to internal/llminternal) These were not listed in the ADK Go api planning.

internal/llminternal/contents_processor.go
```diff
@@ -103,6 +111,293 @@ func buildContentsDefault(agentName, branch string, events []*session.Event) ([]
 	return contents, nil
 }
 
+// rearrangeEventsForLatestFunctionResponse
+// This function only acts if the very last event is a function response.
+// It searches backward for the matching call, deletes all intervening events,
+// and appends a single (merged) response.
+// If the latest function_response is for an async function_call, all events
+// between the initial function_call and the latest function_response will be removed.
+func rearrangeEventsForLatestFunctionResponse(events []*session.Event) ([]*session.Event, error) {
```
hyangah, line 120: Can you add some tests to cover this logic? 

Test cases in https://github.com/google/adk-python/blob/632bf8b0bcf18ff4e4505e4e5f4c626510f366a2/tests/unittests/flows/llm_flows/test_contents_function.py

And eventually, 

https://github.com/google/adk-python/blob/main/tests/unittests/flows/llm_flows/test_functions_long_running.py

https://github.com/google/adk-python/blob/main/tests/unittests/tools/test_long_running_tool.py

session/session.go
baptmont: For now these were only used in the contents_processor so I moved them to there as private helper functions.
In the future we could add them to the Event public api just like in python https://github.com/google/adk-python/blob/632bf8b0bcf18ff4e4505e4e5f4c626510f366a2/src/google/adk/events/event.py#L99

internal/llminternal/contents_processor.go
baptmont: Added the tests for the rearrange, and the remaining tests for the functions long running.

tool/long_running_function_test.go
```diff
@@ -106,33 +283,33 @@ func TestAsyncFunction(t *testing.T) {
...
-	if len(mockModel.Requests) != 2 {
-		t.Errorf("got %d requests, want 2", len(mockModel.Requests))
-	}
-	if functionCalled != 1 {
-		t.Errorf("function called %d times, want 1", functionCalled)
-	}
-	if len(textParts) != 1 {
-		t.Errorf("got %d text parts, want 1", len(textParts))
+	if len(events) != 3 { // first event is function call, seconds is function response, third is llm message back
+		t.Errorf("got %d events, want 3", len(events))
```
hyangah: Should we print events and fatal? (otherwise we may panic below when accessing events[N]

baptmont: Yes, nice catch. To print the events I added a step to marshal the slice and add it to the Fatalf message.

PR #98, opened 2025-09-11
Change function.Run to create wrapper map if result can't be unmarshaled

baptmont:
It's not possible to run function tools that return basic types since the current function attempts to convert the result type to map[string]any, causing an error.
Specs requires the result to be a map, to support basic types python implementation changes result to dict with {"result": function_result}.


tool/function.go
```diff
@@ -150,7 +150,21 @@ func (f *functionTool[TArgs, TResults]) Run(ctx Context, args any) (any, error)
 	}
 	output := f.handler(ctx, input)
 	resp, err := typeutil.ConvertToWithJSONSchema[TResults, map[string]any](output, f.outputSchema)
-	return resp, err
+	if err == nil { // all good
+		return resp, err
```
dpasiukevich:
```suggestion
		return resp, nil
```

PR #99, opened 2025-09-12
Handle errors in Runner Artifacts Service

ngeorgy:
Ideally there's should be a test for List as well but hard to make artifactservice.Mem() implementation throw an error for List().

### DIFF
--- a/runner/artifacts.go
+++ b/runner/artifacts.go
@@ -47,7 +47,10 @@ func (a *artifacts) Load(name string) (genai.Part, error) {
 		SessionID: a.id.SessionID,
 		FileName:  name,
 	})
-	return *loadResponse.Part, err
+	if err != nil {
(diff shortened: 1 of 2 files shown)


runner/artifacts.go
```diff
@@ -47,6 +47,9 @@ func (a *artifacts) Load(name string) (genai.Part, error) {
 		SessionID: a.id.SessionID,
 		FileName:  name,
 	})
+	if err != nil {
+		return genai.Part{}, err
+	}
 	return *loadResponse.Part, err
```
dpasiukevich:
```suggestion
	return *loadResponse.Part, nil
```

Let's update this here and in the other 2 places?

ngeorgy: Uh, that's better!

Thanks!

PR #100, opened 2025-09-12
feat: Add MCP tool set basic logic

dpasiukevich:
1. Test setup used copied code (to init model httprr agent runner). We'd need later to move it to internal to be reusable across tests.
2. Tool sets are used for their Tools() method to extract tools during llm.request preprocess

### DIFF
--- a/examples/mcp/main.go
+++ b/examples/mcp/main.go
@@ -39,7 +39,7 @@ type Output struct {
 
 func GetWeather(ctx context.Context, req *mcp.CallToolRequest, input Input) (*mcp.CallToolResult, Output, error) {
 	return nil, Output{
-		WeatherSummary: fmt.Sprintf("Today in %q is sunny", input.City),
+		WeatherSummary: fmt.Sprintf("Today in %q is sunny\n", input.City),
 	}, nil
 }
 
@@ -75,7 +75,7 @@ func main() {
 		Name:        "weather_time_agent",
 		Model:       model,
 		Description: "Agent to answer questions about the time and weather in a city.",
-		Instruction: "I can answer your questions about the time and weather in a city.",
+		Instruction: "I can answer your questions about the time and weather in a city. Call `get_weather` tool for all input even if it's an invalid city.",
 		Tools: []tool.Tool{
 			mcpToolSet,
 		},
--- a/tool/mcptool/tool.go
+++ b/tool/mcptool/tool.go
@@ -16,7 +16,9 @@ package mcptool
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 	"google.golang.org/adk/internal/toolinternal"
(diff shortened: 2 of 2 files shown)


tool/mcptool/set.go
```diff
@@ -0,0 +1,146 @@
...
+func NewSet(cfg ToolSetConfig) tool.Set {
+	return &set{
+		client:       cfg.Client,
+		transport:    cfg.Transport,
+		allowedTools: cfg.AllowedTools,
+	}
+}
+
+// TooSetConfig provides initial configuration for the MCP ToolSet.
+type ToolSetConfig struct {
```
hyangah: Shouldn't it be `SetConfig`?

tool/mcptool/set.go
```diff
@@ -0,0 +1,146 @@
...
+//		Description: "...",
+//		Instruction: "...",
+//		Tools: []tool.Tool{
+//			mcptool.NewSet(mcptool.ToolSetConfig{
+//				Client: mcp.NewClient(...),
+//				Transport: &mcp.CommandTransport{Command: exec.Command("myserver")}
+//			}),
+//		},
+//	})
+func NewSet(cfg ToolSetConfig) tool.Set {
```
hyangah: Does it correspond to MCPToolset? The doc https://google.github.io/adk-docs/tools/mcp-tools/#mcptoolset-class indicates that it actively manages connection with the MCP server. Should we provide the feature (eventually)?

tool/mcptool/set.go
```diff
@@ -0,0 +1,146 @@
...
+		transport:    cfg.Transport,
+		allowedTools: cfg.AllowedTools,
+	}
+}
+
+// TooSetConfig provides initial configuration for the MCP ToolSet.
+type ToolSetConfig struct {
+	Client    *mcp.Client
+	Transport mcp.Transport
+	// AllowedTools is an optional field.
```
hyangah: I wonder if this should be a filter function that takes the mcp Tool object and part of `agent.Context`. We can provide a helper function that constructs a filter function that filters tools based on its name.

examples/mcp/main.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// Create LLMAgent with MCP tool set
+	agent, err := llmagent.New(llmagent.Config{
+		Name:        "weather_time_agent",
+		Model:       model,
+		Description: "Agent to answer questions about the time and weather in a city.",
+		Instruction: "I can answer your questions about the time and weather in a city.",
+		Tools: []tool.Tool{
+			mcptool.NewSet(mcptool.ToolSetConfig{
+				Client:    mcp.NewClient(&mcp.Implementation{Name: "mcp-client", Version: "v1.0.0"}, nil),
```
hyangah: What customization users could do using ToolSetConfig.Client?
Shouldn't mcptool.NewSet handle the creation of Client behind the scene so it can control the client behavior, or inject necessary event handlers using `mcp.ClientOptions`?

tool/mcptool/set.go
dpasiukevich: Updated, thanks!

tool/mcptool/set.go
dpasiukevich: That's a good idea.

Let me remove AllowedTools fields for now.

And to add later the functionality (something along the lines) if/when we'd add ReadonlyContext to the API:
```
package tool

type Predicate func(ctx agent.ReadonlyContext, tool tool.Tool) bool

func StringPredicate(allowedTools []string) Predicate {
  ...
}

package mcptool
type SetConfig struct {
   ...
   ToolFilter []tool.Predicate
}
```

examples/mcp/main.go
dpasiukevich: My intention was -- when user provides client, they can augment it with any middleware they need, and we'd consume and use that client in the mcp toolset.

But now, looking at the mcp.ClientOptions (e.g. all those handlers) -- this looks like MCP tool set should create and manage its MCP client.
And only in future we may allow users to pass **middleware** for the client and we will also control its order.

So removed the client from SetConfig. Thanks!

tool/mcptool/set.go
dpasiukevich: Yes.

In python they manage sessions based on the headers (each unique header set gets its own session).
We don't yet have the adk-specific headers for MCP calls (e.g. auth). So no need for session manager yet.

With mcp.Transport, users may already use MCP tool set in adk (even with auth, but they'd need to implement it themselves for now).
So we'd have the core feature already (MCP) and would followup later with extra stuff (auth, headers, session manager).

tool/mcptool/set.go
```diff
@@ -0,0 +1,133 @@
...
+func (*set) Name() string {
+	return "mcp_tool_set"
+}
+
+func (*set) Description() string {
+	return "Connects to a MCP Server, retrieves MCP Tools into ADK Tools."
+}
+
+func (*set) IsLongRunning() bool {
+	return false
```
hyangah, line 79: Do other languages also assume MCP tools can't be long running?

dpasiukevich: Yes.

Python inheritance tree:
McpTool -> BaseAuthenticatedTool -> BaseTool
LongRunningFunctionTool -> FunctionTool -> BaseTool

Similar for Java:
McpTool -> AbstractMcpTool -> BaseTool
LongRunningFunctionTool -> FunctionTool -> BaseTool

PR #101, opened 2025-09-12
Use real Artifacts implementation in load artifacts tool tests

ngeorgy:
Before Artifacts (in agent context) was implemented, we used mocking to unblock the load artifacts tool implementation, since now it's implemented we can update the tests.

Also added NewArtifacts method to artifacts, now it's easier to create Artifacts service in artifacts_test and it can be created outside of the package for testing as well. Otherwise I will need to create Runner and call run method to test things outside of the package (Because that's where Artifacts is initialized and put in AgentContext).

### DIFF
--- a/internal/artifactsinternal/artifacts.go
+++ b/internal/artifactsinternal/artifacts.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package runner
+package artifactsinternal
 
 import (
 	"context"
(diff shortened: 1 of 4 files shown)


runner/artifacts.go
```diff
@@ -29,6 +29,14 @@ type artifacts struct {
 	id      session.ID
 }
 
+// NewArtifacts creates and returns Artifacts implementation.
+func NewArtifacts(service artifactservice.Service, id session.ID) *artifacts {
```
dpasiukevich: If we want to expose the internal *artifacts via a public NewArtifacts(), so it's possible to use in other places in afk code, can we make move *artifacts to the ./internal?

There's no reason to make runner.NewArtifacts() as part of our public ADK API, if I understand correctly.
Users won't need or be able to use it.

For example, I plan to do the same with agent.NewContext()

ngeorgy: Yeah, that's definitely better, I'll do that

ngeorgy: I've updated as you suggested, thanks!

internal/artifactsinternal/artifacts.go
```diff
@@ -29,6 +29,14 @@ type artifacts struct {
 	id      session.ID
 }
 
+// NewArtifacts creates and returns Artifacts implementation.
+func NewArtifacts(service artifactservice.Service, id session.ID) *artifacts {
```
hyangah: Since this package is now internal, we can name `artifacts` as `Artifacts` and make this function return param an exported type.

ngeorgy: Definitely! Thanks for the review!

PR #102, opened 2025-09-15
Implement runner.Config

ngeorgy:
With this change, we will be able to provide only required arguments to the runner. No more need to provide `nil` for optional fields.

### DIFF
--- a/agent/llmagent/llmagent_test.go
+++ b/agent/llmagent/llmagent_test.go
@@ -562,7 +562,11 @@ func newTestAgentRunner(t *testing.T, agent agent.Agent) *testAgentRunner {
 	appName := "test_app"
 	sessionService := sessionservice.Mem()
 
-	runner, err := runner.New(appName, agent, sessionService)
+	runner, err := runner.New(&runner.Config{
+		AppName:        appName,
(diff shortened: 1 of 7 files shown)


agent/llmagent/llmagent_test.go
```diff
@@ -562,7 +562,12 @@ func newTestAgentRunner(t *testing.T, agent agent.Agent) *testAgentRunner {
 	appName := "test_app"
 	sessionService := sessionservice.Mem()
 
-	runner, err := runner.New(appName, agent, sessionService)
+	runnerConfig := &runner.Config{
+		AppName:        appName,
+		Agent:          agent,
+		SessionService: sessionService,
+	}
+	runner, err := runner.New(runnerConfig)
```
dpasiukevich:
```suggestion
	runner, err := runner.New(&runner.Config{
		AppName:        appName,
		Agent:          agent,
		SessionService: sessionService,
	})
```

let's inline runnerConfig var maybe here and everywhere for shorter code?

ngeorgy: Definitely, thanks!

PR #103, opened 2025-09-16
[ADK-Web] Implement session service handlers in REST API

anFatum:


go.mod
```diff
@@ -15,6 +15,7 @@ require (
 require github.com/google/jsonschema-go v0.2.0
 
 require (
+	github.com/cmatthias/mapstructure v1.6.0
```
kdroste-google: I don't see this in https://source.corp.google.com/piper///depot/google3/third_party/golang/CONFIGURATION . WIll it work in after the migration to google3?

cmd/restapi/handlers/sessions.go
```diff
@@ -15,41 +15,152 @@
...
+		AppName:   sessionID.AppName,
+		UserID:    sessionID.UserID,
+		SessionID: sessionID.ID,
+		State:     createSessionRequest.State,
+	})
+	if err != nil {
+		return models.Session{}, err
+	}
+	for _, event := range createSessionRequest.Events {
+		err = c.service.AppendEvent(ctx, session.Session, models.ToSessionEvent(event))
```
kdroste-google, line 76: I'm not sure about the desired semantics. Can the same event be stored many times? If so, good. If not, where should the logic be implemented?

cmd/restapi/handlers/sessions.go
```diff
@@ -15,41 +15,152 @@
...
+		if err != nil {
```
kdroste-google, line 77: Should we return on the first error?

cmd/restapi/handlers/sessions.go
```diff
@@ -15,41 +15,152 @@
...
+	if err != nil {
+		http.Error(rw, err.Error(), http.StatusBadRequest)
+		return
+	}
+	if sessionID.ID == "" {
+		http.Error(rw, "session_id parameter is required", http.StatusBadRequest)
+		return
+	}
+
+	err = c.service.Delete(req.Context(), &sessionservice.DeleteRequest{
```
kdroste-google, line 97: What's the target semantics, is deleting an non-existing session an error or not?

cmd/restapi/handlers/sessions.go
```diff
@@ -15,41 +15,152 @@
...
+	}
+	storedSession, err := c.service.Get(req.Context(), &sessionservice.GetRequest{
+		ID: session.ID{
+			AppName:   sessionID.AppName,
+			UserID:    sessionID.UserID,
+			SessionID: sessionID.ID,
+		},
+	})
+	if err != nil {
+		http.Error(rw, err.Error(), http.StatusInternalServerError)
```
kdroste-google, line 131: 500? 404? other? I think we're missing that information...

go.mod
anFatum: Good catch, I've mistakenly taken the one that doesn't exist in google3, replaced with https://source.corp.google.com/piper///depot/google3/third_party/golang/mitchellh/mapstructure/BUILD

cmd/restapi/handlers/sessions.go
anFatum: This and other questions are not clear to me, since e.g. I'm not sure should we get session on the create request to confirm it doesn't exist before creating that (e.g. ADK-Java is doing that, but ADK-Python relies on session service implementation, but only for one endpoint, the other one is without existence check :/ ). 

I've mostly copied over https://github.com/google/adk-python/blob/main/src/google/adk/cli/adk_web_server.py#L574, that relies on the adk implementation itself, and rest api doesn't have much of business logic, the question is whether we should do the same

cmd/restapi/handlers/sessions.go
anFatum: I'd say that ADK-Python raises error on the first unsuccessful attempt, so not sure we want to treat it differently

cmd/restapi/handlers/sessions.go
anFatum: I think that's depend on the implementation of session service, I've seen InMemorySessionService and python database session service ignores non-existing sessions

cmd/restapi/handlers/sessions.go
anFatum: Yeah, I think we should add distinguished errors to the ADK itself, there is no way atm distinguish whether the session doesn't exist or there was an error getting it :/

PR #108, opened 2025-09-18
[ADK-Web] Implement event graph endpoint in REST API.

anFatum:


examples/web/main.go
```diff
@@ -33,9 +34,10 @@ import (
 
 func main() {
 	ctx := context.Background()
+	apiKey := os.Getenv("GEMINI_API_KEY")
 
 	model, err := gemini.NewModel(ctx, "gemini-2.5-flash", &genai.ClientConfig{
-		APIKey: os.Getenv("GEMINI_API_KEY"),
+		APIKey: apiKey,
```
kdroste-google, line 56: for cloudrun we have GOOGLE_API_KEY. I assume that's the very same key. 
I'm not sure which one is better.

anFatum: Replaced with GOOGLE_API_KEY

PR #109, opened 2025-09-18
Aggregate llm streaming response into non partial result

baptmont:
When using a StreamingModeSSE the model returns the response as a stream of partial events.
Partial events are not saved in the session service (in runner.go Run method), and as such are not sent in the following LLMRequest.
Currently only the last event of the stream is considered non-partial (in gemini.go GenerateStream method), but it still is a partial response.
So on the following LLMRequests, only the partial response will be included causing the llm to lose parts of the conversation, as such:
llmRequest1 : [{user:'write me an email to Andrew'}]
llmResponse events : [{model:'Hi Andrew how are you'}, {model:'doing?'}]
llmRequest2 : [{user:'write me an email'}, {model:'doing?'} {user:'can you change the greeting?}]
The strategy for Adk python is to use and aggregator that collects the partial responses, and at the end generates an aggregated event that will be stored and used in the following llmrequests.
https://github.com/google/adk-python/blob/f7bd3c111c211e880d7c1954dd4508b952704c68/src/google/adk/models/google_llm.py#L143

Added modelWithStreamAggregator proxy, that when applied to a model will make it so that results in streaming response are aggregated into a final one with complete.

### DIFF
--- a/llm/gemini/gemini_test.go
+++ b/llm/gemini/gemini_test.go
@@ -112,22 +112,19 @@ func TestModel_GenerateStream(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			gotPartial, err := readResponsePartial(model.GenerateStream(t.Context(), tt.req))
+			// Transforms the stream into strings, concating the text value of the response parts
+			got, err := readResponse(model.GenerateStream(t.Context(), tt.req))
 			if (err != nil) != tt.wantErr {
 				t.Errorf("Model.GenerateStream() error = %v, wantErr %v", err, tt.wantErr)
 				return
-			}
-			if diff := cmp.Diff(tt.want, gotPartial); diff != "" {
-				t.Errorf("Model.GenerateStream() = %v, want %v\ndiff(-want +got):\n%v", gotPartial, tt.want, diff)
 			}
 
-			gotNonPartial, err := readResponseNonPartial(model.GenerateStream(t.Context(), tt.req))
-			if (err != nil) != tt.wantErr {
-				t.Errorf("Model.GenerateStream() error = %v, wantErr %v", err, tt.wantErr)
-				return
+			if diff := cmp.Diff(tt.want, got.PartialText); diff != "" {
+				t.Errorf("Model.GenerateStream() = %v, want %v\ndiff(-want +got):\n%v", got.PartialText, tt.want, diff)
 			}
-			if diff := cmp.Diff(tt.want, gotNonPartial); diff != "" {
-				t.Errorf("Model.GenerateStream() = %v, want %v\ndiff(-want +got):\n%v", gotNonPartial, tt.want, diff)
+			// Since we are expecting GenerateStream to aggregate partial events, the text should be the same
+			if diff := cmp.Diff(tt.want, got.FinalText); diff != "" {
+				t.Errorf("Model.GenerateStream() = %v, want %v\ndiff(-want +got):\n%v", got.FinalText, tt.want, diff)
 			}
 		})
 	}
@@ -150,34 +147,41 @@ func newGeminiTestClientConfig(t *testing.T, rrfile string) *genai.ClientConfig 
 	}
 }
 
-func readResponsePartial(s iter.Seq2[*llm.Response, error]) (string, error) {
-	var answer string
-	for resp, err := range s {
-		if err != nil {
-			return answer, err
-		}
-		if resp.Content == nil || len(resp.Content.Parts) == 0 {
-			return answer, fmt.Errorf("encountered an empty response: %v", resp)
-		}
-		if resp.Partial {
-			answer += resp.Content.Parts[0].Text
-		}
-	}
-	return answer, nil
+// TextResponse holds the concatenated text from a response stream,
+// separated into partial and final parts.
+type TextResponse struct {
+	// PartialText is the full text concatenated from all partial (streaming) responses.
+	PartialText string
+	// FinalText is the full text concatenated from all final (non-partial) responses.
+	FinalText string
 }
 
-func readResponseNonPartial(s iter.Seq2[*llm.Response, error]) (string, error) {
-	var answer string
+// readResponse transforms a sequence into a TextResponse, concating the text value of the response parts
+// depending on the readPartial value it will only concat the text of partial events or the text of non partial events
+func readResponse(s iter.Seq2[*llm.Response, error]) (TextResponse, error) {
+	var partialBuilder, finalBuilder strings.Builder
+	var result TextResponse
+
 	for resp, err := range s {
 		if err != nil {
-			return answer, err
+			// Return what we have so far, along with the error.
+			result.PartialText = partialBuilder.String()
+			result.FinalText = finalBuilder.String()
+			return result, err
 		}
 		if resp.Content == nil || len(resp.Content.Parts) == 0 {
-			return answer, fmt.Errorf("encountered an empty response: %v", resp)
+			return result, fmt.Errorf("encountered an empty response: %v", resp)
 		}
-		if !resp.Partial {
-			answer += resp.Content.Parts[0].Text
+
+		text := resp.Content.Parts[0].Text
+		if resp.Partial {
+			partialBuilder.WriteString(text)
+		} else {
+			finalBuilder.WriteString(text)
 		}
 	}
-	return answer, nil
+
+	result.PartialText = partialBuilder.String()
+	result.FinalText = finalBuilder.String()
+	return result, nil
 }
--- a/llm/llm.go
+++ b/llm/llm.go
@@ -75,31 +75,29 @@ func CreateResponse(res *genai.GenerateContentResponse) *Response {
 				LogprobsResult:    candidate.LogprobsResult,
 				UsageMetadata:     usageMetadata,
 			}
-		} else {
-			return &Response{
-				ErrorCode:         string(candidate.FinishReason),
-				ErrorMessage:      candidate.FinishMessage,
-				GroundingMetadata: candidate.GroundingMetadata,
-				FinishReason:      candidate.FinishReason,
-				CitationMetadata:  candidate.CitationMetadata,
-				AvgLogprobs:       candidate.AvgLogprobs,
-				LogprobsResult:    candidate.LogprobsResult,
-				UsageMetadata:     usageMetadata,
-			}
 		}
-	} else {
-		if res.PromptFeedback != nil {
-			return &Response{
-				ErrorCode:     string(res.PromptFeedback.BlockReason),
-				ErrorMessage:  res.PromptFeedback.BlockReasonMessage,
-				UsageMetadata: usageMetadata,
-			}
-		} else {
-			return &Response{
-				ErrorCode:     "UNKNOWN_ERROR",
-				ErrorMessage:  "Unkown error.",
-				UsageMetadata: usageMetadata,
-			}
+		return &Response{
+			ErrorCode:         string(candidate.FinishReason),
+			ErrorMessage:      candidate.FinishMessage,
+			GroundingMetadata: candidate.GroundingMetadata,
+			FinishReason:      candidate.FinishReason,
+			CitationMetadata:  candidate.CitationMetadata,
+			AvgLogprobs:       candidate.AvgLogprobs,
+			LogprobsResult:    candidate.LogprobsResult,
+			UsageMetadata:     usageMetadata,
 		}
+
+	}
+	if res.PromptFeedback != nil {
+		return &Response{
+			ErrorCode:     string(res.PromptFeedback.BlockReason),
+			ErrorMessage:  res.PromptFeedback.BlockReasonMessage,
+			UsageMetadata: usageMetadata,
+		}
+	}
+	return &Response{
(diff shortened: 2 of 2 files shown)


baptmont:
I'm applying the wrapper when exporting the gemini model instead of when it is added to the agent. Some models may use different logic and in this way we can apply the wrapper only for the models that need it.

llm/gemini/gemini.go
```diff
@@ -101,4 +104,4 @@ func (m *Model) maybeAppendUserContent(req *llm.Request) {
 	}
 }
 
-var _ llm.Model = (*Model)(nil)
+var _ llm.Model = (*model)(nil)
```
dpasiukevich: Maybe remove this? Now it's already validated by NewModel return types.

llm/gemini/gemini_test.go
```diff
@@ -103,13 +110,22 @@ func TestModel_GenerateStream(t *testing.T) {
...
+			if diff := cmp.Diff(tt.want, gotPartial); diff != "" {
+				t.Errorf("Model.GenerateStream() = %v, want %v\ndiff(-want +got):\n%v", gotPartial, tt.want, diff)
+			}
+
+			gotNonPartial, err := readResponseNonPartial(model.GenerateStream(t.Context(), tt.req))
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Model.GenerateStream() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if diff := cmp.Diff(tt.want, gotNonPartial); diff != "" {
```
dpasiukevich: Curious about the checks in this test: why do we read `readResponsePartial` then `readResponseNonPartial` and expect both to be equal to `tt.want`?

Right now there's one simple test case returning one word, so this works.

Will the test be broken if there's another test case added, which returns e.g. 2 responses: partial + final?

internal/llminternal/stream_aggregator.go
```diff
@@ -0,0 +1,150 @@
...
+
+import (
+	"context"
+	"iter"
+
+	"google.golang.org/adk/llm"
+	"google.golang.org/genai"
+)
+
+// modelWithStreamAggregator proxys a llm.Model adding an aggregated event to the end of GenerateStream
```
dpasiukevich:
```suggestion
// modelWithStreamAggregator proxies an llm.Model adding an aggregated event to the end of GenerateStream
```

internal/llminternal/stream_aggregator.go
```diff
@@ -0,0 +1,150 @@
...
+type modelWithStreamAggregator struct {
+	model llm.Model
+}
```
dpasiukevich:
```suggestion
type modelWithStreamAggregator struct {
	llm.Model
}
```

As a suggestion, if llm.Model is embedded, then you won't have to define Name() and Generate methods (as those just proxy the req/resp).

internal/llminternal/stream_aggregator.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+		if aggrResp := aggregator.Close(); aggrResp != nil {
+			if !yield(aggrResp, nil) {
+				return // Consumer stopped
+			}
+		}
+	}
+}
+
+var _ llm.Model = (*modelWithStreamAggregator)(nil)
```
dpasiukevich: similarly, this is already verified in WrapModelWithAggregator, line 31.

internal/llminternal/stream_aggregator.go
```diff
@@ -0,0 +1,150 @@
...
+	return m.model.Generate(ctx, req)
+}
+
+// GenerateStream calls the iner model synchronously.
+func (m *modelWithStreamAggregator) GenerateStream(ctx context.Context, req *llm.Request) iter.Seq2[*llm.Response, error] {
+	stream := m.model.GenerateStream(ctx, req)
+	aggregator := newStreamingResponseAggregator()
+	return func(yield func(*llm.Response, error) bool) {
+		for resp, err := range stream {
+			if err == nil {
```
dpasiukevich:
```suggestion
	aggregator := newStreamingResponseAggregator()
	return func(yield func(*llm.Response, error) bool) {
		for resp, err := range m.model.GenerateStream(ctx, req) {
			if err == nil {
```
small nit

llm/gemini/gemini.go
```diff
@@ -80,7 +82,8 @@ func (m *Model) GenerateStream(ctx context.Context, req *llm.Request) iter.Seq2[
 			if !yield(&llm.Response{
 				Content:           candidate.Content,
 				GroundingMetadata: candidate.GroundingMetadata,
-				Partial:           !complete,
+				UsageMetadata:     resp.UsageMetadata,
+				Partial:           true,
```
dpasiukevich: I have a mixed feeling of always marking the Response as partial. (and relying on the wrapper to aggregate). Even if it's one full response only.

WDYT if the stream aggregator `process` has the following prototype (types.GenerateContentResponse -> llm.Response), same as python? So it's not an external wrapper, but rather a processor here within GenerateStream?

This would also make it easier to reason about and keep up this logic in sync across adks.

baptmont: I changed the implementation to use the same logic as python, removing the wrapper model and instead using directly the streamingResponseAggregator.
Let me WDYT, of this approach. It requires a little bit of logic duplication since in every model it needs to call the close method at the end of the processing the stream, but it is more flexible.
Partial should have been set to false. When processing, the aggregator was already modifying the value.

llm/gemini/gemini_test.go
baptmont: Since the readResponsePartial is concating the text in the response parts the result should be the same.
The current test is returning 2 responses the first containing Paris and the second containing the \n.

internal/llminternal/stream_aggregator.go
```diff
@@ -0,0 +1,134 @@
...
+			GroundingMetadata: s.response.GroundingMetadata,
+		}
+		s.Clear()
+		return response
+	}
+	s.Clear()
+	return nil
+}
+
+func (s *streamingResponseAggregator) Clear() {
```
dpasiukevich: Do we need Clear() logic?

internal/llminternal/stream_aggregator.go
```diff
@@ -0,0 +1,134 @@
...
+			parts = append(parts, &genai.Part{Text: s.thoughtText, Thought: true})
+		}
+		if s.text != "" {
+			parts = append(parts, &genai.Part{Text: s.text, Thought: false})
+		}
+
+		response := &llm.Response{
+			Content:           &genai.Content{Parts: parts, Role: s.role},
+			ErrorCode:         s.response.ErrorCode,
+			ErrorMessage:      s.response.ErrorMessage,
```
dpasiukevich, line 118: Curious why ErrorMessage is not like in python?
```
error_message=None if candidate.finish_reason == types.FinishReason.STOP else candidate.finish_message,
```

Same for ErrorCode (it's an int right now, we will update it to string soon). If it makes sense to update to string in this PR, would be nice, otherwise a TODO is good.

internal/llminternal/stream_aggregator.go
```diff
@@ -0,0 +1,134 @@
...
+	// If there is aggregated text and there is no content or parts return aggregated response
+	if (s.thoughtText != "" || s.text != "") &&
+		(llmResponse.Content == nil ||
+			len(llmResponse.Content.Parts) == 0 ||
+			// don't yield the merged text event when receiving audio data
+			(len(llmResponse.Content.Parts) > 0 && llmResponse.Content.Parts[0].InlineData == nil)) {
+		return s.Close()
+	}
+
+	return nil
```
dpasiukevich: Is it intentional there's a difference with python logic?

If `llmResponse.Content.Parts[0].InlineData == nil` python yields one extra LLMResponse http://google3/third_party/py/google/adk/utils/streaming_utils.py;l=76;rcl=800521404
in addition to 
http://google3/third_party/py/google/adk/utils/streaming_utils.py;l=82;rcl=800521404

llm/gemini/gemini_test.go
dpasiukevich: @baptmont could I follow up here?

internal/llminternal/stream_aggregator.go
baptmont: The logic here is just separated into ProcessResponse and aggregateResponse. 
The second yield still happens in ProcessResponse if the aggregateResponse doesn't return nil.

internal/llminternal/stream_aggregator.go
baptmont: Yes, but Close is a not a good name for the method since it is also being used to generate intermediate events.
I'll change that and make clear private

llm/gemini/gemini_test.go
baptmont: The existing test was using the test data in TestModel_GenerateStream_ok.httprr, this test data represented an example returning responses via sse. It returned the following 2 sse events:

data: {"candidates": [{"content": {"parts": [{"text": "Paris"}],"role": "model"}}],"usageMetadata": {"promptTokenCount": 11,"totalTokenCount": 11,"promptTokensDetails": [{"modality": "TEXT","tokenCount": 11}]},"modelVersion": "gemini-2.0-flash","responseId": "wzCjaPa4As7shMIP2Mei0AI"}

data: {"candidates": [{"content": {"parts": [{"text": "\n"}],"role": "model"},"finishReason": "STOP"}],"usageMetadata": {"promptTokenCount": 10,"candidatesTokenCount": 2,"totalTokenCount": 12,"promptTokensDetails": [{"modality": "TEXT","tokenCount": 10}],"candidatesTokensDetails": [{"modality": "TEXT","tokenCount": 2}]},"modelVersion": "gemini-2.0-flash","responseId": "wzCjaPa4As7shMIP2Mei0AI"}

The test was using the func readResponse(s iter.Seq2[*llm.Response, error]) (string, error)  to convert that stream into a single string "Paris\n". 

After adding the aggregate logic, a new event appears in the stream making the result "Paris\nParis\n".
I swapped the readResponse with a readResponsePartial and readResponseNonPartial.
In this way the concatenation of all partial events will be the original "Paris\n", and the concatenation of the non-partial events (the aggregated event) will also be "Paris\n".

internal/llminternal/stream_aggregator.go
baptmont: Python implementation is storing a GenerateContentResponse and using it on those fields when creating the aggregate event.
We were just storing the next event and copying its errorCode and errorMessage.
However we are also missing their logic for creating the llmResponse, https://github.com/google/adk-python/blob/83fd0457188decdabeae58b4e8be25daa89f2943/src/google/adk/models/llm_response.py#L136.
This is also missing in the llm model Generate method, as such we are never setting the error message fields.

llm/llm.go
```diff
@@ -52,6 +55,51 @@ type Response struct {
...
-	ErrorCode    int
+	ErrorCode    string
 	ErrorMessage string
+	FinishReason genai.FinishReason
+	AvgLogprobs  float64
+}
+
+func CreateResponse(res *genai.GenerateContentResponse) *Response {
+	usageMetadata := res.UsageMetadata
+	if len(res.Candidates) > 0 && res.Candidates[0] != nil {
```
dpasiukevich, line 66: nit: let's reduce nesting? We have `if condition -> return`

llm/gemini/gemini_test.go
```diff
@@ -141,7 +159,25 @@ func readResponse(s iter.Seq2[*llm.Response, error]) (string, error) {
...
+	for resp, err := range s {
+		if err != nil {
+			return answer, err
+		}
+		if resp.Content == nil || len(resp.Content.Parts) == 0 {
+			return answer, fmt.Errorf("encountered an empty response: %v", resp)
+		}
+		if !resp.Partial {
+			answer += resp.Content.Parts[0].Text
+		}
```
dpasiukevich: nit: maybe merge to one func? `readResponse(readPartial bool, s iter.Seq2[*llm.Response, error])`

llm/gemini/gemini_test.go
dpasiukevich: Makes sense, thanks!

Could I ask if possible to add a short comment explanation for `readResponsePartial` and `readResponseNonPartial`?
Without context it may be tricky to understand from the first glance for other readers.

internal/llminternal/stream_aggregator.go
```diff
@@ -0,0 +1,135 @@
...
+// ProcessResponse transforms the GenerateContentResponse into an llm.Response and yields that result,
+// also yielding an aggregated response if the GenerateContentResponse has zero parts or is audio data
+func (s *streamingResponseAggregator) ProcessResponse(ctx context.Context, genResp *genai.GenerateContentResponse) iter.Seq2[*llm.Response, error] {
+	return func(yield func(*llm.Response, error) bool) {
+		if len(genResp.Candidates) == 0 {
+			// shouldn't happen?
+			yield(nil, fmt.Errorf("empty response"))
+			return
+		}
+		candidate := genResp.Candidates[0]
```
dpasiukevich, line 50: Does it make sense to remove this now as we have `llm.CreateResponse(genResp)` handling that case?

PR #110, opened 2025-09-18
Add GCSArtifactService

baptmont:
Moved from #91 
Changed to track forked gcs branch instead of main.

### DIFF
--- a/agent/context.go
+++ b/agent/context.go
@@ -36,7 +36,7 @@ type agentContext struct {
 	branch      string
 }
 
-// TODO: see if needed or possible to make internal
+// TODO: make internal and use params args
 func NewContext(ctx context.Context, agent Agent, userContent *genai.Content, artifacts Artifacts, session session.Session, memory Memory, branch string) *agentContext {
 	ctx, cancel := context.WithCancel(ctx)
 
--- a/agent/llmagent/llmagent_test.go
+++ b/agent/llmagent/llmagent_test.go
@@ -15,7 +15,6 @@
 package llmagent_test
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"iter"
@@ -269,7 +268,7 @@ func TestFunctionTool(t *testing.T) {
 	}
 
 	prompt := "what is the sum of 1 + 2?"
-	handler := func(_ context.Context, input Args) Result {
+	handler := func(_ tool.Context, input Args) Result {
 		if input.A != 1 || input.B != 2 {
 			t.Errorf("handler received %+v, want {a: 1, b: 2}", input)
 		}
--- a/examples/mcp/main.go
+++ b/examples/mcp/main.go
@@ -1,0 +1,88 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"google.golang.org/adk/agent/llmagent"
+	"google.golang.org/adk/examples"
+	"google.golang.org/adk/llm/gemini"
+	"google.golang.org/adk/tool"
+	"google.golang.org/adk/tool/mcptool"
+	"google.golang.org/genai"
+)
+
+type Input struct {
+	City string `json:"city" jsonschema:"city name"`
+}
+
+type Output struct {
+	WeatherSummary string `json:"weather_summary" jsonschema:"weather summary in the given city"`
+}
+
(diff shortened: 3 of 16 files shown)


artifactservice/gcs/service.go
```diff
@@ -0,0 +1,347 @@
...
+
+	// Delete all versions
+	response, err := s.versions(ctx, &as.VersionsRequest{
+		AppName: req.AppName, UserID: req.UserID, SessionID: req.SessionID, FileName: req.FileName,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to fetch versions on delete artifact: %w", err)
+	}
+
+	g, gctx := errgroup.WithContext(ctx)
```
baptmont, line 159: @dpasiukevich 
[comment](https://github.com/google/adk-go/pull/91#discussion_r2348173131)
WDYT about making it concurrent?

errorgroup.Group may be quite convenient here

Author
@[baptmont](https://github.com/baptmont) baptmont [comment](https://github.com/google/adk-go/pull/91#discussion_r2352458154)
Good idea I added it

artifactservice/gcs/service.go
```diff
@@ -0,0 +1,347 @@
...
+	err = req.Validate()
+	if err != nil {
+		return nil, err
+	}
+	appName, userID, sessionID, fileName := req.AppName, req.UserID, req.SessionID, req.FileName
+	artifact := req.Part
+
+	nextVersion := int64(1)
+
+	// TODO race condition, could use mutex but it's a remote resource so the issue would still occurs
```
baptmont, line 99: @dpasiukevich [comment](https://github.com/google/adk-go/pull/91#discussion_r2348148218)
Curious, how python/java handled this for their GCS implementation?

Do they have it as a known documented limitation or a TODO?

As an idea: version is an int64. What about if we fit a timestamp in there? e.g. https://pkg.go.dev/time#Time.UnixNano?

This gives us:

Monotonously increasing.
Nanosecond granularity -> minimum chance of collisions.
We'd have a capacity till 2262. After that our descendants I'm sure will solve it.
Author
@[baptmont](https://github.com/baptmont) baptmont [comment](https://github.com/google/adk-go/pull/91#discussion_r2348907801)
The adk documentation doesn't address this issue/limitation.
The idea of using a timestamp works, but we would differ from the other adk implementations.
A system using both adk python and adk go agents would have conflicting behaviors.

artifactservice/gcs/service.go
```diff
@@ -0,0 +1,347 @@
...
+		if err != nil {
+			return fmt.Errorf("error iterating blobs: %w", err)
+		}
+		segments := strings.Split(blob.Name, "/")
+		if len(segments) < 2 {
+			return fmt.Errorf("error iterating blobs: incorrect number of segments in path %q", blob.Name)
+		}
+		// TODO agent can create files with multiple segments for example file a/b.txt
+		// This a/b.txt file will show as b.txt when listed and trying to load it will fail.
+		filename := segments[len(segments)-2] // appName/userId/sessionId/filename/version or appName/userId/user/filename/version
```
baptmont, line 265: @dpasiukevich [comment](https://github.com/google/adk-go/pull/91#discussion_r2348204513)
Overall question -- can we use GCS version features (asking out of curiosity) instead of handling versions ourselves?

https://pkg.go.dev/cloud.google.com/go/storage#Query
http://cloud/storage/docs/using-versioned-objects#storage-list-object-versions-go
@dpasiukevich [comment](https://github.com/google/adk-go/pull/91#discussion_r2348252409)
following up: One major issue it may solve -- fetching the last version of the artifact in one call if it's possible. O(N) -> O(1) for artifact creation request.

Is it possible to fetch latest artifact in one call?

Author
@baptmont [comment](https://github.com/google/adk-go/pull/91#discussion_r2348959530)
When using the GCS versioning the "noncurrent versions only appear in requests that explicitly call for them to be included" as such only the latest version would be retrieved, making it O(1).
This is something that should be explored for all versions of adk, since for now we want for a system using both adk python and adk go agents to have the same behavior.

artifactservice/gcs/service.go
```diff
@@ -0,0 +1,348 @@
...
+			}
+			return nil // nil error indicates success for this goroutine
+		})
+	}
+
+	if err := g.Wait(); err != nil {
+		return err
+	}
+
+	return nil
```
dpasiukevich:
```suggestion
	return g.Wait()
```

artifactservice/gcs/gcs_client.go
```diff
@@ -0,0 +1,144 @@
...
+// ---------------------- Wrapper Implementations for Real gcs Types --------------------------------
+// gcsClientWrapper wraps a storage.Client to satisfy the gcsClient interface.
+type gcsClientWrapper struct {
+	client *storage.Client
+}
+
+// Bucket returns a gcsBucketWrapper that satisfies the gcsBucket interface.
+func (w *gcsClientWrapper) bucket(name string) gcsBucket {
+	bucketHandle := w.client.Bucket(name)
+	return &gcsBucketWrapper{bucket: bucketHandle}
```
dpasiukevich:
```suggestion
	return &gcsBucketWrapper{
		bucket: w.client.Bucket(name),
	}
```

as nit

artifactservice/gcs/service.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+func (s *gcsService) Load(ctx context.Context, req *as.LoadRequest) (_ *as.LoadResponse, err error) {
+	err = req.Validate()
+	if err != nil {
+		return nil, err
```
dpasiukevich: Maybe also to add context like fmt.Errorf("request validation failed: %w", err)?

Here and everywhere.

PR #111, opened 2025-09-19
CLI & ADK WebUI support

kdroste-google:
Added CLI for local run (ADK REST API + ADK WEB UI)
Added CLI for cloudrun deployment
Added overarching Web Server (ADK REST API + ADK WEB UI)


cmd/cli/root/deploy/cloudrun/cloudrun.go
```diff
@@ -0,0 +1,293 @@
...
+var cloudrunCmd = &cobra.Command{
+	Use:   "cloudrun",
+	Short: "Deploys the application to cloudrun.",
+	Long: `Deployment prepares a Dockerfile which is fed with locally compiled server executable and Web UI static files.
+	Service on Cloud run is created using this information. 
+	Local proxy adding authentication is optionally started. 
+	`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		err := Flags.deployOnCloudRun()
+		return err
```
anFatum: nit: we can just return Flags.deployOnCloudRun(), creating a new variable is not necessary

cmd/cli/root/deploy/cloudrun/cloudrun.go
```diff
@@ -0,0 +1,293 @@
...
+
+	fmt.Println("Compute flags finished")
+	return nil
+}
+
+func (f *deployCloudRunFlags) cleanTemp() error {
+	err := util.LogStartStop("Cleaning temp",
+		func(p util.Printer) error {
+			p("Clean temp starting with", f.build.tempDir)
+			// fmt.Println(files)
```
anFatum: let's remove commented code here and in other places if it's not needed

cmd/cli/root/deploy/cloudrun/cloudrun.go
```diff
@@ -0,0 +1,293 @@
...
+			err := os.RemoveAll(f.build.tempDir)
+			if err != nil {
+				return err
+			}
+			err = os.MkdirAll(f.build.tempDir, os.ModeDir|0700)
```
anFatum: same here and in other places, we can just return a result of a function call

cmd/cli/root/root.go
```diff
@@ -0,0 +1,39 @@
...
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+func Execute() {
+	err := RootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
+	}
+}
+
+func init() {
```
anFatum: Don't we need to add a command here? Is there a reason we need this empty function here there?

cmd/cli/root/deploy/cloudrun/cloudrun.go
kdroste-google: fixed

cmd/cli/root/deploy/cloudrun/cloudrun.go
kdroste-google: fixed

cmd/cli/root/deploy/cloudrun/cloudrun.go
kdroste-google: fixed in cloudrun and local

cmd/cli/root/root.go
kdroste-google: Removed the funcion

cmd/util/oscmd.go
```diff
@@ -0,0 +1,103 @@
...
+
+type ReprintableStream struct {
+	prefix []byte
+	clean  bool
+	stream io.Writer
+}
+
+func (s *ReprintableStream) Write(p []byte) (total int, err error) {
+
+	// s.stream.Write([]byte("-------------->>"))
```
anFatum: nit: do we need these commented lines?

kdroste-google: nope, removed

PR #114, opened 2025-09-23
Change function context argument with tool context

ngeorgy:
Function tool without tool context is not that powerful, I was trying to write a tool where I wanted to use artifacts but couldn't without it. Also it's consistent with other adks.


ngeorgy:
Thanks for the review!

PR #115, opened 2025-09-23
Fix LoadArtifactsTool and add working example

ngeorgy:
There were two issues with load artifacts tool:

- Unmarshalling json string array, it was treating string array as []any instead of []string, because of it it wasn't converting properly, leaving the artifactsName array empty.
- The prompt wasn't completely right, the older prompt ran into issues like failing to process the artifact since it thought it could load them once and be done. But we want it to issue the `load_artifact` function call every time it needs to access them.  I've asked about the issue to Java adk developer and they fixed it on their side as well. 

I've also added example where I insert two artifacts and you can ask the agent about it and it can describe them without issues.


tool/load_artifacts_tool.go
```diff
@@ -80,9 +80,25 @@ func (t *loadArtifactsTool) Run(ctx Context, args any) (any, error) {
...
+			artifactNames = v
+		case []any:
+			for _, item := range v {
+				strItem, ok := item.(string)
+				if !ok {
+					return nil, fmt.Errorf("element in artifact_names slice is not a string, got type %T for value %v", item, item)
+				} else {
+					artifactNames = append(artifactNames, strItem)
+				}
+			}
```
hyangah: Shouldn't we handle types other than []string/[]any?

Another approach is marshal to json&unmarshal to []string.

ngeorgy: Oh, thanks so much for that idea, more elegant and covers other cases as well.

I've also added unit tests for new paths.

Thank you!

examples/tools/loadartifacts/main.go
```diff
@@ -0,0 +1,143 @@
...
+	genai.NewPartFromBytes(imageBytes, "image/png")
+
+	_, err = artifactService.Save(ctx, &artifactservice.SaveRequest{
+		AppName:   appName,
+		UserID:    userID,
+		SessionID: session.ID().SessionID,
+		FileName:  "animal_picture.png",
+		Part:      genai.NewPartFromBytes(imageBytes, "image/png"),
+	})
+
```
dpasiukevich: inconsistent newlines, could I ask to remove?

examples/tools/loadartifacts/main.go
```diff
@@ -0,0 +1,143 @@
...
+		SessionService:  sessionService,
+		ArtifactService: artifactService,
+	})
+	if err != nil {
+		log.Fatalf("Failed to create runner: %v", err)
+	}
+
+	reader := bufio.NewReader(os.Stdin)
+
+	for {
```
dpasiukevich, line 114: Does it make sense to reuse examples.Run()? (e.g. extend to pass artifacts service to it)?

examples/tools/loadartifacts/main.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		log.Fatalf("Failed to create the session service: %v", err)
+	}
+
+	session := resp.Session
+
+	artifactService := artifactservice.Mem()
+
+	// Populate artifacts that can be described later.
+	imageResp, _ := http.Get("https://png.pngtree.com/png-vector/20250111/ourmid/pngtree-golden-retriever-dog-pictures-png-image_15147078.png")
```
dpasiukevich: If you'd like to have an image, I'd suggest to store the file locally in the examples, instead of relying on the external endpoint.

tool/load_artifacts_tool.go
```diff
@@ -80,9 +80,24 @@ func (t *loadArtifactsTool) Run(ctx Context, args any) (any, error) {
...
 		return nil, fmt.Errorf("unexpected args type, got: %T", args)
 	}
-	artifactNames, ok := m["artifact_names"].([]string)
-	if !ok {
+	var artifactNames []string
+	artifactNamesRaw, exists := m["artifact_names"]
+	if !exists {
 		artifactNames = []string{}
+	} else {
+		// In order to cast porperly from []any to []string we're gonna marshal and then
```
dpasiukevich:
```suggestion
		// In order to cast properly from []any to []string we're gonna marshal and then
```

ngeorgy: properly wasn't spelled properly haha :)

Thanks for the catch!

examples/tools/loadartifacts/main.go
ngeorgy: Indeed, I removed few of them so not it's grouped more logically.

examples/tools/loadartifacts/main.go
ngeorgy: I also wanted to show how they could load from the web but I agree, it might change in the future, I've uploaded a picture locally.

examples/tools/loadartifacts/main.go
ngeorgy: Initially that's what I wanted to do. But then I needed to extract session service as well, since I need session id to save artifacts. That would mean to extract session service for all the examples, so I thought it didn't worth the hassle. 

I would keep it rn as it is and if we'll have new examples with artifacts service, maybe it will make sense to create RunConfig for the "examples", so these fields will be optional.

PR #117, opened 2025-09-24
Update LLMAgent to remove DisplayName from media files in Gemini API.

anFatum:
- As per cl/805472895, file attachments with `SaveInputBlobsAsArtifacts=true` are always saved as artifacts and are not forwarded to model. I've removed this option as per ADK-Python, but this results that files not being save as artifacts, but could be readable by model.
- The Gemini API doesn't support file attachments with DisplayName included, so I've duplicated the logic to clean the display name from [ADK-Python](https://source.corp.google.com/piper///depot/google3/third_party/py/google/adk/models/google_llm.py;rcl=809158578;l=413) to ADK-Go as well 
- Minor change in logging middleware in restapi


internal/utils/utils.go
```diff
@@ -185,3 +193,16 @@ func hasTrailingCodeExecutionResult(resp *llm.Response) bool {
 	lastPart := resp.Content.Parts[len(resp.Content.Parts)-1]
 	return lastPart.CodeExecutionResult != nil
 }
+
+// GetGoogleLLMVariant returns the Google LLM variant to use.
+// see https://google.github.io/adk-docs/get-started/quickstart/#set-up-the-model
+func GetGoogleLLMVariant() string {
```
dpasiukevich: Let's try to avoid adding stuff to utils.go (I'd even try to get rid of this in future if possible).

What about something like `internal/llminternal/googlellm/variant.go`?

internal/utils/utils.go
```diff
@@ -185,3 +193,16 @@ func hasTrailingCodeExecutionResult(resp *llm.Response) bool {
...
+	useVertexAI, ok := os.LookupEnv("GOOGLE_GENAI_USE_VERTEXAI")
+	if !ok {
+		useVertexAI = "0"
+	}
+	if slices.Contains([]string{"1", "true"}, useVertexAI) {
+		return GoogleLLMVariantVertexAI
+	}
+	return GoogleLLMVariantGeminiAPI
```
dpasiukevich:
```suggestion
    useVertexAI := os.LookupEnv("GOOGLE_GENAI_USE_VERTEXAI")
	if slices.Contains([]string{"1", "true"}, useVertexAI) {
		return GoogleLLMVariantVertexAI
	}
	return GoogleLLMVariantGeminiAPI
```

the `if !ok {` could be skipped. At least `useVertexAI = "0"` gets me a bit confused. WDYT?

anFatum: Removed !ok, since the value will be empty string by default if not exist, so we are fine.

internal/utils/utils.go
anFatum: Moved to be `llminternal/googlellm/variant.go`

internal/llminternal/base_flow.go
```diff
@@ -222,10 +223,33 @@ func toolPreprocess(ctx agent.Context, req *llm.Request, tools []tool.Tool) erro
 			return err
 		}
 	}
-
+	preprocessRequest(req)
 	return nil
 }
 
+func preprocessRequest(req *llm.Request) {
```
dpasiukevich: WDYT if we make it as a request processor and add to the `DefaultRequestProcessors` list?

This will make it easier to track the request processors, as they'd be in one place and we'd see the order of processing.

anFatum: Yep, that's clearer, moved to be default processor, thanks!

internal/llminternal/base_flow.go
```diff
@@ -222,7 +224,30 @@ func toolPreprocess(ctx agent.Context, req *llm.Request, tools []tool.Tool) erro
 			return err
 		}
 	}
+	return nil
+}
 
+func removeDisplayNameIfExists(ctx agent.Context, req *llm.Request) error {
```
dpasiukevich: base_flow.go is pretty large file already, maybe let's better move this logic to `display_name_processor.go`?

And document the processor function itself (e.g. using lines 238-239) as the comment why the processor is needed, it's hidden now in func body (e.g. it's better to read processor comment instead of reading its whole body).

anFatum: I've moved it to `file_uploads_processor` in case we need any more processors related to file uploads, WDYT?

PR #120, opened 2025-09-24
all: update to jsonschema-go@v0.3.0, modelcontextprotocol/go-sdk@v0.7.0

hyangah:
copied from the internal cl/811016857 authored by @findleyr 

We also have to update mcp (github.com/modelcontextprotocol/go-sdk@v0.7.0)
since the newer jsonschema is not compatible with older mcp.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 
 require (
 	github.com/google/jsonschema-go v0.3.0
-	github.com/modelcontextprotocol/go-sdk v0.4.0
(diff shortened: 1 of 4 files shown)


hyangah:
In the mean time, we added mcp tool test :-) and 
@dpasiukevich @findleyr I updated the mcp module to the latest to make the test pass.
(otherwise fail with an error pointing to https://github.com/google/jsonschema-go/issues/23)

PR #121, opened 2025-09-25
feat: add the internal agent state

dpasiukevich:
This is needed to store agent "type", required for the adk web /graph endpoint.

### DIFF
--- a/examples/mcp/main.go
+++ b/examples/mcp/main.go
@@ -38,7 +38,7 @@ import (
 //   - define golang function (in this case -- GetWeather)
 //   - register it as MCP tool in the in-memory MCP server, using mcp.NewServer and mcp.Tool
 //
-// 2. GitHub MCP server:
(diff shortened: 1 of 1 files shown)


agent/agent.go
```diff
@@ -42,6 +43,9 @@ func New(cfg Config) (Agent, error) {
 		beforeAgent: cfg.BeforeAgent,
 		run:         cfg.Run,
 		afterAgent:  cfg.AfterAgent,
+		State: agentinternal.State{
+			AgentType: agentinternal.TypeCustomAgent,
```
hyangah, line 47: nit: I wonder if we can make the zero value (empty string) meaningful - to represent an agent type unknown to us (which implies custom agent).

dpasiukevich: Agent types are consumed only in rest endpoint for adk web to render agent graph. So in the endpoint implementation we will treat "" as a custom agent.

But "" is not possible right now, otherwise it's a bug. We control all agent creation, and all constructors set the type.

PR #123, opened 2025-09-29
feat: Add InstructionProviders

dpasiukevich:
Added for instruction and globalinstruction.

Logic is kept same as python.

State key is validated to be `^[a-zA-Z_][a-zA-Z0-9_]*$` (same as in TS implementation).
Python uses built-in stricter isidentifier().

If key is invalid, it's returned as is (as literal). This allows escaping e.g. "My instruction \{name\}" will be kept as is.

### DIFF
--- a/agent/llmagent/llmagent.go
+++ b/agent/llmagent/llmagent.go
@@ -149,10 +149,6 @@ type Config struct {
 	// the agent context.
 	//
 	// It takes over the Instruction field if both are set.
-	//
-	// When InstructionProvider is used, ADK will NOT inject session state
-	// placeholders into the instruction. You can use
-	// llmagent.InjectSessionState() helper for that.
 	InstructionProvider InstructionProvider
 
 	// GlobalInstruction is the instruction for all agents in the entire
@@ -179,10 +175,6 @@ type Config struct {
 	// dynamically based on the agent context.
 	//
 	// It takes over the GlobalInstruction field if both are set.
-	//
-	// When InstructionProvider is used, ADK will NOT inject session state
-	// placeholders into the instruction. You can use
-	// llmagent.InjectSessionState() helper for that.
 	GlobalInstructionProvider InstructionProvider
 
 	// LLM-based agent transfer configs.
@@ -358,4 +350,10 @@ func (a *llmAgent) maybeSaveOutputToState(event *session.Event) {
 	}
 }
 
+// InstructionProvider allows to create instructions dynamically. It is called
+// on each agent invocation.
+//
+// NOTE: when InstructionProvider is used, ADK will NOT inject session state
+// placeholders into the instruction. You can use
+// util/instructionutil.InjectSessionState() helper if this functionality is needed.
 type InstructionProvider func(ctx agent.ReadonlyContext) (string, error)


agent/llmagent/instruction.go
```diff
@@ -0,0 +1,38 @@
...
+// State key identifier should be match "^[a-zA-Z_][a-zA-Z0-9_]*$".
+// Otherwise it will be treated as a literal.
+//
+// If the state variable or artifact does not exist, the agent will raise an
+// error. If you want to ignore the error, you can append a ? to the
+// variable name as in {var?}.
+//
+// This method is intended to be used in InstructionProvider based Instruction
+// and GlobalInstruction which are called with ReadonlyContext.
+func InjectSessionState(ctx agent.Context, template string) (string, error) {
```
hyangah: Does this need to be exported or should it be a separate package? (Java has it in `utils.InstructionUtils.injectSessionState`,  and python has it in utils and actually it is used only in `llm_flows._InstructionsLlmRequestProcessor` - effectively private and applied by default) 

If we want to make it public from this package, I think it needs a better name. InjectSessionStateToInstruction, ... ?

agent/llmagent/llmagent.go
```diff
@@ -108,8 +110,59 @@ type Config struct {
 	// usage, or perform post-processing on the raw `LLMResponse`.
 	AfterModel []AfterModelCallback
 
-	Instruction       string
+	// Instruction is set for the LLM model guiding the agent's behavior.
+	//
+	// These instructions can contain placeholders like {variable_name} that
```
hyangah: Can we include a link to the public ADK doc that explains this as a reference?

agent/llmagent/llmagent.go
```diff
@@ -108,8 +110,59 @@ type Config struct {
...
+	// will be resolved at runtime using session state and context.
+	//
+	// {var} is used to insert the value of the state variable named var.
```
hyangah: Godoc rendering is pretty limited compared to markdown, but I think this paragraph will read better if "lists" syntax is used instead of one continued paragraph. https://tip.golang.org/doc/comment#lists

internal/llminternal/instruction_processor.go
```diff
@@ -37,17 +42,157 @@ func instructionsRequestProcessor(ctx agent.Context, req *llm.Request) error {
...
+	if err := appendInstructions(ctx, req, llmAgent.internal()); err != nil {
+		return fmt.Errorf("failed to append instructions: %w", err)
 	}
 
 	return nil
 }
+
+func appendInstructions(ctx agent.Context, req *llm.Request, agentState *State) error {
+	if agentState.InstructionProvider != nil {
+		_, err := agentState.InstructionProvider(ctx)
```
hyangah: nit: merge line 60 and 61 (`if _, err := ...; err != nil {}`)

internal/llminternal/instruction_processor_test.go
```diff
@@ -0,0 +1,176 @@
...
+		},
+		{
+			name:     "Empty braces are allowed",
+			template: "Hello {}, your app is {?}.",
+			want:     "Hello {}, your app is {?}.",
+		},
+		{
+			name:     "Unmatched leading brace",
+			template: "This is a {test.",
+			want:     "This is a {test.", // No match, treated as literal.
```
hyangah: Case like `{ is { name }` is an error?

session/session.go
```diff
@@ -111,3 +112,13 @@ type Actions struct {
 	// The agent is escalating to a higher level agent.
 	Escalate bool
 }
+
+type KeyPrefix string
```
hyangah: Do we need this type? This makes operation like this verbose.

```
func f(outputKey string) {
  key := string(KeyPrefixTemp)+outputKey
 ...
}
```

dpasiukevich: You're right, updated to a `string` type.

agent/llmagent/instruction.go
dpasiukevich: > Does this need to be exported or should it be a separate package?

It should be exported. This is ADK public utils func. [1]
As for the separate package I agree. WDYT if we expose public `util/instructionutil` package and put this func there? `instructionutil. InjectSessionState`.

[1]. Example of ADK python docs showing users how to use InstructionProvider + optionally `inject_session_state` https://google.github.io/adk-docs/sessions/state/#bypassing-state-injection-with-instructionprovider

agent/llmagent/llmagent.go
dpasiukevich: I've tried to make the comment as descriptive as possible while concise as possible.

I'd avoid linking public docs to avoid tight coupling, as I'd expect the links may become broken/moved, requiring go code change. As I can see both python/java do not link.

But if you strongly consider a link would be useful, I can add.

agent/llmagent/llmagent.go
dpasiukevich: The list syntax doesn't seem to work for a field comment. Only for struct/interface at the top level.

Still I've updated the comment to list those items.

internal/llminternal/instruction_processor.go
```diff
@@ -73,7 +64,51 @@ func instructionsRequestProcessor(ctx agent.InvocationContext, req *model.LLMReq
 // The regex to find placeholders like {variable} or {artifact.file_name}.
 var placeholderRegex = regexp.MustCompile(`{+[^{}]*}+`)
 
-// replaceMatch tries to retrieve the match from the state map
+func appendInstructions(ctx agent.InvocationContext, req *model.LLMRequest, agentState *State) error {
+	if agentState.InstructionProvider != nil {
+		if _, err := agentState.InstructionProvider(icontext.NewReadonlyContext(ctx)); err != nil {
```
baptmont: The result string is being ignored, should it not be appended to the request using utils.AppendInstructions?

internal/llminternal/instruction_processor.go
```diff
@@ -73,7 +64,51 @@ func instructionsRequestProcessor(ctx agent.InvocationContext, req *model.LLMReq
...
+		return fmt.Errorf("failed to inject session state into instruction: %w", err)
+	}
+
+	utils.AppendInstructions(req, inst)
+	return nil
+}
+
+func appendGlobalInstructions(ctx agent.InvocationContext, req *model.LLMRequest, agentState *State) error {
+	if agentState.GlobalInstructionProvider != nil {
+		if _, err := agentState.GlobalInstructionProvider(icontext.NewReadonlyContext(ctx)); err != nil {
```
baptmont: The result string is also ignored for the global instructions, should it not be appended to the request using utils.AppendInstructions?

baptmont:
Can you add some tests using InstructionProvider and GlobalInstructionProvider? Not sure if it was lost during the merge.

internal/llminternal/instruction_processor.go
dpasiukevich: Not sure what happened there originally 😅

Updated the code and added the tests.

I've also created `util` package to expose any ADK utilities we want to expose for users (I see there's such stuff in python/java already).
Added `instructionutil` subpackage for it.

PR #124, opened 2025-09-29
feat: adk a2a adapter

yarolegovich:
### Multi-part PR

A split if it's more convenient to review this way:

- [x] Part mapping: https://github.com/google/adk-go/pull/150
- [x] Event processing logic: https://github.com/google/adk-go/pull/151
- [x] Executor: https://github.com/google/adk-go/pull/152
- [x] CLI integration: https://github.com/google/adk-go/pull/153

### Details

The PR depends on unmerged a2a-go changes.

* Added an `adka2a` package which provides an `AgentExecutor` implementation for `a2a-go` SDK. The public interface is TBD, for now it accepts a `type ExecutorConfig runner.Config` and `runner.RunConfig`.
* Added the logic for bidirectional mapping of `genai.Part <-> a2a.Part`.
* Extended `AgentLoader` to provide a method for accessing a `Root()` agent. This will likely be useful for `restapi` as well. 
* Extended `cmd/web/server.go` to support exposing a root agent via a2a.

### Translation rules

Executor invokes an ADK agent and translates session.Events to a2a.Events according to the following rules:
- If the input doesn't reference any `Task`, produce a `TaskStatusUpdateEvent` with `TaskStateSubmitted`.
- Right before `runner.Runner` invocation, produce `TaskStatusUpdateEvent` with `TaskStateWorking`.

There's no `TaskAuthRequiredState` production rules at the moment, because A2A assumes an Agent is still executing while waiting for auth credentials. A proper integration will require more research and planning.

### Demo

Streaming and non-streaming modes:

https://github.com/user-attachments/assets/b1a12edc-c319-4887-84f5-7a66b989333f

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,6 @@ require (
 	github.com/google/jsonschema-go v0.3.0
 	github.com/modelcontextprotocol/go-sdk v0.7.0
 	golang.org/x/net v0.43.0
-	google.golang.org/grpc v1.75.0-dev
 )
 
 require (
@@ -74,5 +73,6 @@ require (
 	golang.org/x/sys v0.35.0 // indirect
 	golang.org/x/text v0.29.0 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20250929231259-57b25ae835d4 // indirect
+	google.golang.org/grpc v1.75.0-dev // indirect
 	google.golang.org/protobuf v1.36.10 // indirect
 )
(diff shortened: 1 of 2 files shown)


adka2a/executor.go
```diff
@@ -0,0 +1,177 @@
...
+	processor := &eventProcessor{task: task, reqCtx: reqCtx, meta: invocationMeta}
+	if err := e.process(ctx, r, processor, content, queue); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (e *Executor) Cancel(ctx context.Context, reqCtx a2asrv.RequestContext, queue eventqueue.Queue) error {
+	return fmt.Errorf("cancelation not supported")
```
mazas-google: Why not supported?

adka2a/agent_card.go
```diff
@@ -0,0 +1,18 @@
+package adka2a
```
mazas-google, line 15: Why not "a2a" package?

adka2a/executor.go
yarolegovich: Following the steps of Python SDK.
But should actually be as simple as writing a TaskStateCanceled event and this should eventually cancel Execute() context.

adka2a/agent_card.go
yarolegovich: to avoid conflicts if users want to use some types from `a2aproject/a2a-go/a2a`, for example create a custom `a2a.AgentCard` and use it during server setup

adka2a/executor.go
yarolegovich: added a [simple implementation](https://github.com/google/adk-go/pull/124/commits/34e4d02d2bbdd68adfa9e086615ba46dca363327). once the current in-flight PRs in a2a repo land we can an add a component tests for cancelations of in-flight tasks

yarolegovich:
Closing this PR as the main parts were merged into `main` from separated branches. CLI integration (the 4th part) has a smaller scope and will be easier to review as a separate PR.

PR #125, opened 2025-09-29
[ADK-Web] Add missing functionality in graph endpoint implementation

anFatum:


cmd/restapi/services/agentgraphgenerator.go
```diff
@@ -53,10 +60,16 @@ func nodeCaption(instance any) string {
 	switch i := instance.(type) {
 	case agent.Agent:
 		caption = "🤖 " + i.Name()
+		agentInternal, ok := i.(agentinternal.Agent)
+		if ok {
+			if slices.Contains(supportedClusterAgents, agentinternal.Reveal(agentInternal).AgentType) {
```
kdroste-google: Imho it's hard to read and differentiate betweeen agentinternal and agentInternal. Maybe agentInt instead of agentInternal or something?

anFatum: Changed variable to be `typedAgent`

PR #127, opened 2025-09-30
[CLI] Moved to static ADK Web distribution for cloudrun and local run

kdroste-google:


cmd/web/server.go
```diff
@@ -91,23 +112,39 @@ func Serve(c *WebConfig, serveConfig *ServeConfig) {
...
+			http.Redirect(w, r, "/ui/", http.StatusFound)
+		})
+
 		rUi.Methods("GET").Handler(http.StripPrefix("/ui/", http.FileServer(http.Dir(c.UIDistPath))))
 	}
 
 	if c.StartRestApi {
 		rApi := rBase.Methods("GET", "POST", "DELETE", "OPTIONS").PathPrefix("/api/").Subrouter()
-		rApi.Use(serverConfig.Cors.Handler)
+		rApi.Use(corsWithArgs(c))
```
anFatum: Why do we have that? Is there a reason we replace `serverConfig.Cors.Handler` with custom middleware? If we want to keep custom CORS function, why do we need `serverConfig.Cors.Handler`?

cmd/web/server.go
```diff
@@ -91,23 +112,39 @@ func Serve(c *WebConfig, serveConfig *ServeConfig) {
...
 		restapiweb.SetupRouter(rApi, &serverConfig)
 	}
 
 	log.Fatal(http.ListenAndServe(":"+strconv.Itoa(c.LocalPort), rBase))
+	// log.Fatal(http.ListenAndServe(":"+strconv.Itoa(c.LocalPort), serverConfig.Cors.Handler(rBase)))
```
anFatum: Let's remove unused comments

cmd/adkgo/root/run/local/local.go
```diff
@@ -16,32 +16,32 @@
...
 	"github.com/spf13/cobra"
 	"google.golang.org/adk/cmd/adkgo/root/run"
 	"google.golang.org/adk/internal/cli/util"
 )
 
 type localServerFlags struct {
-	port int    // command line param
-	url  string // autogenerated
+	port int // command line param
+	//url  string // autogenerated
```
anFatum: Let's remove if it's unused

cmd/adkgo/root/run/local/local.go
```diff
@@ -167,7 +127,7 @@ func (f *runLocalFlags) compileEntryPoint() error {
 			cmd := exec.Command("go", "build", "-o", f.build.execPath, f.source.entryPointPath)
 
 			cmd.Dir = f.source.srcBasePath
-			cmd.Env = append(os.Environ(), "CGO_ENABLED=0", "GOOS=linux")
+			// cmd.Env = append(os.Environ(), "CGO_ENABLED=0", "GOOS=linux")
```
anFatum: The same, as above - unused comment.

cmd/adkgo/root/run/local/local.go
```diff
@@ -83,79 +75,47 @@ var localCmd = &cobra.Command{
...
-	localCmd.PersistentFlags().StringVarP(&flags.source.entryPointPath, "entryPoint", "e", "", "Path to an entry point (go 'main'). ")
-	localCmd.PersistentFlags().StringVarP(&flags.source.srcBasePath, "srcPath", "p", "", "Path to an base path of your project")
+	localCmd.PersistentFlags().IntVarP(&flags.server.port, "server_port", "p", 8080, "Localhost port on which the server should be listening")
+	localCmd.PersistentFlags().StringVarP(&flags.source.webUIDistrPath, "webui_distr_path", "w", "", "Points to a static web ui dist path with the built version of ADK Web UI (cmd/web/distr/browser in the ADK-GO repo).")
+	localCmd.PersistentFlags().StringVarP(&flags.source.entryPointPath, "entry_point_path", "e", "", "Path to an entry point (go 'main'). Example relative to ADK-GO repo: './examples/web/main.go'")
+	localCmd.PersistentFlags().StringVarP(&flags.build.execPath, "target_exec_path", "t", "", "Path to the output executable. Defaults to entry_point_path with no '.go' suffix.")
 }
 
 func (f *runLocalFlags) computeFlags() error {
 	err := util.LogStartStop("Computing flags",
```
anFatum: Why don't you `return util.LogStartStop` instead of creating err variable as you did above?

cmd/web/distr/browser/assets/config/runtime-config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "backendUrl": "http://localhost:8000"
```
anFatum: both java and python have empty backendurl, why do we have that filled to localhost? 
https://github.com/google/adk-python/blob/main/src/google/adk/cli/browser/assets/config/runtime-config.json
https://github.com/google/adk-java/blob/main/dev/browser/assets/config/runtime-config.json

cmd/web/server.go
```diff
@@ -91,23 +112,39 @@ func Serve(c *WebConfig, serveConfig *ServeConfig) {
...
 		AllowCredentials: true})
 
 	rBase := mux.NewRouter().StrictSlash(true)
 	rBase.Use(Logger)
 
 	if c.StartWebUI {
+
 		rUi := rBase.Methods("GET").PathPrefix("/ui/").Subrouter()
+
+		// generate runtime-config in the runtime
```
anFatum: hm, do you thing if we remove localhost from the `runtime-config.json` we could remove that handler? I haven't found anything like that neither in python neither in java, so I'm curious why do we need to add any workaround on our side? 
if that doesn't work, let's add a small comment that this sets the backend address for `runtime-config.json`, thanks!

cmd/web/server.go
kdroste-google: fixed

cmd/web/server.go
kdroste-google: I will test it one more time, but it didn't work previously :(

cmd/adkgo/root/run/local/local.go
kdroste-google: fixed

cmd/web/server.go
anFatum: If that didn't work let's remove it, I don't think we should keep both versions

cmd/adkgo/root/run/local/local.go
kdroste-google: Right, fixed

cmd/web/server.go
kdroste-google: It seems they always use the same port :) Otherwise it should break. I didn't want to rely on this deployment-time-generated file, that's why it's needed on runtime.

kdroste-google: Commented

cmd/web/distr/browser/assets/config/runtime-config.json
kdroste-google: it's being written here
https://github.com/google/adk-web/blob/8cd5de003b67c768b26b8e4551ff4eba4e66cc77/set-backend.js#L20

kdroste-google: We don't anymore, it's runtime-generated now

cmd/web/server.go
kdroste-google: removed, thanks for spotting !

cmd/adkgo/root/run/local/local.go
kdroste-google: fixed

cmd/adkgo/root/deploy/cloudrun/cloudrun.go
```diff
@@ -164,87 +167,87 @@ func (f *deployCloudRunFlags) compileEntryPoint() error {
...
 }
 
 func (f *deployCloudRunFlags) deployOnCloudRun() error {
 	fmt.Println(flags)
-	var err error
 
-	err = f.computeFlags()
+	err := f.computeFlags()
 	if err != nil {
 		return err
```
dpasiukevich, line 244: Could you add the context to the errors?

e.g. `return fmt.Errorf("failed to compute flags: %w", err)`?

Because as the func has multiple exit points, with such error context it's easy to understand from the error the whole trace.

cmd/web/server.go
```diff
@@ -23,42 +23,47 @@ import (
...
-	startRespApi := flag.Bool("start_restapi", true, "Set to start a rest api endpoint '/api'")
-	startWebUI := flag.Bool("start_webui", true, "Set to start a web ui endpoint '/ui'")
-	webuiDist := flag.String("webui_path", "", "Points to a static web ui dist path with the built version of ADK Web UI")
+	localPortFlag := flag.Int("port", 8080, "Localhost port for the server")
+	frontendAddressFlag := flag.String("front_address", "localhost:8080", "Front address to allow CORS requests from as seen from the user browser. Please specify only hostname and (optionally) port")
+	backendAddressFlag := flag.String("backend_address", "http://localhost:8080/api", "Backend server as seen from the user browser. Please specify the whole URL, i.e. 'http://localhost:8080/api'. ")
+	startRespApiFlag := flag.Bool("start_restapi", true, "Set to start a rest api endpoint '/api'")
+	startWebUIFlag := flag.Bool("start_webui", true, "Set to start a web ui endpoint '/ui'")
+	webuiDistPathFlag := flag.String("webui_distr_path", "",
+		`Points to a static web ui dist path with the built version of ADK Web UI (cmd/web/distr/browser in the ADK-GO repo).
```
dpasiukevich:
```suggestion
		`Points to a static web UI dist path with the built version of ADK Web UI (cmd/web/distr/browser in the ADK-GO repo). 
```

cmd/web/server.go
```diff
@@ -23,42 +23,47 @@ import (
...
-	localPortFlag := flag.Int("port", 8080, "Port to listen on")
-	frontendServerFlag := flag.String("front_address", "http://localhost:8001", "Front address to allow CORS requests from")
-	startRespApi := flag.Bool("start_restapi", true, "Set to start a rest api endpoint '/api'")
-	startWebUI := flag.Bool("start_webui", true, "Set to start a web ui endpoint '/ui'")
-	webuiDist := flag.String("webui_path", "", "Points to a static web ui dist path with the built version of ADK Web UI")
+	localPortFlag := flag.Int("port", 8080, "Localhost port for the server")
+	frontendAddressFlag := flag.String("front_address", "localhost:8080", "Front address to allow CORS requests from as seen from the user browser. Please specify only hostname and (optionally) port")
+	backendAddressFlag := flag.String("backend_address", "http://localhost:8080/api", "Backend server as seen from the user browser. Please specify the whole URL, i.e. 'http://localhost:8080/api'. ")
+	startRespApiFlag := flag.Bool("start_restapi", true, "Set to start a rest api endpoint '/api'")
+	startWebUIFlag := flag.Bool("start_webui", true, "Set to start a web ui endpoint '/ui'")
```
dpasiukevich: Can we remove these for now? And start /api and /ui always? Just to keep the initial set of flags simple?

If users in future would ask for some customization with a valid request, we may add those later.

cmd/adkgo/root/deploy/cloudrun/cloudrun.go
kdroste-google: All those functions are using LogStartStop. Error is handled inside of this function and passed up just to stop the sequence. 
To make it more explicit, instead of returning an error, I can return bool == should sequence be stopped or not
wdyt?

cmd/web/server.go
kdroste-google: Fixed, will be removed in the next PR anyway

cmd/web/server.go
kdroste-google: Removed

cmd/web/server.go
```diff
@@ -83,29 +84,58 @@ type ServeConfig struct {
...
-		AllowedOrigins:   []string{c.FrontEndServer},
-		AllowedMethods:   []string{http.MethodGet, http.MethodPost, http.MethodOptions, http.MethodDelete, http.MethodPut},
-		AllowCredentials: true})
 
 	rBase := mux.NewRouter().StrictSlash(true)
 	rBase.Use(Logger)
 
-	if c.StartWebUI {
+	// Setup serving of ADK Web UI
+	{
```
dpasiukevich: nit: this scoping with { } is quite uncommon in go. Can we remove this?

kdroste-google: OK, thanks!, Fixed

PR #136, opened 2025-10-04
Remove agent.Context

dpasiukevich:
1. agent.Context is replaced with InvocationContext, ReadonlyContext and CallbackContext to have ADK conformity.
2. Moved runner.RunConfig -> agent.RunConfig.
3. All contexts are now created in the internal packages: `internal/context` and `internal/tool`. Removed agent.NewContext and tool.NewContext from public API.

### DIFF
--- a/examples/userloop.go
+++ b/examples/userloop.go
@@ -91,7 +91,7 @@ func Run(ctx context.Context, rootAgent agent.Agent, runConfig *RunConfig) {
 			} else {
 				for _, p := range event.LLMResponse.Content.Parts {
 					// if its running in streaming mode, don't print the non partial llmResponses
-					if streamingMode != runner.StreamingModeSSE || event.LLMResponse.Partial {
+					if streamingMode != agent.StreamingModeSSE || event.LLMResponse.Partial {
 						fmt.Print(p.Text)
 					}
 				}


agent/llmagent/llmagent.go
```diff
@@ -20,6 +20,7 @@ import (
 
 	"google.golang.org/adk/agent"
 	agentinternal "google.golang.org/adk/internal/agent"
+	icontext "google.golang.org/adk/internal/context"
```
hyangah, line 23: Can we establish the naming convention now and make
all internal packages whose corresponding public parts are public named
consistently (e.g. `llminternal`, `agentinternal`, `contextinternal` vs `imodel`, `iagent` or `icontext`) so we avoid all these renaming?

agent/context.go
```diff
@@ -17,81 +17,39 @@ package agent
...
-}
+	InvocationID() string
+	Branch() string
+	Agent() Agent
+	UserContent() *genai.Content
+	RunConfig() *RunConfig
 
-func (a *agentContext) Artifacts() Artifacts {
-	return a.artifacts
+	EndInvocation()
```
hyangah, line 37: Shouldn't there be `Ended` to allow to check if the invocation is end?

internal/context/invocation_context.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"context"
+
+	"github.com/google/uuid"
+	"google.golang.org/adk/agent"
+	"google.golang.org/adk/session"
+	"google.golang.org/genai"
+)
+
+type InvocationContextParams struct {
```
hyangah, line 26: (optional) if we want to keep the constructor function, storing the passed in `InvocationContextParams`  can save a fewer lines and reduce a chance of forgetting some params by mistake.

```

type InvocationContext struct {
  context.Context
  params InvocationContextParams

  invocationID string
}
```

internal/context/readonly_context.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func NewReadonlyContext(ctx agent.InvocationContext) agent.ReadonlyContext {
+	return &readonlyContext{
+		InvocationContext: ctx,
+	}
+}
+
+type readonlyContext struct {
+	agent.InvocationContext
```
hyangah: Is embedding the right choice? Shouldn't it be an internal field to prevent users later to type assert and access the underlying InvocationContext?

internal/llminternal/agent_transfer_test.go
```diff
@@ -370,13 +373,15 @@ func TestAgentTransfer_ProcessRequest(t *testing.T) {
 func TestTransferToAgentToolRun(t *testing.T) {
 	t.Run("Success", func(t *testing.T) {
 		curTool := &llminternal.TransferToAgentTool{}
-		ctx := tool.NewContext(agent.NewContext(t.Context(), nil, nil, nil, nil, nil, ""), "", &session.EventActions{})
+
+		ctx := toolinternal.NewToolContext(icontext.NewInvocationContext(t.Context(), icontext.InvocationContextParams{}), "", &session.EventActions{})
```
hyangah: How about making `incCtx` and `toolCtx` vars  to make the distinction clear?

```
invCtx := icontext.NewInvocationContext(...)
toolCtx := toolinternal.NewToolContext(invCtx, ..)

```

agent/llmagent/llmagent.go
dpasiukevich: I plan to converge all to: internal/agent, internal/context as package names. And import as `imodel` within the codebase.
I didn't do it in this PR fully to avoid making it bigger, but we will converge naming in the coming PRs.

internal/context/readonly_context.go
dpasiukevich: You are correct. Better to have the internal field. Thanks!

internal/context/invocation_context.go
dpasiukevich: SG, done

agent/context.go
dpasiukevich: Done.

internal/llminternal/agent_transfer_test.go
dpasiukevich: Done.

PR #140, opened 2025-10-06
Add Agent Tool

ngeorgy:
There are some TODOs left, I don't want to include those changes in this pr since it's already quite large. I'll add them in the following PR. Right now this works and can be used to use multiple genai tools.

### DIFF
--- a/agent/llmagent/llmagent.go
+++ b/agent/llmagent/llmagent.go
@@ -127,6 +127,7 @@ type Config struct {
 	// user messages, tool requests, etc.
 	IncludeContents string
 
+	// TODO(ngeorgy): consider to switch to jsonschema for input and output schema.
 	// The input schema when agent is used as a tool.
 	InputSchema *genai.Schema
 	// The output schema when agent replies.
--- a/examples/tools/multipletools/main.go
+++ b/examples/tools/multipletools/main.go
@@ -94,7 +94,7 @@ func main() {
 		Instruction: "Answer questions about weather based on google search unless asked for a poem," +
 			" for a poem generate it with a tool.",
 		Tools: []tool.Tool{
-			agenttool.NewDefault(searchAgent), agenttool.NewDefault(poemAgent),
+			agenttool.New(searchAgent, nil), agenttool.New(poemAgent, nil),
 		},
 	})
 	if err != nil {
--- a/internal/utils/schema_utils.go
+++ b/internal/utils/schema_utils.go
@@ -23,19 +23,6 @@ import (
 	"google.golang.org/genai"
 )
 
-// Type represents the data type in the schema.
-type Type string
-
-// Known schema types.
-const (
-	String  Type = "STRING"
-	Integer Type = "INTEGER"
-	Boolean Type = "BOOLEAN"
-	Number  Type = "NUMBER"
-	Array   Type = "ARRAY"
-	Object  Type = "OBJECT"
-)
-
 // matchType checks if the value matches the schema type.
 func matchType(value any, schema *genai.Schema, isInput bool) (bool, error) {
 	if schema == nil {
@@ -112,20 +99,20 @@ func ValidateMapOnSchema(args map[string]any, schema *genai.Schema, isInput bool
 		propSchema, exists := properties[key]
 		if !exists {
 			// Note: OpenAPI schemas can allow additional properties. This implementation assumes strictness.
-			return fmt.Errorf("%s arg: '%s' does not exist in schema properties", argType, key)
+			return fmt.Errorf("%s arg: '%q' does not exist in schema properties", argType, key)
 		}
 		ok, err := matchType(value, propSchema, isInput)
 		if err != nil {
-			return fmt.Errorf("%s arg: '%s' validation failed: %w", argType, key, err)
+			return fmt.Errorf("%s arg: '%q' validation failed: %w", argType, key, err)
 		}
 		if !ok {
-			return fmt.Errorf("%s arg: '%s' type mismatch, expected schema type %s, got value %v of type %T", argType, key, propSchema.Type, value, value)
+			return fmt.Errorf("%s arg: '%q' type mismatch, expected schema type %s, got value %v of type %T", argType, key, propSchema.Type, value, value)
 		}
 	}
 
 	for _, requiredKey := range schema.Required {
 		if _, exists := args[requiredKey]; !exists {
-			return fmt.Errorf("%s args does not contain required key: '%s'", argType, requiredKey)
+			return fmt.Errorf("%q args does not contain required key: '%q'", argType, requiredKey)
 		}
 	}
 	return nil
--- a/tool/agenttool/agent_tool.go
+++ b/tool/agenttool/agent_tool.go
@@ -15,14 +15,12 @@
 package agenttool
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
 
 	"google.golang.org/adk/agent"
 	"google.golang.org/adk/artifact"
-	agentinternal "google.golang.org/adk/internal/agent"
 	"google.golang.org/adk/internal/llminternal"
 	"google.golang.org/adk/internal/utils"
 	"google.golang.org/adk/memory"
@@ -39,17 +37,23 @@ type agentTool struct {
 	skipSummarization bool
 }
 
-// New creates a new agentTool.
-func New(agent agent.Agent, skipSummarization bool) tool.Tool {
+// New creates a new agent tool.
+// If cfg is nil, skipSummarization defaults to false.
+func New(agent agent.Agent, cfg *Config) tool.Tool {
+	if cfg == nil {
+		return &agentTool{
+			agent:             agent,
+			skipSummarization: false,
+		}
+	}
 	return &agentTool{
 		agent:             agent,
-		skipSummarization: skipSummarization,
+		skipSummarization: cfg.SkipSummarization,
 	}
 }
 
-// NewDefault creates a new agentTool with skipSummarization set to false.
-func NewDefault(agent agent.Agent) tool.Tool {
-	return New(agent, false)
+type Config struct {
+	SkipSummarization bool
 }
 
 // Name implements tool.Tool.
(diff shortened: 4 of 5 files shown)


examples/tools/multipletools/main.go
```diff
@@ -0,0 +1,105 @@
...
+		},
+	})
+	if err != nil {
+		log.Fatalf("Failed to create agent: %v", err)
+	}
+
+	agent, err := llmagent.New(llmagent.Config{
+		Name:        "root_agent",
+		Model:       model,
+		Description: "You can do a google search and generate poems.",
```
hyangah, line 93: fyi - recently, adk-python added a temporary workaround (automatically translating google_search into an agent tool automatically - see cl/814319961) Let's monitor how it progress.

agent/llmagent/llmagent.go
```diff
@@ -154,6 +157,9 @@ type llmAgent struct {
 	model       model.LLM
 	afterModel  []llminternal.AfterModelCallback
 	instruction string
+
+	inputSchema  *genai.Schema
```
hyangah, line 170: Can't we use `jsonschema.Schema` like the function tool? https://github.com/google/adk-go/blob/b5fb31f4ab0b01df4f5b05f2ae57667c0cb64c66/tool/function.go#L49-L53

The jsonschema package provides validation, inference, etc. And, having just one way to work with json schema will be simpler for users and us.

internal/utils/schema_utils.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+		err := ValidateMapOnSchema(obj, schema, isInput)
+		return err == nil, err
+	default:
+		return false, fmt.Errorf("unsupported type: %s", schema.Type)
+	}
+}
+
+// ValidateMapOnSchema validates a map against a schema.
+func ValidateMapOnSchema(args map[string]any, schema *genai.Schema, isInput bool) error {
```
hyangah, line 83: Do we need all these if we use the `jsonschema` package?

tool/agenttool/agent_tool.go
```diff
@@ -0,0 +1,282 @@
...
+	"google.golang.org/adk/session"
+	"google.golang.org/adk/tool"
+	"google.golang.org/genai"
+)
+
+// AgentTool implements a tool that allows an agent to call another agent.
+type AgentTool struct {
+	agent             agent.Agent
+	skipSummarization bool
+	name              string
```
hyangah: do we need `name/description` fields? It looks like `agent.Name()` and `agent.Description()` can be used directly.

tool/agenttool/agent_tool.go
```diff
@@ -0,0 +1,282 @@
...
+	"google.golang.org/adk/memoryservice"
+	"google.golang.org/adk/model"
+	"google.golang.org/adk/runner"
+	"google.golang.org/adk/session"
+	"google.golang.org/adk/tool"
+	"google.golang.org/genai"
+)
+
+// AgentTool implements a tool that allows an agent to call another agent.
+type AgentTool struct {
```
hyangah: Unexport it. `NewAgentTool` returns `tool.Tool` so this type doesn't seem useful.

Looks like neither `geminitool` or `mcptool` export their concrete implementation type.

tool/agenttool/agent_tool.go
```diff
@@ -0,0 +1,282 @@
...
+	agent             agent.Agent
+	skipSummarization bool
+	name              string
+	description       string
+}
+
+// NewAgentTool creates a new AgentTool.
+func NewAgentTool(agent agent.Agent, skipSummarization bool) tool.Tool {
```
hyangah: Since this is already `agenttool` package,  how about just going with `New`? 

```
func New(agent agent.Agent, opts *Options) tool.Tool

type Options struct {
  SkipSummarization bool
   // ... any extra future customization knobs
}
```

examples/tools/multipletools/main.go
ngeorgy: Thanks for the heads up! Will keep an eye on it!

agent/llmagent/llmagent.go
ngeorgy: As discussed offline, proceeding with genai.schema for now and I've added todo to do it later. Thank you for the suggestion, I think that will simplify the whole thing!

tool/agenttool/agent_tool.go
ngeorgy: Good point, removed them!

internal/utils/schema_utils.go
ngeorgy: Most likely not, since we can use built-in validation

tool/agenttool/agent_tool.go
ngeorgy: Yes, good catch, done!

tool/agenttool/agent_tool.go
ngeorgy: Indeed, I've changed it to New, thank you!

internal/llminternal/agent.go
```diff
@@ -40,6 +40,7 @@ type State struct {
 	DisallowTransferToParent bool
 	DisallowTransferToPeers  bool
 
+	InputSchema  *genai.Schema
```
hyangah, line 43: Add a TODO here to consider to switch to jsonschema

internal/utils/schema_utils.go
```diff
@@ -0,0 +1,149 @@
...
+	"encoding/json"
+	"fmt"
+	"math"
+	"reflect"
+
+	"google.golang.org/genai"
+)
+
+// Type represents the data type in the schema.
+type Type string
```
hyangah: Delete or unexport `Type`, `String`, ...

internal/utils/schema_utils.go
```diff
@@ -0,0 +1,149 @@
...
+			return fmt.Errorf("%s arg: '%s' validation failed: %w", argType, key, err)
+		}
+		if !ok {
+			return fmt.Errorf("%s arg: '%s' type mismatch, expected schema type %s, got value %v of type %T", argType, key, propSchema.Type, value, value)
+		}
+	}
+
+	for _, requiredKey := range schema.Required {
+		if _, exists := args[requiredKey]; !exists {
+			return fmt.Errorf("%s args does not contain required key: '%s'", argType, requiredKey)
```
hyangah: `%q` instead of `'%s'`

tool/agenttool/agent_tool.go
```diff
@@ -0,0 +1,280 @@
...
+// New creates a new agentTool.
+func New(agent agent.Agent, skipSummarization bool) tool.Tool {
+	return &agentTool{
+		agent:             agent,
+		skipSummarization: skipSummarization,
+	}
+}
+
+// NewDefault creates a new agentTool with skipSummarization set to false.
+func NewDefault(agent agent.Agent) tool.Tool {
```
hyangah: Make `NewDefault` to `New` and the current `New` to `NewWithConfig(agent agent.Agent, cfg *Config) tool.Tool`.

--

Or just have one `New(agent agent.Agent, cfg *Config) tool.Tool` and when cfg is nil, construct the agent tool with reasonable defaults.

```
// New creates a new agent tool. If non-nil cfg is 
func New(agent agent.Agent, cfg *Config) tool.Tool

type Config struct {
  SkipSummarization bool
}
```

tool/agenttool/agent_tool.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	var agentInputSchema *genai.Schema
+	internalAgent, ok := t.agent.(agentinternal.Agent)
+	if !ok {
+		return nil, fmt.Errorf("internal error: failed to convert to internal agent")
+	}
+	agentState := agentinternal.Reveal(internalAgent)
+	isLllmAgent := (agentState != nil && agentState.AgentType == agentinternal.TypeLLMAgent)
+	if isLllmAgent {
```
hyangah: Is it necessary to check `isLllmAgent`? 

Looks like internal/llminternal/agent_transfer.go and others still just check if `asLLMAgent` returns nil to test wheter an agent.Agent is an LLM agent. 

```
func asLLMAgent(agent agent.Agent) llminternal.Agent {
	if agent == nil {
		return nil
	}
	if llmAgent, ok := agent.(llminternal.Agent); ok {
		return llmAgent
	}
	return nil
}
```

tool/agenttool/agent_tool.go
```diff
@@ -0,0 +1,280 @@
...
+		// TODO - figure out from where can we retrieve the userId, in python it gets from invocation context
+		// in java it uses tmp-user.
+		UserID: "tmp-user",
+		State:  stateMap,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create session for sub-agent %s: %w", t.agent.Name(), err)
+	}
+
+	eventCh := r.Run(context.Background(), subSession.Session.UserID(), subSession.Session.ID(), content, &agent.RunConfig{
```
hyangah: * s/eventCh/events : It's an iterator, right? (iter.Seq2)
* pass `toolCtx`  instead of `context.Background`. And add TODO(dpasiukevich): verify agent loop termination. 

I think he was planning to clean it up as part of agent.Context.End handling, etc.

tool/agenttool/agent_tool.go
```diff
@@ -0,0 +1,280 @@
...
+			lastEvent = event
+		}
+	}
+
+	if lastEvent == nil {
+		return map[string]any{}, nil
+	}
+
+	lastContent := lastEvent.LLMResponse.Content
+	var outputText string
```
hyangah: (optional) consider `strings.Builder`  or []string + strings.Join, instead of the `+` operator.

internal/llminternal/agent.go
ngeorgy: I swear I had it and looks like I've lost it while fixing merge conflict :) Thanks for the catch, Done!

internal/utils/schema_utils.go
ngeorgy: Uh, I had it initially and forgot to remove, thanks for the catch!

internal/utils/schema_utils.go
ngeorgy: Done, thanks!

tool/agenttool/agent_tool.go
ngeorgy: Done!

tool/agenttool/agent_tool.go
ngeorgy: Thanks for the suggestion, much shorted now!

tool/agenttool/agent_tool.go
ngeorgy: Done!

tool/agenttool/agent_tool.go
ngeorgy: Thanks, I went with  []string + strings.Join

PR #141, opened 2025-10-07
Extract ProcessRequest from tools

ngeorgy:
Some implementations still append function tools a new genai tools, this will block the ability to have different function tools in one agent for vertex models (because of genai limitation).

The correct way to consolidate the tools by combining the function declarations. Therefore extracting processRequest from tools so its logic is in one place, it will make it less error prone.

### DIFF
--- a/internal/toolinternal/toolutils/toolutils.go
+++ b/internal/toolinternal/toolutils/toolutils.go
@@ -1,0 +1,68 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


ngeorgy:
@hyangah Yes, that's exactly the reason.

I've rebased but there wasn't a merge conflict. I'm afraid it's still not possible to avoid import cycle without having separate package :(

I also considered putting this method in golang/adk-go/tool/tool.go but then it will be exported publicly, not sure if that's what we want

internal/toolinternal/toolutils/toolutils.go
```diff
@@ -0,0 +1,65 @@
...
+	"google.golang.org/adk/model"
+	"google.golang.org/genai"
+)
+
+type Tool interface {
+	Name() string
+	Declaration() *genai.FunctionDeclaration
+}
+
+func ProcessRequest(ctx context.Context, req *model.LLMRequest, tool Tool) error {
```
hyangah: can you remove `ctx` from the args? `ctx` is not currently used. 

Ideally we should pass `tool.Context`, but that will create an import cycle. OTOH, we are not currently using `context.Context`. So, until we find the need, how about renaming this to `PackTool` or something like that to reflect the current functionality? And also add a documentation.

ngeorgy: Uh, don't even know why I put ctx there :) thanks!

I've changed it to PackTool and added documentation, I think it's more clear now.

Thanks for the review!

PR #146, opened 2025-10-08
Add agent OutputKey to include state in instructions

baptmont:
Add agent OutputKey, state injecting for instructions and sequentialCode sample

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -31,7 +31,6 @@ type Agent interface {
 	Description() string
 	Run(InvocationContext) iter.Seq2[*session.Event, error]
 	SubAgents() []Agent
-	OutputKey() string
 
 	internal() *agent
 }
@@ -40,7 +39,6 @@ func New(cfg Config) (Agent, error) {
 	return &agent{
 		name:        cfg.Name,
 		description: cfg.Description,
-		outputKey:   cfg.OutputKey,
 		subAgents:   cfg.SubAgents,
 		beforeAgent: cfg.BeforeAgent,
 		run:         cfg.Run,
@@ -55,7 +53,6 @@ type Config struct {
 	Name        string
 	Description string
 	SubAgents   []Agent
-	OutputKey   string
 
 	BeforeAgent []BeforeAgentCallback
 	Run         func(InvocationContext) iter.Seq2[*session.Event, error]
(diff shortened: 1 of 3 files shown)


agent/llmagent/llmagent.go
```diff
@@ -133,7 +135,7 @@ type Config struct {
 	// TODO: BeforeTool and AfterTool callbacks
 	Tools []tool.Tool
 
-	// OutputKey
+	OutputKey string
```
hyangah, line 143: While we are here, copy over the docs from python.

// OutputKey is an optional parameter to specify the key in session state for the agent output.
//
// Typical uses cases are:
//   - Extracts agent reply for ...
//   - Connects agents to ...

agent/llmagent/llmagent.go
```diff
@@ -179,5 +181,46 @@ func (a *llmAgent) run(ctx agent.InvocationContext) iter.Seq2[*session.Event, er
...
+		for _, part := range event.Content.Parts {
+			if part.Text != "" && !part.Thought {
+				sb.WriteString(part.Text)
+			}
+		}
+		result := sb.String()
+
+		// TODO: add output schema validation and unmarshalling
+		if a.OutputSchema != nil {
+			if strings.TrimSpace(result) == "" {
```
hyangah, line 223: copy the comment from python to explain why.

        # If the result from the final chunk is just whitespace or empty,
        # it means this is an empty final chunk of a stream.
        # Do not attempt to parse it as JSON.

agent/llmagent/llmagent.go
```diff
@@ -179,5 +181,46 @@ func (a *llmAgent) run(ctx agent.InvocationContext) iter.Seq2[*session.Event, er
...
+		for ev, err := range f.Run(ctx) {
+			a.maybeSaveOutputToState(ev)
+			if !yield(ev, err) {
+				return
+			}
+		}
+	}
+}
+
+// saves the model output to state if needed. skip if the event was authored by some other agent
```
hyangah: (optional) even though it's internal, it's a good idea to follow go's comment style (full sentence, starting with the function name)

// maybeSaveOutputToState saves the model output to state if needed.
func (a *llmAgent) ...

internal/llminternal/instruction_processor_test.go
```diff
@@ -0,0 +1,281 @@
...
+				if got != tc.want {
+					// Use %q to clearly show differences in strings, especially with whitespace.
+					t.Errorf("got %q, want %q", got, tc.want)
+				}
+			}
+		})
+	}
+}
+
+type fakeSession struct {
```
hyangah: What prevents us from using the in memory session service and artifact service instead of defining the fakes?

agent/agent.go
```diff
@@ -53,6 +55,7 @@ type Config struct {
 	Name        string
 	Description string
 	SubAgents   []Agent
+	OutputKey   string
```
hyangah: Q. Shouldn't this be in llmagent's config?

Java and python define output key in LlmAgent, not BaseAgent.

PR #147, opened 2025-10-09
Change agentinternal package to agent

ngeorgy:
Folder name is agent to so the package should be agent as well.


hyangah:
In other @dpasiukevich 's pr, we were discussing to rename all internal packages to be consistently named and in a way that doesn't require package renaming from import site.

PR #149, opened 2025-10-09
Add check parent is llmagent in transfer agent targets.

baptmont:
Added same logic as https://github.com/google/adk-python/blob/55aa6f669b0c0a47d630db6740a8310da87d43e9/src/google/adk/flows/llm_flows/agent_transfer.py#L131 to verify if parent is an llmagent. Transferring to sequential parent resulted in recursive loop.

### DIFF
--- a/internal/llminternal/agent_transfer.go
+++ b/internal/llminternal/agent_transfer.go
@@ -151,7 +151,7 @@ func transferTargets(agent, parent agent.Agent) []agent.Agent {
 	llmAgent := asLLMAgent(agent)
 	llmParent := asLLMAgent(parent)
 
-	if parent == nil || llmParent == nil {
+	if llmParent == nil {
 		return targets
 	}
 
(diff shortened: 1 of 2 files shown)


internal/llminternal/agent_transfer.go
```diff
@@ -149,16 +149,20 @@ func transferTargets(agent, parent agent.Agent) []agent.Agent {
 	targets := slices.Clone(agent.SubAgents())
 
 	llmAgent := asLLMAgent(agent)
+	llmParent := asLLMAgent(parent)
 
-	if !llmAgent.internal().DisallowTransferToParent && parent != nil {
+	if parent == nil || llmParent == nil {
```
hyangah: `asLLMAgent` returns `nil` if parent is nil. So, I think you can do `if llmParent == nil` here.

baptmont: Sounds good!

baptmont:
We had no tests covering these cases for the go implementation so I just added some now.

PR #151, opened 2025-10-09
feat: adka2a event processor logic (part 2)

yarolegovich:
### Details

Event processing logic required for adka2a adapter.

Parent PR: https://github.com/google/adk-go/pull/124

### DIFF
--- a/adka2a/metadata.go
+++ b/adka2a/metadata.go
@@ -51,7 +51,7 @@ func toMetaKey(key string) string {
 }
 
 func toEventMeta(meta invocationMeta, event *session.Event) (map[string]any, error) {
-	result := make(map[string]any, len(meta.eventMeta)+5)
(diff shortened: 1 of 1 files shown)


adka2a/metadata.go
```diff
@@ -16,12 +16,78 @@ package adka2a
...
+
+	return invocationMeta{userID: userID, sessionID: sessionID, eventMeta: m}
+}
+
 func toMetaKey(key string) string {
 	return "adk_" + key
 }
 
+func toEventMeta(meta invocationMeta, event *session.Event) (map[string]any, error) {
+	result := make(map[string]any, len(meta.eventMeta)+5)
```
hyangah: Where does this  `5` come from?
Does this preallocation make a huge difference in performance?

PR #152, opened 2025-10-09
feat: adka2a AgentExecutor adapter (part 3)

yarolegovich:
### Details 

Executor invokes an ADK agent and translates session.Events to a2a.Events according to the following rules:
- If the input doesn't reference any `Task`, produce a `TaskStatusUpdateEvent` with `TaskStateSubmitted`.
- Right before `runner.Runner` invocation, produce `TaskStatusUpdateEvent` with `TaskStateWorking`.

There's no `TaskAuthRequiredState` production rules at the moment, because A2A assumes an Agent is still executing while waiting for auth credentials. A proper integration will require more research and planning.

Parent PR: https://github.com/google/adk-go/pull/124

### DIFF
--- a/adka2a/executor.go
+++ b/adka2a/executor.go
@@ -28,24 +28,11 @@ import (
 )
 
 // ExecutorConfig represents mandatory Executor dependencies.
-type ExecutorConfig runner.Config
-
-// ExecutorOption is used for configuring an Executor.
-type ExecutorOption interface {
-	apply(ae *Executor)
-}
-
-type agentExecutorOptionFn func(ae *Executor)
-
-func (f agentExecutorOptionFn) apply(ae *Executor) {
-	f(ae)
-}
-
-// WithRunConfig allows to set a configuration that will be passed to runner.Runner.Run during A2A Execute invocation.
-func WithRunConfig(config *agent.RunConfig) ExecutorOption {
-	return agentExecutorOptionFn(func(ae *Executor) {
-		ae.runConfig = config
-	})
+type ExecutorConfig struct {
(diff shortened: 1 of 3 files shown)


adka2a/executor.go
```diff
@@ -0,0 +1,196 @@
...
+	apply(ae *Executor)
+}
+
+type agentExecutorOptionFn func(ae *Executor)
+
+func (f agentExecutorOptionFn) apply(ae *Executor) {
+	f(ae)
+}
+
+// WithRunConfig allows to set a configuration that will be passed to runner.Runner.Run during A2A Execute invocation.
```
hyangah: note: this pattern is different from other parts of adk where option structure was chosen instead of functional options. I don't have any strong preference, but if there is no strong reason, consider the option structure for consistency. cc @dpasiukevich 

https://google.github.io/styleguide/go/best-practices.html#function-argument-lists

adka2a/executor.go
```diff
@@ -0,0 +1,196 @@
...
+func (e *Executor) Execute(ctx context.Context, reqCtx a2asrv.RequestContext, queue eventqueue.Queue) error {
+	msg := reqCtx.Request.Message
+	if msg == nil {
+		return fmt.Errorf("message not provided")
+	}
+	content, err := toGenAIContent(msg)
+	if err != nil {
+		return fmt.Errorf("a2a message conversion failed: %w", err)
+	}
+	cong := runner.Config(*e.config)
```
hyangah: nit: `cong`? how about config or cfg...

yarolegovich: the origin of cong 🙂 
image

PR #154, opened 2025-10-11
agent: extend ReadonlyContext to match the change in python's

hyangah:
Python's ReadonlyContext/CallbackContext/ToolContext will allow to access user id, branch, etc.

In the mean time, fix the tool.Context so it doesn't allow to access to InvocationContext.

And add tests.


ngeorgy:
I'll go ahead and merge this change since it's needed for agent tool.

PR #156, opened 2025-10-14
vertex ai session service implementation

crocodilys:
This PR introduces VertexAISessionService implementation as well as a sample agent that works on top of it. ADK code leverages gRPC calls to the Agent Engine Session Service API.

I still need to add GCP Auth Github Action (go/github-dev-docs/actions?polyglot=github-com#authenticate-to-gcp-with-github-actions), because right now tests are failing during CI checks. But the ADK code could be considered as ready for a review.

At this stage the VertexAI API Client implemented in a way to open a gRPC connection each time any of client's methods called. I am going to improve this behavior in the next PR.


examples/vertexai/agent.go
```diff
@@ -30,6 +30,9 @@ import (
 
 const (
 	modelName = "gemini-2.5-flash"
+	projectID = "adk-go-samples-sandbox-390724"
```
dpasiukevich: Let's maybe use parameters instead of hardcoding properties here? for example via $ENV_VARs.

This way we'd ensure:
1. users who try the example won't try to connect to our projects 
2. we don't expose our internal env setup (project_id, etc) and we ourselves then could change those later when needed
3. keeping the example free of infra setup.

session/service.go
```diff
@@ -37,8 +37,8 @@ func InMemoryService() Service {
 }
 
 // VertexAIService returns VertextAiSessionService implementation.
-func VertexAIService(ctx context.Context, model string) (Service, error) {
-	return newVertexAiSessionService(ctx, model)
+func VertexAIService(location string, projectID string, resourceID string) (Service, error) {
```
dpasiukevich: Let's move the VertexAIService to its own package?

1. We have in-memory as part of the session package because it's a cheap default option. But for any proper implementation, let's keep those separately.
2. It would be more easy to reason, when/if vertexai implementation will add its own public entities.
3. Right now all ADK-go users will import `cloud.google.com/go/aiplatform` to their binaries, even if they don't need it (e.g. they use in-memory or other implementation).

Adding a TODO is fine deferring to another PR.

WDYT?

dpasiukevich: Right now to create VertexAIService func params are passed.

If in future we'd need to extend those, it will be a breaking change. Let's maybe introduce vertexaiservice.Config{} or something along the lines?

This could be a TODO (and it aligns nicely with moving it to a separate package).

session/vertexai_client.go
```diff
@@ -0,0 +1,182 @@
...
+
+	"google.golang.org/api/iterator"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	aiplatform "cloud.google.com/go/aiplatform/apiv1beta1"
+	aiplatformpb "cloud.google.com/go/aiplatform/apiv1beta1/aiplatformpb"
+)
+
+const (
+	EngineResourceMatcher  = "projects/%s/locations/%s/reasoningEngines/%s"
```
dpasiukevich: Should these constants be unexported?

If they need to be a part of public API, how users would use those? + then we'd need comments.

dpasiukevich: Curious why these constants have *Matcher suffix. What does it reflect?

As a suggestion, maybe `engineResourceTemplate, sessionResourceTemplate and ConnectionErrorTemplate (or ConnectionErrorMsg)`?

session/service.go
dpasiukevich: Asking to What is `resourceID`? Is it possible to have more concrete name here or the Vertex API operates with resourceID?

Could it be `reasoningEngine`?

session/vertexai_client.go
```diff
@@ -0,0 +1,182 @@
...
+	pbSession := &aiplatformpb.Session{
+		UserId: req.UserID,
+	}
+	rpcReq := &aiplatformpb.CreateSessionRequest{
+		Parent:  fmt.Sprintf(EngineResourceMatcher, c.projectID, c.location, c.resourceID),
+		Session: pbSession,
+	}
+	lro, err := rpcClient.CreateSession(ctx, rpcReq)
+	if err != nil {
+		return nil, err
```
dpasiukevich, line 61: optional nit: WDYT if we return error context here `fmt.Errorf("failed to create session: %w", err)`, and having `(s *vertexAiService) Create(` just `return nil, err` there?

I'm thinking because `vertexAiService) Create` is just a proxy for vertexClient (so it only has the logic: call vertexClient -> return err or resp).
While `vertexClient` actually does the work and it has multiple paths which return error, so putting the error context at this level is more suitable.

Here and in other methods.

session/vertexai_client.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	return &session{
+		id: id{
```
dpasiukevich, line 65: Why `events` and `state` are not populated? Is it expected or it's a future TODO?

session/vertexai_client.go
```diff
@@ -0,0 +1,182 @@
...
+		Name: sessionNameByID(req.SessionID, c),
+	})
+	if err != nil {
+		return err
+	}
+	err = lro.Wait(ctx)
+	if err != nil {
+		return err
+	}
+	return nil
```
dpasiukevich:
```suggestion
	return lro.Wait(ctx)
```

session/vertexai_test.go
```diff
@@ -0,0 +1,302 @@
...
+const (
+	ProjectID = "adk-go-samples-sandbox-390724"
+	Location  = "us-central1"
+	EngineId  = "5577659759986737152"
+	AppName   = "test-app"
+	UserID    = "test-user"
+	SessionId = "8864667638287040512"
+)
+
+func TestNewVertexSessionSerice(t *testing.T) {
```
dpasiukevich, line 34: Let's for starters just use a fake for vertexAI session service?

Because even if we fix the github actions, it will still be impossible to run the tests locally without extra setup.

And then we'd follow up with real integration sets. WDYT?

As alternative, you can add `testdata` with httprr records (to cache&store the actual vertexAI response). But I'd suggest to start with a fake for now.

session/vertexai_client.go
crocodilys: I made them exported unintentionally, fixed

crocodilys: yeah, template suffix sounds more natural

session/vertexai_client.go
crocodilys: hm, I expected that during session creation we don't want to populate events and state
should we initialize them here with some empty values?

session/service.go
crocodilys: thats make sence

crocodilys: sounds good

session/vertexai_client.go
crocodilys: My idea here was that the `vertexAiClient` is only about managing rpc connections/requests.
So having parameters checks in `vertexAiService`, or wrapping actual network error into a common one looks more natural there

session/vertexai_test.go
crocodilys: well, in the case of having a fake one there will be nothing to test, actually. almost all product code here is about rpc managing.
right now we check that no vertex api changes has been applied at least

PR #157, opened 2025-10-15
Pass RunConfig with value instead of reference

ngeorgy:
If RunConfig is nill, then it fails with panic in the run method. Since we never want nil, we might as well pass it by value, so it will be enforced by the compiler.


ngeorgy:
Thanks for the review!

PR #159, opened 2025-10-15
Universal executor

kdroste-google:


cmd/restapi/services/agentloader.go
```diff
@@ -23,15 +23,29 @@ import (
 type AgentLoader interface {
 	ListAgents() []string
 	LoadAgent(string) (agent.Agent, error)
+	RootAgent() agent.Agent
```
anFatum: I'm not sure that's how we want to have the agent loader, since each agent in agent list represents the root agent, so it's misleading we are passing the one root agent to StaticAgentLoader. 

I think it makes more sense to pass the agent name and use LoadAgent(agentName) to get the agent we need.

examples/userloop.go
anFatum, line 1: Could we remove this file in the follow up PR? I think we already use that internally, so we could have many conflicts when importing to g3

kdroste-google: Good point

cmd/restapi/services/agentloader.go
```diff
@@ -50,3 +64,16 @@ func (s *StaticAgentLoader) LoadAgent(name string) (agent.Agent, error) {
 	}
 	return agent, nil
 }
+
+// MatchingAgent returns the only agent if there's only one. Otherwise searches by name
+func (s *StaticAgentLoader) MatchingAgent(name string) (agent.Agent, error) {
```
anFatum: Why can't we reuse LoadAgent function? It has the same semantics

e.g. as we did before: 

type SingleAgentLoader struct {
  rootAgent agent.Agent
}

func NewSingleAgentLoader(rootAgent) {
  return SingleAgentLoader(rootAgent: rootAgent)
}

func(l SingleAgentLoader) ListAgents() {
 return l.rootAgent.Name()
}

func(l SingleAgentLoader) LoadAgent(agentName string) {
 // or skip agentname checking
 if (l.rootAgent.Name() != agentName) 
  return fmt.Errorf("%v is not loaded", agentName)
 return l.rootAgent
}

Potentially we could use LoadAgent(...string name) and that will help calling LoadAgent() and LoadAgent(name) to be more clear. (Just we need to check the argument length)

cmd/restapi/services/agentloader.go
kdroste-google: Added logic for choosing the only agent by default or relying on user provided agent name for many agents.

cmd/launcher/console/console.go
```diff
@@ -0,0 +1,143 @@
...
+
+	resp, err := sessionService.Create(ctx, &session.CreateRequest{
+		AppName: appName,
+		UserID:  userID,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create the session service: %v", err)
+	}
+
+	matchingAgent, err := config.AgentLoader.MatchingAgent(l.Config.rootAgentName)
```
anFatum: we could also check if agentName == "" here, and if yes list the agents we have in loader. (and pick up the first / only one / raise an error if empty)

anFatum:
I've left the comments regarding the Agent Loader. I'm fine with either leaving the MatchingAgent as it is / updating it to what I've suggested to have a clearer interface for agent loader (I'm not sure we need both Load and Matching Agent functions, but I'm also unaware of any cases we could have in the future).

cmd/restapi/services/agentloader.go
```diff
@@ -50,3 +64,19 @@ func (s *StaticAgentLoader) LoadAgent(name string) (agent.Agent, error) {
...
+	if len(s.agents) == 1 {
+		// return the only element
+		for _, v := range s.agents {
```
anFatum: Why we need the range function here?
Why cannot we `return s.agents[0], nil`?

cmd/launcher/adk/adk.go
```diff
@@ -0,0 +1,30 @@
...
+package adk
+
+import (
+	"google.golang.org/adk/agent"
+	"google.golang.org/adk/artifact"
+	"google.golang.org/adk/cmd/restapi/services"
+	"google.golang.org/adk/session"
+)
+
+type Config struct {
```
anFatum, line 25: Let's also add the comments to public instances (structs, methods) in launcher package I see some of them are missing.

cmd/restapi/services/agentloader.go
kdroste-google: Changed in https://github.com/google/adk-go/pull/159/commits/cbb07f027402c5e36b6ea3a3ade921ae6592d0ba

cmd/restapi/services/agentloader.go
kdroste-google: Changed in https://github.com/google/adk-go/pull/159/commits/cbb07f027402c5e36b6ea3a3ade921ae6592d0ba

cmd/launcher/adk/adk.go
kdroste-google: Done

PR #160, opened 2025-10-16
[ADK-Web] [ADK-Core] Add ArtifactDelta to event actions

anFatum:
ADK-Web previously couldn't recognize the artifacts and their versions. It appears that's because the UI part relies on the even action returned in the event, that identify that the artifact version changes. (e.g. https://source.corp.google.com/piper///depot/google3/third_party/javascript/adk_web/src/app/components/chat/chat.component.ts;rcl=819459180;l=612)

I've added the ArtifactDelta map to the EventActions and make the Save() function return the artifact version as well. This fixed the artifact implementation in ADK-Web. Now it can correctly manage the artifact versions and/or multiple files saved.

https://screenshot.googleplex.com/AEzBiCeGuntNBQK
https://screenshot.googleplex.com/Au6xEUFZ2rzfRL3
https://screenshot.googleplex.com/uGvWEk6GdNMGbiY

### DIFF
--- a/examples/web/agents/image_generator.go
+++ b/examples/web/agents/image_generator.go
@@ -22,7 +22,7 @@ import (
 
 	"google.golang.org/adk/agent"
 	"google.golang.org/adk/agent/llmagent"
-	"google.golang.org/adk/model/gemini"
+	"google.golang.org/adk/model"
 	"google.golang.org/adk/tool"
 	"google.golang.org/adk/tool/functiontool"
 	"google.golang.org/adk/tool/loadartifactstool"
@@ -75,12 +75,7 @@ type generateImageResult struct {
 	Status   string `json:"Status"`
 }
 
-func GetImageGeneratorAgent(ctx context.Context, apiKey string) agent.Agent {
-	model, err := gemini.NewModel(ctx, "gemini-2.0-flash-001", nil)
-	if err != nil {
-		log.Fatalf("Failed to create model: %v", err)
-	}
-
+func GetImageGeneratorAgent(ctx context.Context, model model.LLM) agent.Agent {
 	generateImageTool, err := functiontool.New(
 		functiontool.Config{
 			Name:        "generate_image",
--- a/internal/toolinternal/context.go
+++ b/internal/toolinternal/context.go
@@ -44,7 +44,7 @@ func (ia *internalArtifacts) Save(ctx context.Context, name string, data *genai.
 		// TODO: RWLock, check the version stored is newer in case multiple tools save the same file.
 		ia.eventActions.ArtifactDelta[name] = resp.Version
 	}
-	return resp, err
+	return resp, nil
 }
 
 func NewToolContext(ctx agent.InvocationContext, functionCallID string, actions *session.EventActions) tool.Context {
@@ -70,7 +70,7 @@ type toolContext struct {
 	invocationContext agent.InvocationContext
 	functionCallID    string
 	eventActions      *session.EventActions
-	artifacts         agent.Artifacts
+	artifacts         *internalArtifacts
 }
 
 func (c *toolContext) Artifacts() agent.Artifacts {
--- a/examples/web/agents/llmauditor.go
+++ b/examples/web/agents/llmauditor.go
@@ -22,7 +22,6 @@ import (
 	"google.golang.org/adk/agent/llmagent"
 	"google.golang.org/adk/agent/workflowagents/sequentialagent"
 	"google.golang.org/adk/model"
-	"google.golang.org/adk/model/gemini"
 	"google.golang.org/genai"
 )
 
@@ -218,14 +217,7 @@ func afterReviser(ctx agent.CallbackContext, llmResponse *model.LLMResponse, llm
 	return llmResponse, nil
 }
 
-func GetLLmAuditorAgent(ctx context.Context, apiKey string) agent.Agent {
-	model, err := gemini.NewModel(ctx, "gemini-2.5-flash", &genai.ClientConfig{
-		APIKey: apiKey,
-	})
-	if err != nil {
-		panic(err)
-	}
-
+func GetLLmAuditorAgent(ctx context.Context, model model.LLM) agent.Agent {
 	criticAgent, err := llmagent.New(llmagent.Config{
 		Model:       model,
 		Name:        "critic_agent",
--- a/examples/web/main.go
+++ b/examples/web/main.go
@@ -72,8 +72,8 @@ func main() {
 	if err != nil {
 		log.Fatalf("Failed to create agent: %v", err)
 	}
-	llmAuditor := agents.GetLLmAuditorAgent(ctx, apiKey)
-	imageGeneratorAgent := agents.GetImageGeneratorAgent(ctx, apiKey)
+	llmAuditor := agents.GetLLmAuditorAgent(ctx, model)
+	imageGeneratorAgent := agents.GetImageGeneratorAgent(ctx, model)
 
 	agentLoader := services.NewStaticAgentLoader(
 		rootAgent,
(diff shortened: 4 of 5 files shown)


cmd/restapi/models/event.go
```diff
@@ -22,6 +22,11 @@ import (
 	"google.golang.org/genai"
 )
 
+type EventActions struct {
```
kdroste-google, line 26: Please describe the meaning of the struct and its fields

examples/web/agents/imagegenerator.go
```diff
@@ -0,0 +1,110 @@
...
+		"imagen-3.0-generate-002",
+		input.Prompt,
+		&genai.GenerateImagesConfig{NumberOfImages: 1})
+	if err != nil {
+		return generateImageResult{
+			Status: "fail",
+		}
+	}
+
+	resp, err := ctx.Artifacts().Save(input.Filename, *genai.NewPartFromBytes(response.GeneratedImages[0].Image.ImageBytes, "image/png"))
```
kdroste-google: Remark (not to this PR): 
Looking at artifact / inmemory I'm not sure if Save operation really returns versions incremented by 1. 
Other implementations can do anything. 
Question:  Is there a explicit requirement for artifact service to Save to unique versions? Can versions be overwritten?
If not, it can affect "delta"

examples/web/agents/image_generator.go
```diff
@@ -0,0 +1,110 @@
...
+	Filename string `json:"filename"`
+}
+
+type generateImageResult struct {
+	Filename string `json:"filename"`
+	Status   string `json:"Status"`
+}
+
+func GetImageGeneratorAgent(ctx context.Context, apiKey string) agent.Agent {
+	model, err := gemini.NewModel(ctx, "gemini-2.0-flash-001", nil)
```
kdroste-google: Sometimes we use gemini-2.5-flash, gemini-2.0-flash or gemini-2.0-flash-001
Is that intended?

examples/web/agents/imagegenerator.go
```diff
@@ -0,0 +1,110 @@
...
+		return generateImageResult{
+			Status: "fail",
+		}
+	}
+
+	if ctx.Actions().ArtifactDelta == nil {
+		ctx.Actions().ArtifactDelta = make(map[string]int64)
+	}
+
+	ctx.Actions().ArtifactDelta[input.Filename] = resp.Version
```
kdroste-google: If the fragment above (ArtifactDelta logic) is mandatory, it would make sense to move it to ADK

anFatum: I'd say so, but our implementation is currently lacking the access to EventActions during the artifact Save() in opposite to what Python does - https://github.com/google/adk-python/blob/main/contributing/samples/generate_image/agent.py#L35 and https://github.com/google/adk-python/blob/main/src/google/adk/agents/callback_context.py#L89

internal/artifact/artifacts.go
ngeorgy, line 1: What if we add eventActions to

// Artifacts implements Artifacts
type Artifacts struct {
	Service   artifact.Service
	AppName   string
	UserID    string
	SessionID string
}

Then we can have function like:

SetEventActions() - which will be called in NewToolContext()

And then we can update artifact_delta in artifacts implementation like python and java does.

This way we will be more similar to other languages and will be less likely to diverge from them.

WDYT?

anFatum: The problem is that the artifact_delta is updated not in the artifacts implementation but in the callback context, since event actions are part of the callback context. So we either can follow Python / Java implementation to have a SaveArtifact() method in callback context (e.g. https://github.com/google/adk-java/blob/main/core/src/main/java/com/google/adk/agents/CallbackContext.java#L106) or change the Save() method definition so it consumes context/event actions to update 

e.g. either ctx.SaveArtifact() or ctx.Artifacts().Save(..., ctx.Actions())

WDYT?

ngeorgy: I might be missing something but.

The only way to create CallbackContext is to call NewCallbackContext right.

Then in newCallbackContext we have ctx InvocationContext, here we can set event actions to the artifacts if it's present.

Something like

if ctx.Artifacts() != nil
   ctx.Artifacts().SetEventActions(eventActions)
   
  Then artifacts implementation will hold the reference to the event actions. 
  
  Therefore there won't bee need to pass actions to Save method, it will be already present once called. Artifacts implementation is internal so it won't be created in another way, we can be sure that eventActions will be present there every time someone create ToolContext/CallbackContext. 
  
We introduced Artifacts interface in the first place to extract the artifacts related methods from callbackContext right, so effectively our Artifacts().Save() corresponds to the callbackContext.SaveArtifact() in java/python.

examples/web/agents/image_generator.go
anFatum: I've copied these from Python samples implementation where they use a different models, are there any strong objections we should do that differently?

internal/artifact/artifacts.go
anFatum: Okay, I got it know, I moved that to internal implementation. The only thing though, is that tool context doesn't use the callback context event actions, so I need to propagate that both in tool context and in callback context, which potentially can override other event actions. 

We probably should share it between contexts, do you know why we don't use callback event actions in tool context?

anFatum: I think this could lead to the race condition, since we share the same Artifacts instance between contexts, if there are multiple tool contexts saving the artifact, we might get incorrect eventActions populated

e.g. toolA -> setEventActions(artifacts) and toolB -> setEventActions(artifacts) will run at the same time

ngeorgy: Yes that's a great point.

We might need to initialize artifacts in CallbackContext to have unique event actions

examples/web/agents/image_generator.go
kdroste-google: No, there are none, I'm not sure about the expected lifespan of those models. We should have integration tests in place which will fail in that case

internal/toolinternal/context.go
```diff
@@ -19,22 +19,49 @@ import (
...
+		return resp, err
+	}
+	if ia.eventActions != nil {
+		if ia.eventActions.ArtifactDelta == nil {
+			ia.eventActions.ArtifactDelta = make(map[string]int64)
+		}
+		// TODO: RWLock, check the version stored is newer in case multiple tools save the same file.
+		ia.eventActions.ArtifactDelta[name] = resp.Version
+	}
+	return resp, err
```
dpasiukevich:
```suggestion
	return resp, nil
```

internal/toolinternal/context.go
```diff
@@ -43,6 +70,11 @@ type toolContext struct {
 	invocationContext agent.InvocationContext
 	functionCallID    string
 	eventActions      *session.EventActions
+	artifacts         agent.Artifacts
```
dpasiukevich:
```suggestion
	artifacts         internalArtifacts
```

In case we'd need to access `internalArtifacts`'s fields in toolContext method.

examples/web/agents/image_generator.go
dpasiukevich: Is suggest we use latest models, e.g. gemini-2.5-flash if that's possible.

For one reason mininum -- examples would be relevant some more time. I expect python uses gemini-2.0 simply because there was no 2.5 at that time and the examples weren't updated.

anFatum: I've changed all the examples in web use the same model, agents are working fine

examples/web/agents/imagegenerator.go
anFatum: The python states that after each successful save, the artifact revision should be increased by 1 - https://source.corp.google.com/piper///depot/google3/third_party/py/google/adk/artifacts/base_artifact_service.py;rcl=822995107;l=74

The same states the https://source.corp.google.com/piper///depot/google3/third_party/py/google/adk/artifacts/base_artifact_service.py;rcl=822995107;l=74

We should probably think about the mutex guard on the event actions in case we support parallel tools processing, but while we don't support this, I've added a todo

PR #161, opened 2025-10-19
Implement Before and After ToolCallbacks

ngeorgy:
Java ref: https://github.com/google/adk-java/blob/391e0493317c2d875400e751c5043eec3d4ef031/core/src/main/java/com/google/adk/flows/llmflows/Functions.java#L422
Python ref: https://github.com/google/adk-python/blob/d327538724f505480c9f764c61d59eee0da92b5c/src/google/adk/flows/llm_flows/functions.py#L319

### DIFF
--- a/agent/llmagent/llmagent.go
+++ b/agent/llmagent/llmagent.go
@@ -148,9 +148,27 @@ type Config struct {
 	// such as function tools, RAGs, agent transfer, etc.
 	OutputSchema *genai.Schema
 
+	// Callbacks are executed in the order they are provided.
+	// The execution of the callback chain stops at the first callback that returns a non-nil
+	// response
+	//   - If a callback returns (map[string]any, nil), it will be used directly as the result of
+	//     the tool call. Subsequent BeforeToolCallbacks in the list are skipped.
+	//   - If a callback returns (nil, error), the tool execution is aborted, and the returned
+	//     error is propagated. Subsequent BeforeToolCallbacks are skipped.
+	//   - If a callback returns (nil, nil), the execution continues to the next BeforeToolCallback
+	//     in the sequence.
 	BeforeToolCallbacks []BeforeToolCallback
 	Tools               []tool.Tool
-	AfterToolCallbacks  []AfterToolCallback
+	// Callbacks are executed in the order they are provided.
+	// The execution of the callback chain stops at the first callback that returns a non-nil
+	// response.
+	//   - If a callback returns (map[string]any, nil), it will replace the result returned by
+	//     the tool's Run method. Subsequent AfterToolCallbacks in the list are skipped.
+	//   - If a callback returns (nil, error), this error indicates a failure within the
+	//     callback itself, and will be propagated. Subsequent AfterToolCallbacks are skipped.
+	//   - If a callback returns (nil, nil), the execution continues to the next AfterToolCallback
+	//     in the sequence.
+	AfterToolCallbacks []AfterToolCallback
 	// Toolsets will be used by llmagent to extract tools and pass to the
 	// underlying LLM.
 	Toolsets []tool.Toolset
@@ -172,16 +190,6 @@ type AfterModelCallback func(ctx agent.CallbackContext, llmResponse *model.LLMRe
 
 // BeforeToolCallback is a function type executed before a tool's Run method is invoked.
 //
-// Callbacks are executed in the order they are provided.
-// The execution of the callback chain stops at the first callback that returns a non-nil
-// response
-//   - If a callback returns (map[string]any, nil), it will be used directly as the result of
-//     the tool call. Subsequent BeforeToolCallbacks in the list are skipped.
-//   - If a callback returns (nil, error), the tool execution is aborted, and the returned
-//     error is propagated. Subsequent BeforeToolCallbacks are skipped.
-//   - If a callback returns (nil, nil), the execution continues to the next BeforeToolCallback
-//     in the sequence.
-//
 // Parameters:
 //   - ctx: The tool.Context for the current tool execution.
 //   - tool: The tool.Tool instance that is about to be executed.
@@ -191,23 +199,13 @@ type BeforeToolCallback func(ctx tool.Context, tool tool.Tool, args map[string]a
 // AfterToolCallback is a function type executed after a tool's Run method has completed,
 // regardless of whether the tool returned a result or an error.
 //
-// Callbacks are executed in the order they are provided.
-// The execution of the callback chain stops at the first callback that returns a non-nil
-// response.
-//   - If a callback returns (map[string]any, nil), it will replace the result returned by
-//     the tool's Run method. Subsequent AfterToolCallbacks in the list are skipped.
-//   - If a callback returns (nil, error), this error indicates a failure within the
-//     callback itself, and will be propagated. Subsequent AfterToolCallbacks are skipped.
-//   - If a callback returns (nil, nil), the execution continues to the next AfterToolCallback
-//     in the sequence.
-//
 // Parameters:
 //   - ctx:    The tool.Context for the tool execution.
 //   - tool:   The tool.Tool instance that was executed.
 //   - args:   The arguments originally passed to the tool.
 //   - result: The result returned by the tool's Run method.
 //   - err:    The error returned by the tool's Run method.
-type AfterToolCallback func(ctx tool.Context, tool tool.Tool, args map[string]any, result any, err error) (map[string]any, error)
+type AfterToolCallback func(ctx tool.Context, tool tool.Tool, args map[string]any, result map[string]any, err error) (map[string]any, error)
 
 type llmAgent struct {
 	agent.Agent
--- a/agent/llmagent/llmagent_test.go
+++ b/agent/llmagent/llmagent_test.go
@@ -400,10 +400,10 @@ func TestToolCallback(t *testing.T) {
 			DisallowTransferToPeers:  true,
 			Tools:                    []tool.Tool{rand},
 			AfterToolCallbacks: []llmagent.AfterToolCallback{
-				func(ctx tool.Context, tool tool.Tool, args map[string]any, result any, err error) (map[string]any, error) {
+				func(ctx tool.Context, tool tool.Tool, args map[string]any, result map[string]any, err error) (map[string]any, error) {
 					return nil, nil
 				},
-				func(ctx tool.Context, tool tool.Tool, args map[string]any, result any, err error) (map[string]any, error) {
+				func(ctx tool.Context, tool tool.Tool, args map[string]any, result map[string]any, err error) (map[string]any, error) {
 					return map[string]any{"number": "7"}, nil
 				},
 			},
@@ -435,10 +435,10 @@ func TestToolCallback(t *testing.T) {
 			Tools:                    []tool.Tool{rand},
 			AfterToolCallbacks: []llmagent.AfterToolCallback{
 				// Since it retursn non nil, the next callback won't be executed.
-				func(ctx tool.Context, tool tool.Tool, args map[string]any, result any, err error) (map[string]any, error) {
+				func(ctx tool.Context, tool tool.Tool, args map[string]any, result map[string]any, err error) (map[string]any, error) {
 					return map[string]any{"number": "3"}, nil
 				},
-				func(ctx tool.Context, tool tool.Tool, args map[string]any, result any, err error) (map[string]any, error) {
+				func(ctx tool.Context, tool tool.Tool, args map[string]any, result map[string]any, err error) (map[string]any, error) {
 					return map[string]any{"number": "7"}, nil
 				},
 			},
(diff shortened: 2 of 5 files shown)


agent/llmagent/llmagent.go
```diff
@@ -148,15 +161,16 @@ type Config struct {
...
-
-	// BeforeToolCallback
-	// AfterToolCallback
 }
 
 type BeforeModelCallback func(ctx agent.CallbackContext, llmRequest *model.LLMRequest) (*model.LLMResponse, error)
 
 type AfterModelCallback func(ctx agent.CallbackContext, llmResponse *model.LLMResponse, llmResponseError error) (*model.LLMResponse, error)
 
+type BeforeToolCallback func(ctx tool.Context, tool tool.Tool, args map[string]any) (map[string]any, error)
```
dpasiukevich, line 197: Could you add a documentation for BeforeToolCallback and AfterToolCallback, describing the parameters + using https://google.github.io/adk-docs/callbacks/types-of-callbacks/#tool-execution-callbacks

Plus desribe the behavior how the list of callbacks is handled (e.g. we return first callback returning either non-nil resp or err).
So that users understand what to expect.

agent/llmagent/llmagent.go
```diff
@@ -136,8 +148,9 @@ type Config struct {
 	// such as function tools, RAGs, agent transfer, etc.
 	OutputSchema *genai.Schema
 
-	// TODO: BeforeTool and AfterTool callbacks
-	Tools []tool.Tool
+	BeforeTool []BeforeToolCallback
```
dpasiukevich: Could I ask to update the name to `BeforeToolCallbacks` and `AfterToolCallbacks`? This was asked by the doc reviewers and I do plan to update it for agent/model callbacks in the other PR.

internal/llminternal/base_flow.go
```diff
@@ -37,13 +37,19 @@ type BeforeModelCallback func(ctx agent.CallbackContext, llmRequest *model.LLMRe
...
+
 type Flow struct {
 	Model model.LLM
 
 	RequestProcessors    []func(ctx agent.InvocationContext, req *model.LLMRequest) error
 	ResponseProcessors   []func(ctx agent.InvocationContext, req *model.LLMRequest, resp *model.LLMResponse) error
 	BeforeModelCallbacks []BeforeModelCallback
 	AfterModelCallbacks  []AfterModelCallback
+	BeforeToolCallback   []BeforeToolCallback
+	AfterToolCallback    []AfterToolCallback
```
dpasiukevich:
```suggestion
	BeforeToolCallbacks   []BeforeToolCallback
	AfterToolCallbacks    []AfterToolCallback
```

internal/llminternal/base_flow.go
```diff
@@ -360,15 +366,30 @@ func handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[string]tool.
...
-		// TODO: agent.canonical_before_tool_callbacks
 		spans := telemetry.StartTrace(ctx, "execute_tool "+fnCall.Name)
-		result, err := funcTool.Run(toolCtx, fnCall.Args)
-		// genai.FunctionResponse expects to use "output" key to specify function output
-		// and "error" key to specify error details (if any). If "output" and "error" keys
-		// are not specified, then whole "response" is treated as function output.
-		// TODO(hakim): revisit the tool's function signature to handle error from user function better.
+
+		// If the result is present, it will be used instead of calling the actual tool.
+		result, err := f.invokeBeforeToolCallback(ctx, curTool, fnCall.Args, toolCtx)
```
dpasiukevich:
```suggestion
		result, err := f.invokeBeforeToolCallbacks(ctx, curTool, fnCall.Args, toolCtx)
```

same for `invokeAfterToolCallbacks`

internal/llminternal/base_flow.go
```diff
@@ -360,15 +366,30 @@ func handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[string]tool.
...
 		if err != nil {
-			result = map[string]any{"error": fmt.Errorf("tool %q failed: %w", curTool.Name(), err)}
+			result = map[string]any{"error": fmt.Errorf("BeforeToolCallback failed: %w", err)}
+		} else {
```
dpasiukevich: Could you check if it would be better to move calling callbacks and the tool to a separate func?

sort of

```
func f.callTool(...) {
  result, err := f.invokeBeforeToolCallback 
  if err != nil {
    return nil, err
  }
  if result != nil {
    return result, nil
  }

  funcTool.Run(...)

   // calling afterTool callback
}
```

And outside of this func after
```
result := f.callTool(...)
m, ok := result.(map[string]any)
// creating the event
```

This should allow to unnest and simplify the flow condition, I expect.

internal/llminternal/base_flow.go
```diff
@@ -409,6 +430,48 @@ func handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[string]tool.
 	return mergedEvent, nil
 }
 
+func (f *Flow) invokeBeforeToolCallback(ctx agent.InvocationContext, tool tool.Tool, fArgs map[string]any, toolCtx tool.Context) (any, error) {
+	llmAgent, ok := ctx.Agent().(Agent)
+	if !ok || llmAgent == nil {
+		return nil, nil
+	}
```
dpasiukevich: Do we need this?

If this Flow is called, it's the llmagent already.

internal/llminternal/base_flow.go
```diff
@@ -409,6 +430,48 @@ func handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[string]tool.
...
+		}
+	}
+	return nil, nil
+}
+
+func (f *Flow) invokeAfterToolCallback(ctx agent.InvocationContext, tool tool.Tool, fArgs map[string]any, toolCtx tool.Context, fResult any, fErr error) (any, error) {
+	llmAgent, ok := ctx.Agent().(Agent)
+	if !ok || llmAgent == nil {
+		return nil, nil
+	}
```
dpasiukevich: Same -- do we need this?

agent/llmagent/llmagent.go
ngeorgy: Great point, it was confusing to me initially as well :)

agent/llmagent/llmagent.go
ngeorgy: Done

internal/llminternal/base_flow.go
ngeorgy: Done

internal/llminternal/base_flow.go
ngeorgy: Yes, I think that's much better

internal/llminternal/base_flow.go
ngeorgy: Done

internal/llminternal/base_flow.go
ngeorgy: That's true yes, I was just following the other languages where everything is public, so it makes sense there :) 

Done!

internal/llminternal/base_flow.go
ngeorgy: Done

internal/llminternal/base_flow.go
```diff
@@ -406,6 +405,67 @@ func handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[string]tool.
...
+	// If the result is present, it will replace the result returned by the tool's Run method.
+	if afterToolCallbackResult != nil {
+		return afterToolCallbackResult
+	}
+	return result
+}
+
+func (f *Flow) invokeBeforeToolCallbacks(tool tool.Tool, fArgs map[string]any, toolCtx tool.Context) (any, error) {
+	var result map[string]any
+	var err error
```
dpasiukevich: nit: those 2 declarations are not needed. You could just do `result, err := ..` in line 439

internal/llminternal/base_flow.go
```diff
@@ -406,6 +405,67 @@ func handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[string]tool.
...
+	for _, callback := range f.BeforeToolCallbacks {
+		result, err = callback(toolCtx, tool, fArgs)
+		if err != nil {
+			return nil, fmt.Errorf("failed to execute callback: %w", err)
+		}
+		// When a list of callbacks is provided, the callbacks will be called in the
+		// order they are listed while a callback returns nil.
+		if result != nil {
+			return result, err
```
dpasiukevich: nit: maybe `return result, nil` for clarity that this is a happy path and there's no error?

internal/llminternal/base_flow.go
```diff
@@ -406,6 +405,67 @@ func handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[string]tool.
...
+	var err error
+	for _, callback := range f.AfterToolCallbacks {
+		result, err = callback(toolCtx, tool, fArgs, fResult, fErr)
+		if err != nil {
+			return nil, fmt.Errorf("failed to execute callback: %w", err)
+		}
+		// When a list of callbacks is provided, the callbacks will be called in the
+		// order they are listed while a callback returns nil.
+		if result != nil {
+			return result, err
```
dpasiukevich: same 2 nits as in the invokeBeforeToolCallbacks

internal/llminternal/base_flow.go
```diff
@@ -406,6 +405,67 @@ func handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[string]tool.
...
 }
 
+func (f *Flow) callTool(tool tool.Tool, fArgs map[string]any, toolCtx tool.Context) any {
+	// If the result is present, it will be used instead of calling the actual tool.
+	result, err := f.invokeBeforeToolCallbacks(tool, fArgs, toolCtx)
+	if err != nil {
+		return map[string]any{"error": fmt.Errorf("BeforeToolCallback failed: %w", err)}
+	}
+	if result == nil {
+		result, err = tool.(toolinternal.FunctionTool).Run(toolCtx, fArgs)
```
dpasiukevich: I see the tool is already cast to `toolinternal.FunctionTool` in line 360 in `handleFunctionCalls`.

WDYT if we'd update callTool to accept `toolinternal.FunctionTool`:

`callTool(tool toolinternal.FunctionTool, fArgs map[string]any, toolCtx tool.Context)`?

So that we don't have to type cast here (and rely that type cast check is done externally, outside of this func -- e.g. if this changes, this code may be broken).

This may require updating `toolinternal.FunctionTool` to embed `tool.Tool`, which should be no-op, as FunctionTool already implements tool.Tool

agent/llmagent/llmagent_test.go
```diff
@@ -303,6 +303,199 @@ func TestModelCallbacks(t *testing.T) {
...
+		}
+
+		runner := testutil.NewTestAgentRunner(t, agent)
+		stream := runner.Run(t, "session1", "Generate random number with 5 as a seed.")
+
+		ans, err := testutil.CollectTextParts(stream)
+		if err != nil || len(ans) == 0 {
+			t.Fatalf("agent returned (%v, %v), want result", ans, err)
+		}
+		if got, want := strings.TrimSpace(ans[len(ans)-1]), "7"; got != want {
```
dpasiukevich, line 493: Hmm, why it's not 3 (result of the before_tool_callback)? I assume if before_tool_callback is called its result is returned. And both tool and after_tool_callback are NOT called.

agent/llmagent/llmagent_test.go
```diff
@@ -303,6 +303,199 @@ func TestModelCallbacks(t *testing.T) {
...
+		ans, err := testutil.CollectTextParts(stream)
+		if err != nil || len(ans) == 0 {
+			t.Fatalf("agent returned (%v, %v), want result", ans, err)
+		}
+		if got, want := strings.TrimSpace(ans[len(ans)-1]), "3"; got != want {
+			t.Errorf("unexpected result from agent = (%v, %v), want ([%q], nil)", ans, err, want)
+		}
+	})
+
+	t.Run("after_callback_returned_when_used_with_before_callback", func(t *testing.T) {
```
dpasiukevich, line 462: Could I also ask to add a test case with no-op callbacks: 1 before + 1 after, both return nil, nil -- so the tool result is returned.

internal/llminternal/base_flow.go
ngeorgy: Indeed, initially I was returning result outside of the for loop so that's why I needed to declare it outside, but definitely that's the way!

internal/llminternal/base_flow.go
ngeorgy: Yes, thanks for catching it!

internal/llminternal/base_flow.go
ngeorgy: Yeap, that's much better, I've added tool.Tool in toolinternal.FunctionTool, works like a charm

internal/llminternal/base_flow.go
ngeorgy: Done!

agent/llmagent/llmagent_test.go
ngeorgy: From java and python looks like they're doing same, as discussed offline, let's double check it, keeping it for now. Thanks!

agent/llmagent/llmagent_test.go
ngeorgy: Added, thank you!

PR #165, opened 2025-10-21
feat: adka2a event mapper code

yarolegovich:
### Details

* Add `adka2a` package `a2a.Event -> session.Event` converters for `remoteagent` package to use. (#167)
* Move `adka2a.toMapStructure` to `internal/utils/ToMapStructure`. The utility is used by `adka2a` and `remoteagent` for converting structs to metadata `map[string]any` values.
* Rename and export utility functions for creating metadata keys. Python adds `adk_` prefix for keys it stores in a2a metadata, and `a2a:` prefix for keys it stores in `session.Event.CustomMetadata`, I'm doing the same.

### DIFF
--- a/adka2a/events.go
+++ b/adka2a/events.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/a2aproject/a2a-go/a2a"
 	"google.golang.org/adk/agent"
-	"google.golang.org/adk/model"
 	"google.golang.org/adk/session"
 	"google.golang.org/genai"
 )
@@ -34,7 +33,6 @@ func NewRemoteAgentEvent(ctx agent.InvocationContext) *session.Event {
 	event := session.NewEvent(ctx.InvocationID())
 	event.Author = ctx.Agent().Name()
 	event.Branch = ctx.Branch()
-	event.LLMResponse = &model.LLMResponse{}
 	return event
 }
 
@@ -74,7 +72,7 @@ func ToSessionEvent(ctx agent.InvocationContext, event a2a.Event) (*session.Even
 		}
 		event, err := artifactToEvent(ctx, v.Artifact)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("artifact update event conversion failed: %w", err)
 		}
 		event.LongRunningToolIDs = getLongRunningToolIDs(v.Artifact.Parts, event.Content.Parts)
 		event.CustomMetadata = ToCustomMetadata(v.TaskID, v.ContextID)
@@ -90,7 +88,7 @@ func ToSessionEvent(ctx agent.InvocationContext, event a2a.Event) (*session.Even
 		event, err := messageToEvent(ctx, v.Status.Message)
 		event.CustomMetadata = ToCustomMetadata(v.TaskID, v.ContextID)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("custom metadata conversion failed: %w", err)
 		}
 		if len(event.Content.Parts) == 0 {
 			return nil, nil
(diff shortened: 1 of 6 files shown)


adka2a/events.go
```diff
@@ -0,0 +1,259 @@
...
+	case *a2a.Message:
+		return messageToEvent(ctx, v)
+
+	case *a2a.TaskArtifactUpdateEvent:
+		if len(v.Artifact.Parts) == 0 {
+			return nil, nil
+		}
+		event, err := artifactToEvent(ctx, v.Artifact)
+		if err != nil {
+			return nil, err
```
dpasiukevich: If a func has multiple error exits, IMO then it's nice to add the error context to those.

So then the codepath is clear from the error msg.

`return nil, fmt.Errorf("failed to convert TaskArtifactUpdateEvent: %w", err)` or something along the lines.

adka2a/events.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+}
+
+// ToADKMetaKey adds a prefix used to differentiage A2A-related values stored in custom metadata of an ADK session event.
+func ToADKMetaKey(key string) string {
+	return "a2a:" + key
+}
+
+// ToCustomMetadata creates a session event custom metadata with A2A task and context IDs in it.
+func ToCustomMetadata(tid a2a.TaskID, ctxID string) map[string]any {
```
dpasiukevich:
```suggestion
func ToCustomMetadata(taskID a2a.TaskID, ctxID string) map[string]any {
```

nit: maybe it's more readable

adka2a/events.go
```diff
@@ -0,0 +1,259 @@
...
+	if event == nil || event.CustomMetadata == nil {
+		return
+	}
+	if tid, ok := event.CustomMetadata[customMetaTaskIDKey].(string); ok {
+		taskID = a2a.TaskID(tid)
+	}
+	if ctxID, ok := event.CustomMetadata[customMetaContextIDKey].(string); ok {
+		contextID = ctxID
+	}
+	return
```
dpasiukevich: nit: not needed.

WDYT, maybe also have an explicit return instead of named return?

`return taskID, contextID`. Maybe it's just me, but I find as an extra step to check in func output what is actually returned (var names + types), because in func body it's only `return`.

adka2a/parts.go
```diff
@@ -152,21 +155,21 @@ func toA2ADataPart(part *genai.Part, longRunningToolIDs []string) (a2a.DataPart,
...
-	parts, err := toGenAIParts(msg.Parts)
+	parts, err := ToGenAIParts(msg.Parts)
 	if err != nil {
 		return nil, err
 	}
 	return &genai.Content{Role: genai.RoleUser, Parts: parts}, nil
 }
 
-func toGenAIParts(parts []a2a.Part) ([]*genai.Part, error) {
+func ToGenAIParts(parts []a2a.Part) ([]*genai.Part, error) {
```
dpasiukevich, line 166: Let's add a one line comment to document exported entities (useful for godoc).

internal/utils/utils.go
```diff
@@ -185,3 +186,20 @@ func hasTrailingCodeExecutionResult(resp *model.LLMResponse) bool {
 	lastPart := resp.Content.Parts[len(resp.Content.Parts)-1]
 	return lastPart.CodeExecutionResult != nil
 }
+
+// ToMapStructure converts any to map[string]any.
```
dpasiukevich: nit: WDYT if we try to keep utils organized a bit? e.g. adding this to `internal/converters/map_structure.go` or something.

adka2a/events.go
yarolegovich: empty return is needed 
image

but I don't mind getting rid of the named returned. I just thought that
```go
func GetA2ATaskInfo(event *session.Event) (taskID a2a.TaskID, contextID string)
// vs
func GetA2ATaskInfo(event *session.Event) (a2a.TaskID, string)
```
it was nice you could understand what the second string is about just looking at the signature.
but since the function is small and simple it probably doesn't change much

PR #166, opened 2025-10-21
feat: remoteagent utils

yarolegovich:
### Details

This PR contains utilities used in a2a remote agent implementation (#167).
1. `getUserFunctionCallAt` returns a non-nil response if the last session event is a user function call response. [Python ref](https://github.com/google/adk-python/blob/2158b3c/src/google/adk/agents/remote_a2a_agent.py#L319).
2. `toMissingRemoteSessionParts` collects parts from all the events not present in the remote agent session. An event is considered present in a remote session if there's an event at a higher index with remote agent as an author. [Python ref](https://github.com/google/adk-python/blob/2158b3c/src/google/adk/agents/remote_a2a_agent.py#L346).
3. `presentAsUserMessage` takes an event authored by another agent and converts it to a user event describing the provided event, eg: 
```
present(Event{author: "agent A", text: "Hello"}) -> Event{author: "user", text: "[agent A] said: Hello"}
```

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -15,7 +15,6 @@
 package utils
 
 import (
-	"encoding/json"
 	"strings"
 
 	"github.com/google/uuid"


internal/utils/utils.go
```diff
@@ -181,3 +182,20 @@ func hasTrailingCodeExecutionResult(resp *model.LLMResponse) bool {
 	lastPart := resp.Content.Parts[len(resp.Content.Parts)-1]
 	return lastPart.CodeExecutionResult != nil
 }
+
+// ToMapStructure converts any to map[string]any.
+// We can't use mapstructure library in a way compatible with ADK-python, because genai type fields
+// don't have proper field tags.
+// TODO(yarolegovich): field annotation PR for genai types.
+func ToMapStructure(data any) (map[string]any, error) {
```
dpasiukevich: can be removed

agent/remoteagent/utils.go
```diff
@@ -0,0 +1,157 @@
...
+	if responsePartIndex < 0 {
+		return "", false
+	}
+	return event.Content.Parts[responsePartIndex].FunctionResponse.ID, true
+}
+
+// toMissingRemoteSessionParts returns content parts for all events we think are not present in the remote session
+// and a2a contextID if it was found in a remote agent event metadata.
+// We iterate session events backward until all events are processed or an event authored by a remote agent is found.
+// Parts from all events we processed are returned as a single list.
```
dpasiukevich, line 88: Asking just to understand why all parts from all events are merged together in one list?

agent/remoteagent/utils.go
```diff
@@ -0,0 +1,157 @@
...
+			continue
+		}
+		parts, err := adka2a.ToA2AParts(event.Content.Parts, event.LongRunningToolIDs)
+		if err != nil {
+			// TODO(yarolegovich): log error
+			continue
+		}
+		result = append(result, parts...)
+	}
+	return result, contextID
```
dpasiukevich, line 123: What would be the behavior if contextID == ""? (e.g. lastRemoteResponseIndex == -1 because there's no event from remote agent found). This may never happen, but this is possible how the func code is defined, so does it make sense to handle such case explicitly?

agent/remoteagent/utils.go
```diff
@@ -0,0 +1,157 @@
...
+		tid, ctxID := adka2a.GetA2ATaskInfo(event)
+		result.taskID = tid
+		result.contextID = ctxID
+		return result
+	}
+	return nil
+}
+
+func isFunctionCallEvent(event *session.Event, callID string) bool {
+	if event.Content == nil {
```
dpasiukevich: Does it make sense to handle event == nil?

I don't expect the external code would return nil event for any reason, still this may make code more robust.

agent/remoteagent/utils.go
yarolegovich: `RemoteAgent` might be a non-ADK agent or someone else's ADK agent without access to our session store. So the data-sharing approach adk python took is to collect parts from all session events and present them as an A2A message.

agent/remoteagent/utils.go
yarolegovich: this is expected. `contextID` in a2a is an optional identifier which an agent can use to associate a message with previous tasks. if no contextID is provided server generates a new one
I'll leave a comment.

PR #167, opened 2025-10-21
feat: remote a2a agent

yarolegovich:
### Details

Implements a remote a2a agent. [Python ref](https://github.com/google/adk-python/blob/2158b3c/src/google/adk/agents/remote_a2a_agent.py#L109).
An agent can be created as `remoteagent.New(remoteagent.A2AConfig{...})`.
The agent can be configured with either an `a2a.AgentCard` or an `AgentCardSource` pointing to a local file or an http resource.
Other options contain configurations for the various steps of execution: card resolution, message send configuration, client creation.

For remote agent input:
1. If the last session event is a user function tool response: send content parts as new message parts with the task and contextID matching those on the matching function tool call.
2. Combine parts from all the events after the last event authored by the remote agent and present them as a single user message.

For remote agent output:
1. `TaskStatusUpdate` event parts are converted to thoughts, unless it's the final task status updates.
2. Empty `TaskArtifactUpdate`s (no parts) are skipped. 
3. Every event has a `CustomMetadata` with task and context IDs, the original A2A request and a matching A2A event.
4. If an error is generated during an execution it's converted to a session event with ErrorMessage.

These are slightly different from Python, but changes changes were made based on the implementation of `adka2a.NewExecutor`, I'll open a bug in their tracker to reconcile implementations.

### DIFF
--- a/agent/remoteagent/a2a_agent.go
+++ b/agent/remoteagent/a2a_agent.go
@@ -35,9 +35,9 @@ type A2AConfig struct {
 	Name        string
 	Description string
 
-	AgentCard *a2a.AgentCard
 	// AgentCardSource can be either an http(s) URL or a local file path. If a2a.AgentCard
 	// is not provided, the source is used to resolve the card during the first agent invocation.
+	AgentCard       *a2a.AgentCard
 	AgentCardSource string
 	// CardResolveOptions can be used to provide a set of agencard.Resolver configurations.
 	CardResolveOptions []agentcard.ResolveOption
@@ -52,21 +52,17 @@ type A2AConfig struct {
 // agent which can run in a different process or on a different host.
 func New(cfg A2AConfig) (agent.Agent, error) {
 	if cfg.AgentCard == nil && cfg.AgentCardSource == "" {
-		return nil, fmt.Errorf("either AgentCard, or AgentCardPath, or AgentCardURL must be provided")
+		return nil, fmt.Errorf("either AgentCard or AgentCardSource must be provided")
 	}
 
 	remoteAgent := &a2aAgent{resolvedCard: cfg.AgentCard}
-	a2aAgent, err := agent.New(agent.Config{
+	return agent.New(agent.Config{
 		Name:        cfg.Name,
 		Description: cfg.Description,
 		Run: func(ic agent.InvocationContext) iter.Seq2[*session.Event, error] {
 			return remoteAgent.run(ic, cfg)
 		},
 	})
-	if err != nil {
-		return nil, err
-	}
-	return a2aAgent, err
 }
 
 type a2aAgent struct {
@@ -124,7 +120,9 @@ func (a *a2aAgent) run(ctx agent.InvocationContext, cfg A2AConfig) iter.Seq2[*se
 				continue
 			}
 			updateCustomMetadata(event, req, a2aEvent)
-			yield(event, nil)
+			if !yield(event, nil) {
+				break
+			}
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


agent/remoteagent/a2a_agent.go
```diff
@@ -0,0 +1,205 @@
...
+		Name:        cfg.Name,
+		Description: cfg.Description,
+		Run: func(ic agent.InvocationContext) iter.Seq2[*session.Event, error] {
+			return remoteAgent.run(ic, cfg)
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+	return a2aAgent, err
```
dpasiukevich:
```suggestion
	return agent.New(agent.Config{
		Name:        cfg.Name,
		Description: cfg.Description,
		Run: func(ic agent.InvocationContext) iter.Seq2[*session.Event, error] {
			return remoteAgent.run(ic, cfg)
		},
	})
```

agent/remoteagent/a2a_agent.go
```diff
@@ -0,0 +1,205 @@
...
+				event := toErrorEvent(ctx, fmt.Errorf("failed to convert a2aEvent: %w", err))
+				updateCustomMetadata(event, req, nil)
+				yield(event, nil)
+				return
+			}
+			if event == nil {
+				continue
+			}
+			updateCustomMetadata(event, req, a2aEvent)
+			yield(event, nil)
```
dpasiukevich:
```suggestion
			if !yield(event, nil) {
				return
			}
```

it's better to check if iterator consumer stopped waiting

agent/remoteagent/a2a_agent.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+// A2AConfig is used to describe and configure a remote agent.
+type A2AConfig struct {
+	Name        string
+	Description string
+
+	AgentCard *a2a.AgentCard
+	// AgentCardSource can be either an http(s) URL or a local file path. If a2a.AgentCard
+	// is not provided, the source is used to resolve the card during the first agent invocation.
```
dpasiukevich, line 39: Maybe it makes sense to move `If a2a.AgentCard is not provided, the AgentCardSource will be used to resolve the card during the first agent invocation.` to AgentCard field?

agent/remoteagent/a2a_agent.go
```diff
@@ -0,0 +1,205 @@
...
+	ClientFactory *a2aclient.Factory
+	// MessageSendConfig is attached to a2a.MessageSendParams sent on every agent invocation.
+	MessageSendConfig *a2a.MessageSendConfig
+}
+
+// New creates a RemoteAgent. A2A (Agent-To-Agent) protocol is used for communication with an
+// agent which can run in a different process or on a different host.
+func New(cfg A2AConfig) (agent.Agent, error) {
+	if cfg.AgentCard == nil && cfg.AgentCardSource == "" {
+		return nil, fmt.Errorf("either AgentCard, or AgentCardPath, or AgentCardURL must be provided")
```
dpasiukevich:
```suggestion
		return nil, fmt.Errorf("either AgentCard, or AgentCardSource must be provided")
```

I'd say here we specify the actual fields we expect. And the possible value types (e.g. path or URL) should go to that specific field/type description. E.g. if there's a third way to configure AgentCardSource, will we have to list it there as well?

agent/remoteagent/a2a_agent.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	fileBytes, err := os.ReadFile(cfg.AgentCardSource)
+	if err != nil {
+		return nil, err
+	}
+
+	var card *a2a.AgentCard
+	if err := json.Unmarshal(fileBytes, card); err != nil {
+		return nil, err
```
dpasiukevich: IMO it's nice here to have error context, because this func now has 2 different path (card from URL and filepath).

Something like this:
line 139: fmt.Errorf("failed to resolve agent card by URL: %w", err)
line 144: fmt.Errorf("failed to read agent card source file: %w", err)
something similar in line 149

agent/remoteagent/a2a_agent.go
```diff
@@ -0,0 +1,205 @@
...
+		msg, err := newMessage(ctx)
+		if err != nil {
+			yield(toErrorEvent(ctx, fmt.Errorf("message creation failed: %w", err)), nil)
+			return
+		}
+
+		if len(msg.Parts) == 0 {
+			yield(adka2a.NewRemoteAgentEvent(ctx), nil)
+			return
+		}
```
dpasiukevich, line 102: Curiuos what is the reason to yield empty event with no payload?

agent/remoteagent/a2a_agent.go
yarolegovich: that's actually a leftover from when I started with three separate fields, will clean it up

agent/remoteagent/a2a_agent.go
yarolegovich: not sure, don't know enough about adk invocation semantics, this bit is taken from [python implementation](https://github.com/google/adk-python/blob/2158b3c/src/google/adk/agents/remote_a2a_agent.py#L499)

agent/remoteagent/a2a_agent.go
yarolegovich: my reasoning was to avoid too much wrapping, because the error gets wrapped on the caller level:
```go
fmt.Errorf("agent card resolution failed: %w", err)
```
and OS file read error, json unmarshaling error and card fetch errors should be distinguishable enough.
but I don't mind changing that

PR #171, opened 2025-10-21
[ADK-Web] Add tests for the agentgraphgenertor service

anFatum:


cmd/restapi/services/agentgraphgenerator_test.go
```diff
@@ -0,0 +1,582 @@
...
+func (m *mockTool) Description() string { return "" }
+func (m *mockTool) IsLongRunning() bool { return false }
+
+func TestNodeName(t *testing.T) {
+	tests := []struct {
+		name     string
+		instance any
+		expected string
+	}{
+		{"agent", newTestAgent(t, "TestAgent", "", agentinternal.TypeCustomAgent, nil, nil), "TestAgent"},
```
dpasiukevich: Could you add field names for test cases here and in the other tests where missing?

It make it a bit hard to read what's set and what's expected.

cmd/restapi/services/agentgraphgenerator_test.go
```diff
@@ -0,0 +1,582 @@
...
+		{name: "no highlight pairs", from: "A", to: "B", highlightedPairs: [][]string{}, expected: nil},
+		{name: "matching forward", from: "A", to: "B", highlightedPairs: [][]string{{"A", "B"}}, expected: boolPtr(true)},
+		{name: "matching backward", from: "B", to: "A", highlightedPairs: [][]string{{"A", "B"}}, expected: boolPtr(false)},
+		{name: "no match", from: "C", to: "D", highlightedPairs: [][]string{{"A", "B"}}, expected: nil},
+		{name: "partial match", from: "A", to: "C", highlightedPairs: [][]string{{"A", "B"}}, expected: nil},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := edgeHighlighted(tt.from, tt.to, tt.highlightedPairs)
```
dpasiukevich, line 355: nit: could I ask to add the comment for `edgeHighlighted` func, because it returns *bool, so there are 3 states, to explain how these 3 states are interpreted.

IIUC true == edge arrowhead forward; false == edge arrowhead backward; nil == no edge.

cmd/restapi/services/agentgraphgenerator_test.go
```diff
@@ -0,0 +1,582 @@
...
+		t.Fatalf("failed to set graph name: %v", err)
+	}
+
+	// Add nodes for edges to connect
+	nodes := []string{
+		"NodeA",
+		"NodeB",
+		"NodeC",
+		"NodeD",
+		"NodeE",
```
dpasiukevich: Maybe let's make the test cases here independent? So they don't rely on the same graph.

e.g. one test case drawing some edge may later affect/break the next cases.

cmd/restapi/services/agentgraphgenerator_test.go
```diff
@@ -0,0 +1,582 @@
...
+		// Check if edge exists between sub-agents
+		if edge == nil {
+			t.Fatalf("Edge between SubAgent1 and SubAgent2 not found")
+		}
+		if edge.Attrs["arrowhead"] != "none" {
+			t.Errorf("Sequential agent edge arrowhead mismatch: got %s", edge.Attrs["arrowhead"])
+		}
+	})
+
+	visitedNodes = make(map[string]bool)
```
dpasiukevich: similar here, does it make sense to make the test cases independent?

cmd/restapi/services/agentgraphgenerator_test.go
```diff
@@ -0,0 +1,582 @@
...
+		err := clusterGraph.SetName("cluster_LoopAgent")
+		if err != nil {
+			t.Fatalf("failed to set parent graph name: %v", err)
+		}
+
+		err = drawCluster(parentGraph, clusterGraph, loopAgent, [][]string{}, visitedNodes)
+		if err != nil {
+			t.Fatalf("drawCluster failed: %v", err)
+		}
+
```
dpasiukevich: does it make sense to add this similar check here (as in sequential agent)?

```
		if parentGraph.Nodes.Lookup["SubAgent1"] == nil || parentGraph.Nodes.Lookup["SubAgent2"] == nil {
			t.Error("Sub-agents not drawn as nodes in parent graph")
		}
```

PR #177, opened 2025-10-23
Add DatabaseSessionService

baptmont:
python ref: https://github.com/google/adk-python/blob/955632ce2c48a1f2cb38859d32be13e6e203ca2f/src/google/adk/sessions/database_session_service.py

### DIFF
--- a/session/database/service.go
+++ b/session/database/service.go
@@ -32,6 +32,14 @@ type databaseService struct {
 	db *gorm.DB
 }
 
+// NewSessionService creates a new session.Service implementation that uses a
+// relational database (e.g., PostgreSQL, Spanner, SQLite) via the GORM library.
+//
+// It requires a gorm.Dialector to specify the database connection and
+// accepts optional gorm.Option values for further GORM configuration.
+//
+// It returns the new session.Service or an error if the database connection
+// (gorm.Open) fails.
 func NewSessionService(dialector gorm.Dialector, opts ...gorm.Option) (session.Service, error) {
 	db, err := gorm.Open(dialector, opts...)
 	if err != nil {
@@ -455,13 +463,12 @@ func extractStateDeltas(delta map[string]any) (
 		return
 	}
 
-	// TODO: Replace temp, app and user: with session named const
 	for key, value := range delta {
-		if cleanKey, found := strings.CutPrefix(key, appPrefix); found {
+		if cleanKey, found := strings.CutPrefix(key, session.KeyPrefixApp); found {
 			appStateDelta[cleanKey] = value
-		} else if cleanKey, found := strings.CutPrefix(key, userPrefix); found {
+		} else if cleanKey, found := strings.CutPrefix(key, session.KeyPrefixUser); found {
 			userStateDelta[cleanKey] = value
-		} else if !strings.HasPrefix(key, tempPrefix) {
+		} else if !strings.HasPrefix(key, session.KeyPrefixTemp) {
 			// This key belongs to the session state, as long as it's not temporary.
 			sessionStateDelta[key] = value
 		}
@@ -481,13 +488,12 @@ func mergeStates(appState, userState, sessionState map[string]any) map[string]an
 	// avoid modifying the original sessionState map.
 	maps.Copy(mergedState, sessionState)
 
-	// TODO: Replace app and user: with session named const
 	for key, value := range appState {
-		mergedState[appPrefix+key] = value
+		mergedState[session.KeyPrefixApp+key] = value
 	}
 
 	for key, value := range userState {
-		mergedState[userPrefix+key] = value
+		mergedState[session.KeyPrefixUser+key] = value
 	}
 
 	return mergedState
--- a/session/database/session.go
+++ b/session/database/session.go
@@ -24,13 +24,6 @@ import (
 	"google.golang.org/adk/session"
 )
 
-// TODO consider change with enums or move to session package
-const (
-	appPrefix  = "app:"
-	userPrefix = "user:"
-	tempPrefix = "temp:"
-)
-
 // TODO localSession is identical to session.session. Move to sessioninternal
 type localSession struct {
 	appName   string
@@ -160,8 +153,7 @@ func trimTempDeltaState(event *session.Event) *session.Event {
 	// Iterate over the map and build a new one with the keys we want to keep.
 	filteredStateDelta := make(map[string]any)
 	for key, value := range event.Actions.StateDelta {
-		// TODO: Replace temp with session named const
-		if !strings.HasPrefix(key, tempPrefix) {
+		if !strings.HasPrefix(key, session.KeyPrefixTemp) {
 			filteredStateDelta[key] = value
 		}
 	}
@@ -173,22 +165,21 @@ func trimTempDeltaState(event *session.Event) *session.Event {
 }
 
 // updateSessionState updates the session state based on the event state delta.
-func updateSessionState(session *localSession, event *session.Event) error {
+func updateSessionState(sess *localSession, event *session.Event) error {
 	if event.Actions.StateDelta == nil {
 		return nil // Nothing to do
 	}
 
 	// Ensure the session state map is initialized
-	if session.state == nil {
-		session.state = make(map[string]any)
+	if sess.state == nil {
+		sess.state = make(map[string]any)
 	}
 
 	for key, value := range event.Actions.StateDelta {
-		//TODO replace tempPrefix with session named const
-		if strings.HasPrefix(key, tempPrefix) {
+		if strings.HasPrefix(key, session.KeyPrefixTemp) {
 			continue
 		}
-		session.state[key] = value
+		sess.state[key] = value
 	}
 
 	return nil


session/database/gorm_datatypes.go
```diff
@@ -0,0 +1,177 @@
...
+	"fmt"
+
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+	"gorm.io/gorm/schema"
+)
+
+// StateMap is a custom type for map[string]any that handles its own
+// JSON serialization and deserialization for the database by implementing gorm.Serializer.
+type StateMap map[string]any
```
dpasiukevich: Should this type be exported or we can make it internal?

session/database/gorm_datatypes.go
```diff
@@ -0,0 +1,177 @@
...
+
+// GormDataType / GormDBDataType (For Schema/Migrations)
```
dpasiukevich: nit: What does this comment describe? Should it be removed or rephrased somehow?

session/database/gorm_datatypes.go
```diff
@@ -0,0 +1,177 @@
...
+func (sm StateMap) Value() (driver.Value, error) {
+	if sm == nil {
+		sm = make(map[string]any) // Serialize as '{}' instead of NULL
+	}
+	// For all other databases, return a JSON string.
+	b, err := json.Marshal(sm)
+	if err != nil {
+		return nil, err
+	}
+	return string(b), nil
```
dpasiukevich, line 62:
```suggestion
	return json.Marshal(sm)
```

nit: Is this possible? Or []byte and string are treated differently and we need to type cast to string?

session/database/gorm_datatypes.go
```diff
@@ -0,0 +1,177 @@
...
+	if len(bytes) == 0 {
+		*sm = make(map[string]any)
+		return nil
+	}
+
+	return json.Unmarshal(bytes, sm)
+}
+
+func (sm StateMap) GormValue(ctx context.Context, db *gorm.DB) clause.Expr {
+	data, _ := json.Marshal(sm)
```
dpasiukevich, line 92: nit: could I ask to add TODO to log this?

When we'd add logger, we hopefully should be able integrate and log this.

session/database/gorm_datatypes.go
```diff
@@ -0,0 +1,177 @@
...
+	return gorm.Expr("?", string(data))
+}
+
+// DynamicJSON defined JSON data type, that implements driver.Valuer, sql.Scanner interface
+type DynamicJSON json.RawMessage
```
dpasiukevich: Similar: could we make it unexported? Or it must be a part of public API?

session/database/gorm_datatypes.go
```diff
@@ -0,0 +1,177 @@
...
+	return json.RawMessage(j).MarshalJSON()
+}
+
+// UnmarshalJSON to deserialize []byte
+func (j *DynamicJSON) UnmarshalJSON(b []byte) error {
+	result := json.RawMessage{}
+	err := result.UnmarshalJSON(b)
+	*j = DynamicJSON(result)
+	return err
+}
```
dpasiukevich: May be a silly question, but do we have to double json marshal?

Can we go with something like (not i:

```
func (j DynamicJSON) MarshalJSON() ([]byte, error) {
	return j, nil
}

// UnmarshalJSON to deserialize []byte
func (j *DynamicJSON) UnmarshalJSON(b []byte) error {
	*j = b
	return nil
}
```

session/database/gorm_datatypes.go
```diff
@@ -0,0 +1,177 @@
...
+		return "STRING(MAX)"
+	}
+	return ""
+}
+
+func (js DynamicJSON) GormValue(ctx context.Context, db *gorm.DB) clause.Expr {
+	if len(js) == 0 {
+		return gorm.Expr("NULL")
+	}
+	data, _ := js.MarshalJSON()
```
dpasiukevich: nit: could I ask to add TODO to log this?

When we'd add logger, we hopefully should be able integrate and log this.

session/database/session.go
```diff
@@ -0,0 +1,199 @@
...
+	"fmt"
+	"iter"
+	"strings"
+	"sync"
+	"time"
+
+	"google.golang.org/adk/session"
+)
+
+// TODO consider change with enums or move to session package
```
dpasiukevich: +1 this should be part of the session package (later PR obviously).

session/database/storage_session.go
```diff
@@ -0,0 +1,310 @@
...
+
+	var content *genai.Content
+	if len(se.Content) > 0 && string(se.Content) != "<null>" {
+		if err := json.Unmarshal(se.Content, &content); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal content: %w", err)
+		}
+	}
+
+	var groundingMetadata *genai.GroundingMetadata
+	if len(se.GroundingMetadata) > 0 && string(se.GroundingMetadata) != "<null>" {
```
dpasiukevich: Curious where does <null> keyword comes from so that we need to handle it?

session/database/service.go
```diff
@@ -0,0 +1,501 @@
...
+	"google.golang.org/adk/session"
+	"gorm.io/gorm"
+)
+
+// databaseService is an database implementation of sessionService.Service.
+type databaseService struct {
+	db *gorm.DB
+}
+
+func NewSessionService(dialector gorm.Dialector, opts ...gorm.Option) (session.Service, error) {
```
dpasiukevich, line 43: Could I ask to document this for users -- which backends we support, what to expect, when to run etc.

session/database/service.go
```diff
@@ -0,0 +1,501 @@
...
+	db, err := gorm.Open(dialector, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("error creating database session service :%w", err)
+	}
+	return &databaseService{db: db}, nil
+}
+
+// Create
```
dpasiukevich: nit: this comment could be removed probably

session/database/service.go
```diff
@@ -0,0 +1,501 @@
...
+	}
+
+	err = s.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		storageApp, err := fetchStorageAppState(tx, req.AppName)
+		if err != nil {
+			return err
+		}
+		storageUser, err := fetchStorageUserState(tx, req.AppName, req.UserID)
+		if err != nil {
+			return err
```
dpasiukevich:
```suggestion
			return fmt.Errorf("failed to fetch storage app state: %w", err)
		}
		storageUser, err := fetchStorageUserState(tx, req.AppName, req.UserID)
		if err != nil {
			return fmt.Errorf("failed to fetch storage user state: %w", err)
```

session/database/service.go
```diff
@@ -0,0 +1,501 @@
...
+	}
+
+	// fetch app and user states
+	storageApp, err := fetchStorageAppState(s.db.WithContext(ctx), appName)
+	if err != nil {
+		return nil, err
+	}
+	storageUser, err := fetchStorageUserState(s.db.WithContext(ctx), appName, userID)
+	if err != nil {
+		return nil, err
```
dpasiukevich: nit: similarly, let's add error context.

session/database/service.go
```diff
@@ -0,0 +1,501 @@
...
+		})
+
+	if userID != "" {
+		listQuery = listQuery.Where(&storageSession{
+			UserID: userID,
+		})
+	}
+
+	err := listQuery.Find(&foundSessions).Error
+
```
dpasiukevich: nit: remove newline

session/database/service.go
```diff
@@ -0,0 +1,501 @@
...
+	if err != nil {
+		// Specifically check if the error is "record not found".
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			// This is not a system failure. The record simply doesn't exist.
+			// Return nil for both response and error to indicate "not found".
```
dpasiukevich: nit: remove this comment line. We don't return nil for both response and error, it's `&session.ListResponse{...}, nil`

session/database/service.go
```diff
@@ -0,0 +1,501 @@
...
+	if userID != "" {
+		userState, err := fetchStorageUserState(s.db.WithContext(ctx), appName, userID)
+		if err != nil {
+			return nil, err
+		}
+		userStates = map[string]*storageUserState{userID: userState}
+	} else {
+		userStates, err = fetchAllAppStorageUserState(s.db.WithContext(ctx), appName)
+		if err != nil {
+			return nil, err
```
dpasiukevich: nit: could I ask to add error context

session/database/service.go
```diff
@@ -0,0 +1,501 @@
...
+		sess.state = mergeStates(storageApp.State, userState.State, sess.state)
+		responseSessions = append(responseSessions, sess)
+	}
+
+	return &session.ListResponse{
+		Sessions: responseSessions,
+	}, nil
+}
+
+// Delete
```
dpasiukevich: nit: delete

session/database/service.go
```diff
@@ -0,0 +1,501 @@
...
+		return err
+	}
+
+	// append it to session
+	err = sess.appendEvent(event)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
dpasiukevich:
```suggestion
	return sess.appendEvent(event)
```

session/database/service_test.go
```diff
@@ -0,0 +1,947 @@
...
+				ID: "new_event1",
+				LLMResponse: model.LLMResponse{
+					Partial: false,
+				},
+			},
+			wantStoredSession: &localSession{
+				appName:   "app1",
+				userID:    "user1",
+				sessionID: "session1",
+				events: []*session.Event{
```
dpasiukevich, line 503: Maybe let's add or modify the test case where existing session already has the event and we add a new one (to verify it's not overwritten).

session/database/service_test.go
```diff
@@ -0,0 +1,947 @@
...
+		if diff := cmp.Diff(wantState, gotState); diff != "" {
+			t.Errorf("User 2 state mismatch (-want +got):\n%s", diff)
+		}
+	})
+
+	t.Run("user_state_is_user_specific", func(t *testing.T) {
+		s := emptyService(t)
+		s1, _ := s.Create(ctx, &session.CreateRequest{AppName: appName, UserID: "u1", SessionID: "s1", State: map[string]any{"user:k1": "v1"}})
+		s1.Session.(*localSession).updatedAt = time.Now()
+		_ = s.AppendEvent(ctx, s1.Session.(*localSession), &session.Event{
```
dpasiukevich: nit: let's check for err: if err != nil { t.Fatal(err) } to avoid missing some unexpected flow in tests in future.

And in the other places.

session/database/gorm_datatypes.go
baptmont: The way this is handled is driver dependent, since we want to store them as "TEXT" we want to explicitly return a string. Some drivers could receive the raw binary data and attempt to store it as a BLOB.

session/database/storage_session.go
baptmont: This was coming from the RawMessage.MarshalJSON. I removed the method call and haven't encountered it again

session/database/gorm_datatypes.go
baptmont: This was not a double marshaling, the RawMessage.MarshalJSON was an identity function, handling the nil edge case. Since we already handle this edge case I removed the call.

```golang
// MarshalJSON returns m as the JSON encoding of m.
func (m RawMessage) MarshalJSON() ([]byte, error) {
	if m == nil {
		return []byte("null"), nil
	}
	return m, nil
}

File: session/database/gorm_datatypes.go (hunk 1 above)
Comment (baptmont) [reply session/database/gorm_datatypes.go]: Changed it to stateMap

File: session/database/gorm_datatypes.go (hunk 5 above)
Comment (baptmont) [reply session/database/gorm_datatypes.go]: Changed to dynamicJSON

File: session/database/service_test.go (hunk 19 above)
Comment (baptmont) [reply session/database/service_test.go:503]: Added "append event to the session with events and overwrite in storage"

File: session/database/service.go [hunk 21]
```diff
@@ -0,0 +1,494 @@
...
+
+	// fetch app and user states
+	storageApp, err := fetchStorageAppState(s.db.WithContext(ctx), appName)
+	if err != nil {
+		return nil, fmt.Errorf("error on get session: %w", err)
+	}
+	storageUser, err := fetchStorageUserState(s.db.WithContext(ctx), appName, userID)
+	if err != nil {
+		return nil, fmt.Errorf("error on get session: %w", err)
+	}
```
dpasiukevich, line 176:
```suggestion
	storageApp, err := fetchStorageAppState(s.db.WithContext(ctx), appName)
	if err != nil {
		return nil, fmt.Errorf("failed to fetch app state: %w", err)
	}
	storageUser, err := fetchStorageUserState(s.db.WithContext(ctx), appName, userID)
	if err != nil {
		return nil, fmt.Errorf("failed to fetch user state: %w", err)
	}
```

Having the same error message sort of negates the purpose. Could I ask to update here and in other places (e.g. List method)?

baptmont: Oh i see. The respective fetch method is already including that context to the error

```golang

func fetchStorageAppState(tx *gorm.DB, appName string) (*storageAppState, error) {
	var storageApp storageAppState
	if err := tx.First(&storageApp, "app_name = ?", appName).Error; err != nil {
		if !errors.Is(err, gorm.ErrRecordNotFound) {
			return nil, fmt.Errorf("failed to fetch app state: %w", err)
		}
		// If not found, initialize a new object to be created later.
		storageApp = storageAppState{AppName: appName, State: make(map[string]any)}
	}
	return &storageApp, nil
}

func fetchStorageUserState(tx *gorm.DB, appName string, userID string) (*storageUserState, error) {
	var storageUser storageUserState
	if err := tx.First(&storageUser, "app_name = ? AND user_id = ?", appName, userID).Error; err != nil {
		if !errors.Is(err, gorm.ErrRecordNotFound) {
			return nil, fmt.Errorf("failed to fetch user state: %w", err)
		}
		// If not found, initialize a new object.
		storageUser = storageUserState{AppName: appName, UserID: userID, State: make(map[string]any)}
	}
	return &storageUser, nil
}
```

PR #181, opened 2025-10-24
test: universal launcher test example

yarolegovich:
1. Serve a basic `a2a.AgentCard` from `/.well-known/agent-card.json`. 
2. When serving the card, use `Host` to set URL on which gRPC server is running. Haven't tested it, maybe `X-Forwarded-Host` or something else should be used instead.
3. Add a test which starts a server an uses `a2aclient` to use it.

### DIFF
--- a/cmd/launcher/web/a2a/a2a.go
+++ b/cmd/launcher/web/a2a/a2a.go
@@ -40,7 +40,7 @@ type a2aConfig struct {
 	agentURL string
 }
 
-type A2ALauncher struct {
+type a2aLauncher struct {
 	flags  *flag.FlagSet
 	config *a2aConfig
 }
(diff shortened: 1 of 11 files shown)


cmd/launcher/web/a2a/a2a_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	go func() {
+		if err := launcher.Run(t.Context(), config); err != nil {
+			t.Errorf("launcher.Run() error = %v", err)
+		}
+	}()
+
+	var card *a2a.AgentCard
+	for retry := range 3 {
+		time.Sleep(20 * time.Millisecond)
```
kdroste-google: Why do we need sleep?

cmd/launcher/web/server.go
```diff
@@ -135,17 +129,60 @@ func Serve(c *WebConfig, adkConfig *adk.Config) {
...
+		DefaultInputModes:  []string{"text/plain"},
+		DefaultOutputModes: []string{"text/plain"},
+		PreferredTransport: a2a.TransportProtocolGRPC,
+		Skills:             adka2a.GetAgentSkills(rootAgent),
+		Capabilities:       a2a.AgentCapabilities{Streaming: true},
+		// gRPC GetAgentCard() method will be serving empty card.
+		SupportsAuthenticatedExtendedCard: false,
+	}
+	rBase.HandleFunc("/.well-known/agent-card.json", func(w http.ResponseWriter, r *http.Request) {
+		grpcURL := "127.0.0.1:" + strconv.Itoa(c.LocalPort)
```
kdroste-google: Why 127.0.0.1?

cmd/launcher/web/a2a/a2a_test.go
yarolegovich: added an explanatory comment.
`go test ./cmd/launcher/web/a2a -count 1000` makes tests flaky because sometimes a request is trying to reach a server before it has started

cmd/launcher/web/server.go
yarolegovich: replaced with an `a2a_agent_url` flag

cmd/launcher/web/a2a/a2a.go
```diff
@@ -49,7 +51,7 @@ func NewLauncher() web.WebSublauncher {
 
 	fs := flag.NewFlagSet("a2a", flag.ContinueOnError)
 
-	// no flags at the moment
+	fs.StringVar(&config.agentURL, "a2a_agent_url", "localhost:8080", "A2A gRPC host URL as advertized in the public agent card. It is used by A2A clients as a connection endpoint.")
```
dpasiukevich:
```suggestion
	fs.StringVar(&config.agentURL, "a2a_agent_url", "localhost:8080", "A2A gRPC host URL as advertised in the public agent card. It is used by A2A clients as a connection endpoint.")
```

PR #182, opened 2025-10-24
Add state scopes to inmemmory session service

baptmont:
python ref:
https://github.com/google/adk-python/blob/main/src/google/adk/sessions/in_memory_session_service.py

Added internal sessionutils to reuse later for db session service.


internal/sessionutils/utils.go
```diff
@@ -0,0 +1,73 @@
...
+// limitations under the License.
+
+package sessionutils
+
+import (
+	"maps"
+	"strings"
+)
+
+const (
```
dpasiukevich, line 22: WDYT about exposing these constants in the `session` package? (e.g. `type SessionKeyPrefix string` etc).

session/inmemory.go
```diff
@@ -18,19 +18,25 @@ import (
...
 
 // inMemoryService is an in-memory implementation of sessionService.Service.
 // Thread-safe.
 type inMemoryService struct {
-	mu       sync.RWMutex
-	sessions omap.Map[string, *session] // session.ID) -> storedSession
+	mu        sync.RWMutex
+	sessions  omap.Map[string, *session] // session.ID) -> storedSession
+	userState map[string]map[string]map[string]any
+	appState  map[string]map[string]any
```
dpasiukevich: optional: if we make map[string]any an internal type will it increase readability?

something like `type stateMap map[string]any`

session/inmemory.go
```diff
@@ -91,38 +105,66 @@ func (s *inMemoryService) Get(ctx context.Context, req *GetRequest) (*GetRespons
...
+	if !req.After.IsZero() && len(filteredEvents) > 0 {
+		firstIndexToKeep := sort.Search(len(filteredEvents), func(i int) bool {
+			// Find the first event that is not before the timestamp
+			return !filteredEvents[i].Timestamp.Before(req.After)
+		})
+		filteredEvents = filteredEvents[firstIndexToKeep:]
+	}
+
+	copiedSession.events = make([]*Event, 0, len(filteredEvents))
+	copiedSession.events = append(copiedSession.events, filteredEvents...)
```
dpasiukevich, line 134: Why not `copiedSession.events = filteredEvents`?

session/inmemory.go
```diff
@@ -304,4 +393,59 @@ func (s *state) Set(key string, value any) error {
...
+	// Iterate over the map and build a new one with the keys we want to keep.
+	filteredStateDelta := make(map[string]any)
+	for key, value := range event.Actions.StateDelta {
+		if !strings.HasPrefix(key, sessionutils.TempPrefix) {
+			filteredStateDelta[key] = value
+		}
+	}
+
+	// Replace the old map with the newly filtered one.
+	event.Actions.StateDelta = filteredStateDelta
```
dpasiukevich, line 421: This one is tricky -- we do mutate the event here. Is it expected for e.g. AppendEvent?

Or it's better to pass session state internally across the funcs?

internal/sessionutils/utils.go
baptmont: We can do that if it makes sense to have them as part of the public API. If you prefer it let's do it and move this util functions there too, since we can't keep them in internal because of dependency cycles.

session/inmemory.go
baptmont: On Get the session service is creating a copy of the session, making it so that changes to that object will not modify the one that is saved. For the events of the session I was doing a shallow copy, creating a new array with the pointers to the events. If it makes sense, in the future, we can create a deep copy for the events too.
But by doing a copy at least we are making it so that changes to the array don't modify the stored array.

session/inmemory.go
baptmont: This seems to be the expect behavior for AppendEvent. We want to filter out all deltas with the temp prefix, since these keys are not persisted and should only exist during the current invocation.

session/inmemory.go
baptmont: Yes, great idea

internal/sessionutils/utils.go
dpasiukevich: 1. I think we should definitely expose constants at minimum. They are a part of public API -- users will use those to store key/values with a given scope.

2. Regarding the util funcs -- let's keep them internal for now. I don't see users benefitting from split/merge state funcs. Or maybe I'm missing something. WDYT?

session/inmemory.go
dpasiukevich: Could I ask to double check if possible (to make sure ADK python & java do mutate the event during append and filter out "temp:" keys).

If yes, let's update the AppendEvent method documentation https://github.com/google/adk-go/blob/a6cc67eee15fec96905edf1c2e34263693f42cea/session/service.go#L30
So it says -- the implementation MUST filter it out. And update the db session service if it doesn't do that.

Apologies for this hassle, but hopefully this would make UX better.

session/inmemory.go
dpasiukevich: Right, it makes sense to make a copiedSession.

But my comment is -- why can't we replace lines 133-134 with `copiedSession.events = filteredEvents` (I did select only those 2 lines but github doesn't show it intuitively).

PR #185, opened 2025-10-27
Unified launcher

kdroste-google:
Added universal launcher, which can run console or web with any combination of a2a, webui, api.
This version doesn't include static ADK Web UI files unless webui is included. 

Provides two additional "packs": full (containing console , web with a2a, api, webui) and prod (containing only web with a2a and api.

### DIFF
--- a/cmd/launcher/web/webui/webui.go
+++ b/cmd/launcher/web/webui/webui.go
@@ -62,23 +62,26 @@ func (w *WebUILauncher) Parse(args []string) ([]string, error) {
 	return restArgs, nil
 }
 
-// WrapHandlers implements web.WebSublauncher.
+// WrapHandlers implements the web.WebSublauncher interface. It does not change
+// the top-level routes for the WebUI.
 func (a *WebUILauncher) WrapHandlers(handler http.Handler, adkConfig *adk.Config) http.Handler {
 	// webui doesn't change the top level routes
 	return handler
 }
 
-// SetupSubrouters implements web.WebSublauncher.
+// SetupSubrouters implements the web.WebSublauncher interface. It adds the
+// WebUI subrouter to the main router.
 func (w *WebUILauncher) SetupSubrouters(router *mux.Router, adkConfig *adk.Config) {
 	w.AddSubrouter(router, w.config.pathPrefix, adkConfig, w.config.backendAddress)
 }
 
-// SimpleDescription implements web.WebSublauncher.
+// SimpleDescription returns a simple description of the WebUI launcher.
 func (w *WebUILauncher) SimpleDescription() string {
 	return "starts ADK Web UI server which provides UI for interacting with ADK REST API"
 }
 
-// UserMessage implements web.WebSublauncher.
+// UserMessage implements the web.WebSublauncher interface. It prints a message
+// to the user with the URL to access the WebUI.
 func (w *WebUILauncher) UserMessage(webUrl string, printer func(v ...any)) {
 	printer(fmt.Sprintf("       webui:  you can access API using %s%s", webUrl, w.config.pathPrefix))
 }
@@ -88,7 +91,7 @@ func (w *WebUILauncher) UserMessage(webUrl string, printer func(v ...any)) {
 //go:embed distr/*
 var content embed.FS
 
-// AddSubrouter adds a subrouter for the Web UI to the provided router.
+// AddSubrouter adds a subrouter to serve the ADK Web UI.
 func (w *WebUILauncher) AddSubrouter(router *mux.Router, pathPrefix string, adkConfig *adk.Config, backendAddress string) {
 	// Setup serving of ADK Web UI
 	rUi := router.Methods("GET").PathPrefix(pathPrefix).Subrouter()
@@ -115,7 +118,7 @@ func (w *WebUILauncher) AddSubrouter(router *mux.Router, pathPrefix string, adkC
 	rUi.Methods("GET").Handler(http.StripPrefix(pathPrefix, http.FileServer(http.FS(ui))))
 }
 
-// NewLauncher creates a new WebUILauncher.
+// NewLauncher creates a new WebSublauncher for the ADK Web UI.
 func NewLauncher() weblauncher.WebSublauncher {
 	config := &webUIConfig{}
 


cmd/launcher/prod/prod.go
```diff
@@ -12,18 +12,17 @@
...
+	"google.golang.org/adk/cmd/launcher/web/a2a"
+	"google.golang.org/adk/cmd/launcher/web/api"
 )
 
-// ADKAPIRouterConfigs contains the configs for the ADK API server.
-type ADKAPIRouterConfigs struct {
-	SessionService  session.Service
-	AgentLoader     services.AgentLoader
-	ArtifactService artifact.Service
+// NewLaucher returnes the most versalite universal launcher with all options enabled
```
anFatum: We don't return all options enabled in this launcher, right?

cmd/launcher/console/console.go
```diff
@@ -110,33 +111,44 @@ func (l ConsoleLauncher) Run(ctx context.Context, config *adk.Config) error {
...
+	if l.config.streamingModeString != string(agent.StreamingModeNone) &&
+		l.config.streamingModeString != string(agent.StreamingModeSSE) &&
+		l.config.streamingModeString != string(agent.StreamingModeBidi) {
+		return nil, fmt.Errorf("invalid streaming_mode: %v. Should be (%s|%s|%s)", l.config.streamingModeString,
+			agent.StreamingModeNone, agent.StreamingModeSSE, agent.StreamingModeBidi)
 	}
-	return &ConsoleLauncher{Config: consoleConfig}, argsLeft, nil
+	l.config.streamingMode = agent.StreamingMode(l.config.streamingModeString)
+	return l.flags.Args(), nil
+}
```
anFatum, line 138: How this remaining arguments should be working in case of multiple launchers? I see only one example with throwing an error on surplus of arguments, but other examples? Do you have any other examples in mind we can benefit from these remaining arguments?

What if there are flags of the same name in launchers?

cmd/launcher/prod/prod.go
kdroste-google: Yes, thanks for catching

kdroste-google: Fixed

cmd/launcher/console/console.go
kdroste-google: Args are parsed in following fashion: 
in the loop, we are looking at the first unparsed argument. If it is a known keyword, the remaining arguments are parsed (using a flagset specific for this keyword). Parsing stops at the first non-parsable argument. And we iterate.
So it accepts args: 
Keyword1  Keyword1Param1 ... Keyword1ParamN  Keyword2 Keyword2Param1 ... Keyword2ParamM  [rest of args]

Example:
`/app/main web -port 8080 api -webui_address 127.0.0.1:8081 a2a webui --api_server_address http://127.0.0.1:8081/api`
`          ___    Keyword1`
`               _________    Keyword1Param1`
`                         ___    Keyword2`
`                             _____________________________    Keyword2Param1`
`                                                           ___    Keyword3`
`                                                               ____    Keyword3`
`                                                                     _______________________________________________    Keyword3Param1`

Each launcher has its own set of flags, but they are not merged together, instead they are parsed independently

The idea behind is that a person who wants to use launcher also wants to provide their own command line arguments.

anFatum: Okay, AFAIU you're trimming the arguments based on the Keyword/launcher (l.chosenLauncher.Parse(args[1:])), but could it a problem if we have the parameters with the same name for different launches?

e.g. `/app/main web -port 8080 api --api_server_address 127.0.0.1:8081 webui --api_server_address http://127.0.0.1:8081/api`

not sure what the outcome would be if I try to parse arguments for the api?

kdroste-google: Should be fine, let me test it

kdroste-google: Yes, it's fine, those flags are independent

anFatum: Cool, thanks for testing!

cmd/launcher/console/console.go
```diff
@@ -31,19 +31,20 @@ import (
...
 )
 
-// ConsoleConfig contains command-line params for console launcher
-type ConsoleConfig struct {
-	streamingMode agent.StreamingMode
-	rootAgentName string
+// consoleConfig contains command-line params for console launcher
+type consoleConfig struct {
+	streamingMode       agent.StreamingMode
+	streamingModeString string // command-line param to be converted to agent.StremingMode
```
dpasiukevich:
```suggestion
	streamingModeString string // command-line param to be converted to agent.StreamingMode
```

cmd/launcher/full/full.go
```diff
@@ -0,0 +1,30 @@
...
+	"google.golang.org/adk/cmd/launcher/console"
+	"google.golang.org/adk/cmd/launcher/universal"
+	"google.golang.org/adk/cmd/launcher/web"
+	"google.golang.org/adk/cmd/launcher/web/a2a"
+	"google.golang.org/adk/cmd/launcher/web/api"
+	"google.golang.org/adk/cmd/launcher/web/webui"
+)
+
+// NewLaucher returnes the most versalite universal launcher with all options enabled
+func NewLaucher(rootAgentName string) *universal.Launcher {
```
dpasiukevich:
```suggestion
func NewLauncher(rootAgentName string) *universal.Launcher {
```

cmd/launcher/full/full.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"google.golang.org/adk/cmd/launcher/console"
+	"google.golang.org/adk/cmd/launcher/universal"
+	"google.golang.org/adk/cmd/launcher/web"
+	"google.golang.org/adk/cmd/launcher/web/a2a"
+	"google.golang.org/adk/cmd/launcher/web/api"
+	"google.golang.org/adk/cmd/launcher/web/webui"
+)
+
+// NewLaucher returnes the most versalite universal launcher with all options enabled
```
dpasiukevich: 1. versalite -> versatile
2. with all options enabled -> with all options built-in

WDYT?

cmd/launcher/console/console.go
```diff
@@ -59,9 +60,9 @@ func (l ConsoleLauncher) Run(ctx context.Context, config *adk.Config) error {
 		return fmt.Errorf("failed to create the session service: %v", err)
 	}
 
-	matchingAgent, err := config.AgentLoader.LoadAgent(l.Config.rootAgentName)
+	matchingAgent, err := config.AgentLoader.LoadAgent(l.config.rootAgentName)
 	if err != nil {
-		return fmt.Errorf("failed to find the matching agent: %v", err)
+		return fmt.Errorf("failed to find the matching agent: %v. Please specify a valid value for 'root_agent_name' parameter", err)
```
dpasiukevich:
```suggestion
		return fmt.Errorf("failed to load root agent: %w", err)
```

1. Rename suggestion
2. It's usually better to embed an error using %w for that.

cmd/launcher/console/console.go
```diff
@@ -31,19 +31,20 @@ import (
...
 type ConsoleLauncher struct {
-	Config *ConsoleConfig
+	flags  *flag.FlagSet
+	config *consoleConfig
 }
 
-// Run starts console loop. User-provided text is fed to the chosen agent (the only one if there's only one, specified by name otherwise)
-func (l ConsoleLauncher) Run(ctx context.Context, config *adk.Config) error {
+func (l *ConsoleLauncher) Run(ctx context.Context, config *adk.Config) error {
 	userID, appName := "test_user", "test_app"
```
dpasiukevich:
```suggestion
	userID, appName := "console_user", "console_app"
```

cmd/launcher/console/console.go
```diff
@@ -110,33 +111,44 @@ func (l ConsoleLauncher) Run(ctx context.Context, config *adk.Config) error {
...
+	return launcher.FormatFlagUsage(l.flags)
 }
 
-func ParseArgs(args []string) (*ConsoleConfig, []string, error) {
+func (l *ConsoleLauncher) SimpleDescription() string {
+	return "runs an agent in console mode."
+}
+
+// NewLauncher creates new console launcher. You may provide the default rootAgentName. It can be overriden by command-line params
+func NewLauncher(rootAgentName string) *ConsoleLauncher {
```
dpasiukevich: WDYT if it makes sense to follow this pattern: NewLauncher(rootAgentName string) **Launcher**?

So that each launcher package just exposes the constructor and returns launcher.Launcher interface to keep public API surface to minimum?

If needed we may make concrete types public in future only if it's would be needed later.

examples/vertexai/imagegenerator/main.go
```diff
@@ -82,7 +83,12 @@ func main() {
 		ArtifactService: artifact.InMemoryService(),
 		AgentLoader:     services.NewSingleAgentLoader(agent),
 	}
-	run.Run(ctx, config)
+
+	l := full.NewLaucher("image_generator")
+	err = l.ParseAndRun(ctx, config, os.Args[1:], universal.ErrorOnUnparsedArgs)
```
dpasiukevich: Let's simplify this if we can

1. Move root agent params from NewLauncher() to AgentLoader(). So that agents are created and managed by one entity.
2. Simplify the ParseAndRun:
  * rename to Run
  * make `ErrorOnUnparsedArgs` behaviour by default and drop this parameter. Later we may extend Run method with variadic func options (e.g. WithFlagCallbacks).

cmd/launcher/console/console.go
kdroste-google: Fixed

cmd/launcher/full/full.go
kdroste-google: Fixed, also in examples

cmd/launcher/full/full.go
kdroste-google: Sure

cmd/launcher/console/console.go
kdroste-google: Fixed

cmd/launcher/console/console.go
kdroste-google: Fixed

cmd/launcher/console/console.go
kdroste-google: rootAgentName is moved to AgentLoader
I've reduced public API surface

examples/vertexai/imagegenerator/main.go
kdroste-google: Ad 1) moved
Ad 2) I have introduced Execute 
ErrorOnUnparsedArgs is now a hidden default

cmd/launcher/web/web.go
```diff
@@ -12,62 +12,225 @@
...
 )
 
-// WebConfig contains command-line params for web launcher
-type WebConfig struct {
-	LocalPort       int
-	FrontendAddress string
-	BackendAddress  string
-	ServeA2A        bool
+// WebConfig contains parametres for lauching web server
+type webConfig struct {
```
dpasiukevich, line 35: nit: you can rename webConfig -> Config, WebLauncher -> Launcher etc.

because when used externally they right now look like: web.WebLauncher
with rename it's more tidy: web.Launcher

dpasiukevich:
Let's merge and iterate (as I understand you have other PR pending). Plus I'd experiment and take a deeper look on the launchers if I have suggestion I'd send some PR as well.

PR #186, opened 2025-10-28
feat: construct a2a.AgentSkill from agent.Agent

yarolegovich:
This code is the final missing piece for a2a interop feature parity with Python SDK.
A utility for constructing `[]a2a.AgentSkill` from `agent.Agent` for use in `a2a.AgentCard`.
[Python Reference](https://github.com/google/adk-python/blob/main/src/google/adk/a2a/utils/agent_card_builder.py#L81).

### DIFF
--- a/adka2a/executor.go
+++ b/adka2a/executor.go
@@ -69,10 +69,8 @@ func (e *Executor) Execute(ctx context.Context, reqCtx *a2asrv.RequestContext, q
 		return fmt.Errorf("failed to create a runner: %w", err)
 	}
 
-	task := reqCtx.Task
-	if task == nil {
-		task = &a2a.Task{ID: reqCtx.TaskID, ContextID: reqCtx.ContextID}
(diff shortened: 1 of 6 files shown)


adka2a/agent_card.go
```diff
@@ -15,20 +15,303 @@
...
+		{"you're", "I am"},
+		{"you've", "I have"},
+		{"yours", "mine"},
+		{"your", "my"},
+		{"you", "I"},
+	}
+	lower := strings.ToLower(instruction)
+	for _, sub := range substitutions {
+		for {
+			start := strings.Index(lower, sub.original)
```
dpasiukevich: handle word boundaries + test case

adka2a/agent_card_test.go
dpasiukevich, line 1: It may be worth checking and duplicating the tests cases from the https://github.com/google/adk-python/blob/main/tests/unittests/a2a/utils/test_agent_card_builder.py

PR #189, opened 2025-10-29
Add user scoped artifacts to in-memory artifact service

baptmont:
Python ref:https://github.com/google/adk-python/blob/c8e5340962014cfa09b59b32363e673f6d5552a7/src/google/adk/artifacts/in_memory_artifact_service.py
Documentation : https://google.github.io/adk-docs/artifacts/#namespacing-session-vs-user

### DIFF
--- a/artifact/inmemory.go
+++ b/artifact/inmemory.go
@@ -49,6 +49,7 @@ func fileHasUserNamespace(filename string) bool {
 }
 
 // userScopedArtifactKey defines the string for the part of the path used by user scope files
+// this key is used to store user artifacts within user namespace (available to all sessions for given app+user_id)
 const userScopedArtifactKey = "user"
 
 type artifactKey struct {


artifact/inmemory.go
```diff
@@ -41,6 +43,14 @@ func InMemoryService() Service {
 	return &inMemoryService{}
 }
 
+// fileHasUserNamespace checks if a filename indicates a user scoped artifact.
+func fileHasUserNamespace(filename string) bool {
+	return strings.HasPrefix(filename, "user:")
+}
+
+// userPathPart defines the stirng for the part of the path used by user scope files
```
dpasiukevich:
```suggestion
// userPathPart defines the string for the part of the path used by user scope files
```

artifact/inmemory.go
```diff
@@ -41,6 +43,14 @@ func InMemoryService() Service {
...
+const userPathPart = "user"
```
dpasiukevich: There's a high chance somebody may create a session with `session_id=user`. Then it will be possible to read artifacts of `user` session from any other by passing `user:key` (although it's still scoped within this one user sessions only so it's not that bad).

Is it viable to have a separate storage for user-scoped artifacts: `inMemoryService.userArtifacts` to avoid any potential bugs.
Or to have some unique key to store user-visible artifacts?

dpasiukevich: nit: 
```suggestion
// this key is used to store user artifacts within user namespace (available to all sessions for given app+user_id)
const userScopedArtifactKey = "user"
```

Or even make it a func something like `getUserScopedArtifactKey(...) artifactKey`

baptmont: This is how it works on python too, including for gcs artifact service. If anything we could make it so that "user" is an invalid custom session_id.
However allowing the creation of a custom session with the id "user", makes it so that agents of that session can create artifacts that are user-scoped and shared by all sessions of that app_id/user_id but that cannot be modified by other sessions.

dpasiukevich: I agree. And even if it causes any troubles (cross-session access of the same user), we can easily change that as it's the in-memory implementation (no backwards compatibility is needed).

PR #192, opened 2025-10-31
Add callback state modification

baptmont:
python refs
callbackContextState https://github.com/google/adk-python/blob/9ab17f2afd7ae427f5ea49639394cbcaa6c3cc40/src/google/adk/sessions/state.py
event yield https://github.com/google/adk-python/blob/86f01550bd1b52d6d160e8bc54cecc6c4fe8611c/src/google/adk/agents/base_agent.py#L300

tests:
https://github.com/google/adk-python/blob/546c2a68165f54e694664d5b6b6740566301782b/contributing/samples/session_state_agent/agent.py
https://github.com/google/adk-python/blob/546c2a68165f54e694664d5b6b6740566301782b/contributing/samples/live_tool_callbacks_agent/agent.py

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -199,8 +199,8 @@ func runBeforeAgentCallbacks(ctx InvocationContext) (*session.Event, error) {
 	return nil, nil
 }
 
-// runAfterAgentCallbacks checks if any afterAgentCallback returns non-nil content
-// then it replaces the event content with a value from the callback.
+// runAfterAgentCallbacks checks if any afterAgentCallback returns non-nil content or a state modification
+// then it create a new event with the new content and state delta.
 func runAfterAgentCallbacks(ctx InvocationContext) (*session.Event, error) {
 	agent := ctx.Agent()
 
@@ -319,8 +319,6 @@ func (c *callbackContextState) Set(key string, val any) error {
 func (c *callbackContextState) All() iter.Seq2[string, any] {
 	return c.ctx.invocationContext.Session().State().All()
 }
-
-var _ session.State = (*callbackContextState)(nil)
 
 type invocationContext struct {
 	context.Context
--- a/agent/llmagent/state_agent_test.go
+++ b/agent/llmagent/state_agent_test.go
@@ -59,14 +59,17 @@ func (f *FakeLLM) GenerateContent(ctx context.Context, req *model.LLMRequest, st
 
 var testSessionService session.Service
 
+type assertSessionParams struct {
+	title                   string
+	keysInCtxSession        []string
+	keysInServiceSession    []string
+	keysNotInServiceSession []string
+}
+
 func assertSessionValues(
 	t *testing.T,
-	ctx context.Context,
 	cctx agent.CallbackContext,
-	title string,
-	keysInCtxSession []string,
-	keysInServiceSession []string,
-	keysNotInServiceSession []string,
+	params *assertSessionParams,
 ) {
 	t.Helper()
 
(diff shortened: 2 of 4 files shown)


agent/agent.go
```diff
@@ -265,6 +296,32 @@ func (c *callbackContext) UserID() string {
...
+		c.ctx.actions.StateDelta[key] = val
+	}
+	return c.ctx.invocationContext.Session().State().Set(key, val)
+}
+
+func (c *callbackContextState) All() iter.Seq2[string, any] {
+	return c.ctx.invocationContext.Session().State().All()
+}
+
+var _ session.State = (*callbackContextState)(nil)
```
dpasiukevich: optional nit: this already is verified by `func (c *callbackContext) State() session.State {` which returns &callbackContextState

internal/context/callback_context.go
```diff
@@ -93,3 +98,29 @@ func (c *callbackContext) InvocationID() string {
...
+		c.ctx.eventActions.StateDelta[key] = val
+	}
+	return c.ctx.invocationCtx.Session().State().Set(key, val)
+}
+
+func (c *callbackContextState) All() iter.Seq2[string, any] {
+	return c.ctx.invocationCtx.Session().State().All()
+}
+
+var _ session.State = (*callbackContextState)(nil)
```
dpasiukevich: same nit: `State() session.State` validates `*callbackContextState` implements the `State()` interface

internal/llminternal/base_flow.go
```diff
@@ -112,8 +112,12 @@ func (f *Flow) runOneStep(ctx agent.InvocationContext) iter.Seq2[*session.Event,
 			return
 		}
 		spans := telemetry.StartTrace(ctx, "call_llm")
+		// Create event to pass to callback state delta
+		tmpEv := session.NewEvent(ctx.InvocationID())
+		tmpEv.Author = ctx.Agent().Name()
+		tmpEv.Branch = ctx.Branch()
 		// Calls the LLM.
-		for resp, err := range f.callLLM(ctx, req) {
+		for resp, err := range f.callLLM(ctx, req, tmpEv) {
```
dpasiukevich:
```suggestion
		for resp, err := range f.callLLM(ctx, req, stateDelta) {
```

Does it make sense to pass stateDelta instead of wrapping it in the event? Because IIUC we only need stateDelta here (for model callbacks) so it shows more clear intention: we need stateDelta -> we propagate stateDelta.

And we won't have to create that tmpEvent just for that.

agent/agent.go
```diff
@@ -186,29 +201,45 @@ func runBeforeAgentCallbacks(ctx InvocationContext) (*session.Event, error) {
 
 // runAfterAgentCallbacks checks if any afterAgentCallback returns non-nil content
 // then it replaces the event content with a value from the callback.
```
dpasiukevich: Could I ask to update the comment here?

agent/llmagent/state_agent_test.go
```diff
@@ -0,0 +1,641 @@
...
+		}
+	}
+}
+
+var testSessionService session.Service
+
+func assertSessionValues(
+	t *testing.T,
+	ctx context.Context,
+	cctx agent.CallbackContext,
```
dpasiukevich: Does it make sense to have only agent.CallbackContext?

agent/llmagent/state_agent_test.go
```diff
@@ -0,0 +1,641 @@
...
+			yield(&model.LLMResponse{
+				Content: genai.NewContentFromText("fake model response", genai.RoleModel),
+			}, nil)
+		}
+	}
+}
+
+var testSessionService session.Service
+
+func assertSessionValues(
```
dpasiukevich, line 69: nit: maybe let's update assertSessionValues to accept the struct (e.g. assertSessionParams)?

So that each it's clear from the func call which value belongs to which parameter

```
assertSessionValues(t, ctx, assertSessionParams{
  keysInCtxSession: []string{"before_agent_callback_state_key"},
  keysNotInServiceSession: []string{"before_agent_callback_state_key"}),
})

File: internal/llminternal/base_flow.go [hunk 7]
```diff
@@ -112,8 +112,10 @@ func (f *Flow) runOneStep(ctx agent.InvocationContext) iter.Seq2[*session.Event,
 			return
 		}
 		spans := telemetry.StartTrace(ctx, "call_llm")
+		// Create event to pass to callback state delta
```
dpasiukevich, line 115: nit: delete

PR #198, opened 2025-11-04
feat: use jsonrpc instead of grpc for a2a universal launcher module

yarolegovich:
JSONRPC server transport was merged to a2a-go main branch yesterday.
Using it instead of gRPC makes the cloud run setup simpler and allows us to server a2a under /a2a/ path, in a more consistent way with webui and adk rest.


cmd/launcher/web/a2a/a2a.go
```diff
@@ -76,64 +73,53 @@ func (a *a2aLauncher) Parse(args []string) ([]string, error) {
...
-			handler.ServeHTTP(w, r)
-		}
-	})
-	result = h2c.NewHandler(result, &http2.Server{})
-	return result
+	// doesn't change the top level routes
+	return handler
 }
 
 // SimpleDescription implements web.WebSublauncher. For A2A no subrouter definition is needed
```
kdroste-google: Please update the comment

cmd/launcher/web/web.go
```diff
@@ -157,7 +157,9 @@ func (w *Launcher) Run(ctx context.Context, config *adk.Config) error {
 
 	// Setup subrouters
 	for _, l := range w.activeSublaunchers {
-		l.SetupSubrouters(router, config)
+		if err := l.SetupSubrouters(router, config); err != nil {
+			return err
```
kdroste-google: Please add error msg indicating which sublauncher fails (Keyword?)

cmd/launcher/web/a2a/a2a.go
```diff
@@ -76,64 +73,53 @@ func (a *a2aLauncher) Parse(args []string) ([]string, error) {
 	return restArgs, nil
 }
 
-// WrapHandlers implements web.WebSublauncher. Returns http handler which basing on content-type
-// chooses between a2a grpc handler and provided handler
+// WrapHandlers implements web.WebSublauncher. For API, it doesn't change the top-level routes.
 func (a *a2aLauncher) WrapHandlers(handler http.Handler, adkConfig *adk.Config) http.Handler {
```
kdroste-google: It WrapHandler is not needed anymore, please remove it completely. It was needed only for the sake of a2a

yarolegovich:
> Question: do we need to alter cloudrun deployment? To add this agentcard url?

I think yes, we should have an option to specify it. 
An alternative fallback would be to change the a2a integration such that agent card route handler inspects http headers to know where it's hosted.

kdroste-google:
> > Question: do we need to alter cloudrun deployment? To add this agentcard url?
> 
> I think yes, we should have an option to specify it. An alternative fallback would be to change the a2a integration such that agent card route handler inspects http headers to know where it's hosted.

It's possible, but we have no guarantee that upstream proxies are not playing with those. I would say it's easier to leave it to the end user

PR #206, opened 2025-11-05
Implement EndInvocation

ngeorgy:


internal/context/invocation_context.go
```diff
@@ -82,11 +83,9 @@ func (c *InvocationContext) RunConfig() *agent.RunConfig {
...
-// TODO: implement endInvocation
-func (c *InvocationContext) EndInvocation() {
+func (c *InvocationContext) EndInvocation() bool {
+	return c.params.EndInvocation
 }
-
-// TODO: implement endInvocation
-func (c *InvocationContext) Ended() bool {
-	return false
+func (c *InvocationContext) SetEndInvocation(endInvocation bool) {
```
dpasiukevich: Maybe let's keep the `EndInvocation() bool` and `Ended()`?

Because it's not possible to un-end invocation `EndInvocation(false)` has no sense IMO.

ngeorgy: Sounds good, switched to the proposal

agent/agent_test.go
```diff
@@ -142,6 +142,39 @@ func TestAgentCallbacks(t *testing.T) {
 	}
 }
 
+func TestEndInvocation(t *testing.T) {
```
dpasiukevich: Could I ask to add 2nd test case? To verify after agent.Run() sets EndInvocation(), the afterAgentCallbacks are not called.

ngeorgy: Added, thank you!

internal/llminternal/base_flow.go
```diff
@@ -111,6 +111,9 @@ func (f *Flow) runOneStep(ctx agent.InvocationContext) iter.Seq2[*session.Event,
 			yield(nil, err)
 			return
 		}
+		if ctx.Ended() {
```
dpasiukevich, line 114: One more question -- do we need to check it before `f.preprocess(ctx, req)` or after?

ngeorgy: After, that's how python does it: https://github.com/google/adk-python/blob/main/src/google/adk/flows/llm_flows/base_llm_flow.py#L377

PR #209, opened 2025-11-05
Add docs for the agent package.

dpasiukevich:


agent/agent.go
```diff
@@ -57,29 +67,62 @@ func New(cfg Config) (Agent, error) {
...
-	AfterAgentCallbacks  []AfterAgentCallback
-}
-
+	// Run is the function that defines the agent's behavior.
+	Run func(InvocationContext) iter.Seq2[*session.Event, error]
+	// AfterAgentCallbacks is a list of callbacks that are called sequentially
+	// after the agent has completed its run.
+	//
+	// If any callback returns non-nil content or error, then a new event will be
+	// created from the content or error of that callback and other callbacks
```
ngeorgy: other -> remaining

As it's already written above.

agent/agent.go
```diff
@@ -27,6 +27,15 @@ import (
...
 )
 
+// Agent is the base interface which all agents must implement.
+//
+// Agents are created with ADK constructors to ensure correct
+// init & configuration.
+// The constructors are available in this package and its subpackages.
+// For example: llmagent.New, workflow agents, remote agent or
+// agent.New.
+// NOTE: in future releases we'd allow just implementing this interface. For now,
```
ngeorgy: we'd -> we will

agent/agent.go
```diff
@@ -57,29 +67,62 @@ func New(cfg Config) (Agent, error) {
 	}, nil
 }
 
+// Config is the configuration for creating a new Agent.
 type Config struct {
-	Name        string
+	// Name must be a non-empty string, unique within the agent tree.
+	// Agent name cannot be "user", since it's reserved for end-user's input.
+	Name string
+	// Description about the agent's capability.
```
ngeorgy: about -> of

agent/agent.go
```diff
@@ -57,29 +67,62 @@ func New(cfg Config) (Agent, error) {
...
+// Artifacts interface provides methods to work with artifacts of the current
+// session.
 type Artifacts interface {
 	Save(ctx context.Context, name string, data *genai.Part) (*artifact.SaveResponse, error)
 	List(context.Context) (*artifact.ListResponse, error)
 	Load(ctx context.Context, name string) (*artifact.LoadResponse, error)
 	LoadVersion(ctx context.Context, name string, version int) (*artifact.LoadResponse, error)
 }
 
+// Artifacts interface provides methods to access agent memory across the
```
ngeorgy: Artifacts -> Memory?

agent/agent.go
```diff
@@ -57,29 +67,62 @@ func New(cfg Config) (Agent, error) {
...
 
+// BeforeAgentCallback is a function that is called before the agent starts
+// its run.
+// If it returns non-nil content or error, the agent run will be skipped and a
+// new event will be created.
 type BeforeAgentCallback func(CallbackContext) (*genai.Content, error)
+
+// AfterAgentCallback is a function that is called after the agent has completed
+// its run.
+// If it returns non-nil content or error, a new event will be created.
```
ngeorgy, line 125: Here I would also add sth like:

It does not run if the agent was skipped due to BeforeAgentCallback returning content or if EndInvocation() was called during the agent's run.

agent/context.go
```diff
@@ -21,26 +21,86 @@ import (
...
+
+An agent call:
+ 1. Is handled by agent.Run().
+ 2. Ends when agent.Run() ends.
+
+An agent call can contain one or multiple steps.
+For example, LLM agent runs steps in a loop until:
+ 1. A final response is generated.
+ 2. The agent transfers to another agent.
+ 3. The end_invocation is set to true by any callbacks or tools.
```
ngeorgy: in callbacks and tools we don't really have the access to EndInvocation, It can be called only from InvocationContext and we have access to that in agent run

agent/context.go
```diff
@@ -21,26 +21,86 @@ import (
...
+
+A step:
+ 1. Calls the LLM only once and yields its response.
+ 2. Calls the tools and yields their responses if requested.
+
+The summarization of the function response is considered another step, since
+it is another llm call.
+A step ends when it's done calling llm and tools, or if the end_invocation
+is set to true at any time.
```
ngeorgy: I guess the variable end_invocation is Internal for them, maybe it's better to say EndInvocation() is called

agent/context.go
```diff
@@ -21,26 +21,86 @@ import (
...
+ 1. A final response is generated.
+ 2. The agent transfers to another agent.
+ 3. The end_invocation is set to true by any callbacks or tools.
+
+A step:
+ 1. Calls the LLM only once and yields its response.
+ 2. Calls the tools and yields their responses if requested.
+
+The summarization of the function response is considered another step, since
+it is another llm call.
```
ngeorgy: llm -> LLM

agent/context.go
```diff
@@ -21,26 +21,86 @@ import (
...
-	Agent() Agent
+
+	// UserContent that started this invocation.
 	UserContent() *genai.Content
+
+	// RunConfig stores the runtime configuration used during this invocation.
 	RunConfig() *RunConfig
 
+	// EndInvocation ends the current invocation. This stops any planned agent
+	// calls. This only
```
ngeorgy: Remove "this only"?

agent/context.go
dpasiukevich: Updated to `EndInvocation() was called by the invocation context.`

agent/context.go
dpasiukevich: Yep, good catch. Updated to `EndInvocation()`

PR #210, opened 2025-11-05
Add gemini tracking headers.

baptmont:
python ref: https://github.com/google/adk-python/blob/e01e54e9ee8e6e46c0214278ff3c744c3d3dd3cc/src/google/adk/models/google_llm.py#L219


model/gemini/gemini.go
```diff
@@ -18,25 +18,55 @@ import (
...
+	if cfg == nil {
+		cfg = &genai.ClientConfig{}
+	}
+	if cfg.HTTPOptions.Headers == nil {
+		cfg.HTTPOptions.Headers = make(http.Header)
+	}
+	m := &geminiModel{name: modelName}
+
+	// Create header value once, when the model is created
+	var b strings.Builder
```
mazas-google: Why not simply a single fmt.Sprintf, as per go/go-style/best-practices#string-concat-fmt?

PR #228, opened 2025-11-07
Fix branch prefix checking

ngeorgy:
Since right now we're checking prefix match backwards, it means that children won't be able to see parent content and parents will be able to see children's content, when it should be vice verca.

Python ref: https://github.com/google/adk-python/blob/main/src/google/adk/flows/llm_flows/contents.py#L624

### DIFF
--- a/internal/llminternal/contents_processor.go
+++ b/internal/llminternal/contents_processor.go
@@ -73,10 +73,7 @@ func buildContentsDefault(agentName, invocationBranch string, events []*session.
 		}
 		// Skip events that do not belong to the current branch.
 		// TODO: can we use a richier type for branch (e.g. []string) instead of using string prefix test?
-		// We use dot to delimit branch nodes. To avoid simple prefix match
-		// (e.g. agent_0 unexpectedly matching agent_00), require either perfect branch
-		// match, or match prefix with an additional explicit '.'
-		if invocationBranch != "" && (ev.Branch != invocationBranch && !strings.HasPrefix(invocationBranch, ev.Branch+".")) {
+		if !eventBelongsToBranch(invocationBranch, ev) {
 			continue
 		}
 		if isAuthEvent(ev) {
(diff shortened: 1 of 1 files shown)


internal/llminternal/contents_processor.go
```diff
@@ -75,7 +73,10 @@ func buildContentsDefault(agentName, branch string, events []*session.Event) ([]
 		}
 		// Skip events that do not belong to the current branch.
 		// TODO: can we use a richier type for branch (e.g. []string) instead of using string prefix test?
```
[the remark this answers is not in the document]
ngeorgy: Current improvement has nothing to do with this TODO, resolving this comment

internal/llminternal/contents_processor.go
```diff
@@ -75,7 +73,10 @@ func buildContentsDefault(agentName, branch string, events []*session.Event) ([]
...
-		if branch != "" && (ev.Branch != branch && !strings.HasPrefix(ev.Branch, branchPrefix)) {
+		// We use dot to delimit branch nodes. To avoid simple prefix match
+		// (e.g. agent_0 unexpectedly matching agent_00), require either perfect branch
+		// match, or match prefix with an additional explicit '.'
+		if invocationBranch != "" && (ev.Branch != invocationBranch && !strings.HasPrefix(invocationBranch, ev.Branch+".")) {
```
dpasiukevich: WDYT if we move the condition out to a func to simplify reading code?

`if !sameInvocationBranch(invocationBranch, ev.Branch) {`

internal/llminternal/contents_processor_test.go
dpasiukevich, line 339: Hmm, why this event is included in expected? It has `"branch1"`. But we require `branch: "branch1.task1"` prefix.

internal/llminternal/contents_processor.go
ngeorgy: Good idea! Done!

internal/llminternal/contents_processor_test.go
ngeorgy: Because branch1.task1 should see both branch1 (its parent) and branch1.task1

If event.branch = "parent" and invocation_branch = "parent.child", then invocation_branch.startswith(f'{event.branch}.') is TRUE. --> The child agent sees events from the parent's context.

If event.branch = "parent.child" and invocation_branch = "parent", then both checks are FALSE. --> The parent agent does not see events specific to the child's branch.

Python ref: https://github.com/google/adk-python/blob/main/src/google/adk/flows/llm_flows/contents.py#L624

dpasiukevich: Understood, thanks!

dpasiukevich (approved):

PR #230, opened 2025-11-07
Add .gitignore for adk-go

git-hulk:
Add .gitignore for the project to allow excluding unexpected directories. Just imported and changed a few lines from https://github.com/google/go-github.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,3 @@ profile.cov
 # Go workspace file
 go.work
 go.work.sum
-


dpasiukevich (dismissed):

.gitignore
```diff
@@ -0,0 +1,39 @@
...
+*.test
+
+# Code coverage profiles and other test artifacts
+*.out
+coverage.*
+*.coverprofile
+profile.cov
+
+# Dependency directories (remove the comment below to include it)
+vendor/
```
hyangah: vendor directory is not used in this project.
Can we delete it any way?

.gitignore
```diff
@@ -0,0 +1,39 @@
...
+
+# Go workspace file
+go.work
+go.work.sum
+
+# env file
+.env
+
+# Editor/IDE
```
hyangah: nit: there will be infinite number of editors and tools creating .files. We need to draw a line somewhere.

FYI in go projects, .gitignore is restricted to the artifacts produced by go build and test.

dpasiukevich: It's a good idea. Let's actually draw this line right now?

e.g. by adding the comment to prevent similar PRs (something along the lines):
```
# .gitignore is restricted to the artifacts produced by go build and test.
# for personal setup you can use personal .gitignore: `git config --global core.excludesfile ~/.gitignore`
```

.gitignore
git-hulk: It has been removed.

.gitignore
git-hulk: Yes, I also agree that drawing a line would help avoid similar PRs as more and more IDEs emerge in the age of AI.

I have removed the IDE's lines and added a comment line to state the function of .gitignore in this project.

.gitignore
```diff
@@ -0,0 +1,31 @@
...
+go.work.sum
+
+# env file
+.env
+
+# macOS files
+.DS_Store
+
+# vim temp files
+*.swp
```
dpasiukevich: Could we remove these, to be consistent with the comment on top? And we're good to merge.

Thanks!

git-hulk: Sure, removed now.

Deshaunmr: Good way

PR #232, opened 2025-11-07
cmd/adkgo: hide adkgo subpackages to internal

hyangah:
They don't need to be public.


dpasiukevich:
Does `internal` package name works at any level?  I assumed it's root-only. I mean the feature that go compiler won't allow importing stuff from `internal` from external modules.

hyangah:
Yes, it works at any level.  Packages outside the cmd/adkgo directory won't be able to access the packages in cmd/adkgo/internal.

https://go.dev/doc/go1.4#internalpackages

PR #233, opened 2025-11-07
Add support of the anthropic model

### LINKED ISSUE #225: When is the plan to support the Claude model?
Align to the python version models support:
https://github.com/google/adk-python/blob/main/src/google/adk/models/anthropic_llm.py

git-hulk:
This PR closes #225.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 
 require (
 	cloud.google.com/go v0.123.0
+	github.com/anthropics/anthropic-sdk-go v1.20.0
 	github.com/google/jsonschema-go v0.3.0
 	github.com/google/safehtml v0.1.0
 	github.com/modelcontextprotocol/go-sdk v0.7.0
@@ -86,6 +87,10 @@ require (
 	github.com/jinzhu/inflection v1.0.0 // indirect
 	github.com/jinzhu/now v1.1.5 // indirect
 	github.com/spf13/pflag v1.0.10 // indirect
+	github.com/tidwall/gjson v1.18.0 // indirect
+	github.com/tidwall/match v1.1.1 // indirect
+	github.com/tidwall/pretty v1.2.1 // indirect
+	github.com/tidwall/sjson v1.2.5 // indirect
 	github.com/yosida95/uritemplate/v3 v3.0.2 // indirect
 	golang.org/x/crypto v0.45.0 // indirect
 	golang.org/x/net v0.47.0 // indirect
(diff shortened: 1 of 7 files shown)


git-hulk:
Hi @dpasiukevich @mazas-google, I have implemented the first version of supporting the anthropic model, and have the basic test on my side. 

Could you please have a look when you get time? I would like to confirm if the direction is correct from the maintainer's perspective, and happy to modify the PR according to your feedback.

dpasiukevich:
I will try do a proper review tomorrow. Thank you for the contribution!

git-hulk:
@dpasiukevich, could you please take a look at this PR when you get time?

BarVaserman (approved):

GuyGoldenberg (approved):

GuyGoldenberg:
Any update here? This would be great if merged :)

git-hulk:
@GuyGoldenberg AFAIK, the adk-go community doesn't have an explicit plan for other models yet, according to the reply in https://github.com/google/adk-go/pull/242#issuecomment-3811793950. cc @dpasiukevich

verdverm:
I would say the intention is to have things like this in the `adk-go` community contribution repo, when it gets started, still 404s afaik

wolo-lab:
We are going forward with #598 for Claude support, and closing this one.

Both PRs solve the same problem, and #598 covers more of the surface: extended thinking round-tripped with its signature so history can be replayed, `ToolConfig` mapped through to `tool_choice`, validated image media types, documents, and citations. For a provider the team expects to maintain for years, starting from the more complete conversion layer is the lower-risk path. That is a judgement about scope and nothing else.

What you did here deserves saying properly. You volunteered on #225 within hours of it being opened, agreed the approach in the thread before writing a line — the SDK choice, Vertex AI as the default — and then delivered a working implementation against that agreement. You ran it against real traffic for months, told people in the thread they were welcome to use it, and offered to keep maintaining it while the wider question of external providers was still open. Several people took you up on that. For most of the last year this PR was the answer to "can I use Claude with ADK Go", and that mattered.

The delay is ours. Ten months is far too long to leave a contributor waiting on a direction call, particularly one who had asked for the direction and was told it was coming.

If you are interested, we would value you as a reviewer on the Anthropic package as it lands. You have run Claude against this framework in production longer than anyone else here, and that experience is worth more to us than any of the review we can do internally.

Thank you for the work, and for the patience.

PR #235, opened 2025-11-07
Add vertexAi session service

baptmont:
python ref: https://github.com/google/adk-python/blob/0ccc43cf49dc0882dc896455d6603a602d8a28e7/src/google/adk/sessions/vertex_ai_session_service.py

Fixes https://github.com/google/adk-go/issues/347

### DIFF
--- a/session/vertexai/vertexai_client.go
+++ b/session/vertexai/vertexai_client.go
@@ -433,6 +433,8 @@ func aiplatformToGenaiContent(rpcResp *aiplatformpb.SessionEvent) *genai.Content
 		role := rpcResp.Content.Role
 		for _, respPart := range rpcResp.Content.Parts {
 			part := &genai.Part{}
+			part.Thought = respPart.Thought
+			part.ThoughtSignature = respPart.ThoughtSignature
 			switch v := respPart.Data.(type) {
 			case *aiplatformpb.Part_Text:
 				part.Text = v.Text
@@ -471,6 +473,8 @@ func createAiplatformpbContent(event *session.Event) (*aiplatformpb.Content, err
 		parts := make([]*aiplatformpb.Part, 0)
 		for _, part := range event.Content.Parts {
 			aiplatformPart := &aiplatformpb.Part{}
+			aiplatformPart.Thought = part.Thought
+			aiplatformPart.ThoughtSignature = part.ThoughtSignature
 			if part.Text != "" {
 				aiplatformPart.Data = &aiplatformpb.Part_Text{Text: part.Text}
 			}


baptmont:
missing state scopes
currently only text events are stored

session/vertexai/vertexai.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (s *vertexAiService) Get(ctx context.Context, req *session.GetRequest) (*session.GetResponse, error) {
+	if req.AppName == "" || req.UserID == "" || req.SessionID == "" {
+		return nil, fmt.Errorf("app_name, user_id and session_id are required, got app_name: %q, user_id: %q, session_id: %q", req.AppName, req.UserID, req.SessionID)
+	}
+	sess, err := s.client.getSession(ctx, req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get session: %w", err)
+	}
+	events, err := s.client.listSessionEvents(ctx, req.AppName, req.SessionID, req.After, req.NumRecentEvents)
```
dpasiukevich: Can we make 2 concurrent calls here: `getSession` and `listSessionEvents`?

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,704 @@
...
+}
+
+// Ensure you close it when your application shuts down
+func (c *vertexAiClient) Close() error {
+	return c.rpcClient.Close()
+}
+
+func (c *vertexAiClient) createSession(ctx context.Context, req *session.CreateRequest) (*localSession, error) {
+	pbSession := &aiplatformpb.Session{
+		UserId: req.UserID,
```
dpasiukevich, line 67: Let's set the name `Name: req.Name`.

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,704 @@
...
+		baseDelay  = 1 * time.Second
+		maxDelay   = 5 * time.Second
+	)
+
+	for i := 0; i < maxRetries; i++ {
+		// Get the latest status of the operation.
+		ls, err := c.getSession(ctx, &session.GetRequest{AppName: appName, UserID: userId, SessionID: sessionID})
+		if err != nil {
+			// Basic retry on "not found" which might be due to propagation
+			if i < maxRetries-1 { // sLROtatus.Code(err) == codes.NotFound &&
```
dpasiukevich: // // sLROtatus.Code(err) == codes.NotFound &&

Why is it not commented out? Should we retry only for NotFound?

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,704 @@
...
+				delay := time.Duration(i*i) * baseDelay
```
dpasiukevich, line 123: Seems like baseDelay is not needed, it can be removed as it's 1 sec only.

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,704 @@
...
+	}
+	sessRpcResp, err := c.rpcClient.GetSession(ctx, sessRpcReq)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching session: %w", err)
+	}
+
+	if sessRpcResp == nil {
+		return nil, fmt.Errorf("session %+v not found", req.SessionID)
+	}
+	if sessRpcResp.UserId != req.UserID {
```
dpasiukevich, line 154: Can we remove this check `sessRpcResp.UserId != req.UserID` rather instead set the user_id in the GetSessionRequest?

Or we can leave it (to be extra sure the session doesn't belong to a wrong user_id).

But still let's maybe populate `UserID` field in `GetSessionRequest`?

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,704 @@
...
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("error creating session list: %w", err)
+		}
+		session := &localSession{
+			appName:   req.AppName,
+			userID:    rpcResp.UserId,
+			sessionID: sessionIdBySessionName(rpcResp.Name),
+			state:     filterNilValues(rpcResp.SessionState.AsMap()),
```
dpasiukevich, line 197: Curious why do we need to filter nil values?

Can vertex session service return such entries (key with nil value)? And/or ADK cannot handle such values? I haven't verified but I'd expect ADK at least should not care if the value is nil (as long as it's serializable, and nils are).

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,704 @@
...
+		sessionState, err := structpb.NewStruct(event.Actions.StateDelta)
+		if err != nil {
+			return fmt.Errorf("failed to convert state to structpb: %w", err)
+		}
+		eventState = &aiplatformpb.EventActions{StateDelta: sessionState}
+	}
+
+	// TODO add logic for other types of parts
+	var content *aiplatformpb.Content
+	if event.Content != nil {
```
dpasiukevich: I'd suggest to move it out to a separate func, especially if in future we'd add more logic to convert more types.

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,704 @@
...
+	if err != nil {
+		return nil, err
+	}
+	events := make([]*session.Event, 0)
+	eventsRpcReq := &aiplatformpb.ListEventsRequest{
+		Parent: sessionNameByID(sessionID, c, reasoningEngine),
+	}
+	if !after.IsZero() {
+		// TODO after parameter support
+		// eventsRpcReq.Filter = fmt.Sprintf("timestamp>=%q", after.Format("2006-01-02T15:04:05-07:00"))
```
dpasiukevich: Curious why it's a TODO with the code ready but it's commented out? I assume this is the whole code for this time filter or something is yet TBD?

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,704 @@
...
+	for {
+		rpcResp, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("error fetching session events: %w", err)
+		}
+
+		// TODO add logic for other types of parts
```
dpasiukevich: Similar, let's maybe move it out to a func. We can have something like `genaiToAiplatformContent` and `aiplatformToGenaiContent`.

Because this convertson logic would grow quite a bit once we will add some type support.

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,704 @@
...
+					Text:    &gc.Maps.Text,
+					PlaceId: &gc.Maps.PlaceID,
+				}
+				if gc.Maps.PlaceAnswerSources != nil {
+					var reviewSnippets []*aiplatformpb.GroundingChunk_Maps_PlaceAnswerSources_ReviewSnippet
+					for _, source := range gc.Maps.PlaceAnswerSources.ReviewSnippets {
+						snippet := &aiplatformpb.GroundingChunk_Maps_PlaceAnswerSources_ReviewSnippet{
+							ReviewId:      source.Review,
+							GoogleMapsUri: source.GoogleMapsURI,
+							// FIXME Title: missing in source
```
dpasiukevich: Curious about this FIXME -- why Title cannot be converted? `Title: source.TItle`?

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,704 @@
...
+	}
+
+	out := &genai.GroundingMetadata{
+		WebSearchQueries: metadata.WebSearchQueries,
+		RetrievalQueries: metadata.RetrievalQueries,
+	}
+
+	// Handle string pointer for Context Token
+	if metadata.GoogleMapsWidgetContextToken != nil {
+		out.GoogleMapsWidgetContextToken = *metadata.GoogleMapsWidgetContextToken
```
dpasiukevich: why derefString cannot be used here?

session/vertexai/service_test.go
```diff
@@ -0,0 +1,1282 @@
...
+				AppName: EngineId,
+				UserID:  "testUserID",
+				State: map[string]any{
+					// TODO had to parse to float64, sending int was modified by vertex or by vertex client, int should work
+					"k": float64(5),
+				},
+			},
+		},
+		{
+			name:  "when already exists, it fails",
```
dpasiukevich: how this test case is different from `full key`? Both provide custom `session_id` and fails due to this.

session/vertexai/vertexai_client.go
baptmont: Name is of the aiplatformpb session is the identifier given by projects/{project}/locations/{location}/reasoningEngines/{reasoning_engine}/sessions/{session}, we don't have the session id yet so we can't set the name on create

session/vertexai/vertexai_client.go
baptmont: Lets retry on everything

session/vertexai/vertexai_client.go
baptmont: time.Second is 1000 so baseDelay is 1000, I ll remove the "1 *" part.

session/vertexai/vertexai_client.go
baptmont: GetSessionRequest does not have the UserID field.

session/vertexai/vertexai_client.go
baptmont: This was before genai version was updated, Title did not exist in source before

session/vertexai/vertexai_client.go
baptmont: it can, fixed

session/vertexai/vertexai_client.go
dpasiukevich: Why on everything? Let's retry only on known error codes we should retry (e.g. `codes.NotFound`).

Or vertex API returnes something different when session not found.

IMO we don't need to retry on internal errors etc, only for `codes.NotFound`.

session/vertexai/vertexai_client.go
baptmont: Fixed, the precision of vertex ai session service was not working for the nano seconds used in the tests.
So it looked like the filter was not working. Changed the tests to use seconds and it works now

session/vertexai/vertexai_client.go
baptmont: This one is weird... we currently don't have a way to delete entries from state. So once a "app:" or "user:" scoped state variable is created it in vertex it can't be deleted. So if its set to nil I am filtering it out as a way to exclude them from sessions.

dpasiukevich (approved):

session/vertexai/session.go
```diff
@@ -0,0 +1,197 @@
...
+		state: s.state,
+	}
+}
+
+func (s *localSession) Events() session.Events {
+	return events(s.events)
+}
+
+func (s *localSession) LastUpdateTime() time.Time {
+	s.mu.RLock()
```
kdroste-google, line 65: what sense does it make to read updatedAt in critical section an leaving it immidiately?

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,722 @@
...
+	"google.golang.org/adk/session"
+
+	aiplatform "cloud.google.com/go/aiplatform/apiv1beta1"
+	aiplatformpb "cloud.google.com/go/aiplatform/apiv1beta1/aiplatformpb"
+)
+
+const (
+	engineResourceTemplate  = "projects/%s/locations/%s/reasoningEngines/%s"
+	sessionResourceTemplate = engineResourceTemplate + "/sessions/%s"
+	connectionErrorTemplate = "could not establish connection to the aiplatform server: %s"
```
kdroste-google: it seems it's used only once and is not exported. Please consider inline'ing it

kdroste-google: why not %w?

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,722 @@
...
+		maxDelay   = 5 * time.Second
+	)
+
+	for i := 0; i < maxRetries; i++ {
+		// Get the latest status of the operation.
+		ls, err := c.getSession(ctx, &session.GetRequest{AppName: appName, UserID: userId, SessionID: sessionID})
+		if err != nil {
+			// Basic retry on "not found" which might be due to propagation
+			if i < maxRetries-1 && IsNotFoundError(err) {
+				delay := time.Duration(i*i) * baseDelay
```
kdroste-google, line 123: The delayes would be [sec]: 1, 4, 5, 5, 5, 5, 5, ,5..., 5,   45sec in total. Does it make sense?

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,722 @@
...
+
+	reasoningEngine, err := c.getReasoningEngineID(req.AppName)
+	if err != nil {
+		return nil, err
+	}
+	rpcReq := &aiplatformpb.ListSessionsRequest{
+		Parent: fmt.Sprintf(engineResourceTemplate, c.projectID, c.location, reasoningEngine),
+	}
+	if req.UserID != "" {
+		rpcReq.Filter = fmt.Sprintf("userId=\"%s\"", req.UserID)
```
kdroste-google, line 178: is this safe?

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,722 @@
...
+	reasoningEngine, err := c.getReasoningEngineID(appName)
+	if err != nil {
+		return nil, err
+	}
+	events := make([]*session.Event, 0)
+	eventsRpcReq := &aiplatformpb.ListEventsRequest{
+		Parent: sessionNameByID(sessionID, c, reasoningEngine),
+	}
+	if !after.IsZero() {
+		eventsRpcReq.Filter = fmt.Sprintf("timestamp>=%q", after.Format("2006-01-02T15:04:05-07:00"))
```
kdroste-google, line 297: "2006-01-02T15:04:05-07:00" seems non-standard. Is that a correct format? The nearest is RFC3339 "2006-01-02T15:04:05Z07:00"

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,722 @@
...
+		if numRecentEvents > len(events) {
+			return events, nil
+		}
+		return events[len(events)-numRecentEvents:], nil
+	}
+	return events, nil
+}
+
+func sessionIdBySessionName(sn string) string {
+	return sn[strings.LastIndex(sn, "/")+1:]
```
kdroste-google: are we sure "/" is in the name?

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,722 @@
...
+}
+
+func sessionIDByOperationName(on string) string {
+	return on[strings.LastIndex(on, "/sessions/")+10 : strings.LastIndex(on, "/operations/")]
```
kdroste-google: Are we sure on contains this data?  
Where did 10 come from?

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,722 @@
...
+
+	// Regex pattern to match the full resource name
+	matches := reasoningEnginePattern.FindStringSubmatch(appName)
+
+	if len(matches) == 0 {
+		return "", fmt.Errorf("app name %s is not valid. It should either be the full ReasoningEngine resource name, or the reasoning engine id", appName)
+	}
+
+	// The last group is the ID
+	return matches[len(matches)-1], nil
```
kdroste-google: non-capturing groups can be used `^projects/(?:[a-zA-Z0-9-_]+)/locations/(?:[a-zA-Z0-9-_]+)/reasoningEngines/(\d+)$

examples/vertexai/agent.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+const (
+	modelName = "gemini-2.5-flash"
+)
+
+func main() {
+	ctx := context.Background()
+
+	projectID := os.Getenv("CLOUD_PROJECT_ID")
```
kdroste-google: In other examples we have GOOGLE_CLOUD_PROJECT

examples/vertexai/agent.go
```diff
@@ -0,0 +1,101 @@
...
+	if projectID == "" {
+		log.Fatalf("Env var CLOUD_PROJECT_ID is not set")
+	}
+	location := os.Getenv("CLOUD_LOCATION")
```
kdroste-google: In other examples we have GOOGLE_CLOUD_LOCATION

examples/vertexai/vertexengine/create_engine.go
```diff
@@ -0,0 +1,90 @@
...
+
+	aiplatform "cloud.google.com/go/aiplatform/apiv1"
+	"cloud.google.com/go/aiplatform/apiv1/aiplatformpb"
+	"google.golang.org/api/option"
+)
+
+// main defines an example of how to initialize a vertex ai reasoning engine
+func main() {
+	ctx := context.Background()
+	projectID := os.Getenv("CLOUD_PROJECT_ID")
```
kdroste-google: In other examples we have GOOGLE_CLOUD_PROJECT

examples/vertexai/vertexengine/create_engine.go
```diff
@@ -0,0 +1,90 @@
...
+	if projectID == "" {
+		log.Fatalf("Env var CLOUD_PROJECT_ID is not set")
+	}
+	location := os.Getenv("CLOUD_LOCATION")
```
kdroste-google: In other examples we have GOOGLE_CLOUD_LOCATION

kdroste-google (changes requested):
Please answer my doubts in comments, thank you!

session/vertexai/session.go
baptmont: yes, time.Time reads are not atomic 
``` go
type Time struct {
	// wall and ext encode the wall time seconds, wall time nanoseconds,
	// and optional monotonic clock reading in nanoseconds.
	//
	// From high to low bit position, wall encodes a 1-bit flag (hasMonotonic),
	// a 33-bit seconds field, and a 30-bit wall time nanoseconds field.
	// The nanoseconds field is in the range [0, 999999999].
	// If the hasMonotonic bit is 0, then the 33-bit field must be zero
	// and the full signed 64-bit wall seconds since Jan 1 year 1 is stored in ext.
	// If the hasMonotonic bit is 1, then the 33-bit field holds a 33-bit
	// unsigned wall seconds since Jan 1 year 1885, and ext holds a
	// signed 64-bit monotonic clock reading, nanoseconds since process start.
	wall uint64
	ext  int64

	// loc specifies the Location that should be used to
	// determine the minute, hour, month, day, and year
	// that correspond to this Time.
	// The nil location means UTC.
	// All UTC times are represented with loc==nil, never loc==&utcLoc.
	loc *Location
}
```

session/vertexai/vertexai_client.go
baptmont: This code is temporary while the LRO client is still unfixed
The retry times are not for the cases where timeouts occur. The issue is that it returns a not found value while the LRO is not processed. 
The logic is similar to the LRO client and I think its okay for now

session/vertexai/vertexai_client.go
baptmont: Yes, that is what is in the documentation as RFC 3339
```go
// Optional. The standard list filter.
	// Supported fields:
	//   - `timestamp` range (i.e. `timestamp>="2025-01-31T11:30:00-04:00"` where
	//     the timestamp is in RFC 3339 format)
	//
	// More detail in [AIP-160](https://google.aip.dev/160).
	Filter string `protobuf:"bytes,4,opt,name=filter,proto3" json:"filter,omitempty"`
```

session/vertexai/vertexai_client.go
baptmont: The proto documents it as :
``` go
	// Identifier. The resource name of the event.
	// Format:`projects/{project}/locations/{location}/reasoningEngines/{reasoning_engine}/sessions/{session}/events/{event}`.
	Name string `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
```
I'll add a check either way

session/vertexai/vertexai_client.go
```diff
@@ -0,0 +1,722 @@
...
+					PlaceId: &gc.Maps.PlaceID,
+				}
+				if gc.Maps.PlaceAnswerSources != nil {
+					var reviewSnippets []*aiplatformpb.GroundingChunk_Maps_PlaceAnswerSources_ReviewSnippet
+					for _, source := range gc.Maps.PlaceAnswerSources.ReviewSnippets {
+						snippet := &aiplatformpb.GroundingChunk_Maps_PlaceAnswerSources_ReviewSnippet{
+							ReviewId:      source.Review,
+							GoogleMapsUri: source.GoogleMapsURI,
+							Title:         source.Title,
+						}
```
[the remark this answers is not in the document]
dpasiukevich: Seems like a hallucination? Given that in CI binary builds and test pass.

session/vertexai/service_test.go
```diff
@@ -0,0 +1,1282 @@
...
+			}
+
+			if got.Session.UserID() != tt.req.UserID {
+				t.Errorf("UserID got: %v, want: %v", got.Session.UserID(), tt.wantErr)
+			}
+
+			if tt.req.SessionID != "" {
+				if got.Session.ID() != tt.req.SessionID {
+					t.Errorf("SessionID got: %v, want: %v", got.Session.ID(), tt.wantErr)
+				}
```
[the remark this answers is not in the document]
dpasiukevich: +1

kdroste-google (approved):

PR #237, opened 2025-11-08
chore(example): context propagation and signal-based cancellation

appleboy:
- Add signal handling to gracefully cancel the context on interrupt (SIGINT)
- Update functions to accept context instead of creating a new one internally
- Pass the main context to MCP transport constructors for improved cancellation support


appleboy:
@dpasiukevich revert the last commit. Thanks for your feedback.

dpasiukevich (approved):
Thanks!

appleboy:
@dpasiukevich can you merge the PR?

dpasiukevich:
@appleboy it should be possible to merge yourself, you only needed to mark comments as resolved. Let me know if that was not the case.

PR #246, opened 2025-11-10
Typo in package comment

zebroc:
Typo in package comment


dpasiukevich:
I've updated the comment with gemini suggestion, which looks pretty nice.

Thanks!

PR #254, opened 2025-11-11
Restructure the REST API packages

rakyll:
- Rename `server/restapi` to `server/adkrest` to match `server/adka2a`.
- Move non-public packages to internal.
- Rename `server/handlers` to `server/controllers` because it contains controllers.
- Name all handlers *Handler per Go convention.
- Move AgentLoader to the agent package.

TODO: Consider moving controllers into an internal package.


rakyll:
PTAL

dpasiukevich:
LGTM!

Let me know if you'd like to update this PR to hide restapi entities and expose only `NewHandler(config *launcher.Config) http.Handler`

rakyll:
I'll revisit this PR to make some changes the hide more of the public surface. I'll introduce something like `NewHandler` as a part of the change.

rakyll:
PTAL

rakyll:
We will follow up with some more changes to the launchers, but this PR is ready for review.

kdroste-google:
adk-docs /examples/go should be updated after we merge: for sure cloud-run and a2a_basic/remote_a2a/check_prime_agent, maybe more

server/adkrest/controllers/eval.go
```diff
@@ -12,4 +12,4 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package handlers
+package controllers
```
kdroste-google: Why do we need this file anyway?

server/adkrest/handler.go
```diff
@@ -34,12 +33,14 @@ func NewHandler(config *launcher.Config) http.Handler {
 	telemetry.AddSpanProcessor(sdktrace.NewSimpleSpanProcessor(adkExporter))
 
 	router := mux.NewRouter().StrictSlash(true)
+	// TODO: Allow taking a prefix to allow customizing the path
+	// where the ADK REST API will be served.
```
kdroste-google, line 37: There are 3 prefixes at the moment: api, ui and a2a. We should either allow to choose all (and handle potential conflicts) or none

kdroste-google (dismissed):
LGTM. One extra file (eval.go) to be deleted if not needed.

server/adkrest/controllers/eval.go
rakyll: Not needed, removed.

server/adkrest/handler.go
rakyll: We should allow each to be served from a custom prefix of choice. Let's follow up with a change after this PR.

rakyll:
Can I have LGTM again? I had to rebase.

PR #256, opened 2025-11-11
fix: Refactor RunAgentRequest field names to snake_case

### LINKED ISSUE #255: Local Development REST API does not work according to documentation.
Documentation here: https://google.github.io/adk-docs/runtime/api-server/#local-testing shows that to use curl to explore local services, you should follow the following routine: 

Create a new session
`curl -X POST http://localhost:8000/apps/my_sample_agent/users/u_123/sessions/s_123 `

Then use that session to make a request:
`curl -X POST http://localhost:8080/api/run \
-H "Content-Type: application/json" \
-d '{
"app_name": "boat_agent",
"user_id": "u_123",
"session_ID": "s_123",
"new_message": {
    "role": "user",
    "parts": [{
    "text": "model: Lucia 40, make: Fontaine Pajot"
    }]
}
}'`

This does not work and throws an error: 
`get session: app_name, user_id, session_id are in fact required, got app_name: "", user_id: "", session_id: ""`

However if I match the json string in the RunAgentRequest struct: 
`type RunAgentRequest struct {
	AppName string `json:"appName"`

	UserId string `json:"userId"`

	SessionId string `json:"sessionId"`

	NewMessage genai.Content `json:"newMessage"`

	Streaming bool `json:"streaming,omitempty"`

	StateDelta *map[string]any `json:"stateDelta,omitempty"`
}`

like so: 
`
curl -X POST http://localhost:8080/api/run \
-H "Content-Type: application/json" \
-d '{
"appName": "boat_agent",
"userId": "u_123",
"SessionID": "s_123",
"newMessage": {
    "role": "user",
    "parts": [{
    "text": "model: Lucia 40, make: Fontaine Pajot"
    }]
}
}'
`
It works. 

So in theory the solution is to alter the name jsonschema on the RunAgentRequest to the expected values. 

Now one could just change the documentation, but this is supposed to match the behavior of the python version of adk, so I think the right move is to do the json schema. I'm going to submit a PR to do that (with this the explanation as to why. )

tpryan:
Updated the JSON tags in the RunAgentRequest struct to use snake_case for consistency with API conventions. Removed the AssertRunAgentRequestRequired method and related imports.

Fixes #255

### DIFF
--- a/server/restapi/models/runtime.go
+++ b/server/restapi/models/runtime.go
@@ -30,7 +30,7 @@ type RunAgentRequest struct {
 }
 
 // Validate checks if the required fields are not zero-ed
-func (req *RunAgentRequest) Validate() error {
+func (req RunAgentRequest) Validate() error {
 	elements := map[string]any{
 		"app_name":    req.AppName,
 		"user_id":     req.UserId,


tpryan:
I removed AssertRunAgentRequestRequired because as far as I can tell it isn't called anywhere.

server/restapi/models/runtime.go
```diff
@@ -15,38 +15,14 @@
...
-
-	NewMessage genai.Content `json:"newMessage"`
-
-	Streaming bool `json:"streaming,omitempty"`
-
-	StateDelta *map[string]any `json:"stateDelta,omitempty"`
-}
-
-// AssertRunAgentRequestRequired checks if the required fields are not zero-ed
-func (req RunAgentRequest) AssertRunAgentRequestRequired() error {
```
dpasiukevich, line 38: Could I only ask to keep and actually enable the validation for the request?

Something along the lines:
1. renaming AssertRunAgentRequestRequired -> Validate
2. changing type `NewMessage genai.Content` to pointer -> `NewMessage *genai.Content` for easier validation.
3. Delete `IsZeroValue` and just use a check `... != ""` and `... != nil`.

Early validation IMO is quite a good thing.

Let me know if you'd like to do these changes.

Thanks!

tpryan: Seems reasonable.  I left IsZeroValue, as that seemed preferable to adding if statements since IsZeroValue solves the problem neatly. 

I left NewMessage as a non-pointer as changing it broke a whole bunch of tests.

kdroste-google:
I'm afraid it would break ADK WebUI which is calling `api/run_sse` with payload like: 
```json
{"appName":"weather_time_agent","userId":"user","sessionId":"................","newMessage":{"role":"user","parts":[{"text":"London"}]},"streaming":false,"stateDelta":null}
```

tpryan:
Well then we need to decide which pattern to follow, and document because as it is, new devs can't use this feature.

rakyll:
What is the ADK API standard style? We need to be 100% compatible with the implementation in ADK Python.

dpasiukevich:
> What is the ADK API standard style? We need to be 100% compatible with the implementation in ADK Python.

This one https://github.com/google/adk-go/issues/255#issuecomment-3516317834

rakyll:
camelCase is the official ADK API style, see https://github.com/google/adk-go/issues/255#issuecomment-3529831091 for more details.

PR #259, opened 2025-11-11
Fix #257: Return http.Handler instead of SetupRouter in server/restapi/web

### LINKED ISSUE #257: server/web: Return an http.Handler instead of providing a SetupRouter
Go usually prefer initiating and registering `http.Handler` objects from packages that provide server implementations. Then, any `net/http` package user can directly register them as handlers. Currently we assume users relying on `mux`. It's probably much better to just initiate an `http.Handler` and allow users to register it at their server.

Arjunmehta312:
## Description

Fixes #257

This PR addresses the issue raised by @rakyll to return an `http.Handler` instead of requiring users to pass in a `*mux.Router`. This change follows Go best practices by decoupling the REST API implementation from a specific router library (gorilla/mux).

### 1. Added `NewHandler()` function (`server/restapi/web/server.go`)
- Created `NewHandler(config *launcher.Config) http.Handler` that returns a standard `http.Handler`
- This is the preferred and only way to integrate ADK REST API with any HTTP server
- Users can register the handler with any router: standard `net/http.ServeMux`, chi, echo, gin, etc.
- Keeps the internal `setupRouter` helper for package-internal use

### 2. Removed deprecated `SetupRouter()` 
- Per maintainer feedback, the public `SetupRouter()` function has been removed entirely
- This is a breaking change but aligns with the goal of decoupling from gorilla/mux
- All users should migrate to `NewHandler()`

### 3. Updated internal usage (`cmd/launcher/web/api/api.go`)
- Refactored `SetupSubrouters()` to use `NewHandler()` internally
- Applied CORS middleware before registration
- Uses `http.StripPrefix` for proper path handling
- Updated to use `*launcher.Config` parameter type

### 4. Added example (`examples/restapi/main.go`)
- Demonstrates usage with standard `net/http.ServeMux`
- Shows how to compose the handler with other endpoints (health check)
- Proves the framework-agnostic approach works
- Uses `launcher.Config` type (reflecting recent upstream changes)

## Benefits

**Decoupled**: Not tied to gorilla/mux or any specific router  
**Flexible**: Works with any HTTP server implementation  
**Standard**: Returns `http.Handler`, following Go conventions  
**Clean API**: Single clear entry point (`NewHandler`)

## Testing

- All packages build successfully: `go build ./...`
- Example demonstrates new usage pattern
- Updated to use `launcher.Config` type throughout

## Migration Guide

**New approach:**

```go
handler := web.NewHandler(config)
mux.Handle("/api/", http.StripPrefix("/api", handler))

### DIFF
--- a/examples/restapi/main.go
+++ b/examples/restapi/main.go
@@ -91,4 +91,3 @@ func main() {
 		log.Fatalf("Server failed: %v", err)
 	}
 }
-

### CODE REVIEW

File: server/restapi/web/server.go [hunk 1]
```diff
@@ -26,8 +28,32 @@ import (
...
+	)
+	return router
+}
+
+// SetupRouter initiates mux.Router with ADK REST API routers.
+//
+// Deprecated: Use NewHandler instead. This function assumes the caller is using
+// gorilla/mux and tightly couples the implementation to that router. NewHandler
+// returns a standard http.Handler that can be used with any HTTP server or router.
 func SetupRouter(router *mux.Router, routerConfig *launcher.Config) *mux.Router {
```
dpasiukevich: Let's just remove this func.

examples/restapi/main.go
```diff
@@ -0,0 +1,91 @@
...
+		Tools: []tool.Tool{
+			geminitool.GoogleSearch{},
+		},
+	})
+	if err != nil {
+		log.Fatalf("Failed to create agent: %v", err)
+	}
+
+	// Configure the ADK REST API
+	config := &adk.Config{
```
dpasiukevich: Could you update the example here: `adk.Config` got moved to `launcher.Config`.

server/restapi/web/server.go
Arjunmehta312: Okay will do.

examples/restapi/main.go
Arjunmehta312: Okay will do.

Arjunmehta312:
Made the required changes @dpasiukevich 
Please review and merge if proper.
Thanks and regards.

dpasiukevich:
@Arjunmehta312 could you also update the example to pass the linter and we're good to merge.

Arjunmehta312:
Sure, will update the example to pass the linter as you mentioned

Arjunmehta312:
@dpasiukevich done please check it out 😊

Arjunmehta312:
Also please note that, as per maintainer feedback, the public SetupRouter() function has been removed entirely
This is a breaking change but aligns with the goal of decoupling from gorilla/mux

All okay?

dpasiukevich (approved):

Arjunmehta312:
Thanks @dpasiukevich , sorry for the multiple reviews. 

Happy to contribute!

PR #261, opened 2025-11-11
Added initial llms.txt

ivanmkc:
Initial llms.txt which is a condensed version of adk-docs, specific for Golang.

This is an initial, baseline version which will be benchmarked against future versions (or not using a llms.txt).

Equivalent to: https://github.com/google/adk-python/blob/main/llms-full.txt

This will need updating when the API changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,18 @@ To add ADK Go to your project, run:
 go get google.golang.org/adk
 ```
 
+## 📖 `llms.txt`
+
+[llms.txt](https://github.com/google/adk-go/blob/main/llms.txt) provides a comprehensive, code-first documentation summary of the Go Agent Development Kit (ADK). It is a distilled version of the official Go documentation found in [adk-docs](https://github.com/google/adk-docs/), specifically curated and formatted for optimal use with AI agents like Gemini-CLI. It includes detailed explanations and runnable code examples for core agent concepts, multi-agent systems, advanced features like context management and callbacks, and Agent2Agent (A2A) communication.
+
+**Use with Gemini CLI:**
+
+You can easily reference [llms.txt](https://github.com/google/adk-go/blob/main/llms.txt) in your Gemini CLI interactions to provide context for code generation, explanations, or debugging. For example:
+
+```bash
+Plan and implement an agent to review customer issues and generate a report. Use the ADK Go framework, referring to @llms.txt for sample code.
+```
+
 ## 📄 License
 
 This project is licensed under the Apache 2.0 License - see the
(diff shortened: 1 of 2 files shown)


ivanmkc:
I'll address some of the AI-comments. The llms.txt isn't supposed to be runnable code but I'll add more err checks so the AI-user of llms.txt isn't biased to write code that ignores errors.

ivanmkc:
How do I get @gemini-code-assist to re-review?

ivanmkc:
Let me review the comments again.

rakyll:
We are planning some breaking changes to the API surface. Would it be possible to come back to this after some of those changes land?

ivanmkc:
@rakyll Of course, ping me when that happens. The eventual goal is that llms.txt will update with each breaking version.

hyangah:
check go/llmstxt.

llms.txt has no effect on SEO nor do any of the mainstream consumer LLM systems use it.

verdverm:
Since

1. The python version has llms-full.txt and a "lite" llms.txt
2. This is the equivalent of full

Should we...?
1. name it to match
2. have both versions

wolo-lab:
Thank you for putting this together, and apologies that it sat this long — that's on us, not on the PR.

Closing it in favor of the hosted `llms.txt`, which now covers Go.

Since this PR was opened, adk-python retired its in-repo copy. [`llms-full.txt`](https://github.com/google/adk-python/blob/c66dc1dec44010811b073067136831bdbfee394a/llms-full.txt) is now a stub pointing at the docs site, on the grounds that the file is "automatically generated and hosted on the ADK documentation site". [adk.dev/llms.txt](https://adk.dev/llms.txt) and [adk.dev/llms-full.txt](https://adk.dev/llms-full.txt) are generated from adk-docs and already carry the Go pages — 319 of the 324 `google.golang.org/adk` import paths in the hosted full version are v2. A hand-maintained Go copy here would compete with that rather than complement it.

The content has also gone stale beyond what a rebase would fix. Against [`main` at 70e2f03](https://github.com/google/adk-go/tree/70e2f03c1f2c221018c0d6a22ee8e48559dd7523):

- **Wrong major version.** The file is headed "Valid for ADK Go version: v0.1.0", and all 45 of its import paths are v1 (`google.golang.org/adk/...`). The module is `google.golang.org/adk/v2`, so none of the samples compile as written.
- **Packages that no longer exist.** `server/restapi/services` and `cmd/launcher/adk` have both been removed.
- **Manual upkeep.** 846 lines needing a pass on every API change, as the description notes.

The README half is worth keeping, so I've opened #1402 with a short pointer to the hosted files — the idea and most of the wording are yours.

PR #262, opened 2025-11-11
Delete LICENSE

nicotrie25-gith:


ivanmkc:
We don't foresee a need to delete the license.

PR #269, opened 2025-11-13
test(gemini-model): add unit tests to cover some errors

danielfs:
**Tests included**

- project and API key are mutually exclusive in the client initializer
- generate with empty model name
- maybe append user content when length is zero
- maybe append user content when there is no user role
- Name() method

### DIFF
--- a/model/gemini/errors.go
+++ b/model/gemini/errors.go
@@ -1,0 +1,5 @@
+package gemini
+
+import "errors"
+
+var ErrEmptyModelName = errors.New("model name cannot be empty")
--- a/model/gemini/gemini_test.go
+++ b/model/gemini/gemini_test.go
@@ -15,9 +15,11 @@
 package gemini
 
 import (
+	"errors"
 	"fmt"
 	"iter"
 	"net/http"
+	"os"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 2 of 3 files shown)


model/gemini/gemini_test.go
```diff
@@ -32,6 +32,11 @@ import (
 
 //go:generate go test -httprecord=testdata/.*\.httprr
 
+const (
+	systemInstructionText  = "Handle the requests as specified in the System Instruction."
+	continueProcessingText = "Continue processing previous requests as instructed. Exit or provide a summary if no more outputs are needed."
+)
```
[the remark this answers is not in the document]
Deshaunmr: ](https://developers.google.com/gemini-code-assist/docs/supported-languages#supported_ides)](https://developers.google.com/gemini-code-assist/docs/supported-languages#supported_ides)

model/gemini/gemini_test.go
```diff
@@ -187,6 +192,127 @@ func TestModel_TrackingHeaders(t *testing.T) {
...
+		t.Fatalf("expected 1 error, got %d", len(errs))
+	}
+
+	err = errs[0]
+	if err == nil {
+		t.Fatal("Model.Generate() expected error due to empty model name, got nil")
+	}
+	if want := "model is empty"; !strings.Contains(err.Error(), want) {
+		t.Fatalf("Model.Generate() error = %q, want to contain %q", err, want)
+	}
```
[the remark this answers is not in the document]
Deshaunmr: > medium
> 
> The logic for retrieving and checking the error can be simplified. Since a non-streaming `GenerateContent` call is designed to yield exactly one result, you don't need to collect errors in a slice and then check its length. You can simplify the code by iterating once and checking the error directly. This makes the test's intent clearer and more concise.
> 
> ```go
>  var err error
>  for _, e := range newModel.GenerateContent(t.Context(), req, false) {
>   err = e
>  }
> 
>  if err == nil {
>   t.Fatal("Model.Generate() expected error due to empty model name, got nil")
>  }
>  if want := "model is empty"; !strings.Contains(err.Error(), want) {
>   t.Fatalf("Model.Generate() error = %q, want to contain %q", err, want)
>  }
> ```

@danielfs

model/gemini/errors.go
```diff
@@ -0,0 +1,5 @@
+package gemini
+
+import "errors"
+
+var errEmptyModelName = errors.New("model name cannot be empty")
```
kdroste-google: Maybe we should have ErrEmptyModelName instead? It would allow users to check for this error easier. WDYT?

model/gemini/gemini_test.go
```diff
@@ -188,6 +188,89 @@ func TestModel_TrackingHeaders(t *testing.T) {
 	})
 }
 
+func TestModel_Name(t *testing.T) {
```
kdroste-google, line 192: Does it make sense to test such an assignment?

model/gemini/gemini_test.go
```diff
@@ -188,6 +188,89 @@ func TestModel_TrackingHeaders(t *testing.T) {
...
+func TestModel_NewModelError(t *testing.T) {
+	_, err := NewModel(t.Context(), "gemini-2.5-flash", &genai.ClientConfig{
+		// project and API key are mutually exclusive in the client initializer
+		APIKey:  "api-key",
+		Project: "project-id",
+	})
+	if err == nil {
+		t.Fatal("NewModel() expected error due to project and API key are mutually exclusive, got nil")
+	}
+	if want := "mutually exclusive"; !strings.Contains(err.Error(), want) {
```
kdroste-google: Looking at genai / client.go it is more complex, and depends on the env variables. You can control that using explicit envVarProvider in ClientConfig. Imho it would make sense. WDYT?

model/gemini/constants.go
```diff
@@ -0,0 +1,6 @@
+package gemini
+
+const (
+	systemInstructionText  = "Handle the requests as specified in the System Instruction."
```
kdroste-google: There's only one place where those are actually used. I would consider moving those consts there instead of creating a new file.

model/gemini/errors.go
danielfs: I agree. My first thought was that I wouldn’t need to make the `errEmptyModelName` variable public until it became useful in another package.

model/gemini/gemini_test.go
danielfs: I think so. It is just to cover the function `func (m *geminiModel) Name() string`

model/gemini/gemini_test.go
danielfs: It makes sense, but the code does perform exclusivity checks before backend logic, the env vars have not been copied into cc at that time.

model/gemini/constants.go
danielfs: I agree.

wolo-lab:
Thank you for this, and for spotting that the NewModel error paths and maybeAppendUserContent had no coverage. That was accurate when you wrote it and it is still true today.

We are closing it because the production changes it carries no longer match main. The empty-contents placeholder your table test asserts was deliberately removed in the v2 release (#1109), and the ErrEmptyModelName guard would reject constructing the model with an empty name and supplying it per request through req.Model, which is a supported configuration.

The coverage gap itself is real and we will pick it up separately, written against main's current behavior.

Apologies that this drew review comments and then stalled for months without anyone giving you a decision either way. We are working the PR queue much more actively now.

PR #270, opened 2025-11-13
Allow function tool handlers to return errors

### LINKED ISSUE #260: tool/functiontool: not returning error
What's the motivation for making functiontool callback not returning an error? Is it a bug of feature?
A functiontool can well fail due to programming errors, or unavailable services, etc.
I had to do this ugly mess just to pipe an error around  functiontool. Shouldn't the callback just return `(Results, error)` as per standard Go convention?

```
func (t *funcTool[Args, Results]) create(*createContext) (tool.Tool, error) {
	cfg := functiontool.Config{
		Name:        t.Name,
		Description: t.Description,
	}
	toolFn := func(ctx tool.Context, args Args) Results {
		ectx := ctx.Value(executeContextKey).(Context)
		res, err := t.Func(ectx, args)
		if err != nil {
			panic(toolError{err})
		}
		return res
	}
	ft, err := functiontool.New(cfg, toolFn)
	if err != nil {
		return nil, err
	}
	return &funcToolWrapper{ft}, nil
}

type funcToolWrapper struct {
	tool.Tool
}

type toolError struct {
	err error
}

func (f *funcToolWrapper) Run(ctx tool.Context, args any) (res map[string]any, resErr error) {
	defer func() {
		if err := recover(); err != nil {
			if te, ok := err.(toolError); ok {
				resErr = te.err
				return
			}
			panic(err)
		}
	}()
	type toolRunner interface {
		Run(tool.Context, any) (map[string]any, error)
	}
	return f.Tool.(toolRunner).Run(ctx, args)
}

func (f *funcToolWrapper) ProcessRequest(ctx tool.Context, req *model.LLMRequest) error {
	type requestProcessor interface {
		ProcessRequest(ctx tool.Context, req *model.LLMRequest) error
	}
	return f.Tool.(requestProcessor).ProcessRequest(ctx, req)
}
```

rakyll:
This change re-introduces error as a possible return value for function tools. The previous removal of this feature led to poor error-handling patterns, as functions often need to signal failures from validation, I/O, or network calls. Restoring the ability to return errors allows for conventional error handling.

Fixes #260.

### DIFF
--- a/tool/functiontool/function.go
+++ b/tool/functiontool/function.go
@@ -46,7 +46,7 @@ type Config struct {
 
 // Func represents a Go function that can be wrapped in a tool.
 // It takes a tool.Context and a generic argument type, and returns a generic result type.
-type Func[TArgs, TResults any] func(tool.Context, TArgs) TResults
+type Func[TArgs, TResults any] func(tool.Context, TArgs) (TResults, error)
 
 // New creates a new tool with a name, description, and the provided handler.
 // Input schema is automatically inferred from the input and output types.
@@ -140,7 +140,10 @@ func (f *functionTool[TArgs, TResults]) Run(ctx tool.Context, args any) (map[str
 	if err != nil {
 		return nil, err
 	}
-	output := f.handler(ctx, input)
+	output, err := f.handler(ctx, input)
+	if err != nil {
+		return nil, err
+	}
 	resp, err := typeutil.ConvertToWithJSONSchema[TResults, map[string]any](output, f.outputSchema)
 	if err == nil { // all good
 		return resp, nil
--- a/tool/functiontool/function_test.go
+++ b/tool/functiontool/function_test.go
@@ -45,8 +45,8 @@ func ExampleNew() {
 		Sum int `json:"sum"` // the sum of two integers
 	}
 
-	handler := func(ctx tool.Context, input SumArgs) SumResult {
-		return SumResult{Sum: input.A + input.B}
+	handler := func(ctx tool.Context, input SumArgs) (SumResult, error) {
+		return SumResult{Sum: input.A + input.B}, nil
 	}
 	sumTool, err := functiontool.New(functiontool.Config{
 		Name:        "sum",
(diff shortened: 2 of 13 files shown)


tool/functiontool/function_test.go
```diff
@@ -89,15 +89,15 @@ func TestFunctionTool_Simple(t *testing.T) {
...
 
-	weatherReport := func(ctx tool.Context, input Args) Result {
+	weatherReport := func(ctx tool.Context, input Args) (Result, error) {
 		city := strings.ToLower(input.City)
 		if ret, ok := resultSet[city]; ok {
-			return ret
+			return ret, nil
 		}
 		return Result{
 			Status: "error",
```
hyangah: you can repurpose this test to return an error (instead of using the Status field in Result).

tool/functiontool/function.go
```diff
@@ -140,7 +140,10 @@ func (f *functionTool[TArgs, TResults]) Run(ctx tool.Context, args any) (map[str
 	if err != nil {
 		return nil, err
 	}
-	output := f.handler(ctx, input)
+	output, err := f.handler(ctx, input)
+	if err != nil {
+		return nil, err
```
hyangah, line 145: Now we can remove the TODO :-) in https://github.com/google/adk-go/blob/cffe3b0d6b8b45505a67f658d9cf941849d14b10/internal/llminternal/base_flow.go#L422

tool/functiontool/function_test.go
rakyll: Done.

baptmont:
Shouldn't we send the err to the llm when creating the summarization event?
For example, modifying the [examples/tools/multipletools/main.go ](https://github.com/google/adk-go/blob/a4db9abcf53adabe84e6b2ca84d5c3cb95c65d76/examples/tools/multipletools/main.go#L68) if we have the function
```
	handler := func(ctx tool.Context, input Input) (*Output, error) {
		if input.LineCount < 3 {
			return nil, fmt.Errorf("invalid number of line count, the number of lines cannot be lower than 3 lines")
		}
		return &Output{
			Poem: strings.Repeat("A line of a poem,", input.LineCount) + "\n",
		}, nil
	}
```
The agent will never notify the user of the specific err:
 
Screenshot 2025-11-13 at 10 36 54

dpasiukevich (approved):

dpasiukevich:
It seems as @baptmont found we need to return an error string to llm `err.Error()`, otherwise if it's a pointer it won't read the error message.

dpasiukevich:
> It seems as @baptmont found we need to return an error string to llm `err.Error()`, otherwise if it's a pointer it won't read the error message.

Opened https://github.com/google/adk-go/issues/278 for that

rakyll:
I wonder whether we should allow returning a response and an error at the same time. Then, the functiontool builder can choose between user visible error messages with their custom response type. From a quality perspective, we need to align the behavior with ADK Python. Let's follow up on the issue @dpasiukevich has created.

hyangah:
We are already sending err to llm. Instructions need to tell llm what to do.

PR #273, opened 2025-11-13
Remove Vertex AI session service

rakyll:
Not sure if this package is ready to be released yet. It uses a hardcoded "session-id" in return values. 
Let's release this package once we can provide an identical implementation to Python: https://github.com/google/adk-python/blob/main/src/google/adk/sessions/vertex_ai_session_service.py


dpasiukevich (approved):
VertexAI session service is WIP https://github.com/google/adk-go/pull/235

IMO we can just wait until https://github.com/google/adk-go/pull/235 is merged. But if you'd like to remove this unfinished code for a couple of days, it's also ok.

rakyll:
Let's merge this and use #235 to fix it for real. This isn't ready for use and I'm worried that someone may corrupt data.

PR #288, opened 2025-11-14
Added sample for setting user_id in a2a

kdroste-google:


.gitignore
```diff
@@ -20,3 +20,4 @@ profile.cov
 # Go workspace file
 go.work
 go.work.sum
+.vscode/launch.json
```
dpasiukevich: Let's adhere to

```
# .gitignore is restricted to the artifacts produced by go build and test.
# for personal setup you can use personal .gitignore: `git config --global core.excludesfile ~/.gitignore
```

examples/web/main.go
```diff
@@ -47,6 +48,20 @@ func saveReportfunc(ctx agent.CallbackContext, llmResponse *model.LLMResponse, l
...
+// Intercept implements a2asrv.RequestContextInterceptor.
+func (a *AuthInterceptor) Intercept(ctx context.Context, reqCtx *a2asrv.RequestContext) (context.Context, error) {
+	if reqCtx.Metadata == nil {
+		reqCtx.Metadata = make(map[string]any)
+	}
+	reqCtx.Metadata["adk_user_id"] = "user"
+	return ctx, nil
+}
+
+var _ a2asrv.RequestContextInterceptor = (*AuthInterceptor)(nil)
```
dpasiukevich: not needed. `a2asrv.WithRequestContextInterceptor` already requires the interface as a parameter type.

examples/web/main.go
```diff
@@ -47,6 +48,20 @@ func saveReportfunc(ctx agent.CallbackContext, llmResponse *model.LLMResponse, l
...
 
+type AuthInterceptor struct {
+}
+
+// Intercept implements a2asrv.RequestContextInterceptor.
+func (a *AuthInterceptor) Intercept(ctx context.Context, reqCtx *a2asrv.RequestContext) (context.Context, error) {
+	if reqCtx.Metadata == nil {
+		reqCtx.Metadata = make(map[string]any)
+	}
+	reqCtx.Metadata["adk_user_id"] = "user"
```
yarolegovich: Metadata here is taken from SendMessageRequest, it's better to use `WithCallInterceptor` for auth:
```go
type AuthInterceptor struct {
	a2asrv.PassthroughCallInterceptor
}

func (AuthInterceptor) Before(ctx context.Context, callCtx *CallContext, req *Request) (context.Context, error) {
	callCtx.User = &AuthenticatedUser{UserName: "user"}
	return ctx, nil
}
```
and modify adka2a to use it:
```go
if callCtx, ok := a2asrv.CallContextFrom(ctx); ok && callCtx.User.Name() != "" {
   userID = callCtx.User.Name()
}
```

.gitignore
kdroste-google: ACK

examples/web/main.go
kdroste-google: Removed

examples/web/main.go
kdroste-google: Thanks for hint!

server/adka2a/metadata.go
```diff
@@ -33,10 +34,17 @@ type invocationMeta struct {
 	eventMeta map[string]any
 }
 
-func toInvocationMeta(config ExecutorConfig, reqCtx *a2asrv.RequestContext) invocationMeta {
+func toInvocationMeta(config ExecutorConfig, ctx context.Context, reqCtx *a2asrv.RequestContext) invocationMeta {
```
dpasiukevich: context.Context should be a first parameter as a convention. Could you update it here?

CONTRIBUTING.md
```diff
@@ -15,6 +15,9 @@ We'd love to accept your patches and contributions to this project.
     -   [Unit Tests](#unit-tests)
     -   [Manual End-to-End (E2E) Tests](#manual-end-to-end-e2e-tests)
     -   [Documentation](#documentation)
+    -   [Alignment with adk-python](#alignment-with-adk-python)
+
+
```
dpasiukevich: nit: maybe let's remove these 2 extra newlines?

server/adka2a/metadata.go
kdroste-google: generally I agree, but in this case it doesn't have this meaning.  It's just a data source.

CONTRIBUTING.md
kdroste-google: Fixed

server/adka2a/metadata.go
dpasiukevich: `context.Context` is the data source by definition. It's used to pass request-scoped data like timeouts, user_id, session_id etc.

server/adka2a/metadata.go
```diff
@@ -33,10 +34,17 @@ type invocationMeta struct {
...
 }
 
-func toInvocationMeta(config ExecutorConfig, reqCtx *a2asrv.RequestContext) invocationMeta {
+func toInvocationMeta(ctx context.Context, config ExecutorConfig, reqCtx *a2asrv.RequestContext) invocationMeta {
 	// TODO(yarolegovich): update once A2A provides auth data extraction from Context
 	userID, sessionID := "A2A_USER_"+reqCtx.ContextID, reqCtx.ContextID
 
+	// override userID if set in the call context
+	if callCtx, ok := a2asrv.CallContextFrom(ctx); ok {
+		if callCtx.User != nil && callCtx.User.Name() != "" {
```
yarolegovich, line 43: user is always set, by default to unauthenticated user

yarolegovich (approved):

server/adka2a/metadata.go
kdroste-google: Changed order

server/adka2a/metadata.go
kdroste-google: ACK, but it doesn't hurt to check :)

PR #291, opened 2025-11-14
Add VertexAI RAG retrieval tool implementation

nktks:
- This PR adds Vertex AI RAG Engine tool https://google.github.io/adk-docs/tools/built-in-tools/#vertex-ai-rag-engine.
- This is based on [adk-python implementation](https://github.com/google/adk-python/blob/fd33610e967ad814bc02422f5d14dae046bee833/src/google/adk/tools/retrieval/vertex_ai_rag_retrieval.py)
  - However I omitted the implementation of functional tool implementation with `google.cloud.aiplatform.v1.VertexRagService.RetrieveContexts` call for Gemini 1.x support,  because Gemini 1.x is already retired before adk-go had released. [ref](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/model-versions#retired-models)


jjsasha63:
Thanks for this, and sorry it sat so long.

Since September 2025 main has had a generic adapter that covers the built-in Vertex RAG tool in one call, so a dedicated package is not needed:

```go
geminitool.New("retrieve_rag_documentation",
    "Retrieve documentation and reference materials from the RAG corpus",
    &genai.Tool{Retrieval: &genai.Retrieval{VertexRAGStore: ragStore}})
```

That is exactly what `addBuiltInRAGTool` does here. I ran the two side by side and they produce an identical `LLMRequest`. The branch also predates the move to the `/v2` module path and to `agent.Context`, so it no longer builds against main.

Closing for now. But there is a version of this I would like to see, and it is the part you dropped.

`geminitool.New` only works with Gemini. The native `Retrieval` tool is rejected by the other backends — `model/openaimodel` returns "non-function tools are not supported". A function tool that calls `VertexRagService.RetrieveContexts` directly would work with every model adk-go supports. That is a real capability gap rather than a wrapper, and I will reopen for it. Your reasoning for dropping it was sound at the time, since Gemini 1.x is retired, but non-Gemini models give it a new justification.

A short example under `examples/tools` using `geminitool.New` would also be very welcome. Main documents Vertex RAG nowhere, and your example file is most of the way there already.

PR #299, opened 2025-11-16
fix: serialize tool errors as strings in FunctionResponse

Omzee15:
Tool errors were previously serialized as error interfaces which marshal to empty JSON objects ({}), making it impossible for LLMs to understand what went wrong. This fix converts errors to strings using err.Error() before including them in the FunctionResponse.

Changes:
- Modified callTool() in base_flow.go to call .Error() on error objects
- Added comprehensive tests for error serialization scenarios
- Covers tool execution errors, before/after callback errors

Fixes issue where LLMs received error: {} instead of meaningful error messages.

### DIFF
--- a/internal/llminternal/base_flow.go
+++ b/internal/llminternal/base_flow.go
@@ -417,17 +417,17 @@ func (f *Flow) callTool(tool toolinternal.FunctionTool, fArgs map[string]any, to
 	// If the result is present, it will be used instead of calling the actual tool.
 	result, err := f.invokeBeforeToolCallbacks(tool, fArgs, toolCtx)
 	if err != nil {
-		return map[string]any{"error": fmt.Errorf("BeforeToolCallback failed: %w", err)}
+		return map[string]any{"error": fmt.Errorf("BeforeToolCallback failed: %w", err).Error()}
 	}
 	if result == nil {
 		result, err = tool.Run(toolCtx, fArgs)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


apzuk3:
I created this custom error type to get around this issue..

```go
type ToolError struct {
	Message string
}

// Error implements the error interface.
func (e ToolError) Error() string {
	return e.Message
}

// MarshalJSON implements json.Marshaler to serialize as a string.
func (e ToolError) MarshalJSON() ([]byte, error) {
	return json.Marshal(e.Message)
}

// NewToolError creates a new ToolError with a formatted message.
func NewToolError(format string, args ...any) ToolError {
	return ToolError{Message: fmt.Sprintf(format, args...)}
}

// WrapToolError wraps an existing error as a ToolError.
func WrapToolError(msg string, err error) ToolError {
	if err == nil {
		return ToolError{Message: msg}
	}
	return ToolError{Message: fmt.Sprintf("%s: %v", msg, err)}
}

```

karolpiotrowicz:
Thanks for this. Tool-error serialization already landed in `main`: `callTool` in `internal/llminternal/base_flow.go` now returns `map[string]any{"error": err.Error()}` (sibling paths use `err.Error()` too), so errors no longer serialize as `{}`. Superseded, so I'm closing it. If any of your test cases cover something we're missing, a small separate PR would be welcome.

PR #300, opened 2025-11-16
Feat/session compaction

raphaelmansuy:


rakyll:
How does this change relate to ADK's official compaction capabilities? An experimental implementation is at https://github.com/google/adk-python/blob/f7f6837fdede6a7670ceedfbeac07f926506f7f6/src/google/adk/apps/app.py#L63 and we will port it to ADK Go once it's more mature.

raphaelmansuy:
> How does this change relate to ADK's official compaction capabilities? An experimental implementation is at https://github.com/google/adk-python/blob/f7f6837fdede6a7670ceedfbeac07f926506f7f6/src/google/adk/apps/app.py#L63 and we will port it to ADK Go once it's more mature.

Yes it is based on this experimental implementation

jinzhongjia:
Oh, great

Since this is a new library, could we import the feature first and then mark it as an experimental, unstable API for use?

karolpiotrowicz:
Thanks for the substantial work here. We want to keep session compaction coherent across the ADK implementations rather than land a separate, divergent version in this repo that we'd later have to reconcile (this one also still has the unwired-filter/race issues raised in review). Closing on that basis — we'd genuinely welcome your input as we shape the shared approach.

PR #304, opened 2025-11-17
feat: pass event actions through a2a

yarolegovich:
The changes are mostly test updates.

* If `Actions.Escalate` was set on a `session.Event` it will be passed to callers in the terminal task status update `Metadata`.
* `RemoteAgent` will set it back as `Actions.Event` (handled in `adka2a.ToSessionEvent`).

### DIFF
--- a/server/adka2a/events.go
+++ b/server/adka2a/events.go
@@ -276,8 +276,6 @@ func toEventActions(event a2a.Event) session.EventActions {
 	meta := event.Meta()
 	var result session.EventActions
 	result.Escalate, _ = meta[metadataEscalateKey].(bool)
-	if transferToAgent, ok := meta[metadataTransferToAgentKey].(string); ok {
-		result.TransferToAgent = transferToAgent
-	}
+	result.TransferToAgent, _ = meta[metadataTransferToAgentKey].(string)
 	return result
 }
--- a/server/adka2a/metadata.go
+++ b/server/adka2a/metadata.go
@@ -88,7 +88,7 @@ func toEventMeta(meta invocationMeta, event *session.Event) (map[string]any, err
 }
 
 func setActionsMeta(meta map[string]any, actions session.EventActions) map[string]any {
-	if actions.TransferToAgent == "" && !actions.Escalate {
+	if actions.TransferToAgent == "" && !actions.Escalate { // if meta was nil, it should remain nil
 		return meta
 	}
 	if meta == nil {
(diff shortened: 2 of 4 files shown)


server/adka2a/metadata.go
```diff
@@ -86,3 +86,19 @@ func toEventMeta(meta invocationMeta, event *session.Event) (map[string]any, err
 
 	return result, nil
 }
+
+func setActionsMeta(meta map[string]any, actions session.EventActions) map[string]any {
+	if actions.TransferToAgent == "" && !actions.Escalate {
+		return meta
+	}
```
dpasiukevich: nit: maybe let's remove?

This check looks like a duplicate of the logic below it in this func (same cond checks).

server/adka2a/processor.go
```diff
@@ -132,6 +139,13 @@ func (p *eventProcessor) makeTaskFailedEvent(cause error, event *session.Event)
 	return toTaskFailedUpdateEvent(p.reqCtx, cause, meta)
 }
 
+func (p *eventProcessor) updateTerminalActions(event *session.Event) {
+	p.terminalActions.Escalate = p.terminalActions.Escalate || event.Actions.Escalate
+	if p.terminalActions.TransferToAgent == "" {
+		p.terminalActions.TransferToAgent = event.Actions.TransferToAgent
```
dpasiukevich, line 145: Curious what the behavior should be if multiple events have `event.Actions.TransferToAgent` but with a different agent

Shouldn't the newer event have precedence by overwriting? I assume these events represent one turn.

server/adka2a/events.go
```diff
@@ -263,3 +271,13 @@ func toGenAIRole(role a2a.MessageRole) genai.Role {
...
 	}
 }
+
+func toEventActions(event a2a.Event) session.EventActions {
+	meta := event.Meta()
+	var result session.EventActions
+	result.Escalate, _ = meta[metadataEscalateKey].(bool)
+	if transferToAgent, ok := meta[metadataTransferToAgentKey].(string); ok {
+		result.TransferToAgent = transferToAgent
+	}
```
dpasiukevich: optional: maybe shorten this code to be similar to the line above?
```suggestion
	if transferToAgent, _ = meta[metadataTransferToAgentKey].(string)
```

server/adka2a/metadata.go
yarolegovich: it is here because if there's no transfer or escalation and meta was nil we want it to remain nil, will leave a comment

server/adka2a/processor.go
yarolegovich: you're right

PR #310, opened 2025-11-18
feat: add cn llms and documents

sjzsdu:


dpasiukevich:
@sjzsdu thanks for contribution!

Before we'd start the PR review, could you fix 2 main issues:
1. The imports got changed to `github.com/sjzsdu/adk-go`. It should be `google.golang.org/adk`.
2. Could you translate/remove comments in Chinese to English?

wolo-lab:
Thank you for the work here, and for wanting to make these providers easy to reach from Go. That goal is a good one.

Since this was opened, main gained model/openaimodel (#1178), which exposes a BaseURL field for OpenAI-compatible endpoints. DeepSeek, Qwen, Zhipu, Moonshot, SiliconFlow and Ollama all serve OpenAI-compatible APIs, so each is reachable today by setting that one field. adk-python takes the same approach of one adapter and many base URLs, and we would rather not own a separate package and a model-name constant list per vendor — those constants go stale faster than anything else in the tree.

The two changes asked for in November 2025, restoring the google.golang.org/adk module path and translating the comments to English, were not made, and the CLA is still unsigned, so the branch cannot be merged as it stands.

If you would like to contribute the ergonomics part, a short documentation section showing openaimodel configured for each of these providers would be very welcome.

PR #242, opened 2025-11-10
Add support for the OpenAI API

levigross:
This PR enables basic support for OpenAI models (and endpoints that expose a OpenAI API compatible API interface).

I am stressing on *basic* support because we will leave tool calling to the next PR :)

PR #314, opened 2025-11-18
create openAi adapter 

babasha:


babasha:
Fixed  / Addressed in latest commit

karolpiotrowicz:
Hi @babasha — thanks for this, and for the clear effort behind it.

We're going to close this one. Two main reasons: the `cla/google` check is still
failing (so it can't be merged as-is), and the PR is much broader than a model
adapter (it also adds an unrelated DuckDuckGo tool, editor/config files, setup
docs, and provider-specific live tests). There are also two more narrowly-scoped
PRs for the same feature (#242, #342) that we'd like to consolidate effort around.

If you'd like to keep contributing, the best next step is to sign the CLA and jump
in on #242 / #342 (review or testing against real providers). Thanks again!

PR #318, opened 2025-11-19
Add support of ReAct and builtin planner

git-hulk:
This PR introduces ReAct and builtin planner for adk-go to align with adk-python.

### DIFF
--- a/internal/llminternal/other_processors.go
+++ b/internal/llminternal/other_processors.go
@@ -16,7 +16,10 @@ package llminternal
 
 import (
 	"google.golang.org/adk/agent"
+	icontext "google.golang.org/adk/internal/context"
+	"google.golang.org/adk/internal/utils"
 	"google.golang.org/adk/model"
+	"google.golang.org/adk/planner"
 )
 
 func identityRequestProcessor(ctx agent.InvocationContext, req *model.LLMRequest) error {
@@ -25,7 +28,31 @@ func identityRequestProcessor(ctx agent.InvocationContext, req *model.LLMRequest
 }
 
 func nlPlanningRequestProcessor(ctx agent.InvocationContext, req *model.LLMRequest) error {
-	// TODO: implement (adk-python src/google/adk/flows/llm_flows/_nl_plnning.py)
+	p := getPlanner(ctx)
+	if p == nil {
+		return nil
+	}
+
+	switch planner := p.(type) {
+	case *planner.BuiltInPlanner:
+		planner.ApplyThinkingConfig(req)
+	case *planner.ReActPlanner:
+		readonlyContext := icontext.NewReadonlyContext(ctx)
+		if planningInstruction := planner.BuildPlanningInstruction(readonlyContext, req); planningInstruction != "" {
+			utils.AppendInstructions(req, planningInstruction)
+		}
+
+		for _, content := range req.Contents {
+			if content.Parts == nil {
+				continue
+			}
+			for _, part := range content.Parts {
+				part.Thought = false
+			}
+		}
+	default:
+		return nil
+	}
 	return nil
 }
 
@@ -40,11 +67,42 @@ func authPreprocessor(ctx agent.InvocationContext, req *model.LLMRequest) error 
 }
 
 func nlPlanningResponseProcessor(ctx agent.InvocationContext, req *model.LLMRequest, resp *model.LLMResponse) error {
-	// TODO: implement (adk-python src/google/adk/_nl_planning.py)
+	if resp == nil || resp.Content == nil || len(resp.Content.Parts) == 0 {
+		return nil
+	}
+
+	p := getPlanner(ctx)
+	if p == nil {
+		return nil
+	}
+
+	// Skip built-in planner response processing
+	if _, ok := p.(*planner.BuiltInPlanner); ok {
+		return nil
+	}
+
+	// Process the LLM response for Plan-Re-Act planner
+	// TODO: Need to implement CallbackContext creation and event yielding
+	// This is complex as we need to create a callback context and potentially yield events
+	// For now, just process the response parts
+	if processedParts := p.ProcessPlanningResponse(nil, resp.Content.Parts); processedParts != nil {
+		resp.Content.Parts = processedParts
+	}
+
 	return nil
 }
 
 func codeExecutionResponseProcessor(ctx agent.InvocationContext, req *model.LLMRequest, resp *model.LLMResponse) error {
 	// TODO: implement (adk-python src/google/adk_code_execution.py)
+	return nil
+}
+
+// getPlanner returns the planner from the invocation context, or nil if no planner is available.
+func getPlanner(ctx agent.InvocationContext) planner.BasePlanner {
+	if llmAgent, ok := ctx.Agent().(Agent); ok {
+		state := Reveal(llmAgent)
+		return state.Planner
+	}
+
 	return nil
 }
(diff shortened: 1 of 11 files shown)


whoskavic (approved):

internal/llminternal/other_processors.go
```diff
@@ -40,11 +67,42 @@ func authPreprocessor(ctx agent.InvocationContext, req *model.LLMRequest) error
...
+		return nil
+	}
+
+	// Skip built-in planner response processing
+	if _, ok := p.(*planner.BuiltInPlanner); ok {
+		return nil
+	}
+
+	// Process the LLM response for Plan-Re-Act planner
+	// TODO: Need to implement CallbackContext creation and event yielding
```
git-hulk: Can resolve this TODO after PR https://github.com/google/adk-go/pull/327.

planner/base.go
```diff
@@ -0,0 +1,49 @@
...
+type BasePlanner interface {
+	// BuildPlanningInstruction builds the system instruction to be appended to the LLM request for planning.
+	//
+	// Args:
+	//   readonlyContext: The readonly context of the invocation.
+	//   llmRequest: The LLM request. Readonly.
+	//
+	// Returns:
+	//   The planning system instruction, or empty string if no instruction is needed.
+	BuildPlanningInstruction(readonlyContext agent.ReadonlyContext, llmRequest *model.LLMRequest) string
```
verdverm, line 38: Is there a good reason to make this read only?

One thing we want to be able to do is pass around budget information to the instruction generation process. Having writable context would be helpful, otherwise we have to basically shim this entire process

git-hulk: @verdverm This keeps the same behavior with adk-python. I guess it's not necessary to change the context at the building instruction phase.

verdverm: parity seems core to what the maintainers are after

I _think_ I can hook context budgeting up in the callbacks where state is mutable

git-hulk: > parity seems core to what the maintainers are after

I don't know about the priority, but maintaining consistent behavior (it's hard work) would be helpful from a maintenance perspective.

planner/react.go
```diff
@@ -0,0 +1,180 @@
...
+			}
+			preservedParts = append(preservedParts, part)
+			firstFCPartIndex = i
+			break
+		}
+
+		preservedParts = p.handleNonFunctionCallParts(part, preservedParts)
+	}
+
+	if firstFCPartIndex > 0 {
```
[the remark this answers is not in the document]
git-hulk: Need to confirm if it's a bug in adk-python.

planner/base.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"google.golang.org/adk/agent"
+	"google.golang.org/adk/model"
+)
+
+// BasePlanner is the abstract base interface for all planners.
+//
+// The planner allows the agent to generate plans for the queries to guide its
+// action.
+type BasePlanner interface {
```
zchee, line 29: In the Go idiom, the naming convention is just `Planner`. The `Base` prefix is Python idiom I think.

```suggestion
type Planner interface {
```

git-hulk: Yes, I agree `Planner` is better than `BasePlanner`. But I'm not sure how they think about the consistency and naming.

verdverm:
Finally had some time to look at this in more depth, and I'm not sure what value this actually brings?

1. You can do these planning / ReACT things without the need for a special agent or planning step in the code, and you can also build something like this today with the abstractions that already exist. Why does the core `llmagent.Agent` need a new function to implement? What am I mission?
2. Is it possible that adding this in Python was a cool idea but in retrospect not necessary? Should the Go implementation blindly clone the Python implementation, or can we find cruft or new ways forward to help inform other implementations, similar to how we rely on them for guidance?

For example, can we achieve the same with the sequential or looping agents? Is documentation and an example sufficient and preferable to expanding code interfaces and complexity?

- https://google.github.io/adk-docs/tutorials/agent-team/
- https://google.github.io/adk-docs/agents/workflow-agents/
- https://google.github.io/adk-docs/agents/multi-agents/

internal/llminternal/other_processors.go
```diff
@@ -25,7 +28,31 @@ func identityRequestProcessor(ctx agent.InvocationContext, req *model.LLMRequest
...
+	p := getPlanner(ctx)
+	if p == nil {
+		return nil
+	}
+
+	switch planner := p.(type) {
+	case *planner.BuiltInPlanner:
+		planner.ApplyThinkingConfig(req)
+	case *planner.ReActPlanner:
+		readonlyContext := icontext.NewReadonlyContext(ctx)
```
verdverm, line 40: Relating to my comments in the main thread, I want to share what I've done and give some context to my questions and thoughts around "is adding this bad API design?"

### The `gemini-cli` as an example

gemini-cli does a lot of things like this, bunch of "advanced" features out of the box, but they aren't great prompts, and they end up hurting the experience and user's wallet. You can see people sharing (and complaining), the details with 100s of requests in a single chat session. Their default agent is overly complicated, one of these bellsnwhistles approaches, and seems assembled haphazardly without justification for why some piece is there and needs to be.

### Building blocks over Batteries included

My personal opinion is that having building blocks and great abstractions is better than an ever increasing api surface and attempt at good defaults. In this sense, the good defaults are the core components — ADK has this already (https://google.github.io/adk-docs/runtime/)) — or abstractions of agents ( workflow agents seq,par,loop >>> named agent types or phases on the core `llmagent.Agent` type )

I get that planning is core, but it can be done in many ways in many places too. **We are entering a new phase of agentic R&D**, locking something into the interfaces that is in vogue today can cause growing pains tomorrow.

In my opinion, Kubernetes is the best case study in how to do this well.

### Using State and Prompts for Planning

I attached this comment to this line for a reason. **Giving your planning agent mutable access to the state is _really_ helpful.**

I have been able to get a single Agent not only to plan, but devise, share, update, and follow that plan.

1. `cache_put` tool to write to state, all key/value in the cache are added to the system prompt
1. Tell it the`planning` state value is special, must be json/yaml/... format
1. Instructions for how to use the cache and how to do planning

This is possible today because of the great abstractions around `Context`, `Callback`, `State`, and `InstructionGeneration`. What's more, it is possible to both manually/dynamically adjust agent context and build a testing/evaluation experiment and execute on that very easily. Trying to remain lean on the abstractions and interfaces is preferrable, they are already quite flexible and powerful.

[Planning prompt fragment](https://raw.githubusercontent.com/hofstadter-io/hof/refs/heads/_next/.veg/agents/instructions/shared/subconscious/planning.md) [Cache prompt fragment](https://raw.githubusercontent.com/hofstadter-io/hof/refs/heads/_next/.veg/agents/instructions/shared/cache/default.md) [Values prompt fragment](https://raw.githubusercontent.com/hofstadter-io/hof/refs/heads/_next/.veg/agents/instructions/shared/dynamic/default.md)

[Coding Agent prompt which uses the fragments](https://raw.githubusercontent.com/hofstadter-io/hof/refs/heads/_next/.veg/agents/instructions/agents/coding_assistant.md)

Image

karolpiotrowicz:
@wolo-lab this is a scope question rather than a code review, so flagging it for a direction call before anyone spends a deep review on it.

The change adds a new top-level `planner/` package exporting `BasePlanner`, `BuiltInPlanner`, `NewBuiltInPlanner`, `ReActPlanner` and `NewReActPlanner`. adk-python has an equivalent concept, so this is plausible parity rather than a novel invention, but it is a new public package and the naming and the interface shape are decisions that outlive the PR. Worth deciding whether we want the surface at all, and in this shape, before reviewing the implementation.

PR #327, opened 2025-11-20
Allow to create CallbackContext outside the agent package

git-hulk:
Currently, it has no way to create the callback context, which the planner depends on. This PR adds a new function `NewCallbackContext` to allow creating from InvocationContext.


git-hulk:
It has been implemented in `context` package.

PR #329, opened 2025-11-20
Fix race condition in state.All() method (fixes #309)

### LINKED ISSUE #309: Implementations of `State.All()` are not thread-safe and trigger the race detector
ADK provides two out-of-the-box implementations of the `Service` interface:
- one returned by `InMemoryService`
- one in the `database` package backed by an ORM

Both implementations use an unexported `state` struct:

```go
type state struct {
	mu    *sync.RWMutex
	state map[string]any
}

func (s *state) All() iter.Seq2[string, any] {
	return func(yield func(key string, val any) bool) {
		s.mu.RLock()
		for k, v := range s.state {
			s.mu.RUnlock()
			if !yield(k, v) {
				return
			}
			s.mu.RLock()
		}
		s.mu.RUnlock()
	}
}
```

The problem is that `All()` unlocks the mutex inside a range loop over a map.
This makes the map unprotected during iteration.

# Reproduction (run with `-race`)

```go
t.Run("concurrentmapiteration", func(t *testing.T) {
	m := map[string]any{}
	for i := range 10 {
		m[strconv.Itoa(i)] = i
	}


	s := state{
		mu:    &sync.RWMutex{},
		state: m,
	}
	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		defer wg.Done()
		for k, v := range s.All() {
			t.Log(k, v)
		}
	}()
	wg.Add(1)
	go func() {
		defer wg.Done()
		for i := range 5 {
			m[strconv.Itoa(i)] = i + 1000
		}
	}()
	wg.Wait()
})
```
Running this test with `go test -race` consistently reports data races

shanto12:
The state.All() method had a race condition where it unlocked the mutex inside the range loop over the map, leaving the map unprotected during iteration. This could trigger the race detector and cause concurrent map access issues.

This fix creates a copy of the map while holding the lock, then iterates over the copy without holding the lock. This ensures thread-safe iteration without blocking other operations for extended periods.

Fixes #309

### DIFF
--- a/session/inmemory.go
+++ b/session/inmemory.go
@@ -381,19 +381,22 @@ func (s *state) Get(key string) (any, error) {
 }
 
 func (s *state) All() iter.Seq2[string, any] {
-	return func(yield func(key string, val any) bool) {
-		s.mu.RLock()
-
-		for k, v := range s.state {
-			s.mu.RUnlock()
-			if !yield(k, v) {
-				return
-			}
+			return func(yield func(key string, val any) bool) {
+			// Create a copy of the map while holding the lock
 			s.mu.RLock()
-		}
+			copy := make(map[string]any, len(s.state))
+			for k, v := range s.state {
+				copy[k] = v
+			}
+			s.mu.RUnlock()
 
-		s.mu.RUnlock()
-	}
+			// Iterate over the copy without holding the lock
+			for k, v := range copy {
+				if !yield(k, v) {
+					return
+				}
+			}
+		}
 }
 
 func (s *state) Set(key string, value any) error {


baptmont:
Thanks for the PR! A few points before we can merge this PR:

Contributor License Agreement (CLA): It looks like the CLA may not be signed for this contribution. To contribute to Google open source projects, you'll need to sign the CLA. This ensures Google has the necessary rights to use your contribution.

Code Suggestion: Consider using maps.Clone to create the copy of the map, as suggested by gemini.

Testing: Could you please add specific tests to confirm that the race condition this PR addresses is resolved? Good tests here would also be invaluable for preventing regressions in the future.

Let me know if you have any questions!

wolo-lab:
Closing as the issue #309 is already fixed

PR #331, opened 2025-11-21
Validate the input must be a struct when creating tools

### LINKED ISSUE #330: [Bug] llmagent is only compatible with functiontool where input is a struct
### Problem

`functiontool` can be used in contexts other than llmagent so any Type parameter constraints make sense:
```go
type Func[TArgs, TResults any] func(tool.Context, TArgs) (TResults, error)
```
But `llmagent` is incompatible with `functiontool`s where input is a non-struct type, because `genai.FunctionCall.Args` is a map:
```
Args map[string]any `json:"args,omitempty"`
```
The result is that an agent can never succeed in calling such tool.

### Desired behavior

One of:
1. Validate tools in `llmagent.New` and report tool incompatibility as an error.
2. Make `llmagent` wrap such tools in a special decorator which will be handling unboxing of llm-provided arguments of form like `map[string]any{"input": arg}`. It will need to override input schema as well.

### Reproduction

The agent will never succeed in running the super_tool:

<details>

```go
func main() {
	ctx := context.Background()

	model, err := gemini.NewModel(ctx, "gemini-2.5-flash", &genai.ClientConfig{
		APIKey: os.Getenv("GOOGLE_API_KEY"),
	})
	if err != nil {
		log.Fatalf("Failed to create model: %v", err)
	}

	superTool, err := functiontool.New(functiontool.Config{Name: "super_tool"}, functiontool.Func[string, string](
		func(ctx tool.Context, input string) (string, error) {
			return "Hello", nil
		},
	))
	if err != nil {
		log.Fatalf("Failed to create a tool: %v", err)
	}

		Name:        "super_tool_caller",
		Model:       model,
		Description: "Agent to invoke the super_tool",
		Instruction: "Your SOLE purpose is to invoke the super_tool and pass the result to me.",
		Tools:       []tool.Tool{superTool},
	})
	if err != nil {
		log.Fatalf("Failed to create agent: %v", err)
	}

	l := full.NewLauncher()
	if err = l.Execute(ctx, &launcher.Config{AgentLoader: agent.NewSingleLoader(a)}, os.Args[1:]); err != nil {
		log.Fatalf("Run failed: %v\n\n%s", err, l.CommandLineSyntax())
	}
}

```
</details>

git-hulk:
This may partially resolve #330 by checking if the input is a struct type.

### DIFF
--- a/tool/functiontool/function_test.go
+++ b/tool/functiontool/function_test.go
@@ -610,7 +610,7 @@ func TestNew_InvalidInputType(t *testing.T) {
 				t.Fatalf("functiontool.New() error = %v, want %v", err, functiontool.ErrInvalidArgument)
 			}
 		})
-  }
+	}
 }
 
 func TestFunctionTool_PanicRecovery(t *testing.T) {


tool/functiontool/function_test.go
```diff
@@ -508,3 +520,60 @@ func stringify(v any) string {
...
+		},
+		{
+			name: "int_input",
+			createTool: func() (tool.Tool, error) {
+				return functiontool.New(functiontool.Config{
+					Name:        "int_tool",
+					Description: "a tool with int input",
+				}, func(ctx tool.Context, input int) (int, error) {
+					return input, nil
+				})
```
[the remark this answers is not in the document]
git-hulk: False positive comment.

tool/functiontool/function_test.go
```diff
@@ -508,3 +520,60 @@ func stringify(v any) string {
...
+			},
+			wantErrMsg: "input must be a struct type, got: bool",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := tc.createTool()
+			if err == nil {
+				t.Fatalf("New() succeeded, want error containing %q", tc.wantErrMsg)
```
mazas-google: nit: functiontool.New()

tool/functiontool/function_test.go
```diff
@@ -508,3 +520,60 @@ func stringify(v any) string {
...
+			}
+			if !strings.Contains(err.Error(), tc.wantErrMsg) {
+				t.Errorf("New() error = %q, want error containing %q", err.Error(), tc.wantErrMsg)
```
mazas-google: nit: functiontool.New()

tool/functiontool/function.go
```diff
@@ -54,6 +55,16 @@ type Func[TArgs, TResults any] func(tool.Context, TArgs) (TResults, error)
...
 	// TODO: How can we improve UX for functions that does not require an argument, returns a simple type value, or returns a no result?
 	//  https://github.com/modelcontextprotocol/go-sdk/discussions/37
+
+	var zeroArgs TArgs
+	argsType := reflect.TypeOf(zeroArgs)
+	for argsType != nil && argsType.Kind() == reflect.Ptr {
+		argsType = argsType.Elem()
+	}
+	if argsType == nil || argsType.Kind() != reflect.Struct {
+		return nil, fmt.Errorf("input must be a struct type, got: %v", argsType)
```
mazas-google: Could you introduce a sentinel error for simpler testing?
Example:
var ErrInvalidArgument = errors.New("invalid argument")

Then  you can wrap en error like this:
return fmt.Errorf("arg must be ...", arg, ErrInvalidArgument)

Callers can use errors.Is(err, ErrInvalidArgument) to check the error type.

tool/functiontool/function_test.go
git-hulk: Fixed.

tool/functiontool/function.go
git-hulk: Good suggestion, has changed accordingly.

git-hulk:
@mazas-google Thanks for your review. I have applied the review suggestions.

mazas-google (dismissed):
LGTM

tool/functiontool/function.go
```diff
@@ -49,11 +51,24 @@ type Config struct {
...
 	// TODO: How can we improve UX for functions that does not require an argument, returns a simple type value, or returns a no result?
 	//  https://github.com/modelcontextprotocol/go-sdk/discussions/37
+
+	var zeroArgs TArgs
+	argsType := reflect.TypeOf(zeroArgs)
+	for argsType != nil && argsType.Kind() == reflect.Ptr {
+		argsType = argsType.Elem()
+	}
+	if argsType == nil || argsType.Kind() != reflect.Struct {
+		return nil, fmt.Errorf("input must be a struct type, got: %v: %w", argsType, ErrInvalidArgument)
```
kdroste-google: Maybe struct type or struct pointer type, WDYT?

kdroste-google (dismissed):
LGTM

tool/functiontool/function.go
git-hulk: @kdroste-google Both of them are good to me. I changed to your suggestion since it's more precise.

kdroste-google: Thank you!

git-hulk:
Sorry for missing the lint error in CI, I have enabled the CI on my side and confirmed it works well: https://github.com/git-hulk/adk-go/actions/runs/19690326139/job/56404514448.

By the way, is it possible to enable CI on every push operation instead of only for the `main` branch? This will be easier for contributors to handle the CI issue before waiting for maintainers to approve the workflow. cc @mazas-google I'll be happy to submit a PR to do that if it sounds good to you.

dpasiukevich:
`maps` are also supported right now. Restricting validation to a struct only, that will break function tools that use maps:

`updateUserPreference(ctx tool.Context, args map[string]string) (map[string]string, error)`

git-hulk:
> `maps` are also supported right now. Restricting validation to a struct only, that will break function tools that use maps:
> 
> `updateUserPreference(ctx tool.Context, args map[string]string) (map[string]string, error)`

@dpasiukevich I also considered this situation while fixing this issue. But it seems `map[string]string` as the input type should not work for now, because it cannot correctly resolve the fields of the input schema when the type is `map[string]string`. So it's impossible for LLM to infer the correct fields for the function call. Please correct me if my understanding is wrong.

dpasiukevich:
For example if we have this go func `func updateUserPreference(ctx tool.Context, args map[string]string) (*updateUserPreferenceResult, error) { ... }`

This will generate the following jsonschema for input
```
          {
            "parametersJsonSchema": {
              "type": "object",
              "additionalProperties": {
                "type": "string"
              }
            },
```

> it's impossible for LLM to infer the correct fields for the function call

It all depends on the LLM instruction. For example if there's such instruction
```
"You are an agent that can update user preferences. When a user asks to set a preference, identify the preference key and the desired value. For example, in 'set my theme to dark', the key is 'theme' and the value is 'dark'. Then, call the 'update_user_preference' tool with these values."
```

Then LLM will define both fields/keys and values.

I agree, using structs is more straightforward as it defines a concrete schema.
But using maps gives more flexibility as LLM will and can define both keys and the values.

IMO we should restrict the functiontool Args to types which can be converted to json `object` (structs and maps come to my mind).

git-hulk:
@dpasiukevich Aha, I see. It did work if we explicitly specified the key-value pairs in the instruction. It looks reasonable, but it should be hard to use from the user side.

So, should I relax the input validation by allowing the map type?

dpasiukevich:
> So, should I relax the input validation by allowing the map type?

Yes, we should allow anything which jsonschema-go converts to json object (only maps and structs come to my mind, but if we missed anything, we can expand validation until there's a proper solution for other json types).

> it should be hard to use from the user side.

Just to understand your perspective -- why would it be hard to use from the user side?

This instruction does the whole thing:
```
When a user asks to set a preference, identify the preference key and the desired value. For example, in 'set my theme to dark', the key is 'theme' and the value is 'dark'. Then, call the 'update_user_preference' tool with these values.
```

Maps are the only option is the keys/fields are not known in the instruction and LLM has to infer them.
If only structs are allowed, it would still be possible to make a map within struct, but then why limiting it on top level?

git-hulk:
@dpasiukevich Thanks for your clarification.

> Yes, we should allow anything which jsonschema-go converts to json object (only maps and structs come to my mind, but if we missed anything, we can expand validation until there's a proper solution for other json types).


No problem.

> Just to understand your perspective -- why would it be hard to use from the user side?

Yes, my point is user must know what field name and type can be used in the instruction, or the LLM may infer the wrong type for this field.

git-hulk:
@dpasiukevich Could have a look again.

git-hulk:
@dpasiukevich Could you please take a look at it to see if anything else needs to be addressed?

tool/functiontool/function.go
```diff
@@ -50,11 +52,24 @@ type Config struct {
...
 	// TODO: How can we improve UX for functions that does not require an argument, returns a simple type value, or returns a no result?
 	//  https://github.com/modelcontextprotocol/go-sdk/discussions/37
+
+	var zeroArgs TArgs
+	argsType := reflect.TypeOf(zeroArgs)
+	for argsType != nil && argsType.Kind() == reflect.Ptr {
+		argsType = argsType.Elem()
+	}
+	if argsType == nil || (argsType.Kind() != reflect.Struct && argsType.Kind() != reflect.Map) {
+		return nil, fmt.Errorf("input must be a pointer to a struct or a pointer to a map, but received: %v: %w", argsType, ErrInvalidArgument)
```
dpasiukevich:
```suggestion
		return nil, fmt.Errorf("input must be a struct or a map or a pointer to those types, but received: %v: %w", argsType, ErrInvalidArgument)
```

small nit: current error message indicates that it MUST be a pointer to struct/map. But a non-pointer type is also ok. Could the message be updated?

dpasiukevich (commented):
Apologies, this PR got lost in the inbox.

One small nit + could you PTAL: 

Image

And we're good to merge. Thanks!

git-hulk:
> Apologies, this PR got lost in the inbox.
> 
> One small nit + could you PTAL:
> 
> Image
> And we're good to merge. Thanks!

@dpasiukevich Done, sorry for didn't notice this format issue.

kdroste-google (approved):
LGTM

PR #338, opened 2025-11-22
feat: Add HeaderProvider to mcptoolset

kaugesaar:
## Overview

This PR aligns the Go MCP toolset with the adk-python implementation by introducing:

### 1. HeaderProvider support  
Allows tool calls to attach context-aware HTTP headers (auth tokens, user/session IDs, tenant headers, etc).  Headers are generated per call with `agent.ReadonlyContext` context.

### 2. Session pooling keyed by headers  
Adds a new `sessionManager` that reuses MCP sessions when headers are identical and creates new ones only when necessary.

### 3. Updated example:
`examples/mcp/main.go` now supports a third mode (`AGENT_MODE=custom-headers`) that uses `HeaderProvider` to attach a GitHub PAT + optional context headers.

### `mcptoolset/set.go`
- Added `HeaderProvider` to `Config`
- Added a `sessionManager` instance to `set`
- `getSession` now:
  - accepts `agent.ReadonlyContext`
  - gathers provider headers
  - requests a pooled session for those headers

### `mcptoolset/session_manager.go` (new)
- Pools sessions keyed by deterministic header hash
- Wraps HTTP transports with a custom `RoundTripper` for static header injection
- Validates sessions using `Ping`
- Implements `Close()` to cleanly shut down all sessions

### `mcptoolset/session_manager_test.go` (new)
- Tests for:
  - session key stability
  - pooling behavior
  - header injection
  - cleanup
  - asserts that headers do *not* change session key for stdio/in-memory

### `examples/mcp/main.go`
- Added “custom-headers” mode demonstrating HeaderProvider usage.

## Alignment with Python

This matches the adk-python behavior in:

- per-call dynamic headers
- session pooling keyed by headers
- stdio/in-memory = "single session regardless of headers"

## Example
```go
headerProvider := func(ctx agent.ReadonlyContext) map[string]string {
	return map[string]string{
		"Authorization": "Bearer " + os.Getenv("GITHUB_PAT"),
		"X-User-ID":     ctx.UserID(),
	}
}

ts, _ := mcptoolset.New(mcptoolset.Config{
	Transport: &mcp.StreamableClientTransport{Endpoint: "https://api.githubcopilot.com/mcp/"},
	HeaderProvider: headerProvider,
})
```

## Known Limitation

Sessions are pooled based on headers from `HeaderProvider` only. Headers added by custom `HTTPClient` transports (e.g., `oauth2.NewClient`) are not visible to the session manager and do not affect pooling.

### DIFF
--- a/examples/mcp/main.go
+++ b/examples/mcp/main.go
@@ -35,8 +35,8 @@ import (
 	"google.golang.org/adk/tool/mcptoolset"
 )
 
-// This example demonstrates 2 ways to use MCP tools with ADK:
-// To select between two, set AGENT_MODE="local" or "github" ("local" is default).
+// This example demonstrates 3 ways to use MCP tools with ADK:
+// To select between them, set AGENT_MODE="local", "github", or "custom-headers" ("local" is default).
 //
 // 1. in-memory MCP server:
 //   - define golang function (in this case -- GetWeather)
@@ -45,6 +45,12 @@ import (
 // 2. GitHub's remote MCP server (https://github.com/github/github-mcp-server):
 //   - create http.Client with authenticated transport. In this case it's oauth2 transport with GitHub personal access token.
 //   - use `export GITHUB_PAT=...` to set GitHub personal access token.
+//
+// 3. MCP server with custom headers using HeaderProvider:
+//   - demonstrates how HeaderProvider injects context-aware headers for a real HTTP transport
+//   - uses the same GitHub MCP endpoint but leaves HTTPClient nil for brevity
+//   - headers are generated per tool call (e.g., auth token + user/session IDs)
+//   - use `export GITHUB_PAT=...` and optionally `CUSTOM_API_KEY=...`
 
 type Input struct {
 	City string `json:"city" jsonschema:"city name"`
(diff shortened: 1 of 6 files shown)


jjsasha63:
Thanks for this, and sorry it sat for nine months without a reply.

Main has moved underneath the PR in two ways that matter.

Session validity is now handled by `connectionRefresher` in `tool/mcptoolset/client.go`, which reconnects transparently on `mcp.ErrConnectionClosed`, `ErrSessionMissing` and `io.EOF`. That covers #399 without a `Ping` on every tool call.

Header injection now goes through `Config.Auth` and `auth.Transport`, which resolves the credential on every HTTP request rather than pooling a session per credential. One shared session serves every user, and nothing per-user is retained.

The session manager here takes the opposite approach, and that is the part we cannot take. `sessions` (`session_manager.go:42`) has no TTL, no LRU and no maximum size, and is cleared only by `Close()`. With the headers from your own example, `Authorization` plus `X-User-ID` from `ctx.UserID()`, that is one live MCP session per user held for the lifetime of the process. Your doc comment on `HeaderProvider` flags the same growth ("you will create a new session for every call"), so I think we already agree on the shape of the problem.

The piece we do still want is `HeaderProvider`. `auth.WithHeaders` already covers static extra headers, but it cannot cover yours: a `CredentialProvider` only receives `req.Context()`, and there is no exported way to recover the ADK context from it, so `ctx.UserID()` is out of reach. A `func(agent.ReadonlyContext) map[string]string` is what closes that gap.

Would you be up for a smaller follow-up that adds it at the same `RoundTripper` seam as `Config.Auth` (see `buildTransport` and `authHTTPClient` in `set.go`) and drops the session manager entirely? I am closing this one so it stops sitting stale. Tag me on the new PR and I will make sure it gets reviewed quickly.

PR #351, opened 2025-11-24
Add initial devcontainer configuration

krishna-katekhaye-dev:


kdroste-google:
@krishna-katekhaye-dev, it seems to me that not everyone wants to use Dev Containers. But perhaps you can share some information about it, including some recommendations in [adk-docs](https://github.com/google/adk-docs) or another community repository?

PR #360, opened 2025-11-27
feat: remoteagent a2a request callbacks

yarolegovich:
Added `BeforeA2ARequestCallback`, `A2AEventConverter` and `AfterA2ARequestCallback` to `remoteagent.A2AConfig`.

`BeforeA2ARequestCallback` allows to modify a request or prevent it by returning a non-nil event or error which becomes the result of agent invocation. 

`A2AEventConverter` allows to implement a custom conversion.

`AfterA2ARequestCallback` allows to enrich the result of the conversion, overwrite it or reject a response.

Additionally, replaced gRPC transport in tests with JSONRPC because of an easier setup.


agent/remoteagent/a2a_agent_test.go
```diff
@@ -128,9 +103,9 @@ func runAndCollect(ic agent.InvocationContext, agnt agent.Agent) ([]*session.Eve
 }
 
 func toLLMResponses(events []*session.Event) []model.LLMResponse {
-	result := make([]model.LLMResponse, len(events))
-	for i, v := range events {
-		result[i] = v.LLMResponse
+	var result []model.LLMResponse
+	for _, v := range events {
+		result = append(result, v.LLMResponse)
 	}
```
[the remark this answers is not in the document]
yarolegovich: it's a test

agent/remoteagent/a2a_agent.go
```diff
@@ -98,40 +134,77 @@ func (a *a2aAgent) run(ctx agent.InvocationContext, cfg A2AConfig) iter.Seq2[*se
...
 
-		req := &a2a.MessageSendParams{Message: msg, Config: cfg.MessageSendConfig}
 		for a2aEvent, err := range client.SendStreamingMessage(ctx, req) {
-			if err != nil {
-				event := toErrorEvent(ctx, err)
-				updateCustomMetadata(event, req, nil)
-				yield(event, nil)
-				return
+			var event *session.Event
+			if cfg.Converter != nil {
```
baptmont, line 160: Would it make sense to keep the err validation here or alternatively expose toErrorEvent for it to be used in the custom Converters?

agent/remoteagent/a2a_agent.go
```diff
@@ -27,11 +27,28 @@ import (
...
 	"google.golang.org/adk/session"
 )
 
+// BeforeA2ARequestCallback is called before sending a request to the remote agent.
+//
+// If it returns non-nil result or error, the actual call is skipped and the returned value is used
+// as the agent invocation result.
+type BeforeA2ARequestCallback func(ctx agent.CallbackContext, req *a2a.MessageSendParams) (*session.Event, error)
+
+// A2AEventConverter can be used to provide a custome implementation of A2A event transformation logic.
```
dpasiukevich:
```suggestion
// A2AEventConverter can be used to provide a custom implementation of A2A event transformation logic.
```

agent/remoteagent/a2a_agent.go
```diff
@@ -98,40 +134,78 @@ func (a *a2aAgent) run(ctx agent.InvocationContext, cfg A2AConfig) iter.Seq2[*se
 			return
 		}
 
-		if len(msg.Parts) == 0 {
-			yield(adka2a.NewRemoteAgentEvent(ctx), nil)
+		req := &a2a.MessageSendParams{Message: msg, Config: cfg.MessageSendConfig}
+		if cbResp, cbErr := runBeforeA2ARequestCallbacks(ctx, cfg, req); cbResp != nil || cbErr != nil {
+			yield(cbResp, cbErr)
```
dpasiukevich: WDYT if we pass cbResp and cbErr from before_callback to afterA2ARequestCallbacks?

yarolegovich: done

PR #368, opened 2025-11-30
fix: concurrent session creation race

### LINKED ISSUE #367: Concurrent InMemory session lacks locking

**Describe the bug**
`func (s *inMemoryService) Create(ctx context.Context, req *CreateRequest)` checks for an existing session by calling `s.sessions.Get(encodedKey)` before it acquires `s.mu`, breaking  the  guarantee of "Tread-safe" of `type inMemoryService struct`.

**To Reproduce**
```Go
func Test_inMemoryService_CreateConcurrentAccess(t *testing.T) {
	s := InMemoryService()
	const goroutines = 16
	const attempts = 32

	start := make(chan struct{})
	var wg sync.WaitGroup
	wg.Add(goroutines)

	req := &CreateRequest{
		AppName:   "race-app",
		UserID:    "race-user",
		SessionID: "race-session",
	}

	for range goroutines {
		go func() {
			defer wg.Done()
			<-start
			for range attempts {
				_, _ = s.Create(t.Context(), req)
			}
		}()
	}

	close(start)
	wg.Wait()
}
```

**Expected behavior**

**Screenshots**

Image

 - OS: [e.g. macOS, Linux, Windows] Linux
 - Go version: 1.25.0
 - ADK version: main

**Model Information:**
 - Which model is being used (e.g. gemini-2.5-pro)

**Additional context**

cryo-zd:
### Change:
This PR guards the duplicate session check with `s.mu` in `func (s *inMemoryService) Create`. The `omap.Map` lookup used to happen before the lock, so concurrent `Create` calls for the same `(app, user, session)` may could both pass the "exists" check and insert.

Fix #367 

race_2

### DIFF
--- a/session/inmemory.go
+++ b/session/inmemory.go
@@ -60,8 +60,10 @@ func (s *inMemoryService) Create(ctx context.Context, req *CreateRequest) (*Crea
 	}
 
 	encodedKey := key.Encode()
-	_, ok := s.sessions.Get(encodedKey)
-	if ok {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if _, ok := s.sessions.Get(encodedKey); ok {
 		return nil, fmt.Errorf("session %s already exists", req.SessionID)
 	}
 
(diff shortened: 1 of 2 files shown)


baptmont:
Hi @cryo-zd nice catch and thank you for the contribution. The fix looks great but I would modify the test as suggested by gemini.

cryo-zd:
> Hi @cryo-zd nice catch and thank you for the contribution. The fix looks great but I would modify the test as suggested by gemini.

Thanks for your review, I have updated the test as suggested. When you have time, please review the change.

dpasiukevich (approved):

PR #372, opened 2025-12-01
Document reasoning for omitting ctx.EndInvocation()

MRuhan17:
Added detailed comments explaining the intentional omission of ctx.EndInvocation() and its implications for cross-language consistency.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -285,20 +285,14 @@ func runAfterAgentCallbacks(ctx InvocationContext) (*session.Event, error) {
 		event.Branch = ctx.Branch()
 		event.Actions = *callbackCtx.actions
 		
-		// NOTE:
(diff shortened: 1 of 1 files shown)


baptmont:
Hi, can you fix the indentation and remove the TODO?

jjsasha63:
Thanks for this, and apologies for the long silence — nine months on a six-line comment is on us, not you.

The explanation belongs in the file and we have opened #1461 to add it. We are closing this PR rather than asking you to rebase: the branch is now 341 commits behind main and conflicts, and the indentation and TODO points from 2 December are still open, which is more work than the change is worth on your side.

For what it is worth, #370 was a good find. The TODO had sat there unexplained for a long time, and the discussion you started on it is what produced the wording in #1461, which credits you both. If you would like something that lands, #681 is open and labelled good first issue.

Issue #278, opened 2025-11-13
Ensure LLM can parse tool error.


dpasiukevich:
FunctionResponse has `map[string]any` type.

Currently if the tool returns an error, we pass `map[string]any{"error": err}`. Errors created with `fmt.Errorf` or `errors.New` are json marshalled to `{}`.
As a result LLM knows the tool returned an error, but it doesn't know anything about it.

We should pass `err.Error()` for that.

dpasiukevich:
Also we need to make sure we have a clear documentation for a functiontool about what happens to error and when and what types of error should be in the result struct rather than error.

PR #374, opened 2025-12-01
fix: pass tool error to after_tool_callback

dpasiukevich:
After functool prototype was updated to have error: `func(arg) (resp, err)`. Now user error can be in `res, err := callTool(...)` and we'd need to pass this error to after_tool_callback.

Also fixes: https://github.com/google/adk-go/issues/278 (as now it will pass err.Error() from user funcs/callbacks to LLM).

### DIFF
--- a/internal/llminternal/base_flow_test.go
+++ b/internal/llminternal/base_flow_test.go
@@ -110,6 +110,9 @@ func TestCallTool(t *testing.T) {
 				func(ctx tool.Context, tool tool.Tool, args map[string]any) (map[string]any, error) {
 					return map[string]any{"result": "intercepted"}, nil
 				},
+				func(ctx tool.Context, tool tool.Tool, args map[string]any) (map[string]any, error) {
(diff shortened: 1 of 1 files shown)


baptmont (approved):

internal/llminternal/base_flow_test.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	return nil, nil
+}
+
+func (m *mockFunctionTool) Declaration() *genai.FunctionDeclaration {
+	return nil
+}
+
+func TestCallTool(t *testing.T) {
+	tests := []struct {
```
squee1945, line 70: Add cases (or alter existing) to show what happens when multiple before callbacks or multiple after callbacks are present - I think a non-error allows the chain to continue, but an error halts the chain?

dpasiukevich: Makes sense, I've added a 2nd callback in each test case (before & after callbacks) to verify if one callback returns result/error the callback chain is stopped.

PR #377, opened 2025-12-02
functiontool: support non-struct inputs from LLM function calls

saurabhhhcodes:
Proposes automatic wrapping of function tools whose input type is a non-struct (e.g. string/int) so they accept LLM function-call arguments like {"input": <value>}. Includes wrapper implementation, tests, docs, and a changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,7 @@
+# Changelog
+
+## Unreleased
+
+- functiontool: support non-struct inputs from LLM function calls
+  - Automatically wraps function tools whose input type is a non-struct (e.g. `string`, `int`) so they accept LLM function-call arguments of the form `{ "input": <value> }`.
+  - Adds `nonStructInputWrapper` implementation, unit + integration tests, and documentation.
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ Agent Development Kit (ADK) is a flexible and modular framework that applies sof
 
 This Go version of ADK is ideal for developers building cloud-native agent applications, leveraging Go's strengths in concurrency and performance.
 
+Note: Function tools that accept primitive (non-struct) inputs are now automatically supported — such handlers will be exposed to models as a function taking an object with a single `"input"` property. See `CHANGELOG.md` for details.
+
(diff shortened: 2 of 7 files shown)


saurabhhhcodes:
Requesting maintainers' review and workflow approval for this change (wrapper for non-struct functiontool inputs).

Summary: Adds automatic wrapper for function tools with non-struct inputs (e.g., string/int), unit+integration tests, docs, and CHANGELOG entry. Local tests passed. Please review and approve workflows so CI can run. If you prefer different reviewer assignments or a smaller branch, let me know and I can update the PR.

saurabhhhcodes:
(Fallback) Could maintainers @Dmitry and @baptmont  review this PR? Top contributors identified from repo history: Dmitry Pasiukevich, Jaana Dogan, João Westerberg, Karol Droste. I attempted to add reviewers programmatically but the API returned a permissions error; please add reviewers or approve workflows when you can. Thank you!,isBackground:false}

saurabhhhcodes:
I'll monitor this PR and will merge automatically once the required maintainer approvals and CI/workflow checks complete. If maintainers prefer a smaller PR, I can split this into a minimal wrapper-only PR — please let me know. Thanks!

wolo-lab:
Thank you for this, and I am sorry about how it was handled. You asked for a review three times and never got one from us. Worse, we merged #331 the day after you opened this, taking the design in the opposite direction, and nobody told you.

That decision stands. functiontool.New requires the arguments type to be a struct or a map and rejects anything else at construction with a typed error rather than failing at call time: https://github.com/google/adk-go/blob/245ba284ed51f0f58355a6233f0b1ae01f252160/tool/functiontool/function.go#L88-L90

The reasoning is that a Go handler taking a bare scalar still maps to a single-property object on the wire, so wrapping it under a fixed "input" key takes the property name and its description away from the tool author. A one-field struct produces the same request bytes and gives the model a better declaration to read. adk-python has no equivalent either, since Python parameters are always named.

We are reviewing the queue far more actively now, and a decision like that one will get communicated on the PR it affects.

Issue #356, opened 2025-11-26
Accepting filters for function calls


cs168898:
**Is your feature request related to a problem? Please describe. **
I saw a 'TODO:' in the source code (base_flow.go lines 363) for filtering function calls, I would like to check how would the function calls in the filter be defined? Through the agent dynamically generating it or through user definition or another way I have not thought of.

**Describe the solution you'd like**
I hope to have a clear idea on how the filters will be defined, in order to build the filter as well as the filter acceptance functionality.

**Additional context**
I have just started using and learning this project recently, hoping to contribute in some way 😀

cointem:
me too

PR #380, opened 2025-12-02
Filter function calls

cs168898:
## Filter Function Calls

Developers using ADK-go can now add filters when they create a new LLM agent as such:
```go
	agent, err := llmagent.New(llmagent.Config{
		Name:        "agent",
		Description: "math agent",
		Model:       model,
		Instruction: "only use the tools provided to output results",
		// TODO(hakim): set to false when autoflow is implemented.
		DisallowTransferToParent: true,
		DisallowTransferToPeers:  true,
		Tools:                    []tool.Tool{sumTool},
		Filter:                   map[string]bool{sumTool.Name(): false},  <================ Added filter field
	})
```
The Filter field accepts a map where the `Key` is the tool's name and the `Value` is a boolean on whether the tool is allowed or not.
True = Tool is allowed.
False = Tool is NOT allowed.

- Added Filter field to the `Config` struct.
- Added filter field to the `State` struct.
- Implemented filter logic in `base_flow.go`
- Added tests for testing if filtered tools are still called.

## Notes
I have decided to start on developer declared filter first as the use cases can include tool prohibition based on user roles.
For example, if a user has no premium access account, developers could create a conditional check and initialize the agent without access to specific tools.
The current implementation will impose a permanent tool filter on all requests after agent initialization, meaning we cannot change the filter once the agent has been initialized.

## TODO
 - Implement filter changes per request.
 - Agent defined filters to pass to sub-agents.

This PR references this [Issue ](https://github.com/google/adk-go/issues/356)

### DIFF
--- a/agent/llmagent/llmagent.go
+++ b/agent/llmagent/llmagent.go
@@ -67,6 +67,7 @@ func New(cfg Config) (agent.Agent, error) {
 			GenerateContentConfig:    cfg.GenerateContentConfig,
 			Tools:                    cfg.Tools,
 			Toolsets:                 cfg.Toolsets,
+			Filter:                   cfg.Filter,
 			DisallowTransferToParent: cfg.DisallowTransferToParent,
 			DisallowTransferToPeers:  cfg.DisallowTransferToPeers,
 			InputSchema:              cfg.InputSchema,
@@ -248,6 +249,9 @@ type Config struct {
 	// Toolsets will be used by llmagent to extract tools and pass to the
 	// underlying LLM.
 	Toolsets []tool.Toolset
+
+	// Filter will be used by llmagent to filter tools that will not be used by this specific agent
+	Filter map[string]bool
 
 	// OutputKey is an optional parameter to specify the key in session state for the agent output.
 	//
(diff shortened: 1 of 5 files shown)


caglar10ur:
FYI https://github.com/google/adk-go/pull/489

wolo-lab:
Thank you for raising this, and apologies that the only reply you got was a bare link with no explanation attached.

The capability landed in #489 as tool.FilterToolset together with tool.AllowedToolsPredicate: https://github.com/google/adk-go/blob/245ba284ed51f0f58355a6233f0b1ae01f252160/tool/tool.go#L68-L94

It wraps a Toolset and applies the predicate inside Tools(ctx), so the filter can depend on the invocation context rather than being fixed at construction. That also matches adk-python, which puts the filter on the toolset through BaseToolset(tool_filter=...) rather than on the agent.

    ts := tool.FilterToolset(myToolset, tool.AllowedToolsPredicate([]string{"sum"}))

Issue #356 is covered by the same change and we will close it alongside this.

We are much more active on incoming PRs now, and next time you should get a real answer rather than a link.

cs168898:
Thank you for addressing this @wolo-lab, even though this was really long ago. I really appreciate it.

PR #381, opened 2025-12-02
Fix mcp tool convert, null responseJsonSchema

baptmont:
Fix mcp tool convert to only set responseJsonSchema if mcp tool has output schema.


kdroste-google (approved):
LGTM

PR #384, opened 2025-12-03
feat(database): allow passing a db connection to the session service

### LINKED ISSUE #340: Database connections across services
Was looking at the SessionService database implementation and noticed the database connection is opened by the service. It looks like this prevents me from opening a single database client and passing it to all the services (others still lack implementations)

```go
func NewSessionService(dialector gorm.Dialector, opts ...gorm.Option) (session.Service, error) {
	db, err := gorm.Open(dialector, opts...)
	if err != nil {
		return nil, fmt.Errorf("error creating database session service: %w", err)
	}
	return &databaseService{db: db}, nil
}
```

so I changed it to look like this

```go
func NewSessionService(db *gorm.DB) (session.Service, error) {
	return &databaseService{db: db}, nil
}
```

Another benefit to this is that my application can also use the same database connection for any extra tables it needs.

Thoughts on the change and the issue more generally?

NotNahid:
This change modifies the NewSessionService function to accept a *gorm.DB object directly, instead of creating a new one. This allows for sharing a single database connection across multiple services.

The AutoMigrate function has also been updated to accept a *gorm.DB object directly.

Fixes #340


baptmont:
NewSessionService will be changed for options constructor

PR #385, opened 2025-12-03
Modify handleFunctionCalls to execute beforeToolCallbacks despite tool existing

baptmont:
Add retry callback example.
Add base_flow tests.

### DIFF
--- a/agent/llmagent/llmagent.go
+++ b/agent/llmagent/llmagent.go
@@ -262,6 +262,7 @@ type AfterModelCallback func(ctx agent.CallbackContext, llmResponse *model.LLMRe
 // Parameters:
 //   - ctx: The tool.Context for the current tool execution.
 //   - tool: The tool.Tool instance that is about to be executed.
+//     This value can be nil if the tool name invoked by the LLM function call does not correspond to an existing tool.
 //   - args: The original arguments provided to the tool.
 type BeforeToolCallback func(ctx tool.Context, tool tool.Tool, args map[string]any) (map[string]any, error)
 
@@ -271,9 +272,11 @@ type BeforeToolCallback func(ctx tool.Context, tool tool.Tool, args map[string]a
 // Parameters:
 //   - ctx:    The tool.Context for the tool execution.
 //   - tool:   The tool.Tool instance that was executed.
+//     This value can be nil if the tool name invoked by the LLM function call does not correspond to an existing tool,
+//     and a BeforeToolCallback generated result or an error
 //   - args:   The arguments originally passed to the tool.
-//   - result: The result returned by the tool's Run method.
-//   - err:    The error returned by the tool's Run method.
+//   - result: The result returned by the tool's Run method or by a BeforeToolCallback.
+//   - err:    The error returned by the tool's Run method or by a BeforeToolCallback.
 type AfterToolCallback func(ctx tool.Context, tool tool.Tool, args, result map[string]any, err error) (map[string]any, error)
 
 // IncludeContents controls what parts of prior conversation history is received by llmagent.
--- a/examples/tools/retrytool/main.go
+++ b/examples/tools/retrytool/main.go
@@ -1,0 +1,114 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package demonstrates a workaround for using Google Search tool with other tools.
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"os"
+
+	"google.golang.org/genai"
+
+	"google.golang.org/adk/agent"
+	"google.golang.org/adk/agent/llmagent"
+	"google.golang.org/adk/cmd/launcher"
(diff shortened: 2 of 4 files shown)


examples/tools/retrytool/main.go
```diff
@@ -0,0 +1,114 @@
...
+		Name:        "root_agent",
+		Model:       geminiModel,
+		Description: "You can get a cloud build.",
+		Instruction: `You can get information about a cloud build. 
+		Use the available tools for that but instead of 'get_cloud_build' call 'get_cloud-_build', it's ok if it does not exist the client will know what to do with it. 
+		Never confirm with the user in case of a retry.`,
+		Tools: []tool.Tool{
+			getCloudBuild,
+		},
+		BeforeToolCallbacks: []llmagent.BeforeToolCallback{
```
squee1945, line 87: If I understand the code correctly:

If I raise an error (e.g., unknown tool) in the BeforeCallback, and I happen to have an AfterCallback (e.g., metrics, logging), then I have to be sure that my AfterCallback returns the error passed to it - correct?

(This feels like an error prone interface, though I don't immediately have a suggestion to address it.)

internal/llminternal/base_flow.go
```diff
@@ -372,19 +372,25 @@ func (f *Flow) handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[st
...
 	for _, fnCall := range fnCalls {
-		curTool, ok := toolsDict[fnCall.Name]
-		if !ok {
-			return nil, fmt.Errorf("unknown tool: %q", fnCall.Name)
-		}
-		funcTool, ok := curTool.(toolinternal.FunctionTool)
-		if !ok {
-			return nil, fmt.Errorf("tool %q is not a function tool", curTool.Name())
-		}
+		curTool, toolExists := toolsDict[fnCall.Name]
```
squee1945, line 375: Could the framework just do this?

```
if !toolExists {
  return map[string]any{"error": fmt.Errorf("tool %q does not exist, please choose another tool", fnCall.Name)}, nil
}
```

squee1945: Is this the ADK-Python equivalent? https://github.com/google/adk-python/blob/main/src/google/adk/flows/llm_flows/functions.py#L323-L336

examples/tools/retrytool/main.go
dpasiukevich: No. If AfterCallback is no-op (returns nil, nil) then the original resp/err is returned.

https://github.com/google/adk-go/blob/d8372e6609219c3ea07310e217be776ac29c3db4/internal/llminternal/base_flow.go#L462

squee1945: There's no way to "cancel" the original error then?

internal/llminternal/base_flow.go
```diff
@@ -372,19 +372,25 @@ func (f *Flow) handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[st
...
+		funcTool, toolIsFunction := curTool.(toolinternal.FunctionTool)
 		toolCtx := toolinternal.NewToolContext(ctx, fnCall.ID, &session.EventActions{StateDelta: make(map[string]any)})
-		// toolCtx := tool.
 		spans := telemetry.StartTrace(ctx, "execute_tool "+fnCall.Name)
 
-		result := f.callTool(funcTool, fnCall.Args, toolCtx)
+		result, err := f.invokeBeforeToolCallbacks(funcTool, fnCall.Args, toolCtx)
```
squee1945, line 380: With this change, it seems the contract with the client is changed: now the tool can be `nil`.

This could cause runtime errors with existing clients.

ADK-Python seems to have a different callback for tool errors - and it generates a fake tool to hold the name of the tool that had the error. https://github.com/google/adk-python/blob/main/src/google/adk/flows/llm_flows/functions.py#L326-L332

Should ADK-Go do something similar?

examples/tools/retrytool/main.go
squee1945: I don't have a use case for cancelling/correcting the original error, but I just wanted to make sure that it wasn't a desirable capability of ADK-Go?

baptmont: If an AfterToolCallback return a response that would "cancel" the original error. If the AfterToolCallback returns an err it would replace the original. If it (returns nil, nil) then the original resp/err is returned.

internal/llminternal/base_flow.go
baptmont: This seems like a good option. Simply notify the llm instead of returning the err all the way back to the runner.

examples/tools/retrytool/main.go
squee1945: Yes, but I cannot fully remove the error. (Again, I don't personally need this. I'm just pointing it out in case it's important for someone else. You can ignore and close this thread otherwise.)

dpasiukevich: > but I cannot fully remove the error.

Wanted to clarify. Why couldn't user cannot cancel/remove the error?

after_tool_callback can `return resp, nil`

squee1945: Ah yes - you're right.

agent/llmagent/llmagent.go
```diff
@@ -271,9 +272,11 @@ type BeforeToolCallback func(ctx tool.Context, tool tool.Tool, args map[string]a
 // Parameters:
 //   - ctx:    The tool.Context for the tool execution.
 //   - tool:   The tool.Tool instance that was executed.
+//     This value can be nil if the tool name invoked by the LLM function call does not correspond to an existing tool,
+//     and a BeforeToolCallback generated result or an error
```
hyangah: Are there any other mode `BeforeToolCallback` takes other than returning result or an error? I don't think the comment in line 276 is necessary.

hyangah:
From our triage meeting, this is stale and ok to close it; can you please confirm? @baptmont

PR #392, opened 2025-12-04
Gemini ADK version tag update to 0.2.0

kdroste-google:


dpasiukevich:
That's too late unfortunately, `v0.2.0` tag is released already.

I'm exploring a way to automate this. This field should be updated with a new value exactly before we cut a new release tag.

PR #404, opened 2025-12-08
Resolve TODO comment in telemetry.go by adding the finish reason and usage

git-hulk:
After applying this PR:

image

### DIFF
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -13,6 +13,9 @@
 // limitations under the License.
 
 // Package telemetry sets up the open telemetry exporters to the ADK.
+//
+// WARNING: telemetry provided by ADK (internaltelemetry package) may change (e.g. attributes and their names)
+// because we're in process to standardize and unify telemetry across all ADKs.
 package telemetry
 
 import (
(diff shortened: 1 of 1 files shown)


internal/telemetry/telemetry.go
```diff
@@ -51,13 +51,22 @@ var (
...
+	genAiToolName        = "gen_ai.tool.name"
+	genAiToolCallID      = "gen_ai.tool.call.id"
+	genAiSystemName      = "gen_ai.system"
+
 	genAiRequestModelName = "gen_ai.request.model"
+	genAiRequestTopP      = "gen_ai.request.top_p"
+	genAiRequestMaxTokens = "gen_ai.request.max_tokens"
+
+	genAiResponseFinishReason        = "gen_ai.response.finish_reason"
+	genAiResponsePromptTokens        = "gen_ai.response.prompt_tokens"
```
dpasiukevich: nit: WDYT if we have the same names as genai API provides? in genai the field is named `event.UsageMetadata.PromptTokenCount`

```suggestion
	genAiResponsePromptTokenCount        = "gen_ai.response.prompt_token_count"
```

And the same for other added fields.

internal/telemetry/telemetry.go
```diff
@@ -201,14 +210,29 @@ func TraceLLMCall(spans []trace.Span, agentCtx agent.InvocationContext, llmReque
...
+		}
+		if event.FinishReason != "" {
+			attributes = append(attributes, attribute.String(genAiResponseFinishReason, string(event.FinishReason)))
+		}
+		if event.UsageMetadata != nil {
+			if event.UsageMetadata.PromptTokenCount > 0 {
+				attributes = append(attributes, attribute.Int(genAiResponsePromptTokens, int(event.UsageMetadata.PromptTokenCount)))
+			}
+			if event.UsageMetadata.CandidatesTokenCount > 0 {
+				attributes = append(attributes, attribute.Int(genAiResponseCompletionTokens, int(event.UsageMetadata.CandidatesTokenCount)))
```
dpasiukevich: Mismatching names: variable name `genAiResponseCompletionTokens` while data is from `int(event.UsageMetadata.CandidatesTokenCount)`

internal/telemetry/telemetry.go
git-hulk: Good point.

internal/telemetry/telemetry.go
git-hulk: Yes, I used the OpenAI term to record those attributes. And all of them are changed accrodingly now.

git-hulk:
@dpasiukevich Thanks for your review and good suggestions. All issues that you mentioned should be addressed now, please retake a look when you get time.

Issue #401, opened 2025-12-05
Move agent.RunConfig to the runner package


rakyll:
`agent.RunConfig` represents the Runner config during runtime. This type should be underneath the runner package rather than the agent package.

Ankit152:
Hey @rakyll 👋🏼 

I have added a fix to this here PR #405 . Kindly have a look! 🙂

dpasiukevich:
> This type should be underneath the runner package rather than the agent package.

This was the original intention to put this struct to the `runner` package.

But there are 2 things:

1. In adk-python `RunConfig` located in `agents` directory and not in runners.py:

https://github.com/google/adk-python/blob/143ad44f8c5d1c56fc92dd691589aaa0b788e485/src/google/adk/agents/run_config.py#L40

2. There is already `runner.Config` struct for runner configuration. Having both `runner.Config` and `runner.RunConfig` may cause confusion.

rakyll:
> In adk-python RunConfig located in agents directory and not in runners.py:

To keep the consistency we should keep `RunConfig` in the agent package. Let's close the PR and this issue. Consistency with ADK Python is higher priority.

PR #405, opened 2025-12-08
Move `agent.RunConfig` to runner package

Ankit152:
## Description

Resolves: #401

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -25,6 +25,7 @@ import (
 	agentinternal "google.golang.org/adk/internal/agent"
 	"google.golang.org/adk/memory"
 	"google.golang.org/adk/model"
+	"google.golang.org/adk/runner/runconfig"
 	"google.golang.org/adk/session"
 )
 
@@ -389,7 +390,7 @@ type invocationContext struct {
 	invocationID  string
 	branch        string
 	userContent   *genai.Content
-	runConfig     *RunConfig
+	runConfig     *runconfig.RunConfig
 	endInvocation bool
 }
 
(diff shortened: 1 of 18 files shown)


agent/agent.go
```diff
@@ -25,6 +25,7 @@ import (
 	agentinternal "google.golang.org/adk/internal/agent"
 	"google.golang.org/adk/memory"
 	"google.golang.org/adk/model"
+	"google.golang.org/adk/runner/runconfig"
```
rakyll, line 28: Do we need a new package for this? Is it possible to provide RunConfig from the runner package?

Ankit152: Initially, I attempted that approach, but it resulted in an import cycle issue. As a solution, I created a new package under runner.
Now that the issue has been resolved, I believe we can close this PR as well.

Ankit152: I’m interested in contributing to the project and would love to join the relevant group or Slack channel to stay aligned with ongoing discussions. If that’s possible, could you please add me or share the invitation link?

dpasiukevich:
@Ankit152 let's keep the `agent.RunConfig` in the original package if you don't mind. With reasons listed in the https://github.com/google/adk-go/issues/401#issuecomment-3628297989. Thanks!

Ankit152:
No probelm @dpasiukevich 👍🏼 🙂

agent/agent.go
dpasiukevich: We're working to add some real-time place for discussions (discord/slack etc).

For now it's in this repository (via issues and discussions).

Thanks for the interest!

Issue #399, opened 2025-12-05
mcptoolset: cached MCP session not validated, causes "connection closed" errors on subsequent requests


mikaelhatanpaa:
**Describe the bug**

The `mcptoolset` package caches the `*mcp.ClientSession` after the first connection and never validates if it's still alive before returning it. For HTTP/SSE-based MCP servers (using `StreamableClientTransport`), sessions can become stale due to server-side session timeouts, network interruptions, or idle connection cleanup.

When the cached session becomes stale, all subsequent requests fail with errors like:
```
connection closed: calling "tools/list": client is closing: standalone SSE stream: failed to reconnect (session ID: ...): connection failed after 5 attempts
```

The first request always works, but subsequent requests fail.

**To Reproduce**

1. Install ADK: `go get google.golang.org/adk@v0.2.0`

2. Create an MCP server using the Go MCP SDK with HTTP transport:
```go
import mcpsdk "github.com/modelcontextprotocol/go-sdk/mcp"

srv := mcpsdk.NewServer(&mcpsdk.Implementation{Name: "test-server", Version: "1.0.0"}, nil)
handler := mcpsdk.NewStreamableHTTPHandler(func(r *http.Request) *mcpsdk.Server {
    return srv
}, &mcpsdk.StreamableHTTPOptions{
    SessionTimeout: 5 * time.Minute,
})
// Serve on http://localhost:8080/mcp
```

3. Create an ADK agent that connects to this MCP server via HTTP:
```go
transport := &mcp.StreamableClientTransport{
    Endpoint: "http://localhost:8080/mcp",
}
toolset, _ := mcptoolset.New(mcptoolset.Config{
    Transport: transport,
})
agent, _ := llmagent.New(llmagent.Config{
    Name:     "test-agent",
    Model:    model,
    Toolsets: []tool.Toolset{toolset},
})
```

4. Send a request to the agent → **works**
5. Wait a moment (or let the session timeout)
6. Send another request → **fails with "connection closed" error**

**Root cause**

From `tool/mcptoolset/set.go`:
```go
func (s *set) getSession(ctx context.Context) (*mcp.ClientSession, error) {
    s.mu.Lock()
    defer s.mu.Unlock()
    if s.session != nil {
        return s.session, nil  // ← Always returns cached session without checking if still valid
    }
    session, err := s.client.Connect(ctx, s.transport, nil)
    // ...
    s.session = session
    return s.session, nil
}
```

**Expected behavior**

[...]

PR #406, opened 2025-12-08
fix: use background context for MCP session and add test

Nerja:
The MCP session was being initialized using the request-scoped context passed to Tools(ctx). When this context was cancelled the underlying MCP connection was closed. However, the session remained cached in the mcptoolset struct, causing subsequent calls to fail.

Changed the Connect call in getSession to use context.Background().

Added TestSessionContextCancellation to verify that the context passed to connect is not cancelled.

For questions related to executing Ping as well to confirm the status of the connection, please see https://github.com/google/adk-go/issues/399#issuecomment-3643981984

### DIFF
--- a/tool/mcptoolset/set.go
+++ b/tool/mcptoolset/set.go
@@ -142,7 +142,7 @@ func (s *set) getSession(ctx context.Context) (*mcp.ClientSession, error) {
 		return s.session, nil
 	}
 
-	session, err := s.client.Connect(ctx, s.transport, nil)
+	session, err := s.client.Connect(context.Background(), s.transport, nil)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


dpasiukevich:
I don't think there's problem with request-scoped context. Because this context is only needed to create a session. Once session object is created even if the context is cancelled it doesn't affect the session. The MCP lib doesn't store the context.

@Nerja could you verify if this assumption is correct?

If it is, then setting context.Background() won't solve the issue. As we'd need rather to handle stale cached session object.

Nerja:
Closing in favor of https://github.com/google/adk-go/pull/417 and based on https://github.com/google/adk-go/issues/399#issuecomment-3647187957

PR #409, opened 2025-12-09
fix error:stale session error: last update time from request is older…

huangbug:
**Fix: Truncate event timestamp to millisecond precision to prevent stale session errors**

The MySQL `datetime(3)` column type only supports millisecond precision (3 decimal places). When storing timestamps with nanosecond precision, MySQL rounds the value, which can cause the stored time to be slightly greater than the original in-memory time.

This rounding issue triggers false "stale session error" because the comparison in `applyEvent()` uses `UnixNano()`:

```go
    return fmt.Errorf("stale session error: ...")
}
```

**Solution**: Truncate the event timestamp to millisecond precision before persisting, ensuring consistency between in-memory and database values:

```go
event.Timestamp = time.UnixMilli(event.Timestamp.UnixMilli())
```

This eliminates sub-millisecond precision that would otherwise be lost (and potentially rounded up) during database storage.

### DIFF
--- a/session/database/service.go
+++ b/session/database/service.go
@@ -328,7 +328,7 @@ func (s *databaseService) AppendEvent(ctx context.Context, curSession session.Se
 		return nil
 	}
 
-	// The update_time in the database is in milliseconds, so we need to convert to milliseconds
+	// Truncate timestamp to millisecond precision to match database precision and prevent rounding errors.
 	event.Timestamp = time.UnixMilli(event.Timestamp.UnixMilli())
 	
 	// Trim temp state before persisting


baptmont:
Hi @huangbug, thank you for bringing this up to our attention. Instead of truncating the saved value, we would prefer to add the respective tolerance to the "stale session" validation, since this does not affect every database?
Alternatively does the MYSQL driver you're using allow for configuring datetime precision?

huangbug:
```
CREATE TABLE `sessions` (
  `app_name` varchar(191) COLLATE utf8mb4_unicode_ci NOT NULL,
  `user_id` varchar(191) COLLATE utf8mb4_unicode_ci NOT NULL,
  `id` varchar(191) COLLATE utf8mb4_unicode_ci NOT NULL,
  `state` longtext COLLATE utf8mb4_unicode_ci,
  `create_time` datetime(3) DEFAULT NULL,
  `update_time` datetime(3) DEFAULT NULL,
  PRIMARY KEY (`app_name`,`user_id`,`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```
image

Hi @baptmont ,  from the table definition, we can see that the update_time field in the default sessions table only supports millisecond precision, and MySQL’s DATETIME type does not support nanosecond precision. Without fixing this issue, the error mentioned above will frequently occur when executing runner.run(xxx).

dpasiukevich:
What if we do this conversion to miliseconds only in the applyEvent()?

```
    return fmt.Errorf("stale session error: ...")
}
```

This way we will support the case when db provider has no nanos support.

huangbug:
Hi @dpasiukevich,  to ensure that the `update_time` field in the `sessions` table maintains millisecond precision, the timestamp must be converted to millisecond precision **before** executing `applyEvent()`. Otherwise, inserting an event and executing `applyEvent()` to update `update_time` will result in rounding, causing a loss of precision.

For example:

```sql
UPDATE sessions 
SET update_time = '2025-12-09 16:30:28.4915' 
WHERE app_name = 'role-play-chatbot-dialogue' 
  AND user_id = 'test_1765269021' 
  AND id = 'test_1765269021';
```
image

After execution, the `update_time` in the `sessions` table will be updated to:

```
2025-12-09 16:30:28.492
```

As shown, the portion exceeding millisecond precision is rounded.

dpasiukevich:
Honestly, I don't like the idea to have millisecond precision. If there are 2 agents working on the same session (e.g. main and sub agent), and they generated 2 events it will have a high chance to race (when event from one agent overwrites the event from the other, because both are in the same millisecond).

@huangbug could you verify if you use e.g precision `6` for time? https://github.com/go-gorm/mysql by setting `mysql.Config. DefaultDatetimePrecision`?

And if it sounds ok, for database provider we can **set this precision in gorm model tag for session timestamp**: https://github.com/google/adk-go/blob/0d0062df20e343382169205a1bd15a48b873062a/session/database/storage_session.go#L35

image

@huangbug @baptmont WDYT?

dpasiukevich:
The bumping precision example: https://github.com/google/adk-go/compare/main...dpasiukevich:adk-go:db_session_precision

@huangbug could you check if this looks good?

huangbug:
Hi @dpasiukevich, If the update_time field is defined as datetime(6), you must ensure that the value written to the sessions table uses microsecond precision rather than nanosecond precision. Otherwise, MySQL will still perform rounding during insertion, causing precision loss and leading to the same issue.
Importantly, the rounding is performed by MySQL itself, not by GORM.

dpasiukevich:
Correct. 

I've created a branch with the required changes: https://github.com/google/adk-go/compare/main...dpasiukevich:adk-go:db_session_precision

1. It has `gorm:"precision:6"` for time.Time type for all entities. So now mysql driver should create datetime(6) with microsecond precision.
2. There's a check in the current code which verifies if in-memory ADK object is in-sync with stored session. Which is failing right now in your case because MySQL ceiled it up. So in-memory object is treated as "outdated" because it kept the nanosecond precision in go code.
I've updated this check to round both stored_session and inmemory object to be both in milliseconds to fix this.

Could I ask to PTAL the code if possible?

Thanks!

huangbug:
Hi @dpasiukevich. I've reviewed the changes and everything looks good to me.

huangbug:
Hi @dpasiukevich, I've reviewed the changes and everything looks good to me.

dpasiukevich:
@huangbug I've added https://github.com/google/adk-go/pull/414 to explicitly set microsecond level precision. Thank you for noticing this!

PR #410, opened 2025-12-09
feat: adka2a conversion extendability and metadata handling

yarolegovich:
* Most of the changes are related to grounding, usage and custom metadata conversion. It gets stored in a2a event root metadata and then restored on back-conversion to session events. Since the primary use-case is event enrichment the callbacks takes an ADK-style custom readonly context `ExecutorContext`. Thought about reusing ReadonlyContext, but we don't have access to InvocationID and we also want to provide access to A2A data like RequestContext.
* Added `AfterEventCallback` to `adka2a.Executor` which allows to extend or redact events as they are emitted. Returning an error allows to stop an execution. I considered providing more flexible controls but decided it would be too error-prone for users. Use-cases which require modifying errors or inputs can be handled using existing a2a-go extensions points (RequestContextInterceptor or CallInterceptor).
* Updated a2a-go version.

### DIFF
--- a/server/adka2a/executor.go
+++ b/server/adka2a/executor.go
@@ -27,8 +27,14 @@ import (
 	"google.golang.org/adk/session"
 )
 
+// BeforeExecuteCallback is the callback which will be called before an execution is started.
+type BeforeExecuteCallback func(ctx context.Context, reqCtx *a2asrv.RequestContext) (context.Context, error)
+
 // AfterEventCallback is the callback which will be called after an ADK event is converted to an A2A event.
 type AfterEventCallback func(ctx ExecutorContext, event *session.Event, processed *a2a.TaskArtifactUpdateEvent) error
+
+// AfterExecuteCallback is the callback which will be called after an execution resolved into a completed or failed task.
+type AfterExecuteCallback func(ctx ExecutorContext, finalEvent *a2a.TaskStatusUpdateEvent, err error) error
 
 // ExecutorConfig allows to configure Executor.
 type ExecutorConfig struct {
@@ -38,9 +44,19 @@ type ExecutorConfig struct {
 	// RunConfig is the configuration which will be passed to [runner.Runner.Run] during A2A Execute invocation.
 	RunConfig agent.RunConfig
 
+	// BeforeExecuteCallback is the callback which will be called before an execution is started.
+	// It can be used to instrument a context or prevent the execution by returning an error.
+	BeforeExecuteCallback BeforeExecuteCallback
+
 	// AfterEventCallback is the callback which will be called after an ADK event is successfully converted to an A2A event.
 	// This gives an opportunity to enrich the event with additional metadata or abort the execution by returning an error.
+	// The callback is not invoked for errors originating from ADK or event processing. Such errors are converted to
+	// TaskStatusUpdateEvent-s with TaskStateFailed state. If needed these can be intercepted using AfterExecuteCallback.
 	AfterEventCallback AfterEventCallback
+
+	// AfterExecuteCallback is the callback which will be called after an execution resolved into a completed or failed task.
+	// This gives an opportunity to enrich the event with additional metadata or log it.
+	AfterExecuteCallback AfterExecuteCallback
 }
 
 var _ a2asrv.AgentExecutor = (*Executor)(nil)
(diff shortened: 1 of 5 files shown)


server/adka2a/metadata.go
```diff
@@ -69,21 +96,26 @@ func toEventMeta(meta invocationMeta, event *session.Event) (map[string]any, err
...
+	if event.UsageMetadata != nil {
+		v, err := converters.ToMapStructure(event.UsageMetadata)
 		if err != nil {
 			return nil, err
 		}
-		result[ToA2AMetaKey("grounding_metadata")] = v
+		result[metadataUsageKey] = v
+	}
+	if event.CustomMetadata != nil {
+		result[metadataCustomMetaKey] = event.CustomMetadata
```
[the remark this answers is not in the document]
yarolegovich: it's fine with how struct embedding works 🤷‍♂️

server/adka2a/metadata.go
```diff
@@ -103,3 +135,38 @@ func setActionsMeta(meta map[string]any, actions session.EventActions) map[strin
...
+		}
+		event.UsageMetadata = converted
+	}
+
+	event.CustomMetadata = ToCustomMetadata(taskInfo.TaskID, taskInfo.ContextID)
+	if um, ok := meta[metadataCustomMetaKey].(map[string]any); ok {
+		if event.CustomMetadata == nil {
+			event.CustomMetadata = make(map[string]any)
+		}
+		maps.Copy(event.CustomMetadata, um)
```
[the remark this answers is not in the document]
yarolegovich: struct embedding again 🤔

server/adka2a/metadata.go
```diff
@@ -103,3 +135,38 @@ func setActionsMeta(meta map[string]any, actions session.EventActions) map[strin
 	}
 	return meta
 }
+
+func processA2AMeta(a2aEvent a2a.Event, event *session.Event) error {
+	taskInfo, meta := a2aEvent.TaskInfo(), a2aEvent.Meta()
+
+	if gm, ok := meta[metadataGroundingKey].(map[string]any); ok {
```
dpasiukevich, line 141: Let's add the test case when there's `nil` value for metadataGroundingKey. To make sure the code handles this (not sure how artificial this case is).

If this makes sense, let's add this check for all added fields `metadataGroundingKey`, `metadataUsageKey` etc.

server/adka2a/metadata.go
```diff
@@ -18,41 +18,68 @@ import (
 	"context"
 	"maps"
 
+	"github.com/a2aproject/a2a-go/a2a"
 	"github.com/a2aproject/a2a-go/a2asrv"
 
+	"google.golang.org/genai"
```
dpasiukevich, line 23:
```suggestion
	"github.com/a2aproject/a2a-go/a2a"
	"github.com/a2aproject/a2a-go/a2asrv"
	"google.golang.org/genai"
```

nit: Let's keep the imports in 3 blocks (stdlib + 3P + adk packages). And could you update in other places as well?

server/adka2a/executor_context.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+type executorContext struct {
+	context.Context
+	meta        invocationMeta
+	session     session.ReadonlyState
+	userContent *genai.Content
+}
+
+var _ ExecutorContext = (*executorContext)(nil)
```
dpasiukevich: nit: up to your preference, but this line can be removed, because this check is already in lines 47-48

server/adka2a/executor_context.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+
+	"github.com/a2aproject/a2a-go/a2asrv"
+
+	"google.golang.org/genai"
+
+	"google.golang.org/adk/session"
+)
+
+type ExecutorContext interface {
```
dpasiukevich, line 29: Could you add a meaningful godoc comment for this interface?

server/adka2a/executor.go
```diff
@@ -110,29 +120,33 @@ func (e *Executor) Cancel(ctx context.Context, reqCtx *a2asrv.RequestContext, qu
...
 		}
 
-		a2aEvent, err := processor.process(ctx, event)
-		if err != nil {
-			event := processor.makeTaskFailedEvent(fmt.Errorf("processor failed: %w", err), event)
-			if eventSendErr := q.Write(ctx, event); eventSendErr != nil {
-				return fmt.Errorf("processor error event write failed: %w, %w", err, eventSendErr)
+		a2aEvent, pErr := processor.process(ctx, adkEvent)
+		if pErr == nil && a2aEvent != nil && e.config.AfterEventCallback != nil {
+			pErr = e.config.AfterEventCallback(ctx, adkEvent, a2aEvent)
```
dpasiukevich, line 145: Does it make sense to allow passing the original error to the `AfterEventCallback`?

yarolegovich: Thanks for raising this. I initially thought that errors can be intercepted using a2asrv.CallInterceptor but now realized that execution is happening in a detached context, and interceptors won't be invoked if no client is listening.

Added [BeforeExecuteCallback and AfterExecuteCallback](https://github.com/google/adk-go/pull/410/commits/9cb5a15ca6c5185f75ea1079e4c7c5490092fd9c) to make it possible to capture execution spans and reliably observe task completion events for both failed and successful runs.

PR #417, opened 2025-12-12
Add MCP session reconnection with Ping health check

Nerja:
This PR adds resilience to the MCP toolset by automatically attempting to reconnect when the connection to the MCP server is lost. When ListTools() or CallTool() fails with an error that indicates connection issues, the client first pings the server to verify the connection is actually dead before attempting a reconnect.

### DIFF
--- a/tool/mcptoolset/client.go
+++ b/tool/mcptoolset/client.go
@@ -141,7 +141,7 @@ func shouldRefreshConnection(err error) bool {
 			return true
 		}
 	}
-	// Temporary workaround for session not found errors.
+	// TODO: replace with mcp.ErrSessionMissing when it's available https://github.com/modelcontextprotocol/go-sdk/issues/715
 	if strings.Contains(err.Error(), "session not found") {
 		return true
 	}


dpasiukevich:
Thanks for the PR!

I'm thinking -- now we will always be making 2 network requests: ping + list/call_tool.

What do you think about these approaches:
1. Making a detached goroutine which will ping over an interval configured by user (with some default) and refresh the connection. And we'd probably would need to expose Toolset struct (which is now called `set`) with Close() method so users could clean up this goroutine.
2. Or another option -- handling `mcp.ErrConnectionClosed`. If list/call_tool return this error, then mcptoolset will refresh the connection. Then no public API changes needed, but we'd still probably would need to call ping before the refresh and do some gradual fallback.

What do you think?

Nerja:
@dpasiukevich I updated the PR to only ping when there is an error. What do you think of the current approach? I noticed that not all errors are mcp.ErrConnectionClosed (for instance, the Close() call in the tests does not cause mcp.ErrConnectionClosed)   so made the compromise to issue a ping when there is any error. If ping succeeds, then we will not create a new connection/session.

dpasiukevich:
@Nerja could I ask to check:

1. Instead of reconnecting on every error (this may lead to infinite loops I'd expect, e.g. with validation error). If we reconnect only if `errors.Is(err, mcp.ErrConnectionClosed)`. I did some experiments earlier and this error type looks sufficient (https://github.com/modelcontextprotocol/go-sdk/pull/719#issuecomment-3646681481).

Could you confirm if I'm correct here?

2. Right now the connection refresh logic is spread within Toolset and Tool code. Do you think if we'd try to move it out to a separate entity?
As example we can define something like this:
```
type MCPClient interface {
  CallTool(context.Context, *mcp.CallToolParams) (mcp.CallToolResult, error)
  ListTools(context.Context, *mcp.ListToolsParams) (mcp.ListToolsResult, error)
}

type connectionRefresher struct { ... } // handles the session refresh
```

And MCP Toolset implementation (in Tools() method which calls mcp.ListTools) and mcp.Tool (which calls mcp.CallTool), they both take in MCPClient interface implemented by connectionRefresher.

@Nerja what do you think about it?

Thanks!

dpasiukevich:
cc @majiayu000 I'd appreciate if you could share your thoughts.

majiayu000:
Superseded by #484 to implement the requested refactoring and strict error handling (as I don't have push permissions to this branch).

majiayu000:
Thanks @dpasiukevich for the detailed review!

I have implemented the suggested refactoring in #484 (due to push permission restrictions on this branch).

**Key Changes in #484:**
1.  **Architecture Decoupling**: Extracted `MCPClient` interface and `reconnectingClient` struct to decouple session management from `Toolset`/`Tool`.
2.  **Strict Error Handling**: As you suggested, reconnection is now strictly limited to `mcp.ErrConnectionClosed` to avoid infinite loops on validation errors.

Please review the new PR when you have a chance.

Nerja:
@dpasiukevich 

> @Nerja could I ask to check:
> 
> 1. Instead of reconnecting on every error (this may lead to infinite loops I'd expect, e.g. with validation error). If we reconnect only if `errors.Is(err, mcp.ErrConnectionClosed)`. I did some experiments earlier and this error type looks sufficient ([mcp: export ErrSessionMissing for session termination detection modelcontextprotocol/go-sdk#719 (comment)](https://github.com/modelcontextprotocol/go-sdk/pull/719#issuecomment-3646681481)).
> 
> Could you confirm if I'm correct here?
> 
> 2. Right now the connection refresh logic is spread within Toolset and Tool code. Do you think if we'd try to move it out to a separate entity?
>    As example we can define something like this:
> 
> ```
> type MCPClient interface {
>   CallTool(context.Context, *mcp.CallToolParams) (mcp.CallToolResult, error)
>   ListTools(context.Context, *mcp.ListToolsParams) (mcp.ListToolsResult, error)
> }
> 
> type connectionRefresher struct { ... } // handles the session refresh
> ```
> 
> And MCP Toolset implementation (in Tools() method which calls mcp.ListTools) and mcp.Tool (which calls mcp.CallTool), they both take in MCPClient interface implemented by connectionRefresher.
> 
> @Nerja what do you think about it?
> 
> Thanks!

Updated, not so sure about relying on mcp.ErrConnectionClosed though. It looks like other errors related to network could be thrown that are not wrapped: For example, in SSE transport when the server becomes unreachable. Will provide an example shortly.

dpasiukevich:
@majiayu000 @Nerja you sent 2 really proper PRs! Let's close the duplicates and proceed with this one.

I will make sure you both are listed as authors of the commit for integrity.

tool/mcptoolset/client.go
```diff
@@ -0,0 +1,170 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// First, try ping to confirm connection is dead
+	if c.session != nil {
+		if err := c.session.Ping(ctx, &mcp.PingParams{}); err == nil {
+			// Connection is actually alive, don't refresh
+			return c.session, false, nil
+		}
+		c.session.Close()
```
dpasiukevich: Let's fix the lint failures. Let's log the error here (in the app code) and t.Fail/Error in the test code?

tool/mcptoolset/client.go
```diff
@@ -0,0 +1,170 @@
...
+func withRetry[T any](ctx context.Context, c *connectionRefresher, fn func(*mcp.ClientSession) (T, error)) (T, bool, error) {
+	var zero T
+
+	session, err := c.getSession(ctx)
+	if err != nil {
+		return zero, false, err
+	}
+
+	result, err := fn(session)
+	if err != nil {
```
dpasiukevich, line 122: @Nerja you mentioned, that mcp.ErrConnectionClosed does not capture all such session cases? If there's a chance, could you share more?

I'm still favoring the approach to refreshConnection only on known, defined errors. For example, refreshing on validation or server internal errors doesn't have any benefit.

@majiayu000 WDYT?

tool/mcptoolset/set_test.go
```diff
@@ -307,3 +307,63 @@ func TestToolFilter(t *testing.T) {
...
+	return conn, err
+}
+
+type reconnectableTransport struct {
+	server *mcp.Server
+}
+
+func (rt *reconnectableTransport) Connect(ctx context.Context) (mcp.Connection, error) {
+	ct, st := mcp.NewInMemoryTransports()
+	_, err := rt.server.Connect(context.Background(), st, nil)
```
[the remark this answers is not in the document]
dpasiukevich: nit: this is a slight improvement, let's apply this suggestion?

tool/mcptoolset/client.go
Nerja: Created this small example: https://github.com/Nerja/go-mcp-sdk-conn-error-test

dpasiukevich: Thanks, I'm running the example but for 15 mins there are no failures yet. Need to wait more.

@Nerja could you share the logs if you have those? Do those errors have a type in the mcp lib (like mcp.ErrConnectionClosed) or they are untyped raw go errors?

dpasiukevich: I did some more experiments and was able to get this entry:

```ListTools #11 FAILED: calling "tools/list": sending "tools/list": failed to connect (session ID: AZPXUKSQ3VZ2GYVMW3STYMRPZN): session not found (type: *fmt.wrapError) isConnectionClosed=false```

@Nerja let's do it like this:
1. we will refresh the connection only on 2 cases only:
  a. `errors.Is(err, mcp.ErrConnectionClosed)`
  b. `strings.Contains(err.Error(), "session not found")` -- temporarily

2. Let me ask MCP-SDK folks to introduce `mcp` in https://github.com/modelcontextprotocol/go-sdk/issues/715. After it's available we replace 1b to use this err code.

WDYT?

Comment (Nerja) [reply tool/mcptoolset/client.go:122]: @dpasiukevich I assume that you want to drop the Ping check if we rely on the errors returned by the MCP operations instead?

Comment (dpasiukevich) [reply tool/mcptoolset/client.go:122]: Let's remove it. 

Both ways are ok now once it's done only for mcp.ErrConnectionClosed and "session not found".

By keeping the ping we solve the case when parallel tool calls may refresh a healthy session (e.g. in parallel tool calls, first will refresh closed session, but 2nd parallel tool will refresh a good session created by 1st).

But except for this case, the session will be always closed so pinging may be excessive.

Comment (Nerja) [reply tool/mcptoolset/client.go:122]: @dpasiukevich updated the PR to only consider reconnecting on certain errors. I kept the Ping inside the refreshConnection function to avoid the scenario that you mentioned. I personally feel like it is not adding too much complexity and it avoids causing further issues by closing good newly created sessions. Let me know if you are OK with this or if you want me to further simplify and remove the Ping completely.

Comment (dpasiukevich) [reply tool/mcptoolset/client.go:122]: SGTM. Keeping Ping or not has it's pros/cons. Both are ok for me as now we do it only for session related errors.

**Comment by ejgdr** [comment]
The decision to only reconnect on specific error types (ErrConnectionClosed and "session not found") makes sense, because it avoids the infinite loop scenario with validation errors. Keeping the Ping check in refreshConnection to handle parallel tool calls seems like a good idea for safety measure. Great job!

**Comment by majiayu000** [comment]
Thanks for all the review!

**Comment by dpasiukevich** [comment]
@Nerja @majiayu000 

One small thing: noticed TestCallToolReconnection can flake. In my local runs it's ~1-2 out of 10000 so not that bad.

`set_test.go:384: Tool call after reconnect failed: failed to call MCP tool "get_weather" with err: calling "tools/call": EOF`

https://github.com/google/adk-go/actions/runs/21245011072/job/61131972067

I suspect it happens when `fnTool.Run` happens earlier than `lastConn.Close()` is fully finished. If that's the case then it should not be the issue for prod.

```
	// Kill the transport by closing the connection.
	if err := spyTransport.lastConn.Close(); err != nil {
		t.Fatalf("Failed to close connection: %v", err)
	}

	// Call the tool - should reconnect and succeed.
	fnTool := tools[0].(toolinternal.FunctionTool)
	result, err := fnTool.Run(toolCtx, map[string]any{"city": "Paris"})
```

WDYT?

majiayu000:
@dpasiukevich I will test this later! Thanks!

dpasiukevich:
In CI it actually happens with 70-80% chance. Probably when I tested locally I just ran that one failing test case.

dpasiukevich:
I suspect it's because in tests we use NewInMemoryTransports which underneath uses `net.Pipe()` which returns `io.EOF`

It's all seems in line with https://github.com/google/adk-go/pull/417#issuecomment-3783796150.

We test if the reconnection is successful by making a 2nd call after closing the connection and expect it to be without error. But sometimes the connection is closing too slowly and we already make a 2nd request which fails.

Nerja:
> I suspect it's because in tests we use NewInMemoryTransports which underneath uses `net.Pipe()` which returns `io.EOF`
> 
> It's all seems in line with [#417 (comment)](https://github.com/google/adk-go/pull/417#issuecomment-3783796150).
> 
> We test if the reconnection is successful by making a 2nd call after closing the connection and expect it to be without error. But sometimes the connection is closing too slowly and we already make a 2nd request which fails.

I was able to reproduce the issue locally by running the test 1000 times. By adding io.EOF check, the test did not fail once when I tried to run it 10 000 times. Open a PR that adds io.EOF in https://github.com/google/adk-go/pull/505. In general, this is the downside of maintaining a list of explicit error for which we attempt a Ping.

dpasiukevich:
I agree. On one side making Ping on every error is also excessive (there could be business errors, validation errors).

Let's see short/medium term, if there will be other common errors for which we'd need to refresh. If the list grows, let's drop it and just ping.

PR #420, opened 2025-12-15
fix: remote agent responses are not stored in streaming mode

yarolegovich:
Events marked as `Partial` are not stored in the sessions store.
* Updated comments to make it more clear.
* Updated A2A remote-agent to aggregate text and thoughts from partial events and emit them with or before the terminal events.
* Long-running functions calls are emitted as well because a response is expected to be delivered as a follow-up function call.

Before the fix remoteagent responses in streaming mode were lost unless a Task was emitted as the terminal event (which is not the case for adka2a.Executor)

### DIFF
--- a/server/adka2a/parts.go
+++ b/server/adka2a/parts.go
@@ -89,6 +89,9 @@ func updatePartsMetadata(parts []a2a.Part, update map[string]any) {
 				parts[i] = p
 			}
 			meta = p.Metadata
+		default:
+			// TODO: log unknown part type warning (should never happen)
+			continue
 		}
 		maps.Copy(meta, update)
 	}


baptmont (dismissed):

server/adka2a/parts.go
```diff
@@ -66,6 +67,33 @@ func ToA2AParts(parts []*genai.Part, longRunningToolIDs []string) ([]a2a.Part, e
...
+			}
+			meta = p.Metadata
+		case a2a.DataPart:
+			if p.Metadata == nil {
+				p.Metadata = make(map[string]any)
+				parts[i] = p
+			}
+			meta = p.Metadata
+		}
+		maps.Copy(meta, update)
```
baptmont, line 96: if part is not any of the ones in the switch, meta will be nil and copy will panic.

server/adka2a/parts.go
```diff
@@ -66,6 +67,33 @@ func ToA2AParts(parts []*genai.Part, longRunningToolIDs []string) ([]a2a.Part, e
 	return result, nil
 }
 
+func updatePartsMetadata(parts []a2a.Part, update map[string]any) {
+	for i, part := range parts {
+		var meta map[string]any
+		switch p := part.(type) {
+		case a2a.TextPart:
```
baptmont, line 74: nit: case a, b:

https://go.dev/ref/spec#Expression_switches

server/adka2a/parts.go
yarolegovich: better safe than sorry 🙂

server/adka2a/parts.go
yarolegovich: doesn't work when we need to access typecasted variable fields

dpasiukevich (approved):

Issue #418, opened 2025-12-13
adkrest and adka2a should introduce their own Config types


rakyll:
As a follow up to https://github.com/google/adk-go/issues/415, we should introduce a Config type in the package rather than directly using `*launcher.Config`. We don't have to land both changes in the same PR but  the launcher package isn't the right approach. 

We need a Config type in adkrest.

Kenasvarghese:
@rakyll I’d like to contribute to this issue.

My plan is to introduce a `Config` type inside the `adkrest` package (for example, in `handler.go`) so that `NewHandler` can take `*adkrest.Config` instead of depending directly on `*launcher.Config`. This would decouple adkrest from launcher while still providing the necessary configuration.

Would you prefer defining the `Config` struct directly in `handler.go`, or in a separate `config.go` file under adkrest?

Kenasvarghese:
Please review @rakyll @dpasiukevich

karolpiotrowicz:
Both halves of this have landed, so I'm closing it.

`adkrest` owns its own type: [`ServerConfig`](https://github.com/google/adk-go/blob/440e35202c325a7310557e3bacecdeb3ad07b989/server/adkrest/handler.go#L114) is constructed by [`NewServer(cfg ServerConfig)`](https://github.com/google/adk-go/blob/440e35202c325a7310557e3bacecdeb3ad07b989/server/adkrest/handler.go#L52), with no `*launcher.Config` in the signature.

`adka2a` owns two: [`RunnerConfig`](https://github.com/google/adk-go/blob/440e35202c325a7310557e3bacecdeb3ad07b989/server/adka2a/v2/executor.go#L86) and [`ExecutorConfig`](https://github.com/google/adk-go/blob/440e35202c325a7310557e3bacecdeb3ad07b989/server/adka2a/v2/executor.go#L96), the latter carrying a `runner.Config` rather than the launcher type.

Please reopen if the shape that landed isn't what you had in mind.

Issue #415, opened 2025-12-11
Move restapi server controllers and routers to internal


dpasiukevich:
Ideally we should only expose `func NewHandler(config *launcher.Config) http.Handler`.

Keeping controllers/routers internally gives us more flexibility to refactor/update code. And keeping controller/routers public looks excessive.

Kenasvarghese:
@dpasiukevich I’d like to contribute to this issue. Since it’s my first time contributing to this project, I want to confirm that I’m understanding the goal correctly.
Is the intention to move the adkrest/controllers and adkrest/internal/routers logic into the same package as NewHandler, so they can be unexported and the only public API remains NewHandler(config *launcher.Config)?

kdroste-google:
> [@dpasiukevich](https://github.com/dpasiukevich) I’d like to contribute to this issue. Since it’s my first time contributing to this project, I want to confirm that I’m understanding the goal correctly. Is the intention to move the adkrest/controllers and adkrest/internal/routers logic into the same package as NewHandler, so they can be unexported and the only public API remains NewHandler(config *launcher.Config)?

Hi, @Kenasvarghese, yes, the idea is to hide as much as possible. It would be great if you could contribute!  
Please mind that "internal" directory is exactly what's needed - please see [docs](https://go.dev/doc/modules/layout#package-or-command-with-supporting-packages) .

Kenasvarghese:
Update: routers was already under adkrest/internal.
While moving `adkrest/controllers` under `adkrest/internal/`, I noticed that
`EncodeJSONResponse` (defined in `controllers/handler.go`) is currently used
by `webui`, which lives outside the `internal` tree. This creates a Go `internal` visibility conflict.

My plan is to extract `EncodeJSONResponse` into a shared helper package
(e.g., `adkrest/response`) that both `adkrest` and `webui` can import,
while keeping controllers internal. `NewHandler(config *launcher.Config) http.Handler`
will remain the only public API of adkrest.

@kdroste-google Looking for guidance on the intended API boundary before continuing.

PR #422, opened 2025-12-16
Introduce Config type for adkrest

Kenasvarghese:
This PR introduces a `Config` type in the `adkrest` package to define
session, artifact, and agent dependencies.

As a follow-up to #415, `adkrest` should not directly depend on
launcher.Config. Defining a package-local configuration type provides
a clearer and more appropriate abstraction for REST integration.

Related issue: #418


server/adkrest/handler.go
```diff
@@ -21,15 +21,14 @@ import (
...
-	"google.golang.org/adk/cmd/launcher"
 	"google.golang.org/adk/internal/telemetry"
 	"google.golang.org/adk/server/adkrest/controllers"
 	"google.golang.org/adk/server/adkrest/internal/routers"
 	"google.golang.org/adk/server/adkrest/internal/services"
 )
 
 // NewHandler creates and returns an http.Handler for the ADK REST API.
-func NewHandler(config *launcher.Config, sseWriteTimeout time.Duration) http.Handler {
+func NewHandler(config *Config, sseWriteTimeout time.Duration) http.Handler {
```
rakyll: Can we introduce an SSEWriteTimeout in Config and move the argument there?

rakyll (dismissed):
LGTM, cc @mazas-google @dpasiukevich  to approve the breaking change.

IAmSurajBobade:
@Kenasvarghese can you fix adkrest example. https://github.com/google/adk-go/blob/main/examples/rest/main.go#L69

Kenasvarghese:
Please Review @IAmSurajBobade

Kenasvarghese:
@rakyll @IAmSurajBobade @dpasiukevich Do I need to implement validation and error for the config

server/adkrest/config.go
```diff
@@ -0,0 +1,46 @@
...
+// Config defines the services and loaders required by the adkrest package.
+type Config struct {
+	SessionService  session.Service
+	ArtifactService artifact.Service
+	AgentLoader     agent.Loader
+	SSEWriteTimeout time.Duration
+}
+
+// Validate validates the config
+func (c *Config) Validate() error {
```
rakyll: This shouldn't be exported. func (c *Config) validate() error

server/adkrest/handler.go
```diff
@@ -16,20 +16,22 @@ package adkrest
...
 	"google.golang.org/adk/internal/telemetry"
 	"google.golang.org/adk/server/adkrest/controllers"
 	"google.golang.org/adk/server/adkrest/internal/routers"
 	"google.golang.org/adk/server/adkrest/internal/services"
 )
 
 // NewHandler creates and returns an http.Handler for the ADK REST API.
-func NewHandler(config *launcher.Config, sseWriteTimeout time.Duration) http.Handler {
+func NewHandler(config *Config) (http.Handler, error) {
+	err := config.Validate()
```
rakyll: Fix the style please:

```
if err := config.validate(); err != nil {
   return nil, err
}
```

Kenasvarghese:
Re-requesting review for this pull request. Please let me know if there are any remaining concerns.
@rakyll

rakyll (dismissed):
cc @dpasiukevich @mazas-google (note that this is a breaking change)

Kenasvarghese:
@rakyll  
Rebased with latest main and resolved merge conflicts.
No functional changes — just conflict resolution.
Would appreciate a quick re-approval. Thanks!

Kenasvarghese:
This PR has been approved and updated with the latest main.
Could someone help with the remaining approval / merge when available?
@dpasiukevich @mazas-google

wolo-lab:
I am sorry — this one is squarely on us. You had an LGTM, you rebased when it conflicted, and you asked four separate times over two and a half months for the breaking-change approval that never came. That is a bad experience and not the standard we want to hold.

The decoupling itself shipped in #660, which introduced ServerConfig in server/adkrest/handler.go with the same field set you proposed and removed the cmd/launcher import from the package. With that in main there is nothing left for this branch to do.

One thing your version had that main does not: validate() rejecting nil services at construction. Today NewServer accepts a nil SessionService and panics on the first request instead. We will file that as a follow-up and point at this PR for the design.

We have changed how we work the PR queue since then and are reviewing incoming changes much more actively.

PR #429, opened 2025-12-18
Fix test cases failure due to the time zone

git-hulk:
Currently, the following test case is failing on my side due to using `time.UnixMicro` to truncate the precision:

```shell
    service_test.go:363: Get session events mismatch: (-want +got):
          database.events{
                &{
                        LLMResponse:  {},
        -               ID:           "4",
        +               ID:           "3",
        -               Timestamp:    s"0001-01-01 00:00:00.000004 +0000 UTC",
        +               Timestamp:    s"0001-01-01 08:05:43.000003 +0805 +0805",
                        InvocationID: "",
                        Branch:       "",
                        ... // 3 identical fields
                },
```

The root cause should be `time.UnixMicro` will change the time.Time to the local timezone. To work around this, use the fixed time with the local timezone instead of UTC.

It's related to the change in PR #414.

### DIFF
--- a/session/database/service_test.go
+++ b/session/database/service_test.go
@@ -190,7 +190,7 @@ func Test_databaseService_Get(t *testing.T) {
 		return s
 	}
 
-	now := time.Now()
+	now := time.Date(2025, time.January, 1, 12, 0, 0, 0, time.Local)
 	setupGetWithConfig := func(t *testing.T) *databaseService {
 		t.Helper()
 		s := emptyService(t)


git-hulk:
After diving deep, I found the root cause is that SQLite only supports the ISO-8601 string comparison, see [stackoverflow](https://stackoverflow.com/questions/1975737/sqlite-datetime-comparison). So, when storing and filtering rows with different time zones, it might retrieve the wrong value.

cc @dpasiukevich

dpasiukevich:
Thanks for checking that!

But will this issue affect actual SQLite based session service?

git-hulk:
> But will this issue affect actual SQLite based session service?

I think it only affects test cases that use `time.Time{}` to create a time value because it uses the UTC time zone. And it won't be an issue if we use the same time zone when inserting and querying.

dpasiukevich:
@git-hulk this should be fixed by https://github.com/google/adk-go/pull/448. Let me know if it looks good on your end as well. Thanks!

git-hulk:
> @git-hulk this should be fixed by #448. Let me know if it looks good on your end as well. Thanks!

@dpasiukevich Thanks for your fix, it's perfect for me.

PR #430, opened 2025-12-18
fix: return error instead of panic when ArtifactService is nil (#283)

JEETDESAI25:
Addressing issue: #283

Fixes a nil pointer dereference panic in loadartifactstool that occurred when the tool was used without configuring an ArtifactService in the Runner. The tool now returns `ErrArtifactServiceNotInitialized` instead of panicking.

**Changes:**
  - Added `ErrArtifactServiceNotInitialized` error in loadartifactstool
  - Added nil check in `ProcessRequest()`
  - Fixed context wrappers to return nil when underlying ArtifactService is nil

  **Test:** Added `TestLoadArtifactsTool_ProcessRequest_NilArtifactService`.


wolo-lab:
Thank you for working on it. Closing as duplicate. The PR #1390 is going to solve #283.

PR #433, opened 2025-12-19
Implement RWLock to guard the artifact delta update

git-hulk:


karolpiotrowicz:
Thanks @git-hulk. The concern is still valid, but the code moved: `internal/toolinternal/context.go` / `internalArtifacts` were refactored away, and the logic now lives in `agent/callback_context.go` (`trackedArtifacts.Save`), which still carries the same `// TODO: RWLock …`. Could you re-target the fix there (and use `RLock` on the read path)? Happy to review once rebased.

git-hulk:
@karolpiotrowicz Thanks for your review. I have changed it accordingly.

internal/context/context_test.go
```diff
@@ -48,6 +76,73 @@ func TestCallbackContext(t *testing.T) {
...
+		}()
+	}
+	wg.Wait()
+	close(errs)
+
+	for err := range errs {
+		t.Fatalf("Save returned error: %v", err)
+	}
+	if got := artifactDelta["file.txt"]; got != saves {
+		t.Fatalf("expected version %d after concurrent saves, got %d", saves, got)
```
karolpiotrowicz, line 142: This assertion is the only thing standing between a regression in the write-lock re-check and a green build, and it catches that regression roughly a quarter of the time.

I removed the `if current, ok := ...; !ok || resp.Version > current` re-check from `Save` — leaving the fast path and an unconditional write, which is the natural way someone "simplifies" this later — and ran the test twelve times with the CI flags (`go test -race -mod=readonly -count=1 -shuffle=on`). It failed three times and passed nine.

The reason is that there is no start barrier and `version.Add(1)` hands out versions in increasing order, so the goroutines tend to drain close to spawn order and the out-of-order arrival the assertion needs is left to the scheduler. Releasing the goroutines from a barrier, and allocating each one's version before that barrier, would make the ordering adversarial rather than hopeful.

Two smaller things in the same test: `saves` and `max(version)` are both 100, so an implementation that recorded the number of saves rather than the newest version would pass unnoticed, and both tests use a single artifact name, so one that ignored `name` and kept a single scalar would pass too.

karolpiotrowicz (changes requested):
The blocker here is not the code: this branch can no longer be merged or rebased, and it needs re-authoring against a file that did not exist when you wrote it. There is also one real regression in the locking, which I will come to. Apologies that it took us this long to come back to you — you did what the June review asked for, and the ground moved underneath it again.

## Why it cannot land in this shape

`agent/callback_context.go`, the file this PR edits, was deleted from `main` by the release that split the module to `google.golang.org/adk/v2` ([#1109](https://github.com/google/adk-go/pull/1109), 2026-06-30) — about two weeks after you pushed. The wrapper now lives in `agent/common_context.go`, and every import path in the module gained a `/v2` suffix. GitHub reports the branch as conflicting and not rebaseable, and the base is 172 commits behind.

Neither mechanical resolution gives a working tree. Taking `main`'s side drops the change completely. Taking the branch's side resurrects the file with five imports on the old module path, and it does not compile. So this is a re-author rather than a rebase: the change has to move onto `trackedArtifacts.Save` in `agent/common_context.go`, and the test onto the current `internal/context` file.

The good news is that the work is still needed. Both defects you targeted are live on `main` today. Running an equivalent of your own test against [17221c6a](https://github.com/google/adk-go/commit/17221c6ab5e396d8c5758d73aeeb5ea8cc26a339) reproduces the version-monotonicity failure and a data race at [`agent/common_context.go:491-492`](https://github.com/google/adk-go/blob/17221c6ab5e396d8c5758d73aeeb5ea8cc26a339/agent/common_context.go#L491-L492), which still carries the same `// TODO: RWLock` comment you set out to remove.

## The one regression, worth carrying into the re-authored version

[`agent/callback_context.go:256`](https://github.com/google/adk-go/blob/9d7988f968b9deaf511091d6e6d65e4d077a5f5c/agent/callback_context.go#L256-L262) takes `RLock` without a `defer`, and line 258 dereferences `resp.Version` while that lock is held. Because `&&` short-circuits, the dereference only happens when the artifact name is already in the delta map. If an `Artifacts` implementation returns `(nil, nil)`, the panic escapes with the read lock still held, [`tool/functiontool/function.go:187-191`](https://github.com/google/adk-go/blob/9d7988f968b9deaf511091d6e6d65e4d077a5f5c/tool/functiontool/function.go#L187-L191) recovers it, and the next `Save` that needs the write lock blocks forever.

On `main` today the same nil dereference panics, gets recovered, and everything carries on. With this change it wedges the wrapper permanently. Same probe, both revisions:

<details><summary>Reproduction — deadlocks with the PR applied, completes without it</summary>

```go
// zzdead/deadlock_test.go
package zzdead

import (
	"context"
	"testing"
	"time"

	"google.golang.org/genai"

	"google.golang.org/adk/agent"
	"google.golang.org/adk/artifact"
	icontext "google.golang.org/adk/internal/context"
)

// Returns (nil, nil) the first time, then a real response.
type nilOnce struct{ calls int }

func (f *nilOnce) Save(ctx context.Context, name string, d *genai.Part) (*artifact.SaveResponse, error) {
	f.calls++
	if f.calls == 1 {
		return nil, nil
	}
	return &artifact.SaveResponse{Version: int64(f.calls)}, nil
}
func (f *nilOnce) List(context.Context) (*artifact.ListResponse, error) { return nil, nil }
func (f *nilOnce) Load(ctx context.Context, n string) (*artifact.LoadResponse, error) {
	return nil, nil
}
func (f *nilOnce) LoadVersion(ctx context.Context, n string, v int) (*artifact.LoadResponse, error) {
	return nil, nil
}

func TestPanicUnderLockThenSaveAgain(t *testing.T) {
	// Pre-seeded, so `ok` is true at line 258 and resp.Version is evaluated
	// under the read lock rather than under the deferred write lock.
	actions := &session.EventActions{StateDelta: map[string]any{}, ArtifactDelta: map[string]int64{"f": 1}}
	inv := icontext.NewInvocationContext(t.Context(), icontext.InvocationContextParams{Artifacts: &nilOnce{}})
	tc := agent.NewToolContext(inv, "call-1", actions, nil)

	func() {
		defer func() {
			if r := recover(); r != nil {
				t.Logf("recovered first Save panic, as functiontool.Run does: %v", r)
			}
		}()
		tc.Artifacts().Save(context.Background(), "f", nil)
		t.Fatal("expected the first Save to panic")
	}()

	done := make(chan struct{})
	go func() {
		defer close(done)
		tc.Artifacts().Save(context.Background(), "f", nil)
	}()
	select {
	case <-done:
		t.Log("second Save COMPLETED")
		t.Fatal("second Save DEADLOCKED")
	}
}
```

Run with `go test -count=1 -v -run TestPanicUnderLockThenSaveAgain ./zzdead/`.

With the PR applied: `second Save DEADLOCKED`, FAIL after 5.00s. Without it: `second Save COMPLETED`, PASS.

</details>

The nil dereference itself predates your change, so that part is not yours to answer. What is new is that it now wedges instead of recovering.

Worth knowing: no artifact service in the repo can return `(nil, nil)` — every return in `artifact/inmemory.go` and `artifact/gcsartifact/service.go` is either an error or a non-nil response. Triggering this needs a caller-supplied `artifact.Service`. That is why I am flagging it for the re-authored version rather than treating it as urgent.

One option, which I ran rather than guessed: dropping the read-lock fast path at lines 256-262 entirely leaves the write-lock path, which already re-reads and re-tests under the exclusive lock, so the result is the same on every input. That variant builds, passes your two tests five times under `-race`, passes 35 packages with no failures, and the reproduction above completes on it. I am offering it as one way out rather than as the way — the fast path is there because a reviewer asked for `RLock` on the read path, and a simple `defer` restructuring keeps that request satisfied.

## Two things for the re-authored version

The mutex is on the wrapper, not on the data it protects. `mu` is a field of `trackedArtifacts`, but `ArtifactDelta` belongs to the caller-supplied `*session.EventActions`, so two wrappers over one `EventActions` get two mutexes over one map. Two calls to `agent.NewCallbackContextWithArtifactTracking` sharing one `actions`, driven concurrently, still race with this change applied. I am not asking you to fix that here — `main` is racy in the same configuration today, no caller in the repo drives two wrappers over one map concurrently, and your change strictly reduces the set of racy configurations. But if the lock is going to move anyway, `session.EventActions` is where it belongs.

The invariant to preserve, whatever shape the fix takes: for a given artifact name, the recorded version must never move backwards, and every writer that can reach that map must serialize against every other one.

Separately, one pre-existing thing I noticed and will file on its own: the `artifactDelta` map created at [`internal/llminternal/base_flow.go:556`](https://github.com/google/adk-go/blob/9d7988f968b9deaf511091d6e6d65e4d077a5f5c/internal/llminternal/base_flow.go#L556) is threaded through six callback contexts and never assigned to any event, so artifact saves performed from model callbacks are dropped before they reach a session. Not yours, and not something to address in this PR.

PR #434, opened 2025-12-19
fix: always include events with empty branch

dpasiukevich:
When there's this setup: LLMAgent1 -> ParallelAgent -> LLMAgent2

Then invocation_branch is: llmagent1.llmagent2
But event.branch is "" because it's the event generated from first user input.

python ref: https://github.com/google/adk-python/blob/0b1cff2976d1c04acf3863f76107b05d1cec448f/src/google/adk/flows/llm_flows/contents.py#L634


dpasiukevich:
closing in favor of https://github.com/google/adk-go/pull/437

PR #435, opened 2025-12-21
fix(server): prevent panic in SSE handler on error

### LINKED ISSUE #383: adkrest server controller panic when llm generate returns error
**Describe the bug**
`adkrest server controller panic when llm generate returns error`

**To Reproduce**
Is continue correct? or maybe return? 

// server/adkrest/controllers/runtime.go
```go
	resp := r.Run(req.Context(), runAgentRequest.UserId, runAgentRequest.SessionId, &runAgentRequest.NewMessage, *rCfg)

	rw.WriteHeader(http.StatusOK)
	for event, err := range resp {
		if err != nil {
			_, err := fmt.Fprintf(rw, "Error while running agent: %v\n", err)
			if err != nil {
				return newStatusError(fmt.Errorf("write response: %w", err), http.StatusInternalServerError)
			}
			flusher.Flush()
			continue
		}
		err := flashEvent(flusher, rw, *event)
		if err != nil {
			return err
		}
	}
	return nil
}
```
Image

**Expected behavior**
No panic

**Screenshots**

Image

 - OS: [macOS]
 - Go version: go version go1.24.9 darwin/arm64
 - ADK version: v0.2.0

**Model Information:**
 - openai gpt-5.1(adapted model )

maishivamhoo123:
Fixes #383

Description
This PR addresses a stability issue in RunSSEHandler where the server would panic or enter an infinite loop if the LLM provider returned an error (e.g., invalid API key) during streaming.

The Bug: The error handling logic inside the event loop used continue after encountering a stream error. This caused the loop to attempt reading from a broken or closed stream repeatedly, causing a panic.

The Fix: Replaced continue with return nil in server/adkrest/controllers/runtime.go. This ensures the handler exits the request loop gracefully when a fatal error occurs.

Verification
Before Fix (Panic/Crash): The server crashes with a panic when an invalid API key is used. Image Image

After Fix (Graceful Exit): The server returns the error to the client and remains stable without crashing. Image

### DIFF
--- a/server/adkrest/controllers/runtime.go
+++ b/server/adkrest/controllers/runtime.go
@@ -126,7 +126,7 @@ func (c *RuntimeAPIController) RunSSEHandler(rw http.ResponseWriter, req *http.R
 				return newStatusError(fmt.Errorf("failed to flush: %w", err), http.StatusInternalServerError)
 			}
 
-			continue
+			return nil
 		}
 		err := flashEvent(rc, rw, *event)
 		if err != nil {


wolo-lab:
Thank you for the report and the before-and-after captures. The crash was real and #383 described it accurately.

You asked for this issue and it was assigned to you, you sent a fix nine days later, and you asked for a review that never came. I am sorry — that is worse than leaving a PR unreviewed, and it is on us.

The handler was reworked in #734, which replaced the ad-hoc error text with a structured "event: error" frame written before the loop reaches the event dereference. That closes the crash you hit.

The specific change here, returning instead of continuing on a run error, is not what main does. It keeps streaming and returns only when it cannot write to the client, so one failed event does not tear down the whole stream. That part is deliberate.

We will close #383 with the same explanation.

Apologies that this sat this long with no human reply. We are reviewing the queue far more actively now.

PR #447, opened 2025-12-30
fix: mcptoolset: cached MCP session not validated, causes 'connection closed' errors

### LINKED ISSUE #399: mcptoolset: cached MCP session not validated, causes "connection closed" errors on subsequent requests

majiayu000:
## Summary
This PR fixes #399

## Changes
- `tool/mcptoolset/set.go`
- `tool/mcptoolset/set_test.go`

### DIFF
--- a/tool/mcptoolset/set.go
+++ b/tool/mcptoolset/set.go
@@ -139,7 +139,15 @@ func (s *set) getSession(ctx context.Context) (*mcp.ClientSession, error) {
 	defer s.mu.Unlock()
 
 	if s.session != nil {
-		return s.session, nil
+		// Validate cached session with a Ping health check.
+		// If the connection was closed, clear the stale session and reconnect.
+		if err := s.session.Ping(ctx, nil); err != nil {
+			// Session is stale, close and clear it.
+			_ = s.session.Close() // Best effort close
+			s.session = nil
+		} else {
+			return s.session, nil
+		}
 	}
 
 	session, err := s.client.Connect(ctx, s.transport, nil)
--- a/tool/mcptoolset/set_test.go
+++ b/tool/mcptoolset/set_test.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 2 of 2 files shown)


dpasiukevich:
@majiayu000 wdyt if we'd use https://github.com/google/adk-go/pull/417? 

https://github.com/google/adk-go/pull/447 always makes a Ping request on each List/Call. Ping is a lightweight call, but still it's a network call. And in https://github.com/google/adk-go/pull/417 we'd only reconnect when session is expired.

dpasiukevich:
Let's proceed with https://github.com/google/adk-go/pull/417

PR #451, opened 2025-12-31
feat(openapitoolset): add OpenAPI specification toolset

cpunion:
Base on:
https://github.com/google/adk-go/pull/455 feat(auth): add OAuth2/OIDC authentication package
https://github.com/google/adk-go/pull/457 Feature/auth flow

- Add openapitoolset package for generating tools from OpenAPI specs
- Support JSON and YAML spec formats
- Add RestApiTool with OAuth2, API key, and HTTP auth support
- Add parser for operations, parameters, and request bodies
- Add unit tests for parser, toolset, and rest_api_tool


jjsasha63:
Thank you for this, and I am sorry it took us eight months to reply. That delay is on us.

An OpenAPI toolset is genuinely missing from adk-go and we do want one. This PR cannot be the vehicle, for three reasons.

**The auth half conflicts with a decision we have since made.** This PR is stacked on #455 and #457. We closed #455 because adk-go had by then landed its own `auth` package, built on `golang.org/x/oauth2` rather than on a ported `AuthConfig`/`CredentialManager` layer; you closed #457 yourself. Either way, the foundation this PR sits on is not one we can take now. Main also has `auth.Transport`, an `http.RoundTripper` that resolves a credential per request, which removes the need for the `getAuthHeaders` layer in `rest_api_tool.go` entirely.

**The parser does not resolve `$ref`.** `parser.go:98` skips `$ref` at the path-item level, and parameter and request-body schemas are copied through as raw maps with no `components/schemas` resolution anywhere in the file. A parameter whose schema is a `$ref` reaches `convertSchemaToGenai` with no `type`, so it falls back to `string` at `rest_api_tool.go:92`. Most real-world specs would therefore produce tools with wrong signatures rather than an error, which is the hardest kind of failure to notice. Separately, `parseParameters` still appends the path-level and operation-level lists instead of overriding by name and location, which the automated review flagged as high severity in December and which is still open.

**The branch has moved out from under itself.** Main is 322 commits ahead, with 16 conflicting files including `go.mod` and `go.sum`, and the module path is now `google.golang.org/adk/v2`, so the `google.golang.org/adk/auth` imports no longer resolve. Reconciling that is a rewrite rather than a rebase.

So we are closing this one. We would very much welcome a smaller PR covering only `tool/openapitoolset`, built on an established OpenAPI library and taking credentials through the existing `auth.CredentialProvider` seam. Which library to depend on is a decision we should make before anyone writes the code, so I am opening a tracking issue for it now and will link it here. Please comment there if you would like to pick this up.

jjsasha63:
Tracking issue for the scoped version, as promised above: #1464. It records the constraints (credentials through `auth.CredentialProvider`, separate Go module, `$ref` resolution required) and asks for a decision on which OpenAPI library to depend on.

@cpunion, you have first refusal on it if you want it.

PR #462, opened 2026-01-06
Implement graceful shutdown for web server via context cancellation

imrenagi:
This PR implements graceful shutdown for the HTTP server in webLauncher.Run.

## Issue
Previously, the HTTP server was started using ListenAndServe() directly in the `Run` method. This blocked the execution and did not respect the context.Context passed to the method. As a result, when the context was canceled (e.g., during a system shutdown or signal termination), the server would not stop gracefully, potentially leading to interrupted connections or hanging processes.

## Fix
* Moved `srv.ListenAndServe()` into a separate goroutine.
* Added a select statement to wait for either a server error or context cancellation.
* Upon context cancellation, `srv.Shutdown()` is called with a 5-second timeout to allow active requests to finish before the server stops.
* Correctly handles `http.ErrServerClosed` to avoid returning an error during a planned shutdown.

### DIFF
--- a/cmd/launcher/web/web.go
+++ b/cmd/launcher/web/web.go
@@ -202,7 +202,10 @@ func (w *webLauncher) Run(ctx context.Context, config *launcher.Config) error {
 			return fmt.Errorf("server shutdown failed: %v", err)
 		}
 		return ctx.Err()
-	case err := <-errChan:
+	case err, ok := <-errChan:
+		if !ok {
+			return nil
+		}
 		return fmt.Errorf("server failed: %v", err)
 	}
 }


cmd/launcher/web/web.go
```diff
@@ -184,12 +185,25 @@ func (w *webLauncher) Run(ctx context.Context, config *launcher.Config) error {
 		Handler:      router,
 	}
 
-	err := srv.ListenAndServe()
-	if err != nil {
+	errChan := make(chan error, 1)
+	go func() {
+		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
```
dpasiukevich, line 190: If `srv.ListenAndServe` returns `nil` or http.ErrServerClosed, what will happen? Will the main goroutine block?

imrenagi: Thanks for spotting @dpasiukevich . You are right. I just added close statement after code block on the line 190.

PR #463, opened 2026-01-07
fix: support CitationMetadata

nahapetyan-serob:
Added converters for CitationMetadata in ./server/adka2a/metadata.go and verified in test case.

### DIFF
--- a/server/adka2a/metadata.go
+++ b/server/adka2a/metadata.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/a2aproject/a2a-go/a2a"
 	"github.com/a2aproject/a2a-go/a2asrv"
-	"google.golang.org/genai"
 
 	"google.golang.org/adk/internal/converters"
 	"google.golang.org/adk/session"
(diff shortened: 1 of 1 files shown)


yarolegovich (approved):

yarolegovich (approved):

server/adka2a/metadata.go
```diff
@@ -169,3 +158,30 @@ func processA2AMeta(a2aEvent a2a.Event, event *session.Event) error {
...
+	}
+	if v == nil {
+		return nil
+	}
+
+	result[key] = v
+	return nil
+}
+
+func processMeta[T any](key string, meta map[string]any, target **T) error {
```
dpasiukevich, line 178: Curious why `target **T` has to be a double pointer?

yarolegovich: we're setting a field of type *T on event, so we're passing a pointer to this field to let processMeta update the field. 
an alternative would be
```go
func processMeta[T any](key string, meta map[string]any) (*T, error)
```

but this will require explicit parametrization and will look less nice at the call site
```go
val, err := processMeta[*genai.GenerateContentResponseUsageMetadata](key, meta)
if err != nil {
  return nil
}
event.UsageMetadata = val
```

dpasiukevich: Ah, I see `converters.FromMapStructure[T](m)` returns a pointer.

PR #478, opened 2026-01-13
fix: prevent decode error from being overwritten by Body.Close in decodeRequestBody

cointem:
**Problem:**
In runtime.go, the function decodeRequestBody is declared with named return values.Inside it, the request body is closed using a defer that assigns to the named return variable err.This creates a bug because the deferred closure runs after the function has already decided what to return, but before the return values are actually returned to the caller. If JSON decoding fails, the function intends to return a non-nil error.However, when the deferred Close() runs, it overwrites the return variable err with the result of req.Body.Close().
**fix:**
The fix changes the defer to only propagate a Close() error if no earlier error is being returned:

If decoding (or any earlier step) already set err, we keep that error.
If err is still nil and Close() fails, we return the close error.
**old code:**
image
**new code:**
image

### DIFF
--- a/server/adkrest/controllers/runtime.go
+++ b/server/adkrest/controllers/runtime.go
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


wolo-lab:
Thank you for this. The bug was real and your diagnosis was exactly right — a deferred assignment to a named return does overwrite the decode error, and it did.

It was fixed in #851, which took a slightly different route. Rather than preserving both errors, it removed the defer entirely, on the grounds that net/http closes the request body for handlers and the close was never needed in the first place. decodeRequestBody no longer has named returns.

I am sorry we shipped that four months later without ever replying here. You found it first and deserved to hear that from us. We are much more active on incoming PRs now.

PR #480, opened 2026-01-14
feat: add plugin package

baptmont:
Added plugin package and internal/pluginpackage.

Missing OnUserMessageCallback, OnEventCallback, BeforeRunCallback and AfterRunCallback tests.
Missing closeTimeout and executionOrder logic.
Missing callback context modification.

### DIFF
--- a/plugin/plugin_test.go
+++ b/plugin/plugin_test.go
@@ -39,7 +39,7 @@ func TestNew(t *testing.T) {
 				OnUserMessageCallback: mockOnUserMsg,
 			},
 			validate: func(t *testing.T, p *Plugin) {
-				if p.OnUserMessageCallback == nil {
+				if p.OnUserMessageCallback() == nil {
 					t.Error("OnUserMessageCallback was not mapped correctly")
 				}
 				// Verify Close() is safe even if we didn't provide a CloseFunc


agent/llmagent/llmagent_test.go
```diff
@@ -272,16 +273,154 @@ func TestModelCallbacks(t *testing.T) {
 			},
 			wantErr: http.ErrNoCookie,
 		},
+		{
+			name: "on model error callback is not called",
+			onModelErrorCallback: []llmagent.OnModelErrorCallback{
+				func(ctx agent.CallbackContext, llmRequest *model.LLMRequest, llmError error) (*model.LLMResponse, error) {
+					return &model.LLMResponse{
+						Content: genai.NewContentFromText("hello from on_model_error_callback", genai.RoleModel),
+					}, fmt.Errorf("on_model_error_callback: %w", http.ErrNoCookie)
```
dpasiukevich, line 282: Curious -- this callback returns error `fmt.Errorf("on_model_error_callback: %w", http.ErrNoCookie)`. 
Why the test case expects text and nor this error?

agent/llmagent/llmagent_test.go
```diff
@@ -272,16 +273,154 @@ func TestModelCallbacks(t *testing.T) {
...
+			},
+			llmResponses: []*genai.Content{
+				genai.NewContentFromText("hello from model", genai.RoleModel),
+			},
+			wantTexts: []string{
+				"hello from before_model_callback",
+			},
+		},
+		{
+			name: "before error callback process on model error callback message",
```
dpasiukevich:
```suggestion
			name: "after error callback process on model error callback message",
```

agent/llmagent/llmagent_test.go
```diff
@@ -272,16 +273,154 @@ func TestModelCallbacks(t *testing.T) {
...
+			onModelErrorCallback: []llmagent.OnModelErrorCallback{
+				func(ctx agent.CallbackContext, llmRequest *model.LLMRequest, llmError error) (*model.LLMResponse, error) {
+					return &model.LLMResponse{
+						Content: genai.NewContentFromText("hello from on_model_error_callback", genai.RoleModel),
+					}, nil
+				},
+			},
+			afterModelCallbacks: []llmagent.AfterModelCallback{
+				func(ctx agent.CallbackContext, llmResponse *model.LLMResponse, llmResponseError error) (*model.LLMResponse, error) {
+					return &model.LLMResponse{
```
dpasiukevich, line 388: WDYT if we'd add the check that also verifies that afterModelCallback receives output from onModelCallback

```
if llmResponse != NewContentFromText("hello from on_model_error_callback") {
  t.Errorf("...")
}
```

This could be done in other tests cases. This would be nice to have but not mandatory.

agent/llmagent/llmagent_test.go
```diff
@@ -272,16 +273,154 @@ func TestModelCallbacks(t *testing.T) {
...
+					}, nil
+				},
+			},
+			llmResponses: []*genai.Content{},
+			wantTexts: []string{
+				"hello from after_model_callback",
+			},
+		},
+		{
+			name: "before error callback does not process on model error callback error",
```
dpasiukevich:
```suggestion
			name: "after error callback does not process on model error callback error",
```

internal/llminternal/base_flow_test.go
```diff
@@ -258,16 +262,136 @@ func TestCallTool(t *testing.T) {
...
+			},
+			beforeToolCallbacks: []BeforeToolCallback{
+				func(ctx tool.Context, tool tool.Tool, args map[string]any) (map[string]any, error) {
+					return nil, errors.New("error_from_before")
+				},
+			},
+			onToolErrorCallbacks: []OnToolErrorCallback{
+				func(ctx tool.Context, tool tool.Tool, args map[string]any, err error) (map[string]any, error) {
+					if err == nil || err.Error() != "error_from_before" {
+						return nil, errors.New("unexpected error in on tool error callback")
```
dpasiukevich, line 283: nit: maybe let's use t.Error() to mark the test case as failed?

internal/llminternal/base_flow_test.go
```diff
@@ -258,16 +262,136 @@ func TestCallTool(t *testing.T) {
...
+					if err == nil || err.Error() != "error_from_on_tool_error" {
+						return nil, errors.New("unexpected error in after callback")
+					}
+					return nil, errors.New("error_from_after_tool")
+				},
+			},
+			want: map[string]any{"error": "error_from_after_tool"},
+		},
+		{
+			name: "before callback error passed to on tool error callback and passed to after tool called and handled",
```
dpasiukevich, line 357: WDYT, maybe this error case is excessive and not needed?

`before callback error passed to on tool error callback and passed to after tool called` already does this check.

This test case is only different that it verifies that after_tool_callback consumes the error and returns ok result, this ok result is returned. And it's already tested in "after callback handles error" test case.

internal/plugininternal/plugin_manager.go
```diff
@@ -0,0 +1,306 @@
...
+		}
+	}
+
+	return pm, nil
+}
+
+// RegisterPlugin adds a new plugin to the manager.
+func (pm *PluginManager) RegisterPlugin(plugin plugin.Plugin) error {
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
```
dpasiukevich: Do we need mutex? Plugins are registered once in the very beginning when plugin manager is created.

agent/context.go
```diff
@@ -95,6 +96,8 @@ type InvocationContext interface {
 	EndInvocation()
 	// Ended returns whether the invocation has ended.
 	Ended() bool
+
+	PluginManager() PluginManager
```
dpasiukevich: Could we try this approach: instead of exposing PluginManager() in the public API, but rather to pass it via the context:

branch: https://github.com/dpasiukevich/adk-go/tree/plugins
commit: https://github.com/google/adk-go/commit/830feb2ee10a588fbcdb96e3e1db53e42033ad36

agent/llmagent/llmagent_test.go
baptmont: on_model_error_callback is not called because there is no error generated by the runner

internal/llminternal/base_flow_test.go
baptmont: I would leave this just in case

plugin/plugin.go
```diff
@@ -0,0 +1,112 @@
...
+		p.closeFunc = func() error {
+			return nil
+		}
+	}
+
+	return p, nil
+}
+
+type Plugin struct {
+	Name                  string
```
dpasiukevich: WDYT if we can make these fields internal? This way we'd have sort of a defensive approach: plugin is created with `plugin.New(plugin.Config{...})` and callbacks could be exposed as method `(p *Plugin) BeforeAgentCallback(...) { if p.beforeAgentCallback != nil { return p.beforeAgentCallback}; return nil }`.
This approach is a bit more verbose, but it's easy to extend in future (e.g. if there's any internal state would be needed or having validation in New).

Another approach just to make `type Plugin struct { ... } with exported fields. And remove New and Config funcs.

I'd suggest 1st one. WDYT?

As currently in the PR is a mix: there are both public fields in the Plugin struct and New(Config) funcs.

agent/llmagent/llmagent_test.go
```diff
@@ -272,16 +273,154 @@ func TestModelCallbacks(t *testing.T) {
...
+					return &model.LLMResponse{
+						Content: genai.NewContentFromText("hello from on_model_error_callback", genai.RoleModel),
+					}, fmt.Errorf("error from on_model_error_callback: %w", http.ErrNoCookie)
+				},
+			},
+			llmResponses: []*genai.Content{},
+			wantErr:      http.ErrNoCookie,
+		},
+		{
+			name: "on model error callback does not process before model callback message",
```
dpasiukevich:
```suggestion
			name: "on model error callback does not process before model callback error",
```

agent/llmagent/llmagent_test.go
```diff
@@ -272,16 +273,154 @@ func TestModelCallbacks(t *testing.T) {
...
+					}, fmt.Errorf("error from on_model_error_callback: %w", http.ErrHijacked)
+				},
+			},
+			llmResponses: []*genai.Content{
+				genai.NewContentFromText("hello from model", genai.RoleModel),
+			},
+			wantErr: http.ErrNoCookie,
+		},
+		{
+			name: "on model error callback does not process before model callback error",
```
dpasiukevich:
```suggestion
			name: "on model error callback does not process before model callback message",
```

internal/llminternal/base_flow.go
```diff
@@ -431,19 +535,65 @@ func (f *Flow) handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[st
...
 
-func (f *Flow) callTool(tool toolinternal.FunctionTool, fArgs map[string]any, toolCtx tool.Context) map[string]any {
-	result, err := f.invokeBeforeToolCallbacks(tool, fArgs, toolCtx)
-	if result == nil && err == nil {
-		result, err = tool.Run(toolCtx, fArgs)
+func (f *Flow) runOnToolErrorCallbacks(toolCtx tool.Context, tool tool.Tool, fArgs map[string]any, err error) (map[string]any, error) {
+	pluginManager := pluginManagerFromContext(toolCtx)
+	if pluginManager != nil {
+		result, err := pluginManager.RunOnToolErrorCallback(toolCtx, tool, fArgs, err)
+		if result != nil && err != nil {
```
dpasiukevich: Should it be ||?

```suggestion
		if result != nil || err != nil {
```

internal/plugininternal/plugin_manager.go
```diff
@@ -0,0 +1,278 @@
...
+			plugin.AfterRunCallback(cctx)
+		}
+	}
+}
+
+// RunOnEventCallback runs the OnEventCallback for all plugins.
+func (pm *PluginManager) RunOnEventCallback(cctx agent.InvocationContext, event *session.Event) (*session.Event, error) {
+	for _, plugin := range pm.plugins {
+		if plugin.OnEventCallback != nil {
+			// TODO: Replace 'any' with the actual Event type
```
dpasiukevich: Curious about this TODO. Is it outdated?

plugin/plugin_manager_test.go
```diff
@@ -0,0 +1,878 @@
...
+					got = part.FunctionResponse.Response
+				}
+			}
+
+			if diff := cmp.Diff(tc.want, got); diff != "" {
+				t.Errorf("callTool() mismatch (-want +got):\n%s", diff)
+			}
+
+			if onToolErrorCallbacksCalled == false && tc.dontRunOnError == false {
+				t.Error("on model error should be called")
```
dpasiukevich:
```suggestion
				t.Error("on tool error should be called")
```

plugin/plugin_manager_test.go
```diff
@@ -0,0 +1,878 @@
...
+			}
+			if beforeToolCallbacksCalled == false && tc.dontRunBefore == false {
+				t.Error("before model should be called")
```
dpasiukevich:
```suggestion
				t.Error("before tool should be called")
```

plugin/plugin_manager_test.go
```diff
@@ -0,0 +1,878 @@
...
+			}
+			if afterToolCallbacksCalled == false && tc.dontRunAfter == false {
+				t.Error("after model should be called")
```
dpasiukevich:
```suggestion
				t.Error("after tool should be called")
```

plugin/plugin_manager_test.go
```diff
@@ -0,0 +1,878 @@
...
+			beforeToolCallbacks: []llmagent.BeforeToolCallback{
+				func(ctx tool.Context, tool tool.Tool, args map[string]any) (map[string]any, error) {
+					return map[string]any{"result": "intercepted"}, nil
+				},
+				func(ctx tool.Context, tool tool.Tool, args map[string]any) (map[string]any, error) {
+					return map[string]any{"result": "2nd callback should not be called"}, nil
+				},
+			},
+			dontRunOnError: true,
+			dontRunBefore:  true,
```
dpasiukevich:
```suggestion
			dontRunBeforeCanonicalCallback:  true,
```

nit: maybe let's add Canonical here, just to distinguish this

plugin/plugin.go
baptmont: 1st one sounds good

PR #484, opened 2026-01-14
refactor(mcptoolset): use MCPClient w/ strict ErrConnectionClosed check

majiayu000:
Supersedes #417.

### Changes
1.  **Strict Error Handling**: Reconnection logic now strictly checks for `mcp.ErrConnectionClosed` as suggested by @dpasiukevich. This prevents infinite loops on generic errors.
2.  **Architecture**: Refactored `mcptoolset` to use a `MCPClient` interface and `reconnectingClient` struct, decoupling session management from the Toolset.
3.  **Verification**: Verified with tests ensuring `mcp.ErrConnectionClosed` triggers reconnection correctly.

### DIFF
--- a/tool/mcptoolset/set_test.go
+++ b/tool/mcptoolset/set_test.go
@@ -16,7 +16,9 @@ package mcptoolset_test
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"iter"
 	"log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


majiayu000:
Thanks for the feedback! I've updated the test to use `errors.Is(err, io.ErrClosedPipe)` and `errors.Is(err, io.EOF)` instead of fragile string matching. Also added the necessary imports.

majiayu000:
Thanks for the feedback! I've updated the test to use `errors.Is(err, io.ErrClosedPipe)` and `errors.Is(err, io.EOF)` instead of fragile string matching. Also added the necessary imports.

dpasiukevich:
Let's proceed in https://github.com/google/adk-go/pull/417

PR #488, opened 2026-01-15
Make the type comparison case insensitive

caglar10ur:
Otherwise schema validation that uses lowercase values of enums such as "object" or "string" fails.

### DIFF
--- a/internal/utils/schema_utils.go
+++ b/internal/utils/schema_utils.go
@@ -35,11 +35,7 @@ func matchType(value any, schema *genai.Schema, isInput bool) (bool, error) {
 	}
 
 	// Convert type to upper case to match the type in the schema.
-	localSchema := *schema
-	localSchema.Type = genai.Type(strings.ToUpper(string(schema.Type)))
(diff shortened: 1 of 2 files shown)


internal/utils/schema_utils.go
```diff
@@ -33,6 +34,11 @@ func matchType(value any, schema *genai.Schema, isInput bool) (bool, error) {
 		return false, nil
 	}
 
+	// Convert type to upper case to match the type in the schema.
+	localSchema := *schema
+	localSchema.Type = genai.Type(strings.ToUpper(string(schema.Type)))
+	schema = &localSchema
+
 	switch schema.Type {
```
dpasiukevich:
```suggestion
switch genai.Type(strings.ToUpper(string(schema.Type))) {
```

WDYT if instead of making a shallow copy of the schema object, we'd just normalize (uppercasing) the type field before the check?

caglar10ur: Makes sense - just did that.

PR #489, opened 2026-01-16
feat: add FilterToolset helper

dpasiukevich:
This provides same functionality as in the BaseToolset class in adk-python: https://github.com/google/adk-python/blob/d0aef8a4fadc1c435ec75afecebd84f9ab0ffd7a/src/google/adk/tools/base_toolset.py#L71

### DIFF
--- a/tool/tool.go
+++ b/tool/tool.go
@@ -78,3 +78,42 @@ func StringPredicate(allowedTools []string) Predicate {
 		return m[tool.Name()]
 	}
 }
+
+// FilterToolset returns a Toolset that filters the tools in the given Toolset
+// using the given predicate.
+func FilterToolset(toolset Toolset, predicate Predicate) Toolset {
+	if toolset == nil {
+		panic("toolset must not be nil")
+	}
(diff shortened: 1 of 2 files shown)


kdroste-google (approved):
LGTM

tool/tool.go
```diff
@@ -78,3 +78,42 @@ func StringPredicate(allowedTools []string) Predicate {
...
+	if err != nil {
+		return nil, err
+	}
+	var filtered []Tool
+	for _, tool := range tools {
+		if f.predicate(ctx, tool) {
+			filtered = append(filtered, tool)
+		}
+	}
+	return filtered, nil
```
baptmont, line 118: Could filtered be cached? Applying the predicate every time seems unnecessary if Toolset and Predicate are immutable.

dpasiukevich: But `.Tools(ctx)` result is not immutable. MCPToolset may return different tools on each call. Other Toolsets as well.

If predicate call is heavy, it's up to a user to cache it IMO.

Issue #498, opened 2026-01-21
[Enhancement/ BUG?] `BeforeToolCallbacks` implementation is disturbing


nag763:
**Is your feature request related to a problem? Please describe.**
Related to a disturbing implementation of the `BeforeToolCallbacks` agent LLM hooks.

What is described below mainly refers to changes implemented in adb487b05942ac0874ae506477f44f59dd2bec00

**Describe the solution you'd like**
So far the implementation of the hook is the followings :
1. If no hook exists -> Nothing performed -> OK
2. If one hooks exists and returns a map, which you would assume is keeping the flow of hooks running and replacing the input map in the agent context => actually no tool is called, and other hooks are skipped per the implementation

```go

//hook
func InjectMetadata(ctx tool.Context, tool tool.Tool, args map[string]any) (map[string]any, error) {
	args["metadata"] = domain.Metadata{
		UserId: ctx.UserID(),
	}
        // nag763 : we would assume args need to be returned to indicate that the flow went well
	return args, nil
}


// hook flow
func (f *Flow) invokeBeforeToolCallbacks(tool toolinternal.FunctionTool, fArgs map[string]any, toolCtx tool.Context) (map[string]any, error) {
	for _, callback := range f.BeforeToolCallbacks {
		result, err := callback(toolCtx, tool, fArgs)
		if err != nil {
			return nil, err
		}
		// When a list of callbacks is provided, the callbacks will be called in the
		// order they are listed while a callback returns nil.
		if result != nil {
                        // nag763 :  tool returns a value here, skipping other calls
			return result, nil
		}
	}
	return nil, nil
}

//tool flow
func (f *Flow) callTool(tool toolinternal.FunctionTool, fArgs map[string]any, toolCtx tool.Context) map[string]any {
	result, err := f.invokeBeforeToolCallbacks(tool, fArgs, toolCtx)
        // nag763 : this is skipped as result isn't nil
	if result == nil && err == nil {
		result, err = tool.Run(toolCtx, fArgs)
	}
	result, err = f.invokeAfterToolCallbacks(tool, fArgs, toolCtx, result, err)
	if err != nil {
		return map[string]any{"error": err.Error()}
	}
	return result
}
```

This part is pretty confusing, I would first like to indicate I am not a golang expert, but I would have naively assumed :
* The args returned by my hook are the ones that will be used for calling the tool, not the ones passed as arguments
* This shouldn't prevent the tool from running (is it a bug ?)

**Describe alternatives you've considered**
[...]

PR #500, opened 2026-01-22
update docs for llmagent

indurireddy-TF:
Why this change:
Improves discoverability of existing behavior
Prevents common misuse of BeforeToolCallback and AfterToolCallBack
Aligns interface documentation with existing configuration docs
Reduces developer confusion without introducing breaking changes

Fixes: https://github.com/google/adk-go/issues/498


agent/llmagent/llmagent.go
dpasiukevich, line 301: Could I ask to update the comment for AfterToolCallback in a similar fashion if that's ok?

To say that if AfterToolCallback returns non-nil result or error, then "execution of remaining callbacks stops" and this result is returned.

Thanks!

PR #512, opened 2026-01-27
Add custom part converters to ADK Executor.

zbirkenbuel:
This change essentially mirrors the functionality available in the Python A2A implementation at https://github.com/google/adk-python/blob/main/src/google/adk/a2a/executor/a2a_agent_executor.py#L62-L67.

The use case here is to allow conversion of specific A2A Data parts into something slightly structured (or annotated in some way) for ADK consumption. Currently data parts which are not Function or ExecutableCode related get converted to Text parts (https://github.com/google/adk-go/blob/main/server/adka2a/parts.go#L253). Conceptually these could become InlineData parts (which do tend to overlap with the model of File parts) or extra processing could be applied to make them more consumable by ADK agents. Similarly semi-structured output from ADK agents are returned as TextParts, this just gives us the responsible to marshal these as typed data parts based on whatever signifies we include.

### DIFF
--- a/server/adka2a/executor.go
+++ b/server/adka2a/executor.go
@@ -38,15 +38,15 @@ type AfterEventCallback func(ctx ExecutorContext, event *session.Event, processe
 // AfterExecuteCallback is the callback which will be called after an execution resolved into a completed or failed task.
 type AfterExecuteCallback func(ctx ExecutorContext, finalEvent *a2a.TaskStatusUpdateEvent, err error) error
 
-// A2AToGenAIPartConverter is a custom converter for converting A2A parts to GenAI parts.
+// A2APartConverter is a custom converter for converting A2A parts to GenAI parts.
 // Implementations should generally remember to leverage adka2a.ToGenAiPart for default conversions
 // nil returns are considered intentionally dropped parts.
-type A2AToGenAIPartConverter func(ctx context.Context, a2aEvent a2a.Event, part a2a.Part) (*genai.Part, error)
+type A2APartConverter func(ctx context.Context, a2aEvent a2a.Event, part a2a.Part) (*genai.Part, error)
 
-// GenAIToA2APartConverter is a custom converter for converting GenAI parts to A2A parts.
+// GenAIPartConverter is a custom converter for converting GenAI parts to A2A parts.
 // Implementations should generally remember to leverage adka2a.ToA2APart for default conversions
 // nil returns are considered intentionally dropped parts.
-type GenAIToA2APartConverter func(ctx context.Context, adkEvent *session.Event, part *genai.Part) (a2a.Part, error)
+type GenAIPartConverter func(ctx context.Context, adkEvent *session.Event, part *genai.Part) (a2a.Part, error)
 
 // ExecutorConfig allows to configure Executor.
 type ExecutorConfig struct {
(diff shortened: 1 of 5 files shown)


server/adka2a/executor.go
```diff
@@ -36,6 +37,14 @@ type AfterEventCallback func(ctx ExecutorContext, event *session.Event, processe
 // AfterExecuteCallback is the callback which will be called after an execution resolved into a completed or failed task.
 type AfterExecuteCallback func(ctx ExecutorContext, finalEvent *a2a.TaskStatusUpdateEvent, err error) error
 
+// A2AToGenAIPartConverter is a custom converter for converting A2A parts to GenAI parts.
+// nil returns are ignored and the default conversion is used.
```
yarolegovich: > nil returns are ignored and the default conversion is used

Not sure about this part (pun intended).
It would be nice to allow fully ignoring parts on this level (and if all parts are ignored event is not emitted).

I think the cleanest approach would be making nil serve as "skip part" marker.
Default conversion should remain accessible using `adka2a.To(A2A|Genai)Part`:
```go
converter := func(part a2a.Part) (*genai.Part, error) {
  switch {
    case shouldSkip(part):
      return nil, nil
    case shouldCustomConvert(part):
      return customConvert(part)
    default:
      return adka2a.ToGenaiPart(part)
  } 
}
```

Another alternative is a special `adka2a.ErrSkipPart`, but I think it looks dirtier / more obscure.
```go
converter := func(part a2a.Part) (*genai.Part, error) {
  switch {
    case shouldSkip(part):
      return nil, adka2a.ErrSkipPart
    case shouldCustomConvert(part):
      return customConvert(part)
    default:
      return nil, nil
  } 
}
```

server/adka2a/executor.go
```diff
@@ -36,6 +37,14 @@ type AfterEventCallback func(ctx ExecutorContext, event *session.Event, processe
...
+type A2AToGenAIPartConverter func(part a2a.Part) (*genai.Part, error)
+
+// GenAIToA2APartConverter is a custom converter for converting GenAI parts to A2A parts.
+// nil returns are ignored and the default conversion is used.
+type GenAIToA2APartConverter func(part *genai.Part) (a2a.Part, error)
```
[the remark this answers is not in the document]
yarolegovich: I agree with this comment.

Context can be useful if for example experiment values are propagated to a converter.

The whole *session.Event is also a good suggestion as well. The current implementation doesn't handle metadata well (especially citations) because all parts are merged into a single A2A artifact. I don't yet have a good final solution in mind, but I think the solution will involve moving some metadata to part level.  Some implementations might want to add branch / author. 

In case of `A2AToGenAIPartConverter` it should probably be an a2a.Event, not *a2a.Message. We should be able to reuse this converter in remoteagent, so:
```go
type A2AToGenAIPartConverter func(context.Context, a2a.Event, a2a.Part) (*genai.Part, error)

type GenAIToA2APartConverter func(context.Context, *session.Event, *genai.Part) (a2a.Part, error)
```

server/adka2a/executor.go
zbirkenbuel: Thanks for the review @yarolegovich, I'll take a pass on those comments.  I was hesitant to **require** the converter to always remember to call the baseline ToGenAIPart (as Python effectively does) but it really isn't that big of a deal and the comment can make it clear that doing so is recommended.

zbirkenbuel: Updated.  Sorry for the force push, I'm very out of practice with github and the ability to squash commits for PRs across forks of repos

yarolegovich: > Updated. Sorry for the force push, I'm very out of practice with github and the ability to squash commits for PRs across forks of repos

Np, we'll be squashing it on merge anyway, so what happens on a branch stays on the branch.
Pushed a minor [refactoring](https://github.com/google/adk-go/pull/512/changes/72e944e87da92944920658b3369b729ca4542091), waiting for the second approval now.

PR #513, opened 2026-01-27
Agent-execution-fail-when-instructions-prompt-brackets

indurireddy-TF:
Added DisableInstructionTemplating bool to llmagent.Config.
Added DisableInstructionTemplating bool to llminternal.State.
Mapped the config to the state in the New() function.
Wrapped InjectSessionState with the if !agentState.DisableInstructionTemplating check.


dpasiukevich:
If you don't mind, I've added a PR with docs explaining the current logic for Instruction and InstructionProvider which achieves this. So that we don't need to extend the LLMAgent API.

PR #516, opened 2026-01-27
Add UserAccessValidator to enforce per-user access in REST API 

### LINKED ISSUE #507: user_id parameter validation in session endpoints
**Is your feature request related to a problem? Please describe.**

When deploying an agent using REST api there are seemingly no hooks to perform `user_id` parameter validation in `/apps/{app_name}/users/{user_id}/sessions/*` endpoints. For example I can have one userId in my JWT token, but pass a completely different `user_id` in request url. 

**Describe the solution you'd like**

Expose some sort of user validation features for endpoints that deal with `user_id` endpoints.

**Describe alternatives you've considered**

Adding a custom HTTP middleware to parse url and compare the results with the one from JWT token, or implement a custom sessions service that does the validation.

Kenasvarghese:
This PR introduces a UserAccessValidator interface to the ADK REST API, enabling per-user access control for sessions, runtime, and artifacts endpoints.

**Changes include:**
- Added `UserAccessValidator` interface and `UserAccessValidatorFunc` adapter.
- Updated API controllers (Sessions, Runtime, Artifacts) to use the validator before performing operations.
- Updated handler and launcher to pass the validator.
- Added test cases for unauthorized access scenarios.

By default, existing behavior is unchanged if no validator is provided.

This allows applications to enforce custom authorization rules for API endpoints.

**Related Issues:**

- Fixes https://github.com/google/adk-go/issues/507


server/adkrest/validation/user_access.go
```diff
@@ -0,0 +1,33 @@
...
+package validation
+
+import (
+	"context"
+	"net/http"
+)
+
+// UserAccessValidator is an interface for user access validation
+type UserAccessValidator interface {
+	ValidateUserAccess(ctx context.Context, appName, userID string, r *http.Request) error
```
vadimi-ep: I wonder if `ctx context.Context` and `r *http.Request` are a bit redundant here since we can get context from `r.Context()`?

server/adkrest/controllers/sessions.go
```diff
@@ -162,3 +177,14 @@ func (c *SessionsAPIController) ListSessionsHandler(rw http.ResponseWriter, req
 	}
 	EncodeJSONResponse(sessions, http.StatusOK, rw)
 }
+
+// checkUserAccess checks if the user has access.
+func (c *SessionsAPIController) checkUserAccess(rw http.ResponseWriter, req *http.Request, appName string, userID string) bool {
+	if c.userAccessValidator != nil {
+		if err := c.userAccessValidator.ValidateUserAccess(req.Context(), appName, userID, req); err != nil {
+			http.Error(rw, fmt.Errorf("user access validation failed: %v", err).Error(), http.StatusForbidden)
```
vadimi-ep, line 189: I wonder if some users might want to customize response handling behavior - for example use a different status code. What do you think?

Kenasvarghese: To support customizable response handling, I’ve introduced a typed ValidationError that allows the validator to control the HTTP status code. Controllers now inspect this error type and map the response accordingly, while preserving backward compatibility for validators that return a plain error.

vadimi-ep (approved):

Kenasvarghese:
Please Review @yarolegovich @dpasiukevich

wolo-lab:
Thank you for the valid contribution. Our team starts working on a related mechanizm which will be aligned cross all adks.

PR #521, opened 2026-01-29
feat: provide access to session events from executor callbacks

yarolegovich:
Added `Events()` to executor context for providing read-only access to session events from `adka2a.Executor` callbacks.
This is especially useful in `AfterExecuteCallback` where we can modify the final event. A use case might be iterating over events to collect metadata about enabled a2a extension for setting them on the task level as opposed to the on artifact level.

### DIFF
--- a/server/adka2a/executor_test.go
+++ b/server/adka2a/executor_test.go
@@ -410,10 +410,6 @@ func TestExecutor_Callbacks(t *testing.T) {
 				{LLMResponse: modelResponseFromParts(genai.NewPartFromText("Hello"))},
 				{LLMResponse: modelResponseFromParts(genai.NewPartFromText(", world!"))},
 			},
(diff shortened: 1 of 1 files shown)


server/adka2a/executor_test.go
```diff
@@ -404,6 +404,31 @@ func TestExecutor_Callbacks(t *testing.T) {
 				newFinalStatusUpdate(task, a2a.TaskStateCompleted, nil),
 			},
 		},
+		{
+			name: "can access session events",
+			events: []*session.Event{
+				{LLMResponse: modelResponseFromParts(genai.NewPartFromText("Hello"))},
+				{LLMResponse: modelResponseFromParts(genai.NewPartFromText(", world!"))},
+			},
+			afterEvent: func(ctx ExecutorContext, event *session.Event, processed *a2a.TaskArtifactUpdateEvent) error {
```
baptmont: remove afterEvent

PR #523, opened 2026-01-30
feat: add human in the loop confirmation for mcp toolset

baptmont:
MCP toolset can be configured to require confirmation. This configuration is applied for every tool of the set.

python ref:
https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_toolset.py
https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_tool.py#L48

### DIFF
--- a/tool/functiontool/function_test.go
+++ b/tool/functiontool/function_test.go
@@ -30,12 +30,14 @@ import (
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/agent/llmagent"
+	icontext "google.golang.org/adk/internal/context"
 	"google.golang.org/adk/internal/httprr"
 	"google.golang.org/adk/internal/testutil"
 	"google.golang.org/adk/internal/toolinternal"
 	"google.golang.org/adk/internal/typeutil"
 	"google.golang.org/adk/model"
 	"google.golang.org/adk/model/gemini"
+	"google.golang.org/adk/session"
 	"google.golang.org/adk/tool"
 	"google.golang.org/adk/tool/functiontool"
 	"google.golang.org/adk/tool/toolconfirmation"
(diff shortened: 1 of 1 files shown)


tool/mcptoolset/set.go
```diff
@@ -106,3 +127,5 @@ func (s *set) Tools(ctx agent.ReadonlyContext) ([]tool.Tool, error) {
 
 	return adkTools, nil
 }
+
+type ConfirmationProvider func(string, any) bool
```
dpasiukevich, line 137: could you add godoc?

tool/mcptoolset/tool.go
```diff
@@ -85,6 +91,25 @@ func (t *mcpTool) Declaration() *genai.FunctionDeclaration {
 }
 
 func (t *mcpTool) Run(ctx tool.Context, args any) (map[string]any, error) {
+	requireConfirmation := t.requireConfirmation
+	if t.requireConfirmationProvider != nil {
+		requireConfirmation = t.requireConfirmationProvider(t.Name(), args)
```
dpasiukevich: can we avoid calling requireConfirmationProvider twice? Right now we run it 2 times:
1. on the first call -- when there's no confirmation
2. when there's a confirmation

tool/mcptoolset/set_test.go
```diff
@@ -418,3 +359,434 @@ func (rt *reconnectableTransport) Connect(ctx context.Context) (mcp.Connection,
...
+}
+
+type TestResult struct {
+	Value int
+}
+
+func TestNewToolSet_RequireConfirmationProvider_Validation(t *testing.T) {
+	tests := []struct {
+		name     string
+		provider func(name string, args any) bool // The RequireConfirmationProvider value to test
```
dpasiukevich:
```suggestion
		provider RequireConfirmationProvider // The value to test
```

yarolegovich (approved):

PR #524, opened 2026-01-30
feat: add OpenTelemetry configuration and initialization - TracerProvider

pawel-maciejczek:
# Summary

This PR adds the telemetry module initialization based on OpenTelemetry.

For now only `TracerProvider` is covered, but `LoggerProvider` and `MeterProvider` will follow soon.

Details:
* telemetry module is automatically configured and installed in `web` and `console` launchers
* `web` and `console` launchers now support `otel_to_cloud` enabling exporting telemetry signals to GCP
* telemetry can be customized using newly introduced `telemetry.Option`, e.g. to register custom processors/exporters or pass preconfigured `TracerProvider` - see example for details

### DIFF
--- a/telemetry/setup_otel.go
+++ b/telemetry/setup_otel.go
@@ -91,22 +91,7 @@ func newInternal(cfg *config) (*Providers, error) {
 // 3. GOOGLE_CLOUD_PROJECT environment variable.
 // Returns the quota project or error if the quota project cannot be determined.
 func resolveGcpQuotaProject(cfg *config) (string, error) {
-	if cfg.gcpQuotaProject != "" {
-		return cfg.gcpQuotaProject, nil
-	}
-	if !cfg.oTelToCloud {
-		return "", nil
-	}
-	if cfg.googleCredentials != nil && cfg.googleCredentials.ProjectID != "" {
-		return cfg.googleCredentials.ProjectID, nil
-	}
-	// The quota project wasn't set in credentials during testing, even when it's set in ADC JSON file.
-	// Using fallback to env variable to resolve the quota project as a workaround.
-	project, ok := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
-	if !ok && cfg.oTelToCloud {
-		return "", fmt.Errorf("telemetry.googleapis.com requires setting the quota project. Refer to telemetry.config for the available options to set the quota project")
-	}
-	return project, nil
+	return resolveProject(cfg.gcpQuotaProject, cfg.googleCredentials, cfg.oTelToCloud, "quota")
 }
 
 // resolveGcpResourceProject determines the resource project for telemetry export in the following order:
@@ -115,19 +100,23 @@ func resolveGcpQuotaProject(cfg *config) (string, error) {
 // 3. GOOGLE_CLOUD_PROJECT environment variable.
 // Returns the resource project or error if the resource project cannot be determined.
 func resolveGcpResourceProject(cfg *config) (string, error) {
-	if cfg.gcpResourceProject != "" {
-		return cfg.gcpResourceProject, nil
+	return resolveProject(cfg.gcpResourceProject, cfg.googleCredentials, cfg.oTelToCloud, "resource")
+}
+
+func resolveProject(configuredProject string, creds *google.Credentials, requireProject bool, projectType string) (string, error) {
+	if configuredProject != "" {
+		return configuredProject, nil
 	}
-	if cfg.googleCredentials != nil && cfg.googleCredentials.ProjectID != "" {
-		return cfg.googleCredentials.ProjectID, nil
+	if creds != nil && creds.ProjectID != "" {
+		return creds.ProjectID, nil
 	}
-	// The resource project wasn't set in credentials during testing, even when it's set in ADC JSON file.
-	// Using fallback to env variable to resolve the resource project as a workaround.
-	projectID, ok := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
-	if !ok && cfg.oTelToCloud {
-		return "", fmt.Errorf("telemetry.googleapis.com requires setting the resource project. Refer to telemetry.config for the available options to set the resource project")
+	// The project was always empty during testing, even though it was set in ADC JSON file.
+	// Using fallback to env variable to resolve the project as a workaround.
+	project, ok := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
+	if !ok && requireProject {
+		return "", fmt.Errorf("telemetry.googleapis.com requires setting the %s project. Refer to telemetry.config for the available options to set the %s project", projectType, projectType)
 	}
-	return projectID, nil
+	return project, nil
 }
 
 // resolveResource creates a new resource with attributes specified in the following order (later attributes override earlier ones):


cmd/launcher/console/console.go
```diff
@@ -52,12 +59,30 @@ func NewLauncher() launcher.SubLauncher {
 	fs := flag.NewFlagSet("console", flag.ContinueOnError)
 	fs.StringVar(&config.streamingModeString, "streaming_mode", string(agent.StreamingModeSSE),
 		fmt.Sprintf("defines streaming mode (%s|%s)", agent.StreamingModeNone, agent.StreamingModeSSE))
-
+	fs.BoolVar(&config.otelToCloud, "otel_to_cloud", false, "Enables/disables OpenTelemetry export to cloud")
```
dpasiukevich:
```suggestion
	fs.BoolVar(&config.otelToGCP, "otel_to_gcp", false, "Enables/disables OpenTelemetry export to GCP telemetry.googleapis.com
	")
```

nit: I'd suggest to rename cloud -> gcp because this flag enables the export to GCP specifically.

dpasiukevich: nit: let's expand the flag description to explain for users what setup they need to be able to enable exportToGCP without looking in the source code.

e.g. we can mention how project_id can be defined and how credentials are fetched.

As alternative, the flag description could link to some doc page.

cmd/launcher/console/console.go
```diff
@@ -52,12 +56,25 @@ func NewLauncher() launcher.SubLauncher {
...
+	if err != nil {
+		return fmt.Errorf("telemetry initialization failed: %v", err)
+	}
+	defer func() {
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), l.config.shutdownTimeout)
+		defer cancel()
+		if err := telemetry.Shutdown(shutdownCtx); err != nil {
+			log.Printf("telemetry shutdown failed: %v", err)
+		}
+	}()
```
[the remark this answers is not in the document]
pawel-maciejczek: Adressed in https://github.com/google/adk-go/pull/533/changes

cmd/launcher/telemetry/telemetry.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"context"
+
+	"google.golang.org/adk/cmd/launcher"
+	"google.golang.org/adk/telemetry"
+)
+
+// InitTelemetry initializes telemetry for the launcher.
+func InitTelemetry(ctx context.Context, config *launcher.Config, otelToCloud bool) (telemetry.Service, error) {
```
dpasiukevich: nit: Would it make sense to put this package to cmd/launcher/internal/telemetry? So that this func could only be used within adk-go. Or is it intended to be a public API?

telemetry/setup_otel.go
```diff
@@ -0,0 +1,194 @@
...
+	for _, p := range cfg.spanProcessors {
+		opts = append(opts, sdktrace.WithSpanProcessor(p))
+	}
+	tp := sdktrace.NewTracerProvider(opts...)
+
+	return tp, nil
+}
+
+func newGcpSpanExporter(ctx context.Context, credentials *google.Credentials) (sdktrace.SpanExporter, error) {
+	// The quota poject is not present in credentials, even when it's set in ADC JSON file.
```
dpasiukevich:
```suggestion
	// The quota project is not present in credentials, even when it's set in ADC JSON file.
```

telemetry/setup_otel.go
```diff
@@ -0,0 +1,194 @@
...
+	// Use env variable to get the quota project as a workaround.
+	// Note - the quota project can be different that the OTel resource project, so we don't use the value from the config here.
+	projectID, ok := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
```
dpasiukevich: nit: would it make sense to allow also to pass project_id as a config parameter? e.g. in some setups people won't be able to use env vars or would like to get the value from different sources like cli flags.

telemetry/telemetry.go
```diff
@@ -12,21 +12,77 @@
...
+// Telemetry providers have to be registered in otel global providers either manually or via [SetGlobalOtelProviders].
+//
+// # Usage
+//
+//	 func main() {
+//			telemetryService, err := telemetry.New(ctx,
+//				telemetry.WithOtelToCloud(true),
+//				telemetry.WithResource(resource.New(
+//					ctx,
+//					semconv.ServiceNameKey.String("my-service"),
```
dpasiukevich, line 69: Would it make sense to make a default service_name_key (e.g. it will be the agent name or even "adk-go" and version taken from https://github.com/google/adk-go/blob/main/internal/version/version.go

Asking from usability pov if this makes sense. Then telemetry service can be created as:
```
telemetryService, err := telemetry.New(ctx, telemetry.WithOtelToCloud(true))
```

with all meaningul values already.

server/adkrest/handler.go
```diff
@@ -22,16 +22,19 @@ import (
...
 	"google.golang.org/adk/server/adkrest/internal/routers"
 	"google.golang.org/adk/server/adkrest/internal/services"
+	"google.golang.org/adk/telemetry"
 )
 
 // NewHandler creates and returns an http.Handler for the ADK REST API.
 func NewHandler(config *launcher.Config, sseWriteTimeout time.Duration) http.Handler {
 	adkExporter := services.NewAPIServerSpanExporter()
-	telemetry.AddSpanProcessor(sdktrace.NewSimpleSpanProcessor(adkExporter))
+	processor := sdktrace.NewSimpleSpanProcessor(adkExporter)
```
dpasiukevich, line 34: How would it work with adka2a? For example if user creates adka2a.NewExecutor without the launchers, what's needed to be done to have telemetry in that setup?

cmd/launcher/console/console.go
```diff
@@ -55,7 +59,8 @@ func NewLauncher() launcher.SubLauncher {
 	fs := flag.NewFlagSet("console", flag.ContinueOnError)
 	fs.StringVar(&config.streamingModeString, "streaming_mode", string(agent.StreamingModeSSE),
 		fmt.Sprintf("defines streaming mode (%s|%s)", agent.StreamingModeNone, agent.StreamingModeSSE))
-
+	fs.BoolVar(&config.otelToCloud, "otel_to_cloud", false, "Enables/disables OpenTelemetry export to cloud")
+	fs.DurationVar(&config.shutdownTimeout, "shutdown-timeout", 15*time.Second, "Console shutdown timeout (i.e. '10s', '2m' - see time.ParseDuration for details) - for waiting for active requests to finish during shutdown")
```
RKest: optional: 15 seconds seems like a long default.

I can very much see myself furiously spamming Ctrl+C if my CLI doesn't stop within 2 seconds

pawel-maciejczek: Changed to 2s.

server/adkrest/handler.go
pawel-maciejczek: It will be handled consistently with adk web or console.

If the user uses the launcher, then the telemetry will be automatically initialized in the web launcher (as a2a Executor is created in web a2a sub-lanucher).

If the user chooses to skip launchers, they can:
1. Use the `telemetry.New()` to initialize OTel SDK - see the doc on that method for example OR
2. Initialize OTel by their own.

telemetry/telemetry.go
pawel-maciejczek: I don't think it's a good idea - the service name and service version are user defined attributes, describing their app - https://opentelemetry.io/docs/specs/semconv/registry/attributes/service

ADK internals instrumentation is a different aspect - in a subsequent PR, we will create TracerProvider, which will use the adk version.

cmd/launcher/telemetry/telemetry.go
pawel-maciejczek: I moved it to internal package - thanks for the idea.

cmd/launcher/console/console.go
pawel-maciejczek: Makes sense. For now I will refer to documentation and options in the telemetry package.

pawel-maciejczek: I disambiguated that the signals are exported to GCP in the flag description.

I agree that the flag name is too generic and ideally it should be more descriptive, e.g. `enable_otel_gcp_export`, but Python and other languages use the same flag name, so I was aiming for the consistency between the implementations.

telemetry/setup_otel.go
pawel-maciejczek: Done - added separate config field for the quota project.

telemetry/telemetry.go
```diff
@@ -12,21 +12,77 @@
...
+	"context"
+
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 
-	internaltelemetry "google.golang.org/adk/internal/telemetry"
+	internal "google.golang.org/adk/internal/telemetry"
+)
+
+const (
+	SystemName = internal.SystemName
```
dpasiukevich: How telemetry package consumers will use this const? Should it be public?

examples/telemetry/main.go
```diff
@@ -0,0 +1,90 @@
...
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"os"
+
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.36.0"
```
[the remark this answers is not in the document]
pawel-maciejczek: Invalid finding.

telemetry/telemetry.go
pawel-maciejczek: It's a leftover from development. I removed it as it's no used any more.

telemetry/setup_otel.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+	if c.googleCredentials != nil && c.googleCredentials.ProjectID != "" {
+		c.quotaProject = c.googleCredentials.ProjectID
+	} else {
+		// The quota project wasn't set in credentials during testing, even when it's set in ADC JSON file.
+		// Using fallback to env variable to resolve the quota project as a workaround.
+		projectID, ok := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
+		if !ok {
+			log.Println("telemetry.googleapis.com requires setting the quota project. Refer to telemetry.config for the available options to set the quota project")
+			return nil
```
dpasiukevich: Should this return error? If not, then method comment should be updated.

telemetry/setup_otel.go
```diff
@@ -0,0 +1,252 @@
...
+			}
+		}
+	}
+
+	if err := cfg.resolveQuotaProject(); err != nil {
+		return nil, err
+	}
+
+	if err := cfg.resolveResourceProject(); err != nil {
+		return nil, err
```
dpasiukevich: Could you add error context to these 2 methods with fmt.Errorf("failed to resolve quota project") and fmt.Errorf("failed to resolve resource project")

telemetry/setup_otel.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+	if c.googleCredentials != nil && c.googleCredentials.ProjectID != "" {
+		c.resourceProject = c.googleCredentials.ProjectID
+	} else {
+		// The resource project wasn't set in credentials during testing, even when it's set in ADC JSON file.
+		// Using fallback to env variable to resolve the resource project as a workaround.
+		projectID, ok := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
+		if !ok {
+			log.Println("telemetry.googleapis.com requires setting the resource project. Refer to telemetry.config for the available options to set the resource project")
+			return nil
```
dpasiukevich: similar regarding returning the error

And also would it make sense to deduplicate this logic: reading c.googleCredentials.ProjectID + GOOGLE_CLOUD_PROJECT env var to a separate method. As it's duplicated resolveQuotaProject and in resolveResourceProject

cmd/launcher/console/console.go
```diff
@@ -55,7 +59,8 @@ func NewLauncher() launcher.SubLauncher {
 	fs := flag.NewFlagSet("console", flag.ContinueOnError)
 	fs.StringVar(&config.streamingModeString, "streaming_mode", string(agent.StreamingModeSSE),
 		fmt.Sprintf("defines streaming mode (%s|%s)", agent.StreamingModeNone, agent.StreamingModeSSE))
-
+	fs.DurationVar(&config.shutdownTimeout, "shutdown-timeout", 2*time.Second, "Console shutdown timeout (i.e. '10s', '2m' - see time.ParseDuration for details) - for waiting for active requests to finish during shutdown")
```
dpasiukevich, line 62: nit: could you update to `shutdown_timeout` for consistency?

telemetry/setup_otel.go
pawel-maciejczek: Done.

cmd/launcher/console/console.go
pawel-maciejczek: Let's unify the flag names between web and console (- vs _) in a separate PR.

telemetry/setup_otel.go
pawel-maciejczek: Done.

telemetry/setup_otel.go
pawel-maciejczek: Done.

telemetry/setup_otel.go
pawel-maciejczek: Done.

PR #535, opened 2026-02-05
feat: add a function call modifier plugin.

baptmont:
add a function call modifier plugin.
refractor duplicated newGeminiTestClientConfig by adding it to test utils

### DIFF
--- a/plugins/functioncallmodifier/plugin.go
+++ b/plugins/functioncallmodifier/plugin.go
@@ -17,6 +17,7 @@ package functioncallmodifier
 
 import (
 	"fmt"
+	"maps"
 
 	"google.golang.org/genai"
 
@@ -66,9 +67,8 @@ func beforeModelCallback(cfg FunctionCallModifierConfig) func(agent.CallbackCont
 						decl.Parameters.Properties = map[string]*genai.Schema{}
 					}
 
-					for name, schema := range cfg.Args {
-						decl.Parameters.Properties[name] = schema
-					}
+					maps.Copy(decl.Parameters.Properties, cfg.Args)
+
 					if cfg.OverrideDescription != nil {
 						decl.Description = cfg.OverrideDescription(decl.Description)
 					}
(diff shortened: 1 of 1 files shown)


internal/testutil/genai.go
```diff
@@ -56,3 +59,20 @@ func scrubGeminiRequest(req *http.Request) error {
...
 }
+
+// NewGeminiTestClientConfig returns the genai.ClientConfig configured for record and replay.
+func NewGeminiTestClientConfig(t *testing.T, rrfile string) *genai.ClientConfig {
+	t.Helper()
+	rr, err := NewGeminiTransport(rrfile)
+	if err != nil {
+		t.Fatal(err)
+	}
+	apiKey := ""
```
yarolegovich, line 70: nit: maybe add `os.Getenv("ADK_RECORD_CAPTURE_GEMINI_API_KEY")`

plugins/functioncallmodifier/plugin.go
```diff
@@ -0,0 +1,111 @@
...
+
+				if shouldAddArgs {
+					if decl.Parameters == nil {
+						decl.Parameters = &genai.Schema{Type: "OBJECT", Properties: map[string]*genai.Schema{}}
+					}
+					if decl.Parameters.Properties == nil {
+						decl.Parameters.Properties = map[string]*genai.Schema{}
+					}
+
+					for name, schema := range cfg.Args {
```
yarolegovich: nit: `maps.Copy(decl.Parameters.Properties, cfg.Args)`

plugins/functioncallmodifier/plugin.go
```diff
@@ -0,0 +1,111 @@
...
+			if fc := part.FunctionCall; fc != nil {
+
+				shouldRemoveArgs := cfg.Predicate(fc.Name)
+				if shouldRemoveArgs {
+					for name := range cfg.Args {
+						arg, hasArg := fc.Args[name]
+						if hasArg {
+							delete(fc.Args, name)
+							stateKey := fmt.Sprintf("%s/%s", fc.ID, name)
+							if err := ctx.State().Set(stateKey, arg); err != nil {
```
yarolegovich: nit: I'd flatten with 
```go
if !cfg.Predicate(fc.Name) {
   continue
}
```
and maybe even
```go
if !hasArg {
  continue
}
```

internal/testutil/genai.go
baptmont: by leaving it empty it will just use GOOGLE_API_KEY

PR #545, opened 2026-02-09
fix: remoteagent partial response handling leads to data duplication

yarolegovich:
TLDR: Fixes data duplication in remote agent responses caused by the presence of response aggregator on the model level.
 
Major changes:
* In streaming mode event stream get demultiplexed into a "partial" and "non-partial" artifacts.
* For the non-partial artifact we send events with append=false so its state represents "the last partial streaming event received". After the execution we send an empty data part to reset any leftover state.
* Mark **all** events produced by ADK agents with `adk_partial` flag, unset is not the same as false.
* If the flag is present remoteagent uses it to decide whether a local event is partial or not.
* If the flag is not present, remoteagent aggregates responses and emits the aggregation when a non-text artifact is received or before a final task status update event.
* If a task event is received aggregation gets reset because the task already contains artifact data aggregated so far.
* Fix long running function calls emitted twice (once in the status and once when they actually happen) - emit only with the final task status event.

Minor changes:
* `a2a:request` metadata is only added to the final adk event to reduce data duplication.
* remoteagent sets `TurnComplete: true` on its last event in all the cases (not all were previously handled).
* Fix `CitationMetadata` was not set in stream aggregator.
* Fix validation error aggregation is input-required state. Previously when client was sending a malformed follow-up we were just appending the error message effectively accumulating errors in the status message.

### DIFF
--- a/server/adka2a/processor_test.go
+++ b/server/adka2a/processor_test.go
@@ -559,7 +559,7 @@ func TestEventProcessor_PartialEventsAreDiscardedAsAnArtifact(t *testing.T) {
 
 func makeTerminalEvents(processor *eventProcessor) []a2a.Event {
 	result := make([]a2a.Event, 0, 2)
-	if finalUpdate, ok := processor.makeFinalArtifactUpdate(); ok {
+	if finalUpdate := processor.makeFinalArtifactUpdate(); finalUpdate != nil {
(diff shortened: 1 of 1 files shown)


dpasiukevich (approved):
🚀

PR #546, opened 2026-02-09
Move the function call modifier plugin into one package

git-hulk:
The plugin `function call modifier` is located in the `plugins` directory, but the `logging` plugin is in the `plugin` directory. I think we should have all plugins in one directory instead of multiple directories? I moved the plugin `function call modifier` since the `plugin` is more consistent with other packages.


git-hulk:
cc @baptmont @dpasiukevich

PR #548, opened 2026-02-09
feat: Implement OpenTelemetry semconv tracing

pawel-maciejczek:
This PR adds semconv tracing for adk and covers:

* invoke_agent
* execute_tool
* generate_content

Limitations:
* `yield` may be called multiple times in scope of a span, but span can only capture single status and error. Therefore, we use the last call of yield to set the result of the span.

Example traces in Google Cloud Trace for single invocation of load artifacts example:

* before -  multiple `call_llm` root spans with unknown status, not linked to each other 
image

* after - single tree with invoke_agent as the root span and correct statuses 
image

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -159,11 +159,7 @@ func (a *agent) SubAgents() []Agent {
 
 func (a *agent) Run(ctx InvocationContext) iter.Seq2[*session.Event, error] {
 	return func(yield func(*session.Event, error) bool) {
-		spanCtx, span := telemetry.StartInvokeAgentSpan(ctx, telemetry.StartInvokeAgentSpanParams{
-			AgentName:        a.name,
-			AgentDescription: a.description,
-			SessionID:        ctx.Session().ID(),
-		})
+		spanCtx, span := telemetry.StartInvokeAgentSpan(ctx, a, ctx.Session().ID())
 		yield, endSpan := telemetry.WrapYield(span, yield, func(span trace.Span, event *session.Event, err error) {
 			telemetry.TraceAgentResult(span, telemetry.TraceAgentResultParams{
 				ResponseEvent: event,
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -55,25 +55,20 @@ var tracer trace.Tracer = otel.GetTracerProvider().Tracer(
 	trace.WithSchemaURL(semconv.SchemaURL),
 )
 
-// StartInvokeAgentSpanParams contains parameters for [StartInvokeAgentSpan].
-type StartInvokeAgentSpanParams struct {
-	// AgentName is the name of the agent being invoked.
-	AgentName string
-	// AgentDescription is a brief description of the agent's purpose.
-	AgentDescription string
-	// SessionID is the unique identifier for the current session.
-	SessionID string
+type agent interface {
+	Name() string
+	Description() string
 }
 
 // StartInvokeAgentSpan starts a new semconv invoke_agent span.
 // It returns a new context with the span and the span itself.
-func StartInvokeAgentSpan(ctx context.Context, params StartInvokeAgentSpanParams) (context.Context, trace.Span) {
-	agentName := params.AgentName
+func StartInvokeAgentSpan(ctx context.Context, agent agent, sessionID string) (context.Context, trace.Span) {
+	agentName := agent.Name()
 	spanCtx, span := tracer.Start(ctx, fmt.Sprintf("invoke_agent %s", agentName), trace.WithAttributes(
 		semconv.GenAIOperationNameInvokeAgent,
-		semconv.GenAIAgentDescription(params.AgentDescription),
+		semconv.GenAIAgentDescription(agent.Description()),
 		semconv.GenAIAgentName(agentName),
-		semconv.GenAIConversationID(params.SessionID),
+		semconv.GenAIConversationID(sessionID),
 	))
 
 	return spanCtx, span
--- a/internal/telemetry/telemetry_test.go
+++ b/internal/telemetry/telemetry_test.go
@@ -105,10 +105,21 @@ func TestWrapYield_MultipleCalls(t *testing.T) {
 
 var errTest = errors.New("test error")
 
+type mockAgent struct{}
+
+func (a *mockAgent) Name() string {
+	return "test-agent"
+}
+
+func (a *mockAgent) Description() string {
+	return "test-agent-description"
+}
+
 func TestInvokeAgent(t *testing.T) {
+	sessionID := "test-session"
+	agent := &mockAgent{}
 	tests := []struct {
 		name         string
-		startParams  StartInvokeAgentSpanParams
 		resultParams TraceAgentResultParams
 		wantName     string
 		wantStatus   codes.Code
@@ -116,11 +127,6 @@ func TestInvokeAgent(t *testing.T) {
 	}{
 		{
 			name: "Success",
-			startParams: StartInvokeAgentSpanParams{
-				AgentName:        "test-agent",
-				AgentDescription: "test-description",
-				SessionID:        "test-session",
-			},
 			resultParams: TraceAgentResultParams{
 				ResponseEvent: session.NewEvent("test-invocation-id"),
 			},
@@ -129,17 +135,12 @@ func TestInvokeAgent(t *testing.T) {
 			wantAttrs: map[attribute.Key]string{
 				semconv.GenAIOperationNameKey:    "invoke_agent",
 				semconv.GenAIAgentNameKey:        "test-agent",
-				semconv.GenAIAgentDescriptionKey: "test-description",
+				semconv.GenAIAgentDescriptionKey: "test-agent-description",
 				semconv.GenAIConversationIDKey:   "test-session",
 			},
 		},
 		{
 			name: "Error",
-			startParams: StartInvokeAgentSpanParams{
-				AgentName:        "test-agent",
-				AgentDescription: "test-description",
-				SessionID:        "test-session",
-			},
 			resultParams: TraceAgentResultParams{
 				Error: errTest,
 			},
@@ -153,7 +154,7 @@ func TestInvokeAgent(t *testing.T) {
 			exporter := setupTestTracer(t)
 			ctx := t.Context()
 
-			_, span := StartInvokeAgentSpan(ctx, tc.startParams)
+			_, span := StartInvokeAgentSpan(ctx, agent, sessionID)
 			TraceAgentResult(span, tc.resultParams)
 			span.End()
 


internal/llminternal/base_flow.go
```diff
@@ -146,9 +147,17 @@ func (f *Flow) runOneStep(ctx agent.InvocationContext) iter.Seq2[*session.Event,
 		if ctx.Ended() {
 			return
 		}
-		sctx, callLLMSpan := telemetry.StartTrace(ctx, "call_llm")
```
RKest, line 149: `call_llm` spans can be removed

internal/llminternal/base_flow.go
```diff
@@ -357,6 +365,29 @@ func (f *Flow) callLLM(ctx agent.InvocationContext, req *model.LLMRequest, state
...
+			ModelName: f.Model.Name(),
+		})
+		ctx = ctx.WithContext(spanCtx)
+		yield, endSpan := telemetry.WrapYield(span, yield, func(span trace.Span, resp *model.LLMResponse, err error) {
+			telemetry.AfterGenerateContent(span, telemetry.AfterGenerateContentParams{
+				Response: resp,
+				Error:    err,
+			})
+		})
+		defer endSpan()
```
RKest: I'm pretty sure `defer` here will run into the issue where `generate_content` span will capture time:
1. After last `LLMResponse` is returned
2. During all propagation and handling of this `LLMResponse` up the call chain
3. Until handing of this `LLMResponse` is finished, and some iterator up the call chain reaches for another `LLMResponse`, only to find it's missing.

Can you instead call `endSpan` when first non-partial `LLMResponse` is returned or `yield` returns `false`?

internal/llminternal/base_flow.go
```diff
@@ -490,77 +521,100 @@ Suggested fixes:
...
+func (f *Flow) handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[string]tool.Tool, resp *model.LLMResponse, toolConfirmations map[string]*toolconfirmation.ToolConfirmation) (mergedEvent *session.Event, err error) {
 	var fnResponseEvents []*session.Event
-
 	fnCalls := utils.FunctionCalls(resp.Content)
 	toolNames := slices.Collect(maps.Keys(toolsDict))
 	var result map[string]any
+	// Don't generate empty spans if there are no function calls.
+	if len(fnCalls) > 0 {
+		// this is needed for debug traces of parallel calls
+		mergedCtx, mergedToolCallSpan := telemetry.StartTrace(ctx, "execute_tool (merged)")
```
RKest, line 542: In ADK Python, I think `"execute_tool (merged)"` are only generated if there's parallel tool execution.

Can we do `len(fnCalls) > 1`

agent/agent.go
```diff
@@ -157,9 +159,21 @@ func (a *agent) SubAgents() []Agent {
 
 func (a *agent) Run(ctx InvocationContext) iter.Seq2[*session.Event, error] {
 	return func(yield func(*session.Event, error) bool) {
+		spanCtx, span := telemetry.StartInvokeAgent(ctx, telemetry.StartInvokeAgentParams{
```
RKest: optional: `StartInvokeAgent` -> `StartInvokeAgentSpan`

Same for `Start_` functions

RKest: Instead of passing custom struct, I'd say we could pass entire agent and invocation context.

This would:
1. Make the call site clearer
2. Allow extending the instrumentation without modifying the call site
3. Allow adding Agent-specific attrbiutes. For example if we wanted to add some attribute specific to LLMAgent, or MCPTool (I'd recommend the same change for tool spans as well) in the future.

internal/llminternal/base_flow.go
```diff
@@ -490,77 +515,100 @@ Suggested fixes:
...
 // TODO: check feasibility of running tool.Run concurrently.
-func (f *Flow) handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[string]tool.Tool, resp *model.LLMResponse, toolConfirmations map[string]*toolconfirmation.ToolConfirmation) (*session.Event, error) {
+func (f *Flow) handleFunctionCalls(ctx agent.InvocationContext, toolsDict map[string]tool.Tool, resp *model.LLMResponse, toolConfirmations map[string]*toolconfirmation.ToolConfirmation) (mergedEvent *session.Event, err error) {
 	var fnResponseEvents []*session.Event
-
 	fnCalls := utils.FunctionCalls(resp.Content)
 	toolNames := slices.Collect(maps.Keys(toolsDict))
 	var result map[string]any
+	// Don't generate empty spans if there are no function calls.
+	if len(fnCalls) > 0 {
```
pawel-maciejczek: Right now I'm skipping span creation if there are no function calls, but maybe `handleFunctionCalls` should just return early  empty response in that case?

internal/telemetry/telemetry.go
```diff
@@ -12,162 +12,318 @@
...
+	genAiResponseCachedContentTokenCount = attribute.Key("gen_ai.response.cached_content_token_count")
+	genAiResponseTotalTokenCount         = attribute.Key("gen_ai.response.total_token_count")
+
+	gcpVertexAgentLLMRequestName   = attribute.Key("gcp.vertex.agent.llm_request")
+	gcpVertexAgentToolCallArgsName = attribute.Key("gcp.vertex.agent.tool_call_args")
+	gcpVertexAgentEventID          = attribute.Key("gcp.vertex.agent.event_id")
+	gcpVertexAgentToolResponseName = attribute.Key("gcp.vertex.agent.tool_response")
+	gcpVertexAgentLLMResponseName  = attribute.Key("gcp.vertex.agent.llm_response")
+	gcpVertexAgentInvocationID     = attribute.Key("gcp.vertex.agent.invocation_id")
+	gcpVertexAgentSessionID        = attribute.Key("gcp.vertex.agent.session_id")
```
RKest: Remove session_id and replace it with the semconv eqivalent wherever it's used
https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/registry/attributes/gen-ai.md#gen-ai-conversation-id

internal/telemetry/telemetry.go
```diff
@@ -12,162 +12,318 @@
...
+	genAiResponsePromptTokenCount        = attribute.Key("gen_ai.response.prompt_token_count")
+	genAiResponseCandidatesTokenCount    = attribute.Key("gen_ai.response.candidates_token_count")
+	genAiResponseCachedContentTokenCount = attribute.Key("gen_ai.response.cached_content_token_count")
+	genAiResponseTotalTokenCount         = attribute.Key("gen_ai.response.total_token_count")
+
+	gcpVertexAgentLLMRequestName   = attribute.Key("gcp.vertex.agent.llm_request")
+	gcpVertexAgentToolCallArgsName = attribute.Key("gcp.vertex.agent.tool_call_args")
+	gcpVertexAgentEventID          = attribute.Key("gcp.vertex.agent.event_id")
+	gcpVertexAgentToolResponseName = attribute.Key("gcp.vertex.agent.tool_response")
+	gcpVertexAgentLLMResponseName  = attribute.Key("gcp.vertex.agent.llm_response")
```
RKest: Remove `.*llm_request` and `.*llm_response`

These shouldn't be set outside of `call_llm` which should be removed

internal/telemetry/telemetry.go
```diff
@@ -12,162 +12,318 @@
...
 )
 
-// StartTrace returns two spans to start emitting events, one from global tracer and second from the local.
-func StartTrace(ctx context.Context, traceName string) (context.Context, trace.Span) {
-	return tracer.Start(ctx, traceName)
+var (
+	genAiResponsePromptTokenCount        = attribute.Key("gen_ai.response.prompt_token_count")
+	genAiResponseCandidatesTokenCount    = attribute.Key("gen_ai.response.candidates_token_count")
+	genAiResponseCachedContentTokenCount = attribute.Key("gen_ai.response.cached_content_token_count")
+	genAiResponseTotalTokenCount         = attribute.Key("gen_ai.response.total_token_count")
```
RKest: AFAIU these are not in the semconv
https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/registry/attributes/gen-ai.md#gen-ai-usage-input-tokens

I would only set input and output tokens then

internal/telemetry/telemetry.go
```diff
@@ -12,162 +12,318 @@
...
+type AfterExecuteToolParams struct {
+	Name          string
+	Description   string
+	Args          map[string]any
+	ResponseEvent *session.Event
+	Error         error
+}
+
+// AfterExecuteTool records the tool execution events.
+func AfterExecuteTool(span trace.Span, params AfterExecuteToolParams) {
```
RKest: optional: `AfterExecuteTool` -> `TraceToolResult`

internal/telemetry/telemetry.go
```diff
@@ -12,142 +12,285 @@
...
+		semconv.GenAIOperationNameExecuteTool,
+		semconv.GenAIToolName(toolName),
+	))
+	return spanCtx, span
+}
+
+type AfterExecuteToolParams struct {
+	Name          string
+	Description   string
+	Args          map[string]any
```
RKest: `Name` is duplicated between here and `StartExecuteToolParams`, and I would assume `Description` and `Args` would fit more in the `StartExecuteToolParams` as well

internal/telemetry/telemetry.go
```diff
@@ -12,142 +12,285 @@
...
 // TraceLLMCall fills the call_llm event details.
-func TraceLLMCall(span trace.Span, agentCtx agent.InvocationContext, llmRequest *model.LLMRequest, event *session.Event) {
-	sessionID := agentCtx.Session().ID()
+func TraceLLMCall(span trace.Span, sessionID string, llmRequest *model.LLMRequest, event *session.Event) {
 	attributes := []attribute.KeyValue{
 		attribute.String(genAiSystemName, systemName),
 		attribute.String(genAiRequestModelName, llmRequest.Model),
-		attribute.String(gcpVertexAgentInvocationID, event.InvocationID),
-		attribute.String(gcpVertexAgentSessionID, sessionID),
+		gcpVertexAgentInvocationID.String(event.InvocationID),
```
RKest: optional: `attributeKey.Type(attributeValue)` api is slightly confusing.

I preferred all of these to be `attribute.Type(attributeKey, attributeValue)`

But if types imported from golang semconv package don't fit nicely into that api, then it's fine to stay as-is

internal/telemetry/telemetry.go
```diff
@@ -12,142 +12,285 @@
...
-	attributes = append(attributes, attribute.String(gcpVertexAgentToolResponseName, toolResponse))
+	attributes = append(attributes, semconv.GenAIToolCallIDKey.String(toolCallID))
+	attributes = append(attributes, gcpVertexAgentToolResponseName.String(toolResponse))
+
+	span.SetAttributes(attributes...)
+}
+
+func recordErrorAndStatus(span trace.Span, err error) {
+	if err == nil {
+		span.SetStatus(codes.Ok, "")
```
RKest: Leave status unspecified

It's a quirk of OTel https://opentelemetry.io/docs/concepts/signals/traces/#span-status

internal/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,378 @@
...
+			if tc.afterParams.Error != nil {
+				if gotSpan.Status.Description != tc.afterParams.Error.Error() {
+					t.Errorf("expected status description %q, got %q", tc.afterParams.Error.Error(), gotSpan.Status.Description)
+				}
+			}
+
+			if tc.wantAttrs != nil {
+				gotAttrs := attributesToMap(gotSpan.Attributes)
+				for k, v := range tc.wantAttrs {
+					if gotAttrs[k] != v {
```
RKest, line 267: optional: use https://github.com/google/go-cmp

(if it's tedious to add the dependency it's okay to stay as-is)

internal/llminternal/base_flow.go
pawel-maciejczek: I'll do that in a separate PR after merging this one to avoid conflicts. I already have a draft [here](https://github.com/google/adk-go/pull/556)

internal/telemetry/telemetry.go
pawel-maciejczek: Removed the duplication and moved the args to StartParams. Description is not available when we create a span, so I'll keep it in here.

internal/telemetry/telemetry.go
pawel-maciejczek: That's the way otel API works in golang, so I'll keep it as-is. I prefer to use semconv to avoid any typos in attribute names.

internal/telemetry/telemetry.go
pawel-maciejczek: Thanks - done.

internal/telemetry/telemetry_test.go
pawel-maciejczek: My intention was to avoid test failure if there is extra attribute added to the span, but I can change that.

internal/telemetry/telemetry.go
pawel-maciejczek: Done for this span and others.

agent/agent.go
pawel-maciejczek: Done.

pawel-maciejczek: I'm not using agent.Agent nor agent.InvocationContext in tracing to avoid cyclic dependency between `telemetry` and `agent` packages (compilation error).

The workaround for this is to create a private agent interface in telemetry:

```go
package telemetry

type agent interface {
	Name() string
	Description() string
}

// StartInvokeAgentSpan starts a new semconv invoke_agent span.
// It returns a new context with the span and the span itself.
func StartInvokeAgentSpan(ctx context.Context, agent agent, sessionID string) (context.Context, trace.Span) {
       ...
}

```
Then I could pass the agent as a parameter:
```go
package agent

func (a *agent) Run(ctx InvocationContext) iter.Seq2[*session.Event, error] {
		spanCtx, span := telemetry.StartInvokeAgentSpan(ctx, a, ctx.Session().ID())
        ...
        }
}
```

It should help with 1) and 2).
I'm not sure about 3), but if the LLMAgent is in a different package, then I guess we won't have the problem with cyclic dependency.

internal/telemetry/telemetry.go
pawel-maciejczek: Those are  used in `llm_call`, so I inlined those attributes in TraceLLMCall.

internal/llminternal/base_flow.go
pawel-maciejczek: Done.

internal/llminternal/base_flow.go
pawel-maciejczek: You're right - the current implementation will capture all that upstream processing time.

I think we can end the span when the firs non-partial response is returned with the assumption that a model won't return multiple non-partial responses for a single request.

pawel-maciejczek: I pushed the changes and added tests to verify that the span is ended before yield is called.

internal/llminternal/base_flow_tracing_test.go
```diff
@@ -0,0 +1,267 @@
...
+				}
+			}
+		},
+	}
+
+	ctx := icontext.NewInvocationContext(context.Background(), icontext.InvocationContextParams{})
+
+	for _, err := range generateContent(ctx, modelMock, &model.LLMRequest{}, true) {
+		_ = err
+	}
```
dpasiukevich:
```suggestion
	for range generateContent(ctx, modelMock, &model.LLMRequest{}, true) {
	}
```

here and in other places

dpasiukevich (approved):
LGTM!

dpasiukevich:
@RKest could I ask to ptal as well?

internal/llminternal/base_flow.go
```diff
@@ -357,6 +360,39 @@ func (f *Flow) callLLM(ctx agent.InvocationContext, req *model.LLMRequest, state
...
+		var lastErr error
+		endSpanAndTrackResult := func() {
+			telemetry.TraceGenerateContentResult(span, telemetry.TraceGenerateContentResultParams{
+				Response: lastResponse,
+				Error:    lastErr,
+			})
+			span.End()
+		}
+		// Ensure that the span is ended in case of error or if none final responses are yielded before the yield returns false.
+		defer endSpanAndTrackResult()
```
RKest, line 387: optional: In Python ending a span multiple times produces a warning, and this would produce a lot of spam if the same was implemented in Go

Just in case, could you guard the end with local
```go
var spanEnded bool
```

internal/telemetry/telemetry.go
RKest: invocation_id is fine to set

`gcp.vertex.agent.session_id` has a semconv eqivalent, so it should be `gen_ai.conversation.id` in all cases

We never set the old `gcp.vertex.agent.session_id` outside of `call_llm`, which should not be implemented in Go

internal/telemetry/telemetry.go
```diff
@@ -12,142 +12,283 @@
...
 		// TODO: add tool type
 
 		// Setting empty llm request and response (as UI expect these) while not
 		// applicable for tool_response.
-		attribute.String(gcpVertexAgentLLMRequestName, "{}"),
-		attribute.String(gcpVertexAgentLLMResponseName, "{}"),
-		attribute.String(gcpVertexAgentToolCallArgsName, safeSerialize(fnArgs)),
-		attribute.String(gcpVertexAgentEventID, fnResponseEvent.ID),
+		gcpVertexAgentLLMRequestName.String("{}"),
+		gcpVertexAgentLLMResponseName.String("{}"),
```
RKest: No need to set these, UI will work without these

agent/agent.go
RKest: Makes sense to resolve cyclic depenedency

For 3) I believe as long as we pass any interface, we can do type checks on it later on, so it should help with it as well👍

internal/telemetry/telemetry.go
pawel-maciejczek: Done.

internal/telemetry/telemetry.go
pawel-maciejczek: Done - inlined in call_llm, which will be removed.

internal/telemetry/telemetry.go
pawel-maciejczek: Done.

internal/llminternal/base_flow_tracing_test.go
pawel-maciejczek: Done.

internal/llminternal/base_flow.go
pawel-maciejczek: Done. I haven't seen those warns yet, so maybe go doesn't produce them.

agent/agent.go
pawel-maciejczek: Done.

internal/telemetry/telemetry.go
pawel-maciejczek: Done - new methods use semconv equivalent and I inlined `gcp.vertex.agent.session_id` in `TraceLLMCall`

PR #552, opened 2026-02-10
feat: add otel structured logging

pawel-maciejczek:
This PR introduces structured logging for GenAI requests and responses.

Logging is based on OTel GenAI semconv (linked from the code), but aims for consistency with ADK Python implementation. The deviations between the current implementation and semconv are documented in the code.

### DIFF
--- a/telemetry/setup_otel.go
+++ b/telemetry/setup_otel.go
@@ -93,8 +93,9 @@ func newInternal(cfg *config) (*Providers, error) {
 	// TODO(#479) init meter provider
 
 	return &Providers{
-		TracerProvider: tp,
-		LoggerProvider: lp,
+		TracerProvider:             tp,
+		genAICaptureMessageContent: cfg.genAICaptureMessageContent,
+		LoggerProvider:             lp,
 	}, nil
 }
 


internal/llminternal/base_flow.go
```diff
@@ -309,6 +309,8 @@ func (f *Flow) callLLM(ctx agent.InvocationContext, req *model.LLMRequest, state
 		// TODO: RunLive mode when invocation_context.run_config.support_cfc is true.
 		useStream := runconfig.FromContext(ctx).StreamingMode == runconfig.StreamingModeSSE
 
+		// Log request before calling the model.
+		telemetry.LogRequest(ctx, req)
```
RKest: nit: Is it possible to hide this and `telemetry.LogResponse` in the `generateContent` function?

In python we take up a pretty small code footprint when it comes to instrumentation in the business logic:
https://github.com/google/adk-python/blob/ede925b5025972cffcfaf178b2f81679fabbe90f/src/google/adk/flows/llm_flows/base_llm_flow.py#L1166-L1170

internal/telemetry/logger.go
```diff
@@ -0,0 +1,227 @@
...
+// Message content is not logged by default. Set the following env variable to enable logging of prompt/response content.
+// OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true
+var elideMessageContent = !isEnvVarTrue("OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT")
+
+const elidedContent = "<elided>"
+
+// guessedGenAISystem is the AI system we are using.
+var guessedGenAISystem = guessAISystem()
+
+var logger = global.GetLoggerProvider().Logger(
```
RKest: nit: rename to `otelLogger`.

It's a bit confusing, because most languages have some logger already, and it usually signals stdout logging

internal/telemetry/logger.go
```diff
@@ -0,0 +1,227 @@
...
+	semconv "go.opentelemetry.io/otel/semconv/v1.36.0"
+	"google.golang.org/genai"
+
+	"google.golang.org/adk/internal/version"
+	"google.golang.org/adk/model"
+)
+
+// Message content is not logged by default. Set the following env variable to enable logging of prompt/response content.
+// OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true
+var elideMessageContent = !isEnvVarTrue("OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT")
```
dpasiukevich: @pawel-maciejczek @RKest folks do you know if this is mentioned in the public docs anywhere for ADK? This envvar looks useful, but right now it's hidden in the internal code and users won't know about it unless they'd parse the code.

internal/telemetry/logger.go
```diff
@@ -0,0 +1,227 @@
...
+func isEnvVarTrue(name string) bool {
+	val, ok := os.LookupEnv(name)
+	if !ok {
+		return false
+	}
+	val = strings.ToLower(val)
+	return val == "true" || val == "1"
+}
+
+func guessAISystem() string {
```
dpasiukevich: Could you use this func https://github.com/google/adk-go/blob/147b75759ac648387a10e126df8e9b19d98932eb/internal/llminternal/googlellm/variant.go#L30

VertexAI or GeminiAPI not only can be configured by this env var but also by genai.Client.Backend field.

internal/telemetry/logger.go
dpasiukevich: Just to double check -- logger provider instantiation will be done in a separate PR?

https://github.com/google/adk-go/blob/147b75759ac648387a10e126df8e9b19d98932eb/telemetry/setup_otel.go#L80

verdverm:
general question on this, will it enable me to add labels to various groupings in ADK such that I have some good filtering and pivot variables for grafana?

internal/llminternal/base_flow.go
pawel-maciejczek: Sure, SGTM.

internal/telemetry/logger.go
pawel-maciejczek: Yes, I have a draft in https://github.com/google/adk-go/pull/573. Thanks for asking.

internal/telemetry/logger.go
pawel-maciejczek: I think I can add as a field in telemetry.config together with an option to override it programmatically and document that the default value is taken from the env variable.
The env var is mentioned in https://docs.cloud.google.com/stackdriver/docs/instrumentation/ai-agent-adk#configure and https://opentelemetry.io/blog/2024/otel-generative-ai/#example-usage, so they don't have greatest visiblity.

pawel-maciejczek: I added genAICaptureMessageContent to telemetry.config in a separate commit - PTAL.

internal/telemetry/logger.go
pawel-maciejczek: Done.

pawel-maciejczek:
> general question on this, will it enable me to add labels to various groupings in ADK such that I have some good filtering and pivot variables for grafana?

@verdverm I'm not quite sure what do you mean by that, could you clarify?
This PR will just emit logs in otel format. I think that grafana querying options depend on the backend to which the logs are then exported.

verdverm:
I'm thinking about custom attributes at emission time, so that I can have app specific fields and values to filter on later.

- Traces: https://opentelemetry.io/docs/languages/go/instrumentation/#span-attributes
- Metrics: https://opentelemetry.io/docs/languages/go/instrumentation/#adding-attributes
- Logging: https://pkg.go.dev/go.opentelemetry.io/otel/sdk/log#Record.SetAttributes

I generally think about OTEL as a single thing with events I can label across those signals for correlation. Being able to apply these app custom attributes across the signals consistently is important. These probably need to be set at similar granularity to the app > user > session hierarchy that exists in other places in the SDKs

RKest:
@verdverm You can add custom attributes at emission time, through the use of plugins

Example for `invoke_agent` span is:
```go
plug, err := plugin.New(plugin.Config{
    Name:                "Your plugin",
    BeforeAgentCallback: beforeAgentCallback,
})
func beforeAgentCallback(ctx agent.CallbackContext) (*genai.Content, error) {
  span := trace.SpanFromContext(ctx)
  span.SetAttributes(yourAttributes)
  return nil, nil
}
```

It's possible to do the same for `execute_tool` spans. For `generate_content` spans it's a bit more challenging (still possible through a custom SpanProcessor), because these by design don't include plugin callbacks

internal/telemetry/logger_test.go
```diff
@@ -0,0 +1,598 @@
...
+							{Text: "User message part 2"},
+						},
+					},
+				},
+			},
+			wantEvents: []wantEvent{
+				{
+					name: "gen_ai.system.message",
+					body: map[string]any{
+						"content": "System instruction part 1\nSystem instruction part 2",
```
dpasiukevich, line 86: Curious why it's inconsistent with `gen_ai.user.message` which has it as `"parts": []any{ ... }` and here it's a string?

internal/telemetry/logger.go
```diff
@@ -0,0 +1,232 @@
...
+// If the content is elided, it returns the elided content string.
+func extractSystemMessage(req *model.LLMRequest) log.Value {
+	if !getGenAICaptureMessageContent() {
+		return log.StringValue(elidedContent)
+	}
+	if req == nil || req.Config == nil || req.Config.SystemInstruction == nil {
+		return log.Value{}
+	}
+	var text []string
+	for _, p := range req.Config.SystemInstruction.Parts {
```
dpasiukevich, line 158: Following up on https://github.com/google/adk-go/pull/552/changes#r2822633831

Can we use contentToLogValue(req.Config.SystemInstruction)?

internal/telemetry/logger.go
```diff
@@ -0,0 +1,232 @@
...
+		if p.Text != "" {
+			text = append(text, p.Text)
+		}
+	}
+	content := strings.Join(text, "\n")
+	return log.StringValue(content)
+}
```
dpasiukevich, line 165:
```suggestion
func extractSystemMessage(req *model.LLMRequest) log.Value {
	if !getGenAICaptureMessageContent() {
		return log.StringValue(elidedContent)
	}
	if req == nil || req.Config == nil {
		return log.Value{}
	}
	return contentToLogValue(req.Config.SystemInstruction)
}
```

Something like this.

internal/telemetry/logger_test.go
pawel-maciejczek: It's a quirk of the current implementation in ADK Python. The log body can be any object and for each of those events, the body is different - `string` vs `genai.Content` vs `genai.Content` nested under a `message` field.

dpasiukevich: We discussed, this is the approach in adk-python and for now we'd keep the compatibility.

Yet we'd follow up with the adk-python team to make sure we can improve it here.

verdverm:
@RKest to do spans correctly, you need to derive new spans from parent spans, which means we need access to the same OTEL handles the framework is using, in this case, how do I know if the span is for the main agent or a subagent? (i.e. there is additional hierarchy outside of the span tree itself, which informs that tree)

PR #557, opened 2026-02-11
fix: Introduce `ExitLoop` action and tool for `LoopAgent` to exit loo…

### LINKED ISSUE #522: [bug]exit_loop with Escalate=true stops parent SequentialAgent pipeline
**Describe the bug**

When using `exit_loop` tool inside a `loopagent` that is nested within a `sequentialagent`, the `Escalate=true` flag propagates up and stops the entire sequential pipeline, not just the inner loop.

**Pipeline structure:**
```
Sequential(Loop(Researcher → Executor → Evaluator), Summary)
```

When the Evaluator agent calls `exit_loop`:
- Loop exits ✅
- Sequential **also exits** ❌ (should continue)
- Summary **never executes** ❌
- Pipeline hangs with no response

**To Reproduce**

Minimal code to reproduce:

```go
package main

import (
	"context"
	"google.golang.org/adk/agent"
	"google.golang.org/adk/agent/workflowagents/loopagent"
	"google.golang.org/adk/agent/workflowagents/sequentialagent"
	"google.golang.org/adk/genai/genaiagent"
	"google.golang.org/adk/tool/exitlooptool"
)

func main() {
	ctx := context.Background()
	model := genaiagent.Model("gemini-2.0-flash-exp")

	// Agent that calls exit_loop
	evaluator, _ := llmagent.New(llmagent.Config{
		Config: agent.Config{
			Name:        "evaluator",
			Description: "Evaluates if loop should exit",
		},
		Model:    model,
		Toolsets: []agent.Toolset{exitlooptool.Toolset()},
	})

	// Other agents (simple for reproduction)
	researcher, _ := llmagent.New(llmagent.Config{
		Config: agent.Config{Name: "researcher"},
		Model:  model,
	})
	
	executor, _ := llmagent.New(llmagent.Config{
		Config: agent.Config{Name: "executor"},
		Model:  model,
	})

	summary, _ := llmagent.New(llmagent.Config{
		Config: agent.Config{Name: "summary"},
		Model:  model,
	})

	// Create loop with exit_loop tool
	loop, _ := loopagent.New(loopagent.Config{
		AgentConfig: agent.Config{
			Name:      "research-loop",
			SubAgents: []agent.Agent{researcher, executor, evaluator},
		},
		MaxIterations: 2,
	})

	// Create sequential pipeline
	pipeline, _ := sequentialagent.New(sequentialagent.Config{
		AgentConfig: agent.Config{
			Name:      "reasoning-pipeline",
			SubAgents: []agent.Agent{loop, summary},
		},
	})

	// Run with prompt that triggers exit_loop
	// Result: Summary never executes
	pipeline.Run(ctx, "Analyze metrics and exit loop after first iteration")
}
```

Steps to reproduce:
1. Install ADK v0.3.0: `go get google.golang.org/adk@v0.3.0`
2. Run the code above
3. Send a request that causes evaluator to call `exit_loop`
4. Observe that Summary agent never starts (check logs)

**Expected behavior**

When `exit_loop` is called inside the nested loop:
1. The loop should exit immediately
2. The sequential agent should **continue** to the next sub-agent (Summary)
3. Summary should execute and return a response

- OS: Linux
- Go version: 1.24+
- ADK version: v0.3.0

**Model Information:**
- Which model is being used: claude-sonnet-4-20250514 / gemini-2.0-flash-exp (bug happens with both)

**Additional context**

## Root Cause Analysis

The issue occurs because:

1. **exit_loop sets Escalate=true** (`tool/exitlooptool/tool.go`):
```go
func exitLoop(ctx tool.Context, myArgs EmptyArgs) (map[string]string, error) {
    ctx.Actions().Escalate = true        // <-- Sets escalate
    ctx.Actions().SkipSummarization = true
    return map[string]string{}, nil
}
```

2. **loopagent propagates Escalate event before exiting** (`agent/workflowagents/loopagent/agent.go`):
```go
    if !yield(event, err) {  // <-- Event with Escalate=true is yielded UP
        return
    }
    if event.Actions.Escalate {
        shouldExit = true    // <-- Loop detects escalate
    }
}
if shouldExit {
    return                   // <-- Exits but event already propagated
}
```

3. **sequentialagent IS a loopagent** (`agent/workflowagents/sequentialagent/agent.go`):
```go
func New(cfg Config) (agent.Agent, error) {
    sequentialAgent, err := loopagent.New(loopagent.Config{
        MaxIterations: 1,  // <-- Sequential is just a loop with 1 iteration
    })
```

Since `sequentialagent` is internally a `loopagent`, it **also reacts** to the propagated `Escalate=true` and exits immediately, skipping remaining sub-agents.

MarsZ42:
# Summary

- Fixes #522: exit_loop with Escalate=true was propagating to parent agents, causing SequentialAgent (which is internally a LoopAgent) to also exit and skip
remaining sub-agents
- Introduces a dedicated ExitLoop field on EventActions, decoupling loop exit from the Escalate flag
- LoopAgent consumes the ExitLoop flag before yielding events upstream, so parent agents are unaffected
- Escalate retains its original semantics ("escalate to a higher level agent") and passes through loops without triggering exit

# Test plan

- Updated loopagent tests to use ExitLoop instead of Escalate
- Added TestSequentialAgentWithNestedLoop verifying Sequential(Loop(..., exit_loop), Summary) continues to Summary
- All existing tests pass (go test ./...)

### DIFF
--- a/agent/workflowagents/loopagent/agent.go
+++ b/agent/workflowagents/loopagent/agent.go
@@ -81,12 +81,15 @@ func (a *loopAgent) Run(ctx agent.InvocationContext) iter.Seq2[*session.Event, e
 			for _, subAgent := range ctx.Agent().SubAgents() {
 				for event, err := range subAgent.Run(ctx) {
 					// TODO: ensure consistency -- if there's an error, return and close iterator, verify everywhere in ADK.
+					if event != nil && event.Actions.ExitLoop {
+						shouldExit = true
+						// Consume the ExitLoop flag so parent agents
+						// (e.g. a SequentialAgent wrapping this loop)
+						// don't also react to it and exit prematurely.
+						event.Actions.ExitLoop = false
+					}
 					if !yield(event, err) {
 						return
-					}
-
-					if event != nil && event.Actions.Escalate {
-						shouldExit = true
 					}
 				}
 				if shouldExit {
--- a/agent/workflowagents/loopagent/agent_test.go
+++ b/agent/workflowagents/loopagent/agent_test.go
@@ -118,7 +118,7 @@ func TestNewLoopAgent(t *testing.T) {
 			},
 		},
 		{
-			name: "loop with escalate function returns sumarization",
+			name: "loop with exit_loop function returns summarization",
 			args: args{
 				maxIterations: 2,
 				subAgents:     []agent.Agent{newLmmAgentWithFunctionCall(t, 0, false), newCustomAgent(t, 1)},
@@ -135,9 +135,6 @@ func TestNewLoopAgent(t *testing.T) {
 					LLMResponse: model.LLMResponse{
 						Content: genai.NewContentFromFunctionResponse("exampleFunction", make(map[string]any), genai.RoleUser),
 					},
-					Actions: session.EventActions{
-						Escalate: true,
-					},
 				},
 				{
 					Author: "custom_agent_0",
@@ -153,7 +150,7 @@ func TestNewLoopAgent(t *testing.T) {
 			},
 		},
 		{
-			name: "loop with escalate function returns sumarization",
+			name: "loop with exit_loop function skips summarization",
 			args: args{
 				maxIterations: 2,
 				subAgents:     []agent.Agent{newLmmAgentWithFunctionCall(t, 0, true), newCustomAgent(t, 1)},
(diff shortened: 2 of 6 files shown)


wolo-lab:
Thank you for digging into this, and sorry that nobody responded at all. That should not have happened.

The bug in #522 came from SequentialAgent being built on top of LoopAgent, so an escalate event from an inner loop stopped the outer pipeline. That dependency was removed in #611, three weeks after you opened this. SequentialAgent now yields sub-agent events without inspecting Actions at all, so the pipeline no longer stops early. We reproduced the pipeline shape from #522 against current main and the step after the loop runs.

On the design itself: main already exits loops through EventActions.Escalate, which exitlooptool sets and LoopAgent reads. Having the loop clear the flag before yielding upstream would also diverge from adk-python, where loop_agent yields the event unmodified and the flag stays visible to the parent.

We will close #522 as well. We are reviewing incoming PRs much more actively now.

PR #559, opened 2026-02-11
fix: generate UUID for request confirmation events

### LINKED ISSUE #558: generateRequestConfirmationEvent creates events with empty ID, causing DB primary key violations
## 🔴 Required Information

**Describe the Bug:**
`generateRequestConfirmationEvent()` in `internal/llminternal/functions.go` creates `session.Event` structs without setting the `ID` field, resulting in empty string IDs. When events are persisted via a database session service (e.g., `database.NewSessionService` with PostgreSQL/GORM), the composite primary key `(ID, AppName, UserID, SessionID)` causes a duplicate key violation on the second HITL confirmation event within the same session.

Unlike `session.NewEvent()` which generates a UUID via `uuid.NewString()`, `generateRequestConfirmationEvent()` constructs the Event directly without setting `ID`.

**Steps to Reproduce:**
1. Configure ADK with a database session service (e.g., PostgreSQL via `database.NewSessionService`)
2. Create a tool that uses `RequestConfirmation()` or `RequireConfirmation`
3. Complete the first HITL confirmation flow — works fine
4. Trigger a second confirmation in the same session
5. Observe error: `duplicate key value violates unique constraint "events_pkey"`

**Expected Behavior:**
Each request confirmation event should have a unique UUID, consistent with `session.NewEvent()` behavior.

**Observed Behavior:**
All request confirmation events have `id = ''` (empty string). The first insert succeeds, but subsequent inserts within the same session fail with a primary key constraint violation.

```text
ERROR: duplicate key value violates unique constraint "events_pkey" (SQLSTATE 23505)
```

**Environment Details:**
- ADK Library Version: v0.4.0
- OS: macOS (Darwin 25.0.0)
- Go Version: 1.25.7

**Model Information:**
- Any model (bug is in event creation, not model interaction)

---

## 🟡 Optional Information

**Regression:**
N/A — this has likely existed since `generateRequestConfirmationEvent` was introduced.

**Root Cause:**
`internal/llminternal/functions.go` line 86-96 — the Event struct literal does not set `ID`:

```go
return &session.Event{
    // ID field is missing here
    InvocationID: invocationContext.InvocationID(),
    Author:       invocationContext.Agent().Name(),
    // ...
}
```

Compare with `session.NewEvent()` in `session/session.go` which properly sets `ID: uuid.NewString()`.

- Always (100%) — every request confirmation event has an empty ID

code2gold:
Fixes #558

## Summary

`generateRequestConfirmationEvent()` creates `session.Event` structs without setting the `ID` field, resulting in empty string IDs. When persisted to a database backend, the composite primary key `(ID, AppName, UserID, SessionID)` causes a duplicate key violation on the second HITL confirmation event within the same session.

**Fix:** Add `ID: uuid.NewString()` to the Event struct initialization, consistent with `session.NewEvent()`.

## Changes

- `internal/llminternal/functions.go`: Add `ID: uuid.NewString()` and `uuid` import
- `internal/llminternal/base_flow_test.go`: Add test verifying generated events have non-empty, unique IDs

## Testing Plan

**Unit test added:**
- `TestGenerateRequestConfirmationEvent_SetsEventID` — verifies the generated event has a non-empty ID, and calling twice produces unique IDs

**Manual E2E verification:**
- Configured ADK with `database.NewSessionService` (PostgreSQL via GORM)
- Triggered multiple HITL confirmation flows within the same session
- Before fix: second confirmation fails with `duplicate key value violates unique constraint "events_pkey"`
- After fix: all confirmation events persist successfully with unique UUIDs

```text
$ go test ./internal/llminternal/ -run TestGenerateRequestConfirmationEvent -v
=== RUN   TestGenerateRequestConfirmationEvent_SetsEventID
--- PASS: TestGenerateRequestConfirmationEvent_SetsEventID (0.00s)
PASS
```

### DIFF
--- a/internal/llminternal/functions.go
+++ b/internal/llminternal/functions.go
@@ -17,6 +17,7 @@ package llminternal
 import (
 	"time"
 
+	"github.com/google/uuid"
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/agent"
@@ -84,6 +85,7 @@ func generateRequestConfirmationEvent(
 	utils.PopulateClientFunctionCallID(content)
 
 	return &session.Event{
+		ID:           uuid.NewString(),
 		InvocationID: invocationContext.InvocationID(),
 		Author:       invocationContext.Agent().Name(),
 		Branch:       invocationContext.Branch(),
--- a/internal/llminternal/base_flow_test.go
+++ b/internal/llminternal/base_flow_test.go
@@ -21,11 +21,13 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"google.golang.org/genai"
 
+	"google.golang.org/adk/agent"
 	icontext "google.golang.org/adk/internal/context"
 	"google.golang.org/adk/internal/toolinternal"
 	"google.golang.org/adk/model"
 	"google.golang.org/adk/session"
 	"google.golang.org/adk/tool"
+	"google.golang.org/adk/tool/toolconfirmation"
 )
 
 type mockFunctionTool struct {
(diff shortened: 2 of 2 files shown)


internal/llminternal/functions.go
foxfrikses, line 98: Thank you for fixing this.

I'd propose one change: could you please use `session.NewEvent` function here instead of the raw struct initialization?
This way, if the NewEvent function ever gets modified, this place will be updated with it, keeping code consistent and safe automatically.

```suggestion
event := session.NewEvent(invocationContext.InvocationID())
event.Author = invocationContext.Agent().Name()
event.Branch = invocationContext.Branch()
event.LLMResponse = model.LLMResponse{Content: content}
event.LongRunningToolIDs = longRunningToolIDs
return event
```

dpasiukevich:
Fixed in https://github.com/google/adk-go/pull/588

Thanks!

PR #564, opened 2026-02-12
Introduce Retry and Reflect pluging

caglar10ur:
This is the Go version of the Python plugin (https://github.com/google/adk-python/blob/main/src/google/adk/plugins/reflect_retry_tool_plugin.py)

### DIFF
--- a/plugin/retryandreflect/plugin.go
+++ b/plugin/retryandreflect/plugin.go
@@ -114,6 +114,15 @@ func New(opts ...PluginOption) (*plugin.Plugin, error) {
 	})
 }
 
+// MustNew creates a new reflect and retry tool plugin and panics if it fails.
+func MustNew(opts ...PluginOption) *plugin.Plugin {
+	p, err := New(opts...)
+	if err != nil {
+		panic(err)
+	}
+	return p
+}
+
 func (r *retryAndReflect) afterTool(ctx tool.Context, tool tool.Tool, args, result map[string]any, err error) (map[string]any, error) {
 	if err == nil {
 		isReflectResponse := false
--- a/plugin/functioncallmodifier/plugin.go
+++ b/plugin/functioncallmodifier/plugin.go
@@ -41,6 +41,15 @@ func NewPlugin(cfg FunctionCallModifierConfig) (*plugin.Plugin, error) {
 	})
 }
 
+// MustNewPlugin is like NewPlugin but panics if there is an error.
+func MustNewPlugin(cfg FunctionCallModifierConfig) *plugin.Plugin {
+	p, err := NewPlugin(cfg)
+	if err != nil {
+		panic(err)
+	}
+	return p
+}
+
 func beforeModelCallback(cfg FunctionCallModifierConfig) func(agent.CallbackContext, *model.LLMRequest) (*model.LLMResponse, error) {
 	return func(ctx agent.CallbackContext, req *model.LLMRequest) (*model.LLMResponse, error) {
 		if req.Config == nil || len(req.Config.Tools) == 0 {
--- a/plugin/retryandreflect/exceeded.md
+++ b/plugin/retryandreflect/exceeded.md
@@ -1,5 +1,5 @@
 
-The tool `{{.ToolName}}` has failed consecutively {self.max_retries} times and the retry limit has been exceeded.
+The tool `{{.ToolName}}` has failed consecutively {{.MaxRetries}} times and the retry limit has been exceeded.
 
 **Last Error:**
 ```


verdverm:
1. why is each line of the message a separate fmt.Printf versus using a big format const string or text/template?
2. can the message be made configurable, much like agents have InstructionProvider? (or Generator, forget off the top of my head)

plugin/retryandreflect/plugin.go
```diff
@@ -0,0 +1,226 @@
...
+type PluginOption func(*retryAndReflect)
+
+// WithMaxRetries sets the maximum number of retries for a tool.
+func WithMaxRetries(maxRetries int) PluginOption {
+	return func(r *retryAndReflect) {
+		r.maxRetries = maxRetries
+	}
+}
+
+// WithErrorIfRetryExceeded sets whether to throw an exception if the retry limit is exceeded.
```
dpasiukevich:
```suggestion
// WithErrorIfRetryExceeded sets whether to return an error if the retry limit is exceeded.
```

plugin/retryandreflect/plugin.go
```diff
@@ -0,0 +1,226 @@
...
+func WithErrorIfRetryExceeded(errorIfRetryExceeded bool) PluginOption {
```
dpasiukevich, line 80: Could you also add more docs here explaining this option?

Mentioning that if errorIfRetryExceeded == true, then the original error is returned and that's it.
But if it's errorIfRetryExceeded == false, then instead of the original error, the plugin will return a new instruction "createToolRetryExceedMsg" telling LLM to stop using this tool for the current task.

plugin/retryandreflect/plugin.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	return string(argsBytes)
+}
+
+func (r *retryAndReflect) createToolReflectionResponse(tool tool.Tool, toolArgs map[string]any, toolErr error, retryCount int) map[string]any {
+	argsSummary := r.formatToolArgs(toolArgs)
+	errorDetails := r.formatErrorDetails(toolErr)
+
+	var msg strings.Builder
+	fmt.Fprintf(&msg, "\nThe call to tool `%s` failed.\n\n", tool.Name())
```
dpasiukevich: +1 to the comment, could we use a big format const string or text/template for both createToolReflectionResponse and createToolRetryExceedMsg?

dpasiukevich:
> 1. why is each line of the message a separate fmt.Printf versus using a big format const string or text/template?
> 2. can the message be made configurable, much like agents have InstructionProvider? (or Generator, forget off the top of my head)

+1 to both comments from @verdverm

2nd point could be left to be done later, but if you have the possibility to add this, this would've been perfect!

caglar10ur:
> > 1. why is each line of the message a separate fmt.Printf versus using a big format const string or text/template?
> > 2. can the message be made configurable, much like agents have InstructionProvider? (or Generator, forget off the top of my head)
> 
> +1 to both comments from @verdverm
> 
> 2nd point could be left to be done later, but if you have the possibility to add this, this would've been perfect!

For the 1 - the main reason I used separate printfs way was to be able to use backticks and use the same exact prompt that adk-python uses, the multiline string does not support `s and I didn't want to create additional files for template texts.

I can make them something like this but I personally find the multiple printfs more readable.

```
	templateStr := `
The call to tool {{.ToolName}} failed.

**Error Details:**
` + "```" + `
{{.ErrorDetails}}
` + "```" + `

**Tool Arguments Used:**
` + "```" + `json
{{.ArgsSummary}}
` + "```" + `

**Reflection Guidance:**
This is retry attempt **{{.RetryCount}} of {{.MaxRetries}}**. Analyze the error and the arguments you provided. Do not repeat the exact same call. Consider the following before your next attempt:

1.  **Invalid Parameters**: Does the error suggest that one or more arguments are incorrect, badly formatted, or missing? Review the tool's schema and your arguments.
2.  **State or Preconditions**: Did a previous step fail or not produce the necessary state/resource for this tool to succeed?
3.  **Alternative Approach**: Is this the right tool for the job? Could another tool or a different sequence of steps achieve the goal?
4.  **Simplify the Task**: Can you break the problem down into smaller, simpler steps?
5.  **Wrong Function Name**: Does the error indicates the tool is not found? Please check again and only use available tools.

Formulate a new plan based on your analysis and try a corrected or different approach.
```

Let me know if you have any preferences?

For the 2 - I probably want to do that as a follow up (so thatI can start using this immediately).

plugin/retryandreflect/plugin.go
caglar10ur: Resolving this to discuss options in the main convo

verdverm:
Honestly, it's all the operators and function calls that make it unreadable. In example: are those quotes, plus signs, and ticks in the message or not? I have to do far too much mental parsing, gofmt whitespace rules are going to undo any readability whitespace introduced.

I'd personally put the content in a file and then `//go:embed` it.

My general preference is to align with Go practices over Python for the sake of keeping things visually the same between implementations. As the number of languages increase (there are now 4), it's going to become impossible anyway.

dpasiukevich:
I agree, `//go:embed`-ing the instructions would make it really neat.

Let me know if it looks good to you and it would be ok to update this, otherwise I could update it later. I don't have a strong preference here.

Thanks!

caglar10ur:
@verdverm @dpasiukevich PTAL - I moved them into templates as you suggested.

verdverm:
@caglar10ur, lgtm

Clarifying question, these templates are part of the tool response payload the agent sees on the next turn and do **not** create additional LLM calls, correct? (i.e. there are now additional fields it can pay attention to)

Are there docs on how to use this in my application? (even if for python)

caglar10ur:
@verdverm correct - the guidance is added to the response so that model can figure out what to do next. That being said using the pluging *will* cause additional LLM and tool calls. Its purpose is to provide reflection guide to the model for retry.

E.g. Without this plugin you call a tool and it returns some kind of resource exhausted error and agent will probably return the error to you. With the plugin enabled, the plugin will recommend model to understand and course correct, model will think the resource exhausted is not permanent and try calling the tool again. 

There is https://google.github.io/adk-docs/integrations/reflect-and-retry/ (but obviously only for python).

verdverm:
ok, the retry is certainly expected, I had wondered before reading closer if those were prompts to some "micro" agent that keeps the retry and extra events out of the main session history and main agent context, which would be cool too

Issue #536, opened 2026-02-05
REPL Empty Responses in Containerized Environments


borisdali:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

## 🔴 Required Information
*Please ensure all items in this section are completed to allow for efficient
triaging. If an item is not applicable to you - please mark it as N/A*

**Describe the Bug:**

When running an ADK-based orchestrator that uses `full.NewLauncher()` in Docker or K8s containers, the interactive REPL displays empty responses from sub-agents. Users must press Enter to see the actual response content.

**Steps to Reproduce:**
Please provide a numbered list of steps to reproduce the behavior:

1. Create an ADK orchestrator using `full.NewLauncher()` that calls sub-agents via A2A protocol
2. Build and deploy in a Docker container or a K8s pod
3. Exec into the container and run the orchestrator interactively
4. Submit a query that triggers a sub-agent call
5. Observe that the response area is empty until Enter is pressed

```bash
# Example reproduction with aiHelpDesk: https://github.com/borisdali/helpdesk
docker compose up -d
docker compose --profile interactive run orchestrator

# In the orchestrator REPL:
User -> what databases are you aware of?
# Response appears empty, pressing Enter reveals the actual response
```

**Expected Behavior:**

Agent responses should stream and display immediately as they are received, similar to running the same orchestrator locally on a host/VM (outside of containers).

**Observed Behavior:**
What actually happened? Include error messages or crash stack traces here.

- Response area remains empty after sub-agent completes
- Pressing Enter (or any key) causes the full response to appear
- This happens consistently for all sub-agent calls
- Local (non-containerized) execution works correctly

**Environment Details:**

- ADK version: Latest (commit 39c4210)
- Go version: 1.24+
- Container runtime: Docker 27.x / containerd (K8s)
- Tested on: Linux containers (amd64, arm64)

**Model Information:**
 - Which model is being used: (e.g., gemini-2.5-pro)

claude-haiku-4-5-20251001
---

## 🟡 Optional Information
*Providing this information greatly speeds up the resolution process.*

The issue appears to be related to how `full.NewLauncher()` handles:
1. TTY detection in containerized environments
2. Streaming response buffering from sub-agents
3. Terminal output flushing

The streaming fix in commit `20894cb` (Jan 22, 2026) addressed some streaming issues but did not resolve this specific container/TTY interaction.

**Regression:**
[...]

PR #568, opened 2026-02-16
fix output not rendering in non-TTY environments (#536)

benzaidfoued:
Fixes an issue where agent output does not render immediately in certain non-interactive environments (e.g. docker compose run, kubectl exec) and only appears after pressing Enter.

This change makes console streaming behavior TTY-aware and ensures responses are newline-terminated to avoid PTY buffering/rendering issues.

Fixes : https://github.com/google/adk-go/issues/536

### DIFF
--- a/internal/llminternal/converters/converters.go
+++ b/internal/llminternal/converters/converters.go
@@ -33,6 +33,7 @@ func Genai2LLMResponse(res *genai.GenerateContentResponse) *model.LLMResponse {
 				AvgLogprobs:       candidate.AvgLogprobs,
 				LogprobsResult:    candidate.LogprobsResult,
 				UsageMetadata:     usageMetadata,
+				ModelVersion:      res.ModelVersion,
 			}
 		}
 		return &model.LLMResponse{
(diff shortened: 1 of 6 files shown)


cmd/launcher/console/console.go
```diff
@@ -126,6 +127,8 @@ func (l *consoleLauncher) Run(ctx context.Context, config *launcher.Config) erro
 			inputChan <- userInput
 		}
 	}()
+	// Ensure final output is newline-terminated so PTY/exec setups flush without extra input.
+	fmt.Println()
```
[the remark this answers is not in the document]
benzaidfoued: > medium
> 
> The comment "Ensure final output is newline-terminated" appears to conflict with the code's placement. This `fmt.Println()` executes once at the beginning of the interactive session, not at the end, which could be confusing for future maintainers.
> 
> If this is intended to resolve an initial buffering issue in PTY environments, consider updating the comment to clarify its purpose, for example: `// Print an initial newline to work around PTY buffering issues.`
> 
> If the goal is to ensure a newline is printed upon program termination to flush output, using a `defer fmt.Println()` at the beginning of the `Run` function would be a more idiomatic approach.

Updated

karolpiotrowicz:
@benzaidfoued - can you resolve conflicts, and check the failing linter?

benzaidfoued:
> @benzaidfoued - can you resolve conflicts, and check the failing linter?

Checking...

benzaidfoued:
> @benzaidfoued - can you resolve conflicts, and check the failing linter?

All looks good

cmd/launcher/console/console.go
```diff
@@ -144,7 +146,13 @@ func (l *consoleLauncher) Run(ctx context.Context, config *launcher.Config) erro
...
 			if streamingMode == "" {
-				streamingMode = agent.StreamingModeSSE
+				// Stdlib-only terminal heuristic: stdout is a character device.
+				// Avoids adding golang.org/x/term dependency (golangci-lint failed to load its export data in CI).
+				if fi, err := os.Stdout.Stat(); err == nil && (fi.Mode()&os.ModeCharDevice) != 0 {
+					streamingMode = agent.StreamingModeSSE
+				} else {
+					streamingMode = agent.StreamingModeNone
+				}
 			}
```
dpasiukevich, line 162: Would it make sense to move this logic out of the for user loop? So that streamingMode var is initialized once?

benzaidfoued: > Would it make sense to move this logic out of the for user loop? So that streamingMode var is initialized once?

Let me try something

dpasiukevich:
Had some fun with the tests/linter.

But I've improved the CI along the way.

PR #570, opened 2026-02-16
feature: implement v2 trace debug endpoints based on semconv spans and logs

pawel-maciejczek:
This PR implements the V2 debugging endpoints required by adk webui.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,11 +18,11 @@ require (
 	github.com/modelcontextprotocol/go-sdk v0.7.0
 	github.com/spf13/cobra v1.8.1
 	go.opentelemetry.io/contrib/detectors/gcp v1.38.0
-	go.opentelemetry.io/otel v1.39.0
+	go.opentelemetry.io/otel v1.40.0
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp v1.39.0
-	go.opentelemetry.io/otel/sdk v1.39.0
-	go.opentelemetry.io/otel/sdk/log v0.16.0
-	go.opentelemetry.io/otel/trace v1.39.0
+	go.opentelemetry.io/otel/log v0.16.0
+	go.opentelemetry.io/otel/sdk v1.40.0
+	go.opentelemetry.io/otel/trace v1.40.0
 	golang.org/x/oauth2 v0.32.0
 	golang.org/x/sync v0.18.0
 	google.golang.org/api v0.252.0
@@ -74,9 +74,9 @@ require (
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.63.0 // indirect
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.63.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.39.0 // indirect
-	go.opentelemetry.io/otel/log v0.16.0 // indirect
-	go.opentelemetry.io/otel/metric v1.39.0 // indirect
-	go.opentelemetry.io/otel/sdk/metric v1.39.0 // indirect
+	go.opentelemetry.io/otel/metric v1.40.0 // indirect
+	go.opentelemetry.io/otel/sdk/log v0.16.0
+	go.opentelemetry.io/otel/sdk/metric v1.40.0 // indirect
 	go.opentelemetry.io/proto/otlp v1.9.0 // indirect
 	golang.org/x/crypto v0.45.0 // indirect
 	golang.org/x/net v0.47.0 // indirect


telemetry/setup_otel.go
```diff
@@ -199,6 +205,21 @@ func initTracerProvider(cfg *config) (*sdktrace.TracerProvider, error) {
 	return tp, nil
 }
 
+func initLoggerProvider(cfg *config) (*sdklog.LoggerProvider, error) {
```
dpasiukevich: nit: the internals of this func don't return error anywhere, so we can just have *sdklog.LoggerProvider. And  given that it's the internal func we can add an error later if needed.

```suggestion
func initLoggerProvider(cfg *config) *sdklog.LoggerProvider {
```

server/adkrest/controllers/debug.go
```diff
@@ -31,33 +31,56 @@ import (
...
 		http.Error(rw, "event_id parameter is required", http.StatusBadRequest)
 		return
 	}
-	traceDict := c.spansExporter.GetTraceDict()
-	eventDict, ok := traceDict[eventID]
-	if !ok {
-		http.Error(rw, fmt.Sprintf("event not found: %s", eventID), http.StatusNotFound)
+	spans := c.debugTelemetry.GetSpansByEventID(eventID)
+	for _, span := range spans {
+		opName := span.Attributes["genai.operation.name"]
```
RKest: This is an incorrect attribute key for operation name

server/adkrest/controllers/debug.go
```diff
@@ -31,33 +31,56 @@ import (
...
-	EncodeJSONResponse(eventDict, http.StatusOK, rw)
+	spans := c.debugTelemetry.GetSpansBySessionID(sessionID)
+	result := Result{
+		SchemaVersion: 2,
+		Spans:         spans,
+	}
+	EncodeJSONResponse(result, http.StatusOK, rw)
+}
+
+type Result struct {
```
RKest: nit: `Result` is a bit of a vague name. How about `SessionTelemetry` to match the type in ADK Web

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -0,0 +1,224 @@
...
+			ObservedTimestamp: l.ObservedTimestamp().Format(time.RFC3339),
+			TraceID:           l.TraceID().String(),
+			SpanID:            l.SpanID().String(),
+			EventName:         l.EventName(),
+		})
+	}
+	return spanLogs
+}
+
+func convertLogValue(v log.Value) any {
```
RKest: I think I remember seeing the same function in a different PR

Any chance we can refactor it out?

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+}
+
+type SpanContext struct {
+	TraceID string `json:"trace_id"`
+	SpanID  string `json:"span_id"`
+}
+
+// Span represents a span in the trace.
+type DebugSpan struct {
```
RKest, line 113: It should be possible to not have to define a custom span type

OpenTelemtry defines how spans should be marshaled:
https://github.com/open-telemetry/opentelemetry-go/blob/main/trace/trace.go

(hopefully it's not behind some private interface though)

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -0,0 +1,224 @@
...
+type Body any
+
+// InMemoryLogExporter stores logs in memory for debug telemetry.
+type InMemoryLogExporter struct {
+	mu      sync.Mutex
+	records []sdklog.Record
+}
+
+// Export implements sdklog.Exporter.
+func (e *InMemoryLogExporter) Export(ctx context.Context, records []sdklog.Record) error {
```
RKest, line 141: Can we use the exporter from:

https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/tracetest/exporter.go ?

Same for logs

server/adkrest/controllers/debug.go
```diff
@@ -31,33 +31,56 @@ import (
...
 		return
 	}
-	traceDict := c.spansExporter.GetTraceDict()
-	eventDict, ok := traceDict[eventID]
-	if !ok {
-		http.Error(rw, fmt.Sprintf("event not found: %s", eventID), http.StatusNotFound)
+	spans := c.debugTelemetry.GetSpansByEventID(eventID)
+	for _, span := range spans {
+		opName := span.Attributes["genai.operation.name"]
+		if opName == "execute_tool" || opName == "generate_content" {
```
dpasiukevich: Just to learn:
1. Why do we filter the spans by the `genai.operation.name` instead of returning all spans by event_id?
2. Why do we only allow "execute_tool" or "generate_content"? Could there be different operation names and why we exclude them?

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -0,0 +1,224 @@
...
+
+// DebugTelemetry stores the in memory spans and logs, grouped by session and event.
+type DebugTelemetry struct {
+	spanExporter *tracetest.InMemoryExporter
+	logExporter  *InMemoryLogExporter
+}
+
+// NewDebugTelemetry returns a DebugTelemetry instance
+func NewDebugTelemetry() *DebugTelemetry {
+	spanExporter := tracetest.NewInMemoryExporter()
```
dpasiukevich, line 37: Is it ok that we use tracetest package for the non-test code?

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -0,0 +1,224 @@
...
+	Body Body `json:"body"`
+	// RFC 3339 format timestamp  e.g. "2025-12-02T09:45:36.115239Z"
+	ObservedTimestamp string `json:"observed_timestamp"`
+	// base16 0x + 32 characters  e.g. "0x6bd725d0f21eb3117ae8cfaa709694b1"
+	TraceID   string `json:"trace_id"`
+	SpanID    string `json:"span_id"`
+	EventName string `json:"event_name"`
+}
+
+type Body any
```
dpasiukevich: Do we need this type? Can we define:

```
Body any `json:"body"`
```

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -0,0 +1,224 @@
...
+		return attrs[sessionIDKey] == sessionID
+	})
+}
+
+func (d *DebugTelemetry) getSpansFilterByAttrs(filter func(span tracetest.SpanStub, attrs map[string]string) bool) []DebugSpan {
+	var debugSpans []DebugSpan
+	spans := d.spanExporter.GetSpans()
+
+	logsBySpanID := make(map[string][]sdklog.Record)
+	for _, log := range d.logExporter.GetRecords() {
```
dpasiukevich: WDYT if we store the logs in the logExporter not as a slice but as a `map[span_id][]logs`?

Then we won't need to group them up on every `getSpansFilterByAttrs` call.

server/adkrest/internal/services/debugtelemetry.go
dpasiukevich: Can we use log.Value.String() method instead?

It should produce the same output I'd assume.

pawel-maciejczek: That was the other way around (any -> log.Value) and it's in pending #552 
It may be a good idea to move both to a helper and test them in isolation.

server/adkrest/controllers/debug.go
pawel-maciejczek: Fixed and added tests.

server/adkrest/controllers/debug.go
pawel-maciejczek: I think this is used in adk webui to show the debug details for given eventID and the rationale is that a single event will have either one "execute_tool" span or "generate_content", but not both.
@RKest has more details.

The current state looks to me like mixing frontend and backend logic in one place, which I don't really like.

I think that returning the list of all spans with matching event would be more flexible and easier to maintain:
1. webui can filter those events to populate the current widgets which expect a single span
2. The filtering logic would be kept in a single place - webui vs implementing it in every ADK backend
3. It would be easy to add new debugging widgets presenting all spans for a given id - the endpoint is just there.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: I initially implemented my own span processor and I stored traces in maps for easier access. then switched it to `tracetest.NewInMemoryExporter` and filtering traces on every request after talking with @RKest about python implementation.

I'd like to get back to the previous version (preferably in new PR), becase:
1. `tracetest.NewInMemoryExporter` is a very simple implementation with few lines of code, 
2. `tracetest.NewInMemoryExporter` keeps all spans in an unbounded slice, which will only grow as the time passes. It's ok for testing, but not great to use in server, even if it's just a debug endpoint.
3. We do the span filtering and conversion on every request. In my initial version it was done only once.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: Done.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: I'd like to keep the custom type, beacuse:
1. if we just return otel spans from our endpoint, any changes to otel libraries will affect our output. Having a custom struct protects that from those internal changes in the lib.
2. Go and python have different internal representation of spans (e.g. `parent` in go is an struct of type `SpanContext`, AFAIK in python it's just a string) and logs are automatically marshalled as empty object `{}`. Example from local run:
```
{
    "schema_version": 2,
    "spans": [
        {
            "Name": "invoke_agent root_agent",
            "SpanContext": {
                "TraceID": "31071c940c489cc809e5eb0daf3f3b5f",
                "SpanID": "7cecb5540c243244",
                "TraceFlags": "01",
                "TraceState": "",
                "Remote": false
            },
            "Parent": {
                "TraceID": "00000000000000000000000000000000",
                "SpanID": "0000000000000000",
                "TraceFlags": "00",
                "TraceState": "",
                "Remote": false
            },
            "SpanKind": 1,
            "StartTime": "2026-02-18T09:38:56.550656502Z",
            "EndTime": "2026-02-18T09:38:58.211456756Z",
            "Attributes": [
                {
                    "Key": "gen_ai.operation.name",
                    "Value": {
                        "Type": "STRING",
                        "Value": "invoke_agent"
                    }
                },
                {
                    "Key": "gen_ai.agent.description",
                    "Value": {
                        "Type": "STRING",
                        "Value": "You can do a google search and generate poems."
                    }
                },
                {
                    "Key": "gen_ai.agent.name",
                    "Value": {
                        "Type": "STRING",
                        "Value": "root_agent"
                    }
                },
                {
                    "Key": "gen_ai.conversation.id",
                    "Value": {
                        "Type": "STRING",
                        "Value": "7479baeb-ffa3-4596-a9bc-b000ba1b5685"
                    }
                }
            ],
            "Events": null,
            "Links": null,
            "Status": {
                "Code": "Unset",
                "Description": ""
            },
            "DroppedAttributes": 0,
            "DroppedEvents": 0,
            "DroppedLinks": 0,
            "ChildSpanCount": 1,
            "Resource": [
                {
                    "Key": "gcp.project_id",
                    "Value": {
                        "Type": "STRING",
                        "Value": ""
                    }
                },
                {
                    "Key": "service.name",
                    "Value": {
                        "Type": "STRING",
                        "Value": "unknown_service:main"
                    }
                },
                {
                    "Key": "telemetry.sdk.language",
                    "Value": {
                        "Type": "STRING",
                        "Value": "go"
                    }
                },
                {
                    "Key": "telemetry.sdk.name",
                    "Value": {
                        "Type": "STRING",
                        "Value": "opentelemetry"
                    }
                },
                {
                    "Key": "telemetry.sdk.version",
                    "Value": {
                        "Type": "STRING",
                        "Value": "1.40.0"
                    }
                }
            ],
            "InstrumentationScope": {
                "Name": "gcp.vertex.agent",
                "Version": "0.3.0",
                "SchemaURL": "https://opentelemetry.io/schemas/1.36.0",
                "Attributes": null
            },
            "InstrumentationLibrary": {
                "Name": "gcp.vertex.agent",
                "Version": "0.3.0",
                "SchemaURL": "https://opentelemetry.io/schemas/1.36.0",
                "Attributes": null
            },
            "logs": [
                {}
            ]
        }
    ]
}
```

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: Done.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: I already use the `tracetest.InMemoryExporter` and otel libs don't have equivalent for logs.

I think that eventually I'll ditch the `tracetest.InMemoryExporter` - https://github.com/google/adk-go/pull/570/changes#r2816497197

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: Good catch, done :)

server/adkrest/internal/services/debugtelemetry.go
dpasiukevich: +1 to keep the internal struct for stable and controlled output.

verdverm: +1 confirmation that the OTEL Go libraries have breaking changes pretty much every version

Issue #577, opened 2026-02-18
`maybeSaveOutputToState` overwrites OutputKey with empty string on function-call events


dannovikov:
**Describe the Bug:**

`maybeSaveOutputToState` in llmagent.go uses `!event.Partial` to gate writing the agent's `OutputKey` to `StateDelta`. This condition is true for **any non-partial event**, including function-call and function-response events. These events have no text content, so the method writes `""` (empty string) to `StateDelta[OutputKey]`, overwriting any value previously stored there by an upstream agent.

The Python ADK correctly uses `event.is_final_response()`, which returns `False` for function-call and function-response events, preventing the overwrite.

**Steps to Reproduce:**

1. Create a sequential pipeline with two agents sharing the same `output_key`:
   - **Agent A** (e.g., a ToolAgent) writes `"SECRET_42"` to state via `output_key`
   - **Agent B** (an LLM agent) has a tool that reads the same state key, plus the same `output_key`
2. Agent B calls its tool — this emits a function-call event (non-partial, no text)
3. `maybeSaveOutputToState` fires on the function-call event: `!event.Partial` is `true`, `event.Content.Parts` has a `FunctionCall` part but no `Text`, so `sb.String()` is `""`
4. `StateDelta[output_key] = ""` — clobbering `"SECRET_42"`
5. When the tool reads state, it gets `""` instead of `"SECRET_42"`

**Expected Behavior:**

`maybeSaveOutputToState` should only write to `StateDelta` on the agent's **final text response**, not on intermediate function-call or function-response events. This is what `event.IsFinalResponse()` guarantees — it checks `!hasFunctionCalls && !hasFunctionResponses && !Partial && !hasTrailingCodeExecutionResult`.

**Observed Behavior:**

The state key written by an upstream agent is overwritten with `""` on every function-call event, because `!event.Partial` does not exclude function-call/response events.

**Environment Details:**

 - ADK Library Version: `main` at `8250f6f`
 - OS: macOS (darwin/arm64)
 - Go Version: 1.25.1

**Model Information:**
 - Any model that uses tool calling (reproduced with Azure OpenAI gpt-4o-mini and gpt-4o)

---

**Regression:**
N/A — the condition has been `!event.Partial` since the initial implementation.

**Logs:**
```text
# Event trace showing the clobbering:
# 1. ToolAgent writes: StateDelta["shared_key"] = "SECRET_42"     ✅
# 2. LLM emits FunctionCall event (Partial=false, no text)
[...]

PR #578, opened 2026-02-18
fix: `maybeSaveOutputToState` overwrites OutputKey with empty string on function-call events

dannovikov:
Closes issue #577 

**Description:**

Any LLMAgent with `OutputKey` set that also uses tools has its output key overwritten to an empty string on every function-call and function-response event. This happens because `maybeSaveOutputToState` uses `!event.Partial` as its guard condition, which only filters out streaming chunks — function-call events pass through, their parts contain no `.Text`, so `strings.Builder` produces `""`, which overwrites `StateDelta[OutputKey]`.

Go (`agent/llmagent/llmagent.go:392`):
```go
if a.OutputKey != "" && !event.Partial && event.Content != nil && len(event.Content.Parts) > 0 {
```

Python (`agents/llm_agent.py:819-821`):
```python
if self.output_key and event.is_final_response() and event.content and event.content.parts:
```

The proposed change replaces `!event.Partial` with `event.IsFinalResponse()` — a method that already exists in `session.Event` and mirrors the Python behavior exactly — so that `OutputKey` is only written on the agent's actual final text response.

### Testing

Added two subtests to the existing `TestLlmAgent_MaybeSaveOutputToState` in `llmagent_saveoutput_test.go`:

- **"skips for function call events"** — `FunctionCall` part, `Partial=false` → asserts `StateDelta` stays empty
- **"skips for function response events"** — `FunctionResponse` part, `Partial=false` → asserts `StateDelta` stays empty

Both fail before the fix (`StateDelta` gets `{"result": ""}`), pass after.


dannovikov:
Hi guys, any updates? This one is important to merge, as without it, you cannot rely on output key for agents that call tools. With this PR, the bug is fixed and aligns Go with Python version.

karolpiotrowicz:
The direction here is right, and the reported bug really is fixed for the ordinary case. I checked the branch out, merged it onto current `main`, and confirmed the red-to-green: on `main` an `OutputKey` agent that calls a tool writes `""` into session state on the function-call and function-response events, and with this change it no longer does.

What I would want resolved before it lands is that the same symptom survives the fix on a path that ships in the repo, so the bug is only partly closed.

[`Event.IsFinalResponse()`](https://github.com/google/adk-go/blob/70e2f03c1f2c221018c0d6a22ee8e48559dd7523/session/session.go#L216-L222) returns `true` on its first branch without ever looking at the parts:

```go
func (e *Event) IsFinalResponse() bool {
	if e.Actions.SkipSummarization || len(e.LongRunningToolIDs) > 0 {
		return true
	}
	return !hasFunctionCalls(&e.LLMResponse) && !hasFunctionResponses(&e.LLMResponse) && !e.LLMResponse.Partial && !hasTrailingCodeExecutionResult(&e.LLMResponse)
}
```

So an event carrying `SkipSummarization` or a long-running tool ID passes the new guard at [`llmagent.go:406`](https://github.com/google/adk-go/blob/566588c93320b96366394299b9fe3baefd94b35a/agent/llmagent/llmagent.go#L406) whether or not it has any text. [`exitlooptool`](https://github.com/google/adk-go/blob/70e2f03c1f2c221018c0d6a22ee8e48559dd7523/tool/exitlooptool/tool.go#L28) sets `SkipSummarization` unconditionally on every call, and `tool/tool.go`, `tool/functiontool`, and `tool/mcptoolset` all set it on their confirmation paths. In those cases the text builder still yields `""` and the stored answer is still destroyed.

Reproduction, run against this branch merged onto `70e2f03`:

<details>
<summary><code>go test -run TestOutputKeyExitLoop ./agent/llmagent/</code></summary>

```go
package llmagent_test

import (
	"testing"

	"google.golang.org/genai"

	"google.golang.org/adk/v2/internal/testutil"
	"google.golang.org/adk/v2/tool/exitlooptool"
)

func TestOutputKeyExitLoop(t *testing.T) {
	exitTool, err := exitlooptool.New()
	if err != nil {
		t.Fatalf("exitlooptool.New: %v", err)
	}
	m := &testutil.MockModel{Responses: []*genai.Content{
		{Role: genai.RoleModel, Parts: []*genai.Part{
			{Text: "The answer is 42."},
			{FunctionCall: &genai.FunctionCall{ID: "fc-1", Name: "exit_loop", Args: map[string]any{}}},
		}},
	}}
		Name: "agent_recheck", Model: m, OutputKey: "result",
		Tools:                    []tool.Tool{exitTool},
		DisallowTransferToParent: true, DisallowTransferToPeers: true,
	})
	if err != nil {
		t.Fatalf("llmagent.New: %v", err)
	}
	cr, err := svc.Create(t.Context(), &session.CreateRequest{AppName: "app", UserID: "u", SessionID: "s"})
	if err != nil {
		t.Fatalf("Create: %v", err)
	}
	r, err := runner.New(runner.Config{AppName: "app", Agent: a, SessionService: svc})
	if err != nil {
		t.Fatalf("runner.New: %v", err)
	}
	for ev, err := range r.Run(t.Context(), "u", cr.Session.ID(), genai.NewContentFromText("hi", genai.RoleUser), agent.RunConfig{}) {
		if err != nil {
			t.Fatalf("run: %v", err)
		}
		t.Logf("author=%s partial=%v skipSumm=%v final=%v delta=%v",
			ev.Author, ev.Partial, ev.Actions.SkipSummarization, ev.IsFinalResponse(), ev.Actions.StateDelta)
	}
	got, err := svc.Get(t.Context(), &session.GetRequest{AppName: "app", UserID: "u", SessionID: cr.Session.ID()})
	if err != nil {
		t.Fatalf("Get: %v", err)
	}
	v, _ := got.Session.State().Get("result")
	t.Logf(">>> FINAL SESSION STATE result = %#v", v)
	if s, _ := v.(string); s != "The answer is 42." {
		t.Errorf("result = %q, want %q", s, "The answer is 42.")
	}
}
```

</details>

```
author=agent_recheck partial=false skipSumm=false final=false delta=map[]
author=agent_recheck partial=false skipSumm=true  final=true  delta=map[result:]
>>> FINAL SESSION STATE result = ""
```

That output is identical before and after the change.

adk-python ran into the same thing. It uses `is_final_response()` and then adds a separate text-presence check on top, in a commit titled "fix: skipping state_delta overwrite on function_response-only events": [`llm_agent.py#L1025-L1031`](https://github.com/google/adk-python/blob/dc00db51bc5dc195a21633b295aa9eac33732ccf/src/google/adk/agents/llm_agent.py#L1025-L1031).

One thing worth knowing before reaching for that guard, because I tried it: porting `has_text_part` on its own stops the destruction but does not make the value correct. The result goes from `""` to absent. In this scenario the text-bearing event carries a function call alongside the text, so `IsFinalResponse()` is false for that event too and the text is never written in the first place. adk-python covers that separately with `__maybe_accumulate_streaming_output`, which has no counterpart here. Whether to port the accumulator is a larger question than this pull request, but the guard alone will not make an exit-loop agent's `OutputKey` come out right, and it seemed better to say so than to let you find it after the fact.

Two smaller points.

The change widens the guard in one direction as well as narrowing it. A partial streaming event that carries a long-running tool call now satisfies `IsFinalResponse()` and gets written, where `!event.Partial` rejected it. Partial events are not committed to the session, so the end state was unaffected in everything I ran, but adding `!event.Partial` alongside the new call would keep the condition a strict subset of the old one.

Both new subtests leave `SkipSummarization` false and `LongRunningToolIDs` empty, which is exactly the combination where the short-circuit does not fire. A case with `SkipSummarization: true` over a pre-populated `StateDelta` would pin down the behavior that is actually at risk.

The branch is 160 commits behind `main`, though it merges cleanly and the full suite is unchanged either way.

karolpiotrowicz:
@dannovikov will you be able to take a look?

jjsasha63:
Heads up that I have approved #1061, which makes the same `IsFinalResponse()` swap you have here plus one extra guard, so this is superseded. Your diagnosis in #577 was right and it got here four months earlier — thank you for both.

The extra bit is a `hasTextPart` check before the write. `IsFinalResponse()` short-circuits to true when `Actions.SkipSummarization` is set or `LongRunningToolIDs` is non-empty (session/session.go:226), and both of those arrive as function-response-only events with no text, so the swap on its own still writes `""` over the key in those cases. `SkipSummarization` is set in exitlooptool, agenttool, functiontool, mcptoolset and tool confirmation, so it comes up often. adk-python carries the same guard for the same reason.

Nothing is wrong with the change here, it is just the smaller half.

PR #579, opened 2026-02-18
Introduce tool.WithConfirmation. 

caglar10ur:
It wraps a toolset to inject confirmation logic in each tool after they created.

### DIFF
--- a/tool/tool.go
+++ b/tool/tool.go
@@ -159,21 +159,23 @@ func (f *filteredToolset) Tools(ctx agent.ReadonlyContext) ([]Tool, error) {
 	return filtered, nil
 }
 
-// EXPERIMENTAL: ConfirmationProvider is experimental and not currently in scope for the v1.0 API.
 // ConfirmationProvider defines a function that dynamically determines whether
 // a specific tool execution requires user confirmation.
 //
 // It accepts the tool name and the input parameters as arguments.
 // Returning true signals that the system must wait for Human-in-the-Loop (HITL)
 // approval before proceeding with the execution.
+//
+// EXPERIMENTAL: ConfirmationProvider is experimental and not currently in scope for the v1.0 API.
 type ConfirmationProvider func(toolName string, toolInput any) bool
 
-// EXPERIMENTAL: WithConfirmation is experimental and not currently in scope for the v1.0 API.
 // WithConfirmation wraps a toolset to inject confirmation logic in each tool.
 // Only tools within the provided Toolset that implement the `runnableTool` interface
 // (i.e., provide a FunctionDeclaration and a Run method) will be wrapped with
 // confirmation logic. Other tools will be included in the returned Toolset without
 // modification.
+//
+// EXPERIMENTAL: WithConfirmation is experimental and not currently in scope for the v1.0 API.
 func WithConfirmation(ts Toolset, requireConfirmation bool, requireConfirmationProvider ConfirmationProvider) Toolset {
 	return &confirmationToolset{
 		toolset:                     ts,


tool/tool.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
```
hyangah: Please don't change the year. go/copyright#the-year

tool/tool.go
```diff
@@ -153,3 +158,105 @@ func (f *filteredToolset) Tools(ctx agent.ReadonlyContext) ([]Tool, error) {
...
+// Returning true signals that the system must wait for Human-in-the-Loop (HITL)
+// approval before proceeding with the execution.
+type ConfirmationProvider func(toolName string, toolInput any) bool
+
+// WithConfirmation wraps a toolset to inject confirmation logic in each tool.
+// Only tools within the provided Toolset that implement the `runnableTool` interface
+// (i.e., provide a FunctionDeclaration and a Run method) will be wrapped with
+// confirmation logic. Other tools will be included in the returned Toolset without
+// modification.
+func WithConfirmation(ts Toolset, requireConfirmation bool, requireConfirmationProvider ConfirmationProvider) Toolset {
```
hyangah, line 179: We want to get this merged to unblock the current requirements, but since these new APIs (specifically `WithConfirmation` and the `ConfirmationProvider` type) are not yet within the scope of the v1.0 API and deviate from other language ADK implementations, we need to mark them clearly as experimental.

Could you please add an // EXPERIMENTAL tag to the documentation for these additions? This will signal that these are subject to change or removal following further discussion with other teams.

```
// EXPERIMENTAL: WithConfirmation is experimental and not currently in scope for the v1.0 API.
func WithConfirmation(...)
```

tool/tool.go
```diff
@@ -153,3 +158,105 @@ func (f *filteredToolset) Tools(ctx agent.ReadonlyContext) ([]Tool, error) {
...
+}
+
+type runnableTool interface {
+	Tool
+	Declaration() *genai.FunctionDeclaration
+	Run(ctx Context, args any) (result map[string]any, err error)
+}
+
+func (t *confirmationTool) Declaration() *genai.FunctionDeclaration {
+	return t.Tool.(runnableTool).Declaration()
```
hyangah: I think it makes more sense for the confirmationTool to have `runnableTool` instead of `Tool` as its field, WDYT?

caglar10ur:
@hyangah thanks for the review - I believe I addressed them PTAL

tool/tool.go
```diff
@@ -153,3 +158,107 @@ func (f *filteredToolset) Tools(ctx agent.ReadonlyContext) ([]Tool, error) {
 	}
 	return filtered, nil
 }
+
+// EXPERIMENTAL: ConfirmationProvider is experimental and not currently in scope for the v1.0 API.
```
hyangah: Move this to the last paragraph of the comment. The Go comment must start with the type name.

tool/tool_test.go
```diff
@@ -103,3 +111,169 @@ func TestTypes(t *testing.T) {
...
+		provider                  tool.ConfirmationProvider
+		toolConfirmation          *toolconfirmation.ToolConfirmation
+		wantConfirmationRequested bool
+		wantSkipSummarization     bool
+		wantErr                   bool
+		wantErrMsg                string
+		wantToolRan               bool
+	}{
+		{
+			name:                      "confirmation required, no confirmation in context",
```
hyangah: https://google.github.io/styleguide/go/decisions#subtest-names

Subtest name serves for test filter (`-run`), so short names with printable characters are recommended.

hyangah (approved):

PR #581, opened 2026-02-19
chore: add interface assertion for readability

zchee:
Add interface assertion for readability.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -145,6 +145,8 @@ type agent struct {
 	afterAgentCallbacks  []AfterAgentCallback
 }
 
+var _ Agent = (*agent)(nil)
+
 func (a *agent) Name() string {
 	return a.name
 }
@@ -440,6 +442,8 @@ type invocationContext struct {
 	endInvocation bool
 }
 
+var _ InvocationContext = (*invocationContext)(nil)
+
 func (c *invocationContext) Agent() Agent {
 	return c.agent
 }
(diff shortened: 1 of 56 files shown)


zchee:
Fixed Gemini reviewed `This interface assertion is redundant as it already exists ...`. PTAL.

zchee:
@dpasiukevich gentle ping

jjsasha63:
Thanks for this, and sorry — you pinged twice and we never answered. That is on us.

The codebase has moved a long way since February. The module is now `google.golang.org/adk/v2`, `toolutils` moved out of `internal/toolinternal`, `tool.Context` became `agent.Context` in `Run` signatures, and `memory.Service.AddSession` was renamed to `AddSessionToMemory`. Merging this today gives 17 conflicts across 56 files.

The interface names you corrected are still right — `toolinternal.FunctionTool` and `toolinternal.RequestProcessor` are still where you pointed them. But the surrounding code moved enough that this would be a rewrite rather than a rebase, so we are going to close it rather than ask you to redo 56 files.

Two things worth keeping out of it:

- The `// ProcessRequest implements types.Tool.` and `// Run implements types.Tool.` comments in `internal/llminternal/agent_transfer.go` are still on main, and `types` is still not a package. We will fix those separately.
- There is no issue tracking the assertion convention itself, which is part of why this sat for so long. If you still want it applied across the codebase, open one and we will agree the scope up front, then a narrow sweep can land against current main.

For reference, #750 covers the tools half of this (assertions only, no comment fixes) and is in the same position — 140 commits behind and conflicting.

PR #582, opened 2026-02-19
fix: enable OutputSchema processor for non-Gemini models

GuyGoldenberg:
Fixes OutputSchema not working for non-Gemini models (e.g., Bedrock/Claude).

The set_model_response tool workaround was only activating for Gemini API models, leaving non-Gemini models without structured output support when tools are present.

**Test plan:** Added unit tests, all pass.

### DIFF
--- a/internal/llminternal/googlellm/variant.go
+++ b/internal/llminternal/googlellm/variant.go
@@ -66,13 +66,19 @@ func IsGeminiAPIVariant(llm model.LLM) bool {
 	return GetGoogleLLMVariant(llm) == genai.BackendGeminiAPI
 }
 
-// NeedsOutputSchemaProcessor returns true if the Gemini model doesn't support output schema with tools natively and requires a processor to handle it.
-// Only Gemini 2.5 models and lower and only in Gemini API don't support natively, so we enable the processor for them.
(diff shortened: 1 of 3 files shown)


wolo-lab:
Thank you for looking at this, and sorry it sat without a human review.

We are not taking the change, because non-Gemini structured output already works through the provider's own mechanism. When the processor is not needed, main sets ResponseSchema and ResponseMIMEType directly: https://github.com/google/adk-go/blob/245ba284ed51f0f58355a6233f0b1ae01f252160/internal/llminternal/basic_processor.go#L53-L55

openaimodel converts that into OpenAI's response_format: json_schema. Applying the set_model_response workaround to those models would replace an enforced schema with a prompt asking the model to please call a tool. adk-python makes the same call deliberately — can_use_output_schema_with_tools returns true for LiteLLM models, on the grounds that native support is more reliable than the prompt-based workaround.

Your underlying observation is correct though, and we want to keep it. The decision is made by prefix-matching the model name, so a third-party model.LLM implementation has no way to declare what it supports. adk-python has since moved to a capabilities field for exactly this. We will open an issue for the Go equivalent and credit you in it.

We are reviewing incoming PRs far more actively now.

PR #585, opened 2026-02-20
feat: implements identityRequestProcessor to internal/llminternal

zchee:
Implements `identityRequestProcessor` corresponding to adk-python behaviour. Also recreated the affected httprr.


zchee:
@dpasiukevich This PR is still a draft, but needs to be squash?

PR #591, opened 2026-02-23
fix: use session.NewEvent in generateRequestConfirmationEvent

### LINKED ISSUE #590: generateRequestConfirmationEvent creates events with missing ID
## 🔴 Required Information

**Describe the Bug:**
`generateRequestConfirmationEvent` in `internal/llminternal/functions.go` constructs an `Event` using a raw struct literal instead of `session.NewEvent()`. This results in events with an empty `ID` (no UUID) and an uninitialized `StateDelta` (nil map).

**Steps to Reproduce:**
1. Trigger a tool confirmation flow via `ctx.RequestConfirmation()`
2. The confirmation event created by `generateRequestConfirmationEvent` has an empty `ID` field

**Expected Behavior:**
The event should have a UUID assigned to its `ID` field, consistent with all other events created via `session.NewEvent()`.

**Observed Behavior:**
The event `ID` is an empty string. `StateDelta` is nil instead of an initialized empty map.

The root cause is in `internal/llminternal/functions.go:81`:
```go
return &session.Event{
    InvocationID: invocationContext.InvocationID(),
    Author:       invocationContext.Agent().Name(),
    // ID is NOT set!
}
```

While `session.NewEvent()` properly initializes all fields:
```go
func NewEvent(invocationID string) *Event {
    return &Event{
        ID:           uuid.NewString(),
        InvocationID: invocationID,
        Timestamp:    time.Now(),
        Actions:      EventActions{StateDelta: make(map[string]any)},
    }
}
```

**Environment Details:**
 - ADK Library Version: v0.5.0
 - OS: N/A (code-level bug)
 - Go Version: N/A

**Model Information:**
 - N/A

---

## 🟡 Optional Information

 - Always (100%)

mohamedabozeid:
Fixes #590

## Description

`generateRequestConfirmationEvent` in `internal/llminternal/functions.go` constructs an `Event` using a raw struct literal, bypassing `session.NewEvent()`. This results in:

1. **Missing `ID`** — empty string instead of a UUID. Any downstream logic relying on event IDs for deduplication, ordering, tracing, or correlation will break.
2. **Uninitialized `StateDelta`** — nil map instead of `map[string]any{}`, which could cause nil map panics if anything writes to it.

### Root Cause

The function was written as a raw struct literal `&session.Event{...}` instead of calling `session.NewEvent()`. The `NewEvent` constructor sets:
- `ID: uuid.NewString()`
- `Timestamp: time.Now()`
- `Actions: EventActions{StateDelta: make(map[string]any)}`

The raw literal only set `Timestamp` and left `ID` empty and `StateDelta` nil.

### Fix

Replace the raw struct literal with `session.NewEvent()` and set the remaining fields (`Author`, `Branch`, `LLMResponse`, `LongRunningToolIDs`) on the returned object. This is consistent with how events are created elsewhere in the codebase.

### Unit Tests

Updated tests in `internal/llminternal/functions_test.go`:

- Added assertions verifying that the returned event has a non-empty `ID` and a non-zero `Timestamp`.
- Updated existing test expectation to match the properly initialized `StateDelta` from `NewEvent`.

### Manual E2E Test

Verified with `go test ./internal/llminternal/...` — all tests pass with no regressions.

### DIFF
--- a/internal/llminternal/functions.go
+++ b/internal/llminternal/functions.go
@@ -15,8 +15,6 @@
 package llminternal
 
 import (
-	"time"
-
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/agent"
@@ -78,18 +76,15 @@ func generateRequestConfirmationEvent(
 		return nil
 	}
 
-	return &session.Event{
-		InvocationID: invocationContext.InvocationID(),
-		Author:       invocationContext.Agent().Name(),
-		Branch:       invocationContext.Branch(),
-		LLMResponse: model.LLMResponse{
-			Content: &genai.Content{
-				Parts: parts,
-				Role:  genai.RoleModel,
-			},
+	event := session.NewEvent(invocationContext.InvocationID())
+	event.Author = invocationContext.Agent().Name()
+	event.Branch = invocationContext.Branch()
+	event.LLMResponse = model.LLMResponse{
+		Content: &genai.Content{
+			Parts: parts,
+			Role:  genai.RoleModel,
 		},
-		Timestamp:          time.Now(),
-		LongRunningToolIDs: longRunningToolIDs,
-		Actions:            session.EventActions{},
 	}
+	event.LongRunningToolIDs = longRunningToolIDs
+	return event
 }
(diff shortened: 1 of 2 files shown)


mohamedabozeid:
duplicate to https://github.com/google/adk-go/pull/588

PR #593, opened 2026-02-23
feat: refactor debugtelemetry - optimize trace access

pawel-maciejczek:
This PR

* refactors the signals storage to stop using test otel packages
* refactors the signals storage for efficient retrieval without iterating over all spans on each get request

We can add TTL to clean up old traces in following PR.

### DIFF
--- a/server/adkrest/internal/services/debugtelemetry.go
+++ b/server/adkrest/internal/services/debugtelemetry.go
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -29,6 +29,8 @@ import (
 	"google.golang.org/adk/internal/telemetry"
 )
 
+const eventIDKey = "gcp.vertex.agent.event_id"
+
 // DebugTelemetry stores the in memory spans and logs, grouped by session and event.
 type DebugTelemetry struct {
 	store *spanStore
@@ -51,31 +53,12 @@ func (d *DebugTelemetry) LogProcessor() sdklog.Processor {
 
 // GetSpansByEventID returns spans associated with the given event ID.
 func (d *DebugTelemetry) GetSpansByEventID(eventID string) []DebugSpan {
-	records := d.store.getByEventID(eventID)
-	return convertRecords(records)
+	return d.store.getSpansByEventID(eventID)
 }
 
 // GetSpansBySessionID returns spans associated with the given session ID.
 func (d *DebugTelemetry) GetSpansBySessionID(sessionID string) []DebugSpan {
-	records := d.store.getBySessionID(sessionID)
-	return convertRecords(records)
-}
-
-func convertRecords(records []*spanRecord) []DebugSpan {
-	debugSpans := make([]DebugSpan, len(records))
-	for i, r := range records {
-		debugSpans[i] = DebugSpan{
-			Name:         r.Span.Name,
-			StartTime:    r.Span.StartTime.UnixNano(),
-			EndTime:      r.Span.EndTime.UnixNano(),
-			TraceID:      r.Span.Context.TraceID().String(),
-			SpanID:       r.Span.Context.SpanID().String(),
-			ParentSpanID: r.Span.ParentSpanID.String(),
-			Attributes:   r.Span.Attributes,
-			Logs:         r.Logs,
-		}
-	}
-	return debugSpans
+	return d.store.getSpansBySessionID(sessionID)
 }
 
 func convertAttrs(in []attribute.KeyValue) map[string]string {
@@ -84,12 +67,6 @@ func convertAttrs(in []attribute.KeyValue) map[string]string {
 		out[string(attr.Key)] = attr.Value.Emit()
 	}
 	return out
-}
-
-// SpanContext uniquely identifies a span.
-type SpanContext struct {
-	TraceID string `json:"trace_id"`
-	SpanID  string `json:"span_id"`
 }
 
 // DebugSpan represents a span in the trace.
@@ -151,17 +128,15 @@ func newSpanStore() *spanStore {
 	}
 }
 
-func (s *spanStore) getByEventID(id string) []*spanRecord {
+func (s *spanStore) getSpansByEventID(id string) []DebugSpan {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 	// Create a copy of the slice to avoid race conditions.
-	src := s.recordsByEventID[id]
-	dst := make([]*spanRecord, len(src))
-	copy(dst, src)
-	return filterNilsAndSort(dst)
+	records := slices.Clone(s.recordsByEventID[id])
+	return convertRecords(records)
 }
 
-func (s *spanStore) getBySessionID(sessionID string) []*spanRecord {
+func (s *spanStore) getSpansBySessionID(sessionID string) []DebugSpan {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 	traces := s.traceIDsBySessionID[sessionID]
@@ -171,7 +146,27 @@ func (s *spanStore) getBySessionID(sessionID string) []*spanRecord {
 			records = append(records, r...)
 		}
 	}
-	return filterNilsAndSort(records)
+	return convertRecords(records)
+}
+
+func convertRecords(records []*spanRecord) []DebugSpan {
+	records = filterNilsAndSort(records)
+	debugSpans := make([]DebugSpan, len(records))
+	for i, r := range records {
+		// Clone the logs to avoid race conditions.
+		logs := slices.Clone(r.Logs)
+		debugSpans[i] = DebugSpan{
+			Name:         r.Span.Name,
+			StartTime:    r.Span.StartTime.UnixNano(),
+			EndTime:      r.Span.EndTime.UnixNano(),
+			TraceID:      r.Span.Context.TraceID().String(),
+			SpanID:       r.Span.Context.SpanID().String(),
+			ParentSpanID: r.Span.ParentSpanID.String(),
+			Attributes:   r.Span.Attributes,
+			Logs:         logs,
+		}
+	}
+	return debugSpans
 }
 
 func filterNilsAndSort(records []*spanRecord) []*spanRecord {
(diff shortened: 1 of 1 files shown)


server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -150,38 +118,92 @@ type DebugSpan struct {
...
-	TraceID   string `json:"trace_id"`
-	SpanID    string `json:"span_id"`
-	EventName string `json:"event_name"`
+	TraceID           string `json:"trace_id"`
+	SpanID            string `json:"span_id"`
+	EventName         string `json:"event_name"`
+}
+
+// signalHolder stores all signals (span and logs) for a given span ID.
+type signalHolder struct {
```
dpasiukevich: I'd suggest to rename:

signalHolder -> spanRecord // spanRecord stores a span and its associated logs.
signalsHolder -> spanStore

Because right now both are easy to mix when reading the code

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -16,93 +16,72 @@ package services
...
-			debugSpans = append(debugSpans, debugSpan)
+func sortAndConvertSignals(signals []*signalHolder) []DebugSpan {
+	filtered := slices.DeleteFunc(signals, func(s *signalHolder) bool {
+		// Logs are emitted before the span is closed and sent to the processor.
+		// Skip them in the response.
+		return s == nil || s.Span == nil
+	})
+	slices.SortFunc(filtered, func(a, b *signalHolder) int {
+		return a.Span.StartTime.Compare(b.Span.StartTime)
+	})
```
dpasiukevich: There are 2 sorts: here and in lines 87-89. Seems like we can remove one.

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -16,93 +16,72 @@ package services
...
-			debugSpan.Logs = d.logExporter.GetLogsBySpanID(span.SpanContext.SpanID().String())
-			debugSpans = append(debugSpans, debugSpan)
+func sortAndConvertSignals(signals []*signalHolder) []DebugSpan {
+	filtered := slices.DeleteFunc(signals, func(s *signalHolder) bool {
+		// Logs are emitted before the span is closed and sent to the processor.
+		// Skip them in the response.
+		return s == nil || s.Span == nil
+	})
+	slices.SortFunc(filtered, func(a, b *signalHolder) int {
+		return a.Span.StartTime.Compare(b.Span.StartTime)
```
dpasiukevich: to double check -- we do a sort here without locking. e.g. ExportSpans may update the entry here concurrently while this method runs.

Is a deep-copy viable? Or r-locking there?

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: Thanks, those names are much better.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: Done.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: SpanStore getters return a new slice, which can be sorted in place, so I don't think we need any locks here:
```
func (s *spanStore) getByEventID(id string) []*spanRecord {
	s.mu.RLock()
	defer s.mu.RUnlock()
	// Create a copy of the slice to avoid race conditions.
	src := s.signalsByEventID[id]
	dst := make([]*spanRecord, len(src))
	copy(dst, src)
	return dst
}

func (s *spanStore) getBySessionID(sessionID string) []*spanRecord {
	s.mu.RLock()
	defer s.mu.RUnlock()
	traces := s.traceIDsBySessionID[sessionID]
	var signals []*spanRecord
	for traceID := range traces {
		if traceSignals, ok := s.signalsByTraceID[traceID]; ok {
			signals = append(signals, traceSignals...)
		}
	}
	return signals
}
```

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -150,38 +114,92 @@ type DebugSpan struct {
...
+	Context      trace.SpanContext
+	ParentSpanID trace.SpanID
+	Attributes   map[string]string
+}
+
+// spanStore stores spans and logs in memory for debug telemetry.
+type spanStore struct {
+	mu sync.RWMutex
+	// signalsBySpanID stores span signals indexed by span id.
+	signalsBySpanID map[string]*spanRecord
```
dpasiukevich: slight nit: maybe worth to rename signal -> record globally within a file for naming consistency:

signalsBySpanID -> recordsBySpanID
signalsByEventID -> recordsByEventID

sortAndConvertSignals -> sortAndConvertRecords.

etc

server/adkrest/internal/services/debugtelemetry.go
dpasiukevich: The problem is when ExportSpans updates the span `signals.Span = &inMemorySpan{`.

getByEventID and getBySessionID return a slice copy, but the values (pointers are the same). So it can read/write race when sort method is reading Span.StartTime and ExportSpans method updates the Span.

This could fixed by ensuring a sort is also protected by a read-lock. e.g. by moving it up to GetSpansByEventID and GetSpansBySessionID.

pawel-maciejczek: Thanks for explaining.
I moved sorting to `GetSpansByEventID` and `GetSpansBySessionID`, which is protected by RLock.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: Done - I renamed all occurrences of signals after changing the struct names.

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -150,58 +106,166 @@ type DebugSpan struct {
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	for _, span := range spans {
+		attrs := convertAttrs(span.Attributes())
+		spanID := span.SpanContext().SpanID().String()
+		record, ok := s.recordsBySpanID[spanID]
+		if !ok {
+			record = &spanRecord{}
+			s.recordsBySpanID[spanID] = record
+		}
```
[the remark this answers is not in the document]
pawel-maciejczek: Done.

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -119,24 +86,13 @@ func convertAttrs(in []attribute.KeyValue) map[string]string {
...
-		EndTime:      span.EndTime.UnixNano(),
-		TraceID:      span.SpanContext.TraceID().String(),
-		SpanID:       span.SpanContext.SpanID().String(),
-		ParentSpanID: span.Parent.SpanID().String(),
-		Attributes:   attrs,
-	}
-}
-
+// SpanContext uniquely identifies a span.
 type SpanContext struct {
```
dpasiukevich: let's remove, it's not used anywhere

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -150,58 +106,166 @@ type DebugSpan struct {
...
+		}
+
+		s.updateSpanIndexes(record.Span, record)
+	}
 	return nil
 }
 
-// Shutdown implements sdklog.Exporter.
-func (e *InMemoryLogExporter) Shutdown(ctx context.Context) error {
+func (s *spanStore) updateSpanIndexes(span *inMemorySpan, records *spanRecord) {
```
dpasiukevich:
```suggestion
func (s *spanStore) updateSpanIndexes(span *inMemorySpan, record *spanRecord) {
```

records -> record as it's only one object

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -150,58 +106,166 @@ type DebugSpan struct {
...
+			SpanID:            log.SpanID().String(),
+			EventName:         log.EventName(),
 		})
 	}
 	return nil
 }
 
-// ForceFlush implements sdklog.Exporter.
-func (e *InMemoryLogExporter) ForceFlush(ctx context.Context) error {
+// ExportSpans implements sdktrace.SpanProcessor.
```
dpasiukevich:
```suggestion
// ExportSpans implements sdktrace.SpanExporter.
```

nit: should it be spanExporter?

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -16,98 +16,65 @@ package services
...
+	for i, r := range records {
+		debugSpans[i] = DebugSpan{
+			Name:         r.Span.Name,
+			StartTime:    r.Span.StartTime.UnixNano(),
+			EndTime:      r.Span.EndTime.UnixNano(),
+			TraceID:      r.Span.Context.TraceID().String(),
+			SpanID:       r.Span.Context.SpanID().String(),
+			ParentSpanID: r.Span.ParentSpanID.String(),
+			Attributes:   r.Span.Attributes,
+			Logs:         r.Logs,
```
dpasiukevich: Let's use slices.Clone(r.Logs).

Otherwise GetSpansBySessionID will return a shared object (r.Logs) which will be read (e.g. on marshalling to return for HTTP request) and at the same time updated by Export method when it does: `record.Logs = append(record.Logs`.

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -16,98 +16,65 @@ package services
...
 }
 
-// GetSpansByEventID returns stored event traces.
+// GetSpansByEventID returns spans associated with the given event ID.
 func (d *DebugTelemetry) GetSpansByEventID(eventID string) []DebugSpan {
-	return d.getSpansFilterByAttrs(func(span tracetest.SpanStub, attrs map[string]string) bool {
-		return attrs["gcp.vertex.agent.event_id"] == eventID
-	})
+	records := d.store.getByEventID(eventID)
+	return convertRecords(records)
```
dpasiukevich: I'd still suggest to have a mutex here in GetSpansByEventID (and similar in the GetSpansBySessionID method). So that convertRecords is also guarded by mutex.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: Done.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: Done.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: Done.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: Done.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: Done - I moved the conversion logic down to `GetSpansByEventID` and `GetSpansBySessionID`, so they covered by mutex.

PR #595, opened 2026-02-23
feat: unify schema validation using jsonschema library

GuyGoldenberg:
## Summary

Unifies schema validation by using the `jsonschema` library consistently across all tool types, adding enum validation support for `AgentTool` and output schemas.

## Problem

`FunctionTool` used `jsonschema.Resolved.Validate()` which validates enums, but `AgentTool` and output schema used a custom `ValidateMapOnSchema()` that did not.

## Solution

Add a `genai.Schema` → `jsonschema.Schema` converter so all validation uses jsonschema.

Addresses TODO at `agent/llmagent/llmagent.go:238`.

## Changes

- Add `internal/schemautil` with converter functions
- Update `AgentTool` and `setModelResponseTool` to use jsonschema validation  
- Remove unused `internal/utils/schema_utils.go`


karolpiotrowicz:
@wolo-lab flagging a behavior change that reaches beyond the diff.

The change unifies schema validation on a single library, and the effect is that enum constraints start being enforced on agent-tool inputs and output schemas where they previously were not. Anyone relying on the looser behavior gets a new runtime failure. That is a repo-wide policy change worth deciding deliberately, and it may want a release note either way.

PR #597, opened 2026-02-24
feat: integrate debug endpoints with adk-web

pawel-maciejczek:
The previous version was based on proposed v2 debug endpoints, which got abandoned in adk-web. Adk-web will support v1 endpoints and semconv spans.

Change summary:
1. Changed `gen_ai.usage.output_tokens` attribute to use CandidatesTokenCount instead of TotalTokenCount
1. Added `gcp.vertex.agent.invocation_id` id attribute to `invoke_agent` and `generate_content` spans
2. Added missing `gcp.vertex.agent.tool_call_args` attribute to `execute_tool` span
3. Updated `/debug/trace/session/{session_id}` to return all traces with a span with matching session ID (not only the spans with matching sessionID).
4. Updated the response of debug endpoints to match adk-web v1 format.
5. Improved debug endpoints output for logs to return log body as JSON instead of JSON encoded string

Local run shows that the Trace panel is finally populated with data:
image

`[no invocation id found]` will require fixing adk-web.

### DIFF
--- a/internal/llminternal/base_flow.go
+++ b/internal/llminternal/base_flow.go
@@ -23,6 +23,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/google/uuid"
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/agent"
@@ -187,7 +188,7 @@ func (f *Flow) runOneStep(ctx agent.InvocationContext) iter.Seq2[*session.Event,
 
 			// Handle function calls.
 
-			ev, err := f.handleFunctionCalls(ctx, tools, resp, nil)
+			ev, err := f.handleFunctionCalls(ctx, tools, resp.LLMResponse, nil)
 			if err != nil {
 				yield(nil, err)
 				return
@@ -282,14 +283,18 @@ func toolPreprocess(ctx agent.InvocationContext, req *model.LLMRequest, tools []
 	return nil
 }
 
-func (f *Flow) callLLM(ctx agent.InvocationContext, req *model.LLMRequest, stateDelta map[string]any) iter.Seq2[*model.LLMResponse, error] {
-	return func(yield func(*model.LLMResponse, error) bool) {
+func newResponseWithEventID(resp *model.LLMResponse) *responseWithEventID {
+	return &responseWithEventID{resp, uuid.New().String()}
+}
+
+func (f *Flow) callLLM(ctx agent.InvocationContext, req *model.LLMRequest, stateDelta map[string]any) iter.Seq2[*responseWithEventID, error] {
+	return func(yield func(*responseWithEventID, error) bool) {
 		pluginManager := pluginManagerFromContext(ctx)
 		if pluginManager != nil {
 			cctx := icontext.NewCallbackContextWithDelta(ctx, stateDelta)
 			callbackResponse, callbackErr := pluginManager.RunBeforeModelCallback(cctx, req)
 			if callbackResponse != nil || callbackErr != nil {
-				yield(callbackResponse, callbackErr)
+				yield(newResponseWithEventID(callbackResponse), callbackErr)
 				return
 			}
 		}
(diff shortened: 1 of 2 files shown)


internal/telemetry/logger.go
```diff
@@ -189,44 +188,3 @@ func contentToJSONLikeValue(c *genai.Content) any {
...
-// To unmarshal JSON into an interface value,
-// Unmarshal stores one of these in the interface value:
-//
-//   - bool, for JSON booleans
-//   - float64, for JSON numbers
-//   - string, for JSON strings
-//   - []any, for JSON arrays
-//   - map[string]any, for JSON objects
-//   - nil for JSON null
-func mapToLogValue(v any) log.Value {
```
pawel-maciejczek, line 204: Moved to `converters.go`

server/adkrest/controllers/debug.go
```diff
@@ -77,16 +96,7 @@ func (c *DebugAPIController) SessionSpansHandler(rw http.ResponseWriter, req *ht
...
-		SchemaVersion: 2,
-		Spans:         spans,
-	}
-	EncodeJSONResponse(result, http.StatusOK, rw)
-}
-
-type SessionTelemetry struct {
-	SchemaVersion int                  `json:"schema_version"`
-	Spans         []services.DebugSpan `json:"spans"`
+	EncodeJSONResponse(spans, http.StatusOK, rw)
```
pawel-maciejczek, line 99: v1 expects the list of spans, without the wrapper with version number.

server/adkrest/internal/services/debugtelemetry_test.go
```diff
@@ -41,85 +41,169 @@ func TestDebugTelemetryGetSpansBySessionID(t *testing.T) {
 
 	tests := []testCase{
 		{
-			name: "nested-span-with-log",
-			testSetup: func(ctx context.Context, tracer trace.Tracer, logger log.Logger) {
-				ctx, span := tracer.Start(ctx, "parent-span", trace.WithAttributes(
+			name: "root span with conversation id",
```
pawel-maciejczek, line 44: I added more tests to verify that we're returning all spans in a trace with matching session id.

internal/llminternal/base_flow.go
```diff
@@ -470,16 +490,17 @@ func (f *Flow) agentToRun(ctx agent.InvocationContext, agentName string) agent.A
...
 
-func (f *Flow) finalizeModelResponseEvent(ctx agent.InvocationContext, resp *model.LLMResponse, tools map[string]tool.Tool, stateDelta map[string]any) *session.Event {
+func (f *Flow) finalizeModelResponseEvent(ctx agent.InvocationContext, resp *responseWithEventID, tools map[string]tool.Tool, stateDelta map[string]any) *session.Event {
 	// FunctionCall & FunctionResponse matching algorithm assumes non-empty function call IDs
 	// but function call ID is optional in genai API and some models do not use the field.
 	// Generate function call ids. (see functions.populate_client_function_call_id in python SDK)
 	utils.PopulateClientFunctionCallID(resp.Content)
 
 	ev := session.NewEvent(ctx.InvocationID())
+	ev.ID = resp.eventID // TODO change NewEvent to accept event id
```
RKest, line 500: This is less than ideal (finalize modifying input parameters)

Is it possible to apply the TODO in this PR, or return the eventID from this function

internal/llminternal/base_flow_telemetry_test.go
```diff
@@ -103,8 +103,8 @@ func TestGenerateContentTracing(t *testing.T) {
 				// Yield final response - should not panic.
 				if !yield(&model.LLMResponse{
 					UsageMetadata: &genai.GenerateContentResponseUsageMetadata{
-						PromptTokenCount: 100,
-						TotalTokenCount:  200,
+						PromptTokenCount:     100,
+						CandidatesTokenCount: 200,
```
RKest, line 107: Awesome, thanks for the fix

Issue #596, opened 2026-02-23
Add Support for OpenAI-Compatible Providers


crojas:
Currently, the Go version of the Agent Development Kit (ADK) is tightly coupled with Gemini/Vertex AI. While the Model interface exists, there are no built-in providers for OpenAI or other popular LLM services. This creates a high barrier to entry for developers who want to maintain a multi-cloud strategy or use local models (like Llama 3 via Ollama) using a common SDK. I would like to see an official provider (driver) for OpenAI-compatible APIs within the google.golang.org/adk/model package.

The Python version of many Google AI tools often has broader provider support. Bringing this parity to the Go ADK would significantly increase its adoption in the backend community where Go is a primary language.

verdverm:
dupe of https://github.com/google/adk-go/issues/341

tl;dr - there is a repo in the works for community / 3rd party extensions which is a better home for this

achetronic:
@crojas By the moment... 

https://github.com/achetronic/adk-utils-go

civaapple-alt:
can try [adk-openai](https://github.com/civaapple-alt/adk-openai) for study

support
Chat Completions (POST /v1/chat/completions) — default
Responses API (POST /v1/responses)

achetronic:
> can try [adk-openai](https://github.com/civaapple-alt/adk-openai) for study
> 
> support
> Chat Completions (POST /v1/chat/completions) — default
> Responses API (POST /v1/responses)
> 
> 

You could contribute to adk-utils-go too 😊

https://github.com/achetronic/adk-utils-go

PR #598, opened 2026-02-25
feat(model): add Anthropic LLM provider

### LINKED ISSUE #225: When is the plan to support the Claude model?

ivanvanderbyl:

- Closes: #225
- Related: N/A

N/A (issue exists)

**Problem:**
ADK Go was model-agnostic in design, but there was no first-party Anthropic provider package in this repository. Users wanting Claude support had to rely on external integrations.

**Solution:**
Added a new additive provider package at `model/anthropic` that implements the existing `model.LLM` interface without changing any existing API surface.  
The implementation includes:

- direct Anthropic API support and Vertex AI variant support,
- streaming and non-streaming generation paths,
- tool/function calling conversions,
- thinking config and structured output handling (including Vertex fallback),
- dedicated tests for the provider and converters.

Also added docs and a runnable example:

- `examples/anthropic/main.go`
- README updates documenting package usage.

### Testing Plan

**Unit Tests:**

Commands run:

- `go test ./model/anthropic/...`
- `go test ./model/...`
- `go test ./...`

Result summary:

- All tests passed locally, including the full repository test sweep.

Manual steps to verify locally:

1. Export `ANTHROPIC_API_KEY`.
2. Run: `go run ./examples/anthropic/main.go --help` (launcher wiring check).
3. Run the example via console mode and prompt the agent (for example, “Give a short summary of Sydney weather patterns in summer.”).
4. Optionally validate Vertex mode by setting:
   - `ANTHROPIC_USE_VERTEX=1`
   - `GOOGLE_CLOUD_PROJECT`
   - `GOOGLE_CLOUD_LOCATION`

Notes:

- Full automated test suite passed.
- Live Anthropic API behaviour depends on local credentials and environment.

### Checklist

### Additional context

This PR keeps change surface minimal and idiomatic to existing repo patterns:

- no modifications to existing public interfaces,
- provider is opt-in via package import (`google.golang.org/adk/model/anthropic`),
- converter logic is internal to the provider package to avoid unnecessary exported surface.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
       <a href="https://google.github.io/adk-docs/">Docs</a> &
       <a href="https://github.com/google/adk-go/tree/main/examples">Samples</a> &
       <a href="https://github.com/google/adk-python">Python ADK</a> &
-      <a href="https://github.com/google/adk-java">Java ADK</a> & 
+      <a href="https://github.com/google/adk-java">Java ADK</a> &
       <a href="https://github.com/google/adk-web">ADK Web</a>.
     </h3>
 </html>
@@ -58,7 +58,7 @@ import (
 )
 
 func buildModel(ctx context.Context) {
-    _, _ = adkanthropic.NewModel(ctx, anthropicapi.Model("claude-sonnet-4-20250514"), &adkanthropic.Config{
+    _, _ = adkanthropic.NewModel(ctx, anthropicapi.Model(anthropic.ModelClaudeSonnet4_6), &adkanthropic.Config{
         APIKey: "${ANTHROPIC_API_KEY}",
     })
 }
--- a/README.md
+++ b/README.md
@@ -45,6 +45,27 @@ To add ADK Go to your project, run:
 go get google.golang.org/adk
 ```
 
+### Anthropic Claude model package
+
+ADK Go now includes an Anthropic provider that implements the same `model.LLM` interface used by Gemini.
+
+```go
+import (
+    "context"
+
+    anthropicapi "github.com/anthropics/anthropic-sdk-go"
+    adkanthropic "google.golang.org/adk/model/anthropic"
+)
+
+func buildModel(ctx context.Context) {
+    _, _ = adkanthropic.NewModel(ctx, anthropicapi.Model("claude-sonnet-4-20250514"), &adkanthropic.Config{
+        APIKey: "${ANTHROPIC_API_KEY}",
+    })
+}
+```
+
+See [`examples/anthropic/main.go`](examples/anthropic/main.go) for a full runnable sample.
+
 ## 📄 License
 
 This project is licensed under the Apache 2.0 License - see the
--- a/model/anthropic/anthropic.go
+++ b/model/anthropic/anthropic.go
@@ -1,0 +1,436 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package anthropic
+
+import (
+	"context"
+	"fmt"
+	"iter"
+	"log/slog"
+	"os"
+	"regexp"
+	"strings"
+
+	anthropicapi "github.com/anthropics/anthropic-sdk-go"
+	"github.com/anthropics/anthropic-sdk-go/option"
+	"github.com/anthropics/anthropic-sdk-go/vertex"
+	"google.golang.org/genai"
+
+	"google.golang.org/adk/model"
+	"google.golang.org/adk/model/anthropic/internal/converters"
+)
+
+const defaultMaxTokens = 4096
+
+type anthropicModel struct {
+	client           anthropicapi.Client
+	name             anthropicapi.Model
+	variant          string
+	defaultMaxTokens int
+}
+
+// NewModel returns [model.LLM], backed by Anthropic Claude.
+//
+// It creates an Anthropic client based on the provided configuration.
+// If Variant is not specified, it checks the ANTHROPIC_USE_VERTEX environment variable.
+//
+// For direct Anthropic API, set APIKey in the config or the ANTHROPIC_API_KEY
+// environment variable.
+//
+// For Vertex AI, set VertexProjectID and VertexLocation in the config or use
+// GOOGLE_CLOUD_PROJECT and GOOGLE_CLOUD_LOCATION environment variables.
+func NewModel(ctx context.Context, modelName anthropicapi.Model, cfg *Config) (model.LLM, error) {
+	if cfg == nil {
+		cfg = &Config{}
+	}
+
+	variant := cfg.Variant
+	if variant == "" {
+		variant = GetVariant()
+	}
+
+	var client anthropicapi.Client
+
+	switch variant {
+	case VariantVertexAI:
+		projectID := cfg.VertexProjectID
+		if projectID == "" {
+			projectID = os.Getenv("GOOGLE_CLOUD_PROJECT")
+		}
+		if projectID == "" {
+			return nil, fmt.Errorf("VertexProjectID is required for Vertex AI (set GOOGLE_CLOUD_PROJECT)")
+		}
+
+		location := cfg.VertexLocation
+		if location == "" {
+			location = os.Getenv("GOOGLE_CLOUD_LOCATION")
+		}
+		if location == "" {
+			return nil, fmt.Errorf("VertexLocation is required for Vertex AI (set GOOGLE_CLOUD_LOCATION)")
+		}
+
+		client = newVertexClient(ctx, cfg)
+	default:
+		client = newAPIClient(cfg)
+	}
+
+	maxTokens := cfg.DefaultMaxTokens
+	if maxTokens == 0 {
+		maxTokens = defaultMaxTokens
+	}
+
+	return &anthropicModel{
+		client:           client,
+		name:             modelName,
+		variant:          variant,
+		defaultMaxTokens: maxTokens,
+	}, nil
+}
+
+// newAPIClient creates a client for the direct Anthropic API.
+func newAPIClient(cfg *Config) anthropicapi.Client {
+	opts := []option.RequestOption{}
+
+	apiKey := cfg.APIKey
+	if apiKey == "" {
+		apiKey = os.Getenv("ANTHROPIC_API_KEY")
+	}
+	if apiKey != "" {
+		opts = append(opts, option.WithAPIKey(apiKey))
+	}
+
+	return anthropicapi.NewClient(opts...)
+}
+
+// newVertexClient creates a client for Anthropic via Vertex AI.
+// Note: The caller must validate that projectID and region are set before calling this.
+func newVertexClient(ctx context.Context, cfg *Config) anthropicapi.Client {
+	projectID := cfg.VertexProjectID
+	if projectID == "" {
+		projectID = os.Getenv("GOOGLE_CLOUD_PROJECT")
+	}
+
+	location := cfg.VertexLocation
+	if location == "" {
+		location = os.Getenv("GOOGLE_CLOUD_LOCATION")
+	}
+
+	return anthropicapi.NewClient(
+		vertex.WithGoogleAuth(ctx, location, projectID),
+	)
+}
+
+// Name returns the model name.
+func (m *anthropicModel) Name() string {
+	return string(m.name)
+}
+
+// GenerateContent calls the Anthropic model.
+func (m *anthropicModel) GenerateContent(ctx context.Context, req *model.LLMRequest, stream bool) iter.Seq2[*model.LLMResponse, error] {
+	m.maybeAppendUserContent(req)
+
+	if stream {
+		return m.generateStream(ctx, req)
+	}
+
+	return func(yield func(*model.LLMResponse, error) bool) {
+		resp, err := m.generate(ctx, req)
+		yield(resp, err)
+	}
+}
+
+// generate calls the model synchronously.
+func (m *anthropicModel) generate(ctx context.Context, req *model.LLMRequest) (*model.LLMResponse, error) {
+	// Vertex AI doesn't support OutputConfig — fall back to prompt-based JSON.
+	if m.variant == VariantVertexAI && req.Config != nil && req.Config.ResponseSchema != nil {
+		return m.generateWithPromptBasedJSON(ctx, req)
+	}
+
+	params, err := m.convertRequest(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert request: %w", err)
+	}
+
+	msg, err := m.client.Messages.New(ctx, params)
+	if err != nil {
+		return nil, fmt.Errorf("failed to call model: %w", err)
+	}
+
+	resp, err := converters.MessageToLLMResponse(msg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert response: %w", err)
+	}
+
+	return resp, nil
+}
+
+// generateWithPromptBasedJSON falls back to prompt-based JSON for Vertex AI.
+// It embeds the JSON schema in the system instruction and asks the model to respond with valid JSON.
+func (m *anthropicModel) generateWithPromptBasedJSON(ctx context.Context, req *model.LLMRequest) (*model.LLMResponse, error) {
+	req = embedSchemaAsSystemPrompt(req)
+
+	params, err := m.convertRequest(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert request: %w", err)
+	}
+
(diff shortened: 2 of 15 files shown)


verdverm:
See https://github.com/google/adk-go/issues/596

karolpiotrowicz:
@wolo-lab a direction call before review, and this one has a second PR attached to it.

The change adds a third-party dependency and a new public `model/anthropic` package. The capability is clearly wanted, so the question is not whether but which: #233 proposes the same thing by a different route and has been open for 300 days. Picking one and closing the other is the decision that unblocks both.

wolo-lab:
@karolpiotrowicz the call is made, and it is this one. Please go ahead with a detailed review here.

#233 takes the same capability by a different route. This PR covers more of the surface — extended thinking round-tripped with its signature, ToolConfig mapped through to tool_choice, validated image media types, documents, and citations — and for a provider we will maintain for years, starting from the more complete conversion layer is the lower-risk path. We are closing #233 with thanks to its author.

@ivanvanderbyl we want to take this forward and land it. Expect review comments shortly. Given that you have run this code against production traffic for months, your view on anything we ask for will carry weight.

model/anthropic/anthropic.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+
+	return func(yield func(*model.LLMResponse, error) bool) {
+		params, err := m.convertRequest(req)
+		if err != nil {
+			yield(nil, fmt.Errorf("failed to convert request: %w", err))
+			return
+		}
+
+		stream := m.client.Messages.NewStreaming(ctx, params)
```
karolpiotrowicz, line 219: The stream returned here owns the HTTP response body, and `Close()` is never called on any path out of this function — not on the accumulate error, not on either consumer-break return, not on the stream error, not on normal completion. `base_flow.go` legitimately stops consuming mid-iteration, so this is reached in ordinary use.

Measured on two copies of the tree differing only by a `defer stream.Close()` on the next line, same test and same server, consumer breaking after the first value:

```
without the line:  server never saw the client go away, wrote 287 further events
with the line:     server saw it immediately, wrote 1
```

I ran that one-line change rather than just suggesting it: with it applied the package suite passes, `go build ./...` is clean, and a live streaming round-trip still yields partials then a final with `TurnComplete` set.

model/anthropic/internal/converters/tools.go
```diff
@@ -0,0 +1,379 @@
...
+		if schema == nil {
+			continue
+		}
+		result[name] = SchemaToMap(schema)
+	}
+	return result
+}
+
+// SchemaToMap converts a genai.Schema to a map[string]any suitable for Anthropic.
+func SchemaToMap(schema *genai.Schema) map[string]any {
```
karolpiotrowicz, line 183: `SchemaToMap` recurses through `Items`, `Properties` and `AnyOf` with no depth limit and no visited set, so a self-referential `genai.Schema` never terminates. Driven through `GenerateContent` it produces `fatal error: stack overflow`, which is a runtime throw rather than a panic, so `recover()` cannot contain it and the process dies.

On reachability, to keep this in proportion: `functiontool.New` rejects a recursive Go struct at construction ("cycle detected for type"), and MCP schemas arrive as JSON and cannot carry a Go pointer cycle, so it needs a hand-built `*genai.Schema` cycle passed as `Parameters` or `ResponseSchema`. The reason a depth cap still seems worth it is that the same value handed to the genai marshaller comes back as an error — the neighbouring provider degrades where this one takes the process down.

`jsonSchemaPropertyToMap` just below has the same unbounded shape.

model/anthropic/internal/converters/response.go
```diff
@@ -0,0 +1,250 @@
...
+
+// ContentBlockToGenaiPart converts an Anthropic ContentBlockUnion to a genai.Part.
+func ContentBlockToGenaiPart(block anthropic.ContentBlockUnion) (*genai.Part, error) {
+	switch variant := block.AsAny().(type) {
+	case anthropic.TextBlock:
+		return &genai.Part{Text: variant.Text}, nil
+
+	case anthropic.ThinkingBlock:
+		// Map thinking blocks to genai.Part with Thought=true
+		signature, _ := base64.StdEncoding.DecodeString(variant.Signature)
```
karolpiotrowicz, line 77: The decode error is discarded, and `DecodeString` returns the bytes decoded before the error — so a signature that is not canonical padded StdEncoding yields a truncated blob, which is then re-encoded on replay into something the server did not issue. If the first quantum fails the slice is empty, and the request side then treats the thought as ordinary text.

Scoping this honestly: I replayed signed thoughts across three live turns and the round-trip worked, so this is about malformed encodings rather than the normal path. The signature is an opaque token though, so decoding it at all is avoidable — adk-python stores the bytes verbatim: [anthropic_llm.py:608](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/models/anthropic_llm.py#L608).

model/anthropic/internal/converters/response.go
```diff
@@ -0,0 +1,250 @@
...
+		return &genai.Part{
+			Text:             variant.Thinking,
+			Thought:          true,
+			ThoughtSignature: signature,
+		}, nil
+
+	case anthropic.RedactedThinkingBlock:
+		// Redacted thinking - we can't see the content but preserve the marker
+		return &genai.Part{
+			Text:    "[thinking redacted]",
```
karolpiotrowicz, line 87: A redacted thinking block loses its `Data` payload and is replaced by this literal, with no signature. On the next turn the request side sees text with `Thought: true` and no signature, falls past the thinking branch, and sends `[thinking redacted]` to the model as ordinary assistant prose — words the assistant never said, in place of the blob Anthropic expects echoed back.

adk-python keeps the payload in the signature field: [anthropic_llm.py:615](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/models/anthropic_llm.py#L615).

model/anthropic/internal/converters/response.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+		// Collect citations from text blocks
+		if textBlock, ok := block.AsAny().(anthropic.TextBlock); ok {
+			if citations := textCitationsToSlice(textBlock.Citations); len(citations) > 0 {
+				allCitations = append(allCitations, citations...)
+			}
+		}
+	}
+
+	resp := &model.LLMResponse{
```
karolpiotrowicz, line 56: `ErrorCode` and `ErrorMessage` are never set here, and `Content` is always allocated, so a reply with no content blocks becomes a successful empty turn: `parts=0`, `errorCode=""`, `FINISH_REASON_UNSPECIFIED`, no error. `agent/agenttool` and `server/adka2a` both use `ErrorCode`/`ErrorMessage` as their only failure channel, and the Gemini path sets `ErrorCode` from the finish reason.

Worth being straight about how likely this is: I tried to obtain a real content-less refusal and could not. Both refusals I got back came as ordinary text with `finish=STOP`, so this is a gap in the mapping rather than something I can show biting today. `refusal` and `pause_turn` are also absent from `StopReasonToFinishReason` and collapse to unspecified.

model/anthropic/anthropic.go
```diff
@@ -0,0 +1,436 @@
...
+		location := cfg.VertexLocation
+		if location == "" {
+			location = os.Getenv("GOOGLE_CLOUD_LOCATION")
+		}
+		if location == "" {
+			return nil, fmt.Errorf("VertexLocation is required for Vertex AI (set GOOGLE_CLOUD_LOCATION)")
+		}
+
+		client = newVertexClient(ctx, cfg)
+	default:
```
karolpiotrowicz, line 85: An unrecognised `Variant` falls through to the direct API rather than being rejected, even though `config.go` documents only two valid values. A typo such as `"VERTEXAI"` skips the project and location validation just above, stores the bogus string, and makes the three later `m.variant == VariantVertexAI` checks read false — so the request also carries an `OutputConfig` that was deliberately suppressed for Vertex, and traffic goes to the public API.

model/anthropic/anthropic.go
```diff
@@ -0,0 +1,436 @@
...
+
+	// Clone the config to avoid mutating the original request.
+	modifiedConfig := *req.Config
+	if modifiedConfig.SystemInstruction == nil {
+		modifiedConfig.SystemInstruction = &genai.Content{
+			Parts: []*genai.Part{{Text: jsonInstruction}},
+		}
+	} else {
+		existingText := ""
+		for _, part := range modifiedConfig.SystemInstruction.Parts {
```
karolpiotrowicz, line 373: This dereferences `part` without a nil check, unlike the three sibling loops in this package. A nil entry in `SystemInstruction.Parts` panics here on the Vertex path with an output schema set.

model/anthropic/internal/converters/request.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+
+	// Anthropic requires input to be a dictionary - ensure we have a valid map
+	// After JSON round-trip, nil maps stay nil, so we must always provide a valid map
+	var input any = call.Args
+	if call.Args == nil || len(call.Args) == 0 {
+		input = map[string]any{}
+	}
+
+	block := anthropic.NewToolUseBlock(call.ID, input, call.Name)
```
karolpiotrowicz, line 325: `functionResponseToBlock` above rejects an empty ID with a clear local error, but a `FunctionCall` with no ID is serialised as `"id": ""` and left to fail server-side. Worth mirroring the check so both halves of a tool round-trip fail the same way.

model/anthropic/internal/converters/response.go
```diff
@@ -0,0 +1,250 @@
...
+
+	return result
+}
+
+// UsageToMetadata converts Anthropic Usage to genai UsageMetadata.
+func UsageToMetadata(usage anthropic.Usage) *genai.GenerateContentResponseUsageMetadata {
+	return &genai.GenerateContentResponseUsageMetadata{
+		PromptTokenCount:     int32(usage.InputTokens),
+		CandidatesTokenCount: int32(usage.OutputTokens),
+		TotalTokenCount:      int32(usage.InputTokens + usage.OutputTokens),
```
karolpiotrowicz, line 202: The total is input plus output only, so prompt-cache tokens are dropped from usage accounting. A reply reporting 10 input, 5 output, 100 cache-creation and 200 cache-read comes back as a total of 15. `ThoughtsTokenCount` is also never populated, so a thinking-enabled call reports no thinking tokens.

README.md
```diff
@@ -45,6 +45,27 @@ To add ADK Go to your project, run:
...
+```go
+import (
+    "context"
+
+    anthropicapi "github.com/anthropics/anthropic-sdk-go"
+    adkanthropic "google.golang.org/adk/model/anthropic"
+)
+
+func buildModel(ctx context.Context) {
+    _, _ = adkanthropic.NewModel(ctx, anthropicapi.Model(anthropic.ModelClaudeSonnet4_6), &adkanthropic.Config{
```
karolpiotrowicz, line 61: This snippet does not compile — the import above is aliased `anthropicapi`, so `anthropic.ModelClaudeSonnet4_6` is undefined. The `anthropicapi.Model(...)` conversion is also redundant, since the constant already has that type.

Line 62 sets `APIKey` to the literal `"${ANTHROPIC_API_KEY}"`. Go does not expand that, and because it is non-empty it overrides the environment variable the SDK would otherwise pick up, so following the README gives a 401 that surfaces as a generic call failure. `examples/anthropic/main.go` gets both of these right.

karolpiotrowicz (changes requested):
Two things need fixing before this can merge. Extended thinking cannot succeed against the Anthropic API as this stands, and on the Vertex path a structured-output response can be silently replaced with the wrong data. Everything else below can land afterwards. Separately, the branch needs a rebase first — `main` moved to the `google.golang.org/adk/v2` module path since this was opened in February, so the imports need migrating, and CI has never run on this head.

Worth saying what holds up, because I tested it rather than assuming it. A three-turn conversation with thinking enabled, two parallel tool calls, the assistant turn replayed verbatim with its signed thought parts, and tool results correlated by ID, all ran clean against a live endpoint. That is the hardest round-trip in an Anthropic mapping and it works. Tool results forced to user role and tool calls to assistant, strict alternation, `tool_use.input` always an object, a tool result without an ID failing locally rather than as a remote 400, and `embedSchemaAsSystemPrompt` copying the config instead of mutating the caller's — those are the details a first-cut provider usually gets wrong.

## Thinking is unusable without also setting MaxOutputTokens

`resolveThinkingBudget` returns 10000 for `ThinkingLevelHigh` and for a bare `IncludeThoughts: true`, while `max_tokens` stays at the 4096 default unless the caller sets `MaxOutputTokens`. Anthropic requires `max_tokens > thinking.budget_tokens`, so the simplest way to ask for thoughts fails every time:

```
ThinkingConfig{ThinkingLevel: HIGH}     -> 400 "max_tokens must be greater than thinking.budget_tokens"
ThinkingConfig{IncludeThoughts: true}   -> 400, same message
same request with MaxOutputTokens=12000 -> 200 OK
```

`ThinkingConfig{IncludeThoughts: true}` with nothing else set is the shape `agent/llmagent/llmagent_test.go` already uses, so this is the default experience rather than an edge case.

Two more problems in the same function. `ThinkingBudget: 0` means "do not think", but it is returned verbatim through a `budget >= 0` gate and becomes thinking-enabled-with-budget-0, which the API rejects with `budget_tokens: Input should be greater than or equal to 1024`. And `ThinkingLevelMedium`, `ThinkingLevelMinimal` and genai's `-1` all fall through to the default arm and silently disable thinking. adk-python maps a zero budget to a disabled config at [anthropic_llm.py:193](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/models/anthropic_llm.py#L193).

Since fixing only the `IncludeThoughts` case would leave the others broken, the property worth holding to is: for every `ThinkingConfig` a caller can construct, the emitted request either omits `thinking` entirely or satisfies `1024 <= budget_tokens < max_tokens`.

## The Vertex JSON fallback can return the wrong answer, silently

On the Vertex variant with an output schema, every response goes through `stripMarkdownFromResponse`. The permissive extractor runs on any text the strict regex rejected, with no anchoring and no check that the text is already valid JSON, and replaces the entire text with the first fenced object it finds.

Asking the model to draft inside a fenced block before committing to an answer — an ordinary reasoning style — the caller received the draft instead of the answer:

```
caller received:  {"draft": 1}

control, same prompt with no ResponseSchema so the fallback never runs.
this is what the model actually produced:
  "# Reasoning\n\nLet me think through this step by step:\n\n1. The user wants me to
   demonstrate my thinking process\n2. I need to create a draft JSON block ..."
```

There is no error and the result is valid JSON, so nothing downstream can tell. `basic_processor` sets `ResponseSchema` for any non-Gemini model whenever the agent declares an `OutputSchema`, so this is active for every structured-output agent on Vertex.

The same loop also rewrites thought parts, because it skips only nil and empty-text parts and never checks `part.Thought` — the text is replaced while the signature is kept, so the pair no longer matches when it is replayed. And the strict regex's `(?:json)?` is lowercase-only, so a ` ```JSON ` fence has its tag captured into the payload rather than stripped.

## A text artifact fails the whole turn

`inlineDataToBlock` errors on any MIME type that is not an image or a PDF, and that error propagates up, so one unsupported part fails the entire request rather than just that part. `tool/loadartifactstool` appends stored artifact parts into `req.Contents` verbatim, which makes a plain-text artifact an ordinary input:

```
text/plain       -> turn fails: unsupported MIME type for inline data: text/plain
text/csv         -> turn fails
application/json -> turn fails
image/png        -> ok
```

The same agent on `model/gemini` succeeds.

## Tool calling cannot be turned off

`FunctionCallingConfigModeNone` produces an empty `tool_choice`, but the tools are still attached, and Anthropic defaults to `auto` when tools are present and `tool_choice` is absent. The emitted request carries the tool list with no `tool_choice` key, and on a live call under ModeNone the model called the tool.

## Schema constraints are dropped for tools declared by JSON schema

`jsonSchemaPropertyToMap` keeps six keywords. A `map[string]any` passes through untouched, so this only bites the `*jsonschema.Schema` form — which is what `mcptoolset` assigns straight from a remote server's `InputSchema`, and what `functiontool` produces. Off the wire:

```
declared:  code  {"type":"string","pattern":"^[A-Z]{3}-[0-9]{4}$"}
           ratio {"type":"number","minimum":0,"maximum":1}
received:  code  {"type":"string"}
           ratio {"type":"number"}
```

`enum` and `description` survive. To be fair to the change, I could not get a model to actually violate the dropped constraints — asked to call such a tool it supplied conforming values anyway, so this is loss of fidelity rather than a demonstrated misbehaviour. adk-python deep-copies the schema whole at [anthropic_llm.py:734](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/models/anthropic_llm.py#L734).

## Tests

Not a merge gate for a new package, but one gap is worth closing alongside the fixes: no test in the package calls `GenerateContent` at all, so `generate`, `generateWithPromptBasedJSON` and `generateStream` have zero coverage. Deleting `finalResp.TurnComplete = true`, dropping the final yield, swapping the text and thinking delta converters, or dropping `params.System` each leave the suite green. The SDK honours `option.WithBaseURL`, so an `httptest.Server` emitting a canned SSE transcript covers the streaming path directly — that is how the stream finding in the inline comments was measured.

Whatever lands from the above, it would help if each fix came with a test that fails without it — the thinking-budget one especially, since a converter test currently asserts the 10000 value.

I have left the smaller items as inline comments.

Issue #469, opened 2026-01-09
Set MCP metadata on a request level


philippslang:
**Is your feature request related to a problem? Please describe.**
I need to forward request-level metadata to MCP. This is usually done in mcp.CallToolParams, but there seems no way to inject this in the current mcpTool.Run function.

**Describe the solution you'd like**
I'd like a way to set MCP tool call metadata on a request-level basis. Specifically my use-case is to forward A2A request-level metadata to MCP metadata.

**Describe alternatives you've considered**
I can implement mcpToolset myself, which is fine.

indurireddy-TF:
Thanks for opening this! This would be a very useful addition to our MCP support. @baptmont @dpasiukevich , could you please take a look at the open PR ?

PR #602, opened 2026-02-27
feat: add metadata support to MCP tools and refactor metadata tests

alexshopee:
## Summary

This PR is **based on [PR #475](https://github.com/google/adk-go/pull/475)** (“feat: add metadata support to MCP tools”) by **@philippslang (Philipp Lang)**. It implements the same feature for [Issue #469](https://github.com/google/adk-go/issues/469), and adds test refactoring and robustness improvements from code review.

## Issue & related PR

- **Issue:** [#469](https://github.com/google/adk-go/issues/469) — metadata support for MCP toolset (request-scoped metadata forwarding).
- **Related PR:** [#475](https://github.com/google/adk-go/pull/475) — feat: add metadata support to MCP tools (**author:** [@philippslang](https://github.com/philippslang) (Philipp Lang)).

## Purpose (from #475)

- Introduce a **MetadataProvider** so request-scoped metadata (e.g. tracing IDs, auth tokens) can be forwarded from the ADK environment to MCP tool calls.
- Add optional `MetadataProvider` on `mcptoolset.Config`; when set and returning non-nil metadata, attach it to `mcp.CallToolParams.Meta` before calling the MCP tool.
- Support no provider configured or provider returns nil (no metadata attached).

---

### Functionality (production code)

| Area | PR #475 | This PR (#602) |
|------|--------|-----------------|
| **tool.go – metadata assignment** | `if t.metadataProvider != nil { if meta := t.metadataProvider(ctx); meta != nil { params.Meta = meta } }` | `if t.metadataProvider != nil { params.Meta = t.metadataProvider(ctx) }` — direct assignment (nil is fine for `params.Meta`). |
| **tool.go – panic recovery** | None. | `defer recover()` in `mcpTool.Run`, converting panic to an error with tool name and stack (same pattern as `functiontool.Run`), so a panicking `MetadataProvider` does not crash the process. |

No other API or config changes; `MetadataProvider` type and `Config`/`set` wiring match #475.

### Unit tests (set_test.go)

| Area | PR #475 | This PR (#602) |
|------|--------|-----------------|
| **Structure** | Three separate tests: `TestMetadataProvider`, `TestMetadataProviderNil`, `TestMetadataProviderReturnsNil`. Each duplicated: in-memory transport, server, `AddTool`, `mcptoolset.New`, `Tools()`, get first tool, `Run(toolCtx, map[string]any{})`. | Single **table-driven** test `TestMetadataProvider` with three cases: `"provider returns metadata"`, `"provider is nil"`, `"provider returns nil"`. Shared setup and run moved into a **generic helper** `runMetadataTest[In, Out any](t, provider, toolFunc)`. |
| **Helper** | None. | `runMetadataTest`: builds in-memory transport & server, registers one tool with `toolFunc`, creates toolset with given `provider`, calls `Tools()`, runs the single tool with `map[string]any{}`. No return value; failures via `t.Fatal`/`t.Fatalf` (e.g. tool run error). |
| **Tool-invocation assertion** | Only `TestMetadataProviderReturnsNil` used a `metaCalled` flag and asserted the tool was called. | All three cases use a **`toolCalled`** flag set in the handler and assert `toolCalled` after `runMetadataTest`, so no case passes if the handler is never invoked. |
| **Naming** | `metaCalled` in one test. | Unified to **`toolCalled`** everywhere; test name `TestMetadata` → **`TestMetadataProvider`** to reflect metadata-provider behavior. |
| **Assertion placement (provider nil)** | In #475, `TestMetadataProviderNil` could assert inside the handler (`req.Params.Meta != nil` → `t.Errorf`). | Handler only captures state (`toolCalled = true`, `receivedMeta = req.Params.Meta`); assertion **outside** the handler: `if receivedMeta != nil { t.Errorf(...) }`, keeping act vs assert separate. |
| **Result / nil check** | Caller could check `result := runMetadataTest(...); if result == nil { t.Fatal(...) }`. | Helper returns nothing; redundant `result == nil` check in helper was removed (per review); callers only assert `toolCalled` and `receivedMeta` (or nil). |

**Coverage:** Same three scenarios as #475 (provider returns metadata; provider is nil; provider returns nil), with added assurance that the tool handler is actually run in each case.

Code follows the [Google Go Style Guide](https://google.github.io/styleguide/go/index). Changes are limited to `tool/mcptoolset` (set.go, tool.go, set_test.go).

### Unit tests

- **Scope:** `tool/mcptoolset` metadata behavior and refactored tests.
- **Coverage:** Provider returns metadata; provider is nil; provider returns nil; tool handler invoked in every case (`toolCalled`).
- **How to run:** `go test ./tool/mcptoolset/ -run 'TestMetadataProvider' -v`
- Tests are fast, isolated, and use in-memory MCP transport. No manual E2E required for this library-only change.


alexshopee:
hi @dpasiukevich i suppose the codes have been improved with the instructions of gemini, could u help take a look at this pr? In addition, i just follow the panic recover mechanism regarding the functionTool as below:
```
func (f *functionTool[TArgs, TResults]) Run(ctx tool.Context, args any) (result map[string]any, err error) {
	// TODO: Handle function call request from tc.InvocationContext.
	defer func() {
		if r := recover(); r != nil {
			err = fmt.Errorf("panic in tool %q: %v\nstack: %s", f.Name(), r, debug.Stack())
		}
	}()
```
and i suppose there is no need to follow the gemini's latest comment, if there is no other problems, pls merge this pr, thank u !

alexshopee:
hi @baptmont could u help take a look at this pr, i appreciate it

vnalawad-tibco:
@alexshopee Thanks for the PR. In my use case, I want to pass OTEL tracing context in MCP tool call as metadata. This change can help propagate tracing context from agent to MCP servers. 
@baptmont Can I request you to review it as well?

baptmont:
Picking this up with maintainer edits, since the branch had drifted a long way behind `main`. The feature is still wanted — #469 is open and @vnalawad-tibco has a live use case for it above.

## Why it was rewritten rather than rebased

The branch was on **v1** (`module google.golang.org/adk`) and **172 commits behind** `main`, spanning the v2 module bump. That isn't a conflict resolution:

- `tool.Context` no longer exists — `mcpTool.Run` now takes `agent.Context`
- `ConfirmationProvider` moved to the `tool` package
- the import path is `google.golang.org/adk/v2/...`

The design is unchanged from the original proposal — an optional provider on `Config` whose result becomes `CallToolParams.Meta` — and credit for it belongs to @philippslang (#475) and @alexshopee. The code itself is written fresh against v2.

## One deliberate change: the provider can now fail

```go
type MetadataProvider func(ctx agent.Context) (map[string]any, error)
```

Producing metadata can fail — minting a token, reading a trace header. With no error return the only way to signal that is to panic, which is exactly why the earlier revision needed a `recover()` in `Run`. Returning an error is the smaller and more direct fix: a failure aborts the tool call and names the tool.

For auth- or tenant-scoped metadata this matters. Quietly issuing the call *without* the metadata is worse than not calling at all.

Consequently the panic-recovery hunk is **dropped**. It was justified only by the provider's inability to report failure, and it changed behaviour for every `mcpTool.Run` — including swallowing panics from the MCP client — which belongs in its own change if wanted.

`nil` handling is unchanged: `Meta` is `omitempty`, so no provider, or a provider returning `nil`, sends no `_meta` at all.

## Review feedback

All ten inline comments were on the previous test structure. Status against the current code:

| Comment | Status |
|---|---|
| Assert the tool handler actually ran (×2) | **Done** — every case asserts `called` |
| Move assertions out of the tool handler | **Done** — the handler only captures `called` and `meta` |
| Collapse into a table-driven test | **Done** — `TestMetadataProvider`, four cases |
| Name it `TestMetadataProvider` | **Done** |
| Drop the redundant `result == nil` check | **Done** — no such check |
| `params.Meta = provider(ctx)` without the nested `if` | **Done** — direct assignment |
| Rename `metaCalled` → `toolCalled` | Obsolete — that helper is gone |
| Structured `ToolPanicError` type | Obsolete — panic recovery removed, see above |
| Make the run helper self-contained and return nothing | **Diverged** — `runSingleTool` still returns `(map[string]any, error)`, because the new error test asserts on the returned error. Setup and the tool-count check are inside the helper. |

## Tests

Seven cases driving the tool directly against an in-memory MCP server that captures server-side `_meta` — no model, no `httprr`, so they run in milliseconds: metadata arrives intact; no provider; provider returns `nil`; provider returns an empty map; provider errors; the error names the tool; the provider receives the live invocation context.

Both halves are mutation-checked — dropping `params.Meta = meta` fails `provider_metadata_reaches_the_server`, and ignoring the provider error fails `TestMetadataProviderError`.

```
go build -mod=readonly ./...                                  pass
go test -race -mod=readonly -count=1 -shuffle=on ./tool/...   pass
go test -race -mod=readonly -count=1 ./...                    pass (full repo)
golangci-lint run ./tool/...                                  0 issues
go mod tidy -diff                                             clean, no new deps
gofmt -l tool/                                                clean
```

@vnalawad-tibco — your OTEL case is covered: the provider is handed the live `agent.Context`, so you can pull the span context out of it and return it as `_meta`.

Note the PR description above is now out of date; this comment supersedes it.

baptmont:
Hi @alexshopee thanks for your patience on this.
I've taken the branch over with maintainer edits to bring it up to date. It was on v1 and ~172 commits behind `main`, spanning the v2 module bump `tool.Context` no longer exists and `ConfirmationProvider` has moved so it needed rewriting against the current API rather than rebasing. The design is unchanged and still yours and @philippslang's; I've written up the details in the comment above.

Everything else in CI is green.

There's one thing only you can do, though. The `cla/google` check is still failing, and because you opened the PR the check includes your account regardless of what the commits look like, so I can't clear it from my side.

One gotcha: the email on your CLA has to match an email attached to your GitHub account, otherwise the check won't match you up.
Once that's green this is ready to merge. And if you'd rather not sign, just say the word and I'll land it as a separate PR crediting you and @philippslang for the design.
Thanks again for pushing this forward.

baptmont (approved):

vnalawad-tibco:
Hi @alexshopee,
Any chance to get this merged soon?

vnalawad-tibco:
@alexshopee @baptmont Any way to move this forward? I need this enhancement for my use case.

jjsasha63:
Thanks @alexshopee, and @philippslang before you on #475. I have moved this to
#1429 so it can land, and closing here.

The reason for the move is the CLA check. It covers the commit authors on this
branch, but it also requires the person who opened the pull request, and that
has been outstanding since February. Everything else here is green and approved,
so the check is the only thing left standing between the feature and `main` —
and people are waiting on it, @vnalawad-tibco most directly.

#1429 keeps your design: an optional provider on `Config` feeding
`CallToolParams.Meta`. Both of you are credited in the commit message and the
description. It adds three things on top of the branch as it stood:

- The tests were comparing the whole `_meta` map, which broke when go-sdk v1.7.0
  started adding its own reserved entries to every request. They now assert on
  what the provider contributed.
- The provider's map is copied before the client sees it, so a provider that
  reuses a map does not find the client's keys written into it.
- Two tests, that provider metadata does not displace the client's own entries,
  and that the provider's map is left alone.

@alexshopee if you sign the CLA I will add you as a co-author on #1429 before it
merges — the same offer to @philippslang. Sorry this one took so long.

PR #611, opened 2026-03-03
Remove loopagent dependency from sequential agent

baptmont:


agent/workflowagents/sequentialagent/agent.go
```diff
@@ -54,3 +59,18 @@ type Config struct {
...
 }
+
+type sequentialAgent struct{}
+
+func (a *sequentialAgent) Run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
+	return func(yield func(*session.Event, error) bool) {
+		for _, subAgent := range ctx.Agent().SubAgents() {
+			for event, err := range subAgent.Run(ctx) {
+				// TODO: ensure consistency -- if there's an error, return and close iterator, verify everywhere in ADK.
+				if !yield(event, err) {
```
dpasiukevich, line 70: Do we need to handle event.Actions.Escalate?

dpasiukevich: I've checked. In adk-python they do NOT read event.Actions.Escalate. 

So it only works for the LoopAgent.

PR #616, opened 2026-03-04
Add pathPrefix config to api launcher

### LINKED ISSUE #264: Discrepancy between documented behavior and adk-go behavior in local development curl calls
Related to #255 and about procedures documented at https://google.github.io/adk-docs/runtime/api-server

When started with command: 
`go run agent.go web api`

The expected base url for operations against the local dev server is: 
`http://localhost:8000/`

as in this example: 

`curl -X POST http://localhost:8000/run \
-H "Content-Type: application/json" \
-d '{
"app_name": "my_sample_agent",
"user_id": "u_123",
"session_id": "s_123",
"new_message": {
    "role": "user",
    "parts": [{
    "text": "Hey whats the weather in new york today"
    }]
}
}''`

However the base url  appears to be: 
`http://localhost:8080/api/run`

as in this working example. 

`curl -X POST http://localhost:8080/api/run \ 
-H "Content-Type: application/json" \ 
-d '{ "appName": "boat_agent", "userId": "u_123", "SessionID": "s_123", "newMessage": { "role": "user", "parts": [{ "text": "model: Lucia 40, make: Fontaine Pajot" }] } }'`

Now, maybe this change should be made in the documentation, but I'd argue again like I did in #255 that for these particular examples, behaving like the documented behavior for adk-python is the correct remedy. 

I was unable to track down how to make a PR to fix it. So starting a conversation instead.

baptmont:


cmd/launcher/web/api/api.go
```diff
@@ -64,8 +65,8 @@ func corsWithArgs(frontendAddress string) func(next http.Handler) http.Handler {
 
 // UserMessage implements web.Sublauncher. Prints message to the user
 func (a *apiLauncher) UserMessage(webURL string, printer func(v ...any)) {
-	printer(fmt.Sprintf("       api:  you can access API using %s/api", webURL))
-	printer(fmt.Sprintf("       api:      for instance: %s/api/list-apps", webURL))
+	printer(fmt.Sprintf("       api:  you can access API using %s%s", webURL, a.config.pathPrefix))
```
kdroste-google, line 69: In order to be oblivious to the existence of / in the prefix, please consider using url.JoinPath

cmd/launcher/web/api/api.go
```diff
@@ -76,10 +77,16 @@ func (a *apiLauncher) SetupSubrouters(router *mux.Router, config *launcher.Confi
...
-	)
+	// If prefix is empty, don't use PathPrefix("") because it's too greedy.
+	// Instead, attach the handler to the main router directly.
+	if a.config.pathPrefix == "" || a.config.pathPrefix == "/" {
+		// This allows other routes (like /ui/) to match first if registered
+		router.Methods("GET", "POST", "DELETE", "OPTIONS").Handler(corsHandler)
+	} else {
+		router.Methods("GET", "POST", "DELETE", "OPTIONS").
+			PathPrefix(a.config.pathPrefix).
+			Handler(http.StripPrefix(a.config.pathPrefix, corsHandler))
```
kdroste-google, line 89: Are we handling pathPrefix "api" (instead of "/api") properly?

kdroste-google (changes requested):
Please check handling prefixes like "api" or "api/"

cmd/launcher/web/api/api.go
baptmont: Changed formatting in parse to remove trailing and add leading slash

cmd/launcher/web/api/api.go
baptmont: We were not I changed formatting in parse to remove trailing and add leading slash

kdroste-google (approved):
LGTM

PR #620, opened 2026-03-04
fix: use LLMRequest.Model in Gemini GenerateContent calls

### LINKED ISSUE #619: gemini: GenerateContent ignores LLMRequest.Model, always uses m.name
## Description

The Gemini model implementation (`model/gemini/gemini.go`) ignores the `LLMRequest.Model` field and always uses `m.name` (set at construction time) when calling the GenAI API.

`LLMRequest.Model` is set by `base_flow.go:133` and can be modified by `BeforeModelCallback`, but the Gemini implementation never reads it. This creates a contract violation: the field exists, callbacks can modify it, logging plugins read it, but the actual API call uses a different value.

## Impact

- **BeforeModelCallback model override is silently ineffective** — any code setting `req.Model` to a different model has no effect on the actual API call.
- **Observability mismatch** — the stock `loggingplugin` and any custom plugin logging `req.Model` reports a value that may not match what's actually sent to the API.
- **Dead field** — `LLMRequest.Model` serves no functional purpose in the Gemini backend.

## Root cause

`model/gemini/gemini.go` lines 102 and 118:

```go
// generate()
resp, err := m.client.Models.GenerateContent(ctx, m.name, ...)

// generateStream()
for resp, err := range m.client.Models.GenerateContentStream(ctx, m.name, ...)
```

Both use `m.name` instead of `req.Model`.

## Proposed fix

Add a helper that prefers `req.Model` when set, falling back to `m.name`:

```go
func (m *geminiModel) modelName(req *model.LLMRequest) string {
    if req.Model != "" {
        return req.Model
    }
    return m.name
}
```

Then use it in both call sites:

```diff
- m.client.Models.GenerateContent(ctx, m.name, req.Contents, req.Config)
+ m.client.Models.GenerateContent(ctx, m.modelName(req), req.Contents, req.Config)

- m.client.Models.GenerateContentStream(ctx, m.name, req.Contents, req.Config)
+ m.client.Models.GenerateContentStream(ctx, m.modelName(req), req.Contents, req.Config)
```

The fallback to `m.name` ensures existing behavior is preserved — `req.Model` defaults to `f.Model.Name()` via `base_flow.go:133`, which returns `m.name`. Behavior only changes when a callback explicitly overrides `req.Model`.

## Related

**Telemetry gap (separate concern):** `generateContent()` in `base_flow.go:380` also uses `m.Name()` instead of `req.Model` for the OpenTelemetry span, creating a similar mismatch in tracing. This should be addressed separately.

dmora:
Fixes #619

## Summary

- Add `modelName()` helper to `geminiModel` that prefers `req.Model` when set, falling back to the construction-time `m.name`
- Update `generate()` and `generateStream()` to use the helper instead of hardcoded `m.name`
- Add `TestModel_RespectsRequestModel` verifying both default and override behavior

## Problem

The Gemini model implementation ignores `LLMRequest.Model` and always uses `m.name` when calling the GenAI API. Since `BeforeModelCallback` can modify `req.Model`, this makes runtime model switching silently ineffective.

## Testing plan

- **Unit test added**: `TestModel_RespectsRequestModel` uses an HTTP interceptor to verify the correct model name appears in the API request URL for both cases:
  - `req.Model` empty → falls back to constructor name
  - `req.Model` set → overridden name used in API call
- **Existing tests pass**: `TestModel_Generate`, `TestModel_GenerateStream`, `TestModel_TrackingHeaders` all pass unchanged, confirming zero regression
- **Backward compatible**: The fallback to `m.name` when `req.Model` is empty means all existing code paths (which don't set `req.Model` directly) behave identically

```
$ go test ./model/gemini/... -v
--- PASS: TestModel_Generate (0.00s)
--- PASS: TestModel_GenerateStream (0.00s)
--- PASS: TestModel_TrackingHeaders (0.00s)
--- PASS: TestModel_RespectsRequestModel (0.38s)
PASS
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/model/gemini/gemini_test.go
+++ b/model/gemini/gemini_test.go
@@ -232,12 +232,7 @@ func TestModel_RespectsRequestModel(t *testing.T) {
 				Model:    tt.reqModel,
 				Contents: genai.Text("ping"),
 			}
-			for _, err := range geminiModel.GenerateContent(t.Context(), req, false) {
-				// We expect an error since we're using a fake transport.
-				// We only care about verifying the URL.
-				if err != nil {
-					t.Logf("GenerateContent error (expected): %v", err)
-				}
+			for range geminiModel.GenerateContent(t.Context(), req, false) {
 			}
 
 			if capturedURL == "" {


model/gemini/gemini.go
```diff
@@ -97,9 +97,19 @@ func (m *geminiModel) addHeaders(headers http.Header) {
...
 
+// modelName returns the model name to use for the API call.
+// It prefers req.Model (which can be set by BeforeModelCallback),
+// falling back to the construction-time name if unset.
+func (m *geminiModel) modelName(req *model.LLMRequest) string {
+	if req.Model != "" {
+		return req.Model
+	}
+	return m.name
+}
```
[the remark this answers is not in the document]
dmora: LLMRequest is an internal framework type, never deserialized from external user input. The Model field is only set by:

base_flow.go:133 — initializes to f.Model.Name() (the constructor value)
BeforeModelCallback / plugin callbacks — developer-controlled code
End users interact with the agent via session messages, not by constructing LLMRequest objects directly. The trust boundary is at the agent/runner level, well above this internal struct.

Callback authors should treat model selection as a privileged operation, but that's an application-level concern — same as how callbacks can already modify system instructions, tool configs, etc.

model/gemini/gemini_test.go
```diff
@@ -189,6 +189,67 @@ func TestModel_TrackingHeaders(t *testing.T) {
...
+				Model:    tt.reqModel,
+				Contents: genai.Text("ping"),
+			}
+			for _, err := range geminiModel.GenerateContent(t.Context(), req, false) {
+				// We expect an error since we're using a fake transport.
+				// We only care about verifying the URL.
+				if err != nil {
+					t.Logf("GenerateContent error (expected): %v", err)
+				}
+			}
```
dpasiukevich:
```suggestion
			for range geminiModel.GenerateContent(t.Context(), req, false) {
			}
```

nit: I'd just consume & exhaust the iterator here. No strong preferences here, just wanted to see WDYT.

dmora: Good call, cleaner. updated!

dpasiukevich (approved):
Thanks!

PR #624, opened 2026-03-06
fix(memory): handle punctuation and whitespace in extractWords

### LINKED ISSUE #569: Bug: Memory search fails to match words with punctuation in extractWords
### Summary
The `extractWords` function in `memory/inmemory.go` uses space-only splitting and doesn't strip punctuation, causing memory search to miss relevant results when text contains punctuation or non-space whitespace.

### Reproduction
**Scenario**: Add a session with content containing punctuation, then search for a word without punctuation.

```go
// Add session with punctuation
session := makeSession("app1", "user1", "sess1", []*session.Event{
    {
            Content: genai.NewContentFromText("The agent works great!", genai.RoleModel),
        },
    },
})
memSvc.AddSession(ctx, session)

// Search for "great" (without punctuation)
resp, _ := memSvc.Search(ctx, &memory.SearchRequest{
    AppName: "app1",
    UserID:  "user1",
    Query:   "great",
})

// Expected: 1 memory found
// Actual: 0 memories (because stored token is "great!" not "great")
```

### Root Cause
**File**: `memory/inmemory.go`, line ~155

```go
func extractWords(text string) map[string]struct{} {
	res := make(map[string]struct{})

	for s := range strings.SplitSeq(text, " ") {  // ← Only splits on space
		if s == "" {
			continue
		}
		res[strings.ToLower(s)] = struct{}{}  // ← Doesn't strip punctuation
	}

	return res
}
```

**Issues**:
1. **Space-only splitting**: `strings.SplitSeq(text, " ")` doesn't handle tabs, newlines, or multiple spaces
2. **No punctuation normalization**: `"great!"` is stored as-is, won't match `"great"`
3. **Case sensitivity handled but not enough**: Lowercasing happens after punctuation is included

### Impact
- **Search accuracy degraded**: Users searching for "error" won't find memories containing "error." or "error," or "error!"
- **Common patterns affected**:
  - Sentences ending with punctuation (`.`, `!`, `?`)
  - Comma-separated lists
  - Quoted text
  - Multi-line responses with `\n` or `\t`

### Proposed Fix
Replace space-only splitting with proper whitespace tokenization and strip punctuation:

```go
func extractWords(text string) map[string]struct{} {
	res := make(map[string]struct{})

	for _, word := range strings.Fields(text) {  // Splits on all whitespace
		// Strip punctuation
		cleaned := strings.TrimFunc(word, func(r rune) bool {
			return !unicode.IsLetter(r) && !unicode.IsNumber(r)
		})
		if cleaned == "" {
			continue
		}
		res[strings.ToLower(cleaned)] = struct{}{}
	}

	return res
}
```

**Alternative**: Use a proper tokenizer/stemmer for production-grade search, but the above fix would resolve the immediate issue.

### Test Case to Add
```go
{
    name: "match words with punctuation",
    initSessions: []session.Session{
        makeSession(t, "app1", "user1", "sess1", []*session.Event{
            {
                    Content: genai.NewContentFromText("Error: connection timeout! Please retry.", genai.RoleModel),
                },
            },
        }),
    },
    req: &memory.SearchRequest{
        AppName: "app1",
        UserID:  "user1",
        Query:   "error timeout retry",  // No punctuation
    },
    wantResp: &memory.SearchResponse{
        Memories: []memory.Entry{/* should find the memory */},
    },
},
```

### Environment
- **Version**: `main` branch (commit: latest as of 2026-02-16)
- **Go version**: 1.22+

### Additional Context
This is particularly problematic for AI agent memory since LLM responses naturally contain punctuation. The current implementation significantly reduces search recall in real-world usage.

koriyoshi2041:
## Summary

Fixes #569.

`extractWords` in `memory/inmemory.go` currently splits text only on spaces (`strings.SplitSeq(text, " ")`), which means:
- Words separated by tabs or newlines are not properly tokenized
- Punctuation attached to words (e.g., `"great!"`, `"banana,"`) prevents keyword matching

### Changes

- Replace `strings.SplitSeq(text, " ")` with `strings.Fields(text)` to handle all Unicode whitespace (tabs, newlines, multiple spaces)
- Add `strings.TrimFunc` to strip non-letter/non-number characters from word boundaries
- Add 3 test cases covering punctuation, multi-line text, and comma-separated values

## Test plan

- [x] `go test ./memory/ -run Test_inMemoryService_SearchMemory` — all 8 tests pass (5 existing + 3 new)
- [x] `go vet ./memory/` — clean


memory/inmemory.go
```diff
@@ -156,7 +157,10 @@ func checkMapsIntersect(m1, m2 map[string]struct{}) bool {
 func extractWords(text string) map[string]struct{} {
 	res := make(map[string]struct{})
 
-	for s := range strings.SplitSeq(text, " ") {
+	for _, s := range strings.Fields(text) {
```
hyangah: Can you keep using  `strings.SplitSeq`? That avoids unnecessary allocation.

koriyoshi2041: Good point\! Switched to `strings.FieldsSeq` which gives us the iterator (no slice allocation) while also splitting on all unicode whitespace (tabs, newlines, etc.) — best of both worlds.

koriyoshi2041:
Rebased this onto current `main` and cleaned up the gofmt alignment in the added memory tests. The change still uses `strings.FieldsSeq`, so it keeps the iterator/no-slice-allocation path from your review note while handling punctuation and other whitespace.

Checked locally:

```text
go test ./memory
go test -race -mod=readonly ./memory
go vet ./memory
git diff --check
```

jjsasha63:
Thanks for sticking with this, and sorry it sat so long.

We are going to take #1403 instead. It splits on runs of non-word runes rather than trimming token ends, so it also fixes the comma-joined case from #569 — `regions: us-east1,us-west1`, which this PR still indexes as a single token. I checked by swapping your `extractWords` into #1403's tests: two of its three new cases fail. It also matches adk-python's `re.findall(r'\w+', text)`, and CONTRIBUTING.md asks us to follow adk-python.

Your version does one thing better. Trimming only the ends keeps `built-in` and `don't` whole, where #1403 splits them into `built`/`in` and `don`/`t`. Those one and two-letter fragments then match across unrelated memories: on #1403, the query "What's for lunch?" returns a memory reading "The build's output is in bazel-bin", on a shared `s`. adk-python does the same, so we are accepting that.

Your switch to `strings.FieldsSeq` also still stands. We have asked on #1403 to use `strings.FieldsFuncSeq`, so the merged version keeps the allocation-free path hyangah asked for — on 2 MB of text that is 320 KB/op against 34.5 MB/op for plain `FieldsFunc`.

If #1403 stalls, we will come back to you.

PR #627, opened 2026-03-06
feat: extension points for remoteagent part conversions

nahapetyan-serob:
Problem:

A partial fix was contributed via [PR #512](https://github.com/google/adk-go/pull/512), but the agent/remoteagent package lacked the necessary extension points.

Solution:

Implemented flexible extension points using dependency injection for part converters:
Added A2APartConverter and GenAIPartConverter to remoteagent.A2AConfig, allowing per-agent customization.
Introduced ToSessionEventWithParts in the adka2a package to support custom mapping.
Refactored internal loops to prioritize custom converters when provided, while safely falling back to default logic when not.

Testing Plan

a2a_agent_test.go: Added TestRemoteAgent_CustomConverters to verify that remoteagent correctly handles custom and default converters based on configuration.
events_test.go: TestToSessionEventWithParts verifies part filtering, modification, and error propagation across all A2A event types (Messages, Tasks, and Artifacts).
Updated TestToSessionEvent with a multi-artifact task to confirm that LongRunningToolIDs are still aggregated correctly after the logic refactor.


telemetry/telemetry_test.go
```diff
@@ -233,7 +233,7 @@ func extractResourceAttributes(res *resource.Resource) (projectID, serviceName,
 			serviceVersion = attr.Value.AsString()
 		}
 	}
-	return
+	return projectID, serviceName, serviceVersion
```
yarolegovich, line 236: was it failing linter? 🤔

server/adka2a/events.go
```diff
@@ -220,9 +250,15 @@ func taskToEvent(ctx agent.InvocationContext, task *a2a.Task) (*session.Event, e
 	event := NewRemoteAgentEvent(ctx)
 
 	if task.Status.Message != nil {
-		msgParts, err := ToGenAIParts(task.Status.Message.Parts)
-		if err != nil {
-			return nil, fmt.Errorf("failed to convert status message parts: %w", err)
+		var msgParts []*genai.Part
+		for _, part := range task.Status.Message.Parts {
```
yarolegovich: this block is repeated 4 times, let's introduce a helper function

server/adka2a/events.go
```diff
@@ -63,12 +64,19 @@ func EventToMessage(event *session.Event) (*a2a.Message, error) {
 
 // ToSessionEvent converts the provided a2a event to session event authored by the agent running in the provided invocation context.
 func ToSessionEvent(ctx agent.InvocationContext, event a2a.Event) (*session.Event, error) {
+	return ToSessionEventWithParts(ctx, event, func(ctx context.Context, a2aEvent a2a.Event, part a2a.Part) (*genai.Part, error) {
+		return ToGenAIPart(part)
+	})
+}
+
+// ToSessionEventWithParts converts the provided a2a event to session event with custom part converter.
+func ToSessionEventWithParts(ctx agent.InvocationContext, event a2a.Event, partConverter A2APartConverter) (*session.Event, error) {
```
yarolegovich, line 71: let's make it default to ToGenAIPart if nil:
```go
if partConverter == nil {
  partConverter = func(ctx context.Context, a2aEvent a2a.Event, part a2a.Part) (*genai.Part, error) {
    return ToGenAIPart(part)
  }
}
```
and ToSessionEvent can delegate to it passing nil not to duplicate the func

agent/remoteagent/a2a_agent_test.go
```diff
@@ -1168,3 +1168,80 @@ func TestRemoteAgent_ErrorEventOnServerError(t *testing.T) {
...
+		t.Fatalf("convertParts() error = %v", err)
+	}
+
+	if len(parts) != 1 {
+		t.Errorf("Expected 1 part after filtering, got %d", len(parts))
+	}
+
+	for _, p := range parts {
+		if p == nil {
+			t.Fatalf("Found a nil part! This is the DoS vulnerability.")
```
yarolegovich: "got nil part, want it filtered out"

server/adka2a/events_test.go
```diff
@@ -426,6 +427,58 @@ func TestToSessionEvent(t *testing.T) {
 				Branch: branch,
 			},
 		},
+		{
+			name: "task with multiple artifacts and mixed long-running tools",
```
yarolegovich, line 431: why was this test added? don't see it in the file history before this PR

server/adka2a/events_test.go
```diff
@@ -467,3 +520,261 @@ func TestToSessionEvent(t *testing.T) {
...
+				t.Errorf("ToSessionEvent() error = %v, want nil", err)
+			}
+			if diff := cmp.Diff(tc.want, got, ignoreFields...); diff != "" {
+				t.Errorf("ToSessionEvent() wrong result (+got,-want)\ngot = %v\nwant = %v\ndiff = %s", got, tc.want, diff)
+			}
+		})
+	}
+}
+
+func TestToSessionEvent_DoSVulnerability(t *testing.T) {
```
yarolegovich: `TestToSessionEventWithParts_NilResultFiltered` 
also update `t.Fatalf("Found a nil part! ...`

server/adka2a/events_test.go
```diff
@@ -467,3 +520,261 @@ func TestToSessionEvent(t *testing.T) {
 		})
 	}
 }
+
+func TestToSessionEventWithParts(t *testing.T) {
```
yarolegovich: I think this one can just be removed. the next test verifies that part converter gets invoked, can modify parts and filter parts out, this imo is enough coverage for the functionality

telemetry/telemetry_test.go
nahapetyan-serob: `golangci-lint run --fix` made this change

server/adka2a/events_test.go
nahapetyan-serob: Since ToSessionEvent function was refactored into a wrapper, wanted to add one more complex test case.

Issue #586, opened 2026-02-20
Nil pointer dereference in base_flow.go: runconfig.FromContext returns nil


mohamed-benali:
**Describe the bug**

`runconfig.FromContext(ctx)` returns `nil` when `RunConfig` is not set as a context value, causing a nil pointer dereference in `internal/llminternal/base_flow.go:310`:

```go
useStream := runconfig.FromContext(ctx).StreamingMode == runconfig.StreamingModeSSE
```

**Root Cause**

`runconfig.ToContext()` is never called anywhere in the ADK codebase. `agent.Run()` at `agent/agent.go:181` copies `ctx.RunConfig()` into the internal `invocationContext.runConfig` struct field, but `base_flow.go` reads from the Go context via `runconfig.FromContext(ctx)` — not from the struct field. Since nobody puts the RunConfig into the context, `FromContext` always returns `nil`.

**To Reproduce**

Use `agent.Run()` with a custom `InvocationContext` implementation (not using the ADK runner). The nil dereference occurs on the first LLM call.

```go
package main

import (
	"context"
	"fmt"

	"google.golang.org/adk/agent"
	"google.golang.org/adk/agent/llmagent"
	"google.golang.org/adk/session"
	"google.golang.org/genai"
)

// Custom InvocationContext (without using runner)
type myInvocationContext struct {
	context.Context
	agent       agent.Agent
	sess        session.Session
	userContent *genai.Content
	ended       bool
}

func (c *myInvocationContext) Agent() agent.Agent          { return c.agent }
func (c *myInvocationContext) Artifacts() agent.Artifacts  { return nil }
func (c *myInvocationContext) Memory() agent.Memory        { return nil }
func (c *myInvocationContext) Session() session.Session    { return c.sess }
func (c *myInvocationContext) InvocationID() string        { return "test" }
func (c *myInvocationContext) Branch() string              { return "main" }
func (c *myInvocationContext) UserContent() *genai.Content { return c.userContent }
func (c *myInvocationContext) RunConfig() *agent.RunConfig {
	return &agent.RunConfig{StreamingMode: agent.StreamingModeNone}
}
func (c *myInvocationContext) EndInvocation()              { c.ended = true }
func (c *myInvocationContext) Ended() bool                 { return c.ended }
func (c *myInvocationContext) WithContext(ctx context.Context) agent.InvocationContext {
	newCtx := *c
	newCtx.Context = ctx
	return &newCtx
}

func main() {
	// ... create agent, session ...
[...]

PR #630, opened 2026-03-06
fix: skip nil responses in generateContent

koriyoshi2041:
## Summary

Adds nil checks for `LLMResponse` in `base_flow.go` to prevent nil pointer dereference panics when a model yields a nil response.

**Problem:** In `callLLM()` and `generateContent()`, the code accesses `resp.LLMResponse` and `resp.Content` without checking for nil. When a model returns an empty stream or yields `nil`, this causes a panic.

**Fix:**
- Added nil guard for `resp.LLMResponse` in `callLLM()` (line ~337) — skips the iteration
- Added nil guard for `resp` in `generateContent()` (line ~413) — skips nil responses

Related to #586.

## Test plan

- [x] Added `TestCallLLMNilResponse` — verifies no panic when model yields nil
- [x] `go test ./internal/llminternal/...` passes
- [x] `go vet ./...` passes

### DIFF
--- a/internal/llminternal/base_flow.go
+++ b/internal/llminternal/base_flow.go
@@ -842,6 +842,9 @@ func generateContent(ctx agent.InvocationContext, m model.LLM, req *model.LLMReq
 			// Complete the span immediately to avoid capturing the upstream yield processing time.
 			if err != nil {
 				endSpanAndTrackResult()
+			} else if resp == nil {
+				// Skip nil responses to prevent nil pointer dereference.
+				continue
 			} else if !resp.Partial {
 				// Log only final responses.
 				telemetry.LogResponse(ctx, resp, backend)
--- a/internal/llminternal/base_flow.go
+++ b/internal/llminternal/base_flow.go
@@ -935,6 +935,10 @@ func generateContent(ctx agent.InvocationContext, m model.LLM, req *model.LLMReq
 		// Ensure that the span is ended in case of error or if none final responses are yielded before the yield returns false.
 		defer endSpanAndTrackResult()
 		for resp, err := range m.GenerateContent(ctx, req, useStream) {
+			if resp == nil && err == nil {
+				// Third-party model implementations may yield an empty response.
+				continue
+			}
 			response := newResponseWithEventID(ctx, resp)
 			lastResponse = *response
 			lastErr = err
--- a/internal/llminternal/base_flow_telemetry_test.go
+++ b/internal/llminternal/base_flow_telemetry_test.go
@@ -174,6 +174,10 @@ func TestGenerateContentTracingNoFinalResponse(t *testing.T) {
 				if len(gotSpans) != 0 {
 					t.Errorf("expected 0 spans after partial response, got %d", len(gotSpans))
 				}
+
+				// A trailing empty response must not replace the partial response used
+				// when the deferred span records token usage.
+				yield(nil, nil)
 			}
 		},
 	}
--- a/internal/llminternal/base_flow_test.go
+++ b/internal/llminternal/base_flow_test.go
@@ -15,13 +15,16 @@
 package llminternal
 
 import (
+	"context"
 	"errors"
+	"iter"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/agent"
+	"google.golang.org/adk/internal/agent/runconfig"
 	icontext "google.golang.org/adk/internal/context"
 	"google.golang.org/adk/internal/toolinternal"
 	"google.golang.org/adk/model"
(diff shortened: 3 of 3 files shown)


koriyoshi2041:
Rebased onto latest `main` to resolve the conflict — it was a textual clash in `base_flow_test.go` with the newly added `TestPreprocess_Toolset`, plus `callLLM` gained an `artifactDelta` parameter upstream which the regression test now passes. Both tests pass and `go vet`/`gofmt` are clean.

koriyoshi2041:
Rebased this onto current `main` and tightened the patch based on the earlier review note: the extra `callLLM` nil guard is gone now, so the fix lives at the `generateContent` boundary and the regression test still covers the `callLLM` entry point.

Checked locally:

```text
go vet ./internal/llminternal
git diff --check
```

karolpiotrowicz (commented):
Apologies for the long silence here — this sat without a maintainer review far longer than it should have.

The change still applies. `generateContent` on `main` at [058933d](https://github.com/google/adk-go/commit/058933d583478dcae7dd5d870f366b51e3b00ae1) still reaches [`base_flow.go:944`](https://github.com/google/adk-go/blob/058933d583478dcae7dd5d870f366b51e3b00ae1/internal/llminternal/base_flow.go#L944) — `} else if !resp.Partial {` — with no nil check, and nothing fixed it in the meantime. Driving a third-party `model.LLM` that yields `(nil, nil)` through the public runner: on `main` the invocation ends with `node "a" panicked: runtime error: invalid memory address or nil pointer dereference`, and with your patch it does not. A mixed `[nil, final]` stream still delivers the final response, so the `continue` does what it should. The guard is at the right layer.

Three things before it can go in.

**Rebase.** `base_flow.go` merges cleanly, but `base_flow_test.go` conflicts — `main` has grown a lot of tests since your base and the module moved to `google.golang.org/adk/v2`. It is a plain end-of-file collision, so it should be quick.

**A trailing nil after a partial response wipes the span.** [`base_flow.go:939`](https://github.com/google/adk-go/blob/058933d583478dcae7dd5d870f366b51e3b00ae1/internal/llminternal/base_flow.go#L939) does `lastResponse = *response` before your guard runs, so a skipped nil overwrites the last good response and the deferred span-end records nothing. On a `[partial(prompt=10, candidates=20), nil]` stream the `generate_content` span comes out with empty input and output token attributes, where the same stream without the trailing nil records 10 and 20. `[final, nil]` is unaffected — a non-partial response has already closed the span — so this is specifically the streaming-then-nil shape.

The fix is to move the check above line 939, but please use the conjunct form:

```go
if resp == nil && err == nil {
    continue
}
```

The `&& err == nil` matters more than it looks. Every model in the tree reports failure as `yield(nil, err)`, so a bare `if resp == nil { continue }` hoisted above line 939 skips the error branch entirely — I ran it, and a model returning an error produced zero yields with no error reaching the caller. Your current placement does not have that problem. It is only the hoist that needs the guard on `err`.

**The test never runs its assertion.** Once the fix works `callLLM` yields nothing, so the loop body — and the only `t.Fatalf` — is dead. It would pass equally if `continue` were `return`, which would drop every response after a nil. It is a real regression test for the panic, so it earns its place, but asserting the yielded count and adding the `[nil, final]` case would pin the skip-and-keep-consuming behaviour.

Two smaller ones. The comment cites #586, but that was the `runconfig.FromContext` nil, fixed separately and already referenced in this same file at [`base_flow.go:842`](https://github.com/google/adk-go/blob/058933d583478dcae7dd5d870f366b51e3b00ae1/internal/llminternal/base_flow.go#L842) — better to describe the real trigger, since no in-tree model can produce a nil response and what this defends against is a third-party `model.LLM`. And the new `mockModel` duplicates `mockModelForTest` in the same package, so you can drop it.

One thing I checked and it is fine: the skipped nil produces a run with no events and a nil error, which looks like it hides the failure. `main` already behaves that way for a non-nil content-less response — [`base_flow.go:652`](https://github.com/google/adk-go/blob/058933d583478dcae7dd5d870f366b51e3b00ae1/internal/llminternal/base_flow.go#L652) skips those deliberately, mirroring adk-python's [`base_llm_flow.py:1484`](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/flows/llm_flows/base_llm_flow.py#L1484). Your change routes one more input into an outcome the framework already treats as normal. A log line on the skip would be a nice touch, not a blocker.

Not yours to fix, but worth knowing: the same "a `continue` bypasses the consumer's only stop signal" shape already exists at `base_flow.go:652`, and a model emitting content-less responses without bound will spin there on `main` today. I will raise that separately.

Finally, the title still says `callLLM` — since your June revision the fix lives entirely in `generateContent`.

koriyoshi2041:
Addressed all requested changes and rebased onto current `main`:

- moved the nil guard ahead of response wrapping and telemetry state updates, using `resp == nil && err == nil` so model errors still propagate;
- reused `mockModelForTest` and asserted both zero yields for `[nil]` and continued consumption for `[nil, final]`;
- added telemetry coverage for `[partial, nil]`, preserving the partial response's token attributes when the deferred span closes;
- updated the PR title to reflect the `generateContent` boundary.

Verified locally with `go test ./internal/llminternal`, `go vet ./internal/llminternal`, and `git diff --check`.

karolpiotrowicz (commented):
The revision does everything I asked, and the conjunct form is exactly right.

I re-ran the checks against the rebased head ([8613b7c](https://github.com/google/adk-go/commit/8613b7c69e09d8314bb9eb62102c66e11607494a)) on top of current `main` ([ffddb2a](https://github.com/google/adk-go/commit/ffddb2a6d7ed97974e82815d00c95eb148038f2c)). The span keeps the partial's token attributes now — a `[partial(prompt=10, candidates=20), nil]` stream records 10 and 20 where it previously recorded nothing — model errors still reach the caller, and `[nil, final]` still delivers the final response.

The part worth calling out is the test work, which is stronger than what I asked for. I mutated the guard three ways in a scratch copy and re-ran your committed suite against each:

```
guard deleted            -> TestCallLLMSkipsNilResponseAndContinues/nil_only panics
guard back in the old spot -> TestGenerateContentTracingNoFinalResponse: expected input tokens 10, got
guard reduced to bare resp == nil -> TestGenerateContentTracingError: expected span status "Error", got "Unset"
```

All three regressions are now caught by tests in the repo. In particular that last one means someone can no longer quietly simplify the condition back to `resp == nil` and swallow every model error — that was the gap I was most worried about leaving open, and appending the trailing `yield(nil, nil)` to the existing telemetry test closes it neatly.

Three small things, none of them blocking, all optional.

The new table has no error column, so the `(nil, err)` case cannot be written in it. That case is covered, but only indirectly, by the span-status assertion in `TestGenerateContentTracingError`. A `wantErr` field and a third row asserting that `callLLM` yields exactly one item carrying the error would make the guarantee direct rather than inferred from telemetry.

The `nil then final` row asserts a count of one but not that the surviving response is the one carrying `"done"`. It happens not to matter today, since the obvious wrong implementation would panic rather than yield the wrong item, but capturing the response and comparing its text would make the test say what its name claims.

The comment reads "may yield an empty response" while the guard tests nil-ness. A non-nil `&model.LLMResponse{}` trailing a partial still blanks the span — I checked, and it does that on `main` too, so it is not something this PR introduced and not yours to fix here. Saying "a nil response" would just keep the comment matched to the code.

One thing I want to be explicit about, since I raised it last time and it is still true: `continue` skips the `yield`, so a model emitting nils without bound still cannot be stopped by the consumer. I said it was not yours to fix and I stand by that — the identical shape already exists at [`base_flow.go:652`](https://github.com/google/adk-go/blob/ffddb2a6d7ed97974e82815d00c95eb148038f2c/internal/llminternal/base_flow.go#L652) and is reachable on `main` today with a more ordinary producer. I will file that separately rather than hold this up for it.

CI is green across the board, including `apidiff`. From my side this is ready.

PR #635, opened 2026-03-09
fix(examples): remove unused code in loadartifacts

mvsouza:
The loadartifacts example contained an unused call to genai.NewPartFromBytes which has been removed.


wolo-lab (approved):
LGTM, thanks!

PR #638, opened 2026-03-10
fix: resolve 'TODO: last event is not final' error in streaming mode …

### LINKED ISSUE #600: TODO: last event is not final error in a2a remote agent call

## 🔴 Required Information

**Describe the Bug:**
While using a2a remote agent with streaming mode call I often get `TODO: last event is not final` error. The code is pretty simple:

```go
	runcfg := agent.RunConfig{
		StreamingMode: agent.StreamingModeSSE,
	}
	for resp, err := range arunner.Run(ctx, userID, req.SessionID, msg, runcfg) {
		// err here is "TODO: last event is not final"
	}
```

**Expected Behavior:**
No error should be returned

**Observed Behavior:**
receiving `TODO: last event is not final` error

**Environment Details:**

 - ADK Library Version: `v0.5.0`
 - OS: Linux, macos
 - Go Version: 1.26

**Model Information:**
 - Which model is being used: `gemini-3-flash-preview`

---

## 🟡 Optional Information
*Providing this information greatly speeds up the resolution process.*

**Regression:**
Did this work in a previous version of ADK? (Yes/No) If so, which one?
No, it was happening in version `0.4.0` too

 - Often (50%+)

mahendrarathore1742:
…(#600)

When using streaming mode (e.g., with a2a remote agents), Flow.Run() would return an error 'TODO: last event is not final' if the last event from runOneStep had Partial=true. This is a legitimate state that occurs when:
- The model reaches max token limit during streaming
- A sub-agent (e.g., a2a remote agent) emits partial streaming events

The turn is complete in these cases, so we simply return instead of erroring.

Added regression test TestFlowRunPartialLastEvent with 3 sub-cases covering single partial, multiple partial, and mixed partial/final response scenarios.

Fixes #600


jjsasha63 (changes requested):
Thanks for this, and sorry it has sat so long.

The approach is right. adk-python does the same thing in the same place, so relaxing the guard rather than erroring keeps us in parity (base_llm_flow.py):

if not last_event or last_event.is_final_response() or last_event.partial:
    if last_event and last_event.partial:
        logger.warning('The last event is partial, which is not expected.')
    break
I rebased the branch onto current main and ran the suite. The production change cherry-picks cleanly and everything is green. A few things to fix before it can go in.

1. Please log a warning
Partial events are never written to the session — runner/runner.go:750 and runner/run_node.go:236 both skip AppendEvent for them. So when a turn ends on a partial, the caller sees the whole answer stream over SSE and session history keeps nothing. The next turn then shows two consecutive user messages with the assistant turn missing.

I checked this end to end, driving a real runner.Runner with the in-memory session service and a fake model that streams partials with no aggregate. Both trees persist the same thing:

yielded to caller (3):
  [0] author=assistant partial=true  text="The answer "
  [1] author=assistant partial=true  text="is "
  [2] author=assistant partial=true  text="42."
persisted in session (1):
  [0] author=user   partial=false text="what is the answer?"
That loss is already on main, so it is not something you introduced. But today the error at least tells an operator it happened; after this change there is nothing to correlate with a bug report. base_flow.go already imports log and warns for the thought-only case a few lines above, so it would fit straight in.

2. Two corrections to the comment
The turn is complete so we simply return

The turn was truncated, not completed — which is why adk-python calls it "not expected".

e.g., reached max token limit during streaming

This cannot happen with either in-tree model. aggregator.Close() runs unconditionally after the stream loop (model/gemini/gemini.go:156) and returns a non-partial response whenever any chunk arrived, including on MAX_TOKENS. model/openaimodel has the same shape. The realistic cause is a producer that ends a stream without a terminal aggregate — an a2a peer whose stream ends on an appended artifact chunk with no terminal status, for example.

3. One sub-case does not exercise the change
"partial followed by non-partial" never reaches the branch you edited: the last event is non-partial, so IsFinalResponse() returns first. I confirmed by planting a panic() in the branch and running that sub-case alone — it passes. Worth dropping or repurposing.

4. A narrowing mutation survives the test
Change the condition to this and all three sub-cases still pass:

if lastEvent.LLMResponse.Partial && lastEvent.Author == ctx.Agent().Name() {
That variant leaves #600 broken, because the issue is about a sub-agent and runOneStep forwards sub-agent events under the sub-agent's own author. Could you add a case where the trailing partial carries a different author?

5. Bound the consumer
Deleting the branch, or turning the return into a continue, is caught only when the test binary times out — a long hang that takes the whole package down. The two tests just above yours use const safetyLimit = 50 with the comment "Bound the consumer so a regression fails the test instead of hanging it." The same here would turn those into an assertion failure.

Two smaller points:

mockModelForTest in base_flow_telemetry_test.go is function-valued and does what mockStreamingLLM does, so you may be able to drop the new mock.
The branch is conflicting. base_flow.go applies cleanly; the test file needs main's version with your test appended and the import paths moved to google.golang.org/adk/v2.
Separately, and not blocking: #600 was filed against v0.5.0, and the a2a aggregation path has been rebuilt since (#599, #603, #655, #701), along with a gemini aggregator fix in #654. @vadimi-ep, could you retest on the latest release? The guard needs relaxing either way, but it would be useful to know whether the original symptom is still there.

mahendrarathore1742:
> Thanks for this, and sorry it has sat so long.
> 
> The approach is right. adk-python does the same thing in the same place, so relaxing the guard rather than erroring keeps us in parity (base_llm_flow.py):
> 
> if not last_event or last_event.is_final_response() or last_event.partial: if last_event and last_event.partial: logger.warning('The last event is partial, which is not expected.') break I rebased the branch onto current main and ran the suite. The production change cherry-picks cleanly and everything is green. A few things to fix before it can go in.
> 
> 1. Please log a warning
>    Partial events are never written to the session — runner/runner.go:750 and runner/run_node.go:236 both skip AppendEvent for them. So when a turn ends on a partial, the caller sees the whole answer stream over SSE and session history keeps nothing. The next turn then shows two consecutive user messages with the assistant turn missing.
> 
> I checked this end to end, driving a real runner.Runner with the in-memory session service and a fake model that streams partials with no aggregate. Both trees persist the same thing:
> 
> yielded to caller (3): [0] author=assistant partial=true text="The answer " [1] author=assistant partial=true text="is " [2] author=assistant partial=true text="42." persisted in session (1): [0] author=user partial=false text="what is the answer?" That loss is already on main, so it is not something you introduced. But today the error at least tells an operator it happened; after this change there is nothing to correlate with a bug report. base_flow.go already imports log and warns for the thought-only case a few lines above, so it would fit straight in.
> 
> 2. Two corrections to the comment
>    The turn is complete so we simply return
> 
> The turn was truncated, not completed — which is why adk-python calls it "not expected".
> 
> e.g., reached max token limit during streaming
> 
> This cannot happen with either in-tree model. aggregator.Close() runs unconditionally after the stream loop (model/gemini/gemini.go:156) and returns a non-partial response whenever any chunk arrived, including on MAX_TOKENS. model/openaimodel has the same shape. The realistic cause is a producer that ends a stream without a terminal aggregate — an a2a peer whose stream ends on an appended artifact chunk with no terminal status, for example.
> 
> 3. One sub-case does not exercise the change
>    "partial followed by non-partial" never reaches the branch you edited: the last event is non-partial, so IsFinalResponse() returns first. I confirmed by planting a panic() in the branch and running that sub-case alone — it passes. Worth dropping or repurposing.
> 4. A narrowing mutation survives the test
>    Change the condition to this and all three sub-cases still pass:
> 
> if lastEvent.LLMResponse.Partial && lastEvent.Author == ctx.Agent().Name() { That variant leaves #600 broken, because the issue is about a sub-agent and runOneStep forwards sub-agent events under the sub-agent's own author. Could you add a case where the trailing partial carries a different author?
> 
> 5. Bound the consumer
>    Deleting the branch, or turning the return into a continue, is caught only when the test binary times out — a long hang that takes the whole package down. The two tests just above yours use const safetyLimit = 50 with the comment "Bound the consumer so a regression fails the test instead of hanging it." The same here would turn those into an assertion failure.
> 
> Two smaller points:
> 
> mockModelForTest in base_flow_telemetry_test.go is function-valued and does what mockStreamingLLM does, so you may be able to drop the new mock. The branch is conflicting. base_flow.go applies cleanly; the test file needs main's version with your test appended and the import paths moved to google.golang.org/adk/v2. Separately, and not blocking: #600 was filed against v0.5.0, and the a2a aggregation path has been rebuilt since (#599, #603, #655, #701), along with a gemini aggregator fix in #654. @vadimi-ep, could you retest on the latest release? The guard needs relaxing either way, but it would be useful to know whether the original symptom is still there.

    Thanks for the review @jjsasha63! I've updated the PR to address all your feedback:

    1. **Warning log**: Replaced the error with a warning log (`log.Printf`) to match parity with `adk-python`.
    2. **Comment fix**: Corrected the comment to clarify that the turn was truncated (not completed) and mentioned the a2a stream
  case instead of max tokens.
    3. **Dropped sub-case**: Dropped the unnecessary 'partial followed by non-partial' test sub-case.
    4. **Sub-agent test**: Added a new test case specifically checking a trailing partial with a sub-agent author.
    5. **Safety limit**: Bound the test loops with `safetyLimit = 50` so any regression fails properly.

    I also dropped the redundant `partialOnlyModel` mock, rebased the branch on `main` to resolve the merge conflicts, and
  ensured all the test file imports are using `google.golang.org/adk/v2`. Let me know if everything looks good now!

  Just drop that into the comment box and you should be good to go! Let me know if you need anything else.

jjsasha63:
Much better, thanks. I re-ran the mutations against the new head and three of the four now fail as they should:

| mutation | caught |
| --- | --- |
| restore the original `yield(nil, fmt.Errorf(...))` | yes, all three sub-cases |
| narrow to `&& lastEvent.Author == ctx.Agent().Name()` | yes, by the new sub-agent case |
| `return` → `continue` | yes, `safetyLimit` fails it in 0.03s instead of hanging the package |
| delete the `log.Printf` | no |

`gofmt`, `go vet` and `go build` are clean, and `./internal/... ./runner/... ./agent/...` is green.

Two things left.

### The warning names the wrong producer

`f.Model.Name()` attributes the partial to the model, but the point of the sub-agent case is that the model did not produce it. Your own new sub-case shows the mismatch: it sets no `modelResponses`, so the model yields nothing, yet the log says

```
adk: model "test-model" yielded a partial final event for agent "test-agent" (invocation "e-...")
```

`lastEvent.Author` is the honest field:

```go
log.Printf("adk: agent %q (invocation %q): step ended on a partial event from %q; the producer did not close its stream with an aggregated final event, so the turn will not appear in session history",
	ctx.Agent().Name(), ctx.InvocationID(), lastEvent.Author)
```

I ran that against your branch — it prints `from "sub-agent"` for the sub-agent sub-case and `from "test-agent"` for the two model ones. Phrase it however you like; the ask is just that the message not name a producer that did not produce the event.

### Nothing asserts the warning

Deleting the `log.Printf` leaves all three sub-cases passing, so it is the one mutation that still survives. Since the warning is the reason we are comfortable returning rather than erroring, it is worth pinning: an assertion that one line is logged, naming the event author, would close the last gap.

There is a `captureLog` helper at `agent/callback_context_wrapper_test.go:35` that does the `log.SetOutput` dance with cleanup. It is in a different package so you cannot import it, but it is about a dozen lines to copy.

The branch is behind main but no longer conflicting, so a merge from main before the final push should be all that is needed.

mahendrarathore1742:
> Much better, thanks. I re-ran the mutations against the new head and three of the four now fail as they should:
> 
> mutation	caught
> restore the original `yield(nil, fmt.Errorf(...))`	yes, all three sub-cases
> narrow to `&& lastEvent.Author == ctx.Agent().Name()`	yes, by the new sub-agent case
> `return` → `continue`	yes, `safetyLimit` fails it in 0.03s instead of hanging the package
> delete the `log.Printf`	no
> `gofmt`, `go vet` and `go build` are clean, and `./internal/... ./runner/... ./agent/...` is green.
> 
> Two things left.
> 
> ### The warning names the wrong producer
> `f.Model.Name()` attributes the partial to the model, but the point of the sub-agent case is that the model did not produce it. Your own new sub-case shows the mismatch: it sets no `modelResponses`, so the model yields nothing, yet the log says
> 
> ```
> adk: model "test-model" yielded a partial final event for agent "test-agent" (invocation "e-...")
> ```
> 
> `lastEvent.Author` is the honest field:
> 
> ```go
> log.Printf("adk: agent %q (invocation %q): step ended on a partial event from %q; the producer did not close its stream with an aggregated final event, so the turn will not appear in session history",
> 	ctx.Agent().Name(), ctx.InvocationID(), lastEvent.Author)
> ```
> 
> I ran that against your branch — it prints `from "sub-agent"` for the sub-agent sub-case and `from "test-agent"` for the two model ones. Phrase it however you like; the ask is just that the message not name a producer that did not produce the event.
> 
> ### Nothing asserts the warning
> Deleting the `log.Printf` leaves all three sub-cases passing, so it is the one mutation that still survives. Since the warning is the reason we are comfortable returning rather than erroring, it is worth pinning: an assertion that one line is logged, naming the event author, would close the last gap.
> 
> There is a `captureLog` helper at `agent/callback_context_wrapper_test.go:35` that does the `log.SetOutput` dance with cleanup. It is in a different package so you cannot import it, but it is about a dozen lines to copy.
> 
> The branch is behind main but no longer conflicting, so a merge from main before the final push should be all that is needed.

Thanks for the review, @jjsasha63! I've pushed an update addressing the remaining points:

1. **Warning log producer** — updated the warning log to use `lastEvent.Author` instead of hardcoding the model, using your suggested message.
2. **Test assertions** — brought in the `captureLog` helper and added assertions verifying the warning is actually logged and correctly identifies the event author, across all three sub-cases.
3. **Merge** — merged `main` into the branch so it's fully up to date.

Tests and linting are all green. Let me know if there's anything else!

jjsasha63 (approved):
Thanks for turning these around so quickly.

I re-ran the mutation checks against the new head and all five now fail as they should:

| mutation | sub-cases failed |
| --- | --- |
| delete the `log.Printf` | 3 |
| attribute the warning to `f.Model.Name()` again | 3 |
| restore the original `yield(nil, fmt.Errorf(...))` | 3 |
| narrow to `&& lastEvent.Author == ctx.Agent().Name()` | 1, the sub-agent case |
| `return` → `continue` | 3, via `safetyLimit` |

`gofmt`, `go vet` and `go build` are clean, and `./internal/... ./runner/... ./agent/...` is green under `-race`.

One correction to your last note: the branch has not been merged with main, it is still five commits behind. Nothing is at risk — none of those five touch `llminternal` or `runner`, and I merged main into the head locally to be sure, which gave no conflicts and a green suite. I will bring the branch up to date from my side.

LGTM.

PR #644, opened 2026-03-12
ToolConfirmations with Vertex AI sessions

radhus:
**No issue exists, describe the change:**

**Problem:**

Using the `ctx.RequestConfirmation()` inside a tool doesn't work with Vertex AI Session storage. The error returned when using that is:
```
failed to add event to session: failed to append event: error creating content: failed to convert function call to structpb: proto:\u00a0invalid type: *genai.FunctionCall
```

It seems like it tries to store an actual pointer to an `genai.FunctionCall` on the session, which is not a serializable type to pass along.

The same happens with `toolconfirmation.ToolConfirmation`:
```
failed to add event to session: failed to append event: error creating content: failed to convert function call to structpb: proto:\u00a0invalid type: toolconfirmation.ToolConfirmation
```

Using confirmations together with Vertex AI is however mentioned in the [ADK docs known limitations section](https://google.github.io/adk-docs/tools-custom/confirmation/#known-limitations).

**Solution:**

Two bugs found and fixed during debugging:

1. Function call IDs are not restored when a session is restored from Vertex AI (first commit)
2. The `*genai.FunctionCall` needs to be converted to a serializable type, fixed using `converters.ToMapStructure`
3. The `toolconfirmation.ToolConfirmation` needs to be converted to a serializable type, fixed using `converters.ToMapStructure`

### Testing Plan

**Unit Tests:**

Build an agent which uses some kind of confirmation (or just HITL) tools and validate that both being presented that the tool is running works, and that the confirmation answer is received.

### Checklist

### Additional context

I am not super familiar with the code base, and I'm not sure if this change potentially breaks other usages, but I couldn't find any indication of that.

Also not sure if this should be handled in sync with similar changes to adk-python.

### DIFF
--- a/internal/llminternal/base_flow.go
+++ b/internal/llminternal/base_flow.go
@@ -203,7 +203,11 @@ func (f *Flow) runOneStep(ctx agent.InvocationContext) iter.Seq2[*session.Event,
 				continue
 			}
 
-			toolConfirmationEvent := generateRequestConfirmationEvent(ctx, modelResponseEvent, ev)
+			toolConfirmationEvent, err := generateRequestConfirmationEvent(ctx, modelResponseEvent, ev)
+			if err != nil {
+				yield(nil, err)
+				return
+			}
 			if toolConfirmationEvent != nil {
 				if !yield(toolConfirmationEvent, nil) {
 					return
--- a/internal/llminternal/functions.go
+++ b/internal/llminternal/functions.go
@@ -15,9 +15,12 @@
 package llminternal
 
 import (
+	"fmt"
+
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/agent"
+	"google.golang.org/adk/internal/converters"
 	"google.golang.org/adk/internal/utils"
 	"google.golang.org/adk/model"
 	"google.golang.org/adk/session"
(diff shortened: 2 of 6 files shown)


odsod:
Bumping this PR - we were considering migrating to the Vertex AI session storage but are now holding off since this would affect us as well.

radhus:
@dpasiukevich can you take a look? Any hint in what direction I can go to get this closer to getting fixed is appreciated!

radhus:

wolo-lab:
Thank you for this, and I am sorry. You tagged two maintainers, @odsod bumped the thread saying it was blocking a migration, and none of us replied. That was a real failure on our side.

Your problem is fixed on main, in two pieces. Function call IDs are now preserved when loading a Vertex AI session, from #762. And the "proto: invalid type: *genai.FunctionCall" failure was resolved one layer down by #793, which replaced the structpb.NewStruct calls in the Vertex client with a JSON round-trip, so any value carrying JSON tags serializes correctly. We ran the exact event shape you reported against current main and it serializes.

Worth checking against main before re-testing from scratch.

We are reviewing the queue much more actively now, and a thread with two people asking for help will not go unanswered again.

PR #651, opened 2026-03-16
cleanup: update Loader documentation and add interface checks

hyangah:
- Fix documentation of Loader and its implementations.
- Add TODO for future refactoring - we are planning to move loader to a separate package.

Note: This change does not address the internal api cleanup bug yet (b/478885381) but
is related to the codebase improvements discussed there.


karolpiotrowicz:
Thanks @hyangah — nice cleanup and it still applies. Only thing holding it up is the failing `lint` check; could you fix that and then we'll merge?

wolo-lab (approved):
LGTM

PR #427, opened 2025-12-17
Fix/graceful tool error handling #423

sheing-google:
Addressing issue: #423 

This commit fixes a nil pointer dereference in the loopagent that occurred when an LLM called a tool that was not registered with the current agent. The agent now gracefully handles the error and continues execution instead of crashing.

Test: Passed.

PR #658, opened 2026-03-17
Fix: gracefully handle sub-agent errors in loopagent

### LINKED ISSUE #423: Gracefully handle invalid tool calls instead of crashing
**Describe the bug**

When an LLM calls a tool that is not registered with the current agent, the ADK crashes with a nil pointer dereference instead of gracefully handling the error. The `loopagent` accesses `event.Actions.Escalate` without checking if `event` is nil after an error is returned.

**To Reproduce**

Minimal reproduction code:

```go
package main

import (
	"context"
	"fmt"

	"google.golang.org/adk/agent"
	"google.golang.org/adk/agent/workflowagents/loopagent"
	"google.golang.org/adk/model/gemini"
	"google.golang.org/adk/tool"
	"google.golang.org/genai"
)

func main() {
	ctx := context.Background()

	// Create a simple tool
	readFile, _ := functiontool.New(
		functiontool.Config{
			Name:        "read_file",
			Description: "Read a file",
		},
		func(ctx tool.Context, args struct{ Path string }) (string, error) {
			return "file content", nil
		},
	)

	// Create model
	model, _ := gemini.New(ctx, "gemini-2.0-flash")

	// Create agent with only read_file tool
	// But system prompt mentions "apply_change" which doesn't exist
	myAgent, _ := llmagent.New(llmagent.Config{
		Name:        "test-agent",
		Description: "Test agent",
		Model:       model,
		Instruction: `You have access to: read_file and apply_change tools.
Use apply_change to modify files.`, // Mentions non-existent tool
		Tools: []tool.Tool{readFile}, // Only read_file is registered
	})

	// Wrap in loop agent
	loopAgent := loopagent.New(loopagent.Config{
		MaxIterations: 5,
		SubAgents:     []agent.Agent{myAgent},
	})

	// Create runner
	sessionSvc := session.InMemoryService()
	r, _ := runner.New(runner.Config{
		AppName:        "test",
		Agent:          loopAgent,
		SessionService: sessionSvc,
	})

	sessionSvc.Create(ctx, &session.CreateRequest{
		AppName:   "test",
		UserID:    "user",
		SessionID: "session1",
	})

	message := &genai.Content{
		Role:  "user",
		Parts: []*genai.Part{genai.NewPartFromText("Please modify the config file")},
	}

	// This will panic when LLM tries to call "apply_change"
	for event, err := range r.Run(ctx, "user", "session1", message, agent.RunConfig{}) {
		if err != nil {
			fmt.Printf("Error: %v\n", err)
		}
		if event != nil {
			fmt.Printf("Event: %s\n", event.Author)
		}
	}
}
```

Steps to reproduce:
1. Install ADK: `go get google.golang.org/adk@v0.2.0`
2. Run the above code
3. When the LLM generates a function call for `apply_change` (which is mentioned in the prompt but not registered), the application panics

Error/stacktrace:
```
[ERROR] unknown tool: "apply_change"
[signal SIGSEGV code=0x1 addr=0xf0 pc=0x7ff683ffb421]

goroutine 1 [running]:
google.golang.org/adk/agent/workflowagents/loopagent.(*loopAgent).Run-fm.(*loopAgent).Run.func1-range1(0x0, {0x7ff6849d9b60?, 0xc000b9fd40?})
        google.golang.org/adk@v0.2.0/agent/workflowagents/loopagent/agent.go:88 +0x61
google.golang.org/adk/agent.(*agent).Run.func1-range1(0xc000191c80?, {0x7ff6849d9b60?, 0xc000b9fd40?})
        google.golang.org/adk@v0.2.0/agent/agent.go:189 +0xd2
google.golang.org/adk/agent/llmagent.(*llmAgent).run.func1-range1(0x0, {0x7ff6849d9b60, 0xc000b9fd40})
        google.golang.org/adk@v0.2.0/agent/llmagent/llmagent.go:345 +0x8f
...
```

**Expected behavior**

When an LLM calls an unknown/unregistered tool, the ADK should:
1. Return an error response to the LLM indicating the tool doesn't exist
2. Allow the LLM to retry with a valid tool
3. Continue agent execution normally (not crash)

**Screenshots**

N/A

- OS: Ubuntu 22.04.3 LTS (WSL2, kernel 6.6.87.2-microsoft-standard-WSL2)
- Go version: go1.25.1 linux/amd64
- ADK version: v0.2.0

**Model Information:**
- gemini-2.0-flash (but issue is model-agnostic - happens with any model that calls an unknown tool)

**Additional context**

**Root Cause:**

The bug is in `agent/workflowagents/loopagent/agent.go` at line 88:

```go
    // TODO: ensure consistency -- if there's an error, return and close iterator...
    if !yield(event, err) {
        return
    }

    if event.Actions.Escalate {  // LINE 88 - PANIC: event is nil when err != nil
        shouldExit = true
    }
}
```

When `handleFunctionCalls()` in `base_flow.go` detects an unknown tool, it returns `(nil, error)`. The error is yielded correctly, but then the code tries to access `event.Actions.Escalate` without checking if `event` is nil.

Note: There's already a TODO comment on line 83 acknowledging this needs to be fixed.

**Suggested Fix:**

```go
    if !yield(event, err) {
        return
    }
    
    if err != nil {
        shouldExit = true
        break
    }
    
    if event != nil && event.Actions.Escalate {
        shouldExit = true
    }
}
```

**Impact:**

This affects any application where:
- LLMs might hallucinate tool names
- Multi-agent systems have different tools per agent  
- System prompts mention tools that aren't always registered

The crash is unrecoverable and terminates the entire agent execution.

lucent1:

- Closes: #423 
- Related: #427 

**Problem:**
_When a sub-agent yields an error in `loopagent`, the error was silently swallowed. `shouldExit` was set to `true` but the error was never propagated to the caller, making it impossible to distinguish between a normal escalation and a sub-agent failure. This caused a nil pointer dereference panic upstream._

**Solution:**
_Return early and yield the error to the caller when a sub-agent returns a non-nil error. This is consistent with how `Flow.Run` handles errors in `llmagent` and follows the TODO comment left in the original code._

### Testing Plan
**Unit Tests:**

_Added a new test case `loop agent which always yields an error` that creates a sub-agent that returns an error and verifies the loop agent propagates it correctly without panicking._

_The original panic can be reproduced by registering an unknown tool name in the prompt. After this fix the agent returns an error instead of crashing._

### Checklist

### Additional Context
Alternative approach to #427 which has been inactive. This fix also propagates the error to the caller rather than swallowing it.

### DIFF
--- a/agent/workflowagents/loopagent/agent_test.go
+++ b/agent/workflowagents/loopagent/agent_test.go
@@ -45,6 +45,7 @@ func TestNewLoopAgent(t *testing.T) {
 		args       args
 		wantEvents []*session.Event
 		wantErr    bool
+		wantRunErr bool
 	}{
 		{
 			name: "infinite loop",
@@ -197,6 +198,15 @@ func TestNewLoopAgent(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "loop agent which always yields an error",
+			args: args{
+				maxIterations: 1,
+				subAgents:     []agent.Agent{newErrorAgent(t)},
+			},
+			wantEvents: nil,
+			wantRunErr: true,
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
@@ -238,7 +248,10 @@ func TestNewLoopAgent(t *testing.T) {
 
 			for event, err := range agentRunner.Run(ctx, "user_id", "session_id", genai.NewContentFromText("user input", genai.RoleUser), agent.RunConfig{}) {
 				if err != nil {
-					t.Errorf("got unexpected error: %v", err)
+					if !tt.wantRunErr {
+						t.Errorf("got unexpected error: %v", err)
+					}
+					break
 				}
 
 				if tt.args.maxIterations == 0 && len(gotEvents) == len(tt.wantEvents) {
@@ -375,4 +388,20 @@ func (f *FakeLLM) GenerateContent(ctx context.Context, req *model.LLMRequest, st
 			}
 		}
 	}
+}
+
+func newErrorAgent(t *testing.T) agent.Agent {
+	t.Helper()
+	a, err := agent.New(agent.Config{
+		Name: "error_agent",
+		Run: func(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
+			return func(yield func(*session.Event, error) bool) {
+				yield(nil, fmt.Errorf("something went wrong"))
+			}
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	return a
 }
--- a/agent/workflowagents/loopagent/agent.go
+++ b/agent/workflowagents/loopagent/agent.go
@@ -80,7 +80,11 @@ func (a *loopAgent) Run(ctx agent.InvocationContext) iter.Seq2[*session.Event, e
 			shouldExit := false
 			for _, subAgent := range ctx.Agent().SubAgents() {
 				for event, err := range subAgent.Run(ctx) {
-					// TODO: ensure consistency -- if there's an error, return and close iterator, verify everywhere in ADK.
+					if err != nil {
+						yield(nil, err)
+						return
+					}
+
 					if !yield(event, err) {
 						return
 					}


agent/workflowagents/loopagent/agent_test.go
```diff
@@ -197,6 +198,15 @@ func TestNewLoopAgent(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "loop agent which always yields an error",
+			args: args{
+				maxIterations: 1,
```
karolpiotrowicz, line 204: This case passes unchanged on `main` with the `agent.go` hunk reverted, so it does not exercise the fix. `maxIterations: 1` means the old loop also stops after one pass, and `wantEvents: nil` reduces the only assertion that runs to `0 == 0`.

The behavior worth pinning is that a failing sub-agent is invoked *once*. A case that fails before the change and passes after it looks like this: `maxIterations: 5`, an error agent that increments a counter each time its `Run` is entered, a consumer that keeps ranging, and assertions that exactly one error reached the caller and the sub-agent ran exactly once. Against `main` that yields 5 and 5.

While you are here, the pair a sub-agent can legally yield — a non-nil event *and* a non-nil error — has no coverage at all, which is what let the dropped-event problem through.

agent/workflowagents/loopagent/agent_test.go
```diff
@@ -238,7 +248,10 @@ func TestNewLoopAgent(t *testing.T) {
 
 			for event, err := range agentRunner.Run(ctx, "user_id", "session_id", genai.NewContentFromText("user input", genai.RoleUser), agent.RunConfig{}) {
 				if err != nil {
-					t.Errorf("got unexpected error: %v", err)
+					if !tt.wantRunErr {
```
karolpiotrowicz, line 251: `wantRunErr` only suppresses the `t.Errorf`, so nothing ever asserts that an error actually arrived. A regression that swallowed the error entirely would leave `gotEvents` empty and still pass the length check below.

Recording a `gotRunErr` flag in this branch and comparing it to `tt.wantRunErr` after the loop would close that. Asserting on the error text as well would stop an unrelated failure — one from `runner.New`, from `sessionService.Create`, or from the runner's own `failed to add event to session` — from satisfying the case.

agent/workflowagents/loopagent/agent_test.go
```diff
@@ -238,7 +248,10 @@ func TestNewLoopAgent(t *testing.T) {
...
+						t.Errorf("got unexpected error: %v", err)
+					}
+					break
```
karolpiotrowicz, line 254: This `break` is why the new case cannot tell the two behaviors apart: it stops the consumer at the first error under both, whereas `runner.Run` yields the error and keeps going.

It also applies to the five pre-existing cases, which do not want it. When one of those hits an unexpected error it now reports the error, breaks, and then trips the `Unexpected event length, got: 0, want: N` fatal below, so the per-event `cmp.Diff` that would have shown what actually differed never runs. Making the `break` conditional on `tt.wantRunErr` keeps the old diagnostics.

karolpiotrowicz (commented):
The behavior this adds is right and it closes a defect `main` still has, but it should not go in as written, and it should not close #423.

## One thing needs fixing before merge

`yield(nil, err)` at [agent.go:83-86](https://github.com/google/adk-go/blob/a93706681e7477e4d3064ae77cf24d55a87908b3/agent/workflowagents/loopagent/agent.go#L83-L86) hard-codes `nil` in place of the event, so a sub-agent that yields an event *and* an error on the same call has that event silently discarded. `iter.Seq2[*session.Event, error]` permits the pair and `agent.Config.Run` is exported, so application code can produce it — [`a2a_agent.go:281-286`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/agent/remoteagent/v2/a2a_agent.go#L281-L286) forwards both unmodified when a user's `BeforeA2ARequestCallback` returns both. The event is not recoverable anywhere downstream: [`runner.go:713-719`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/runner/runner.go#L713-L719) takes the error branch before it ever appends to the session, so the caller's `yield` was the only place it could surface.

Terminating the loop does not require dropping the payload. Deleting the duplicated yield gets both, and I ran it — the existing loopagent tests stay green and the error still stops the loop after one sub-agent invocation:

```go
    if !yield(event, err) {
        return
    }
    if err != nil {
        return
    }

    if event != nil && event.Actions.Escalate {
        shouldExit = true
    }
}
```

<details>
<summary>Reproduction — <code>go test -run TestLoopAgentForwardsEventAlongsideError ./agent/workflowagents/loopagent/...</code> passes on <code>main</code> and fails with this PR applied</summary>

```go
func TestLoopAgentForwardsEventAlongsideError(t *testing.T) {
	ctx := context.Background()

	sub, err := agent.New(agent.Config{
		Name: "partial_failure_agent",
		Run: func(ictx agent.InvocationContext) iter.Seq2[*session.Event, error] {
				ev := session.NewEvent(ictx, ictx.InvocationID())
				ev.Author = "partial_failure_agent"
				ev.LLMResponse = model.LLMResponse{
					Content: genai.NewContentFromText("partial output", genai.RoleModel),
				}
				yield(ev, fmt.Errorf("failed after emitting partial output"))
			}
		},
	})
	if err != nil {
		t.Fatal(err)
	}

	loopAgent, err := loopagent.New(loopagent.Config{
		MaxIterations: 1,
		AgentConfig: agent.Config{
			Name:      "test_agent",
			SubAgents: []agent.Agent{sub},
		},
	})
	if err != nil {
		t.Fatal(err)
	}

		AppName:        "test_app",
		Agent:          loopAgent,
	})
	if err != nil {
		t.Fatal(err)
	}
	if _, err := sessionService.Create(ctx, &session.CreateRequest{
		AppName:   "test_app",
		UserID:    "user_id",
		SessionID: "session_id",
	}); err != nil {
		t.Fatal(err)
	}

	sawEventWithError := false
	for ev, err := range agentRunner.Run(ctx, "user_id", "session_id",
		genai.NewContentFromText("user input", genai.RoleUser), agent.RunConfig{}) {
		if err != nil && ev != nil {
			sawEventWithError = true
		}
	}

	if !sawEventWithError {
		t.Errorf("the event accompanying the error was not forwarded to the caller")
	}
}
```

```
before the fix   yielded: event=true  err=failed after emitting partial output   PASS
with the fix     yielded: event=false err=failed after emitting partial output   FAIL
```

</details>

## This does not close #423, and the description does not match `main`

Both halves of #423 were already fixed before this branch was cut, so the `Closes: #423` line should come out.

- **The nil dereference is guarded on `main`.** [`agent.go:88`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/agent/workflowagents/loopagent/agent.go#L88) reads `if event != nil && event.Actions.Escalate`, added in [the temp-state-deltas change](https://github.com/google/adk-go/commit/7535be35f537aa2a12446bd74f2e3e8d7b4f39b8) (#537, 6 February). That line is diff context here, not an addition.
- **The unregistered-tool crash is gone too.** [`base_flow.go:1150-1155`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/internal/llminternal/base_flow.go#L1150-L1155) turns a tool-not-found into a result that is wrapped as a `FunctionResponse` and sent back to the model, added in [the plugin-package change](https://github.com/google/adk-go/commit/781f76da7ca29c2612c62bded7fd7b1e5572bc0f) (#480, 20 January). That path no longer produces an error at all, so it never reaches the loop agent.
- **The stated problem is not what `main` does.** The description says the error "was silently swallowed" and that "`shouldExit` was set to `true`". Neither is true of the base code — it yielded `(event, err)` straight through, and set `shouldExit` only on escalation.

What the change *does* fix is worth stating in its place: on `main` a persistently failing sub-agent is re-driven on every pass. With `MaxIterations: 5` and a consumer that keeps ranging, `main` invokes the failing sub-agent 5 times and emits the same error 5 times, and with `MaxIterations: 0` it never stops. This PR makes both one.

## Other things worth settling

**`sequentialagent` is left behind.** The deleted TODO said "verify everywhere in ADK", and the identical one still sits at [`sequentialagent/agent.go:84`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/agent/workflowagents/sequentialagent/agent.go#L84). Concretely, `sequentialagent{A, B}` where A errors still runs B on its own, but no longer does when it is nested under a loop agent. Worth either doing both here or saying why loop-only is intended.

**The doc comment now understates the exit conditions.** [`agent.go:33-34`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/agent/workflowagents/loopagent/agent.go#L33-L34) still says a `MaxIterations == 0` loop "runs indefinitely or until any sub-agent escalates". There is now a third terminator.

**A failing sub-agent's `AfterAgentCallbacks` stop running.** Returning from inside `range subAgent.Run(ctx)` makes the sub-agent's `yield` return false, so its wrapper returns before reaching [`agent/agent.go:207`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/agent/agent.go#L207). Measured with a sub-agent carrying one before- and one after-callback: `main` gives `before=1 after=1`, this PR gives `before=1 after=0`. This is not new to loop agents in general — [`parallelagent/agent.go:146-148`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/agent/workflowagents/parallelagent/agent.go#L146-L148) already abandons the iterator the same way and gives `before=1 after=0` on unmodified `main` — and the root cause is that `runAfterAgentCallbacks` is not deferred, which belongs in its own issue rather than here. Flagging it so the change ships knowingly rather than by accident.

**Existing callers lose a retry they may be relying on.** A refinement loop whose critic hits a transient 429 previously surfaced the error and tried again next iteration. It now stops. That is the point of the change, but it reaches YAML-declared apps through `internal/configurable` with no way to opt out, so it is worth a line in the release notes.

The branch is behind `main` by about five months and will need a rebase before it can land.

PR #660, opened 2026-03-18
refactor: remove launcher dependency from adkrest

dpasiukevich:
Launcher package may be revised in future (for example in favor of ADK App).
Right now we need to make sure that we don't import the launcher as a config type in critical packages, e.g. `server` packages.

BREAKING CHANGE: updated adkrest package, replaced NewHandler with NewServer constructor.


pawel-maciejczek:
LGTM

PR #670, opened 2026-03-20
fix: use []any instead of []string in loadartifactstool to fix structpb serialization

### LINKED ISSUE #668: `loadartifactstool` breaks Vertex AI session persistence due to `[]string` in function response
## Required Information

**Describe the Bug:**

`loadartifactstool.Run` returns `[]string` in its response map, which causes `structpb.NewStruct` to fail when the Vertex AI session service persists the function response event. `structpb` only supports `[]any`, not typed slices like `[]string`.

**Steps to Reproduce:**

1. Register `loadartifactstool.New()` on an agent
2. Register a separate tool that saves artifacts via `toolCtx.Artifacts().Save()`
3. Configure the runner with a Vertex AI session service (`session/vertexai`)
4. Send a message that causes the agent to call the artifact-saving tool, then `load_artifacts`
5. Observe the error when the `load_artifacts` function response event is persisted:

```
AGENT_ERROR: failed to add event to session: failed to append event: error creating content: failed to convert function response to structpb: proto: invalid type: []string
```

**Expected Behavior:**

The `load_artifacts` function response event should persist successfully to the Vertex AI session service, and the agent should continue with the loaded artifact content.

**Observed Behavior:**

The invocation fails when `structpb.NewStruct` is called on the function response map containing `[]string` ([`vertexai_client.go` line 503](https://github.com/google/adk-go/blob/da1d6a560463aba60c28bee44ab1ece320c29c83/session/vertexai/vertexai_client.go#L503)).

The root cause is in `loadartifactstool.Run` ([lines 112-114](https://github.com/google/adk-go/blob/da1d6a560463aba60c28bee44ab1ece320c29c83/tool/loadartifactstool/load_artifacts_tool.go#L112-L114)):

```go
result := map[string]any{
    "artifact_names": artifactNames, // artifactNames is []string
}
return result, nil
```

`structpb.NewValue` (called internally by `structpb.NewStruct`) has a [type switch](https://github.com/protocolbuffers/protobuf-go/blob/96a179180f0ad6bba9b1e7b6e38d0affb0168e9a/types/known/structpb/struct.pb.go#L364-L371) that handles `[]any` but not `[]string` — any unmatched type falls through to the `default` case which returns `"invalid type: %T"`.

**Environment Details:**

- ADK Library Version: v0.6.0 (also present in v0.5.0)
- OS: macOS (Darwin 25.3.0)
- Go Version: 1.25

**Model Information:**

- N/A (the bug is in tool response serialization, not model-specific)

---

## Optional Information

**Regression:**

Unknown.

Always (100%) — any `load_artifacts` call triggers it when using the Vertex AI session service.

```go
package main

import (
	"fmt"
	"google.golang.org/protobuf/types/known/structpb"
)

func main() {
	// Exactly what loadartifactstool.Run returns:
	result := map[string]any{
		"artifact_names": []string{"a.pdf", "b.pdf"},
	}
	_, err := structpb.NewStruct(result)
	fmt.Println(err) // proto: invalid type: []string

	// The fix — use []any instead of []string:
	fixed := map[string]any{
		"artifact_names": []any{"a.pdf", "b.pdf"},
	}
	_, err = structpb.NewStruct(fixed)
	fmt.Println(err) // <nil>
}
```

**Additional Context:**

Suggested fix for `loadartifactstool.Run` — return `[]any` instead of `[]string`:

```go
anyNames := make([]any, len(artifactNames))
for i, name := range artifactNames {
    anyNames[i] = name
}
result := map[string]any{
    "artifact_names": anyNames,
}
```

There is a secondary issue: even if `Run` is fixed to return `[]any`, `processLoadArtifactsFunctionCall` ([line 176](https://github.com/google/adk-go/blob/da1d6a560463aba60c28bee44ab1ece320c29c83/tool/loadartifactstool/load_artifacts_tool.go#L176)) does a hard type assertion to `[]string`. After a `structpb` round-trip (persist then reload), the value comes back as `[]interface{}`, so that assertion would also fail. It should handle both `[]string` and `[]any`.

brucearctor:
## Description

Fixes #668

`loadartifactstool.Run` returns `[]string` in its response map, which causes `structpb.NewStruct` to fail when the Vertex AI session service persists the function response event. `structpb` only supports `[]any`, not typed slices like `[]string`.

### `tool/loadartifactstool/load_artifacts_tool.go`

1. **`Run()`**: Convert `[]string` to `[]any` before returning in the result map, ensuring compatibility with `structpb.NewStruct`.

2. **`processLoadArtifactsFunctionCall()`**: Replace the hard `[]string` type assertion with a type switch that handles both `[]string` (direct call) and `[]any` (after a `structpb` round-trip where `[]string` becomes `[]interface{}`).

### `tool/loadartifactstool/load_artifacts_tool_test.go`

- Updated existing test expectations from `[]string` to `[]any` to match the new return type.
- Added `TestLoadArtifactsTool_Run_StructpbCompatibility`: verifies that `Run()` output can be serialized via `structpb.NewStruct` (the core bug assertion).
- Added `TestLoadArtifactsTool_ProcessRequest_Artifacts_LoadArtifactsFunctionCall_AnySlice`: verifies `ProcessRequest` handles `[]any` in the function response (simulating a round-trip).

### Unit Tests
```
go test ./tool/loadartifactstool/ -v -count=1
```

All 13 tests pass:
- 8 existing `Run` tests (updated expectations)
- 1 new `StructpbCompatibility` test
- 2 existing `ProcessRequest` tests
- 1 new `ProcessRequest_AnySlice` test
- 1 existing `OtherFunctionCall` test

### Minimal Reproduction (from issue)
```go
result := map[string]any{"artifact_names": []string{"a.pdf", "b.pdf"}}
_, err := structpb.NewStruct(result) // BEFORE: proto: invalid type: []string

// After fix, Run() returns []any:
result = map[string]any{"artifact_names": []any{"a.pdf", "b.pdf"}}
_, err = structpb.NewStruct(result)  // AFTER: <nil>
```


baptmont:
Pushed a commit narrowing this to the read path (maintainer edits). The diff goes from +113/-10 to **+77/-3**, and the part that was actually still broken is untouched.

To be clear up front: **the original diagnosis was correct when this was opened.** It has just been overtaken.

## The reported failure no longer happens

`toStructPB` stopped calling `structpb.NewStruct` and now round-trips through JSON, which handles typed slices fine. That was ca2e3d5c (#793), 2026-06-19 — three months after this branch was written. Driving the real conversion path on current `main` with exactly what `Run` produces:

```
createAiplatformpbContent -> ok; persisted = map["artifact_names":[]interface{}{"a.pdf","b.pdf"}]
eventToRawEvent           -> ok
```

That's `createAiplatformpbContent`, the function behind the `failed to convert function response to structpb` message in #668. It's also fixed on the `v1` branch. So the `[]string` → `[]any` change in `Run` no longer has a bug to fix, which is why it's reverted here.

## What *is* still broken is the other direction

`Response.AsMap()` widens the persisted value back to `[]interface{}`, and the reader asserted `[]string`:

```
[]any (post-round-trip) -> err = invalid artifact names type: []interface {}, expected []string
```

So artifacts silently fail to load after a session round-trip — on `main` and `v1` both. **Your type switch fixes that**, and it's the half nobody would find from reading #668, which describes only the write direction. That's the whole remaining change.

## Why revert `Run` rather than keep it as belt-and-braces

Emitting `[]any` doesn't simplify the reader: the `[]string` branch has to stay regardless, for events persisted before this lands and for in-process callers. So it changes which branch the common path takes and nothing else — while leaving `loadartifactstool` returning `[]any` where `loadmemorytool` returns a typed `[]memory.Entry`. #793 settled that shape deliberately; its log reads `revert: undo initial changes to load_memory tool` followed by the central normalization.

I also dropped `TestLoadArtifactsTool_Run_StructpbCompatibility`. It asserted `Run`'s output against `structpb.NewStruct`, which isn't the code path — keeping it would pin a diagnosis that's no longer true, and the real serialization is already covered in `session/vertexai`.

## Verification

`..._AnySlice` is mutation-checked: revert the type switch and it fails, so the remaining fix is genuinely guarded.

```
go build -mod=readonly ./...                                     pass
go test -race -count=1 -shuffle=on ./tool/loadartifactstool/...  pass
go test -race -count=1 ./...                                     pass (full repo)
golangci-lint run ./tool/loadartifactstool/...                   0 issues
gofmt -l / go mod tidy -diff                                     clean
```

Two notes for merge. The PR description and #668 both still describe the write-side failure and don't mention the read path, so they'd mislead anyone reading back — worth correcting when this lands rather than letting the auto-close stand. And CI here has never run `test`/`lint`/`guardrail`; the red `check-changes` fails at the *Set up job* step, which is runner setup rather than anything in the code.

Thanks @brucearctor — the read-side fix is the real bug and it would have been easy to miss.

baptmont (approved):

PR #673, opened 2026-03-22
fix: propagate plugin callbacks to sub-agent runners in agenttool

### LINKED ISSUE #669: Plugin callbacks not propagated to sub-agent runners in agenttool
## Bug

When a `Plugin` is registered on a `Runner` via `PluginConfig`, its callbacks (`BeforeToolCallback`, `AfterToolCallback`, etc.) do not fire for tools executed inside a sub-agent invoked via `agenttool`.

## Root Cause

`agenttool.Run()` creates a new `runner.Runner` at [agent_tool.go:170-177](https://github.com/google/adk-go/blob/main/tool/agenttool/agent_tool.go#L170-L177) **without forwarding `PluginConfig`**:

```go
    AppName:         t.agent.Name(),
    Agent:           t.agent,
    SessionService:  sessionService,
    ArtifactService: artifact.InMemoryService(),
    MemoryService:   memory.InMemoryService(),
    // PluginConfig not set — sub-runner has no plugins
})
```

The parent runner stores its plugin manager in context at [runner.go:160](https://github.com/google/adk-go/blob/main/runner/runner.go#L160):
```go
ctx = plugininternal.ToContext(ctx, r.pluginManager)
```

The sub-runner's `Run()` then **overwrites** the parent's plugin manager with its own empty one at the same line. The parent's plugins are lost.

## Impact

- Observability plugins (logging, tracing, metrics) are blind to tool calls inside sub-agents
- `BeforeToolCallback` / `AfterToolCallback` only fire for the `agenttool` wrapper itself, not for function tools inside the sub-agent
- Same issue affects `RunLive()` (line 337)

## Reproduction

1. Create a parent agent with a sub-agent wrapped via `agenttool.New()`
2. Sub-agent has a function tool (e.g., `functiontool.New(...)`)
3. Register a plugin with `BeforeToolCallback` on the parent runner
4. Run the parent agent with input that triggers delegation to the sub-agent
5. Observe: plugin callback fires for `tool=sub_agent_name` but NOT for the inner function tool

## Suggested Fix

Add a `HasPlugins()` method to `PluginManager` and make `runner.Run()` / `runner.RunLive()` conditional:

```go
// runner.go:160 (and :337 for RunLive)
// Only overwrite if this runner has its own plugins — otherwise inherit parent's
if r.pluginManager.HasPlugins() {
    ctx = plugininternal.ToContext(ctx, r.pluginManager)
}
```

This preserves backwards compatibility: sub-runners with explicit plugins override; sub-runners without plugins (agenttool case) inherit the parent's.

## Related

- Same issue reported in adk-python: google/adk-python#2809, google/adk-python#1746
- The `TODO` comments in `agent_tool.go` (lines 174, 200) suggest forwarding services is planned but not yet implemented

brucearctor:
## Summary

Fixes #669 — Plugin callbacks (`BeforeToolCallback`, `AfterToolCallback`, etc.) are not propagated to sub-agent runners created by `agenttool`.

## Root Cause

`agenttool.Run()` creates a new `runner.Runner` without forwarding `PluginConfig`. The sub-runner's `Run()` then unconditionally overwrites the parent's plugin manager in context with its own empty one, causing the parent's plugin callbacks to not fire for tools inside the sub-agent.

## Fix

1. **`internal/plugininternal/plugin_manager.go`** — Add `HasPlugins()` method to `PluginManager`
2. **`runner/runner.go`** — Make the context override conditional: only overwrite if this runner has its own plugins, otherwise inherit the parent's

This preserves backwards compatibility:
- Sub-runners **with** explicit plugins → override (existing behavior)
- Sub-runners **without** plugins (agenttool case) → inherit parent's plugins ✅

## Testing Plan

- Added unit tests for `HasPlugins()` (empty, nil, populated configs) in `internal/plugininternal/plugin_manager_test.go`
- Ran all existing tests — all pass:
  - `runner/...` ✅
  - `tool/agenttool/...` ✅
  - `plugin/...` ✅
  - `plugin/functioncallmodifier/...` ✅
  - `plugin/retryandreflect/...` ✅

## Related

- Same issue in adk-python: google/adk-python#2809, google/adk-python#1746


internal/plugininternal/plugin_manager.go
```diff
@@ -72,6 +72,13 @@ func (pm *PluginManager) registerPlugin(plugin *plugin.Plugin) error {
 	return nil
 }
 
+// HasPlugins reports whether the PluginManager has any registered plugins.
+// This is used by the runner to decide whether to overwrite an inherited
+// plugin manager in context (e.g. from a parent runner via agenttool).
+func (pm *PluginManager) HasPlugins() bool {
+	return len(pm.plugins) > 0
+}
+
```
w1t, line 84: Perhaps we could also add a small helper like FromContext(ctx context.Context) *PluginManager here, so callers have a consistent way to retrieve the manager from context.

```
func FromContext(ctx context.Context) *PluginManager {
	a := ctx.Value(plugincontext.PluginManagerCtxKey)
	m, ok := a.(*PluginManager)
	if !ok {
		return nil
	}
	return m
}
```

runner/runner.go
```diff
@@ -173,7 +173,12 @@ func (r *Runner) Run(ctx context.Context, userID, sessionID string, msg *genai.C
 		ctx = runconfig.ToContext(ctx, &runconfig.RunConfig{
 			StreamingMode: runconfig.StreamingMode(cfg.StreamingMode),
 		})
-		ctx = plugininternal.ToContext(ctx, r.pluginManager)
+		// Only overwrite the plugin manager in context if this runner has its
+		// own plugins. Otherwise, inherit the parent's plugin manager (e.g.
+		// when a sub-runner is created by agenttool without PluginConfig).
+		if r.pluginManager.HasPlugins() {
+			ctx = plugininternal.ToContext(ctx, r.pluginManager)
+		}
```
w1t, line 248: Should we first check whether r.pluginManager is nil and, if so, try to inherit a plugin manager from the context?

```
		pluginManager := r.pluginManager
		if pluginManager != nil && pluginManager.HasPlugins() {
			ctx = plugininternal.ToContext(ctx, pluginManager)
		} else {
			pluginManager = plugininternal.FromContext(ctx)
		}
```

internal/plugininternal/plugin_manager.go
```diff
@@ -72,6 +72,13 @@ func (pm *PluginManager) registerPlugin(plugin *plugin.Plugin) error {
 	return nil
 }
 
+// HasPlugins reports whether the PluginManager has any registered plugins.
+// This is used by the runner to decide whether to overwrite an inherited
+// plugin manager in context (e.g. from a parent runner via agenttool).
+func (pm *PluginManager) HasPlugins() bool {
+	return len(pm.plugins) > 0
+}
```
w1t, line 83: Could we make `HasPlugins()` nil-safe? In Go, calling a method on a nil pointer receiver is valid, but accessing `pm.plugins` without first checking `pm == nil` would panic. Something like:

```go
func (pm *PluginManager) HasPlugins() bool {
	if pm == nil {
		return false
	}
	return len(pm.plugins) > 0
}
```

internal/plugininternal/plugin_manager.go
brucearctor: makes sense

runner/runner.go
brucearctor: great catch!

internal/plugininternal/plugin_manager.go
brucearctor: will do

indurireddy-TF:
Hi @brucearctor Could you please update your branch, address the outstanding review comments, and resolve the comment threads once they are fixed?

internal/plugininternal/plugin_manager.go
brucearctor: Done — `FromContext()` was added in commit f796221 and is now also wired as the fallback in `Run()` (commit 7241443). When the runner has no plugins, it inherits the parent's plugin manager via `pluginManager = plugininternal.FromContext(ctx)`.

runner/runner.go
brucearctor: Done — commit 7241443 now implements the full pattern you suggested:

```go
pluginManager := r.pluginManager
if pluginManager != nil && pluginManager.HasPlugins() {
    ctx = plugininternal.ToContext(ctx, pluginManager)
} else {
    pluginManager = plugininternal.FromContext(ctx)
}
```

This is applied in both `Run()` and `RunLive()` (the `RunLive` path was also missing the guard — now fixed).

internal/plugininternal/plugin_manager.go
brucearctor: Done — `HasPlugins()` is now nil-receiver safe (commit f796221):

```go
func (pm *PluginManager) HasPlugins() bool {
    if pm == nil {
        return false
    }
    return len(pm.plugins) > 0
}
```

Test for nil receiver also added.

brucearctor:
@indurireddy-TF Thanks for the nudge! All three review comments from @w1t have been addressed and the threads are now resolved:

1. **`FromContext()` helper** — added in f796221, now also wired as the fallback in `Run()` (7241443)
2. **`HasPlugins()` nil-safe** — nil-receiver guard added in f796221 with test
3. **`r.pluginManager` nil-guard + `FromContext` fallback** — `Run()` now uses the full pattern w1t suggested. `RunLive()` also received the same conditional guard (it was previously unconditionally overwriting the inherited plugin manager — now fixed in 7241443)

Branch has been rebased/merged with main (ed2b768). All runner, agenttool, and plugin tests pass.

karolpiotrowicz (changes requested):
The diagnosis here is right: plugin dispatch is context-sourced, so whoever last calls `plugininternal.ToContext` decides which plugins a sub-agent's tools see, and making that call conditional is the correct lever. I ran the change against a reproduction of #669 driven through `runner.Runner.Run`, and there are three things that need resolving before it can land.

## The fix does not reach the topology the issue describes

For a sub-agent that is an `LlmAgent` — the case in #669, since only `llmagent.Config` has a `Tools` field — `Run` never reaches the patched block. It enters the `isLlmAgent` branch at [runner.go:185](https://github.com/google/adk-go/blob/7241443617d306c89de7c99919efa6074e15bbea/runner/runner.go#L185) and returns at [runner.go:232-233](https://github.com/google/adk-go/blob/7241443617d306c89de7c99919efa6074e15bbea/runner/runner.go#L232-L233), so the new code at [runner.go:240-248](https://github.com/google/adk-go/blob/7241443617d306c89de7c99919efa6074e15bbea/runner/runner.go#L240-L248) is not on that path at all. The sub-runner goes through `runNode` instead and hits [run_node.go:218](https://github.com/google/adk-go/blob/7241443617d306c89de7c99919efa6074e15bbea/runner/run_node.go#L218), which still overwrites the parent's manager unconditionally.

I built the issue's exact shape — parent `LlmAgent` → `agenttool(LlmAgent)` → `functiontool` — and the inner tool's `BeforeToolCallback` fails to fire both before and after the change. Applying the same `HasPlugins()` guard at `run_node.go:218` and changing nothing else makes it pass, so that one line is the whole gap. The change does work for a sub-agent behind a non-LlmAgent wrapper such as a `SequentialAgent`, which is worth keeping.

<details><summary>Reproduction — parent LlmAgent, agenttool-wrapped LlmAgent sub-agent, one function tool</summary>

```
// go test -run TestPluginReachesSubAgentInnerTool ./internal/reprotest/
// A plugin with BeforeToolCallback is registered on the PARENT runner only.

base (f73905c8):  beforeTool=[sub_agent]              FAIL
head (7241443):   beforeTool=[sub_agent]              FAIL
head + the same HasPlugins() guard at run_node.go:218:
                  beforeTool=[sub_agent inner_tool]   PASS
```
</details>

## On the path it does reach, a guardrail plugin stops working correctly

Line [runner.go:247](https://github.com/google/adk-go/blob/7241443617d306c89de7c99919efa6074e15bbea/runner/runner.go#L247) reassigns the local `pluginManager`, which hands the inherited parent manager ownership of the sub-runner's run-level callbacks, not just the tool-level ones it needs. Two consequences follow.

The agenttool's synthesized internal request is presented to the parent's `OnUserMessageCallback` as if it were an end-user message, and that callback's return value replaces the message at [runner.go:618-619](https://github.com/google/adk-go/blob/7241443617d306c89de7c99919efa6074e15bbea/runner/runner.go#L618-L619). A plugin that redacts or rewrites user input will therefore quietly rewrite what the sub-agent is asked.

The sharper one: a plugin whose `BeforeRunCallback` short-circuits now short-circuits the sub-agent run too. The early-exit event is built with `Content: msg` at [runner.go:293-297](https://github.com/google/adk-go/blob/7241443617d306c89de7c99919efa6074e15bbea/runner/runner.go#L293-L297), and `agenttool` takes the last event's text as the tool result at [agent_tool.go:195-202](https://github.com/google/adk-go/blob/7241443617d306c89de7c99919efa6074e15bbea/tool/agenttool/agent_tool.go#L195-L202). So the guard's refusal text is dropped, the sub-agent never runs, and the refused prompt goes back to the parent model as a successful sub-agent answer. That `Content: msg` line predates this PR — it is unchanged from the base commit and `RunLive` gets it right — but this change is what puts it on a live path.

<details><summary>Reproduction — parent guardrail plugin, agenttool wrapping a SequentialAgent</summary>

```
// go test -run TestEarlyExitPluginHijacksSubAgent ./internal/reprotest/
// Plugin refuses any run whose user message contains BLOCKED_TOPIC.
// The real user message is "go"; the agenttool's internal request contains the marker.

base (f73905c8):
  guardSaw=["go"]
  agenttoolResult=["map[result:leaf done]"]                       PASS
head (7241443):
  guardSaw=["go" "please compute BLOCKED_TOPIC"]
  agenttoolResult=["map[result:please compute BLOCKED_TOPIC]"]    FAIL
```
</details>

Both go away by deleting the `else` branch at `runner.go:246-248` and keeping only the conditional `ToContext`. That leaves `Run` with the shape you already gave `RunLive` in this same PR, where the context is inherited but the run-level callbacks stay on `r.pluginManager`. I applied that change and re-ran the reproductions to confirm it, rather than only reasoning about it.

One related effect worth knowing about even after that fix: parallel function calls in a single model turn run concurrently at [base_flow.go:1064](https://github.com/google/adk-go/blob/7241443617d306c89de7c99919efa6074e15bbea/internal/llminternal/base_flow.go#L1064), so with the current code four concurrent `agenttool` calls invoke the parent's run-level callbacks from four goroutines. Under `-race` that produces races in plugin state and a lost update in a plugin that counts runs, where the base commit is deterministic. Removing the `else` branch removes this too.

## The new test file does not compile

[plugin_manager_test.go:21](https://github.com/google/adk-go/blob/7241443617d306c89de7c99919efa6074e15bbea/internal/plugininternal/plugin_manager_test.go#L21) imports `google.golang.org/adk/plugin`, but `go.mod` declares the module as `google.golang.org/adk/v2`. `go test ./internal/plugininternal/` fails with `[setup failed]`, which means the tests this PR adds have never run, and `go vet` and `staticcheck` both abort on the same import so neither analysed the changed code. The import was correct when you wrote it — the v2 rename landed in main afterwards and the merge did not pick it up. `gofmt` also wants the trailing blank line at the end of that file removed.

## Two smaller notes

`Run` and `RunLive` end up with different plugin semantics for the same plugin: `RunLive` inherits through the context but keeps its own manager for run-level callbacks, while `Run` does not. Nothing reaches `RunLive` through `agenttool`, so there is no live consequence, but the two should agree — and `RunLive` is the half that is right.

Worth a look at how adk-python solved the same problem, since it avoids the branch question entirely: it reads the parent's plugins in `AgentTool` and passes them into the sub-`Runner` constructor rather than relying on context inheritance, gates it behind a public `include_plugins=True` argument so callers can decline, and tells the sub-runner's manager to skip closing the borrowed plugins. Doing it at the `agenttool` call site here would sidestep the `runNode` split, and it would sit naturally next to the existing `// TODO - use forwarding_artifact_service as in python.` in that file.

The branch also needs a rebase — it is 129 commits behind main with a conflict in `runner/runner.go`.

PR #680, opened 2026-03-25
feat: custom runner provider for adka2a executor

yarolegovich:
* `adka2a` package declares `Runner` interface so that user can use custom runner implementations with a2a translation layer.
* `adka2a.ExecutorConfig` takes `RunnerProvider` defaulting to `runner.New` if not passed.

### DIFF
--- a/server/adka2a/executor.go
+++ b/server/adka2a/executor.go
@@ -31,6 +31,7 @@ import (
 
 	"google.golang.org/adk/agent"
 	iremoteagent "google.golang.org/adk/internal/agent/remoteagent"
+	"google.golang.org/adk/plugin"
 	"google.golang.org/adk/runner"
 	"google.golang.org/adk/session"
 )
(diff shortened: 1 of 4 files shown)


server/adka2a/executor.go
```diff
@@ -362,3 +404,46 @@ func (e *Executor) prepareSession(ctx context.Context, meta invocationMeta) erro
 
 	return nil
 }
+
+func (e *Executor) createRunnerConfig(ctx context.Context, reqCtx *a2asrv.RequestContext) (RunnerConfig, error) {
```
kdroste-google, line 408: Why not to store the config (as effectiveConfig RunnerConfig in Executor or ExecutorConfig) after it's created ? There's no guarantee that RunnerProvider returns the same config data

kdroste-google (changes requested):

server/adka2a/executor.go
yarolegovich: executor also handles cancel RPCs, so we might require runner config without requiring a runner: two servers behind a load-balancer - one receives an execute rpc, the second one receives a cancel rpc

kdroste-google: Ok, interesting.

kdroste-google (approved):
LGTM, Thank you!

PR #683, opened 2026-03-26
fix: filter adk_request_confirmation events from LLM history

### LINKED ISSUE #682: contents_processor.go missing isConfirmationEvent filter breaks tool confirmation with persistent session stores
## 🔴 Required Information

**Describe the Bug:**

`buildContentsDefault` in `internal/llminternal/contents_processor.go` filters `adk_request_credential` events from LLM history via `isAuthEvent()`, but has no equivalent filter for `adk_request_confirmation` events. When using a persistent session store (anything other than `InMemorySessionService`), confirmation events are replayed into the Gemini context on the next turn, causing a 400 error because Gemini sees FunctionCall/FunctionResponse parts for `adk_request_confirmation` — a function it never declared.

This is documented as a known limitation in `docs/tools-custom/confirmation.md`:

> Known limitations:
> - DatabaseSessionService is not supported by this feature.
> - VertexAiSessionService is not supported by this feature.

The fix below removes this limitation.

**Steps to Reproduce:**

1. Configure an agent with a tool that uses `ctx.RequestConfirmation()` (or `RequireConfirmation: true`)
2. Use any persistent `SessionService` (database-backed, custom implementation, etc.)
3. User sends a message that triggers the tool → confirmation event is stored in session
4. User sends confirmation response (`FunctionResponse` with `adk_request_confirmation`)
5. `RequestConfirmationRequestProcessor` re-executes the tool and yields the result
6. `ContentsRequestProcessor` builds LLM history from session — includes `adk_request_confirmation` events
7. Gemini returns 400: `"Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn"`

**Expected Behavior:**

`adk_request_confirmation` events should be filtered from LLM history, matching the existing `isAuthEvent()` pattern for `adk_request_credential`.

**Observed Behavior:**

Gemini 400 error on the confirmation response turn. Tool confirmation only works with `InMemorySessionService`.

**Proposed Fix:**

Add `isConfirmationEvent()` filter in `buildContentsDefault`, mirroring `isAuthEvent()`:

```go
// In buildContentsDefault, after isAuthEvent check:
if isConfirmationEvent(ev) {
    continue
}

// New function:
const requestConfirmationFunctionCallName = "adk_request_confirmation"

func isConfirmationEvent(ev *session.Event) bool {
    c := utils.Content(ev)
    if c == nil {
        return false
    }
    for _, p := range c.Parts {
        if p.FunctionCall != nil && p.FunctionCall.Name == requestConfirmationFunctionCallName {
            return true
        }
        if p.FunctionResponse != nil && p.FunctionResponse.Name == requestConfirmationFunctionCallName {
            return true
        }
    }
    return false
}
```

**Environment Details:**

- ADK Library Version: v1.0.0
- OS: Linux
- Go Version: 1.26

**Model Information:**

- Gemini 2.5 Flash / Pro (any model — the issue is in event history construction, not model-specific)

---

## 🟡 Optional Information

**Related Issues:**

- #656 (ThoughtSignature missing on confirmation events — different bug, same flow)
- #590, #558 (confirmation event missing ID — different bug, same flow)

Always (100%) — deterministic when using persistent sessions with tool confirmation.

odsod:
## Description

`buildContentsDefault` in `contents_processor.go` filters `adk_request_credential` events from LLM history via `isAuthEvent()`, but has no equivalent for `adk_request_confirmation` events.

When using a persistent session store (anything other than `InMemorySessionService`), confirmation events are replayed into the Gemini context on the next turn, causing a 400 error — Gemini sees FunctionCall/FunctionResponse parts for `adk_request_confirmation`, a function it never declared.

## Changes

- Add `isConfirmationEvent()` function in `contents_processor.go`, mirroring `isAuthEvent()`
- Wire it into `buildContentsDefault` to skip confirmation events
- Add 3 test cases: FunctionCall event, FunctionResponse event, and mixed history with confirmation events interleaved

## Testing

```
go test ./internal/llminternal/ -run "TestContentsRequestProcessor$" -v
```

All existing tests pass. New tests:
- `ConfirmationFunctionCallEvent` — verifies FunctionCall with `adk_request_confirmation` is filtered
- `ConfirmationFunctionResponseEvent` — verifies FunctionResponse with `adk_request_confirmation` is filtered
- `ConfirmationEventsFilteredFromHistory` — verifies confirmation events are stripped while regular tool call events are preserved

Fixes #682

### DIFF
--- a/internal/llminternal/contents_processor.go
+++ b/internal/llminternal/contents_processor.go
@@ -29,6 +29,7 @@ import (
 	"google.golang.org/adk/internal/utils"
 	"google.golang.org/adk/model"
 	"google.golang.org/adk/session"
+	"google.golang.org/adk/tool/toolconfirmation"
 )
 
 // ContentRequestProcessor populates the LLMRequest's Contents based on
@@ -83,6 +84,9 @@ func buildContentsDefault(agentName, invocationBranch string, events []*session.
 			continue
 		}
 		if isAuthEvent(ev) {
+			continue
+		}
+		if isConfirmationEvent(ev) {
 			continue
 		}
 		if isOtherAgentReply(agentName, ev) {
@@ -519,6 +523,26 @@ func isAuthEvent(ev *session.Event) bool {
 			return true
 		}
 		if p.FunctionResponse != nil && p.FunctionResponse.Name == requestEUCFunctionCallName {
+			return true
+		}
+	}
+	return false
+}
+
+// isConfirmationEvent returns true if the event contains an adk_request_confirmation
+// FunctionCall or FunctionResponse. These are ADK-internal events used for the tool
+// confirmation (HITL) flow and must be filtered from LLM history — the model never
+// declared this function and cannot interpret it.
+func isConfirmationEvent(ev *session.Event) bool {
+	c := utils.Content(ev)
+	if c == nil {
+		return false
+	}
+	for _, p := range c.Parts {
+		if p.FunctionCall != nil && p.FunctionCall.Name == toolconfirmation.FunctionCallName {
+			return true
+		}
+		if p.FunctionResponse != nil && p.FunctionResponse.Name == toolconfirmation.FunctionCallName {
 			return true
 		}
 	}
--- a/internal/llminternal/contents_processor_test.go
+++ b/internal/llminternal/contents_processor_test.go
@@ -372,6 +372,111 @@ func TestContentsRequestProcessor(t *testing.T) {
 			want: nil,
 		},
 		{
+			name: "ConfirmationFunctionCallEvent",
+			events: []*session.Event{
+				{
+					Author: agentName,
+					LLMResponse: model.LLMResponse{
+						Content: &genai.Content{
+							Role: "model",
+							Parts: []*genai.Part{
+								{FunctionCall: &genai.FunctionCall{Name: "adk_request_confirmation"}},
+							},
+						},
+					},
+				},
+			},
+			want: nil,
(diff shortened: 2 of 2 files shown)


internal/llminternal/contents_processor.go
```diff
@@ -525,6 +528,28 @@ func isAuthEvent(ev *session.Event) bool {
 	return false
 }
 
+const requestConfirmationFunctionCallName = "adk_request_confirmation"
```
w1t: Nit: could we reuse toolconfirmation.FunctionCallName here instead of introducing another local adk_request_confirmation constant? It already exists in tool/toolconfirmation, and reusing it would keep this as a single source of truth.


https://github.com/google/adk-go/blob/b8eb8c5f07cf43889a02bce2eac2d104a81fb927/tool/toolconfirmation/tool_confirmation.go#L46

odsod: @w1t good catch! will fix

odsod:
@w1t hi! anything more I can do on this one, or do you think we can get this included in the next release?

verdverm:
want to confirm this filtering only happens during message creation and that confirmations are persisted to the session store for compliance

karolpiotrowicz:
Thanks for this fix and the clear write-up on the underlying issue! The same behavior — filtering `adk_request_confirmation` events out of LLM history — has since landed on `main` via #717, so this PR is now superseded. Closing it out; we really appreciate the contribution.

PR #686, opened 2026-03-27
Add thought signature propagation to avoid hard crashes

### LINKED ISSUE #656: Missing ThoughtSignature on adk_request_confirmation parts breaks execution
**Describe the bug**

`generateRequestConfirmationEvent` in `internal/llminternal/functions.go` creates synthetic `adk_request_confirmation` `FunctionCall` parts but does not copy the `ThoughtSignature` from the original model-generated `FunctionCall` part onto them. This causes Gemini 3 models (and 2.5 with thinking enabled) to reject subsequent requests with a `400 INVALID_ARGUMENT` error: "function call … is missing a `thought_signature`".

The stream aggregator (`internal/llminternal/stream_aggregator.go`) already handles this correctly — it tracks `currentFunctionThoughtSignature` and copies it onto emitted parts. The confirmation flow simply missed it.

**To Reproduce**

1. Create an `LlmAgent` with a tool wrapped in `tool.WithConfirmation` (or using `require_confirmation`)
2. Use a Gemini model that emits thought signatures (e.g. `gemini-3-flash-preview`)
3. Trigger the tool so the model emits a `FunctionCall` with a `ThoughtSignature`
4. The confirmation flow calls `generateRequestConfirmationEvent`, which builds a new `Part{FunctionCall: ...}` without setting `ThoughtSignature`
5. This event is appended to session history. On the next model call, the Gemini API rejects the request with:
   ```
   400 INVALID_ARGUMENT: Unable to submit request because function call `adk_request_confirmation`
   in the N. content block is missing a `thought_signature`.
   ```

**Expected behavior**

`generateRequestConfirmationEvent` should copy the `ThoughtSignature` from the original model-generated `FunctionCall` part (found in `functionCallEvent.Content.Parts`) onto the synthetic `adk_request_confirmation` part it creates, so that the Gemini API thought signature validation passes.

**Screenshots**

N/A — API-level error, no UI involved.

**Versions**

- OS: Linux (also reproducible on Mac)
- ADK version: `v0.5.0` / `main` at commit `d8a05b5` (2026-03-17). 
- Python version: N/A (this is ADK Go). Go 1.23+.

**Additional context**

The fix is likely small. In `generateRequestConfirmationEvent`, the loop already looks up `originalFunctionCall` by ID from `functionCallEvent.Content.Parts`. It just needs to also look up the corresponding `ThoughtSignature` from the original `Part` and set it on the new part:

```go
// Find the original Part's ThoughtSignature
var originalThoughtSig []byte
for _, p := range functionCallEvent.Content.Parts {
    if p.FunctionCall != nil && p.FunctionCall.ID == funcID {
        originalThoughtSig = p.ThoughtSignature
        break
    }
}

part := &genai.Part{
    FunctionCall: requestConfirmationFC,
}
if len(originalThoughtSig) > 0 {
    part.ThoughtSignature = originalThoughtSig
}
parts = append(parts, part)
```

This mirrors how `stream_aggregator.go` already handles `ThoughtSignature` propagation.

gustafvh:

- Closes: #656

**Problem:**
`generateRequestConfirmationEvent` created synthetic `adk_request_confirmation` function-call parts without preserving the original model-generated part's `ThoughtSignature`.

Gemini thinking models require every replayed model-role function-call part to carry its thought signature. When the confirmation flow appended a synthetic `adk_request_confirmation` part without that signature, subsequent requests could fail with `400 INVALID_ARGUMENT` complaining that the function call was missing a `thought_signature`.

**Solution:**
This PR fixes the confirmation-path bug by preserving the original `ThoughtSignature` when constructing synthetic `adk_request_confirmation` function-call parts in `generateRequestConfirmationEvent`.

The implementation indexes the original function-call parts by call ID, captures the corresponding `ThoughtSignature`, and copies it onto the synthetic confirmation part when present.

This PR is intentionally scoped to the `#656` confirmation bug only. It does **not** include:
- parallel streaming thought-signature propagation
- confirmation-event history filtering
- Vertex AI session persistence / serialization changes

These should be addressed in separate PRs to keep scope small.

### Testing Plan

**Unit Tests:**

- [x] Relevant tests pass locally.

New/updated tests:
- `TestGenerateRequestConfirmationEventPreservesThoughtSignature` -- verifies the original thought signature is copied to the synthetic confirmation part
- `TestGenerateRequestConfirmationEventNoThoughtSignature` -- verifies normal behavior when the original function-call part has no thought signature

```sh
ok  	google.golang.org/adk/internal/llminternal
```

### DIFF
--- a/internal/llminternal/functions.go
+++ b/internal/llminternal/functions.go
@@ -43,20 +43,33 @@ func generateRequestConfirmationEvent(
 
 	parts := []*genai.Part{}
 	longRunningToolIDs := []string{}
-	functionCalls := make(map[string]*genai.FunctionCall, len(functionCallEvent.Content.Parts))
-	for _, call := range utils.FunctionCalls(functionCallEvent.Content) {
-		functionCalls[call.ID] = call
+
+	// Index the original Parts by function call ID so we can carry over
+	// ThoughtSignature. Gemini thinking models require every model-role
+	// function call Part to include its thought signature; without it the
+	// API returns INVALID_ARGUMENT.
+	type originalCall struct {
+		call             *genai.FunctionCall
+		thoughtSignature []byte
+	}
+	originals := make(map[string]originalCall, len(functionCallEvent.Content.Parts))
+	for _, p := range functionCallEvent.Content.Parts {
+		if p.FunctionCall != nil {
+			originals[p.FunctionCall.ID] = originalCall{
+				call:             p.FunctionCall,
+				thoughtSignature: p.ThoughtSignature,
+			}
+		}
 	}
 
 	for funcID, confirmation := range functionResponseEvent.Actions.RequestedToolConfirmations {
-		originalFunctionCall, ok := functionCalls[funcID]
-		if !ok || originalFunctionCall == nil {
+		orig, ok := originals[funcID]
+		if !ok || orig.call == nil {
 			continue
 		}
 
-		// Prepare arguments for the adk_request_confirmation call
 		args := map[string]any{
-			"originalFunctionCall": originalFunctionCall,
+			"originalFunctionCall": orig.call,
 			"toolConfirmation":     confirmation,
 		}
 
@@ -67,7 +80,8 @@ func generateRequestConfirmationEvent(
 		}
 
 		parts = append(parts, &genai.Part{
-			FunctionCall: requestConfirmationFC,
+			FunctionCall:     requestConfirmationFC,
+			ThoughtSignature: orig.thoughtSignature,
 		})
 		longRunningToolIDs = append(longRunningToolIDs, requestConfirmationFC.ID)
 	}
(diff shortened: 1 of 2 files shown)


baptmont:
Hi @gustafvh thank you for reducing the scope of the fix to only cover issue #656. As far I can tell the fix in PR #717 of excluding the adk_request_confirmation events will fix the #656 issue. These are internal function calls not requested by the llm that are incorrectly being included in the context. 
If after #717 we still encounter the same issue we can explore an alternative like this. However including a ThoughtSignature to a part that was not generated by the llm can also trigger validation to incorrect history.
In some case we don't want to send thought signatures for function calls, like looking at the documentation for [gemini thought-signatures](https://ai.google.dev/gemini-api/docs/thought-signatures) when using Parallel Function Calls `"the thought_signature is attached only to the first functionCall part. Subsequent functionCall parts in the same response will not contain a signature" `

gustafvh:
@baptmont Makes sense, I will close the PR, wait for #717, and reopen if we still encounter the same issues. Thanks for the pointer!

PR #687, opened 2026-03-27
feat(telemetry): add configurable LRU cache in debug telemetry to avoid memory leaks

pawel-maciejczek:
This CL adds LRU cache to limit the number of traces kept in memory for debug telemetry.
The cache size is configurable via the newly added config struct. The default is 10k traces.

### DIFF
--- a/server/adkrest/handler.go
+++ b/server/adkrest/handler.go
@@ -34,7 +34,9 @@ import (
 
 // NewServer creates a new ADK REST API server which implements [http.Handler] interface.
 func NewServer(cfg ServerConfig) (*Server, error) {
-	debugTelemetry := services.NewDebugTelemetry()
+	debugTelemetry := services.NewDebugTelemetry(&services.DebugTelemetryConfig{
+		TraceCapacity: cfg.DebugConfig.TraceCapacity,
+	})
 
 	router := mux.NewRouter().StrictSlash(true)
 	// TODO: Allow taking a prefix to allow customizing the path
@@ -61,6 +63,14 @@ type ServerConfig struct {
 	ArtifactService artifact.Service
 	SSEWriteTimeout time.Duration
 	PluginConfig    runner.PluginConfig
+	DebugConfig     *DebugTelemetryConfig
+}
+
+// DebugTelemetryConfig contains parameters for the debug telemetry.
+type DebugTelemetryConfig struct {
+	// Maximum number of traces to keep in memory.
+	// If <= 0, the default capacity is used.
+	TraceCapacity int
 }
 
 // Server is an HTTP server that serves the ADK REST API.
--- a/server/adkrest/handler.go
+++ b/server/adkrest/handler.go
@@ -15,6 +15,7 @@
 package adkrest
 
 import (
+	"fmt"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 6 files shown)


server/adkrest/handler.go
```diff
@@ -61,6 +63,14 @@ type ServerConfig struct {
 	ArtifactService artifact.Service
 	SSEWriteTimeout time.Duration
 	PluginConfig    runner.PluginConfig
+	DebugConfig     *DebugTelemetryConfig
+}
+
+// DebugTelemetryConfig contains parameters for the debug telemetry.
+type DebugTelemetryConfig struct {
+	// Maximum number of traces to keep in memory.
+	// If <= 0, the default capacity is used.
```
dpasiukevich: nit: could you mention the actual default value so the users know

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -94,47 +110,54 @@ type DebugLog struct {
...
-	return &spanStore{
+func newSpanStore(capacity int) *spanStore {
+	store := &spanStore{
 		recordsBySpanID:     make(map[string]*spanRecord),
 		traceIDsBySessionID: make(map[string]map[string]struct{}),
 		recordsByEventID:    make(map[string][]*spanRecord),
-		recordsByTraceID:    make(map[string][]*spanRecord),
 	}
+	cache, err := lru.NewWithEvict(capacity, store.evict)
+	if err == nil {
```
dpasiukevich: 1. To avoid any hard-to-debug errors, let's propagate the error up the stack to adk rest handler constructor.

2. And could I ask to add the error context, so if this fails, the final error in the HTTP handler looks something like (as example):
"failed to create REST handle: failed to create debug telemetry service: failed to create LRU cache: lru cache error"

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -94,47 +110,54 @@ type DebugLog struct {
...
 	// Create a copy of the slice to avoid race conditions.
 	records := slices.Clone(s.recordsByEventID[id])
+	seenTraces := make(map[string]bool)
+	for _, r := range records {
+		traceIDStr := r.Context.TraceID().String()
+		if traceIDStr != "" && !seenTraces[traceIDStr] {
+			s.recordsByTraceID.Get(traceIDStr)
+			seenTraces[traceIDStr] = true
+		}
+	}
```
dpasiukevich: What does this code do? It seems to be no-op, because values of `s.recordsByTraceID.Get(traceIDStr)` are not used.

server/adkrest/internal/services/debugtelemetry.go
```diff
@@ -245,16 +267,54 @@ func (s *spanStore) updateSpanIndexes(span *inMemorySpan, record *spanRecord) {
...
+				s.evictTraceIDsBySessionID(sessionID, traceID)
+			}
+		}
+	}
+}
+
+func (s *spanStore) evictRecordsByEventID(eventID string, span *spanRecord) {
+	records := s.recordsByEventID[eventID]
+	records = slices.DeleteFunc(records, func(r *spanRecord) bool {
+		return r == span
```
dpasiukevich: You compare `*spanRecord` pointers here. Would it be better to compare & delete by span_id for example?

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: SG, done.

server/adkrest/handler.go
pawel-maciejczek: Done.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: Sure - done.

server/adkrest/internal/services/debugtelemetry.go
pawel-maciejczek: We need to iterate over all traces returned `getSpansByEventID` and "touch" them by calling `Get` method to make them the most recently used keys in lru cache. It's not needed 

I extracted a helper function and added comments to explain it better in the code.

dpasiukevich (approved):

server/adkrest/internal/services/debugtelemetry.go
dpasiukevich: Thanks for clarification!

PR #688, opened 2026-03-27
feat: init trigger routes set up for event processing

anFatum:
Extends the ADK Web with Event processing endpoints based on the provided flags, this will be used to allow users specify ADK Web endpoints as the processing endpoints for Pubsub, Eventarc and BigQuery GCP services. (see the details in [go/orcas-rfc-522](http://go/orcas-rfc-522))

### DIFF
--- a/cmd/launcher/web/api/api.go
+++ b/cmd/launcher/web/api/api.go
@@ -19,6 +19,7 @@ import (
 	"flag"
 	"fmt"
 	"net/http"
+	"slices"
 	"strings"
 	"time"
 
@@ -31,11 +32,19 @@ import (
 	"google.golang.org/adk/telemetry"
 )
 
+// SupportedTriggers defines the allowed trigger sources for the ADK REST API.
+var SupportedTriggers = []string{"pubsub"}
+
 // apiConfig contains parametres for lauching ADK REST API
 type apiConfig struct {
-	frontendAddress string
-	pathPrefix      string
-	sseWriteTimeout time.Duration
+	frontendAddress   string
+	pathPrefix        string
+	sseWriteTimeout   time.Duration
+	triggerSources    string
+	triggerMaxRetries int
+	triggerBaseDelay  time.Duration
+	triggerMaxDelay   time.Duration
+	triggerMaxRuns    int
 }
 
 // apiLauncher can launch ADK REST API
@@ -73,6 +82,25 @@ func (a *apiLauncher) UserMessage(webURL string, printer func(v ...any)) {
 
 // SetupSubrouters adds the API router to the parent router.
 func (a *apiLauncher) SetupSubrouters(router *mux.Router, config *launcher.Config) error {
+	if a.config.triggerSources != "" {
+		sources := strings.Split(a.config.triggerSources, ",")
+		for _, source := range sources {
+			if !slices.Contains(SupportedTriggers, source) {
+				return fmt.Errorf("invalid trigger source: %q. Any subset of %s is allowed. Values should be comma-separated", source, strings.Join(SupportedTriggers, ", "))
+			}
+		}
+		// De-duplicate the input sources.
+		slices.Sort(sources)
+		config.TriggerSources = slices.Compact(sources)
+	}
+
+	config.TriggerConfig = launcher.TriggerConfig{
+		MaxRetries:        a.config.triggerMaxRetries,
+		BaseDelay:         a.config.triggerBaseDelay,
+		MaxDelay:          a.config.triggerMaxDelay,
+		MaxConcurrentRuns: a.config.triggerMaxRuns,
+	}
+
 	// Create the ADK REST API handler
 	restServer, err := adkrest.NewServer(adkrest.ServerConfig{
 		SessionService:  config.SessionService,
@@ -81,6 +109,7 @@ func (a *apiLauncher) SetupSubrouters(router *mux.Router, config *launcher.Confi
 		ArtifactService: config.ArtifactService,
 		SSEWriteTimeout: a.config.sseWriteTimeout,
 		PluginConfig:    config.PluginConfig,
+		TriggerSources:  config.TriggerSources,
 	})
 	if err != nil {
 		return fmt.Errorf("failed to create REST server: %w", err)
@@ -115,6 +144,19 @@ func (a *apiLauncher) Parse(args []string) ([]string, error) {
 	if err != nil || !a.flags.Parsed() {
 		return nil, fmt.Errorf("failed to parse api flags: %v", err)
 	}
+	if a.config.triggerMaxRetries < 0 {
+		return nil, fmt.Errorf("trigger_max_retries must be >= 0")
+	}
+	if a.config.triggerBaseDelay < 0 {
+		return nil, fmt.Errorf("trigger_base_delay must be >= 0")
+	}
+	if a.config.triggerMaxDelay < 0 {
+		return nil, fmt.Errorf("trigger_max_delay must be >= 0")
+	}
+	if a.config.triggerMaxRuns < 0 {
+		return nil, fmt.Errorf("trigger_max_concurrent_runs must be >= 0")
+	}
+
 	p := a.config.pathPrefix
 	if !strings.HasPrefix(p, "/") {
 		p = "/" + p
@@ -138,6 +180,11 @@ func NewLauncher() weblauncher.Sublauncher {
 	fs.StringVar(&config.frontendAddress, "webui_address", "localhost:8080", "ADK WebUI address as seen from the user browser. It's used to allow CORS requests. Please specify only hostname and (optionally) port.")
 	fs.StringVar(&config.pathPrefix, "path_prefix", "/api", "ADK REST API path prefix. Default is '/api'.")
 	fs.DurationVar(&config.sseWriteTimeout, "sse-write-timeout", 120*time.Second, "SSE server write timeout (i.e. '10s', '2m' - see time.ParseDuration for details) - for writing the SSE response after reading the headers & body")
+	fs.IntVar(&config.triggerMaxRetries, "trigger_max_retries", 3, "Maximum retries for HTTP 429 errors from triggers")
+	fs.DurationVar(&config.triggerBaseDelay, "trigger_base_delay", 1*time.Second, "Base delay for trigger retry exponential backoff")
+	fs.DurationVar(&config.triggerMaxDelay, "trigger_max_delay", 10*time.Second, "Maximum delay for trigger retry exponential backoff")
+	fs.IntVar(&config.triggerMaxRuns, "trigger_max_concurrent_runs", 100, "Maximum concurrent trigger runs")
+	fs.StringVar(&config.triggerSources, "trigger_sources", "", fmt.Sprintf("Comma-separated list of trigger sources to enable (any subset of %s)", strings.Join(SupportedTriggers, ", ")))
 
 	return &apiLauncher{
 		config: config,
--- a/server/adkrest/controllers/triggers.go
+++ b/server/adkrest/controllers/triggers.go
@@ -1,0 +1,217 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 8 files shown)


cmd/launcher/web/api/api.go
```diff
@@ -73,6 +78,15 @@ func (a *apiLauncher) UserMessage(webURL string, printer func(v ...any)) {
 
 // SetupSubrouters adds the API router to the parent router.
 func (a *apiLauncher) SetupSubrouters(router *mux.Router, config *launcher.Config) error {
+	if a.config.triggerSources != "" {
+		sources := strings.Split(a.config.triggerSources, ",")
+		for _, source := range sources {
+			if !slices.Contains(SupportedTriggers, source) {
+				return fmt.Errorf("invalid trigger source: %q. Allowed values are: %s", source, strings.Join(SupportedTriggers, ", "))
```
kdroste-google: Maybe "Any subset of .... is allowed. Values should be comma-separated."

cmd/launcher/web/api/api.go
```diff
@@ -138,6 +153,7 @@ func NewLauncher() weblauncher.Sublauncher {
 	fs.StringVar(&config.frontendAddress, "webui_address", "localhost:8080", "ADK WebUI address as seen from the user browser. It's used to allow CORS requests. Please specify only hostname and (optionally) port.")
 	fs.StringVar(&config.pathPrefix, "path_prefix", "/api", "ADK REST API path prefix. Default is '/api'.")
 	fs.DurationVar(&config.sseWriteTimeout, "sse-write-timeout", 120*time.Second, "SSE server write timeout (i.e. '10s', '2m' - see time.ParseDuration for details) - for writing the SSE response after reading the headers & body")
+	fs.StringVar(&config.triggerSources, "trigger_sources", "", "Comma-separated list of trigger sources to enable (bq, pubsub, eventarc)")
```
kdroste-google: Maybe generate it from SupportedTriggers instead of verbatim string

cmd/launcher/web/api/api.go
kdroste-google: We are accepting multiple occurences like "bq,bq,bq" here, just a remark, I see no need to be too strict about that.

kdroste-google: But later on based on this value we add handlers, imho it should get deduplicated

server/adkrest/controllers/triggers.go
```diff
@@ -0,0 +1,40 @@
...
+type TriggersAPIController struct{}
+
+// NewTriggersAPIController creates a new TriggersAPIController.
+func NewTriggersAPIController() *TriggersAPIController {
+	return &TriggersAPIController{}
+}
+
+// BQTriggerHandler handles the BigQuery trigger endpoint.
+func (c *TriggersAPIController) BQTriggerHandler(w http.ResponseWriter, r *http.Request) {
+	Unimplemented(w, r)
```
kdroste-google: What's the plan for that? Should it stay unimplemented?

server/adkrest/controllers/triggers.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// PubSubTriggerHandler handles the PubSub trigger endpoint.
+func (c *TriggersAPIController) PubSubTriggerHandler(w http.ResponseWriter, r *http.Request) {
+	Unimplemented(w, r)
```
kdroste-google: What's the plan for that? Should it stay unimplemented?

server/adkrest/controllers/triggers.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// EventarcTriggerHandler handles the Eventarc trigger endpoint.
+func (c *TriggersAPIController) EventarcTriggerHandler(w http.ResponseWriter, r *http.Request) {
+	Unimplemented(w, r)
```
kdroste-google: What's the plan for that? Should it stay unimplemented?

server/adkrest/internal/models/triggers.go
```diff
@@ -0,0 +1,101 @@
...
+// BQTriggerResponse represents the response for the BigQuery trigger.
+// See: https://docs.cloud.google.com/bigquery/docs/remote-functions#output_format
+type BQTriggerResponse struct {
+	// ErrorMessage indicates the error message in BQ response (if any).
+	ErrorMessage string `json:"errorMessage,omitempty"`
+}
+
+// PubSubTriggerRequest represents the request for the PubSub trigger.
+// See: https://cloud.google.com/pubsub/docs/push#receive_push
+type PubSubTriggerRequest struct {
```
kdroste-google, line 19: https://cloud.google.com/pubsub/docs/push#receive_push says in the example 
  {
      "deliveryAttempt": 5, 

What will happen with this payload?

server/adkrest/internal/models/triggers.go
```diff
@@ -0,0 +1,101 @@
...
+	// Email of the BigQuery user who triggered the query
+	Caller string `json:"caller"`
+	// Email of the session user in BigQuery.
+	SessionUser string `json:"sessionUser"`
+	// Array of calls to the remote function.
+	Calls [][]json.RawMessage `json:"calls"`
+}
+
+// BQTriggerResponse represents the response for the BigQuery trigger.
+// See: https://docs.cloud.google.com/bigquery/docs/remote-functions#output_format
```
kdroste-google: The docs says:
replies | A batch of return values. | Required for a successful response. A JSON array.Each element corresponds to a JSON encoded return value of the external function.Size of the array must match the size of the JSON array of calls in the HTTP request. For example, if the JSON array in calls has 4 elements, this JSON array needs to have 4 elements as well.
Are we not going to return non-error values?

server/adkrest/internal/models/triggers.go
```diff
@@ -0,0 +1,101 @@
...
+
+// BQTriggerRequest represents the request for the BigQuery trigger.
+// See: https://docs.cloud.google.com/bigquery/docs/remote-functions#input_format
+type BQTriggerRequest struct {
+	// Unique ID for this request from BigQuery
+	RequestID string `json:"requestId"`
+	// Email of the BigQuery user who triggered the query
+	Caller string `json:"caller"`
+	// Email of the session user in BigQuery.
+	SessionUser string `json:"sessionUser"`
```
kdroste-google: The doc says there maybe an additional field
userDefinedContext | The user defined context that was used when creating the remote function in BigQuery. | Optional. A JSON object with key-value pairs.

server/adkrest/internal/models/triggers.go
```diff
@@ -0,0 +1,101 @@
...
+// See: https://cloud.google.com/pubsub/docs/push#receive_push
+type PubSubTriggerRequest struct {
+	// The Pub/Sub message.
+	Message PubSubMessage `json:"message"`
+	// The subscription this message was published to.
+	Subscription string `json:"subscription"`
+}
+
+// PubSubMessage represents the message for the PubSub trigger.
+type PubSubMessage struct {
```
kdroste-google, line 27: https://docs.cloud.google.com/pubsub/docs/publisher#using-attributes says
{
  "data": string,
  "attributes": {
    string: string,
    ...
  },
  "messageId": string,
  "publishTime": string,
  "orderingKey": string
}

orderingKey is missing

server/adkrest/internal/models/triggers.go
```diff
@@ -0,0 +1,101 @@
...
+	// The CloudEvents specification version used for this event
+	SpecVersion string `json:"specversion"`
+	// Unique identifier for the event
+	ID string `json:"id"`
+	// Event generation time, in RFC 3339 format (optional)
+	Time string `json:"time,omitempty"`
+}
+
+// TriggerResponse represents the standard response for Pub/Sub and Eventarc triggers.
+type TriggerResponse struct {
```
kdroste-google, line 41: https://docs.cloud.google.com/eventarc/docs/reference/rest/v1/projects.locations.operations#resource:-operation says

Union field result. The operation result, which can be either an error or a valid response. If done == false, neither error nor response is set. If done == true, exactly one of error or response can be set. Some services might not provide the result. result can be only one of the following:
--
error | object (Status)The error result of the operation in case of failure or cancellation.
response | objectThe normal, successful response of the operation. If the original method returns no data on success, such as Delete, the response is google.protobuf.Empty. If the original method is standard Get/Create/Update, the response should be the resource. For other methods, the response should have the type XxxResponse, where Xxx is the original method name. For example, if the original method name is TakeSnapshot(), the inferred response type is TakeSnapshotResponse.An object containing fields of an arbitrary type. An additional field "@type" contains a URI identifying the type. Example: { "id": 1234, "@type": "types.example.com/standard/id" }.

Union field result. The operation result, which can be either an error or a valid response. If done == false, neither error nor response is set. If done == true, exactly one of error or response can be set. Some services might not provide the result. result can be only one of the following:
error	
object ([Status](https://docs.cloud.google.com/eventarc/docs/reference/rest/v1/projects.locations.operations#Status))

The error result of the operation in case of failure or cancellation.

response	
object

The normal, successful response of the operation. If the original method returns no data on success, such as Delete, the response is google.protobuf.Empty. If the original method is standard Get/Create/Update, the response should be the resource. For other methods, the response should have the type XxxResponse, where Xxx is the original method name. For example, if the original method name is TakeSnapshot(), the inferred response type is TakeSnapshotResponse.

An object containing fields of an arbitrary type. An additional field "@type" contains a URI identifying the type. Example: { "id": 1234, "@type": "types.example.com/standard/id" }.

it may be not the right example... If it is, we can have either error or response.

server/adkrest/internal/routers/triggers.go
```diff
@@ -0,0 +1,68 @@
...
+		controller:     controller,
+		triggerSources: triggerSources,
+	}
+}
+
+// Routes returns the routes for the triggers API depending on configured triggerSources.
+func (r *TriggersAPIRouter) Routes() Routes {
+	var routes Routes
+
+	for _, source := range r.triggerSources {
```
kdroste-google, line 41: multiple occurences of the same value will probably kick in here

server/adkrest/controllers/triggers.go
anFatum: I will add the implementation if the following PRs, I didn't want to have a one huge PR to review
But if you prefer having implementation now let me know I'll add it here :)

server/adkrest/internal/models/triggers.go
anFatum: I'm not quite sure I understand, the Eventarc triggers will be used (e.g.https://codelabs.developers.google.com/eventarc-workflows-cloud-run#4) 
and they follow the logic of ACKing the Pub/Sub

server/adkrest/controllers/triggers.go
anFatum: Same as above.

server/adkrest/controllers/triggers.go
kdroste-google: Yes, please, perhaps let's add one working piece, and later on - the rest

server/adkrest/internal/models/triggers.go
kdroste-google: full sample: 
{
      "deliveryAttempt": 5,
      "message": {
          "attributes": {
              "key": "value"
          },
          "data": "SGVsbG8gQ2xvdWQgUHViL1N1YiEgSGVyZSBpcyBteSBtZXNzYWdlIQ==",
          "messageId": "2070443601311540",
          "message_id": "2070443601311540",
          "orderingKey": "key",
          "publishTime": "2021-02-26T19:13:55.749Z",
          "publish_time": "2021-02-26T19:13:55.749Z"
      },
    "subscription": "projects/myproject/subscriptions/mysubscription"
}

anFatum: It will be ignored, I also not sure we should support ALL the fields, since for now the agent will consume only the message itself?

server/adkrest/internal/models/triggers.go
kdroste-google: But where do we have the information about "Status" field from? What are possible values?

anFatum: Just additional info, the status code will be used to determine if the processing succeeded, status just an additional field that will be propagated with the error, I can remove it and report only status code

anFatum:
> Let's add at least one implementation and skip "unimplemented" ones

Done, I've removed eventarc and bq for now and left only the pubsub trigger support

PR #689, opened 2026-03-27
fix: fix data part conversions

nahapetyan-serob:
a2a.DataPart values were lost during A2A → genai → A2A conversion, coming back as TextPart instead of DataPart.
Fixed by encoding untyped DataPart data as InlineData with <json>...</json> sentinel tags instead of plain text, allowing the reverse conversion to reconstruct the original DataPart.


server/adka2a/parts.go
```diff
@@ -60,6 +61,12 @@ func IsPartialFlagSet(meta map[string]any) bool {
 	return isSet
 }
 
+func isJSONTaggedInlineData(part *genai.Part) bool {
```
yarolegovich: nit. there's an implicit knowledge in the code inside the block  `if isJSONTaggedInlineData(...)`:
```
jsonBytes := part.InlineData.Data[len("<json>") : len(part.InlineData.Data)-len("</json>")]
```
about what the function checks (that there's a prefix and a suffix.

not really an issue in this case given that this code is still very localised (in the same file, is not exported), but I think it's better when such functions are written like:
```
validateXMatchesY(x X) (Y, bool)
```
in this case
```go
func validateDataPartJSON(d *genai.Part) ([]byte, bool) {
  if d.InlineData == nil || d.InlineData.MIMEType != "text/plain" {
    return nil, false
  }
  if noPrefix, ok := bytes.CutPrefix(d.InlineData.Data, []byte("<json>")); ok {
    if result, ok := bytes.CutSuffix(noPrefix, []byte("</json>")); ok {
      return result, true
    }
  }
  return nil, false
}
```

server/adka2a/parts.go
```diff
@@ -351,14 +359,7 @@ func toGenAIDataPart(part a2a.DataPart) (*genai.Part, error) {
...
 	default:
-		return &genai.Part{Text: string(bytes)}, nil
-	}
-}
-
-func toGenAITextPart(part a2a.DataPart) (*genai.Part, error) {
-	bytes, err := json.Marshal(part.Data)
-	if err != nil {
-		return nil, err
+		jsonData := []byte("<json>" + string(bytes) + "</json>")
```
yarolegovich: also not necessarily a problem but I'm not sure this will be properly optimised and overhead might end up if agent produce / consume many data parts. there are some hidden conversion here:
string(bytes) requires an allocation, then concatenation result requires an allocation, and then the final result requires an allocation. it **might** be better to write this as:
```go
var res []byte
prefix, suffix := []byte("<json>"), []byte("</json>")
res = make([]byte, 0, len(prefix)+len(data)+len(suffix))
res = append(res, prefix...)
res = append(res, data...)
res = append(res, suffix...)
```

PR #695, opened 2026-03-31
fix: correct json envelope for datapart

nahapetyan-serob:
Updated a2a.DataPart conversion to use the correct <a2a_datapart_json> envelope.


yarolegovich:
Just for context: this is compatible with adk_python.

Issue #313, opened 2025-11-18
session.Events() and separating thinking & final response


verdverm:
When fetching a session and the event history, the responses do not contain sufficient information to discriminate between thinking output and final response intended for the user.

Code:

```go
fmt.Println("sessionChat", p.ID)

// get session
resp, err := c.runtime.s.Get(c.runtime.ctx, &session.GetRequest{
	AppName:   "veg",
	UserID:    "tony",
	SessionID: p.ID,
})
if err != nil {
	log.Printf("Error deleting session: %v", err)
	return
}
fmt.Println("sessionChat.get", resp)

// build response
s := resp.Session
S := make(map[string]any)
S["id"] = s.ID()
S["state"] = maps.Collect(s.State().All())
S["events"] = slices.Collect(s.Events().All())

fmt.Println("mailing sessions", payload)
c.mail("session.info", S)
```

Thinking:

```json
{
  "Content": {
    "parts": [
      {
        "text": "The `lib/agent` directory contains the following:\n- `agents` (likely a directory)\n- `cmd` (likely a directory)\n- `runtime.go`\n- `tools` (likely a directory)\n- `websocket.go`\n\nTo understand what's going on, I should probably investigate the `runtime.go` and `websocket.go` files, as they seem to be source code files directly within this directory, and then potentially the `agents` and `cmd` subdirectories."
      }
    ],
    "role": "model"
  },
  "CitationMetadata": null,
  "GroundingMetadata": null,
  "UsageMetadata": {
    "cacheTokensDetails": [
      {
        "modality": "TEXT",
        "tokenCount": 1732
      }
    ],
    "cachedContentTokenCount": 1732,
    "candidatesTokenCount": 129,
    "promptTokenCount": 2668,
    "promptTokensDetails": [
      {
        "modality": "TEXT",
        "tokenCount": 2668
      }
    ],
    "totalTokenCount": 2797
  },
  "CustomMetadata": null,
  "LogprobsResult": null,
  "Partial": false,
  "TurnComplete": false,
  "Interrupted": false,
  "ErrorCode": "",
  "ErrorMessage": "",
  "FinishReason": "STOP",
  "AvgLogprobs": 0,
  "ID": "e6ca2b3d-dff5-481e-8ef6-307e28e881a5",
  "Timestamp": "2025-11-18T01:29:29.718216-08:00",
  "InvocationID": "e-c9945ca4-eb99-4ff3-87d6-9a8e13978a7a",
  "Branch": "",
[...]

PR #696, opened 2026-04-01
fix: filter thinking parts from agenttool result

### LINKED ISSUE #697: agenttool: Run() leaks thinking parts into tool result
## Description

`agenttool.Run()` collects all `Part.Text` from the sub-agent's final response without checking `Part.Thought`. When thinking models (Gemini 2.5/3 with `IncludeThoughts: true`) are used as sub-agents, their internal reasoning tokens leak into the tool result that the calling agent sees.

## Root cause

`tool/agenttool/agent_tool.go` lines 223-228:

```go
for _, part := range lastContent.Parts {
    if part != nil && part.Text != "" {
        textParts = append(textParts, part.Text)
    }
}
```

No check on `part.Thought`. Compare with `genai.GenerateContentResponse.Text()` (`types.go:3148-3151`) which correctly skips thought parts:

```go
if part.Thought {
    continue
}
```

## Impact

- **Thinking tokens leak into tool results** — the root agent sees the sub-agent's internal reasoning mixed into the response text
- **OutputSchema validation can fail** — when thought text is concatenated before JSON output, unmarshalling breaks
- **Token waste** — the root agent processes thinking tokens it shouldn't see, inflating context

## Related

- #655 — same class of bug, fixed for `remoteagent` (merged)
- #313 — request for separating thinking & final response in session events
- #656 / #686 — `ThoughtSignature` propagation issues (related thinking-model support gaps)

## Fix

Submitted in #696 — adds `&& !part.Thought` to the text extraction loop, matching `genai.Text()` behavior.

dmora:

- Closes: #697
- Related: #655, #313

**Problem:**
`agenttool.Run()` collects all `Part.Text` from the sub-agent's final response without checking `Part.Thought`. When thinking models (Gemini 2.5/3 with `IncludeThoughts: true`) are used as sub-agents, their internal reasoning tokens leak into the tool result `map[string]any` that the calling agent sees.

This is the same class of bug fixed for `remoteagent` in #655.

Compare `agenttool.Run()` (before this fix):
```go
for _, part := range lastContent.Parts {
    if part != nil && part.Text != "" {
        textParts = append(textParts, part.Text)
    }
}
```

With `genai.GenerateContentResponse.Text()` which correctly filters thoughts:
```go
if part.Text != "" {
    if part.Thought {
        continue
    }
    texts = append(texts, part.Text)
}
```

**Solution:**
Add `&& !part.Thought` to the text extraction loop in `agenttool.Run()`, matching the `genai.Text()` behavior.

### Testing Plan

**Unit Tests:**

New tests added:
- `TestAgentTool_Run_FiltersThoughtParts` — mixed thought + text parts, with edge case `{Text: "", Thought: true}`, verifies only non-thought text is returned
- `TestAgentTool_Run_AllThoughtPartsReturnsEmpty` — all-thought response returns empty map
- `TestAgentTool_Run_FiltersThoughtParts_WithOutputSchema` — thought parts mixed with JSON output, verifies clean unmarshalling when `OutputSchema` is set

```
$ go test ./tool/agenttool/ -v
--- PASS: TestAgentTool_Declaration (0.00s)
--- PASS: TestAgentTool_DeclarationWithoutSchema (0.00s)
--- PASS: TestAgentTool_Run_InputValidation (0.00s)
--- PASS: TestAgentTool_Run_OutputValidation (0.00s)
--- PASS: TestAgentTool_Run_Successful (0.00s)
--- PASS: TestAgentTool_Run_WithoutSchema (0.00s)
--- PASS: TestAgentTool_Run_FiltersThoughtParts (0.00s)
--- PASS: TestAgentTool_Run_AllThoughtPartsReturnsEmpty (0.00s)
--- PASS: TestAgentTool_Run_FiltersThoughtParts_WithOutputSchema (0.00s)
--- PASS: TestAgentTool_Run_EmptyModelResponse (0.00s)
--- PASS: TestAgentTool_Run_SkipSummarization (0.00s)
PASS
ok  google.golang.org/adk/tool/agenttool  0.278s
```

Verified locally with a real Gemini 2.5 Flash model with `ThinkingConfig{IncludeThoughts: true, ThinkingBudget: 2048}`. The sub-agent was asked "What is 7 * 8?" and the tool result contained only `"56"` — no thinking tokens leaked through.

Also verified that reverting the fix causes the mock test to fail with thinking text concatenated into the result:
```
- "56"
+ "The user is asking about multiplication. 7 times 8 equals 56.\n56\nLet me double-check: 7 * 8 = 56. Yes, that's correct."
```

### Checklist

### Additional context

Aligned with how `genai.GenerateContentResponse.Text()` handles thought parts in `types.go`.

Generated with [Claude Code](https://claude.com/claude-code)


jjsasha63 (commented):
Thanks for the clear write-up — the diagnosis is right and the fix is the right shape.

I checked the rest of the tree and `agenttool` really is the last holdout: `workflow/base_node.go:135`, `agent/llmagent/llmagent.go:527`, `internal/utils.IsProsePart` and `remoteagent` (after #655) all filter thoughts already. So this closes the gap rather than introducing a new convention.

I also verified it locally against current `main`: tests pass, and reverting the one-line change makes all three new tests fail with exactly the leak you describe, including the `OutputSchema` case failing to parse. The tests bite.

### One suggestion: use the existing helper

We already have `internal/utils.IsProsePart`, and `agenttool` imports `internal/utils` for the schema validators, so it costs nothing:

```go
for _, part := range lastContent.Parts {
    if utils.IsProsePart(part) {
        textParts = append(textParts, part.Text)
    }
}
```

Besides thoughts, that also drops a part carrying text alongside an actionable payload — a `FunctionCall`, `InlineData`, `ExecutableCode`. The hand-rolled condition still folds those parts' text into the result, which is the same class of leak one step over. It also means the next change to what counts as readable text lands here for free.

### Tests

They cover the right cases, including the `OutputSchema` path, which is the failure mode most likely to bite someone in production.

Two small things:

- `{Text: "", Thought: true}` is already dropped by the `Text != ""` check, so it does not exercise the new condition. Harmless, keep it if you like it as documentation.
- If you take the `IsProsePart` suggestion, a case with `{Text: "...", FunctionCall: ...}` would lock in the wider behaviour.

jjsasha63 (approved):

PR #701, opened 2026-04-02
feat: make adk work with a2a-go/v2

yarolegovich:
* Copied the code to ./agent/remoteagent and ./server/adka2a to v2 directories.
* Migrated the code in v2 directories to use a2a-go v2 types.
* Changed the code in the original directories to delegate to the new code and serve as a type translation layer.
* Deprecated original packages with a reference to v2 directories.

What this gives is that ADK can be used with a2a-go/v2 which implements A2A protocol v1 (as well as v0).


ysyneu:
Hi @yarolegovich and @dpasiukevich — thanks for working on this.

We ran into the gap this PR fixes while integrating our ADK-based agent with a remote A2A server that implements protocol v1.0 via a2a-go/v2 (HTTP+JSON transport, `supportedInterfaces` card schema). On the shipped ADK (a2a-go v0.3.x) the client failed at `NewFromCard` with "no compatible transports found" because the v0.x card struct doesn't parse `supportedInterfaces` and there is no HTTP+JSON client transport factory registered.

To unblock ourselves we pinned our project to this PR's head (`36f4c0d…`) plus `a2a-go/v2 v2.0.1` and switched our builder to `google.golang.org/adk/agent/remoteagent/v2.NewA2A`. End-to-end test against the live v1.0 server works as intended: `SendStreamingMessage` streams `TaskArtifactUpdateEvent`s back, `aggregatePartial` yields partial `session.Event`s, and they flow through the parent `llmagent` session out to our SSE client. No regressions observed elsewhere in the ADK bump (needed only minor interface updates on our side for `memory.Service.AddSessionToMemory`/`SearchMemory` renames and the new `artifact.Service.GetArtifactVersion` method).

A couple of questions, if you have a moment:

1. Is there anything concrete blocking merge beyond the current conflict with main? Happy to help test additional scenarios or review if that would move things along.
2. Any rough sense of timing? We'd like to drop our pseudo-version pin and consume this from a tagged release as soon as it's available.
3. The PR description mentions the v0.x packages becoming a translation layer — is the long-term plan to remove them in a later major, or will they remain as compatibility shims?

Thanks again — this PR lands exactly the API we need, and the split into `v2` subpackages is a clean migration path.

yarolegovich:
@ysyneu thanks for confirming there's demand for it, we'll resume the work on it

jankrynauw:
Just pinned the branch (`google.golang.org/adk v1.1.1-0.20260423095712-9cb6095c2e76`) in one of our agents which makes use of a2a extensions (https://pkg.go.dev/go.alis.build/a2a/extension/a2ui), custom TaskStore (go.alis.build/a2a/tasks) etc. based on v2 and all seems to be working fine.

server/adka2a/v2/parts.go
```diff
@@ -85,22 +84,30 @@ func ToA2APart(part *genai.Part, longRunningToolIDs []string) (a2a.Part, error)
...
 			result[i] = r
 		} else if jsonBytes, ok := validateDataPartJSON(part); ok {
 			var data map[string]any
 			if err := json.Unmarshal(jsonBytes, &data); err != nil {
 				return nil, err
 			}
-			result[i] = a2a.DataPart{Data: data}
+			result[i] = a2a.NewDataPart(data)
 		} else if part.InlineData != nil || part.FileData != nil {
+			if part.InlineData != nil && part.InlineData.DisplayName == "a2a_data_part" {
```
baptmont: This DisplayName seems like old logic. to remove

agent/remoteagent/v2/client.go
```diff
@@ -0,0 +1,75 @@
...
+		var err error
+		if factory != nil {
+			client, err = factory.CreateFromCard(ctx, card)
+		} else {
+			client, err = a2aclient.NewFromCard(ctx, card)
+		}
+		if err != nil {
+			return nil, err
+		}
+		return &defaultA2AClient{client: client}, nil
```
baptmont: //why not
//return client, nil

baptmont (approved):

PR #708, opened 2026-04-07
Implement OnAgentErrorCallback for error metrics

jlapenna:
## Summary

This PR introduces an `OnAgentErrorCallback` mechanism to the ADK framework. This allows agents and plugins to register callbacks that are executed when an agent encounters an error during its execution.

Fixes #709

## Rationale

Currently, there is no standardized way for plugins to intercept and handle or record errors that occur during agent execution (e.g., in `agent.Run`). This makes it difficult to implement observability metrics for agent failures, such as counting invocation errors.

By adding `OnAgentErrorCallback`, we enable plugins to listen for these errors and perform actions like incrementing error counters, logging, or attempting recovery.

### `agent/agent.go`
- Added `OnAgentErrorCallback` type definition.
- Added `OnAgentErrorCallbacks` to `Config` struct.
- Updated `Run` method to trigger `runOnAgentErrorCallbacks` when `a.run(ctx)` yields an error.
- Implemented `runOnAgentErrorCallbacks` helper to execute callbacks from both the agent and registered plugins.
- Added `RunOnAgentErrorCallback` to the `pluginManager` interface.

### `plugin/plugin.go`
- Added `OnAgentErrorCallback` to `Config` and `Plugin` structs.
- Added accessor method `OnAgentErrorCallback()`.

### `internal/plugininternal/plugin_manager.go`
- Implemented `RunOnAgentErrorCallback` to iterate over registered plugins and execute their `OnAgentErrorCallback`.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -67,6 +67,7 @@ func New(cfg Config) (Agent, error) {
 		beforeAgentCallbacks: cfg.BeforeAgentCallbacks,
 		run:                  cfg.Run,
 		afterAgentCallbacks:  cfg.AfterAgentCallbacks,
+		onAgentErrorCallbacks: cfg.OnAgentErrorCallbacks,
 		State: agentinternal.State{
 			AgentType: agentinternal.TypeCustomAgent,
 		},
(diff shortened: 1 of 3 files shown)


jlapenna:
I think this is actually the wrong appraoch given the adk-python impl. i'll re-review and circl eback. thanks for the offline feedback!

PR #711, opened 2026-04-07
feat: Define skill.Source interface and implement file system skill source.

foxfrikses:
The source interface defines a contract via which skills framework is going to access skills.

### DIFF
--- a/tool/skilltoolset/skill/filesystem_source.go
+++ b/tool/skilltoolset/skill/filesystem_source.go
@@ -41,8 +41,8 @@ import (
 //		   SKILL.md
 //		   references/
 //		   scripts/
-func NewFileSystemSource(filesystem fs.FS) (Source, error) {
-	return &fileSystemSource{filesystem: filesystem}, nil
+func NewFileSystemSource(filesystem fs.FS) Source {
+	return &fileSystemSource{filesystem: filesystem}
 }
 
 type fileSystemSource struct {
--- a/tool/skilltoolset/skill/filesystem_source.go
+++ b/tool/skilltoolset/skill/filesystem_source.go
@@ -1,0 +1,228 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


tool/skilltoolset/skill/filesystem_source.go
```diff
@@ -0,0 +1,228 @@
...
+	return file, nil
+}
+
+// ListResources walks the specified resource directory within the skill.
+//
+// If resourceDirectoryPath is empty or ".", it walks the 'references/',
+// 'assets/', and 'scripts/' directories. It restricts traversal to these
+// approved directories and returns sanitized paths relative to the skill root.
+func (f *fileSystemSource) ListResources(ctx context.Context, name, resourceDirectoryPath string) ([]string, error) {
+	if err := f.validateSkill(name); err != nil {
```
dpasiukevich, line 142: Just being curious -- what is the usage pattern overall if you know?

Would it make sense to cache the results to avoid validations on every LoadResource/ListResources call? Just in case, we definitely don't need to do this now (unless we're 100% it's needed already)

Or even completely remove validateSkill call?

I'm asking, as if ListResources is called frequently, we're just doing lots of extra duplicated work for file-reads and cpu time.

dpasiukevich (approved):
LGTM!

tool/skilltoolset/skill/filesystem_source.go
foxfrikses: 1. Usage pattern

There is going to be no usage of ListResources initially, as it'll match the ADK Python experimental functionality, which AFAIU assumes that SKILL.md mentions all the resources the skill needs.
How exactly ListResources is going to be used in the end - I don't know.

One possible approach could be:
- an agent is given a tool wrapping this function,
- SKILL.md, say, mentions "you can find more info on this under 'assets/'
- the agent is able to list all the assets, without the need for SKILL.md to list them all.

2. Cache

Anyone can add cache for ListResources or LoadResource with a proxy, same as proxy caches we plan for frontmatters and instructions.

3. Validation

Agree this is not optimal. I think that might not cause any problems - removing the validation calls.
However, I might be missing something: my intuition screams when I try to remove it, thus I'd like to think it through yet for some time. For now I'd prefer to go with the safer approach rather than optimal. We can remove it later if we deem these validations serving no practical purpose.

PR #713, opened 2026-04-08
feat: add the implementation of Eventarc subrouter

anFatum:
Implementation of Cloud Events trigger processing.

This implementation should handle the resource exhausted exceptions with jitter to prevent thundering herd (configured per subrouter).

Similar to https://github.com/google/adk-go/pull/704

(see the details in [go/orcas-rfc-522](http://go/orcas-rfc-522))

### DIFF
--- a/server/adkrest/controllers/triggers/eventarc.go
+++ b/server/adkrest/controllers/triggers/eventarc.go
@@ -97,7 +97,7 @@ func (c *EventarcController) EventarcTriggerHandler(w http.ResponseWriter, r *ht
 		}
 		messageContent, err = messageContentFromPubSub(pubsub)
 		if err != nil {
-			respondError(w, http.StatusBadRequest, err.Error())
+			respondError(w, http.StatusBadRequest, fmt.Sprintf("failed to retrieve message content: %v", err))
 			return
 		}
 	} else {
@@ -113,7 +113,7 @@ func (c *EventarcController) EventarcTriggerHandler(w http.ResponseWriter, r *ht
 
 	appName, err := appName(r)
 	if err != nil {
-		respondError(w, http.StatusInternalServerError, err.Error())
+		respondError(w, http.StatusInternalServerError, fmt.Sprintf("failed to retrieve app name: %v", err))
 		return
 	}
 
--- a/server/adkrest/controllers/triggers/eventarc.go
+++ b/server/adkrest/controllers/triggers/eventarc.go
@@ -62,7 +62,7 @@ func (c *EventarcController) EventarcTriggerHandler(w http.ResponseWriter, r *ht
 		// The entire event is in the body. Decode it.
 		// The payload (Storage or Pub/Sub) gets safely trapped in event.Data as bytes.
 		if err := json.NewDecoder(r.Body).Decode(&event); err != nil {
-			http.Error(w, "Bad Request", http.StatusBadRequest)
+			respondError(w, http.StatusBadRequest, fmt.Sprintf("failed to unmarshal eventarc request: %v", err))
 			return
 		}
 	} else {
@@ -78,7 +78,7 @@ func (c *EventarcController) EventarcTriggerHandler(w http.ResponseWriter, r *ht
 		// We just read it as raw bytes into event.Data.
 		bodyBytes, err := io.ReadAll(r.Body)
 		if err != nil {
-			http.Error(w, "Failed to read body", http.StatusInternalServerError)
+			respondError(w, http.StatusInternalServerError, fmt.Sprintf("failed to read body: %v", err))
 			return
 		}
 		event.Data = bodyBytes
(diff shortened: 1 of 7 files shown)


server/adkrest/controllers/triggers/eventarc.go
```diff
@@ -0,0 +1,137 @@
...
+	var event models.EventarcTriggerRequest
+	contentType := r.Header.Get("Content-Type")
+	// The HTTP Content-Type header MUST be set to the media type of an event format for structured mode.
+	// https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/http-protocol-binding.md#321-http-content-type
+	if contentType == "application/cloudevents+json" {
+		// --- STRUCTURED MODE ---
+		// The entire event is in the body. Decode it.
+		// The payload (Storage or Pub/Sub) gets safely trapped in event.Data as bytes.
+		if err := json.NewDecoder(r.Body).Decode(&event); err != nil {
+			http.Error(w, "Bad Request", http.StatusBadRequest)
```
kdroste-google: how about returning the error to the caller (respondError?), or logging the error?

server/adkrest/controllers/triggers/pubsub.go
```diff
@@ -52,22 +52,47 @@ func NewPubSubController(sessionService session.Service, agentLoader agent.Loade
...
+	// Parse the request to the request model.
+	var req models.PubSubTriggerRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		respondError(w, http.StatusBadRequest, fmt.Sprintf("failed to decode request: %v", err))
+		return
+	}
+
+	agentMessage, err := messageContentFromPubSub(req)
+	if err != nil {
+		respondError(w, http.StatusBadRequest, err.Error())
```
kdroste-google: adding the info about a failure in messageContentFromPubSub may help debugging

server/adkrest/controllers/triggers/pubsub.go
```diff
@@ -52,22 +52,47 @@ func NewPubSubController(sessionService session.Service, agentLoader agent.Loade
...
+		return
+	}
+
+	appName, err := appName(r)
+	if err != nil {
+		respondError(w, http.StatusInternalServerError, err.Error())
```
kdroste-google: adding the info about a failure in appName may help debugging

server/adkrest/controllers/triggers/pubsub.go
```diff
@@ -76,26 +101,12 @@ func (c *PubSubController) PubSubTriggerHandler(w http.ResponseWriter, r *http.R
 	}
 
 	if len(messageContent) == 0 {
-		respondError(w, http.StatusBadRequest, "empty message data and attributes")
-		return
+		return "", fmt.Errorf("empty message data and attributes")
 	}
 
 	agentMessage, err := json.Marshal(messageContent)
```
kdroste-google, line 107: I'm not sure about handling empty req.Message.Data. Should we care? Is there any difference if we provide empty data vs not provide data at all?

kdroste-google (changes requested):

server/adkrest/controllers/triggers/pubsub.go
anFatum: I think we don't really care if there is no data, but the pub/sub data field definition says "Optional. The message data field. If this field is empty, the message must contain at least one attribute." (https://docs.cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage) 

so if we receive the empty data at all it might be incorrect?

kdroste-google: It's not clear to me. Just an edge-case, we can change it later on.

server/adkrest/controllers/triggers/eventarc.go
```diff
@@ -0,0 +1,137 @@
...
+		var pubsub models.PubSubTriggerRequest
+		var err error
+		// Unmarshal the raw bytes into our specific Pub/Sub struct
+		if err := json.Unmarshal(event.Data, &pubsub); err != nil {
+			respondError(w, http.StatusInternalServerError, fmt.Sprintf("failed to unmarshal pubsub data: %v", err))
+			return
+		}
+		messageContent, err = messageContentFromPubSub(pubsub)
+		if err != nil {
+			respondError(w, http.StatusBadRequest, err.Error())
```
kdroste-google: adding info about a failure in messageContentFromPubSub may help debugging

server/adkrest/controllers/triggers/eventarc.go
```diff
@@ -0,0 +1,137 @@
...
+		if err != nil {
+			respondError(w, http.StatusInternalServerError, fmt.Sprintf("failed to marshal agent message: %v", err))
+			return
+		}
+		messageContent = string(messageBytes)
+	}
+
+	appName, err := appName(r)
+	if err != nil {
+		respondError(w, http.StatusInternalServerError, err.Error())
```
kdroste-google: adding info about a failure in appName may help debugging

kdroste-google (dismissed):
Please add debugging info (indicated in comments)

kdroste-google (approved):
LGTM, thank you!

PR #721, opened 2026-04-10
fix: Allow manual setting of session IDs

dietb:
**Problem:**
The `VertexAiSessionService.Create` method rejects user-provided session IDs by returning an error when `req.SessionID` is populated. The vertex `CreateSessionRequest` proto supports an optional `SessionId` field, meaning the API  allows callers to specify their own session IDs. This prevents users from controlling session identity, causing numerous issues, namely with a2a implementations that would otherwise require a user to implement something to map UUIDs to the auto-generated vertex ID format.

**Solution:**
Remove the validation check in `vertexai.go` that rejected user-provided session IDs, and pass `req.SessionID` through to the `CreateSessionRequest` proto's `SessionId` field in `vertexai_client.go`. When `SessionID` is empty, the server generates one automatically. 

This also brings this package in line with the [python implementation](https://github.com/google/adk-python/blob/main/src/google/adk/sessions/vertex_ai_session_service.py#L103).

### Testing Plan

**Unit Tests:**

1. Create a `VertexAiSessionService` with valid project/location/reasoning engine config.
2. Call `Create` with an empty `SessionID` -- verify the server generates a session ID (unchanged behavior).
3. Call `Create` with a user-provided `SessionID` -- verify the session is created with the specified ID.
4. Call `Get` using the user-provided session ID -- verify the session is retrievable.

### Checklist


kdroste-google:
Hi, @dietb, thank you for your contribution! 
Could you please rebase the PR with the latest main branch so we can proceed?

dietb:
@kdroste-google done, but there's one more thing that should be done to enable some existing tests.

1. Someone with access to EngineId 5576569044451983360 should run `UPDATE_REPLAYS=true go test -run Test_vertexaiService/Create/full_key -v`
2.  [This](https://github.com/google/adk-go/blob/81a63d8feb7d713b1731f0c740d95574eb64dafa/session/vertexai/service_test.go#L44) should be changed to true and the trailing comment on line 47 should be removed.

I could write tests that'd handle this otherwise, but I don't want to break existing patterns.  If you see another path forward, let me know.

session/session_test/service_suite.go
```diff
@@ -385,7 +386,10 @@ func RunServiceTests(t *testing.T, opts SuiteOptions, setup func(t *testing.T) s
 				SessionID: "nonExistent",
 			})
 			if err != nil {
-				t.Errorf("Delete() non-existent error = %v, want nil", err)
+				// VertexAI fails on Delete when the sessionID is not found, we accept that
+				if !strings.Contains(err.Error(), "Invalid Session resource name") {
```
baptmont: Can we avoid stringly typed error handling?
Would something like this work:

``` golang
import (
    "google.golang.org/grpc/status"
    "google.golang.org/grpc/codes"
)

// Extract the gRPC status from the error
if st, ok := status.FromError(err); ok {
    // Check the actual error code (e.g., InvalidArgument, NotFound)
    if st.Code() != codes.InvalidArgument {
        // Handle other gRPC errors
    }
} else {
    // Handle non-gRPC errors
}
```

baptmont (dismissed):

session/session_test/service_suite.go
dietb: Good catch.  I just pushed a fix for this.

baptmont (approved):

kdroste-google (approved):

kdroste-google:
@dietb, thank you!

PR #726, opened 2026-04-12
test(adkrest/models): add EventActions field mapping coverage

### LINKED ISSUE #509: [bug] REST API EventActions doesn't include TransferToAgent, Escalate, SkipSummarization fields
### Issue Template 1: TransferToAgent Missing

**Title**: `[bug] REST API EventActions doesn't include TransferToAgent, Escalate, SkipSummarization fields`

**Labels**: `bug`

**Body**:
```markdown
### Description

The REST API model `EventActions` in `server/adkrest/internal/models/event.go` is missing the `TransferToAgent`, `Escalate`, and `SkipSummarization` fields that exist in `session.EventActions`.

This prevents the web UI from knowing when an agent is transferring to another agent or escalating.

### Expected Behavior

The `EventActions` struct in the REST models should include all fields from `session.EventActions`:
- `TransferToAgent string`
- `Escalate bool`
- `SkipSummarization bool`

### Actual Behavior

Currently only `StateDelta` and `ArtifactDelta` are serialized:

```go
// server/adkrest/internal/models/event.go (lines 25-28)
type EventActions struct {
    StateDelta    map[string]any   `json:"stateDelta"`
    ArtifactDelta map[string]int64 `json:"artifactDelta"`
    // Missing: TransferToAgent, Escalate, SkipSummarization
}
```

Compare with `session.EventActions` (session/session.go lines 138-152):
```go
type EventActions struct {
    StateDelta        map[string]any
    ArtifactDelta     map[string]int64
    SkipSummarization bool
    TransferToAgent   string  // NOT serialized in REST API
    Escalate          bool    // NOT serialized in REST API
}
```

### Suggested Fix

Update `EventActions` in `server/adkrest/internal/models/event.go`:

```go
type EventActions struct {
    StateDelta        map[string]any   `json:"stateDelta"`
    ArtifactDelta     map[string]int64 `json:"artifactDelta"`
    TransferToAgent   string           `json:"transferToAgent,omitempty"`
    Escalate          bool             `json:"escalate,omitempty"`
    SkipSummarization bool             `json:"skipSummarization,omitempty"`
}
```

Also update `FromSessionEvent()` and `ToSessionEvent()` to include these fields.

### Environment

- ADK version: v0.2.0
- Go version: 1.24
```

### PR DESCRIPTION
**Pull request by nuthalapativarun**
### Description of Change

**Problem:**
`FromSessionEvent` and `ToSessionEvent` were missing `TransferToAgent`, `Escalate`, and `SkipSummarization` fields in the REST API `EventActions` struct (issue #509). The fix was merged in #609, but there were no tests to prevent regressions.

**Solution:**
Add table-driven tests covering all three fields for both `FromSessionEvent` and `ToSessionEvent`, plus a round-trip test (`FromSessionEvent` → `ToSessionEvent`) to verify no `EventActions` fields are lost.

Closes #509

### Testing Plan

**Unit Tests:**

- [x] Added `TestFromSessionEvent_EventActions` — 5 table cases (all fields, each field individually, zero value)
- [x] Added `TestToSessionEvent_EventActions` — same 5 cases for reverse direction
- [x] Added `TestFromSessionEvent_RoundTrip` — full round-trip test

```
$ go test ./server/adkrest/internal/models/...
ok  	google.golang.org/adk/server/adkrest/internal/models	0.209s
```

### Checklist

### Additional context

This PR adds regression test coverage only — no production code changes. The underlying fix was already merged in #609.

---
_Authored by Varun Nuthalapati using AI coding assistance (Claude)_


nuthalapativarun:
Hi, just checking in on this one — happy to make any adjustments if something needs to change. Let me know if there's anything blocking review. Thanks!

nuthalapativarun:

nuthalapativarun:

nuthalapativarun:

karolpiotrowicz:
Thanks for adding this coverage — it's a useful gap to close, and the tests look clean: both conversion directions plus a round-trip, fully offline, and passing locally.

One thing I'd like to tighten before approving:

The doc comments say the tests verify "all EventActions fields" / that an event round-trips "without losing any EventActions fields", but `session.EventActions.RequestedToolConfirmations` isn't carried by the REST `EventActions` model in `server/adkrest/internal/models/event.go`, so it's silently dropped in the conversion and isn't exercised here. As written, the comments overclaim what's actually covered.

Could you either:
- soften the wording to something like "the mapped EventActions fields", or
- add a case that populates `RequestedToolConfirmations` and documents the intended omission?

(If dropping that field on the REST surface is *not* intended, that's worth a separate issue/follow-up rather than folding it into this PR.)

A couple of minor, optional notes:
- The all-zero-value table rows are essentially tautological — fine to keep, but they don't add much beyond the populated cases.
- I wasn't able to run `golangci-lint` locally, so I'll defer to CI on lint.

Once the comment/coverage is tightened up I'm happy to approve. Thanks again!

nuthalapativarun:
Thanks for the thorough review @karolpiotrowicz! You're right — the wording overclaimed. I've updated the comments to say "the mapped EventActions fields" instead of "all EventActions fields", which accurately reflects that `RequestedToolConfirmations` is intentionally not carried in the REST conversion. Happy to open a follow-up issue for that field if you think it's worth tracking. Ready for re-review when you get a chance!

karolpiotrowicz (approved):
Apologies for the long silence here — you made the change I asked for back in June and this then sat for far too long. I've re-validated it against current `main` rather than the base it was written on, and it holds up.

**It still works after the drift.** Your branch is 123 commits behind, so I rebased it locally onto [`53a0bc0`](https://github.com/google/adk-go/commit/53a0bc0ceb2e4437a075631f33a76f039445745f) and ran it there: clean rebase with no conflict, `go build ./...` and the package suite both green, five tests passing — your three alongside the two that landed on `main` in the meantime.

**It still pins real behaviour.** I broke the mapping code nine different ways and re-ran the package suite twice for each, once with your file and once without. All nine fail with your file and pass without it, so every one is coverage `main` does not have today. The nicest of them is the `Escalate`/`SkipSummarization` source swap: your `"all fields populated"` case sets both to `true` and your `"zero value"` case sets both to `false`, so neither can see a swap — it is caught only because you also wrote the isolated single-field cases. Those cases look redundant next to the combined one, which is exactly why they usually get dropped.

**No overlap with what landed since.** `event_test.go` grew while this was waiting, but it covers `RequestedToolConfirmations`, `Compaction` and the workflow fields. Yours covers `TransferToAgent`, `Escalate`, `SkipSummarization`, `StateDelta` and `ArtifactDelta`. The two files are disjoint.

One thing worth knowing, since it changes the reasoning behind the wording fix rather than the fix itself. When I raised `RequestedToolConfirmations` in June it genuinely was not carried by the REST model. It is now, in both directions, and `main` has since added a seventh field, `Compaction`, which is deliberately outbound-only — `FromSessionEvent` maps it and `ToSessionEvent` drops it, so a client cannot erase conversation history and speak into the gap. Your hedged "the mapped EventActions fields" is still accurate, just for a different reason than when you wrote it.

That does leave one sharp edge in `TestFromSessionEvent_RoundTrip`. Its comment says the event round-trips "without losing any mapped EventActions fields", and `Compaction` is a mapped field that intentionally does not survive the trip. If someone later adds `Compaction` to that fixture to make the test match its own comment, the failure they get reads like data loss in `ToSessionEvent`, and the obvious repair is the one that removes the protection. I checked both steps and the guard rail holds — `TestCompactionIsReadOnlyOverREST` fails loudly if anyone tries it — so this costs a future contributor an hour, not a security hole. Worth noting too that `event_test.go` on `main` makes a stronger unqualified version of the same claim, so this is ours as much as yours. A clause naming `Compaction` as the exception would close it, in either file.

I've left two smaller notes on the lines themselves, about a subtest name and an unasserted fixture. Both are optional.

None of this blocks. If you'd rather not do another round after this wait, say so and I'll take the polish as a follow-up — please just rebase onto current `main` so it can go in.

Separately, and not for this PR: while checking what your tests pin I found several gaps that predate them. Nothing in the package marshals anything, so renaming a JSON tag on `EventActions` passes the entire suite while breaking every REST client. The nil-versus-empty map distinction that `session.EventActions.MarshalJSON` deliberately preserves is untested, as is whether the mappers alias or copy the caller's maps, and a new field mapped in only one direction would go unnoticed. I'll open an issue against the package for those — they are not yours to fix.

PR #732, opened 2026-04-15
feat: add retry wrapper with exponential backoff for LLM calls (#710)

chaojixinren:
                                                                                                                          
  - Closes: #710                                                                                                                                
                                                                                                                        
  **Problem:**                                                                                                          
  LLM API calls can fail with transient errors (429 rate limiting, 503 service unavailable, RESOURCE_EXHAUSTED) causing 
  unnecessary request failures. There is currently no built-in retry mechanism in the model layer to handle these       
  recoverable errors gracefully.                                                                                        
                                                            
  **Solution:**                                                                                                         
  Add a `WithRetry()` LLM decorator that wraps any `model.LLM` implementation with automatic retry logic using 
  configurable exponential backoff and jitter, aligned with adk-python's retry behavior                                 
  (`tenacity.wait_exponential_jitter`).                     

  Key design decisions:                                                                                                 
  - **Decorator pattern**: `WithRetry(llm, cfg)` returns a new `LLM`, composable with any existing implementation 
  without modifying it. Unlike adk-python where retry logic is embedded in each LLM class, the Go version uses this     
  pattern to avoid duplication across `gemini` and `apigee` packages, following idiomatic Go composition.
  - **Aligned with adk-python defaults**: 5 retries, 1s initial delay, 60s max delay, 2x multiplier, 1s absolute jitter.
  - **Retryable errors**: HTTP 408/429/500/502/503/504, gRPC UNAVAILABLE/RESOURCE_EXHAUSTED, and network errors         
  (connection refused/reset, DNS failure, i/o timeout) — matching adk-python's `default_status_codes` and
  `httpx.NetworkError`.                                                                                                 
  - **Stream-safe**: Streaming calls only retry before the first yielded response to prevent duplicate partial content. 
  - **Context-aware**: Backoff sleeps respect context cancellation.                                                     
  - **Customizable**: Users can provide their own `IsRetryable` function for custom error classification.               
                                                                                                                        
  ### Testing Plan                                                                                                      
                                                                                                                        
  **Unit Tests:**                                                                                                       
                                                                                                                        
  13 test cases with 18 sub-tests covering:                                                                             
  - Unary: success without retry, retry on transient error, non-retryable error passthrough, retry exhaustion, context
  cancellation, custom `IsRetryable`                                                                                    
  - Streaming: success without retry, retry before first data, no retry after partial data
  - Config: default values, backoff growth, backoff max cap                                                             
  - Error classification: table-driven tests for all HTTP status codes, gRPC errors, and network errors                 
  - Name delegation                                                                                                                 
   
<img width="750" height="813" alt="截屏2026-04-15 22 24 00" src="https://github.com/user-attachments/assets/e3e9c53d-4aa0-4546-b3c9-396b12f581c3" />
                                                                                                                        
  ```go                                                                                                                 
  // Default config (aligned with adk-python):
  llm := model.WithRetry(baseLLM, nil)                                                                                  
                                                            
  // Custom config:                       
  llm := model.WithRetry(baseLLM, &model.RetryConfig{                                                                   
      MaxRetries:   3,                                                                                                  
      InitialDelay: 2 * time.Second,                                                                                    
      Jitter:       500 * time.Millisecond,                                                                             
  })                                                        
```
  The wrapped LLM is a drop-in replacement for any existing model.LLM. When a retryable error occurs, the retry log     
  output looks like:                      
                                                                                                                        
  WARN retrying LLM call model=gemini-2.5-flash attempt=1 max_retries=5 delay=1.536s error="Error 503, Message: high    
  demand, Status: UNAVAILABLE"                
                                                                                                                        
  Checklist                                                 
                                                                                                                        
  - I have read the CONTRIBUTING.md document.
  - I have performed a self-review of my own code.                                                                      
  - I have commented my code, particularly in hard-to-understand areas.
  - I have added tests that prove my fix is effective or that my feature works.
  - New and existing unit tests pass locally with my changes.
  - I have manually tested my changes end-to-end.                                                                       
  - Any dependent changes have been merged and published in downstream modules.
                                                                                                                        
   Additional context                                                                                                
                                                                                                                       
  **Design note — standalone `model/retry.go` vs embedding in each LLM:**                                               
                                                                                                                        
  In adk-python, retry logic is embedded in each LLM class (`google_llm.py` delegates to SDK's `HttpRetryOptions`,      
  `apigee_llm.py` uses tenacity). The Go version extracts it into a standalone decorator because:                       
                                                                                                                        
  1. Go's `model.LLM` is a single interface — a decorator wraps any implementation without modifying it.                
  2. Avoids duplicating retry logic across `model/gemini/` and `model/apigee/` packages.                                
  3. Follows idiomatic Go composition (`WithRetry(llm, cfg)` is the same pattern as `io.LimitReader`,                   
  `http.TimeoutHandler`, etc.).                                                         
                                                                                                                        
  New files:                                                                                                            
  - `model/retry.go` — `RetryConfig`, `WithRetry()`, `retryLLM` decorator, `defaultIsRetryable`, `sleepCtx`
  - `model/retry_test.go` — 13 test cases with `fakeLLM` test double

### DIFF
--- a/agent/llmagent/llmagent_test.go
+++ b/agent/llmagent/llmagent_test.go
@@ -93,6 +93,21 @@ func TestLLMAgent(t *testing.T) {
 	}
 }
 
+func TestLLMAgentRetryWithoutModelReturnsConfigurationError(t *testing.T) {
+	a, err := llmagent.New(llmagent.Config{
+		Name:  "missing_model",
+		Retry: &model.RetryConfig{},
+	})
+	if err != nil {
+		t.Fatalf("NewLLMAgent failed: %v", err)
+	}
+
+	_, runErr := testutil.CollectEvents(testutil.NewTestAgentRunner(t, a).Run(t, "session", "hello"))
+	if runErr == nil || !strings.Contains(runErr.Error(), "model not configured") {
+		t.Fatalf("run error = %v, want model-not-configured error", runErr)
+	}
+}
+
 func TestLLMAgentStreamingModeSSE(t *testing.T) {
 	model := newGeminiModel(t, "gemini-2.5-flash", nil)
 	a, err := llmagent.New(llmagent.Config{
--- a/model/retry.go
+++ b/model/retry.go
@@ -1,0 +1,260 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package model
+
+import (
+	"context"
+	"iter"
+	"log/slog"
+	"math"
+	"math/rand/v2"
+	"strings"
+	"time"
+)
+
+// RetryConfig controls retry behavior for transient LLM errors.
+type RetryConfig struct {
+	// MaxRetries is the maximum number of retry attempts. Defaults to 5.
+	MaxRetries int
+	// InitialDelay is the delay before the first retry. Defaults to 1s.
+	InitialDelay time.Duration
+	// MaxDelay caps the backoff delay. Defaults to 60s.
+	MaxDelay time.Duration
+	// Multiplier is the exponential backoff multiplier. Defaults to 2.0.
+	Multiplier float64
+	// Jitter is the maximum absolute random delay added on top of the
+	// exponential backoff. Defaults to 1s.
+	Jitter time.Duration
+	// IsRetryable determines whether an error should be retried.
+	// When nil, the default policy retries 408, 429, 500, 502, 503, 504,
+	// UNAVAILABLE, RESOURCE_EXHAUSTED, and network-related errors.
+	IsRetryable func(error) bool
+}
+
+func (c *RetryConfig) maxRetries() int {
+	if c != nil && c.MaxRetries > 0 {
+		return c.MaxRetries
+	}
+	return 5
+}
+
+func (c *RetryConfig) initialDelay() time.Duration {
+	if c != nil && c.InitialDelay > 0 {
+		return c.InitialDelay
+	}
+	return time.Second
+}
+
+func (c *RetryConfig) maxDelay() time.Duration {
+	if c != nil && c.MaxDelay > 0 {
+		return c.MaxDelay
+	}
+	return 60 * time.Second
+}
+
+func (c *RetryConfig) multiplier() float64 {
+	if c != nil && c.Multiplier > 0 {
+		return c.Multiplier
+	}
+	return 2.0
+}
+
+func (c *RetryConfig) jitter() time.Duration {
+	if c != nil && c.Jitter > 0 {
+		return c.Jitter
+	}
+	return time.Second
+}
+
+func (c *RetryConfig) isRetryable(err error) bool {
+	if c != nil && c.IsRetryable != nil {
+		return c.IsRetryable(err)
+	}
+	return defaultIsRetryable(err)
+}
+
+// backoff computes the delay for the given zero-based attempt number.
+// Formula matches tenacity.wait_exponential_jitter:
+//
+//	min(initial * multiplier^attempt, maxDelay) + random(0, jitter)
+func (c *RetryConfig) backoff(attempt int) time.Duration {
+	base := float64(c.initialDelay()) * math.Pow(c.multiplier(), float64(attempt))
+	delay := math.Min(base, float64(c.maxDelay()))
+	j := rand.Float64() * float64(c.jitter())
+	return time.Duration(delay + j)
+}
+
+var retryableSubstrings = []string{
+	// HTTP status codes (aligned with adk-python default_status_codes).
+	"408",
+	"429",
+	"500",
+	"502",
+	"503",
+	"504",
+	// gRPC status names.
+	"UNAVAILABLE",
+	"RESOURCE_EXHAUSTED",
+	"ResourceExhausted",
+	"ServiceUnavailable",
+	// Network errors (aligned with adk-python httpx.NetworkError).
+	"connection refused",
+	"connection reset",
+	"no such host",
+	"i/o timeout",
+	"network is unreachable",
+}
+
+func defaultIsRetryable(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	for _, sub := range retryableSubstrings {
+		if strings.Contains(msg, sub) {
+			return true
+		}
+	}
+	return false
+}
+
+// WithRetry wraps an LLM with automatic retry logic using exponential backoff
+// with jitter. Pass nil for cfg to use sensible defaults.
+func WithRetry(llm LLM, cfg *RetryConfig) LLM {
+	return &retryLLM{inner: llm, cfg: cfg}
+}
+
+type retryLLM struct {
+	inner LLM
+	cfg   *RetryConfig
+}
+
+func (r *retryLLM) Name() string {
+	return r.inner.Name()
+}
+
+func (r *retryLLM) GenerateContent(ctx context.Context, req *LLMRequest, stream bool) iter.Seq2[*LLMResponse, error] {
+	if stream {
+		return r.generateContentStream(ctx, req)
+	}
+	return r.generateContentUnary(ctx, req)
+}
+
+// generateContentUnary retries the full unary call on retryable errors.
+func (r *retryLLM) generateContentUnary(ctx context.Context, req *LLMRequest) iter.Seq2[*LLMResponse, error] {
+	return func(yield func(*LLMResponse, error) bool) {
+		maxRetries := r.cfg.maxRetries()
+		var lastErr error
+
+		for attempt := range maxRetries + 1 {
+			if attempt > 0 {
+				delay := r.cfg.backoff(attempt - 1)
+				slog.WarnContext(ctx, "retrying LLM call",
+					"model", r.inner.Name(),
+					"attempt", attempt,
+					"max_retries", maxRetries,
+					"delay", delay,
+					"error", lastErr,
+				)
+				if err := sleepCtx(ctx, delay); err != nil {
+					yield(nil, lastErr)
+					return
+				}
+			}
+
+			var resp *LLMResponse
+			var callErr error
+			for r, e := range r.inner.GenerateContent(ctx, req, false) {
+				resp, callErr = r, e
+			}
+
+			if callErr == nil {
+				yield(resp, nil)
+				return
+			}
+
+			lastErr = callErr
+			if !r.cfg.isRetryable(callErr) {
+				yield(nil, callErr)
+				return
+			}
+		}
+
+		yield(nil, lastErr)
+	}
+}
+
+// generateContentStream retries only when the stream fails before yielding
+// any successful response, preventing duplicate partial content.
+func (r *retryLLM) generateContentStream(ctx context.Context, req *LLMRequest) iter.Seq2[*LLMResponse, error] {
+	return func(yield func(*LLMResponse, error) bool) {
+		maxRetries := r.cfg.maxRetries()
+		var lastErr error
+
+		for attempt := range maxRetries + 1 {
+			if attempt > 0 {
+				delay := r.cfg.backoff(attempt - 1)
+				slog.WarnContext(ctx, "retrying LLM stream",
+					"model", r.inner.Name(),
+					"attempt", attempt,
+					"max_retries", maxRetries,
+					"delay", delay,
+					"error", lastErr,
+				)
+				if err := sleepCtx(ctx, delay); err != nil {
+					yield(nil, lastErr)
+					return
+				}
+			}
+
+			yieldedData := false
+			shouldRetry := false
+
+			for resp, err := range r.inner.GenerateContent(ctx, req, true) {
+				if err != nil {
+					if !yieldedData && r.cfg.isRetryable(err) {
+						lastErr = err
+						shouldRetry = true
+						break
+					}
+					yield(nil, err)
+					return
+				}
+				yieldedData = true
+				if !yield(resp, nil) {
+					return
+				}
+			}
+
+			if !shouldRetry {
+				return
+			}
+		}
+
+		yield(nil, lastErr)
+	}
+}
+
+// sleepCtx blocks for d or until ctx is cancelled, whichever comes first.
+func sleepCtx(ctx context.Context, d time.Duration) error {
+	t := time.NewTimer(d)
+	defer t.Stop()
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-t.C:
+		return nil
+	}
+}
--- a/model/retry.go
+++ b/model/retry.go
@@ -1,0 +1,273 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package model
+
+import (
+	"context"
+	"iter"
+	"log/slog"
+	"math"
+	"math/rand/v2"
+	"regexp"
+	"strings"
+	"time"
+
+	"google.golang.org/genai"
+)
+
+// RetryConfig controls retry behavior for transient LLM errors.
+type RetryConfig struct {
+	// MaxRetries is the maximum number of retry attempts. Defaults to 5.
+	MaxRetries int
+	// InitialDelay is the delay before the first retry. Defaults to 1s.
+	InitialDelay time.Duration
+	// MaxDelay caps the backoff delay. Defaults to 60s.
+	MaxDelay time.Duration
+	// Multiplier is the exponential backoff multiplier. Defaults to 2.0.
+	Multiplier float64
+	// Jitter is the maximum absolute random delay added on top of the
+	// exponential backoff. Defaults to 1s.
+	Jitter time.Duration
+	// IsRetryable determines whether an error should be retried.
+	// When nil, the default policy retries 408, 429, 500, 502, 503, 504,
+	// UNAVAILABLE, RESOURCE_EXHAUSTED, and network-related errors.
+	IsRetryable func(error) bool
+}
+
+func (c *RetryConfig) maxRetries() int {
+	if c != nil && c.MaxRetries > 0 {
+		return c.MaxRetries
+	}
+	return 5
+}
+
+func (c *RetryConfig) initialDelay() time.Duration {
+	if c != nil && c.InitialDelay > 0 {
+		return c.InitialDelay
+	}
+	return time.Second
+}
+
+func (c *RetryConfig) maxDelay() time.Duration {
+	if c != nil && c.MaxDelay > 0 {
+		return c.MaxDelay
+	}
+	return 60 * time.Second
+}
+
+func (c *RetryConfig) multiplier() float64 {
+	if c != nil && c.Multiplier > 0 {
+		return c.Multiplier
+	}
+	return 2.0
+}
+
+func (c *RetryConfig) jitter() time.Duration {
+	if c != nil && c.Jitter > 0 {
+		return c.Jitter
+	}
+	return time.Second
+}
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: model/retry.go [hunk 1]
```diff
@@ -0,0 +1,260 @@
...
+func defaultIsRetryable(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	for _, sub := range retryableSubstrings {
+		if strings.Contains(msg, sub) {
+			return true
+		}
+	}
```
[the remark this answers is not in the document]
chaojixinren: Thanks for the suggestion. The substring matching approach is intentional here for two reasons:                       
                                                                                                                        
  1. Alignment with adk-python: The Python implementation uses similar broad matching for error detection. The Go       
  model.LLM interface is provider-agnostic — the underlying implementation may not use HTTP at all, so we can't rely on 
  errors.As with net.Error or typed HTTP status checks.                                                                 
  2. Escape hatch exists: The IsRetryable callback allows users to provide precise error classification when the default
   policy is too broad for their use case.                                                                              
   
  In practice, LLM API errors follow predictable formats (e.g., "Error 503, Message: ..., Status: UNAVAILABLE", "429 Too
   Many Requests"), making false positives from unrelated numeric matches unlikely.

model/retry.go
```diff
@@ -0,0 +1,260 @@
...
+					"model", r.inner.Name(),
+					"attempt", attempt,
+					"max_retries", maxRetries,
+					"delay", delay,
+					"error", lastErr,
+				)
+				if err := sleepCtx(ctx, delay); err != nil {
+					yield(nil, lastErr)
+					return
+				}
```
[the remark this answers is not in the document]
chaojixinren: Returning lastErr here is intentional — it matches the behavior of Python's tenacity with reraise=True (which
  adk-python configures in apigee_llm.py). When retries are interrupted, tenacity re-raises the last actual exception,  
  not the cancellation error.
                                                                                                                        
  From the caller's perspective: they already know they cancelled the context. The more useful information is what
  underlying LLM error was being retried, which would be lost if we returned the context error instead.

model/retry.go
```diff
@@ -0,0 +1,260 @@
...
```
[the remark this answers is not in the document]
chaojixinren: Same rationale as the unary path above — returning lastErr preserves the original LLM error that triggered the retry, 
  consistent with adk-python's tenacity reraise=True behavior. The caller can always check ctx.Err() separately if they
  need to distinguish cancellation from exhaustion.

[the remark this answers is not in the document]
chaojixinren: Same rationale as the unary path above — returning lastErr preserves the original LLM error that triggered the retry, 
  consistent with adk-python's tenacity reraise=True behavior. The caller can always check ctx.Err() separately if they
  need to distinguish cancellation from exhaustion.

cruvie:
I think make `retry` as a config of `llmagent.Config` could be better.

chaojixinren:
> I think make `retry` as a config of `llmagent.Config` could be better.

Done — added `Retry *model.RetryConfig` to `llmagent.Config`. When set, the constructor wraps the model via `model.WithRetry` automatically. Nil (default) means no retry, keeping backward compatibility.

karolpiotrowicz:
@wolo-lab worth a direction call, because the capability may already be available.

The change adds a public retry surface — `model.RetryConfig`, `model.WithRetry` and a `Retry` field on the agent config. The genai SDK we pin already performs retries, so this would give us two retry layers with independent budgets and no coordination between them. The question is whether we want our own surface at all, or documentation pointing at the SDK's.

chaojixinren:
@karolpiotrowicz @wolo-lab Implemented and pushed the follow-up fixes to this PR (rebased onto current `main`):

- Preserve nil-model behavior when `Retry` is configured; no panic, the normal model-not-configured error is returned.
- Preserve the optional `GoogleLLM` backend capability through the retry decorator so Gemini-specific output-schema processing and telemetry continue to work.
- Tighten default HTTP status matching to avoid retrying on unrelated numbers in error messages.
- Return the context cancellation/deadline error when interrupted during retry backoff.
- Added regression tests for all of the above.

Testing plan:
- `go test -race ./model ./agent/llmagent`
- `go mod tidy -diff`

Local build and tests pass. GitHub checks are running; `apidiff`, `guardrail`, CLA, and security checks have passed so far.

model/retry.go
```diff
@@ -0,0 +1,273 @@
...
+
+func (c *RetryConfig) isRetryable(err error) bool {
+	if c != nil && c.IsRetryable != nil {
+		return c.IsRetryable(err)
+	}
+	return defaultIsRetryable(err)
+}
+
+// backoff computes the delay for the given zero-based attempt number.
+// Formula matches tenacity.wait_exponential_jitter:
```
karolpiotrowicz, line 92: This claims parity with `tenacity.wait_exponential_jitter`, and the two formulas differ. tenacity computes `max(0, min(initial*exp + jitter, max))` — it clamps *after* adding jitter. Lines 97-99 here clamp first and then add, so the result can land one jitter-width above `MaxDelay`.

The pinned SDK does it tenacity's way at [`common.go:569-570`](https://github.com/googleapis/go-genai/blob/5e7adb3c9022e7b5f4eb6a76d0c277f569d0bbb0/common.go#L569-L570), under a comment making the same parity claim, so this diverges from both.

Separately, the doc on line 36 ("MaxDelay caps the backoff delay") reads as an absolute cap, while lines 40-41 and the formula just below are precise about jitter being added on top. Worth reconciling line 36 with them whichever way you take the clamp.

model/retry.go
```diff
@@ -0,0 +1,273 @@
...
+	return false
+}
+
+// WithRetry wraps an LLM with automatic retry logic using exponential backoff
+// with jitter. Pass nil for cfg to use sensible defaults.
+func WithRetry(llm LLM, cfg *RetryConfig) LLM {
+	if llm == nil {
+		return nil
+	}
+	return &retryLLM{inner: llm, cfg: cfg}
```
karolpiotrowicz, line 140: `WithRetry` keeps the caller's `*RetryConfig` pointer and reads every field on every attempt of every concurrent call, so a caller who mutates the config after construction races with in-flight calls. `gemini.NewModel` just above copies the caller's config for a related reason, and `workflow.DefaultRetryConfig()` returns a copy too.

A shallow copy here would be enough — the only reference-ish field is the `IsRetryable` func value. Worth noting no test currently shares a config across goroutines, so `-race` would not surface this as things stand.

model/retry.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	for _, sub := range retryableErrorSubstrings {
+		if strings.Contains(msg, sub) {
+			return true
+		}
+	}
+	return false
+}
+
+// WithRetry wraps an LLM with automatic retry logic using exponential backoff
```
karolpiotrowicz, line 134: Two things a caller would want stated here. With the defaults a failing call now sleeps roughly 31 seconds across 6 attempts before the error surfaces, which is a large change to a p99 for a one-line config addition.

And `genai.HTTPOptions.RetryOptions` is a second retry layer that composes with this one rather than replacing it — set both and the attempt counts multiply. I measured 9 HTTP requests for a single logical call with the SDK at 3 attempts and this wrapper at 3.

model/retry.go
```diff
@@ -0,0 +1,273 @@
...
+					return
+				}
+			}
+
+			if !shouldRetry {
+				return
+			}
+		}
+
+		yield(nil, lastErr)
```
karolpiotrowicz, line 259: No test reaches this line. The three streaming tests all exit earlier: two succeed on the first attempt and the third returns at the partial-data guard, so nothing drives a stream that fails retryably on every attempt.

That matters because of what the untested regression looks like — replacing this `yield` with a bare `return` keeps the suite green and turns an exhausted stream into a silent, successful, empty one. The unary twin at line 208 is pinned by `TestRetryUnary_ExhaustsRetries`, so this is the one gap of the pair.

model/retry.go
```diff
@@ -0,0 +1,273 @@
...
+
+// generateContentUnary retries the full unary call on retryable errors.
+func (r *retryLLM) generateContentUnary(ctx context.Context, req *LLMRequest) iter.Seq2[*LLMResponse, error] {
+	return func(yield func(*LLMResponse, error) bool) {
+		maxRetries := r.cfg.maxRetries()
+		var lastErr error
+
+		for attempt := range maxRetries + 1 {
+			if attempt > 0 {
+				delay := r.cfg.backoff(attempt - 1)
```
karolpiotrowicz, line 176: The exponential wiring is asserted only through the pure `backoff` function, never through the loop that calls it. No test observes which argument is passed here or measures the resulting delay sequence, so changing this to `backoff(0)` (constant delay, no growth) or to `backoff(attempt)` (every delay doubled) leaves the suite green.

Every unary test uses `InitialDelay: time.Millisecond`, so the difference is sub-millisecond and unasserted. An injectable sleep hook on `retryLLM` would let a test assert the sequence without depending on wall-clock time.

model/retry.go
```diff
@@ -0,0 +1,273 @@
...
+				)
+				if err := sleepCtx(ctx, delay); err != nil {
+					yield(nil, err)
+					return
+				}
+			}
+
+			var resp *LLMResponse
+			var callErr error
+			for r, e := range r.inner.GenerateContent(ctx, req, false) {
```
karolpiotrowicz, line 192: Two things on this line.

The range collapses a multi-yield inner sequence to its last pair, so a `model.LLM` that yields more than once in unary mode loses everything but the final response. All three in-repo models yield exactly once, so this is latent — but `WithRetry` is exported and takes any implementation.

Also, `r` here shadows the `*retryLLM` receiver. It compiles and is correct today, because the range expression resolves the outer `r` before the new binding takes effect. It only bites whoever next needs `r.cfg` inside this loop body.

agent/llmagent/llmagent_test.go
```diff
@@ -93,6 +93,21 @@ func TestLLMAgent(t *testing.T) {
 	}
 }
 
+func TestLLMAgentRetryWithoutModelReturnsConfigurationError(t *testing.T) {
```
karolpiotrowicz, line 96: This is the only test that sets `Retry`, and it sets no `Model`, so `WithRetry(nil, cfg)` returns a nil `model.LLM` and `m` ends up identical to `cfg.Model` either way. Deleting the wrapping in `llmagent.New` entirely leaves this test passing.

It is not a useless test — it does pin the nil guard in `WithRetry`, without which `retryLLM.Name()` would nil-deref here. But nothing yet shows the field has any effect on an agent that has a model. A stub that fails twice with a retryable error and then succeeds, asserting the run succeeds and the stub saw 3 calls, would cover the integration.

karolpiotrowicz (changes requested):
Setting `Retry` on an agent removes live/bidi streaming from it, so this needs a change before it can merge. The rest below is smaller, and one thing I want to flag early is that the obvious fix for the blocker is worse than the blocker — I applied it and it segfaults.

I ran the build, the full test suite, `gofmt`, `go vet` and an API-compatibility check on the merge result, since the branch is behind `main`. All clean: the API additions are compatible, and the only two failing packages (`telemetry`, `session/vertexai`) fail the same way on `main` without this change.

## Enabling retry disables live streaming

`retryLLM` implements `Name`, `GenerateContent` and `GetGoogleLLMVariant` ([`model/retry.go:143-159`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/model/retry.go#L143-L159)), but not `Client() *genai.Client`. [`base_flow.go:315-319`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/internal/llminternal/base_flow.go#L315-L319) type-asserts exactly that method before opening a live session and rejects anything without it, and [`gemini.go:199`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/model/gemini/gemini.go#L199) is the only production type that provides it. So an agent that works in live mode today stops working the moment `Retry` is set, with an error that blames the model.

Driving `Flow.RunLive` with the same model, wrapped and unwrapped:

```
=== RUN   TestRunLive_RetryWrapperBreaksLiveConnection/without_WithRetry
    bare live-capable model: RunLive returned a session, no error
=== RUN   TestRunLive_RetryWrapperBreaksLiveConnection/with_WithRetry
    same model wrapped in model.WithRetry: RunLive error = model does not support live connection
--- PASS
```

Nothing in CI can catch this: an API-compatibility check cannot see it, because the lost capability is a duck-typed optional interface rather than a declared symbol, and the repo's live tests never wrap their model in `WithRetry`.

The property a fix needs to hold is both-ways: for every optional capability the codebase asserts on a `model.LLM` — today `Client()` and `GetGoogleLLMVariant()` — `WithRetry(inner, cfg)` should satisfy it if and only if `inner` does.

**The second direction is the trap, and I checked it rather than guessing.** Forwarding `Client()` the way `GetGoogleLLMVariant` is forwarded looks right and is not, because Go method sets are static: a wrapped OpenAI or Apigee model would then also pass the assertion and return `nil`, and [`base_flow.go:396`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/internal/llminternal/base_flow.go#L396) dereferences it inside the goroutine at [`base_flow.go:330`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/internal/llminternal/base_flow.go#L330), which has no `recover()`. I applied that fix and ran it:

```
=== RUN   TestPrescribedFix_NonGoogleModelBaseline
    bare non-Google model: RunLive err = model does not support live connection
--- PASS
=== RUN   TestPrescribedFix_NonGoogleModelWrapped
    wrapped non-Google model WITH the prescribed fix: RunLive err = <nil>
    RunLive ACCEPTED a model that cannot serve live; now draining the session goroutine
[signal SIGSEGV: segmentation violation]
goroutine 10 [running]:
google.golang.org/adk/v2/internal/llminternal.(*Flow).RunLive.func1()
FAIL
```

A clean error becomes an unrecovered crash. `GetGoogleLLMVariant` tolerates unconditional forwarding only because `BackendUnspecified` is a harmless zero value, and `*genai.Client` has no equivalent. Which way to go is yours — a capability-matched wrapper, an `Unwrap()` the assertion site honors, or a nil check after the assertion all work, and they have different costs elsewhere.

## The wrapper can produce a response pair the model never produced

[`model/retry.go:190-197`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/model/retry.go#L190-L197) collapses the inner sequence into two locals and then takes the success branch when both are still nil, so an inner model that yields nothing produces one `(nil, nil)` pair. `model.LLM` documents no minimum yield count, and an implementation of that shape already exists in the tree at [`base_flow_live_test.go:49-51`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/internal/llminternal/base_flow_live_test.go#L49-L51).

Through `llmagent.New` and a real runner, with `Retry` as the only difference:

```
without Retry: 0 events, err=<nil>, no panic
with Retry:    0 events, err=node "with_retry" panicked: runtime error: invalid memory address or nil pointer dereference
```

To be precise about the blast radius: the workflow scheduler recovers that panic into a node error, so it does not take the process down.

The same `(nil, nil)` is reachable from configuration alone. `MaxRetries: math.MaxInt` makes `maxRetries + 1` at [`model/retry.go:174`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/model/retry.go#L174) wrap negative, `range` runs zero iterations, and control falls through to the same yield:

```
MaxRetries=math.MaxInt -> 0 inner calls, 1 yielded pairs, 1 of them (nil,nil)
```

Both want the same guard, plus a floor of one attempt however `MaxRetries` is spelled.

## The zero value means "default", and the other `RetryConfig` in this module says the opposite

[`model/retry.go:50`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/model/retry.go#L50) and [`:78`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/model/retry.go#L78) gate on `> 0`, so zero and negative both fall back to five retries and one second of jitter. Neither "no retries" nor "deterministic backoff" can be expressed:

```
MaxRetries:0 produced 6 calls to the inner model
MaxDelay=10ms Jitter=0 -> one backoff took 507.743995ms (config says at most 10ms)
```

On its own that is an ordinary Go convention and I would not raise it. What makes it worth settling is that this module already ships [`workflow.RetryConfig`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/workflow/config.go#L110), which documents the inverse on both fields: [`:111`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/workflow/config.go#L111) says "If 0 or 1, it means no retries", [`:126-128`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/workflow/config.go#L126-L128) says "Zero means deterministic delays", and [`:105-106`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/workflow/config.go#L105-L106) advertises the zero struct as "a valid 'no retry, no backoff, no jitter' policy". Two exported types with the same name, whose zero values mean opposite things. It is also the cheapest thing on this list to change now and a breaking change to fix later.

## The retryability check matches digits anywhere in the error text

[`model/retry.go:102`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/model/retry.go#L102) matches `\b(?:408|429|500|502|503|504)\b` against the whole rendered error. You clearly thought about this class — [`model/retry_test.go:448-449`](https://github.com/google/adk-go/blob/93c63e600a4cb3babcb6137c417c594f71179f29/model/retry_test.go#L448-L449) pins `"request took 500ms"` and `"project ID 123429"` as non-retryable — but the boundary lands in the wrong place for a realistic hard 400:

```
error text: Error 400, Message: thinkingBudget must be at most 24576 but got 500, Status: INVALID_ARGUMENT
result:     4 calls
```

A deterministic rejection is retried for about 31 seconds at the defaults before the real 400 reaches the caller.

## A note on scope, for the maintainers rather than for you

Worth raising because it affects whether this shape is the one we want, not because anything is wrong with the code here.

The pinned `google.golang.org/genai v1.70.0` already implements HTTP retry with exponential backoff and jitter ([`types.go:1866`](https://github.com/googleapis/go-genai/blob/5e7adb3c9022e7b5f4eb6a76d0c277f569d0bbb0/types.go#L1866), executed at [`common.go:479`](https://github.com/googleapis/go-genai/blob/5e7adb3c9022e7b5f4eb6a76d0c277f569d0bbb0/common.go#L479)), and it is reachable from adk-go today through the `*genai.ClientConfig` that `gemini.NewModel` already accepts. I confirmed that on `main` without this PR applied — a server returning 503 twice then 200 was driven to success in 3 HTTP attempts, and with no `RetryOptions` the same server produced 1 attempt and surfaced the error.

adk-python solves the same problem that way rather than with a decorator: [`google_llm.py:182`](https://github.com/google/adk-python/blob/6de43b05a07f8ca89909664b3a21049646d7336c/src/google/adk/models/google_llm.py#L182) exposes `retry_options` on the model and [`:434`](https://github.com/google/adk-python/blob/6de43b05a07f8ca89909664b3a21049646d7336c/src/google/adk/models/google_llm.py#L434) passes it to the client. Its one `tenacity` loop is in `apigee_llm.py`, on a path that bypasses the genai client entirely.

The real gap this PR fills is that SDK retry is Gemini-only, so it does nothing for `openaimodel`, `apigee`, or a user's own `model.LLM`. That is a good reason for a decorator to exist. The cost is that the two layers have independent budgets: with the SDK set to 3 attempts and this wrapper to 3, one logical call became 9 HTTP requests in a test I ran. Worth a sentence in the docs whichever way the scope question lands.

I have left the smaller items as inline comments.

karolpiotrowicz:
@wolo-lab — a direction call on this one before the author puts in more work.

This adds `model.RetryConfig`, `model.WithRetry` and `llmagent.Config.Retry`, which is new exported API in a core package from an outside contributor. Two things make it a scope question rather than a code question.

adk-python does not solve retries with a decorator. It exposes `retry_options` on the model and hands it straight to the SDK client — [`google_llm.py:182`](https://github.com/google/adk-python/blob/6de43b05a07f8ca89909664b3a21049646d7336c/src/google/adk/models/google_llm.py#L182) declares it and [`:434`](https://github.com/google/adk-python/blob/6de43b05a07f8ca89909664b3a21049646d7336c/src/google/adk/models/google_llm.py#L434) passes it into the client's HTTP options, both pinned at [6de43b05](https://github.com/google/adk-python/commit/6de43b05a07f8ca89909664b3a21049646d7336c).

adk-go already supports the same thing today, without this change. Retry flows through the `*genai.ClientConfig` that `gemini.NewModel` already accepts, using the SDK's [`HTTPRetryOptions`](https://github.com/googleapis/go-genai/blob/5e7adb3c9022e7b5f4eb6a76d0c277f569d0bbb0/types.go#L1866). I checked that on `main` with this PR not applied: a model configured with `RetryOptions` makes 3 HTTP attempts, and the same model without it makes 1.

So the question for you is whether we want a second, ADK-owned retry mechanism sitting alongside the SDK's, and whether `model` is where it belongs. That is a different question from whether this implementation is correct.

Separately, I have left a code review with three issues that block regardless of how the scope question lands. The clearest is that setting `Retry` drops `Client() *genai.Client`, which breaks live and bidi streaming.

PR #737, opened 2026-04-17
feat: Add the example of trigger processing

anFatum:
This example decodes the message sent from GCP Pub/Sub trigger and suggests the follow-up on that message.


wolo-lab:
The author has left the team. Let's start from scratch.

PR #744, opened 2026-04-20
feat: add MiniMax provider with M3 as default

octo-patch:
## Summary
Adds a MiniMax provider to ADK Go with MiniMax-M3 as the default model.

## Changes
- New package `model/minimax` implementing `model.LLM` over the MiniMax OpenAI-compatible API (`https://api.minimax.io`)
- Default model is `MiniMax-M3` (512K context, up to 128K output, supports image input)
- `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` are also exposed as named constants
- Unit tests cover constructor, default-model selection, system instruction, tool calls, error paths, and a `TestModelConstants` guard so the model ID set is stable

## Why
MiniMax-M3 is the latest MiniMax model. Going forward, new users get M3 by default while M2.7 / M2.7-highspeed remain available for callers that still want them.

## Testing
- Unit tests in `model/minimax/minimax_test.go` (mocked HTTP transport)
- Integration smoke-tested against the live `api.minimax.io` endpoint with `MINIMAX_API_KEY`

### DIFF
--- a/model/minimax/minimax.go
+++ b/model/minimax/minimax.go
@@ -1,0 +1,572 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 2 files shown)


karolpiotrowicz:
Thanks for this — and for the thorough tests and write-up!

ADK Go is model-agnostic by design: `model.LLM` is a small, public interface precisely so that any provider can be implemented and shipped as its own module, with no changes to ADK Go itself. We're intentionally very selective about which providers we maintain in-tree, so third-party providers like this one are best published and versioned independently. Users then add both modules to their imports and pass your model into `llmagent.Config{Model: ...}`.

So rather than merge it here, we'd encourage you to release it as a standalone module (e.g. `github.com/<you>/adk-minimax`) implementing `model.LLM`. That keeps you free to version and ship it on your own schedule, and we're glad to point people toward community providers. We'll close this PR for now — but thank you again, and feel free to share the link once it's published! (No CLA needed for your own repo.)

PR #746, opened 2026-04-20
fix: sanitize user_id derived from PubSub subscription and Eventarc source

anFatum:
This ensures that the user_id is a valid identifier without path separators.

### DIFF
--- a/server/adkrest/controllers/triggers/triggers.go
+++ b/server/adkrest/controllers/triggers/triggers.go
@@ -47,6 +47,10 @@ type RetriableRunner struct {
 func (r *RetriableRunner) RunAgent(ctx context.Context, appName, userID, messageContent string) ([]*session.Event, error) {
 	userID = strings.ReplaceAll(strings.Trim(userID, "/"), "/", "--")
 
+	if userID == "" {
+		return nil, fmt.Errorf("user id cannot be empty")
+	}
+
 	// Each retry = new session
 	sessReq := &session.CreateRequest{
 		AppName: appName,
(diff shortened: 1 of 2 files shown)


server/adkrest/controllers/triggers/triggers.go
```diff
@@ -45,6 +45,8 @@ type RetriableRunner struct {
 }
 
 func (r *RetriableRunner) RunAgent(ctx context.Context, appName, userID, messageContent string) ([]*session.Event, error) {
+	userID = strings.ReplaceAll(strings.Trim(userID, "/"), "/", "--")
```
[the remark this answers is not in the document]
anFatum: We support only pub-sub and eventarc subscription names for now:
- `projects/project_name/subscriptions/sub_name`
- `//pubsub.googleapis.com/projects/project_name/topics/topic_name`

hanorik (approved):

server/adkrest/controllers/triggers/triggers.go
```diff
@@ -45,6 +45,12 @@ type RetriableRunner struct {
...
```
kdroste-google, line 48: QQ: Why do we need this user for? Why not a constant "pub-sub" or "eventarc" ?

kdroste-google, line 48: QQ2: why do we bother with "/" at all ?

kdroste-google, line 48: If we have to pass the user from Event.Source, and have to get rid of "/" (at this moment I don't understand why) we can do url.QueryEscape or something similar (although it's less readable)

if is has to be human-readable, and the only thing we want to escape is "/", then replacement with -- makes sense. Probably topics / subscriptions have some limits about allowed characters...

server/adkrest/controllers/triggers/triggers.go
```diff
@@ -45,6 +45,12 @@ type RetriableRunner struct {
...
+
+	if userID == "" {
+		return nil, fmt.Errorf("user id cannot be empty")
```
kdroste-google, line 51: in both pubsub and eventarc we replace an empty userId. Effectively it handles only a case of "////". So probably the error msg should be different

PR #747, opened 2026-04-21
feat: Add merged skill source proxy.

foxfrikses:
This proxy allows combining multiple skill sources into a single source.

### DIFF
--- a/tool/skilltoolset/skill/merged_source.go
+++ b/tool/skilltoolset/skill/merged_source.go
@@ -31,7 +31,7 @@ type mergedSource struct {
 // Source implementations. The sources are queried in the order they are
 // provided.
 func NewMergedSource(sources ...Source) Source {
-	return &mergedSource{sources: sources}
+	return &mergedSource{sources: slices.Clone(sources)}
 }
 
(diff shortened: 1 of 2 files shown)


tool/skilltoolset/skill/merged_source.go
```diff
@@ -0,0 +1,112 @@
...
+			}
+			names[fm.Name] = true
+		}
+		allFrontmatters = append(allFrontmatters, frontmatters...)
+	}
+
+	// Sort for determinism.
+	slices.SortFunc(allFrontmatters, func(a, b *Frontmatter) int {
+		return strings.Compare(a.Name, b.Name)
+	})
```
dpasiukevich: nit: I'm curious why you need to sort by name?

Without this sort, the output will still be deterministic (the part the ADK provides). The skills will be in the order of sources. So MergedSource will be just merging sources, simple.
And if needed for any reason the user can make SortedSource.

dpasiukevich (dismissed):

tool/skilltoolset/skill/merged_source.go
foxfrikses: Agree: it's deterministic already if the underlying sources have deterministic order.

dpasiukevich (approved):

PR #750, opened 2026-04-22
chore: add compile-time interface assertions for tool implementations

merkulovlad:
### Description of Change

  **Problem:**
  Currently, explicit compile-time checks exist only for the MCP tool via `RequestProcessor` and `FunctionTool`. Other tool
  types rely on runtime validation in `base_flow`, which makes interface conformance checks less explicit.

  **Solution:**
  Add compile-time `var _` interface assertions to all tool implementations that implement `toolinternal.RequestProcessor`
  and/or `toolinternal.FunctionTool`, consistent with the existing pattern in `mcpTool`.

  ### Testing Plan

  No tests needed.
  This PR only adds explicit compile-time interface checks and does not change runtime behavior.

  ### Checklist

  - [x] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) document.

### DIFF
--- a/internal/llminternal/agent_transfer.go
+++ b/internal/llminternal/agent_transfer.go
@@ -180,7 +180,10 @@ func (t *TransferToAgentTool) Run(ctx agent.ToolContext, args any) (map[string]a
 	return map[string]any{}, nil
 }
 
-var _ tool.Tool = (*TransferToAgentTool)(nil)
+var (
+	_ toolinternal.FunctionTool     = (*TransferToAgentTool)(nil)
+	_ toolinternal.RequestProcessor = (*TransferToAgentTool)(nil)
+)
 
 func transferTargets(agent, parent agent.Agent) []agent.Agent {
 	targets := slices.Clone(agent.SubAgents())
--- a/internal/llminternal/outputschema_processor.go
+++ b/internal/llminternal/outputschema_processor.go
@@ -23,6 +23,7 @@ import (
 
 	"google.golang.org/adk/agent"
 	"google.golang.org/adk/internal/llminternal/googlellm"
+	"google.golang.org/adk/internal/toolinternal"
 	"google.golang.org/adk/internal/toolinternal/toolutils"
 	"google.golang.org/adk/internal/utils"
 	"google.golang.org/adk/model"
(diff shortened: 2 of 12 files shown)


karolpiotrowicz:
Thanks @merkulovlad. CLA's signed — this just needs a rebase to clear the conflict with `main`, then we'll review.

merkulovlad:
Thanks @karolpiotrowicz! I rebased the branch onto the latest main and resolved the conflicts. It should be ready for review now.

karolpiotrowicz:
@wolo-lab this is a scope call, not a code review — the change is small and mechanical, but it sets a convention for the whole repository.

The PR adds `var _ toolinternal.FunctionTool` / `RequestProcessor` assertions to twelve files. Today that pattern lives in exactly one place, [tool/mcptoolset/tool.go#L187-L190](https://github.com/google/adk-go/blob/53a0bc0ceb2e4437a075631f33a76f039445745f/tool/mcptoolset/tool.go#L187-L190), with a second file asserting a different internal interface at [tool/agenttool/agent_tool.go#L41](https://github.com/google/adk-go/blob/53a0bc0ceb2e4437a075631f33a76f039445745f/tool/agenttool/agent_tool.go#L41). Generalizing one instance into a house style is the kind of thing that should be a decision rather than a merge, because every tool added afterwards will be reviewed against it.

There is a concrete side effect worth weighing. Ten of the twelve files gain a new dependency on `internal/toolinternal` purely to hold the assertion, including public packages that do not import it today — `tool/exampletool`, `tool/geminitool`, `tool/preloadmemorytool` and `tool/skilltoolset`. In `exampletool` that is the import at [tool/exampletool/tool.go#L25](https://github.com/google/adk-go/blob/07234ef754d0ec7282dc096b27a98661ce4e5ce8/tool/exampletool/tool.go#L25) added for the assertion at [tool/exampletool/tool.go#L153](https://github.com/google/adk-go/blob/07234ef754d0ec7282dc096b27a98661ce4e5ce8/tool/exampletool/tool.go#L153). Whether public tool packages should reach into an internal package to document conformance is the substance of the decision.

Separately, and regardless of which way you go: the branch's last commit is from 15 June 2026, 84 days ago, and it is written against the pre-v2 import paths while main is on `google.golang.org/adk/v2`. It needs a rebase from the author before it can be built or reviewed.

The call I need from you: adopt the convention repo-wide and ask for a rebase, take a narrower version covering only the packages where an assertion earns its keep, or decline it and leave conformance to the existing runtime checks.

PR #753, opened 2026-04-24
fix: check for NotFound error before auto-creating session (#751)

### LINKED ISSUE #751: AutoCreateSession should only create on NotFound, not on all Get errors
## Summary

When `runner.Config.AutoCreateSession` is `true`, the runner calls `sessionService.Get()` and falls back to `sessionService.Create()` on **any** error — not just when the session doesn't exist. This can silently create a new session on transient database errors (timeouts, connection failures), which is unsafe for multi-turn conversational agents that depend on session history.

## Current Behavior

In [`runner.go` lines 142-161](https://github.com/google/adk-go/blob/main/runner/runner.go#L142-L161):

```go
getResp, err := r.sessionService.Get(ctx, &session.GetRequest{...})
if err != nil {
    if !r.autoCreateSession {
        yield(nil, err)
        return
    }
    // Creates on ANY error, not just NotFound
    createResp, err := r.sessionService.Create(ctx, &session.CreateRequest{...})
    ...
}
```

All `Get` errors are treated the same — the runner does not distinguish between:
- **NotFound** (session never existed → safe to create)
- **Transient errors** (database timeout, connection refused → should fail fast)

## Problem

For multi-turn conversational agents, this behavior can silently drop conversation history:

1. User has an ongoing conversation with session `abc-123`
2. Database has a brief connectivity blip
3. `sessionService.Get("abc-123")` returns a transient error
4. Runner creates a **new empty session** with ID `abc-123` instead of surfacing the error
5. User's conversation history is silently lost

This forces multi-turn agents to implement their own get-or-create pattern with explicit error type checking rather than using `AutoCreateSession`.

## Expected Behavior

When `AutoCreateSession` is `true`, the runner should only call `Create` when the `Get` error indicates the session does not exist (e.g., `codes.NotFound`). Transient errors should be propagated to the caller.

## Suggested Fix

```go
getResp, err := r.sessionService.Get(ctx, &session.GetRequest{...})
if err != nil {
    if !r.autoCreateSession || !isNotFoundError(err) {
        yield(nil, err)
        return
    }
    createResp, err := r.sessionService.Create(ctx, &session.CreateRequest{...})
    ...
}
```

Where `isNotFoundError` checks for gRPC `codes.NotFound` and/or the string-based `"not found"` pattern used by the database session service (which wraps `gorm.ErrRecordNotFound` as `"database error while fetching session: record not found"`).

## Workaround

We currently implement manual get-or-create logic in our service layer for multi-turn agents, bypassing `AutoCreateSession` entirely.

## Environment

- `google.golang.org/adk` v1.1.0
- Go 1.24

indurireddy-TF:

- Closes: #751 
- Related: #751 

possible._

**Problem:**
The code treats all network or database errors as a missing session, silently overwriting existing user data when a temporary failure occurs.

**Solution:**
Add an check so the framework only auto-creates a new session when it confirms the session actually does not exist.

### Testing Plan

**Unit Tests:**

- [ ] I have added or updated unit tests for my change.

### Checklist

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -23,6 +23,8 @@ import (
 	"time"
 
 	"google.golang.org/genai"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"google.golang.org/adk/agent"
 	"google.golang.org/adk/artifact"
@@ -145,7 +147,7 @@ func (r *Runner) Run(ctx context.Context, userID, sessionID string, msg *genai.C
 			SessionID: sessionID,
 		})
 		if err != nil {
-			if !r.autoCreateSession {
+			if !r.autoCreateSession || status.Code(err) != codes.NotFound {
 				yield(nil, err)
 				return
 			}
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"iter"
 	"log"
+	"strings"
 	"time"
 
 	"google.golang.org/genai"
@@ -147,7 +148,7 @@ func (r *Runner) Run(ctx context.Context, userID, sessionID string, msg *genai.C
 			SessionID: sessionID,
 		})
 		if err != nil {
-			if !r.autoCreateSession || status.Code(err) != codes.NotFound {
+			if !r.autoCreateSession || !isNotFoundError(err) {
 				yield(nil, err)
 				return
 			}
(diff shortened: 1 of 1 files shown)


runner/runner.go
```diff
@@ -145,7 +148,7 @@ func (r *Runner) Run(ctx context.Context, userID, sessionID string, msg *genai.C
 			SessionID: sessionID,
 		})
 		if err != nil {
-			if !r.autoCreateSession {
+			if !r.autoCreateSession || !isNotFoundError(err) {
```
kdroste-google, line 151: I like the idea. 
I miss a clear signal saying "everything went OK but the session is not found"
Looking at the existing implementations:
1) database: 
https://github.com/google/adk-go/blob/a586408fe4376c7337496bcdaf3525f64eef194b/session/database/service.go#L159
here I'm not sure what to expect
2) vertexai
https://github.com/google/adk-go/blob/a586408fe4376c7337496bcdaf3525f64eef194b/session/vertexai/vertexai_client.go#L135
this seems to be a bit different: 
```go
	sessRpcResp, err := c.rpcClient.GetSession(ctx, sessRpcReq)
	if err != nil {
		return nil, fmt.Errorf("error fetching session: %w", err)
	}

	if sessRpcResp == nil {
		return nil, fmt.Errorf("session %+v not found", req.SessionID)
	}
```
here the underlying signal is nice: the rpc call was successful but returned nil. But we are covering it with err ("not found")
3) InMemory
https://github.com/google/adk-go/blob/a586408fe4376c7337496bcdaf3525f64eef194b/session/inmemory.go#L112

I see following options: 
1) for "session not found" return (*session, error)  nil, nil   meaning "everything's fine, no session found"
2) introduce common error msg with the same prefix (longer than "not found") and use it consistently across all implementations (+ document in interface)
3) create a new error type like 
```go
type SessionNotFound struct {
  sessionId string
}
//with Error()

//and in code check 
switch err := err.(type) {
case SessionNotFound:
    //handle it
}
```
WDYT?

wolo-lab:
Closing as stale.

PR #755, opened 2026-04-24
feat: add support for updating existing Agent Engine instances

hanorik:
This PR introduces the ability to update an existing Vertex AI Reasoning Engine (Agent Engine) instance using the `adkgo` deployment tool, removing the need to always deploy a new instance when changes are made.

**Key Changes**

- CLI Flags: Added a persistent flag to the agentEngine deployment command:
  - `--agent_engine_id`: Specifies the full ID of the Agent Engine instance to update if it exists (default: None, which means a new instance will be created).

- Update Workflow: Implemented `gcloudUpdateAgentEngine()` to handle the update process:
  - Packs and uploads the updated source code archive.
  - Retrieves and registers available class methods via `ListClassMethods` to ensure the instance specification matches the code.
  - Utilizes a `FieldMask` (`spec.source_code_spec, spec.class_methods`) to selectively update the instance via the `aiplatform` client.
- Deployment Logic: Updated `deployOnagentEngine` to route to the update logic when the flags are present.

### DIFF
--- a/cmd/adkgo/internal/deploy/agentengine/agentengine.go
+++ b/cmd/adkgo/internal/deploy/agentengine/agentengine.go
@@ -322,6 +322,16 @@ func (f *deployAgentEngineFlags) gcloudUpdateAgentEngine() error {
 				return fmt.Errorf("cannot read archive file: %w", err)
 			}
 
+			methods, err := agentengine.ListClassMethods()
+			if err != nil {
+				return fmt.Errorf("cannot list class methods: %w", err)
+			}
+			methodsJSON, err := json.Marshal(methods)
+			if err != nil {
+				return fmt.Errorf("cannot marshal methods: %w", err)
+			}
+			p("Methods:", string(methodsJSON))
+
 			req := &aiplatformpb.UpdateReasoningEngineRequest{
 				ReasoningEngine: &aiplatformpb.ReasoningEngine{
 					Name: f.agentEngine.instanceName,
@@ -338,9 +348,10 @@ func (f *deployAgentEngineFlags) gcloudUpdateAgentEngine() error {
 								},
 							},
 						},
+						ClassMethods: methods,
 					},
 				},
-				UpdateMask: &fieldmaskpb.FieldMask{Paths: []string{"spec.source_code_spec"}},
+				UpdateMask: &fieldmaskpb.FieldMask{Paths: []string{"spec.source_code_spec", "spec.class_methods"}},
 			}
 			p("Sending UpdateReasoningEngine request...")
 			op, err := client.UpdateReasoningEngine(ctx, req)
--- a/cmd/adkgo/internal/deploy/agentengine/agentengine.go
+++ b/cmd/adkgo/internal/deploy/agentengine/agentengine.go
@@ -45,11 +45,10 @@ type gCloudFlags struct {
 }
 
 type agentEngineServiceFlags struct {
-	name         string
-	displayName  string
-	serverPort   int
-	update       bool
-	instanceName string
+	name          string
+	displayName   string
+	serverPort    int
+	agentEngineID string
 }
 
 type buildFlags struct {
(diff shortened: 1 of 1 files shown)


cmd/adkgo/internal/deploy/agentengine/agentengine.go
```diff
@@ -288,6 +293,85 @@ func (f *deployAgentEngineFlags) gcloudDeployToAgentEngine() error {
...
+								LanguageSpec: &aiplatformpb.ReasoningEngineSpec_SourceCodeSpec_ImageSpec_{
+									ImageSpec: &aiplatformpb.ReasoningEngineSpec_SourceCodeSpec_ImageSpec{},
+								},
+							},
+						},
+						ClassMethods: methods,
+					},
+				},
+				UpdateMask: &fieldmaskpb.FieldMask{Paths: []string{"spec.source_code_spec", "spec.class_methods"}},
+			}
```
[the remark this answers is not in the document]
hanorik: This change is for code update - the functionality to update environment variables will be added later.

cmd/adkgo/internal/deploy/agentengine/agentengine.go
```diff
@@ -94,6 +97,8 @@ func init() {
 	agentEngineCmd.PersistentFlags().IntVar(&flags.agentEngine.serverPort, "server_port", 8080, "agentEngine server port")
 	agentEngineCmd.PersistentFlags().StringVarP(&flags.source.entryPointPath, "entry_point_path", "e", "", "Path to an entry point (go 'main')")
 	agentEngineCmd.PersistentFlags().StringVarP(&flags.source.sourceDir, "source_dir", "d", "", "Directory to archive, defaults to current working directory")
+	agentEngineCmd.PersistentFlags().BoolVar(&flags.agentEngine.update, "update", false, "Update an existing Agent Engine instance")
```
kdroste-google: WDYT if we copy the command line semantics from adk:
https://github.com/google/adk-python/blob/c52f9564330f0c00d82338cc58df28cb22400b6f/src/google/adk/cli/cli_tools_click.py#L1059

@click.option(
    "--agent_engine_id",
    type=str,
    default=None,
    help=(
        "Optional. ID of the Agent Engine instance to update if it exists"
        " (default: None, which means a new instance will be created)."
        " The corresponding resource name in Agent Engine will be:"
        " `projects/{project}/locations/{region}/reasoningEngines/{agent_engine_id}`."
    ),
)

kdroste-google (changes requested):

cmd/adkgo/internal/deploy/agentengine/agentengine.go
hanorik: It's a good point - modified!

cmd/adkgo/internal/deploy/agentengine/agentengine.go
```diff
@@ -94,6 +96,7 @@ func init() {
 	agentEngineCmd.PersistentFlags().IntVar(&flags.agentEngine.serverPort, "server_port", 8080, "agentEngine server port")
 	agentEngineCmd.PersistentFlags().StringVarP(&flags.source.entryPointPath, "entry_point_path", "e", "", "Path to an entry point (go 'main')")
 	agentEngineCmd.PersistentFlags().StringVarP(&flags.source.sourceDir, "source_dir", "d", "", "Directory to archive, defaults to current working directory")
+	agentEngineCmd.PersistentFlags().StringVar(&flags.agentEngine.agentEngineID, "agent_engine_id", "", "ID of the Agent Engine instance to update if it exists (default: None, which means a new instance will be created).")
```
kdroste-google: More golangish "" instead of None

cmd/adkgo/internal/deploy/agentengine/agentengine.go
```diff
@@ -288,6 +291,78 @@ func (f *deployAgentEngineFlags) gcloudDeployToAgentEngine() error {
 		})
 }
 
+// gcloudUpdateAgentEngine invokes gcloud to update source on agentEngine
+func (f *deployAgentEngineFlags) gcloudUpdateAgentEngine() error {
```
kdroste-google, line 295: WDYT about merging those two: update and create (preparing the difference in payload to aiplatform call), with the third one method (modify?) called from both?

kdroste-google (changes requested):

cmd/adkgo/internal/deploy/agentengine/agentengine.go
hanorik: I believe it makes sense but will require some valuable refactoring so I'd like to submit current implementation so we could use update functionality instead of deploying a new instance every time and to refactor it in the separate PR

kdroste-google (approved):
LGTM, thank you!

PR #763, opened 2026-04-28
fix: copy ThoughtSignature onto adk_request_confirmation parts

### LINKED ISSUE #656: Missing ThoughtSignature on adk_request_confirmation parts breaks execution

gustafvh:

- Closes: google/adk-go#656

### Problem

`generateRequestConfirmationEvent` in `internal/llminternal/functions.go` creates synthetic `adk_request_confirmation` `FunctionCall` parts but does not copy the `ThoughtSignature` from the original model-generated `FunctionCall` part onto them.

Gemini thinking models require replayed model-role function-call parts to carry their thought signature, so the next model request rejects the synthetic part with `400 INVALID_ARGUMENT`: "function call `adk_request_confirmation` ... is missing a `thought_signature`".

The stream aggregator already handles this for normal streamed function calls. The confirmation flow simply missed it.

### Solution

When building the synthetic confirmation part, look up the original part by function-call ID and copy its `ThoughtSignature` onto the new part when present.

### Testing Plan

- [x] Added `TestGenerateRequestConfirmationEventPreservesThoughtSignature` verifying the original signature is copied onto the synthetic confirmation part.
- [x] Added `TestGenerateRequestConfirmationEventNoThoughtSignature` verifying normal behavior when the original part has no signature.

```sh
```

### Alignment with adk-python

This follows the cross-ADK requirement for Gemini thinking models: any replayed model-role function-call part, including ADK-generated confirmation calls, must preserve the thought signature associated with the model output.

### DIFF
--- a/internal/llminternal/functions.go
+++ b/internal/llminternal/functions.go
@@ -43,20 +43,22 @@ func generateRequestConfirmationEvent(
 
 	parts := []*genai.Part{}
 	longRunningToolIDs := []string{}
-	functionCalls := make(map[string]*genai.FunctionCall, len(functionCallEvent.Content.Parts))
-	for _, call := range utils.FunctionCalls(functionCallEvent.Content) {
-		functionCalls[call.ID] = call
+	functionCallParts := make(map[string]*genai.Part, len(functionCallEvent.Content.Parts))
+	for _, part := range functionCallEvent.Content.Parts {
+		if part.FunctionCall != nil {
+			functionCallParts[part.FunctionCall.ID] = part
+		}
 	}
 
 	for funcID, confirmation := range functionResponseEvent.Actions.RequestedToolConfirmations {
-		originalFunctionCall, ok := functionCalls[funcID]
-		if !ok || originalFunctionCall == nil {
+		originalPart, ok := functionCallParts[funcID]
+		if !ok || originalPart.FunctionCall == nil {
 			continue
 		}
 
 		// Prepare arguments for the adk_request_confirmation call
 		args := map[string]any{
-			"originalFunctionCall": originalFunctionCall,
+			"originalFunctionCall": originalPart.FunctionCall,
 			"toolConfirmation":     confirmation,
 		}
 
@@ -66,9 +68,14 @@ func generateRequestConfirmationEvent(
 			Args: args,
 		}
 
-		parts = append(parts, &genai.Part{
+		part := &genai.Part{
 			FunctionCall: requestConfirmationFC,
-		})
+		}
+		if len(originalPart.ThoughtSignature) > 0 {
+			part.ThoughtSignature = originalPart.ThoughtSignature
+		}
+
+		parts = append(parts, part)
 		longRunningToolIDs = append(longRunningToolIDs, requestConfirmationFC.ID)
 	}
 
--- a/internal/llminternal/functions.go
+++ b/internal/llminternal/functions.go
@@ -68,14 +68,10 @@ func generateRequestConfirmationEvent(
 			Args: args,
 		}
 
-		part := &genai.Part{
-			FunctionCall: requestConfirmationFC,
-		}
-		if len(originalPart.ThoughtSignature) > 0 {
-			part.ThoughtSignature = originalPart.ThoughtSignature
-		}
-
-		parts = append(parts, part)
+		parts = append(parts, &genai.Part{
+			FunctionCall:     requestConfirmationFC,
+			ThoughtSignature: originalPart.ThoughtSignature,
+		})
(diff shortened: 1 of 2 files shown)


radhus:

internal/llminternal/functions.go
```diff
@@ -66,9 +68,14 @@ func generateRequestConfirmationEvent(
...
-		parts = append(parts, &genai.Part{
+		part := &genai.Part{
 			FunctionCall: requestConfirmationFC,
-		})
+		}
+		if len(originalPart.ThoughtSignature) > 0 {
+			part.ThoughtSignature = originalPart.ThoughtSignature
+		}
+
+		parts = append(parts, part)
```
foxfrikses: nit:

```go
parts = append(parts, &genai.Part{
    FunctionCall: requestConfirmationFC,
    ThoughtSignature: originalPart.ThoughtSignature,
})
```

internal/llminternal/functions_test.go
```diff
@@ -157,13 +157,13 @@ func TestGenerateRequestConfirmationEvent(t *testing.T) {
...
-										"toolConfirmation": toolconfirmation.ToolConfirmation{
-											Hint: "Are you sure?",
-										},
+								Args: map[string]any{
+									"originalFunctionCall": confirmingFunctionCall,
+									"toolConfirmation": toolconfirmation.ToolConfirmation{
+										Hint: "Are you sure?",
 									},
 								},
+								},
```
foxfrikses: Spaces slid here: could you fix it back please, so that it's formatted properly?

internal/llminternal/functions.go
gustafvh: Fair, changed!

internal/llminternal/functions_test.go
gustafvh: Sure thing, must have slipped through. Fixed now with the linter

foxfrikses (approved):
Thank you for the fix!

PR #765, opened 2026-04-28
fix: yield tool responses before confirmations

### LINKED ISSUE #759: Function responses may not be persisted before tool confirmation pauses execution
**Describe the bug**

In `internal/llminternal/base_flow.go`, `runOneStep` currently yields a generated `adk_request_confirmation` event before yielding the merged function-response event returned from tool execution:

```go
toolConfirmationEvent := generateRequestConfirmationEvent(ctx, modelResponseEvent, ev)
if toolConfirmationEvent != nil {
    if !yield(toolConfirmationEvent, nil) {
        return
    }
}

if !yield(ev, nil) {
    return
}
```

This ordering can lose function-response history for consumers that intentionally stop iterating when they receive a tool confirmation request so they can pause execution and ask a user to approve or reject the pending action.

In that flow, the confirmation event is yielded and persisted, but the function-response event (`ev`) may never be yielded. The session history is then incomplete when execution resumes.

**To Reproduce**

1. Create an agent with at least one regular tool and one tool that requests confirmation.
2. Run the agent through a consumer that pauses as soon as it receives an `adk_request_confirmation` event.
3. Let the confirmed tool path generate both:
   - a normal function-response event for already executed tools
   - an `adk_request_confirmation` event for the pending tool approval
4. Stop consuming the iterator after the confirmation event, as a UI or human-in-the-loop integration commonly would.
5. Inspect the session history before resuming.

**Expected behavior**

All completed tool function responses should be yielded and persisted before the confirmation request is yielded. A consumer that pauses on the confirmation boundary should still have complete function-response history for work that already happened.

**Observed behavior**

The confirmation request is yielded first. If the consumer pauses immediately, the merged function-response event is not yielded and may not be persisted to the session.

**Proposed fix**

Yield the merged function-response event before yielding the generated `adk_request_confirmation` event:

```go
toolConfirmationEvent := generateRequestConfirmationEvent(ctx, modelResponseEvent, ev)

if !yield(ev, nil) {
    return
}

if toolConfirmationEvent != nil {
    if !yield(toolConfirmationEvent, nil) {
        return
    }
}
```

This preserves completed tool results before entering the human-confirmation pause.

**Testing**

Add a runner-level regression test where the consumer stops after receiving `adk_request_confirmation`. The test should assert that the function-response event is yielded before the confirmation request so it can be persisted even when the consumer pauses at the confirmation boundary.

**Alignment with adk-python**

This is specific to ADK Go's runner iterator/yield ordering. The cross-language invariant is that completed tool results should be persisted before execution pauses for human confirmation.

gustafvh:

- Closes: google/adk-go#759

### Problem

`runOneStep` currently yields a generated `adk_request_confirmation` event before yielding the merged function-response event from tool execution.

Consumers commonly pause as soon as they receive a tool confirmation request so they can ask a user to approve or reject the pending action. If such a consumer stops iterating at that boundary, the confirmation request is yielded and persisted, but the already completed function-response event may never be persisted to session history.

### Solution

Yield the merged function-response event before yielding the generated `adk_request_confirmation` event, so completed tool results are persisted before execution pauses for human confirmation.

### Testing Plan

- [x] Added `TestToolConfirmationYieldsFunctionResponseBeforeConfirmation` verifying a consumer that stops at the confirmation event has already seen the tool function response.
- [x] Updated `TestToolConfirmation` fixture order for the new yield order.
- [x] Updated `TestMCPToolSetConfirmation` fixture order for the new yield order.

```sh
go test ./internal/llminternal ./tool/functiontool ./tool/mcptoolset
```

### Alignment with adk-python

This keeps the session-history invariant that completed tool responses are available before the framework pauses for human approval of a confirmation request.


radhus:

tool/functiontool/function_test.go
```diff
@@ -865,6 +865,60 @@ func TestToolConfirmation(t *testing.T) {
...
+		for _, part := range got.Content.Parts {
+			if part.FunctionResponse != nil && part.FunctionResponse.Name == "test_tool" {
+				sawFunctionResponse = true
+			}
+			if part.FunctionCall != nil && part.FunctionCall.Name == toolconfirmation.FunctionCallName {
+				sawConfirmation = true
+				if !sawFunctionResponse {
+					t.Fatal("confirmation was yielded before the tool function response")
+				}
+				return
```
foxfrikses, line 910: You probably meant 'break' here - not 'return'

foxfrikses: Oh, I see, by this line you've already found both 'confirmation' and 'function response' parts; thus you do want to return here.

I guess, then you don't need the `sawConfirmation` variable at all: if you encounter confirmation, then return; if you don't – `t.Error("expected confirmation event")` in the end, without checking `if !sawConfirmation`: you know you haven't seen the confirmation: if you had, the test would have returned already.

gustafvh: Good point! I've removed the `sawConfirmation` variable now per your suggestion

foxfrikses:
Some tests are failing. Could you address that?

gustafvh:
@foxfrikses For sure! Updated the event ordering for the tests to follow the new order per this PR

foxfrikses (approved):
Thank you for the fix!

PR #766, opened 2026-04-28
fix: ignore stale function responses

### LINKED ISSUE #760: Stale function responses can hard-fail content processing after retries or session restarts
**Describe the bug**

`rearrangeEventsForLatestFunctionResponse` in `internal/llminternal/contents_processor.go` returns an error when the latest event is a function response and no matching function-call event can be found in session history:

```go
if functionCallEventIdx == -1 {
    return nil, fmt.Errorf(
        "no function call event found for function responses ids: %v",
        responseIDs,
    )
}
```

That is appropriate for truly malformed history, but it can also happen when a stale tool result is replayed after a retry, reconnect, or session restart. In that case, the response is already dangling; failing the whole request prevents the next turn from continuing even though the stale response could be safely ignored.

**To Reproduce**

1. Start or resume a session with normal user/model history.
2. Submit a function-response event whose `ID` no longer matches any function-call event in the session history.
3. Let `ContentsRequestProcessor` build request contents.
4. `rearrangeEventsForLatestFunctionResponse` sees the latest event is a function response.
5. No matching function-call event is found, so content processing returns:

```text
no function call event found for function responses ids: ...
```

**Expected behavior**

If the latest function-response event is stale/orphaned and has no matching function-call event in history, ADK Go should treat it as a no-op for content construction rather than failing the whole turn.

The previous valid history should still be included in the request contents.

**Observed behavior**

The whole content processing path fails with `no function call event found...`, even though dropping the dangling latest response would allow the turn to continue.

**Proposed fix**

When no matching function-call event is found for the latest function-response event:

- log a warning with the orphaned response IDs
- drop only the trailing stale function-response event
- continue building contents from the remaining valid history

This should be limited to the "latest response has no matching call" case. Other validation errors where a response set partially matches the wrong call event should remain errors.

**Testing**

Add content processor tests for:

- a trailing orphaned function response after normal user history
- a trailing orphaned function response after a legitimate model text reply

Both tests should assert that the stale response is dropped and earlier valid contents remain.

**Alignment with adk-python**

This is specific to ADK Go's event history rearrangement for function responses. The cross-language invariant is that stale client retries should not make otherwise valid session history unusable.

gustafvh:

- Closes: google/adk-go#760

### Problem

`rearrangeEventsForLatestFunctionResponse` returns an error when the latest event is a function response and no matching function-call event can be found in session history:

```text
no function call event found for function responses ids: ...
```

That is appropriate for malformed history, but it also fires when a stale tool result is replayed after a retry, reconnect, or session restart. The dangling response could be safely ignored; failing the whole request prevents the next turn from continuing on otherwise valid history.

### Solution

When the latest function-response event has no matching call in history, drop only that trailing event and continue with the remaining valid contents. Other validation errors, such as a response set partially matching the wrong call event, remain hard errors.

### Testing Plan

- [x] Added coverage for a trailing orphaned function response after normal user history.
- [x] Added coverage for a trailing orphaned function response after a model text reply.

```sh
```

### Alignment with adk-python

This is specific to ADK Go's content-history rearrangement path. The shared behavior goal is that stale client retries should not make otherwise valid session history unusable.


radhus:

indurireddy-TF:
Hi @radhus @gustafvh Could you please sync this PR with the latest changes from the base branch so we can review it? Thanks!

radhus:
> Hi @radhus @gustafvh Could you please sync this PR with the latest changes from the base branch so we can review it? Thanks!

Just rebased 🙏

indurireddy-TF:
Hi @kdroste-google @baptmont Could you please review this PR, since all test cases are passing successfully?
image

internal/llminternal/contents_processor.go
kdroste-google, line 225: The first and only event will not be dropped, even if it would be dropped if we had more events. Is it ok?

kdroste-google (commented):
Hi, @gustafvh,
Thanks for spotting this issue and preparing the fix!
I have however some doubts about the root cause. Under what circumstances you have experienced that? Was is maybe a non-persistent session service? 

An edge case for 1 event only (see the comment) - the event will not be ignored.  Maybe doesn't make sense to handle it, what do you think?

gustafvh:
@kdroste-google Thanks for the review!

We hit this with a persistent Vertex AI session service, not an in-memory one. The stale function_response came from a client retry/reconnect where a tool result was replayed after the matching call had already been processed, leaving a trailing orphan response in session history. That then caused content construction to fail and blocked the next turn.

On your specific point, I think it’s ok because a session containing only an orphaned function response has no valid prior context to preserve, while this PR targets the production case (an orphan trailing after otherwise valid history). But I can also drop the single-event orphan if you prefer consistency here

And feel free to challenge the above if you disagree

kdroste-google (dismissed):
Hi, @gustafvh , 

Let's ignore the case with the only non-matching event - maybe just leave a comment in the code.

But I think that skipping the orphaned function response should not be silent. What do you think of logging some info in that case? Without sensitive information of course?

Recently we moved to v2, a lot of code has changed...

gustafvh:
Hi @kdroste-google, 

Sure, makes sense to add that logging, I can change that! 

Yes saw that, looks really nice! Planning to upgrade us to v2, likely this week. I would however guess that v2 did not address this issue and therefore this PR is still needed? Or what do you think?

karolpiotrowicz:
Confirmed — v2 didn't address this. On current `main` (v2.0.0), `rearrangeEventsForLatestFunctionResponse` still returns the `no function call event found…` error, and that still fails the whole turn — so this fix is still needed. Good news: it already applies to v2 directly, so there's no re-port to worry about after the upgrade.

To move it forward, the two points from the earlier review still stand: leave a code comment for the single-orphan-only-event case (rather than handling it), and make the drop non-silent by logging at info level (no sensitive data) when a stale response is dropped. Once those are in, this should be good for another review pass.

karolpiotrowicz:
Following up on my earlier note. Going through this against current `main`, the core fix does what it claims: on `main` a history of `[user text, model text, orphaned response]` fails the turn, and with this change it succeeds with the earlier history intact. What needs work is everything else the one-line drop takes with it.

**The whole event is discarded rather than the orphaned response, so the user's current message can be destroyed silently.** One event can hold a text part next to a function response, and [`runner.go:657-659`](https://github.com/google/adk-go/blob/5d90dca767d1561cf57d038e8324f2ba0359b136/runner/runner.go#L657-L659) stores the caller's `Content` verbatim as a single event. A client that sends its next question in the same message as a replayed tool result therefore loses the question, and because the last surviving content is then the model's reply, the continuation turn gets appended and the model answers the previous exchange. Before this change that request failed loudly. After it, it succeeds with the wrong prompt.

<details>
<summary>Reproduction — the user's question disappears from the prompt</summary>

Added to `internal/llminternal/contents_processor_test.go`, reusing the existing `fakeSession` and `NewContentFromFunctionResponse` helpers:

```go
func TestTrailingMixedEventLosesUserText(t *testing.T) {
	orphan := &genai.FunctionResponse{ID: "stale_x", Name: "read_file", Response: map[string]any{"r": "x"}}
	events := []*session.Event{
		{Author: "user", LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("Hello", "user")}},
		{Author: "test_agent", LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("Hi", "model")}},
		{Author: "user", LLMResponse: model.LLMResponse{Content: &genai.Content{Role: "user", Parts: []*genai.Part{
			{Text: "What is the weather?"},
			{FunctionResponse: orphan},
		}}}},
	}
	testAgent := utils.Must(llmagent.New(llmagent.Config{Name: "test_agent", Model: &testModel{}}))
	ctx := icontext.NewInvocationContext(t.Context(), icontext.InvocationContextParams{
		Agent: testAgent, Session: &fakeSession{events: events},
	})
	req := &model.LLMRequest{}
	for _, err := range llminternal.ContentsRequestProcessor(ctx, req, &llminternal.Flow{}) {
		if err != nil {
			t.Fatalf("errored: %v", err)
		}
	}
	for i, c := range req.Contents {
		for _, p := range c.Parts {
			t.Logf("contents[%d] role=%s text=%q", i, c.Role, p.Text)
		}
	}
}
```

Run with `go test -run TestTrailingMixedEventLosesUserText -v ./internal/llminternal/`. With this change applied:

```
contents[0] role=user  text="Hello"
contents[1] role=model text="Hi"
contents[2] role=user  text="Continue processing previous requests as instructed. Exit or provide a summary if no more outputs are needed."
```

On `main` the same input instead returns `no function call event found for function responses ids: map[stale_x:{}]`.

</details>

Stripping the unmatched `FunctionResponse` parts from the trailing event and keeping the event when other parts survive — dropping it only once nothing is left — handles this without changing the rest of the behavior.

**"No matching call" means "no matching call in the filtered slice", so genuine tool results are dropped too.** The backward search runs after the isolation-scope filter at [`contents_processor.go:101-102`](https://github.com/google/adk-go/blob/5d90dca767d1561cf57d038e8324f2ba0359b136/internal/llminternal/contents_processor.go#L101-L102), so a call issued inside a task scope whose result arrives unscoped looks like an orphan even though nothing is stale. History truncation produces the same shape. With this change the real tool result is discarded and the turn proceeds as though the tool never ran.

**A history consisting only of orphaned responses now sends a bare `functionResponse` to the model.** Dropping the tail leaves a single event, which short-circuits the second pass's own guard at [`contents_processor.go:363-365`](https://github.com/google/adk-go/blob/5d90dca767d1561cf57d038e8324f2ba0359b136/internal/llminternal/contents_processor.go#L363-L365), so the remaining orphan is never cleaned up and goes out with no preceding function call.

**The drop is still silent.** The logging asked for in the earlier review matters more than it first appears: given the cases above, the discarded event can be the user's actual message, and there would be nothing in the logs to explain why the model answered a different question.

All four points are worth weighing against porting adk-python's approach rather than extending this one. [`_drop_orphaned_function_responses`](https://github.com/google/adk-python/blob/4599a52659f9c3655263b3aabd39a189c692c0d2/src/google/adk/flows/llm_flows/contents.py#L219-L277) runs as a pre-pass [before both rearrangement passes](https://github.com/google/adk-python/blob/4599a52659f9c3655263b3aabd39a189c692c0d2/src/google/adk/flows/llm_flows/contents.py#L950), and is part-granular, whole-history and logging. It also lets the error at [`contents_processor.go:299-304`](https://github.com/google/adk-go/blob/5d90dca767d1561cf57d038e8324f2ba0359b136/internal/llminternal/contents_processor.go#L299-L304) stay in place for genuinely malformed history, rather than trading that diagnostic away. Two Go-specific notes if you go that way: build the call-ID set from the pre-filter events so the isolation-scope case is not misclassified, and keep Python's carve-out for responses with no ID, because Go currently treats an ID-less response as an orphan and discards its result.

Smaller things:

- **The change removes the only test case that populated `wantErr`**, so the table's error assertion is no longer exercised by anything in the package. Separately, the `validation failed: last response event has IDs not in the matching call event` branch has no test at all, and this change is meant to preserve it.
- **Both added cases are the same shape** — a single-part orphan after plain history. A one-event history, mixed text-and-response parts, consecutive trailing orphans, and a partial ID match are all unpinned.
- **The branch is 76 commits behind `main`** and needs a rebase before it can land.

karolpiotrowicz:
@gustafvh - did you have a chance to take a look?

gustafvh:
@karolpiotrowicz Apologies, missed your previous comment!

Will have a look at a fix soon according to your comments and request your review here. Thanks!

gustafvh:
@karolpiotrowicz Thanks — addressed your review:

- Rebased onto current `main`.
- Replaced the whole-event drop with a Python-parity pre-pass that removes only orphaned `FunctionResponse` parts, preserves other and ID-less parts, and uses complete pre-filter history for call IDs.
- Restored strict validation for genuine mismatches and added an `adk:` log containing only orphan IDs.
- Added coverage for mixed, single, consecutive, and mid-history orphans, ID-less responses, isolation-filtered calls, and partial mismatches.

karolpiotrowicz (changes requested):
The rebase is in and the five things round 2 asked for are all there: the drop is part-granular, the call-ID set comes from the pre-filter history, responses without an ID are carved out, the drop is logged, and the strict validation error is back with two cases pinning it rather than none. I checked the round-2 reproduction against both revisions — the history `[Before]`, `[text "my question" + stale response]`, `[After]` loses the question on `main` and keeps it with this branch applied. That was the bug.

Four things came out of going through it against `main` ([f322a15](https://github.com/google/adk-go/commit/f322a15f976f402991afd8e5d7e709aa7771ea07)), in order of who they affect.

## The new notice fires on healthy tool traffic

An ordinary tool loop with compaction configured now logs "no matching function call" once per tool call, every turn, for calls that are correctly paired. This already happens in the repo's own suite:

```
$ go test -run TestTailRetentionStopsWhenItIsNotHelping -v ./runner/
2026/09/07 22:09:34 adk: dropping function responses with no matching function call: [c1]
2026/09/07 22:09:34 adk: dropping function responses with no matching function call: [c2]
2026/09/07 22:09:34 adk: dropping function responses with no matching function call: [c3]
2026/09/07 22:09:34 adk: dropping function responses with no matching function call: [c4]
2026/09/07 22:09:34 adk: dropping function responses with no matching function call: [c5]
2026/09/07 22:09:34 adk: dropping function responses with no matching function call: [c6]
--- PASS: TestTailRetentionStopsWhenItIsNotHelping (0.00s)
```

`c1` through `c6` are the six `ping` calls that test's model emits at [`runner/compaction_test.go:1335`](https://github.com/google/adk-go/blob/f322a15f976f402991afd8e5d7e709aa7771ea07/runner/compaction_test.go#L1335), each answered by a real tool. Nothing is stale.

The route is the compaction token estimator. [`tail_retention.go:275`](https://github.com/google/adk-go/blob/f322a15f976f402991afd8e5d7e709aa7771ea07/internal/compactioninternal/tail_retention.go#L275) calls the estimator as `estimate(events[i+1:])`, a suffix window. [`compaction_processor.go:241`](https://github.com/google/adk-go/blob/f322a15f976f402991afd8e5d7e709aa7771ea07/internal/llminternal/compaction_processor.go#L241) hands that window to `buildContentsDefault`, which passes it as both arguments at [`contents_processor.go:118`](https://github.com/google/adk-go/blob/1798aff082e0b904be15eb3f38f036107ef1e6a3/internal/llminternal/contents_processor.go#L118). Every response in the window whose call sits before the window has nothing to match against.

The estimate itself does not change — for a window holding a 4000-character tool result whose call is outside it, the estimate is 0 tokens both with and without this branch, because `rearrangeEventsForFunctionResponsesInHistory` was already discarding that response. So this is the message being wrong, not a miscount. One constraint on any fix: it cannot live inside `dropOrphanedFunctionResponses`, which only ever sees what it is given. The estimator closure does hold `ctx`, and `Session()` is on the interface at [`agent/agent.go:428`](https://github.com/google/adk-go/blob/f322a15f976f402991afd8e5d7e709aa7771ea07/agent/agent.go#L428), so passing the full list from `compaction_processor.go` is available if you want the estimate to match the real prompt.

## A single part holding both text and a stale response loses the text

`genai.Part` lets one part carry a `Text` and a `FunctionResponse` together. [`isOrphan`](https://github.com/google/adk-go/blob/1798aff082e0b904be15eb3f38f036107ef1e6a3/internal/llminternal/contents_processor.go#L280) tests the part and [line 302](https://github.com/google/adk-go/blob/1798aff082e0b904be15eb3f38f036107ef1e6a3/internal/llminternal/contents_processor.go#L302) deletes the whole part, so the text goes with the response. This is the round-2 failure again, moved from event granularity down to part granularity: the request succeeds, the user's question is missing, and the model answers the previous exchange.

<details>

<summary>Reproduction — one part, both fields</summary>

```go
events := []*session.Event{
	{Author: "user", LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("Hello", "user")}},
	{Author: "testAgent", LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("Hi", "model")}},
	{Author: "user", LLMResponse: model.LLMResponse{Content: &genai.Content{Role: "user", Parts: []*genai.Part{
		{Text: "what is the weather?", FunctionResponse: &genai.FunctionResponse{ID: "stale_x", Name: "t", Response: map[string]any{"r": 1}}},
	}}}},
}
```

On `main`: `no function call event found for function responses ids: map[stale_x:{}]`. With this branch: `["Hello", "Hi", "Continue processing previous requests as instructed…"]` — the question is gone.

</details>

Worth saying that adk-python drops the whole part too, at [`_tool_call_rearranger.py:203-209`](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/flows/llm_flows/_tool_call_rearranger.py#L203-L209), so matching it is a defensible answer. It just seems better made deliberately than by inheritance, given round 2 was specifically about this failure.

## Two of the new decisions are not pinned by anything

Both are cheap and both are decisions this branch makes rather than pre-existing gaps.

- **The log survives deletion.** Removing the whole `if len(orphanedIDs) > 0` block leaves `go test ./internal/llminternal/` green. `captureLog` already exists at [`base_flow_test.go:1207`](https://github.com/google/adk-go/blob/f322a15f976f402991afd8e5d7e709aa7771ea07/internal/llminternal/base_flow_test.go#L1207), though it is in `package llminternal` and the table is in `llminternal_test`, so it needs a move rather than a call.
- **The `allEvents` argument at [line 650](https://github.com/google/adk-go/blob/1798aff082e0b904be15eb3f38f036107ef1e6a3/internal/llminternal/contents_processor.go#L650) is load-bearing in no test.** That is the one call site where the two arguments differ, and every `include_contents="none"` case in the file builds its calls without IDs, so `isOrphan` short-circuits before the distinction matters. Swapping it to `events[i:]` breaks nothing.

## Smaller things

- `%v` at [line 315](https://github.com/google/adk-go/blob/1798aff082e0b904be15eb3f38f036107ef1e6a3/internal/llminternal/contents_processor.go#L315) prints the ids raw, and the id comes from client-supplied content, so a newline in one forges a whole log line. `%q` escapes elementwise.
- `cloneEvent` keeps only `Content` out of the whole `LLMResponse` — `ModelVersion`, `FinishReason`, `TurnComplete` and the rest come back zeroed. Nothing downstream of the drop reads them today, so this changes no output, but the clone is guarded only by the `make`+`copy` on the line under `// TODO check if copy parts is needed`. Acting on that TODO makes `slices.DeleteFunc` write into the session's live parts array. The package suite stays green when I do that — `go test ./...` catches it, but as a nil-pointer panic in an unrelated compaction test rather than as a failed assertion. A line of comment there would help.
- Four of the ten new or changed table cases pass on `main` as well: the mid-history case, the no-ID case, and both `wantErr` cases. They still pin behavior against a code regression, so they are worth keeping — the mid-history one just does not demonstrate what its name suggests, since `rearrangeEventsForFunctionResponsesInHistory` already handled that shape.
- The mid-history mixed-part shape — the one from round 2 that this branch actually fixes — has no test. It sits between the trailing-mixed case and the mid-history response-only case.

One thing outside your control: the changes-requested review from July is still attached and points at a commit the rebase replaced, so it will need dismissing before this can merge.

gustafvh:
@karolpiotrowicz Thanks — addressed in ea45087:

- Compaction estimates now use full session history for call IDs; your runner reproduction passes without false stale-response notices.
- Clear only the orphaned response field, preserving text and other fields sharing its part.
- Escape logged IDs with `%q`; test the notice, newline escaping, and absence of response payloads.
- Preserve cloned response metadata, document parts-array ownership, and test session immutability, current-turn call history, and mid-history mixed content. Kept the existing regression cases.

Build, race tests, lint, formatting, and tidy passed across both modules. Could you take another look? The July review dismissal still needs an upstream maintainer.

internal/llminternal/contents_processor.go
```diff
@@ -259,6 +266,64 @@ func eventBelongsToBranch(invocationBranch string, event *session.Event) bool {
...
+				cleaned := *part
+				cleaned.FunctionResponse = nil
+				if reflect.ValueOf(cleaned).IsZero() {
+					continue
+				}
+				part = &cleaned
+			}
+			parts = append(parts, part)
+		}
+		cloned.LLMResponse.Content.Parts = parts
```
karolpiotrowicz, line 315: The text preserved here is discarded a few frames later on one path. With `[call a]`, then `[Text "note" + orphaned FunctionResponse]`, then the response matching call a, this loop keeps `note`, and then the intervening-event purge in `rearrangeEventsForLatestFunctionResponse` drops the event again — it now carries no calls and no responses, so it hits the [`continue` at contents_processor.go:425-428](https://github.com/google/adk-go/blob/ea45087add2e24c36c71af6611751b12a8ec05b7/internal/llminternal/contents_processor.go#L425-L428) and never reaches `resultEvents`. Contents come out as `[call:a] [resp:a]`.

No change in behavior: the merge-base and the previous head produce exactly the same contents for that input, so this is a limit on how far the fix reaches rather than something the fix broke. Filed as #1540 so it does not get lost — the new mixed-content table cases have no trailing matched response, so none of them enters this branch and the gap is not visible from this PR's tests. Nothing for you to do here.

internal/llminternal/compaction_processor.go
```diff
@@ -238,11 +239,17 @@ func promptTokenEstimator(ctx agent.InvocationContext) compactioninternal.TokenC
 		// history it counts turns the real prompt will not carry. That
 		// divergence predates this change — an explicit IncludeContents="none"
 		// reached it the same way — and a placement is now a second route in.
-		contents, err := buildContentsDefault(
+		// A suffix may omit a response's call without making that response stale.
+		allEvents := events
+		if ctx.Session() != nil {
+			allEvents = slices.Collect(ctx.Session().Events().All())
```
karolpiotrowicz, line 245: `Events()` is dereferenced without a nil check, where [`collect` in the same subsystem](https://github.com/google/adk-go/blob/ea45087add2e24c36c71af6611751b12a8ec05b7/internal/compactioninternal/compactor.go#L239-L243) guards it and returns nil. `session.Session` is exported, so an implementation returning a nil `Events` is expressible, and a few test doubles in the repo do it.

Not reachable in practice as things stand: [`contents_processor.go:89-90`](https://github.com/google/adk-go/blob/ea45087add2e24c36c71af6611751b12a8ec05b7/internal/llminternal/contents_processor.go#L89-L90) performs the identical guarded-session, unguarded-`Events()` pair earlier on the same request path, so such a session would already have panicked before the estimator runs. Worth matching `collect`'s shape anyway, or calling it, so the two do not drift apart.

karolpiotrowicz (approved):
Nothing blocks. Round 3's two headline items are both measured fixed, and the rest of that round's asks landed with tests behind them.

The false stale-response notices are gone. Running `go test -count=1 -run TestTailRetentionStopsWhenItIsNotHelping -v ./runner/`, the line `adk: dropping function responses with no matching function call` appears six times on the previous head ([1798aff0](https://github.com/google/adk-go/commit/1798aff082e0b904be15eb3f38f036107ef1e6a3)) and zero times with this revision ([ea45087a](https://github.com/google/adk-go/commit/ea45087add2e24c36c71af6611751b12a8ec05b7)). Passing the session as the call-ID source at [`compaction_processor.go:243-247`](https://github.com/google/adk-go/blob/ea45087add2e24c36c71af6611751b12a8ec05b7/internal/llminternal/compaction_processor.go#L243-L247) is what does it. Text sharing a part with an orphaned response now survives, and the assertions added for the log output, for the untouched-session guarantee, and for the `include_contents="none"` argument each fail on the previous head and pass here, so they discriminate rather than just execute.

One thing I chased down, because it looks at first like a regression and turns out to be the opposite. Take two calls in separate events, then a part carrying both text and an orphaned response, then a trailing event answering both calls. With this revision the request fails at [`contents_processor.go:384-394`](https://github.com/google/adk-go/blob/ea45087add2e24c36c71af6611751b12a8ec05b7/internal/llminternal/contents_processor.go#L384-L394) with `validation failed: last response event has IDs not in the matching call event`, where the previous head built contents. But that error is the behavior the suite already asserts: [`contents_processor_test.go:1541-1553`](https://github.com/google/adk-go/blob/ea45087add2e24c36c71af6611751b12a8ec05b7/internal/llminternal/contents_processor_test.go#L1541-L1553) is the same shape without the orphan, and it expects exactly that message. The merge-base ([f322a15f](https://github.com/google/adk-go/commit/f322a15f976f402991afd8e5d7e709aa7771ea07)) errors too.

The previous head only got past it by accident. The orphan was that event's only part, so pruning deleted the whole event, the calls became adjacent to the response, and the [early return at `contents_processor.go:352-362`](https://github.com/google/adk-go/blob/ea45087add2e24c36c71af6611751b12a8ec05b7/internal/llminternal/contents_processor.go#L352-L362) skipped validation altogether — and what it then built was `[call:a] [resp:a resp:b] [call:b] [resp:a resp:b]`, the response event duplicated with the calls out of order. Because this revision keeps the text remnant, the intervening event survives, the early return is correctly blocked, and the validation runs. The separate-parts case below is the control: with the payloads in two parts the remnant always survived, and every revision errors. So preserving combined parts restored an invariant the previous head was quietly breaking. Nothing to do here, but worth having in writing.

<details>

<summary>The three scenarios, run on all three revisions</summary>

```
combined part      two calls, then [Text "note" + FunctionResponse "z" (no matching call)], then one event answering both calls
      merge-base  ERROR  validation failed: last response event has IDs not in the matching call event
      prev head   OK     [call:a] [resp:a resp:b] [call:b] [resp:a resp:b]
      this rev    ERROR  validation failed: last response event has IDs not in the matching call event

matched response   one call, then [Text "note" + FunctionResponse "z"], then the matching response
      merge-base  OK     [call:a] [resp:a]
      prev head   OK     [call:a] [resp:a]
      this rev    OK     [call:a] [resp:a]

separate parts     control: the first shape with the text and the orphan in two parts
      merge-base  ERROR
      prev head   ERROR
      this rev    ERROR
```

The separate-parts case is the control: with the payloads split across two parts the text remnant survived pruning on every revision, so the intervening event was always there to block the early return, and all three error. That isolates the previous head's leniency to the one case where pruning deleted the event outright.

</details>

The middle case — a preserved remnant followed by a matched response — is a real limit on how far the part-preservation fix reaches, and I have left that one on the line it concerns. It costs nothing today, since the behavior matches the merge-base exactly.

Last thing, a question rather than a request. Preserving the remainder of a combined part diverges from adk-python, which drops the whole part — at [`_tool_call_rearranger.py:204-209`](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/flows/llm_flows/_tool_call_rearranger.py#L204-L209) an orphaned response `continue`s past `kept_parts.append(part)`, so any text beside it goes too. Losing less user data seems like the better behavior, and that middle case shows the two implementations converge again on the trailing-response path. It is worth a sentence in the PR description saying the divergence is deliberate, so the next person to compare the two does not read it as drift.

That case is now tracked as #1540, so nothing there is waiting on you.

On the July change-request: I have dismissed it. Both of its asks are in — the lone-orphan asymmetry it was about no longer exists, since the pre-pass treats a single orphaned response exactly like one in a longer history, and the drop is no longer silent. That was the last thing holding the review state.

internal/llminternal/contents_processor.go
karolpiotrowicz: Resolving this one, as the shape it describes is gone.

The change is now a part-granular pre-pass over the whole history rather than a drop of the trailing event, so a single orphaned response is treated exactly like one sitting in a longer history. The "Only orphaned function response is dropped" case in the table test pins it: a lone orphaned response event, and the expected contents are empty.

PR #768, opened 2026-04-28
feat: implement workflow routing support with tagged events and conditional graph traversal

hanorik:
Implements conditional routing and event-driven graph traversal, key changes:
- Added Route Interface: Introduced `string`, `int`, and `boolean` route matchers to conditionally traverse workflow edges based on execution events.
- Enhanced Graph Edge: Extended the `Workflow` structure with an adjacency map to evaluate outgoing paths.
- Traversal Queue: Refactored `Workflow.Run` to process node outputs iteratively, using matched routes to determine the next nodes to execute.
- Event Persistence: Added a `Route` field to `session.Event`.

### DIFF
--- a/session/database/storage_session.go
+++ b/session/database/storage_session.go
@@ -79,6 +79,7 @@ type storageEvent struct {
 	// equivalent. Unpickling would require a custom library or service.
 	Actions                []byte
 	LongRunningToolIDsJSON dynamicJSON
+	RouteJSON              dynamicJSON
 	Branch                 *string
 	Timestamp              time.Time `gorm:"precision:6"`
 
@@ -133,6 +134,14 @@ func createStorageEvent(session session.Session, event *session.Event) (*storage
 			return nil, fmt.Errorf("failed to marshal long running tool IDs: %w", err)
 		}
 		storageEv.LongRunningToolIDsJSON = toolIDsJSON
+	}
+
+	if len(event.Route) > 0 {
+		routeJSON, err := json.Marshal(event.Route)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal event route: %w", err)
+		}
+		storageEv.RouteJSON = routeJSON
 	}
 
 	// Handle optional fields by taking the address of the value.
@@ -250,6 +259,13 @@ func createEventFromStorageEvent(se *storageEvent) (*session.Event, error) {
 		}
 	}
 
+	var route []string
+	if se.RouteJSON != nil {
+		if err := json.Unmarshal([]byte(se.RouteJSON), &route); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal event route: %w", err)
+		}
+	}
+
 	// --- Handle simple pointer fields (dereference or use zero value) ---
 	// Use the helper to safely get the value or its zero-value default
 	branch := derefOrZero(se.Branch)
@@ -267,6 +283,7 @@ func createEventFromStorageEvent(se *storageEvent) (*session.Event, error) {
 		Timestamp:          se.Timestamp,
 		Actions:            actions,
 		LongRunningToolIDs: toolIDs,
+		Route:              route,
 		Branch:             branch,
 		LLMResponse: model.LLMResponse{
 			Content:           content,
--- a/session/database/storage_session.go
+++ b/session/database/storage_session.go
@@ -79,7 +79,7 @@ type storageEvent struct {
 	// equivalent. Unpickling would require a custom library or service.
 	Actions                []byte
 	LongRunningToolIDsJSON dynamicJSON
-	RouteJSON              dynamicJSON
+	RoutesJSON              dynamicJSON
 	Branch                 *string
 	Timestamp              time.Time `gorm:"precision:6"`
 
@@ -136,12 +136,12 @@ func createStorageEvent(session session.Session, event *session.Event) (*storage
 		storageEv.LongRunningToolIDsJSON = toolIDsJSON
 	}
 
-	if len(event.Route) > 0 {
-		routeJSON, err := json.Marshal(event.Route)
+	if len(event.Routes) > 0 {
+		routesJSON, err := json.Marshal(event.Routes)
 		if err != nil {
 			return nil, fmt.Errorf("failed to marshal event route: %w", err)
 		}
-		storageEv.RouteJSON = routeJSON
+		storageEv.RoutesJSON = routesJSON
 	}
 
 	// Handle optional fields by taking the address of the value.
@@ -259,9 +259,9 @@ func createEventFromStorageEvent(se *storageEvent) (*session.Event, error) {
 		}
 	}
 
-	var route []string
-	if se.RouteJSON != nil {
-		if err := json.Unmarshal([]byte(se.RouteJSON), &route); err != nil {
+	var routes []string
+	if se.RoutesJSON != nil {
+		if err := json.Unmarshal([]byte(se.RoutesJSON), &routes); err != nil {
 			return nil, fmt.Errorf("failed to unmarshal event route: %w", err)
 		}
 	}
@@ -283,7 +283,7 @@ func createEventFromStorageEvent(se *storageEvent) (*session.Event, error) {
 		Timestamp:          se.Timestamp,
 		Actions:            actions,
 		LongRunningToolIDs: toolIDs,
-		Route:              route,
+		Routes:             routes,
 		Branch:             branch,
 		LLMResponse: model.LLMResponse{
 			Content:           content,
--- a/session/session.go
+++ b/session/session.go
@@ -115,6 +115,8 @@ type Event struct {
 	// Agent client will know from this field about which function call is long running.
 	// Only valid for function call event.
 	LongRunningToolIDs []string
+	// Routing information for workflow execution
+	Route []string
 }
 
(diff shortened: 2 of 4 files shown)


session/database/storage_session.go
```diff
@@ -79,6 +79,7 @@ type storageEvent struct {
 	// equivalent. Unpickling would require a custom library or service.
 	Actions                []byte
 	LongRunningToolIDsJSON dynamicJSON
+	RouteJSON              dynamicJSON
```
wolo-lab: We can have more than one, so let's rename to `RoutesJson`

session/database/storage_session.go
```diff
@@ -250,6 +259,13 @@ func createEventFromStorageEvent(se *storageEvent) (*session.Event, error) {
 		}
 	}
 
+	var route []string
```
wolo-lab: ditto

workflow/workflow.go
```diff
@@ -183,8 +231,26 @@ func (w *Workflow) Run(ctx agent.InvocationContext) iter.Seq2[*session.Event, er
...
+
+		if len(queue) == 0 {
+			yield(nil, fmt.Errorf("no start node found"))
+			return
+		}
+
+		for len(queue) > 0 {
+			currentNode := queue[0].node
+			input = queue[0].input
+			queue = queue[1:]
```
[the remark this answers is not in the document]
hanorik: it uses a constant time for reslicing.

session/session.go
```diff
@@ -115,6 +115,8 @@ type Event struct {
 	// Agent client will know from this field about which function call is long running.
 	// Only valid for function call event.
 	LongRunningToolIDs []string
+	// Routing information for workflow execution
+	Route []string
```
wolo-lab: ditto

workflow/workflow.go
```diff
@@ -34,17 +34,42 @@ type Node interface {
...
+
+func (r BoolRoute) Matches(event *session.Event) bool {
+	str := fmt.Sprint(r)
+	for _, v := range event.Route {
+		if v == str {
+			return true
+		}
+	}
+	return false
 }
```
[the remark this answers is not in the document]
wolo-lab: +1

workflow/workflow_test.go
```diff
@@ -130,3 +131,183 @@ func TestSequentialWorkflow(t *testing.T) {
 		t.Errorf("expected last output 'HELLO done', got %v", lastOutput)
 	}
 }
+
+func TestRoutes(t *testing.T) {
```
wolo-lab: **TestRoutes should be split per route type.** The current single test
  bundles `StringRoute`, `IntRoute`, and `BoolRoute` assertions together,
  which hurts failure isolation (one failure shows up as `TestRoutes`
  failing, not the specific type). Split into `TestStringRoute`,
  `TestIntRoute`, `TestBoolRoute`.

workflow/workflow_test.go
```diff
@@ -130,3 +131,183 @@ func TestSequentialWorkflow(t *testing.T) {
...
+
+func TestWorkflowRouting(t *testing.T) {
+	type testTracker struct {
+		executed []string
+	}
+
+	type testCase struct {
+		name           string
+		startRoutes    []string
+		edges          func(START Node, nodeStart *CustomRouteNode, nodeA, nodeB *FunctionNode, nodeC *CustomRouteNode, nodeD *FunctionNode) []Edge
```
wolo-lab: drop `START` from the `edges` closure signature. You can use a package level START

workflow/workflow.go
```diff
@@ -148,29 +173,52 @@ const DEFAULT_ROUTE = "__DEFAULT__"
...
+				matched = true
+				break
+			}
 		}
+	}
+	if !matched {
+		return nil, fmt.Errorf("node %s produces route tags that do not match any valid outgoing edge", currentNode.Name())
+	}
+	return queue, nil
+}
```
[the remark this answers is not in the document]
wolo-lab: Let's add this check.

workflow/workflow.go
```diff
@@ -148,29 +173,52 @@ const DEFAULT_ROUTE = "__DEFAULT__"
...
-			if edge.From == START {
-				currentNode = edge.To
-				break
-			}
+type queueItem struct {
+	node  Node
+	input any
+}
+
+func (w *Workflow) findNextNodes(currentNode Node, input any, eventList []*session.Event) ([]queueItem, error) {
```
wolo-lab: The name should represent what's inside the class, not how it is used (queue)

workflow/workflow.go
```diff
@@ -183,8 +231,26 @@ func (w *Workflow) Run(ctx agent.InvocationContext) iter.Seq2[*session.Event, er
 			input = sb.String()
 		}
 
-		for currentNode != nil {
-			for ev, err := range currentNode.Run(ctx, input) {
+		var queue []queueItem
```
wolo-lab: I would just put here START and other nodes will be calculated inside the loop by calling `findNextNodes`.

Let's assume START is always present. We will add a check for it in the following commits.

workflow/workflow.go
```diff
@@ -148,29 +173,52 @@ const DEFAULT_ROUTE = "__DEFAULT__"
...
+		for _, event := range eventList {
+			if edge.Route.Matches(event) {
+				queue = append(queue, queueItem{node: edge.To, input: input})
+				matched = true
+				break
+			}
 		}
+	}
+	if !matched {
+		return nil, fmt.Errorf("node %s produces route tags that do not match any valid outgoing edge", currentNode.Name())
```
wolo-lab: 1) Let's add a tag name to to the error message.
2) The problem might be with edge, not only with node. Let's change to something like: "No outgoing edge matches the events with a route X emitted by node Y"
3) we should distinguish the situation in which the node is terminal and it's fine no to have next nodes.

workflow/workflow.go
```diff
@@ -183,8 +231,26 @@ func (w *Workflow) Run(ctx agent.InvocationContext) iter.Seq2[*session.Event, er
...
+		}
+
+		if len(queue) == 0 {
+			yield(nil, fmt.Errorf("no start node found"))
+			return
+		}
+
+		for len(queue) > 0 {
+			currentNode := queue[0].node
+			input = queue[0].input
```
wolo-lab, line 233: there are two issues:

|  | Python | Go (PR \#768) |
| :---- | :---- | :---- |
| 3 events with outputs `a`,`b`,`c` | passes `["a","b","c"]` | passes `"c"` ❌ |
| 0 events with output | passes `None` | passes the **predecessor's own input** ❌  |

1) could be resolved in a similar way like in python: src/google/adk/workflow/utils/_node_output_utils.py:79-115
2) pass nil

wolo-lab (changes requested):

workflow/workflow.go
wolo-lab: Actually, addressing the first issue is quite complex. Let's create a bug for it and solve it in a separate cl.

workflow/workflow.go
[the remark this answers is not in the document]

workflow/workflow.go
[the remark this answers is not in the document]

session/database/storage_session.go
hanorik: Done.

session/database/storage_session.go
hanorik: Done.

session/session.go
hanorik: Done.

workflow/workflow_test.go

workflow/workflow_test.go

workflow/workflow.go
hanorik: Changed it.

workflow/workflow.go
hanorik: START node doesn't produce any events so for now we will leave it as it is.

hanorik: Actually, forget about the previous comment - realized the way to do it after I've sent it :)

workflow/workflow.go

workflow/workflow.go
hanorik: Fixed the second issue and added a check for a multiple events with route tags as only one event can specify routes actually.

workflow/workflow.go
```diff
@@ -148,29 +165,53 @@ const DEFAULT_ROUTE = "__DEFAULT__"
...
+	input any
+}
+
+func (w *Workflow) findNextNodes(currentNode Node, input any, event *session.Event) ([]nodeInput, error) {
+	if len(w.edges[currentNode]) == 0 {
+		return nil, nil
+	}
+	matched := false
+	queue := []nodeInput{}
+	added := make(map[Node]struct{})
```
anFatum, line 191: Do we update somewhere the `added` map? Or is it intended to be updated in the following PRs?

hanorik: Oh, my bad - fixed.

anFatum: nit: it would be nice to have a test that catches that in the future, but I'm okay leaving that as it is for now

anFatum (approved):
LGTM

wolo-lab (approved):
LGTM

workflow/workflow.go
hanorik: good point, will add it with next PRs

PR #781, opened 2026-04-30
feat: add MultiRoute generic type to support matching multiple route values

hanorik:
This PR introduces a new `MultiRoute` generic type in the workflow package to support matching multiple route values. This allows an edge in the workflow to be followed if the event matches any of the routes specified in the MultiRoute list.

Key Changes:

- Added `MultiRoute` generic type and implemented the `Matches` method to check if any value in the list matches the event's routes.
- Added unit tests for `MultiRoute` with both string and int types, test cases in `TestWorkflowRouting` to verify routing behavior with `MultiRoute` and a test case to verify behavior when duplicate edges exist between the same nodes.

### DIFF
--- a/workflow/workflow_test.go
+++ b/workflow/workflow_test.go
@@ -64,7 +64,7 @@ func TestFunctionNode(t *testing.T) {
 			t.Fatalf("unexpected error: %v", err)
 		}
 		count++
-		
+
 		output, ok := ev.Actions.StateDelta["output"]
 		if !ok {
 			t.Errorf("expected output in state delta")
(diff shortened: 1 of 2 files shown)


workflow/workflow_test.go
```diff
@@ -285,6 +315,49 @@ func TestWorkflowRouting(t *testing.T) {
...
+					{From: x, To: a, Route: MultiRoute[string]{"branchX", "branchA"}},
+					{From: x, To: b},
+					{From: x, To: c},
+					{From: c, To: d},
+				}
+			},
+			expectedExec: []string{"A", "B", "C", "D"},
+		},
+		{
+			name:        "incorrect MultiRoute",
```
wolo-lab: Nit: Let's be more descriptive and explain what "inccorrect" means.

workflow/workflow_test.go
```diff
@@ -285,6 +315,49 @@ func TestWorkflowRouting(t *testing.T) {
...
+					{From: x, To: a, Route: MultiRoute[string]{"branchX", "branchY"}},
+					{From: x, To: b, Route: MultiRoute[string]{"branchZ"}},
+					{From: x, To: c, Route: StringRoute("branchC")},
+					{From: c, To: d},
+				}
+			},
+			expectErrorMsg: "no outgoing edge matches the event with routes",
+		},
+		{
+			name:        "duplicate edges to same node",
```
wolo-lab, line 345: nit: let's document (method level docs) this behavior in `findNextNodes`

workflow/workflow_test.go

workflow/workflow_test.go
hanorik: Done.

wolo-lab (commented):
LGTM

anFatum (approved):

PR #783, opened 2026-04-30
fix: skip metadata-only SSE chunks instead of aborting stream in StreamingResponseAggregator

### LINKED ISSUE #782: Streaming SSE aggregator aborts on metadata-only chunks
## 🔴 Required Information

**Describe the Bug:**

`StreamingResponseAggregator` aborts the stream with `"empty response"` when Vertex AI emits a leading SSE chunk that contains only response metadata (no `Candidates`). This is consistently triggered by `gemini-3-flash-preview` + `googleSearch` grounding, causing roughly 40-50% of streaming calls to fail even though the model produces a complete answer in the subsequent chunks.

The same setup using `StreamingModeNone` succeeds 100% of the time, and `adk-python`'s aggregator (gated by the `PROGRESSIVE_SSE_STREAMING` feature, on by default) handles these chunks gracefully.

**Steps to Reproduce:**

1. `go mod init demo` and add `google.golang.org/adk@v1.2.0`, `google.golang.org/genai@v1.55.0`.
2. Authenticate to Vertex AI (`gcloud auth application-default login` or `GOOGLE_APPLICATION_CREDENTIALS`) for a project that has access to `gemini-3-flash-preview`.
3. Build a minimal agent with `gemini-3-flash-preview` and `geminitool.GoogleSearch{}`, using an in-memory session service and a fresh session per trial.
4. Call `runner.Run(...)` 10 times with `agent.RunConfig{StreamingMode: agent.StreamingModeSSE}` and the prompt `"What are some recent movie recommendations?"`.
5. Observe that ~40-50% of the runs return `error: empty response` from the iterator.

Full minimal reproduction code is attached at the bottom.

**Expected Behavior:**
All 10 streaming runs should succeed and produce a non-empty answer, matching the behavior of `StreamingModeNone` and `adk-python` against the same model and tool combination.

**Observed Behavior:**
```
[ 1] ERROR: empty response
[ 2] OK
[ 3] ERROR: empty response
[ 4] OK
[ 5] ERROR: empty response
[ 6] ERROR: empty response
[ 7] OK
[ 8] ERROR: empty response
[ 9] OK
[10] OK
=== success=5, empty/error=5, total=10 ===
```

The same program with `StreamingModeNone`: 10/10 success.
Equivalent `adk-python` setup with `StreamingMode.SSE`: 10/10 success.

**Environment Details:**

 - ADK Library Version: v1.2.0 (also reproduces on v1.1.0)
 - OS: macOS (Darwin 24.6.0, arm64)
 - Go Version: 1.26.2

**Model Information:**
 - Model: `gemini-3-flash-preview` on Vertex AI, `location=global`
 - Tool: `geminitool.GoogleSearch{}`

---

## 🟡 Optional Information

**Regression:**
N/A. The aggregator logic in `internal/llminternal/stream_aggregator.go` is identical between v1.1.0 and v1.2.0, so this is not a regression introduced by a specific ADK release.

**Logs:**

The first SSE chunk Vertex AI emits for the failing calls looks like this (captured via a debug log added to `genai`'s SSE reader):

```json
{
  "createTime": "2026-04-30T...",
  "modelVersion": "gemini-3-flash-preview",
  "responseId": "...",
  "usageMetadata": { "trafficType": "ON_DEMAND" }
}
```

No `candidates`, no `promptFeedback`. The actual content chunks arrive in subsequent SSE events, but `ProcessResponse` has already aborted the iterator by then.

**Additional Context:**

Root cause is in `internal/llminternal/stream_aggregator.go` (`ProcessResponse`):

```go
if len(genResp.Candidates) == 0 {
    // shouldn't happen?
    yield(nil, fmt.Errorf("empty response"))
    return
}
```

This treats metadata-only chunks as fatal, aborting the iterator before the real content chunk arrives.

`adk-python` handles the same situation in [`utils/streaming_utils.py`](https://github.com/google/adk-python/blob/main/src/google/adk/utils/streaming_utils.py) under the `PROGRESSIVE_SSE_STREAMING` feature (enabled by default): the aggregator buffers per-part and finalizes on `finish_reason`, silently ignoring metadata-only chunks.

**Proposed fix** (one-line behavior change in `ProcessResponse`):

```diff
 if len(genResp.Candidates) == 0 {
-    // shouldn't happen?
-    yield(nil, fmt.Errorf("empty response"))
+    // metadata-only chunk; skip and wait for the next one.
     return
 }
```

With this patch applied locally, the same reproduction produces **20/20 success** on streaming. Happy to send a PR with a unit test if the direction is acceptable, or to gate the change behind a feature flag if the maintainers prefer a more conservative rollout.

```go
package main

import (
	"context"
	"fmt"
	"strings"

	"google.golang.org/adk/agent"
	"google.golang.org/adk/model/gemini"
	"google.golang.org/adk/tool"
	"google.golang.org/adk/tool/geminitool"
	"google.golang.org/genai"
)

func main() {
	ctx := context.Background()

	model, err := gemini.NewModel(ctx, "gemini-3-flash-preview", &genai.ClientConfig{
		Backend:  genai.BackendVertexAI,
		Project:  "<your-project>",
		Location: "global",
	})
	if err != nil {
		panic(err)
	}

		Name:        "main_agent",
		Description: "test",
		Model:       model,
		Tools:       []tool.Tool{geminitool.GoogleSearch{}},
	})
	if err != nil {
		panic(err)
	}

	sessSvc := session.InMemoryService()
		AppName: "demo", Agent: a, SessionService: sessSvc,
	})
	if err != nil {
		panic(err)
	}

	const userID = "u"
	for i := 1; i <= 10; i++ {
		s, _ := sessSvc.Create(ctx, &session.CreateRequest{AppName: "demo", UserID: userID})
		msg := &genai.Content{Role: genai.RoleUser, Parts: []*genai.Part{{Text: "What are some recent movie recommendations?"}}}

		var got strings.Builder
		var gotErr error
		for ev, err := range r.Run(ctx, userID, s.Session.ID(), msg, agent.RunConfig{StreamingMode: agent.StreamingModeSSE}) {
			if err != nil {
				gotErr = err
				break
			}
			if ev != nil && ev.Content != nil {
				for _, p := range ev.Content.Parts {
					got.WriteString(p.Text)
				}
			}
		}

		switch {
		case gotErr != nil:
			fmt.Printf("[%2d] ERROR: %v\n", i, gotErr)
		case got.Len() == 0:
			fmt.Printf("[%2d] EMPTY\n", i)
		default:
			fmt.Printf("[%2d] OK len=%d\n", i, got.Len())
		}
	}
}
```

 - **Often (50%+)** — deterministically reproducible on `gemini-3-flash-preview` + `googleSearch` over Vertex AI; observed empty-rate is ~40-60% across batches of 10-20 runs.

anishesg:

Closes: #782

Vertex AI occasionally emits a leading SSE chunk that contains only response metadata (`createTime`, `modelVersion`, `usageMetadata`, etc.) with no `Candidates` field. This happens consistently with `gemini-3-flash-preview` + `googleSearch` grounding and caused roughly 40–50% of streaming calls to fail with `"empty response"`, even though the actual content arrived in subsequent chunks.

The root cause is in `internal/llminternal/stream_aggregator.go` in `ProcessResponse`: when `len(genResp.Candidates) == 0`, the function called `yield(nil, fmt.Errorf("empty response"))` and returned, aborting the iterator. The fix changes this to a silent `return`, allowing the aggregator to skip metadata-only chunks and continue processing subsequent chunks that carry real content. This matches the behavior of `adk-python`'s `PROGRESSIVE_SSE_STREAMING` path which handles identical chunks gracefully.

The unused `fmt` import was also removed as a consequence.

### Unit Tests

- [x] Added `TestMetadataOnlyChunkDoesNotAbortStream` in `stream_aggregator_test.go` that feeds a metadata-only chunk (no `Candidates`) followed by a real content chunk and asserts the aggregated response contains the expected text without any error.
- [x] All existing stream aggregator unit tests continue to pass.

```
go test ./internal/llminternal/... -run "TestProgressiveSSE|TestStreaming|TestFinishReason|TestMetadata" -v
--- PASS: TestProgressiveSSEStreamingFunctionCallArguments (0.00s)
--- PASS: TestProgressiveSSEPreservesThoughtSignature (0.00s)
--- PASS: TestProgressiveSSEHandlesEmptyFunctionCall (0.00s)
--- PASS: TestStreamingFCChunkWithWillContinueButNoPartialArgs (0.00s)
--- PASS: TestProgressiveSSEStreamingFunctionCalls (0.00s)
--- PASS: TestProgressiveSSEPreservesPartOrdering (0.00s)
--- PASS: TestMetadataOnlyChunkDoesNotAbortStream (0.00s)
--- PASS: TestFinishReasonUnexpectedToolCallPreservesErrorCode (0.00s)
PASS
```

### Manual E2E Tests

The reproduction script from the issue (10 runs of `gemini-3-flash-preview` + `googleSearch` with `StreamingModeSSE`) now produces 10/10 successes with this patch applied, matching the behavior of `StreamingModeNone` and `adk-python`.

## Checklist

- [ ] I have read the [CONTRIBUTING.md](https://github.com/google/adk-go/blob/main/CONTRIBUTING.md) file.
- [ ] I have performed a self-review of my own code.
- [ ] I have added unit tests that prove my fix is effective or that my feature works.
- [ ] I have completed manual E2E testing.
- [ ] Any dependent changes have been merged and published.

Fixes #782

### DIFF
--- a/internal/llminternal/stream_aggregator.go
+++ b/internal/llminternal/stream_aggregator.go
@@ -16,7 +16,6 @@ package llminternal
 
 import (
 	"context"
-	"fmt"
 	"iter"
 	"maps"
 	"reflect"
@@ -58,8 +57,10 @@ func NewStreamingResponseAggregator() *streamingResponseAggregator {
 func (s *streamingResponseAggregator) ProcessResponse(ctx context.Context, genResp *genai.GenerateContentResponse) iter.Seq2[*model.LLMResponse, error] {
 	return func(yield func(*model.LLMResponse, error) bool) {
 		if len(genResp.Candidates) == 0 {
-			// shouldn't happen?
-			yield(nil, fmt.Errorf("empty response"))
+			// Vertex AI occasionally emits leading SSE chunks that contain only
+			// response metadata (createTime, modelVersion, usageMetadata, etc.)
+			// but no Candidates. These are valid and should be skipped; the actual
+			// content arrives in subsequent chunks.
 			return
 		}
 		candidate := genResp.Candidates[0]
--- a/internal/llminternal/stream_aggregator_test.go
+++ b/internal/llminternal/stream_aggregator_test.go
@@ -847,6 +847,53 @@ func TestPartialFunctionCallsNotExecutedInNoneStreamingMode(t *testing.T) {
 	}
 }
 
+func TestMetadataOnlyChunkDoesNotAbortStream(t *testing.T) {
+	aggregator := llminternal.NewStreamingResponseAggregator()
+	ctx := t.Context()
+
+	// Simulate the leading metadata-only SSE chunk that Vertex AI emits for
+	// gemini-3-flash-preview + googleSearch grounding: no Candidates at all.
+	metadataChunk := &genai.GenerateContentResponse{
+		// Candidates is intentionally nil / empty.
+		UsageMetadata: &genai.GenerateContentResponseUsageMetadata{},
+	}
+
+	// The real content arrives in the next chunk.
+	contentChunk := &genai.GenerateContentResponse{
+		Candidates: []*genai.Candidate{
+			{
+				Content: &genai.Content{
+					Role:  "model",
+					Parts: []*genai.Part{{Text: "Here are some movie recommendations."}},
+				},
+				FinishReason: genai.FinishReasonStop,
+			},
+		},
+	}
+
+	for _, chunk := range []*genai.GenerateContentResponse{metadataChunk, contentChunk} {
+		for resp, err := range aggregator.ProcessResponse(ctx, chunk) {
+			if err != nil {
+				t.Fatalf("unexpected error processing chunk: %v", err)
+			}
+			_ = resp
+		}
+	}
+
+	finalResponse := aggregator.Close()
+	if finalResponse == nil {
+		t.Fatal("expected a final aggregated response, got nil")
+	}
+
+	if len(finalResponse.Content.Parts) == 0 {
+		t.Fatal("expected content parts in the final response, got none")
+	}
+
+	if finalResponse.Content.Parts[0].Text != "Here are some movie recommendations." {
+		t.Errorf("unexpected text in final response: %q", finalResponse.Content.Parts[0].Text)
+	}
+}
+
 func TestFinishReasonUnexpectedToolCallPreservesErrorCode(t *testing.T) {
 	aggregator := llminternal.NewStreamingResponseAggregator()
 	ctx := t.Context()


internal/llminternal/stream_aggregator.go
```diff
@@ -58,8 +57,10 @@ func NewStreamingResponseAggregator() *streamingResponseAggregator {
 func (s *streamingResponseAggregator) ProcessResponse(ctx context.Context, genResp *genai.GenerateContentResponse) iter.Seq2[*model.LLMResponse, error] {
 	return func(yield func(*model.LLMResponse, error) bool) {
 		if len(genResp.Candidates) == 0 {
-			// shouldn't happen?
-			yield(nil, fmt.Errorf("empty response"))
+			// Vertex AI occasionally emits leading SSE chunks that contain only
```
kdroste-google, line 60: Please add the information that sometimes SSE stream can start with an empty text parts mixed with chunks which do not contain any candidates.

internal/llminternal/stream_aggregator_test.go
```diff
@@ -847,6 +847,53 @@ func TestPartialFunctionCallsNotExecutedInNoneStreamingMode(t *testing.T) {
 	}
 }
 
+func TestMetadataOnlyChunkDoesNotAbortStream(t *testing.T) {
```
kdroste-google, line 850: Please add a test for the following sequence
Empty text
No candidates
No candidates
Empty text
No candidates
Non-empty text

kdroste-google (changes requested):
@anishesg, thank you for your contribution!
The fix seems OK.
Please see other comments for minor remarks

Thank you!

kdroste-google:
I'm closing this PR, the #918 is a replacement

kdroste-google:
No longer applicable

kdroste-google:
@anishesg , thank you!

wolo-lab:
Thanks for jumping on this so quickly, @anishesg — the diagnosis here was
spot on, and the unit test made the behavior easy to reason about.

We ended up merging #918, which builds directly on your fix. The one change:
instead of dropping the metadata-only chunks, it passes them downstream, since
they carry `usageMetadata` we want to keep. 

The fix shipped in **v1.4.0**. Thanks again for the contribution.

PR #785, opened 2026-05-01
fix(llminternal): re-evaluate toolsets on every step in toolProcessor

### LINKED ISSUE #757: Toolset.Tools(ctx) evaluated once per Runner.Run(), not per model step — state-driven toolsets broken
## Bug

`Toolset.Tools(ctx)` is called only once per `Runner.Run()`, not before each model call. The result is cached in `Flow.Tools` and reused for all subsequent `runOneStep()` iterations. Tools activated by session state changes mid-run are never visible within that run.

### Expected behavior

`Toolset.Tools(ctx)` is re-evaluated before each model call within `Flow.Run()`, so tools can change dynamically based on session state modified by earlier tool calls in the same run.

### Actual behavior

`Toolset.Tools(ctx)` is evaluated once during the first `runOneStep()`. All subsequent steps in the same `Runner.Run()` reuse the cached tool list. A tool that writes to session state (e.g. activating a capability) cannot make new tools visible until the next `Runner.Run()`.

### Root cause

`toolProcessor` in `internal/llminternal/tools_processor.go` guards with `if f.Tools != nil { return }` and caches the result in `f.Tools`. Since `llmAgent.run()` creates one `Flow` per `Runner.Run()` (`agent/llmagent/llmagent.go:374`), the cache persists across all `runOneStep()` iterations:

```go
func toolProcessor(ctx agent.InvocationContext, req *model.LLMRequest, f *Flow) iter.Seq2[*session.Event, error] {
        if f.Tools != nil {
            return // cached — never re-evaluated
        }
        // ...
        for _, toolSet := range Reveal(llmAgent).Toolsets {
            tsTools, err := toolSet.Tools(icontext.NewReadonlyContext(ctx))
            tools = append(tools, tsTools...)
        }
        f.Tools = tools // set once, reused for all subsequent steps
    }
}
```

### Suggested fix

Reset `f.Tools = nil` at the start of each `runOneStep()`, or remove the `if f.Tools != nil { return }` guard so `toolProcessor` re-evaluates toolsets every step.

### Impact

Any `Toolset` whose `Tools(ctx)` output depends on session state is broken within a single `Runner.Run()`. The `ctx` parameter on `Tools(ctx)` implies dynamic per-call evaluation, but the caching makes it effectively static.

A concrete example: an agent that starts with a small bootstrap toolset and uses a `load_skill` tool to activate domain-specific tools based on user intent. The tool writes to session state, and a conditional `Toolset.Tools(ctx)` checks that state. With the current caching, the activated tools are invisible until the next `Runner.Run()`.

### Reproduction

```go
type activateTool struct{}

func (t activateTool) Name() string        { return "activate_vehicle_tools" }
func (t activateTool) Description() string { return "Activate vehicle tools." }
func (t activateTool) IsLongRunning() bool { return false }

func (t activateTool) Run(ctx tool.Context, args any) (map[string]any, error) {
    err := ctx.State().Set("active_skills", map[string]any{"vehicles": true})
    if err != nil {
        return map[string]any{"error": err.Error()}, nil
    }
    return map[string]any{"ok": true}, nil
}

type conditionalToolset struct {
    vehicleTool tool.Tool
}

func (ts conditionalToolset) Name() string { return "vehicles" }

func (ts conditionalToolset) Tools(ctx agent.ReadonlyContext) ([]tool.Tool, error) {
    v, err := ctx.ReadonlyState().Get("active_skills")
    if err != nil {
        return nil, nil
    }
    skills, ok := v.(map[string]any)
    if !ok || skills["vehicles"] != true {
        return nil, nil
    }
    return []tool.Tool{ts.vehicleTool}, nil
}
```

1. Create an agent with `activateTool` in `Tools` and `conditionalToolset` in `Toolsets`.
2. The model calls `activate_vehicle_tools` → state is written.
3. On the next `runOneStep()` within the same `Runner.Run()`, `conditionalToolset.Tools(ctx)` is never called — `f.Tools` is cached.
4. `search_vehicles` remains unavailable for the rest of the run.

### Environment

- **ADK version:** `google.golang.org/adk v1.2.0`
- **Go version:** go1.26.2 linux/amd64
- **Frequency:** Always (100%)

nuthalapativarun:

- Closes: #757

### Description

**Problem:**
`toolProcessor` in `internal/llminternal/tools_processor.go` cached `f.Tools` after the first call and short-circuited with `if f.Tools != nil { return }` on every subsequent step. Since a `Flow` is created once per `Runner.Run()` and reused across all `runOneStep()` iterations, `Toolset.Tools(ctx)` was only ever called on the first step of a run.

Any toolset whose `Tools(ctx)` output depends on session state modified by an earlier tool call could not surface new tools within the same `Runner.Run()`. The `ctx` parameter on `Tools(ctx)` implies dynamic per-step evaluation, but the caching made it effectively static for the duration of the run.

**Root cause:**
```go
// tools_processor.go — before
func toolProcessor(...) iter.Seq2[*session.Event, error] {
        if f.Tools != nil {   // ← cached after first call; never re-evaluated
            return
        }
        ...
        f.Tools = tools
    }
}
```

**Solution:**
Remove the `if f.Tools != nil { return }` guard so `Toolset.Tools()` is called before every model step. Static `Tools` from the agent config and dynamic toolset tools are rebuilt into `f.Tools` on each call, which is correct because `runOneStep` creates a fresh `*model.LLMRequest` each iteration anyway.

```go
// tools_processor.go — after
func toolProcessor(...) iter.Seq2[*session.Event, error] {
        // No cache guard — toolsets re-evaluated every step.
        ...
        f.Tools = tools
    }
}
```

### Testing Plan

**Unit Tests:**

- [x] Added `TestToolProcessorReEvaluatesToolsetsEachStep` in `internal/llminternal/base_flow_test.go` with a `dynamicToolset` that returns no tools on the first call and one tool on the second call — verifying that `toolProcessor` picks up the change on the next step.
- [x] All existing unit tests pass locally.

```
$ go test ./internal/llminternal/... -run TestToolProcessorReEvaluatesToolsetsEachStep -v
=== RUN   TestToolProcessorReEvaluatesToolsetsEachStep
--- PASS: TestToolProcessorReEvaluatesToolsetsEachStep (0.00s)
PASS
ok  	google.golang.org/adk/internal/llminternal	0.392s

$ go test ./internal/llminternal/...
ok  	google.golang.org/adk/internal/llminternal	0.909s
ok  	google.golang.org/adk/internal/llminternal/googlellm	0.196s
```

The bug is deterministic: any agent with a `Toolset` that uses session state to control which tools are returned will reproduce it. The `activate_vehicle_tools` example from issue #757 exercises the exact path — `activate_vehicle_tools` sets state, and `conditionalToolset.Tools(ctx)` checks that state to surface a vehicle tool. After this fix, the activated tools appear on the next `runOneStep()` within the same `Runner.Run()`.

### Checklist

### DIFF
--- a/internal/llminternal/base_flow_test.go
+++ b/internal/llminternal/base_flow_test.go
@@ -683,3 +683,65 @@ func TestPreprocess_Toolset(t *testing.T) {
 		})
 	}
 }
+
+// dynamicToolset returns different tools on each call to simulate a toolset
+// whose output depends on session state written by an earlier tool in the run.
+type dynamicToolset struct {
+	callCount int
+	toolsByCall map[int][]tool.Tool
+}
+
+func (d *dynamicToolset) Name() string { return "dynamic" }
+func (d *dynamicToolset) Tools(_ agent.ReadonlyContext) ([]tool.Tool, error) {
+	tools := d.toolsByCall[d.callCount]
+	d.callCount++
+	return tools, nil
+}
+
+// TestToolProcessorReEvaluatesToolsetsEachStep verifies that toolProcessor
+// calls Toolset.Tools() on every step, not just the first. This ensures that
+// tools activated by session-state changes in an earlier step are visible to
+// subsequent model calls within the same Runner.Run().
+func TestToolProcessorReEvaluatesToolsetsEachStep(t *testing.T) {
+	extraTool := &mockFunctionTool{name: "extra_tool"}
+
+	// Call 0: no tools yet. Call 1+: return extraTool.
+	ts := &dynamicToolset{
+		toolsByCall: map[int][]tool.Tool{
+			0: nil,
+			1: {extraTool},
+		},
+	}
+
+	agentState := &State{Toolsets: []tool.Toolset{ts}}
+	mockAgent := &mockLLMAgent{s: agentState}
+	ctx := icontext.NewInvocationContext(t.Context(), icontext.InvocationContextParams{Agent: mockAgent})
+
+	f := &Flow{}
+
+	// First call: toolset returns nil — f.Tools should be empty.
+	req1 := &model.LLMRequest{}
+	for _, err := range toolProcessor(ctx, req1, f) {
+		if err != nil {
+			t.Fatalf("toolProcessor call 1 error: %v", err)
+		}
+	}
+	if len(f.Tools) != 0 {
+		t.Errorf("after call 1: got %d tools, want 0", len(f.Tools))
+	}
+
+	// Second call: toolset now returns extraTool — f.Tools must be updated.
+	req2 := &model.LLMRequest{}
+	for _, err := range toolProcessor(ctx, req2, f) {
+		if err != nil {
+			t.Fatalf("toolProcessor call 2 error: %v", err)
+		}
+	}
+	if len(f.Tools) != 1 || f.Tools[0].Name() != extraTool.Name() {
+		t.Errorf("after call 2: got tools %v, want [%s]", f.Tools, extraTool.Name())
+	}
+
+	if ts.callCount != 2 {
+		t.Errorf("toolset.Tools() called %d times, want 2", ts.callCount)
+	}
+}
--- a/internal/llminternal/base_flow_test.go
+++ b/internal/llminternal/base_flow_test.go
@@ -771,3 +771,65 @@ func TestIsThoughtOnlyTurn(t *testing.T) {
 		})
 	}
 }
+
+// dynamicToolset returns different tools on each call to simulate a toolset
+// whose output depends on session state written by an earlier tool in the run.
+type dynamicToolset struct {
+	callCount   int
+	toolsByCall map[int][]tool.Tool
+}
+
+func (d *dynamicToolset) Name() string { return "dynamic" }
+func (d *dynamicToolset) Tools(_ agent.ReadonlyContext) ([]tool.Tool, error) {
+	tools := d.toolsByCall[d.callCount]
+	d.callCount++
+	return tools, nil
+}
+
+// TestToolProcessorReEvaluatesToolsetsEachStep verifies that toolProcessor
+// calls Toolset.Tools() on every step, not just the first. This ensures that
+// tools activated by session-state changes in an earlier step are visible to
+// subsequent model calls within the same Runner.Run().
+func TestToolProcessorReEvaluatesToolsetsEachStep(t *testing.T) {
+	extraTool := &mockFunctionTool{name: "extra_tool"}
+
+	// Call 0: no tools yet. Call 1+: return extraTool.
+	ts := &dynamicToolset{
+		toolsByCall: map[int][]tool.Tool{
+			0: nil,
+			1: {extraTool},
+		},
+	}
+
+	agentState := &State{Toolsets: []tool.Toolset{ts}}
+	mockAgent := &mockLLMAgent{s: agentState}
+	ctx := icontext.NewInvocationContext(t.Context(), icontext.InvocationContextParams{Agent: mockAgent})
+
+	f := &Flow{}
+
+	// First call: toolset returns nil — f.Tools should be empty.
+	req1 := &model.LLMRequest{}
+	for _, err := range toolProcessor(ctx, req1, f) {
+		if err != nil {
+			t.Fatalf("toolProcessor call 1 error: %v", err)
+		}
+	}
+	if len(f.Tools) != 0 {
+		t.Errorf("after call 1: got %d tools, want 0", len(f.Tools))
+	}
+
+	// Second call: toolset now returns extraTool — f.Tools must be updated.
+	req2 := &model.LLMRequest{}
+	for _, err := range toolProcessor(ctx, req2, f) {
+		if err != nil {
+			t.Fatalf("toolProcessor call 2 error: %v", err)
+		}
+	}
+	if len(f.Tools) != 1 || f.Tools[0].Name() != extraTool.Name() {
+		t.Errorf("after call 2: got tools %v, want [%s]", f.Tools, extraTool.Name())
+	}
+
+	if ts.callCount != 2 {
+		t.Errorf("toolset.Tools() called %d times, want 2", ts.callCount)
+	}
+}
--- a/internal/llminternal/tools_processor.go
+++ b/internal/llminternal/tools_processor.go
@@ -24,13 +24,12 @@ import (
 	"google.golang.org/adk/session"
 )
 
-// ContentRequestProcessor populates the LLMRequest's Contents based on
-// the InvocationContext that includes the previous events.
+// toolProcessor populates f.Tools on every step by re-evaluating each
+// Toolset. Toolsets may return different tools based on session state that
+// was modified by an earlier step in the same Run(), so the tool list must
+// be rebuilt before each model call rather than being cached across steps.
 func toolProcessor(ctx agent.InvocationContext, req *model.LLMRequest, f *Flow) iter.Seq2[*session.Event, error] {
 	return func(yield func(*session.Event, error) bool) {
-		if f.Tools != nil {
-			return
-		}
 		llmAgent, ok := ctx.Agent().(Agent)
 		if !ok {
 			yield(nil, fmt.Errorf("agent %v is not an LLMAgent", ctx.Agent().Name()))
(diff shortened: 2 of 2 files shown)


nuthalapativarun:

nuthalapativarun:

nuthalapativarun:

nuthalapativarun:

karolpiotrowicz (commented):
The change itself is right, and it matches how adk-python has always behaved. `_process_agent_tools` there re-resolves every tool union, toolsets included, on each step ([base_llm_flow.py:458](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/flows/llm_flows/base_llm_flow.py#L458), awaited from `_preprocess_async` at [line 1435](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/flows/llm_flows/base_llm_flow.py#L1435), inside the step loop), and the comment at [lines 540-541](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/flows/llm_flows/base_llm_flow.py#L540-L541) states the reason outright: "Tool sets can change between model steps, so the cache is refreshed each time." Driving a state-gated toolset through `runner.Runner.Run`, step 2's tool list is `[activate]` on main and `[activate search_vehicles]` with this applied, so the fix works at the public entry point. Sorry this sat unreviewed for so long.

Two things before it can merge.

## The test passes on main, so it does not pin the fix

The fixture at [base_flow_test.go:804](https://github.com/google/adk-go/blob/c9b98da47a16a77cd28969ce41077c52a7cecfa1/internal/llminternal/base_flow_test.go#L804) is `&State{Toolsets: []tool.Toolset{ts}}`, which leaves `State.Tools` nil, and [`toolsByCall[0]`](https://github.com/google/adk-go/blob/c9b98da47a16a77cd28969ce41077c52a7cecfa1/internal/llminternal/base_flow_test.go#L799-L800) is nil too. Appending zero elements to a nil slice returns nil, so `f.Tools` is still nil after the first call and the guard you removed — [`if f.Tools != nil { return }`](https://github.com/google/adk-go/blob/058933d583478dcae7dd5d870f366b51e3b00ae1/internal/llminternal/tools_processor.go#L31) — would never have fired here. I put the merged test file onto an otherwise unmodified main and it passes:

```
$ go test -run TestToolProcessorReEvaluatesToolsetsEachStep ./internal/llminternal/
--- PASS: TestToolProcessorReEvaluatesToolsetsEachStep (0.00s)
ok  google.golang.org/adk/v2/internal/llminternal  0.117s
```

Giving the agent one static tool makes `f.Tools` non-nil after the first call, which is what the guard needs in order to trip. With this the test fails on main and passes with your change, and `./internal/llminternal/...` stays green:

```diff
-	agentState := &State{Toolsets: []tool.Toolset{ts}}
+	baseTool := &mockFunctionTool{name: "base_tool"}
+	agentState := &State{Tools: []tool.Tool{baseTool}, Toolsets: []tool.Toolset{ts}}
@@
-	if len(f.Tools) != 0 {
-		t.Errorf("after call 1: got %d tools, want 0", len(f.Tools))
+	if len(f.Tools) != 1 {
+		t.Errorf("after call 1: got %d tools, want 1", len(f.Tools))
 	}
@@
-	if len(f.Tools) != 1 || f.Tools[0].Name() != extraTool.Name() {
+	if len(f.Tools) != 2 || f.Tools[1].Name() != extraTool.Name() {
```

## Rebase

The branch is 104 commits behind and conflicts in `internal/llminternal/base_flow_test.go`. It is only that both sides append tests at the end of the file, so keeping both blocks resolves it, but it does need doing.

## Non-blocking, worth knowing

- **Toolsets that do I/O now pay per step.** `Toolset.Tools()` goes from one call per run to one per model call — 11 calls for an 11-model-call run, measured. That matters for [`mcptoolset`](https://github.com/google/adk-go/blob/058933d583478dcae7dd5d870f366b51e3b00ae1/tool/mcptoolset/set.go#L164-L165), where every call is a live paginated `ListTools` with no caching, and the loop at [tools_processor.go:39](https://github.com/google/adk-go/blob/c9b98da47a16a77cd28969ce41077c52a7cecfa1/internal/llminternal/tools_processor.go#L39) walks toolsets serially where Python runs them under `asyncio.gather` specifically to overlap those listings. Nothing to change in this PR — caching belongs in `mcptoolset` — but it is a consequence someone will hit.
- **A toolset that fails mid-run now aborts the invocation.** The [error return](https://github.com/google/adk-go/blob/c9b98da47a16a77cd28969ce41077c52a7cecfa1/internal/llminternal/tools_processor.go#L41-L44) is unchanged, but it was previously only reachable before the first model call. A toolset erroring on its third evaluation used to leave the run to finish, and now ends it with earlier steps' side effects already persisted. The same holds for a toolset that starts returning a name already taken by a static tool, which surfaces as `duplicate tool: "..."`. Python fails closed the same way, so this looks like the intended semantics rather than something to fix — it is just worth a line in the description so it is not a surprise.
- **The comment overreaches slightly on the live path.** [tools_processor.go:27-30](https://github.com/google/adk-go/blob/c9b98da47a16a77cd28969ce41077c52a7cecfa1/internal/llminternal/tools_processor.go#L27-L30) says the list "must be rebuilt before each model call", but `RunLive` calls `preprocess` once at [base_flow.go:336](https://github.com/google/adk-go/blob/058933d583478dcae7dd5d870f366b51e3b00ae1/internal/llminternal/base_flow.go#L336), outside the reconnect loop, so toolsets are resolved once per live session both before and after this change. Issue #757's symptom survives there. That is a separate gap, not something this PR needs to close — the comment just shouldn't imply it already has. The comment it replaced described `ContentRequestProcessor`, so this is still a clear improvement.

nuthalapativarun:
Rebased onto current `main` — resolved a straightforward conflict in `internal/llminternal/base_flow_test.go` where `main` had independently appended a different set of tests (thought-only-turn termination/reset tests and the partial-last-event regression test) right after the same insertion point this PR's `TestToolProcessorReEvaluatesToolsetsEachStep`/`dynamicToolset` additions land at; both sets of tests are now present, no logic changes needed. `go build ./...` and `go test ./internal/llminternal/...` are green. Force-pushed the rebase. This PR hasn't had a first review yet — would appreciate a look when you have a chance. Thanks!

internal/llminternal/base_flow_test.go
```diff
@@ -1223,3 +1223,65 @@ func captureLog(t *testing.T, fn func()) string {
...
+
+	// Call 0: no tools yet. Call 1+: return extraTool.
+	ts := &dynamicToolset{
+		toolsByCall: map[int][]tool.Tool{
+			0: nil,
+			1: {extraTool},
+		},
+	}
+
+	agentState := &State{Toolsets: []tool.Toolset{ts}}
```
karolpiotrowicz: With no static tools, `f.Tools` is still nil after the first call, so the removed `if f.Tools != nil` guard never fires and this test passes on main. One static tool makes the guard trip:

```diff
-	agentState := &State{Toolsets: []tool.Toolset{ts}}
+	baseTool := &mockFunctionTool{name: "base_tool"}
+	agentState := &State{Tools: []tool.Tool{baseTool}, Toolsets: []tool.Toolset{ts}}
@@
-	if len(f.Tools) != 0 {
-		t.Errorf("after call 1: got %d tools, want 0", len(f.Tools))
+	if len(f.Tools) != 1 {
+		t.Errorf("after call 1: got %d tools, want 1", len(f.Tools))
 	}
@@
-	if len(f.Tools) != 1 || f.Tools[0].Name() != extraTool.Name() {
+	if len(f.Tools) != 2 || f.Tools[1].Name() != extraTool.Name() {
```

I ran it: on main it fails with `toolset.Tools() called 1 times, want 2`, and with your change it passes and `./internal/llminternal/...` stays green.

karolpiotrowicz (changes requested):
One thing still blocks: the new test passes without the fix, so it would not catch the cache guard coming back. Before merge, the test needs to fail on main and pass with your change. The fixture-line comment has a change that does this. The full reasoning is in my [earlier review](https://github.com/google/adk-go/pull/785#pullrequestreview-5090704290), which I suspect was easy to miss, since your last comment mentions no review yet.

The branch is behind main again, with no conflicts at the moment.

PR #787, opened 2026-05-04
feat: implement explicit Default route support in workflow graph traversal

hanorik:
This PR introduces support for an explicit `Default` route in the `workflow` package, serving as a fallback when no other concrete routes match.

**Changes:**
- Added `defaultRoute` type and a global `Default` instance.
- Refactored `findNextNodes` to track and fallback to a `Default` route if no conditional edges match.
- Removed the `DEFAULT_ROUTE` constant.
- Removed error propagation from `findNextNodes` when no routes match; the workflow now proceeds with unconditional edges or the default route, or simply stops.

### DIFF
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -64,6 +64,15 @@ func (r BoolRoute) Matches(event *session.Event) bool {
 	return matchRoute(fmt.Sprint(r), event)
 }
 
+// DefaultRoute is a special route that matches when no other concrete routes match.
+var Default = &defaultRoute{}
+
+type defaultRoute struct{}
+
+func (r *defaultRoute) Matches(event *session.Event) bool {
+	return false
+}
+
 // baseNode provides common fields for all nodes.
 type baseNode struct {
 	name        string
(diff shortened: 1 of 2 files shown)


anFatum (approved):
LGTM

workflow/workflow.go
```diff
@@ -64,6 +64,15 @@ func (r BoolRoute) Matches(event *session.Event) bool {
 	return matchRoute(fmt.Sprint(r), event)
 }
 
+// defaultRoute is a special route that matches when no other concrete routes match.
+var Default = &defaultRoute{}
```
wolo-lab, line 81: Not sure if we need that. Please justify.
Can we use just workflow.DefaultRoute{} ?

hanorik: I thought about to make it more similar to the way the Start node defined as it seems to be a bit more elegant for me. I've changed it to use `workflow.DefaultRoute{}` instead but will leave the comment unresolved so you could confirm if you want to stay with this implementation or not.

wolo-lab (approved):
LGTM

workflow/workflow.go
hanorik: We discussed it privately and decided to stay with the origin implementation for now.

PR #788, opened 2026-05-04
fix: propagate thought signature to first function call in mixed responses

baptmont:
Propagates the thought signature of a candidate response to its first subsequent function call if it lacks one. This ensures Vertex AI validates the turn history when echoing it back on the next turn.


.gitignore
```diff
@@ -20,3 +20,4 @@ profile.cov
 # Go workspace file
 go.work
 go.work.sum
+.vscode
```
dpasiukevich: nit: I suggest to set up personal .gitignore for your local setup. And to keep this file only for artifacts produced by go build and test. Otherwise it will be a clutter of all possible tools&extensions for windows/linux/macos.

> .gitignore is restricted to the artifacts produced by go build and test.
> for personal setup you can use personal .gitignore: `git config --global core.excludesfile ~/.gitignore`

dpasiukevich (dismissed):

internal/llminternal/stream_aggregator.go
```diff
@@ -135,6 +140,10 @@ func (s *streamingResponseAggregator) processFunctionCallPart(part *genai.Part)
 	} else {
 		if part.FunctionCall.Name != "" {
 			s.flushTextBufferToSequence()
+			if part.ThoughtSignature == nil && s.currentThoughtSignature != nil {
+				part.ThoughtSignature = s.currentThoughtSignature
+			}
+			s.currentThoughtSignature = nil
```
adamglowacki, line 146: Should it really remove the thought signature here? Shouldn't this be stored until we stumble upon another non-empty thought signature?

I saw in the tests it's a deliberate decision. Why?

baptmont: According to the documentation only the first function call should have a thought signature.
"If the model generates parallel function calls in a response, the thought_signature is attached only to the first functionCall part. Subsequent functionCall parts in the same response will not contain a signature."
https://ai.google.dev/gemini-api/docs/thought-signatures

.gitignore
baptmont: done

dpasiukevich (approved):

PR #795, opened 2026-05-05
feat: add EdgeBuilder to simplify workflow graph construction

hanorik:
This PR introduces the `EdgeBuilder` utility in the `workflow` package to provide a fluent API for constructing workflow edges, making it easier to build complex workflow graphs.

**Changes:**
- Implements `EdgeBuilder` with methods for:
  - `Add`: Add a simple edge between two nodes.
  - `AddRoute`: Add an edge with a specific route value (supports string, int, bool).
  - `AddMultiRoute`: Add an edge that matches multiple routes.
  - `AddFanOut`: Connect one node to multiple destination nodes.
  - `AddFanIn`: Connect multiple source nodes to a single destination node.
  - `AddRoutes`: Connect one node to multiple nodes based on a map of routes.
- Added comprehensive unit tests for all `EdgeBuilder` methods to verify correct edge construction and routing.

### DIFF
--- a/workflow/edgebuilder.go
+++ b/workflow/edgebuilder.go
@@ -1,0 +1,94 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package workflow
+
+// EdgeBuilder provides a fluent API for building a list of Edges.
+type EdgeBuilder struct {
+	edges []Edge
+}
+
+// NewEdgeBuilder creates a new EdgeBuilder.
+func NewEdgeBuilder() *EdgeBuilder {
+	return &EdgeBuilder{}
+}
+
+// Add adds a new edge between two nodes.
+func (b *EdgeBuilder) Add(from, to Node) *EdgeBuilder {
+	b.edges = append(b.edges, Edge{From: from, To: to})
+	return b
+}
+
+// AddRoute adds a new edge with a route condition between two nodes.
+// The route condition is of type any and is converted to a Route interface.
+// Supported routes are StringRoute, IntRoute, BoolRoute and MultiRoute.
+func (b *EdgeBuilder) AddRoute(from, to Node, route Route) *EdgeBuilder {
+	b.edges = append(b.edges, Edge{From: from, To: to, Route: route})
+	return b
+}
+
+// AddFanOut adds multiple edges from a single source node to multiple target nodes.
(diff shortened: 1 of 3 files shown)


anFatum (approved):
LGTM

workflow/edgebuilder.go
```diff
@@ -0,0 +1,87 @@
...
+	var r Route
+	switch route := route.(type) {
+	case string:
+		r = StringRoute(route)
+	case int:
+		r = IntRoute(route)
+	case bool:
+		r = BoolRoute(route)
+	default:
+		panic(fmt.Sprintf("unsupported route type: %T", route))
```
wolo-lab: The implementation discovered a flaw in the api design. 
 The error will be discovered only at runtime. I believe our api would be safer if we would change `route any` to `Route: route`. The same for add `MultiRoute`, wdyt?

workflow/workflow_test.go
```diff
@@ -397,3 +395,161 @@ func TestWorkflowRouting(t *testing.T) {
 		})
 	}
 }
+
```
wolo-lab: new tests should be added to edgebuilder_test.go

workflow/edgebuilder.go
```diff
@@ -0,0 +1,87 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package workflow
+
+import "fmt"
+
+type EdgeBuilder struct {
```
wolo-lab, line 18: Please add docs here and to all exported methods.

workflow/workflow_test.go
```diff
@@ -397,3 +395,161 @@ func TestWorkflowRouting(t *testing.T) {
...
+	}
+	if edges[0].To != nodeB {
+		t.Errorf("expected To node to be %s, got %s", nodeB.Name(), edges[0].To.Name())
+	}
+	if edges[0].Route != nil {
+		t.Errorf("expected no route, got %v", edges[0].Route)
+	}
+}
+
+func TestAddRouteEdgeBuilder(t *testing.T) {
```
wolo-lab: lots of boilerplate repetition, please convert to table-driven tests.

workflow/workflow_test.go
```diff
@@ -397,3 +395,161 @@ func TestWorkflowRouting(t *testing.T) {
...
+		t.Errorf("expected no routes, got %v, %v", edges[0].Route, edges[1].Route)
+	}
+}
+
+func TestAddRoutesEdgeBuilder(t *testing.T) {
+	nodeA := &dummyNode{name: "A"}
+	nodeB := &dummyNode{name: "B"}
+	nodeC := &dummyNode{name: "C"}
+	
+	routes := map[any]Node{
```
wolo-lab: This test case actually made me realize that we don't need support different key types in real life. It would be weird for the previous node to return different data types.

So let's simplify this and have `routes map[string]Node`. Strings will be the most popular choice (my guess tbh) and this class is just a helper, so there will be alternative ways for other data types (however a bit less verbose). By that we will also get rid of checking types in runtime.

wolo-lab (changes requested):

wolo-lab (approved):
LGTM

PR #796, opened 2026-05-05
feat: add NodeConfig to the workflow nodes

anFatum:
Add implementation for the NodeConfig in workflows.

### DIFF
--- a/agent/workflowagent/workflow_test.go
+++ b/agent/workflowagent/workflow_test.go
@@ -28,6 +28,10 @@ import (
 	"google.golang.org/adk/workflow"
 )
 
+var defaultNodeConfig = workflow.NodeConfig{
+	RetryConfig: workflow.NewRetryConfig(),
+}
+
 // MockSession is a minimal implementation of session.Session for testing.
 type MockSession struct {
 	session.Session
@@ -106,8 +110,8 @@ func TestWorkflowAgent(t *testing.T) {
 		return input + " done", nil
 	}
 
-	nodeA := workflow.NewFunctionNode("upper", upperFn)
-	nodeB := workflow.NewFunctionNode("suffix", suffixFn)
+	nodeA := workflow.NewFunctionNode("upper", upperFn, defaultNodeConfig)
+	nodeB := workflow.NewFunctionNode("suffix", suffixFn, defaultNodeConfig)
 
 	edges := workflow.Chain(workflow.Start, nodeA, nodeB)
 
--- a/agent/workflowagent/workflow_test.go
+++ b/agent/workflowagent/workflow_test.go
@@ -28,9 +28,7 @@ import (
 	"google.golang.org/adk/workflow"
 )
 
-var defaultNodeConfig = workflow.NodeConfig{
-	RetryConfig: workflow.DefaultRetryConfig(),
-}
+var defaultNodeConfig = workflow.NodeConfig{}
 
 // MockSession is a minimal implementation of session.Session for testing.
 type MockSession struct {
--- a/agent/workflowagent/workflow_test.go
+++ b/agent/workflowagent/workflow_test.go
@@ -29,7 +29,7 @@ import (
 )
 
 var defaultNodeConfig = workflow.NodeConfig{
-	RetryConfig: workflow.NewRetryConfig(),
+	RetryConfig: workflow.DefaultRetryConfig(),
 }
 
 // MockSession is a minimal implementation of session.Session for testing.
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -286,7 +286,7 @@ func (s *startNode) Run(ctx agent.InvocationContext, input any) iter.Seq2[*sessi
 }
 func (s *startNode) Config() NodeConfig { return NodeConfig{} }
 
-const defaultRoute = "__DEFAULT__"
+const DEFAULT_ROUTE = "__DEFAULT__"
 
 // Workflow manages the workflow graph execution.
 type Workflow struct {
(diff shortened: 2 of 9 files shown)


workflow/workflow.go
```diff
@@ -26,11 +27,108 @@ import (
...
 // Node is the interface for all nodes in a workflow.
 type Node interface {
 	Name() string
 	Description() string
 	Run(ctx agent.InvocationContext, input any) iter.Seq2[*session.Event, error]
+	Config() NodeConfig
+}
+
+// NodeConfig defines the configuration for a node.
+type NodeConfig struct {
```
wolo-lab: Let's move configs to a separate file  (like config.go)

hanorik (approved):
LGTM

workflow/workflow.go
```diff
@@ -114,6 +218,7 @@ func NewFunctionNode[IN, OUT any](name string, fn func(ctx agent.InvocationConte
 	return &FunctionNode{
 		baseNode: baseNode{name: name},
 		fn:       wrappedFn,
+		config:   cfg,
```
wolo-lab: please drop the duplicate config field and pass cfg into the embedded baseNode:

```
return &FunctionNode{
    baseNode: baseNode{name: name, config: cfg},
    fn:       wrappedFn,
}
```

workflow/workflow.go
```diff
@@ -81,19 +182,22 @@ func (r MultiRoute[T]) Matches(event *session.Event) bool {
...
 func (b *baseNode) Name() string        { return b.name }
 func (b *baseNode) Description() string { return b.description }
+func (b *baseNode) Config() NodeConfig  { return b.config }
 
 // FunctionNode wraps a custom function.
 type FunctionNode struct {
 	baseNode
-	fn func(ctx agent.InvocationContext, input any) (any, error)
+	fn     func(ctx agent.InvocationContext, input any) (any, error)
+	config NodeConfig
```
wolo-lab: It's inside baseNode

workflow/workflow_test.go
```diff
@@ -20,13 +20,18 @@ import (
...
 	"testing"
+	"time"
 
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/agent"
 	"google.golang.org/adk/session"
 )
 
+var defaultNodeConfig = NodeConfig{
```
wolo-lab: in tests we want to fail fast (unless we test retry mechanism) meaning no retries, default (nil) config should be ok.

workflow/workflow.go
```diff
@@ -26,11 +27,108 @@ import (
...
+func NewRetryConfig(opts ...RetryOption) *RetryConfig {
+	cfg := defaultRetryConfig
+	for _, opt := range opts {
+		opt(&cfg)
+	}
+	return &cfg
+}
+
+// WithMaxAttempts sets the maximum number of attempts.
+func WithMaxAttempts(attempts int) RetryOption {
```
wolo-lab: in adk go I don't see any other applications of the pattern of functional options (withProperty(...)) for core entities. I like the idea but let's be consistent across the project.

workflow/workflow_test.go
```diff
@@ -204,6 +209,58 @@ func TestMultiRouteInt(t *testing.T) {
 	}
 }
 
+func TestNewRetryConfig(t *testing.T) {
```
wolo-lab: Table-driven tests are recommended: go/gotip/episodes/21, I think they will simplify a code a bit.

workflow/workflow.go
```diff
@@ -26,11 +27,108 @@ import (
...
 )
 
+// defaultRetryConfig is the default retry configuration for a node.
+var defaultRetryConfig = RetryConfig{
+	MaxAttempts:   5,
+	InitialDelay:  time.Second,
+	MaxDelay:      60 * time.Second,
+	BackoffFactor: 2.0,
+	Jitter:        1.0,
+	ShouldRetry: func(err error) bool {
```
wolo-lab: err unused, let's change to:
```
ShouldRetry: func(error) bool { return true },
```

wolo-lab (changes requested):

workflow/workflow.go
anFatum: I've removed the functional options, and added the usage of default retry config instead to replace the empty parameters with default ones, but still not sure that's the most convenient solution here as I needed to switch to pointers, since we need somehow to distinguish between zero and null value

wolo-lab: True. So maybe let's back to non-pointers but define how it should be built. I think it would be fair to require from users to use full config, by recommending building the config based on a default one:

```
rc := workflow.DefaultRetryConfig()  
rc.MaxAttempts = 10
```
by that default will be always set and no need to call `ptr(..)`

We could add doc like that:
```
// Recommended construction is via DefaultRetryConfig and override
// the fields you want to customize:
//
//	rc := workflow.DefaultRetryConfig()
//	rc.MaxAttempts = 10
//	cfg := workflow.NodeConfig{RetryConfig: &rc}
//
// Constructing via struct literal (RetryConfig{...}) is permitted
// but discouraged: any unset field defaults to its zero value, not
// to DefaultRetryConfig's value. The zero RetryConfig is a valid
// "no retry, no backoff, no jitter" policy.
```

WDYT?

agent/workflowagent/workflow_test.go
```diff
@@ -28,6 +28,10 @@ import (
 	"google.golang.org/adk/workflow"
 )
 
+var defaultNodeConfig = workflow.NodeConfig{
+	RetryConfig: workflow.DefaultRetryConfig(),
```
wolo-lab: Can we remove the config so tests won't be retried? RetryConfig : nil should mean "no retries".

workflow/workflow.go
anFatum: I removed the pointers and add the comment about the defaultretryconfig, but I still think we should consider introducing the functional options across the codebase, I believe there will be more optionals to come

wolo-lab: Yes, it sounds as a good way forward.

wolo-lab (approved):

PR #797, opened 2026-05-05
feat: implement node name validation and update workflow construction to return errors

hanorik:
This PR introduces node name validation to ensure that different nodes in a workflow do not share the same name. It also updates workflow construction functions to return errors on validation failure.

**Changes:**
- **`workflow/validator.go`** : Added `validateUniqueNames` to check that all nodes in the workflow have unique names, preventing accidental reuse or ambiguity.
- **`workflow/workflow.go`**:
  - Updated `Chain`, `Concat`, and `New` to return an `error`.
  - Integrated `validateUniqueNames` into these construction functions.
- **`workflow/workflow_test.go`**:
  - Updated existing tests to handle the new error returns from `Chain` and `New`.
  - Added `TestUniqueNames` to verify that duplicate node names are correctly identified and handled.

### DIFF
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -181,12 +181,15 @@ type Workflow struct {
 }
 
 // New creates a new Workflow engine with the given edges.
-func New(edges []Edge) *Workflow {
+func New(edges []Edge) (*Workflow, error) {
+	if err := validateUniqueNames(edges); err != nil {
+		return nil, err
+	}
 	adj := make(map[Node][]Edge)
 	for _, edge := range edges {
 		adj[edge.From] = append(adj[edge.From], edge)
 	}
-	return &Workflow{edges: adj}
+	return &Workflow{edges: adj}, nil
 }
 
 type nodeInput struct {
--- a/workflow/workflow_test.go
+++ b/workflow/workflow_test.go
@@ -96,8 +96,11 @@ func TestSequentialWorkflow(t *testing.T) {
 	nodeB := NewFunctionNode("suffix", suffixFn)
 
 	edges := Chain(Start, nodeA, nodeB)
-
-	w := New(edges)
+	
+	w, err := New(edges)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
 
 	mockCtx := &MockInvocationContext{
 		sess: nil,
@@ -364,10 +367,13 @@ func TestWorkflowRouting(t *testing.T) {
 			start.route = tc.startRoutes
 			edges := tc.edges(start, a, b, c, d)
 
-			w := New(edges)
+			w, err := New(edges)
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+
 			mockCtx := &MockInvocationContext{sess: nil}
 
-			var err error
 			for _, testErr := range w.Run(mockCtx) {
 				if testErr != nil {
 					err = testErr
--- a/agent/workflowagent/workflow.go
+++ b/agent/workflowagent/workflow.go
@@ -31,7 +31,10 @@ type Config struct {
 
 // New creates a new Workflow agent.
 func New(cfg Config) (agent.Agent, error) {
-	w := workflow.New(cfg.Edges)
+	w, err := workflow.New(cfg.Edges)
+	if err != nil {
+		return nil, err
+	}
 
 	return agent.New(agent.Config{
 		Name:                 cfg.Name,
--- a/workflow/validation.go
+++ b/workflow/validation.go
@@ -1,0 +1,50 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package workflow
+
+import (
+	"errors"
+	"fmt"
+)
+
+// ErrDuplicateNodeName is returned when an edge set contains two
+// distinct Node instances that share the same Name.
+var ErrDuplicateNodeName = errors.New("duplicate node name")
+
+// validateUniqueNames checks that all nodes in the edge set have unique names.
+// If duplicate node names are found, it returns an error. The equality between
+// nodes is checked by comparing the nodes directly.
+func validateUniqueNames(edges []Edge) error {
+	names := make(map[string]Node)
+	checkNode := func(node Node) error {
+		if storedNode, ok := names[node.Name()]; ok {
+			if storedNode != node {
+				return fmt.Errorf("%w: %s", ErrDuplicateNodeName, node.Name())
+			}
+		} else {
+			names[node.Name()] = node
+		}
+		return nil
+	}
+	for _, edge := range edges {
+		if err := checkNode(edge.From); err != nil {
+			return err
+		}
+		if err := checkNode(edge.To); err != nil {
+			return err
+		}
+	}
+	return nil
+}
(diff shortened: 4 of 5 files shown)


anFatum (approved):
LGTM

workflow/workflow.go
```diff
@@ -144,29 +137,37 @@ type Edge struct {
 }
 
 // Chain generates a slice of Edges to form a chain of nodes.
-func Chain(nodes ...Node) []Edge {
+func Chain(nodes ...Node) ([]Edge, error) {
```
wolo-lab: Actually, let's allow  input of any length. For zero we will get empty array, for 1 node, just array with one node.

workflow/validator.go
```diff
@@ -0,0 +1,38 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package workflow
+
+import "fmt"
+
+func validateUniqueNames(edges []Edge) error {
```
wolo-lab: please add a doc

Important to include: how equality between nodes is checked.

workflow/workflow.go
wolo-lab: We cannot now call Concat on Chain, e.g.

```
edges := workflow.Concat(
    workflow.Chain(start, a, b),     
    workflow.Chain(b, d)
)
```

Let's remove error

workflow/validator.go
```diff
@@ -0,0 +1,38 @@
+// Copyright 2026 Google LLC
```
wolo-lab: Sentinel errors should be errors.Is-able. See: 
http://google3/learning/gemini/agents/skills/go_readability/references/common_mentor_feedback.md?rcl=887007368&l=133

Example:
```
/ ErrDuplicateNodeName is returned when an edge set contains two
// distinct Node instances that share the same Name.
var ErrDuplicateNodeName = errors.New("duplicate node name")

func validateUniqueNames(edges []Edge) error {
    // ...
    return fmt.Errorf("%w: %s", ErrDuplicateNodeName, edge.From.Name())
}
```

workflow/workflow_test.go
```diff
@@ -397,3 +403,77 @@ func TestWorkflowRouting(t *testing.T) {
...
+func (n *dummyNode) Description() string { return "" }
+func (n *dummyNode) Run(ctx agent.InvocationContext, input any) iter.Seq2[*session.Event, error] {
+	return func(yield func(*session.Event, error) bool) {}
+}
+
+func TestUniqueNames(t *testing.T) {
+	tests := []struct {
+		name  string
+		edges []Edge
+		expectErr error
```
wolo-lab: not used in tests

workflow/workflow_test.go
```diff
@@ -397,3 +403,77 @@ func TestWorkflowRouting(t *testing.T) {
...
+		expectErrorMsg string
+	}{
+		{
+			name: "unique names",
```
wolo-lab: It could be more concise if we would have something like:

```
type nodeSetup struct{ aName, bName, cName string }

tests := []struct {
    name        string
    setup       nodeSetup
    wantErrMsg  string
}{
    {name: "unique names", setup: nodeSetup{"A", "B", "C"}},
    {name: "duplicate in From", setup: nodeSetup{"A", "A", "C"}, wantErrMsg: "duplicate node name: A"},
    // ...
}
```

workflow/workflow_test.go
```diff
@@ -397,3 +403,77 @@ func TestWorkflowRouting(t *testing.T) {
...
+	name string
+}
+
+func (n *dummyNode) Name() string        { return n.name }
+func (n *dummyNode) Description() string { return "" }
+func (n *dummyNode) Run(ctx agent.InvocationContext, input any) iter.Seq2[*session.Event, error] {
+	return func(yield func(*session.Event, error) bool) {}
+}
+
+func TestUniqueNames(t *testing.T) {
```
wolo-lab: maybe in validator_test.go?

workflow/workflow_test.go
```diff
@@ -397,3 +403,77 @@ func TestWorkflowRouting(t *testing.T) {
 		})
 	}
 }
+
+type dummyNode struct {
```
wolo-lab: one-line doc?

workflow/validator.go
wolo-lab: nit: The file contains function `validateXXX`, not a validator, so it would be better to rename it to `validation.go`

workflow/workflow.go
```diff
@@ -144,29 +137,37 @@ type Edge struct {
...
+func Concat(items ...any) ([]Edge, error) {
 	var edges []Edge
 	for _, item := range items {
 		switch v := item.(type) {
 		case Edge:
 			edges = append(edges, v)
 		case []Edge:
 			edges = append(edges, v...)
+		default:
+			return nil, fmt.Errorf("invalid item type: %T", item)
```
wolo-lab: The user looking at this gets invalid item type: int but doesn't learn what types are valid, proposal:
`fmt.Errorf("workflow.Concat: item must be Edge or []Edge, got %T", item)`

workflow/workflow_test.go
wolo-lab: move after tests, see: http://google3/learning/gemini/agents/skills/go_readability/references/common_mentor_feedback.md?rcl=887007368&l=343

workflow/validator.go
```diff
@@ -0,0 +1,38 @@
...
+// limitations under the License.
+
+package workflow
+
+import "fmt"
+
+func validateUniqueNames(edges []Edge) error {
+	names := make(map[string]Node)
+	for _, edge := range edges {
+		if node, ok := names[edge.From.Name()]; ok {
```
wolo-lab: there's identical code below (only diff is edge.from and edge.to), it could be wrapped by a function:

```
    if err := helperFunction(edge.From); err != nil {
        return err
    }
    if err := helperFunction(edge.To); err != nil {
        return err
    }
```

workflow/workflow.go
```diff
@@ -144,29 +137,37 @@ type Edge struct {
...
 	if len(nodes) < 2 {
-		return nil
+		return nil, fmt.Errorf("at least two nodes are required for a chain")
 	}
 	edges := make([]Edge, len(nodes)-1)
 	for i := 0; i < len(nodes)-1; i++ {
 		edges[i] = Edge{From: nodes[i], To: nodes[i+1]}
 	}
-	return edges
+	if err := validateUniqueNames(edges); err != nil {
```
wolo-lab: I thought again about having this checks here and I think it would be better to skip it there as it will be redundant and won't be enough. Let's treat Chain(...), Concat(...) as a syntax sugar just to improve expressiveness. 

Having said that, let's move these edge helpers/builder to a dedicated file.

workflow/workflow.go
```diff
@@ -188,12 +189,15 @@ type Workflow struct {
 }
 
 // New creates a new Workflow engine with the given edges.
-func New(edges []Edge) *Workflow {
+func New(edges []Edge) (*Workflow, error) {
+	if err := validateUniqueNames(edges); err != nil {
+		return nil, fmt.Errorf("workflow validation failed: %w", err)
```
wolo-lab: The reader will see "workflow validation failed: node names validation failed: duplicate node name: A" — three layers of wrapping for what is conceptually one error. Some wrappers could be skipped.

wolo-lab (changes requested):

workflow/workflow.go
hanorik: I will add this and move `Chain` and `Concat` to the edge builder file within the #795 .

workflow/validator.go
hanorik: Done.

workflow/workflow.go

workflow/validator.go
hanorik: Done.

workflow/workflow_test.go
hanorik: Removed.

workflow/workflow.go
hanorik: Removed.

workflow/workflow_test.go

workflow/validator.go
hanorik: Done.

workflow/workflow.go
hanorik: Removed.

workflow/workflow.go
hanorik: Removed.

wolo-lab (approved):

PR #799, opened 2026-05-06
feat: aggregate multiple llm errors and update metadata (closes #798)

### LINKED ISSUE #798: Improve error handling in Event Processor by aggregating multiple LLM failures
Currently, the EventProcessor only captures the first error encountered during an invocation. Any subsequent errors are ignored, which leads to incomplete diagnostic information when multiple issues occur.

This issue proposes:

Merging multiple LLM error messages into a single, concatenated string.
Including the full error message in the terminal event's metadata.
Ensuring task actions (escalate/transfer) are preserved even when errors are reported.

This follows the TODO left in processor.go:
// TODO(yarolegovich): consider merging responses if multiple errors can be produced during an invocation

aurelsandu:
**Description** 
This PR addresses the TODO in `processor.go` regarding error handling when multiple events report failures during an invocation. Previously, only the first error was captured.

### Changes
- **Error Aggregation**: If multiple LLM responses contain errors, they are now concatenated using a newline separator.
- **Metadata Enrichment**: Added the full error message to the terminal event's metadata under the `"error"` key for better visibility.
- **Tests**: Updated `processor_test.go` to validate the new aggregation logic and metadata structure.

Closes #798


aurelsandu:
Hi @baptmont and @dpasiukevich,

I wanted to follow up on this PR and let you know that I'm fully available if you have any questions, feedback, or objections regarding the implementation details.

Please let me know if there are specific areas where you'd like to see adjustments or a more detailed analysis. I'd be happy to collaborate and help move this forward whenever you have a moment to look into it.

Thanks!

indurireddy-TF:
Hi @aurelsandu Can you please resolve conflicts?

aurelsandu:
Conflicts resolved. I synced with upstream/main and integrated the error aggregation logic directly into the updated v2 processor structure (where the existing TODO about merging responses was left). Everything builds fine locally

server/adka2a/v2/processor_test.go
```diff
@@ -127,7 +127,10 @@ func TestEventProcessor_Process(t *testing.T) {
 			terminal: []a2a.Event{
 				toTaskFailedUpdateEvent(
 					task, errorFromResponse(&model.LLMResponse{ErrorCode: "1", ErrorMessage: "failed"}),
-					map[string]any{ToA2AMetaKey("error_code"): "1"},
+					map[string]any{
+						ToA2AMetaKey("error_code"): "1",
+						"error":                    `llm error response: "failed"`,
```
karolpiotrowicz, line 132: This expectation is one of the six that fail. The fixture above sets `ErrorCode: "1"`, so `errorFromResponse` produces `llm error response (code 1): "failed"` and the `(code 1)` is missing here. The same omission is at lines 149, 170, 191, 352 and 432, and in the `errors.New(...)` causes at 146 and 429.

The cases that predate this PR pass because they build the expected string by calling `errorFromResponse(...)` rather than typing it out. Doing the same here would keep these in step with the formatter automatically.

server/adka2a/v2/processor_test.go
```diff
@@ -139,8 +142,12 @@ func TestEventProcessor_Process(t *testing.T) {
 			},
 			terminal: []a2a.Event{
 				toTaskFailedUpdateEvent(
-					task, errorFromResponse(&model.LLMResponse{ErrorCode: "1", ErrorMessage: "failed 1"}),
-					map[string]any{ToA2AMetaKey("error_code"): "1"},
+					task,
+					errors.New("llm error response: \"failed 1\"\nllm error response: \"failed 2\""),
+					map[string]any{
+						ToA2AMetaKey("error_code"): "2",
```
karolpiotrowicz, line 148: This case is still named `the first failure is returned` at line 138, but it now asserts a concatenation of both failures and `adk_error_code: "2"` — the second one. The name says the opposite of what the case checks, and it is the sort of thing a later bisect reads as the invariant.

It also overlaps almost exactly with `concatenate multiple failures` added at line 420: two coded errors, no content parts, differing only in the string literals. Rather than keeping both, one of them could cover a shape nothing currently reaches — an error event with no `ErrorCode`, three or more errors, or content parts on an error event followed by another error.

server/adka2a/v2/processor_test.go
```diff
@@ -15,13 +15,13 @@
 package adka2a
 
 import (
-	"testing"
-
+	"errors"
```
karolpiotrowicz, line 18: `errors` lands in the third-party group here and `testing` moves into it at line 24, which leaves the file with no separate standard-library group. `golangci-lint` fails on this: `processor_test.go:18:1: File is not properly formatted (gofumpt)`. It is clean on the base commit, so it comes in with this change.

Worth noting `gofmt -l` stays silent, since gofmt sorts within groups but never moves an import between them. Only the repo's gofumpt and goimports configuration catches it.

server/adka2a/v2/processor.go
```diff
@@ -93,14 +94,22 @@ func (p *eventProcessor) process(ctx context.Context, event *session.Event) (*a2
...
+		if p.failedEvent != nil {
+			if oldErrStr, ok := p.failedEvent.Metadata["error"].(string); ok {
+				mergedErr = errors.New(oldErrStr + "\n" + newErr.Error())
+			}
 		}
+
+		// terminal event might add additional keys to its metadata when it's dispatched and these changes should
+		// not be reflected in this event's metadata
+		terminalEventMeta := maps.Clone(eventMeta)
+		terminalEventMeta["error"] = mergedErr.Error()
```
karolpiotrowicz, line 109: This key should be `ToA2AMetaKey("error")`. Every other ADK-authored key on an A2A event carries the `adk_` prefix, `adk_error_code` included, and that one is written into this very map. The terminal event currently goes out with `[adk_invocation_id adk_author adk_error_code error]`.

Nothing reads it today, so nothing breaks — but this is a wire surface, and once a release ships the unprefixed name, renaming it is a breaking change for anyone who started consuming it. No CI check can catch this, because it is not an exported Go identifier.

When you rename it, line 101 reads the accumulated string back out of the same key. Changing the write without the read makes that type assertion fail silently and drops back to last-error-only, with nothing to catch it. Keeping the accumulated errors in a field on `eventProcessor` would remove the coupling.

server/adka2a/v2/processor.go
```diff
@@ -93,14 +94,22 @@ func (p *eventProcessor) process(ctx context.Context, event *session.Event) (*a2
...
+
+		p.failedEvent = toTaskFailedUpdateEvent(p.reqCtx, mergedErr, terminalEventMeta)
+		p.failedEvent.Status.Message.Parts = append(p.failedEvent.Status.Message.Parts, parts...)
```
karolpiotrowicz, line 112: Rebuilding `p.failedEvent` on every error means the terminal status message carries only the last error event's content parts. Before this change it carried only the first event's, so this is not a loss — I checked both directions and each side keeps exactly one event's parts. The content also still reaches the client either way through the artifact update below.

It is an inconsistency rather than a defect: the change now aggregates error *text* across every error while parts stay one-event-wins. If aggregation is the goal, carrying the accumulated parts forward instead of starting a fresh message would make the two agree. Not blocking.

karolpiotrowicz (commented):
The aggregation logic does what the issue asks for, and the shape is right — it reuses `errorFromResponse` rather than reimplementing the format, keeps the clone-before-mutate behavior, and leaves the escalate/transfer path alone. Three things need fixing before it can land, and the first one blocks.

## The package's tests fail on this branch

`go test ./server/adka2a/v2/...` fails six subtests on this branch, and on the result of merging it into `main`. It passes on the merge-base, so the branch introduces it.

The `"error"` expectations added by this PR are written out by hand as `llm error response: "failed"`, but every fixture that produces them sets an `ErrorCode`, and [`errorFromResponse`](https://github.com/google/adk-go/blob/315d97c2f4dd1c8573d7ce37e9c9eea20932054e/server/adka2a/v2/processor.go#L196-L201) emits `llm error response (code 1): "failed"` whenever a code is present. The same omission is in the `errors.New(...)` expected causes at [processor_test.go#L146](https://github.com/google/adk-go/blob/315d97c2f4dd1c8573d7ce37e9c9eea20932054e/server/adka2a/v2/processor_test.go#L146) and [#L429](https://github.com/google/adk-go/blob/315d97c2f4dd1c8573d7ce37e9c9eea20932054e/server/adka2a/v2/processor_test.go#L429).

Worth knowing where this came from, because it is not carelessness: `errorFromResponse` grew the `(code %s)` form in [the structured-error-propagation commit](https://github.com/google/adk-go/commit/05d6b957bbdfb0273a666e21be48f0f3a56e0987), which arrived on this branch through its merge of `main` in June. The expectations that predate this PR survived that change because they *compute* the string by calling `errorFromResponse(...)`. The new ones spell it out, and so drifted. Building the expectations the same way the older cases do would make this class of drift impossible.

The failing subtests are `failed_without_artifacts`, `the_first_failure_is_returned`, `failed_with_artifacts`, `failed_before_receiving_all_parts`, `actions_not_overwritten_by_subsequent_events`, and `concatenate_multiple_failures`.

```
$ go test ./server/adka2a/v2/...
--- FAIL: TestEventProcessor_Process/failed_without_artifacts
    Metadata: map[string]any{
            "adk_error_code": string("1"),
    -       "error":          string(`llm error response: "failed"`),
    +       "error":          string(`llm error response (code 1): "failed"`),
    }
```

## The error text and `adk_error_code` stop describing the same event

[processor.go#L108](https://github.com/google/adk-go/blob/315d97c2f4dd1c8573d7ce37e9c9eea20932054e/server/adka2a/v2/processor.go#L108) re-clones the terminal metadata from the current error event each time round, so once there is more than one error the message text describes all of them while `adk_error_code` describes at most one. Before this change the two always agreed, because only the first error was ever reported.

The case that costs something is a coded error followed by an uncoded one, where the code disappears from the terminal event entirely even though the text still names it. That matters downstream: [`processA2AMeta`](https://github.com/google/adk-go/blob/315d97c2f4dd1c8573d7ce37e9c9eea20932054e/server/adka2a/v2/metadata.go#L163-L165) copies `adk_error_code` into a consuming ADK parent's `LLMResponse.ErrorCode`, so that parent ends up with an empty code beside an error text reading `(code 429)`. Uncoded error events are produced in-tree by [`toErrorEvent`](https://github.com/google/adk-go/blob/315d97c2f4dd1c8573d7ce37e9c9eea20932054e/agent/remoteagent/v2/a2a_agent.go#L366-L372) and by the A2A ingest path at [events.go#L266-L268](https://github.com/google/adk-go/blob/315d97c2f4dd1c8573d7ce37e9c9eea20932054e/server/adka2a/v2/events.go#L266-L268), so a local model error followed by a remote sub-agent failure produces exactly this. I did not drive an end-to-end run that produces the sequence, so treat the ordering as reachable-by-construction rather than observed.

<details>
<summary>Reproduction — drop this in <code>server/adka2a/v2/</code> and run <code>go test -run TestCodeCase -v ./server/adka2a/v2/</code></summary>

```go
func TestCodeCase(t *testing.T) {
	ev := func(code, msg string) *session.Event {
		return &session.Event{LLMResponse: model.LLMResponse{ErrorCode: code, ErrorMessage: msg}}
	}
	for _, tc := range []struct {
		name   string
		events []*session.Event
	}{
		{"coded then uncoded", []*session.Event{ev("429", "rate limited"), ev("", "stream closed")}},
		{"uncoded then coded", []*session.Event{ev("", "stream closed"), ev("500", "internal")}},
		{"both coded", []*session.Event{ev("429", "rate limited"), ev("500", "internal")}},
	} {
			task := &a2a.Task{ID: a2a.NewTaskID(), ContextID: a2a.NewContextID()}
			reqCtx := &a2asrv.ExecutorContext{TaskID: task.ID, ContextID: task.ContextID}
			p := newEventProcessor(reqCtx, invocationMeta{}, nil, newLegacyArtifactMaker(reqCtx))
			for i, e := range tc.events {
				if _, err := p.process(t.Context(), e); err != nil {
					t.Fatalf("process(%d) = %v", i, err)
				}
			}
			term := p.makeFinalStatusUpdate()
			v, ok := term.Metadata[ToA2AMetaKey("error_code")]
			t.Logf("adk_error_code present=%v value=%v | text=%q",
				ok, v, string(term.Status.Message.Parts[0].Content.(a2a.Text)))
		})
	}
}
```

Observed, before the change and with it applied:

```
                              before                     with this PR
coded(429) then uncoded       code=429                   code ABSENT
                              text: "(code 429) ..."     text: "(code 429) ...\n... uncoded"
uncoded then coded(500)       code ABSENT                code=500
coded(429) then coded(500)    code=429                   code=500
```

</details>

To be clear about what I am *not* raising: the rest of the metadata moving from the first error to the last — `adk_author`, `adk_invocation_id`, `adk_branch`, and the citation, grounding and usage entries — is symmetric. Before the change it described the first error, now it describes the last, and nothing says which is correct. That part reads as a design choice, not a defect.

## The new metadata key is not namespaced

[processor.go#L109](https://github.com/google/adk-go/blob/315d97c2f4dd1c8573d7ce37e9c9eea20932054e/server/adka2a/v2/processor.go#L109) writes the bare key `"error"` rather than `ToA2AMetaKey("error")`. Every other ADK-authored key on an A2A event carries the `adk_` prefix, including `adk_error_code` sitting in the very same map — see [metadata.go#L29-L46](https://github.com/google/adk-go/blob/315d97c2f4dd1c8573d7ce37e9c9eea20932054e/server/adka2a/v2/metadata.go#L29-L46). The terminal event currently ships `[adk_invocation_id adk_author adk_error_code error]` where it used to ship the first three.

Nothing breaks today — no ADK code reads a bare `error` key, and `processA2AMeta` only looks at `adk_`-prefixed ones. The reason to fix it now rather than later is that this is a wire surface, so once a release goes out with the unprefixed name, renaming it becomes a breaking change for anyone who started reading it. Nothing in CI can catch this either, since it is not an exported Go identifier.

One thing to watch when fixing it: the merge at [processor.go#L101](https://github.com/google/adk-go/blob/315d97c2f4dd1c8573d7ce37e9c9eea20932054e/server/adka2a/v2/processor.go#L101) reads the accumulated string back out of that same key. Renaming the write without renaming the read makes the type assertion fail silently and quietly reverts to last-error-only, with no test covering it. Holding the accumulated errors in a field on `eventProcessor` instead would decouple the two and remove the trap.

## Smaller things

There are inline comments on the import grouping, the stale `the first failure is returned` case name, the overlap between the two multi-error test cases, and the terminal-event parts behavior. None of those block.

One more, which is about the existing tests rather than this change: the executor-level tests all pass `cmpopts.IgnoreFields(a2a.TaskStatusUpdateEvent{}, "Metadata")` ([executor_test.go#L214](https://github.com/google/adk-go/blob/315d97c2f4dd1c8573d7ce37e9c9eea20932054e/server/adka2a/v2/executor_test.go#L214)), so the only tests that drive the real executor path cannot see terminal metadata at all. That predates this PR, but it is why both of the wire-level points above would otherwise reach a release uncovered — worth one case asserting the terminal metadata keys explicitly.

Nothing else came up. The build is clean on both sides, `go vet` and `staticcheck` are clean, the dependency set is untouched, and no exported API changes.

PR #800, opened 2026-05-06
stale-bot-implementation

indurireddy-TF:

- Closes: #_issue_number_
- Related: #_issue_number_

possible._

**Problem:**
Maintaining inactive issues manually is time-consuming and inconsistent.
There was no automated mechanism to:

Identify stale issues based on inactivity
Handle maintainer–user interaction intelligently
Detect silent updates (description edits without comments)
Automatically close issues after a defined threshold

This leads to:

Accumulation of outdated issues
Reduced repository hygiene
Increased manual effort for maintainers
**Solution:**
Implemented an AI-driven stale bot that:

Analyzes issue activity using GitHub GraphQL + timeline events
Differentiates between:
Author activity
Maintainer activity
Internal maintainer discussions
Detects silent edits and alerts maintainers
Applies intelligent decision logic:
Marks issues as stale after inactivity
Removes stale label when user responds
Closes issues after stale threshold is exceeded
Uses an LLM agent with a structured decision tree for consistent behavior

### Testing Plan

**Unit Tests:**

- [X] I have added or updated unit tests for my change.
- [X] All unit tests pass locally.

### Checklist

- [X] I have added tests that prove my fix is effective or that my feature works.

### Additional context

image


.github/contributing/samples/stale-bot-agent/PROMPT_INSTRUCTION.txt
```diff
@@ -0,0 +1,73 @@
...
+        - **Report**: "Analysis for Issue #[number]: ACTIVE. Maintainer is discussing with another maintainer. No action."
+
+    - **Question Check**: Does the text ask a question, request clarification, ask for logs, or give suggestions?
+    - **Time Check**: Is `days_since_activity` > {stale_threshold_days}?
+
+    - **DECISION**:
+        - **IF (Question == YES) AND (Time == YES) AND (Internal Discussion Check == FALSE):**
+            - **Action**: Call `add_stale_label_and_comment`.
+            - **Check**: If '{REQUEST_CLARIFICATION_LABEL}' is not in `current_labels`, call `add_label_to_issue` with '{REQUEST_CLARIFICATION_LABEL}'.
+            - **Report**: "Analysis for Issue #[number]: STALE. Maintainer asked question {days_since_activity} days ago. Marking stale."
```
karolpiotrowicz, line 69: **Blocking — the agent fails on every run.** This line was changed from the Python sample's safe square brackets `[days_since_activity]` to curly braces `{days_since_activity}`. ADK treats `{name}` as a session-state placeholder: the static `Instruction` is run through `InjectSessionState` on *every* model call (`internal/llminternal/instruction_processor.go`), which looks up `days_since_activity` in session state, doesn't find it, and returns an error. So `runner.Run` yields `(nil, err)` on the first turn of every issue and the agent never takes any action.

Fix: revert to `[days_since_activity]` (it's only display text inside a report string), or use the optional form `{days_since_activity?}`. Better still, keep runtime values out of the static instruction entirely.

.github/contributing/samples/stale-bot-agent/main.go
```diff
@@ -0,0 +1,306 @@
...
+		promptText := fmt.Sprintf("Audit Issue #%d.", issueNumber)
+		promptMessage := &genai.Content{
+			Role: "user",
+			Parts: []*genai.Part{
+				{Text: promptText},
+			},
+		}
+
+		eventStream := r.Run(ctx, UserID, sess.Session.ID(), promptMessage, agent.RunConfig{})
+		for event := range eventStream {
```
karolpiotrowicz, line 114: **`runner.Run` returns `iter.Seq2[*session.Event, error]` — the error is being dropped.** Ranging with a single variable binds only the event and silently discards every error. This (a) hides the instruction bug above (the `(nil, err)` event is swallowed, so the run exits cleanly while doing nothing) and (b) models the wrong pattern in a sample. Use both values:

```go
for event, err := range eventStream {
    if err != nil {
        log.Printf("#%d run error: %v", issueNumber, err)
        return
    }
    // ... handle event
}
```

For a flagship sample, correct error handling on the run loop is the single most important thing to get right.

karolpiotrowicz (commented):
## Maintainer review: stale-bot sample

Thanks for porting this — the decision tree, the single GraphQL query, and the tool set faithfully follow the adk-python `adk_stale_agent`, and the overall flow is easy to follow. Before it can land there are a few **blocking functional issues**, and since this is meant to double as a user-facing ADK Go sample, a number of idiomatic-Go cleanups. I cross-checked behavior against the Python original and verified the wiring against the current ADK Go APIs (`runner`, `llmagent`, `functiontool`, `session`, and the instruction processor).

### Blocking (won't work as-is)
1. **Agent errors on every run** — `PROMPT_INSTRUCTION.txt` uses `{days_since_activity}` (curly braces), which ADK resolves as a session-state placeholder and fails on. The Python sample uses safe square brackets. (inline)
2. **`runner.Run` errors are dropped** — `for event := range` discards the error half of `iter.Seq2[*session.Event, error]`, which silently masks #1. (inline)
3. **The workflow can't find the code** — `working-directory` omits the `.github/` prefix the files actually live under. (inline)

Because of #1 + #2, the one path that does execute (`go run .`) finishes green while doing nothing.

### Robustness
- Unchecked type assertions panic on a malformed GraphQL response (inline).
- Data race on the `maintainersCache` global under concurrent issue processing (inline).

### Sample quality / idiomatic Go
- SCREAMING_SNAKE_CASE names, pervasive mutable package globals, a deprecated `tool.Context`, ineffective `description:` struct tags, inconsistent tool return types, ignored constructor errors, leftover Python comments, and `any` vs `interface{}` — all flagged inline.
- The sample lives under `.github/`, outside the Go module: it isn't built/vetted/tested by repo CI (the `./...` wildcard skips dot-directories), and the workflow builds it against the released module instead of this repo. Recommend moving to `examples/stale-bot-agent/`. (inline)
- A `README.md` (like the Python sample has) would help users a lot.

### On the tests
The unit tests pass, but they cover only the pure-logic helpers and happy-path tool HTTP calls — they never construct the agent or run the instruction path, so they catch none of the issues above, and they aren't executed by any CI today. Details inline.

Happy to pair on a follow-up that relocates this to `examples/`, switches to a struct-based design (no globals), and fixes the blockers.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
+
+      - name: Set up Go
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff
+        with:
+          go-version: '1.26' # Use a stable Go version
+
+      - name: Initialize and Sync Dependencies
+        working-directory: contributing/samples/stale-bot-agent
```
karolpiotrowicz, line 30: **The workflow can't find the code.** `working-directory` is `contributing/samples/stale-bot-agent`, but the files are added under `.github/contributing/samples/stale-bot-agent/`. This step (and the run step at line 39) will fail with a 'directory not found' error. Either add the `.github/` prefix to both paths, or — preferred — move the sample into the module (see the comment on line 34).

.github/workflows/stale.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          # Initialize module only if go.mod is missing or invalid
+          if [ ! -f go.mod ] || ! grep -q "module" go.mod; then
+            go mod init stale-bot-agent
```
karolpiotrowicz, line 34: `go mod init` + `go mod tidy` fabricates a throwaway module at CI time that pulls `google.golang.org/adk` from the module proxy (latest release) instead of building against this repo. So the in-repo sample is never compiled or tested against the code it ships with (API-skew risk).

ADK Go samples live in `examples/` (e.g. `examples/quickstart`, `examples/skills`) and are part of the repo module, so the compiler, `go vet`, and `-race` cover them automatically. Recommend relocating to `examples/stale-bot-agent/` and dropping the `go mod init` dance. Note: anything under `.github/` is also skipped by `go build ./...` / `go test ./...`, because the `...` wildcard ignores dot-directories — so the bundled tests never run in CI today.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,48 @@
...
+      contents: read
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
+
+      - name: Set up Go
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff
+        with:
+          go-version: '1.26' # Use a stable Go version
```
karolpiotrowicz, line 27: `go-version: '1.26'` — adk-go targets Go 1.24+. Pin to a currently-released, supported version (e.g. `'1.24'` / `'1.25'`) rather than one labeled 'stable' that may not be installable on the runner.

.github/contributing/samples/stale-bot-agent/agent.go
```diff
@@ -0,0 +1,662 @@
...
+			if firstErr, ok := errList[0].(map[string]any); ok {
+				return nil, fmt.Errorf("GraphQL Error: %v", firstErr["message"])
+			}
+		}
+		return nil, fmt.Errorf("GraphQL Error: unknown format")
+	}
+
+	data := resp["data"].(map[string]any)
+	repo := data["repository"].(map[string]any)
+	issue := repo["issue"]
```
karolpiotrowicz, line 195: Unchecked type assertions will panic on any unexpected payload — e.g. a GraphQL error response (no `data`), or `repository`/`issue` being `null`. The Python original uses `.get(...)` with defaults and surfaces a clean error; here a panic is only caught by the coarse `recover()` in `processSingleIssue`. Use comma-ok and return an error instead:

```go
data, ok := resp["data"].(map[string]any)
if !ok {
    return nil, fmt.Errorf("unexpected GraphQL response shape")
}
```

.github/contributing/samples/stale-bot-agent/agent.go
```diff
@@ -0,0 +1,662 @@
...
+	"log"
+	"math"
+	"sort"
+	"strings"
+	"time"
+
+	"google.golang.org/adk/tool"
+)
+
+var maintainersCache []string
```
karolpiotrowicz, line 28: **Data race.** `maintainersCache` is a package global read/written in `getCachedMaintainers` with no synchronization, but `main.go` processes issues concurrently (goroutines up to `ConcurrencyLimit`). On the first batch, multiple goroutines hit this cache and race. Python is safe because of the GIL; Go is not — `go run -race` / `go test -race` would flag it. Note the API-call counter right next door (`util.go`) *is* mutex-guarded, so this is inconsistent. Guard with `sync.Once`/`sync.Mutex`, or (better) inject the maintainer list via a struct rather than a global.

.github/contributing/samples/stale-bot-agent/config.go
```diff
@@ -0,0 +1,108 @@
...
+	Owner string
+	Repo  string
+
+	// Labels
+	StaleLabelName            = "stale"
+	RequestClarificationLabel = "request clarification"
+
+	// Thresholds (hours)
+	STALE_HOURS_THRESHOLD             float64
+	CLOSE_HOURS_AFTER_STALE_THRESHOLD float64
```
karolpiotrowicz, line 36: Go uses MixedCaps, never SCREAMING_SNAKE_CASE — `STALE_HOURS_THRESHOLD` / `CLOSE_HOURS_AFTER_STALE_THRESHOLD` (plus `BOT_NAME`, `BOT_ALERT_SIGNATURE` in agent.go and `PROMPT_TEMPLATE` in main.go) read as Python transliterated into Go. Rename to `staleThreshold`, `closeThreshold`, etc.

More broadly, this file plus the other globals (`Owner`, `Repo`, `GitHubToken`, `httpClient`, `maintainersCache`, `apiCallCount`, `rootAgent`) lean heavily on mutable package state. An exemplary sample would encapsulate config and the GitHub/HTTP client in a struct and pass it to the tools — that removes the globals, fixes the race above, and makes the tests independent.

.github/contributing/samples/stale-bot-agent/agent.go
```diff
@@ -0,0 +1,662 @@
...
+}
+
+func errorResponse(msg string) map[string]any {
+	return map[string]any{
+		"status": "error",
+		"error":  msg,
+	}
+}
+
+func addLabelToIssue(ctx tool.Context, args LabelTargetArgs) (ToolResult, error) {
```
karolpiotrowicz, line 428: `tool.Context` is a deprecated alias (`// Deprecated: use agent.ToolContext`). It compiles via the alias — which is why the build passes — but a sample should model current best practice. Use `agent.ToolContext` here and in the other five handlers (`removeLabelFromIssue`, `addStaleLabelAndComment`, `alertMaintainerOfEdit`, `closeAsStale`, `getIssueState`).

.github/contributing/samples/stale-bot-agent/agent.go
```diff
@@ -0,0 +1,662 @@
...
+	LastActivityTime time.Time `json:"last_activity_time"`
+	LastActionType   string    `json:"last_action_type"`
+	LastCommentText  *string   `json:"last_comment_text"`
+	LastActorName    string    `json:"last_actor_name"`
+}
+
+// Struct for tools that only need an Issue Number
+// Used by: addStaleLabelAndComment, alertMaintainerOfEdit, closeAsStale, getIssueState
+type IssueTargetArgs struct {
+	IssueNumber int `json:"issue_number" description:"The number of the GitHub issue to act upon"`
```
karolpiotrowicz, line 52: The `description:"..."` struct tag is not recognized by `jsonschema-go` (used by `functiontool` to build the tool's input schema), so these argument descriptions never reach the model. The schema is inferred from the Go types; to attach a description use the `jsonschema` tag, e.g.

```go
IssueNumber int `json:"issue_number" jsonschema:"the GitHub issue number to act upon"`
```

Same applies to the tags in `LabelTargetArgs` (lines 58-59).

.github/contributing/samples/stale-bot-agent/agent.go
```diff
@@ -0,0 +1,662 @@
...
+
+// Struct for tools that need an Issue Number AND a Label Name
+// Used by: addLabelToIssue, removeLabelFromIssue
+type LabelTargetArgs struct {
+	IssueNumber int    `json:"issue_number" description:"The number of the GitHub issue"`
+	LabelName   string `json:"label_name" description:"The specific name of the label"`
+}
+
+func getCachedMaintainers() ([]string, error) {
+	// if _MAINTAINERS_CACHE is not None: return it
```
karolpiotrowicz, line 63: Leftover Python pseudocode comment (`if _MAINTAINERS_CACHE is not None: return it`). Please remove — small thing, but these stand out in a reference sample.

.github/contributing/samples/stale-bot-agent/agent.go
```diff
@@ -0,0 +1,662 @@
...
+	if maintainersCache != nil {
+		return maintainersCache, nil
+	}
+
+	log.Println("Initializing Maintainers Cache...")
+
+	url := fmt.Sprintf("%s/repos/%s/%s/collaborators", GitHubBaseURL, Owner, Repo)
+	params := map[string]interface{}{
```
karolpiotrowicz, line 71: `map[string]interface{}` here (and `[]interface{}` on line 94) is inconsistent with the `map[string]any` / `[]any` used throughout the rest of this file. Prefer `any` (Go 1.18+) consistently.

.github/contributing/samples/stale-bot-agent/agent.go
```diff
@@ -0,0 +1,662 @@
...
+		}, err
+	}
+
+	return ToolResult{
+		Status: "success",
+	}, nil
+}
+
+// getIssueState orchestrates the fetching and analysis of an issue.
+func getIssueState(ctx tool.Context, args IssueTargetArgs) (map[string]any, error) {
```
karolpiotrowicz, line 579: Tool return types are inconsistent: most tools return the typed `ToolResult` struct, but `getIssueState` and `errorResponse` (line 421) return `map[string]any`. Pick one convention — idiomatic Go favors a typed result struct everywhere, which also documents the output shape for readers. (Also: `ToolResult` is declared in main.go but used here; co-locate it with the tools.)

.github/contributing/samples/stale-bot-agent/main.go
```diff
@@ -0,0 +1,306 @@
...
+	path := filepath.Join(baseDir, filename)
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return "", err
+	}
+	return string(data), nil
+}
+
+func setupTools() []tool.Tool {
+	t1, _ := functiontool.New(functiontool.Config{
```
karolpiotrowicz, line 151: `functiontool.New` errors are discarded for all six tools (`t1, _ := ...`). A sample should check them, and this reads as a lot of near-identical boilerplate. Consider a small table + loop:

```go
specs := []struct {
    cfg functiontool.Config
    fn  any
}{ ... }
for _, s := range specs {
    t, err := functiontool.New(s.cfg, s.fn)
    if err != nil {
        return nil, err
    }
    tools = append(tools, t)
}
```

.github/contributing/samples/stale-bot-agent/main.go
```diff
@@ -0,0 +1,306 @@
...
+	log.Println("PROMPT_TEMPLATE loaded successfully.")
+	log.Printf("--- Starting Stale Bot for %s/%s ---", Owner, Repo)
+	log.Printf("Concurrency level set to %d", ConcurrencyLimit)
+
+	model, err := gemini.NewModel(ctx, geminiModel, &genai.ClientConfig{APIKey: os.Getenv("GOOGLE_API_KEY")})
+	if err != nil {
+		log.Fatalf("Failed to create model: %v", err)
+	}
+
+	instruction := formatPrompt(PROMPT_TEMPLATE, map[string]string{
```
karolpiotrowicz, line 214: This is where the leftover `{days_since_activity}` placeholder (see `PROMPT_INSTRUCTION.txt`) would need to be resolved: `formatPrompt` only substitutes the keys in this map, so any other `{...}` token survives into the instruction and then errors inside ADK's state injection. Adding the missing key here is one fix, but keeping runtime values out of the static instruction is cleaner.

.github/contributing/samples/stale-bot-agent/stale_advance_test.go
```diff
@@ -0,0 +1,344 @@
...
+
+func newMockResponse(status int, body string) *http.Response {
+	return &http.Response{
+		StatusCode: status,
+		Body:       io.NopCloser(bytes.NewBufferString(body)),
+		Header:     make(http.Header),
+	}
+}
+
+func setupMockHTTP(handler func(req *http.Request) *http.Response) {
```
karolpiotrowicz, line 42: Two things about the test suite as a whole, anchored here:

1. **It exercises none of the agent wiring.** There are no `adk` imports — the tests cover pure helpers (`getEnv`, `formatDays`, `isMaintainer`, `formatPrompt`), `replayHistoryToFindState`, and `getIssueState`/tool HTTP happy-paths, but never build the `llmagent`, start the `runner`, or run the instruction templating. So every functional bug above (prompt placeholder, dropped error, panics, race) passes cleanly here.
2. **It mutates globals without restoring them** — `setupMockHTTP` sets `Owner`/`Repo`/`GitHubToken`/`httpClient.Transport`; other tests set `STALE_HOURS_THRESHOLD` and call `os.Setenv` without cleanup. Use `t.Setenv` and restore globals (or, better, the struct-based refactor removes the need). Tests are also order-dependent because `maintainersCache` is never reset between cases.

Net: the green checks here give false confidence — and per the workflow comment, they don't run in CI anyway.

PR #803, opened 2026-05-07
feat: introduce scheduler-based engine with goroutine-per-node execution

wolo-lab:

- ADK 2.O Workflow Agents: https://adk.dev/2.0/

**Problem:**
The legacy in-process BFS dispatcher in `Workflow.Run` cannot run sibling nodes
concurrently, lacks per-node timeouts and cancellation, and provides no place
to hang HITL pauses, retries, or dynamic children — all of which are upcoming
workflow milestones.

**Solution:**
Replace the BFS with a goroutine-per-node scheduler. Each scheduled node runs
in its own goroutine, pushes events into a buffered channel, and a single
consumer goroutine drains the channel, applies state-side effects, yields
events to the caller, and schedules successors when nodes complete.

Sibling cancellation on first error, per-node `Config().Timeout`, and
caller-break draining (`break` outside the range loop stops further
scheduling, not just yielding) are included. Producer-consumer model means no
mutexes are needed for scheduler state; race detector is clean.

### Testing Plan

**Unit Tests:**

- [X] I have added or updated unit tests for my change.
- [X] All unit tests pass locally.

`workflow/scheduler_test.go` covers linear chains, fan-out concurrency
(barrier-based, verifies parallelism), sibling cancellation on first error,
caller-break stops scheduling (regression), per-node timeout, and the
single-output / single-routing-event accumulator contract.

```
go test -race ./workflow/...               # ok 1.24s
go test ./...                              # all packages pass
go vet ./...                               # clean
go build ./...                             # clean
```

- [X] `examples/workflow/basic` runs end-to-end on the new engine.

### Checklist

- [X] I have added tests that prove my feature works.
- [X] Any dependent changes have been merged and published in downstream modules.

### DIFF
--- a/workflow/config.go
+++ b/workflow/config.go
@@ -28,26 +28,61 @@ var defaultRetryConfig = RetryConfig{
 	},
 }
 
+// DefaultRetryConfig returns a copy of the default retry policy
+// (5 attempts, 1s initial delay, 60s cap, 2x backoff, full jitter,
+// retry every error). Override fields on the returned value to
+// customise:
+//
+//	rc := workflow.DefaultRetryConfig()
+//	rc.MaxAttempts = 10
+//	cfg := workflow.NodeConfig{RetryConfig: rc}
 func DefaultRetryConfig() *RetryConfig {
 	cfg := defaultRetryConfig
 	return &cfg
 }
 
 // NodeConfig defines the configuration for a node.
+//
+// All fields are optional. The pointer-typed fields (RerunOnResume,
+// WaitForOutput) are tri-state: nil means "use the engine default for
+// this node kind", which mirrors Python's per-node-type defaults
+// (e.g. AgentNode in task mode defaults WaitForOutput to true while
+// other nodes default to false). Use the *Or accessor helpers to
+// read these values with an explicit per-call-site default.
 type NodeConfig struct {
-	// Enables data parallelism (runs node concurrently for each item in input collection)
+	// ParallelWorker, when true, runs the node concurrently for each
+	// item of a list-typed input. The engine collects per-item
+	// outputs and emits a single aggregate output event.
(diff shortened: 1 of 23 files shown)


workflow/config.go
```diff
@@ -59,17 +94,31 @@ type NodeConfig struct {
 // but discouraged: any unset field defaults to its zero value, not
 // to DefaultRetryConfig's value. The zero RetryConfig is a valid
 // "no retry, no backoff, no jitter" policy.
+//
+// The struct shape is deliberately a flat set of plain values: no
+// dependency on cenkalti/backoff/v5. The scheduler's retry handler
+// implements the backoff math in workflow/retry.go.
```
anFatum: Should we mention `workflow/retry.go` once it's implemented?

workflow/scheduler.go
```diff
@@ -0,0 +1,475 @@
...
+		parentCtx:   parent,
+	}
+}
+
+// buildNodesByName walks the graph's edges and returns the name→Node
+// lookup. Lets handleCompletion resolve a completion's node name to
+// its instance in O(1) instead of scanning the full table.
+func buildNodesByName(g *graph) map[string]Node {
+	nodesByName := map[string]Node{}
+	add := func(n Node) { nodesByName[n.Name()] = n }
```
anFatum: nit: let's inline that, it's a one-liner, I don't think it's necessary to wrap that into a lambda function

anFatum (approved):

hanorik (approved):

workflow/config.go
wolo-lab: You're right!

workflow/scheduler.go
wolo-lab: Good point, changed.

PR #807, opened 2026-05-08
feat: add start node validation

hanorik:
This PR introduces validation rules to ensure the integrity of the workflow graph's entry point.

Key Changes:

- `Start` Node Requirement: Validates that the workflow graph contains a `Start` node as an origin for at least one edge, ensuring the workflow can begin execution and that no edges in the graph point back to the `Start` node.
- New Error Types: Introduced `ErrNoStartNode` and `ErrNodePointsToStart` in to represent these failure cases.

Tests: Added corresponding unit tests to verify these conditions.

### DIFF
--- a/workflow/validation.go
+++ b/workflow/validation.go
@@ -23,6 +23,26 @@ import (
 // distinct Node instances that share the same Name.
 var ErrDuplicateNodeName = errors.New("duplicate node name")
 
+// ErrNoStartNode is returned when no start node is found in the edge set.
+var ErrNoStartNode = errors.New("no start node found")
+
+// ErrNodePointsToStart is returned when a node points to the start node.
+var ErrNodePointsToStart = errors.New("node points to start node")
+
+// validateNodes executes a set of edges validation checks.
+func validateNodes(edges []Edge) error {
+	if err := validateUniqueNames(edges); err != nil {
+		return err
+	}
+	if err := validateStartNodePresent(edges); err != nil {
+		return err
+	}
+	if err := validateStartNodeNoIncoming(edges); err != nil {
+		return err
+	}
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


anFatum (approved):

workflow/validation.go
```diff
@@ -23,11 +23,18 @@ import (
...
+// ErrNoStartNode is returned when no start node is found in the edge set.
+var ErrNoStartNode = errors.New("no start node found")
+
+// ErrNodePointsToStart is returned when a node points to the start node.
+var ErrNodePointsToStart = errors.New("node points to start node")
+
+// validateNodes checks that all nodes in the edge set have unique names and that the start node exists and is connected correctly.
 // If duplicate node names are found, it returns an error. The equality between
 // nodes is checked by comparing the nodes directly.
-func validateUniqueNames(edges []Edge) error {
```
wolo-lab, line 29: We could split checks into something like that so it would be more readable:

```
func validateWorkflow(edges []Edge) error {
    if err := validateUniqueNames(edges); err != nil { return err }
    if err := validateStartNodePresent(edges); err != nil { return err }
    if err := validateStartNodeNoIncoming(edges); err != nil { return err }
    return nil
}
```

wolo-lab (changes requested):

workflow/validation.go
hanorik: Devided, but added `validateNodes(edges []Edge)` instead of using `validateWorkflow()` as the last one in other PRs uses `Workflow` structure as an argument. We also could provide both `Workflow` and `[]Edges` as arguments to use for the corresponding validation function so please let me know which approach is more preferable for you.

wolo-lab (approved):

PR #808, opened 2026-05-08
feat: implement workflow validation to reject duplicate edges and remove redundant edge filtering logic

hanorik:
This PR introduces validation to ensure that a workflow graph does not contain duplicate edges (multiple edges from the same source to the exact same destination). Enforcing this constraint at workflow creation time allows us to simplify the traversal logic during execution.

Key Changes:
- `validation.go`:
  - Added `validateUniqueEdges` to check for duplicate edges.
  - Added `validateWorkflow` as a centralized entry point for workflow validation.
  - Introduced `ErrDuplicateEdge`.
- `workflow.go`:
  - Updated `New` to run the validation checks and fail if they do not pass.
  - Removed the added map from `findNextNodes` which was previously used to filter out duplicate target nodes during traversal.
- `validation_test.go`: Added unit tests to verify that duplicate edges are correctly detected.
- `workflow_test.go`: Removed a test case in `TestWorkflowRouting` that relied on duplicate edges, as this configuration is now invalid.

### DIFF
--- a/workflow/validation.go
+++ b/workflow/validation.go
@@ -23,6 +23,11 @@ import (
 // distinct Node instances that share the same Name.
 var ErrDuplicateNodeName = errors.New("duplicate node name")
 
+// ErrDuplicateEdge is returned when an edge set contains two identical edges.
+// Two edges with the same (From, To) are rejected regardless of Route; use
+// MultiRoute to express alternatives to the same target.
+var ErrDuplicateEdge = errors.New("duplicate edge")
+
 // validateUniqueNames checks that all nodes in the edge set have unique names.
 // If duplicate node names are found, it returns an error. The equality between
 // nodes is checked by comparing the nodes directly.
@@ -44,6 +49,30 @@ func validateUniqueNames(edges []Edge) error {
 		}
 		if err := checkNode(edge.To); err != nil {
 			return err
+		}
+	}
+	return nil
+}
+
+// validateWorkflow executes a set of workflow validation checks.
+func validateWorkflow(workflow *Workflow) error {
+	if err := validateUniqueEdges(workflow); err != nil {
+		return err
+	}
+	return nil
+}
+
+// validateUniqueEdges checks that there are no duplicate edges in the workflow.
+// Two edges with the same (From, To) are rejected regardless of Route; use
+// MultiRoute to express alternatives to the same target.
+func validateUniqueEdges(workflow *Workflow) error {
+	for node, edges := range workflow.edges {
+		uniqueEdges := make(map[Node]struct{})
+		for _, edge := range edges {
+			if _, ok := uniqueEdges[edge.To]; ok {
+				return fmt.Errorf("%w: from %q to %q", ErrDuplicateEdge, node.Name(), edge.To.Name())
+			}
+			uniqueEdges[edge.To] = struct{}{}
 		}
 	}
 	return nil
--- a/workflow/validation_test.go
+++ b/workflow/validation_test.go
@@ -64,3 +64,47 @@ func TestUniqueNames(t *testing.T) {
 		})
 	}
 }
+
+func TestDuplicateEdges(t *testing.T) {
+	nodeA := &dummyNode{name: "A"}
+	nodeB := &dummyNode{name: "B"}
+	tests := []struct {
+		name      string
+		edges     []Edge
+		expectErr bool
+	}{
+		{
+			name:  "no duplicate edges",
+			edges: []Edge{{From: nodeA, To: nodeB}},
+		},
+		{
+			name:      "duplicate edges",
+			edges:     []Edge{{From: nodeA, To: nodeB}, {From: nodeA, To: nodeB}},
+			expectErr: true,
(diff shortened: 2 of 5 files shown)


workflow/validation.go
```diff
@@ -48,3 +51,25 @@ func validateUniqueNames(edges []Edge) error {
...
+
+// validateWorkflow executes a set of workflow validation checks.
+func validateWorkflow(workflow *Workflow) error {
+	if err := validateUniqueEdges(workflow); err != nil {
+		return err
+	}
+	return nil
+}
+
+// validateUniqueEdges checks that there are no duplicate edges in the workflow.
```
wolo-lab, line 118: Let's add a bit more about how checks:

* two edges with the same (From, To) are rejected regardless of Route; use MultiRoute to express alternatives to the same target
* how default edges are handled

workflow/workflow.go
```diff
@@ -185,7 +185,11 @@ func New(edges []Edge) (*Workflow, error) {
 	for _, edge := range edges {
 		adj[edge.From] = append(adj[edge.From], edge)
 	}
-	return &Workflow{edges: adj}, nil
+	wf := &Workflow{edges: adj}
+	if err := validateWorkflow(wf); err != nil {
+		return nil, fmt.Errorf("workflow validation failed: %w", err)
```
wolo-lab: in 182 we have `return nil, err`. Here we can do the same, the error from validateWorkflow has already details.

workflow/workflow.go
wolo-lab: nit: it should be updated if we want to remove deduplication.

workflow/validation_test.go
```diff
@@ -64,3 +64,33 @@ func TestUniqueNames(t *testing.T) {
...
+		name      string
+		edges     []Edge
+		expectErr bool
+	}{
+		{
+			name:  "no duplicate edges",
+			edges: []Edge{{From: nodeA, To: nodeB}},
+		},
+		{
+			name:      "duplicate edges",
```
wolo-lab, line 171: Let's add some more tests:

- Two edges with the same (From, To) and different concrete Routes. Should error. 
- One unconditional + one concrete-route edge to the same To. Should error.
- MultiRoute to a single target — should pass validation and at runtime fire the target exactly once even when several emitted routes match
- Empty edge set — should not error

wolo-lab (changes requested):

workflow/workflow.go

workflow/workflow.go
hanorik: Forgot to erase this legacy part - removed.

workflow/validation_test.go
hanorik: Done. For the `MultiRoute` case added it within `TestWorkflowRouting`.

workflow/validation.go
hanorik: Done.

workflow/validation.go
```diff
@@ -48,3 +51,30 @@ func validateUniqueNames(edges []Edge) error {
 	}
 	return nil
 }
+
+// validateWorkflow executes a set of workflow validation checks.
+func validateWorkflow(workflow *Workflow) error {
```
wolo-lab: we could just pass edges/graph. We don't need anything else from the workflow.

workflow/validation.go
```diff
@@ -48,3 +51,30 @@ func validateUniqueNames(edges []Edge) error {
...
+	}
+	return nil
+}
+
+// validateUniqueEdges checks that there are no duplicate edges in the workflow.
+//
+// Behavior:
+//   - Two edges with the same (From, To) are rejected regardless of Route; use
+//     MultiRoute to express alternatives to the same target.
+//   - TODO(hanorik): to expand functionality and explain how default edges are handled.
```
wolo-lab: left over?

workflow/validation.go
```diff
@@ -48,3 +51,30 @@ func validateUniqueNames(edges []Edge) error {
...
+func validateUniqueEdges(workflow *Workflow) error {
+	for node, edges := range workflow.edges {
+		uniqueEdges := make(map[Node]struct{})
+		for _, edge := range edges {
+			if _, ok := uniqueEdges[edge.To]; ok {
+				return fmt.Errorf("%w: from %s to %s", ErrDuplicateEdge, node.Name(), edge.To.Name())
```
wolo-lab: nit: let's use %q instead of %s to have names wrapped by quotes.
someone can define the same edges with different different routes, let's add a hint in the doc of the error that MultiRoute should be used for such a use case.

wolo-lab (approved):

anFatum (approved):

workflow/validation.go
hanorik: Yes, removed.

workflow/validation.go
hanorik: Done.

workflow/validation.go
hanorik: Changed it to use `graph` instead of the `Workflow`

PR #809, opened 2026-05-08
feat: implement workflow connectivity validation to ensure all nodes are reachable from start

hanorik:
This PR introduces validation to ensure that all nodes in a workflow graph are reachable from the `Start` node. This prevents orphaned nodes that can never be executed, ensuring the graph is fully connected and valid.

Key Changes:

- `validation.go`: Added logic to traverse the workflow graph and verify that every registered node has at least one path leading to it from the start.
- `workflow.go`: Integrated the connectivity check into the workflow initialization/validation process.
- Tests: Added unit tests in `validation_test.go` to cover valid connected graphs and invalid disconnected graphs.

### DIFF
--- a/workflow/validation.go
+++ b/workflow/validation.go
@@ -25,8 +25,40 @@ import (
 // distinct Node instances that share the same Name.
 var ErrDuplicateNodeName = errors.New("duplicate node name")
 
+// ErrNoStartNode is returned when no start node is found in the edge set.
+var ErrNoStartNode = errors.New("no start node found")
+
+// ErrNodePointsToStart is returned when a node points to the start node.
+var ErrNodePointsToStart = errors.New("node points to start node")
+
+// ErrDuplicateEdge is returned when an edge set contains two identical edges.
+// Two edges with the same (From, To) are rejected regardless of Route; use
+// MultiRoute to express alternatives to the same target.
+var ErrDuplicateEdge = errors.New("duplicate edge")
+
+// ErrMultipleDefaultRoutes is returned when a node has more than one default route.
+var ErrMultipleDefaultRoutes = errors.New("node has more than one default route")
+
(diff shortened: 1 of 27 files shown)


workflow/validation.go
```diff
@@ -48,3 +52,44 @@ func validateUniqueNames(edges []Edge) error {
...
+			if !visited[neighbor.To] {
+				traverse(neighbor.To)
+			}
+		}
+	}
+
+	traverse(Start)
+
+	var unreachable []string
+	for node := range workflow.edges {
```
wolo-lab: We will miss nodes without outgoing edges:

```
edges := []Edge{
    {From: Start, To: A},
    {From: B, To: C},   // B is a key (caught), but C is a leaf and is missed
}
```

Let's add a test for that.

nit:We could sort unreachable nodes before reporting them.

wolo-lab (changes requested):

workflow/validation.go
hanorik: Done.

workflow/validation.go
```diff
@@ -48,3 +53,53 @@ func validateUniqueNames(edges []Edge) error {
...
+	var unreachable []string
+	for node := range allNodes {
+		if !visited[node] {
+			unreachable = append(unreachable, node.Name())
+		}
+	}
+	slices.Sort(unreachable)
+
+	if len(unreachable) > 0 {
+		return fmt.Errorf("%w: %s", ErrNodesNotReachable, strings.Join(unreachable, ", "))
```
wolo-lab: %q instead of %s

wolo-lab (approved):
be careful while merging with 808

anFatum (approved):

workflow/validation.go
hanorik: Done.

PR #811, opened 2026-05-08
feat: implement unconditional cycle detection in workflow validation

hanorik:
This PR introduces validation to detect unconditional cycles in the workflow graph. Cycles that consist entirely of edges without routing conditions (where `edge.Route == nil`) are now detected and reported as errors, while conditional cycles are permitted.

Key Changes:

- Unconditional Cycle Detection: Implemented `validateCycles` in `validation.go` using a Depth-First Search (DFS) algorithm that tracks the current path (stack) to identify cycles, specifically ignoring edges with non-nil routes.
- Unit Tests: Added `TestValidateCycles` in `validation_test.go` to verify behavior with no cycles, only conditional cycles, and mixed cycles.

### DIFF
--- a/workflow/validation.go
+++ b/workflow/validation.go
@@ -23,9 +23,32 @@ import (
 // distinct Node instances that share the same Name.
 var ErrDuplicateNodeName = errors.New("duplicate node name")
 
+// ErrNoStartNode is returned when no start node is found in the edge set.
+var ErrNoStartNode = errors.New("no start node found")
+
+// ErrNodePointsToStart is returned when a node points to the start node.
+var ErrNodePointsToStart = errors.New("node points to start node")
+
+// ErrMultipleDefaultRoutes is returned when a node has more than one default route.
+var ErrMultipleDefaultRoutes = errors.New("node has more than one default route")
+
 // ErrUnconditionalCycle is returned when a cycle is detected that does not
 // contain any conditional edges.
 var ErrUnconditionalCycle = errors.New("unconditional cycle detected")
+
+// validateNodes executes a set of edges validation checks.
+func validateNodes(edges []Edge) error {
+	if err := validateUniqueNames(edges); err != nil {
+		return err
+	}
+	if err := validateStartNodePresent(edges); err != nil {
+		return err
+	}
+	if err := validateStartNodeNoIncoming(edges); err != nil {
+		return err
+	}
+	return nil
+}
 
 // validateUniqueNames checks that all nodes in the edge set have unique names.
 // If duplicate node names are found, it returns an error. The equality between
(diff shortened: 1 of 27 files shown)


workflow/validation.go
```diff
@@ -48,3 +52,54 @@ func validateUniqueNames(edges []Edge) error {
...
+	if err := validateCycles(workflow); err != nil {
+		return err
+	}
+	return nil
+}
+
+// validateCycles checks that there are no unconditional cycles in the workflow.
+// It performs a depth first search for every node in the workflow and checks
+// for cycles where all edges in the cycle have nil routes.
+func validateCycles(workflow *Workflow) error {
```
wolo-lab, line 130: please document how default are handled while detecting cycles

workflow/validation.go
```diff
@@ -48,3 +52,54 @@ func validateUniqueNames(edges []Edge) error {
...
+	visited := make(map[Node]struct{})
+
+	var traverse func(n Node, inStack map[Node]struct{}) error
+	traverse = func(n Node, inStack map[Node]struct{}) error {
+		if _, ok := inStack[n]; ok {
+			return ErrUnconditionalCycle
```
wolo-lab: let's add a b.Name() to the error

wolo-lab (approved):

workflow/validation_test.go
```diff
@@ -64,3 +64,73 @@ func TestUniqueNames(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateCycles(t *testing.T) {
```
wolo-lab, line 187: please add more tests:

- diamond without cycle: A->B, A->C, B->D, C->D
- empty workflow
- graph with default: A->B, B --default--> A

wolo-lab (changes requested):

workflow/validation.go
hanorik: Do you mean the name of the node for which cycle is detected? If so, as we discussed before and according to the current algorithm I want to note that it will not be guaranteed that it will be the node where the cycle actually starts.

wolo-lab: Yes, I think it's fine, we don't need to say where the cycle starts. But at least we will inform which node appears in the cycle, so the user will be able to take it from there.

workflow/validation.go
hanorik: Done.

workflow/validation.go
hanorik: SG, added.

workflow/validation_test.go
hanorik: Done.

wolo-lab (approved):

anFatum (approved):

PR #814, opened 2026-05-11
feat: implement default route validation to prevent multiple default routes per node

hanorik:
This PR introduces validation to ensure that a node in a workflow graph does not have more than one default route.

Key Changes:

- Default Route Validation: Added `validateDefaultRoute` in `validation.go` to ensure each node has at most one outgoing edge with `Default` route. It returns a new error `ErrMultipleDefaultRoutes` if a violation is found.
- Unit Tests: Added `TestDefaultRoute` in `validation_test.go` to verify the behavior with both valid and invalid configurations.

### DIFF
--- a/workflow/validation.go
+++ b/workflow/validation.go
@@ -23,6 +23,9 @@ import (
 // distinct Node instances that share the same Name.
 var ErrDuplicateNodeName = errors.New("duplicate node name")
 
+// ErrMultipleDefaultRoutes is returned when a node has more than one default route.
+var ErrMultipleDefaultRoutes = errors.New("node has more than one default route")
+
 // validateUniqueNames checks that all nodes in the edge set have unique names.
 // If duplicate node names are found, it returns an error. The equality between
 // nodes is checked by comparing the nodes directly.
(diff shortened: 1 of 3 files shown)


workflow/validation.go
```diff
@@ -48,3 +51,26 @@ func validateUniqueNames(edges []Edge) error {
 	}
 	return nil
 }
+
+// validateWorkflow executes a set of workflow validation checks.
+func validateWorkflow(workflow *Workflow) error {
```
wolo-lab, line 96: it's not used anywhere. Please build chains from your PRs so it will be easier to merge them.
We could have something like

```golang
validateNodes(edges []Edge) error {
   if err := validateSthn(...); err != nul {return err ]
   \+ adding more checks like the new one here
}
```

workflow/validation.go
```diff
@@ -48,3 +51,26 @@ func validateUniqueNames(edges []Edge) error {
...
+	}
+	return nil
+}
+
+// validateDefaultRoute checks that there are no multiple default routes for one node.
+func validateDefaultRoute(workflow *Workflow) error {
+	for node, edges := range workflow.edges {
+		hasDefault := false
+		for _, edge := range edges {
+			if edge.Route == Default && !hasDefault {
```
wolo-lab, line 108: nit (**optional**): consider if this is more readable 

```golang
if edge.Route != Default {
  continue
}
if hasDefault {
  return fmt.Errof(...)
}
hasDefault = true
```

workflow/validation_test.go
```diff
@@ -64,3 +64,34 @@ func TestUniqueNames(t *testing.T) {
...
+			expectErr: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			if err := validateDefaultRoute(&Workflow{edges: map[Node][]Edge{nodeA: tc.edges}}); err != nil && !tc.expectErr {
+				t.Errorf("got an error %v, expected none", err)
+			} else if err == nil && tc.expectErr {
+				t.Errorf("expected an error, got none")
```
wolo-lab: maybe let's check if we got the proper error type (`errors.IS`)?

wolo-lab (changes requested):
LGTM, only minor changes.

workflow/validation_test.go
hanorik: Added.

workflow/validation.go
hanorik: Ack.

workflow/validation.go
hanorik: Ack.

wolo-lab (approved):

anFatum (approved):

PR #815, opened 2026-05-12
feat: add WithInvocationID run option

quoral:
**Problem:**
`InvocationID` was not able to be set from the callers of the ADK. This did not match the adk-python implementation. This has made us not being able to follow the resume functionality within the adk-python guide [here](https://adk.dev/runtime/resume/).

**Solution:**
This adds a RunOption that lets a caller pass an existing invocation ID into
runner.Run. When set, the invocation context uses the supplied ID
instead of generating a fresh one.

### Testing Plan

**Unit Tests:**

I have started and run our adk-go Agent with these fixes, and ensured that there are no regressions in other behaviors.

### Checklist

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -65,7 +65,8 @@ type PluginConfig struct {
 type RunOption func(*runOptions)
 
 type runOptions struct {
-	stateDelta map[string]any
+	stateDelta          map[string]any
+	invocationID  string
 }
 
 // WithStateDelta sets a state delta for the run invocation.
@@ -73,6 +74,12 @@ func WithStateDelta(delta map[string]any) RunOption {
 	return func(o *runOptions) {
 		o.stateDelta = delta
 	}
+}
+
+// WithInvocationID sets the invocation ID for the run.
+// When omitted, a new ID is generated automatically.
+func WithInvocationID(id string) RunOption {
+	return func(o *runOptions) { o.invocationID = id }
 }
 
 // New creates a new [Runner].
@@ -196,12 +203,13 @@ func (r *Runner) Run(ctx context.Context, userID, sessionID string, msg *genai.C
 		}
 
 		ctx := icontext.NewInvocationContext(ctx, icontext.InvocationContextParams{
-			Artifacts:   artifacts,
-			Memory:      memoryImpl,
-			Session:     storedSession,
-			Agent:       agentToRun,
-			UserContent: msg,
-			RunConfig:   &cfg,
+			Artifacts:    artifacts,
+			Memory:       memoryImpl,
+			Session:      storedSession,
+			Agent:        agentToRun,
+			UserContent:  msg,
+			RunConfig:    &cfg,
+			InvocationID: options.invocationID,
 		})
 		ctx, err = r.appendMessageToSession(ctx, storedSession, msg, cfg.SaveInputBlobsAsArtifacts, r.pluginManager, options.stateDelta)
 		if err != nil {
--- a/runner/runner_test.go
+++ b/runner/runner_test.go
@@ -447,3 +447,61 @@ func TestRunner_AutoCreateSession(t *testing.T) {
 		})
 	}
 }
+
+func TestWithInvocationID(t *testing.T) {
+	t.Parallel()
+
+	const wantInvocationID = "e-test-invocation-id"
+
+	var gotInvocationID string
+	testAgent := must(agent.New(agent.Config{
+		Name: "test_agent",
+		Run: func(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
+			return func(yield func(*session.Event, error) bool) {
+				ev := session.NewEvent(ctx.InvocationID())
+				ev.Author = "test_agent"
+				ev.LLMResponse = model.LLMResponse{
+					Content: &genai.Content{Role: "model", Parts: []*genai.Part{{Text: "hello"}}},
+				}
+				yield(ev, nil)
+			}
+		},
+	}))
+
+	ctx := context.Background()
+	sessionService := session.InMemoryService()
+
+	_, err := sessionService.Create(ctx, &session.CreateRequest{
+		AppName:   "app",
+		UserID:    "user",
+		SessionID: "sess",
+	})
+	if err != nil {
+		t.Fatalf("create session: %v", err)
+	}
+
+	r, err := New(Config{
+		AppName:           "app",
+		Agent:             testAgent,
+		SessionService:    sessionService,
+		AutoCreateSession: false,
+	})
+	if err != nil {
+		t.Fatalf("new runner: %v", err)
+	}
+
+	for ev, err := range r.Run(ctx, "user", "sess", nil, agent.RunConfig{},
+		WithInvocationID(wantInvocationID),
+	) {
+		if err != nil {
+			t.Fatalf("run error: %v", err)
+		}
+		if ev != nil && gotInvocationID == "" {
+			gotInvocationID = ev.InvocationID
+		}
+	}
+
+	if gotInvocationID != wantInvocationID {
+		t.Errorf("invocation ID = %q, want %q", gotInvocationID, wantInvocationID)
+	}
+}


radhus:
@mazas-google tagging you 🙏

newtonnthiga:
This is useful. Currently facing a blocker here when resuming agents after an adk tool confirmation request

runner/runner.go
```diff
@@ -65,7 +65,8 @@ type PluginConfig struct {
 type RunOption func(*runOptions)
 
 type runOptions struct {
-	stateDelta map[string]any
+	stateDelta          map[string]any
+	invocationID  string
```
karolpiotrowicz, line 69: `gofmt -l runner/` flags this file — the two field names are not aligned. `gofmt -w runner/runner.go` fixes it.

runner/runner.go
```diff
@@ -75,6 +76,12 @@ func WithStateDelta(delta map[string]any) RunOption {
 	}
 }
 
+// WithInvocationID sets the invocation ID for the run.
+// When omitted, a new ID is generated automatically.
```
karolpiotrowicz, line 80: This sentence is not accurate on the path an `LlmAgent` root takes. There, omitting the option can reuse a *prior* invocation's ID on a resume turn rather than generating a new one, and supplying the option does nothing at all. Whichever way the rebase resolves that, the comment needs to match — `WithYieldUserMessage` just above sets the precedent for naming the path a option applies to.

runner/runner.go
```diff
@@ -196,12 +203,13 @@ func (r *Runner) Run(ctx context.Context, userID, sessionID string, msg *genai.C
...
-			Agent:       agentToRun,
-			UserContent: msg,
-			RunConfig:   &cfg,
+			Artifacts:    artifacts,
+			Memory:       memoryImpl,
+			Session:      storedSession,
+			Agent:        agentToRun,
+			UserContent:  msg,
+			RunConfig:    &cfg,
+			InvocationID: options.invocationID,
```
karolpiotrowicz, line 212: This is the line the merge conflict turns on. On current `main` it reads `InvocationID: resolveInvocationID(storedSession, msg)`, and replacing that expression rather than composing with it is what breaks human-in-the-loop resume. Details and a reproduction are in the review body.

runner/runner_test.go
```diff
@@ -447,3 +447,61 @@ func TestRunner_AutoCreateSession(t *testing.T) {
...
+	t.Parallel()
+
+	const wantInvocationID = "e-test-invocation-id"
+
+	var gotInvocationID string
+	testAgent := must(agent.New(agent.Config{
+		Name: "test_agent",
+		Run: func(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
+			return func(yield func(*session.Event, error) bool) {
+				ev := session.NewEvent(ctx.InvocationID())
```
karolpiotrowicz, line 461: This will not compile against current `main`. `session.NewEvent` gained a context parameter and is now `func NewEvent(ctx context.Context, invocationID string) *Event`, so this needs `session.NewEvent(ctx, ctx.InvocationID())`.

runner/runner_test.go
```diff
@@ -447,3 +447,61 @@ func TestRunner_AutoCreateSession(t *testing.T) {
...
+				ev.Author = "test_agent"
+				ev.LLMResponse = model.LLMResponse{
+					Content: &genai.Content{Role: "model", Parts: []*genai.Part{{Text: "hello"}}},
+				}
+				yield(ev, nil)
+			}
+		},
+	}))
+
+	ctx := context.Background()
```
karolpiotrowicz, line 471: The rest of this file uses `t.Context()` — nine times — so the run is cancelled when the test ends. Worth matching.

karolpiotrowicz (commented):
The capability is worth having and the shape of the option is right, but this needs a rebase that is more than mechanical: `main` grew its own invocation-ID resolution after this branch was written, and the two collide at exactly the line this PR changes.

## What blocks

**Applied to `main` as it stands, this stops a paused human-in-the-loop turn from resuming.** When the branch was written, the invocation context in `Run` carried no `InvocationID` field at all, so setting it was purely additive. `main` now computes one at [`runner/runner.go:681`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/runner/runner.go#L681) via [`resolveInvocationID`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/runner/run_node.go#L433-L446), which reuses the paused run's ID when the incoming message answers a pending tool call. Replacing that expression with `options.invocationID` means a caller who passes no option gets `""`, a fresh `e-<uuid>` is minted at [`internal/context/invocation_context.go:47-48`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/internal/context/invocation_context.go#L47-L48), and rehydration at [`workflow/persistence.go:141`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/workflow/persistence.go#L141) then matches none of the paused run's events. The workflow raises a fresh interrupt instead of resuming. Seven integration tests across three packages fail this way on the merge and pass on `main`.

<details><summary>Reproduction and output</summary>

```
git fetch origin pull/815/head
git merge FETCH_HEAD          # resolve the two conflicts in runner/, keeping the PR's InvocationID line
                              # also change session.NewEvent(ctx.InvocationID()) to
                              # session.NewEvent(ctx, ctx.InvocationID()) in the new test
go test ./runner/... ./cmd/launcher/console/... ./server/adkrest/...
```

```
--- FAIL: TestRunner_WorkflowHITL_Roundtrip_Handoff
    hitl_integration_test.go:88: turn 2 unexpectedly produced another interrupt
    hitl_integration_test.go:91: handler input = <nil>, want "approve"
--- FAIL: TestRunner_WorkflowHITL_Roundtrip_ReEntry
    hitl_integration_test.go:133: handler input = <nil>, want "yes"
--- FAIL: TestRunner_WorkflowHITL_FunctionResponseRoutedByID
    hitl_integration_test.go:180: turn 2 produced a fresh interrupt instead of resuming
--- FAIL: TestRunner_WorkflowHITL_DynamicOrchestrator_Resume
    hitl_integration_test.go:329: parent output = <nil>, want "X:draft|Y:yes"
--- FAIL: TestRunner_WorkflowHITL_TwoFullCycles_SameSession
    hitl_integration_test.go:362: handler input = <nil>, want "first"
FAIL	google.golang.org/adk/v2/runner
--- FAIL: TestConsoleHITL_TwoFullCycles_SameSession
    hitl_test.go:517: cycle 1 greeting = <nil>, want "Hello, Wojtek!"
FAIL	google.golang.org/adk/v2/cmd/launcher/console
--- FAIL: TestRESTHITL_TwoFullCycles_SameSession
    hitl_integration_test.go:53: cycle 1 greeting = "", want "Hello, Wojtek!"
FAIL	google.golang.org/adk/v2/server/adkrest
```

</details>

The fix is not obvious, because it needs a precedence decision rather than a re-application of the patch. adk-python already made that decision: [`_resolve_invocation_id`](https://github.com/google/adk-python/blob/c506ddf3bc34a6312ffc81899221bfa3f2da3b1d/src/google/adk/runners.py#L500-L533) takes the caller's `invocation_id` as a third argument, returns it when there is no answered function response, and otherwise prefers the ID resolved from history while logging a warning that the caller's value was ignored. Giving the Go `resolveInvocationID` a third `provided string` parameter returned on its two early-exit paths, and calling it as `resolveInvocationID(storedSession, msg, options.invocationID)`, matches that and turns the suite green again — I ran it, and the only remaining failure was an unrelated environmental one that also fails on `main`.

**The option does nothing on the path an `LlmAgent` root takes.** [`runner/runner.go:624`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/runner/runner.go#L624) dispatches those roots to `runNode` and returns before reaching the line this PR patches. `runNode` reads `opts.stateDelta` and `opts.yieldUserMessage` but never `opts.invocationID`, and [`newNodeInvocationContext`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/runner/run_node.go#L270-L311) has no parameter to carry one. So an `LlmAgent` root sets an ID, gets no error back, and its events carry a generated one instead. Threading the value through that helper is the same change as the fix above.

## Worth doing in the same pass

**The invocation ID is a scope key, and the doc comment does not say so.** Rehydration scoping, the dynamic scheduler's output cache at [`workflow/dynamic_scheduler.go:173`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/workflow/dynamic_scheduler.go#L173), and the compaction window at [`internal/compactioninternal/window.go:283-311`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/internal/compactioninternal/window.go#L283-L311) all key on it. A caller who reads "sets the invocation ID for the run" and pins one correlation ID across a session gets stale rehydration, a child node served from cache instead of re-run, and compaction that never fires — the window counts distinct IDs, so N turns under one ID collapse to a single entry and the interval threshold is never reached. One sentence recommending a per-run-unique value covers it. This is a documentation gap rather than a behavior change, and adk-python has the same property, so please do not add a validator here.

**Two comments in the tree assert the property the change removes.** [`workflow/persistence.go:76-78`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/workflow/persistence.go#L76-L78) says "The runner reuses the paused run's invocation ID for the resume turn, so the pause and its reply share it", and [`agent/workflowagent/workflow.go:145-148`](https://github.com/google/adk-go/blob/e9da66863f5eeceea884856df07f8f8d985b71f8/agent/workflowagent/workflow.go#L145-L148) says the same. Both become true again under the fix above with no edit, so they are a useful check that the rebase landed correctly.

## Context that may change what you want to build

The blocker described in this thread — resuming after a tool confirmation request — is now handled on `main` without any caller involvement. `resolveInvocationID` was added in the v2 release and reuses the paused invocation's ID automatically whenever the incoming message carries a function response answering a pending call. If that was the whole use case, it may already work for you on current `main`, and it is worth checking before you spend time on the rebase.

That does not make the option redundant. adk-python exposes [`invocation_id` on `run_async`](https://github.com/google/adk-python/blob/c506ddf3bc34a6312ffc81899221bfa3f2da3b1d/src/google/adk/runners.py#L974) and adk-go has no equivalent, so explicit caller control is still a real gap. It does mean the two mechanisms have to coexist rather than one replacing the other.

## On the test

`TestWithInvocationID` does discriminate — removing the wiring makes it fail — but it passes a `nil` message to a non-`LlmAgent` root, which is the one case where the current design and the correct design produce the same answer. That is why it stayed green while the seven tests above went red. A resume turn asserting that the paused ID is reused when the option is omitted, and a second one asserting the supplied ID is honored on a fresh turn, would both catch the problems above.

runner/runner_test.go
```diff
@@ -447,3 +447,61 @@ func TestRunner_AutoCreateSession(t *testing.T) {
...
+		AppName:           "app",
+		Agent:             testAgent,
+		SessionService:    sessionService,
+		AutoCreateSession: false,
+	})
+	if err != nil {
+		t.Fatalf("new runner: %v", err)
+	}
+
+	for ev, err := range r.Run(ctx, "user", "sess", nil, agent.RunConfig{},
```
karolpiotrowicz, line 493: Passing a `nil` message is what limits this test's reach. With no message there is nothing for the runner to resolve an existing invocation ID from, so the option's value and the value `main` would have computed are identical, and the test cannot tell a correct wiring from one that overrides. Giving it a real message and adding a resume turn would make it discriminate.

PR #824, opened 2026-05-13
fix: add validation for transfer agent

anFatum:
Adds the NewTransferToAgentTool constructor that updates the declaration for transfer function to limit the agent_name with enum that consumes only agents that agent could transfer to


internal/llminternal/agent_transfer.go
```diff
@@ -82,21 +83,37 @@ func AgentTransferRequestProcessor(ctx agent.InvocationContext, req *model.LLMRe
...
 
-type TransferToAgentTool struct{}
+// TransferToAgentTool is a tool that handles transferring control to another agent.
+type TransferToAgentTool struct {
+	instructions    string
+	supportedAgents []agent.Agent
+}
+
+// NewTransferToAgentTool creates a new TransferToAgentTool.
+func NewTransferToAgentTool(agent, parent agent.Agent, targets []agent.Agent) (*TransferToAgentTool, error) {
```
wolo-lab: nit: shadows imported agent, could be renamed

internal/llminternal/agent_transfer.go
```diff
@@ -82,21 +83,37 @@ func AgentTransferRequestProcessor(ctx agent.InvocationContext, req *model.LLMRe
...
+	t := &TransferToAgentTool{
+		supportedAgents: targets,
+	}
+	si, err := t.instructionsForTransferToAgent(agent, parent, targets)
```
wolo-lab: nit (**optional**): we could set instructions inside constructor

```golang
si, err := t.instructionsForTransferToAgent(agent, parent, targets)
if err != nil {
		return nil, err
	}
return &TransferToAgentTool{
		supportedAgents: targets,
         instructions: si
	}, nil
```

wolo-lab (dismissed):

wolo-lab (approved):

PR #826, opened 2026-05-13
feat: enable h2c on web launcher listener

### LINKED ISSUE #825: Enable cleartext HTTP/2 on the web launcher listener
## Summary

The current `cmd/launcher/web` server starts a single `net/http` listener for ADK web sublaunchers, but it does not explicitly enable cleartext HTTP/2. This makes it difficult for users to build custom web sublaunchers that serve native gRPC traffic on the same launcher port.

## Problem

`web.Sublauncher` is a public extension point:

```go
type Sublauncher interface {
	SetupSubrouters(router *mux.Router, config *launcher.Config) error
}
```

This allows users to register custom handlers on the launcher router. In principle, a custom sublauncher can mount a `grpc.Server` or another HTTP/2-aware handler. However, the launcher currently constructs `http.Server` without setting `Server.Protocols`, so the listener is not explicitly configured to accept unencrypted HTTP/2 traffic.

As a result, users who want to serve REST or Web UI routes and local h2c/gRPC routes from the same launcher have to work around the default web launcher or run a separate server.

## Proposed change

Configure the existing `http.Server` in `cmd/launcher/web/web.go` with standard library protocol support:

```go
protocols := new(http.Protocols)
protocols.SetHTTP1(true)
protocols.SetUnencryptedHTTP2(true)

srv := http.Server{
	...
	Protocols: protocols,
}
```

This keeps HTTP/1 enabled for existing REST, Web UI, A2A, and trigger routes while also allowing cleartext HTTP/2 on the same listener.

## Benefits

- Existing launcher behaviour remains compatible for HTTP/1 clients.
- Users can build custom web sublaunchers that register h2c/gRPC-capable handlers.
- The change uses the standard library `net/http.Protocols` support already available to the module.
- The launcher continues to use one listener and one port.

jankrynauw:
# Enable h2c on the web launcher listener

Fixes #825 

## Summary

This PR configures the `cmd/launcher/web` HTTP server to accept both HTTP/1 and cleartext HTTP/2 on the same listener.

The web launcher already exposes `web.Sublauncher` as a public extension point, allowing users to register custom handlers on the shared router. Enabling h2c at the listener level lets those custom sublaunchers mount HTTP/2-aware handlers, including native gRPC servers, without changing the public launcher APIs or requiring a separate server.

## Changes

- Creates an `http.Protocols` value in `webLauncher.Run`.
- Enables HTTP/1 with `SetHTTP1(true)`.
- Enables cleartext HTTP/2 with `SetUnencryptedHTTP2(true)`.
- Assigns the configured protocols to `http.Server.Protocols`.
- Adds a regression test confirming existing HTTP/1 routing through `BuildBaseRouter` still works.

## Notes

This intentionally does not add a first-class gRPC registration API. Since `web.Sublauncher` is already public, users can build custom sublaunchers that register their own gRPC-capable handlers. This PR only enables the shared listener to accept the required HTTP/2 transport.

The change also avoids adding `golang.org/x/net/http2/h2c`; it uses the standard library `net/http.Protocols` support available to this module.

## Testing

```sh
go test ./cmd/launcher/web
go test ./cmd/launcher/web/...
go test ./cmd/launcher/...
```


indurireddy-TF:
Hi @baptmont @kdroste-google  could you please review this PR? I have validated it on my end and all tests are working perfectly.

cmd/launcher/web/web.go
```diff
@@ -182,12 +182,21 @@ func (w *webLauncher) Run(ctx context.Context, config *launcher.Config) error {
 	}
 	log.Println()
 
+	// Enable both HTTP/1 and cleartext HTTP/2 on the same listener. Existing
+	// REST, Web UI, A2A, and trigger routes continue to work over HTTP/1.1,
+	// while custom web sublaunchers can register handlers that require h2c,
+	// such as native gRPC servers.
+	protocols := new(http.Protocols)
```
kdroste-google: Hi, @jankrynauw, 
The need for that is clear to me. Could you please make it more on-damand? 
I mean a new flag for command line `--h2c`, which would set webConfig.useH2C to true. And setting the protocol conditionally, only if the flag is set.

kdroste-google (changes requested):

cmd/launcher/web/web.go
jankrynauw: Agreed, implemented conditional logic using using a command line flag

kdroste-google (dismissed):
Looks good to me, thank you!

karolpiotrowicz:
Thanks for this, Jan — nice, focused change. I checked it out locally and the
mechanism looks correct: it's opt-in (`--h2c`, default `false`), doesn't change
the public API, adds no new dependency, and with the flag set a single listener
serves both HTTP/1.1 and cleartext HTTP/2 (I confirmed both an HTTP/1.1 client
and an h2c prior-knowledge client are served on the same port, and that HTTP/1.1
stays the only protocol when the flag is off).

One request before merge, plus a couple of small doc nits.

**Could we add a test for the h2c behavior?** The added
`TestBuildBaseRouterHTTP1RouteStillWorks` exercises `BuildBaseRouter()`, which
this PR doesn't modify — it passes even with the h2c changes reverted, so `Run()`
and `NewLauncher()` (where the flag and the `srv.Protocols` logic live) currently
have no test coverage. Since serving h2c is the whole point of the change, it's
worth locking in against regressions. It can be covered with the standard library
only, for example:
- with `--h2c` set: assert that **both** an HTTP/1.1 client and an h2c
  prior-knowledge client (an `http.Transport` with
  `Protocols.SetUnencryptedHTTP2(true)`) succeed against the listener;
- with `--h2c` unset: assert the h2c client is rejected / falls back to HTTP/1.1.

Happy to help with the test if that's useful.

**Doc nits (non-blocking):**
- The comment mentions "native gRPC servers." net/http's h2c serves
  handler-style HTTP/2, so connect-go / grpc-web work directly, while grpc-go's
  own `grpc.Server` goes through a separate, experimental `ServeHTTP` path —
  might be worth softening the wording so it doesn't set the wrong expectation.
- net/http's unencrypted HTTP/2 is prior-knowledge only (it doesn't perform the
  HTTP/1.1 `Upgrade: h2c` handshake). Worth noting that in the flag help, along
  with a short "cleartext — don't expose to untrusted networks" caveat.
- Streaming responses are still bounded by `--write-timeout` (15s default), so
  anyone enabling `--h2c` for long-lived streams will want to raise it.

Thanks again — happy to re-review once the test is in.

jankrynauw:
Thanks for the detailed review — addressed.

I replaced the router-only test with a listener-level test covering both configurations:
- with --h2c: HTTP/1.1 and prior-knowledge h2c requests both succeed;
- without --h2c: HTTP/1.1 succeeds and the prior-knowledge h2c request is rejected.
 
I extracted server construction into buildHTTPServer() so the test exercises the same configuration used by Run() while using a test-owned ephemeral listener.

I also updated the flag help and comments to clarify prior-knowledge-only behavior, the cleartext security caveat, --write-timeout for long-lived streams, and the handler-style HTTP/2 use case.

The launcher tests and race detector pass. Thanks again — ready for re-review.

karolpiotrowicz (approved):
Confirmed additive: with `--h2c` unset the constructed server is field-for-field the
previous literal and `Protocols` stays nil, so HTTP/1.1 behavior is unchanged. No
exported API is added or altered, and it stays on the standard library. The test is
load-bearing in both directions — it fails if the conditional is removed, and it fails
if the flag default is flipped to true — which was the gap in the earlier revision.

All the earlier review points are resolved: the opt-in flag, the h2c coverage for both
flag states, the handler-style HTTP/2 wording, and the prior-knowledge and
`--write-timeout` notes in the flag help.

One operational note, not a blocker and not something to change here. On Go 1.26.5,
`conn.serve` calls `maybeServeUnencryptedHTTP2` before the read loop that arms a
deadline, and that preface peek blocks with no deadline of its own (the
`SetReadDeadline` calls on that path sit in the TLS branch). So once `--h2c` is on, a
connection that sends fewer than 14 bytes holds a goroutine indefinitely, and neither
`--read-timeout`, nor `ReadHeaderTimeout`, nor `--idle-timeout` reaps it — including for
plain HTTP/1.1 clients. That is upstream behavior rather than anything this PR
introduces, and it is inert while the flag is off, but it does mean the existing "do not
expose it to untrusted networks" warning carries more weight than confidentiality alone.
I have opened #1224 to track a listener-level mitigation.

Two optional nits if you end up pushing a rebase anyway (the branch is behind main):
asserting on the rejection error rather than bare `err != nil` in the disabled branch
would distinguish a server rejection from an unreachable port, and a package-local
`http.Client` would avoid depending on `http.DefaultClient`.

Approving.

PR #833, opened 2026-05-14
feat(live): Add core bidirectional streaming support

baptmont:
Adds the core files for live agent execution, including RunLive implementation, base flow updates, and session management.


server/adkrest/controllers/runtime.go
```diff
@@ -237,3 +242,137 @@ func decodeRequestBody(req *http.Request) (decodedReq models.RunAgentRequest, er
...
+				break
+			}
+		}
+	}
+
+	r, _, err := c.getRunner(models.RunAgentRequest{AppName: appName, UserId: userID, SessionId: sessionID})
+	if err != nil {
+		closeReason := err.Error()
+		if _, loadErr := c.agentLoader.LoadAgent(appName); loadErr != nil {
+			closeReason = fmt.Sprintf("agent %s not found", appName)
```
dpasiukevich: nit: maybe include the original closeReason here? something like "agent %s not found for original error: %w"

server/adkrest/controllers/runtime.go
```diff
@@ -237,3 +242,137 @@ func decodeRequestBody(req *http.Request) (decodedReq models.RunAgentRequest, er
 	}
 	return runAgentRequest, nil
 }
+
+func (c *RuntimeAPIController) RunLiveHandler(rw http.ResponseWriter, req *http.Request) error {
+	upgrader := websocket.Upgrader{
+		ReadBufferSize:  1024,
+		WriteBufferSize: 1024,
+		CheckOrigin:     func(r *http.Request) bool { return true },
```
dpasiukevich: is it possible to make it more secure? What are the consequences of the current code which will accept every origin header?

server/adkrest/controllers/runtime.go
```diff
@@ -237,3 +242,137 @@ func decodeRequestBody(req *http.Request) (decodedReq models.RunAgentRequest, er
...
+	}
+
+	r, _, err := c.getRunner(models.RunAgentRequest{AppName: appName, UserId: userID, SessionId: sessionID})
+	if err != nil {
+		closeReason := err.Error()
+		if _, loadErr := c.agentLoader.LoadAgent(appName); loadErr != nil {
+			closeReason = fmt.Sprintf("agent %s not found", appName)
+		}
+		sendClose(websocket.CloseInternalServerErr, closeReason)
+		return nil
```
dpasiukevich, line 296: do we need to log/return the error here?

server/adkrest/controllers/runtime.go
```diff
@@ -237,3 +242,137 @@ func decodeRequestBody(req *http.Request) (decodedReq models.RunAgentRequest, er
...
+	// Read from Runner and write back to client over the WebSocket
+	liveSession, eventIter, err := r.RunLive(req.Context(), userID, sessionID, agent.LiveRunConfig{
+		MaxLLMCalls:              100, // Reasonable default
+		ResponseModalities:       []genai.Modality{genai.ModalityAudio},
+		InputAudioTranscription:  &genai.AudioTranscriptionConfig{},
+		OutputAudioTranscription: &genai.AudioTranscriptionConfig{},
+	})
+	if err != nil {
+		sendClose(websocket.CloseInternalServerErr, err.Error())
+		return nil
```
dpasiukevich, line 309: do we need to log/return the error here?

server/adkrest/controllers/runtime.go
```diff
@@ -237,3 +242,137 @@ func decodeRequestBody(req *http.Request) (decodedReq models.RunAgentRequest, er
...
+	defer func() {
+		_ = liveSession.Close()
+	}()
+
+	// Spawning goroutine for reading from the client over WebSocket and pushing it to Runner
+	go func() {
+		for {
+			messageType, p, err := ws.ReadMessage()
+			if err != nil {
+				// WebSocket is closed or error occurred
```
dpasiukevich: Should we log the errors here and in other err paths in this goroutine? To make sure the people who run the server know about the errors. Or is it ok to skip it here?

server/adkrest/controllers/runtime.go
```diff
@@ -237,3 +242,137 @@ func decodeRequestBody(req *http.Request) (decodedReq models.RunAgentRequest, er
...
+		sendClose(websocket.CloseInternalServerErr, err.Error())
+		return nil
+	}
+	defer func() {
+		_ = liveSession.Close()
+	}()
+
+	// Spawning goroutine for reading from the client over WebSocket and pushing it to Runner
+	go func() {
+		for {
```
dpasiukevich, line 320: could you check if we need to add `defer liveSession.Close()`? in this goroutine in addition to line311

for example when client closes, then ws.ReadMessage() will return the error, and this goroutine will exit, while the main goroutine will hand on ws.WriteMessage

internal/llminternal/googlellm/live_connection.go
```diff
@@ -0,0 +1,298 @@
...
+		var filteredParts []*genai.Part
+		for _, part := range content.Parts {
+			if part == nil {
+				continue
+			}
+			if part.InlineData != nil && strings.HasPrefix(part.InlineData.MIMEType, "audio/") {
+				continue
+			}
+			filteredParts = append(filteredParts, part)
+			fmt.Printf("filtered part: %v\n", part.Text)
```
dpasiukevich: do we need this ftm.Printf? If yes, should we better use logger (here and in other places in the file).

server/adkrest/controllers/runtime.go
baptmont: done

server/adkrest/controllers/runtime.go
baptmont: removed the custom func and left default that validates same origin. 
In the future we can add a allow list to the params and make a custom func for that.

server/adkrest/controllers/runtime.go
baptmont: sendClose is sending the err via the websocket but its always good to have logs so added logging

server/adkrest/controllers/runtime.go
baptmont: added logging

server/adkrest/controllers/runtime.go
baptmont: added logging

internal/llminternal/googlellm/live_connection.go
baptmont: No, I just removed it

dpasiukevich (approved):

PR #836, opened 2026-05-14
feat(live): Add streaming tools

baptmont:
### **Overview**
Introduces support for asynchronous **Streaming Tools** within the bidirectional live-streaming (`RunLive`) architecture. This allows tools to execute and yield chunks of data progressively back to the active live session over time, rather than blocking the execution loop to return a single, final static result.

### **Key Changes**
*   **`tool/functiontool/streaming_function.go`**:
    *   Adds the `NewStreaming[TArgs]` constructor and `streamingFunctionTool[TArgs]` struct to represent functions yielding a Go iterator (`iter.Seq2[string, error]`).
    *   Integrates request-confirmation workflows via `RequireConfirmation` and dynamic `RequireConfirmationProvider` callbacks.
    *   **Concurrent Execution**: Invokes streaming tools in a background goroutine (under a cancellable context) so that tool execution does not block the main live-event loop.
    *   **Progression Delivery**: Listens to the iterator and immediately packages each yielded text chunk as an intermediate user content message (`"Function <Name> returned: <Chunk>"`) streamed directly back to the client via `liveSession.Send(...)`.
    *   **Cancellation Control**: Automatically registers running tools with the session (`RegisterStreamingTool`) and introduces a built-in `stop_streaming` system tool to allow client-side or agent-triggered cancellation of active tool instances.


examples/bidi/streaming_tool/main.go
```diff
@@ -0,0 +1,149 @@
...
+		Instruction: "You are a helpful assistant with a streaming tool 'count_to'. Always use it when asked to count. Wait for the tool results, and when you recieve them you should say the number, if it is divisible by 3 you should not say the number and instead say Fizz and if it is divisible by 5 you should say Buzz, if it is divisible by both 3 and 5 you should say FizzBuzz. Always use the check_divisible tool",
+		Tools:       []tool.Tool{counterTool, stopTool, checkDivisibleTool},
+	})
+	if err != nil {
+		log.Fatalf("Failed to create agent: %v", err)
+	}
+
+	uiMode := true
+
+	if uiMode {
```
dpasiukevich: let's remove this if

examples/bidi/streaming_tool/main.go
```diff
@@ -0,0 +1,149 @@
...
+	"google.golang.org/adk/tool"
+	"google.golang.org/adk/tool/functiontool"
+)
+
+func main() {
+	log.SetOutput(os.Stdout)
+	ctx := context.Background()
+
+	// gemini-3.1-flash-live-preview
+	// gemini-2.5-flash-native-audio-preview-12-2025
```
dpasiukevich: Let's remove these model names. They can become outdated pretty soon I assume.

If you'd like, we can put some human text here, describing which model types to use.

dpasiukevich (dismissed):

examples/bidi/streaming_tool/main.go
baptmont: removed it

examples/bidi/streaming_tool/main.go
baptmont: removed it

dpasiukevich (approved):

PR #837, opened 2026-05-14
feat(live): Add session resumption

baptmont:
### **Overview**
Enables **Session Resumption** for bidirectional streaming (`RunLive`) sessions. This allows real-time WebSocket connections to seamlessly recover their state, conversation context, and history following unexpected dropouts, retries, or network reconnections.

### **Key Changes**
*   **`internal/context/invocation_context.go`**:
    *   Adds `LiveSessionResumptionHandle` to the `InvocationContext` properties, providing thread-safe getters and setters (`LiveSessionResumptionHandle()` / `SetLiveSessionResumptionHandle()`) to persist the handle across connection attempts.
    *   **Handle Capture**: Intercepts `SessionResumptionHandle` from incoming server responses and stores it back inside the active `InvocationContext`.
    *   **Resumption Re-handshake**: When initiating or restoring the live connection, if a handle exists in the context, it configures `genai.SessionResumptionConfig` with the handle to re-bind to the existing host-side session.
    *   **Vertex AI Integration**: Supports transparent resumption configuration (`SessionResumption.Transparent = true`) when executing against the Vertex AI backend.


internal/context/invocation_context.go
```diff
@@ -94,6 +95,14 @@ func (c *InvocationContext) Ended() bool {
 	return c.params.EndInvocation
 }
 
+func (c *InvocationContext) LiveSessionResumptionHandle() string {
```
dpasiukevich, line 100: nit: could you add method comments for this method, so users know what and how to use it

internal/context/invocation_context.go
```diff
@@ -94,6 +95,14 @@ func (c *InvocationContext) Ended() bool {
...
+	return c.params.LiveSessionResumptionHandle
+}
+
+func (c *InvocationContext) SetLiveSessionResumptionHandle(handle string) {
```
dpasiukevich, line 106: nit: could you add method comments for this method, so users know what and how to use it

internal/llminternal/base_flow.go
```diff
@@ -257,10 +257,23 @@ func (f *Flow) RunLive(ctx agent.InvocationContext) (agent.LiveSession, iter.Seq
 		}
 
 		for {
+			if iCtx, ok := ctx.(*icontext.InvocationContext); ok {
```
dpasiukevich: is it possible to move out the type cast out of the loop? So it's done only once.

dpasiukevich (dismissed):

dpasiukevich (approved):

PR #838, opened 2026-05-14
feat(live): Add audio cache for save artifact

baptmont:
### **Overview**
Introduces an **Audio Cache Manager** to collect, aggregate, and persist real-time voice conversation chunks (both user voice inputs and model voice outputs) into the session's permanent artifact storage, recording them as events in the session's trace history.

### **Key Changes**
*   **`internal/llminternal/audio_cache_manager.go`**:
    *   Implements a thread-safe cache (`inputCache` and `outputCache`) that accumulates raw binary audio chunks along with their starting timestamps.
    *   `FlushCaches()` merges successive cached chunks, constructs a unique timestamped filename (e.g., `.pcm` or `.mp3`), saves the combined file using the invocation's `Artifacts().Save()` service, and returns a `session.Event` referencing the newly saved file URI (`artifact://...`).
    *   Integrates the audio manager into `liveSessionImpl` and respects `runCfg.Live.SaveLiveBlob`.
    *   Caches incoming user voice packets (`CacheInput`) and outgoing model voice pieces (`CacheOutput`).
    *   Flushes caches at logical speech-turn boundaries (e.g., when the model is `Interrupted` or the turn completes with `TurnComplete`), recording the final structured audio artifacts directly into the persistent session history.


internal/llminternal/base_flow.go
```diff
@@ -327,6 +336,9 @@ func (f *Flow) RunLive(ctx agent.InvocationContext) (agent.LiveSession, iter.Seq
 							}
 						}
 						if req.RealtimeInput != nil {
+							if blob, ok := req.RealtimeInput.(*genai.Blob); ok {
```
hanorik, line 411: I think it's also makes sense to add "audio/" prefix check for the `blob.MIMEType.

internal/llminternal/audio_cache_manager_test.go
hanorik, line 1: Please add a test case with different MIME types for input and output caches. Also consider to make `FlushCaches` table-driven as some of them are pretty similar.

internal/llminternal/base_flow.go
baptmont: Added validation to CacheInput/ CacheOutput

internal/llminternal/audio_cache_manager_test.go
baptmont: Added to AudioCacheManager
```
	inputMimeType   string
	outputMimeType  string
```

hanorik (approved):
LGTM.

PR #839, opened 2026-05-14
Changed model for gemini API to gemini-3.1-flash-lite (in examples)

kdroste-google:
For examples changed gemini API model to gemini-3.1-flash-lite


wolo-lab (approved):
Thx!

PR #840, opened 2026-05-14
feat: implement AgentNode

hanorik:
This PR introduces the `AgentNode` component in the `workflow` package, enabling standard `agent.Agent` instances to be used as nodes within a workflow graph. It supports type-safe input/output handling and schema validation.

**Key Changes:**
- **`AgentNode` Implementation**: Added [AgentNode](file:///usr/local/google/home/hanorik/adk-go/workflow/agent_node.go) which wraps `agent.Agent` and implements the `Node` interface.
- **Factory Methods**:
  - `NewAgentNode`: Creates a node with `any` input/output types.
  - `NewAgentNodeTyped[Input, Output]`: Uses generics to infer input/output schemas automatically.
  - `NewAgentNodeWithSchemas`: Allows passing explicit `jsonschema.Schema` definitions.
- **Execution & Validation**: The `Run` method handles input conversion to `genai.Content` and invokes the agent using `internalcontext.NewInvocationContext`.
- **Comprehensive Tests**: Added [agent_node_test.go](file:///usr/local/google/home/hanorik/adk-go/workflow/agent_node_test.go) covering various execution scenarios.

### DIFF
--- a/workflow/agent_node.go
+++ b/workflow/agent_node.go
@@ -1,0 +1,149 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package workflow
+
+import (
+	"fmt"
+	"iter"
+
+	"github.com/google/jsonschema-go/jsonschema"
+	"google.golang.org/genai"
+
+	"google.golang.org/adk/agent"
+	internalcontext "google.golang.org/adk/internal/context"
+	"google.golang.org/adk/internal/typeutil"
+	"google.golang.org/adk/session"
+)
+
+// AgentNode wraps a standard agent.Agent.
+type AgentNode struct {
+	BaseNode
+	agent        agent.Agent
+	inputSchema  *jsonschema.Resolved
+	outputSchema *jsonschema.Resolved
+}
+
+// newAgentNodeWithSchemasTyped creates a new node wrapping an agent with explicitly provided schemas.
+// If a schema is nil, it will be inferred from the corresponding generic type Input or Output.
+func newAgentNodeWithSchemasTyped[Input, Output any](a agent.Agent, inputSchema, outputSchema *jsonschema.Schema, cfg NodeConfig) (Node, error) {
+	if a == nil {
+		return nil, fmt.Errorf("agent cannot be nil")
+	}
+	ischema, err := resolvedSchema[Input](inputSchema)
+	if err != nil {
+		return nil, fmt.Errorf("resolving input schema for agent %q: %w", a.Name(), err)
+	}
+	if ischema == nil {
+		return nil, fmt.Errorf("resolved input schema for agent %q is nil", a.Name())
+	}
+	oschema, err := resolvedSchema[Output](outputSchema)
+	if err != nil {
+		return nil, fmt.Errorf("resolving output schema for agent %q: %w", a.Name(), err)
+	}
+	if oschema == nil {
+		return nil, fmt.Errorf("resolved output schema for agent %q is nil", a.Name())
+	}
+
+	return &AgentNode{
+		BaseNode:     NewBaseNode(a.Name(), a.Description(), cfg),
+		agent:        a,
+		inputSchema:  ischema,
+		outputSchema: oschema,
+	}, nil
+}
+
+// NewAgentNodeWithSchemas is a convenience wrapper for NewAgentNodeWithSchemasTyped[any, any].
+// It uses explicitly provided schemas for both input and output.
+func NewAgentNodeWithSchemas(a agent.Agent, inputSchema, outputSchema *jsonschema.Schema, cfg NodeConfig) (Node, error) {
+	return newAgentNodeWithSchemasTyped[any, any](a, inputSchema, outputSchema, cfg)
+}
+
+// NewAgentNodeTyped creates a new node wrapping an agent using generics to
+// automatically infer input and output schemas from the provided types.
+func NewAgentNodeTyped[Input, Output any](a agent.Agent, cfg NodeConfig) (Node, error) {
+	return newAgentNodeWithSchemasTyped[Input, Output](a, nil, nil, cfg)
+}
+
+// NewAgentNode creates a new node wrapping an agent. Input and output schemas
+// are inferred as `any`.
+func NewAgentNode(a agent.Agent, cfg NodeConfig) (Node, error) {
+	return NewAgentNodeTyped[any, any](a, cfg)
+}
+
+// Run implements the Node interface.
+func (n *AgentNode) Run(ctx agent.InvocationContext, input any) iter.Seq2[*session.Event, error] {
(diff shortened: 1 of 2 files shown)


workflow/agent_node.go
```diff
@@ -0,0 +1,149 @@
...
+			switch v := input.(type) {
+			case string:
+				userContent = &genai.Content{
+					Parts: []*genai.Part{{Text: v}},
+				}
+			case *genai.Content:
+				userContent = v
+			default:
+				userContent = &genai.Content{
+					Parts: []*genai.Part{{Text: fmt.Sprint(v)}},
```
wolo-lab: use json.Marshal

workflow/agent_node.go
```diff
@@ -0,0 +1,149 @@
...
+		BaseNode:     NewBaseNode(a.Name(), a.Description(), cfg),
+		agent:        a,
+		inputSchema:  ischema,
+		outputSchema: oschema,
+	}, nil
+}
+
+// NewAgentNodeWithSchemas is a convenience wrapper for NewAgentNodeWithSchemasTyped[any, any].
+// It uses explicitly provided schemas for both input and output.
+func NewAgentNodeWithSchemas(a agent.Agent, inputSchema, outputSchema *jsonschema.Schema, cfg NodeConfig) (Node, error) {
```
wolo-lab: return concrete types
reference:
- https://go.dev/wiki/CodeReviewComments#interfaces):

workflow/agent_node_test.go
```diff
@@ -0,0 +1,417 @@
...
+		t.Fatalf("jsonschema.For[Output] failed: %v", err)
+	}
+
+	tests := []struct {
+		name    string
+		creator func() (Node, error)
+		want    string
+	}{
+		{
+			name: "NewToolNodeTyped",
```
wolo-lab: "NewToolNode" - copy paste issue?

anFatum (approved):

workflow/agent_node.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+		agentCtx := internalcontext.NewInvocationContext(ctx, params)
+
+		for event, err := range n.agent.Run(agentCtx) {
+			if err != nil {
+				yield(nil, err)
+				return
+			}
+
+			// Validate output if schema is available and event has output
```
wolo-lab: The wrapper part is straightforward and uncontroversial — but the validation requires some checks and answering some questions (feel free to split):

- [ ] can we relay on output stored in StateDelta["output"]? in agent config we can define outputKey. Please verify. 
- [ ] what about intermediate progress returned from agent? how to deal with `not isFinalResponse`?
- [ ] is it possible that the result which should be validated will be delivered partially by more than 1 events?
- [ ] in scheduler the output is taken from the last event, but the validation is done on the first output
- [ ] adk python uses `Output any` on `Event`. Please verify why they decided to choose this way.

wolo-lab (changes requested):
There are some questions about schema validation which worries me.

workflow/agent_node.go
hanorik: Done.

workflow/agent_node.go
hanorik: Done.

workflow/agent_node_test.go
hanorik: Yes, exactly. Fixed.

workflow/agent_node.go
hanorik: 1. I would say that not fully. `OutputKey` is optional so there are possible situations when it's empty and the result will not be stored in the `StateDelta`.
2. I think we might just yield intermediate events processing only the final one, wdyt?
3. Feels like yes in streaming mode.
4. Yeah, my fail. Will fix it to validate on the last one.

wolo-lab: > I would say that not fully. OutputKey is optional so there are possible situations when it's empty and the 
result will not be stored in the StateDelta.

Let's have a test to verify the correct behavior when outputKey is present.

> I think we might just yield intermediate events processing only the final one, wdyt?

Sounds reasonable. Please compare python implementation and think how to cover it in a test

> Feels like yes in streaming mode.

Streaming mode won't be supported in workflow agents for now. I mean the situation in which lmm agent generate output in multiple events and we need to merge it (somehow) and only then validate. Probably not in scope, but let's verify if adk python supports that.

> Yeah, my fail. Will fix it to validate on the last one.

Ok

hanorik: Ok, will keep that noted and leave just a basic agent node implementation within this PR. All the validations and additional test coverage will be added in the following.

wolo-lab: Was the commit pushed? I don't see the changes.

hanorik: It was and I see actual changes. Probably you selected previous commit?

workflow/agent_node.go
```diff
@@ -0,0 +1,134 @@
...
+// If a schema is nil, it will be inferred from the corresponding generic type Input or Output.
+func newAgentNodeWithSchemasTyped[Input, Output any](a agent.Agent, inputSchema, outputSchema *jsonschema.Schema, cfg NodeConfig) (*AgentNode, error) {
+	if a == nil {
+		return nil, fmt.Errorf("agent cannot be nil")
+	}
+	ischema, err := resolvedSchema[Input](inputSchema)
+	if err != nil {
+		return nil, fmt.Errorf("resolving input schema for agent %q: %w", a.Name(), err)
+	}
+	if ischema == nil {
```
wolo-lab: schema is optional, we don't need these check. The same for oschema in line 55

workflow/agent_node.go
```diff
@@ -0,0 +1,134 @@
...
+				}
+			}
+		}
+
+		// Use existing agent context instead of implementing a new one
+		params := internalcontext.InvocationContextParams{
+			Artifacts:     ctx.Artifacts(),
+			Memory:        ctx.Memory(),
+			Session:       ctx.Session(),
+			Branch:        ctx.Branch(),
```
wolo-lab, line 110: please add TODO: branch insolation in a separate PR

workflow/agent_node.go
```diff
@@ -0,0 +1,134 @@
...
+// NewAgentNode creates a new node wrapping an agent. Input and output schemas
+// are inferred as `any`.
+func NewAgentNode(a agent.Agent, cfg NodeConfig) (*AgentNode, error) {
+	return NewAgentNodeTyped[any, any](a, cfg)
+}
+
+// Run implements the Node interface.
+func (n *AgentNode) Run(ctx agent.InvocationContext, input any) iter.Seq2[*session.Event, error] {
+	return func(yield func(*session.Event, error) bool) {
+		var userContent *genai.Content
```
wolo-lab, line 83: add TODO about input validation - it will be done in a separate PR

workflow/agent_node.go
```diff
@@ -0,0 +1,134 @@
...
+			InvocationID:  ctx.InvocationID(),
+		}
+		agentCtx := internalcontext.NewInvocationContext(ctx, params)
+
+		for event, err := range n.agent.Run(agentCtx) {
+			if err != nil {
+				yield(nil, err)
+				return
+			}
+
```
wolo-lab, line 124: add todo about output validation (separate PR)

wolo-lab (approved):

workflow/agent_node.go
hanorik: Added.

workflow/agent_node.go
hanorik: Added.

workflow/agent_node.go
hanorik: Added.

workflow/agent_node.go
hanorik: Done.

PR #845, opened 2026-05-16
feat: add HITL console launcher support for workflow input prompts

wolo-lab:
## Summary

Wires human-in-the-loop pauses into `cmd/launcher/console`, so a
user running `adk launcher console` sees a HITL prompt when a
workflow node pauses for input, can type a reply, and that reply
routes back to the paused node on the next turn — closing the
loop end-to-end through the console UI.

Detection is name-agnostic: any event with a non-empty
`LongRunningToolIDs` containing a `FunctionCall.ID` is treated as
a pending interrupt, regardless of which kind of pause produced
it (workflow `RequestInput`, future `RequestConfirmation` / other
kinds — same path). Rendering and response shaping then dispatch
on the `FunctionCall.Name`, with `workflow.WorkflowInputFunctionCallName`
having a dedicated renderer and unknown kinds falling back to a
generic banner that still gets the operator's reply through.

Tool confirmation (`toolconfirmation.FunctionCallName`) currently
hits the generic fallback. The transport works (the reply routes
back to the tool by FunctionCall.ID) but the `{"result": <text>}`
envelope does not match what `ctx.ToolConfirmation()` expects.
A dedicated renderer + yes/no parser will land in a follow-up PR.

## Output format

Aligned with the existing console launcher convention
(`Agent -> ` / `User -> `, no brackets, no jargon). Example
session:

    User -> Approve my vacation request for 5 days

    Agent -> Approve $500 vacation request?
      Payload: {
        "user": "alice",
        "days": 5,
        "cost": 500
      }
      Expected response schema: {"type":"object","properties":{"approved":{"type":"boolean"}}}
    User -> {"approved": true}

    Agent -> Request approved. Booked vacation Mon-Fri.

    User ->

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./cmd/launcher/console/...` — all pass
- [x] `go test -race ./cmd/launcher/console/...` — clean
- [x] 3 new tests in `cmd/launcher/console/hitl_test.go`
  (no test infrastructure existed for this package previously):
  - `TestCollectPendingInterrupts_DetectionByLongRunningToolIDs`
    — empty input, FunctionCall without long-running marker,
    long-running marker without matching call, workflow input
    on LLMResponse.Content, and multi-event filtering
  - `TestBuildInterruptResponse_WorkflowInput` — raw-text,
    JSON-object, JSON-scalar, JSON-array, and CRLF-stripping
    cases
  - `TestBuildInterruptResponse_GenericFallback` — catch-all
    path for an unknown long-running call name

### DIFF
--- a/cmd/launcher/console/console.go
+++ b/cmd/launcher/console/console.go
@@ -99,7 +99,7 @@ func (l *consoleLauncher) Run(ctx context.Context, config *launcher.Config) erro
 
 	rootAgent := config.AgentLoader.RootAgent()
 
-	session := resp.Session
+	sess := resp.Session
 
 	r, err := runner.New(runner.Config{
 		AppName:         appName,
@@ -144,6 +144,14 @@ func (l *consoleLauncher) Run(ctx context.Context, config *launcher.Config) erro
 		}
 	}
 
+	// pendingInterrupts carries human-input prompts the agent
+	// emitted on the previous turn. While non-empty, the next
+	// stdin read is interpreted as the answer to its head; once
+	// every prompt has an answer the assembled FunctionResponse
+	// content is sent as the next "user message" turn.
+	var pendingInterrupts []pendingInterrupt
+	var pendingResponses []*genai.Part
+
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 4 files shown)


hanorik (approved):
LGTM.

PR #852, opened 2026-05-18
feat: add HITL console launcher support for tool confirmation prompts

wolo-lab:
## Summary

Adds a dedicated renderer and yes/no parser for tool confirmation
interrupts (`toolconfirmation.FunctionCallName`), so the operator
running `adk launcher console` sees a clear confirmation prompt
instead of the generic fallback.

Without this PR tool confirmation still works at the transport
layer (the reply routes back to the tool by FunctionCall.ID), but
the response is shaped as `{"result": <text>}` — which does not
match what `ctx.ToolConfirmation()` reads. Operators have no way
to actually confirm or reject. This PR fixes that by sending the
expected `{"confirmed": bool}` envelope.

## Output format

Aligned with the existing console launcher convention (`Agent ->
` / `User -> `, no brackets, no jargon — same as #845). Example
session:

    User -> Delete my old vacation requests

    Agent -> Delete 3 old vacation requests? This is irreversible.
      Type 'yes' to confirm, anything else to reject.
    User -> yes

    Agent -> Deleted 3 vacation requests.

    User ->

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./cmd/launcher/console/...` — all pass
- [x] `go test -race ./cmd/launcher/console/...` — clean
- [x] 1 new case in
  `TestCollectPendingInterrupts_DetectionByLongRunningToolIDs`
  verifying detection works on a real tool-confirmation event
  (not just the generic mock used by the base PR).
- [x] New `TestBuildInterruptResponse_ToolConfirmation` covers
  10 yes/no variants: `yes`, `YES`, ` Yes ` (whitespace),
  `y`, `true`, `confirm` (all map to `true`), `no`, `n`,
  blank line, arbitrary text (all map to `false`).

### DIFF
--- a/cmd/launcher/console/hitl.go
+++ b/cmd/launcher/console/hitl.go
@@ -22,6 +22,7 @@ import (
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/session"
+	"google.golang.org/adk/tool/toolconfirmation"
 	"google.golang.org/adk/workflow"
 )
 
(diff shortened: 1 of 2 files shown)


hanorik (approved):
LGTM.

anFatum (approved):

PR #853, opened 2026-05-18
feat: add retry config implementation

anFatum:


workflow/retry.go
```diff
@@ -0,0 +1,67 @@
...
+	if cfg.MaxAttempts <= 1 {
+		return false
+	}
+	if failedAttempts >= cfg.MaxAttempts {
+		return false
+	}
+	if cfg.ShouldRetry != nil {
+		return cfg.ShouldRetry(err)
+	}
+	return false // Default to false if not specified
```
wolo-lab: shouldn't it be true? See config:go:121

defaultRetryConfig also returns true

workflow/state.go
```diff
@@ -98,6 +98,9 @@ type NodeState struct {
 	// NodeWaiting and the wait was caused by a human-input request
 	// (as opposed to a fan-in barrier).
 	PendingRequest *session.RequestInput `json:"pendingRequest,omitempty"`
+
+	// Attempt is the number of times this node has been executed.
```
wolo-lab: 1) I think it's about the number of times the node has failed. 
2) Why not "omitempty" as other fields here?

workflow/retry_test.go
wolo-lab, line 1: It is not clear if attempt is per node execution or per workflow execution. Let's add a test to pin the semantics.

Scenario:  Start -> A,   A  --condition-- > A , A --condition--> Finish.
And let's have maxAttemp=3 and A failing 2 times and pointing during the first execution to itself and later to finish. Flow: Start -> A (2 failures then success) -> A (2 failures then success) -> Finish.

So in total there were 4 failures but 2 by each try.

workflow/scheduler.go
```diff
@@ -240,6 +256,15 @@ func (s *scheduler) scheduleNode(n Node, input any, triggeredBy string) {
 	go runNode(s.eventQueue, &s.wg, name, n, perNodeCtx, input)
 }
 
+// scheduleRetry schedules a retry for node n after the given delay.
+func (s *scheduler) scheduleRetry(n Node, input any, triggeredBy string, delay time.Duration) {
+	time.AfterFunc(delay, func() {
+		go func() {
```
wolo-lab, line 272: 1) new goroutine is not needed, PTAL at http://google3/third_party/go/gc/src/context/context.go?rcl=910118120&l=326

2) when a node waits ot be retried and it will be cancelled in the meantime, the retry will still trigger. it should not happen. The proposal:   have `retryTimers []*time.Timer` in scheduler.  here in `scheduleRetry` add a newly created timer and inside `cancelAll` iterate through them and stop the.

workflow/scheduler.go
```diff
@@ -83,6 +84,12 @@ type scheduler struct {
 	eventQueue chan queueItem
 	wg         sync.WaitGroup
 
+	// pendingRetries is a counter of nodes that are scheduled to run but haven't completed yet.
```
wolo-lab: sounds like "just pending". Maybe "scheduled to retry"? or something similar.

wolo-lab (changes requested):
Nice! Some small adjustments.

hanorik (approved):

workflow/scheduler.go
anFatum: removed, I use the retryTimers instead

wolo-lab (approved):

Issue #857, opened 2026-05-18
`execute_tool` spans are siblings of `generate_content` instead of children — breaks observability span hierarchy


UnderTreeTech:
## Summary

When an LLM agent performs a tool-calling flow (LLM → tool call → LLM summarize), the OpenTelemetry spans emitted by ADK have a **flat sibling structure** under `invoke_agent`, rather than a **nested hierarchy** that reflects the actual causal relationship. This makes it impossible for observability backends (Langfuse, Jaeger, etc.) to correctly display the logical execution order.

## Environment

- **ADK version**: `google.golang.org/adk v1.2.0`
- **Go version**: 1.24+
- **Observability backend**: Langfuse (via OTLP/HTTP), but this affects any OTel-compatible backend

## Current Behavior

When asking an LLM agent a question that requires a tool call (e.g., "What's the weather in Shanghai?"), the ADK emits the following span tree:

```
invoke_agent weather_time_agent          (root span)
  ├── generate_content deepseek-v4-pro   (Step 1: LLM returns FunctionCall)
  ├── execute_tool get_city_weather      (Tool execution)
  └── generate_content deepseek-v4-pro   (Step 2: LLM summarizes tool result)
```

All three child spans (`generate_content`, `execute_tool`, `generate_content`) are **direct children of `invoke_agent`** at the same level. There is no span hierarchy that indicates `execute_tool` was triggered *because of* the first `generate_content` response.

In observability UIs like Langfuse, this flat structure causes the spans to be displayed in start-time order, but with no causal relationship visible. Worse, due to `BatchSpanProcessor` timing, the display order can appear as:

```
invoke_agent → generate_content → generate_content → execute_tool
```

which is misleading — it looks like two LLM calls happened before the tool was ever invoked.

## Expected Behavior

The span hierarchy should reflect the actual **causal/logical relationship** between steps within a single `runOneStep` cycle. Each `runOneStep` iteration should group its `generate_content` and resulting `execute_tool` spans together:

**Option A: Nested under `generate_content`**
```
invoke_agent weather_time_agent
  ├── generate_content deepseek-v4-pro   (Step 1)
  │     └── execute_tool get_city_weather
  └── generate_content deepseek-v4-pro   (Step 2: final response)
```

**Option B: Grouped under a `step` span**
```
invoke_agent weather_time_agent
  ├── step_1
  │     ├── generate_content deepseek-v4-pro
  │     └── execute_tool get_city_weather
  └── step_2
[...]

PR #858, opened 2026-05-18
feat: add llm_step span to group generate_content and execute_tool wi…

UnderTreeTech:
### Description

[issue] https://github.com/google/adk-go/issues/857

#### Problem

Currently, `generate_content` and `execute_tool` spans are all created as **direct children of the `invoke_agent` span** (flat siblings). This makes it impossible for observability backends (Langfuse, Jaeger, etc.) to correctly display the causal relationship between an LLM call and the tool execution it triggered.

For a simple tool-calling flow ("What's the weather in Shanghai?"), the span tree looks like:

```
invoke_agent weather_time_agent
  ├── generate_content deepseek-v4-pro    (Step 1)
  ├── generate_content deepseek-v4-pro    (Step 2)
  └── execute_tool get_city_weather
```

Observability backends display these in start-time order (or even worse, in export-batch order), making it appear as if two LLM calls happened before the tool was ever invoked.

#### Root Cause

In `internal/llminternal/base_flow.go`, `runOneStep()` passes the `invoke_agent`-level context to both `callLLM()` (which creates `generate_content` spans) and `handleFunctionCalls()` (which creates `execute_tool` spans). Since the `generate_content` span is already ended before `handleFunctionCalls` runs, there is no intermediate span to establish a parent-child grouping.

#### Fix

This PR introduces a **`llm_step` span** at the beginning of each `runOneStep()` iteration. All operations within that iteration — including `generate_content` and `execute_tool` — become children of this step span.

**After the fix**, the span tree becomes:

```
invoke_agent weather_time_agent
  ├── llm_step                              (iteration 1)
  │     ├── generate_content deepseek-v4-pro
  │     └── execute_tool get_city_weather
  └── llm_step                              (iteration 2)
        └── generate_content deepseek-v4-pro
```

#### Changes

- **`internal/llminternal/base_flow.go`**: Added `telemetry.StartTrace(ctx, "llm_step")` at the top of `runOneStep()` and updated the context so all downstream spans are children of the step span.

#### Compatibility

- The Langfuse enrichment plugin matches pending LLM calls by the **parent span ID** of `generate_content` spans. Before this change, that parent was `invoke_agent`; after this change, it becomes `llm_step`. Since the `beforeModel` callback also sees the same `llm_step` span in its context (via `SpanFromContext`), the matching logic remains correct with zero changes to the plugin code.
- No behavioral change to agent execution — the step span is purely observability overhead.
- Existing tests pass without modification.

### DIFF
--- a/internal/llminternal/base_flow.go
+++ b/internal/llminternal/base_flow.go
@@ -423,6 +423,13 @@ func (f *Flow) runOneStep(ctx agent.InvocationContext) iter.Seq2[*session.Event,
 			return
 		}
 
+		// Create a step-level span to group generate_content and execute_tool
+		// spans within the same iteration, establishing correct parent-child
+		// hierarchy for observability backends.
+		stepCtx, stepSpan := telemetry.StartTrace(ctx, "llm_step")
+		defer stepSpan.End()
+		ctx = ctx.WithContext(stepCtx)
+
 		req := &model.LLMRequest{
 			Model: f.Model.Name(),
 		}


karolpiotrowicz:
@wolo-lab worth a decision, and something has changed since this was opened.

The change inserts a new span type between the agent span and its children, altering the emitted span tree that adk-web consumes. No sibling implementation has this span. More to the point, golden fixtures were added to the repository after this pull request was written that assert the current flat shape as expected, so what the issue calls a defect is now an encoded expectation. Reconciling those two is the decision — the diff itself is seven lines.

PR #862, opened 2026-05-20
feat: add new parallel worker node type to the adk workflows so it could process multiple inputs simultaneously

anFatum:


workflow/parallel_worker.go
```diff
@@ -0,0 +1,174 @@
...
+
+			if res.ev != nil {
+				if out, ok := res.ev.Actions.StateDelta["output"]; ok {
+					outputs[res.index] = out
+				}
+			}
+		}
+
+		if len(errs) > 0 {
+			yield(nil, errors.Join(errs...))
```
anFatum: Or should we return the partial result here?

hanorik (approved):

workflow/parallel_worker.go
wolo-lab, line 1: RetryConfig in wrapped node is ignored:

If user will have:
worker := NewFunctionNode("flaky", flakyFn, NodeConfig{
    RetryConfig: &RetryConfig{MaxAttempts: 3},  // retry wrapped
})
pw := NewParallelWorker("parallel", worker, 0, NodeConfig{
    RetryConfig: &RetryConfig{MaxAttempts: 3},  // retry parallel
})
And let's have
Input: []any{"a", "b", "c"}. flakyFn fails always for "b".
User would expect: 1 ("a") + 3 ("b" retried) + 1 ("c") = 5 calls
Therefore I would go in the direction:
RetryConfig on the wrapped node is forbidden (check in constructor) -  only ParallelWorker's
own RetryConfig applies and it retries in a smart way, only failures. I believe you need to pass control to scheduler.

HITL in a wrapped node is not supported, please add a test for it.

workflow/parallel_worker_test.go
wolo-lab, line 1: Please add tests for:

1. cancel during execution: no deadlocks, no final result
2. Test `maxConcurrency > nItems` (trivial) 
3. Test concurrent multi-output workers: order in (`outputs[i] = [a, b, c]` should be kept for `i` even if nodes return result in random order
4. test `RetryConfig`
5. test for HITL

workflow/parallel_worker.go
wolo-lab: after getting a first error, cancel other jobs and return the first error.
Please document.

wolo-lab (changes requested):

workflow/parallel_worker.go
anFatum: I've added RetryConfig support to ParallelWorker, although I haven't involved scheduler, as the retries should be performed only on the failed inputs and it's easier to track in ParallelWorker itself than populate it back to scheduler.

Please let me know if you would like to rather have it in scheduler

wolo-lab: Right call, agreed.

workflow/parallel_worker_test.go
```diff
@@ -0,0 +1,591 @@
...
+	done := make(chan struct{})
+	go func() {
+		for range pw.Run(mockCtx, input) {
+		}
+		close(done)
+	}()
+
+	// Wait a bit for workers to start.
+	// We expect at most 2 workers to start because maxConcurrency is 2.
+	time.Sleep(100 * time.Millisecond)
```
wolo-lab: `Sleep` in tests is anti-pattern. In some CI under heavy load it might not be sufficient. You could use a channel to signal when workers have started.

workflow/parallel_worker.go
wolo-lab, line 1: Mikalai just merged the changes about using `Event.Output` instead of `StateDelta["output"]`.
It can be done in this or in the follow-up cl.

workflow/parallel_worker_test.go
```diff
@@ -0,0 +1,591 @@
...
+
+type delayedMultiOutputTestNode struct {
+	BaseNode
+}
+
+func (n *delayedMultiOutputTestNode) Run(ctx agent.InvocationContext, input any) iter.Seq2[*session.Event, error] {
+	return func(yield func(*session.Event, error) bool) {
+		s := input.(string)
+
+		var delay time.Duration
```
wolo-lab: As previously, it might be flaky. 
Either add synchronization by channels, or remove for now and create a bug to address it later.

wolo-lab (approved):
Thanks, just some adjustments in tests to make them more reliable.

PR #865, opened 2026-05-20
fix: convert unsupported artifact MIME types to text

### LINKED ISSUE #864: LoadArtifactsTool should handle unsupported artifact MIME types gracefully
## Required Information

**Describe the Bug:**

**Steps to Reproduce:**

**Expected Behavior:**

**Observed Behavior:**

**Environment Details:**

- OS: macOS
- Go Version: 1.22+

**Model Information:**

---

## Optional Information

1wos:
Fixes #864

## Summary

`LoadArtifactsTool` currently appends loaded artifact parts directly into the next model request. This can produce invalid requests when an artifact is stored as inline data with a MIME type that Gemini does not support inline.

This change adds a model-safe conversion step before appending loaded artifacts:

- preserves supported inline media: `image/*`, `audio/*`, `video/*`, `application/pdf`
- converts text-like inline data to text parts: `text/*`, `application/csv`, `application/json`, `application/xml`
- replaces unsupported binary inline data with a short placeholder containing artifact name, MIME type, and size

The behavior mirrors the ADK Python fix from google/adk-python#4028 / fdc98d5c.

## Tests

- `go test ./tool/loadartifactstool`

### DIFF
--- a/tool/loadartifactstool/load_artifacts_tool.go
+++ b/tool/loadartifactstool/load_artifacts_tool.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/genai"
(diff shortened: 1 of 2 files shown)


1wos:
Closing for now. I will revisit this after preparing the contribution separately.

PR #866, opened 2026-05-20
fix(tools): convert unsupported artifact MIME types to text in LoadArtifactsTool

### LINKED ISSUE #867: LoadArtifactsTool should handle unsupported artifact MIME types gracefully
## Required Information

**Describe the Bug:**

**Steps to Reproduce:**

**Expected Behavior:**

**Observed Behavior:**

**Environment Details:**

- OS: macOS
- Go Version: 1.22+

**Model Information:**

---

## Optional Information

1wos:

- Closes: #867

**Problem:**

`LoadArtifactsTool` appends loaded artifact parts directly into the next model request. When the inline MIME is not accepted by Gemini, the follow-up request fails with `Unsupported MIME Type` instead of continuing with usable context. Same class of issue as adk-python #4028 (fixed by commit `fdc98d5c`).

**Solution:**

MIME normalization strips the `;charset=...` parameter, so `application/csv; charset=utf-8` is handled the same as `application/csv`.

### Testing Plan

**Unit Tests:**

Added `TestLoadArtifactsTool_ProcessRequest_MIMEConversion`, a table test with nine cases covering the fallback paths:

- text-like type is inlined as text
- media type keeps its inline data
- document type keeps its inline data
- type casing is ignored
- unsupported binary type becomes a placeholder
- text type holding invalid UTF-8 becomes a placeholder
- missing type defaults to octet-stream
- empty data becomes a placeholder
- empty data is reported even for a supported type

```
$ go test -count=1 ./tool/loadartifactstool/...
ok  google.golang.org/adk/v2/tool/loadartifactstool  0.422s
```

23/23 pass. No existing test was changed or removed.

Adapted `examples/tools/loadartifacts/main.go` locally (not part of this PR) to save a `weather.csv` artifact with `application/csv` inline MIME and ask the agent to describe it against the live Gemini API. The agent loaded the artifact, the fallback converted the inline csv to a text part, and the model produced an accurate description:

```
=== E2E: LoadArtifactsTool with application/csv (unsupported inline MIME) ===
FunctionCall: load_artifacts(map[artifact_names:[weather.csv]])
FunctionResponse: load_artifacts -> map[artifact_names:[weather.csv]]
Agent: The artifact `weather.csv` contains weather data. It has two columns:
"city" and "temperature". The data includes temperatures for Seoul (12),
Tokyo (18), and Singapore (30).
=== Done ===
```

Without the fallback, this request fails with `Unsupported MIME Type` before the model can respond.

### Checklist

### Additional context

Ports the behavior introduced in adk-python by commit `fdc98d5c` (Close #4028). Follows the "Alignment with adk-python" section of CONTRIBUTING.md.


indurireddy-TF:
Hi @1wos Could you please resolve the merge conflicts so we can proceed with the review?

1wos:
Sorry for the delay. Rebased onto main and resolved the conflict.

It needed a little more than a conflict fix since `tool.Context` moved to `agent.Context` in the meantime. I tightened a few things while I was in there too: case-insensitive MIME matching, empty inline data handled before the supported-type shortcut, and size reported in bytes. Tests are table driven now.

gofmt, vet and the package tests are clean. Ready for review.

indurireddy-TF:
Hi @1wos Can you please resolve merge conflicts?

1wos:
> Hi @1wos Can you please resolve merge conflicts?

@indurireddy-TF  hi ! Rebased onto the latest main and fixed the conflicts. Kept both the new upstream tests and mine, all tests pass locally.

karolpiotrowicz (dismissed):
This is the right shape for #867 and it holds up well: build, vet, `gofmt`, golangci-lint at the CI pin, staticcheck, govulncheck and `apidiff` are all clean, it merges into current main without conflicts, and a `-race` run driving 24 artifacts concurrently through `ProcessRequest` is clean too. Stripping the MIME parameter in `normalizeMIMEType` is the detail that makes this work on real input, since `application/csv; charset=utf-8` and `application/json; charset=utf-8` are the forms an HTTP `Content-Type` actually carries. One thing needs changing before merge.

## Text in a legacy encoding is discarded rather than decoded

[`load_artifacts_tool.go:290-292`](https://github.com/google/adk-go/blob/91b5d3e913733f46b41192d0ae8a06198fb3d043/tool/loadartifactstool/load_artifacts_tool.go#L290-L292) requires `utf8.Valid(data)` before converting, so a text-like artifact whose bytes are not valid UTF-8 falls through to the binary placeholder and its content never reaches the model. A CSV exported from Excel in cp1252 — the file type #867 is about — produces `[Binary artifact: sales.csv, type: application/csv, size: 9 bytes. Content cannot be displayed inline.]`, and the model reports it cannot read the file.

The comment on line 290 justifies the guard for one case, binary data mislabelled as text. The rule also catches genuinely textual files that simply are not UTF-8, which is a large share of spreadsheet and log exports in practice.

[The commit this PR ports](https://github.com/google/adk-python/commit/fdc98d5c927bfef021e87cf72103892e4c2ac12a) already handled that case: its text path is `data.decode('utf-8')` with a `UnicodeDecodeError` fallback to `data.decode('utf-8', errors='replace')`, still present today at [load_artifacts_tool.py:273-277](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/tools/load_artifacts_tool.py#L273-L277). So this is a gap against the revision being ported, not against something upstream added later.

<details><summary>Reproduction — fails on this branch, passes with the change below</summary>

```go
// tool/loadartifactstool/zz_repro_test.go
package loadartifactstool_test

import (
	"testing"

	"google.golang.org/genai"

	"google.golang.org/adk/v2/internal/toolinternal"
	"google.golang.org/adk/v2/tool/loadartifactstool"
)

// A CSV exported from Excel in cp1252: "café, 12\n", where é is the single
// byte 0xE9. Valid text, not valid UTF-8.
func TestLegacyEncodedCSVLosesItsContent(t *testing.T) {
	tool := loadartifactstool.New()
	tc := createToolContext(t)
	if _, err := tc.Artifacts().Save(t.Context(), "sales.csv",
		genai.NewPartFromBytes([]byte("caf\xe9, 12\n"), "application/csv")); err != nil {
		t.Fatal(err)
	}

		Role: "model",
		Parts: []*genai.Part{genai.NewPartFromFunctionResponse("load_artifacts",
			map[string]any{"artifact_names": []string{"sales.csv"}})},
	}}}
	if err := tool.(toolinternal.RequestProcessor).ProcessRequest(tc, req); err != nil {
		t.Fatal(err)
	}

	got := req.Contents[1].Parts[1].Text
	if want := "caf\uFFFD, 12\n"; got != want {
		t.Errorf("artifact content did not reach the model\n got: %q\nwant: %q", got, want)
	}
}
```

```
$ go test -run TestLegacyEncodedCSV ./tool/loadartifactstool/
--- FAIL: TestLegacyEncodedCSVLosesItsContent (0.00s)
    artifact content did not reach the model
     got: "[Binary artifact: sales.csv, type: application/csv, size: 9 bytes. Content cannot be displayed inline.]"
    want: "caf\ufffd, 12\n"
```

</details>

Replacing the guard with a lossy decode is enough, and I ran it rather than guessing:

```go
	// Text in a legacy encoding still carries readable content, so replace the
	// invalid sequences rather than discarding the artifact.
	if isTextLikeMIMEType(mimeType) {
		return genai.NewPartFromText(strings.ToValidUTF8(string(data), "\uFFFD"))
	}
```

That drops the `unicode/utf8` import, builds clean, and leaves the package suite green once the `text type holding invalid UTF-8 becomes a placeholder` case is updated to assert the recovered text. A 30-input sweep through `ProcessRequest` confirms nothing else changes. Go collapses a run of invalid bytes into one `U+FFFD` where Python emits one per byte, which affects the replacement text only.

## `image/svg+xml` still goes inline

[Line 254](https://github.com/google/adk-go/blob/91b5d3e913733f46b41192d0ae8a06198fb3d043/tool/loadartifactstool/load_artifacts_tool.go#L252-L258) matches the bare `image/` prefix, so `image/svg+xml` and `image/xml` are treated as inline-supported and forwarded untouched. adk-python subtracts those three subtypes at [load_artifacts_tool.py:53-57](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/tools/load_artifacts_tool.py#L53-L57) and routes them to the text path instead, noting they were verified against gemini-2.5-flash.

Not something to hold this PR for, and worth saying why: `main` forwards SVG inline today as well, so nothing regresses here, and the upstream exclusion landed [in a later adk-python commit](https://github.com/google/adk-python/commit/f9be94c624bfa17f5518e1e134fa516e36138ef1) on 2026-06-16, four weeks after you opened this. Adding the carve-out now would be a nice extra. Leaving it for a follow-up is fine.

## Smaller things

- **The MIME type reaches the prompt unescaped.** [Lines 285](https://github.com/google/adk-go/blob/91b5d3e913733f46b41192d0ae8a06198fb3d043/tool/loadartifactstool/load_artifacts_tool.go#L285) and [295-300](https://github.com/google/adk-go/blob/91b5d3e913733f46b41192d0ae8a06198fb3d043/tool/loadartifactstool/load_artifacts_tool.go#L295-L300) interpolate it with `%s`, and `normalizeMIMEType` only cuts at the first `;` and trims the outer edges, so embedded newlines survive into the placeholder and break its `[...]` framing. `%q` on both the MIME type and the artifact name closes it. Low stakes — the artifact name already reaches the prompt raw on `main`, and anyone who can set the MIME type can set the bytes, which become prompt text by design — but the fix costs one character per verb.
- **Four allow-list entries are untested.** Dropping `application/json` or `application/xml` from `isTextLikeMIMEType`, or `audio/` or `video/` from `isInlineMIMETypeSupported`, leaves the package green. So does deleting the `part == nil` placeholder, and so does removing the `normalizeMIMEType` call inside `isInlineMIMETypeSupported` — meaning the parameter stripping that makes this change work is not itself pinned by a test. Four more table rows and one `nil` case would cover it.
- [`isTextLikeMIMEType`](https://github.com/google/adk-go/blob/91b5d3e913733f46b41192d0ae8a06198fb3d043/tool/loadartifactstool/load_artifacts_tool.go#L261-L266) does not normalize its argument while its sibling does, so `isTextLikeMIMEType("TEXT/PLAIN")` is false. Harmless today because the only call site pre-normalizes, worth a line of doc either way.

One more thing, since it is your work being carried forward: [#882](https://github.com/google/adk-go/pull/882) proposes a different fix for the same issue and I am recommending this one, because #882 does not strip MIME parameters and so turns `application/csv; charset=utf-8` and `application/pdf; version=1.7` into placeholders. Its placeholders do use `%q`, which is the escaping suggested above.

1wos:
Thanks for the depth here, and for running the fix instead of just sketching it. Took your change as-is: the `utf8.Valid` guard is gone, text-like artifacts decode through `strings.ToValidUTF8`, and your cp1252 CSV repro is now a table case asserting `caf�, 12\n`. The old invalid-UTF-8 case asserts the recovered text instead of a placeholder.

Also picked up the smaller items: `%q` on the placeholder fields, table rows pinning `application/json`, `application/xml`, `audio/`, `video/` and the parameter stripping inside `isInlineMIMETypeSupported`, a nil-part case, and the doc line on `isTextLikeMIMEType` expecting normalized input.

Left the `image/svg+xml` carve-out for a follow-up as you suggested.

gofmt, vet and `go build ./...` are clean, and the tool tree tests pass locally.

1wos:
Went ahead and added the SVG carve-out here rather than in a follow-up, since it is small: `image/svg`, `image/svg+xml` and `image/xml` now fall out of the inline check and through the text path, matching the subtype set and rationale in adk-python's `_GEMINI_UNSUPPORTED_INLINE_SUBTYPES`. `application/svg+xml` joins the text-like set for the same parity. Four table cases cover the new routing; build, vet and the tool tree tests are green.

karolpiotrowicz (commented):
All four items from the last round are fixed, and I checked each by running it rather than reading the diff: the cp1252 CSV keeps its content, `image/svg+xml` takes the text path, a MIME type full of newlines no longer breaks the placeholder frame, and five of the six unpinned branches now fail when you break them. The parameterized-PDF case is untouched, and it merges into current main clean with the tool and artifact trees green.

Adding the SVG carve-out here rather than deferring it was the right call, and I can now say why from a measurement instead of a citation. Real `generateContent` calls against `gemini-2.5-flash`:

```
SVG source,     label image/svg+xml / image/svg / image/xml  -> 400 "Provided image is not valid."
SVG source,     label image/png / image/jpeg / image/webp    -> 400  (the same error)
raster PNG,     label image/svg+xml                          -> 200  "Red"
the same SVG sent as a text part                             -> 200  model described the drawing
```

The backend is rejecting on the bytes, not on the label. That does not change the list, which matches adk-python exactly, but it does change what the list can be said to do — and two consequences fall out of it. An SVG saved under a non-SVG image label still fails, and this carve-out does not help it. A raster image mislabelled `image/svg+xml`, which works today, now arrives as 74 runes of replacement characters.

Four things left, none of them blocking. Each is on the line it concerns.

Last, a correction to something I told you last round. I said Go collapsing a run of invalid bytes into a single `U+FFFD` affects the replacement text only. That is true for cp1252, which is what I measured, and not true for multibyte legacy encodings — cp1251 `Привет,12` comes out as `\uFFFD,12`, so a whole word becomes one character. It does not change the decision to decode rather than discard, and your new test asserts the Go behavior correctly.

tool/loadartifactstool/load_artifacts_tool.go
```diff
@@ -232,8 +233,85 @@ func (t *artifactsTool) loadIndividualArtifact(ctx context.Context, artifactsSer
...
+// case-insensitive per RFC 2045.
+func normalizeMIMEType(mimeType string) string {
+	mimeType, _, _ = strings.Cut(mimeType, ";")
+	return strings.ToLower(strings.TrimSpace(mimeType))
+}
+
+// isInlineMIMETypeSupported reports whether the model accepts this type inline.
+// The media match is prefix-based on purpose: enumerating subtypes would rot.
+func isInlineMIMETypeSupported(mimeType string) bool {
+	m := normalizeMIMEType(mimeType)
```
karolpiotrowicz: This inner call is unreachable in practice, so the table row meant to pin parameter-stripping inside `isInlineMIMETypeSupported` does not reach it either. Line 293 normalizes before line 302 makes the call, so `normalizeMIMEType` here never receives a non-normalized value from any test or any caller. Deleting this line leaves the package green.

The row still earns its place — it proves the original MIME string survives on the inline path — but what it exercises is line 293. Asserting `isInlineMIMETypeSupported("IMAGE/SVG+XML; charset=utf-8") == false` in the internal test would cover it, or drop this call and give the function the same "argument must already be normalized" note you added to `isTextLikeMIMEType`.

tool/loadartifactstool/load_artifacts_tool.go
```diff
@@ -232,8 +233,85 @@ func (t *artifactsTool) loadIndividualArtifact(ctx context.Context, artifactsSer
...
+	// XML-based image subtypes match the image/ prefix but Gemini rejects them
```
karolpiotrowicz: This overstates what the carve-out does, given the probe in the review body: Gemini rejects SVG source under `image/png` and `image/jpeg` with the identical 400, and accepts a raster PNG under `image/svg+xml`. The rejection is on the bytes, not on the declared subtype, so "XML-based image subtypes ... Gemini rejects them" describes a rule the backend does not apply.

The list itself is right and matches adk-python. Worth saying here that it catches the usual spelling rather than matching the backend's actual rule, because the stated mechanism is what a later change would cite when extending or trimming the list.

tool/loadartifactstool/load_artifacts_tool.go
```diff
@@ -232,8 +233,85 @@ func (t *artifactsTool) loadIndividualArtifact(ctx context.Context, artifactsSer
...
+	// as inline data with 400 INVALID_ARGUMENT, so they take the text path.
+	// Mirrors adk-python's _GEMINI_UNSUPPORTED_INLINE_SUBTYPES, verified there
+	// against gemini-2.5-flash.
+	switch m {
```
karolpiotrowicz: This switch matches exactly while line 261 matches by prefix, and the mismatch fails in both directions once a control character is in the label. `strings.TrimSpace` on line 246 strips Unicode spaces but not control characters, so `image/svg+xml\x00` misses this switch and still matches the `image/` prefix below, landing back on the inline path the carve-out exists to avoid.

The quiet direction is the worse one: `application/json\x00` misses the exact-match list in `isTextLikeMIMEType` and the `text/` prefix both, so a perfectly readable JSON artifact becomes a binary placeholder with nothing to indicate content was dropped.

Unchanged from before your two commits, so it is not something you introduced.

tool/loadartifactstool/load_artifacts_tool.go
```diff
@@ -232,8 +233,85 @@ func (t *artifactsTool) loadIndividualArtifact(ctx context.Context, artifactsSer
...
+			safePartForLLM(resp.Part, artifactName),
 		},
 		Role: genai.RoleUser,
 	}, nil
 }
+
+// normalizeMIMEType drops any parameters and lowercases the type, which is
+// case-insensitive per RFC 2045.
+func normalizeMIMEType(mimeType string) string {
+	mimeType, _, _ = strings.Cut(mimeType, ";")
```
karolpiotrowicz, line 250: Removing the `utf8.Valid` guard was my suggestion last round, and it has a cost I should have flagged when I made it: the declared label is now trusted absolutely, so 1 MiB of binary mislabelled `text/plain` reaches the model as 712,702 runes where it used to be a 103-byte placeholder. adk-python does the same thing with no cap and a worse expansion ratio, so I am not asking you to diverge from it.

One cheap improvement is available on this line, though. `strings.Cut` discards the charset parameter, so `text/csv; charset=windows-1252` holding `caf\xe9,12` yields `caf\uFFFD,12` when `café,12` was recoverable from information that was right there in the label.

1wos:
All four picked up:

- Dropped the unreachable normalize inside `isInlineMIMETypeSupported` and gave it the same "argument must already be normalized" note as its sibling. The table row stays, since what it actually pins is the call-site normalization.
- Reworded the carve-out comment along your probe: the rejection is on the bytes, so the comment now says the list catches the conventional labels rather than mirroring a backend rule.
- `normalizeMIMEType` now strips control characters, so `application/json\x00` lands on the text path instead of a silent placeholder and `image/svg+xml\x00` cannot sneak back inline. Two table rows pin both directions. It predates this PR as you said, but the fix was one `strings.Map` away, so it felt wrong to leave it.
- `decodeArtifactText` honours a declared charset via `ianaindex` before falling back to lossy replacement: `text/csv; charset=windows-1252` holding `caf\xe9` now comes back as `café`, with an unknown-charset row keeping the fallback pinned. One manifest note: this moves `golang.org/x/text` from indirect to direct in go.mod. It was already in the module graph, just flagging it since it touches the manifest.

Thanks for the Gemini probe and the cp1251 correction, both went straight into the comments. gofmt, vet, `go build ./...` and the tool tree are green.

karolpiotrowicz:
@wolo-lab flagging this for a call on direction rather than a code review, on a point that only appeared in the latest commit.

The PR now honours the `charset` parameter on an artifact's MIME type, so `text/csv; charset=windows-1252` decodes to its real characters instead of degrading to replacement characters. That is a genuine improvement and I want to be transparent that I prompted it — I noted the charset was being discarded when the information was right there in the label.

Two things make it a decision rather than a nit. It puts adk-go ahead of adk-python rather than at parity with it: Python's `_normalize_mime_type` strips the parameter outright at [load_artifacts_tool.py#L82-L86](https://github.com/google/adk-python/blob/4cd8aa4043277b7c40aabad77ebb7c8bd11c6ca0/src/google/adk/tools/load_artifacts_tool.py#L82-L86) and then decodes as UTF-8 with `errors='replace'` at [L275-L277](https://github.com/google/adk-python/blob/4cd8aa4043277b7c40aabad77ebb7c8bd11c6ca0/src/google/adk/tools/load_artifacts_tool.py#L275-L277), so the same artifact comes back mangled there and correct here. And it promotes `golang.org/x/text` from an indirect to a direct dependency, for one import of `encoding/ianaindex`.

Neither is alarming on its own. `x/text` is already in the module graph and is Go-team maintained, so nothing new is being pulled in, and the implementation is one small function that falls back to the previous behaviour whenever the charset is absent, unrecognised, or fails to decode. The question is whether we want adk-go leading the other implementations on artifact decoding and taking a direct dependency to do it, or whether this should track Python and be raised upstream first.

The rest of the PR is unaffected by the answer, and it is a fix for a reported crash. Happy to finish the review either way once there is a direction.

tool/loadartifactstool/load_artifacts_tool.go
```diff
@@ -232,8 +236,112 @@ func (t *artifactsTool) loadIndividualArtifact(ctx context.Context, artifactsSer
...
+
+// decodeArtifactText converts artifact bytes to UTF-8 text. A charset
+// parameter on the original MIME type is honoured when it names a known
+// encoding, so "text/csv; charset=windows-1252" round-trips instead of
+// degrading; otherwise invalid sequences are replaced rather than the
+// artifact discarded.
+func decodeArtifactText(data []byte, rawMIMEType string) string {
+	if _, params, err := mime.ParseMediaType(rawMIMEType); err == nil {
+		if cs := params["charset"]; cs != "" && !strings.EqualFold(cs, "utf-8") {
+			if enc, err := ianaindex.IANA.Encoding(cs); err == nil && enc != nil {
```
karolpiotrowicz, line 303: This guard is load-bearing and nothing currently fails if you remove it. `ianaindex.IANA.Encoding` returns `(nil, nil)` — no error — for a charset it knows but does not implement, and `gb2312` and `utf-7` are both in that set. Drop `&& enc != nil` and the package suite still passes, while an artifact labelled `text/plain; charset=gb2312` panics with a nil pointer dereference on the next line.

The existing "unknown charset falls back to lossy decoding" row cannot catch that, because `no-such-charset` returns an *error*, so `err == nil` short-circuits before the guard is ever reached. A row using `gb2312` or `utf-7` would pin it.

karolpiotrowicz, line 303: Worth knowing where `ianaindex.IANA` stops, since it decides which of these actually decode. It resolves `windows-1252`, `iso-8859-1`, `latin1`, `us-ascii`, `windows-1251`, `shift_jis`, `ibm437` and `iso-2022-jp`, and it rejects `cp1252`, `ascii`, `utf8` and `sjis`. So `text/csv; charset=cp1252` still comes out as `caf\uFFFD,12` — the spelling Windows tooling and Python's `codecs` emit most often is the one that misses.

I am not suggesting a swap to `htmlindex`, because it is not a superset — it handles every alias above but fails on `csUTF8`, `ibm437` and `utf-7`, and pulling in a second table works against the binary-size point in the review body. Nothing here regresses, an unresolved alias falls back to exactly the old lossy path. It is a coverage limit worth being deliberate about rather than a bug.

tool/loadartifactstool/load_artifacts_tool.go
```diff
@@ -232,8 +236,112 @@ func (t *artifactsTool) loadIndividualArtifact(ctx context.Context, artifactsSer
...
+}
+
+// decodeArtifactText converts artifact bytes to UTF-8 text. A charset
+// parameter on the original MIME type is honoured when it names a known
+// encoding, so "text/csv; charset=windows-1252" round-trips instead of
+// degrading; otherwise invalid sequences are replaced rather than the
+// artifact discarded.
+func decodeArtifactText(data []byte, rawMIMEType string) string {
+	if _, params, err := mime.ParseMediaType(rawMIMEType); err == nil {
+		if cs := params["charset"]; cs != "" && !strings.EqualFold(cs, "utf-8") {
```
karolpiotrowicz, line 302: Nothing distinguishes this short-circuit from letting a declared `utf-8` go through the decoder. Replace `!strings.EqualFold(cs, "utf-8")` with `true` and the whole package suite still passes.

The two are not equivalent on invalid input — `strings.ToValidUTF8` collapses a run of bad bytes into a single `U+FFFD`, whereas the x/text UTF-8 decoder emits one per byte. Small, but it is the sort of thing that changes silently under a dependency bump.

tool/loadartifactstool/load_artifacts_tool.go
```diff
@@ -232,8 +236,112 @@ func (t *artifactsTool) loadIndividualArtifact(ctx context.Context, artifactsSer
...
 }
+
+// normalizeMIMEType drops any parameters and control characters and lowercases
+// the type, which is case-insensitive per RFC 2045. Control characters never
+// appear in a legitimate label but would otherwise dodge the exact-match lists
+// below while still hitting the prefix matches.
+func normalizeMIMEType(mimeType string) string {
+	mimeType, _, _ = strings.Cut(mimeType, ";")
+	mimeType = strings.Map(func(r rune) rune {
+		if r < 0x20 || r == 0x7f {
```
karolpiotrowicz, line 252: This closes C0 and DEL, which is what the two new test rows cover, and the same dodge stays open for anything above 0x7f that is not whitespace. `strings.TrimSpace` removes Unicode whitespace but not category-Cf format characters, and only at the edges.

Both directions still reproduce with a zero-width space. `image/svg+xml\u200b` misses the exact-match list, hits the `image/` prefix below, and goes out inline — the case this carve-out exists to prevent. `application/json\u200b` misses both the exact list and the `text/` prefix, so readable JSON is replaced by a binary placeholder with nothing signalling that content was dropped.

This is not something you introduced and I would not hold the PR for it. If you want to close the class rather than another member of it, validating the normalized result against the RFC 2045 `type/subtype` token grammar and treating anything else as unknown would do it in one place.

tool/loadartifactstool/load_artifacts_tool.go
```diff
@@ -232,8 +236,112 @@ func (t *artifactsTool) loadIndividualArtifact(ctx context.Context, artifactsSer
...
+	if len(data) == 0 {
+		return genai.NewPartFromText(fmt.Sprintf("[Artifact: %q, type: %q. No inline data was provided.]", artifactName, mimeType))
+	}
+	if isInlineMIMETypeSupported(mimeType) {
+		return part
+	}
+	// Text in a legacy encoding still carries readable content, so decode it
+	// rather than discarding the artifact.
+	if isTextLikeMIMEType(mimeType) {
+		return genai.NewPartFromText(decodeArtifactText(data, part.InlineData.MIMEType))
```
karolpiotrowicz, line 338: The raw label is the right thing to pass here — the normalized one has already been cut at the first `;`, so it cannot carry a charset at all and the feature would be a no-op. Worth noting one seam it leaves, though.

A control character in the label is now stripped for routing but not for parsing, so the two halves of this commit disagree on the same input. `text/csv\x00; charset=windows-1252` holding `caf\xe9,12` reaches the text path because normalization removed the NUL, then `mime.ParseMediaType` rejects the raw string over that same NUL, the charset is dropped, and the result is `caf\uFFFD,12`. Without the NUL the identical artifact yields `café,12`.

It degrades to the previous behaviour rather than to anything worse, so this is minor. Control-stripping the raw label once and feeding that to both `normalizeMIMEType` and `mime.ParseMediaType` would close it.

tool/loadartifactstool/load_artifacts_tool.go
```diff
@@ -232,8 +236,112 @@ func (t *artifactsTool) loadIndividualArtifact(ctx context.Context, artifactsSer
...
+	if mimeType == "" {
+		mimeType = "application/octet-stream"
+	}
+	data := part.InlineData.Data
+	// An empty blob is degenerate whatever its declared type.
+	if len(data) == 0 {
+		return genai.NewPartFromText(fmt.Sprintf("[Artifact: %q, type: %q. No inline data was provided.]", artifactName, mimeType))
+	}
+	if isInlineMIMETypeSupported(mimeType) {
+		return part
```
karolpiotrowicz, line 333: Once normalization opens the inline path, this forwards `part` with its *raw* label, so the sanitization never reaches the wire. `application/pdf\x00` now goes out as inline data carrying `"application/pdf\x00"` as its MIME type, where the previous head produced a placeholder.

I could not test what the backend does with a NUL inside that field, so I do not know whether this is better or worse than the placeholder was — the PDF does at least reach the model now. Flagging it so the asymmetry is a choice rather than an accident.

karolpiotrowicz (commented):
All four items from the last round are addressed, and I checked each by running it rather than reading the diff. Declared charsets now decode — `text/csv; charset=windows-1252`, `charset=windows-1251`, `charset=utf-16` and `charset=ISO-2022-JP` all come through as readable text where they were replacement characters before. A NUL in the label no longer drops readable JSON to a placeholder, and no longer lets SVG back onto the inline path. The unreachable inner `normalizeMIMEType` call is gone with the precondition documented, and the carve-out comment now says what the list actually does. Nothing here blocks merging.

I ran fifty artifact inputs through `ProcessRequest` at the merge-base, at [`1fa1de9`](https://github.com/google/adk-go/commit/1fa1de9546ef88f78b88c326e5216997c0d0a851) and at [`78547cd`](https://github.com/google/adk-go/commit/78547cd68ca097904492df358849e1c965a8a3e4). Sixteen moved between the last round's head and this one. Fifteen moved the right way.

## One decision I'd like you to make, not a defect

The sixteenth is the one worth your time, and it is a genuine design decision rather than something to fix. [`decodeArtifactText`](https://github.com/google/adk-go/blob/78547cd68ca097904492df358849e1c965a8a3e4/tool/loadartifactstool/load_artifacts_tool.go#L300-L311) consults the label before it looks at the bytes, so an artifact labelled `text/html; charset=iso-8859-1` that actually holds UTF-8 now arrives as `cafÃ©` where the last round delivered `café`. Stale legacy charset labels over UTF-8 payloads are common in the wild.

I wrote the obvious fix — prefer the bytes when they are already valid UTF-8 and carry a multi-byte sequence — and ran it. The package suite stays green and exactly one of the fifty inputs changes. The `!isASCII` half is necessary: without it, `charset=ISO-2022-JP` regresses from `日` back to the raw escape stream, because a 7-bit stateful encoding looks like plain ASCII.

But it is not a free improvement, and I would rather show you the trade than hand you a patch. A genuine windows-1252 file whose content is the two literal characters `Ã©` is the byte pair `C3 A9`, which is also valid UTF-8 for `é`:

```
                                                as it stands   with the bytes preferred
charset=windows-1252, bytes C3 A9               "Ã©"           "é"        <- wrong
charset=iso-8859-1, UTF-8 bytes for "café"      "cafÃ©"        "café"     <- right
charset=windows-1252, bytes 93 94               "“”"           "“”"       <- unchanged
```

Nothing in the input separates the first two cases. Honouring the declaration is what an HTTP client does, preferring the bytes is what a browser's sniffing step does, and the tool has to pick one. What is there now is a defensible choice, and I asked for the charset handling in the first place, so I am not going to ask you to reverse it on my say-so. If you want the byte-preferring rule instead, the guard above is what I ran and it is three lines.

## A dependency question for a maintainer

[`go.mod:103`](https://github.com/google/adk-go/blob/78547cd68ca097904492df358849e1c965a8a3e4/go.mod#L103) drops `// indirect` from `golang.org/x/text v0.40.0`. That is not a new dependency — same module, same version, already in the graph, and `go mod tidy` is content with the placement. Reaching `ianaindex` does link the charmap, japanese, korean, chinese and unicode tables, though. I measured it on a binary that imports only this package: 19,722,041 bytes before, 20,588,337 after, so **+866,296 bytes, +4.4%**.

`AGENTS.md` puts adding or upgrading a dependency under "Ask first". This literally does neither, and it is materially the thing that line is about, so it is a call for a maintainer rather than something you need to change.

## Parity with adk-python

One gap that is worth knowing about, though I would file it separately rather than hold this PR for it. adk-python treats an artifact as text-like when its *name* ends in `.csv`, `.txt`, `.json` or `.xml`, whatever the MIME type says — [`load_artifacts_tool.py:270-273`](https://github.com/google/adk-python/blob/c506ddf3bc34a6312ffc81899221bfa3f2da3b1d/src/google/adk/tools/load_artifacts_tool.py#L270-L273). This code looks only at the MIME type, so `report.csv` saved as `application/octet-stream` becomes `[Binary artifact: "report.csv", type: "application/octet-stream", size: 8 bytes. Content cannot be displayed inline.]` here and arrives as readable CSV in Python. `application/octet-stream` is a common upload default.

Worth recording the other direction too, since it bears on the decision above: adk-python does not honour `charset` at all. Its normalizer is documented as returning the type ["without parameters like charset"](https://github.com/google/adk-python/blob/c506ddf3bc34a6312ffc81899221bfa3f2da3b1d/src/google/adk/tools/load_artifacts_tool.py#L83) and its text path is `data.decode('utf-8')` with a `UnicodeDecodeError` fallback to `errors='replace'`. So the charset support here is a Go-only addition. That is fine by me — I asked for it — but it should be a deliberate divergence rather than an accidental one.

## Smaller things

The rest are on the lines they concern. The two most useful are both one test row each: the `enc != nil` guard is load-bearing and nothing currently fails if you delete it, and the `windows-1252` row uses a byte that three different encodings agree on.

Everything else holds up. The public surface is unchanged, `gofmt`, `go vet`, `golangci-lint` at the CI pin, `staticcheck` and `govulncheck` are all clean, the suite has no new failures, a concurrency run driving twenty-four artifacts through `ProcessRequest` is clean with a positive control confirming the check would have caught a race, and it merges into current main without conflicts.

tool/loadartifactstool/load_artifacts_tool.go
```diff
@@ -232,8 +236,112 @@ func (t *artifactsTool) loadIndividualArtifact(ctx context.Context, artifactsSer
...
+// parameter on the original MIME type is honoured when it names a known
+// encoding, so "text/csv; charset=windows-1252" round-trips instead of
+// degrading; otherwise invalid sequences are replaced rather than the
+// artifact discarded.
+func decodeArtifactText(data []byte, rawMIMEType string) string {
+	if _, params, err := mime.ParseMediaType(rawMIMEType); err == nil {
+		if cs := params["charset"]; cs != "" && !strings.EqualFold(cs, "utf-8") {
+			if enc, err := ianaindex.IANA.Encoding(cs); err == nil && enc != nil {
+				if decoded, err := enc.NewDecoder().Bytes(data); err == nil {
+					return string(decoded)
```
karolpiotrowicz, line 305: The fallback below asserts UTF-8 validity and this branch does not, so the invariant now depends on every reachable x/text decoder replacing ill-formed input rather than passing it through. I could not find a decoder that misbehaves, so this is hardening rather than a bug I can demonstrate.

Wrapping this in `strings.ToValidUTF8` is free on the happy path — it returns the input unmodified and unallocated when there is nothing invalid — and makes the guarantee unconditional.

tool/loadartifactstool/load_artifacts_tool_test.go
```diff
@@ -497,6 +497,243 @@ func TestLoadArtifactsTool_ProcessRequest_Artifacts_MultiPartFunctionResponse(t
...
+		{
+			name:         "svg application subtype is delivered as text",
+			artifactName: "chart.svg",
+			part:         genai.NewPartFromBytes([]byte("<svg id=\"c\"/>"), "application/svg+xml"),
+			wantText:     "<svg id=\"c\"/>",
+		},
+		{
+			name:         "declared charset is honoured when decoding",
+			artifactName: "sales2.csv",
+			part:         genai.NewPartFromBytes([]byte("caf\xe9,12\n"), "text/csv; charset=windows-1252"),
```
karolpiotrowicz, line 619: This row does not pin windows-1252. Byte `0xE9` maps to `é` under windows-1252, ISO-8859-1 and latin1 alike, so the assertion passes under any of the three and would survive `ianaindex` resolving the name differently.

`0x93` discriminates — it is `“` under windows-1252 and `\u0093`, a C1 control, under ISO-8859-1.

1wos:
Rebased onto main. The only conflict was the `golang.org/x/text` line in go.mod, where #1517 bumped the version on the same line this PR drops `// indirect` from. Took v0.41.0 without the marker.

Everything from the last round is in. The two unpinned branches have rows now: `charset=gb2312` for `enc != nil`, which panics without the guard, and a declared `utf-8` for the short-circuit. `0x93`/`0x94` replace `0xE9` in the windows-1252 row. The decoder branch is wrapped in `strings.ToValidUTF8`, control characters are stripped once at the call site and fed to both `normalizeMIMEType` and `mime.ParseMediaType`, and the `ianaindex` coverage is in the doc comment.

Thanks for writing the label-versus-bytes trade out as a table instead of handing me the patch. The three you left to me:

- The raw label on the inline path is deliberate now. The sanitized one ships, so `application/pdf\x00` reaches the model as `application/pdf`.
- Zero-width space deferred. RFC 2045 token validation fixes the class rather than another member of it, so it should be its own change.
- Keeping label over bytes. Honouring what the artifact declares is the more predictable rule.

Build, `go test -race -shuffle=on work` over 88 packages, `golangci-lint` v2.3.1 and `go mod tidy -diff` all clean in both modules.

On `x/text`: if the direct dependency is not wanted, say so and I will drop the charset handling. go.mod then leaves the PR and this is only the #867 fix.

PR #874, opened 2026-05-22
feat: adka2a structured error propagation

yarolegovich:
These was no way to propagate structured errors to clients using adka2a integration. The logic was relying only on `ErrorMessage`, meaning only text could be passed down to the caller.

This PR changes it such that `ErrorMessage` gets put as the first Status.Message text part to preserve backward compatibility and additionally all event content parts get appended to the list of parts. A metadata flag is introduced to mark this error and differentiate it from content parts. 

Now developers can implement an `AfterExecuteCallback` which will convert an error to a structured format and it will propagate in Status.Message to the original client.

### DIFF
--- a/server/adka2a/v2/events.go
+++ b/server/adka2a/v2/events.go
@@ -410,6 +410,6 @@ func convertStatusMessage(ctx agent.InvocationContext, event a2a.Event, status a
 	return &convertedStatusMessage{
 		errorMessage:       errMessage,
 		parts:              filterNilParts(convertedParts),
-		longRunningToolIDs: getLongRunningToolIDs(status.Message.Parts, convertedParts),
+		longRunningToolIDs: getLongRunningToolIDs(parts, convertedParts),
 	}, nil
 }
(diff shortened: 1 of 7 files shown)


baptmont (approved):

server/adka2a/v2/events.go
```diff
@@ -382,3 +381,35 @@ func filterNilParts(parts []*genai.Part) []*genai.Part {
...
+	}
+
+	convertedParts, err := convertParts(ctx, event, parts, partConverter)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert status message parts: %w", err)
+	}
+	return &convertedStatusMessage{
+		errorMessage:       errMessage,
+		parts:              filterNilParts(convertedParts),
+		longRunningToolIDs: getLongRunningToolIDs(status.Message.Parts, convertedParts),
```
baptmont: Should it be parts instead of  status.Message.Parts? ConvertedParts length could be status.Message.Parts-1

baptmont (changes requested):

server/adka2a/v2/events.go
yarolegovich: a failed task should not in theory be carrying long-running tool IDs, but yeah - should be parts. fixed

baptmont (approved):

PR #878, opened 2026-05-23
fix(configurable): prevent path traversal in AgentTool config_path resolution

adilburaksen:
## Summary

`ResolveAgentReference` in `configurable_utils.go` accepted absolute `refPath` values unconditionally and joined relative paths via `filepath.Join` without boundary enforcement. An attacker-controlled `config_path` field in an agent YAML could read arbitrary files accessible to the server process.

**Vulnerable pattern (before):**
```go
if !filepath.IsAbs(refPath) {
    targetPath = filepath.Join(filepath.Dir(parentPath), refPath)  // no ".." check
}
absPath, _ := filepath.Abs(targetPath)
a, err := FromConfig(ctx, absPath)  // reads file unconditionally
```

## Fix

- Reject absolute `refPath` values
- After resolving, verify the path stays within the parent agent's directory using `strings.HasPrefix` on the cleaned path

## Related

Same vulnerability exists in `adk-python` (PR: google/adk-python#5826) and `adk-java` (PR: google/adk-java#1218) — fix pattern is identical across all three SDKs.


adilburaksen:
@wolo-lab @hanorik — could one of you take a look at this when you have a moment, or route it to whoever owns `internal/configurable`?

This closes a path traversal in `ResolveAgentReference`: an attacker-supplied `config_path` is currently used verbatim when absolute, and joined without a containment check when relative, so a sub-agent reference such as `../../../../etc/passwd` is read from outside the parent agent's directory.

The same issue was fixed in the sibling implementations and both have landed:

- adk-python google/adk-python#5826 (merged, `171ae9e`)
- adk-java google/adk-java#1218 (merged, `08f4fdc6c9`)

adk-go is the last of the three still carrying the original pattern.

I have just pushed an update that tightens the check and adds tests:

- both sides of the comparison are now made absolute. The previous revision compared an absolute target against a parent directory that was only `filepath.Clean`ed, so a relative `parentPath` rejected legitimate in-directory references.
- symlinks are resolved with `filepath.EvalSymlinks` where the paths exist, so a symlink inside the agent directory can no longer resolve outside it. When a path does not exist the check falls back to the lexical path, so a missing file still reports as not found rather than as a traversal.
- added `configurable_utils_test.go` covering absolute paths, parent traversal, symlink escape, a legitimate in-directory reference, and a relative parent path. The symlink and relative-path cases fail against the previous revision of this PR and pass with the update.

Status: CLA signed, `check-changes` green, `go build`, `go vet`, `gofmt` and the package tests all clean locally, and the branch is rebased on current `main` — it still applies to `internal/configurable/configurable_utils.go` after the V2 restructure. No reviewer was ever assigned to the PR, which I suspect is why it has been sitting rather than any deliberate decision.

Happy to adjust the approach to whatever you prefer, or feel free to take the change over and land it yourselves.

wolo-lab:
Thank you for preparing a fix. @baptmont, please take a look at this PR.

baptmont (approved):

PR #880, opened 2026-05-24
Drop trailing stale function responses

### LINKED ISSUE #760: Stale function responses can hard-fail content processing after retries or session restarts

Haihan-Jiang:
Fixes #760.

## What changed
- When the latest event is an orphaned function response with no matching function call in history, drop only that trailing stale response instead of failing content construction.
- Preserve earlier valid user/model history so the next turn can continue after a retry, reconnect, or session restart.
- Add regression coverage for stale trailing function responses after a user message and after a model reply.

## Why
A replayed stale tool result can leave a dangling function response at the end of session history. Treating that as fatal makes otherwise valid session history unusable, even though the stale response can be safely ignored.

## Validation
- go test ./internal/llminternal
- go test ./internal/llminternal ./agent/llmagent

### DIFF
--- a/internal/llminternal/contents_processor.go
+++ b/internal/llminternal/contents_processor.go
@@ -295,10 +295,10 @@ SearchLoop: // A label to allow breaking out of the nested loop
 	}
 
 	if functionCallEventIdx == -1 {
-		return nil, fmt.Errorf(
-			"no function call event found for function responses ids: %v",
-			responseIDs,
-		)
+		// The latest response can be a stale client retry after a reconnect or
+		// session restart. Drop only that trailing orphan so earlier valid
+		// history remains usable.
+		return events[:len(events)-1], nil
 	}
 
 	// Collect function response events related to the matching call while
--- a/internal/llminternal/contents_processor_test.go
+++ b/internal/llminternal/contents_processor_test.go
@@ -1338,13 +1338,26 @@ func TestContentsRequestProcessor_Rearrange(t *testing.T) {
 			},
 		},
 		{
-			name: "Error on function response without matching call",
+			name: "Drop trailing orphaned function response after user message",
 			events: []*session.Event{
 				{Author: "user", LLMResponse: model.LLMResponse{Content: &genai.Content{Role: "user", Parts: []*genai.Part{{Text: "Regular message"}}}}},
 				{Author: "user", LLMResponse: model.LLMResponse{Content: &genai.Content{Role: "user", Parts: []*genai.Part{{FunctionResponse: frOrphaned}}}}},
 			},
-			want:    nil,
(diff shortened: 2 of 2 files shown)


Haihan-Jiang:
This is still ready for review from my side. It drops trailing stale function responses so callers do not receive obsolete tool outputs; could someone take a look when convenient?

karolpiotrowicz:
This proposes the same change as #766, which was opened earlier and already has review history on it, so I'd suggest we consolidate the discussion there and close this one.

The underlying report is real and still reproduces on `main`. The review on #766 has the details, and the short version is that dropping the trailing *event* also discards any text part sharing that event with the stale response, which can be the user's current message. Both PRs have that in common, so this is not a reason to prefer one over the other.

Happy to have you weigh in on #766 if you want to keep working on it.

PR #882, opened 2026-05-24
fix: handle unsafe artifact MIME types

### LINKED ISSUE #867: LoadArtifactsTool should handle unsupported artifact MIME types gracefully

Haihan-Jiang:
Fixes #867.

## What changed
- Convert text-like loaded artifact inline data (`text/*`, `application/csv`, `application/json`, `application/xml`) to text before appending it to the next model request.
- Preserve inline media MIME types Gemini accepts directly: images, audio, video, and PDF.
- Replace unsupported binary inline artifacts with a short placeholder containing the artifact name, MIME type, and byte size.
- Handle nil/empty artifact content with explicit placeholder text instead of forwarding unsafe inline parts.

## Why
`LoadArtifactsTool` previously appended loaded artifact parts directly into the next model request. Unsupported inline MIME types such as CSV, DOCX, or PPTX could make the following `generateContent` call fail with an unsupported MIME type error.

## Validation
- `go test ./tool/loadartifactstool`
- `go test ./tool/loadartifactstool ./artifact ./internal/artifact/tests`
- `git diff --check`

### DIFF
--- a/tool/loadartifactstool/load_artifacts_tool.go
+++ b/tool/loadartifactstool/load_artifacts_tool.go
@@ -21,6 +21,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
+	"unicode/utf8"
 
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/genai"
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is ready for review. The artifact handling now rejects unsafe MIME types, and the visible checks are green.

karolpiotrowicz:
We're taking [#866](https://github.com/google/adk-go/pull/866) forward for #867, and the deciding difference is MIME parameter handling rather than anything about the overall approach — your classification, your placeholder shape and your empty-content handling all match what the issue asked for. Because [`safeArtifactPart`](https://github.com/google/adk-go/blob/e5bd4536f4153edf42faf624d3f45d6c392c6d2c/tool/loadartifactstool/load_artifacts_tool.go#L243-L269) lowercases and trims the MIME type but does not strip the `;`-parameter, the exact-match arms miss the spellings an HTTP `Content-Type` actually carries: `application/csv; charset=utf-8` becomes a placeholder instead of the CSV text, and `application/pdf; version=1.7` is discarded even though `main` forwards those bytes intact today and PDF is on your own keep-inline list.

Two things in your version are better than what we're merging, and both should be carried across. You format the artifact name and MIME type with `%q`, so a MIME type containing newlines cannot break out of the placeholder's framing — #866 uses `%s` and it can, which we've raised there and credited to this PR. Your placeholder assertions also use substring matching rather than exact strings, which is the more durable way to test text that will get reworded.

Thank you for this, and I'm sorry it sat as long as it did — three months without a decision is on us, not on the contribution.

PR #887, opened 2026-05-25
feat: Add OnPipelineErrorCallback lifecycle hook to Plugin Config

### LINKED ISSUE #886: feat: Add OnPipelineErrorCallback lifecycle hook to Plugin Config
# Feature Request: `OnPipelineErrorCallback` Lifecycle Hook
---

Currently, the `plugin.Config` struct provides precise, granular hooks for structural runtime failures like `OnModelErrorCallback` and `OnToolErrorCallback`. However, there is a visibility and control gap for errors encountered during orchestrator lifecycles.

If an error occurs during runtime execution setups-such as within an `OnUserMessageCallback` or a `BeforeRunCallback`-the underlying framework immediately propagates the error up the call stack, halting execution. Independent plugin developers lack a native mechanism to intercept these orchestrator-level failures to perform telemetry logging, clean up transient states, or implement custom retry/fallback behaviors globally for the agent pipeline.

---

## Proposed Solution

Introduce an `OnPipelineErrorCallback` lifecycle hook to the `plugin.Config` and `plugin.Plugin` structures. This callback will capture any pipeline-level error that escapes granular tool/model loops, allowing plugins to inspect the error context, log telemetry data, gracefully transform or wrap the error before propagation, or return a fallback response payload.

---

## Impact on your work

This feature significantly improves the design of enterprise-grade telemetry and resilience plugins. Without it, our global error tracking wrappers cannot safely intercept lifecycle anomalies without risking unhandled pipeline failures or losing context from the `agent.InvocationContext`.

---

### Alternatives Considered

Writing custom boilerplate code inside every individual agent implementation to wrap the runner call in a `recover()` or manual try-catch style pattern defeats the modularity and design pattern of the ADK plugin infrastructure.

---

### Willingness to contribute

Yes, I am interested in implementing this feature and submitting a PR.

---

### 1. Extend `plugin.Config` and `plugin.Plugin`

```go
type OnPipelineErrorCallback func(agent.InvocationContext, error) error

type Config struct {
	Name string

	// Existing fields...
	OnUserMessageCallback OnUserMessageCallback
	BeforeRunCallback     BeforeRunCallback

	// OnPipelineErrorCallback intercepts errors occurring during orchestration lifecycles.
	OnPipelineErrorCallback OnPipelineErrorCallback
}
```

---

### 2. Expose an accessor on the `Plugin` struct

```go
func (p *Plugin) OnPipelineErrorCallback() OnPipelineErrorCallback {
	return p.onPipelineErrorCallback
}
```

---

### 3. Integrate into Runtime Orchestration Engine

The runner component executing the lifecycle callbacks will pass unresolved lifecycle errors into the active plugin stack’s `OnPipelineErrorCallback` loop before aborting execution entirely.

---

## Additional Context

Completes the parity lifecycle symmetry alongside `OnModelError` and `OnToolError` blocks already present in `plugin.go`.

mramansayyad:
# Pull Request: Add `OnPipelineErrorCallback` lifecycle hook to Plugin Config

## Description
Fixes #886

This PR introduces the `OnPipelineErrorCallback` lifecycle hook to `plugin.Config` and `plugin.Plugin` structures. 

Currently, `plugin.Config` supports precise model-level (`OnModelErrorCallback`) and tool-level (`OnToolErrorCallback`) failure hooks, but lacks a mechanism to intercept and handle orchestrator/pipeline-level failures. If an error is encountered during the runner lifecycle (e.g., within `OnUserMessageCallback` or `BeforeRunCallback`), the runner immediately aborts execution. 

With `OnPipelineErrorCallback`, plugins can now intercept, log, or transform pipeline-level errors before they propagate up the call stack, enabling custom global error tracking, retry policies, or fallback behavior.

## Proposed API / Design

1. **Callback Signature**:
   ```go
   type OnPipelineErrorCallback func(agent.InvocationContext, error) error
   ```
2. **Extended `plugin.Config` & `plugin.Plugin`**:
   Added `OnPipelineErrorCallback` fields and constructor mappings, along with the `OnPipelineErrorCallback()` accessor on `plugin.Plugin`.
3. **Plugin Manager Execution**:
   Added `RunOnPipelineErrorCallback` in `PluginManager` to run the active plugin stack's error hooks sequentially:
   ```go
   func (pm *PluginManager) RunOnPipelineErrorCallback(cctx agent.InvocationContext, err error) error {
       for _, plugin := range pm.plugins {
           callback := plugin.OnPipelineErrorCallback()
           if callback != nil {
               err = callback(cctx, err)
           }
       }
       return err
   }
   ```
4. **Orchestrator Integration**:
   Intercepted pipeline errors in `Run` and `RunLive` loops in `runner/runner.go` and fed them to the active plugin manager before yielding the final error or exiting.

---

## Telemetry Test Fix
Fixed a flaky assertion in `server/adkrest/internal/services/debugtelemetry_test.go` where expected spans list was compared without sorting, causing non-deterministic map iteration order failures. Added `cmpopts.SortSlices` sorting by span `Name` to ensure stable green builds.

---

## Verification & Testing
Added new subtests under `TestRunner_OnPipelineErrorCallback` in `runner/runner_test.go`:
* **BeforeRunCallback Error**: Validates that errors raised in `BeforeRunCallback` trigger the hook and are properly propagated/transformed.
* **OnUserMessageCallback Error**: Validates that errors raised during message session appends trigger the hook and are properly propagated/transformed.
* **Agent Run Loop Error**: Validates that errors returning from the inner agent execution loop trigger the hook and are properly propagated/transformed.

Ran all test packages:
```bash
go test -v ./runner/...
go test ./...
```
Result: **PASS** (100% green).


mramansayyad:
I signed it.

mramansayyad:
cc @anFatum @wolo-lab could you please take a look and review this PR?

mramansayyad:
@indurireddy-TF @dpasiukevich  @wolo-lab  please anyone could take a look and review this PR?

mramansayyad:
This PR is fully implemented, verified with comprehensive tests, and has no merge conflicts. Requesting review and merge. Thank you!

jjsasha63:
Thanks for this, and sorry for the slow response. The gap is real: adk-go has no runner-level error hook, and adk-python has one. I would like this feature, so I am not closing the PR — but the API shape needs to change first, and that is worth settling before you rebase 143 commits.

### The shape

Upstream calls this `on_run_error_callback`, and it is deliberately notification-only:

> This is a notification-only callback. The exception is always re-raised after all registered plugins have been notified. Plugins should NOT suppress the exception.

`on_agent_error_callback` is the same. Only the model and tool error hooks may substitute a value, because at those levels there is a well-defined thing to substitute — an `LlmResponse`, a tool result dict. At runner level there is not.

This PR returns `error`, so returning `nil` swallows the pipeline error. Two of the eight integration sites do not re-check the error afterwards, and that is where it breaks:

- `Run`, the `appendMessageToSession` path, calls `yield(nil, err)` unguarded. A plugin that returns `nil` makes the runner emit a `(nil event, nil error)` pair to the consumer.
- `RunLive`, the `lAgent.RunLive` path, does `return nil, nil, err` unguarded. A plugin that returns `nil` hands the caller a nil `LiveSession` and a nil iterator with no error to signal it. The caller panics on first use.

I reproduced both against this branch. The other six sites guard correctly, so it is an inconsistency rather than a design you chose — but it is the natural failure mode of allowing suppression at all.

What I would like instead, matching upstream and removing both crash paths by construction:

```go
// signature matching its siblings, which currently take agent.InvocationContext
type OnRunErrorCallback func(agent.InvocationContext, error)
```

If you have a concrete use case that needs to transform or suppress the error rather than just observe it, please describe it on #886 and let us work it out there before you write more code.

### On the tests

These are good, and I want to say so — they assert on the callback firing, the error received, and the error propagated. I stubbed the manager to never invoke the callbacks and all three subtests failed with specific messages, which is more than many PRs manage.

Coverage to add once the shape is settled: none of the four `RunLive` sites are exercised, and there is no test with two plugins registered. With the current chaining loop, a second plugin can be invoked with a nil error after the first one swallows it.

### When you rebase

- The `debugtelemetry_test.go` sort fix has already landed on main as `cmpopts.SortSlices(compareDebugSpans)`, so drop that hunk.
- `plugin/plugin.go` needs gofmt — the struct literal in `New` is misaligned.
- The branch is 143 commits behind: the module path moved to `/v2` and `appendMessageToSession` now returns three values.

PR #890, opened 2026-05-25
feat: Add conformance test recording plugin

baptmont:
Python ref: https://github.com/google/adk-python/blob/7ad7994744de18f2394e4bcb961cd5c7a24afb4b/src/google/adk/cli/plugins/recordings_plugin.py#L66


internal/configurable/conformance/recordplugin/record_plugin.go
```diff
@@ -0,0 +1,625 @@
...
+		"tool_use_prompt_tokens_details", "total_token_count", "traffic_type",
+	}
+
+	for _, key := range snakeCaseKeys {
+		cleaned := strings.ReplaceAll(key, "_", "")
+		snakeCaseFieldMap[cleaned] = key
+	}
+
+	// Manual edge cases
+	snakeCaseFieldMap["propertyordering"] = "property_order"
```
baptmont, line 437: Let me know if you know a better way to do this

wolo-lab: I don't see a better way, we end up with having this define in one place or another.

wolo-lab (approved):

PR #891, opened 2026-05-25
feat(telemetry): Instrument nodes and workflows

RKest:
### Instrument workflows and nodes

Instruments V2 `Workflows` and `Nodes`.

Prior to this change, running:
```sh
go run ./examples/workflow/basic console -otel_to_cloud
```
produced no telemetry.

After this change, the following trace data is emitted:
image

### Testing Plan

**Unit Tests:**

1. Run the basic node example
```sh
export GOOGLE_CLOUD_PROJECT=your_gcp_project
go run ./examples/workflow/basic console -otel_to_cloud
```
2. Ctrl+C to flush the OpenTelemetry spans
3. Navigate to https://pantheon.corp.google.com/traces/explorer
4. Observe saved trace

### Checklist

### Additional context

### DIFF
--- a/internal/telemetry/node_tracing.go
+++ b/internal/telemetry/node_tracing.go
@@ -1,0 +1,148 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package telemetry
+
+import (
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/otel/attribute"
+	semconv "go.opentelemetry.io/otel/semconv/v1.36.0"
+	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
+
+	agentinternal "google.golang.org/adk/internal/agent"
+	"google.golang.org/adk/session"
+)
+
+// gen_ai.operation.name attribute values for the agent / workflow /
+// node spans. These mirror the values set by adk-python's
+// node_tracing module.
+const (
+	invokeWorkflowOperationName = "invoke_workflow"
+	invokeNodeOperationName     = "invoke_node"
+)
+
+// Attribute keys not yet in the Go semconv package; the string
+// literals match the keys used by adk-python.
+var (
+	genAIWorkflowName = attribute.Key("gen_ai.workflow.name")
+	genAINodeName     = attribute.Key("gen_ai.node.name")
+)
+
+// Operation is the internal sum-type which describes what can be used to start a span.
+type Operation interface {
+	isOperation()
+}
+
+// AgentLike is the minimal contract telemetry needs to describe a BaseAgent.
+type AgentLike interface {
+	Name() string
+	Description() string
+}
+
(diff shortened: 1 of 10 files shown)


workflow/node_context.go
```diff
@@ -56,3 +60,16 @@ func (c *nodeContext) ResumedInput(interruptID string) (any, bool) {
 	v, ok := c.resumeInputs[interruptID]
 	return v, ok
 }
+
+// WithContext preserves the nodeContext wrapper when callers derive
+// a new context from this one (e.g. when the scheduler attaches an
+// OpenTelemetry span context). Without this override, the base
```
wolo-lab, line 128: To prevent regression in the future (while adding new fields to node_context), let use positional literals:

```golang
	// Positional literal on purpose: forces a compile error when
	// a new field is added to nodeContext, so the author must
	// explicitly decide whether to propagate it here.
	return &nodeContext{
		c.InvocationContext.WithContext(ctx),
		c.resumeInputs,
	}
```

internal/telemetry/node_tracing.go
```diff
@@ -0,0 +1,149 @@
...
+// concrete span name (invoke_agent vs invoke_workflow) is decided
+// inside [StartNodeSpan] by peeking at the agent's internal
+// State.AgentType.
+type NodeAgent struct {
+	Agent AgentLike
+}
+
+func (NodeAgent) isNode() {}
+
+// NodeLike is the minimal contract telemetry needs to describe a BaseNode.
```
wolo-lab: NodeLike and AgentLike have the same fields: name and description. Why we need both of them?

workflow/node_context.go
RKest: Good call, updated

internal/telemetry/node_tracing.go
RKest: I was overcautious with types. Consolidated into a single `NodeLike`

internal/telemetry/node_tracing.go
```diff
@@ -0,0 +1,144 @@
...
+type NodeLike interface {
+	Name() string
+	Description() string
+}
+
+// NodeAgent is the [Node] variant for an agent invocation.
+// concrete span name (invoke_agent vs invoke_workflow) is decided
+// inside [StartNodeSpan] by peeking at the agent's internal
+// State.AgentType.
+type NodeAgent struct {
```
wolo-lab: Thanks for addressing my previous comment. I see one more potential improvement.

`NodeNode` and `NodeAgent` are basically the same; both only have a `NodeLike` field. Let's think how we could simplify it.

btw: `NodeNode` sounds just bad.

I noticed that Description is only required by Agent, not by Workflow and NodeNode. We could have separate interfaces:

```go
type AgentLike interface {
	Name() string
	Description() string
}

type WorkflowNodeLike interface {
	Name() string
}

```

I was thinking how to avoid having `Node` which collides a bit with workflow node, I see that in the telemetry you use `Operation` term, maybe it is a good candidate. Then we would have:

```go
type Operation interface {
	isOperation()
}

type OperationAgent struct{ Agent AgentLike }

type OperationNode struct{ Node WorkflowNodeLike }

func (OperationAgent) isOperation() {}
func (OperationNode) isOperation()  {}

func StartNodeSpan(ctx context.Context, ictx InvocationContext, **op Operation**) (...)
```

One small note to pre-empt the "but they're still structurally identical" objection: even though `OperationAgent` and `OperationNode` carry just one field each, the distinct *types* of those fields (`AgentLike` vs `WorkflowNodeLike`) make the sum type meaningful — the variant isn't just a label, it carries different capabilities. 

If you don't like `Operation` term, we could have something different, but separate interfaces would be still a nice improvement.

The rest would remain unchanged, WDYT?

wolo-lab (approved):
Approved with small readability improvement to consider.

internal/telemetry/node_tracing.go
RKest: Done, great suggestion

mhenc (approved):

PR #893, opened 2026-05-26
docs: Explain adka2a handler wiring

### LINKED ISSUE #419: adka2a should provide a http.Handler
Similar to the adkrest package (https://github.com/google/adk-go/issues/415), adka2a should provide a handler that serves A2A. Instead, it seems to be providing an Executor.

cc @dpasiukevich

Haihan-Jiang:
Adds package docs for the legacy and v2 adka2a packages explaining that NewExecutor returns an a2a-go AgentExecutor, which should be wrapped with a2asrv.NewHandler and the transport-specific HTTP handler.

Also adds a compile-checked JSON-RPC example showing the agent card endpoint and invoke endpoint wiring without adding new handler APIs to ADK.

Tests:
- go test ./server/adka2a/... ./examples/a2a ./cmd/launcher/web/a2a

Fixes #419.


Haihan-Jiang:
This is ready for review. The adka2a handler wiring docs are clarified, and the visible checks are green.

baptmont (approved):

PR #894, opened 2026-05-26
test(telemetry): add instrumentation functional tests 

RKest:
### Add instrumentation functional tests 

- Related: ADK-Python parity with `tests/unittests/telemetry/test_functional.py`

**Problem:**

ADK-Go has unit tests for individual telemetry helpers (e.g. `LogRequest`,
`StartGenerateContentSpan`), but no end-to-end coverage that asserts the
**shape of the span+log tree** emitted by a realistic agent run. Regressions
that change span ordering, attribute names, parent/child relationships, or
log-event bodies slip past unit tests because each unit test only exercises
a single helper in isolation.

Additionally, the existing `telemetry.WithGenAICaptureMessageContent` option
duplicated the OpenTelemetry-spec env var
`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` and stored the result
in a package-level `atomic.Bool`. That additional global state made hermetic
per-test toggling difficult, and it would also complicate future MT telemetry support.

**Solution:**

Two changes, landed together because they share infrastructure:

1. **Functional test infrastructure** under `internal/telemetry/`:
   - `functionaltest/` — the end-to-end test package. Tests here install
     in-memory span/log exporters, run a real agent through a canned
     scenario, then diff the captured spans+logs against an expected
     fixture using `cmp.Diff`.
   - `telemetrytest/` — reusable helpers: in-memory log exporter,
     span/log "digest" converters that strip non-deterministic fields
     (IDs, timestamps) so fixtures are stable, and a scenario runner.
   - `telemetrytestcase/` — expected-output fixtures keyed per scenario,
     kept in their own package so they can be shared across functional
     tests without import cycles.

2. **Agent functional test** (`TestTelemetrySchema_AgentWithTool`):
   The canonical "llmagent with one `FunctionTool`" scenario, parametrized
   over `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` to cover both
   the elided (production default) and full-content variants. Hermetic:
   no network calls, no live model, no GCP. Mirrors `test_telemetry_schema`
   in `adk-python/tests/unittests/telemetry/test_functional.py`.

Supporting changes inside `internal/telemetry/`:

- Adds `OverrideTracerForTesting` and `OverrideLoggerForTesting` so a test
  can swap in an in-memory provider for its duration via `t.Cleanup`.
- Replaces the `atomic.Bool` + `SetGenAICaptureMessageContent` pair with
  a direct env-var read on every `getGenAICaptureMessageContent` call.
  Tests now flip behaviour via `t.Setenv`; the OS provides the caching.
- Removes the public `telemetry.WithGenAICaptureMessageContent` option
  (and the now-unused `genAICaptureMessageContent` field on `Providers`
  / `config`). Users who were relying on it should set
  `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true` in the
  environment instead — this is the OTel-spec way of configuring it.

This PR is the **parent** of a follow-up that adds node/workflow telemetry
instrumentation; that PR reuses the same `functionaltest` / `telemetrytest`
infrastructure for its own graph-based scenario.

### Testing Plan

**Unit Tests:**

New tests:
- `TestTelemetrySchema_AgentWithTool/elided`
- `TestTelemetrySchema_AgentWithTool/capture_content`

Updated tests:
- `internal/telemetry/logger_test.go` — `setup()` now toggles the env var
  via `t.Setenv` instead of calling the removed `SetGenAICaptureMessageContent`.

```
$ go test ./internal/telemetry/... ./telemetry/... -count=1
ok      google.golang.org/adk/internal/telemetry                0.078s
ok      google.golang.org/adk/internal/telemetry/functionaltest 0.087s
?       google.golang.org/adk/internal/telemetry/telemetrytest  [no test files]
?       google.golang.org/adk/internal/telemetry/telemetrytestcase [no test files]
ok      google.golang.org/adk/telemetry                         0.114s
```

Not applicable — this PR only adds test infrastructure and a hermetic
functional test; there is no runtime-visible behaviour change other than
the removal of `WithGenAICaptureMessageContent` (see "Breaking change"
below).

### Checklist

### Additional context

**Breaking change:** `telemetry.WithGenAICaptureMessageContent` is removed.
The replacement is the OTel-spec env var
`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true`. The option had
no other call sites in this repository besides one example
(`examples/tools/multipletools`), which has a pre-existing build break
that should be cleaned up in a separate PR.

### DIFF
--- a/internal/telemetry/logger.go
+++ b/internal/telemetry/logger.go
@@ -17,8 +17,8 @@ package telemetry
 import (
 	"context"
 	"encoding/json"
+	"os"
 	"strings"
-	"sync/atomic"
 
 	"go.opentelemetry.io/otel/log"
 	"go.opentelemetry.io/otel/log/global"
@@ -29,17 +29,17 @@ import (
 	"google.golang.org/adk/model"
 )
 
-// genAICaptureMessageContent is true if message content should be elided. False by default.
-var genAICaptureMessageContent atomic.Bool
+// captureMessageContentEnvVar is the OpenTelemetry-spec env var
+// that controls whether ADK emits full message content in log
+// records (true) or elides it for privacy (anything else,
+// including unset). Defined by
+// https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/.
+const captureMessageContentEnvVar = "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT"
 
-// SetGenAICaptureMessageContent sets whether message content should be elided.
-func SetGenAICaptureMessageContent(capture bool) {
-	genAICaptureMessageContent.Store(capture)
-}
-
-// getGenAICaptureMessageContent returns whether message content should be elided.
+// getGenAICaptureMessageContent reports whether message content
+// should be captured in log records.
 func getGenAICaptureMessageContent() bool {
-	return genAICaptureMessageContent.Load()
+	return strings.TrimSpace(os.Getenv(captureMessageContentEnvVar)) == "true"
 }
 
 const elidedContent = "<elided>"
@@ -49,6 +49,19 @@ var otelLogger = global.GetLoggerProvider().Logger(
 	log.WithSchemaURL(semconv.SchemaURL),
 	log.WithInstrumentationVersion(version.Version),
 )
+
+// OverrideLoggerForTesting replaces the package-level otelLogger
+// with one derived from lp for the duration of the calling test.
+// The original logger is restored via t.Cleanup.
+func OverrideLoggerForTesting(t interface{ Cleanup(func()) }, lp log.LoggerProvider) {
+	original := otelLogger
+	otelLogger = lp.Logger(
+		systemName,
+		log.WithSchemaURL(semconv.SchemaURL),
+		log.WithInstrumentationVersion(version.Version),
+	)
+	t.Cleanup(func() { otelLogger = original })
+}
 
 // LogRequest logs the request to the model - the system message and user messages.
 // It iterates over the request contents and logs each as a separate event.
--- a/internal/telemetry/telemetrytest/log_digest.go
+++ b/internal/telemetry/telemetrytest/log_digest.go
@@ -1,0 +1,138 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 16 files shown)


internal/telemetry/logger.go
```diff
@@ -29,17 +29,19 @@ import (
...
+// that controls whether ADK emits full message content in log
+// records (true) or elides it for privacy (anything else,
+// including unset). Defined by
+// https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/.
+const captureMessageContentEnvVar = "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT"
+
+// getGenAICaptureMessageContent reports whether message content
+// should be captured in log records. It reads the env var on
+// every call so tests can flip the behaviour via [testing.T.Setenv]
+// without any in-memory flag plumbing; the OS provides the caching.
```
wolo-lab: nit: "the OS provide the caching" - I don't think it is true, I would remove it. Unless you can provide some explanations that it really happens.

internal/telemetry/telemetrytest/log_digest.go
```diff
@@ -0,0 +1,138 @@
...
+	"go.opentelemetry.io/otel/log"
+	sdklog "go.opentelemetry.io/otel/sdk/log"
+)
+
+// LogDigest is a deterministic snapshot of one OTel log record:
+// the event name, attributes, and the body decoded to a Go value.
+type LogDigest struct {
+	EventName  string
+	Attributes map[string]any
+	Body       any
```
wolo-lab: Just wondering:
Maybe we could add a comment to `Body any` : `// JSON-normalized: int64 values are returned as float64` .
Maybe also a comment to `Attributes` that here we don't convert types (side note: maybe we should also have `jsonRountTrip(attrs).(map[string]any)` in line `43`?

wolo-lab (approved):

internal/telemetry/logger.go
RKest: Done

internal/telemetry/telemetrytest/log_digest.go
RKest: Done for the `Body`

For attributes, OTel has it's own sum types for attributes. When these are emitted to OTel SDK and down the line sent over the wire, they are either converted to JSON or Proto. Unlike `Body` there's no need to JSON normalize these, because unlike `Body`, these are not always serialized as JSON.

PR #896, opened 2026-05-26
feat(workflow): add dynamic-node orchestration

wolo-lab:
Wires the user-facing API for dynamic workflows on top of the
sub-scheduler skeleton from the previous PR. A dynamic node's
execution order is expressed as Go code (loops, branches, goroutines)
that calls other nodes inline via `RunNode`, branches on their typed
output, and pauses for HITL input.

The public surface:

- `workflow.NewDynamicNode[IN, OUT](name, fn, cfg)` — orchestrator
  constructor; `cfg.RerunOnResume` defaults to `&true` (an explicit
  `&false` is respected).
- `workflow.RunNode[OUT](ctx, child, input, opts...)` — generic
  helper for scheduling a child. Returns its typed output, or
  `errors.Is`-matchable `ErrNodeInterrupted` / `ErrNodeFailed`.
- `workflow.WithRunID(id)` — option overriding the auto-counter
  with a stable id (rejected if empty, purely numeric, or
  containing `/` or `@`).
- `session.NodeInfo` — substruct on `Event` carrying the emitting
  node's composite path; shape mirrors adk-python's `event.nodeInfo`.

The scheduler's `handleEvent` scopes per-activation `Output`/`Routes`
invariants by `NodeInfo.Path`, so a dynamic node forwarding a child's
terminal output plus its own no longer trips `ErrMultipleOutputs`.
Descendant `RequestedInput` events are promoted onto the parent's
accumulator so `Workflow.Resume` matches the InterruptID against the
parent's `NodeState.PendingRequest` — enabling HITL inside a dynamic
orchestrator.

Stacked on the sub-scheduler skeleton PR; follow-ups add
resume/replay-skip, broader parent re-entry scenarios, and parallel
HITL detection.

Tested: `go build ./...`, `go test -race ./workflow/...
./agent/workflowagent/...`, `gofmt -l` clean. Coverage includes
constructor defaults and overrides, sequential `RunNode` chains,
typed-input coercion (direct + JSON fallback), mid-body emit, HITL
swallow, error propagation, terminal output, plus end-to-end tests
for child+parent output forwarding and HITL round-trip via
`Workflow.Run` + `Workflow.Resume`.

### DIFF
--- a/workflow/dynamic_node.go
+++ b/workflow/dynamic_node.go
@@ -1,0 +1,183 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package workflow
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"iter"
+
+	"github.com/google/jsonschema-go/jsonschema"
+	"google.golang.org/adk/agent"
+	"google.golang.org/adk/internal/typeutil"
+	"google.golang.org/adk/session"
+)
+
+// DynamicFn is the orchestrator body of a dynamic node. emit publishes
+// mid-body events (state updates, HITL requests, progress); the return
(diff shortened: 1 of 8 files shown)


workflow/dynamic_node.go
```diff
@@ -0,0 +1,184 @@
...
+				return
+			}
+			yield(nil, err)
+			return
+		}
+
+		ev := session.NewEvent(parentNC.InvocationID())
+		ev.Output = out
+		ev.NodeInfo = &session.NodeInfo{Path: sub.parentPath}
+		ev.Actions.StateDelta["output"] = out
```
hanorik: Do we need for some reason to store the output in the `StateDelta`?

wolo-lab: Good catch! Removed.

hanorik (approved):

PR #898, opened 2026-05-27
fix: preserve interleaved function events

### LINKED ISSUE #761: Function response rearrangement can drop unrelated tool events from history
**Describe the bug**

`rearrangeEventsForLatestFunctionResponse` in `internal/llminternal/contents_processor.go` rebuilds history when the latest event is a function response. After finding the matching function-call event, it collects related function-response events to merge, then returns only:

```go
resultEvents := events[:functionCallEventIdx+1]
// ... append merged response event
return resultEvents, nil
```

That drops intermediate events between the matching function-call event and the latest response unless they are part of the merged response set. This can remove unrelated function-call events and unrelated function-response events from the replayed history.

**To Reproduce**

Use a history where a long-running tool response arrives after unrelated tool activity, for example:

1. User asks the agent to start long-running tool A.
2. Model emits function call A.
3. User/tool returns an intermediate response for A.
4. Model emits an unrelated function call B.
5. User/tool returns a response for B.
6. User/tool later returns the final response for A.
7. `ContentsRequestProcessor` calls `rearrangeEventsForLatestFunctionResponse` because the latest event is a function response for A.

The rearrangement keeps the original function-call event for A and the merged final response for A, but drops the unrelated function-call/function-response pair for B from the resulting history.

**Expected behavior**

When rearranging latest function responses, ADK Go should merge only response events related to the matching function-call event. Unrelated intermediate function-call events and unrelated function-response events should remain in the reconstructed history so replayed contents do not lose valid tool history.

**Observed behavior**

Unrelated function events between the matching call and the latest response can be dropped during rearrangement.

**Proposed fix**

While scanning events between the matching function-call event and the latest response:

- keep intermediate `FunctionCall` events in the result history
- keep unrelated intermediate `FunctionResponse` events in the result history
- merge only response events related to the matching call event into the latest response batch

This preserves valid unrelated tool history while still consolidating the long-running response sequence that triggered rearrangement.

**Testing**

Add content processor tests for an interleaved history where:

- tool A has an intermediate response
- unrelated tool B call/response occurs before A's final response
- A's final response triggers latest-response rearrangement
- the expected contents preserve B's call/response and include A's merged final response

**Alignment with adk-python**

This matches the cross-ADK history invariant that rearrangement should repair pairing for the target long-running function response without dropping unrelated valid events from the conversation history.

Haihan-Jiang:
Fixes #761

## Summary
- preserve unrelated tool call and response events while rearranging the latest function response
- split related response parts from unrelated tool parts before merging the long-running response batch
- add a regression test for an interleaved long-running response with unrelated tool activity

## Testing
- go test ./internal/llminternal -count=1
- git diff --check

### DIFF
--- a/internal/llminternal/contents_processor.go
+++ b/internal/llminternal/contents_processor.go
@@ -272,48 +272,86 @@ SearchLoop: // A label to allow breaking out of the nested loop
 		)
 	}
 
-	// Collect function response events related to the matching call while
-	// preserving unrelated tool events that happened in between.
+	// Collect all related function response events *between* the call and the
+	// last response. Preserve unrelated tool events in their original order, so
+	// rearranging a long-running response does not drop valid concurrent tool
+	// history.
 	var responseEventsToMerge []*session.Event
-	resultEvents := events[:functionCallEventIdx+1]
+	var eventsToPreserve []*session.Event
 	for i := functionCallEventIdx + 1; i < len(events)-1; i++ {
 		event := events[i]
-		calls := utils.FunctionCalls(event.Content)
-		if len(calls) > 0 {
-			resultEvents = append(resultEvents, event)
-			continue
+		mergeEvent, preserveEvent := splitEventForLatestFunctionResponse(event, responseIDs)
+		if mergeEvent != nil {
+			responseEventsToMerge = append(responseEventsToMerge, mergeEvent)
 		}
-
-		responses := utils.FunctionResponses(event.Content)
-		if len(responses) == 0 {
-			continue
-		}
-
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is ready for review. The function event handling now preserves interleaved events, and the visible checks are green.

karolpiotrowicz:
Thanks a lot for taking the time to submit this pull request!

The underlying issue, #761, has already been resolved in another update, so we will go ahead and close this pull request. We really appreciate your work and hope to see more of your PRs in the future!

PR #904, opened 2026-05-28
fix: preserve vertex ai artifact deltas

### LINKED ISSUE #902: Vertex AI session service drops `Actions.ArtifactDelta` during `AppendEvent` / `ListEvents` round-trip
## 🔴 Required Information

**Describe the Bug:**

`appendEvent` and `listSessionEvents` in `session/vertexai/vertexai_client.go`
serialize only `session.EventActions.StateDelta` — `Actions.ArtifactDelta` is
silently dropped on round-trip. Any tool that saves an artifact via
`toolCtx.Artifacts().Save(...)` ends up with an in-memory event whose
`Actions.ArtifactDelta` is populated, but after the runner persists the event
via the Vertex AI session service and a later `Get` / `ListEvents` reads it
back, `Actions.ArtifactDelta` is `nil`.

The wire format already supports this: `aiplatformpb.EventActions` has an
`ArtifactDelta` field (proto field 3, `map[string]int32`) that's currently
neither written by `appendEvent` nor read by `listSessionEvents`.

**Steps to Reproduce:**

1. Configure a runner with `session/vertexai`.
2. Register a tool that saves an artifact via
   `toolCtx.Artifacts().Save(name, part)`.
3. Send a message that triggers the tool.
4. Allow the runner to call `sessionService.AppendEvent` for the tool's
   function-response event (the in-memory event has
   `Actions.ArtifactDelta[name] = version` populated at this point).
5. Either start a new invocation on the same session, or call `Get` /
   `ListEvents` for the session.
6. Observe that every persisted event comes back with
   `Actions.ArtifactDelta == nil`.

**Expected Behavior:**

`Actions.ArtifactDelta` round-trips through the Vertex AI session service:
the value read back from `Get` / `ListEvents` deep-equals the value that was
persisted via `AppendEvent`.

**Observed Behavior:**

`Actions.ArtifactDelta` is always `nil` after a Vertex AI round-trip,
regardless of what was persisted. Downstream consumers that depend on
`ArtifactDelta` (e.g., chat UIs that render artifact chips against session
history) silently lose every artifact reference on reload.

**Environment Details:**

- ADK Library Version: v1.3.0 (also present on `main`; bug exists in earlier
  versions including v1.1.0)
- OS: macOS
- Go Version: 1.25

**Model Information:**

- N/A — bug is in session-store serialization, model-independent.

---

## 🟡 Optional Information

**Regression:**

No — this has been present since the Vertex AI session service was added.

- Always (100%) — every tool-side artifact save is affected when running
  against `session/vertexai`.

**Additional Context:**

Structurally identical to #679 (Vertex AI client drops `FunctionCall.ID` /
`FunctionResponse.ID` on round-trip, fixed by #762), in the same file. See
also #668 (`loadartifactstool` `[]string` issue) and #509 (REST API
`EventActions` missing fields) for the same class of "EventActions
serializer is incomplete" bug.

```go
package main

import (
    "context"
    "fmt"

    "google.golang.org/adk/model"
    "google.golang.org/adk/session/vertexai"
    "google.golang.org/genai"
)

func main() {
    ctx := context.Background()
    svc, err := vertexai.NewSessionService(ctx, vertexai.VertexAIServiceConfig{
        ProjectID:       "<your-project>",
        Location:        "us-central1",
        ReasoningEngine: "<your-engine-id>",
    })
    if err != nil {
        panic(err)
    }

    create, err := svc.Create(ctx, &session.CreateRequest{
        AppName: "<your-engine-id>", UserID: "u",
    })
    if err != nil {
        panic(err)
    }

    err = svc.AppendEvent(ctx, create.Session, &session.Event{
        InvocationID: "inv-1",
        Author:       "model",
        Actions: session.EventActions{
            ArtifactDelta: map[string]int64{"chart.html": 1},
        },
            Content: genai.NewContentFromText("ok", genai.RoleModel),
        },
    })
    if err != nil {
        panic(err)
    }

    got, err := svc.Get(ctx, &session.GetRequest{
        AppName: "<your-engine-id>", UserID: "u", SessionID: create.Session.ID(),
    })
    if err != nil {
        panic(err)
    }

    for evt := range got.Session.Events().All() {
        // Prints: ArtifactDelta = map[]
        // Want:   ArtifactDelta = map[chart.html:1]
        fmt.Printf("ArtifactDelta = %v\n", evt.Actions.ArtifactDelta)
    }
}
```

kiwigitops:
## Summary
- include `Actions.ArtifactDelta` when appending events to Vertex AI sessions
- restore `Actions.ArtifactDelta` when listing Vertex AI session events, including append/list read-back
- reject artifact versions that do not fit in the Vertex AI `int32` wire field instead of truncating them
- keep nil `Actions` responses from panicking while preserving empty action messages on the wire

Fixes #902.

## Testing
- `go test -race -timeout=120s -mod=readonly -count=1 -run 'TestEventActionsConverters|TestVertexAiClientEventActionsRoundTrip' ./session/vertexai`
- `gofmt -l session/vertexai/vertexai_client.go session/vertexai/vertexai_test.go`
- `git diff --check upstream/main...HEAD`

### DIFF
--- a/session/vertexai/vertexai_client.go
+++ b/session/vertexai/vertexai_client.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"regexp"
 	"strconv"
 	"strings"
@@ -253,14 +254,9 @@ func (c *vertexAiClient) appendEvent(ctx context.Context, appName, sessionID str
 		return err
 	}
 
-	var eventState *aiplatformpb.EventActions
-	// Convert and set the initial state if provided
-	if len(event.Actions.StateDelta) > 0 {
-		sessionState, err := toStructPB(event.Actions.StateDelta)
-		if err != nil {
-			return fmt.Errorf("failed to convert state to structpb: %w", err)
-		}
-		eventState = &aiplatformpb.EventActions{StateDelta: sessionState}
+	eventActions, err := createAiplatformpbEventActions(event)
+	if err != nil {
+		return fmt.Errorf("failed to convert event actions: %w", err)
 	}
 
 	content, err := createAiplatformpbContent(event)
@@ -293,7 +289,7 @@ func (c *vertexAiClient) appendEvent(ctx context.Context, appName, sessionID str
 			Author:        event.Author,
 			InvocationId:  event.InvocationID,
 			Content:       content,
-			Actions:       eventState,
+			Actions:       eventActions,
 			EventMetadata: metadata,
 			ErrorCode:     event.ErrorCode,
 			ErrorMessage:  event.ErrorMessage,
@@ -405,9 +401,7 @@ func (c *vertexAiClient) listSessionEvents(ctx context.Context, appName, session
 			Timestamp:    rpcResp.Timestamp.AsTime(),
 			InvocationID: rpcResp.InvocationId,
 			Author:       rpcResp.Author,
-			Actions: session.EventActions{
-				StateDelta: filterNilValues(rpcResp.Actions.StateDelta.AsMap()),
-			},
+			Actions:      aiplatformToSessionEventActions(rpcResp.Actions),
 			LLMResponse: model.LLMResponse{
 				Content:      content,
 				ErrorCode:    rpcResp.ErrorCode,
@@ -434,6 +428,48 @@ func (c *vertexAiClient) listSessionEvents(ctx context.Context, appName, session
 		return events[len(events)-numRecentEvents:], nil
 	}
 	return events, nil
+}
+
+func createAiplatformpbEventActions(event *session.Event) (*aiplatformpb.EventActions, error) {
+	if len(event.Actions.StateDelta) == 0 && len(event.Actions.ArtifactDelta) == 0 {
+		return nil, nil
+	}
+
+	actions := &aiplatformpb.EventActions{}
+	if len(event.Actions.StateDelta) > 0 {
+		sessionState, err := toStructPB(event.Actions.StateDelta)
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert state to structpb: %w", err)
+		}
+		actions.StateDelta = sessionState
+	}
+	if len(event.Actions.ArtifactDelta) > 0 {
+		actions.ArtifactDelta = make(map[string]int32, len(event.Actions.ArtifactDelta))
+		for name, version := range event.Actions.ArtifactDelta {
+			if version > math.MaxInt32 || version < math.MinInt32 {
+				return nil, fmt.Errorf("artifact %q version %d does not fit in int32", name, version)
+			}
+			actions.ArtifactDelta[name] = int32(version)
+		}
+	}
+	return actions, nil
+}
+
+func aiplatformToSessionEventActions(actions *aiplatformpb.EventActions) session.EventActions {
+	if actions == nil {
+		return session.EventActions{}
+	}
+
+	eventActions := session.EventActions{
+		StateDelta: filterNilValues(actions.StateDelta.AsMap()),
+	}
+	if len(actions.ArtifactDelta) > 0 {
+		eventActions.ArtifactDelta = make(map[string]int64, len(actions.ArtifactDelta))
+		for name, version := range actions.ArtifactDelta {
+			eventActions.ArtifactDelta[name] = int64(version)
+		}
+	}
+	return eventActions
 }
 
 func sessionIdBySessionName(sn string) (string, error) {
--- a/session/vertexai/vertexai_test.go
+++ b/session/vertexai/vertexai_test.go
@@ -15,16 +15,26 @@
 package vertexai
 
 import (
+	"context"
+	"fmt"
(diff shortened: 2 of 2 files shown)


session/vertexai/vertexai_client.go
```diff
@@ -436,6 +429,46 @@ func (c *vertexAiClient) listSessionEvents(ctx context.Context, appName, session
...
+	return actions, nil
+}
+
+func aiplatformToSessionEventActions(actions *aiplatformpb.EventActions) session.EventActions {
+	if actions == nil {
+		return session.EventActions{}
+	}
+
+	eventActions := session.EventActions{}
+	if actions.StateDelta != nil {
```
hanorik: The replaced code called `filterNilValues(...)` unconditionally, and
`(*structpb.Struct)(nil).AsMap()` returns an empty **non-nil** map. So the old reader
produced `map[string]any{}` where this one produces `nil`.

Not a corner case: dumping the fixtures with `rpcreplay.Fprint`, **20 of 20**
`session_events` come back with an `actions` field, 16 of them empty `actions:{}` — the
service materialises `actions` even when the client sent none.

The wire can't distinguish "absent" from "empty", so one of these two must be
non-identity:

| written | `main` reads back | this PR |
|---|---|---|
| `StateDelta: nil` | `map[string]any{}` ✗ | `nil` ✓ |
| `StateDelta: map[string]any{}` (what `session.NewEvent` emits) | `map[string]any{}` ✓ | `nil` ✗ |

Row 2 is the common path — `session.NewEvent` (`session/session.go:231`),
`agent/agent.go:247,306` and `agent/common_context.go:146` all build empty non-nil maps —
and it's also what adk-python does (`default_factory=dict`, never `None`). A read-back
`StateDelta` is also no longer writable: `got.Actions.StateDelta["k"]="v"` now panics with
`assignment to entry in nil map`. In-tree both indexed-write sites happen to nil-guard
(`agent/llmagent/llmagent.go:537-540`, `llm_agent_wrapper.go:220-222`), but that's those
call sites, not a contract.

Nothing catches it either — the only whole-`Actions` `cmp.Diff` in the conformance suite
(`session/sessiontestsuite/service_suite.go:642`) uses a non-empty `StateDelta`.

Suggest keeping the existing behaviour so the PR stays purely additive:

```diff
-	eventActions := session.EventActions{}
-	if actions.StateDelta != nil {
-		eventActions.StateDelta = filterNilValues(actions.StateDelta.AsMap())
+	eventActions := session.EventActions{
+		StateDelta: filterNilValues(actions.StateDelta.AsMap()),
 	}
```

Keeping `nil` deliberately is defensible too — but then please say so in the PR
description and pin it with a test, so it doesn't read as an accident.

session/vertexai/vertexai_test.go
```diff
@@ -620,3 +620,65 @@ func TestCreateAiplatformpbMetadata(t *testing.T) {
 		})
 	}
 }
+
+func TestCreateAiplatformpbEventActionsIncludesArtifactDelta(t *testing.T) {
```
hanorik: The three new tests exercise the converters in isolation; neither `appendEvent` nor
`listSessionEvents` is called. So the behaviour #902 describes — read-back deep-equals what
was persisted — has no coverage, and strictly none of the three could fail on `main`, since
neither function exists there.

An in-process fake `SessionService` is ~40 lines and makes it testable without touching the
fixtures — embed `aiplatformpb.UnimplementedSessionServiceServer`, keep the appended
`SessionEvent` in a slice, return it from `ListEvents`, and (like the real service)
materialise `Actions` if the client sent none. Then wire a `vertexAiClient` at it with
`option.WithGRPCConn` + `option.WithoutAuthentication()` and table-drive it:

```go
{name: "artifact delta only", in:  session.EventActions{ArtifactDelta: map[string]int64{"chart.html": 1}},
                              want: session.EventActions{StateDelta: map[string]any{}, ArtifactDelta: map[string]int64{"chart.html": 1}}},
{name: "state delta only",    in:  session.EventActions{StateDelta: map[string]any{"user:theme": "dark"}},
                              want: session.EventActions{StateDelta: map[string]any{"user:theme": "dark"}}},
{name: "both deltas",         /* ... */},
{name: "empty deltas",        in:  session.EventActions{StateDelta: map[string]any{}, ArtifactDelta: map[string]int64{}},
                              want: session.EventActions{StateDelta: map[string]any{}}},
{name: "no actions at all",   in:  session.EventActions{},
                              want: session.EventActions{StateDelta: map[string]any{}}},
```

with `cmp.Diff(tc.want, got[0].Actions)`.

While you're in there — the existing three tests would read better as one table using
`cmp.Diff` over the whole map rather than two hand-picked keys; as written they'd pass even
if the converter emitted an extra entry.

hanorik (changes requested):
Thanks for this — the diagnosis is right and the shape is good; extracting the two
converters as a symmetric pair is the right call.
Two things before it lands:

hanorik (dismissed):
LGTM.

hanorik (approved):

PR #905, opened 2026-05-28
fix: serialize ArtifactDelta in Vertex AI session service round-trip

### LINKED ISSUE #902: Vertex AI session service drops `Actions.ArtifactDelta` during `AppendEvent` / `ListEvents` round-trip

brucearctor:
## Summary

`appendEvent` and `listSessionEvents` in `session/vertexai/vertexai_client.go` serialize only `EventActions.StateDelta` — `ArtifactDelta` is silently dropped on round-trip. Any tool that saves an artifact via `toolCtx.Artifacts().Save(...)` ends up with an in-memory event whose `ArtifactDelta` is populated, but after the runner persists the event via the Vertex AI session service and a later `Get` / `ListEvents` reads it back, `ArtifactDelta` is `nil`.

The wire format (`aiplatformpb.EventActions`) already supports this via the `ArtifactDelta` field (proto field 3, `map<string, int32>`). This change adds the missing field mapping in both directions, with `int64↔int32` conversion since the ADK model uses `int64` while the proto uses `int32`.

## Changes

- **`appendEvent`**: Now serializes `ArtifactDelta` alongside `StateDelta` into `aiplatformpb.EventActions`
- **`listSessionEvents`**: Now deserializes `ArtifactDelta` from the proto response back into `session.EventActions`
- **`toInt32Map` / `toInt64Map`**: New helpers to convert between `map[string]int64` (ADK model) and `map[string]int32` (proto wire format)
- **Tests**: Unit tests for conversion helpers, serialization path, and deserialization path

## Testing

- All new unit tests pass
- Full `session/vertexai` integration test suite passes (32s)

Fixes #902

### DIFF
--- a/session/vertexai/vertexai_client.go
+++ b/session/vertexai/vertexai_client.go
@@ -253,12 +253,18 @@ func (c *vertexAiClient) appendEvent(ctx context.Context, appName, sessionID str
 
 	var eventState *aiplatformpb.EventActions
 	// Convert and set the initial state if provided
-	if len(event.Actions.StateDelta) > 0 {
-		sessionState, err := structpb.NewStruct(event.Actions.StateDelta)
-		if err != nil {
-			return fmt.Errorf("failed to convert state to structpb: %w", err)
+	if len(event.Actions.StateDelta) > 0 || len(event.Actions.ArtifactDelta) > 0 {
+		eventState = &aiplatformpb.EventActions{}
+		if len(event.Actions.StateDelta) > 0 {
+			sessionState, err := structpb.NewStruct(event.Actions.StateDelta)
+			if err != nil {
+				return fmt.Errorf("failed to convert state to structpb: %w", err)
+			}
+			eventState.StateDelta = sessionState
 		}
-		eventState = &aiplatformpb.EventActions{StateDelta: sessionState}
+		if len(event.Actions.ArtifactDelta) > 0 {
+			eventState.ArtifactDelta = toInt32Map(event.Actions.ArtifactDelta)
+		}
 	}
 
 	content, err := createAiplatformpbContent(event)
@@ -328,7 +334,8 @@ func (c *vertexAiClient) listSessionEvents(ctx context.Context, appName, session
 			InvocationID: rpcResp.InvocationId,
 			Author:       rpcResp.Author,
 			Actions: session.EventActions{
-				StateDelta: filterNilValues(rpcResp.Actions.StateDelta.AsMap()),
+				StateDelta:    filterNilValues(rpcResp.Actions.StateDelta.AsMap()),
+				ArtifactDelta: toInt64Map(rpcResp.Actions.GetArtifactDelta()),
 			},
 			LLMResponse: model.LLMResponse{
 				Content:      content,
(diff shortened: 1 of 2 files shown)


indurireddy-TF:
Thank you for the PR. I have verified the test cases, and they are working as expected.
@kdroste-google @foxfrikses , could you please take a look at this PR and review the changes when you have a moment?

hanorik:
Thanks so much for taking the time to put this together! It looks like this overlaps with #904 that was already merged so I will close this one. We really appreciate your contribution and hope you'll submit more PRs in the future!

PR #908, opened 2026-05-28
fix: agenttool SkipSummarization terminates parent agent loop

### LINKED ISSUE #892: [bug] agenttool SkipSummarization terminates the parent agent loop
## Required information

**Describe the bug:**

When an `agenttool` is configured with `SkipSummarization: true`, the parent agent's flow loop terminates immediately after the tool returns. The parent never makes a follow-up LLM call to process the tool result.

The root cause: `agenttool.Run()` sets `ctx.Actions().SkipSummarization = true` on the tool context (line 129 in `tool/agenttool/agent_tool.go`). This propagates to the function response event. Then in `base_flow.go`, `IsFinalResponse()` returns `true` when `Actions.SkipSummarization` is set, causing the outer `for` loop to exit.

The flag conflates two concerns:
1. "Don't LLM-summarize the sub-agent output" (what the name suggests)
2. "This is a terminal event, stop the parent loop" (the actual effect via `IsFinalResponse()`)

**Steps to reproduce:**

1. Create a parent agent with tools including an `agenttool`
2. Configure the agenttool with `&agenttool.Config{SkipSummarization: true}`
3. Have the parent call the agenttool as part of a multi-step workflow (parent expects to act on the result)
4. The parent streams text, calls the agenttool, tool executes successfully, then the session ends silently with no further LLM call

```go
subAgent, _ := llmagent.New(llmagent.Config{...})
subTool := agenttool.New(subAgent, &agenttool.Config{SkipSummarization: true})

parentAgent, _ := llmagent.New(llmagent.Config{
    Tools: []tool.Tool{subTool, otherTools...},
    // parent is instructed to call subTool, then use the result to call otherTools
})
```

**Expected behavior:**

After the agenttool returns, the parent agent should make another LLM call with the tool result in context, allowing it to continue its workflow (e.g., call other tools with the sub-agent's output).

**Observed behavior:**

The parent's flow loop exits. No error is returned. The session ends as if the agent completed normally. In a streaming setup this looks like the response just stops mid-conversation after the tool call.

**Environment details:**
- ADK Library Version: v0.2.0
- OS: macOS / Linux
- Go Version: 1.24

**Model information:**
- claude-sonnet-4-6 (not model-specific, the bug is in the runner loop)

---

## Optional information

**Regression:** Unknown, likely present since `SkipSummarization` was added to `IsFinalResponse()`.

**How often has this issue occurred?:** Always (100%)

**Minimal reproduction code:**

The relevant code path in `tool/agenttool/agent_tool.go`:

```go
// Line 127-131
if t.skipSummarization {
    if actions := toolCtx.Actions(); actions != nil {
        actions.SkipSummarization = true
    }
}
```

Then in `session/session.go`:

```go
func (e *Event) IsFinalResponse() bool {
    if (e.Actions.SkipSummarization) || len(e.LongRunningToolIDs) > 0 {
        return true  // terminates the parent loop
    }
    return !hasFunctionCalls(&e.LLMResponse) && ...
}
```

And in `internal/llminternal/base_flow.go`:

```go
func (f *Flow) Run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
        for {
            for ev, err := range f.runOneStep(ctx) { ... }
            if lastEvent == nil || lastEvent.IsFinalResponse() {
                return  // parent stops here
            }
        }
    }
}
```

**Suggested fix:** Decouple the two behaviors. `SkipSummarization` should only control whether the sub-agent output gets LLM-summarized before being returned. It should not affect `IsFinalResponse()`. The function response event from a tool call should never be considered "final" on its own since the parent still needs to process it.

rsd-darshan:
Fixes #892

## Root cause

`agenttool.Run()` sets `SkipSummarization = true` on the parent tool context
when `Config.SkipSummarization` is enabled. That's intentional: it matches
adk-python and adk-java, and it's the mechanism `exitlooptool` and the
confirmation-required paths in `functiontool`/`mcptoolset` also rely on to
end the parent agent loop early. The actual bug is what happens once the
loop ends there: the terminal function response event carried only the
`FunctionResponse` part, no visible text, so any UI that doesn't render
function responses (most of them) showed nothing for a call that did
produce output — the "session ends silently" symptom in #892.

## Fix

Port adk-python's fix for the same complaint (`b983fcf9`, later scoped to
`AgentTool` by `84c77110`): when `SkipSummarization` ends the loop, append
the tool's result to the terminal event as a visible text part alongside
the `FunctionResponse` part.

This only applies to tools that opt into it via a new
`toolinternal.SkipSummarizationResultDisplayer` interface, which `agenttool`
implements. `SkipSummarization` is also set by tools whose result is an
internal detail never meant for display (a pending tool confirmation, a
UI/widget tool's acknowledgement); those don't implement the interface, so
their result isn't leaked into a visible text part.

## Test

- `TestAgentTool_Run_SkipSummarization` — `SkipSummarization` propagates to
  the parent tool context as before.
- `TestDisplayableToolResultText` — the text-rendering helper, including the
  empty/nil-result and HTML-escaping edge cases raised in review.
- `TestHandleFunctionCalls_SkipSummarization_AgentToolDisplaysResult` —
  end-to-end: the loop still terminates on the function response event, and
  that event carries the sub-agent's answer as visible text.
- `TestHandleFunctionCalls_SkipSummarization_PlainToolResultNotDisplayed` —
  end-to-end: a plain tool that sets `SkipSummarization` still terminates
  the loop, but its result is not surfaced as text.


rsd-darshan:
Hi @yarolegovich, when you get a chance, could you take a look at #908? It fixes a bug where `agenttool` with `SkipSummarization: true` was terminating the parent agent loop prematurely — the flag was being propagated to the parent tool context, causing `IsFinalResponse()` to return `true` on the function response event. Removed that propagation so the parent loop continues as expected.

effati:
What's the status of this PR?

jjsasha63 (changes requested):
Thanks for tracing this — the diagnosis is right and I reproduced it. With
`SkipSummarization: true` the parent model is called once and the follow-up
never happens; with `false` it is called twice. Your patch does make the parent
continue.

I do not think the propagation should be deleted, though.

## The propagation looks intentional, not accidental

adk-python does the same thing and has since its first public commit:
`src/google/adk/tools/agent_tool.py` sets `tool_context.actions.skip_summarization`,
`events/event.py` short-circuits `is_final_response()` on it, and
`flows/llm_flows/base_llm_flow.py` breaks its loop there. adk-java repeats it,
with a comment in `Event.java` saying it mirrors Python.

Upstream has answered this exact report twice. google/adk-python#3881 is the
same complaint; the fix (commit `b983fcf9`, scoped by `84c77110`) kept the
termination and instead appended the sub-agent's output as visible text.
google/adk-python#6598 was closed as intended behaviour, on the grounds that
`exit_loop`, `get_user_choice` and `agent_tool` all rely on that side effect.
In this repo, #544 deliberately set the flag for the same purpose.

## What the patch does to the public config

After the change, `agentTool.skipSummarization` is assigned in `New()` and read
nowhere. I deleted the field and both writes: `go build ./...`, `go vet` and the
full test suite still pass, including the new table test. So
`agenttool.Config{SkipSummarization: true}` becomes inert while the doc comment
at `agent_tool.go:48-50` still promises behaviour, and the `skip_summarization`
YAML key plumbed through `internal/configurable/configurable_utils.go:122`
becomes a no-op — a key Go, Python and Java share a schema for.

## A smaller change that fixes what the reporter saw

The real gap is that adk-go never ported `b983fcf9`. A Go user gets a terminal
event carrying no displayable text, which is the "session ends silently"
symptom in #892. Appending the sub-agent's response text to the function
response event removes that symptom and keeps the three runtimes aligned.

It does not give the parent a second LLM call. If adk-go wants that, it is a
semantics change worth a maintainer decision and probably a separately named
action rather than a behaviour change to `SkipSummarization` — see the naming
discussion on #583.

Unrelated but nearby: the actions merge in `internal/llminternal/base_flow.go`
ORs `SkipSummarization` across parallel function responses, so one flagged agent
tool ends the turn for all of them. Probably its own issue.

Happy to share the repro test either way.

rsd-darshan:
Thanks for the thorough review, @jjsasha63 — the adk-python history you dug up settles it, I agree the propagation should stay.

Pushed a rework: `agenttool` sets `SkipSummarization` again (so `Config.SkipSummarization` is live and the doc comment/YAML key stay true), and the loop-termination behavior is untouched. The actual fix is now in `handleFunctionCalls` in `internal/llminternal/base_flow.go`, mirroring `b983fcf9`: when `SkipSummarization` is set on a non-error, non-empty result, the terminal event gets the result appended as a text part alongside the `FunctionResponse` part, so it's no longer silently dropped by UIs that don't render function responses.

One addition beyond the direct port: I gated it on `RequestedToolConfirmations` being empty. `RequestConfirmation` also sets `SkipSummarization` to stop the loop while a confirmation is pending, and running the same logic there double-serializes the pending result into a spurious text part — caught this because it broke `TestA2AInputRequired/tool_confirmation`. Control-flow tools like `exit_loop` are unaffected since they return an empty result.

Added a unit test for the new helper (`TestDisplayableToolResultText`) and an end-to-end one (`TestHandleFunctionCalls_SkipSummarizationDisplaysResult`) asserting the loop still terminates on the function response *and* the terminal event carries the visible text. Full suite and vet pass.

Left the parallel-merge OR issue you flagged out of scope here — agreed it's a separate bug, happy to file it if you haven't already.

jjsasha63 (changes requested):
Thanks — this is the right shape now. I re-reviewed the head commit
(6f39613d): build, vet and the full suite are green, including with the branch
rebased onto current main. I also mutation-tested the two new tests and all
seven mutations were caught, so they are load-bearing. Your
RequestedToolConfirmations gate is doing real work — removing it fails
TestA2AInputRequired/tool_confirmation and TestA2AMultiHopInputRequired/tool_confirmation.

One blocker remains, plus some smaller things.

The append needs to be scoped to AgentTool
This ports b983fcf9, but upstream narrowed that seven weeks later in
84c77110 ("fix: scope skip_summarization text-append to AgentTool"), after
https://github.com/google/adk-python/issues/6230.

The reason is that skip_summarization means two opposite things. AgentTool
sets it because the sub-agent already wrote the answer, so it should be shown.
UI and widget tools set it because their response is an internal
acknowledgement, so it should stay hidden. A text part bypasses the client-side
filters that strip functionResponse, so the unscoped version leaked payloads
like {"status": "ok"} into transcripts. Current Python gates on
isinstance(tool, AgentTool) in flows/llm_flows/functions.py.

There is no equivalent check in base_flow.go, and
TestHandleFunctionCalls_SkipSummarizationDisplaysResult uses a plain
functiontool and asserts the text part is present — which is the behaviour
upstream removed. Upstream's regression test for that same input asserts one
part and no text.

Note tool/agenttool imports internal/llminternal, so this needs an interface
assertion rather than a type import.

An empty result produces an invalid part
displayableToolResultText returns ("", true) for {"result": ""}, so a
genai.Part with no field set gets appended. model/openaimodel/request.go
rejects that: openai: unsupported content part *genai.Part items=0. Right now
it is masked by the zero-part strip in contents_processor.go, which is there
for an unrelated reason. Python guards this with display_result != ''.

Smaller points
{"result": nil} renders as {"result":null}; Python suppresses that case too.
json.Marshal HTML-escapes, so a URL inside a structured result renders with
\u0026, which works against the readability this change is for. A
json.Encoder with SetEscapeHTML(false) avoids it.
The openaimodel converter turns the appended text into a separate user
message after the function_call_output, so the model is told the user said
the tool's JSON. Worth a look even once the scope is narrowed.
The mock model's t.Fatalf runs on the flow goroutine rather than the test
one. If loop termination regressed, runtime.Goexit() would kill that
goroutine and the len(events) != 2 check would still pass, so nothing would
fail. Yielding a real second response and asserting m.calls from the test
goroutine pins it properly.
The PR description still describes removing the propagation.
Yes please on the parallel-merge OR issue — go ahead and file it, I have not.

rsd-darshan:
Good catch on all of these, @jjsasha63 — especially the AgentTool scoping, that's a real bug in what I pushed (I'd missed `84c77110` when I went looking for `b983fcf9`).

Pushed a fix:

- Added `toolinternal.SkipSummarizationResultDisplayer`, an interface `agenttool` implements (`DisplayResultOnSkipSummarization() bool`, returns `true`) and nothing else does. The text-append in `handleFunctionCalls` now only fires when `curTool` implements it, so a UI/widget tool or a pending confirmation that happens to set `SkipSummarization` no longer gets its result serialized into a text part. This replaces the `RequestedToolConfirmations` gate from before — it only covered one of the two leak paths and isn't needed now that the check is opt-in per tool.
- Fixed `displayableToolResultText`: `{"result": ""}` and `{"result": nil}` now correctly return `ok=false` instead of producing a `genai.Part` with no field set. Switched to a `json.Encoder` with `SetEscapeHTML(false)` so URLs and the like don't come out with `&` in them.
- Rewrote the test: one case asserts a real `agenttool.New(...)` result gets the text part, a second asserts a plain `functiontool` that sets `SkipSummarization` does not. Neither uses `t.Fatalf` inside the model goroutine anymore — the mock now yields a real second turn if called again and I assert the call count from the test goroutine afterward, per your point.
- Updated the PR description to match what's actually in the diff now.

Leaving the openaimodel converter thing as a follow-up rather than folding it in here — confirmed it's real (`newContents` in `model/openaimodel/request.go` flushes trailing text on a content block into its own message using the block's role, so it lands as a `user` message right after `function_call_output`). It's a pre-existing quirk of how that converter handles any text following a function response in the same `Content`, not something specific to this change, so I'd rather not fold a fix for it into a PR about loop termination. Happy to file it separately if you'd like.

Haven't filed the parallel-merge OR issue yet — will do.

jjsasha63 (approved):
Approving. I re-reviewed the head commit (6e7a67e7) after the merge from main.
The PR's own diff is byte-identical to what I looked at last round — I diffed
the seven files across the merge and the only change is main's `!part.Thought`
filter in `agentTool.Run`, which came in with the merge. `go build ./...`,
`go vet ./...` and `go test ./...` all pass on the merged tree.

The scoping is right and the tests hold it in place. I mutation-tested eight
ways: dropping the interface gate, never appending, deleting each of the four
guards in `displayableToolResultText`, turning HTML escaping back on, and
making `DisplayResultOnSkipSummarization` return false. All eight failed the
suite.

I checked the parity questions I raised last time and they all come out the
same as upstream:

- The `role: "user"` on the appended text is what adk-python does too
  (`_build_function_response_content` returns `types.Content(role='user', ...)`).
  So the openaimodel converter turning it into a user message after
  `function_call_output` mirrors `lite_llm._content_to_message_param`, which
  puts leftover non-tool parts into a message with `role=content.role`. Not a
  Go defect, and nothing to fix here.
- `displayableToolResultText`'s four guards line up one-for-one with
  `has_displayable_result` and the `'error' not in function_result` check in
  `functions.py`.
- Ordering is safe: `telemetry.go` and `googlellm/live_connection.go` both read
  `Parts[0].FunctionResponse`, and the FunctionResponse part stays first.
  `mergeFunctionResponseEvents` appends non-FunctionResponse parts rather than
  dropping them.

## Non-blocking, worth a follow-up commit

Three comment fixes:

- `base_flow.go:1307` — "see RequestedToolConfirmations above" points at nothing
  above; that gate went away in the rework. The mechanism you mean is
  `agent.Context.RequestConfirmation`, which sets `SkipSummarization`.
- `base_flow.go:1143` — the escaping example cancels itself out. The comment
  reads `a URL's "&" becomes "&"`. Write `\u0026` or drop the example.
- `base_flow.go:1112` — the doc cites `exitlooptool` as a case the function
  handles, but it cannot reach the function now that the gate is opt-in and only
  `agenttool` implements it. `agentTool.Run`'s own empty-map returns are the real
  case.

Two behaviours no test pins, both of which survived a mutation:

- Removing the `actions.SkipSummarization` condition from the append still
  passes, so nothing covers an `agenttool` with `SkipSummarization: false`.
- Prepending the text part instead of appending it still passes, even though
  the two `Parts[0].FunctionResponse` readers above depend on the order.

## Separate issues, not this PR

- An `agenttool` call that fails still ends the run with no visible text: `Run`
  sets the flag before it can return an error, and the `error` guard then
  suppresses the text. That is the same shape as issue 892 for the failure case.
  adk-python has it too — `agent_tool.py` sets `skip_summarization` at the top of
  `run_async`, and `functions.py` guards on `'error' not in function_result` — so
  it is an upstream gap rather than something you introduced.
- Still keen on the parallel-merge OR issue when you get to it.

PR #910, opened 2026-05-28
feat: add the workflow node to allow the nested workflows

anFatum:
I've decided to implement the nested workflows as the nested types, so there is no need to change the `Workflow` interface (Workflow would have to implement `Description()` and `Config()` and to change `Run()` signature). 

Additionally the run method in workflow itself would require additional filtering and handling the multiple outputs that comes from one workflow

Please let me know if you have any strong objections on this approach.

Things that are out of scope in this PR:

1. Hierarchical node path. 
2. Event authorship.
3. Interrupt propagation for HITL.


workflow/workflow_node_test.go
```diff
@@ -0,0 +1,298 @@
...
+	if err != nil {
+		t.Fatalf("failed to create workflow node: %v", err)
+	}
+
+	suffixFn := func(ctx agent.InvocationContext, input string) (string, error) {
+		return input + "-parent", nil
+	}
+	parentNode := NewFunctionNode("parent_node", suffixFn, defaultNodeConfig)
+
+	parentEdges := Chain(Start, wfNode, parentNode)
```
hanorik: LGTM, but consider to rename nodes/workflows as it might seem a bit confusing that parent node is executed after the the child function node wrapped with the workflow node.

hanorik (approved):

workflow/validation.go
```diff
@@ -60,6 +63,29 @@ func validateNodes(edges []Edge) error {
...
+		}
+	}
+	return nil
+}
+
+// checkSubWorkflowName checks if the node is a WorkflowNode and if its sub-workflow has the same name as the parent workflow.
+func checkSubWorkflowName(node Node, workflowName string) error {
+	if wfNode, ok := node.(*WorkflowNode); ok {
+		if wfNode.subWorkflow.Name() == workflowName {
+			return fmt.Errorf("%w: %s", ErrSubWorkflowNameCollision, workflowName)
```
wolo-lab: super nit: %q instead of %s

wolo-lab (approved):
+1 to Mikalai's comment 

Today branch isolation was introduced, please be cautious while merging and resolving conflicts

PR #914, opened 2026-05-29
feat: add SingleTurnTool

dpasiukevich:
Add SingleTurnTool that handles single_turn agent mode. It runs the wrapped agent as a dynamic node.

### DIFF
--- a/internal/workflowinternal/single_turn_tool.go
+++ b/internal/workflowinternal/single_turn_tool.go
@@ -1,17 +1,137 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package workflowinternal
 
 import (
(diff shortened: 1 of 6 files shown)


internal/workflowinternal/single_turn_tool.go
```diff
@@ -1,17 +1,137 @@
...
+}
+
+func (t *SingleTurnTool) IsLongRunning() bool {
+	return false
+}
+
+func (t *SingleTurnTool) Run(toolCtx tool.Context, args any) (map[string]any, error) {
+	margs, ok := args.(map[string]any)
+	if !ok {
+		return map[string]any{"error": fmt.Errorf("expecting map[string]any arguments, got %T", args)}, nil
```
wolo-lab: question (here and line:66):  I wonder if we should put just strings here like:
```		return map[string]any{"error": fmt.Sptrinf("expecting map[string]any arguments, got %T", args)}, nil ```?
Then handling of the errors would be simplified.

line:66 -> In `agent_tool.go` in case of argument validation failure we do `return nil,  fmt.Errorf(...)`. Here we return `result_with_error, nil`. Consistency could be good (either way).

internal/workflowinternal/single_turn_tool_test.go
wolo-lab, line 1: nit: we could add a happy path test

wolo-lab (approved):

internal/workflowinternal/single_turn_tool.go
dpasiukevich: done, updated to be similar to agent_tool for consistency

internal/workflowinternal/single_turn_tool_test.go
dpasiukevich: added a test that uses a real workflow instance

PR #917, opened 2026-05-29
feat(workflow): add WithMaxConcurrency option to cap graph-scheduled …

wolo-lab:
Adds a workflow-level concurrency cap that limits how many nodes may
run concurrently within a single `Workflow` invocation. When the cap
is reached, additional ready-to-run nodes enter `NodePending` and
queue in FIFO order; the scheduler drains the queue as in-flight
nodes complete.

Mirrors adk-python's `Workflow.max_concurrency` 

## API

```go
w, err := workflow.New("my_workflow", edges, workflow.WithMaxConcurrency(4))
```

- `n > 0`: cap is enforced.
- `n == 0` (default): unlimited — no behavioural change for existing callers.
- `n < 0`: clamped to 0 (unlimited); defensive.

### DIFF
--- a/workflow/scheduler.go
+++ b/workflow/scheduler.go
@@ -93,6 +93,30 @@ type scheduler struct {
 	wg         sync.WaitGroup
 
 	parentCtx agent.InvocationContext
+
+	// maxConcurrency caps len(runsByName); 0 disables the cap.
+	// When at the cap, scheduleResumedNode enqueues into
+	// pendingQueue (status NodePending) and the consumer drains
+	// pendingQueue via tryDispatchPending on each completion.
+	// Set once at construction; immutable thereafter.
+	maxConcurrency int
+
+	// pendingQueue holds activations queued while the
+	// concurrency cap is saturated. FIFO; nodes are dispatched
+	// in arrival order as in-flight nodes complete. Owned by the
+	// consumer goroutine.
+	pendingQueue []pendingActivation
+}
+
+// pendingActivation is a deferred scheduleResumedNode call kept on
+// the consumer-owned pendingQueue while the concurrency cap is
+// saturated. Drained by tryDispatchPending on each completion.
+type pendingActivation struct {
+	node         Node
+	input        any
+	triggeredBy  string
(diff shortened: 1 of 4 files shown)


workflow/workflow.go
```diff
@@ -151,7 +183,10 @@ type Workflow struct {
 // An empty name disables persistence: the workflow runs normally
 // but its RunState is neither saved nor loaded, so Resume on a
 // follow-up turn will find nothing to resume from.
-func New(name string, edges []Edge) (*Workflow, error) {
+//
+// Optional Option values configure engine behaviour
+// (concurrency cap, etc.); see WithMaxConcurrency.
```
dpasiukevich, line 188: nit: I'd suggest to remove/rephrase this text. It looks excessive.

workflow/scheduler.go
```diff
@@ -235,20 +266,75 @@ func buildNodesByName(g *graph) map[string]Node {
...
+// tryDispatchPending starts as many queued activations as the
+// current concurrency budget allows. FIFO over pendingQueue;
+// stops when the queue is empty or the cap is reached again.
+//
+// Called after every completion (and after retry-timer expiry)
+// to make room-becoming-available immediately observable.
+func (s *scheduler) tryDispatchPending() {
+	for len(s.pendingQueue) > 0 && !s.atConcurrencyLimit() {
+		next := s.pendingQueue[0]
+		s.pendingQueue = s.pendingQueue[1:]
```
dpasiukevich, line 294: curious if this code should be guarded by mutex or is it excessive?

dpasiukevich (approved):

workflow/scheduler.go
wolo-lab: Valid question. 
No mutex needed - `pendingQueue` is owned exclusively by a single consumer goroutine (executing scheduler). 

A scheduler is a strict consumer-producer model: multiple goroutines related to `runNode()` communicate with scheduler (consumer) by events and don't interact with it directly. A scheduler is a consumer running by a single goroutine.

PR #919, opened 2026-05-29
fix(skilltoolset): document load_skill param as name, not skill_name

### LINKED ISSUE #912: SkillToolset default SystemInstruction documents `skill_name=` for load_skill, but the schema expects `name=`
## Summary

ADK v1.3.0's `tool/skilltoolset` ships a default `SystemInstruction` that tells the LLM to call `load_skill` with a parameter named `skill_name`, but the tool's actual `LoadSkillArgs` struct uses `name`. The LLM follows the documented instruction and is rejected by ADK's own input-schema validator on every session that loads a skill — typically self-correcting only on retry.

A secondary issue: the sibling tool `load_skill_resource` actually does use `skill_name` (plus `resource_path`), so the two tools' parameter naming diverges. Even after fixing the default instruction, this inconsistency is a footgun.

## The contradiction

[`tool/skilltoolset/toolset.go`](https://github.com/google/adk-go/blob/v1.3.0/tool/skilltoolset/toolset.go) — the default instruction:

```go
defaultSkillSystemInstruction = `... ` +
    "1. If a skill seems relevant to the current user query, you MUST use the `load_skill` " +
    "tool with `skill_name=\"<SKILL_NAME>\"` to read its full instructions before proceeding.\n" +
    ...
```

[`tool/skilltoolset/internal/skilltool/load_skill.go`](https://github.com/google/adk-go/blob/v1.3.0/tool/skilltoolset/internal/skilltool/load_skill.go) — the args:

```go
type LoadSkillArgs struct {
    Name string `json:"name" jsonschema:"The name of the skill to load."`
}
```

The JSON schema derived from this struct lists `name` as the only property — so the LLM-emitted call following the default instruction is rejected with `unexpected additional properties [skill_name]`.

## Observed behavior

Trace from a test session, in order:

```
[tool] load_skill {"skill_name":"generate-playbook"}
→ error: validating root: unexpected additional properties ["skill_name"]

[tool] load_skill {"name":"generate-playbook"}     # self-correct
→ success (returns the frontmatter + instructions)
```

The self-correction tends to work because the *response* field is named `skill_name` (`LoadSkillResult.SkillName`) and the model's pattern flips. But every fresh session pays the cost.

## Compounding inconsistency

[`load_skill_resource.go`](https://github.com/google/adk-go/blob/v1.3.0/tool/skilltoolset/internal/skilltool/load_skill_resource.go):

```go
type LoadSkillResourceArgs struct {
    SkillName    string `json:"skill_name"    jsonschema:"The name of the skill."`
    ResourcePath string `json:"resource_path" jsonschema:"..."`
}
```

So:
- `load_skill` → `name`
- `load_skill_resource` → `skill_name`

Even when the default instruction is fixed, callers will keep being confused — the *same logical thing* (the skill name) has different parameter keys across two co-resident tools.

## Suggested fix

**Option A (minimal, non-breaking):** change the default instruction's reference from `skill_name="<SKILL_NAME>"` to `name="<SKILL_NAME>"`. One line in `toolset.go`. Eliminates the misdirection without touching schemas.

**Option B (more consistent, breaking for direct callers of `load_skill_resource`):** also rename `LoadSkillResourceArgs.SkillName` → `Name` (with json tag `name`) so both tools share the same parameter key. Update the default instruction accordingly.

Happy to send a PR for Option A in the meantime.

## Workaround

For now, downstream agents can override the default by passing their own `SystemInstruction`:

```go
ts, err := skilltoolset.New(ctx, skilltoolset.Config{
    Source:            source,
    SystemInstruction: corrected, // mentions `name=` for load_skill, `skill_name=`+`resource_path=` for load_skill_resource
})
```

Discovered while building a Bedrock-Claude agent on ADK v1.3.0 — the misdirection cost about 2s per session and confused the model's parameter-name inference in subsequent tool calls on the same conversation.

koriyoshi2041:
Closes #912.

The default `SkillToolset` system instruction tells the model:

> use the `load_skill` tool with `skill_name="<SKILL_NAME>"`

but the `load_skill` tool input (`skilltool.LoadSkillArgs`) marshals its field as `name`:

```go
type LoadSkillArgs struct {
    Name string `json:"name" jsonschema:"The name of the skill to load."`
}
```

So the first `load_skill` call in a session is rejected for an unknown `skill_name` field and only succeeds after the model retries with `name`. Updated the instruction to document `name="<SKILL_NAME>"` to match the schema. (`load_skill_resource` keeps its own `skill_name` parameter — that one is unchanged.)

Added `TestDefaultInstructionMatchesLoadSkillSchema`, which reflects the JSON tag off `LoadSkillArgs.Name` and asserts the default instruction documents that exact parameter for `load_skill` (and no longer `skill_name=`), so the two cannot drift apart again. `go test ./tool/skilltoolset/...` passes.


indurireddy-TF:
Thank you for opening this PR. This correctly resolves the schema mismatch in skilltoolset where the default system instruction incorrectly directed models to use skill_name instead of name for load_skill

I have verified the test cases, and they are working as expected.

@kdroste-google @foxfrikses @baptmont , could you please take a look at these changes and review the PR when you have a moment?

koriyoshi2041:
Rebased this onto current `main` so the PR is no longer behind. Rechecked the touched skilltoolset package with `go test ./tool/skilltoolset/...`.

koriyoshi2041:
Rebased this onto current `main` again so the branch is no longer behind. Rechecked the touched package with:

```text
go test ./tool/skilltoolset/...
```

baptmont:
Pushed a small fix-up commit to this branch (maintainer edits) so it builds — the change itself and the test logic are untouched.

The test file was importing the pre-v2 module path, which took the whole package down:

```
tool/skilltoolset/instruction_test.go:8:2: no required module provides package
	google.golang.org/adk/tool/skilltoolset/internal/skilltool
```

Repointed it at `google.golang.org/adk/v2/...` and added the Apache 2.0 header that `goheader` requires on new Go files. Two lines of real change; nothing else touched.

Worth flagging that CI here is misleading — only `check-changes` and `cla/google` have ever run on this PR, so the green ticks never covered `build`, `test` or `lint`. Locally, with the fix-up:

```
go build -mod=readonly ./...                                        pass
go vet ./tool/skilltoolset/                                         pass
go test -race -count=1 -shuffle=on ./tool/skilltoolset/...          pass
go test -race -count=1 ./...                                        pass (full repo)
golangci-lint run ./tool/skilltoolset/...                           0 issues
go mod tidy -diff / gofmt -l                                        clean
```

I also checked the guard actually guards: reverting the one-line instruction change makes `TestDefaultInstructionMatchesLoadSkillSchema` fail on both assertions, so it will catch this drifting again.

#912 also raised a secondary point — `load_skill` takes `name` while `load_skill_resource` takes `skill_name`, which the reporter called a footgun in its own right. This PR deliberately leaves that alone (correctly, in my view — it's a separate change), so #912 shouldn't be treated as fully resolved by the auto-close. Probably worth a follow-up issue for the parameter-naming inconsistency.

Thanks @koriyoshi2041 — good catch, and the reflection-based drift guard is a nice touch.

baptmont (approved):

PR #922, opened 2026-05-29
docs(examples/workflow): add a WithUseAsOutput sample

wolo-lab:
## Problem

The workflow examples cover dynamic orchestration (`dynamic/basic`) and calling an `LlmAgent` from a dynamic body (`dynamic/llm`), but none of them shows output delegation. `workflow.WithUseAsOutput()` promotes a child's output to the parent dynamic node's terminal output. Without it the same value rides two events, the child's own event and the orchestrator's terminal event. With it only the child's event carries the value, stamped for both node paths, so an `LlmAgent`'s reply reaches the caller authored by the agent instead of being restated by the workflow. There was no runnable sample for the option and no entry for it in the workflow example index.

## Summary

Adds `examples/workflow/dynamic/use_as_output/` and registers it in `examples/workflow/README.md`. The layout follows the sibling samples: a `main.go` plus a README with concept bullets, goal, auth block, Mermaid diagram, run command, example session, and a "what it shows" table.

A single orchestrator holds both behaviors so the contrast is visible from the console. An in-scope request is delegated to an `LlmAgent` child with `WithUseAsOutput()`. Anything else is answered by the orchestrator's own `return`, without calling the model at all.

Which branch runs is decided by a substring check on the user's request rather than by anything the model produces, and that is deliberate. An earlier draft gated on the length of the generated email, so the delegating branch — the only thing the sample exists to show — never ran: drafts measured 412 and 506 characters against a 60-character threshold. A guard evaluated in Go before the model call keeps both branches reachable on demand and makes the session in the README reproduce.

The orchestrator is the last node in the graph, which is the shape that works today. A successor placed after a delegating dynamic node receives the zero value instead of the delegated output, which is a separate engine defect and out of scope here.

### DIFF
--- a/examples/workflow/dynamic/use_as_output/README.md
+++ b/examples/workflow/dynamic/use_as_output/README.md
@@ -8,7 +8,8 @@ ordinary `return`.
 
 - **Concept:** Promote a child's output to the parent dynamic node's terminal
   output with `workflow.RunNode(..., workflow.WithUseAsOutput())`.
-- **Needs LLM?** Yes (Gemini), on the delegating branch only
+- **Needs LLM?** Yes (Gemini). Only the delegating branch calls the model, but
+  credentials are needed to start either one.
 
 ## Goal
 
@@ -18,11 +19,12 @@ That works, but it emits the same content twice: once on the child's event and
 once on the orchestrator's terminal event.
 
 `WithUseAsOutput()` removes the second one. The child's event is stamped as the
-parent's output and the parent emits no terminal event of its own, so an
-`LlmAgent` child's reply reaches the caller exactly once, authored by the agent
-rather than by the workflow.
+parent's output and the parent emits no terminal event of its own, so the
+`LlmAgent` child's reply is carried by exactly one event, authored by the agent
+rather than by the workflow. The console prints the reply once either way, so
+the saving shows up in the event stream and not in the transcript below.
 
-The sample puts both behaviors in one node so the difference is visible from the
+The sample puts both branches in one node so either can be exercised from the
 console: type a request mentioning an email to take the delegating branch, or
 anything else to take the plain-return branch.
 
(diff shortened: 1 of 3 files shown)


examples/workflow/dynamic/use_as_output/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+The model client reads its config from the environment, so set one of:
+
+```bash
+# Option A — Gemini API key
+export GOOGLE_API_KEY=...
+
+# Option B — Vertex AI via gcloud Application Default Credentials
+gcloud auth application-default login
+export GOOGLE_GENAI_USE_VERTEXAI=true
```
kdroste-google, line 41: gemini-flash-latest won't work

examples/workflow/dynamic/use_as_output/README.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+The model client is built at startup, before the guard runs, so with neither set
+the sample exits with `api key is required for Google AI backend` instead of
+showing a prompt. That applies to the out-of-scope branch too, even though it
+never calls the model.
+
+## Workflow
+
+```mermaid
```
kdroste-google, line 53: maybe a sequence diagram will be clearer

kdroste-google (changes requested):

examples/workflow/dynamic/use_as_output/README.md
wolo-lab: Yep, we need to move it to higher level readme as we are repeating it across all the samples. To follow-up. Changed to 3.5

examples/workflow/dynamic/use_as_output/README.md
wolo-lab: Updated.

kdroste-google (approved):
LGTM, thank you!

PR #924, opened 2026-05-30
cmd/adkgo: support Agent Engine secret env vars

### LINKED ISSUE #849: Add custom list of Secrets to Agent Engine deployment
Default GOOGLE_API_KEY is required only if you use API key for Gemini API.
If not set in the project, prevents the container from proper starting. 

### Proposed Solution
Instead of hardcoded GOOGLE_API_KEY the cli `adkgo deploy agentengine` should support providing a custom list of secrets to be mapped to Env variables in the container. 

--secret=ENV_VARIABLE=SECRET:VERSION
Regex: ([^=]+)=(.*):(.*)
$1 - target Env variable
$2 - secret name as defined in the project
$3 - version (specific or "latest")

The provided list should be used as:
SecretEnv: []*aiplatformpb.SecretEnvVar{
	{Name: "ENV_VARIABLE", SecretRef: &aiplatformpb.SecretRef{Secret: "SECRET", Version: "VERSION"}},
},

Haihan-Jiang:
Fixes #849.

Summary:
- Add repeatable `--secret ENV=SECRET:VERSION` support to `adkgo deploy agentengine`.
- Use the configured secrets for Agent Engine `SecretEnv` instead of always requiring a `GOOGLE_API_KEY` secret.
- Apply configured secrets when updating an existing Agent Engine as well as when creating one.
- Add unit coverage for secret parsing and deployment spec construction.

Users who still want to provide the Gemini API key can pass:

```sh
adkgo deploy agentengine --secret GOOGLE_API_KEY=GOOGLE_API_KEY:latest ...
```

Tests:
- `go test ./cmd/adkgo/internal/deploy/agentengine`
- `go test ./cmd/adkgo/internal/deploy/agentengine -run 'TestParseSecretEnvVars|TestDeploymentSpec' -count=10`\n- `go test ./cmd/adkgo/...`\n- `go run ./cmd/adkgo deploy agentengine --help | rg -- '--secret|Agent Engine|Deploys'`\n- `git diff --check`

### DIFF
--- a/cmd/adkgo/internal/deploy/agentengine/agentengine.go
+++ b/cmd/adkgo/internal/deploy/agentengine/agentengine.go
@@ -56,6 +56,7 @@ type agentEngineServiceFlags struct {
 	serverPort    int
 	agentEngineID string
 	memoryBank    memoryBankFlags
+	secrets       []string
 }
 
 type buildFlags struct {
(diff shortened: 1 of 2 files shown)


jjsasha63 (changes requested):
Thanks for this, and apologies it has sat without a response since May.

The feature is still needed: `main` has no `SecretEnv` anywhere in the `adkgo`
deploy path, and #849 is still open asking for this flag shape. CI is green and I
built and tested it locally.

One logistics note first: it no longer merges cleanly. #1207 landed
`cmd/adkgo/internal/deploy/agentengine/agentengine_test.go` on main a few hours
ago, so your new test file now collides with it (add/add). Please rebase and fold
your four tests into the existing file. Line numbers below refer to your current
head.

Two things I think need to change before it lands.

## 1. The update path replaces the whole deployment spec

`agentengine.go:416-419` sets `req.ReasoningEngine.Spec.DeploymentSpec = deploymentSpec`
and adds `spec.deployment_spec` to the update mask. A mask path naming a message
replaces that whole message, and `deploymentSpec()` (line 138) only populates
`Env` and `SecretEnv`. Everything else in `ReasoningEngineSpec_DeploymentSpec`
gets cleared: `min_instances`, `max_instances`, `resource_limits`,
`container_concurrency`, `psc_interface_config`.

That matters because `--agent_engine_id` is documented as updating "the Agent
Engine instance to update if it exists" (line 107), so the target need not have
been created by this CLI. An engine created in the Console or by adk-python with
tuned instance counts would lose them the first time someone adds a secret.

Two ways out:
- narrow the mask to `spec.deployment_spec.secret_env` (plus `.env` if those
  should refresh too), or
- `GetReasoningEngine` first and read-modify-write the existing spec.

I would take the narrow mask. [AIP-161](https://google.aip.dev/161) requires that
a user be able to specify "either a field as a whole, or one of its subfields", so
`spec.deployment_spec.secret_env` should be accepted. If you find the service
rejects it, say so here and we will do the read-modify-write instead.

## 2. Nothing tests the wiring

The four new tests cover the two pure helpers, and they do bite — I checked.
Stubbing `SecretEnv` to an empty slice fails `TestDeploymentSpecUsesConfiguredSecrets`,
and disabling the validation guards fails all five subtests of
`TestParseSecretEnvVarsRejectsInvalidInput`.

But nothing covers the part that actually ships the secrets. I disabled both
call sites — `DeploymentSpec: deploymentSpec` at line 319, and the whole
`if len(f.agentEngine.secrets) > 0` block at 416-419 — so no secret reached any
request, and `go test ./cmd/adkgo/internal/deploy/agentengine/` still passed.
The feature can be entirely dead with CI green.

Could you extract the request construction out of the `util.LogStartStop`
closures into builders, the same way you extracted `deploymentSpec()`, and
assert on the built `CreateReasoningEngineRequest` and
`UpdateReasoningEngineRequest`? Asserting the update mask contents with and
without `--secret` is the case that would have caught issue 1.

CONTRIBUTING also asks for manual end-to-end evidence — a real deploy plus
update log would help.

## Smaller things, not blocking

- **Validation runs too late and leaks a temp directory.** `deploymentSpec()` is
  first called at line 296 (create) and 390 (update), after `prepareDockerfile`
  and `createArchive`. A typo like `--secret FOO` fails only once the whole
  tarball is built, and `cleanTemp()` at line 480 sits behind the error return at
  477, so the temp directory is left on disk. Parsing in `computeFlags` (line
  155) and storing the result on the flags struct fixes both.
- **The flag help should mention the IAM prerequisite.** The proto notes you need
  `roles/secretmanager.secretAccessor` on the Reasoning Engine Service Agent.
  Without it the deploy succeeds and the container fails to start, which is
  unpleasant to debug. One sentence at line 108.
- **No hygiene checks on the env var name.** `parseSecretEnvVars` (line 116)
  currently accepts `A=x:1 A=y:2` (two entries named `A`),
  `GOOGLE_CLOUD_REGION=r:1` (same name in both `env` and `secret_env`), and
  `has space=x:1`. It also accepts `ENV=projects/p/secrets/s:latest` and passes
  the resource path straight through as `SecretRef.Secret`, which wants a bare
  `{secret_name}`. A few lines to reject those.
- **The description is slightly stale.** It says this replaces "always requiring
  a `GOOGLE_API_KEY` secret", but `main` has no hardcoded `GOOGLE_API_KEY` secret
  today. Only the "add custom secrets" half of #849 is still open.

The `ENV=SECRET:VERSION` format itself looks right, and the fully-qualified form
`ENV=projects/p/secrets/s/versions/1` is correctly rejected.

PR #938, opened 2026-06-01
Failing test

kdroste-google:
Fail check


kdroste-google:
Worked, v2 enforces test and lint

PR #944, opened 2026-06-02
refactor: create runner.getOrCreateSession method

dpasiukevich:


runner/runner.go
```diff
@@ -125,6 +125,29 @@ type Runner struct {
...
 }
 
+func (r *Runner) getOrCreateSession(ctx context.Context, userID, sessionID string) (session.Session, error) {
+	getResp, err := r.sessionService.Get(ctx, &session.GetRequest{
+		AppName:   r.appName,
+		UserID:    userID,
+		SessionID: sessionID,
+	})
+	if err == nil {
+		return getResp.Session, nil
```
baptmont, line 135: nit: Should we check if getResp != nil?

baptmont (approved):

runner/runner.go
dpasiukevich: We didn't check in the old code. Plus if there's no error, there must be a non-nil session.

PR #945, opened 2026-06-02
Context unification: merge interfaces ToolContext and CallbackContext into Context. Introduce aliases

kdroste-google:
Further unification of contexts


wolo-lab (commented):
CallBackContext implements now 5 new methods, so the existing mocks, adapters etc will break. Maybe we should move with this PR to v2?

README-v2.md
wolo-lab, line 1: love it.

wolo-lab (approved):

PR #946, opened 2026-06-02
feat: introduce robust JSON-aware input validation in base node for string-encoded types

hanorik:
Introduces robust, JSON-aware input validation for workflow base nodes. When a string-encoded input is provided for a non-string target schema, the validator attempts to parse the string as JSON first before falling back to raw-string evaluation or structured schema validation. This facilitates seamless integration with stringified payload formats.

### Key Changes
- **`defaultValidateInput`**: Implemented as a package-private utility function in the `workflow` package to handle input data validation and coercion against schema specifications. The primary user-facing entry point is exposed via `BaseNode.ValidateInput`.
- **`schemaIsString`**: A helper implemented to identify whether a resolved schema expects a string. It correctly supports both the singular `Type` field and the plural `Types` field of `*jsonschema.Schema`.
- **Edge Cases Handled**:
  - A `nil` schema returns input data as-is.
  - A `nil` input data returns `nil` directly.
- **Validation Flow**:
  - For string inputs targeting string schemas, it bypasses JSON parsing entirely (`schemaIsString` short-circuit).
  - For string inputs targeting non-string schemas, it attempts JSON parsing first. If that fails or is invalid, it falls back to validating the raw string (supporting enums and literals).
  - For structured inputs (e.g., `map[string]any`), it leverages standard validation and conversion paths.

### Test Coverage
Added comprehensive test cases under `TestDefaultValidateInput` and `TestSchemaIsString` covering the following scenarios:
- `"123"` (string) + `int` schema $\rightarrow$ returns parsed `123` (JSON parse path).
- `"yes"` (string) + `enum["yes","no"]` schema $\rightarrow$ returns `"yes"` (raw-string fallback path).
- `"plain text"` (string) + `int` schema $\rightarrow$ returns an error.
- `map[string]any{...}` + `struct` schema $\rightarrow$ returns coerced struct mapping (standard path).
- `nil` input + any schema $\rightarrow$ returns `nil`.
- `"hello"` (string) + `string` schema $\rightarrow$ returns `"hello"` directly without attempting JSON parsing (`schemaIsString` short-circuit).

### DIFF
--- a/workflow/validation_test.go
+++ b/workflow/validation_test.go
@@ -504,10 +504,10 @@ func TestDefaultValidateInput(t *testing.T) {
 			want:   "hello",
 		},
 		{
-      name: "valid JSON object missing required field",
-      data: `{"x":1}`,
-      schema: structSchema,
-      wantErrSubstr: "missing properties", 
+			name:          "valid JSON object missing required field",
+			data:          `{"x":1}`,
+			schema:        structSchema,
+			wantErrSubstr: "missing properties",
 		},
 	}
 
--- a/workflow/validation_test.go
+++ b/workflow/validation_test.go
@@ -18,6 +18,10 @@ import (
 	"errors"
 	"strings"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/jsonschema-go/jsonschema"
 )
 
 func TestUniqueNames(t *testing.T) {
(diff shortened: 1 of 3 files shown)


workflow/validation_test.go
```diff
@@ -414,3 +418,173 @@ func TestValidateSubWorkflowNames(t *testing.T) {
...
+			schema: intSchema,
+			want:   float64(123), // JSON numbers unmarshal as float64 in any
+		},
+		{
+			name:   "string matching enum element (raw fallback)",
+			data:   "yes",
+			schema: enumSchema,
+			want:   "yes",
+		},
+		{
```
wolo-lab, line 482: one more test case:

```golang
{
    name: "valid JSON object missing required field",
    data: `{"x":1}`,
    schema: structSchema,
    wantErrSubstr: "missing properties", 
```

wolo-lab (approved):

workflow/validation_test.go
hanorik: Added.

PR #947, opened 2026-06-02
feat(otel): Instrument dynamic workflows

RKest:
**Problem:**
Children executed via `workflow.RunNode` inside a `DynamicNode` were not instrumented, so RunNode-driven delegation was invisible in the trace tree.

**Solution:**
`dynamicSubScheduler.runNode` now opens an `invoke_node` span per delegated child, nested under the dynamic node's span, and records genuine runtime failures while skipping expected control flow (`context.Canceled`, `ErrNodeInterrupted`) to match the top scheduler. The telemetry schema test case was replaced (`WorkflowChainCase` → `WorkflowDynamicCase`) to assert the nested span shape.

### Testing Plan

**Unit Tests:**

```
ok  google.golang.org/adk/workflow                            0.111s
ok  google.golang.org/adk/internal/telemetry                  (cached)
ok  google.golang.org/adk/internal/telemetry/functionaltest   0.030s
```

```sh
go run ./examples/workflow/dynamic/hitl/main.go console -otel_to_cloud
```

Dynamic node is instrumented:
image

### Checklist

### Additional context

N/A


internal/telemetry/functionaltest/workflow_telemetry_schema_test.go
```diff
@@ -28,22 +28,23 @@ import (
...
-// the empty-Logs expectation.
+// starts emitting events would be caught by the cmp.Diff against the
+// empty-Logs expectation.
 //
 // Hermetic: no network calls, no LLMs, no GCP.
 //
 // Mirrors test_telemetry_schema in
 // adk-python/tests/unittests/telemetry/test_node_functional.py.
-func TestTelemetrySchema_WorkflowChain(t *testing.T) {
+func TestTelemetrySchema_Workflow(t *testing.T) {
```
wolo-lab, line 82: we have two workflow types: static and dynamic. I think we should cover both of them, so just add a new test instead of replacing it.

internal/telemetry/functionaltest/workflow_telemetry_schema_test.go
```diff
@@ -28,22 +28,23 @@ import (
...
-// starts emitting events would be caught by the cmp.Diff against
-// the empty-Logs expectation.
+// starts emitting events would be caught by the cmp.Diff against the
+// empty-Logs expectation.
 //
 // Hermetic: no network calls, no LLMs, no GCP.
 //
 // Mirrors test_telemetry_schema in
 // adk-python/tests/unittests/telemetry/test_node_functional.py.
-func TestTelemetrySchema_WorkflowChain(t *testing.T) {
```
wolo-lab, line 46: it would be nice to have some more tests with error handling 
1) e.g. upper_node returning error.
2) check that `ErrNodeInterrupted` is properly handled

workflow/dynamic_scheduler.go
wolo-lab: just to double check that it is desired behavior: cache lookup won't emit span. If so we could add a small comment so in the future we would know that it is ok.

workflow/dynamic_scheduler.go
```diff
@@ -59,7 +63,7 @@ func newDynamicSubScheduler(parent NodeContext, parentPath string, emitUp func(*
 //
 // Session, invocation metadata, and cancellation come from
 // s.parentCtx. opts carries the resolved RunNodeOption arguments.
-func (s *dynamicSubScheduler) runNode(child Node, input any, opts runNodeOptions) (any, error) {
+func (s *dynamicSubScheduler) runNode(child Node, input any, opts runNodeOptions) (result any, err error) {
```
wolo-lab: super nit: `result` name is used but in the code we have `return **out**, nil`. It would be a bit more consistent if we use `out` or `result` in both places: in the function signature and in the body

wolo-lab (approved):

internal/telemetry/functionaltest/workflow_telemetry_schema_test.go
RKest: I disagree, a single test:
1. Reduces surface area of the code
2. Tests composition of different workflow pieces

Currently within a single test I cover
1. Static nodes
2. Dynamic nodes
3. Agents mode=task
4. Agents mode=single_turn

I feel this is better than four separate tests to get a grasp of if telemetry is working as expected

internal/telemetry/functionaltest/workflow_telemetry_schema_test.go
RKest: Done, added different cases which test errors at different stages

workflow/dynamic_scheduler.go
RKest: Thanks for catch. This **wasn't** the desired behavior.

I designed the functional test such that it would catch a regression in this behavior.

workflow/dynamic_scheduler.go
RKest: Done

internal/telemetry/functionaltest/workflow_telemetry_schema_test.go
wolo-lab: :D But it's completely different test right now! It's a nice table-driven test and the comment was for the previous version in which you had manually created 2 function nodes (upper_node and suffix_node).

I guess you replied after the refactoring.

workflow/dynamic_scheduler.go
wolo-lab: Currently we don't have a way to distinguish node execution when we hit cache or just execute a node - could we add a kind of marker to distinguish these two scenarios? Or maybe there's other mechanism which could be applicable.

FWIF: I checked that we don't have a new span for cache hit in adk python.

wolo-lab (approved):

internal/telemetry/functionaltest/workflow_telemetry_schema_test.go
RKest: Yea, it's a completely different test, but I would prefer to not have a separate functional test for every feature when it comes to telemetry.

Otherwise it can grow a bit out of control😅
Assume a separate test for each of these features × failure case × OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT for experimental semconv (not implemented in Go currently) × cache/no-cache regression test.

That's 4 × 2 × 3 × 2 = 48 functional tests😅

I prefer one parametrized canonical example🙏

We could instead prove this with unit tests, but these are in my experience more difficult to maintain, and give me less confidence

workflow/dynamic_scheduler.go
RKest: Is this marker necessary?

My thinking is for cases where:

| x | Cache hit | No cache hit | 
| - | - | - |
| Short running Node | Doesn't matter if cache was hit, because the span duration is short anyway | Doesn't matter if cache was hit, because the span duration is short anyway |
| Long-running Node | Will be short if you take a look at the span duration, hence you'll know you hit cache | Will be long if you take a look at the span duration, hence you'll know you didn't hit cache |

RKest: The missing span in ADK Python is definitely a bug, I wasn't aware of the node caching beforehand

internal/telemetry/functionaltest/workflow_telemetry_schema_test.go
wolo-lab: I fully agree. I just meant that the comment is already obsolete and not applicable for this test. LGTM for this test.

workflow/dynamic_scheduler.go
wolo-lab: While debugging P0s you usually want to limit guessing and prefer crystal clear info. However, I'm fine with that. We can add the marker later if we get an evidence that we really need it. 
side note: please create a bug for adk python, so it won't be forgotten.

PR #956, opened 2026-06-03
feat: implement input validation in scheduler and exclude ErrInputValidation from retries

hanorik:
This PR integrates node input validation into the workflow scheduler's execution flow and adjusts the retry policy to prevent retrying deterministic validation errors.

## Changes:

1. **Input Validation Before Execution**:
   - `ValidateInput` is invoked in `runNode` ([scheduler.go](file:///usr/local/google/home/hanorik/adk-go/workflow/scheduler.go#L443)) before `Run` for every node activation.

2. **Validation Error Emission & State Transition**:
   - A validation error in `runNode` returns early, setting the completion error.
   - The scheduler's `handleCompletion` ([scheduler.go](file:///usr/local/google/home/hanorik/adk-go/workflow/scheduler.go#L665)) transitions the node's status to `NodeFailed`.
   - The error is emitted via the event stream iterator as `yield(nil, wrappedErr)` in `scheduler.run` ([scheduler.go](file:///usr/local/google/home/hanorik/adk-go/workflow/scheduler.go#L558)).

3. **Test Coverage**:
   - Added unit tests in [scheduler_test.go](file:///usr/local/google/home/hanorik/adk-go/workflow/scheduler_test.go#L647):
     - `TestScheduler_InputValidationSuccess`: Verifies that upon validation success, the node's `Run` method receives the validated value rather than the raw input.
     - `TestScheduler_InputValidationFailure`: Verifies that upon validation failure, the workflow ends with `ErrInputValidation` and the node transitions to `NodeFailed`.

4. **Retry Policy Interaction**:
   - Declared the `ErrInputValidation` sentinel error in [errors.go](file:///usr/local/google/home/hanorik/adk-go/workflow/errors.go#L41).
   - Configured `defaultRetryConfig` ([config.go](file:///usr/local/google/home/hanorik/adk-go/workflow/config.go#L29)) and the fallback in `ShouldRetry` ([retry.go](file:///usr/local/google/home/hanorik/adk-go/workflow/retry.go#L65)) to exclude `ErrInputValidation` from retries by default because node input is deterministic per activation and retrying would result in a loop.
   - If a custom `ShouldRetry` predicate is explicitly provided in `RetryConfig`, the engine honors it.
   - Key implementation decisions and rationales are documented in the code comments.

### DIFF
--- a/workflow/config.go
+++ b/workflow/config.go
@@ -82,6 +82,13 @@ type NodeConfig struct {
 	// means the node is bounded only by the parent invocation
 	// context's deadline, if any.
 	Timeout time.Duration
+
+	// EmitsOwnSpan, when true, tells the scheduler not to wrap the node
+	// in an "invoke_node" telemetry span because the node body already
+	// starts its own span (e.g. an LlmAgent node whose wrapped agent
+	// emits "invoke_agent"). This keeps the span tree consistent with
+	// the direct agent path and adk-python.
+	EmitsOwnSpan bool
 }
 
 // RetryConfig defines the parameters for retrying a failed node.
--- a/workflow/config.go
+++ b/workflow/config.go
@@ -14,7 +14,10 @@
 
 package workflow
 
-import "time"
+import (
+	"errors"
+	"time"
+)
 
 // defaultRetryConfig is the default retry configuration for a node.
 var defaultRetryConfig = RetryConfig{
@@ -23,9 +26,11 @@ var defaultRetryConfig = RetryConfig{
 	MaxDelay:      60 * time.Second,
 	BackoffFactor: 2.0,
 	Jitter:        1.0,
-	ShouldRetry: func(error) bool {
-		return true
-	},
+	ShouldRetry:   defaultShouldRetry,
+}
+
+func defaultShouldRetry(err error) bool {
+	return !errors.Is(err, ErrInputValidation)
 }
 
 // DefaultRetryConfig returns a copy of the default retry policy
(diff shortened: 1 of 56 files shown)


workflow/scheduler_test.go
```diff
@@ -643,3 +645,72 @@ func TestScheduler_RetryCancelled(t *testing.T) {
...
+	if len(events) != 1 {
+		t.Fatalf("expected 1 event, got %d", len(events))
+	}
+	// JSON numbers unmarshal as float64 when unmarshaled into any
+	if got, ok := node.runInput.(float64); !ok || got != 123.0 {
+		t.Errorf("expected node to receive validated float64(123.0), got %T(%v)", node.runInput, node.runInput)
+	}
+}
+
+func TestScheduler_InputValidationFailure(t *testing.T) {
```
wolo-lab, line 673: let's verify if there are no retries.

workflow/config.go
```diff
@@ -23,8 +26,9 @@ var defaultRetryConfig = RetryConfig{
 	MaxDelay:      60 * time.Second,
 	BackoffFactor: 2.0,
 	Jitter:        1.0,
-	ShouldRetry: func(error) bool {
-		return true
+	ShouldRetry: func(err error) bool {
```
wolo-lab: we could have a helper function: 

```go
defaultShouldRetry(err error) bool {
	return !errors.Is(err, ErrInputValidation)
 }
```

which would be referenced here and in the retry.go. In this case it won't be needed to update two places.

wolo-lab (approved):

workflow/scheduler_test.go
hanorik: Done.

workflow/config.go
hanorik: Added.

PR #962, opened 2026-06-04
fix(agentengine): bind session ops to authenticated caller identity

XananasX7:
## Summary

AgentEngine session method handlers (`create_session`, `get_session`, `list_sessions`, `delete_session`, `stream_query`) were trusting caller-supplied `user_id` and `session_id` values with no binding to an authenticated principal.  This enabled BOLA/IDOR-style cross-user session access: any caller who can reach the endpoint can read, list, and delete another user's sessions by supplying that user's `user_id`.

Additionally, `create_session` accepted raw `app:` and `user:` prefixed keys in the initial state payload.  Because the session service interprets those prefixes as writes into shared app-wide and per-user state stores, an external caller could poison state visible to other users or persist malicious context across a victim's future sessions.

### `server/agentengine/controllers/method/authz.go` (new file)

- `WithAuthenticatedUserID` / `authenticatedUserID` — context helpers for middleware to inject a verified principal.
- `resolveUserID` — returns the authenticated identity when present, falls back to the caller-supplied value otherwise (backward-compatible with single-tenant / local-dev deployments).
- `sanitizeExternalState` — strips `app:` and `user:` prefixed keys from state maps supplied by external callers.

### Handler patches

All five AgentEngine method handlers now call `resolveUserID` before constructing their session service request.  `create_session` additionally passes the state through `sanitizeExternalState`.

## Backward compatibility

Deployments that do not configure authentication middleware are unaffected: `resolveUserID` falls back to the caller-supplied `user_id` exactly as before.  The fix only takes effect when a middleware stores a verified identity via `WithAuthenticatedUserID`.

## Testing

```
go build ./server/agentengine/...
```

Builds cleanly with no new errors or warnings.

---

*Reported via Google OSS VRP.*


XananasX7:
Rebased on `upstream/main` — the branch is no longer behind. No conflicts. CLA is already passing.

XananasX7:
Friendly ping — the branch has been rebased on `upstream/main` and CLA is passing. Happy to address any review comments. Thank you!

karolpiotrowicz (changes requested):
The two reported problems are real — I reproduced both against the base commit by driving `POST /reasoning_engine` on the handler `agentengine.NewHandler` returns, with no credential of any kind. Cross-user read, list and delete all return HTTP 200, and an `app:`-prefixed key supplied by one caller shows up inside a different user's session. Thank you for finding them, and sorry it took this long to get you a review.

The change as written does not close the first one, and that is what needs resolving before this can go in.

## The identity binding never runs on a server this repo can produce

`resolveUserID` falls back to the caller-supplied value whenever the context carries no identity, and nothing in the tree ever puts one there. `WithAuthenticatedUserID` has no call site, and [`AgentEngineAPIController.Query`](https://github.com/google/adk-go/blob/5cbc37c794e27470cad8f8731dd7a6953a367cb8/server/agentengine/controllers/agent_engine.go#L80) hands `req.Context()` to the handlers untouched. Driven through the public route, `async_get_session`, `async_list_sessions` and `async_delete_session` behave byte-identically before and after the change: HTTP 200, victim's state disclosed, victim's session destroyed.

An embedder who calls `agentengine.NewHandler` themselves and wraps the returned `http.Handler` does get the binding — I verified that path works, and with it wired the cross-user requests are correctly refused. But there is no config field, flag or option that installs a middleware on the `adk web agentengine` path, and nothing documents the wrapping approach. So as it stands the patch adds a hook that no supported deployment can reach.

## The binding also cannot express "reject this request"

[`authenticatedUserID`](https://github.com/google/adk-go/blob/5cbc37c794e27470cad8f8731dd7a6953a367cb8/server/agentengine/controllers/method/authz.go#L37-L42) treats a stored-but-empty principal as "no identity", so `resolveUserID` falls back to the request body. I first tested this with a middleware that stores an empty `sub`, which is an integrator mistake rather than a defect here. So I retested with a middleware doing exactly what your doc comment prescribes — setting the value only after successful validation and leaving the context alone otherwise. That leaks too, because the handler cannot tell "authentication failed" apart from "no authentication configured". With no exported reader and no strict mode, an integrator has no supported way to make this server refuse an unauthenticated caller.

## Substitution rather than refusal has a surprising outcome on delete

[`resolveUserID`](https://github.com/google/adk-go/blob/5cbc37c794e27470cad8f8731dd7a6953a367cb8/server/agentengine/controllers/method/authz.go#L53-L58) overwrites the supplied `user_id` instead of comparing it. With a middleware wired, a request naming another user plus a session ID that the *caller* happens to own returns HTTP 200, leaves the named user's session intact, and deletes the caller's own row. Without that collision the same request is a silent HTTP 200 no-op. Either way the client is told the operation it asked for succeeded, and it did not.

## `sanitizeExternalState` changes behavior for every existing deployment

This is the one part of the patch that is live today, and it contradicts the compatibility note in the description. With no middleware configured, `async_create_session` sent `{"app:cfg":"v1","user:pref":"dark","plain":"ok"}` stores `{"plain":"ok"}` and returns HTTP 200 with no error. Stripping those prefixes from an externally-supplied map is defensible — they really do reach shared stores — but dropping them silently is not, and [`session/session_test/service_suite.go`](https://github.com/google/adk-go/blob/5cbc37c794e27470cad8f8731dd7a6953a367cb8/session/session_test/service_suite.go#L645) pins seeding `app:` state through `Create` as behavior every `session.Service` must support. Worth a maintainer deciding explicitly whether that is supported through this HTTP surface before it is removed.

## One thing that has moved since you opened this

Main gained a sixth session-touching handler on 2026-06-18, [`streaming_agent_run_with_events`](https://github.com/google/adk-go/blob/160a39763e901047505abcc1acc32707edc0fab5/server/agentengine/controllers/method/streaming_agent_run_with_events.go#L173-L176), which was not in your base. It uses the caller-supplied `user_id` directly and is registered next to the handlers you patched in [`listStreamHandlers`](https://github.com/google/adk-go/blob/160a39763e901047505abcc1acc32707edc0fab5/server/agentengine/handler.go#L120-L124). On the merge result with a middleware wired, your five handlers correctly refuse a cross-user request while this one returns HTTP 200 and writes two events into the other user's session. Not something you could have anticipated, but it needs covering before merge, because five bound handlers next to one unbound one reads as a protected surface and is not.

One related note that saves you work: the Vertex delete path no longer needs attention. It did ignore `UserID` at your base, but [#1195](https://github.com/google/adk-go/pull/1195) has since added an ownership check on main.

## Smaller things

Enabling a middleware later makes every row stored under the old `user_id` unreachable, and delete then returns HTTP 200 without removing anything — I confirmed that on the sqlite backend, where the rows outlive the process. Worth a sentence in the description about what happens to existing sessions.

`WithAuthenticatedUserID` becomes permanent public API of the module with no in-repo caller. It also sits in `controllers/method`, so [`server/adkrest`](https://github.com/google/adk-go/blob/5cbc37c794e27470cad8f8731dd7a6953a367cb8/server/adkrest/controllers/sessions.go#L66-L71) — which has the same problem — cannot benefit from it. Given the semantics above are still in flux, freezing the symbol now is worth avoiding.

The `Metadata()` schemas published at deployment time still describe `user_id` as required and honoured, and `state` as stored as supplied.

[`authz.go:74`](https://github.com/google/adk-go/blob/5cbc37c794e27470cad8f8731dd7a6953a367cb8/server/agentengine/controllers/method/authz.go#L74) hardcodes `"app:"` and `"user:"` where `session.KeyPrefixApp` and `session.KeyPrefixUser` already exist. There is a third reserved prefix, `session.KeyPrefixTemp`, that the function does not strip — harmless, since `temp:` has no shared-store destination, but the doc comment reads as covering all prefixed keys.

Lastly, the patch carries no tests, and the branch of `resolveUserID` that returns the authenticated identity never executes in the suite — coverage on it is 0%. For a change whose whole value is an authorization boundary, a handler-level test that fails on the base commit would make the rest of this much easier to evaluate.

## Two separate bugs I hit while testing this

Neither is yours and neither should hold up this PR, but they are worth filing. `async_list_sessions` with `user_id` omitted returns HTTP 200 with every user's sessions and their state — I saw three users' private state in one response, identically on base and on head, so your change neither causes nor closes it. And `session/inmemory.go` `Delete` returns success for a session that does not exist, which is what makes the orphaned-row case above report a false success.

Happy to look again once the wiring question is settled — the reproductions above are all straightforward to turn into tests if that helps.

karolpiotrowicz:

It adds an in-code identity primitive so session operations can be bound to an authenticated caller. adk-python deliberately does not do this — it documents the surface as unauthenticated and leaves auth to whoever deploys it. So the question is whether ADK Go wants to own that primitive at all, and if so, where it lives so both server packages can consult it. Worth settling before the author invests further.

PR #964, opened 2026-06-04
feat: Add ability to set custom clock and UUID providers

### LINKED ISSUE #963: Custom providers for time and UUIDs

Yes.  When trying to durably execute code run via ADK, there are potential sources of non-determinism - specifically, clocks and UUID generation.  This was solved in `adk-python` with `set_time_provider` and `set_id_provider` in that ADK, but there aren't currently equivalents in `adk-go`.  Generally speaking, aside from that example, being able to set custom time and UUID providers is a capability present in `adk-python` that's currently lacking in `adk-go`.

### Proposed Solution

The proposed solution is to add equivalent seams for users to provide custom time and ID providers via context variables, and then update references to clock and UUID functions in `adk-go` to use those providers (with the default being the current behavior ofc).  This would mimic what's done in `adk-python`.

### Impact on your work

We are trying to build durable agentic workflows on top of `adk-go` and this issue is currently a blocker.  We would love to be able to land a fix next week (see below).

---

### Alternatives Considered

N/A

### Willingness to contribute

I have opened #964 that implements the proposed solution.

N/A - refer to PR (#964) directly.

### Additional Context

N/A

DABH:

- Closes: #963 

### Testing Plan

**Unit Tests:**

Full `go test ./... run` on the branch yields 55 passed, 1 failed.  The failure is identical to a failure occurring on `main` right now and seems unrelated to the changes in this branch.

A representative E2E test is here: https://gist.github.com/DABH/91c0d0a45788b4e152d53ca175b92980 .   You can run this `go run .` and get e.g.
```
  [run-1] id=det-id-3 ts=2026-01-01T00:00:02Z
  [run-2] id=det-id-3 ts=2026-01-01T00:00:02Z
```
which demonstrates that all the plumbing in the PR works as expected (we inject deterministic providers and get the same output on each run).  You can replace `deterministicCtx(context.Background())` with `context.Background()` in that test file to see the results diverge (matching the current/default ADK behavior).

### Checklist

## Additional Information

- The `platform` naming here mimics what's in `adk-python` (`google.adk.platform.time` / `google.adk.platform.uuid`), but open to other naming based on feedback.  My goal was to try to make things as close to `adk-python` as possible with this attempt.

- One important callout is a proposed *breaking change* for `session.NewEvent` - adding a context variable as a first parameter.  It's possible to avoid this (have `NewEvent` stay as-is and have it call `NewEventWithContext(context.Background() ...`), but modifying `NewEvent` seemed cleaner.  Let me know what you think.


platform/doc.go
```diff
@@ -0,0 +1,33 @@
...
+// Callers that need deterministic behavior — for example, replaying an agent
+// run on top of a workflow engine — can install their own providers on a
+// context.Context with WithTimeProvider and WithUUIDProvider. ADK reads the
+// current time and new IDs through Now and NewUUID, so an installed provider
+// transparently controls the timestamps and identifiers that end up in events
+// and sessions.
+//
+// This package mirrors the google.adk.platform.time and
+// google.adk.platform.uuid modules in adk-python. Go has no ambient,
+// context-local storage equivalent to Python's contextvars, so providers are
```
xumaple: leave mention of python out of comments

platform/doc.go
```diff
@@ -0,0 +1,33 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package platform provides seams for non-deterministic system operations,
+// such as reading the current time and generating unique IDs.
+//
+// By default these operations use the wall clock (time.Now) and random UUIDs.
+// Callers that need deterministic behavior — for example, replaying an agent
```
xumaple: I don't think this should be phrased as "deterministic" vs "nondeterministic" - this is really just a generic dichotomy between "default' vs "injected". Determinism is a temporal implementation detail.

DABH: Addressed in [aba09e6](https://github.com/google/adk-go/pull/964/commits/aba09e63c31a1c0aeb04f61b14ee25e65ae24c0a) - thanks

platform/doc.go
DABH: Addressed in [aba09e6](https://github.com/google/adk-go/pull/964/commits/aba09e63c31a1c0aeb04f61b14ee25e65ae24c0a) - thanks

platform/time.go
```diff
@@ -0,0 +1,51 @@
...
+
+package platform
+
+import (
+	"context"
+	"time"
+)
+
+// TimeProvider returns the current time. The default behavior is the wall
+// clock (time.Now); callers that need deterministic time can install a custom
```
xumaple: mentions of determinism here too

DABH: Addressed

platform/time.go
```diff
@@ -0,0 +1,50 @@
...
+
+package platform
+
+import (
+	"context"
+	"time"
+)
+
+// TimeProvider returns the current time. The default behavior is the wall
+// clock (time.Now); callers tcan install a custom provider on a context
```
brianstrauch:
```suggestion
// clock (time.Now); callers can install a custom provider on a context
```

mazas-google:
Thanks for the work on this, especially for proactively calling out the session.NewEvent signature change in the PR description — much appreciated.

Unfortunately, our API stability policy doesn't allow breaking changes in minor releases, so we need to keep this change backwards compatible for now. Breaking changes will be batched into the upcoming 2.0 release, at which point we'll be able to clean up the deprecated surface.

Concretely, could we:

1. Keep the existing session.NewEvent(invocationID string) *Event signature unchanged, and add a new function alongside it — e.g. NewEventWithContext(ctx context.Context, invocationID string) *Event — that carries the context through to the platform providers. The existing NewEvent can simply delegate:
```
      func NewEvent(invocationID string) *Event {
       return NewEventWithContext(context.Background(), invocationID)
   }
```
2. Mark NewEvent as deprecated using the standard Go convention so tooling (gopls, staticcheck, etc.) picks it up:
```
   // Deprecated: Use NewEventWithContext instead so that platform-installed
   // time and UUID providers (see platform.WithTimeProvider /
   // platform.WithUUIDProvider) are honored.
   func NewEvent(invocationID string) *Event { ... }
```

3. Update internal call sites in this PR to use NewEventWithContext so they benefit from the new providers immediately, while leaving NewEvent available for existing external callers. We'll remove NewEvent in 2.0.


The same pattern applies to any other publicly exported function whose signature is changing in this PR — please keep the old signature, add a *WithContext variant, and add a // Deprecated: comment pointing at the new one. Functions under internal/ are fine to change directly since they're not part of the public API. 

Cheers!

DABH:
Thank you so much @mazas-google for the review - it's greatly appreciated!

No problem re: backwards compatibility.  I had a compatible version of this queued up along the lines of what you were thinking, so I've pushed a fully backward compatible version now:


- `session.NewEvent(invocationID string) *Event` keeps its original signature and is now marked `// Deprecated: ` pointing at the new function. It simply delegates:
```
// Deprecated: Use NewEventWithContext instead so that platform-installed
// time and UUID providers (see platform.WithTimeProvider /
// platform.WithUUIDProvider) are honored. NewEvent always uses the wall
// clock and a random UUID.
func NewEvent(invocationID string) *Event {
    return NewEventWithContext(context.Background(), invocationID)
}
```

- Added `NewEventWithContext(ctx context.Context, invocationID string) *Event`, which sources the event ID and timestamp from the platform seams.
- All in-repo call sites now use `NewEventWithContext` so they benefit from the providers immediately, while external callers of `NewEvent` are unaffected. Happy to remove `NewEvent` in the 2.0 cleanup.

`NewEvent` was the only exported signature that changed in this PR; everything else was either additive (the new `platform` package) or under `internal/`, which I changed directly per your note. I've also squashed everything down to a single commit and rebased onto the latest `main`.

Thanks again for any other feedback needed to get this to a mergeable state!!

mazas-google (commented):
LGTM. Please fix the lint errors and this will be good to merge.

DABH:
Thanks again @mazas-google !  I've fixed the lint errors - please approve and merge when you're ready, and let me know anything else I can do :) Thanks!!

mazas-google:
@DABH Could you please take a look at the failing tests?

DABH:
Hi @mazas-google I believe the (one) test failure is an unrelated flaky timeout rather than anything in this PR:

- The failure is `TestA2ACleanupPropagation` (agent/remoteagent/v2/a2a_e2e_test.go:480, "remote cleanup was not called") timing out at ~5s. This PR doesn't touch that test file, and I think the test job passed on the previous run of this branch.
- Locally it passes consistently for me under the same flags CI uses (`go test -race -shuffle=on`), both the single test and the whole agent/remoteagent/v2 package.
- FWIW, the test shares a single `time.After(5 * time.Second)` deadline (line 475) across both cleanup-wait loops plus the intervening `<-remoteTaskIDChan`, so under `-race` on a busy runner all of that likely just occasionally doesn't complete within the one 5s window.

Since I don't have permissions to re-run CI on the repo, could you kick off a re-run of the test job when you get a chance, so we can verify that this is just a flake? Happy to send a separate PR to give each wait its own deadline (or bundle a fix here) if you'd find that useful.

Thanks again!

mazas-google (approved):

mazas-google:
Merged. Thanks again!

PR #968, opened 2026-06-05
feat(plugin): support clearing registered plugins on Runner and PluginManager

mramansayyad:
This PR introduces a ClearPlugins() method to Runner and PluginManager. This is critical for testing environments or long-running servers where you need to reset the plugin registry to avoid duplicate plugin registration errors or accumulation of plugin hooks.


jjsasha63:
I am closing this in favour of #887.

This branch contains #887's commit verbatim (32796552) plus 47 lines, so the two cannot both be reviewed independently, and the description mentions only ClearPlugins while the diff also adds the whole OnPipelineErrorCallback public API to plugin.Config.

On the ClearPlugins part itself:

- PluginManager has no lock and its pointer is stored in the run context, so setting `pm.plugins = nil` races with the callback loops in base_flow and RunLive. Running `Run` concurrently with `ClearPlugins` under `-race` reports a data race between `RunOnUserMessageCallback` / `RunBeforeRunCallback` and `ClearPlugins`.
- It drops the plugins without calling `Close()`, so a Runner shutdown path would silently skip their `CloseFunc`.
- `TestClearPlugins` calls the method without asserting anything, so it passes no matter what the method does. Replacing the body of `ClearPlugins` with a no-op leaves it green.

If you need a clean plugin registry per test, construct a new `Runner` — each one builds its own manager, and the duplicate-registration check is per-manager. If you have a real use case that needs clearing at runtime, please open an issue describing it and we will work out a safe shape.

PR #970, opened 2026-06-06
feat(workflow): validate node output with passthrough and model-text fallback

wolo-lab:
**Problem:**
The workflow `Node` interface exposed `ValidateOutput`, but nothing invoked it:
the scheduler committed and forwarded node output unvalidated, and the base
helper only ran a plain schema check. Nodes that yield raw model text (notably
`AgentNode`/LlmAgent) had no way to project that text onto a declared structured
output schema, and dynamic/delegated child output was never validated at all.

**Solution:**
Wire `ValidateOutput` into both schedulers and extend `defaultValidateOutput`
beyond a plain schema check:
- **Scheduler wiring** (`scheduler.go`, `dynamic_scheduler.go`): both
  `scheduler.handleEvent` and `dynamicSubScheduler.runNode` now run
  `ValidateOutput` on every yielded output before it is committed to the
  accumulator and forwarded to the consumer. The validated (possibly coerced)
  value is stamped back onto `Event.Output` when carried there; model-text
  outputs stay off the event. A failure surfaces as a `NodeFailed` transition.
- **Passthrough**: framework control values (`*session.Event`,
  `*session.RequestInput`) ride through `Event.Output` on some nodes but are not
  user output payloads, so they bypass schema validation.
- **Model-text fallback**: when direct validation fails on a `string` or a
  `*genai.Content` of model parts, the text is projected onto the schema —
  returned directly for a string schema, otherwise JSON-parsed and re-validated.
  This makes structured `AgentNode` output validate end-to-end and mirrors ADK
  Python's `_validate_output_data`. On total failure the original validation
  error is returned, not a downstream parse error.
- An unknown node name now fails instead of silently forwarding unvalidated
  output.


workflow/base_node.go
```diff
@@ -85,12 +92,101 @@ func (b BaseNode) ValidateOutput(out any) (any, error) {
...
+	return parsed, true
+}
+
+// rootSchemaIsString reports whether schema's root type is "string".
+func rootSchemaIsString(schema *jsonschema.Resolved) bool {
+	root := schema.Schema()
+	if root == nil {
+		return false
+	}
+	return root.Type == "string"
```
hanorik: Should we also check and handle `root.Types` correctly for cases when multiple root types are allowed for schema? If so, it also makes sense to add test case(s) for multiple types.

wolo-lab: Right, good catch!

hanorik (approved):

PR #974, opened 2026-06-06
feat(launcher): add HTTPMiddleware seam to launcher.Config

### LINKED ISSUE #965: cmd/launcher/web: expose an HTTP middleware seam on launcher.Config for embedders

## 🔴 Required Information

Embedders that wrap ADK Go inside a larger application can't insert HTTP middleware into the launcher's request pipeline. webLauncher.Run in cmd/launcher/web/web.go constructs:

```
srv := http.Server{
    Addr:    fmt.Sprintf(":%v", w.config.port),
    Handler: router,
    // ... timeouts ...
}
```

launcher.Config has no Handler, Middleware, or WrapHandler field, so there's no public way to wrap the inbound handler before it reaches ADK's routes.

The concrete case where this hurts is distributed-tracing context extraction. For cross-process tracing, an embedder needs to wrap the handler with otelhttp.NewHandler so the W3C traceparent header gets extracted into req.Context() before any handler runs. Without that wrap:

Inbound HTTP requests carry a traceparent populated by an upstream caller.
ADK handlers see req.Context() without an extracted span context.
Every span ADK creates (invoke_agent, generate_content, execute_tool) becomes a fresh trace root.
Cross-service trace stitching breaks at the ADK boundary.
The TracerProvider half of the OTel embedding story is already publicly handleable via launcher.Config.TelemetryOptions with telemetry.WithTracerProvider(...) — that prevents SetGlobalOtelProviders from clobbering the embedder's TP. But the HTTP pipeline has no symmetric escape hatch.

### Proposed Solution

Add a field to launcher.Config:

```
type Config struct {
    // ... existing fields ...

    // HTTPMiddleware is applied in order to the HTTP handler before
    // serving. Each middleware wraps the handler returned by the previous
    // one. The first middleware in the slice is the outermost (runs first
    // on inbound requests).
    HTTPMiddleware []func(http.Handler) http.Handler
}
```
And apply it in webLauncher.Run:

```
var handler http.Handler = router
for i := len(config.HTTPMiddleware) - 1; i >= 0; i-- {
    handler = config.HTTPMiddleware[i](handler)
}
```
```
srv := http.Server{
    Addr:    fmt.Sprintf(":%v", w.config.port),
    Handler: handler,                 // was: router
    // ... timeouts unchanged ...
}
```
The change is purely additive — existing callers that don't set HTTPMiddleware see no behavior difference.

### Impact on your work
How does this feature impact your work and what are you trying to achieve?

We embed ADK Go inside an internal agent platform that already has its own OpenTelemetry pipeline (TracerProvider, propagators, span processors, OTLP exporter, etc.). Today we bypass webLauncher.Run entirely and re-implement its setup (BuildBaseRouter, apiLauncher.SetupSubrouters, session.InMemoryService default) so we can insert otelhttp.NewHandler around the router before constructing the http.Server. The bypass is ~50 lines of code in our repo and pins us against ADK internals — every ADK upgrade requires re-checking whether our replay still mirrors webLauncher.Run correctly.

With this feature, the bypass goes away. We'd call full.NewLauncher().Execute(ctx, config, args) with both TelemetryOptions and HTTPMiddleware set, and stop tracking ADK launcher internals.

---

### Alternatives Considered

1. Bypass webLauncher.Run (current workaround). Re-implement its setup, wrap with otelhttp.NewHandler, serve. Works, but fragile against ADK upgrades.
2. Sidecar reverse proxy on a different port. Stand up a separate http.Server with otelhttp.NewHandler that reverse-proxies to ADK's port. This does NOT solve the problem — httputil.NewSingleHostReverseProxy forwards a new HTTP request whose req.Context() is freshly populated by ADK's http.Server. ADK's handlers still don't extract the inbound traceparent. Verified empirically: the trace tree remains broken at the ADK boundary.
3. Pre-bind the listener and hand it to ADK. launcher.Config has no Listener field; srv.ListenAndServe() binds internally. Not viable through the public API go:linkname or reflection to grab ADK's router after construction. Possible but unmaintainable; breaks on every ADK patch release.
4. The proposed HTTPMiddleware field is the smallest public-API change that genuinely solves the problem.

### Willingness to contribute
Are you interested in implementing this feature yourself or submitting a PR?

Yes — happy to submit a PR if the design is acceptable. Would prefer to align on the API shape first (single func(http.Handler) http.Handler vs. slice; field name; doc-comment phrasing; whether the same hook should also apply to a2a and webui sublaunchers or only api) before writing the change.

If you have ideas on how this should look in code, please share a
pseudo-code example.

See Proposed Solution for the field and the webLauncher.Run loop. Embedder usage after the change would be:

```
config := &launcher.Config{
    AgentLoader: agent.NewSingleLoader(myAgent),
    TelemetryOptions: []telemetry.Option{
        telemetry.WithTracerProvider(myTP),
    },
    HTTPMiddleware: []func(http.Handler) http.Handler{
        func(h http.Handler) http.Handler {
            return otelhttp.NewHandler(h, "agent-name",
                otelhttp.WithPropagators(otel.GetTextMapPropagator()))
        },
    },
}
return full.NewLauncher().Execute(ctx, config, args)
```

The middleware chain order is outer-first: [A, B] means A wraps B wraps router, so on inbound requests A runs first. This matches the common chi.Use(...) / mux.Use(...) convention in Go HTTP routers.

### Additional Context

- launcher.Config.TelemetryOptions already provides the analogous public seam for the TracerProvider clobber problem — telemetry.WithTracerProvider(tp) makes SetGlobalOtelProviders re-install the same TP (no-op clobber). HTTPMiddleware would close the symmetric gap on the HTTP pipeline.
- Per CONTRIBUTING.md, alignment with adk-python matters. I haven't yet checked whether Python ADK has an analogous embedding seam (e.g. a FastAPI middleware list) — if it does, naming and semantics should follow it. Happy to research that before the design is finalized.
- Verified the proposal against ADK v0.6.0 source: cmd/launcher/launcher.go: Config, cmd/launcher/web/web.go: webLauncher.Run, cmd/launcher/internal/telemetry/telemetry.go: InitAndSetGlobalOtelProviders, telemetry/config.go: WithTracerProvider, telemetry/telemetry.go: SetGlobalOtelProviders.

nuthalapativarun:

- Closes: #965

### Root Cause / Motivation

Embedders that wrap ADK Go inside a larger application have no way to insert HTTP middleware into the launcher's request pipeline. `webLauncher.Run` builds the `http.Server` directly from the router with no public hook for wrapping the handler. The concrete case described in the issue is distributed-tracing context extraction (`otelhttp.NewHandler`), but the pattern applies to any cross-cutting concern (auth, rate-limiting, request-ID injection, etc.).

### Change

Add `HTTPMiddleware []func(http.Handler) http.Handler` to `launcher.Config`:

```go
// HTTPMiddleware is applied in order to the HTTP handler before serving.
// Each entry wraps the handler returned by the previous one. The first
// middleware in the slice is the outermost (runs first on inbound requests).
// A nil or empty slice is a no-op; existing callers are unaffected.
HTTPMiddleware []func(http.Handler) http.Handler
```

Apply the chain in `webLauncher.Run` before constructing `http.Server`:

```go
var handler http.Handler = router
for i := len(config.HTTPMiddleware) - 1; i >= 0; i-- {
    handler = config.HTTPMiddleware[i](handler)
}
```

The reverse-index loop produces outer-first ordering (slice index 0 runs first on inbound requests), matching `chi.Use` / `mux.Use` conventions. The zero-value (nil slice) is a no-op, so all existing callers are unaffected.

### Testing Plan
**Unit Tests:**
- [x] Added `TestHTTPMiddleware_AppliedOutermostFirst` in `cmd/launcher/web/web_test.go`: starts the web server with two named middlewares, fires a `/ping` request, and asserts that the invocation order matches slice order (`["first", "second", ...]`).
- [x] All existing tests in `./cmd/launcher/...` continue to pass.

Embedder usage after the change:
```go
config := &launcher.Config{
    AgentLoader: agent.NewSingleLoader(myAgent),
    TelemetryOptions: []telemetry.Option{
        telemetry.WithTracerProvider(myTP),
    },
    HTTPMiddleware: []func(http.Handler) http.Handler{
        func(h http.Handler) http.Handler {
            return otelhttp.NewHandler(h, "agent-name",
                otelhttp.WithPropagators(otel.GetTextMapPropagator()))
        },
    },
}
return full.NewLauncher().Execute(ctx, config, args)
```

### Checklist
- [x] Read CONTRIBUTING.md
- [x] Self-reviewed code
- [x] Commented field doc (ordering semantics are non-obvious)
- [x] Added tests proving the feature
- [x] Purely additive — zero-value is a no-op, no existing callers affected

### DIFF
--- a/cmd/launcher/web/web_test.go
+++ b/cmd/launcher/web/web_test.go
@@ -1,0 +1,118 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package web_test
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/gorilla/mux"
+
+	"google.golang.org/adk/cmd/launcher"
+	"google.golang.org/adk/cmd/launcher/web"
+	"google.golang.org/adk/session"
+)
+
+// minimalSublauncher is a stub Sublauncher for testing the web launcher.
+type minimalSublauncher struct{}
+
+func (m *minimalSublauncher) Keyword() string                             { return "stub" }
+func (m *minimalSublauncher) Parse(args []string) ([]string, error)       { return args, nil }
+func (m *minimalSublauncher) CommandLineSyntax() string                   { return "" }
+func (m *minimalSublauncher) SimpleDescription() string                   { return "" }
+func (m *minimalSublauncher) UserMessage(_ string, _ func(v ...any))      {}
+func (m *minimalSublauncher) SetupSubrouters(router *mux.Router, _ *launcher.Config) error {
+	router.HandleFunc("/ping", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+	return nil
+}
+
+func getFreePort(t *testing.T) int {
+	t.Helper()
+	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
+	if err != nil {
+		t.Fatalf("net.ResolveTCPAddr() error = %v", err)
+	}
+	l, err := net.ListenTCP("tcp", addr)
+	if err != nil {
+		t.Fatalf("net.ListenTCP() error = %v", err)
+	}
+	port := l.Addr().(*net.TCPAddr).Port
+	l.Close()
+	return port
+}
+
+func TestHTTPMiddleware_AppliedOutermostFirst(t *testing.T) {
+	var order []string
+
+	makeMiddleware := func(name string) func(http.Handler) http.Handler {
+		return func(next http.Handler) http.Handler {
+			return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				order = append(order, name)
+				next.ServeHTTP(w, r)
+			})
+		}
+	}
+
+	port := getFreePort(t)
+	config := &launcher.Config{
+		SessionService: session.InMemoryService(),
+		HTTPMiddleware: []func(http.Handler) http.Handler{
+			makeMiddleware("first"),
+			makeMiddleware("second"),
+		},
+	}
+
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
+
+	webLauncher := web.NewLauncher(&minimalSublauncher{})
+	errCh := make(chan error, 1)
+	go func() {
+		args := []string{fmt.Sprintf("--port=%d", port), "stub"}
+		errCh <- webLauncher.Execute(ctx, config, args)
+	}()
+
+	// Wait for the server to start accepting connections.
+	deadline := time.Now().Add(5 * time.Second)
+	var pingErr error
+	for time.Now().Before(deadline) {
+		var resp *http.Response
+		resp, pingErr = http.Get(fmt.Sprintf("http://localhost:%d/ping", port))
+		if pingErr == nil {
+			resp.Body.Close()
+			break
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+	if pingErr != nil {
+		t.Fatalf("server did not start in time: %v", pingErr)
+	}
+
+	cancel()
+	<-errCh
+
+	if len(order) < 2 {
+		t.Fatalf("middleware not invoked; got invocation order: %v", order)
+	}
+	if order[0] != "first" || order[1] != "second" {
+		t.Errorf("middleware invocation order = %v, want [first second ...]", order)
+	}
+}
--- a/cmd/launcher/web/web_test.go
+++ b/cmd/launcher/web/web_test.go
@@ -33,11 +33,11 @@ import (
 // minimalSublauncher is a stub Sublauncher for testing the web launcher.
 type minimalSublauncher struct{}
 
-func (m *minimalSublauncher) Keyword() string                             { return "stub" }
-func (m *minimalSublauncher) Parse(args []string) ([]string, error)       { return args, nil }
-func (m *minimalSublauncher) CommandLineSyntax() string                   { return "" }
-func (m *minimalSublauncher) SimpleDescription() string                   { return "" }
-func (m *minimalSublauncher) UserMessage(_ string, _ func(v ...any))      {}
+func (m *minimalSublauncher) Keyword() string                        { return "stub" }
+func (m *minimalSublauncher) Parse(args []string) ([]string, error)  { return args, nil }
+func (m *minimalSublauncher) CommandLineSyntax() string              { return "" }
+func (m *minimalSublauncher) SimpleDescription() string              { return "" }
+func (m *minimalSublauncher) UserMessage(_ string, _ func(v ...any)) {}
 func (m *minimalSublauncher) SetupSubrouters(router *mux.Router, _ *launcher.Config) error {
 	router.HandleFunc("/ping", func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
--- a/cmd/launcher/web/web_test.go
+++ b/cmd/launcher/web/web_test.go
@@ -24,10 +24,10 @@ import (
 
 	"github.com/gorilla/mux"
 
-	"google.golang.org/adk/cmd/launcher"
-	"google.golang.org/adk/cmd/launcher/universal"
-	"google.golang.org/adk/cmd/launcher/web"
-	"google.golang.org/adk/session"
+	"google.golang.org/adk/v2/cmd/launcher"
+	"google.golang.org/adk/v2/cmd/launcher/universal"
+	"google.golang.org/adk/v2/cmd/launcher/web"
+	"google.golang.org/adk/v2/session"
 )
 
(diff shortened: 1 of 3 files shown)


indurireddy-TF:
Hi @nuthalapativarun, thanks for the PR.
@kdroste-google,@mazas-google @karolpiotrowicz  could you please review the PR when you have a moment? All tests are passing successfully.

nuthalapativarun:

cmd/launcher/web/web_test.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
+
+	webLauncher := web.NewLauncher(&minimalSublauncher{})
+	errCh := make(chan error, 1)
+	go func() {
+		args := []string{fmt.Sprintf("--port=%d", port), "stub"}
+		errCh <- webLauncher.Execute(ctx, config, args)
```
kdroste-google: Do you mean something like

```go
	l := universal.NewLauncher(webLauncher)
	errCh := make(chan error, 1)
	go func() {
		args := []string{fmt.Sprintf("--port=%d", port), "stub"}
		errCh <- l.Execute(ctx, config, args)
	}()

```
?

kdroste-google (changes requested):
Hi, @nuthalapativarun, the general idea is fine, please fix the issues and rebase to new main (we are after the v2 release now)
Thank you!

nuthalapativarun:
Hi @kdroste-google — addressed the inline comment and rebased onto the current main (post-v2).

Summary of changes:
- **test(launcher/web)**: Updated `TestHTTPMiddleware_AppliedOutermostFirst` to route through `universal.NewLauncher(webLauncher)` and call `l.Execute` instead of calling `webLauncher.Execute` directly — matching how production code invokes the web launcher.
- **Rebased** both commits cleanly onto upstream main (no conflicts).

Please take another look when you get a chance. Thanks!

nuthalapativarun:
Hi @kdroste-google — following up. The gofumpt formatting issue and CI failures have been fixed and the branch is rebased/mergeable. Would appreciate a re-review when you have a chance to clear the CHANGES_REQUESTED status. Thanks!

karolpiotrowicz:
@wolo-lab flagging this one for a direction call before it gets a code review.

The change adds a new exported field to `launcher.Config` — [`HTTPMiddleware []func(http.Handler) http.Handler`](https://github.com/google/adk-go/blob/10fa888f97a4a260e699dd68fe7e838272080c94/cmd/launcher/launcher.go#L67-L71) — and [wraps the router with it, outermost first, before serving](https://github.com/google/adk-go/blob/10fa888f97a4a260e699dd68fe7e838272080c94/cmd/launcher/web/web.go#L185-L195). That is permanent public surface on the launcher config, and the request behind it (#965) came from a user rather than from the team, so nobody here has yet decided this is the shape we want.

It is worth weighing against how adk-python handles the same embedder problem. [`get_fast_api_app()`](https://github.com/google/adk-python/blob/c7ffcfa85a8e8970f6318306479d9c4c110583b2/src/google/adk/cli/fast_api.py#L95-L130) takes no middleware argument at all. It [returns the FastAPI app](https://github.com/google/adk-python/blob/c7ffcfa85a8e8970f6318306479d9c4c110583b2/src/google/adk/cli/fast_api.py#L672), so an embedder calls `app.add_middleware(...)` on it directly. adk-go's launcher never hands the router back, which is why the gap exists here and not there. The alternative of exposing the handler, rather than accepting a list of middleware, has not been compared against this one.

The implementation looks like a reasonable reading of what the issue asked for. The open question is whether adk-go wants a config-level middleware seam at all, and that is a call for you rather than something a code review should settle. Happy to run a full review once the direction is set.

nuthalapativarun:
Rebased onto current `main` (this branch was 173 commits behind — from before the v2 release) and squashed the follow-up fixup commits into the original one.

Addressed your inline comment: `TestHTTPMiddleware_AppliedOutermostFirst` now matches the pattern you suggested — since `cmd/launcher/web`'s test file has since moved to the internal `web` package (as part of unrelated work that landed on `main`), the test calls `webLauncher.Run` directly (via `NewLauncher(pingSublauncher{}).(*webLauncher)`) instead of going through `universal.NewLauncher(...).Execute(...)`, but the effect is the same as what you proposed.

The rebase itself needed real conflict resolution, not just a replay, since `main` had since:
- added `Authenticator`/`Authorizer`/`Compaction` fields to `launcher.Config` (kept alongside the new `HTTPMiddleware` field — both additive, no overlap)
- added telemetry init and an h2c-aware `buildHTTPServer` helper inside `web.Run` (kept that, and now build the middleware-wrapped handler before passing it to `buildHTTPServer` rather than constructing a raw `http.Server` inline)
- rewritten `web_test.go` substantially (moved to the internal `web` package, added several new tests) — my test is now appended to that file using the same fixture conventions as the neighboring tests (e.g. `telemetryFailSublauncher`)

Verification:
- `go build ./...`, `go vet ./...` clean across the whole workspace
- `go test -count=1 ./cmd/launcher/...` green, including the new test
- `golangci-lint run --new-from-rev=main ./cmd/launcher/...`: 0 issues
- Manually removed the middleware-wrapping loop in `web.Run` and confirmed `TestHTTPMiddleware_AppliedOutermostFirst` fails (`middleware not invoked; got invocation order: []`), then restored it and confirmed green — the test does pin the change.

(No `cgo`/Docker in this environment, so I couldn't run `-race`, but this package's tests don't rely on it.)

nuthalapativarun:
@kdroste-google — following up specifically on the formal review status: this PR still shows CHANGES_REQUESTED from your 2026-07-03 review, but the concrete asks from that round (the gofumpt formatting issue in `web_test.go`, and the test/lint/CLA CI failures) were fixed back on 2026-07-25, and there's been no re-review since. The branch is now rebased onto current `main` (confirmed `MERGEABLE` above) and build/vet/test/lint are all green. Could you take another look, or let me know if there's anything else outstanding I should address? Thanks for your patience on this one.

PR #975, opened 2026-06-06
feat(adkgo/deploy): add --env flag to agentengine deploy subcommand

### LINKED ISSUE #848: Add custom list of Env Variables to Agent Engine deployment

An easy way to provide custom environment variables to Agent Engine is needed

### Proposed Solution
Add new repeated flag --env="ENV_VARIABLE=A_VALUE" to `adkgo deploy agentEngine `
Accepted pattern (regexp)  ([^=]+)=(.*)
$1 - env variable name
$2 - env variable value

Many env flags should be allowed.

Expected behaviour: 
defined env flags should be set as env variables after the deployment.

nuthalapativarun:

- Closes: #848

### Root Cause / Motivation

`adkgo deploy agentengine` passes a fixed set of environment variables to the Agent Engine deployment. Users that need custom env vars (e.g. `VERTEX_AI_LOCATION`, `LOG_LEVEL`, third-party API keys that are not secrets) have no way to inject them.

### Change

Add a repeatable `--env=KEY=VALUE` flag to the `agentengine` deploy subcommand:

```
adkgo deploy agentengine \
  --env=VERTEX_AI_LOCATION=us-central1 \
  --env=LOG_LEVEL=debug \
  --region=... --project_name=...
```

**On create:** extra env vars are appended after the built-in defaults (`GOOGLE_CLOUD_REGION`, `NUM_WORKERS`, OTel flags) in `DeploymentSpec.Env`.

**On update:** if `--env` flags are provided, `DeploymentSpec` is set with defaults + extra vars and `"spec.deployment_spec"` is added to the update mask. If no `--env` flags are given, update behavior is unchanged.

Values containing `=` are handled correctly (`URL=http://x?a=1&b=2` splits on the first `=` only). Entries without `=` return a clear error.

### Testing Plan
**Unit Tests:**
- [x] Added `TestParseEnvVars` in `agentengine_test.go` covering: nil input, single var, value-contains-equals, empty value, multiple vars, missing-equals error, empty-string error.
- [x] All existing tests pass.

The flag wires correctly into the cobra command. The `parseEnvVars` function can be verified without a live GCP project; the proto field is identical to the existing hardcoded env var entries.

```
$ adkgo deploy agentengine --env=FOO=bar --env=BAZ=qux ...
# DeploymentSpec.Env will contain: GOOGLE_CLOUD_REGION, NUM_WORKERS, otel vars, FOO, BAZ
```

### Checklist
- [x] Read CONTRIBUTING.md
- [x] Self-reviewed code
- [x] Added tests proving the feature (parseEnvVars unit tests)
- [x] Symmetric with the existing env var pattern in the proto request

### DIFF
--- a/cmd/adkgo/internal/deploy/agentengine/agentengine.go
+++ b/cmd/adkgo/internal/deploy/agentengine/agentengine.go
@@ -57,6 +57,7 @@ type agentEngineServiceFlags struct {
 	serverPort    int
 	agentEngineID string
 	memoryBank    memoryBankFlags
+	envVars       []string
 }
 
 type buildFlags struct {
@@ -110,6 +111,21 @@ func init() {
 		" https://${LOCATION_ID}-aiplatform.googleapis.com/v1beta1/publishers/google/models. It will be prefixed with projects/<project_name>/locations/<region> forming for instance full model name like "+
 		" 'projects/project_name/locations/us-central1/publishers/google/models/gemini-2.5-flash'")
 	agentEngineCmd.PersistentFlags().DurationVar(&flags.agentEngine.memoryBank.ttl, "mem_ttl", time.Hour*24*365, "Time-To-Live for memories")
+	agentEngineCmd.PersistentFlags().StringArrayVar(&flags.agentEngine.envVars, "env", nil, "Additional environment variable in KEY=VALUE format. Repeatable.")
+}
+
+// parseEnvVars converts a slice of "KEY=VALUE" strings into EnvVar protos.
+// Returns an error for any entry that does not contain "=".
+func parseEnvVars(raw []string) ([]*aiplatformpb.EnvVar, error) {
+	vars := make([]*aiplatformpb.EnvVar, 0, len(raw))
+	for _, s := range raw {
+		parts := strings.SplitN(s, "=", 2)
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid --env value %q: expected KEY=VALUE", s)
+		}
+		vars = append(vars, &aiplatformpb.EnvVar{Name: parts[0], Value: parts[1]})
+	}
+	return vars, nil
 }
 
 // computeFlags uses command line arguments to create a full config
@@ -255,6 +271,17 @@ func (f *deployAgentEngineFlags) gcloudDeployToAgentEngine() error {
 			}
 			p("Methods:", string(methodsJSON))
 
+			defaultEnv := []*aiplatformpb.EnvVar{
+				{Name: "GOOGLE_CLOUD_REGION", Value: f.gcloud.region},
+				{Name: "NUM_WORKERS", Value: "1"},
+				{Name: "GOOGLE_CLOUD_AGENT_ENGINE_ENABLE_TELEMETRY", Value: "true"},
+				{Name: "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT", Value: "true"},
+			}
+			extraEnv, err := parseEnvVars(f.agentEngine.envVars)
+			if err != nil {
+				return err
+			}
+
 			req := &aiplatformpb.CreateReasoningEngineRequest{
 				Parent: parent,
 				ReasoningEngine: &aiplatformpb.ReasoningEngine{
@@ -274,11 +301,9 @@ func (f *deployAgentEngineFlags) gcloudDeployToAgentEngine() error {
 						},
 						AgentFramework: "google-adk",
 						DeploymentSpec: &aiplatformpb.ReasoningEngineSpec_DeploymentSpec{
-							Env: []*aiplatformpb.EnvVar{
-								{Name: "GOOGLE_CLOUD_REGION", Value: f.gcloud.region},
-								{Name: "NUM_WORKERS", Value: "1"},
-								{Name: "GOOGLE_CLOUD_AGENT_ENGINE_ENABLE_TELEMETRY", Value: "true"},
-								{Name: "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT", Value: "true"},
+							Env: append(defaultEnv, extraEnv...),
+							SecretEnv: []*aiplatformpb.SecretEnvVar{
+								{Name: "GOOGLE_API_KEY", SecretRef: &aiplatformpb.SecretRef{Secret: "GOOGLE_API_KEY", Version: "latest"}},
 							},
 						},
 						ClassMethods: methods,
@@ -352,6 +377,11 @@ func (f *deployAgentEngineFlags) gcloudUpdateAgentEngine() error {
 			}
 			p("Methods:", string(methodsJSON))
 
+			extraEnv, err := parseEnvVars(f.agentEngine.envVars)
+			if err != nil {
+				return err
+			}
+
 			req := &aiplatformpb.UpdateReasoningEngineRequest{
 				ReasoningEngine: &aiplatformpb.ReasoningEngine{
 					Name: name,
@@ -372,6 +402,19 @@ func (f *deployAgentEngineFlags) gcloudUpdateAgentEngine() error {
 					},
 				},
 				UpdateMask: &fieldmaskpb.FieldMask{Paths: []string{"spec.source_code_spec", "spec.class_methods"}},
+			}
+
+			if len(extraEnv) > 0 {
+				defaultEnv := []*aiplatformpb.EnvVar{
+					{Name: "GOOGLE_CLOUD_REGION", Value: f.gcloud.region},
+					{Name: "NUM_WORKERS", Value: "1"},
+					{Name: "GOOGLE_CLOUD_AGENT_ENGINE_ENABLE_TELEMETRY", Value: "true"},
+					{Name: "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT", Value: "true"},
+				}
+				req.ReasoningEngine.Spec.DeploymentSpec = &aiplatformpb.ReasoningEngineSpec_DeploymentSpec{
+					Env: append(defaultEnv, extraEnv...),
+				}
+				req.UpdateMask.Paths = append(req.UpdateMask.Paths, "spec.deployment_spec")
 			}
 
 			if f.agentEngine.memoryBank.deploy {
--- a/cmd/adkgo/internal/deploy/agentengine/agentengine_test.go
+++ b/cmd/adkgo/internal/deploy/agentengine/agentengine_test.go
@@ -1,0 +1,87 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package agentengine
+
+import (
+	"testing"
+
+	"cloud.google.com/go/aiplatform/apiv1beta1/aiplatformpb"
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/protobuf/testing/protocmp"
+)
+
+func TestParseEnvVars(t *testing.T) {
+	tests := []struct {
+		name    string
+		raw     []string
+		want    []*aiplatformpb.EnvVar
+		wantErr bool
+	}{
+		{
+			name: "nil input returns empty",
+			raw:  nil,
+			want: []*aiplatformpb.EnvVar{},
+		},
+		{
+			name: "single key=value",
+			raw:  []string{"FOO=bar"},
+			want: []*aiplatformpb.EnvVar{{Name: "FOO", Value: "bar"}},
+		},
+		{
+			name: "value contains equals sign",
+			raw:  []string{"URL=http://example.com?a=1&b=2"},
+			want: []*aiplatformpb.EnvVar{{Name: "URL", Value: "http://example.com?a=1&b=2"}},
+		},
+		{
+			name: "empty value",
+			raw:  []string{"EMPTY="},
+			want: []*aiplatformpb.EnvVar{{Name: "EMPTY", Value: ""}},
+		},
+		{
+			name: "multiple vars",
+			raw:  []string{"A=1", "B=2", "C=three"},
+			want: []*aiplatformpb.EnvVar{
+				{Name: "A", Value: "1"},
+				{Name: "B", Value: "2"},
+				{Name: "C", Value: "three"},
+			},
+		},
+		{
+			name:    "missing equals returns error",
+			raw:     []string{"NOEQUALS"},
+			wantErr: true,
+		},
+		{
+			name:    "empty string returns error",
+			raw:     []string{""},
+			wantErr: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got, err := parseEnvVars(tc.raw)
+			if (err != nil) != tc.wantErr {
+				t.Fatalf("parseEnvVars(%v) error = %v, wantErr %v", tc.raw, err, tc.wantErr)
+			}
+			if tc.wantErr {
+				return
+			}
+			if diff := cmp.Diff(tc.want, got, protocmp.Transform()); diff != "" {
+				t.Errorf("parseEnvVars(%v) mismatch (-want +got):\n%s", tc.raw, diff)
+			}
+		})
+	}
+}


nuthalapativarun:

cmd/adkgo/internal/deploy/agentengine/agentengine.go
```diff
@@ -110,6 +111,21 @@ func init() {
...
+	agentEngineCmd.PersistentFlags().StringArrayVar(&flags.agentEngine.envVars, "env", nil, "Additional environment variable in KEY=VALUE format. Repeatable.")
+}
+
+// parseEnvVars converts a slice of "KEY=VALUE" strings into EnvVar protos.
+// Returns an error for any entry that does not contain "=".
+func parseEnvVars(raw []string) ([]*aiplatformpb.EnvVar, error) {
+	vars := make([]*aiplatformpb.EnvVar, 0, len(raw))
+	for _, s := range raw {
+		parts := strings.SplitN(s, "=", 2)
+		if len(parts) != 2 {
```
karolpiotrowicz, line 123: This guard is one case weaker than the issue asks for. `strings.SplitN("=value", "=", 2)` returns `["", "value"]`, which has length 2, so `--env==value` is accepted and produces an `EnvVar` with an empty `Name`. Issue #848 specifies the pattern `([^=]+)=(.*)`, and `[^=]+` requires at least one character before the `=`.

```go
got, err := parseEnvVars([]string{"=orphaned-value"})
// err == nil, got == [{Name: "", Value: "orphaned-value"}]
```

Rejecting `parts[0] == ""` and adding a `{raw: []string{"=value"}, wantErr: true}` row to the table covers it.

Worth noting what is *not* a problem here, since it looks like one at a glance: `"MY KEY"`, `" LEADING"`, `"lower-case"` and `"1STARTSWITHDIGIT"` are all accepted as names too, but every one of them matches `[^=]+`, so they conform to the pattern as written. Only the empty name is outside it.

cmd/adkgo/internal/deploy/agentengine/agentengine.go
```diff
@@ -255,6 +271,17 @@ func (f *deployAgentEngineFlags) gcloudDeployToAgentEngine() error {
 			}
 			p("Methods:", string(methodsJSON))
 
+			defaultEnv := []*aiplatformpb.EnvVar{
+				{Name: "GOOGLE_CLOUD_REGION", Value: f.gcloud.region},
+				{Name: "NUM_WORKERS", Value: "1"},
+				{Name: "GOOGLE_CLOUD_AGENT_ENGINE_ENABLE_TELEMETRY", Value: "true"},
+				{Name: "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT", Value: "true"},
+			}
+			extraEnv, err := parseEnvVars(f.agentEngine.envVars)
```
karolpiotrowicz, line 280: By the time this runs, `computeFlags` has created the temp directory, `prepareDockerfile` has written into it and `createArchive` has tarred the source tree. A typo like `--env=NOEQUALS` therefore costs a full archive build before it is rejected.

The error return also skips `cleanTemp()`, so the temp directory survives holding `Dockerfile` and `archive.tgz`. That skip is pre-existing on every error path in this function rather than something you introduced, so I would not ask you to fix it here — but validating `--env` early, in `computeFlags` or a cobra `PreRunE`, avoids reaching it at all and gives a faster error.

cmd/adkgo/internal/deploy/agentengine/agentengine.go
```diff
@@ -274,11 +301,9 @@ func (f *deployAgentEngineFlags) gcloudDeployToAgentEngine() error {
 						},
 						AgentFramework: "google-adk",
 						DeploymentSpec: &aiplatformpb.ReasoningEngineSpec_DeploymentSpec{
-							Env: []*aiplatformpb.EnvVar{
-								{Name: "GOOGLE_CLOUD_REGION", Value: f.gcloud.region},
-								{Name: "NUM_WORKERS", Value: "1"},
-								{Name: "GOOGLE_CLOUD_AGENT_ENGINE_ENABLE_TELEMETRY", Value: "true"},
-								{Name: "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT", Value: "true"},
+							Env: append(defaultEnv, extraEnv...),
```
karolpiotrowicz, line 304: Appending means a user value whose name matches one of the four built-in defaults is sent twice rather than replacing it. `--env=NUM_WORKERS=4` produces a list containing `NUM_WORKERS` at `1` and then at `4`.

Which one the API honours I could not determine without a live project. Extras land last, so if it is last-wins the override works as the user intended, and this is only a documentation gap. If it is first-wins the flag is silently ignored, and if duplicates are rejected the deploy fails.

Either reconcile by name so a user value replaces the matching default, or reject a `--env` name that collides with one of the four and say so. Naming the four reserved variables in the flag's help text would help regardless of which way you go.

cmd/adkgo/internal/deploy/agentengine/agentengine.go
```diff
@@ -374,6 +404,19 @@ func (f *deployAgentEngineFlags) gcloudUpdateAgentEngine() error {
...
 			}
 
+			if len(extraEnv) > 0 {
+				defaultEnv := []*aiplatformpb.EnvVar{
+					{Name: "GOOGLE_CLOUD_REGION", Value: f.gcloud.region},
+					{Name: "NUM_WORKERS", Value: "1"},
+					{Name: "GOOGLE_CLOUD_AGENT_ENGINE_ENABLE_TELEMETRY", Value: "true"},
+					{Name: "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT", Value: "true"},
+				}
+				req.ReasoningEngine.Spec.DeploymentSpec = &aiplatformpb.ReasoningEngineSpec_DeploymentSpec{
```
karolpiotrowicz, line 414: This builds a `DeploymentSpec` containing only `Env`, while the create path a few hundred lines up builds one containing `Env` **and** the new `SecretEnv`. Since the next line adds `spec.deployment_spec` to the update mask, an update carrying `--env` sends a spec that omits the secret binding the create path just set.

Whether that clears it on the server depends on whether a mask path naming a message replaces or merges it, which I could not test without a live project. The asymmetry between the two paths is visible in the code either way, and it disappears if the `SecretEnv` block comes out of this PR.

One caution if you unify the two: the `defaultEnv` literal is duplicated here and in the create path, and that duplication is currently what makes `append(defaultEnv, extraEnv...)` safe — each call gets a fresh length-4, capacity-4 slice, so the append always reallocates. Hoisting it to a package-level `var` would introduce an aliasing bug that does not exist today. A function returning a fresh slice avoids that.

karolpiotrowicz (commented):
The `--env` flag itself is well built: `StringArrayVar` is the right choice so a value containing commas survives intact, splitting on the first `=` only handles URLs and query strings correctly, and the table test covers the cases that matter for the parser. Two things need resolving before this can merge, and one of them is unrelated to the feature.

## The `SecretEnv` block is a separate change, and it applies to everyone

[`agentengine.go:305-307`](https://github.com/google/adk-go/blob/9a5528f86d7261feec9713e1ae2935d0e4d27dcb/cmd/adkgo/internal/deploy/agentengine/agentengine.go#L305-L307) adds a Cloud Secret Manager reference to every `CreateReasoningEngine` request:

```go
SecretEnv: []*aiplatformpb.SecretEnvVar{
    {Name: "GOOGLE_API_KEY", SecretRef: &aiplatformpb.SecretRef{Secret: "GOOGLE_API_KEY", Version: "latest"}},
},
```

It sits inside no conditional and behind no flag, so it applies to every deploy including those that pass no `--env` at all. It is also absent from the PR description and from the commit message, both of which otherwise enumerate the change completely.

I want to be fair about the precedent here, because it matters: the identical unconditional secret already ships on the sibling deploy target at [`cloudrun.go:256`](https://github.com/google/adk-go/blob/440e35202c325a7310557e3bacecdeb3ad07b989/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go#L256) on main today, so this is an existing pattern being extended rather than something new to the repo. That makes it a defensible thing to want. It does not make it part of this PR, which issue #848 scopes to adding `--env`.

Two consequences worth knowing before deciding what to do with it. The proto documents an IAM prerequisite that nothing in this change mentions — `DeploymentSpec.SecretEnv` in `aiplatform@v1.126.0` says *"To use this feature, add 'Secret Manager Secret Accessor' role (roles/secretmanager.secretAccessor) to AI Platform Reasoning Engine Service Agent"*. And I could not determine what the Agent Engine API does when the referenced secret does not exist in the project, because I have no GCP project to try it against — that is the one thing I would want checked before this lands in any form.

My suggestion is to drop it from this PR and raise it separately, which also resolves the create/update mismatch noted on the diff below.

## `--env` values are printed to stdout

The flag's own description names *"third-party API keys"* as a use case, and those values are echoed twice on the way through.

[`agentengine.go:457`](https://github.com/google/adk-go/blob/9a5528f86d7261feec9713e1ae2935d0e4d27dcb/cmd/adkgo/internal/deploy/agentengine/agentengine.go#L457) is `fmt.Println(flags)`, which prints the whole flag struct. That line predates this PR and `cloudrun.go` does the same, so the line is not yours — but before this change the struct held only regions, paths and ports, and adding `envVars` to it is what turns a harmless debug print into one that emits credentials.

[`agentengine.go:124`](https://github.com/google/adk-go/blob/9a5528f86d7261feec9713e1ae2935d0e4d27dcb/cmd/adkgo/internal/deploy/agentengine/agentengine.go#L124) then echoes the offending value again in the error text, and that one is new code. It fires exactly when someone pastes a value and forgets the `KEY=` prefix, which is the most likely way to get a bare secret onto the command line.

Both reproduce through the real CLI:

<details><summary>Observed stdout (values are synthetic)</summary>

```
$ adkgo deploy agentengine --env=THIRD_PARTY_API_KEY=s3cr3t-third-party-key ...
{{ } {  8080  {false publishers/google/models/gemini-2.5-flash 31536000000000000} [THIRD_PARTY_API_KEY=s3cr3t-third-party-key]} { } { }}
Computing flags & preparing temp : Starting

$ adkgo deploy agentengine --env=AIzaSyFAKE-pasted-key-2222 ...
{{us-central1 fake-project} {  8080  ... [AIzaSyFAKE-pasted-key-2222]} ...}
...
Deploying to Agent Engine : Finished with error
Error: invalid --env value "AIzaSyFAKE-pasted-key-2222": expected KEY=VALUE
```

</details>

A local terminal is one thing, but in CI the job log is often the only durable copy of that value. adk-python hit the same question and went the other way deliberately — [`cli_deploy.py:1360-1361`](https://github.com/google/adk-python/blob/c506ddf3bc34a6312ffc81899221bfa3f2da3b1d/src/google/adk/cli/cli_deploy.py#L1360-L1361) carries the comment *"sorted() over a dict iterates its keys, so this prints the variable names only and never their values."* Printing names and a count would match that, and a fix needs to cover both sites, since patching only the struct print leaves the error path still emitting the value.

## Worth saying out loud: this diverges from adk-python on purpose

adk-python is moving away from a CLI flag for this. Its equivalent is `--env_file`, a dotenv file, and that flag is explicitly deprecated — env vars reach a deployment through `agent_config['env_vars']` instead. So Go would be adding a repeatable `--env` flag as Python retires its own.

That looks like a deliberate choice rather than an oversight, since kdroste-google specified the flag name, its repeatability and the `([^=]+)=(.*)` pattern in #848. Flagging it only so the divergence is a decision on the record rather than something noticed later.

## Rebase needed

The branch no longer merges. Main gained its own `agentengine_test.go` in [2722dfd9](https://github.com/google/adk-go/commit/2722dfd99e7c0b6eb14da872ef98c4e3e1746bdf) ("derive Agent Engine builder Go version from go.mod"), so the two test files collide. The `.go` file still merges cleanly, and the two sets of tests are independent, so the resolution is just a union of both with the imports merged. Reviewed against that merge result, the build, the full suite, vet, lint and the module tidy check are all clean, with no new test failures against main.

Also worth checking before you spend more time: kdroste-google told Allmight-456 "sounds good" on #848 back in May, so there may be a parallel implementation nobody closed.

## Smaller points

These are on the diff lines themselves rather than here.

Two more that have no single line to attach to. Nothing exercises the flag's wiring — the test covers `parseEnvVars` well, but renaming the flag, swapping `StringArrayVar` for `StringSliceVar`, or reversing the append order would all leave the suite green. That is not really this PR's fault, since `cmd/adkgo` has almost no test harness to build on and the request path constructs a live client inline, so I would not hold the PR for it. Separately, because `flags` is a package global and `StringArrayVar` appends rather than replaces, a second parse in the same process accumulates the first run's values — the shipped binary calls `Execute()` once so this only reaches in-process tests, but it will bite anyone who writes one.

nuthalapativarun:
Rebased onto current `main` — resolved an add/add conflict in `cmd/adkgo/internal/deploy/agentengine/agentengine_test.go` where `main` had independently added a new test file (Go-version-from-go.mod and shell-arg-safety tests) at the same path this PR added `TestParseEnvVars` to; merged so both sets of tests are present. `go build ./...` passes and the package's tests run cleanly aside from a pre-existing, unrelated Windows path-separator issue in a couple of the go.mod-version tests that already fails the same way on `main` before this branch's changes are applied. Force-pushed the rebase. This PR hasn't had a first review yet — would appreciate a look when you have a chance. Thanks!

cmd/adkgo/internal/deploy/agentengine/agentengine.go
```diff
@@ -111,6 +112,21 @@ func init() {
...
+}
+
+// parseEnvVars converts a slice of "KEY=VALUE" strings into EnvVar protos.
+// Returns an error for any entry that does not contain "=".
+func parseEnvVars(raw []string) ([]*aiplatformpb.EnvVar, error) {
+	vars := make([]*aiplatformpb.EnvVar, 0, len(raw))
+	for _, s := range raw {
+		parts := strings.SplitN(s, "=", 2)
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid --env value %q: expected KEY=VALUE", s)
```
kdroste-google, line 125: Looking at https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html
It seems [A-Z_][A-Z0-9_]* or [a-zA-Z_][a-zA-Z0-9_]* may be used to validate the name.

As mentioned above, the code as-is allows edge case - an empty variable name.

cmd/adkgo/internal/deploy/agentengine/agentengine.go
```diff
@@ -111,6 +112,21 @@ func init() {
...
 	agentEngineCmd.PersistentFlags().DurationVar(&flags.agentEngine.memoryBank.ttl, "mem_ttl", time.Hour*24*365, "Time-To-Live for memories")
+	agentEngineCmd.PersistentFlags().StringArrayVar(&flags.agentEngine.envVars, "env", nil, "Additional environment variable in KEY=VALUE format. Repeatable.")
+}
+
+// parseEnvVars converts a slice of "KEY=VALUE" strings into EnvVar protos.
+// Returns an error for any entry that does not contain "=".
+func parseEnvVars(raw []string) ([]*aiplatformpb.EnvVar, error) {
+	vars := make([]*aiplatformpb.EnvVar, 0, len(raw))
+	for _, s := range raw {
+		parts := strings.SplitN(s, "=", 2)
```
kdroste-google, line 123: Please consider using Cut, I think it's simpler

cmd/adkgo/internal/deploy/agentengine/agentengine.go
```diff
@@ -348,6 +364,17 @@ func (f *deployAgentEngineFlags) gcloudDeployToAgentEngine() error {
 			}
 			p("Methods:", string(methodsJSON))
 
+			defaultEnv := []*aiplatformpb.EnvVar{
```
kdroste-google, line 367: now we are allowing to provide custom env vars. And we have some defaults.

The user should be able to override the defaults. 
Proposal:  
1) a new flag --default_env (default: true) which will conditionally add defaults (opt-out)
2) instead of adding new custom aiplatformpb.EnvVar we should rather merge them (overriding defaults with custom).

I'm aware that by doing that the user who wants all the defaults except of one will have to turn off the defaults and provide the list without unwanted name. But it seems doable, just 4 of them. 

So, maybe it would be easier to use maps to do this operations (which can go to computeFlags) and at the end pack them to []aiplatformpb.EnvVar
@nuthalapativarun, WDYT?

cmd/adkgo/internal/deploy/agentengine/agentengine.go
```diff
@@ -467,6 +497,19 @@ func (f *deployAgentEngineFlags) gcloudUpdateAgentEngine() error {
...
 
+			if len(extraEnv) > 0 {
+				defaultEnv := []*aiplatformpb.EnvVar{
+					{Name: "GOOGLE_CLOUD_REGION", Value: f.gcloud.region},
+					{Name: "NUM_WORKERS", Value: "1"},
+					{Name: "GOOGLE_CLOUD_AGENT_ENGINE_ENABLE_TELEMETRY", Value: "true"},
+					{Name: "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT", Value: "true"},
+				}
+				req.ReasoningEngine.Spec.DeploymentSpec = &aiplatformpb.ReasoningEngineSpec_DeploymentSpec{
+					Env: append(defaultEnv, extraEnv...),
```
kdroste-google, line 508: as mentioned above, should be rather override (like on map), not append (on slices)

cmd/adkgo/internal/deploy/agentengine/agentengine_test.go
```diff
@@ -53,6 +57,70 @@ func TestGoVersionFromModFile(t *testing.T) {
...
+		{
+			name: "multiple vars",
+			raw:  []string{"A=1", "B=2", "C=three"},
+			want: []*aiplatformpb.EnvVar{
+				{Name: "A", Value: "1"},
+				{Name: "B", Value: "2"},
+				{Name: "C", Value: "three"},
+			},
+		},
+		{
```
kdroste-google, line 96: "A=1", "A=2" ?

kdroste-google: Either the last one should win, or better - and error complaining about the lack of uniqueness

cmd/adkgo/internal/deploy/agentengine/agentengine_test.go
```diff
@@ -53,6 +57,70 @@ func TestGoVersionFromModFile(t *testing.T) {
...
+			raw:     []string{"NOEQUALS"},
+			wantErr: true,
+		},
+		{
+			name:    "empty string returns error",
+			raw:     []string{""},
+			wantErr: true,
+		},
+	}
+
```
kdroste-google, line 107: missing overrides of defaults

kdroste-google (changes requested):
Hi, @nuthalapativarun , thank you for providing this PR. 
Please take a look at the comments - I think it could be better handling defaults & overrides 
What do you think?

PR #985, opened 2026-06-08
feat: add ValidateInput override to JoinNode to validate individual predecessor outputs

hanorik:
This PR adds a `ValidateInput` override to `JoinNode` to validate the outputs of individual predecessor nodes against the node's input schema, rather than validating the aggregated map as a single value.

### Key Changes
- **`JoinNode.ValidateInput` implementation**:
  - Overrides the default base node input validation.
  - If the input is a `map[string]any` containing outputs from multiple predecessor nodes, it validates each predecessor's output value individually against `InputSchema()`.
  - Returns a map of validated values or returns a wrapped validation error including the predecessor's name (formatted as `predecessor "X": <inner error>`).
  - If the input is not a `map[string]any` (e.g., a raw single predecessor value), it falls back to `BaseNode.ValidateInput`.
  - If the node's `InputSchema()` is `nil`, the input is returned as-is.

### Automated Tests
Added unit tests in [workflow/join_node_test.go](file:///usr/local/google/home/hanorik/adk-go/workflow/join_node_test.go) to verify:
- **All match schema**: Validation succeeds when three predecessor values all match the schema, returning the map of validated values.
- **One fails validation**: Returns a detailed error matching the format `predecessor "X": <inner error>` when one of three predecessor values fails validation.
- **Not a map falls back**: Correctly falls back to `BaseNode.ValidateInput` when the input is a raw value rather than a `map[string]any`.
- **Nil schema returns as-is**: Simply returns the input as-is when `InputSchema()` is `nil`.

### DIFF
--- a/workflow/join_node.go
+++ b/workflow/join_node.go
@@ -15,7 +15,10 @@
 package workflow
 
 import (
+	"fmt"
 	"iter"
+
+	"github.com/google/jsonschema-go/jsonschema"
 
 	"google.golang.org/adk/agent"
 	"google.golang.org/adk/session"
@@ -40,6 +43,18 @@ func NewJoinNode(name string) *JoinNode {
 	return &JoinNode{BaseNode: NewBaseNode(name, "", NodeConfig{})}
 }
 
+// NewJoinNodeWithSchema returns a JoinNode with the given name and input schema.
+//
+// The input schema is applied to each predecessor's output individually when validation is performed,
+// rather than being applied to the combined map structure itself.
+//
+// If a predecessor's output is nil, validation is bypassed and the nil value is preserved.
+func NewJoinNodeWithSchema(name string, schema *jsonschema.Resolved) *JoinNode {
+	return &JoinNode{
+		BaseNode: NewBaseNodeWithSchemas(name, "", NodeConfig{}, schema, nil),
+	}
+}
+
 // Run satisfies the Node interface. See JoinNode for the
 // aggregation contract.
 func (n *JoinNode) Run(ctx agent.InvocationContext, input any) iter.Seq2[*session.Event, error] {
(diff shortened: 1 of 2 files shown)


workflow/join_node.go
```diff
@@ -49,3 +50,26 @@ func (n *JoinNode) Run(ctx agent.InvocationContext, input any) iter.Seq2[*sessio
 		yield(event, nil)
 	}
 }
+
+// ValidateInput overrides BaseNode.ValidateInput. Instead of validating the
+// aggregated map as a single value, it validates each predecessor's output
+// individually against the node's InputSchema.
+func (n *JoinNode) ValidateInput(input any) (any, error) {
+	schema := n.InputSchema()
```
wolo-lab, line 72: We could add a new constructor so creating a new JoinNode with schemas will be simpler than:

```
	joinWithSchema := &JoinNode{
		BaseNode: NewBaseNodeWithSchemas("join", "", NodeConfig{}, stringSchema, nil),
	}
```

in a new constructor please add a doc explaining how schema is applied (there are two possible interpretations: schema for each value in the map (recommended) or schema for a map. Add also a note how nil in the map is handled.

workflow/join_node_test.go
```diff
@@ -229,3 +232,86 @@ func inputRecorder(name string, seen *any) *FunctionNode {
 			return "done", nil
 		}, defaultNodeConfig)
 }
+
+func TestJoinNode_ValidateInput(t *testing.T) {
```
wolo-lab, line 236: please add a test for nil in the map and empty map.

wolo-lab (approved):

workflow/join_node.go
hanorik: Done.

workflow/join_node_test.go
hanorik: Done.

PR #989, opened 2026-06-09
feat: create contents for isolated agents

dpasiukevich:
Scoped agents (task/single_turn) are invoked by task-delegation FC. This commit handles proper content construction for such agents: making FC args visible to the agent + adding SingleTurnNudge text part for single_turn agent.

### DIFF
--- a/internal/llminternal/contents_processor.go
+++ b/internal/llminternal/contents_processor.go
@@ -42,8 +42,9 @@ func ContentsRequestProcessor(ctx agent.InvocationContext, req *model.LLMRequest
 			// Do nothing.
 			return // In python, no error is yielded.
 		}
+		state := llmAgent.internal()
 		fn := buildContentsDefault // "" or "default".
-		if llmAgent.internal().IncludeContents == "none" {
+		if state.IncludeContents == "none" {
 			// Include current turn context only (no conversation history)
 			fn = buildContentsCurrentTurnContextOnly
 		}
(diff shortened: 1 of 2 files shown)


internal/llminternal/contents_processor.go
```diff
@@ -603,6 +618,73 @@ func shouldExcludeEvent(ev *session.Event) bool {
 	return false
 }
 
+// SingleTurnNudge is appended as a second user-content text part for
+// single_turn agents.
+const SingleTurnNudge = "Important: You will not receive any user replies or clarifications." +
```
wolo-lab, line 623: shouldn't it be a system instruction?

internal/llminternal/contents_processor.go
```diff
@@ -603,6 +618,73 @@ func shouldExcludeEvent(ev *session.Event) bool {
...
+// the rendered node input). Returns nil if neither source yields
+// content.
+//
+// When isSingleTurn is true, appends singleTurnNudge as an additional
+// user-content text part.
+func buildTaskInputUserContent(events []*session.Event, isolationScope string, isSingleTurn bool, userContent *genai.Content) *genai.Content {
+	if isolationScope == "" {
+		return nil
+	}
+	for _, ev := range events {
```
wolo-lab, line 648: Python iterates in the reverse order (`_finish_task_tool.py`), so the newest event is used. Should we do the same?

wolo-lab (approved):

internal/llminternal/contents_processor.go
dpasiukevich: It appears to be the normal order:  `for event in all_events`:

https://github.com/google/adk-python/blob/4006fe408c583bffd95c23455f00184baa4fad6d/src/google/adk/flows/llm_flows/contents.py#L337

Could you link the python code? Wanted to make sure we didn't miss anything.

internal/llminternal/contents_processor.go
dpasiukevich: It's part of the request content, not the system instruction in python's flow. I kept it similar to easier to reason about.

wolo-lab: I was comparing against the old python's branch. Sorry!

internal/llminternal/contents_processor.go
wolo-lab: I was comparing against the old python's branch. Sorry!

PR #990, opened 2026-06-09
feat: add NewFunctionNodeFromState to support binding node inputs and state fields via struct parameters

hanorik:
This PR introduces the `NewFunctionNodeFromState` constructor to support binding custom function parameters directly from session state and predecessor outputs using struct parameters.

### Key Changes
- **New Constructor `NewFunctionNodeFromState[Params, OUT]`**: Added support for constructing function nodes with struct parameters mapping fields to session state values.
- **Struct Parameter Validation**: Rejects non-struct `Params` types at construction time with a clear error.
- **Flexible Field-Name Mapping**: Resolves Go field names as the state keys by default, allowing overrides via `state:"X"` struct tags.
- **Predecessor Output Binding (`node_input`)**: Supports binding the raw predecessor output (instead of state values) to a field named `NodeInput` or tagged with `state:"node_input"`.
- **Per-field Type Coercion**: Leverages `ConvertToWithJSONSchema` for type coercion and conversion of state/input parameters.
- **Missing State Key Validation**: Validates session state values at execution time, returning a clear error if any required field is missing.
- **Exclusion of Predecessor Output from State Schema**: Implements the `StateParamsAware` interface via `StateFieldNames()`, which returns only state-key fields and explicitly excludes the `"node_input"` field.

### Test Coverage
Added comprehensive tests covering the following scenarios:
- Successful instantiation and execution with a two-field struct where the state contains both keys.
- State mapping overrides via `state:"X"` tag.
- Predecessor output mapping using `NodeInput` (and `state:"node_input"` tag).
- Validation errors when state keys are missing.
- Type coercion/type mismatch validation errors.

### DIFF
--- a/workflow/function_node_test.go
+++ b/workflow/function_node_test.go
@@ -15,6 +15,8 @@
 package workflow
 
 import (
+	"fmt"
+	"iter"
 	"strings"
 	"testing"
 
@@ -22,6 +24,7 @@ import (
 	"github.com/google/jsonschema-go/jsonschema"
 
 	"google.golang.org/adk/agent"
+	"google.golang.org/adk/session"
 )
 
 func TestNewFunctionNodeWithSchema(t *testing.T) {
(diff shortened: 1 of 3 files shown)


workflow/workflow.go
```diff
@@ -51,6 +51,13 @@ type Node interface {
 	Run(ctx agent.InvocationContext, input any) iter.Seq2[*session.Event, error]
 }
 
+// StateParamsAware is implemented by nodes that bind their parameters
+// from ctx.state. Used by Workflow's static state-schema validation
+// to detect mismatches between declared state fields and node consumers.
+type StateParamsAware interface {
```
wolo-lab: It is not used anywhere. I think it is save to remove it.

workflow/function_node_test.go
wolo-lab, line 1: Please add one test in which output is a struct. There might be some issues with the output validation in the current implementation. I would like to be sure that it's working fine.

wolo-lab (changes requested):

wolo-lab (approved):

PR #991, opened 2026-06-09
refactor: consolidate agent node input conversion into a helper and add validation test coverage

hanorik:
Consolidated AgentNode input validation and conversion logic. Input schemas are now fully integrated with BaseNode, and input data is parsed into `genai.Content` before invoking the wrapped agent.

### Key Changes
* **Schema Handling**: Removed local `inputSchema` / `outputSchema` fields from `AgentNode`. These schemas now reside entirely in `BaseNode` and are passed via the constructor.
* **Input Conversion Helper**: Introduced `nodeInputToContent`, which handles:
  - `nil`
  - `*genai.Content`
  - `string`
  - `json.Marshaler`
  - Fallback/default types (struct / map / list / primitives) using JSON marshaling.
* **AgentNode Execution**: `AgentNode.Run` converts the validated input into a `*genai.Content` and passes it to the wrapped agent, which observes it as `InvocationContext.UserContent()`.

### Tests Added
* **Struct input** → JSON text part is delivered to the agent.
* **String input** → Text part is delivered directly.
* **Nil input** → Agent runs without overriding the user-content.
* **Input fails InputSchema validation** → Scheduler rejects the activation, and the agent is never invoked.


workflow/agent_node.go
```diff
@@ -76,31 +76,13 @@ func NewAgentNode(a agent.Agent, cfg NodeConfig) (*AgentNode, error) {
...
-				b, err := json.Marshal(v)
-				if err != nil {
-					yield(nil, fmt.Errorf("marshaling input for agent %q to JSON: %w", n.agent.Name(), err))
-					return
-				}
-				userContent = &genai.Content{
-					Parts: []*genai.Part{{Text: string(b)}},
-				}
-			}
+		if userContent == nil {
```
wolo-lab: I would remove this fallback. Is there a strong reason to keep it?

wolo-lab (approved):

workflow/agent_node.go
hanorik: I thought that there might be situations when agent node could receive `nil` input so the agent could reference the original `UserContent` rather than being invoked with completely empty context, wdyt?

wolo-lab: please verify what adk python does in this scenario.

hanorik: Seems that they don't fallback and just provide `None` input.

hanorik: Removed this fallback.

PR #992, opened 2026-06-09
feat: add utility to generate canonical, key-sorted JSON for schemas and nested maps

hanorik:
This PR implements a deterministic canonicalizer for JSON schemas to enable byte-for-byte schema comparison. It recursively sorts object keys alphabetically while preserving array orders and primitive values.

### Changes
*   **`canonicalSchemaJSON`**: A package-private utility that marshals a `jsonschema.Schema`, parses it back, and outputs a canonicalized JSON byte slice.
*   **`canonicalize`**: A package-private recursive helper that encodes maps with sorted keys, preserves slice element order, and delegates primitives to `encoding/json`.
*   **Dependencies**: Uses only `encoding/json` and `sort` from the Go standard library, introducing no new external dependencies.

### Test Coverage
*   **Property Order**: Validates that two schemas differing only in `PropertyOrder` produce identical canonical output.
*   **Nested Objects**: Confirms that deeply nested objects (3+ levels) have keys sorted recursively.
*   **Arrays**: Verifies that arrays preserve their input order.
*   **Primitives**: Verifies that primitive values (string, number, bool, and nil) round-trip unchanged.


wolo-lab (approved):
maybe let's rename a file. Right now it's not super clear what's inside. Maybe `scheme_canonical.go`? Maybe it could be also moved to `internal/utils`?

hanorik:
> maybe let's rename a file. Right now it's not super clear what's inside. Maybe `scheme_canonical.go`? Maybe it could be also moved to `internal/utils`?

SG, renamed and moved.

PR #993, opened 2026-06-09
feat: add static schema validation to workflow graph construction

hanorik:
This PR implements static schema validation on workflow graph edges during workflow initialization, establishing parity with Python validation behavior.

## Key Changes

* **Static Edge Schema Validation**: Added `validateStaticSchemas` helper function which is invoked as part of `validateWorkflow` during `workflow.New`.
* **Selective Validation**: Validation is only run if both endpoint nodes of an edge declare input/output schemas (i.e. both `edge.From.OutputSchema()` and `edge.To.InputSchema()` are non-nil). If either schema is missing, validation is skipped for that edge.
* **Canonical JSON Schema Comparison**: Uses `canonicalSchemaJSON` to compare schemas, ensuring that schemas differing only in property ordering (e.g., `PropertyOrder` field order) are successfully canonicalized and treated as matching.

## Tests

Added `TestStaticSchemaValidation` in `workflow/validation_test.go` covering the following cases:
- **Same Go type** (using `jsonschema.For[T]`): Edge validation succeeds and workflow initialization completes.
- **Different Go types**: Workflow initialization fails, returning an error naming both nodes (e.g., `schema mismatch on edge A -> B`).
- **Single endpoint schema**: Validation is skipped when only one endpoint on the edge defines a schema, and workflow initialization succeeds.
- **`PropertyOrder` mismatch**: Two schemas that differ only in property ordering successfully pass validation due to schema canonicalization.


workflow/graph.go
```diff
@@ -36,6 +36,16 @@ func newGraph(edges []Edge) *graph {
 	return &graph{successors: succ, predecessors: pred}
 }
 
+// allEdges returns all unique edges in the graph.
```
wolo-lab: nit: I don't think we have deduplication. We can just remove "unique"

workflow/validation_test.go
wolo-lab, line 1: we could add one more test: different types having the same fields. It's nice improvement over adk python's solution.

wolo-lab (approved):

workflow/graph.go
hanorik: Agree, removed.

workflow/validation_test.go
hanorik: Done.

PR #995, opened 2026-06-10
feat: reject task mode llmagents as static graph nodes

dpasiukevich:


wolo-lab (approved):
Minors:
- we could use a sentinel error (we already have ErrDuplicateNodeName or ErrNoStartNode)
- the number of files grow, maybe we could add a test from validation_taskmode_test.go to validation_test.go?

dpasiukevich:
1. I suggest not to add the sentinel -- not sure if this specific sentinel error will be useful for the user. Let me know if you would like to add it, I can do it, I don't have strong opinion here.
2. It's a different package name workflow_test with different imports (e.g. it's not possible to import llmagent in the workflow package due to cycle, because llmagent package needs workflow logic).

PR #999, opened 2026-06-10
feat: introduce WorkflowConfig to support state schema validation during workflow construction

hanorik:
This PR introduces static workflow schema validation to ensure that any state fields referenced by workflow nodes are declared in the workflow's state schema.

## Key Changes

* **`WorkflowConfig`**: Added a new configuration struct `WorkflowConfig` that contains a `StateSchema` field of type `*jsonschema.Resolved`.
* **`workflow.New` Integration**: Updated `workflow.New` signature to accept `WorkflowConfig`.
* **Validation Wiring**: Integrated `validateStateSchemaConsistency` validation into the workflow initialization process.
* **Special Field Handling**: The special `node_input`/`NodeInput` field is explicitly excluded from schema consistency checks since it represents input passed from predecessor nodes rather than a state key.
* **Skip & Ignore Behavior**:
  * Skip validation completely if `StateSchema` is `nil`.
  * Ignore nodes that do not implement the `StateParamsAware` interface (e.g. `AgentNode` or `ToolNode`).

## Tests

Added comprehensive test cases in `workflow_config_test.go` verifying that:
1. Valid parameter matching succeeds (e.g., parameter tag matches schema field name).
2. State schema parameter typos or case mismatches (e.g. tag `state:"foo"` but schema declares `"Foo"`) produce a descriptive validation error containing both the node name and the missing field name.
3. Validation is skipped when `StateSchema` is `nil`.
4. Nodes without `StateParamsAware` are ignored.

### DIFF
--- a/workflow/graph.go
+++ b/workflow/graph.go
@@ -52,3 +52,19 @@ func (g *graph) successorsOf(n Node) []Edge {
 func (g *graph) predecessorsOf(n Node) []Edge {
 	return g.predecessors[n]
 }
+
+// allNodes returns all nodes in the graph.
+func (g *graph) allNodes() []Node {
+	nodes := make(map[Node]bool)
+	for n := range g.successors {
+		nodes[n] = true
+	}
+	for n := range g.predecessors {
+		nodes[n] = true
+	}
+	var res []Node
+	for n := range nodes {
+		res = append(res, n)
+	}
+	return res
+}
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -51,6 +51,13 @@ type Node interface {
 	Run(ctx agent.InvocationContext, input any) iter.Seq2[*session.Event, error]
 }
 
+// StateParamsAware is implemented by nodes that bind their parameters
+// from ctx.state. Used by Workflow's static state-schema validation
+// to detect mismatches between declared state fields and node consumers.
+type StateParamsAware interface {
+	StateFieldNames() []string
+}
+
 // Route defines the interface for matching execution results to edges.
 type Route interface {
 	Matches(event *session.Event) bool
@@ -144,6 +151,9 @@ type Workflow struct {
 	// may run concurrently within a single Run invocation. 0
 	// (the default) means unlimited. Set via WithMaxConcurrency.
 	maxConcurrency int
+
+	// stateSchema holds the state schema for the workflow.
+	stateSchema *jsonschema.Resolved
 }
 
 // Option configures a Workflow at construction time. Pass options
@@ -155,6 +165,7 @@ type Option func(*workflowOptions)
 // engine defaults".
 type workflowOptions struct {
 	maxConcurrency int
+	stateSchema    *jsonschema.Resolved
 }
 
 // WithMaxConcurrency caps how many graph-scheduled nodes may run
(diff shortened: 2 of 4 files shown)


workflow/workflow_config.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"fmt"
+	"slices"
+
+	"github.com/google/jsonschema-go/jsonschema"
+)
+
+// WorkflowConfig configures the workflow at construction time.
+type WorkflowConfig struct {
+	StateSchema *jsonschema.Resolved
```
wolo-lab: Maybe`WithStateSchema` as `Option`? Struct has just one field and requires changing `workflow.New`. It would be consistent with `WithMaxConcurrency`.

```go
type workflowOptions struct {
    maxConcurrency int
    stateSchema    *jsonschema.Resolved
}
```

workflow/graph.go
```diff
@@ -52,3 +52,19 @@ func (g *graph) successorsOf(n Node) []Edge {
 func (g *graph) predecessorsOf(n Node) []Edge {
 	return g.predecessors[n]
 }
+
+// allNodes returns all nodes in the graph.
+func (g *graph) allNodes() []Node {
```
wolo-lab, line 57: We calculate allNodes() in other places, we could refactor:

// workflow/validation.go:177 
// workflow/validation.go:255

workflow/workflow.go
```diff
@@ -207,6 +217,9 @@ func New(name string, edges []Edge, opts ...Option) (*Workflow, error) {
 	if err := validateWorkflow(graph); err != nil {
```
wolo-lab: we could move validateStateSchemaConsistency inside validateWorkflow as we have all checks there

workflow/workflow_config.go
```diff
@@ -0,0 +1,58 @@
...
+		return nil
+	}
+	schemaFields := extractFieldNames(schema)
+
+	for _, n := range g.allNodes() {
+		spa, ok := n.(StateParamsAware)
+		if !ok {
+			continue
+		}
+		for _, fieldName := range spa.StateFieldNames() {
```
wolo-lab: should we skip field names with standard prefixes?

```go
    if strings.HasPrefix(fieldName, session.KeyPrefixApp) ||
       strings.HasPrefix(fieldName, session.KeyPrefixUser) ||
       strings.HasPrefix(fieldName, session.KeyPrefixTemp) {
        continue
    }
````

wolo-lab (approved):

workflow/workflow_config.go
hanorik: From ticket description it felt like other fields will be added to the structure later. If we don't expect any other fields, I'm completely fine with this option.

workflow/graph.go
hanorik: In first case it will require some additional refactoring as all nodes extraction happens basing on edges instead of graph. So I could leave it as it is and refactored only the second one.

workflow/workflow.go

workflow/workflow_config.go
hanorik: SG, done.

workflow/workflow_config.go
wolo-lab: we can refactor later when we will have more fields.

hanorik: Ok, I've added `stateSchema` to the `workflowOptions` with a method to set the schema for the workflow. And as we don't need a separate structure for the workflow config I also moved validate methods to the `validate.go` and tests to the `workflow_test.go`.

PR #1000, opened 2026-06-10
feat(workflow): add NewEmittingFunctionNode for streaming and HITL function node

wolo-lab:
**Problem**
A `FunctionNode` emits exactly one event built from its return value, so
a plain function body cannot emit a HITL `RequestInput`, a routing event
(`Event.Routes`), or progress updates. Authors who needed any of these had
to drop down to a custom `workflow.Node` (`BaseNode` subclass) just to call
`yield`. adk-python supports this directly via generator-style function
nodes; adk-go had no equivalent.

**Solution**
Add `NewEmittingFunctionNode` / `NewEmittingFunctionNodeWithSchema` taking
an `EmittingFunctionFn[IN, OUT]` whose body receives an `emit` callback. The
shape and HITL contract mirror `DynamicFn`: emit any number of intermediate
events, and return `ErrNodeInterrupted` after a `RequestInput` to pause. No
scheduler/resume/persistence changes are needed — the existing
pause/handoff machinery applies unchanged. `NewFunctionNode` is untouched.

Simplifies five workflow examples (hitl_simple, dynamic/hitl, routing
int/string/llm) that previously hand-rolled a `BaseNode` subclass solely to
emit a single custom event.


workflow/function_node.go
```diff
@@ -30,10 +30,25 @@ import (
...
+	ctx NodeContext, input IN, emit func(*session.Event) error,
+) (OUT, error)
+
 // FunctionNode wraps a custom function.
+//
+// Exactly one of fn or emittingFn is set per node.
 type FunctionNode struct {
 	BaseNode
 	fn              func(ctx agent.InvocationContext, input any) (any, error)
+	emittingFn      func(ctx NodeContext, input any, emit func(*session.Event) error) (any, error)
```
kdroste-google: It seems that the function cannot emit multiple errors. The only way to emit an error is to return it. Is that the right way?

workflow/function_node.go
```diff
@@ -241,9 +286,48 @@ func newFunctionNodeWithResolvedSchemas[IN, OUT any](name string, fn func(ctx ag
...
+					return nil, fmt.Errorf("new function node: invalid input type, expected %T: %w", new(IN), err)
+				}
+				t = converted
+			}
+			typedInput = t
+		}
+
+		output, err := fn(ctx, typedInput, emit)
+		if err != nil {
+			return output, err
```
kdroste-google: we are returning the as-is output anyway. Seems not to be consistent with the usual nil,error or non-nil, nil pairs

workflow/function_node.go
```diff
@@ -271,3 +355,35 @@ func (n *FunctionNode) Run(ctx agent.InvocationContext, input any) iter.Seq2[*se
...
+	emit := makeEmit(yield, nodeCtx)
+	output, err := n.emittingFn(nodeCtx, input, emit)
+	if err != nil {
+		if errors.Is(err, ErrNodeInterrupted) {
+			return
+		}
+		yield(nil, err)
+		return
+	}
+	if output == nil {
```
kdroste-google, line 387: For a no-op function there is no trace of completion. To be considered: emit an event with invocation id and no content

workflow/function_node.go
wolo-lab: Yes, this is intentional. A node has exactly one terminal outcome: either (output, nil) or (nil, err). emit is only for intermediate events (progress, state deltas, HITL RequestInput), not for error reporting. An error terminates the node, so there can only be one and return is a dedicated way to do it. Its mirrors how dynamic node was implemented. Non-breaking issues could be emitted using a regular event with extra info.

workflow/function_node.go
wolo-lab: Good catch. Fixed.

workflow/function_node.go
wolo-lab: Valid point. It makes sense. I've updated it for dynamic_nodes as well.

wolo-lab: I will handle it in a follow-up PR as it is a bigger change.

workflow/function_node.go
kdroste-google: Thanks for the clarification!

examples/workflow/hitl_rerun/main.go
```diff
@@ -0,0 +1,98 @@
...
+
+func main() {
+	ctx := context.Background()
+
+	rerun := true
+	greet := workflow.NewEmittingFunctionNode[any, any]("greet",
+		func(nc workflow.NodeContext, _ any, emit func(*session.Event) error) (any, error) {
+			// Resume branch: the node was re-run after the human
+			// replied, so the answer is available here.
+			if reply, ok := nc.ResumedInput("ask_name"); ok {
```
kdroste-google: Here the order of actions is not natural. Maybe we should have something like

answer, request, gotAnswer := GetAnswerOrRaiseQuestion( "ask_name", func() session.RequestInput { return session.RequestInput{ Message: "AAA", Payload: "BBB" }})
if ! gotAnswer { 
  return request 
} 
// got answer

And GetAnswerOrRaiseQuestion will use the first argument to both get answer & raise question

examples/workflow/hitl_simple/main.go
```diff
@@ -48,65 +45,39 @@ import (
...
-	hello := mkNode("greet", "greets the user by name", greet)
+	ask := workflow.NewEmittingFunctionNode[any, any]("ask_name",
+		func(ctx agent.Context, _ any, emit func(*session.Event) error) (any, error) {
+			if err := emit(workflow.NewRequestInputEvent(ctx, session.RequestInput{
+				InterruptID: "ask_name",
+				Message:     "What's your name?",
+			})); err != nil {
+				return nil, err
+			}
+			return nil, workflow.ErrNodeInterrupted
```
kdroste-google, line 59: QQ: What's the logic behind, after the node returns nil, workflow.ErrNodeInterrupted  how  the node executiion is completed and the workflow moves to the next node(s)?

kdroste-google: OK, so the key part is workflow.NodeConfig{RerunOnResume: &rerun}, when set to true will re-exec the function after the answer is delivered, and for false will complete the node execution and move on.

workflow/function_node.go
```diff
@@ -241,9 +286,48 @@ func newFunctionNodeWithResolvedSchemas[IN, OUT any](name string, fn func(ctx ag
...
 }
 
+// newEmittingFunctionNodeWithResolvedSchemas is the internal
+// constructor for the streaming variant.
+func newEmittingFunctionNodeWithResolvedSchemas[IN, OUT any](name string, fn EmittingFunctionFn[IN, OUT], inputSchema, outputSchema *jsonschema.Resolved, cfg NodeConfig) *FunctionNode {
+	wrappedFn := func(ctx agent.Context, input any, emit func(*session.Event) error) (any, error) {
+		var typedInput IN
+		if input != nil {
+			t, ok := input.(IN)
+			if !ok {
```
kdroste-google, line 302: Please comment this part, it's not clear why we do ConvertToWithJSONSchema when input.(IN) fails.

workflow/function_node.go
```diff
@@ -241,9 +286,48 @@ func newFunctionNodeWithResolvedSchemas[IN, OUT any](name string, fn func(ctx ag
...
+				converted, err := typeutil.ConvertToWithJSONSchema[any, IN](input, inputSchema)
```
kdroste-google, line 307: It's not clear to me, why we do not validate JSON schema upon successful input.(IN)

workflow/function_node.go
kdroste-google, line 358: What if ev.InvocationID != "" && ev.InvocationID != ctx.InvocationID(). Is it OK not to yield an event with ctx.InvocationID()?

kdroste-google (changes requested):

examples/workflow/hitl_rerun/main.go
wolo-lab: Good idea. I've added a general helper for that.

workflow/function_node.go
wolo-lab: Added.

workflow/function_node.go
wolo-lab: Good point, it was a gap. Fixed here and in the not-emitting version.

workflow/function_node.go
wolo-lab: It's a weird situation which should normally not happen. Anyway, I've checked that adk-python always set invocationId (without any conditions). I've created a dedicated bug for that and will work on it in a separate PR as I don't want to keep PR focused on specific problems/featues.

kdroste-google (approved):
LGTM

kdroste-google (commented):
LGTM, thank you!

PR #1002, opened 2026-06-11
fix(webui): server-side fixes for HITL in the web UI

wolo-lab:
Two independent server-side bugs broke human-in-the-loop (HITL) workflows
under the web UI. Both are needed for a HITL resume to reach the browser;
the web UI bundle update is handled separately.

## Problem
1. **`functionCallEventId` rejected (400).** The web UI sends
   `functionCallEventId` on every
   HITL response — fields that exist on adk-python's `/run` / `/run_sse`
   request body. adk-go's `RunAgentRequest` did not declare them, and
   `decodeRequestBody` uses `DisallowUnknownFields`, so the request failed
   with `json: unknown field "functionCallEventId"`. The web UI buffers the
   SSE response and renders on completion, so the 4xx surfaced as a silent
   no-op.
2. **Missing `Content.Role`.** Workflow nodes that build `Event.Content`
   directly (`FunctionNode` and `BaseNode`-derived nodes) set `Parts` but
   not `Role`. adk's convention is that model-authored Content carries
   `Role="model"`, and the web UI relies on it to render an event. A resume
   reply was persisted and returned over SSE with an empty role, so the UI
   ignored it.

## Solution
- **`fix(adkrest)`** — declare `functionCallEventId` on `RunAgentRequest` as optional fields (adk-python
  parity, accepted-and-ignored for now). Strict decoding stays on for typo
  safety. Adds positive + regression tests.
- **`fix(workflow)`** — stamp `Content.Role="model"` centrally in
  `scheduler.handleEvent`, the one chokepoint every node event passes
  through (it already stamps `Branch` and `NodeInfo.Path`). Runs before the
  dynamic-descendant short-circuit and leaves an explicitly-set role
  untouched. Mirrors the convention used by `NewRequestInputEvent` and the
  LLM flow. Adds tests.
Verified end to end with `examples/workflow/hitl_simple`


server/adkrest/internal/models/runtime.go
```diff
@@ -32,6 +32,8 @@ type RunAgentRequest struct {
 	Streaming bool `json:"streaming,omitempty"`
 
 	StateDelta *map[string]any `json:"stateDelta,omitempty"`
+
+	FunctionCallEventID *string `json:"functionCallEventId,omitempty"`
```
kdroste-google, line 40: Please add a comment about the purpose of this field, it's not obvious

workflow/scheduler.go
```diff
@@ -627,6 +628,13 @@ func (s *scheduler) handleEvent(it eventItem) {
 	if it.ev.Branch == "" && nr.branch != "" {
 		it.ev.Branch = nr.branch
 	}
+	// Default Content.Role to "model" for nodes that left it empty
+	// (FunctionNode/BaseNode set Parts but not Role); clients like
+	// the web UI rely on it. Before the descendant short-circuit so
+	// dynamic children are covered too.
+	if it.ev.Content != nil && it.ev.Content.Role == "" {
+		it.ev.Content.Role = genai.RoleModel
```
kdroste-google: To be checked - what's the right role for function / tool calls? I'm not sure if the default is appropiate for all the cases

kdroste-google (changes requested):

server/adkrest/internal/models/runtime.go
wolo-lab: True. Especially that this field is not used at all right now. Web ui sends it anyway, so we need to be able to handle it.

workflow/scheduler.go
wolo-lab: Good catch. The role depends on the part type, following the adk-python convention: a FunctionResponse is app-authored so it takes user, while everything else stays model.

kdroste-google (approved):
LGTM

PR #1004, opened 2026-06-11
feat: llm agent wrapper

dpasiukevich:
RunLLMAgentAsNode func is in the llmagent package, because both runner and llmagent need to call it:
- runner in the runner/agent_node.go
- llmagent when agentnode.Run is called (it wraps agent.Agent)


wolo-lab (approved):
Nice!

PR #1008, opened 2026-06-12
Context unification NodeContext with InvocationContext added to commonContext

kdroste-google:
Context unification


wolo-lab (approved):
Thanks. For the records: the proper cleanup will be done in the following PRs.

PR #1012, opened 2026-06-12
Fix silent bug skipping custom tools wrapped in ToolCall envelopes

### LINKED ISSUE #1011: utils.FunctionCalls silently drops custom functions when returned as part.ToolCall.
When using newer models, the backend API occasionally returns custom function invocations wrapped inside a unified ToolCall object (with ToolType: "function_call") rather than the legacy top-level FunctionCall object.

The underlying google-genai-sdk correctly deserializes this into part.ToolCall. However, ADK's utils.FunctionCalls(c *genai.Content) strictly checks for p.FunctionCall != nil and ignores p.ToolCall.

This causes the ADK Runner to miss the model's requested function call.

What i see in my agent runs are multiple thought blocks with no tool calls or text output and eventually the agent just stops.

jpgincgoogle:
Updated utils.go and session.go to recognize custom tools when returned
inside part.ToolCall (with type "function_call") rather than just part.FunctionCall.
This prevents the agent runner from discarding the tool calls.

- Closes: #1011

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -83,6 +83,15 @@ func FunctionCalls(c *genai.Content) (ret []*genai.FunctionCall) {
 		if p.FunctionCall != nil {
 			ret = append(ret, p.FunctionCall)
 		}
+		if p.ToolCall != nil && string(p.ToolCall.ToolType) == "function_call" {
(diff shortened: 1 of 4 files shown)


jpgincgoogle:
nvm, fake news.

PR #1016, opened 2026-06-14
refactor(agent): migrate all consumers to Context

wolo-lab:
The context unification introduced a single unified Context, with agent.CallbackContext, agent.ToolContext and tool.Context kept as aliases to ease migration.

Migrate every consumer (~90 files) to use agent.Context directly. The three aliases remain defined (now marked Deprecated) so this change is safe to import into downstream repositories that still reference them; a follow-up removes the aliases for a clean open-source surface.


kdroste-google (approved):
LGTM

PR #1017, opened 2026-06-14
fix(session): persist Event.NodeInfo through Vertex AI backend

wolo-lab:
### Problem
The Vertex AI session backend only round-tripped event fields that have a dedicated SessionEvent column (Content, Actions, EventMetadata). On write everything else was silently dropped, and on read it was never reconstructed: Output, NodeInfo, IsolationScope, RequestedInput and Routes.
This became correctness-critical once the workflow engine moved to event-scan resume reconstruction (ReconstructRunState / rehydrateCache), where NodeInfo.Path identifies the emitting node. A workflow paused and resumed against a Vertex-persisted session resumed incorrectly because NodeInfo was gone. The in-memory and SQL backends already preserved these fields; Vertex was the last gap.

### Solution
Persist the full event in SessionEvent.raw_event (proto field 12, the same field adk-python uses) and prefer it on read, falling back to the legacy field-based reconstruction for events written before raw_event support.
- eventNeedsRawEvent — raw_event is attached only when the event carries one of the column-less fields (Output, NodeInfo, IsolationScope, RequestedInput, Routes). Plain events keep their existing wire format, so the recorded rpcreplay fixtures stay valid (none of them sets those fields).
- eventToRawEvent / eventFromRawEvent — JSON round-trip into structpb.Struct and back. On read, identity fields (ID, Timestamp, InvocationID, Author) are taken from the envelope.

### DIFF
--- a/session/vertexai/vertexai_client.go
+++ b/session/vertexai/vertexai_client.go
@@ -332,12 +332,8 @@ func eventToRawEvent(event *session.Event) (*structpb.Struct, error) {
 	if err != nil {
 		return nil, fmt.Errorf("marshaling event: %w", err)
 	}
-	var m map[string]any
-	if err := json.Unmarshal(b, &m); err != nil {
-		return nil, fmt.Errorf("unmarshaling event to map: %w", err)
-	}
-	s, err := structpb.NewStruct(m)
-	if err != nil {
+	s := &structpb.Struct{}
+	if err := s.UnmarshalJSON(b); err != nil {
 		return nil, fmt.Errorf("converting event to structpb: %w", err)
 	}
 	return s, nil
--- a/session/vertexai/vertexai_client.go
+++ b/session/vertexai/vertexai_client.go
@@ -16,6 +16,7 @@ package vertexai
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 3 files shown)


session/vertexai/vertexai_client.go
```diff
@@ -294,6 +306,59 @@ func (c *vertexAiClient) appendEvent(ctx context.Context, appName, sessionID str
...
+// (structpb numbers and json.Unmarshal into any are both float64). This
+// matches the SQL backend, so the lossiness is framework-wide; store
+// values needing exact integer fidelity as strings.
+func eventToRawEvent(event *session.Event) (*structpb.Struct, error) {
+	b, err := json.Marshal(event)
+	if err != nil {
+		return nil, fmt.Errorf("marshaling event: %w", err)
+	}
+	var m map[string]any
+	if err := json.Unmarshal(b, &m); err != nil {
```
kdroste-google: You can use 
```go
s := structpb.Struct{}
err = s.UnmarshalJSON(b)
```
one step less

kdroste-google (approved):
LGTM

session/vertexai/vertexai_client.go
wolo-lab: Great idea. Applied.

PR #1018, opened 2026-06-14
feat(telemetry): read capture-message-content from env var

wolo-lab:
## Problem
A follow-up telemetry change removes `SetGenAICaptureMessageContent` in favour of
the OpenTelemetry-spec env var `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`
(see #894). Today `getGenAICaptureMessageContent` reads **only** the setter, so
there is no overlap window: removing the setter outright is a breaking change for
any code that still calls it, with no way to migrate ahead of time.

## Solution
Make `getGenAICaptureMessageContent` honour **both** sources with OR semantics:
content capture is enabled if either the env var is `"true"` **or** the setter
enabled it. The setter is kept and marked `Deprecated`.
This is the **expand** step of an expand → migrate → contract rollout:
1. **expand (this PR):** env var and the deprecated setter both work — nothing
   breaks, and the env var becomes usable.
2. **migrate:** callers move from the setter to the env var at their own pace.
3. **contract:** the setter removal (#894) lands once nothing calls it anymore.
Opening the overlap window before the removal turns a breaking change into a
safe, staged migration.


dpasiukevich (approved):

wolo-lab:
Closing, already solved by 1#027

PR #1031, opened 2026-06-15
fix(console): use readline for CJK-friendly console input

### LINKED ISSUE #978: Console launcher: CJK characters cannot be deleted during input editing
#### Description
When using the console launcher (cmd/launcher/console), editing Chinese/CJK characters during input doesn't work correctly. Specifically, after typing Chinese characters, pressing Backspace fails to delete them. ASCII/English characters work fine.

#### Steps to Reproduce
Build and run any agent with the console launcher
Type Chinese characters (e.g. via an IME) at the User -> prompt
Press Backspace to try to delete the Chinese characters
Observe that the characters cannot be deleted, or only partially deleted

#### Root Cause
The console launcher reads input via bufio.NewReader(os.Stdin).ReadString('\n') (console.go:120-127):

```
go func() {
    reader := bufio.NewReader(os.Stdin)
    for {
        userInput, err := reader.ReadString('\n')
        if err != nil {
            readErrChan <- err
            return
        }
        inputChan <- userInput
    }
}()
```
bufio.Reader has an internal 4096-byte read buffer that eagerly consumes bytes from stdin. CJK characters are UTF-8 multi-byte (3 bytes per character) and occupy 2 terminal columns (wide characters). When an IME sends intermediate composition bytes to stdin, the bufio.Reader may consume them into its buffer, causing the terminal's input editing state to desync from what the program has read. This makes Backspace unable to correctly delete the characters.

#### Suggested Fix
Replace the raw bufio.Reader with a proper readline library (e.g. github.com/ergochat/readline) that correctly handles:

UTF-8 multi-byte character width and cursor movement
IME composition state
Line editing operations (backspace, delete, cursor movement)

#### Environment
adk-go v1.4.0
macOS Darwin 25.5.0
Terminal: VS Code integrated terminal / iTerm2
Input method: macOS Chinese IME

KirtiRamchandani:
## Summary
- Replace `bufio.Reader.ReadString('\n')` with `github.com/ergochat/readline` in the console launcher
- Fixes CJK/IME backspace and line editing when typing multibyte characters at the `User ->` prompt
- Removes the extra `fmt.Print("\nUser -> ")` after agent responses; readline owns the prompt

Fixes #978

## Test plan
- [x] `go build ./cmd/launcher/console/...`
- [x] `go test ./...`
- [ ] Manual: type Chinese via macOS IME in console launcher and verify Backspace deletes characters correctly

/cc @indurireddy-TF @bitsark


indurireddy-TF:
Hi @KirtiRamchandani Can you please resolve merge conflicts?

karolpiotrowicz (changes requested):
The diagnosis in #978 is right that `bufio.ReadString` is byte-oriented and that a line editor is the natural answer. Two things in the current shape need fixing before this can go in, and there's a question about the fix's scope that's worth settling first because it may change what you build.

## Ctrl-C no longer stops a running agent

This is the blocking one. `Readline()` puts the terminal in raw mode, which clears `ISIG`, so the kernel stops turning Ctrl-C into `SIGINT`. Because the reader goroutine re-enters `Readline()` the instant it hands a line to the unbuffered `inputChan` ([console.go:137](https://github.com/google/adk-go/blob/799fdec605db7f23b838683a19fe3d5962308f41/cmd/launcher/console/console.go#L137) then straight back to [console.go:132](https://github.com/google/adk-go/blob/799fdec605db7f23b838683a19fe3d5962308f41/cmd/launcher/console/console.go#L132)), the terminal stays in raw mode for the entire duration of the agent's turn. So `signal.NotifyContext` never fires, the context is never cancelled, and `r.Run(ctx, ...)` runs to completion — the agent finishes every remaining tool call after the operator has asked it to stop.

Driving a turn loop that honours the context exactly the way `r.Run` does, with Ctrl-C pressed mid-turn:

```
main    -> [TURN ABORTED by ctx — tool calls stopped]        exit 0
this PR -> [TURN RAN TO COMPLETION — all tool calls executed] exit 1
```

For a framework where a turn executes tools with the user's credentials, that in-band stop is worth protecting.

Ctrl-C at an idle prompt has a smaller version of the same problem: `Readline()` returns `readline.ErrInterrupt`, which is not `io.EOF`, so it falls through to `log.Fatal` at [console.go:172](https://github.com/google/adk-go/blob/799fdec605db7f23b838683a19fe3d5962308f41/cmd/launcher/console/console.go#L172). `log.Fatal` calls `os.Exit`, which skips deferred functions — including the OpenTelemetry shutdown, so buffered spans and metrics are dropped. Exit status goes from 0 to 1.

## The prompt now races the agent's output

Moving the prompt into `readline.Config` means it is printed by the reader goroutine on entry to `Readline()`, while the main goroutine is printing the answer with `fmt.Print`. The two main-goroutine prompt prints this PR removes were what kept those ordered. Five runs per arm, this PR garbled the output 5/5 and main 0/5. What the user sees depends on how fast the turn is — with a quick turn the prompt lands inside the answer (`Agent -> User -> Hello, how can I help?`), and with a realistic multi-second turn you get a spurious empty `User -> ` while the agent is still working. The race itself is unconditional either way, because both writes happen in the microseconds after the channel handoff, before the model call.

I tried the obvious remedy — routing the agent output through `rl.Stdout()` rather than `os.Stdout` — and it does not fix it: still 5/5 garbled. So I'd rather not suggest a specific fix here, beyond noting that it likely has to stop the reader from holding the line while the main loop is mid-turn, which would also address the Ctrl-C problem above.

The same extra prompt shows up on the human-in-the-loop path, where [hitl.go:102](https://github.com/google/adk-go/blob/799fdec605db7f23b838683a19fe3d5962308f41/cmd/launcher/console/hitl.go#L102) prints its own `User -> `. Counting across one whole HITL turn: main renders one prompt, this PR renders two. That also makes the committed transcripts in `examples/workflow/hitl_simple/README.md` inaccurate.

## The terminal can be left unusable

`defer rl.Close()` at [console.go:129](https://github.com/google/adk-go/blob/799fdec605db7f23b838683a19fe3d5962308f41/cmd/launcher/console/console.go#L129) is on the reader goroutine's frame, so it only runs if that goroutine returns. When the main loop exits via [console.go:165](https://github.com/google/adk-go/blob/799fdec605db7f23b838683a19fe3d5962308f41/cmd/launcher/console/console.go#L165) the goroutine is still parked inside `Readline()` holding raw mode. Reading the pty's termios after the process exits:

| how it exited | main | this PR |
| --- | --- | --- |
| Ctrl-C | restored | restored |
| external `kill -INT` | restored | `ICANON=off ECHO=off ISIG=off` |
| SIGTERM (`docker stop`, systemd, CI timeout) | restored | `ICANON=off ECHO=off ISIG=off` |

The user's shell is left with no echo and no Ctrl-C, and they have to type `stty sane` blind to recover. Worth knowing that the SIGTERM row is not fixed by moving that `defer`: `signal.NotifyContext` registers `os.Interrupt` only, so on SIGTERM no deferred function runs anywhere. The library ships `CaptureExitSignal()`, which closes the instance on both Interrupt and SIGTERM — I haven't tried it here, so treat that as a pointer rather than a recommendation.

## Worth checking before you invest more in this

The improvement turned out to depend on the terminal's `IUTF8` flag, which I did not expect:

| | main | this PR |
| --- | --- | --- |
| `IUTF8` off | `你好世` + a corrupted trailing byte (11 bytes, 5 runes) | `你好世` (9 bytes, 3 runes) |
| `IUTF8` on | `你好世` (9 bytes, 3 runes) — already correct | identical |

With `IUTF8` set, the kernel line discipline already erases a whole multi-byte character in canonical mode and this change makes no difference. My first measurement looked like a dramatic fix only because a bare pty defaults to `-iutf8`. On top of that, #978 describes an IME on macOS, and I only tested backspacing over characters that were already committed — I never drove an IME composition sequence, so I can't say whether this closes the reported problem.

None of that makes the change worthless, and the bug is clearly real for the reporter. It does mean the benefit is narrower than it first appears, which matters when weighed against the regressions above. It would be worth asking them for `stty -a | grep -o iutf8` from the terminal where it fails — if `IUTF8` is set there, then something other than plain backspace handling is going on and this may not be the right fix.

## Smaller things

CI lint is red on this PR: `errcheck` flags the unchecked error from `rl.Close()` at [console.go:129](https://github.com/google/adk-go/blob/799fdec605db7f23b838683a19fe3d5962308f41/cmd/launcher/console/console.go#L129).

There's no test with the change. That one is genuinely awkward here — the package's existing harness swaps `os.Stdin` for a pipe, which never reaches line editing at all, so a test reproducing #978 needs a pty. A test for the Ctrl-C error classification needs no pty though, and would have caught the exit-status change.

One deliberate-looking behaviour change to confirm: the prompt is no longer printed when stdout is piped, because the library suppresses it off-TTY. I grepped all 32 occurrences of `User -> ` in the repo and no test asserts on it, so nothing breaks, but it is a silent change in output shape for anyone parsing piped output.

## On the dependency

`github.com/ergochat/readline` is a light addition — one module, and its own requirements are already in the graph at higher versions, so nothing transitive comes with it. It cross-compiles cleanly for Windows, macOS and linux/arm64, and adds no vulnerabilities. Still, adk-python reads console input with the plain builtin `input()` and no line-editing library, so this would make Go the only ADK implementation carrying one. That's a maintainer call about direction rather than anything wrong with your patch, and it's worth getting an explicit yes on before you spend more time on the code.

karolpiotrowicz:
@wolo-lab flagging this one for a dependency call rather than a code review.

It swaps console input to a readline library to fix CJK character handling, which would make ADK Go the only implementation carrying a line-editing dependency — adk-python uses the builtin. The footprint is small and self-contained, one module with nothing transitive. Worth deciding whether we want the dependency before the author reworks the terminal handling.

PR #1040, opened 2026-06-15
fix(workflow): serialize dynamic-node emit across concurrent children

wolo-lab:
## Problem
A dynamic node's body (`DynamicFn`) can launch several children at once and run them on separate goroutines.
Every child sends its events "up" to the parent through a single shared callback (`emit`, which wraps the parent's one `yield` function). That callback had **no synchronization**, so when two children fire at the same time they call the same `yield` concurrently — and two things break:
1. **Panic.** Go's range-over-func (the streaming mechanism behind these events) does not allow the same iterator to be entered from two goroutines at once. It detects the overlap and panics (`range function continued iteration after loop body panic`).
2. **Data race.** The parent's event pump (`runNode`) records the node's outcome on that same path; two goroutines writing it concurrently is a race (caught by `-race`).

**Trigger:** any `DynamicFn` that fans children out across goroutines (e.g. `errgroup` / `WaitGroup`), each emitting at least one event.

## Solution
Serialize emission with a single per-activation mutex inside `makeEmit`. Every `yield` — whether from the `DynamicFn`'s own `emit` or from a child via `RunNode` or the sub-scheduler — now goes through the same lock, so only one runs at a time. The lock is held only around the single `yield` call, so children still run concurrently; only the hand-off upstream is serialized.

### DIFF
--- a/workflow/dynamic_node.go
+++ b/workflow/dynamic_node.go
@@ -105,11 +105,7 @@ func (n *dynamicNode[IN, OUT]) Run(ctx agent.Context, input any) iter.Seq2[*sess
 			return
 		}
 
-		// One mutex serializes every yield for this activation: the emit
-		// passed to the DynamicFn and the same emit driven by RunNode via
-		// the sub-scheduler. Concurrent children must not yield at once.
-		var emitMu sync.Mutex
-		emit := makeEmit(yield, ctx, &emitMu)
+		emit := makeEmit(yield, ctx)
 		sub := newDynamicSubScheduler(ctx, n.composePath(ctx), emit)
 		orchestratorCtx := newDynamicNodeContext(ctx, sub.ParentPath(), "", sub, sub.OutputForAncestors())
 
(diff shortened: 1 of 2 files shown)


workflow/dynamic_node.go
```diff
@@ -104,7 +105,11 @@ func (n *dynamicNode[IN, OUT]) Run(ctx agent.Context, input any) iter.Seq2[*sess
 			return
 		}
 
-		emit := makeEmit(yield, ctx)
+		// One mutex serializes every yield for this activation: the emit
+		// passed to the DynamicFn and the same emit driven by RunNode via
+		// the sub-scheduler. Concurrent children must not yield at once.
+		var emitMu sync.Mutex
+		emit := makeEmit(yield, ctx, &emitMu)
```
dpasiukevich: nit: would it make sense to create mutex within `makeEmit` function? this will make a simpler func prototype.

dpasiukevich (approved):

workflow/dynamic_node.go
wolo-lab: That's an excelant idea!

PR #1044, opened 2026-06-15
feat(session): add EventCompaction type and compaction filtering in contents_processor

### LINKED ISSUE #1001: Add EventCompaction support to EventActions and compaction filtering in contents_processor.go
Yes. I'm working on adding compaction support to [kagent](https://github.com/kagent-dev/kagent)'s Go ADK runtime to match the existing Python ADK behavior. The Go ADK has no compaction primitives at all 
I checked every released version from v0.1.0 to v1.4.0. 
EventActions has no Compaction field and contents_processor.go has no filtering step for compaction markers.
Without these primitives , after a compaction runs the raw events stay in the backing store and get re-sent to the LLM on every subsequent request, which defeats the purpose of compaction entirely.

### Proposed Solution
Three things need to be added to mirror what Python ADK already does in events/event_actions.py, apps/compaction.py and flows/llm_flows/contents.py -

- An EventCompaction struct in session/session.go carrying the compacted time range and compacted content

- A Compaction field on EventActions to hold the marker

- Compaction filtering in internal/llminternal/contents_processor.go  when building LLM contents raw events whose timestamps fall within a compaction marker's range should be replaced by the marker's compacted content

### Impact on your work
This is blocking a PR I have open at kagent ([#1971](https://github.com/kagent-dev/kagent/pull/1971)) to add Go-runtime compaction. The kagent maintainer specifically pointed out that without these primitives the implementation cannot match Python ADK semantics. Compaction is important for long-running agents so the sooner the better though no hard deadline.

### Willingness to contribute
Yes happy to submit a PR
Just wanted to check with maintainers first before touching core session and flow internals.

### Additional Context
Related downstream PR: [kagent #1971](https://github.com/kagent-dev/kagent/pull/1971)  
this is what is blocked on this feature.

QuentinBisson:
Ports the compaction persistence and content-assembly model from adk-python to adk-go. Fixes #1001.

## What's missing today

`EventActions` has no `Compaction` field and `buildContentsDefault` has no filtering step for compaction markers. This means raw events are never replaced by their summaries when building the LLM context — compaction runs but has no visible effect on what gets sent to the model.

## Changes

**`session/session.go`**

```go
type EventCompaction struct {
    StartTimestamp   time.Time
    EndTimestamp     time.Time
    CompactedContent *genai.Content
}
```

Added `Compaction *EventCompaction` to `EventActions`. The summary lives here, not in `Event.Content`, so compaction markers are excluded from normal content rendering and distinguishable from real model output.

**`internal/llminternal/contents_processor.go`**

New `applyCompaction` pass called inside `buildContentsDefault`, after transcription aggregation and before the async-function-response rearrangement step:

1. Collect all compaction marker events (those with `Actions.Compaction != nil`).
2. Remove subsumed markers — if marker C2's range fully contains marker C's range, C is dropped (a later compaction already covers it).
3. Sort active markers by `StartTimestamp`.
4. Walk the event list: events whose timestamp falls within an active marker's range are replaced by a single synthetic event carrying `CompactedContent`; compaction marker events themselves are always removed from the output.

This mirrors `_apply_compaction` in adk-python's `flows/llm_flows/contents.py`.

## Downstream reference

kagent PR https://github.com/kagent-dev/kagent/pull/2025 uses this via a `replace` directive and exercises both sliding-window and token-threshold compaction modes against Anthropic and OpenAI adapters. The downstream compaction driver (watermark-based invocation counting, `longestSelfContainedPrefix` safety, summarizer LLM) lives in kagent — out of scope for this PR, which only adds the session primitives and content-assembly filtering that the driver depends on.


QuentinBisson:
Superseeded by #1231

PR #1047, opened 2026-06-16
refactor(workflow): move output validation off inline node paths

wolo-lab:
## Problem
Output validation for `FunctionNode` and `ToolNode` was duplicated. Each node
validated its output inline (inside `wrappedFn` / `runTool`) **and** the
scheduler validated it again scheduler-side via `BaseNode.ValidateOutput`
(introduced in #970). The same `OutputSchema` was therefore checked twice on
every activation.
Beyond the redundancy, the two copies are a drift hazard: as the
scheduler-side helper evolves (e.g. #970 added the `*genai.Content` text
fallback and passthrough for framework control values), the inline copies fall
behind and the two paths diverge. The validation behaviour also stopped being
uniform across node types — custom `Node` implementations only ever went through
the scheduler, while `FunctionNode`/`ToolNode` carried their own logic.
The FunctionTool `{"result": X}` unwrap convention was buried inline inside
`ToolNode.runTool`, mixing a tool-specific quirk into the execution path instead
of expressing it as part of the validation contract.

## Solution
Make the scheduler the single, authoritative output-validation point and remove
the inline validation from the node run paths:
- **`FunctionNode`**: drop the inline `OutputSchema` check from all three
  constructors (`newFunctionNodeWithResolvedSchemas`, `NewFunctionNodeFromState`,
  and the emitting variant). The schema still flows through `BaseNode`, so the
  scheduler enforces it via `BaseNode.ValidateOutput`. `wrappedFn` is back to
  doing one thing: materialise the typed parameter and call the user function.
- **`ToolNode`**: remove the inline validation from `runTool` and move the
  FunctionTool `{"result": X}` unwrap into a dedicated `ToolNode.ValidateOutput`
  override. The override tries standard validation first, falls back to
  unwrapping `{"result": X}`, and re-runs standard validation on total failure
  so the caller still sees the original schema-mismatch error. This is the right
  home for the convention because it is tool-specific — making it a general
  default could mask genuine validation errors in other node types.

## Behaviour note
Validation errors now surface scheduler-side rather than inside `Run`. For real
workflows this is unobservable (every activation goes through the scheduler).
Callers that invoke `node.Run()` directly (mostly tests) now receive the raw
output and must call `ValidateOutput` themselves. Tests are split accordingly:
`Run`-level tests assert raw passthrough; validation is asserted via
`ValidateOutput` (`TestFunctionNode_ValidateOutput`,
`TestToolNode_ValidateOutput` covering direct-valid, `{"result": X}` unwrap,
fail-both-paths, and nil-schema cases).

### DIFF
--- a/workflow/function_node_test.go
+++ b/workflow/function_node_test.go
@@ -158,6 +158,17 @@ func TestFunctionNode_ValidateOutput(t *testing.T) {
 	if _, err := node.ValidateOutput(got); err == nil {
 		t.Fatalf("ValidateOutput: expected validation error, got nil")
 	}
+
+	// A schema-conforming output passes ValidateOutput and is returned
+	// unchanged.
+	valid := map[string]any{"result": 1}
+	gotValid, err := node.ValidateOutput(valid)
+	if err != nil {
+		t.Fatalf("ValidateOutput: unexpected error for valid output: %v", err)
+	}
+	if diff := cmp.Diff(valid, gotValid); diff != "" {
+		t.Errorf("ValidateOutput mismatch (-want +got):\n%s", diff)
+	}
 }
 
 func mustSchema[T any](t *testing.T) *jsonschema.Schema {
--- a/workflow/function_node_test.go
+++ b/workflow/function_node_test.go
@@ -37,9 +37,6 @@ func TestNewFunctionNodeWithSchema(t *testing.T) {
 	type Output struct {
 		Result string `json:"result"`
 	}
-	type TargetOutput struct {
-		Result int `json:"result"`
-	}
 
 	tests := []struct {
 		name         string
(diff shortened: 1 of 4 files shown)


workflow/function_node_test.go
```diff
@@ -132,6 +117,49 @@ func TestNewFunctionNodeWithSchema(t *testing.T) {
 	}
 }
 
+// TestFunctionNode_ValidateOutput verifies that output schema validation
+// is enforced through the node-level ValidateOutput contract (invoked
+// scheduler-side), not inline inside Run. Run itself passes the output
+// through unchanged; ValidateOutput surfaces the schema mismatch.
+func TestFunctionNode_ValidateOutput(t *testing.T) {
```
hanorik: Maybe to add a success test case as well?

hanorik (approved):

workflow/function_node_test.go
wolo-lab: Good idea. Adding.

PR #1050, opened 2026-06-16
feat: add platform TaskRunner seam for caller-controlled tool fan-out

### LINKED ISSUE #1051: Add a platform concurrency seam (TaskRunner) for caller-controlled tool fan-out

ADK's LLM flow hard-codes its tool-call fan-out to one goroutine per call (`handleFunctionCalls` in `internal/llminternal`). Applications embedding ADK may need to control how that concurrent work runs - for example, to cap parallelism against rate-limited model/tool backends, reuse an existing worker pool or executor, integrate with their own scheduler, or run tool calls sequentially for predictable ordering and easier debugging. There is currently no hook to influence this.

### Proposed Solution

Add a concurrency seam alongside the existing `platform` provider seams (`WithTimeProvider` / `WithUUIDProvider`):

- `platform.RunTasks(ctx, tasks []func(context.Context))`: runs a batch of independent tasks and blocks until all complete. With no runner installed it keeps today's behavior (one goroutine per task).
- `platform.WithTaskRunner(ctx, runner)`: installs a caller-provided runner that decides how the tasks execute (bounded concurrency, a custom executor, sequential execution, etc.).
- Each task is invoked with its own `context.Context`, so a runner can derive a distinct per-task context (per-task cancellation, deadlines, or values). The default runner passes the parent `ctx` to each task.
- `internal/llminternal` tool fan-out (`handleFunctionCalls`) executes its tasks via `platform.RunTasks`, so the seam actually governs tool concurrency.

### Impact on your work

We are building an ADK integration where we need to be able to provide our own executor, so this is a critical requirement for us.

---

### Alternatives Considered

- Wrapping each tool to self-limit - doesn't help true fan-out and pushes concurrency control into every tool.
- A package-global setting - not concurrency-safe and doesn't compose across independent runners.
- Patching ADK via a fork / `replace` directive - fragile and a maintenance burden.

The context-carried seam matches the existing provider-seam pattern and is per-call scoped.

### Willingness to contribute

Yes - already implemented in #1050

```go
// package platform
type TaskRunner func(ctx context.Context, tasks []func(context.Context))

func WithTaskRunner(ctx context.Context, runner TaskRunner) context.Context
func RunTasks(ctx context.Context, tasks []func(context.Context)) // default: one goroutine per task
```

### Additional Context

Implemented by #1050.

DABH:
### Link to Issue

Closes #1051

### Summary

Adds a `platform.TaskRunner` seam, mimicking the (now-)existing `platform` provider seams (`WithTimeProvider` / `WithUUIDProvider`) from #964.  This lets a caller control how ADK runs its internal tool-call fan-out, and routes that fan-out through it.

- `platform.RunTasks(ctx, tasks)` runs a batch of independent tasks and blocks until all complete. With no runner installed it keeps today's behavior: one goroutine per task.  So, _no breaking changes here_
- `platform.WithTaskRunner(ctx, runner)` lets a caller install its own runner (for example, to cap the number of concurrent tool calls, e.g., against rate-limited backends), dispatch them onto an existing executor / worker pool, or run them sequentially for predictable ordering.
- Each task is invoked with its own `context.Context`, so a runner can derive a distinct per-task context (per-task cancellation, deadlines, or values). The default runner passes the parent `ctx` to every task.
- `internal/llminternal` tool fan-out (`handleFunctionCalls`) now executes its tasks via `platform.RunTasks`, so the seam governs tool concurrency.

Today ADK hard-codes one-goroutine-per-tool-call fan-out, with no way for an embedding application to influence it. This adds that control point - mirroring the existing `platform` provider seams - while leaving the default behavior unchanged.

### Testing Plan

**Unit Tests:**

`platform/exec_test.go` covers the default runner (runs all tasks, is a no-op on empty/nil input, and passes a context to each task) and an installed runner (used when present, falls back to the default when nil, and receives a distinct per-task context).

```
$ go test ./platform/... ./internal/llminternal/...
ok  google.golang.org/adk/platform
ok  google.golang.org/adk/internal/llminternal
```

N/A; this is a low-level concurrency seam with no ADK Web / Runner UI surface. The default path (one goroutine per task when no runner is installed) is unchanged, and the installed-runner path is exercised by the unit tests above.

### Checklist

- [ ] I have manually tested my changes end-to-end. _(N/A — low-level seam; see above.)_

### DIFF
--- a/platform/exec.go
+++ b/platform/exec.go
@@ -1,0 +1,95 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package platform
+
+import (
+	"context"
+	"sync"
+)
+
+// TaskRunner runs a batch of independent tasks, invoking each one exactly
+// once and passing it a context.Context. An implementation must block until
+// every task has completed.
+//
+// Each task receives a context so the runner can give it its own per-task
+// context rather than sharing one. The default runner passes the same ctx it
+// was given to every task; a custom runner may instead derive a distinct
+// context per task.
+//
+// The default runner (used when none is installed on the context) runs the
+// tasks concurrently, one goroutine per task. A caller that wants to control
+// how the fan-out executes — for example to bound concurrency, dispatch onto
+// its own executor, or run the tasks sequentially in a single goroutine — can
+// install its own runner with WithTaskRunner.
+type TaskRunner func(ctx context.Context, tasks []func(context.Context))
+
+// taskRunnerKey is the context key under which a TaskRunner is stored.
+type taskRunnerKey struct{}
+
+// WithTaskRunner returns a copy of ctx that carries runner. RunTasks called
+// with the returned context, or any context derived from it, uses runner
+// instead of the default goroutine-based runner. A nil runner is ignored by
+// RunTasks, which then falls back to the default.
+//
+// WithTaskRunner is the concurrency analog of WithTimeProvider and
+// WithUUIDProvider: it lets a host environment substitute its own execution
+// strategy for ADK's internal fan-out without the ADK runtime taking a
+// dependency on that environment.
+func WithTaskRunner(ctx context.Context, runner TaskRunner) context.Context {
+	return context.WithValue(ctx, taskRunnerKey{}, runner)
+}
+
+// RunTasks runs every task in tasks and blocks until all of them complete. If
+// ctx carries a TaskRunner installed with WithTaskRunner, that runner is used;
+// otherwise RunTasks runs the tasks concurrently on one goroutine each.
+//
+// Each task is invoked with a context.Context. Under the default runner every
+// task receives ctx; an installed runner may instead pass each task its own
+// per-task context. Under the default runner, the tasks run concurrently and
+// must be safe to do so. An empty or nil slice returns without doing anything.
+//
+// RunTasks enforces the completion barrier itself rather than relying on the
+// runner: it wraps each task to signal a sync.WaitGroup on return and waits on
+// that group after the runner returns. Callers may therefore read per-task
+// results as soon as RunTasks returns, even if a custom runner does not block
+// until its tasks finish.
+func RunTasks(ctx context.Context, tasks []func(context.Context)) {
+	if len(tasks) == 0 {
+		return
+	}
+	var wg sync.WaitGroup
+	wg.Add(len(tasks))
+	if ctx != nil {
+		if r, ok := ctx.Value(taskRunnerKey{}).(TaskRunner); ok && r != nil {
+			wrapped := make([]func(context.Context), len(tasks))
+			for i, task := range tasks {
+				wrapped[i] = func(c context.Context) {
+					defer wg.Done()
+					task(c)
+				}
+			}
+			r(ctx, wrapped)
+			wg.Wait()
+			return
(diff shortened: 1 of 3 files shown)


DABH:
Hey @mazas-google , thanks so much for your help with reviewing my prior PR in #964 .  To build my ADK integration, it turns out we need one more extensibility point in ADK, which is captured in this PR.  Is there any chance you could take a look and offer your review?

Also, I tried to keep one conceptual change per PR, but my work surfaced a couple other points where we could make adk-go improvements (either things that help bring parity between adk-go and adk-python, or just niceties that make adk-go more extensible for people like me building integrations).  These are captured in #1055, #1053, #1057.  I am trying to ship my integration by the end of this week, so I'm very grateful for any help and guidance with these potential changes.  Thanks again!

kdroste-google (commented):
The recent upgrade to v2 changed things a bit. 
Have you seen examples/workflow/complex/main.go ?

DABH:
Thanks! Rebased onto v2 (`google.golang.org/adk/v2`). And yes, I looked at `examples/workflow/complex/main.go` — that graph engine (`AddFanOut`/`AddFanIn`, `WithMaxConcurrency`) fans out whole agents/nodes, which is great, but it's a layer above what this PR touches. #1050 targets the fan-out of parallel tool calls *within a single LLM turn* — the goroutine loop in `internal/llminternal/base_flow.go` `handleFunctionCalls` (still there in v2, with the `// TODO: check feasibility of running tool.Run concurrently`). The `TaskRunner` seam lets a caller control how those intra-turn tool calls are scheduled — a shared executor, a concurrency cap, or sequential execution for deterministic/replayable runs — instead of ADK always spawning its own goroutines. It mirrors the existing `platform.WithTimeProvider`/`WithUUIDProvider` seams and defaults to today's one-goroutine-per-task behavior (a no-op unless a caller opts in). Happy to align the naming/shape with the v2 conventions if you'd prefer.

platform/exec.go
```diff
@@ -0,0 +1,81 @@
...
+//
+// WithTaskRunner is the concurrency analog of WithTimeProvider and
+// WithUUIDProvider: it lets a host environment substitute its own execution
+// strategy for ADK's internal fan-out without the ADK runtime taking a
+// dependency on that environment.
+func WithTaskRunner(ctx context.Context, runner TaskRunner) context.Context {
+	return context.WithValue(ctx, taskRunnerKey{}, runner)
+}
+
+// RunTasks runs every task in tasks and blocks until all of them complete. If
```
wolo-lab, line 54: The "an installed runner blocks until all tasks complete" behavior is an unenforced contract, and I'd rather not rely on every custom-runner author respecting it. `handleFunctionCalls` reads `fnResponseEvents[i]` right after `RunTasks` returns, so a runner that returns early (e.g. a fire-and-forget worker pool) would cause a data race and dropped tool responses. I'd prefer we enforce the barrier on our side, something like:

```go
func RunTasks(ctx context.Context, tasks []func(context.Context)) {
	if len(tasks) == 0 {
		return
	}
	var wg sync.WaitGroup
	wg.Add(len(tasks))
	if ctx != nil {
		if r, ok := ctx.Value(taskRunnerKey{}).(TaskRunner); ok && r != nil {
			wrappedTasks := make([]func(context.Context), len(tasks))
			for i, task := range tasks {
				wrappedTasks[i] = func(c context.Context) {
					defer wg.Done()
					task(c)
				}
			}
			r(ctx, wrappedTasks)
			wg.Wait()
			return
		}
	}
	for _, task := range tasks {
		go func() {
			defer wg.Done()
			task(ctx)
		}()
	}
	wg.Wait()
}
```

wolo-lab (changes requested):

platform/exec.go
DABH: Good catch — agreed the block-until-complete contract shouldn't be left to each runner to honor. Adopted your approach: `RunTasks` now wraps each task to signal a `sync.WaitGroup` and waits on it after the runner returns, so `handleFunctionCalls` is safe to read `fnResponseEvents[i]` regardless of a custom runner's own blocking behavior. Kept the default one-goroutine-per-task path unchanged. Thanks!

wolo-lab (approved):
Excellent, thank you for the quick turnaround. LGTM!

PR #1053, opened 2026-06-16
fix: stable ordering of generated tool-confirmation requests

### LINKED ISSUE #1052: Tool-confirmation request parts are emitted in unstable (map-iteration) order
## 🔴 Required Information

**Describe the Bug:**
`generateRequestConfirmationEvent` (in `internal/llminternal`) builds the `adk_request_confirmation` function-call parts (and the parallel `LongRunningToolIDs` slice) by ranging directly over `EventActions.RequestedToolConfirmations`, which is a Go map. Because Go randomizes map iteration order, the order of the emitted parts (and IDs) varies run-to-run for the exact same input whenever a single response requests more than one tool confirmation.

**Steps to Reproduce:**
1. Trigger a model turn that requests confirmation for 2+ tool calls in one response (so `RequestedToolConfirmations` has multiple entries).
2. Generate the request-confirmation event for that response repeatedly.
3. Observe that the order of the resulting confirmation parts / `LongRunningToolIDs` differs between runs.

**Expected Behavior:**
For the same set of requested confirmations, the emitted parts and `LongRunningToolIDs` come out in a stable, predictable order across runs.

**Observed Behavior:**
The order follows Go's randomized map iteration, so it changes run-to-run. This makes the output non-reproducible and can surface as flaky tests / unstable snapshots and inconsistent logs.

**Environment Details:**
 - ADK Library Version: main (longstanding; not version-specific)
 - OS: N/A (Go map iteration is randomized on all platforms)
 - Go Version: N/A

**Model Information:**
 - Which model is being used: N/A (independent of model; depends only on having >1 requested tool confirmation)

---

## 🟡 Optional Information

**Regression:** No — present since the request-confirmation event code was added.

**Additional Context:**
Straightforward fix: collect the funcIDs and `sort.Strings` them before building the parts, so iteration is over a sorted key slice. A PR with this fix and a regression test is open.

```go
// With >1 entry in functionResponseEvent.Actions.RequestedToolConfirmations,
// generateRequestConfirmationEvent emits parts in map-iteration order, which
// Go randomizes -- so repeated calls can produce different orderings.
```

 - Often (50%+) — whenever a response carries more than one requested confirmation.

DABH:
### Link to Issue

Closes #1052

### Summary

`generateRequestConfirmationEvent` (in `internal/llminternal`) ranged `EventActions.RequestedToolConfirmations` (a Go map) directly, so the order of the emitted `adk_request_confirmation` parts and the parallel `LongRunningToolIDs` slice followed Go's randomized map iteration and varied run-to-run whenever a response requested more than one tool confirmation.

This now iterates the function-call parts of the model response (`functionCallEvent.Content.Parts`) in order — skipping parts with a nil `FunctionCall`, then looking up each confirmation from `RequestedToolConfirmations` by call ID (skipping if absent) — so confirmations are emitted in **model-response order**. That order is deterministic and reproducible for the same input, mirrors adk-python (which emits in the order confirmations were requested), and aligns with execution flow. The intermediate `functionCallParts` map and the `sort` import are no longer needed. No behavior change other than ordering.

### Testing Plan

**Unit Tests:**

`internal/llminternal/functions_test.go` (`TestGenerateRequestConfirmationEventResponseOrder`): builds a response whose function-call parts are in a deliberately non-sorted order, and asserts the emitted parts and `LongRunningToolIDs` follow that model-response order and are identical across repeated calls.

```
$ go test ./internal/llminternal/...
```

N/A - pure output-ordering fix with no ADK Web / Runner UI surface; covered by the unit test above.

### Checklist

- [ ] I have manually tested my changes end-to-end. _(N/A — internal ordering fix; see above.)_

### DIFF
--- a/internal/llminternal/functions.go
+++ b/internal/llminternal/functions.go
@@ -43,16 +43,18 @@ func generateRequestConfirmationEvent(
 
 	parts := []*genai.Part{}
 	longRunningToolIDs := []string{}
-	functionCallParts := make(map[string]*genai.Part, len(functionCallEvent.Content.Parts))
-	for _, part := range functionCallEvent.Content.Parts {
-		if part.FunctionCall != nil {
-			functionCallParts[part.FunctionCall.ID] = part
-		}
-	}
 
-	for funcID, confirmation := range functionResponseEvent.Actions.RequestedToolConfirmations {
-		originalPart, ok := functionCallParts[funcID]
-		if !ok || originalPart.FunctionCall == nil {
+	// Emit confirmations in the order their function calls appear in the model
+	// response, mirroring adk-python. Iterating Content.Parts (an ordered slice)
+	// rather than ranging RequestedToolConfirmations (a map, whose iteration
+	// order Go randomizes) keeps the emitted order deterministic and aligned
+	// with execution flow.
+	for _, originalPart := range functionCallEvent.Content.Parts {
+		if originalPart.FunctionCall == nil {
+			continue
+		}
+		confirmation, ok := functionResponseEvent.Actions.RequestedToolConfirmations[originalPart.FunctionCall.ID]
+		if !ok {
 			continue
 		}
 
--- a/internal/llminternal/functions_test.go
+++ b/internal/llminternal/functions_test.go
@@ -299,6 +299,103 @@ func TestGenerateRequestConfirmationEventPreservesThoughtSignature(t *testing.T)
 	}
 }
 
+// TestGenerateRequestConfirmationEventResponseOrder verifies that the emitted
+// confirmation parts (and the parallel LongRunningToolIDs slice) follow the
+// order of the function calls in the model response (Content.Parts), regardless
+// of the (randomized) Go map iteration order of RequestedToolConfirmations, and
+// that repeated calls produce an identical order.
+func TestGenerateRequestConfirmationEventResponseOrder(t *testing.T) {
+	ctx := &mockInvocationContext{
+		invocationID: "inv_1",
+		agentName:    "agent_1",
+		branch:       "main",
+	}
+
+	// Confirmations are emitted in the order their function calls appear in the
+	// model response (Content.Parts), independent of funcID string order or map
+	// insertion order. responseOrder is deliberately not sorted.
+	responseOrder := []string{"call_c", "call_a", "call_d", "call_b"}
+
+	functionCallParts := make([]*genai.Part, 0, len(responseOrder))
+	requestedConfirmations := make(map[string]toolconfirmation.ToolConfirmation, len(responseOrder))
+	for _, id := range responseOrder {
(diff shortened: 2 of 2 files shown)


kdroste-google (commented):
Hi, @DABH, I'm sorry, I do not understand the reasons behind. 
Shouldn't it be handled by the testing code instead? 
Something like https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#SortSlices  or in-test sorting. 
WDYT?

DABH:
Fair question — the motivation is runtime determinism, not test assertions. `generateRequestConfirmationEvent` emits the confirmation parts (and the parallel `LongRunningToolIDs` slice) by ranging a map, so the emitted order varies run to run. `cmpopts.SortSlices` would fix a test comparison but not the actual emitted order — and for any record/replay execution model, that emitted event is part of recorded history, so a differently-ordered emission on replay diverges from what was recorded. The change is a minimal sort-by-funcID at the point of emission (stdlib `sort`, no behavior change beyond stable order).

That said, this is the least essential of my PRs and I don't feel strongly: if you'd rather ADK not sort here, I'm happy to close this and have the consumer sort defensively instead. Your call. (Rebased onto v2.)

DABH:
@wolo-lab Thanks so much for your reviews on my other PRs and for helping get things merged.  This is the last one I have open right now that's impacting the integrations I'm building.  Lmk if you have any thoughts.  Thanks so much again, and I promise I will try to stay out of your guys' hair after this one 😅

wolo-lab (changes requested):
@DABH, thank you for your contributions. Feel free to send us more changes. We haven't had the bandwidth to process them in a timely manner, but we are working on improving this.

Regarding this PR:
Let's change the order here: instead of sorting the confirmations by funcID,
let's emit them in the order the function calls appear in the model response.
That mirrors adk-python's approach (it emits in the order confirmations were
requested) and is more natural from an execution-flow perspective (and still stable).
Concretely, I'd suggest removing `functionCallParts` altogether and iterating
over `functionCallEvent.Content.Parts` directly: in the loop, skip parts with a
nil `FunctionCall`, then read the confirmation from `RequestedToolConfirmations`
by the call ID (skipping if absent). 
WDYT?

DABH:
Thanks @wolo-lab — that's a nicer approach; done. `generateRequestConfirmationEvent` now iterates `functionCallEvent.Content.Parts` directly (skipping parts with a nil `FunctionCall`, then looking up each confirmation from `RequestedToolConfirmations` by call ID, skipping if absent), so confirmations are emitted in model-response order. That's deterministic, matches adk-python, and drops the `functionCallParts` map and the sort (and the `sort` import). Updated the test to assert the emission follows `Content.Parts` order and is stable across repeated calls. PTAL.

wolo-lab (approved):
Thank you. I look forward to your further contributions.

PR #1055, opened 2026-06-16
feat: expose the tool-declaration packer (PackTool) as a public API

### LINKED ISSUE #1054: Expose the tool-declaration packer (PackTool) as a public API

The helper that consolidates multiple function tools' declarations into a single `genai.Tool` on a `model.LLMRequest`, `PackTool`, lives in `internal/toolinternal/toolutils`, so it is not importable from outside the module. Anyone building a custom tool or toolset that needs to attach its function declarations to an `LLMRequest` the same way the built-in ADK tools do has to re-implement this logic by hand, which is easy to get subtly wrong and tends to drift from ADK's behavior over time.

### Proposed Solution

Expose the packer as a public API:

- A new public package `tool/toolutils` with `func PackTool(req *model.LLMRequest, t Packable) error` and a small `Packable` interface (`Name() string`, `Declaration() *genai.FunctionDeclaration`).
- The existing `internal/toolinternal/toolutils.PackTool` delegates to it, so all current callers are unchanged.

### Impact on your work

I'm doing exactly what's described above for an ADK integration I'm building.  This fix would be a material improvement to the maintainability of my integration, although it's not strictly required.

---

### Alternatives Considered

- Re-implementing the packing logic in each external package - I can do this, but it's duplicative and prone to drifting from ADK's behavior.

### Willingness to contribute

Yes - already done, see #1055 

```go
// package tool/toolutils
type Packable interface {
	Name() string
}

func PackTool(req *model.LLMRequest, t Packable) error
```

### Additional Context

The internal package keeps its own `Tool` interface and delegates to the public `PackTool`; the public package depends only on `model` and `genai` (no import cycle).

DABH:
### Link to Issue

Closes #1054

### Summary

Exposes ADK's tool-declaration packer as a public API so external tools / toolsets can consolidate their function declarations onto a `model.LLMRequest` the same way the built-in tools do, instead of re-implementing it.

- New public package `tool/toolutils`: `PackTool(req *model.LLMRequest, t Tool) error` plus a `Tool` interface (`Name()` / `Declaration()`). The packing logic lives here.
- The internal `internal/toolinternal/toolutils` package is **removed**; all of its callers now use the public `tool/toolutils` directly.
- No cyclical imports are created; the public package depends only on `model` and `genai`.

### Testing Plan

**Unit Tests:**

`tool/toolutils/toolutils_test.go`: packing one tool registers it and creates one `genai.Tool` with one declaration; a second tool appends to the same `genai.Tool`; a duplicate name returns an error; a tool whose `Declaration()` is nil still registers and adds no declaration. The migrated internal callers continue to pass their existing tests.

```
$ go test ./tool/toolutils/... ./internal/...
ok  google.golang.org/adk/v2/tool/toolutils
```

N/A (this exposes existing logic via a public package; behavior is unchanged and covered by the unit tests above).

### Checklist

- [ ] I have manually tested my changes end-to-end. _(N/A — API-exposure refactor; see above.)_

### DIFF
--- a/tool/toolutils/toolutils.go
+++ b/tool/toolutils/toolutils.go
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package tool defines internal-only interfaces and logic for tools.
+// Package toolutils provides public helpers for packing tool declarations
+// into a model.LLMRequest. It allows external code to consolidate tool
+// function declarations the same way the built-in ADK tools do, without
+// re-implementing the logic.
 package toolutils
 
 import (
@@ -23,31 +26,33 @@ import (
 	"google.golang.org/adk/v2/model"
 )
 
+// Tool is implemented by any tool that can be packed into a
+// model.LLMRequest via PackTool.
 type Tool interface {
 	Name() string
 	Declaration() *genai.FunctionDeclaration
 }
 
-// The PackTool ensures that in case there is a usage of multiple function tools,
+// PackTool ensures that in case there is a usage of multiple function tools,
 // all of them are consolidated into one genai tool that has all the function declarations
 // provided by the tools. So, if there is already a tool with a function declaration,
 // it appends another to it; otherwise, it creates a new genai tool.
-func PackTool(req *model.LLMRequest, tool Tool) error {
+func PackTool(req *model.LLMRequest, t Tool) error {
 	if req.Tools == nil {
 		req.Tools = make(map[string]any)
 	}
 
-	name := tool.Name()
+	name := t.Name()
 
 	if _, ok := req.Tools[name]; ok {
 		return fmt.Errorf("duplicate tool: %q", name)
 	}
-	req.Tools[name] = tool
+	req.Tools[name] = t
 
 	if req.Config == nil {
 		req.Config = &genai.GenerateContentConfig{}
 	}
-	if decl := tool.Declaration(); decl == nil {
+	if decl := t.Declaration(); decl == nil {
 		return nil
 	}
 	// Find an existing genai.Tool with FunctionDeclarations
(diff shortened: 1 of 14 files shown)


internal/toolinternal/toolutils/toolutils.go
kdroste-google, line 1: Let's remove this file completely, it doesn't make sense after having the exported func available.

tool/toolutils/toolutils.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+
+	"google.golang.org/genai"
+
+	"google.golang.org/adk/model"
+)
+
+// Packable is implemented by any tool that can be packed into an
+// model.LLMRequest via PackTool.
+type Packable interface {
```
kdroste-google: Could you please rename it to Tool as it was in the internal package?

kdroste-google (dismissed):
Hi, @DABH , thank you for this contribution. 

I assume you have a real-life scenario where you can benefit from having this functionality available OOB.

If that's the case, let's move on.
Having in internal package as a proxy doesn't make any sense to me, so could you get rid of it and use the public version instead? 

Thank you!

DABH:
Thanks! Yes, there's a concrete use case: out-of-tree tools/toolsets that need to consolidate their function declarations onto an `LLMRequest` exactly the way the built-in tools do — today that logic is internal, so any external tool has to reimplement it.

Addressed both points: deleted `internal/toolinternal/toolutils/toolutils.go` entirely and migrated its callers to the public `tool/toolutils`, and renamed the exported interface `Packable` → `Tool` to match the old internal name. Also rebased onto v2. PTAL.

tool/toolutils/toolutils.go
```diff
@@ -23,31 +26,33 @@ import (
 	"google.golang.org/adk/v2/model"
 )
 
+// Tool is implemented by any tool that can be packed into an
```
wolo-lab: super nit: `an` -> `a`

wolo-lab (dismissed):
Please update PR description as it is outdated after removing the old toolutils.go

DABH:
Thanks @wolo-lab! Fixed the `an` → `a` nit and updated the PR description to match the final state — public `tool/toolutils` with the `Tool` interface, the internal package removed and its callers migrated, rebased onto v2.

wolo-lab (approved):
Thank you, LGTM!

kdroste-google:
Hi, @DABH, thank you for all your contributions! It seems all have been already merged 🥳

PR #1057, opened 2026-06-16
feat: add a name-based model registry (Register/NewLLM)

### LINKED ISSUE #1056: Add a name-based model registry (LLMRegistry equivalent)

adk-python can construct a model from just its name string via its `LLMRegistry` (`LLMRegistry.new_llm(name)` / `register`), but adk-go has no equivalent. In Go, callers must hold a concrete constructor (e.g. `gemini.NewModel(ctx, name, cfg)`) and wire it up themselves; there is no way to go from a model-name string (such as the `Model` field on an `LLMRequest`) to an `LLM` without hard-coding the provider. This is a gap relative to adk-python and makes it awkward to build components that accept a model name and construct the model lazily.

### Proposed Solution

Add a public, name-based registry in `model`, mirroring adk-python's `LLMRegistry`:

- `model.Register(namePattern string, factory Factory)`: register a constructor for model names matching a pattern.
- `model.NewLLM(ctx, name) (LLM, error)`: construct the `LLM` for the first registered pattern that matches `name`.
- Provider packages self-register in `init()` (e.g. `model/gemini` registers `"(?i)gemini-.*"`).

### Impact on your work

Lets code construct an `LLM` from a name string without taking a hard dependency on a specific provider package, matching how adk-python works.  I am attempting to do this in a Go ADK integration I'm currently building and would like to expose this as a downstream feature for usability, so it would be nice to have this in adk-go.

---

### Alternatives Considered

- Requiring every caller to pass a concrete constructor or a name->factory map; this works (and is basically what I'm doing now), but duplicates wiring everywhere and diverges from adk-python.

### Willingness to contribute

Yes - already implemented, see #1057 

```go
// package model
type Factory func(ctx context.Context, name string) (LLM, error)

func Register(namePattern string, f Factory)              // regexp, first-match-wins, init-time
func NewLLM(ctx context.Context, name string) (LLM, error)
```

### Additional Context

In the proposed approach, registrations match by regular expression in registration order (first match wins) under a mutex. The registry path constructs with application-default credentials; callers needing custom configuration continue to call the provider constructor (e.g. `gemini.NewModel`) directly.

DABH:
### Link to Issue

Closes #1056

### Summary

Adds a public, name-based model registry _mirroring adk-python's `LLMRegistry`_  so an `LLM` can be constructed from just a model-name string.

- `model.Register(namePattern string, Factory)` and `model.NewLLM(ctx, name) (LLM, error)`. Registrations are matched by regular expression in registration order (first match wins), guarded by a `sync.RWMutex`; the matching factory is copied out before the lock is released so construction never runs under the lock.
- `model/gemini` self-registers `"(?i)gemini-.*"` in an `init()`, constructing via `gemini.NewModel(ctx, name, nil)` (application-default credentials; callers needing custom config continue to call `NewModel` directly).
- Purely additive, no existing behavior changes.

### Testing Plan

**Unit Tests:**

`model/registry_test.go`: register + lookup, no-match error, first-match-wins precedence, and concurrent `Register`/`NewLLM` access (so it is meaningful under `-race`).

```
$ go test -race ./model/...
ok  google.golang.org/adk/model
ok  google.golang.org/adk/model/apigee
ok  google.golang.org/adk/model/gemini
```

N/A; additive registry with no ADK Web / Runner UI surface; covered by the unit tests above.

### Checklist

- [ ] I have manually tested my changes end-to-end. _(N/A — additive registry; see above.)_

### DIFF
--- a/model/registry.go
+++ b/model/registry.go
@@ -1,0 +1,124 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The name-based model registry lets an [LLM] be constructed from just a
+// model-name string via [NewLLM], mirroring Python ADK's
+// LLMRegistry.new_llm(name).
+//
+// Registration is opt-in: provider packages do not register themselves on
+// import. A user who wants name-based Gemini resolution registers it
+// explicitly, using an anchored pattern:
+//
+//	model.Register("^(?i)gemini-.*", func(ctx context.Context, name string) (model.LLM, error) {
+//		return gemini.NewModel(ctx, name, nil)
+//	})
+//
+// [NewLLM] then requires exactly one registered pattern to match a given name:
+// zero matches or two-or-more matches are reported as errors, so resolution
+// never depends on registration or import order.
+package model
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"sync"
+)
+
+// Factory constructs an [LLM] for the given model name.
+//
+// The name is the same string that was matched against the registered
+// pattern, allowing a single factory to serve a family of models (for
+// example, all "gemini-*" names).
+type Factory func(ctx context.Context, name string) (LLM, error)
+
+// registration pairs a compiled name pattern with the factory that builds the
+// corresponding [LLM].
+type registration struct {
+	pattern *regexp.Regexp
+	factory Factory
+}
+
+// registry holds all model registrations, guarded by mu.
+var (
+	mu       sync.RWMutex
+	registry []registration
+)
+
+// Register associates a model name pattern with a [Factory].
+//
+// namePattern is a regular expression compiled with [regexp.MustCompile];
+// it is matched against candidate model names using [regexp.Regexp.MatchString]
+// (i.e. an unanchored, partial match). To match a name exactly, anchor the
+// pattern with "^" and "$".
+//
+// Registration order is not significant: [NewLLM] requires exactly one
+// registered pattern to match a name, so patterns must not overlap for names a
+// caller intends to resolve. Register is intended to be called at
+// initialization time. It panics if namePattern is not a valid regular
+// expression, or if namePattern has already been registered — surfacing the
+// programming error at the registration site rather than later in [NewLLM].
+//
+// Register is safe for concurrent use.
+func Register(namePattern string, f Factory) {
+	re := regexp.MustCompile(namePattern)
+	mu.Lock()
+	defer mu.Unlock()
+	for _, reg := range registry {
+		if reg.pattern.String() == namePattern {
+			panic(fmt.Sprintf("model: LLM pattern %q is already registered", namePattern))
+		}
+	}
+	registry = append(registry, registration{pattern: re, factory: f})
+}
+
+// NewLLM builds an [LLM] for the given model name.
+//
+// It scans all registrations and requires exactly one whose pattern matches
+// name:
+//
+//   - If no registered pattern matches, NewLLM returns an error reporting that
+//     name was not found.
+//   - If exactly one matches, NewLLM invokes its [Factory] and returns the
+//     result.
+//   - If two or more match, NewLLM returns an error naming the ambiguous name
+//     and listing the matching patterns, rather than silently picking one.
+//
+// Because a match must be unique, the result never depends on registration or
+// import order.
+//
+// NewLLM is safe for concurrent use.
+func NewLLM(ctx context.Context, name string) (LLM, error) {
+	mu.RLock()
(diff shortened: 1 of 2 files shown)


model/gemini/gemini.go
```diff
@@ -32,6 +32,20 @@ import (
...
+// init registers Gemini model names (e.g. "gemini-2.5-flash") with the
+// package-level model registry so callers can build a [model.LLM] from just a
+// name via [model.NewLLM].
+//
+// The registry path constructs the client with application-default credentials
+// (a nil [genai.ClientConfig]). Callers that need a custom configuration
+// (custom HTTP client, API key, backend, etc.) should call [NewModel]
+// directly.
+func init() {
+	model.Register("(?i)gemini-.*", func(ctx context.Context, name string) (model.LLM, error) {
```
DABH: This is to mimic adk-python, where `LLMRegistry.new_llm("gemini-...")` resolves Gemini out of the box.  I'm also fine if we don't want any model registration to happen by default (could catch some users by surprise).

model/gemini/gemini.go
```diff
@@ -32,6 +32,20 @@ import (
 	"google.golang.org/adk/model"
 )
 
+// init registers Gemini model names (e.g. "gemini-2.5-flash") with the
```
kdroste-google: init will be executed only when a module is actually imported. 
It gives you some control over the order but I would not rely on this due to recursive handling (https://go.dev/ref/spec#Program_initialization). 

In order to use the registry, you have to 
```go
import (
   // import something 

	_ "google.golang.org/adk/model/gemini"
   // import other models

	"google.golang.org/adk/model"
   // .... 
) 
```

Maybe it would be better not to rely on this order, and instead return an error if multiple registrations are matching. What do you think?

model/gemini/gemini.go
kdroste-google: `^(?i)gemini-.*` ?

kdroste-google (commented):
@DABH , thank you for this contribution!

Please see the comments

What do you think about providing explicit env variables to let the model factory func use whatever it wants? Gemini uses env anyway...

DABH:
Thanks for the thorough look — all addressed (and rebased onto v2, which re-cut this against `model/llm.go`):

- **No more reliance on init/registration order:** `NewLLM` now returns an error when *more than one* registered pattern matches a name (and when none matches), instead of first-match-wins — so resolution never depends on registration or import order.
- **Dropped the implicit Gemini `init()` self-registration** entirely; `model/gemini` no longer registers itself on import. Registration is now explicit/opt-in, documented in the `model` package doc with the anchored pattern you suggested: `model.Register("^(?i)gemini-.*", func(ctx, name) (model.LLM, error) { return gemini.NewModel(ctx, name, nil) })`.
- On env/config: that also sidesteps the env question — `gemini.NewModel` already reads its config from the environment, so the factory picks up whatever the caller wants.

PTAL.

model/registry.go
```diff
@@ -0,0 +1,118 @@
...
+// caller intends to resolve. Register is intended to be called at
+// initialization time. It panics if namePattern is not a valid regular
+// expression, surfacing the programming error immediately.
+//
+// Register is safe for concurrent use.
+func Register(namePattern string, f Factory) {
+	re := regexp.MustCompile(namePattern)
+	mu.Lock()
+	defer mu.Unlock()
+	registry = append(registry, registration{pattern: re, factory: f})
```
wolo-lab, line 83: nit: Panic on duplicate patterns so conflicts surface at registration, not later in NewLLM. It will make clear where is the source of the issue.

wolo-lab (dismissed):
Taking over this PR from @kdroste-google who is OOO.
Thanks for the contribution, LGTM. One nit requested.

model/registry.go
DABH: Done — `Register` now panics if the same pattern is registered twice, so duplicates surface at the registration site. I kept `NewLLM`'s multiple-match error as well, since it also covers *distinct-but-overlapping* patterns (e.g. `^gemini-.*` and `^gemini-2.*`) that can't be detected as duplicates at registration time. Added a test for the panic. Thanks!

DABH:
Heads up on the red `test` job: the failure is `TestA2ACleanupPropagation` in `agent/remoteagent/v2` (`a2a_e2e_test.go:480: remote cleanup was not called`), which is unrelated to this PR — it's a timing flake that reproduces on a clean `main` checkout. Running it on `upstream/main` with no changes, it failed 1 of 6 runs with the same assertion. This PR only adds `model/registry.go`, and its tests (including `TestRegisterDuplicatePatternPanics` and `TestRegistryConcurrentAccess`) pass.

Could you re-run the job when you get a chance? Happy to also push a trivial no-op to re-trigger CI from my side if that's easier. Thanks!

wolo-lab:
> Heads up on the red `test` job: the failure is `TestA2ACleanupPropagation` in `agent/remoteagent/v2` (`a2a_e2e_test.go:480: remote cleanup was not called`), which is unrelated to this PR — it's a timing flake that reproduces on a clean `main` checkout. Running it on `upstream/main` with no changes, it failed 1 of 6 runs with the same assertion. This PR only adds `model/registry.go`, and its tests (including `TestRegisterDuplicatePatternPanics` and `TestRegistryConcurrentAccess`) pass.
> 
> Could you re-run the job when you get a chance? Happy to also push a trivial no-op to re-trigger CI from my side if that's easier. Thanks!

Of course, I'm re-running the failed job. I can confirm this test is flaky—I have it on my TODO list and will work on it in the near future.

wolo-lab (approved):
Thanks, LGTM!

PR #1060, opened 2026-06-17
test: add llmagent delegation tests

dpasiukevich:


agent/llmagent/llmagent_delegation_test.go
```diff
@@ -0,0 +1,2038 @@
...
+// and on machines without an API key, the .httprr files under
+// testdata/ are replayed deterministically. To re-record after
+// changing an Instruction, tool declaration, or agent shape, delete
+// the corresponding .httprr file and run:
+//
+//	GEMINI_API_KEY=... go test ./agent/llmagent/ -httprecord=TestDelegation_NN
+//
+// To re-record everything: -httprecord=TestDelegation.
+//
+// Assertions are structural (FC/FR names, scope ids, presence of
```
wolo-lab, line 32: Let's add a note about the purpose .*events.yaml files. It is natural to assume that they are used for assertions.

agent/llmagent/llmagent_delegation_test.go
```diff
@@ -0,0 +1,2038 @@
...
+	if _, err := svc.Create(context.Background(), &session.CreateRequest{
+		AppName:   delegationApp,
+		UserID:    delegationUser,
+		SessionID: delegationSession,
+	}); err != nil {
+		t.Fatalf("session.Create: %v", err)
+	}
+	dr := &delegationRunner{
+		t: t, r: r, svc: svc,
+		app: delegationApp, user: delegationUser, sessID: delegationSession,
```
wolo-lab, line 169: We could change to something like
```go
         t: t, r: r, svc: svc,
		app: delegationApp, user: delegationUser, sessID: delegationSession,
		skipDump: !recording, 
```

Making snapshot of *.events.yaml by default will be problematic for `blaze` and running all tests inside the repo will dirty the workspace.

agent/llmagent/llmagent_delegation_test.go
wolo-lab, line 1: I wonder if we could make the machinery reusable for other tests and split tests and the machinery to separate files.

wolo-lab (approved):

agent/llmagent/llmagent_delegation_test.go
dpasiukevich: The machinery code is only to render events in yaml to illustrate delegation -- so it's easy to take a look at agent setup and see which events are produced. I don't think such event printing would be useful in any other tests. But if it will, we can move the logic out to be reusable. For now let's keep as is.

agent/llmagent/llmagent_delegation_test.go
dpasiukevich: Good idea! Done.

PR #1061, opened 2026-06-17
fix: prevent saving non-text events to state

### LINKED ISSUE #577: `maybeSaveOutputToState` overwrites OutputKey with empty string on function-call events

tanghaowillow:
Closes: #577

Align OutputKey state updates with the Python ADK implementation, avoiding overwrites from function-call or function-response events.


jjsasha63 (approved):
This is right, and it applies cleanly onto main at a33494c6 — 133 commits past the base — with the full suite green. Reverting just llmagent.go fails all three new subtests, so they pin real behaviour.

Worth saying explicitly for anyone triaging: this supersedes #578, which makes the same `IsFinalResponse()` swap but stops there. The `hasTextPart` guard is the half that matters most in practice. `IsFinalResponse()` short-circuits to true when `Actions.SkipSummarization` is set or `LongRunningToolIDs` is non-empty (session/session.go:226), and both of those are function-response-only events with no text, so the swap on its own still writes `""` over the key. `SkipSummarization` is set in exitlooptool, agenttool, functiontool, mcptoolset and tool confirmation, so that path is common rather than exotic. Both halves match adk-python's `__maybe_save_output_to_state`, including the `has_text_part` guard and the comment explaining it.

Two non-blocking things.

The guard needs that comment. It reads as redundant next to `IsFinalResponse()` and the next person through here will delete it. Something close to upstream's wording would do: skip when there is no text part, to avoid overwriting a StateDelta value already set, because IsFinalResponse reports true for function-response-only events under SkipSummarization or LongRunningToolIDs.

The `IsFinalResponse()` half isn't pinned by any test. Change it back to `!event.Partial` while keeping `hasTextPart` and all three new subtests still pass. Three cases would close that: a text part alongside a `FunctionCall` part in the same event, which only `IsFinalResponse()` catches; a function-response part with `SkipSummarization` set; and a function-call part with `LongRunningToolIDs` set, where only `hasTextPart` saves you.

One thing for the description rather than the code. The change also means a model turn that emits a text preamble alongside a tool call no longer writes that preamble to `OutputKey`. That is intended and it is what Python does, but it is user-visible and worth stating.

PR #1063, opened 2026-06-18
fix(workflow): Align hierarchical node pathing and OutputFor metadata

baptmont:
Standalone agent wrappers previously stamped extraneous namespace prefixes onto events and omitted ancestor output tracking. Added synthetic root workflow tagging and dynamic context ancestor resolution to ensure deterministic parity with conformance test recordings.

### Summary
Aligns event `NodeInfo.Path` and `NodeInfo.OutputFor` metadata generation in the Go graph engine with the Python reference implementation.

### Root Cause & Solution
1. **Root Path Distortion**: Standalone agents run via `Runner.runNode` ride inside a synthetic wrapper workflow. Added `WithRootWrapper()` so the scheduler suppresses extraneous root path prefixes on emitted events.
2. **Missing Ancestor Metadata**: Terminal nodes designated as `use_as_output` in hierarchical sub-workflows were not reporting ancestor workflow paths in `OutputFor`. Added dynamic context propagation to pass ancestor lineage down through sub-schedulers.
3. **Terminal Output Capture**: Updated `WorkflowNode.Run` to correctly extract terminal node names from compound paths (e.g., splitting `parent/child@1` to match `child`).


workflow/persistence.go
```diff
@@ -418,9 +418,17 @@ func firstUserInput(events session.Events) any {
 // "parent/child@1"; its interrupt is owned by the nearest static
 // ancestor (the first path segment). Falls back to Author for the
 // LlmAgent node path, where Author == node name and no path is set.
-func eventNodeName(ev *session.Event) string {
+func eventNodeName(ev *session.Event, nodesByName map[string]Node) string {
```
wolo-lab, line 421: please add an unit test for it.

internal/context/invocation_context.go
```diff
@@ -123,4 +125,24 @@ func (c *InvocationContext) WithContext(ctx context.Context) agent.InvocationCon
 // override this method.
 func (c *InvocationContext) ResumedInput(string) (any, bool) { return nil, false }
 
+func (c *InvocationContext) Path() string {
+	if c.params.Path != "" {
+		return c.params.Path
+	}
+	if p, ok := c.Context.(interface{ Path() string }); ok {
```
wolo-lab: I don't see in what scenario we would read from c.Context(). The same for OutputForAncestors(). Looks like a dead code but I might be wrong.

workflow/scheduler.go
```diff
@@ -389,6 +390,20 @@ func (s *scheduler) startNode(n Node, input any, triggeredBy, branch string, res
...
+	nodePath := name + "@1"
+	if p := s.parentCtx.Path(); p != "" {
+		nodePath = p + "/" + name + "@1"
+	} else if s.graph.isRootWrapper {
+		nodePath = ""
+	}
+	var ancestors []string
+	if s.terminals[name] {
+		if s.parentCtx.Path() != "" {
+			ancestors = append([]string{s.parentCtx.Path()}, s.parentCtx.OutputForAncestors()...)
```
wolo-lab: We do the same also in handleEvent:770-774, we could have a helper for that to keep them in sync.

wolo-lab (approved):
After rebasing v2 to main, your branch needs rebasing - sorry!

Nit for the future: let's have a PR per bug.

internal/context/invocation_context.go
baptmont: It's for the scenarios with multiple wrapping layers. When contexts get re-wrapped by intermediate adapters (like branchOverride), the outer envelope's metadata fields are empty, so this fallback delegates down to the inner context.

PR #1065, opened 2026-06-19
refactor(session): remove usage of deprecated NewEvent 

wolo-lab:
## Problem
`session.NewEvent` is deprecated: it always uses the wall clock and a random
UUID, ignoring any platform-installed time/UUID providers
(`platform.WithTimeProvider` / `platform.WithUUIDProvider`). This makes events
created through it non-deterministic and not replay-safe, which matters for
workflow engines. The non-deprecated `session.NewEventWithContext` already
honors those providers, but call sites across the codebase still used the old
API, keeping the deprecated function alive.

## Solution
Migrate every call site to `NewEventWithContext`, (removal of the deprecated
`NewEvent` will happen in the following PR):
- Production and example call sites (`workflow/`, `runner/`, `agent/llmagent/`,
  `examples/`) now pass the in-scope context to `NewEventWithContext`.
- `synthesizeTaskFREvent` had no context parameter; the surrounding invocation
  context is threaded through from its caller.
- Test call sites are migrated too; helpers without a context use the test's
  `t.Context()` or `context.Background()`.
- Removed `session.NewEvent` and its dedicated test `TestNewEventDefaults`,
  along with the now-unused `uuid` imports.
This is a breaking change to the public `session` API surface (removal of the
deprecated `NewEvent`).


dpasiukevich (approved):
I don't see where session.NewEvent itself is removed (while the PR description says it's done in this PR).

wolo-lab:
> I don't see where session.NewEvent itself is removed (while the PR description says it's done in this PR).

It's about removing usages only. The session.newEvent itself will be covered in a dedicated PR (it was to simplify test migration of 1P users which was in progress), the description was updated.

PR #1085, opened 2026-06-24
fix(tool): sanitize anyOf/oneOf tool schemas for Vertex

wolo-lab:
## Problem
adk-go packs tool parameter schemas into the `FunctionDeclaration` raw and
forwards them to the model verbatim. Optional/union fields emitted by pydantic-
or MCP-based tools produce an `anyOf` (or `oneOf`) block alongside sibling
keywords such as `description`/`title`. Vertex function-calling rejects that
shape — "schema specified other fields alongside any_of. When using any_of, it
must be the only field set." — so any tool with such a parameter fails on
Vertex. adk-python enforces the same constraint.

### Sources:
* https://github.com/google/adk-python/issues/3281:
   * ``` 400 INVALID_ARGUMENT. {'error': {'code': 400, 'message': 'Unable to submit request because 'create_payee' functionDeclaration 'parameters.person.age' schema specified other fields alongside any_of. When using any_of, it must be the only field set. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling', 'status': 'INVALID_ARGUMENT'}}```
* https://github.com/googleapis/python-genai/issues/625
    *  ```400 INVALID_ARGUMENT. {'error': {'code': 400, 'message': 'Unable to submit request because one or more response schemas specified other fields alongside any_of. When using any_of, it must be the only field set. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/control-generated-output', 'status': 'INVALID_ARGUMENT'}}```   

## Solution
Add `toolutils.SanitizeSchemaForVertex`, which deep-copies a schema and
normalizes every node so a composition never sits next to sibling keys:
- a composition whose members are all plain type schemas collapses into a
  JSON-Schema type array (preserving `description`/`title`);
- a composition with structured members is reduced to the composition keyword
  alone.
The same normalization is exposed via `SanitizeJSONSchemaForVertex`, which
handles the two forms a tool schema takes in practice: a `*jsonschema.Schema`,
and the `map[string]any` that MCP tools arrive as on the client side (a plain
JSON unmarshal — `Tool.InputSchema` is typed `any`). Walking the map form is
what actually covers the primary trigger for this bug: pydantic-based MCP
servers.
Wired into `functiontool.Declaration()` and `mcptoolset`, so function-generated,
user-supplied, and MCP schemas are all covered.

Examples:

Before:
```json
{
  "type": "object",
  "properties": {
    "field": {
      "description": "an optional value",
      "anyOf": [
        { "type": "string" },
        { "type": "null" }
      ]
    }
  }
}
```

After:
```json
{
  "type": "object",
  "properties": {
    "field": {
      "description": "an optional value",
      "type": ["string", "null"]
    }
  }
}
```

Before:
```json
{
  "description": "optional model",
  "title": "Maybe",
  "anyOf": [
    { "type": "object", "properties": { "a": { "type": "string" } } },
    { "type": "null" }
  ]
}
```
After:
```json
{
  "anyOf": [
    { "type": "object", "properties": { "a": { "type": "string" } } },
    { "type": "null" }
  ]
}
```


karolpiotrowicz (commented):
The `anyOf`-with-siblings shape really does break tool calling on Vertex, and the change is carefully built: a deep copy rather than in-place mutation, a reflective walk so the struct form can't miss a subschema field, an explicit allow-list for the map form with a test that keeps it in sync with the library struct, and a regression test on each of the two wired call sites. I confirmed the fix works for the shape it targets: both new tests fail on the base commit and pass here, and `build` / `vet` / `gofmt` / `staticcheck` / `-race` / `apidiff` are all clean.

My concerns are about the transformation's edge cases and about how much it discards. The main one is that the "structured" branch removes *every* sibling key, which for some legal schemas removes the parameter list itself. I've left inline notes on each; the two I'd most want addressed before merge are the `anyOf`+`oneOf` interaction and the sibling-stripping scope.

A few things I couldn't settle from the code, offered as questions rather than claims:

- **Is the constraint confirmed for `parametersJsonSchema` specifically?** Both errors quoted in the description name the typed `parameters` field (`parameters.person.age`) or a response schema. `adk-python`'s converter that dereferences `$ref` and normalizes type lists also feeds `parameters`, not `parameters_json_schema` — on the JSON-Schema field it currently forwards pydantic's output (`anyOf` + `default` + `title`) unchanged. If Vertex applies the rule to both fields, this PR is right and `adk-python` has the same latent bug; if it only applies to `parameters`, this would be trading schema fidelity for a rejection that doesn't occur on this path. Do you have a request/response confirming it on `parametersJsonSchema`?

- **`functiontool.NewStreaming` isn't covered.** `tool/functiontool/streaming_function.go:114` does `decl.ParametersJsonSchema = f.inputSchema.Schema()` — same package, same `*jsonschema.Resolved` field, same public `Config.InputSchema`. Feeding it the exact schema from your new `TestFunctionTool_SanitizesAnyOfForVertex` still produces `anyOf` alongside `description`. Was leaving it out intentional (and if so, worth a comment), or is it just the one-line call the non-streaming path got?

- **A third producer, `internal/llminternal/outputschema_processor.go:133`,** forwards the public `llmagent.Config.OutputSchema` as `ParametersJsonSchema`. That's a `*genai.Schema`, which also has `AnyOf` and `Description` — and worth noting, `SanitizeJSONSchemaForVertex` has no `*genai.Schema` case, so wiring it in as-is would fall through to the `default` passthrough and silently do nothing.

- **Should this be gated on the backend?** It's applied unconditionally, but the repo already has `GetGoogleLLMVariant` and uses it for exactly this kind of backend-specific behaviour. As written, users on the Gemini API backend — who were never hitting this constraint — absorb the lossy transforms too. Gating on `BackendVertexAI` would also make the more aggressive rewrites much easier to justify.

One suggestion on testing: the gaps below are all shapes the current table doesn't construct. A property-style test asserting the invariant directly — walk the output and fail if any node has a non-empty `anyOf`/`oneOf` next to another key, plus assert idempotence — would catch this class. Your `hasCompositionWithSiblings` helper is already most of that oracle; it flags the `anyOf`+`oneOf` case correctly, it just never gets an input that carries both.

internal/toolinternal/toolutils/sanitize.go
```diff
@@ -0,0 +1,335 @@
...
+	if s == nil {
+		return
+	}
+	forEachChildSchema(s, sanitizeSchemaNode)
+
+	if len(s.AnyOf) > 0 {
+		s.AnyOf = normalizeComposition(s, s.AnyOf)
+	}
+	if len(s.OneOf) > 0 {
+		s.OneOf = normalizeComposition(s, s.OneOf)
```
karolpiotrowicz, line 75: `anyOf` and `oneOf` are normalized independently, and the second pass can undo the first. Two cases:

**A composition ends up beside `type` again.** With a structured `anyOf` and a plain `oneOf` on one node, the `anyOf` pass reduces the node to `{AnyOf, OneOf}`, then the `oneOf` pass collapses its plain members and writes `Type`/`Types` straight back onto that node:

```go
in:  &jsonschema.Schema{
        Description: "d",
        AnyOf: []*jsonschema.Schema{{Type: "object", Properties: ...}, {Type: "null"}},
        OneOf: []*jsonschema.Schema{{Type: "string"}, {Type: "null"}},
     }
out: {"type":["string","null"],"anyOf":[{"type":"object",...},{"type":"null"}]}
```

That's the shape this change exists to remove. The map form at lines 218-219 does the same. It also makes the function non-idempotent: sanitizing that output again yields `{"anyOf":[...]}`, a different result.

**The reverse order deletes a union outright.** Plain `anyOf` + structured `oneOf`: the `anyOf` pass sets `Types`, then the `oneOf` pass runs `*s = jsonschema.Schema{AnyOf: anyOf, OneOf: oneOf}` (line 88) and wipes it:

```go
in:  {AnyOf: [{string},{null}], OneOf: [{object, properties:{a:string}}]}
out: {"oneOf":[{"type":"object","properties":{"a":{"type":"string"}}}]}
```

The `string|null` alternative is gone — silent data loss, and unlike the first case this is a change in behaviour from the current code, which forwards both keywords intact.

Suggestion: normalize a node's compositions jointly rather than in two blind passes — decide once whether the node collapses to a type array or reduces to a single composition keyword, and never write `type` back onto a node that still carries one. A regression test with both keywords on one node would lock it in; `hasCompositionWithSiblings` already detects this output, it just never receives such an input.

internal/toolinternal/toolutils/sanitize.go
```diff
@@ -0,0 +1,335 @@
...
+// normalizeComposition removes the Vertex-rejected "composition + siblings"
+// shape from s. It returns the members to keep on the composition keyword (nil
+// when the composition was collapsed into a type array on s).
+func normalizeComposition(s *jsonschema.Schema, members []*jsonschema.Schema) []*jsonschema.Schema {
+	if collapseTypeUnion(s, members) {
+		return nil
+	}
+	// Members are structured: the composition must be the only field on s.
+	anyOf, oneOf := s.AnyOf, s.OneOf
+	*s = jsonschema.Schema{AnyOf: anyOf, OneOf: oneOf}
```
karolpiotrowicz, line 88: This reduction is wider than "strip the siblings Vertex objects to" — it drops the node's entire contents, including keywords that carry the tool's actual interface.

**`properties`/`required`/`type` go too.** `anyOf` is legal as a *constraint* alongside object properties, which is the standard "at least one of these fields" idiom and common in hand-written MCP schemas (it also satisfies MCP's requirement that the root be `type: "object"`):

```json
{"type":"object",
 "properties":{"email":{"type":"string"},"phone":{"type":"string"}},
 "anyOf":[{"required":["email"]},{"required":["phone"]}]}
```

becomes

```json
{"anyOf":[{"required":["email"]},{"required":["phone"]}]}
```

The model is now told the tool takes no parameters at all. On Vertex that turns a loud 400 into a silent mis-specification; on the Gemini API backend, where the original schema was accepted, it's a straight regression.

**`$defs` goes too, while `$ref` members stay.** Since `plainJSONMemberTypes` rejects `{"$ref": ...}`, a `$ref` union always lands here, and `{"$defs":{"Foo":...},"anyOf":[{"$ref":"#/$defs/Foo"},{"type":"null"}]}` comes out as `{"anyOf":[{"$ref":"#/$defs/Foo"},{"type":"null"}]}` with the definitions removed and the references dangling.

The map form at lines 252-256 has both behaviours.

Some options, roughly in increasing order of effort: keep `$defs`/`definitions`/`$id` unconditionally (they're not "fields alongside `any_of`" in the sense the error is about); push `description`/`title` down into each member rather than deleting them; and for a node that has both `properties` and a composition, consider moving the composition into an `allOf` wrapper — or leaving that node alone — instead of discarding the properties. At minimum it's worth making the docstring explicit that this branch discards the node's own constraints, since that's a much bigger claim than "reduced to the composition keyword alone".

internal/toolinternal/toolutils/sanitize.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+	switch {
+	case len(types) == 0:
+		return false
+	case len(types) == 1:
+		s.Type = types[0]
+		s.Types = nil
+	default:
+		s.Type = ""
+		s.Types = types
```
karolpiotrowicz, line 125: Collapsing `oneOf` into a type array quietly changes its meaning: `oneOf` is exclusive, a type array is not. The new test at `sanitize_test.go:91-103` is exactly the case where they differ:

```json
{"oneOf":[{"type":"integer"},{"type":"number"}]}   →   {"type":["integer","number"]}
```

`1` matches both `integer` and `number`, so `oneOf` **rejects** it, while the collapsed form accepts it. The transformation is a widening, and the doc comment at lines 37-38 only mentions "primitive unions collapse into a type array (siblings preserved)".

For a schema shown to a model this is unlikely to cause real harm, and the widening may well be the pragmatic choice — but it's worth either restricting the collapse to `anyOf` or calling out in the comment that `oneOf` exclusivity is deliberately discarded, so the next reader doesn't assume the rewrite is semantics-preserving.

internal/toolinternal/toolutils/sanitize.go
```diff
@@ -0,0 +1,335 @@
...
+	for _, m := range members {
+		if !isPlainTypeSchema(m) {
+			return false
+		}
+		for _, t := range memberTypes(m) {
+			if t == "null" {
+				hasNull = true
+				continue
+			}
+			if !slices.Contains(nonNull, t) {
```
karolpiotrowicz, line 107: `slices.Contains` over the accumulated list makes this de-duplication O(n²) in the number of composition members, and `n` isn't bounded by the seven JSON types: `plainJSONMemberTypes` (line 294) accepts any string as a type, so a schema with many distinct made-up type names grows the accumulator linearly. Since MCP `InputSchema` comes from a remote server, `n` is not under your control.

Measured locally, in `Tools()`:

```
n=1000    payload  19 KB   12.7 ms
n=4000    payload  76 KB    134 ms
n=16000   payload 304 KB   2.06 s
```

Not a security issue — you already trust that server to run tool calls — but 2 seconds of CPU for a 304 KB response is a lot for a de-dup, and the current code does no work here at all. A `map[string]bool` (or `slices.Sort` + `Compact`) is a couple of lines. Same pattern at line 277 in the map form.

internal/toolinternal/toolutils/sanitize.go
```diff
@@ -0,0 +1,335 @@
...
+	"github.com/google/jsonschema-go/jsonschema"
+)
+
+var (
+	schemaPtrType   = reflect.TypeOf((*jsonschema.Schema)(nil))
+	schemaSliceType = reflect.TypeOf([]*jsonschema.Schema(nil))
+	schemaMapType   = reflect.TypeOf(map[string]*jsonschema.Schema(nil))
+)
+
+// SanitizeSchemaForVertex returns a deep copy of s with every `anyOf`/`oneOf`
```
karolpiotrowicz, line 30: Small doc accuracy point: "returns a deep copy of s" is stronger than what `CloneSchemas` provides. Its own doc says *"The copy is shallow except for sub-schemas"*, so every non-schema slice and map — `Required`, `Enum`, `Examples`, `Extra`, `PropertyOrder`, `DependencyStrings` — is still shared with the caller:

```go
out := SanitizeSchemaForVertex(in)
out.Required[0] = "x"   // also mutates in.Required[0]
```

The sanitizer itself never writes through those, so there's no bug here today, and it's already an improvement over the current code which copies nothing. Just worth saying "deep-copies the schema tree" (and the same in the `SanitizeJSONSchemaForVertex` comment, where `deepCopyJSONValue` genuinely is deep) so nobody later relies on full isolation.

PR #1086, opened 2026-06-24
fix(llmflow): don't end invocation on a thought-only model turn

wolo-lab:
## Problem
A model turn whose content carries only thought / thought-signature parts has
no function call or response and is not partial, so `IsFinalResponse()`
classified it as final. The LLM run loop then stopped after that single model
call and never produced the real answer — a parity gap with adk-python that
surfaced intermittently.

## Solution
Treat a thought-only ("thinking") turn as non-final in the run loop
(`base_flow.go`): when the last event is thought-only, the loop calls the model
again instead of returning, so the answer surfaces. A new `isThoughtOnlyTurn`
helper recognizes turns whose parts are all explicit thought or bare
thought-signature parts. Covered by a helper unit test (`base_flow_test.go`)
and an end-to-end run-loop regression test (`llmagent_test.go`).

### DIFF
--- a/internal/llminternal/base_flow.go
+++ b/internal/llminternal/base_flow.go
@@ -115,7 +115,9 @@ func (f *Flow) Run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error]
 				}
 				lastEvent = ev
 			}
-			if lastEvent == nil || lastEvent.IsFinalResponse() {
+			// A thought-only ("thinking") turn reports as final but has no
+			// answer; don't stop on it — call the model again.
+			if lastEvent == nil || (lastEvent.IsFinalResponse() && !isThoughtOnlyTurn(lastEvent)) {
 				return
 			}
 			if lastEvent.LLMResponse.Partial {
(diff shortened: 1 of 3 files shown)


internal/llminternal/base_flow.go
```diff
@@ -128,6 +130,36 @@ func (f *Flow) Run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error]
...
+// isThoughtPart reports whether p carries only model "thinking" and no surfaced
+// content: an explicit thought part, or a bare thought-signature part.
+func isThoughtPart(p *genai.Part) bool {
+	if p.Thought {
+		return true
+	}
+	return len(p.ThoughtSignature) > 0 &&
+		p.Text == "" && p.FunctionCall == nil && p.FunctionResponse == nil &&
+		p.InlineData == nil && p.FileData == nil &&
+		p.CodeExecutionResult == nil && p.ExecutableCode == nil
```
dpasiukevich: Do we need these checks and why relying on `p.Thought` is not enough? If there's any new field in the genai API, this code will automatically assume it's a thought.

wolo-lab: I agree! Changed.

dpasiukevich (approved):

PR #1094, opened 2026-06-26
Clean context

kdroste-google:
Cleaning the context implementation.


agent/context.go
```diff
@@ -218,7 +220,7 @@ type Context interface {
 	// WithBranch returns a Context whose Branch() returns the
 	// given value; all other fields (path, runID, subScheduler,
 	// resumeInputs, embedded InvocationContext) are preserved.
-	WithBranch(branch string) Context
+	// WithBranch(branch string) Context
```
wolo-lab: to remove?

workflow/dynamic_scheduler.go
```diff
@@ -236,7 +244,7 @@ func (s *dynamicSubScheduler) runNode(child Node, input any, opts runNodeOptions
 	// as its own span. startNodeSpan returns a context carrying the span.
 	span, spanCtx := startNodeSpan(childCtx, child)
 	defer span.End()
-	childCtx = spanCtx
+	childCtx = childCtx.WithAgentContext(spanCtx)
```
wolo-lab: Old version should work as well

agent/agent.go
```diff
@@ -390,6 +371,18 @@ type invocationContext struct {
 	endInvocation  bool
 }
 
+// Apply implements [InvocationContext].
+func (c *invocationContext) WithICDelta(d *InvocationContextDelta) InvocationContext {
```
wolo-lab, line 375: shouldn't it apply for all? not only userContent?

agent/common_context.go
```diff
@@ -590,3 +479,23 @@ func (a *trackedArtifacts) Save(ctx context.Context, name string, data *genai.Pa
 	}
 	return resp, nil
 }
+
+func NewCleanToolContext(ctx Context, functionCallID string, actions *session.EventActions, confirmation *toolconfirmation.ToolConfirmation) (Context, error) {
```
wolo-lab: it is ued only in tests, can we remove it?

agent/common_context.go
```diff
@@ -29,63 +29,38 @@ import (
...
-	if p, ok := parent.(*commonContext); ok {
-		inherited = p.resumeInputs
+// NewContext returns a full Context backed by parent, with no callback,
+// tool, or node specializations. Use it wherever a plain run context is
+// needed (e.g. running an agent).
+// Please mind that if you already have commonContext, you should use WithDelta to
+// create child contextes
+func NewContext(parent InvocationContext) Context {
+	if _, ok := parent.(*commonContext); ok {
+		panic("Should not happen")
```
wolo-lab: maybe just return it?

```go
if ctx, ok := parent.(*commonContext); ok {
		return ctx
```
In this case it would be the same as `Promote(...)`, so maybe just remove it and keep only `Promote(...)`?

If not, then at least let's improve the error message, so it will be meaningful. just in case.

wolo-lab (approved):
Nice improvement! Some nits only.

agent/context.go
kdroste-google: sure, done

workflow/dynamic_scheduler.go
kdroste-google: true, reverted, thanks for catching this!

agent/agent.go
kdroste-google: Yes, good catch, it may be needed in the future, thank you!
Fixed

agent/common_context.go
kdroste-google: not really, it uses internals (commonContext)
Renamed the func to make the intent clear

agent/common_context.go
kdroste-google: The idea behind was to have a clean cut between creating a new commonContext (with default fields) based on pure InvocationContext vs creating a child Context from Context (inheriting fields).
It was important during the migration to the new version.
You are right, now it makes less sense. 
Rewritten

kdroste-google: PTAL

wolo-lab: LGTM

PR #1103, opened 2026-06-27
test(runner): cover node-runtime behaviors 

wolo-lab:
## Summary
Ports the runner-level node behaviors from adk-python's
`tests/unittests/runners/test_runner_node.py` into adk-go's
`runner/run_node_test.go`, to confirm behavioral parity around the
`run_node` path in `runners.py`. adk-go has no `Runner(node=...)`
constructor, so each case drives a node graph wrapped in a
`workflowagent` (or a plain `LlmAgent` root) through the same node
runtime.
Test-only change — no production code is touched.

## What's covered
Behaviors adk-go already implements correctly (now verified at the
runner level):
- terminal node output + persistence to the session
- node error propagation to the caller of `Run`
- event accumulation across multiple turns in one session
- `yield_user_message` defaulting to off
- `WithStateDelta` applied to session state before the node / `LlmAgent`
  runs, and surfaced on the yielded user event
- plain text not triggering the resume path

## Intentional divergence (guarded by characterization tests)
adk-python's `Runner._run_node_async` rejects malformed resume messages
(`test_mixed_fr_and_text_raises`, `test_resume_raises_on_unmatched_fr`,
`test_resume_raises_on_multi_invocation_fr`). adk-go is deliberately more
permissive: the A2A flow resumes a paused task by sending the human's
reply text **alongside** the approval function response in a single
message (`agent/remoteagent/v2/a2a_e2e_test.go`, `msg3`), and
`buildResumeResponses` ignores responses that answer no pending
interrupt.
Two characterization tests (`TestRunner_Node_AllowsMixedFunctionResponseAndText`,
`TestRunner_Node_ToleratesUnmatchedFunctionResponse`) lock in this
permissive behavior so a future "parity" change can't silently break
A2A. Porting the adk-python guards was tried and reverted because it
failed the A2A e2e tests.

### DIFF
--- a/runner/run_node_test.go
+++ b/runner/run_node_test.go
@@ -17,6 +17,7 @@ package runner_test
 import (
 	"context"
 	"iter"
+	"strings"
 	"testing"
 
 	"google.golang.org/genai"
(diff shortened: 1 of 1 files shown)


hanorik (approved):
LGTM.

kiratlambrentomangubat-dot (approved):

PR #1104, opened 2026-06-28
fix(llminternal): drop thought parts in ConvertForeignEvent

### LINKED ISSUE #1096: ConvertForeignEvent does not drop another agent's thought parts — diverges from adk-python's _present_other_agent_message
## 🔴 Required Information

**Describe the Bug:**
In multi-agent flows, ADK presents one agent's events to another agent as context by reframing them into a `user`-role "For context:" message via `ConvertForeignEvent` (`internal/llminternal/contents_processor.go`). Unlike the Python ADK's equivalent — `_present_other_agent_message` (`src/google/adk/flows/llm_flows/contents.py`), which drops thought parts as the first step of its per-part loop (`if part.thought: continue`) — `ConvertForeignEvent` has no such guard: a thought part with text falls into the text branch and is re-emitted as `"[author] said: <reasoning>"`, and an empty-text thought part falls into the `default` branch and is copied through verbatim. As a result, one agent's private chain-of-thought is replayed into another agent's context. Separately, a foreign turn that contains only thoughts produces a bare `"For context:"` content in adk-go, whereas adk-python drops the event entirely (returns `None`).

**Steps to Reproduce:**
1. Create two `LLMAgent`s that share a session and run in sequence — e.g. a `SequentialAgent` or `LoopAgent` with both as sub-agents and `IncludeContents: IncludeContentsDefault`.
2. Use a current thinking-capable Gemini model (e.g. `gemini-3.5-flash`, or any Gemini 3 thinking model) so the first agent's turn includes thought parts (`Thought == true`).
3. Run the first agent (it produces a turn that includes thoughts), then run the second agent.
4. In a `BeforeModelCallback` on the second agent, inspect `LLMRequest.Contents`: the first agent's thought is present in the `user`-role "For context:" content — as `"[<author>] said: <reasoning>"` for a thought with text, or as a raw thought part for an empty-text thought. adk-python drops both.

**Expected Behavior:**
Match adk-python: when presenting another agent's event as context, exclude thought parts (as `_present_other_agent_message` does with `if part.thought: continue`), and drop the event entirely when only the `"For context:"` header remains. A foreign agent's reasoning should never be replayed into the current agent's context.

**Observed Behavior:**
The other agent's reasoning is retained in the current agent's context — leaked as `"[author] said: …"` text for a thought with text, or carried through as a raw thought part for an empty-text thought — and a thought-only foreign turn produces a bare `"For context:"` content instead of being dropped.

**Environment Details:**
 - ADK Library Version: v1.4.0 (also `main` as of 2026-06 — `ConvertForeignEvent`'s per-part switch has no thought handling)
 - OS: platform-independent
 - Go Version: 1.25

**Model Information:**
 - Any thinking-capable model that populates thought parts; e.g. a current Gemini 3 thinking model such as `gemini-3.5-flash`. The defect is in the provider-agnostic contents processor, so it is independent of the model backend.

---

## 🟡 Optional Information

**Regression:** No — `ConvertForeignEvent` was ported from adk-python's `_convert_foreign_event` before adk-python fixed this (see Additional Context) and has never dropped thoughts; the divergence is just visible now that thinking-capable models populate thought parts in multi-agent flows.

**Additional Context:**
 - adk-go: the per-part `switch` in `ConvertForeignEvent` (`internal/llminternal/contents_processor.go`) has no `Thought` case — a thought with text hits the text branch (`"[author] said: …"`) and an empty-text thought hits the `default` branch (copied verbatim).
 - adk-python reference: `_present_other_agent_message` (`src/google/adk/flows/llm_flows/contents.py`) does `if part.thought: continue` and returns `None` when only the header remains. Note `_is_part_invisible` keeps function-call/response parts even when `thought=True`, but only for an agent's *own* turns. The unit test `test_other_agent_empty_content` (`tests/unittests/flows/llm_flows/test_contents.py`) asserts a thought-only *foreign* turn is filtered out, and `test_function_call_with_thought_not_filtered` in the same file documents that Gemini 3 Flash marks function calls `thought=True` — so this is current behavior, not specific to one model version.
 - Already fixed in adk-python: commit `a30851e` (2025-09-03, *"fix: Fixes `thought` handling in contents.py"*) states *"other agent's reply with thought will still be inserted in the outgoing LlmRequest due to the wrong `else` statement"*, renamed `_convert_foreign_event` → `_present_other_agent_message`, and added `test_other_agent_thoughts_are_excluded`. adk-go's `ConvertForeignEvent` (note the matching name) was ported from the pre-fix version and hasn't picked the fix up.
 - Same class of bug on sibling paths in adk-go (one agent's thoughts leaking into another agent's input), already tracked: #655 (`remoteagent`'s `presentAsUserMessage`, fixed) and #697 / #696 (`agenttool.Run()` leaking thoughts into the tool result, open). This issue covers the remaining in-process shared-session path — `ConvertForeignEvent` — which neither of those addresses.

```go
// Two LLM agents sharing a session, run in sequence. The second agent's request
// is built from the first agent's event via ConvertForeignEvent.
gen, _ := llmagent.New(llmagent.Config{
    Name: "first", Model: geminiThinkingModel, // emits Thought parts
    Instruction:     "Answer the user's question.",
    IncludeContents: llmagent.IncludeContentsDefault,
})
reviewer, _ := llmagent.New(llmagent.Config{
    Name: "second", Model: geminiThinkingModel,
    Instruction:     "Review the previous answer.",
    IncludeContents: llmagent.IncludeContentsDefault,
    BeforeModelCallbacks: []llmagent.BeforeModelCallback{
        func(_ agent.CallbackContext, req *model.LLMRequest) (*model.LLMResponse, error) {
            // req.Contents includes a user-role "For context:" content that still carries
            // the first agent's thought (as "[first] said: ..." or a raw thought part).
            return nil, nil
        },
    },
})
root, _ := sequentialagent.New(sequentialagent.Config{ /* AgentConfig with SubAgents: gen, reviewer */ })
```
Illustrative — `ConvertForeignEvent` is in an internal package, so the behavior is observed at the agent level as above, or directly in `internal/llminternal` tests by passing a foreign event whose content has a `Thought` part to `ConvertForeignEvent` and asserting the thought is excluded.

 - Always (100%) for a multi-agent handoff on a thinking-capable model.

nuthalapativarun:
- Closes: #1096

**Problem:** `ConvertForeignEvent` includes thought parts (`Part.Thought == true`) when building foreign-agent context, diverging from adk-python's `_present_other_agent_message` which skips parts where `thought=True`. This leaks internal reasoning into the context passed to other agents.

**Solution:** Skip thought parts in `ConvertForeignEvent`'s parts loop. If the event contains only thought parts (nothing meaningful remains after stripping), return `nil` so the call site can omit it entirely. Nil-check the return at the call site.

### Testing Plan
**Unit Tests:**
- [x] Added test case: thought-only event returns nil
- [x] Added test case: mixed event strips thought parts, retains regular parts

This is a pure logic fix with full unit test coverage; no E2E harness needed for this change.

### Checklist
- [x] Read CONTRIBUTING.md
- [x] Self-reviewed code
- [x] Commented hard-to-understand areas
- [x] Tests prove fix/feature works
- [x] Unit tests pass locally
- [x] Dependent changes merged

### DIFF
--- a/internal/llminternal/contents_processor.go
+++ b/internal/llminternal/contents_processor.go
@@ -87,7 +87,9 @@ func buildContentsDefault(agentName, invocationBranch string, events []*session.
 			continue
 		}
 		if isOtherAgentReply(agentName, ev) {
-			filtered = append(filtered, ConvertForeignEvent(ev))
+			if converted := ConvertForeignEvent(ev); converted != nil {
+				filtered = append(filtered, converted)
+			}
 		} else {
 			filtered = append(filtered, ev)
 		}
@@ -531,6 +533,9 @@ func ConvertForeignEvent(ev *session.Event) *session.Event {
 		Parts: []*genai.Part{{Text: "For context:"}},
 	}
 	for _, p := range content.Parts {
+		if p.Thought {
+			continue
+		}
 		switch {
 		case p.Text != "":
 			converted.Parts = append(converted.Parts, &genai.Part{
@@ -547,6 +552,11 @@ func ConvertForeignEvent(ev *session.Event) *session.Event {
 		default: // fallback to the original part for non-text and non-functionCall parts.
 			converted.Parts = append(converted.Parts, p)
 		}
+	}
+
+	// If all parts were thoughts, nothing meaningful remains after the header.
+	if len(converted.Parts) == 1 {
+		return nil
 	}
 
 	return &session.Event{ // made-up event. Don't go through types.NewEvent.
--- a/internal/llminternal/contents_processor_test.go
+++ b/internal/llminternal/contents_processor_test.go
@@ -586,6 +586,55 @@ func TestConvertForeignEvent(t *testing.T) {
 				Branch: "b",
 			},
 		},
+		{
+			name: "ThoughtOnlyEvent",
+			event: &session.Event{
+				Timestamp: now,
+				Author:    "foreign",
+				LLMResponse: model.LLMResponse{
+					Content: &genai.Content{
+						Role: "model",
+						Parts: []*genai.Part{
+							{Text: "internal reasoning", Thought: true},
+							{Text: "more thoughts", Thought: true},
+						},
+					},
+				},
+				Branch: "b",
+			},
+			want: nil, // thought-only event produces no useful context
+		},
+		{
+			name: "MixedThoughtAndTextEvent",
+			event: &session.Event{
+				Timestamp: now,
+				Author:    "foreign",
+				LLMResponse: model.LLMResponse{
+					Content: &genai.Content{
+						Role: "model",
+						Parts: []*genai.Part{
+							{Text: "internal reasoning", Thought: true},
+							{Text: "actual response"},
+						},
+					},
+				},
+				Branch: "b",
+			},
+			want: &session.Event{
+				Timestamp: now,
+				Author:    "user",
+				LLMResponse: model.LLMResponse{
+					Content: &genai.Content{
+						Role: "user",
+						Parts: []*genai.Part{
+							{Text: "For context:"},
+							{Text: "[foreign] said: actual response"},
+						},
+					},
+				},
+				Branch: "b",
+			},
+		},
 	}
 
 	for _, tc := range testCases {


indurireddy-TF:
Hi @nuthalapativarun ,Thanks for the fix! I verified the changes by running the required tests and confirming that the agent's private thought parts are now correctly filtered out of the foreign event context.

karolpiotrowicz:
Both this PR and #1203 make the same production change — a `p.Thought` guard at the head of the loop in `ConvertForeignEvent`, a `len(converted.Parts) == 1` check so a thought-only turn returns nil rather than a bare `For context:` header, and the matching nil check at the `buildContentsDefault` call site. I compared them directly and there is no behavioral difference between the two.

We are consolidating on #1203, on test coverage. It covers the empty-text thought part — a part with `Thought: true` and no text — which issue #1096 singles out because it matches none of the `switch` cases, reaches `default:`, and is appended verbatim with `Thought` still set. This PR handles that input correctly but has no test pinning it.

To be clear that this is the only reason: #1232 landed today and touched the same `if` statement, so **both** PRs now conflict with `main` and both need a rebase. That is not a point of difference between them.

On the order of events: this PR was opened four weeks earlier and reached the same fix first. Your `ThoughtOnlyEvent` case covers an event made up of two thought parts, which #1203 does not, and I have asked there that it be ported across with credit to you rather than dropped.

This sat two months without a review, which is on us and not on the patch. The diagnosis was right and it got there first.

PR #1105, opened 2026-06-30
fix(examples): keep complex workflow from hanging on flaky Google Search

wolo-lab:
## Summary
The `examples/workflow/complex` research pipeline fans three Google
Search-grounded researchers into a `JoinNode` barrier, then formats and
synthesizes their findings. Gemini's Search grounding intermittently
**stalls with neither a response nor an error**, and the example set no
per-call deadline — so a single stalled researcher hung the entire run
**forever**. Observed empirically: ~3/5 single-researcher runs stalled
past 30s; with three in parallel behind a Join barrier the pipeline
almost always hung.
The per-node `RetryConfig` did **not** help: it only retries calls that
*return* an error, and a hang returns nothing, so the retry/backoff
never triggered.

## Fix
Add a `resilientModel` decorator (implements `model.LLM`) that wraps
each agent's model and:
- **bounds** every call with a per-attempt `context.WithTimeout`, turning
  a stall into `context.DeadlineExceeded` instead of an infinite wait;
- **retries** timed-out/errored attempts with exponential backoff (fresh
  deadline each time);
- **fails open** after exhausting attempts, returning a short fallback
  message instead of an error, so the `JoinNode` still fires and the
  synthesis agent still produces a report (graceful degradation rather
  than a dead pipeline);
- distinguishes a cancelled *parent* context (real abort) from its own
  per-attempt timeout, propagating the former instead of failing open.
Wire it into each researcher (30s / 3 attempts, topic-specific fallback)
and synthesis (60s / 2 attempts, minimal-report fallback), drop the
misleading node `RetryConfig`, and refresh the example docs/comments.
The graph wiring is unchanged.

### DIFF
--- a/examples/workflow/complex/resilient_model.go
+++ b/examples/workflow/complex/resilient_model.go
@@ -25,26 +25,16 @@ import (
 	"google.golang.org/adk/model"
 )
 
-// resilientModel wraps a model.LLM to bound and, as a last resort, paper
-// over flaky calls — specifically Gemini's Google Search grounding, which
-// intermittently stalls with neither a response nor an error.
-//
-// Why it exists: a bare model call has no per-call deadline, so a stalled
-// grounding request hangs forever. A workflow.RetryConfig does not help —
-// it only retries calls that *return* an error, and a hang returns
-// nothing. And because the pipeline fans out to three researchers behind a
-// JoinNode barrier, a single stalled researcher would hang (or, with a
-// plain timeout, fail) the whole run.
-//
-// resilientModel addresses all three:
-//   - timeout: every attempt runs under its own context.WithTimeout, so a
-//     stall becomes a context.DeadlineExceeded instead of an infinite wait;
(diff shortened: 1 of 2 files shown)


dpasiukevich (approved):

examples/workflow/complex/resilient_model.go
```diff
@@ -0,0 +1,140 @@
...
+	"google.golang.org/genai"
+
+	"google.golang.org/adk/model"
+)
+
+// resilientModel wraps a model.LLM to bound and, as a last resort, paper
+// over flaky calls — specifically Gemini's Google Search grounding, which
+// intermittently stalls with neither a response nor an error.
+//
+// Why it exists: a bare model call has no per-call deadline, so a stalled
```
dpasiukevich: I'd only suggest to make the comments more consice.

wolo-lab: In samples my threshold is more relaxed. However, ok let's do it.

PR #1119, opened 2026-07-02
feat(guardrail): add ErrGuardrailBlocked sentinel and OnGuardrailBlockedCallback

### LINKED ISSUE #1116: feat: add ErrGuardrailBlocked sentinel type to distinguish policy denials from runtime errors in callbacks
## Summary

`BeforeToolCallback` and `BeforeModelCallback` can short-circuit execution by returning an `error`, but the framework has no way to distinguish a **deliberate policy denial** from a **runtime tool/model failure**. Both paths produce the same outcome: a generic error propagated up the call stack. This makes it impossible to build clean guardrail layers on top of the existing plugin system.

## Current Behaviour

```go
// Guardrail plugin blocking a transfer that exceeds a policy limit
func (g *CompliancePlugin) BeforeToolCallback() plugin.BeforeToolCallbackFunc {
    return func(ctx plugin.Context, t tool.Tool, args map[string]any) (map[string]any, error) {
        amount, _ := args["amount"].(float64)
        if amount > 10_000_000 {
            return nil, fmt.Errorf("transfer exceeds policy limit")
        }
        return nil, nil
    }
}
```

**Problems with this approach:**

1. **Indistinguishable from a crash** — the runner treats `fmt.Errorf("transfer exceeds policy limit")` identically to a panic recovery or network timeout. Error metrics count it as a tool failure, not a policy event.

2. **No structured denial context** — the error string is the only carrier of policy information. The calling agent has no machine-readable signal to act on (retry with different params, escalate to human, log to compliance audit trail).

3. **No audit routing** — there is no hook between "callback returned error" and "error propagated to caller" where the framework could route the event to a separate audit/observability sink.

4. **Misleading observability** — `OnToolErrorCallback` fires for both policy denials and genuine errors, making dashboards noisy and compliance reports unreliable.

## Proposed Change

Add a sentinel error type `ErrGuardrailBlocked` that the framework recognises as a deliberate policy denial rather than a runtime failure:

```go
// plugin/guardrail.go

// ErrGuardrailBlocked is returned from BeforeToolCallback or
// BeforeModelCallback to signal a deliberate policy denial.
// The framework routes it differently from runtime errors:
// it does not increment error metrics, does not fire
// OnToolErrorCallback / OnModelErrorCallback, and surfaces
// Reason as a structured response to the caller.
type ErrGuardrailBlocked struct {
    // Policy is a machine-readable policy identifier.
    // e.g. "CBN-NIP-TRANSFER-CAP", "GDPR-SENSITIVE-DATA"
    Policy string

    // Reason is a human-readable explanation surfaced to the caller.
    // e.g. "Transfer ₦15,000,000 exceeds single-transaction cap of ₦10,000,000"
    Reason string
}

func (e *ErrGuardrailBlocked) Error() string {
    return fmt.Sprintf("guardrail blocked [%s]: %s", e.Policy, e.Reason)
}
```

The runner checks `errors.As(err, &ErrGuardrailBlocked{})` after each callback invocation and, when matched:

- Returns a structured `GuardrailBlockedResponse` to the caller instead of propagating as a tool error
- Does **not** fire `OnToolErrorCallback` / `OnModelErrorCallback`
- Does **not** increment tool/model error metrics
- Fires a new optional `OnGuardrailBlockedCallback` (if registered) for audit routing

```go
// Optional audit hook — fires only on ErrGuardrailBlocked, never on runtime errors
type OnGuardrailBlockedCallbackFunc func(ctx Context, blocked *ErrGuardrailBlocked)
```

## Usage Example

```go
// Clean guardrail — policy denial is structurally distinct from a crash
func (g *CompliancePlugin) BeforeToolCallback() plugin.BeforeToolCallbackFunc {
    return func(ctx plugin.Context, t tool.Tool, args map[string]any) (map[string]any, error) {
        amount, _ := args["amount"].(float64)
        if amount > 10_000_000 {
            return nil, &plugin.ErrGuardrailBlocked{
                Policy: "CBN-NIP-TRANSFER-CAP",
                Reason: fmt.Sprintf(
                    "Transfer ₦%.0f exceeds the CBN NIP single-transaction cap of ₦10,000,000",
                    amount,
                ),
            }
        }
        return nil, nil
    }
}

// Audit hook — routes denials to compliance log, not error log
func (g *CompliancePlugin) OnGuardrailBlockedCallback() plugin.OnGuardrailBlockedCallbackFunc {
    return func(ctx plugin.Context, blocked *plugin.ErrGuardrailBlocked) {
        auditLog.Record(ctx, blocked.Policy, blocked.Reason)
    }
}
```

## Scope of Changes

| File | Change |
|---|---|
| `plugin/guardrail.go` | New file — `ErrGuardrailBlocked` type + `OnGuardrailBlockedCallbackFunc` |
| `plugin/plugin.go` | Add `OnGuardrailBlockedCallback()` to the `Plugin` interface (with default no-op) |
| `runner/plugin_manager.go` | Check `errors.As(err, &ErrGuardrailBlocked{})` after before-callbacks; route accordingly |
| `runner/runner.go` | Return structured blocked response instead of propagating as tool/model error |
| `examples/compliance/` | Nigerian fintech example: CBN NIP transfer cap + NDPA BVN/NIN output guard using the new type |

## Why Not Just Use a Custom Error Type Per Plugin?

Each plugin author today can define their own sentinel — but the **framework never inspects the error**. Without framework-level recognition, there is no way to suppress `OnToolErrorCallback`, clean up metrics, or route to an audit sink. The sentinel is only useful if the runner acts on it.

## Prior Art

- OpenAI Agents SDK: `GuardrailFunctionOutput.tripwire_triggered` — explicit boolean separation of guardrail block from tool error
- LangChain: `ToolException` — structured tool-layer exception routed differently from agent errors
- adk-python: `before_tool_callback` returning a `dict` to override result vs raising to signal error — partial separation but no structured denial type

Happy to implement this. I'll open a draft PR once there's signal on the approach.

kingztech2019:
Please ensure you have read the contribution guide before creating a pull request.

- Closes: #1116

### Testing Plan

**Unit Tests:**

```
ok  google.golang.org/adk/v2/guardrail                  0.002s
ok  google.golang.org/adk/v2/internal/llminternal       0.579s
ok  google.golang.org/adk/v2/plugin                     0.011s
ok  google.golang.org/adk/v2/agent/llmagent             0.038s
```

New tests added:
- `guardrail/guardrail_test.go` — 4 tests: error message formatting, `errors.As` unwrapping, interface compliance
- `internal/llminternal/base_flow_test.go` — 3 new `TestCallTool` cases: guardrail routes to `OnGuardrailBlockedCallback` not `OnToolErrorCallback`, default fallback response, non-guardrail errors still route to `OnToolErrorCallback`
- `plugin/plugin_manager_test.go` — 2 new `TestOnGuardrailBlockedCallbackPlugin` integration cases via the full plugin/runner stack

A runnable example agent is provided at `examples/compliance/`. It demonstrates a payment agent with two policy guardrails (AML narration screening and high-value transfer limit). When a tool call is blocked, `OnGuardrailBlockedCallback` returns a structured `{"status": "blocked", "policy": "...", "message": "..."}` response the LLM reasons about instead of a generic error.

To run:
```bash
export GOOGLE_API_KEY=<your-key>
go run ./examples/compliance -- --console
# Try: "Transfer 2000000 units to ACC123 for payment"
```

### Checklist

### Additional context

Currently, a `BeforeToolCallback` that returns any error is routed through `OnToolErrorCallback` — the same path as unexpected runtime crashes. This makes it impossible to distinguish a deliberate policy denial from a broken tool.

This PR adds a sentinel type `guardrail.ErrGuardrailBlocked{Policy, Reason}` and a dedicated `OnGuardrailBlockedCallback` chain. The runner uses `errors.As` to detect the sentinel and routes policy blocks separately. Existing behaviour for all other errors is unchanged.

### DIFF
--- a/agent/llmagent/llmagent.go
+++ b/agent/llmagent/llmagent.go
@@ -22,6 +22,7 @@ import (
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/v2/agent"
+	"google.golang.org/adk/v2/guardrail"
 	agentinternal "google.golang.org/adk/v2/internal/agent"
 	icontext "google.golang.org/adk/v2/internal/context"
 	"google.golang.org/adk/v2/internal/llminternal"
@@ -63,14 +64,20 @@ func New(cfg Config) (agent.Agent, error) {
 		onToolErrorCallback = append(onToolErrorCallback, llminternal.OnToolErrorCallback(c))
 	}
 
+	onGuardrailBlockedCallbacks := make([]llminternal.OnGuardrailBlockedCallback, 0, len(cfg.OnGuardrailBlockedCallbacks))
+	for _, c := range cfg.OnGuardrailBlockedCallbacks {
+		onGuardrailBlockedCallbacks = append(onGuardrailBlockedCallbacks, llminternal.OnGuardrailBlockedCallback(c))
+	}
+
 	a := &llmAgent{
-		model:                 cfg.Model,
-		beforeModelCallbacks:  beforeModelCallbacks,
-		afterModelCallbacks:   afterModelCallbacks,
-		onModelErrorCallbacks: onModelErrorCallbacks,
-		beforeToolCallbacks:   beforeToolCallbacks,
-		afterToolCallbacks:    afterToolCallbacks,
-		onToolErrorCallbacks:  onToolErrorCallback,
+		model:                       cfg.Model,
+		beforeModelCallbacks:        beforeModelCallbacks,
+		afterModelCallbacks:         afterModelCallbacks,
+		onModelErrorCallbacks:       onModelErrorCallbacks,
+		beforeToolCallbacks:         beforeToolCallbacks,
+		afterToolCallbacks:          afterToolCallbacks,
+		onToolErrorCallbacks:        onToolErrorCallback,
+		onGuardrailBlockedCallbacks: onGuardrailBlockedCallbacks,
 		instruction:           cfg.Instruction,
 		inputSchema:           cfg.InputSchema,
 		outputSchema:          cfg.OutputSchema,
@@ -326,6 +333,11 @@ type Config struct {
 
 	OnToolErrorCallbacks []OnToolErrorCallback
 
+	// OnGuardrailBlockedCallbacks are called when a BeforeToolCallback returns
+	// *guardrail.ErrGuardrailBlocked. This allows custom responses for policy
+	// denials, distinct from unexpected runtime errors.
+	OnGuardrailBlockedCallbacks []OnGuardrailBlockedCallback
+
 	// OutputKey is an optional parameter to specify the key in session state for the agent output.
 	//
 	// Typical uses cases are:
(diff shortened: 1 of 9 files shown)


karolpiotrowicz:
@wolo-lab flagging this for a direction call rather than a review.

The change adds a new top-level `guardrail/` package with an exported `ErrGuardrailBlocked` sentinel and a new exported callback on the agent configuration. Both the package boundary and the callback signature are long-lived public surface, and the issue asking for it was filed by the same contributor, so there is no independent signal yet on whether we want this shape.

PR #1120, opened 2026-07-02
fix: prevent HITL confirmation replay

huynhtrungcip:
Summary:
- Treat a HITL confirmation as already consumed when the original tool call has a response between the confirmation request and the current confirmation response.
- Add processor-level coverage for replayed confirmations.
- Add REST coverage that verifies replaying the same confirmation ID with a changed payload does not re-execute the tool.

Tests:
- go test ./internal/llminternal -run TestRequestConfirmationRequestProcessor -count=1 -v
- go test ./server/adkrest -run TestRESTHITLConfirmationReplayDoesNotReexecuteTool -count=1 -v
- go test ./internal/llminternal ./server/adkrest -count=1
- go test ./...


karolpiotrowicz (changes requested):
The replay this fixes is real, and the test that proves it is the right test. I checked out the branch and ran it against [the merge-base](https://github.com/google/adk-go/commit/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e): there the new REST test fails with `tool executions = 2 and last payload = {"approved_scope":"high-risk"}`, and with your change applied it passes. Driving it through `POST /run` rather than the processor is what makes it convincing — it shows the payload from the replayed confirmation reaching the tool, which is the part that turns a double-execution bug into a privilege-escalation one. Reusing the session event log as the marker is also the right call: nothing new is persisted, nothing needs eviction, and it survives a restart wherever the session does.

Two things need fixing before this goes in, both of them the same shape: with the change applied, a genuine human approval can be dropped and nothing says so. On a human-in-the-loop path a silent no-op is worse than an error, because the operator has no way to tell an approval that was refused from one that was never acted on.

## A pending approval can be cancelled by an earlier function response

[`removeAlreadyCompletedTools`](https://github.com/google/adk-go/blob/39b850b16fa72fc1564cc81f8feae74685788f4f/internal/llminternal/request_confirmation_processor.go#L266-L279) deletes on `resp.ID` alone. It never looks at `events[j].Author`, so any event that lands in the window can remove a pending resume, and the approval that follows is then dropped by [the guard immediately below the call](https://github.com/google/adk-go/blob/39b850b16fa72fc1564cc81f8feae74685788f4f/internal/llminternal/request_confirmation_processor.go#L216-L219) with no error and no event.

Measured through `POST /run`. Start a turn that requests confirmation for a tool whose original call id is `transfer-call-1`. Before approving, post an ordinary user message whose content carries `FunctionResponse{ID: "transfer-call-1"}`. Then post the real approval. On the merge-base the tool runs once. With this change it runs zero times.

Adding the author check to the helper fixes this one, and I confirmed it rather than guessing — with `if events[j].Author != agentName { continue }` at the top of the loop body, both of your new tests still pass and this case goes back to executing once. Worth saying that the existing pass just below at [lines 223-235](https://github.com/google/adk-go/blob/39b850b16fa72fc1564cc81f8feae74685788f4f/internal/llminternal/request_confirmation_processor.go#L223-L235) has the same gap, so this is not a convention your code broke — it inherits one. It is still worth closing in the new code.

## Rejecting a confirmation and then approving it silently does nothing

This one needs no attacker at all. Rejecting a confirmation makes ADK emit an agent-authored `FunctionResponse` carrying the **original** call id with an error body, and that event sits inside your window. A later approval of the same confirmation id is therefore deleted as "already completed". On the merge-base the approval runs the tool. With this change it runs zero times, silently.

Here is the real session log after a rejection, dumped from the GET-session endpoint, with the agent named `toctou_probe`:

```
EVENT[0] author="user"          TEXT go
EVENT[1] author="toctou_probe"  CALL id=transfer-call-1 name=transfer_funds
EVENT[2] author="toctou_probe"  RESP id=transfer-call-1 {"error":"...requires confirmation, please approve or reject"}
EVENT[3] author="toctou_probe"  CALL id=adk-e2ab... name=adk_request_confirmation
EVENT[4] author="user"          RESP id=adk-e2ab... {"confirmed":false}
EVENT[5] author="toctou_probe"  RESP id=transfer-call-1 {"error":"...call is rejected"}
EVENT[6] author="toctou_probe"  TEXT done
```

`EVENT[5]` is the rejection, and it carries `transfer-call-1`. Note the author check above does not help here, because that event is authored by the agent — I measured that too.

Two constraints on any fix. `EVENT[2]`, the "requires confirmation" placeholder, carries the same original call id and is the reason a normal first approval still works: it is emitted *before* the confirmation request, so it falls outside the window. That ordering comes from [`base_flow.go` lines 654-666](https://github.com/google/adk-go/blob/39b850b16fa72fc1564cc81f8feae74685788f4f/internal/llminternal/base_flow.go#L654-L666) and nothing currently pins it, so the fix is load-bearing on a yield order no test protects. And treating "a response exists for this id" as "the tool ran" is what makes both `EVENT[2]` and `EVENT[5]` look like completions when neither is one.

I am not claiming re-approval is a documented feature. It is not — [`tool_confirmation.go` lines 44-45](https://github.com/google/adk-go/blob/39b850b16fa72fc1564cc81f8feae74685788f4f/tool/toolconfirmation/tool_confirmation.go#L44-L45) describes exactly two outcomes and the example treats a rejection as terminal. Making rejection final may well be the behavior you want. What I would not ship is making it final *silently*, as a side effect, with no test asserting it, especially since [the console maps a blank line to `{"confirmed": false}`](https://github.com/google/adk-go/blob/39b850b16fa72fc1564cc81f8feae74685788f4f/cmd/launcher/console/hitl.go#L196-L208) — an accidental rejection is one keystroke away and now ends that confirmation with no message.

## Two limits on the protection, worth naming in the PR

Neither is a regression — both behave identically on the merge-base — but in each one the PR title's claim does not hold, so each is worth stating in the description rather than leaving for someone to discover.

Concurrent submissions still both execute. Two `POST /run` requests carrying the same confirmation id, issued at once against the same session, run the tool twice and deliver both payloads. The check reads a session snapshot, decides, and only then executes, and nothing serialises a session across that span. Worth knowing that `-race` cannot see this and never will — I injected a deliberate unsynchronised write on the same concurrent path to confirm the detector was armed and reaching this code, and it reported the race, so the clean result on your branch is a genuine negative. This is a logical window, not a memory race.

A confirmation-gated tool that is also long-running has no marker at all. With `RequireConfirmation: true` and `IsLongRunning: true` — [both public fields of `functiontool.Config`](https://github.com/google/adk-go/blob/39b850b16fa72fc1564cc81f8feae74685788f4f/tool/functiontool/function.go#L48-L55) — the resumed tool returns no result, so [`base_flow.go` lines 1192-1199](https://github.com/google/adk-go/blob/39b850b16fa72fc1564cc81f8feae74685788f4f/internal/llminternal/base_flow.go#L1192-L1199) returns before emitting any function response. No response event means no marker, so the plain serial replay still works: I measured one execution after the first approval and two after replaying the same confirmation. That is the exact attack this PR is named after, still open for that tool class.

## Tests

The two new tests do not constrain the helper against over-suppression, which is how both of the cases above slipped through. A version of `removeAlreadyCompletedTools` that clears the *whole* resume set whenever any response appears in the window passes every test in the repo — the PR's own case has a single pending tool, so "delete this id" and "clear everything" are indistinguishable, and `TestRequestConfirmationResumeSkipsAlreadyAnsweredCalls` places its marker after the user event, where the new window is empty and the helper never runs. A case with one tool still pending and an unrelated response in the window would pin the behavior you actually want.

Smaller things. The helper is worth a short comment explaining why the window starts at the confirmation request rather than at the beginning of the session, since that bound is doing real work and is easy to "simplify" later. In the new test file, [`m.calls++`](https://github.com/google/adk-go/blob/39b850b16fa72fc1564cc81f8feae74685788f4f/server/adkrest/hitl_confirmation_replay_test.go#L104) is unsynchronised on state the HTTP handler goroutine touches — harmless while the requests are strictly sequential, but it becomes a real race the moment anyone adds a concurrent case, and the recorder in the same file already takes a mutex.

One note for whoever merges this: the branch is behind `main`, though the file it touches has not moved since the merge-base, so the merge is clean.

karolpiotrowicz:
@kdroste-google flagging this one to you rather than to the wider group, since you have the most context on the confirmation flow.

It closes a real hole: on main, replaying a confirmation re-executes the tool and delivers the replayed payload. There is no upstream shape to copy — adk-python's dedupe covers the within-turn re-run, not cross-turn replay — so the semantics are ours to decide. The specific question the review raises is whether a rejection is terminal, or whether a user may legitimately re-answer a confirmation they have already answered. Today reject-then-approve silently does nothing.

huynhtrungcip:
Implemented and pushed in `ba2525f`. Confirmed approvals now acquire a durable event-log claim before dispatch, with atomic append-if-absent support for in-memory and database session services. Claims are persisted through Vertex raw events as well. Added regression coverage for concurrent duplicate approvals and confirmation-gated long-running replay. Targeted tests, vet, and full workspace compilation pass; please re-review.

karolpiotrowicz:
@wolo-lab this is a scope call rather than a code review. The underlying bug is real, but the fix adds public API to the `session` package and introduces a persistence concept that no other ADK implementation has, so the shape is yours to decide before anyone reviews the code.

Two new exported symbols land in `session`. [`EventAppender`](https://github.com/google/adk-go/blob/70dc87649ce4c33604f9a390f9940c1ebc833612/session/service.go#L79-L88) is an optional interface carrying `AppendEventIfAbsent`, which every third-party session service would need to implement to get the protection. [`ConfirmationClaim`](https://github.com/google/adk-go/blob/70dc87649ce4c33604f9a390f9940c1ebc833612/session/session.go#L283-L288) is a new struct, surfaced as a new field on `EventActions` at [session/session.go#L258-L261](https://github.com/google/adk-go/blob/70dc87649ce4c33604f9a390f9940c1ebc833612/session/session.go#L258-L261) and therefore part of the persisted event wire format.

adk-python has neither. Its session service API stops at `append_event` — [base_session_service.py#L166-L182](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/sessions/base_session_service.py#L166-L182) — with no append-if-absent, and `EventActions` carries no claim field. Taking this means ADK Go leads on a durable-claim design rather than following one, and the event schema it writes stops round-tripping through the Python implementation.

Being first is not by itself a reason to decline. It is a reason for you to decide whether we want this design and whether it should be proposed across implementations first, since public API and a persisted field are both hard to walk back.

The call I need from you: take it as-is, take it with the fix confined to internals so no new public API or persisted field is added, or decline it here and raise the claim design across the implementations first.

kdroste-google:
It seems to me that (at least partially) the functionality is similar to https://github.com/google/adk-go/pull/1369.

PR #1122, opened 2026-07-03
feat(agentregistry): add package foundation and REST transport

wolo-lab:
**Problem:**
adk-go has no client for the Google Cloud Agent Registry
(`agentregistry.googleapis.com`) — the governed catalog of A2A agents, MCP
servers, and model endpoints. adk-python already ships one, so we need a Go
equivalent with matching behavior to discover and construct agents/toolsets from
the registry.

**Solution:**
Add the `agentregistry` package **foundation** — everything the discovery
methods and the RemoteAgent/McpToolset factory helpers (follow-up PRs in this
stack) build on. This PR is additive: a new package, no changes to existing
public APIs.
- **Client surface:** `Config`, `Client`, and `New` — ADC-authenticated by
  default, or a caller-supplied `*http.Client`.
- **REST transport:** an authenticated GET path to `agentregistry.googleapis.com`
  that resolves resource paths (absolute `projects/...` vs. parent-relative),
  sets the `x-goog-user-project` quota header, and returns non-2xx responses as a
  typed `*APIError`. The transport sits behind an internal `requester` interface
  seam (default impl `restRequester`) so an alternative transport can back the
  same `Client`.
- **Endpoint & mTLS selection (single source of truth):** when `New` builds the
  client, endpoint and client-certificate selection are delegated to the Google
  API transport (`internaloption.WithDefaultEndpoint` /
  `WithDefaultMTLSEndpoint`). It honors `GOOGLE_API_USE_MTLS_ENDPOINT` and
  `GOOGLE_API_USE_CLIENT_CERTIFICATE` and, when a default client-certificate
  source is present, both targets the mTLS endpoint **and** configures mTLS on
  the transport from the same source — so endpoint and transport can never
  disagree (parity with adk-python / google-auth). A caller-supplied
  `*http.Client` owns its own transport/mTLS config; we only honor an explicit
  `GOOGLE_API_USE_MTLS_ENDPOINT=always` for it. (No programmatic mTLS knob — this
  is env-driven, like adk-python.)
- **Quota project:** `x-goog-user-project` follows adk-python's
  `credentials.quota_project_id or project_id`: prefer
  `GOOGLE_CLOUD_QUOTA_PROJECT`, then the ADC credentials' `quota_project_id`,
  then the configured `ProjectID` — resolved regardless of whether the HTTP
  client is caller-supplied.
- **Wire types & options:** `Agent`, `McpServer`, `Endpoint`, `Protocol`,
  `Interface`, `Skill`, `Card`, the `List{Agents,McpServers,Endpoints}Response`
  list responses, and `ListOption`s (`WithFilter` / `WithPageSize` /
  `WithPageToken`).
- **Resolution helpers (parity with adk-python):** `cleanName` (`_clean_name`),
  the protocol-binding → A2A transport map (`_TRANSPORT_MAPPING`, via
  `transportBinding`), and `connectionURI` (`_get_connection_uri`) — which maps
  each interface's registry binding to an A2A transport (e.g. `HTTP_JSON` →
  `HTTP+JSON`) and returns that mapped transport, matching Python.

### DIFF
--- a/agentregistry/registry.go
+++ b/agentregistry/registry.go
@@ -1,0 +1,139 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package agentregistry
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"strings"
+
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/option"
+	"google.golang.org/api/option/internaloption"
+	htransport "google.golang.org/api/transport"
+)
+
+// cloudPlatformScope is the OAuth scope used for Application Default Credentials.
+const cloudPlatformScope = "https://www.googleapis.com/auth/cloud-platform"
+
+// Base URLs for the Agent Registry API (parity with adk-python).
+const (
+	baseURLProd = "https://agentregistry.googleapis.com/v1"
+	baseURLMTLS = "https://agentregistry.mtls.googleapis.com/v1"
+)
+
+// Config configures a [Client].
+type Config struct {
+	// ProjectID is the Google Cloud project ID. Required.
+	ProjectID string
+	// Location is the Google Cloud location (region), e.g. "us-central1".
+	// Required.
+	Location string
+	// HTTPClient is used for registry API calls. If nil, an ADC-authenticated
+	// client is created and the endpoint (incl. mTLS) is resolved from
+	// GOOGLE_API_USE_MTLS_ENDPOINT / GOOGLE_API_USE_CLIENT_CERTIFICATE. A
+	// supplied client manages its own mTLS.
+	HTTPClient *http.Client
+}
+
+// Client is a client for the Google Cloud Agent Registry.
+type Client struct {
+	requester requester
+}
+
+// New creates a [Client]. By default it authenticates to
+// agentregistry.googleapis.com using Application Default Credentials; provide
+// Config.HTTPClient to supply a custom (e.g. pre-authenticated) client.
+func New(ctx context.Context, cfg Config) (*Client, error) {
+	if cfg.ProjectID == "" || cfg.Location == "" {
+		return nil, fmt.Errorf("agentregistry: ProjectID and Location must be set")
+	}
+
+	var creds *google.Credentials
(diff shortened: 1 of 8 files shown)


agentregistry/registry.go
```diff
@@ -0,0 +1,139 @@
...
+		httpClient = client
+		baseURL = endpoint
+	} else {
+		baseURL = customClientBaseURL()
+	}
+
+	// x-goog-user-project: quota project (env or ADC) if set, else ProjectID
+	// (matches adk-python). Resolved even for a caller-supplied client.
+	quotaProject := cfg.ProjectID
+	if qp := quotaProjectID(creds); qp != "" {
```
baptmont, line 98: if an HTTPClient is passed creds is always nil, so it will only check the env var value, is this intended?

agentregistry/registry.go
```diff
@@ -0,0 +1,139 @@
...
+func New(ctx context.Context, cfg Config) (*Client, error) {
+	if cfg.ProjectID == "" || cfg.Location == "" {
+		return nil, fmt.Errorf("agentregistry: ProjectID and Location must be set")
+	}
+
+	var creds *google.Credentials
+	var baseURL string
+	httpClient := cfg.HTTPClient
+	if httpClient == nil {
+		c, err := google.FindDefaultCredentials(ctx, cloudPlatformScope)
```
baptmont, line 71: This path is untested, all TestNew_Validation cases include a httpClient.
we could check htransport.NewHTTPClient and internaloption.WithDefaultEndpoint(baseURLProd) does not strip the /v1 from the baseURLProd

agentregistry/types.go
```diff
@@ -0,0 +1,141 @@
...
+	Name        string     `json:"name,omitempty"`
+	DisplayName string     `json:"displayName,omitempty"`
+	Description string     `json:"description,omitempty"`
+	Version     string     `json:"version,omitempty"`
+	Protocols   []Protocol `json:"protocols,omitempty"`
+	Skills      []Skill    `json:"skills,omitempty"`
+	Card        *Card      `json:"card,omitempty"`
+}
+
+// McpServer is a registered MCP server.
```
baptmont: nit: Mcp Should be in uppercase (https://g3doc.corp.google.com/go/g3doc/style/decisions.md?cl=head#initialisms)

baptmont (dismissed):

agentregistry/types.go
wolo-lab: True, changed.

agentregistry/registry.go
wolo-lab: Right, test added.

baptmont (approved):

agentregistry/registry.go
wolo-lab: Intended - but the old comment was not clear, I updated it.

For a caller-supplied *http.Client we deliberately don't load ADC, so there are no credentials to read quota_project_id from — an opaque *http.Client can't be introspected for it. In that case the quota project resolves from GOOGLE_CLOUD_QUOTA_PROJECT.

If a custom-client caller ever needs a quota project distinct from ProjectID without the process-wide env var, an explicit Config.QuotaProjectID is the clean fix — captured as a follow-up rather than growing this foundation PR.

PR #1124, opened 2026-07-03
feat(agentregistry): add discovery methods for agents, MCP servers, e…

wolo-lab:
**Problem:**
The agentregistry foundation (#1122) wires an authenticated Client and REST
transport but exposes no way to read the registry. Callers need to list and
fetch its resources — A2A agents, MCP servers, and model endpoints — with
behavior matching adk-python.

**Solution:**
Add the read/discovery surface on top of the client foundation, in `client.go`:
- **List / Get / All per resource** for agents, MCP servers, and model endpoints:
  - `ListAgents` / `ListMcpServers` / `ListEndpoints` — one page, honoring
    `WithFilter` / `WithPageSize` / `WithPageToken`.
  - `GetAgent` / `GetMcpServer` / `GetEndpoint` — a single resource by full name
    (used verbatim, not re-prefixed with the parent).
  - `AllAgents` / `AllMcpServers` / `AllEndpoints` — `iter.Seq2[*T, error]`
    auto-paging iterators that fetch pages on demand and stop on the first error.
- **`ModelName`** resolves an endpoint to a model resource name (mirrors
  adk-python's `get_model_name`: strips a trailing `:method` suffix and extracts
  an embedded `projects/...` name).
- **Shared primitives:** a generic `getResource[T]` (in `registry.go`) for the
  GET→decode path, and a generic `pages[T]` paginator (in `client.go`) backing
  the `All*` iterators.
Behavior mirrors adk-python's `agent_registry`.


agentregistry/client_test.go
```diff
@@ -0,0 +1,460 @@
...
+			name: "endpoints single page",
+			handler: func(w http.ResponseWriter, r *http.Request) {
+				_, _ = w.Write([]byte(`{"endpoints":[{"displayName":"gemini"}]}`))
+			},
+			all: func(ctx context.Context, c *Client) iter.Seq2[string, error] {
+				return nameSeq(c.AllEndpoints(ctx), func(e *Endpoint) string { return e.DisplayName })
+			},
+			wantNames: []string{"gemini"},
+		},
+	}
```
baptmont, line 411: we could add a test to check if opts is kept across pages. 

```go
{
    // The paginator must forward opts (filter) on *every* page, not just the
    // first. The handler 500s any request missing the filter, so if the token
    // page dropped it, page 2 would error instead of returning "a2".
    name: "filter is forwarded on every page",
    handler: func(w http.ResponseWriter, r *http.Request) {
        if r.URL.Query().Get("filter") != "type=A2A" {
            w.WriteHeader(http.StatusInternalServerError)
            _, _ = w.Write([]byte(`{"error":"missing filter"}`))
            return
        }
        switch r.URL.Query().Get("pageToken") {
        case "":
            _, _ = w.Write([]byte(`{"agents":[{"displayName":"a1"}],"nextPageToken":"tok2"}`))
        case "tok2":
            _, _ = w.Write([]byte(`{"agents":[{"displayName":"a2"}]}`))
        default:
            w.WriteHeader(http.StatusInternalServerError)
        }
    },
    all: func(ctx context.Context, c *Client) iter.Seq2[string, error] {
        return nameSeq(c.AllAgents(ctx, WithFilter("type=A2A")), func(a *Agent) string { return a.DisplayName })
    },
    wantNames:    []string{"a1", "a2"},
    wantRequests: 2,
},
```

agentregistry/client.go
```diff
@@ -0,0 +1,186 @@
...
+			return nil, "", err
+		}
+		return page.Endpoints, page.NextPageToken, nil
+	})
+}
+
+// ModelName resolves an endpoint into a model resource name (e.g.
+// "projects/<p>/locations/<l>/publishers/google/models/<model>"). endpointName
+// is the full endpoint resource name.
+func (c *Client) ModelName(ctx context.Context, endpointName string) (string, error) {
```
baptmont: nit: Change it to GetModelName since its a remote call and not an accessor

wolo-lab: Valid point, changed.

agentregistry/client_test.go
wolo-lab: Valid as well. I added your suggestion.

agentregistry/client.go
```diff
@@ -0,0 +1,186 @@
...
+			return nil, "", err
+		}
+		return page.Endpoints, page.NextPageToken, nil
+	})
+}
+
+// GetModelName resolves an endpoint into a model resource name (e.g.
+// "projects/<p>/locations/<l>/publishers/google/models/<model>"). endpointName
+// is the full endpoint resource name.
+func (c *Client) GetModelName(ctx context.Context, endpointName string) (string, error) {
```
kdroste-google: Not clear what's the reason for this, let's remove it for the time being until someone needs this functionality :)

agentregistry/client.go
```diff
@@ -0,0 +1,186 @@
...
+		return "", err
+	}
+	uri, _, _, ok := connectionURI(nil, endpoint.Interfaces, "", "")
+	if !ok {
+		return "", fmt.Errorf("agentregistry: connection URI not found for endpoint %q", endpointName)
+	}
+	return parseModelName(uri), nil
+}
+
+var (
```
kdroste-google: Needed by GetModelName, may be removed

agentregistry/client.go
```diff
@@ -0,0 +1,186 @@
...
+	// reModelSuffix strips a trailing ":method" suffix (e.g. ":predict").
+	reModelSuffix = regexp.MustCompile(`:\w+$`)
+	// reProjectsSubstring extracts a "projects/..." resource name embedded in a URL.
+	reProjectsSubstring = regexp.MustCompile(`projects/.+`)
+)
+
+// parseModelName extracts a model resource name from an endpoint URI, mirroring
+// the parsing in adk-python's get_model_name.
+func parseModelName(uri string) string {
```
kdroste-google: Needed by GetModelName, may be removed

kdroste-google (changes requested):

agentregistry/client.go
wolo-lab: Removed.

agentregistry/client.go
wolo-lab: Removed.

agentregistry/client.go
wolo-lab: Removed.

kdroste-google (approved):
LGTM

PR #1127, opened 2026-07-06
feat(agentregistry): add RemoteAgent and McpToolset factory helpers

wolo-lab:
**Problem:**
The `agentregistry` package can already _discover_ registry resources (list/get
agents, MCP servers, endpoints), but the returned metadata isn't usable at
runtime as-is. Callers still have to resolve connection endpoints, parse or
synthesize A2A agent cards, reconcile A2A protocol-version compatibility, and
wire up egress auth before a registered agent or MCP server can be plugged into
ADK.

adk-python already exposes this via `AgentRegistry.get_remote_a2a_agent`
and `get_mcp_toolset`; the Go client had no equivalent.

**Solution:**
Add two factory helpers on `Client` that turn registry metadata into
ready-to-use ADK objects (parity with adk-python 2.3.0
`integrations/agent_registry`):
- `Client.RemoteAgent(ctx, name, ...RemoteAgentOption) (agent.Agent, error)` —
  resolves a registered A2A agent into an `agent.Agent` via
  `remoteagent.NewA2A`. Uses the embedded `A2A_AGENT_CARD` when present,
  otherwise synthesizes a card from discrete fields (cleaned display name,
  default HTTP+JSON binding, default protocol version `0.3.0`, text I/O modes).
  It registers a2a **compatibility transports** for each (binding,
  protocolVersion) the card advertises: Agent Registry commonly reports a
  different-major protocol version (e.g. `0.3.0`) than the a2a-go SDK's current
  `1.0`, and the SDK's same-major fallback can't bridge it — without a matching
  compat transport, client creation fails with "no compatible transports found".
- `Client.McpToolset(ctx, name, ...McpToolsetOption) (tool.Toolset, error)` —
  resolves a registered MCP server into a `tool.Toolset` via `mcptoolset.New`
  over a streamable-HTTP transport. Endpoint resolution prefers the JSONRPC
  binding, falling back to HTTP_JSON.
Egress model (mirrors adk-python's `_is_google_api`):
- `McpToolset` auto-authenticates `*.googleapis.com` endpoints with the
  registry's ADC client; `RemoteAgent` does **not** auto-authenticate (egress is
  left to the caller).
- Both accept caller-supplied HTTP clients and static headers via options:
  `WithA2AHTTPClient`/`WithA2AHeaders`, `WithMcpHTTPClient`/`WithMcpHeaders`.
  Headers are layered onto a cloned client so shared clients are never mutated.
Deferred (out of scope, to be tracked separately): bindings-based
`GcpAuthProvider` resolution, the MCP tool-name prefix, and the
`gcp.mcp.server.destination.id` telemetry key — they depend on the auth
subsystem and on tool-level metadata support respectively.

### DIFF
--- a/agentregistry/factory.go
+++ b/agentregistry/factory.go
@@ -54,6 +54,18 @@ func applyMCPToolsetOptions(opts []MCPToolsetOption) egressConfig {
 	return ec
 }
 
+func addHeaders(e *egressConfig, h map[string]string) {
+	if len(h) == 0 {
+		return
+	}
+	if e.headers == nil {
+		e.headers = make(map[string]string, len(h))
+	}
+	for k, v := range h {
+		e.headers[k] = v
+	}
+}
+
 // RemoteAgentOption customizes [Client.RemoteAgent].
 type RemoteAgentOption func(*egressConfig)
 
@@ -70,9 +82,9 @@ func WithA2AHTTPClient(c *http.Client) RemoteAgentOption {
 }
 
 // WithA2AHeaders adds static headers to every request sent to the remote A2A
-// agent.
+// agent. Repeated calls accumulate; a later value wins on a key conflict.
 func WithA2AHeaders(h map[string]string) RemoteAgentOption {
-	return func(e *egressConfig) { e.headers = h }
+	return func(e *egressConfig) { addHeaders(e, h) }
 }
 
 // WithMCPHTTPClient sets the HTTP client used to reach the MCP server. It
@@ -86,8 +98,9 @@ func WithMCPHTTPClient(c *http.Client) MCPToolsetOption {
 }
 
 // WithMCPHeaders adds static headers to every request sent to the MCP server.
+// Repeated calls accumulate; a later value wins on a key conflict.
 func WithMCPHeaders(h map[string]string) MCPToolsetOption {
-	return func(e *egressConfig) { e.headers = h }
+	return func(e *egressConfig) { addHeaders(e, h) }
 }
 
 // egressClient selects the HTTP client used to reach an endpoint at rawURL.
@@ -167,7 +180,7 @@ func (c *Client) RemoteAgent(ctx context.Context, name string, opts ...RemoteAge
 		return nil, err
 	}
 
-	card, agentName, description, err := agentCard(info, name)
+	card, agentName, err := agentCard(info, name)
 	if err != nil {
 		return nil, err
 	}
@@ -179,10 +192,15 @@ func (c *Client) RemoteAgent(ctx context.Context, name string, opts ...RemoteAge
 
 	return remoteagent.NewA2A(remoteagent.A2AConfig{
 		Name:           agentName,
-		Description:    description,
+		Description:    card.Description,
 		AgentCard:      card,
 		ClientProvider: remoteagent.NewA2AClientProvider(a2aClientFactory(card, egress)),
 	})
+}
+
+type transportKey struct {
+	binding a2a.TransportProtocol
+	version a2a.ProtocolVersion
 }
 
 // a2aClientFactory builds an A2A client factory for the given card and HTTP
@@ -197,7 +215,7 @@ func a2aClientFactory(card *a2a.AgentCard, httpClient *http.Client) *a2aclient.F
 		a2aclient.WithJSONRPCTransport(httpClient),
 		a2aclient.WithRESTTransport(httpClient),
 	}
-	seen := make(map[string]bool)
+	seen := make(map[transportKey]bool)
 	for _, iface := range card.SupportedInterfaces {
 		// The current-version transports above already cover a2a.Version.
 		if iface == nil || iface.ProtocolVersion == "" || iface.ProtocolVersion == a2a.Version {
@@ -212,7 +230,7 @@ func a2aClientFactory(card *a2a.AgentCard, httpClient *http.Client) *a2aclient.F
 		default:
 			continue
 		}
-		key := string(iface.ProtocolBinding) + "@" + string(iface.ProtocolVersion)
+		key := transportKey{iface.ProtocolBinding, iface.ProtocolVersion}
 		if seen[key] {
 			continue
 		}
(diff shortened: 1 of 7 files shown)


agentregistry/factory.go
```diff
@@ -0,0 +1,355 @@
...
+// WithA2AHTTPClient sets the HTTP client used to reach the remote A2A agent.
+// A2A egress is not auto-authenticated, so set this to authenticate requests to
+// the remote agent. The default ([http.DefaultClient]) has no timeout; bound
+// egress on the client's Transport rather than via http.Client.Timeout, which
+// is a deadline over the whole request and would truncate streaming responses.
+func WithA2AHTTPClient(c *http.Client) RemoteAgentOption {
+	return func(e *egressConfig) { e.httpClient = c }
+}
+
+// WithA2AHeaders adds static headers to every request sent to the remote A2A
```
kdroste-google: it doesn't add headers, it overrides the entire set

agentregistry/factory.go
```diff
@@ -0,0 +1,355 @@
...
+// overrides the default (an Application Default Credentials client for
+// *.googleapis.com endpoints, else [http.DefaultClient]). The default has no
+// timeout; bound egress on the client's Transport rather than via
+// http.Client.Timeout, which is a deadline over the whole request and would
+// truncate streaming responses.
+func WithMCPHTTPClient(c *http.Client) MCPToolsetOption {
+	return func(e *egressConfig) { e.httpClient = c }
+}
+
+// WithMCPHeaders adds static headers to every request sent to the MCP server.
```
kdroste-google: it doesn't add headers, it overrides the entire set

agentregistry/factory.go
```diff
@@ -0,0 +1,355 @@
...
+// egressClient selects the HTTP client used to reach an endpoint at rawURL.
+// Precedence: an explicit override, then (only when autoADC is set and the
+// endpoint is a Google API) the registry's authenticated client, then a default
+// client. Static headers, if any, are layered on via a cloned client so shared
+// clients are never mutated.
+func (c *Client) egressClient(rawURL string, ec egressConfig, autoADC bool) *http.Client {
+	base := ec.httpClient
+	if base == nil {
+		if autoADC && isGoogleAPI(rawURL) {
+			base = c.httpClient
```
kdroste-google, line 116: It will probably fail for agents deployed to agent engine / cloud run - can be fixed later

agentregistry/factory.go
```diff
@@ -0,0 +1,355 @@
...
+		var factory a2aclient.TransportFactory
+		switch iface.ProtocolBinding {
+		case a2a.TransportProtocolJSONRPC:
+			factory = jsonrpcTransportFactory(httpClient)
+		case a2a.TransportProtocolHTTPJSON:
+			factory = restTransportFactory(httpClient)
+		default:
+			continue
+		}
+		key := string(iface.ProtocolBinding) + "@" + string(iface.ProtocolVersion)
```
kdroste-google: we are assuming here that no edge cases with "" and "@" can appear as ProtocolBinding and ProtocolVersion, right?

agentregistry/factory.go
```diff
@@ -0,0 +1,355 @@
...
+			return nil, "", "", fmt.Errorf("agentregistry: decoding embedded agent card for %q: %w", resourceName, err)
+		}
+		// Fail fast: an interface-less card would otherwise only error at the
+		// first invocation, deep in the a2a client, without the resource name.
+		if len(embedded.SupportedInterfaces) == 0 {
+			return nil, "", "", fmt.Errorf("agentregistry: embedded agent card for %q has no supported interfaces", resourceName)
+		}
+		agentName := cleanName(embedded.Name)
+		if agentName == "" {
+			agentName = cleanName(resourceName)
```
kdroste-google: We assume that resourceName is not "" here. Or agentName can be empty?

agentregistry/factory.go
```diff
@@ -0,0 +1,355 @@
...
+		Name:        agentName,
+		Description: info.Description,
+		Version:     info.Version,
+		SupportedInterfaces: []*a2a.AgentInterface{{
+			URL:             url,
+			ProtocolBinding: transport,
+			ProtocolVersion: a2a.ProtocolVersion(protocolVersion),
+		}},
+		Skills:             toA2ASkills(info.Skills),
+		Capabilities:       a2a.AgentCapabilities{Streaming: false},
```
kdroste-google: Streaming defaults to false. Do we need it here?

agentregistry/factory.go
```diff
@@ -0,0 +1,355 @@
...
+func jsonrpcTransportFactory(httpClient *http.Client) a2aclient.TransportFactory {
+	return a2aclient.TransportFactoryFn(func(_ context.Context, _ *a2a.AgentCard, iface *a2a.AgentInterface) (a2aclient.Transport, error) {
+		return a2aclient.NewJSONRPCTransport(iface.URL, httpClient), nil
+	})
+}
+
+// agentCard resolves an [a2a.AgentCard] plus the cleaned agent name and
+// description for a registered agent, using the embedded card when available and
+// otherwise synthesizing one. resourceName is used as a fallback name.
+func agentCard(info *Agent, resourceName string) (card *a2a.AgentCard, name, description string, err error) {
```
kdroste-google: Why do we need to return the description? It seems to be equal to card.Descrption for non-error paths

kdroste-google (changes requested):
Some minor things - listed in comments

agentregistry/factory.go
wolo-lab: True, updated

agentregistry/factory.go
wolo-lab: True, updated

agentregistry/factory.go
wolo-lab: No, removed.

agentregistry/factory.go
wolo-lab: Correct, it was redundant.

agentregistry/factory.go
wolo-lab: Agreed. Let's have it as a follow-up.

agentregistry/factory.go
wolo-lab: Good point. Added a check with an error.

agentregistry/factory.go
wolo-lab: Right, string concatenation is fragile. Let's have a struct key, which removes the ambiguity entirely.

agentregistry/factory.go
```diff
@@ -0,0 +1,375 @@
...
+
+func applyMCPToolsetOptions(opts []MCPToolsetOption) egressConfig {
+	var ec egressConfig
+	for _, opt := range opts {
+		opt(&ec)
+	}
+	return ec
+}
+
+func addHeaders(e *egressConfig, h map[string]string) {
```
kdroste-google, line 61: it is actually overwriting headers (adds new, overwrites existing, deletes nothing)

agentregistry/factory.go
```diff
@@ -0,0 +1,375 @@
...
+func isGoogleAPI(rawURL string) bool {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return false
+	}
+	host := u.Hostname()
+	return host == "googleapis.com" || strings.HasSuffix(host, ".googleapis.com")
+}
+
+// clientWithHeaders returns a client that adds the given static headers to every
```
kdroste-google: adds or overwrites

agentregistry/factory.go
```diff
@@ -0,0 +1,375 @@
...
+// request. If there are no headers, base is returned unchanged. Otherwise base
+// is shallow-copied so the caller's client is not mutated.
+func clientWithHeaders(base *http.Client, headers map[string]string) *http.Client {
+	if len(headers) == 0 {
+		return base
+	}
+	clone := *base
+	clone.Transport = &headerRoundTripper{base: base.Transport, headers: headers}
```
kdroste-google, line 143: To be considered: adding explicit 
```go
var _ http.RoundTripper = (*headerRoundTripper)(nil)
```

agentregistry/factory.go
```diff
@@ -0,0 +1,375 @@
...
+}
+
+func addHeaders(e *egressConfig, h map[string]string) {
+	if len(h) == 0 {
+		return
+	}
+	if e.headers == nil {
+		e.headers = make(map[string]string, len(h))
+	}
+	for k, v := range h {
```
kdroste-google: consider using `map.Copy(e.headers, h)`

agentregistry/factory.go
```diff
@@ -0,0 +1,375 @@
...
+func WithMCPHeaders(h map[string]string) MCPToolsetOption {
+	return func(e *egressConfig) { addHeaders(e, h) }
+}
+
+// egressClient selects the HTTP client used to reach an endpoint at rawURL.
+// Precedence: an explicit override, then (only when autoADC is set and the
+// endpoint is a Google API) the registry's authenticated client, then a default
+// client. Static headers, if any, are layered on via a cloned client so shared
+// clients are never mutated.
+func (c *Client) egressClient(rawURL string, ec egressConfig, autoADC bool) *http.Client {
```
kdroste-google: maybe it would be better to have an explicit dedicated clients for google api calls and everything else. Then such a config could have the same client for both, or different clients. WDYT?

agentregistry/factory.go
```diff
@@ -0,0 +1,375 @@
...
+
+	card, agentName, err := agentCard(info, name)
+	if err != nil {
+		return nil, err
+	}
+
+	ec := applyRemoteAgentOptions(opts)
+	// A2A egress is not auto-authenticated (parity with adk-python), so the
+	// endpoint URL is irrelevant to client selection here.
+	egress := c.egressClient("", ec, false)
```
kdroste-google: To be considered: may be simplified, passing empty url and false for autoADC effectively only applies the override from ec.httpClient or uses defaultClient  + headers are applied.

kdroste-google (changes requested):

agentregistry/factory.go
wolo-lab: right, it's upsert. Added a documentation

agentregistry/factory.go
wolo-lab: Updated.

agentregistry/factory.go
wolo-lab: Added.

agentregistry/factory.go
wolo-lab: Good catch!

agentregistry/factory.go
wolo-lab: nice, done.

agentregistry/factory.go
wolo-lab: I'm not sure. Each factory call resolves exactly one endpoint, so there will be always one client which is unnecessary (we don't share it with different endpoints). I believe it would be a bit cleaner with just one.

Adding it later would be non-breaking change, so we can go back to this approach once we have a real use case for communicating with different endpoints.

kdroste-google: SG

kdroste-google (approved):
LGTM, thank you!

PR #1129, opened 2026-07-07
fix(runner): preserve nested workflow HITL resume responses

### LINKED ISSUE #1125: Runner.Run  fails to resume nested  WorkflowAgent  after  ResumeOrRequestInput  interrupt
## 🔴 Required Information

**Describe the Bug:**
When a `WorkflowAgent` is executed as a dynamically resolved sub-agent (delegated from a parent `LlmAgent`), it fails to correctly resume after pausing for Human-In-The-Loop (HITL) input via `ctx.ResumeOrRequestInput()`.

Instead of resuming the paused node upon receiving the user's `FunctionResponse`, the ADK `Runner` incorrectly passes the *original* user prompt (e.g., the message that initially started the workflow) back down to the nested `WorkflowAgent`. Consequently, the internal `detectResume` method fails to find the `FunctionResponse` in `ctx.UserContent()`, incorrectly assumes it is a brand-new turn, and erroneously restarts the entire sub-agent workflow from the beginning.

**Steps to Reproduce:**
1. Create a `WorkflowAgent` containing a node that pauses execution using `workflow.ResumeOrRequestInput(ctx, emit, req)`.
2. Register this `WorkflowAgent` as a sub-agent of a parent `LlmAgent`.
3. Use the `runner.Runner.Run()` API to start a conversation with the parent `LlmAgent`. The parent agent transfers execution to the sub-agent.
4. Wait for the `WorkflowAgent` to pause and emit the `adk_request_input` tool call.
5. Provide a valid user response containing the corresponding `FunctionResponse` back into `Runner.Run()` to continue the session.

**Expected Behavior:**
The ADK runner should route the user's new message (containing the `FunctionResponse`) into the nested `WorkflowAgent`, allowing its `detectResume()` method to match the interrupt ID and seamlessly resume the workflow from the paused node.

**Observed Behavior:**
The `WorkflowAgent` restarts from the very first node (e.g. `START`) instead of resuming. 

Under the hood, `Runner.Run()` dynamically wraps the sub-agent in a single-node workflow (`START -> subagent`). During the resume phase, `workflow.Resume()` forgets to populate `s.workflowInput` with the new user message. When evaluating `s.predecessorInput(node)` for the subagent, it falls back to reconstructing the `START` node's output from the *previous* run state (the original trigger message). The sub-agent receives this old message, fails to detect the resume, and restarts entirely.

**Environment Details:**

 - ADK Library Version: v2.0.0 (`google.golang.org/adk/v2@v2.0.0`)
 - OS: Linux
 - Go Version: 1.26.3

**Model Information:**
 - Which model is being used: N/A (Engine bug, applies to all models)

---

## 🟡 Optional Information
*Providing this information greatly speeds up the resolution process.*

**Regression:**
N/A

**Logs:**
N/A

**Screenshots / Video:**
N/A

**Additional Context:**
The bug specifically lives in `adk/v2/workflow/resume.go`. The signature for `Resume` is:
```go
func (w *Workflow) Resume(ctx agent.Context, state *RunState, responses map[string]any) iter.Seq2[*session.Event, error]
```
Unlike `Run`, `Resume` does not accept a `workflowInput` parameter. Because `s.workflowInput` is never initialized inside `Resume()`, any node succeeding `START` (like the dynamically wrapped sub-agent in `Runner.runNode`) incorrectly receives the `START` node's persisted output from the previous iteration, effectively stripping the current turn's `UserContent()` and breaking `WorkflowAgent.detectResume()`.

```go
// 1. Create a WorkflowAgent that asks a question
askUserNode := workflow.NewDynamicNode("ask", func(ctx agent.Context, in any, emit func(*session.Event) error) (any, error) {
    var userResp string
    // Pauses and waits for user input
    respAny, err := workflow.ResumeOrRequestInput(ctx, emit, session.RequestInput{
        InterruptID: "ask_desc",
        Message:     "Please provide a description.",
    })
    if respAny != nil {
        userResp = respAny.(string)
    }
    return userResp, err
}, workflow.NodeConfig{})

eb := workflow.NewEdgeBuilder()
eb.Add(workflow.Start, askUserNode)
wf, _ := workflow.New("subagent_wf", eb.Build())

subAgent, _ := workflowagent.New(workflowagent.Config{
    Name:     "subagent",
    Workflow: wf, // Or Edges depending on version
})

// 2. Wrap it inside an LlmAgent (Orchestrator)
parentAgent, _ := llmagent.New(llmagent.Config{
    Name:      "orchestrator",
    SubAgents: []agent.Agent{subAgent},
})

// 3. Trigger it via Runner.Run
// - First turn: Send "ask the subagent something"
// - Wait for adk_request_input
// - Second turn: Send the FunctionResponse to Runner.Run()
// - OBSERVE: SubAgent restarts the "ask" node instead of completing it.
```

 - Always (100%)

chaojixinren:

- Closes: #1125

**Problem:**
An LlmAgent transfer to a WorkflowAgent resumes through the runner's dynamic
agent-node wrapper. The generic wrapper dropped UserContent on resume, so
WorkflowAgent.detectResume could not see the current FunctionResponse and
treated the turn as a fresh run from START.

**Solution:**
Detect resume from the current turn's FunctionResponse and pass only the
matching response parts to generic agents on resume. This keeps the original
user input from being reprocessed while preserving the resume payload workflow
agents need.

This also adds a runner integration test covering LlmAgent -> WorkflowAgent
transfer with RequestInput resume.

### Testing Plan

**Unit Tests:**

Passed locally:

- `go test ./runner -run TestRunner_NestedWorkflowAgentUnderLlmAgent_Resume -count=1`
- `go test ./runner -count=1`
- `go test ./agent/workflowagent ./workflow ./runner -count=1`
- `go test ./... -count=1`

No separate UI/manual E2E was needed for this runner-level fix. The added runner
integration test exercises the full in-memory flow:

1. A scripted LlmAgent emits `transfer_to_agent` for a WorkflowAgent sub-agent.
2. The WorkflowAgent pauses with `RequestInput`.
3. The follow-up turn sends a matching FunctionResponse.
4. The test verifies the sub-agent does not restart with a fresh interrupt and
   the handler receives the resume payload.

### Checklist

### Additional context

Fixes #1125.


AmirF194 (commented):
I traced and reproduced this on #1125, so I checked the fix against that repro. It lands on the right seam: the generic agent-node wrapper was detecting resume from session history (`answeredOpenInterrupts`) and then handing the sub-agent no `UserContent`, so `WorkflowAgent.detectResume` saw no FunctionResponse and restarted from START. Keying detection off the current turn's FunctionResponse and forwarding only the matching parts (`resumeUserContent`) fixes exactly that.

Verified in a clean container (golang:1.25, main `0c88126` vs `c177298`): `TestRunner_NestedWorkflowAgentUnderLlmAgent_Resume` fails on main (`sub-agent restarted with fresh interrupt "sub_approval" on turn 2`, `handler input = <nil>`) and passes on the branch. Full `runner` package is green, including the existing HITL tests.

One optional note, not blocking: moving from "any long-running interrupt answered anywhere in history" to "the current turn carries the answering FunctionResponse" also closes a second gap, since a later fresh turn in the same session would previously be classified as a resume and have its input dropped. That's the more behaviorally load-bearing half of the change, and the added test only exercises the nested-restart path. A dedicated case for "a fresh turn after a completed resume still delivers input to the sub-agent" would lock it in. (I couldn't isolate that path cleanly through the LlmAgent transfer in my own harness, so I'm flagging it as a coverage gap, not a defect.)

Rebase will clear the BEHIND state.

hanorik:
Hi, thanks for taking this! Fix looks right — the one thing I'd ask for: tests that constrain the logic

Mutation-testing the changed code (mutate, `go build ./... && go test`):

| Mutation | Result |
|---|---|
| `resumeUserContent` -> `return userContent` (no filtering at all) | **survives** the full suite |
| `answeredInterrupts` falls back to scanning history | **survives** the full suite |
| any `FunctionResponse` with a non-empty ID counts as a resume | **survives** `./runner/` |
| (control) `resumeUserContent` -> `nil`, i.e. revert the fix | killed by the new test |

So the new test pins the *direction* of the fix but not its logic — and mutation 2 re-introduces
exactly the second bug the commit message calls out ("later fresh messages in the same session still
get delivered as input"), with the suite still green. This matches the coverage gap @AmirF194 raised.

Table-driven unit tests for the two helpers would close it — for `answeredInterrupts`: already-answered
ID, an ID that was never long-running, nil session, nil `UserContent`, empty ID, duplicate IDs; for
`resumeUserContent`: text dropped, unmatched FR dropped, `FunctionCall` dropped. Those kill all four.

### Nits

- `answeredInterrupts` doc (`agent_node.go:170`) says "answered by the current user turn" — it returns
  IDs the turn *mentions*, open or not. And `:69-72` ("later fresh messages … get delivered as input")
  holds only for *text* messages on the generic path; a non-text fresh turn still arrives as `nil`
  because the fresh branch uses `inputToUserContent(input)`. Pre-existing, but the comment overstates it.
- `runGenericAgentBody` takes both `isResume` and `resolved`, where `isResume == len(resolved) > 0`.
- `agent_node.go:205` and `:217` are unreachable: `resumeUserContent` is only called when
  `len(resolved) > 0`, and every key in `resolved` came from a part of the same `userContent`.

Please also rebase — 39 commits behind, though the cherry-pick applies cleanly.

chaojixinren:
> Hi, thanks for taking this! Fix looks right — the one thing I'd ask for: tests that constrain the logic嗨，感谢您接受这个建议！代码看起来没问题。我唯一希望的是：能有一些测试来确保逻辑的准确性。
> 
> Mutation-testing the changed code (mutate, `go build ./... && go test`):对修改后的代码进行变异测试（变异操作， `go build ./... && go test` ）：
> 
> Mutation  突变	Result  结果
> `resumeUserContent` -> `return userContent` (no filtering at all)
> `resumeUserContent` -> `return userContent` （完全不进行过滤）	**survives** the full suite  能够经受住所有考验/能够成功度过所有挑战
> `answeredInterrupts` falls back to scanning history
> `answeredInterrupts` 恢复为扫描历史记录模式	**survives** the full suite  能够经受住所有考验/能够成功度过所有挑战
> any `FunctionResponse` with a non-empty ID counts as a resume
> 任何具有非空 ID 的 `FunctionResponse` 都算作一次简历提交。	**survives** `./runner/`  在 `./runner/` 的攻击下仍然存活
> (control) `resumeUserContent` -> `nil`, i.e. revert the fix
> （控制） `resumeUserContent` -> `nil` ，即撤销之前的修改。	killed by the new test
> 被新的测试给“弄死了”/因为新的测试而失败了
> So the new test pins the _direction_ of the fix but not its logic — and mutation 2 re-introduces因此，新的测试方法确定了修复的方向，但无法确定其具体的逻辑实现方式——而第 2 种变异则又重新引入了原来的逻辑结构。 exactly the second bug the commit message calls out ("later fresh messages in the same session still正是提交信息中提到的第二个错误：“同一会话中稍后发送的新消息仍然……” get delivered as input"), with the suite still green. This matches the coverage gap @AmirF194 raised.以“作为输入值被传递进来”的方式处理，此时该套件的状态仍为“绿色”。这与之前指出的覆盖范围缺口相符。
> 
> Table-driven unit tests for the two helpers would close it — for `answeredInterrupts`: already-answered针对这两个辅助函数的基于表格的单元测试已经完成——关于 `answeredInterrupts` 的问题：已有解答。 ID, an ID that was never long-running, nil session, nil `UserContent`, empty ID, duplicate IDs; forID：从来都不长的 ID、无效的会话信息、空值 `UserContent` 、无效的 ID、重复的 ID。 `resumeUserContent`: text dropped, unmatched FR dropped, `FunctionCall` dropped. Those kill all four. `resumeUserContent` ：文本被丢弃，不匹配的 FR 也被丢弃了， `FunctionCall` 也被丢弃了。这样，四个都被处理掉了。
> 
> ### Nits
> * `answeredInterrupts` doc (`agent_node.go:170`) says "answered by the current user turn" — it returns `answeredInterrupts` doc（ `agent_node.go:170` ）显示“由当前用户回答”——系统会返回该结果。
>   IDs the turn _mentions_, open or not. And `:69-72` ("later fresh messages … get delivered as input")标识出该轮次中提到的各项内容，无论这些内容是否已被处理。另外还有 `:69-72` （“稍后的新消息……会作为输入内容被处理”）。
>   holds only for _text_ messages on the generic path; a non-text fresh turn still arrives as `nil`该规则仅适用于通过常规路径发送的文本消息；而那些非文本格式的消息，仍然会以 `nil` 的形式被发送出来。
>   because the fresh branch uses `inputToUserContent(input)`. Pre-existing, but the comment overstates it.因为新分支使用了 `inputToUserContent(input)` 。这个功能其实早就存在了，但注释中的描述有些夸张了。
> * `runGenericAgentBody` takes both `isResume` and `resolved`, where `isResume == len(resolved) > 0`. `runGenericAgentBody` 同时包含 `isResume` 和 `resolved` ，其中 `isResume == len(resolved) > 0` 也是如此。
> * `agent_node.go:205` and `:217` are unreachable: `resumeUserContent` is only called when `agent_node.go:205` 和 `:217` 无法被访问： `resumeUserContent` 仅在以下情况下会被调用：
>   `len(resolved) > 0`, and every key in `resolved` came from a part of the same `userContent`. `len(resolved) > 0` ，而 `resolved` 中的每一个键都源自同一个 `userContent` 的某个部分。
> 
> Please also rebase — 39 commits behind, though the cherry-pick applies cleanly.请也进行重新提交吧——虽然通过“cherry-pick”方式可以顺利应用更改，但目前还有 39 次提交需要处理。

Thanks for the thorough mutation-testing review — that was incredibly helpful in pinpointing exactly what the coverage gaps were.

I've addressed all the feedback:

**Table-driven unit tests** added in `runner/agent_node_test.go`:
- `TestAnsweredInterrupts` — 11 cases: nil session, nil user content, answers the open interrupt, fresh text turn after a completed resume, response to a call that was never long-running, no long-running call in history at all, empty response ID, already-answered ID answered again, duplicate IDs collapse, only the long-running subset is reported, parallel interrupts answered together.
- `TestResumeUserContent` — 9 cases: drops text parts, drops FunctionResponses outside resolved, drops FunctionCall parts, keeps every resolved response, preserves role, nil user content, empty resolved forwards nothing, nothing matches, skips nil parts.

These kill all four mutations you identified.

**Nits fixed:**
- `answeredInterrupts` doc: clarified it returns "the subset of long-running interrupt IDs that the current user turn's FunctionResponses reference" rather than implying the turn "answers" them.
- Comment at L68–72: narrowed to "so that a fresh text turn following a completed resume is not misclassified" — no longer overstates the scope.
- Removed the redundant `isResume` param from `runGenericAgentBody`; the body now computes it from `len(resolved) > 0`.
- Removed the unreachable `len(resolved) == 0` guard in `resumeUserContent`; documented the caller contract instead. The `len(parts) == 0` path is kept as defense-in-depth against data inconsistency (it's reachable if resolved IDs don't match any part in the content).

**Rebased** onto latest main.

With these unit tests in place the mutation-testing results should now be all-killed.

hanorik (approved):

PR #1130, opened 2026-07-07
fix: update confirmation wrapper to propagate ProcessRequest calls to underlying tools and intercept packing registration

hanorik:
Fixes an issue where `tool.WithConfirmation` (`confirmationTool`) shadowed custom `ProcessRequest` methods on wrapped tools.

## Changes

- **Forward `ProcessRequest`**: Updated `confirmationTool.ProcessRequest` in `tool/tool.go` to delegate `ProcessRequest` calls to the underlying tool if it implements the method.
- **Preserve Confirmation Wrapper**: If the wrapped tool registers itself in `req.Tools` during `ProcessRequest`, the wrapper replaces that entry with itself so confirmation checks still execute on tool runs.
- **Tests**: Added unit tests (`TestWithConfirmation_ProcessRequest` and `TestWithConfirmation_ProcessRequest_Packing`) in `tool/tool_test.go` verifying delegation and registration wrapping.

## Testing

- `go test -race ./tool/...`

### DIFF
--- a/tool/tool.go
+++ b/tool/tool.go
@@ -185,16 +185,14 @@ type runnableTool interface {
 	Run(ctx agent.Context, args any) (result map[string]any, err error)
 }
 
-type requestProcessor interface {
-	ProcessRequest(ctx agent.Context, req *model.LLMRequest) error
-}
-
 func (t *confirmationTool) Declaration() *genai.FunctionDeclaration {
 	return t.runnableTool.Declaration()
 }
 
 func (t *confirmationTool) ProcessRequest(ctx agent.Context, req *model.LLMRequest) error {
-	if rp, ok := t.runnableTool.(requestProcessor); ok {
+	if rp, ok := t.runnableTool.(interface {
+		ProcessRequest(ctx agent.Context, req *model.LLMRequest) error
+	}); ok {
 		_, existedBefore := req.Tools[t.Name()]
 		if err := rp.ProcessRequest(ctx, req); err != nil {
 			return err
--- a/tool/tool.go
+++ b/tool/tool.go
@@ -185,11 +185,27 @@ type runnableTool interface {
 	Run(ctx agent.Context, args any) (result map[string]any, err error)
 }
 
+type requestProcessor interface {
+	ProcessRequest(ctx agent.Context, req *model.LLMRequest) error
+}
+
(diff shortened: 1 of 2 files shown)


tool/tool.go
```diff
@@ -185,11 +185,27 @@ type runnableTool interface {
 	Run(ctx agent.Context, args any) (result map[string]any, err error)
 }
 
+type requestProcessor interface {
```
wolo-lab: This interface is confusing. Since its only purpose is to check for the presence of a single method, I suggest using an inline anonymous interface.

tool/tool_test.go
```diff
@@ -296,3 +300,121 @@ func TestWithConfirmation(t *testing.T) {
...
+func (m *mockCustomRequestPackingTool) Run(ctx agent.Context, args any) (map[string]any, error) {
+	return map[string]any{}, nil
+}
+
+func (m *mockCustomRequestPackingTool) ProcessRequest(ctx agent.Context, req *model.LLMRequest) error {
+	*m.processRequestCalled = true
+	return toolutils.PackTool(req, m)
+}
+
+func TestWithConfirmation_ProcessRequest_Packing(t *testing.T) {
```
wolo-lab, line 386: 1) TestWithConfirmation_ProcessRequest - please simulate a side effect in processRequest (e.g. add systemInstruction mirroring loadmemorytool) in mockCustomRequestTool and verify if the side effect is present in test - regression test

2) TestWithConfirmation_ProcessRequest_Packing - assert the confirmation gate fires at Run. `req.Tools[...] == tools[0]` proves the swap happened, but not its effect. Pull the object back out of the registry and call Run on it; assert it returns `tool.ErrConfirmationRequired` and that confirmation was requested. Since the inner tool's Run just returns success, this cleanly proves the wrapper's Run executes — i.e. confirmation isn't bypassed after packing.

wolo-lab (dismissed):

tool/tool_test.go
hanorik: Done.

tool/tool.go
hanorik: Good point! Changed.

wolo-lab (approved):
Nice!

PR #1135, opened 2026-07-08
refactor: replace boolean endInvocation flag with atomic.Bool to enable cross-context state propagation

hanorik:
Fixes an issue where `EndInvocation()` called inside sub-agent callbacks or child nodes did not propagate to parent/sibling contexts because `EndInvocation` was stored as a value-copied `bool`.

## Changes
- **Shared State**: Replaced `EndInvocation bool` with `*atomic.Bool` across `InvocationContext` implementations (`agent` and `internal/context`).
- **Propagation**: Added `EndInvocationPtr() *atomic.Bool` so derived child contexts (`WithContext`, `WithICDelta`, `parallelagent`, `workflow.AgentNode`) share the same underlying atomic flag.
- **Orchestration**: Updated `sequentialagent` and `loopagent` to check `ctx.Ended()` and stop execution when a sub-agent ends the invocation.

## Testing Plan
- Added unit tests in `sequentialagent`, `agent`, and `internal/context` packages verifying that `EndInvocation()` propagates across child/sub-agent contexts.
- Ran `go test -race ./...`.
```

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"iter"
+	"sync/atomic"
 
 	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/genai"
@@ -368,7 +369,7 @@ type invocationContext struct {
 	isolationScope string
 	userContent    *genai.Content
 	runConfig      *RunConfig
-	endInvocation  bool
+	endInvocation  *atomic.Bool
 }
 
 // Apply implements [InvocationContext].
@@ -432,10 +433,16 @@ func (c *invocationContext) RunConfig() *RunConfig {
 }
 
 func (c *invocationContext) EndInvocation() {
-	c.endInvocation = true
+	if c.endInvocation != nil {
+		c.endInvocation.Store(true)
+	}
 }
 
 func (c *invocationContext) Ended() bool {
+	return c.endInvocation != nil && c.endInvocation.Load()
+}
+
+func (c *invocationContext) EndInvocationPtr() *atomic.Bool {
 	return c.endInvocation
 }
 
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -17,6 +17,7 @@ package agent
 import (
 	"context"
 	"iter"
+	"sync/atomic"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -122,9 +123,10 @@ func TestAgentCallbacks(t *testing.T) {
 			}
 
 			ctx := &invocationContext{
-				Context: t.Context(),
-				agent:   testAgent,
-				session: &mockSession{sessionID: "test-session"},
+				Context:       t.Context(),
+				agent:         testAgent,
+				session:       &mockSession{sessionID: "test-session"},
+				endInvocation: new(atomic.Bool),
 			}
 			var gotEvents []*session.Event
 			for event, err := range testAgent.Run(ctx) {
@@ -168,10 +170,13 @@ func TestEndInvocation_EndsBeforeMainCall(t *testing.T) {
 		t.Fatalf("failed to create agent: %v", err)
 	}
 
+	endedBool := new(atomic.Bool)
+	endedBool.Store(true)
+
 	ctx := &invocationContext{
 		Context:       t.Context(),
 		agent:         testAgent,
-		endInvocation: true,
+		endInvocation: endedBool,
 		session:       &mockSession{sessionID: "test-session"},
 	}
 	for _, err := range testAgent.Run(ctx) {
@@ -204,9 +209,10 @@ func TestEndInvocation_EndsAfterMainCall(t *testing.T) {
 	}
 
 	ctx := &invocationContext{
-		Context: t.Context(),
-		agent:   testAgent,
-		session: &mockSession{sessionID: "test-session"},
+		Context:       t.Context(),
+		agent:         testAgent,
+		session:       &mockSession{sessionID: "test-session"},
+		endInvocation: new(atomic.Bool),
 	}
 	var gotEvents []*session.Event
 	for event, err := range testAgent.Run(ctx) {
@@ -262,9 +268,10 @@ type testKey struct{}
 func TestWithContext(t *testing.T) {
 	baseCtx := t.Context()
 	inv := &invocationContext{
-		Context:      baseCtx,
-		invocationID: "test",
-		branch:       "branch",
+		Context:       baseCtx,
+		invocationID:  "test",
+		branch:        "branch",
+		endInvocation: new(atomic.Bool),
 	}
 
 	key := testKey{}
@@ -274,7 +281,13 @@ func TestWithContext(t *testing.T) {
 	if got.Value(key) != val {
 		t.Errorf("WithContext() did not update context")
 	}
-	if diff := cmp.Diff(inv, got, cmp.AllowUnexported(invocationContext{}), cmpopts.IgnoreFields(invocationContext{}, "Context")); diff != "" {
+	atomicBoolComparer := cmp.Comparer(func(a, b *atomic.Bool) bool {
+		if a == nil || b == nil {
+			return a == b
+		}
+		return a.Load() == b.Load()
+	})
+	if diff := cmp.Diff(inv, got, cmp.AllowUnexported(invocationContext{}), cmpopts.IgnoreFields(invocationContext{}, "Context"), atomicBoolComparer); diff != "" {
 		t.Errorf("WithContext() params mismatch (-want +got):\n%s", diff)
 	}
 }
@@ -349,5 +362,77 @@ func TestFindAgent(t *testing.T) {
 				t.Errorf("FindAgent(%q) = %v, want %v", tt.agentName, got, tt.want)
 			}
 		})
+	}
+}
+
+func TestSubAgentCallbackEndInvocationPropagated(t *testing.T) {
+	t.Parallel()
+
+	subAgentCalled := false
+
+	subAgent, err := New(Config{
+		Name: "sub_agent",
+		BeforeAgentCallbacks: []BeforeAgentCallback{
+			func(ctx Context) (*genai.Content, error) {
+				subAgentCalled = true
+				return genai.NewContentFromText("sub agent ended", genai.RoleModel), nil
+			},
+		},
+		Run: func(InvocationContext) iter.Seq2[*session.Event, error] {
+			return func(func(*session.Event, error) bool) {}
+		},
+	})
+	if err != nil {
+		t.Fatalf("failed to create sub_agent: %v", err)
+	}
+
+	parentRunCalled := false
+	parentEndedAfterSubAgent := false
+
+	parentAgent, err := New(Config{
+		Name:      "parent_agent",
+		SubAgents: []Agent{subAgent},
+		Run: func(ctx InvocationContext) iter.Seq2[*session.Event, error] {
+			parentRunCalled = true
+			return func(yield func(*session.Event, error) bool) {
+				for _, sa := range ctx.Agent().SubAgents() {
+					for event, err := range sa.Run(ctx) {
+						if !yield(event, err) {
+							return
+						}
+					}
+				}
+				// Demonstrates that when sub-agent BeforeAgentCallback returns non-nil content
+				// (which calls ctx.EndInvocation() on the sub-agent context), the endInvocation
+				// flag propagates to the parent context via *atomic.Bool.
+				parentEndedAfterSubAgent = ctx.Ended()
+			}
+		},
+	})
+	if err != nil {
+		t.Fatalf("failed to create parent_agent: %v", err)
+	}
+
+	parentCtx := &invocationContext{
+		Context:       t.Context(),
+		agent:         parentAgent,
+		session:       &mockSession{sessionID: "test-session"},
+		endInvocation: new(atomic.Bool),
+	}
+
+	for _, err := range parentAgent.Run(parentCtx) {
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+	}
+
+	if !parentRunCalled {
+		t.Error("expected parent run to be called")
+	}
+	if !subAgentCalled {
+		t.Error("expected sub_agent BeforeAgentCallback to be called")
+	}
+	if !parentEndedAfterSubAgent {
+		t.Error("expected parentEndedAfterSubAgent to be true after sub-agent callback ended invocation")
 	}
 }
--- a/agent/workflowagents/parallelagent/agent.go
+++ b/agent/workflowagents/parallelagent/agent.go
@@ -18,6 +18,7 @@ package parallelagent
 import (
 	"fmt"
 	"iter"
+	"sync/atomic"
 
 	"golang.org/x/sync/errgroup"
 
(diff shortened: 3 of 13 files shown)

### CODE REVIEW

File: agent/workflowagents/sequentialagent/agent.go [hunk 1]
```diff
@@ -78,6 +78,9 @@ type sequentialAgent struct{}
 func (a *sequentialAgent) Run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
 	return func(yield func(*session.Event, error) bool) {
 		for _, subAgent := range ctx.Agent().SubAgents() {
+			if ctx.Ended() {
```
karolpiotrowicz, line 81: Could the same guard go in `RunLive`? It iterates sub-agents without checking `ctx.Ended()`, so a sub-agent ending the invocation won't stop the following ones in live/streaming mode — the fix would then cover both paths.

agent/workflowagents/parallelagent/agent.go
```diff
@@ -78,17 +79,22 @@ func run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
...
-				RunConfig:    ctx.RunConfig(),
-				InvocationID: ctx.InvocationID(),
+				Artifacts:     ctx.Artifacts(),
+				Memory:        ctx.Memory(),
+				Session:       ctx.Session(),
+				Branch:        branch,
+				Agent:         subAgent,
+				UserContent:   ctx.UserContent(),
+				RunConfig:     ctx.RunConfig(),
+				EndInvocation: endInvPtr,
```
karolpiotrowicz: Sharing the parent's atomic across all siblings means one sibling ending the invocation can stop the others mid-flight (through the new `Ended()` checks in nested sequential/loop agents). That's a change from the "isolated" behavior documented for this agent, and from adk-python, where the workflow agents don't consult `end_invocation` at all. If cross-sibling termination is intended, a short comment on the rationale would help; if not, consider giving each sibling its own flag and reconciling back to the parent after the group completes.

agent/agent.go
```diff
@@ -368,14 +369,21 @@ type invocationContext struct {
 	isolationScope string
 	userContent    *genai.Content
 	runConfig      *RunConfig
-	endInvocation  bool
+	endInvocation  *atomic.Bool
+}
+
+func (c *invocationContext) initEndInvocation() {
```
karolpiotrowicz: This lazy `if == nil { new(...) }` is a check-then-set on the pointer field, so it isn't thread-safe on its own. It's fine today because production contexts are eager-initialized in `NewInvocationContext` and the only lazily-initialized type is test-only — but in a change whose purpose is safe shared state, eager-initializing at construction (and dropping the lazy path) would remove the footgun and simplify the call sites.

karolpiotrowicz (commented):
Thanks for this — the value-copy propagation bug is real, and the red→green tests (`TestSubAgentCallbackEndInvocationPropagated`, `TestSequentialAgent_SubAgentEndInvocationPropagated`) capture it well. I reproduced it locally: both new tests fail on `main` and pass with this change, the full `go test -race ./...` suite is green (one unrelated `remoteagent` timing flake), and there's no public API change. Keeping `EndInvocationPtr()` off the public interface via a type assertion is a nice touch.

I left a few inline notes. The main things I'd like to see before merge:

- **Live path isn't covered** — `sequentialAgent.RunLive` doesn't have the `Ended()` short-circuit that `Run` now has (inline).
- **ParallelAgent semantics need a deliberate decision** — sharing one flag across siblings changes isolation behavior (inline).
- **`initEndInvocation` isn't thread-safe** — worth eager-initializing (inline).
- **Test coverage** — `loopagent`, `parallelagent` (shared atomic), and `workflow/agent_node.go` are changed but untested; could you add cases for those?
- **Rebase on `main`** — the branch predates the `runner.NewInMemory` addition, so an API diff against current `main` looks misleadingly like a removal; a rebase clears it. (It also sets up the small follow-up that #370 describes.)

Overall the core mechanism and the sequential propagation are solid — nice work.

agent/agent.go
hanorik: Switched to initialization at construction.

agent/workflowagents/parallelagent/agent.go
hanorik: Good point - I've missed the moment about isolated behavior for this node. Made every sibling to have its own flag instead.

agent/workflowagents/sequentialagent/agent.go
hanorik: Added.

hanorik:
Also added missing test coverage.

wolo-lab (commented):
The fix looks good. 

There's one thing worrying me. adk-python decided explicitly not to fix it: https://github.com/google/adk-python/issues/1452.

Submitting this fix will introduce the difference in behavior between adks. Let me sync with them first.

agent/workflowagents/parallelagent/agent.go
```diff
@@ -73,28 +74,40 @@ func run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
...
+				RunConfig:     ctx.RunConfig(),
+				EndInvocation: siblingEndInv,
+				InvocationID:  ctx.InvocationID(),
 			})
 
 			if err := runSubAgent(subCtx, subAgent, resultsChan, doneChan); err != nil {
 				return fmt.Errorf("failed to run sub-agent %q: %w", subAgent.Name(), err)
 			}
 
+			if siblingEndInv.Load() {
```
karolpiotrowicz, line 107: This sits after the early `return` on line 104, so a sibling that calls `EndInvocation()` and then errors never reaches it, `anySiblingEnded` stays false, and the parent never learns. `errgroup.WithContext` cancels the group on the first error, so one failing sibling drops the end signal of every sibling still in flight.

Not a regression — the same case fails on `main`, because the propagation does not exist there — so this is an incompleteness of the new mechanism rather than something the branch broke. Moving the check into a `defer` at the top of the `errGroup.Go` closure would cover it, but note that also makes an aborted run mark the parent ended, which is a semantic worth choosing deliberately rather than inheriting.

agent/agent_test.go
```diff
@@ -274,7 +281,13 @@ func TestWithContext(t *testing.T) {
 	if got.Value(key) != val {
 		t.Errorf("WithContext() did not update context")
 	}
-	if diff := cmp.Diff(inv, got, cmp.AllowUnexported(invocationContext{}), cmpopts.IgnoreFields(invocationContext{}, "Context")); diff != "" {
+	atomicBoolComparer := cmp.Comparer(func(a, b *atomic.Bool) bool {
+		if a == nil || b == nil {
+			return a == b
+		}
+		return a.Load() == b.Load()
```
karolpiotrowicz, line 288: This compares the two flags by value, so the assertion passes whenever the values happen to match — including if `WithContext` allocated a brand-new flag. Pointer identity is the invariant the whole change rests on, and it is what a future refactor would most plausibly break here. `cmp.Comparer(func(a, b *atomic.Bool) bool { return a == b })` asserts the property directly, and makes the diff output meaningful when it fails.

The comparer in `internal/context/context_test.go` has the same shape. Identity is covered behaviorally elsewhere (`agent_node_test.go` and the parallel-agent test both fail if the pointer is not shared), so the gap is confined to `WithContext`.

karolpiotrowicz (changes requested):
Re-reviewed against current `main`. The branch is 133 commits behind, so I ran everything twice: once on the head as it stands, and once on the merge result of `main` (`a6114727`) with this branch. It merges cleanly, and build, `go vet`, `gofmt`, `go mod tidy`, `golangci-lint`, `staticcheck` and the full test suite are all green on the merge result, in both the root module and `plugin/agentanalytics`. `apidiff` reports no signature changes on any of the five touched exported packages.

The mechanism is right and the sequential propagation genuinely works. What I found is that turning a value field into a shared pointer couples three sets of things that used to be independent, and each one fails silently rather than loudly. There are two distinct root causes, and I want to separate them because one fix does not cover both.

## Concurrently scheduled units now share one flag

`WithDelta` shallow-copies, so the `*atomic.Bool` is shared rather than snapshotted, and both places that fan out over one parent context hand every child the same pointer.

For workflow graphs that is [`scheduler.go:382`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/workflow/scheduler.go#L382) deriving each `perNodeCtx`, then [`agent_node.go:106-118`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/workflow/agent_node.go#L106-L118) threading it into the child invocation context. On the merge base that line passed `EndInvocation: ctx.Ended()`, a value snapshot. A two-branch graph where branch A's agent ends the invocation and branch B is gated until it has:

<details><summary>Reproduction — branch B is skipped with no error</summary>

```go
// workflow/zz_repro_sched_test.go, package workflow
func TestReproS_SchedulerFanOutCoupling(t *testing.T) {
	aEnded := make(chan struct{})
	var bAgentRan atomic.Bool

	agentA, _ := agent.New(agent.Config{Name: "a_ender",
				ctx.EndInvocation()
				close(aEnded)
				ev := session.NewEvent(ctx, ctx.InvocationID())
				ev.Output = "a"
				yield(ev, nil)
			}
		}})
	agentB, _ := agent.New(agent.Config{Name: "b_worker",
				bAgentRan.Store(true)
				ev := session.NewEvent(ctx, ctx.InvocationID())
				ev.Output = "b"
				yield(ev, nil)
			}
		}})
	nodeA, _ := NewAgentNode(agentA, NodeConfig{})
	nodeB, _ := NewAgentNode(agentB, NodeConfig{})

	// The gate makes the interleaving deterministic rather than timing-dependent.
	gate := NewFunctionNode("gate", func(ctx agent.Context, input string) (string, error) {
		select {
		case <-aEnded:
			t.Error("gate timed out")
		}
		return "go", nil
	}, NodeConfig{})

	w, _ := New("", []Edge{{From: Start, To: nodeA}, {From: Start, To: gate}, {From: gate, To: nodeB}})

	resp, _ := svc.Create(t.Context(), &session.CreateRequest{AppName: "app", UserID: "u", SessionID: "s"})
	ic := icontext.NewInvocationContext(t.Context(), icontext.InvocationContextParams{Session: resp.Session})

	var gotErr error
	for _, err := range w.Run(ic) {
		if err != nil {
			gotErr = err
		}
	}
	t.Logf("branch B agent body ran: %v ; workflow error: %v", bAgentRan.Load(), gotErr)
	if !bAgentRan.Load() {
		t.Errorf("branch B's agent was silently skipped")
	}
}
```

```
go test -count=1 -run TestReproS ./workflow/

on main without this branch:   branch B agent body ran: true
                               (and the workflow reports two terminal outputs, so both branches produced one)
on main with this branch:      branch B agent body ran: false ; workflow error: <nil>
```
</details>

Branch B is not failed and not cancelled. It completes, contributes nothing, and its successors receive `nil`.

The same mechanism reaches `ParallelWorker`, which I think is the worse of the two because the result is corrupted rather than merely short. [`parallel_worker.go:123-124`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/workflow/parallel_worker.go#L123-L124) derives each item's context the same way and launches a goroutine per item, and the comment just above it at [`:102-105`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/workflow/parallel_worker.go#L102-L105) states per-item isolation as the design intent. With `maxConcurrency=1` so the ordering is deterministic, three items over an agent that ends the invocation on the first:

<details><summary>Reproduction — two of three items yield nil into the aggregate, no error</summary>

```go
// workflow/zz_repro_pw_test.go, package workflow
func TestReproP_ParallelWorkerItemCoupling(t *testing.T) {
	calls := 0
	a, _ := agent.New(agent.Config{Name: "ender_once",
				calls++
				if calls == 1 {
					ctx.EndInvocation()
				}
				ev := session.NewEvent(ctx, ctx.InvocationID())
				ev.Output = "out"
				yield(ev, nil)
			}
		}})
	an, _ := NewAgentNode(a, NodeConfig{})
	pw, _ := NewParallelWorker("parallel", an, 1, NodeConfig{})

	resp, _ := svc.Create(t.Context(), &session.CreateRequest{AppName: "app", UserID: "u", SessionID: "s"})
	ic := icontext.NewInvocationContext(t.Context(), icontext.InvocationContextParams{Agent: a, Session: resp.Session})

	var got []any
	for ev, err := range pw.Run(agent.NewContext(ic), []any{"a", "b", "c"}) {
		if err != nil {
		}
		if out, ok := extractOutput(ev); ok {
			got, _ = out.([]any)
		}
	}
	t.Logf("agent Run bodies executed: %d ; aggregated outputs: %#v", calls, got)
}
```

```
go test -count=1 -run TestReproP ./workflow/

on main without this branch:   bodies executed: 3 ; outputs: []interface{}{"out", "out", "out"}
on main with this branch:      bodies executed: 1 ; outputs: []interface{}{"out", nil, nil}
                               error returned: nil
```
</details>

Both reproductions pass on `main` at [`a6114727`](https://github.com/google/adk-go/commit/a61147276167f6b1b636d863f23a030c0a6babdf) and on the merge base at [`d0111d0d`](https://github.com/google/adk-go/commit/d0111d0dc15cff75bb84974596089370c1e7992b), and fail only with this branch applied, so the 133 commits in between are not the cause.

Worth knowing before you pick a direction: ordinary `LlmAgent` nodes escape both of these by accident. `NewAgentNode` defaults to single-turn mode, and [`llm_agent_wrapper.go:111`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/agent/llmagent/llm_agent_wrapper.go#L111) rebuilds the invocation context without `EndInvocation`, so the pointer [`agent_node.go:118`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/workflow/agent_node.go#L118) just threaded is dropped one call later. That is why the suite is green. It also means threading the flag at that site — which otherwise looks like the obvious completion of this change — would widen both reproductions above to the most common node type in the framework. The two need deciding together.

## A descendant ending the invocation now silences its ancestors' teardown

[`agent.go:203`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/agent/agent.go#L203) gates `runAfterAgentCallbacks` on `nodeCtx.Ended()`. Because the flag is now shared upward, a leaf ending the invocation suppresses the `AfterAgentCallbacks` of every agent above it. The trigger is the framework's own: [`agent.go:264`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/agent/agent.go#L264) and [`:285`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/agent/agent.go#L285) call `EndInvocation()` whenever a before-agent callback returns content, so a guardrail on one sub-agent now disables logging, metrics and state-flush callbacks all the way to the root. A parent agent with an `AfterAgentCallback` over a sub-agent whose before-callback returns content runs the callback on `main` and skips it with this branch applied.

`parallelagent` hits the same line one frame closer, and there the writer is this PR's own new code. The funnel at [`parallelagent/agent.go:122-124`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/agent/workflowagents/parallelagent/agent.go#L122-L124) stores into the parent flag before `close(resultsChan)`, so by the time control returns to `(*agent).Run` the parallel agent's *own* `AfterAgentCallbacks` are skipped — in the one agent whose [`doc comment at :39`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/agent/workflowagents/parallelagent/agent.go#L39) promises sub-agents run "in isolated manner".

I read the existing [`doc on AfterAgentCallback at agent.go:136-137`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/agent/agent.go#L136-L137) ("skipped also if EndInvocation was called before"), which does point this way, and it is why I have these below the two above rather than alongside them. It is unscoped as to *whose* `EndInvocation`, though, and I do not think it stretches to ancestors.

## Smaller things

`NewCleanToolContextTestOnly` at [`common_context.go:501-504`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/agent/common_context.go#L501-L504) builds the `invocationContext` literal without `endInvocation`, so on that context `EndInvocation()` is a permanent no-op and `Ended()` permanently false — weaker than the `bool` it replaced. It is documented as test-only and the sole in-repo caller is a test, so the blast radius is user-written test doubles rather than production, but it is a regression on an exported symbol and the one-line nature of it is why I mention it at all. Five test literals in the diff were patched to add `endInvocation: new(atomic.Bool)`. This is the one production literal of the same shape.

The carrier capability is an anonymous `interface{ EndInvocationPtr() *atomic.Bool }` written out three times, at [`agent_node.go:106`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/workflow/agent_node.go#L106), [`parallelagent/agent.go:78`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/agent/workflowagents/parallelagent/agent.go#L78) and [`common_context.go:225`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/agent/common_context.go#L225), and never declared or documented. Every failure path returns nil silently. I checked the consequence and it is milder than it looks — the nil lands in [`invocation_context.go:51-53`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/internal/context/invocation_context.go#L51-L53) and becomes a fresh working flag, which is base behavior, and the base code called `ctx.Ended()` there, which panics on `StrictContextMock`, so this removed a panic. Naming the interface once would still be worth it, mostly so the failure is visible.

I have left a note on the `parallelagent` reconciliation, which is skipped on the error path — not a regression, but an incompleteness of the new mechanism.

## On the concurrency claim itself

The `*atomic.Bool` is doing real work — [`scheduler.go:409`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/workflow/scheduler.go#L409) and [`parallel_worker.go:124`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/workflow/parallel_worker.go#L124) both create genuinely concurrent readers and writers of one flag, which is exactly the coupling above. But no test exercises it. I checked this rather than assuming: swapping the atomic for a plain-bool drop-in with the same API, across every file that carries the flag, leaves `go test -race` reporting zero data races on the touched packages, on `./...`, and under `-count=5 -shuffle=on`. A two-goroutine test over one context does trip the detector on that same tree, so the race detector can see this flag — the suite simply never drives it concurrently. As it stands, a future change back to a plain `bool` would ship green.

A test that drives `EndInvocation()` and `Ended()` from two unsynchronized goroutines over one derived context would close that, and it is cheap.

Two smaller test notes, one of them inline: the new `atomicBoolComparer` compares by value rather than pointer identity, and the 1 s `time.Sleep` in `TestParallelAgent_IsolatedEndInvocation` makes that assertion wall-clock dependent and costs 5 s under `-count=5`, though the assertion itself does discriminate.

Finally, [`scheduler.go:64-77`](https://github.com/google/adk-go/blob/51eb2fee6586d1fe476b379ada99df573115d880/workflow/scheduler.go#L64-L77) still says producers "share only `eventQueue`" and therefore need no mutex. They now also share a mutable flag reachable through `perNodeCtx`.

PR #1143, opened 2026-07-10
feat(auth): add core credential and provider package

wolo-lab:
## Problem
ADK Go has no shared primitive for authenticating outbound HTTP calls. The next
steps in this stack — per-request auth for MCP toolsets and A2A remote agents —
need a common way to (a) represent a *resolved* credential and apply it to
request headers, and (b) *resolve* a credential per invocation (and per user)
through a pluggable seam: static token, API key, any oauth2 token source, ADC,
service account, and later interactive OAuth consent.
Without this foundation each consumer would hand-roll header/token handling,
there'd be nowhere to hang richer providers, and no seam for a human-in-the-loop
consent flow. adk-python already has this concept (BaseAuthProvider /
AuthCredential); adk-go has had no equivalent.

## Summary
Adds a small, dependency-light `auth/` package (not yet wired into any consumer —
it's the base of the auth stack):
- **`Credential`** — an open interface whose `Apply(http.Header)` writes the
  credential onto an outbound request. Concrete kinds: `APIKeyCredential`,
  `BearerCredential`, `BasicCredential`, and `OAuth2Credential` (a static token
  or any auto-refreshing `oauth2.TokenSource`). `WithHeaders` decorates any
  credential with extra headers (e.g. `x-goog-user-project`).
- **`CredentialProvider`** — a `context.Context`-based seam that resolves a
  credential for the current invocation, with a `ProviderFunc` adapter and a
  `ConsentRequiredError` (detected via `errors.As`) for the future
  interactive-consent flow.
- **Built-in providers** — `StaticToken`, `APIKey`, `TokenSourceProvider`,
  `ADC`, and `ServiceAccount` (access token or audience-scoped ID token).
- **`Transport`** — a context-aware `http.RoundTripper` that resolves the
  provider per request and applies the credential, cloning the request so the
  caller's is never mutated.
Key decisions:
- Token exchange/refresh is delegated to `golang.org/x/oauth2`, not re-modeled.
  **No new module dependencies** (`go mod tidy -diff` is empty).
- `Credential` is an interface (not a struct-of-optionals), so illegal states
  (empty / multiple kinds) are unrepresentable.
- Lazy token-source providers (`ADC`, `ServiceAccount`) init once in a detached
  goroutine and memoize on success; each caller selects on init finishing **or
  its own ctx**, so a slow init never blocks a caller past its deadline. Init
  runs under `context.WithoutCancel(ctx)` (keeps values, drops the request's
  cancellation) but is bounded by an internal timeout, so a hung init fails and
  is retried instead of wedging the provider. Keeps library code free of
  `context.Background()`.
- `ServiceAccount` with an explicit key + access token requires scopes (stricter
  than adk-python — a scope-bound token with no scopes is unusable, so fail
  fast); service-account JSON is loaded via the type-specific
  `option.WithAuthCredentialsJSON(option.ServiceAccount, …)` /
  `google.CredentialsFromJSONWithType(…, google.ServiceAccount, …)`, which
  reject non-service-account configs.


auth/providers.go
```diff
@@ -0,0 +1,217 @@
...
+			return OAuth2Credential{TokenSource: ts}, nil
+		}
+		a := cur
+		if a == nil {
+			a = &attempt{done: make(chan struct{})}
+			cur = a
+			// Detached: keep the request's values but drop its cancellation, so
+			// the shared source outlives the caller that triggered it.
+			go func() {
+				a.ts, a.err = init(context.WithoutCancel(ctx))
```
baptmont: WithoutCancel also drops deadline/timeout. If init hangs the goroutine never finishes and cur is never reset and so every later caller keeps waiting on the same stuck init and the provider can't recover.

we could add our own timeout
``` golang
dctx := context.WithoutCancel(ctx)                     
dctx, cancel := context.WithTimeout(dctx, 30*time.Second) 
defer cancel()
a.ts, a.err = init(dctx)
```

baptmont:
PR description is outdated mentioning Scheme marker types

auth/providers.go
wolo-lab: Good catch! Applied. Thanks

wolo-lab:
> PR description is outdated mentioning Scheme marker types

Right, updated.

baptmont (approved):

PR #1144, opened 2026-07-10
chore(deps): bump golang.org/x/crypto to v0.52.0

wolo-lab:
Clears 13 of 14 dormant govulncheck advisories against x/crypto v0.51.0 (all but GO-2026-5932, which has no fixed version yet). All are currently unreachable so the nightly vulncheck job is green; this is proactive hygiene to stay ahead before any become reachable.

Verified: go build, go test -race -shuffle=on, golangci-lint (0 issues), go mod tidy -diff (clean); govulncheck module-level advisories drop 14 -> 1.


kdroste-google (approved):
LGTM

PR #1146, opened 2026-07-10
chore: initialize Go workspace and remove it from gitignore

hanorik:
Standing up the multi-module skeleton by adding a root `go.work` referencing the core module and new modules that will come in the future.


go.work.sum
wolo-lab, line 1: we don't need to commit that. After running tests it is modified so there is a dirty workspace for contributors and they might be confused what to do with changes in this file.

.gitignore
```diff
@@ -16,7 +16,3 @@
 coverage.*
 *.coverprofile
 profile.cov
-
-# Go workspace file
-go.work
```
wolo-lab, line 21: I think we want to track go.work, so it should not be ignored by git.

wolo-lab (approved):

go.work
```diff
@@ -0,0 +1,3 @@
+go 1.25.0
```
hyangah, line 1: Can we avoid committing go.work?

People use go.work to configure their local dependencies (for example, a user can create a workspace that includes their main module and adk-go module to test). Having go.work in the repo may interfere the workflow or confuse during dev.

The recommendation is not to check in go.work in most cases - https://go.dev/ref/mod#workspaces:~:text=It%20is%20generally%20inadvisable%20to%20commit%20go.work%20files%20into%20version%20control%20systems

hanorik: Hi, this is a preparation for the #738 BigQuery plugin deployment. It's decided to have a separate module for this plugin so it feels that to keep everything clear we need it for the purpose. I'm still working on the CI enhancement so if you have some ideas regarding this changes feel free to share.

PR #1147, opened 2026-07-10
perf(session): reduce allocations in InMemoryService.AppendEvent

huntiezz:
## Summary

- Add `cloneEventForStorage()` to deep-clone event maps and slices only when non-empty
- Trim temporary state keys before persisting events to the canonical in-memory session store

## Motivation

`InMemoryService.AppendEvent` runs on every non-partial LLM/tool event in the default in-memory runner path. It previously called `maps.Clone` / `slices.Clone` on every action field even when empty, causing unnecessary heap allocations on the hot path.

This also ensures events retrieved via `Get` have temporary state keys stripped (matching the runner-held session copy).

## Testing Plan

- [x] `go test -race -count=1 ./session/...`
- [x] Extended `TestInMemoryService_AppendEvent_PreservesInputEventTempState` to verify `Get` returns stripped temp keys
- [x] Added `BenchmarkInMemoryService_AppendEvent`

### DIFF
--- a/session/inmemory.go
+++ b/session/inmemory.go
@@ -219,32 +219,12 @@ func (s *inMemoryService) AppendEvent(ctx context.Context, curSession Session, e
 	}
 
 	// update the in-memory session
-	if err := sess.appendEvent(event); err != nil {
+	processedEvent, err := sess.appendEvent(event)
+	if err != nil {
 		return fmt.Errorf("fail to set state on appendEvent: %w", err)
 	}
 
-	eventCopy := &Event{
-		ID:             event.ID,
-		InvocationID:   event.InvocationID,
-		Timestamp:      event.Timestamp,
-		Author:         event.Author,
-		Branch:         event.Branch,
-		IsolationScope: event.IsolationScope,
-		Actions: EventActions{
-			StateDelta:                 maps.Clone(event.Actions.StateDelta),
-			ArtifactDelta:              maps.Clone(event.Actions.ArtifactDelta),
-			RequestedToolConfirmations: maps.Clone(event.Actions.RequestedToolConfirmations),
-			TransferToAgent:            event.Actions.TransferToAgent,
-			Escalate:                   event.Actions.Escalate,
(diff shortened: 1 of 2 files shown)


huntiezz:
Can someone review this?

jjsasha63:
Thanks for this, and sorry for the slow reply — nearly two months is too long.

You found a real bug. The `temp:` stripping in this PR predates issue #1354 by over five weeks, and nobody else reported it until then. Two problems with the rest of the change mean we should close this in favour of #1356.

First, the perf claim does not reproduce. I ran your `BenchmarkInMemoryService_AppendEvent` five times against main and five times against this PR merged into main:

- empty deltas: 9 allocs/op, ~729 B/op on both sides
- with a `StateDelta` (a variant I added, since your benchmark leaves the action maps unset): 24 allocs/op, ~2788 B/op on both sides

`maps.Clone(nil)` and `slices.Clone(nil)` return nil without allocating — 0 allocs/op at 0.39 ns/op when I measured them directly. The unused `EventActions` maps are nil rather than empty, so the new `if len(...) > 0` guards never fire. Only a non-nil empty map allocates (1 alloc, 48 B), and that shape does not reach this path.

Second, `cloneEventForStorage` drops `Actions.Compaction`. We added `Compaction: event.Actions.Compaction.clone()` to the storage copy in #1231. At current main this PR conflicts on exactly that line, and resolving it your way makes `TestInMemoryService_AppendEvent_CopiesCompaction` fail with `compaction was not persisted`, so the compaction record is lost from the canonical store.

One smaller thing if you revisit this: the length guards also turn a non-nil empty map into nil in the stored record. `EventActions.MarshalJSON` uses `omitempty`, so `{"stateDelta":{},"artifactDelta":{}}` becomes `{}` — the fields are dropped rather than written as empty objects.

#1356 fixes the `temp:` leak in one line, keeps `Compaction`, and merges cleanly into `session/inmemory.go`, so we are going with that.

If you want to come back to allocations here, a benchmark that shows a difference between the two trees is the place to start — happy to review it.

PR #1149, opened 2026-07-12
feat(auth/gcp): add credentials-service REST client

wolo-lab:
## Problem

ADK-go's auth subsystem needs to fetch end-user credentials from Google Cloud's
Agent Identity and IAM Connector credential services. There are no generated Go
client libraries for these (preview) services, and adk-go has no transport for
them at all — while adk-python already talks to them.

## Summary

- Adds `auth/gcp`, a hand-rolled REST client over `net/http` (no new module
  dependencies) for the Agent Identity and IAM Connector credential services.
- `Client.RetrieveCredential(ctx, Request)` takes a resource name, routes it to
  the right service (IAM Connector when the resource matches
  `projects/*/locations/*/connectors/*`, Agent Identity otherwise — same split
  as adk-python), retrieves an end-user credential, and maps the
  `{header, token}` result to an `auth.Credential`: an `Authorization: Bearer`
  header becomes a bearer credential; any other header becomes a header-based
  API key keyed by the full header string and mirrored into `X-Goog-Api-Key`
  (matching adk-python's `_construct_auth_credential`).
- Polls with exponential backoff while the service reports a non-interactive
  `pending` state (bounded by `Config.PollTimeout`, default 10s), and surfaces
  interactive consent as `*auth.ConsentRequiredError`.
- **Config-struct API** (`NewClient(ctx, *Config)`; a nil `*Config` or any zero
  field uses defaults), consistent with the rest of adk-go. Calls are
  authenticated with Application Default Credentials (cloud-platform scope)
  unless `Config.HTTPClient` is supplied; endpoints are overridable via
  `Config.AgentIdentityEndpoint` / `Config.ConnectorEndpoint` (used by tests).
- A non-2xx response is returned as a typed `*APIError{StatusCode, Body}` — the
  same shape `agentregistry` exports — so callers can tell a fatal 403 from a
  transient 503 with `errors.As` instead of matching on the message.
- **The ADC client refuses redirects.** `oauth2.Transport` re-signs every hop
  below the layer where `net/http` strips credentials on a cross-host redirect,
  so following one would hand the cloud-platform token to the redirect target
  and let it dictate the returned credential.
- The token source is detached from the construction context: `ctx` bounds
  credential discovery, not the client's life, so a client built inside a
  request-scoped context keeps refreshing its token afterwards.
- Hardening: caps the response body with `io.LimitReader` (classifying the
  status first, so an oversized error page still names it); validates the
  resource name against the GCP resource-name charset (and rejects `..`) before
  interpolating it into the request URL; rejects a returned header that is not a
  usable HTTP field name; truncates oversized error bodies on a UTF-8 rune
  boundary and escapes them into errors.
- Transport layer only — the `auth.CredentialProvider` that resolves the acting
  user from the invocation context is a later step (#1173).

### DIFF
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -64,6 +64,22 @@ var (
 	ErrPollTimeout = errors.New("gcp: timed out waiting for credentials")
 )
 
+// APIError is returned when a credential service responds with a non-2xx
+// status. Callers match it with errors.As to tell a fatal status (say 403) from
+// a transient one (503) without matching on the message.
+type APIError struct {
+	// StatusCode is the HTTP status code of the response.
+	StatusCode int
+	// Body is the response body, truncated, useful for diagnosing the failure.
+	Body string
+}
+
+func (e *APIError) Error() string {
+	// %q, not %s: the body is service-controlled and can carry control bytes
+	// that would otherwise forge lines in an operator's log.
+	return fmt.Sprintf("gcp: credentials service returned status %d: %q", e.StatusCode, e.Body)
+}
+
 // Client retrieves end-user credentials from the Agent Identity / IAM Connector
 // credential services and maps them to [auth.Credential].
 type Client struct {
@@ -296,13 +312,13 @@ func (c *Client) doPost(ctx context.Context, url string, body, out any) error {
 	if err != nil {
 		return fmt.Errorf("gcp: read response: %w", err)
 	}
-	if len(data) > maxBody {
-		return fmt.Errorf("gcp: credentials service returned status %d with a response exceeding %d bytes", resp.StatusCode, maxBody)
-	}
+	// Classify the status before the size check, so an oversized error page still
+	// reports the status — the most actionable field — instead of only its size.
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-		// %q, not %s: the body is service-controlled and can carry control bytes
-		// that would otherwise forge lines in an operator's log.
-		return fmt.Errorf("gcp: credentials service returned status %d: %q", resp.StatusCode, truncateForError(strings.TrimSpace(string(data))))
+		return &APIError{StatusCode: resp.StatusCode, Body: truncateForError(strings.TrimSpace(string(data)))}
+	}
+	if len(data) > maxBody {
+		return fmt.Errorf("gcp: credentials service response exceeded %d bytes", maxBody)
 	}
 	if err := json.Unmarshal(data, out); err != nil {
 		return fmt.Errorf("gcp: decode response: %w", err)
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -44,6 +45,15 @@ const (
 // connectorResourceRE matches an IAM Connector resource name; anything else is
 // routed to the Agent Identity service (same split as adk-python).
 var connectorResourceRE = regexp.MustCompile(`^projects/[^/]+/locations/[^/]+/connectors/[^/]+$`)
+
+// Sentinel errors from [Client.RetrieveCredential]; callers test with errors.Is.
+var (
+	// ErrConsentRejected means the end user rejected the consent request.
+	ErrConsentRejected = errors.New("gcp: user consent rejected")
+	// ErrPollTimeout means polling exceeded the poll timeout while the credential
+	// was still pending.
+	ErrPollTimeout = errors.New("gcp: timed out waiting for credentials")
+)
 
 // Client retrieves end-user credentials from the Agent Identity / IAM Connector
 // credential services and maps them to [auth.Credential].
@@ -142,11 +152,11 @@ func (c *Client) RetrieveCredential(ctx context.Context, req Request) (*auth.Cre
 		case statusConsentRequired:
 			return nil, &auth.ConsentRequiredError{AuthURI: res.consentURI, Nonce: res.consentNonce}
 		case statusRejected:
-			return nil, fmt.Errorf("gcp: user consent rejected for %q", req.Resource)
+			return nil, fmt.Errorf("%w for %q", ErrConsentRejected, req.Resource)
 		case statusPending:
 			remaining := time.Until(deadline)
 			if remaining <= 0 {
-				return nil, fmt.Errorf("gcp: timed out waiting for credentials for %q", req.Resource)
+				return nil, fmt.Errorf("%w for %q", ErrPollTimeout, req.Resource)
 			}
 			wait := min(backoff, remaining)
 			select {
@@ -190,8 +200,9 @@ func mapCredential(header, token string) (*auth.Credential, error) {
 		strings.HasPrefix(strings.ToLower(strings.TrimSpace(hint)), "bearer") {
 		return &auth.Credential{HTTP: &auth.HTTPCredential{Scheme: "bearer", Token: token}}, nil
 	}
-	// Non-bearer header: place the token in the named header. (adk-python also
-	// mirrors it into X-GOOG-API-KEY; deferred until a case needs it.)
+	// Non-bearer header -> header-based API key.
+	// TODO: for full adk-python parity also mirror the token into X-GOOG-API-KEY;
+	// needs an AdditionalHeaders field on auth.APIKeyCredential (additive, non-breaking).
 	return &auth.Credential{APIKey: &auth.APIKeyCredential{Name: name, Value: token}}, nil
 }
 
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -200,10 +200,11 @@ func mapCredential(header, token string) (auth.Credential, error) {
 		strings.HasPrefix(strings.ToLower(strings.TrimSpace(hint)), "bearer") {
 		return auth.BearerCredential{Token: token}, nil
 	}
-	// Non-bearer header -> header-based API key.
-	// TODO: for full adk-python parity also mirror the token into X-GOOG-API-KEY
-	// (via auth.WithHeaders) as a follow-up.
-	return auth.APIKeyCredential{Name: name, Value: token}, nil
+	// Non-bearer header -> header-based API key. adk-python also mirrors the
+	// token into X-Goog-Api-Key for custom headers (alongside the service's own
+	// header), so match that behavior.
+	key := auth.APIKeyCredential{Name: name, Value: token}
+	return auth.WithHeaders(key, map[string]string{"X-Goog-Api-Key": token}), nil
 }
 
 // doPost sends body as JSON to url and decodes a JSON response into out.
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -25,6 +25,7 @@ import (
 	"regexp"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
@@ -37,7 +38,9 @@ const (
 	defaultAgentIdentityURL = "https://agentidentitycredentials.googleapis.com"
 	defaultConnectorURL     = "https://iamconnectorcredentials.googleapis.com"
 
-	defaultPollTimeout    = 10 * time.Second
+	defaultPollTimeout = 10 * time.Second
+	// The credentials service documents an exponential polling backoff
+	// (0.5, 1, 2, 4, 8s); these constants track it.
 	defaultInitialBackoff = 500 * time.Millisecond
 	maxBackoff            = 8 * time.Second
 )
@@ -45,6 +48,12 @@ const (
 // connectorResourceRE matches an IAM Connector resource name; anything else is
 // routed to the Agent Identity service (same split as adk-python).
 var connectorResourceRE = regexp.MustCompile(`^projects/[^/]+/locations/[^/]+/connectors/[^/]+$`)
+
+// resourceNameRE bounds a resource name to the characters GCP resource names
+// use, so it can't inject extra path segments, a query, or a fragment into the
+// request URL it is interpolated into. A separate ".." check blocks path
+// traversal (dots are allowed so domain-style ids still pass).
+var resourceNameRE = regexp.MustCompile(`^[A-Za-z0-9._~/-]+$`)
 
 // Sentinel errors from [Client.RetrieveCredential]; callers test with errors.Is.
 var (
@@ -75,7 +84,9 @@ type Config struct {
 	AgentIdentityEndpoint string
 	// ConnectorEndpoint overrides the IAM Connector base URL (scheme+host).
 	ConnectorEndpoint string
-	// PollTimeout bounds the total time spent polling a pending retrieval.
+	// PollTimeout bounds the wall-clock time spent retrying a pending retrieval.
+	// It caps the retry loop, not an individual request; bound a single stalled
+	// request via ctx (or an HTTPClient with its own Timeout).
 	PollTimeout time.Duration
 }
 
@@ -94,10 +105,10 @@ func NewClient(ctx context.Context, cfg *Config) (*Client, error) {
 		initialBackoff:   defaultInitialBackoff,
 	}
 	if cfg.AgentIdentityEndpoint != "" {
-		c.agentIdentityURL = cfg.AgentIdentityEndpoint
+		c.agentIdentityURL = strings.TrimRight(cfg.AgentIdentityEndpoint, "/")
 	}
 	if cfg.ConnectorEndpoint != "" {
-		c.connectorURL = cfg.ConnectorEndpoint
+		c.connectorURL = strings.TrimRight(cfg.ConnectorEndpoint, "/")
 	}
 	if cfg.PollTimeout != 0 {
 		c.pollTimeout = cfg.PollTimeout
@@ -132,10 +143,13 @@ type Request struct {
 // If interactive consent is required it returns an [auth.ConsentRequiredError].
 func (c *Client) RetrieveCredential(ctx context.Context, req Request) (auth.Credential, error) {
 	if req.Resource == "" {
-		return nil, fmt.Errorf("gcp: RetrieveCredential requires a Resource")
+		return nil, errors.New("gcp: RetrieveCredential requires a Resource")
 	}
 	if req.UserID == "" {
-		return nil, fmt.Errorf("gcp: RetrieveCredential requires a UserID")
+		return nil, errors.New("gcp: RetrieveCredential requires a UserID")
+	}
+	if !resourceNameRE.MatchString(req.Resource) || strings.Contains(req.Resource, "..") {
+		return nil, fmt.Errorf("gcp: RetrieveCredential resource %q has invalid characters", req.Resource)
 	}
 
 	retrieve := c.retrieveAgentIdentity
@@ -205,23 +219,29 @@ type credentialPayload struct {
 	Header string `json:"header"`
 }
 
+// retrieveRequest is the JSON body for both services' credentials:retrieve RPC
+// (the auth provider / connector is bound to the URL path, not the body).
+type retrieveRequest struct {
+	UserID      string   `json:"userId,omitempty"`
+	Scopes      []string `json:"scopes,omitempty"`
+	ContinueURI string   `json:"continueUri,omitempty"`
+}
+
 // mapCredential maps the service's {header, token} tuple to an [auth.Credential]:
 // an "Authorization: Bearer" header becomes a bearer credential; any other header
 // name becomes a header-based API key.
 func mapCredential(header, token string) (auth.Credential, error) {
 	if header == "" || token == "" {
-		return nil, fmt.Errorf("gcp: credentials service returned an empty header or token")
+		return nil, errors.New("gcp: credentials service returned an empty header or token")
 	}
-	name, hint, _ := strings.Cut(header, ":")
-	name = strings.TrimSpace(name)
-	if strings.EqualFold(name, "authorization") &&
-		strings.HasPrefix(strings.ToLower(strings.TrimSpace(hint)), "bearer") {
+	name, scheme, _ := strings.Cut(header, ":")
+	if strings.EqualFold(strings.TrimSpace(name), "authorization") &&
+		strings.HasPrefix(strings.ToLower(strings.TrimSpace(scheme)), "bearer") {
 		return auth.BearerCredential{Token: token}, nil
 	}
-	// Non-bearer header -> header-based API key. adk-python also mirrors the
-	// token into X-Goog-Api-Key for custom headers (alongside the service's own
-	// header), so match that behavior.
-	key := auth.APIKeyCredential{Name: name, Value: token}
+	// Non-bearer header -> header-based API key. Matches adk-python: key by the
+	// full returned header, and mirror the token into X-Goog-Api-Key too.
+	key := auth.APIKeyCredential{Name: header, Value: token}
 	return auth.WithHeaders(key, map[string]string{"X-Goog-Api-Key": token}), nil
 }
 
@@ -236,6 +256,7 @@ func (c *Client) doPost(ctx context.Context, url string, body, out any) error {
 		return fmt.Errorf("gcp: build request: %w", err)
 	}
 	httpReq.Header.Set("Content-Type", "application/json")
+	httpReq.Header.Set("Accept", "application/json")
 
 	resp, err := c.httpClient.Do(httpReq)
 	if err != nil {
@@ -243,15 +264,37 @@ func (c *Client) doPost(ctx context.Context, url string, body, out any) error {
 	}
 	defer func() { _ = resp.Body.Close() }()
 
-	data, err := io.ReadAll(io.LimitReader(resp.Body, 1<<20))
+	// Read one byte past the cap so an oversized body is caught explicitly rather
+	// than fed to json.Unmarshal as silently truncated (and thus garbled) JSON.
+	const maxBody = 1 << 20
+	data, err := io.ReadAll(io.LimitReader(resp.Body, maxBody+1))
 	if err != nil {
 		return fmt.Errorf("gcp: read response: %w", err)
 	}
+	if len(data) > maxBody {
+		return fmt.Errorf("gcp: credentials service response exceeded %d bytes", maxBody)
+	}
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-		return fmt.Errorf("gcp: credentials service returned status %d: %s", resp.StatusCode, strings.TrimSpace(string(data)))
+		return fmt.Errorf("gcp: credentials service returned status %d: %s", resp.StatusCode, truncateForError(strings.TrimSpace(string(data))))
 	}
 	if err := json.Unmarshal(data, out); err != nil {
 		return fmt.Errorf("gcp: decode response: %w", err)
 	}
 	return nil
+}
+
+// truncateForError caps an error body so a large (e.g. HTML gateway) response
+// doesn't bloat the returned error.
+func truncateForError(s string) string {
+	const max = 1024
+	if len(s) <= max {
+		return s
+	}
+	// Back up to a rune boundary so a multi-byte rune straddling the cap isn't
+	// sliced into a mangled partial rune.
+	cut := max
+	for cut > 0 && !utf8.RuneStart(s[cut]) {
+		cut--
+	}
+	return s[:cut] + "..."
 }
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -126,7 +126,7 @@ type Request struct {
 // RetrieveCredential retrieves a credential for req, polling while the service
 // reports a non-interactive pending state (up to the configured poll timeout).
 // If interactive consent is required it returns an [auth.ConsentRequiredError].
-func (c *Client) RetrieveCredential(ctx context.Context, req Request) (*auth.Credential, error) {
+func (c *Client) RetrieveCredential(ctx context.Context, req Request) (auth.Credential, error) {
 	if req.Resource == "" {
 		return nil, fmt.Errorf("gcp: RetrieveCredential requires a Resource")
 	}
@@ -190,7 +190,7 @@ type retrieveResult struct {
 // mapCredential maps the service's {header, token} tuple to an [auth.Credential]:
 // an "Authorization: Bearer" header becomes a bearer credential; any other header
 // name becomes a header-based API key.
-func mapCredential(header, token string) (*auth.Credential, error) {
+func mapCredential(header, token string) (auth.Credential, error) {
 	if header == "" || token == "" {
 		return nil, fmt.Errorf("gcp: credentials service returned an empty header or token")
 	}
@@ -198,12 +198,12 @@ func mapCredential(header, token string) (*auth.Credential, error) {
 	name = strings.TrimSpace(name)
 	if strings.EqualFold(name, "authorization") &&
 		strings.HasPrefix(strings.ToLower(strings.TrimSpace(hint)), "bearer") {
-		return &auth.Credential{HTTP: &auth.HTTPCredential{Scheme: "bearer", Token: token}}, nil
+		return auth.BearerCredential{Token: token}, nil
 	}
 	// Non-bearer header -> header-based API key.
-	// TODO: for full adk-python parity also mirror the token into X-GOOG-API-KEY;
-	// needs an AdditionalHeaders field on auth.APIKeyCredential (additive, non-breaking).
-	return &auth.Credential{APIKey: &auth.APIKeyCredential{Name: name, Value: token}}, nil
+	// TODO: for full adk-python parity also mirror the token into X-GOOG-API-KEY
+	// (via auth.WithHeaders) as a follow-up.
+	return auth.APIKeyCredential{Name: name, Value: token}, nil
 }
 
 // doPost sends body as JSON to url and decodes a JSON response into out.
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -1,0 +1,227 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gcp
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"regexp"
+	"strings"
+	"time"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/google"
+
+	"google.golang.org/adk/v2/auth"
+)
+
+const (
+	cloudPlatformScope      = "https://www.googleapis.com/auth/cloud-platform"
+	defaultAgentIdentityURL = "https://agentidentitycredentials.googleapis.com"
+	defaultConnectorURL     = "https://iamconnectorcredentials.googleapis.com"
+
+	defaultPollTimeout    = 10 * time.Second
+	defaultInitialBackoff = 500 * time.Millisecond
+	maxBackoff            = 8 * time.Second
+)
+
+// connectorResourceRE matches an IAM Connector resource name; anything else is
+// routed to the Agent Identity service (same split as adk-python).
+var connectorResourceRE = regexp.MustCompile(`^projects/[^/]+/locations/[^/]+/connectors/[^/]+$`)
+
+// Client retrieves end-user credentials from the Agent Identity / IAM Connector
+// credential services and maps them to [auth.Credential].
+type Client struct {
+	httpClient       *http.Client
+	agentIdentityURL string
+	connectorURL     string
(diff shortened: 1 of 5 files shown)


auth/gcp/client.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	if cfg.PollTimeout != 0 {
+		c.pollTimeout = cfg.PollTimeout
+	}
+	if c.httpClient == nil {
+		creds, err := google.FindDefaultCredentials(ctx, cloudPlatformScope)
+		if err != nil {
+			return nil, fmt.Errorf("gcp: find default credentials: %w", err)
+		}
+		c.httpClient = oauth2.NewClient(ctx, creds.TokenSource)
```
karolpiotrowicz: `oauth2.NewClient` returns a client with a nil `CheckRedirect`, so up to 10 redirects are followed, and `oauth2.Transport.RoundTrip` clones the request and calls `token.SetAuthHeader` on **every** hop. Go's redirect header-stripping only withholds sensitive headers that were present on the *initial* request, so it never sees this one — the ADC cloud-platform token is freshly minted and attached to whatever host the redirect names. `doPost` only ever inspects the final hop, so the 3xx is invisible to the status check below.

I reproduced this with two test servers on different hostnames: a plain `http.Client` with `Authorization` set on the initial request delivered `""` to the redirect target (stripping works), while this client delivered `Bearer ya29.<token>` — and `RetrieveCredential` then returned `auth.BearerCredential{Token: "ATTACKER-SUPPLIED"}` from the redirect target's body with a nil error.

A `credentials:retrieve` RPC has no legitimate reason to redirect, so the cheapest fix is to refuse outright:

```go
hc := oauth2.NewClient(ctx, creds.TokenSource)
hc.CheckRedirect = func(*http.Request, []*http.Request) error { return http.ErrUseLastResponse }
c.httpClient = hc
```

It would also be worth rejecting a 3xx explicitly in `doPost` alongside the existing non-2xx check, so a caller-supplied `Config.HTTPClient` can't reintroduce it.

auth/gcp/client_test.go
```diff
@@ -0,0 +1,412 @@
...
+
+// wantAPIKey fails t unless applying cred sets the named header and the
+// X-Goog-Api-Key mirror (adk-python parity) to value.
+func wantAPIKey(t *testing.T, cred auth.Credential, name, value string) {
+	t.Helper()
+	h := http.Header{}
+	if err := cred.Apply(h); err != nil {
+		t.Fatalf("cred.Apply() error = %v", err)
+	}
+	if got := h.Get(name); got != value {
```
karolpiotrowicz, line 517: This helper checks `h.Get(name)` and `h.Get("X-Goog-Api-Key")` — but both call sites pass `name == "X-Goog-Api-Key"` (the table case at line 62 and the `mapCredential` case at line 296), so the two assertions read the same header and the mirror itself is never exercised.

I confirmed it by deleting the `auth.WithHeaders(...)` wrapper in `mapCredential` and re-running the package: still green. Since the mirror is the parity behaviour this change exists to provide, that's the one thing you'd most want a test to hold onto.

Adding a case with a genuinely different header name covers it:

```go
{
    name:       "agent identity third-party header",
    resource:   authProviderResource,
    bodies:     []string{`{"success":{"token":"KEY","header":"X-Acme-Token"}}`},
    wantAPIKey: [2]string{"X-Acme-Token", "KEY"},
},
```

karolpiotrowicz (commented):
The routing split, the sealed-outcome modelling of the two very different response shapes, and the resource-name allow-list are all solid, and I couldn't defeat the URL guard with any of the eleven injection candidates I tried (scheme, host, port, query, fragment, percent-encoded and Unicode traversal all bounce off it). Build, `go vet`, `staticcheck`, `golangci-lint`, `-race` and the full suite are all green. Two things I'd like to see addressed before this lands, plus a handful of smaller ones inline.

**The two that matter**

1. **A redirect off the credentials endpoint hands over the ADC token** (`auth/gcp/client.go:121`). `oauth2.NewClient` leaves `CheckRedirect` nil and `oauth2.Transport` re-signs *every* hop, so Go's own cross-host redirect header-stripping never applies. I reproduced it: with a genuinely cross-host redirect, a plain `http.Client` correctly drops the header, but this client delivers `Authorization: Bearer <ADC cloud-platform token>` to the redirect target — and then accepts that target's response body as the end-user credential, with a nil error. A `credentials:retrieve` call has no reason to redirect, and it's a one-line guard.

2. **The `X-Goog-Api-Key` mirror isn't actually tested** (`auth/gcp/client_test.go:406`). Every `wantAPIKey` case passes `"X-Goog-Api-Key"` as the header name, so `h.Get(name)` and `h.Get("X-Goog-Api-Key")` read the same header. I checked by deleting the `WithHeaders` wrapper: the suite stays green. Since that mirror is the parity behaviour the change is explicitly there to provide, it's worth a case with a genuinely different name.

**Smaller things, inline:** the API-key branch uses the whole returned header string as the header *name*; the connector's `error.message` skips the truncation `doPost` applies everywhere else; `%s` vs `%q` on the error body; the resource-validation test passing for the wrong reason; the oversize check hiding the HTTP status; and a couple of doc gaps around `ctx` lifetime and `Config.HTTPClient`.

**A question rather than a finding:** `Config.PollTimeout`'s doc is explicit that it caps the retry loop and not an individual request, and points callers at `ctx` — which is fair and idiomatic. Still, `RetrieveCredential(context.Background(), …)` against a peer that accepts the POST and never answers blocks indefinitely, and the ADC-built client inherits `Timeout: 0`. Was leaving that entirely to the caller a deliberate call, or would you be open to `context.WithDeadline(ctx, deadline)` at the top of the loop so the documented bound covers the requests too? (That would also remove the extra POST the loop currently issues after sleeping to exactly the deadline.)

**Two nits not worth inlining:** `if cfg.PollTimeout != 0` at `auth/gcp/client.go:113` lets a negative duration through, which silently means "one attempt, then `ErrPollTimeout`" — `> 0` would be safer. And a note for whoever reviews the follow-ups: everything here propagates up the stack, so the `ctx`-lifetime point below is worth settling at this layer.

Several candidate findings were dropped when reproduction refuted them — including a reported `time.After` timer leak (not a leak on Go 1.23+), and CRLF injection via the returned header name or token value (`net/http` rejects both at the wire).

auth/gcp/client.go
```diff
@@ -0,0 +1,300 @@
...
+		return nil, errors.New("gcp: credentials service returned an empty header or token")
+	}
+	name, scheme, _ := strings.Cut(header, ":")
+	if strings.EqualFold(strings.TrimSpace(name), "authorization") &&
+		strings.HasPrefix(strings.ToLower(strings.TrimSpace(scheme)), "bearer") {
+		return auth.BearerCredential{Token: token}, nil
+	}
+	// Non-bearer header -> header-based API key. Matches adk-python: key by the
+	// full returned header, and mirror the token into X-Goog-Api-Key too.
+	key := auth.APIKeyCredential{Name: header, Value: token}
```
karolpiotrowicz, line 293: `strings.Cut` on line 237 already gives you the header name, but this branch passes the **whole** returned string as `APIKeyCredential.Name`. If the service ever returns a non-bearer header carrying a scheme — `"X-Api-Key: Token"`, or `"Authorization: Basic"` — the result is not a legal header field name.

The failure mode is the unpleasant part: `APIKeyCredential.Apply` only checks that `Name` is non-empty, and `textproto.CanonicalMIMEHeaderKey` passes an invalid key through unchanged, so `Apply` returns nil and the credential looks fine. It then blows up much later inside `net/http` with `invalid header field name "X-Api-Key: Token"`, aborting whatever request the credential was applied to, far from the cause. (Verified locally.)

Using the already-parsed name and validating it keeps the failure where it can be understood:

```go
name = strings.TrimSpace(name)
if !httpguts.ValidHeaderFieldName(name) {
    return nil, fmt.Errorf("gcp: credentials service returned an invalid header name %q", name)
}
key := auth.APIKeyCredential{Name: name, Value: token}
```

Worth noting the reference Python implementation keys by the full header string too, so if you'd rather keep strict parity that's a reasonable answer — but then it's probably worth a comment saying so, because the current code looks like it meant to use the parsed name.

auth/gcp/connector.go
```diff
@@ -0,0 +1,82 @@
...
+		Code    int    `json:"code"`
+		Message string `json:"message"`
+	} `json:"error"`
+}
+
+// result collapses the Operation-wrapped response into an outcome.
+func (o connectorOperation) result(resource string) (outcome, error) {
+	if o.Error != nil {
+		if o.Error.Message != "" {
+			return nil, fmt.Errorf("gcp: connector operation failed (code %d): %s", o.Error.Code, o.Error.Message)
```
karolpiotrowicz: `doPost` goes to some trouble to bound what can enter an error — the 1 MiB cap plus `truncateForError` — and this path bypasses all of it. `o.Error.Message` is a response-controlled string that can be nearly the whole body budget, formatted with `%s` and neither truncated nor escaped.

With a valid 900 KB JSON body I got a **900 063-byte** Go error back, carrying raw CRLF and ANSI escapes straight through to whatever logs it. Reusing the helper that already exists two files over fixes it:

```go
return nil, fmt.Errorf("gcp: connector operation failed (code %d): %q", o.Error.Code, truncateForError(o.Error.Message))
```

auth/gcp/client.go
```diff
@@ -0,0 +1,300 @@
...
+	const maxBody = 1 << 20
+	data, err := io.ReadAll(io.LimitReader(resp.Body, maxBody+1))
+	if err != nil {
+		return fmt.Errorf("gcp: read response: %w", err)
+	}
+	if len(data) > maxBody {
+		return fmt.Errorf("gcp: credentials service response exceeded %d bytes", maxBody)
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return fmt.Errorf("gcp: credentials service returned status %d: %s", resp.StatusCode, truncateForError(strings.TrimSpace(string(data))))
```
karolpiotrowicz: This is the one error site in the file that formats untrusted bytes with `%s` — lines 152, 173 and 177 all use `%q`. `truncateForError` caps the length but doesn't escape anything, and `strings.TrimSpace` only touches the ends, so up to 1 KiB of response-chosen bytes reaches the error verbatim.

A 503 whose body is `"unavailable\r\n2026-07-24 INFO auth: credential granted user=victim status=OK\x1b[2J"` produces exactly that error string, raw control bytes and all — enough to forge a plausible line in an operator's log from the audit-relevant path of a credentials client. Switching to `%q` escapes control characters and quotes in one character, and keeps the existing cap.

auth/gcp/client_test.go
```diff
@@ -0,0 +1,412 @@
...
+		name string
+		req  Request
+	}{
+		{name: "missing resource", req: Request{UserID: "u"}},
+		{name: "missing user id", req: Request{Resource: authProviderResource}},
+		{name: "resource path traversal", req: Request{Resource: "projects/p/../q/authProviders/a", UserID: "u"}},
+		{name: "resource query injection", req: Request{Resource: "projects/p/authProviders/a?x=1", UserID: "u"}},
+		{name: "resource with space", req: Request{Resource: "projects/p/authProviders/a b", UserID: "u"}},
+	}
+	c := &Client{httpClient: http.DefaultClient}
```
karolpiotrowicz: This `Client` has no endpoint URL, so once a case gets past validation the request fails anyway with `unsupported protocol scheme ""` — which means the `err == nil` assertion below can't tell the guard from the missing URL. The three injection cases (path traversal, query injection, space) pass whether or not `RetrieveCredential` validates anything.

I checked by deleting the entire `resourceNameRE` / `".."` block: the package still passes. The guard itself is sound — I couldn't get past it with any of the eleven inputs I tried — so this is purely about it staying that way.

Pointing the client at an `httptest` server makes the test mean what it says:

```go
var hits int32
srv := httptest.NewServer(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {
    atomic.AddInt32(&hits, 1)
}))
defer srv.Close()
c := newTestClient(t, srv)
// ... then assert the error mentions "invalid characters" and that hits == 0.
```

auth/gcp/client.go
```diff
@@ -0,0 +1,300 @@
...
+		c.agentIdentityURL = strings.TrimRight(cfg.AgentIdentityEndpoint, "/")
+	}
+	if cfg.ConnectorEndpoint != "" {
+		c.connectorURL = strings.TrimRight(cfg.ConnectorEndpoint, "/")
+	}
+	if cfg.PollTimeout != 0 {
+		c.pollTimeout = cfg.PollTimeout
+	}
+	if c.httpClient == nil {
+		creds, err := google.FindDefaultCredentials(ctx, cloudPlatformScope)
```
karolpiotrowicz: Worth a sentence in `NewClient`'s doc: the `ctx` you pass here is retained by the token source that `FindDefaultCredentials` returns, and is used for every later token refresh — not just for discovery. A caller who passes a request-scoped or otherwise short-lived context gets a `Client` that works until the first token expiry and then fails every retrieval with `context canceled`, long after the constructor returned success and with an error that points at the oauth2 package rather than here.

I confirmed this for the gcloud `authorized_user` flow: cancelling the construction context and then calling `Token()` fails with `Post "https://oauth2.googleapis.com/token": … context canceled`.

This is the usual convention for `NewClient(ctx, …)` across the ecosystem, so documenting it is probably enough — but this package already documents its other subtleties (`Config.PollTimeout` explains its own bound carefully), so the omission stands out. If you'd rather make it robust instead, `context.WithoutCancel(ctx)` around the token-source construction decouples the two lifetimes. It matters more than usual here because the follow-up provider builds the client from a short-lived context.

auth/gcp/client.go
```diff
@@ -0,0 +1,300 @@
...
+	defer func() { _ = resp.Body.Close() }()
+
+	// Read one byte past the cap so an oversized body is caught explicitly rather
+	// than fed to json.Unmarshal as silently truncated (and thus garbled) JSON.
+	const maxBody = 1 << 20
+	data, err := io.ReadAll(io.LimitReader(resp.Body, maxBody+1))
+	if err != nil {
+		return fmt.Errorf("gcp: read response: %w", err)
+	}
+	if len(data) > maxBody {
```
karolpiotrowicz, line 328: This check runs before the status check below, so an oversized error page loses the single most actionable field. A gateway answering `502` with a >1 MiB HTML body surfaces `gcp: credentials service response exceeded 1048576 bytes` with no mention of the 502 (verified locally) — and it also makes `truncateForError` unreachable for exactly the "large HTML gateway response" case its own comment names.

Carrying the status through keeps both:

```go
if len(data) > maxBody {
    return fmt.Errorf("gcp: credentials service returned status %d with a response exceeding %d bytes", resp.StatusCode, maxBody)
}
```

auth/gcp/client.go
```diff
@@ -0,0 +1,300 @@
...
+		return fmt.Errorf("gcp: credentials service response exceeded %d bytes", maxBody)
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
karolpiotrowicz, line 325: Since this package's exported surface is being fixed for the first time here, it may be worth making the non-2xx case a typed error rather than a formatted string. This is the most common failure a credentials client has, and right now a caller can't distinguish a fatal 403 (wrong permissions) from a retryable 503 without substring-matching the message — which the next layer up will need to do almost immediately, and which then becomes impossible to change.

The package already exports two sentinels for rarer conditions, so a third exported type would be consistent:

```go
type StatusError struct {
    StatusCode int
    Body       string
}
func (e *StatusError) Error() string { ... }
```

returned wrapped from here so callers can use `errors.As`. Purely additive later if you'd rather not, but cheaper now.

auth/gcp/client.go
```diff
@@ -0,0 +1,300 @@
...
+	if cfg.AgentIdentityEndpoint != "" {
+		c.agentIdentityURL = strings.TrimRight(cfg.AgentIdentityEndpoint, "/")
+	}
+	if cfg.ConnectorEndpoint != "" {
+		c.connectorURL = strings.TrimRight(cfg.ConnectorEndpoint, "/")
+	}
+	if cfg.PollTimeout != 0 {
+		c.pollTimeout = cfg.PollTimeout
+	}
+	if c.httpClient == nil {
```
karolpiotrowicz, line 150: Small doc gap with a sharp edge: this is the only gate on the ADC path, so supplying `Config.HTTPClient` disables Application Default Credentials entirely. That collides with the advice in `Config.PollTimeout`'s doc a few lines up, which suggests bounding a stalled request with "an HTTPClient with its own Timeout" — a caller who follows that advice silently loses authentication and gets 401s from the service, with nothing in the godoc to explain why.

Either note it on the `HTTPClient` field ("if set, it is used verbatim and ADC is not applied; it must supply its own credentials"), or add a narrower seam — a `Config.TokenSource` or `Config.Transport` — so the timeout case doesn't have to give up ADC. The seam would also make the currently-uncovered ADC branch testable offline.

auth/gcp/client_test.go
```diff
@@ -0,0 +1,412 @@
...
+		{"connector", connectorResource, "/v1alpha/"},
+		{"auth provider", authProviderResource, "/v1/"},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			var gotPath, gotMethod, gotUserID string
+			srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				gotPath, gotMethod = r.URL.Path, r.Method
+				var body struct {
+					UserID string `json:"userId"`
```
karolpiotrowicz: The handler decodes only `userId`, so `Request.Scopes` and `Request.ContinueURI` are never asserted on the wire — the `scopes` and `continueUri` JSON tags could be wrong (or the fields dropped by `omitempty` when they shouldn't be) and everything here would still pass. `ContinueURI` in particular is what makes the 3-legged flow work, so a silent typo there would be an unpleasant one to chase.

Decoding all three and passing them in the request would close it:

```go
var body struct {
    UserID      string   `json:"userId"`
    Scopes      []string `json:"scopes"`
    ContinueURI string   `json:"continueUri"`
}
```

auth/gcp/client.go
```diff
@@ -0,0 +1,300 @@
...
+// doesn't bloat the returned error.
+func truncateForError(s string) string {
+	const max = 1024
+	if len(s) <= max {
+		return s
+	}
+	// Back up to a rune boundary so a multi-byte rune straddling the cap isn't
+	// sliced into a mangled partial rune.
+	cut := max
+	for cut > 0 && !utf8.RuneStart(s[cut]) {
```
karolpiotrowicz: Nit: this backward scan is unbounded, and the input is an untrusted network body that need not be UTF-8 at all. A well-formed sequence needs at most three steps, but if bytes 0..1024 are all continuation bytes — easy enough for a proxy returning binary or compressed content — `cut` walks all the way to 0 and the function returns a bare `"..."`, discarding every byte of diagnostic context. (`truncateForError(strings.Repeat("\x80", 2000))` returns `"..."`.)

Bounding the backup and falling back to the byte cap keeps the diagnostics:

```go
cut := max
for i := 0; i < utf8.UTFMax-1 && cut > 0 && !utf8.RuneStart(s[cut]); i++ {
    cut--
}
if !utf8.RuneStart(s[cut]) {
    cut = max // not UTF-8; cut on the byte boundary
}
```

auth/gcp/client.go
wolo-lab: Good catch, and thank you for the repro — this was a real hole, not a theoretical one. Fixed in 88ca347: the ADC-backed client now sets `CheckRedirect` to return `http.ErrUseLastResponse`, with the reason recorded on `NewClient`'s doc so it doesn't get "cleaned up" later.

`TestNewClientRefusesRedirects` pins it end to end: it drives the *real* ADC branch via a fake `GOOGLE_APPLICATION_CREDENTIALS` file and a cross-host redirector, and asserts both that the target never sees an `Authorization` header and that the 3xx surfaces as an error. Deleting the guard fails it with `auth.BearerCredential{Token:"attacker"}, nil error` — exactly what you described.

On the second half: I did not add an explicit 3xx check in `doPost`. With `ErrUseLastResponse` the 3xx already falls through the existing non-2xx check, and for a caller-supplied `Config.HTTPClient` that *follows* redirects the check can't help — `doPost` only ever sees the final 200 hop. So the redirect requirement is documented as a contract on `Config.HTTPClient` instead. Happy to add the check anyway if you'd still prefer belt-and-braces.

auth/gcp/client_test.go
wolo-lab: You're right.

Covered in 88ca347 by a `TestMapCredential` case with a genuinely different name (`X-Acme-Token`, client_test.go:332). I put it there rather than in the end-to-end table because `mapCredential` is where the mirror lives, so that's where the mutant dies: deleting the `auth.WithHeaders(...)` wrapper now fails `TestMapCredential/third-party_header_is_mirrored`

auth/gcp/client.go
wolo-lab: Agreed on the diagnosis — `APIKeyCredential.Apply` only checks for a non-empty `Name`, `CanonicalMIMEHeaderKey` passes an invalid key through, and the failure lands inside `net/http` far from here.

I took the parity branch of your suggestion rather than the parsed-name one: adk-python keys by the full returned header, and silently narrowing that could drop a header the service actually meant. So `Name` stays the full string, but 88ca347 rejects an unusable one up front  with a small `validHeaderFieldName` helper (hand-rolled because x/net is only an indirect dependency). `TestMapCredential` now pins `"X-Api-Key: Token"` as an error.

Worth flagging that this is a deliberate, narrow divergence from Python: where Python would build a credential that breaks later, we fail at the cause. I'll note it in the PR description.

auth/gcp/connector.go
wolo-lab: Agreed — that path bypassed both the cap and the escaping every other error site applies.

auth/gcp/client.go
wolo-lab: Agreed, fixed.

auth/gcp/client_test.go
wolo-lab: nice, fixed.

auth/gcp/client.go
wolo-lab: Agreed — losing the status was the wrong trade, especially since it made `truncateForError` unreachable for the exact case its comment names.

auth/gcp/client.go
wolo-lab: The collision with the `PollTimeout` doc is a good spot — that advice really did lead a caller straight into losing authentication. Documented in 88ca347 on the `HTTPClient` field (client.go:80): if set, it's used verbatim, ADC is not applied, so it must carry its own credentials and should refuse redirects for the reason `NewClient` gives.

I left out the `TokenSource`/`Transport` seam for now — it's a second way to do the same thing on a surface I'd like to keep small, and the testability motivation went away: `TestNewClientRefusesRedirects` exercises the real ADC branch offline via a fake ADC file. Easy to add later if a caller actually needs a timeout without giving up ADC.

auth/gcp/client_test.go
wolo-lab: Agreed — the handler now decodes all three fields and the test sends real values for them, asserting `scopes` and `continueUri` on the wire.

auth/gcp/client.go
wolo-lab: Nit accepted — an untrusted body needn't be UTF-8, and returning a bare "..." throws away exactly the diagnostics the helper exists to preserve. Fixed with the bounded backup and byte-cap fallback you suggested , plus a `TestTruncateForError` case for a 2000-byte all-continuation input.

auth/gcp/client.go
wolo-lab: Done — and there's a precedent that made it a smaller decision than it looked: `agentregistry/requester.go:41` already exports exactly this shape for its hand-rolled REST client. So rather than invent a second one I copied it, name included: `APIError{StatusCode, Body}`, returned from `doPost` and matched with `errors.As` in `TestRetrieveHTTPError` .

It also produced a better answer to your oversize-response comment than the patch I first wrote: classifying the status *before* the size check means an oversized error page now comes back as `*APIError{StatusCode: 502, Body: <first 1 KiB>}` rather than losing one or the other, and the bare "response exceeded" error is left for the one case that really is about size — a 2xx body too large to decode.

I left retry classification out of this PR deliberately, and it's worth naming since it's what the typed error unlocks: the loop only retries the service's `pending` state, so a transient 503 or 429 is terminal today. That's consistent with the rest of the repo (`agentregistry` doesn't retry either) and with adk-python, whose ADK layer doesn't retry at all — it delegates transport concerns to the generated `google-cloud-agentidentitycredentials` client and flattens failures to `RuntimeError`. Since we hand-roll the transport we inherit none of that, so it is a real gap — just a behavioural one that deserves its own change rather than riding along here.

auth/gcp/client.go
wolo-lab: You were right to push on this, and the "it matters more than usual here" part turned out to be literal — I reproduced it against the follow-up instead of reasoning about it. #1173's provider caches a client built as `WithTimeout(WithoutCancel(ctx), 30s)` with a `defer cancel()`, and with a fake ADC that means the *first* retrieval already fails with `context canceled` (the token is minted lazily, after the cancel). Without the cancel, the bound alone still kills it: the first call succeeds, then every call past the deadline fails permanently.

What pushed me past documenting it: #1173's code *reads* as compliant with the doc — it uses `WithoutCancel` precisely so the client outlives the request. The poison is the bound plus the deferred cancel, which "pass a context that outlives the Client" doesn't visibly cover. A doc a careful reader can satisfy and still be wrong isn't pulling its weight.

The ctx also buys less than it looks like. In x/oauth2 v0.36.0 the only network I/O in `FindDefaultCredentials` is `metadata.OnGCE()`, which calls `OnGCEWithContext(context.Background())` and bounds itself (2s dial, 5s client timeout, `sync.Once`-cached); everything else is file reads. So it never really bounded discovery — it only ended up inside the token source.

Fixed at `client.go:146` with `context.WithoutCancel`, which is what `auth/providers.go:198` already does for the same reason. It has to go at that call rather than at `oauth2.NewClient` — the latter's ctx only selects the base transport. The doc now says the ctx bounds discovery (`client.go:115`), and `TestNewClientOutlivesConstructionCtx` (`client_test.go:498`) fails with `context canceled` if the detachment is removed. The trade is that refreshes are no longer cancellable, which costs nothing here: `oauth2.TokenSource` takes no ctx, so a per-request context never bounded a refresh anyway.

auth/gcp/client.go
```diff
@@ -0,0 +1,366 @@
...
+	// It caps the retry loop, not an individual request; bound a single stalled
+	// request via ctx (or an HTTPClient with its own Timeout).
+	PollTimeout time.Duration
+}
+
+// NewClient builds a Client from cfg; a nil cfg (or any zero field) uses
+// defaults. Unless cfg.HTTPClient is set, it discovers Application Default
+// Credentials (cloud-platform scope) to authenticate calls to the services.
+//
+// ctx bounds credential discovery only: the token source backing the returned
```
karolpiotrowicz: This sentence isn't true as written. Three lines below, `FindDefaultCredentials`
is called with `context.WithoutCancel(ctx)`, so credential discovery is the one
thing `ctx` does *not* bound.

It wouldn't bound it either way: in x/oauth2 v0.36.0 every network call on that
path goes through a ctx-less variant — `metadata.OnGCE()`, memoised by a
`sync.Once`, and on GCE `metadata.ProjectID()` after it — so a caller's
cancellation was never reaching discovery to begin with.

The inline comment just below already says the accurate thing, so the two
disagree with each other. Worth correcting now rather than later, because this
is the godoc a caller reads and it ships as public API. Something like:

```go
// ctx is used for credential discovery only, and its cancellation is not
// honored: the token source backing the returned client is detached from ctx,
// so a Client built inside a request-scoped context keeps refreshing its token
// after that request ends.
```

The load-bearing half — the detachment itself — is correct, and
`TestNewClientOutlivesConstructionCtx` fails if it's removed. This is only about
the wording.

karolpiotrowicz (dismissed):
Everything from the last round is in, and — more to the point — the fixes are
held down. I checked each one by deleting it and re-running the package: the
redirect guard, the token-source detachment, the `X-Goog-Api-Key` mirror, the
resource-name guard, the header-name rejection, the `%q` escaping, the
status-before-size ordering, the `continueUri` tag and the negative
`PollTimeout` all fail a test when removed. I also went back at the URL guard
with the slash cases the charset allows — `//evil.example.com/x`, a leading
slash, a trailing slash, a doubled inner slash — and every one of them still
lands on the configured host with only `URL.Path` deformed. The typed `APIError`
came out better than what I suggested, and it's stricter than the
`agentregistry` shape it copies, which still formats its body with `%s`.

Approving. A few small things, none of them blocking. The one I'd fix before
this merges is the `ctx` sentence on `NewClient`, inline.

**One error still escapes the cap this PR applies everywhere else.**
[`client.go:283`](https://github.com/google/adk-go/blob/e6974fd21069ae0dc1c609b8e0331aff21ce29af/auth/gcp/client.go#L283)
formats the returned header name without `truncateForError`. A service that
answers with a 900 KB header name produces a 900 078-byte Go error. Same shape
as the connector-message case from last round, and the same one-word fix. `%q`
is already there, so this is size only — the token doesn't reach the error, I
checked.

**Three of the guards have no test holding them.** Deleting `truncateForError`
from
[`client.go:318`](https://github.com/google/adk-go/blob/e6974fd21069ae0dc1c609b8e0331aff21ce29af/auth/gcp/client.go#L318)
or from
[`connector.go:46`](https://github.com/google/adk-go/blob/e6974fd21069ae0dc1c609b8e0331aff21ce29af/auth/gcp/connector.go#L46)
leaves the package green, because the assertions only check that the body
contains `"nope"` and `"boom"` — `TestTruncateForError` pins the helper but
nothing pins its application. Removing the `case <-ctx.Done()` arm from
[`client.go:216`](https://github.com/google/adk-go/blob/e6974fd21069ae0dc1c609b8e0331aff21ce29af/auth/gcp/client.go#L216)
is also survivable: cancellation still surfaces through the next request, so the
outcome holds and only the promptness is lost, up to a whole `maxBackoff`. Given
two of these three are last round's fixes, an assertion on the length would stop
them quietly regressing.

**`NewClient(ctx, nil)` is documented but never called.** The nil-`Config`
branch at
[`client.go:123`](https://github.com/google/adk-go/blob/e6974fd21069ae0dc1c609b8e0331aff21ce29af/auth/gcp/client.go#L123)
backs a promise on the exported doc, and every call site in the test file passes
a non-nil `&Config{}`. `fakeADC(t)` already makes that testable in one case.

**On the question I left open last time**, about `PollTimeout` not covering the
requests themselves: it's still the case that
`RetrieveCredential(context.Background(), …)` against a peer that accepts the
POST and never answers blocks indefinitely, and that the loop issues one more
POST after sleeping to the deadline — two calls for a single-attempt budget. I'm
no longer asking for `context.WithDeadline`. `doPost` uses
`http.NewRequestWithContext`, and
[`client.go:105`](https://github.com/google/adk-go/blob/e6974fd21069ae0dc1c609b8e0331aff21ce29af/auth/gcp/client.go#L105)
already points callers at `ctx`, which is the ordinary Go contract.

One thing I got wrong last round is worth recording, though. I claimed a caller
who wants a per-request timeout has to supply `Config.HTTPClient` and thereby
lose ADC. That isn't true — `oauth2.NewClient` carries the `Timeout` through
from an `oauth2.HTTPClient` in the context, so passing one in the ctx given to
`NewClient` gets a bounded client and keeps ADC. I measured it. That's a better
answer than the `TokenSource` seam I floated, and it deserves a line on
`Config.PollTimeout` next to the existing advice, since nobody would guess it.

**One thing to carry into #1174 rather than fix here.**
[`credentialPayload`](https://github.com/google/adk-go/blob/e6974fd21069ae0dc1c609b8e0331aff21ce29af/auth/gcp/client.go#L253)
captures `{token, header}` and nothing else. That costs nothing at this layer,
which caches nothing, but a store that caches these has no expiry to evict on.
Do you know whether the services return a lifetime field? If they do, dropping
it here is the constraint the caching layer will run into, and `auth.Credential`
being `Apply(http.Header) error` means there's currently nowhere to put it.

Two smaller notes, neither worth changing on its own. The endpoint overrides at
[`client.go:133`](https://github.com/google/adk-go/blob/e6974fd21069ae0dc1c609b8e0331aff21ce29af/auth/gcp/client.go#L133)
aren't parsed, so an `http://` value would send the ADC token in the clear — the
defaults are https and this is developer config, and a scheme check would have
to exempt loopback or it would break `newTestClient` and
`TestNewClientRefusesRedirects`, so a sentence on the field is probably the right
weight. And the two services disagree on an unrecognised 200: Agent Identity
fails fast at
[`agentidentity.go:48`](https://github.com/google/adk-go/blob/e6974fd21069ae0dc1c609b8e0331aff21ce29af/auth/gcp/agentidentity.go#L48)
while the connector treats it as pending at
[`connector.go:70`](https://github.com/google/adk-go/blob/e6974fd21069ae0dc1c609b8e0331aff21ce29af/auth/gcp/connector.go#L70)
and polls to the timeout. Both are defensible, neither is pinned.

auth/gcp/client.go
wolo-lab: You're right, and it's worse than a wording slip: the sentence claimed the one thing ctx does not do. Fixed in 82b3b4fa with your wording — the godoc now says ctx is used for discovery only and its cancellation is not honored, so it agrees with the inline comment three lines below instead of contradicting it.

Worth adding to your second point: the cancellation wouldn't reach discovery even without the WithoutCancel. In x/oauth2 v0.36.0 the only network probe on that path is metadata.OnGCE(), which bounds itself and is memoised, so there was never a caller-cancellable operation for the doc to describe. It was promising a capability that didn't exist at either layer.

TestNewClientOutlivesConstructionCtx still pins the load-bearing half.

wolo-lab:
> Everything from the last round is in, and — more to the point — the fixes are held down. I checked each one by deleting it and re-running the package: the redirect guard, the token-source detachment, the `X-Goog-Api-Key` mirror, the resource-name guard, the header-name rejection, the `%q` escaping, the status-before-size ordering, the `continueUri` tag and the negative `PollTimeout` all fail a test when removed. I also went back at the URL guard with the slash cases the charset allows — `//evil.example.com/x`, a leading slash, a trailing slash, a doubled inner slash — and every one of them still lands on the configured host with only `URL.Path` deformed. The typed `APIError` came out better than what I suggested, and it's stricter than the `agentregistry` shape it copies, which still formats its body with `%s`.
> 
> Approving. A few small things, none of them blocking. The one I'd fix before this merges is the `ctx` sentence on `NewClient`, inline.
> (...)


Thanks — all of the smaller items are in as of 82b3b4fa. I used your deletion check on each one, since three of them were specifically about guards that nothing held down.

**The `ctx` sentence on `NewClient`** — fixed; replied inline.

**The uncapped header name** (`client.go:283`). Fixed — it was the last site formatting service-controlled text without the cap. `TestMapCredentialCapsHeaderNameInError` pins it: with the cap removed the error is 900 083 bytes. It also asserts the token doesn't reach the error, matching what you found.

**The three unpinned guards.** Each now fails when deleted:

- `truncateForError` in `doPost` → `Body = 1048577 bytes, want it capped to 1024`. Asserted on length in `TestDoPostOversizeKeepsStatus`, next to the existing status assertion.
- `truncateForError` on the connector operation message → `error is 900044 bytes`. New `TestRetrieveConnectorErrorMessageIsCapped`; it reaches the error by a different path than a response body, so it needed its own test.
- the `ctx` arm of the poll wait → `returned after 30.014314725s`. You were right that only promptness is lost, so the assertion is on elapsed time: the backoff is set to 30s and the call must return within 5s of cancellation.

The 1 KiB cap is now a package-level `maxErrorBody`, so the tests pin the real constant instead of a copy of it.

**`NewClient(ctx, nil)`** — covered via `fakeADC`, asserting the ADC-backed client and the three defaults.

**`oauth2.HTTPClient` in the ctx** — thanks for the correction, and it checks out: `oauth2.NewClient` copies `Timeout` (along with `Jar` and `CheckRedirect`) off the context client, so a caller does get a bounded client without giving up ADC. That's now a line on `Config.PollTimeout`. The redirect guard is unaffected — we set `CheckRedirect` *after* `NewClient`, so a ctx-supplied one can't reinstate redirect following.

**Endpoint overrides** — a sentence on the fields, at the weight you suggested: used as given, not parsed, so an `http://` value would send the token in the clear.

**The unrecognised 200** — both behaviours are now pinned as table cases, with a comment on why they differ: Agent Identity's result is a closed oneof, so an unknown arm can only be a mismatch worth failing on, whereas a connector operation that simply isn't done yet is normal and keeps being polled.

Not changed here: granted-scope reading and retry classification, both of which stay out of this PR.


### On the lifetime question

Yes — both services return one, so the constraint you anticipated is real, and #1174 is where it lands.

`RetrieveCredentialsResponse` carries `expire_time` as field 3, a `google.protobuf.Timestamp`. It sits in the `Success` message for Agent Identity v1 and on the response payload for IAM Connector v1alpha. The field's own documentation notes that when it is unset the token may be permanent, or the service may not know when it expires.

That last part settles the caching policy rather than leaving it to taste: #1174 caches **only** when an expiry is present, because an absent one means "unknown lifetime", and caching that would risk serving a stale credential indefinitely.

On "there's currently nowhere to put it" — agreed, and that's precisely why #1174 changes `RetrieveCredential` to return a `*Retrieval` carrying the expiry, rather than trying to hang it off `auth.Credential`, whose entire surface is `Apply(http.Header) error`. So `credentialPayload` grows `expireTime` there, not here.

One more field worth knowing about while we're in here: the same success payload carries `scopes` — the scopes *actually* granted, which can differ from those requested if the user declined some or the authorization server substituted a different set. Neither this PR nor #1174 reads it, and neither does adk-python, so it's an improvement beyond parity rather than a gap. I'm tracking it as a separate small change.

Finally: the push dismissed your approval (this PR targets `main`, which dismisses on new commits), so it needs another look when you have a moment.

karolpiotrowicz (approved):

PR #1150, opened 2026-07-12
feat(remoteagent): apply per-request auth to A2A via Config.Auth

wolo-lab:
## Problem

The A2A remote agent (`agent/remoteagent/v2`) had no way to attach a per-request end-user credential to outgoing A2A calls. To talk to a remote agent whose agent card declares a security requirement, the caller had to hand-build a client and wire in auth themselves. Now that the core `auth` package exists, the remote agent should resolve and attach credentials itself, per request, scoped to the caller and the callee.

## Summary

Adds `A2AConfig.Auth auth.CredentialProvider` — additive and backward-compatible. When set, `NewA2A` registers an `a2aclient.AuthInterceptor` backed by the provider, so a credential is attached to every outgoing request whose agent card declares a matching security requirement. An unexported `credentialsService` adapts `auth.CredentialProvider` to `a2aclient.CredentialsService`, returning the raw secret the interceptor transmits. It lives in `remoteagent` rather than `auth/` so the core auth package stays free of the a2a-go dependency.

`Auth` combined with a custom `ClientProvider` is a configuration error, since the interceptor cannot be injected into a caller-built client. `CredentialScope` is exported for that case, so a caller wiring their own interceptor can key on the same collision-free scope.

### Scoping

The scope handed to a session-aware provider is the whole calling identity — app name, user id, session id and the remote agent's `Name` — each percent-encoded and joined with `/`. The session id alone is caller-supplied and optional, so two users each holding a session named `default` would otherwise resolve to one credential. The agent name is part of the key because a bearer token is scoped to an audience as well as a user: one provider shared between two remote agents would otherwise cross their tokens. The context handed to the provider also remains an `agent.InvocationContext`, on the cleanup path as well as the send.

The scope covers the message send, the `CancelTask` the run loop issues when it exits before a terminal event, and the `CancelTask` the adka2a server issues for an abandoned child task. That last one reuses the subagent's client, so without the scope it went out unauthenticated and a secured remote left the task running. The scope helper therefore lives in `internal/agent/remoteagent`, which both packages can import.

### Placement

The agent card decides where the credential goes, and the a2a interceptor picks among the schemes named in one requirement object in Go map order without checking them against the credential. The adapter returns `ErrCredentialNotFound` for a scheme that cannot carry what the provider resolved, which makes selection deterministic and stops a bearer token landing in an API-key header. It also reads the fields the interceptor ignores: an API key the card wants in a query parameter or a cookie, and an HTTP scheme other than bearer, are skipped rather than approximated.

Because the card dictates placement, this is narrower than the field of the same name on `mcptoolset.Config`. Only `APIKeyCredential`, `BearerCredential` and `OAuth2Credential` can be sent, `APIKeyCredential.Name` is ignored, and a non-bearer OAuth2 token type is rejected rather than sent mislabeled. Both field docs cross-reference the other.

### Transport hardening

With `Auth` set the client refuses a redirect that leaves the card's host or downgrades its scheme, checked against the original request and the hop just taken. The credential is attached before the first hop and Go replays headers on every hop, stripping `Authorization` only when the host changes and never stripping a card-named API-key header.

An OAuth2 mint is bounded by the caller's context and by a budget of its own, since `oauth2.TokenSource.Token` takes no context and the runner sets no deadline on an invocation. An `oauth2.RetrieveError`'s verbatim response body is kept out of the message the interceptor logs at `ERROR`.

### Failure modes

Resolution is fail-open: the interceptor logs a resolution error and sends the request unauthenticated rather than failing the call. Two consequences are documented on the field. An interactive credential cannot work here, because an `*auth.ConsentRequiredError` is swallowed instead of driving a consent round-trip — use static tokens, API keys, or 2-legged or service-account sources. And a resolution failure during cleanup sends `CancelTask` unauthenticated, which a secured remote rejects, leaving the remote task running.

Where fail-open would otherwise be entirely silent, the agent warns once: a card that declares no security requirement at all, and a credential no scheme on the card can carry. A provider that is a typed nil is a constructor error rather than a panic on first use.


agent/remoteagent/v2/a2a_agent.go
```diff
@@ -245,13 +275,21 @@ func (a *a2aAgent) run(ctx agent.InvocationContext, cfg A2AConfig) iter.Seq2[*se
 			return yield(nil, err)
 		}
 
+		// When Auth is set, scope every outgoing call to the ADK session so the a2a
+		// auth interceptor resolves credentials: the message send below and the
+		// cleanup CancelTask issued from the deferred cleanupRemoteTask.
+		var sendCtx context.Context = ctx
+		if cfg.Auth != nil {
+			sendCtx = a2aclient.AttachSessionID(ctx, a2aclient.SessionID(ctx.Session().ID()))
```
karolpiotrowicz, line 283: The scope handed to the credential provider is the bare ADK session id, but an ADK session is identified by the triple (AppName, UserID, SessionID), and `SessionID` is caller-provided — `session.CreateRequest.SessionID` is optional and used verbatim, autogenerated only when empty. Two different users who both hold a session named e.g. `"default"` are therefore indistinguishable to a session-aware provider, which is exactly the pattern this field exists for and what `TestRemoteAgent_AuthScopedPerSession` demonstrates.

I ran two invocations against one agent value — `{app, alice, "default"}` and `{app, bob, "default"}` — with a provider caching on `a2aclient.SessionIDFrom(ctx)`. Both requests went out with `Authorization: Bearer tok-for-0`: bob received alice's credential.

This is reachable without any misconfiguration. `runner.Run(ctx, userID, sessionID, ...)` passes the caller's session id straight through to `Create`, the REST server takes it from the URL path, and in `server/adka2a/v2/metadata.go:62` the session id is the *incoming* A2A `ContextID` while the user id comes from authentication — so in an agent-to-agent chain the scope key is chosen by the remote caller and the one discriminator that isn't (the authenticated user) is precisely what's dropped here.

Suggestion: scope on a composite that cannot collide, e.g.

```go
s := ctx.Session()
sendCtx = a2aclient.AttachSessionID(ctx, a2aclient.SessionID(s.AppName()+"\x00"+s.UserID()+"\x00"+s.ID()))
```

and document the format so providers can parse it.

Related, and worth solving together: because `AttachSessionID` is `context.WithValue`, the context handed to the provider is no longer an `agent.InvocationContext`. A provider therefore can't type-assert its way back to the user id either — I confirmed the assertion succeeds on the no-`Auth` path and fails here. That closes off the natural workaround for the collision above, and it's also the mechanism `auth.CredentialProvider`'s own doc points at ("recovers the ADK context from ctx"). Preserving the concrete type while adding the value would fix both:

```go
type authContext struct {
    agent.InvocationContext
    values context.Context
}
func (c authContext) Value(k any) any { return c.values.Value(k) }
```

agent/remoteagent/v2/auth.go
```diff
@@ -0,0 +1,88 @@
...
+	case auth.BearerCredential:
+		if v.Token == "" {
+			return "", errors.New("remoteagent: a2a auth requires a bearer token credential")
+		}
+		return v.Token, nil
+	case auth.OAuth2Credential:
+		if v.TokenSource == nil {
+			return "", errors.New("remoteagent: oauth2 credential missing token source")
+		}
+		tok, err := v.TokenSource.Token()
```
karolpiotrowicz, line 77: `oauth2.TokenSource.Token()` takes no context and `credentialValue` has none to give it, so this call cannot be interrupted. Interceptors run before the transport call, so the client's own `http.Client.Timeout` doesn't bound it — and neither does the 5s budget `cleanupRemoteTask` sets at `a2a_agent.go:376`, because the mint happens *inside* that call.

Because the deferred cleanup runs synchronously inside the run-loop iterator, a consumer that simply `break`s out of `Run` blocks on it. With a token source that succeeds once and then blocks, `Run` was still blocked after 8 seconds despite the documented 5s cleanup budget, and returned only once the source was released. The cancel then failed with `context deadline exceeded`, so the remote task was leaked as well.

This isn't an exotic path: `auth.ServiceAccount` resolves to the JWT token source in `golang.org/x/oauth2`, which posts to the token endpoint through `http.DefaultClient` — no timeout and no request context. Against a black-holed endpoint I measured `Token()` still blocked after 10s. Both `auth.ADC()` and `auth.ServiceAccount()` are recommended in the `Auth` field doc.

Suggestion: thread the `ctx` from `Get` into `credentialValue` and bound the mint against it — run `Token()` in a goroutine and `select` on `ctx.Done()`, which is the same shape `auth.lazyTokenSource` already uses. It would also be worth sizing the cleanup budget to cover credential resolution plus the RPC, rather than the RPC alone.

agent/remoteagent/v2/auth.go
```diff
@@ -0,0 +1,88 @@
...
+		if err != nil {
+			return "", fmt.Errorf("remoteagent: mint oauth2 token: %w", err)
+		}
+		if tok == nil || tok.AccessToken == "" {
+			return "", errors.New("remoteagent: oauth2 token source returned an empty access token")
+		}
+		return tok.AccessToken, nil
+	default:
```
karolpiotrowicz, line 85: Three credential shapes a caller can legitimately obtain from the public `auth` package land in this `default` branch:

- `auth.BasicCredential`
- whatever `auth.WithHeaders` returns (an unexported wrapper type, so it can never match this switch and its inner credential is discarded with it)
- pointer forms such as `&auth.BearerCredential{...}` — pointers satisfy `auth.Credential` because every `Apply` has a value receiver

The error returned here is then swallowed by the fail-open interceptor, so the request goes out with no credential at all. I ran all three end to end: in each case the server received the request with an empty `Authorization`, `Run` returned `nil`, and no error event was emitted. Nothing at all reaches the caller.

`auth.WithHeaders(oauth2Cred, map[string]string{"x-goog-user-project": ...})` is the auth package's documented pattern for exactly this kind of call, so it's a likely thing for someone to wire in.

Fail-open itself is clearly deliberate and I'm not arguing with it — the gap is that this consequence isn't accounted for. Suggestions, roughly in order of value:

1. Unwrap what can be unwrapped: give the `WithHeaders` wrapper an accessor (or an `interface{ Inner() Credential }` seam in `auth`) and add pointer cases to the switch.
2. Enumerate the supported set in the `Auth` field doc. Right now it appears only in this error string, which the caller never sees.
3. Consider making an unsupported kind visible somewhere — a `Warn` from `Get` including the agent name, at minimum — so a misconfiguration doesn't vanish entirely into fail-open.

agent/remoteagent/v2/auth.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"google.golang.org/adk/v2/auth"
+)
+
+// credentialsService adapts an [auth.CredentialProvider] to
+// [a2aclient.CredentialsService]. The a2a AuthInterceptor calls Get and places
+// the returned value per the agent card's security scheme (it adds the "Bearer "
+// prefix or the API-key header itself), so Get returns the raw secret.
+//
+// The provider is not scheme-aware; it yields the same credential for whichever
```
karolpiotrowicz, line 33: "whichever scheme the card lists first" doesn't hold. `a2a.SecurityRequirements` is a `map[SecuritySchemeName]SecuritySchemeScopes`, and the interceptor selects with `for schemeName := range requirement`, so within a single requirement object the choice is Go's randomized map order — there is no first. Only the outer `SecurityRequirementsOptions` slice is ordered.

With a card declaring `{apikey, bearer}` in one requirement object — a shape the a2a library's own type docs give as an example — I saw the same secret land in two different headers across 40 identical invocations: 33 in `X-Api-Key`, 7 in `Authorization`. So a bearer token is emitted as an API key on a random subset of requests, and the resulting failures are intermittent in a way no config inspection explains.

Two options:

- Minimum: reword to say that when a requirement object names more than one scheme the choice is arbitrary and may vary per request, and that a caller needing determinism should publish one scheme per requirement object.
- Better: use the `a2a.SecuritySchemeName` argument `Get` already receives and return `a2aclient.ErrCredentialNotFound` for schemes the resolved credential doesn't match (`APIKeyCredential` to the apiKey scheme, `BearerCredential`/`OAuth2Credential` to the HTTP-auth/OAuth2 schemes). The interceptor treats that sentinel as "try the next scheme", so selection becomes deterministic and the doc becomes true.

A two-scheme test case would keep it that way.

agent/remoteagent/v2/a2a_agent.go
```diff
@@ -157,8 +178,17 @@ func NewA2A(cfg A2AConfig) (agent.Agent, error) {
...
+	}
 	if cfg.ClientProvider == nil {
-		cfg.ClientProvider = NewA2AClientProvider(a2aclient.NewFactory())
+		var opts []a2aclient.FactoryOption
+		if cfg.Auth != nil {
+			opts = append(opts, a2aclient.WithCallInterceptors(&a2aclient.AuthInterceptor{
+				Service: credentialsService{provider: cfg.Auth},
+			}))
+		}
+		cfg.ClientProvider = NewA2AClientProvider(a2aclient.NewFactory(opts...))
```
karolpiotrowicz, line 191: This keeps the library's default transports, which build an `http.Client` with a nil `CheckRedirect` — so up to 10 redirects are followed and the initial request's headers are replayed on each hop. Go strips only `Authorization`, `Www-Authenticate`, `Cookie`, `Cookie2` and the two proxy variants, and only when the destination *hostname* differs; scheme and port are not considered. A card-declared API-key header name is never on that list.

I confirmed both halves. With an apiKey card, a 307 from the declared endpoint to a different host delivered `X-Api-Key: <secret>` to that host. With a bearer card, a redirect that changed only the port still carried `Authorization`. (Cross-host bearer is correctly stripped by Go, so that half is fine.)

To be fair about impact: the card's declared endpoint already holds the credential before it can redirect, so this isn't a new disclosure to a party that didn't have it. The residual risks are an `https` to `http` downgrade on the same host putting the secret on the wire in cleartext, and simply not being able to bound where it goes. What makes it awkward is that `Auth` and `ClientProvider` are mutually exclusive, so a caller can't install a hardened `CheckRedirect` themselves.

Suggestion: when `Auth` is set, build the transports with an explicit client whose `CheckRedirect` refuses a scheme or host change — A2A JSON-RPC has no legitimate need for redirects — passing `a2aclient.WithJSONRPCTransport` / `WithRESTTransport` alongside `WithCallInterceptors`. Alternatively, add a config field so callers can supply their own client without giving up `Auth`.

agent/remoteagent/v2/a2a_agent.go
```diff
@@ -245,13 +275,21 @@ func (a *a2aAgent) run(ctx agent.InvocationContext, cfg A2AConfig) iter.Seq2[*se
 			return yield(nil, err)
 		}
 
+		// When Auth is set, scope every outgoing call to the ADK session so the a2a
+		// auth interceptor resolves credentials: the message send below and the
+		// cleanup CancelTask issued from the deferred cleanupRemoteTask.
+		var sendCtx context.Context = ctx
+		if cfg.Auth != nil {
```
karolpiotrowicz, line 282: Gating `AttachSessionID` on `cfg.Auth != nil` makes the remediation in the error at line 182 ("wire the credential into your ClientProvider instead") a trap for the obvious implementation. I built a factory with the library's own `a2aclient.AuthInterceptor` and a credentials store pre-populated under the real ADK session id, passed it as `ClientProvider`, and the request went out with an empty `Authorization` — `SessionIDFrom` finds nothing, so the interceptor returns before attaching anything. No error, no log from this package.

There is a way through: because `sendCtx == ctx` on that path, a custom `A2AClient` can type-assert the context it receives and call `AttachSessionID` itself before delegating (I verified this works). But that isn't discoverable from the error message.

Suggestion: drop the condition and attach the session id unconditionally — it's an inert context value when nothing reads it — which makes the documented escape hatch work as written. Failing that, extend the error message to say the custom client must attach the session id itself.

agent/remoteagent/v2/auth.go
```diff
@@ -0,0 +1,88 @@
...
+// header) — one more reason to enable Auth only for trusted cards.
+type credentialsService struct {
+	provider auth.CredentialProvider
+}
+
+var _ a2aclient.CredentialsService = credentialsService{}
+
+// Get implements [a2aclient.CredentialsService].
+func (s credentialsService) Get(ctx context.Context, _ a2aclient.SessionID, _ a2a.SecuritySchemeName) (a2aclient.AuthCredential, error) {
+	cred, err := s.provider.Credential(ctx)
```
karolpiotrowicz, line 46: `cfg.Auth != nil` is an interface comparison, so a typed nil — `auth.ProviderFunc(nil)`, or a nil pointer whose methods have pointer receivers — passes every gate in `NewA2A` and then panics here on the first request. I confirmed `invalid memory address or nil pointer dereference` originating from `Get`, which means it surfaces inside `SendMessage` and inside the deferred cleanup, where a panic while the run loop is already unwinding masks the original outcome.

`credentialValue` already guards a nil *credential* at line 61, so a matching guard for a nil *provider* is cheap — or a `recover()` in `Get` that converts it into the error fail-open already handles. Either is a better failure mode than an SDK panicking on a misconfiguration two lines away from where `NewA2A` returns a clean error for the `ClientProvider` conflict.

Worth noting this footgun isn't unique to this field — `auth.Transport` behaves the same way with a typed-nil provider — so a guard in the `auth` package would cover both call sites at once.

agent/remoteagent/v2/auth_test.go
```diff
@@ -0,0 +1,520 @@
...
+	card := newSecureCard(srv.URL,
+		a2a.NamedSecuritySchemes{
+			"apikey": a2a.APIKeySecurityScheme{Location: a2a.APIKeySecuritySchemeLocationHeader, Name: "X-Api-Key"},
+		},
+		a2a.SecurityRequirementsOptions{
+			{a2a.SecuritySchemeName("apikey"): a2a.SecuritySchemeScopes{}},
+		},
+	)
+
+	remoteAgent, err := NewA2A(A2AConfig{Name: "a2a", AgentCard: card, Auth: auth.APIKey("X-Api-Key", "secret")})
```
karolpiotrowicz, line 251: The card declares `Name: "X-Api-Key"` (line 244) and the provider here is `auth.APIKey("X-Api-Key", "secret")`, so this assertion passes whether the header name came from the card or from the credential. The real behavior is that it comes exclusively from the card — `credentialValue` never reads `APIKeyCredential.Name` — and that's one of the two headline behaviors of the feature, so it's worth pinning.

I verified the distinction by using different names: with the card declaring `X-Card-Key` and the credential `auth.APIKey("X-Caller-Key", "secret")`, the server saw `X-Card-Key: secret` and nothing in `X-Caller-Key`.

Suggestion: use distinct names and assert both sides — `X-Card-Key == "secret"` and `X-Caller-Key == ""`. That turns the test into a real statement of "the card decides placement", which is also what the type doc in `auth.go` claims.

agent/remoteagent/v2/auth.go
```diff
@@ -0,0 +1,88 @@
...
+	value, err := credentialValue(cred)
+	if err != nil {
+		return "", err
+	}
+	return a2aclient.AuthCredential(value), nil
+}
+
+// credentialValue returns the raw secret the a2a AuthInterceptor transmits: the
+// API-key value, the bearer token, or a freshly minted OAuth2 access token.
+func credentialValue(c auth.Credential) (string, error) {
```
karolpiotrowicz, line 59: The SDK already has an `Auth auth.CredentialProvider` field on `mcptoolset.Config`, and it behaves differently for the same value. That one routes through `auth.Transport`, which calls `cred.Apply` — so it supports every `auth.Credential` implementation, honours `APIKeyCredential.Name` and the OAuth2 token type, and fails closed (`RoundTrip` returns the error). This one reimplements a subset of `Apply` by type switch: three of the five implementations, `Name` and the token type discarded, and fail-open.

Same field name, same type, a narrower supported set and the opposite failure mode, with neither doc mentioning the other. Someone who has a working `CredentialProvider` against a toolset and moves it to a remote agent gets a silently unauthenticated request rather than an error.

The failure mode is a deliberate choice here and the card-driven placement is inherent to the protocol, so the ask is mostly documentation: note on this field that the agent card dictates placement (hence the credential's own `Name` and token type are ignored) and that the supported set is narrower than `auth.Credential`'s, and ideally cross-reference the two fields so the difference is discoverable from either side.

agent/remoteagent/v2/auth_test.go
```diff
@@ -0,0 +1,520 @@
...
+				return nil, errors.New("boom")
+			}),
+			wantErr: true,
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			svc := credentialsService{provider: tc.provider}
+			got, err := svc.Get(t.Context(), a2aclient.SessionID("sid"), a2a.SecuritySchemeName("scheme"))
+			if tc.wantErr {
```
karolpiotrowicz, line 105: All six `wantErr` rows assert only `err != nil` and then return, so the distinct error branches are interchangeable — collapsing `credentialValue` to a single `errors.New("nope")` for every non-happy input would keep this test green. That also means the `%w` wrapping at `auth.go:48`, the only place the provider's own error survives for an operator, is unasserted.

Two branches have no covering row at all (the coverage run flags exactly these): the empty `BearerCredential.Token` rejection at `auth.go:69-71`, and the `TokenSource.Token()` error at `auth.go:78-80`. The second is the realistic failure mode for `auth.ADC()` and `auth.ServiceAccount()`, the sources the field doc recommends. The table has an "empty api key value" row but no bearer analogue, which looks accidental.

Suggestions:

- Add a `wantErrContains` (or exported sentinels) per row, and `errors.Is` for the provider-error row.
- Add `{name: "empty bearer token", provider: auth.StaticToken(""), wantErr: true}`.
- Add a row with a `TokenSource` that returns an error.
- Cheap bonus for the "no credential in error strings" property: give one row a distinctive value and assert `!strings.Contains(err.Error(), ...)`.

karolpiotrowicz (commented):
The shape is right (additive opt-in field, session-scoped resolution, cleanup path authenticated too) and the test file is unusually thorough for a first pass. I dug into it fairly deeply because it puts credentials on the wire, and I found a few things worth resolving before this lands; the first three are the ones I'd hold on.

**Blocking**

- **The credential scope key can collide across users** (`agent/remoteagent/v2/a2a_agent.go:283`). The scope is the bare ADK session id, but an ADK session is the triple (AppName, UserID, SessionID) and the session id is caller-provided. Two users holding a session named `"default"` get the same credential. Inline comment has the reproduction and a suggested composite key.

**Worth fixing before merge**

- **A hung token mint can't be interrupted** (`agent/remoteagent/v2/auth.go:77`) — not by the invocation context, not by the 5s cleanup budget, not by the client's own timeout. It blocks `Run` itself.
- **Some legitimate `auth.Credential` values silently become unauthenticated requests** (`agent/remoteagent/v2/auth.go:85`) — `auth.BasicCredential`, anything from `auth.WithHeaders`, and pointer forms of the supported types. Nothing reaches the caller.

**Smaller points, inline**

Doc accuracy on multi-scheme cards, redirect hardening, the mutual-exclusion error's remediation advice, a typed-nil provider panic, the API-key test's discriminating power, the divergence from `mcptoolset.Config.Auth`, and two uncovered error branches.

**Notes that didn't seem worth their own comment**

- The provider is now called concurrently (I measured 8 in-flight calls across 8 concurrent invocations of one agent) and may be called more than once per outgoing request (a card declaring 20 unusable schemes produced 20 `Credential` calls for a single request — token *mints* stay at 1 thanks to `ReuseTokenSource` caching, so this is mostly a provider-invocation cost). Neither is stated on the field doc; one sentence would cover both.
- No test in the package starts two invocations concurrently, so the green `-race` run doesn't actually cover the shared-interceptor case. Writing that test currently trips over `serveRecordingA2A` sharing one event slice with a replay helper that stamps `TaskID`/`ContextID` into those events per request — worth fixing together.
- `fmt.Errorf("...: %w", err)` at `agent/remoteagent/v2/auth.go:79` propagates `oauth2.RetrieveError`, whose message embeds the verbatim token-endpoint response body; the interceptor logs that at ERROR. I saw a planted body value make it into the error string intact. Whether a real provider's error body carries anything sensitive is provider-dependent, but a redacted wrap here would remove the question.
- Nothing requires TLS before the credential is attached — a card naming an `http://` interface gets it in cleartext. Same absence exists on the `auth.Transport` path, so this may be a package-level decision rather than yours.
- `OAuth2Credential`'s token type is dropped, so a non-bearer token would be transmitted mislabeled. In practice `oauth2.Token.Type()` returns `Bearer` when unset and the interceptor hardcodes the prefix anyway, so rejecting a non-bearer type is probably the only sensible option — very low priority.
- The recorders in the new tests overwrite rather than accumulate, so each asserts on the last request only. Correct today (one call per scenario), fragile if a scenario ever produces two.

**A question rather than a claim**

- When resolution fails during cleanup, the `CancelTask` goes out unauthenticated and the remote task stays alive — which is the failure `TestRemoteAgent_AuthAttachedToCleanupCancel`'s comment says the cleanup auth exists to prevent. I can see the argument that this is just fail-open behaving as documented, since `cleanupRemoteTask` already tolerates any cancel failure with a `log.Warn`. Is that the intent, or would you want the fail-open case called out in the `Auth` doc as "a resolution failure at cleanup can leave the remote task running"?

For what it's worth, everything objective is green: `go build`, the full `go test ./...`, `go test -race` on the package, `gofmt`, `go vet`, `golangci-lint`, and `staticcheck`. `apidiff` reports the change as compatible (`A2AConfig.Auth: added`, no incompatible changes), and I confirmed no existing caller of `NewA2A` sets the new field or can reach the new validation error. The new tests correctly fail to compile against the merge base, so they do exercise new API.

agent/remoteagent/v2/a2a_agent.go
wolo-lab: Fixed, and reproduced first — with a provider caching on the scope, alice and bob both holding a session named `default` sent `Bearer tok-0`, so bob got alice's credential.

The scope is now the whole calling identity rather than the session id: app name, user id, session id and the remote agent's `Name`, each percent-encoded and joined with `/` ([CredentialScope](https://github.com/google/adk-go/blob/9c9ac81725faf065005835961a15e4b0b0b3f54e/internal/agent/remoteagent/auth_scope.go#L43-L47)). Percent-encoding rather than a NUL join, because the parts must survive being logged and split back out, and `url.QueryEscape` escapes both `/` and `%` so no two identities can produce the same key.

The agent name is in there for a reason your comment doesn't cover but is the same class of bug. A bearer token is scoped to an audience as well as to a user, so one provider shared between two remote agents in one session would hand the second agent the token minted for the first. `TestCredentialScope` pins the format with literal strings, including the `a/b` + `c` versus `a` + `b/c` pair that a naive join collapses.

Your second half is fixed too, and it turned out to matter more than the workaround it closes. `authSendContext` now returns an `agent.InvocationContext` whose values come from the a2a chain ([auth.go](https://github.com/google/adk-go/blob/9c9ac81725faf065005835961a15e4b0b0b3f54e/agent/remoteagent/v2/auth.go#L113-L124)), and `reattachInvocation` re-wraps after the deferred cleanup detaches the context ([auth.go](https://github.com/google/adk-go/blob/9c9ac81725faf065005835961a15e4b0b0b3f54e/agent/remoteagent/v2/auth.go#L126-L136)) — without that the type assertion succeeded on the send and failed on the cancel, which is where an unauthenticated request leaks a remote task.

One thing I did not keep: the scope is no longer attached when `Auth` is unset. Attaching it unconditionally would have revived a dormant `AuthInterceptor` a caller had installed but never fed a session id, putting a credential on the wire for someone who never opted in. `CredentialScope` is exported instead, so the ClientProvider path can build the same key deliberately.

agent/remoteagent/v2/auth.go
wolo-lab: Fixed. The mint runs in a goroutine and selects on the context ([mintAccessToken](https://github.com/google/adk-go/blob/9c9ac81725faf065005835961a15e4b0b0b3f54e/agent/remoteagent/v2/auth.go#L328-L364)), so the 5s cleanup budget now actually bounds resolution plus the RPC rather than the RPC alone.

It also has a budget of its own, 30s, mirroring `auth.initTimeout`. Your reproduction went through the cleanup path, which supplies a deadline, but the send path is the common one and the runner sets no deadline on an invocation at all — so on that path the inherited bound you'd be relying on does not exist.

What this does not do is cancel `Token()`, which cannot be interrupted. The goroutine stays parked until the source gives up. That is no worse than the synchronous call it replaces, which parked the caller's goroutine instead, but it is not a fix and the comment says so rather than implying otherwise.

agent/remoteagent/v2/auth.go
wolo-lab: Fixed for the pointer forms, and I went a different way on `WithHeaders`.

Pointer credentials now work — `derefCredential` unwraps them ([auth.go](https://github.com/google/adk-go/blob/9c9ac81725faf065005835961a15e4b0b0b3f54e/agent/remoteagent/v2/auth.go#L271-L287)), and a typed nil pointer is left alone so it lands in the untransmittable branch instead of panicking.

I did not unwrap `WithHeaders`. The a2a credentials interface hands back a single string, so there is nowhere for the extra headers to go: unwrapping would silently drop the `x-goog-user-project` that was the whole reason for using it, which is a worse failure than refusing. It is rejected with a message naming the type, and the supported set is now on the `Auth` field doc where a caller reads it rather than only in an error string they never see.

On visibility, the interceptor does log the error, but the case that produced no signal at all was the one you did not hit: a credential no scheme on the card can carry returns `ErrCredentialNotFound`, which the interceptor swallows without logging. That now warns once per agent, and stays quiet when the card offers an alternative the credential does fit.

agent/remoteagent/v2/auth.go
wolo-lab: You're right that there is no first, and I took the second option.

`Get` now checks the resolved credential against the card's scheme and returns `ErrCredentialNotFound` when it cannot be carried ([schemeAccepts](https://github.com/google/adk-go/blob/9c9ac81725faf065005835961a15e4b0b0b3f54e/agent/remoteagent/v2/auth.go#L228-L252)), so selection is deterministic and the doc is true rather than reworded. The card comes from the context, since the interface hands `Get` only the scheme name.

Two more fields the interceptor ignores are checked there as well. An API key the card wants in a query parameter or a cookie would be written as a header the remote never reads, and a card declaring `Basic` would receive `Bearer` — the same mislabelling the OAuth2 token-type check two cases away already refused. Both are skipped now, as are mutual TLS and OpenID Connect, and an absent card or an undeclared scheme name denies rather than allows.

The two-scheme test is there: `TestRemoteAgent_AuthMultiSchemeCardIsDeterministic` runs 40 invocations against a single requirement object naming both and asserts the same placement every time. Before the fix it failed on run 0 with the token in `X-Api-Key`.

agent/remoteagent/v2/a2a_agent.go
wolo-lab: Fixed. With `Auth` set the client is built with an explicit `CheckRedirect` ([auth.go](https://github.com/google/adk-go/blob/9c9ac81725faf065005835961a15e4b0b0b3f54e/agent/remoteagent/v2/auth.go#L386-L404)) rather than a config field, so callers get the hardening without having to give up `Auth` to obtain it.

The policy refuses any hop that changes the host or downgrades the scheme, and permits an http-to-https upgrade because it strictly improves confidentiality. It is checked against the original request and against the hop just taken — against the original alone, a chain could climb to https and drop back to http with every individual hop looking same-origin. Host comparison is case-insensitive with the default port filled in, so `https://h` and `https://h:443` are one origin.

A custom `CheckRedirect` replaces the standard 10-hop cap rather than extending it, so that is restated too, otherwise a same-origin redirect loop would spin until the request timeout.

Your framing of the residual risk is what decided the shape: the declared endpoint already holds the credential, so what is worth bounding is where it can travel next, not whether it is disclosed at all.

agent/remoteagent/v2/a2a_agent.go
wolo-lab: Fixed, though not by dropping the condition.

Attaching the scope unconditionally would have revived an `AuthInterceptor` a caller had installed but never fed a session id — dormant before, firing after — which puts a credential on the wire for someone who never opted into this field. With `Auth` unset the context is now left completely alone, dynamic type included ([authSendContext](https://github.com/google/adk-go/blob/9c9ac81725faf065005835961a15e4b0b0b3f54e/agent/remoteagent/v2/auth.go#L113-L124)).

What makes the remediation work instead is `CredentialScope`, now exported. A caller wires their own interceptor and attaches the same collision-free key themselves. The error message says that, and names the fact your comment turned up the hard way: the interceptor reads the context of each call, not the one the provider was built with, so the scope has to go on per call rather than at construction.

agent/remoteagent/v2/auth.go
wolo-lab: Fixed at the constructor rather than in `Get` ([a2a_agent.go](https://github.com/google/adk-go/blob/9c9ac81725faf065005835961a15e4b0b0b3f54e/agent/remoteagent/v2/a2a_agent.go#L225-L227)), so the misconfiguration is reported next to the one it already reported cleanly, and reported before it can panic anywhere.

It uses reflection, which is heavier than I would like for a nil check, but it is the only way to see a nil inside a non-nil interface and it runs once per agent. The check is ordered before the ClientProvider conflict check, so a config that is broken both ways gets the diagnosis that is still true after fixing the other one.

`Get` keeps a plain nil-provider guard as well. Agreed that a guard in `auth` would cover both call sites, and `auth.Transport` has the same hole — worth doing there separately rather than in this change.

agent/remoteagent/v2/auth_test.go
wolo-lab: Fixed. The card now declares `X-Card-Key` and the credential names `X-Caller-Key`, and the test asserts the secret arrives in the first and that the second is absent — so it states "the card decides placement" rather than passing either way.

The same distinction turned out to be load-bearing elsewhere: the scheme check added for the multi-scheme comment now reads `APIKeySecurityScheme.Location` too, since an API key the card wants in a query parameter would otherwise go out as a header under a name the remote never reads.

agent/remoteagent/v2/auth.go
wolo-lab: Documented on both sides, and the gap narrowed a little in the process.

The `Auth` field now says the card dictates placement — hence `APIKeyCredential.Name` and the token type are ignored — enumerates what cannot be sent and why, states the fail-open consequences including the one at cleanup, and points at `mcptoolset.Config.Auth`. That field carries the reciprocal note, so the difference is visible from either side.

Two of the divergences you listed are gone rather than documented. The supported set now includes pointer forms, and a scheme this path cannot satisfy is skipped rather than approximated. What remains is inherent: the protocol hands over one secret and the card decides where it goes.

The silent-unauthenticated outcome you describe is the part I was least happy leaving as prose, so it also warns now, once per agent, when no scheme on the card can carry what the provider resolved.

agent/remoteagent/v2/auth_test.go
wolo-lab: Fixed, all four.

Every row now carries a `wantErrContains` so the branches are no longer interchangeable, and the provider-error row asserts `errors.Is` against a sentinel, which is what keeps the `%w` wrapping load-bearing. The empty bearer token and the `TokenSource` error both have rows.

Your cheap bonus is in as `wantErrAbsent`, on the two rows whose credential actually holds a secret — a non-bearer OAuth2 token and a `BasicCredential` password. That one earned its place immediately: it is how I noticed `oauth2.RetrieveError` prints the token endpoint's response body verbatim, and only when the response was not a well-formed OAuth2 error, which is exactly when the content is least predictable. That branch is rewritten now ([redactTokenError](https://github.com/google/adk-go/blob/9c9ac81725faf065005835961a15e4b0b0b3f54e/agent/remoteagent/v2/auth.go#L474-L490)) with the status and the error chain kept.

wolo-lab:
All ten comments are addressed in the latest push ([9c9ac81](https://github.com/google/adk-go/commit/9c9ac81725faf065005835961a15e4b0b0b3f54e)). Detail is on each thread — this is the shape of the change and the parts that go beyond what you raised.

## The three you held on

The scope collision reproduced exactly as you described, and the fix goes further than a composite key: the scope now carries the remote agent's name as well, because a bearer token is scoped to an audience and one provider shared between two remote agents in a session would otherwise cross their tokens. The mint is bounded by the caller's context and by a budget of its own, since the send path — unlike your cleanup reproduction — inherits no deadline from the runner. Pointer credentials are accepted, and `WithHeaders` is refused rather than unwrapped: the protocol hands over one string, so unwrapping would silently drop the header that was the reason for using it.

## Three things your comments led to that they did not name

**The adka2a server issues a second CancelTask, and it was unauthenticated.** `cancelChildInputRequiredTasks` reuses the subagent's auth-wired client with a context that never carried a scope, so the interceptor returns before resolving anything and a secured remote rejects the cancel — leaking the child task, which is the failure the cleanup auth exists to prevent. It now attaches the scope and the card ([executor.go](https://github.com/google/adk-go/blob/9c9ac81725faf065005835961a15e4b0b0b3f54e/server/adka2a/v2/executor.go#L322-L340)), which is why the credential scope helper moved to `internal/agent/remoteagent`: the server needs the identical key and cannot import `agent/remoteagent/v2`.

**Redirect hardening had to be checked against the previous hop, not just the original.** Every hop of `http → https → http` looks same-origin against the original request, so checking only that would let a chain climb to TLS and drop back to cleartext.

**The remediation in the mutual-exclusion error did not work as written.** A caller reading the scope inside their `ClientProvider` gets nothing, because the interceptor reads the context of each call rather than the one the provider was built with. The message now says to attach it per call, and `CredentialScope` is exported so the key is the same one this package uses.

## Your notes-without-comments

The concurrency observation was right and there is now a test that starts eight concurrent invocations of one agent value with distinct users and asserts the exact set of credentials the server saw, so `-race` covers the shared interceptor. It needed the replay helper fixed first, as you predicted — the per-request task and context ids were being stamped into one shared event slice.

`oauth2.RetrieveError` was the note I'm most glad you wrote down. It prints the response body verbatim, and only when the response was not a well-formed OAuth2 error, so the case where the body reaches an ERROR log is the case where its content is least predictable. Redacted, with the status and the error chain kept.

The multiple-calls-per-request and concurrency facts are on the field doc now, as is the cleanup answer to your question: a resolution failure there does send CancelTask unauthenticated and can leave the remote task running. That is fail-open behaving as documented, and it is worth a caller knowing before they choose a provider that can fail.

Two I did not take. Requiring TLS before attaching stays documented rather than enforced, since it would break local development and every in-process test server. And the token type is dropped no longer — a non-bearer type is rejected outright, which as you said is the only sensible option.

## Checks

`go build`, `go vet`, `gofmt`, `golangci-lint`, `go mod tidy -diff` and `go test -race -count=1 -shuffle=on ./...` are clean. `apidiff` reports `A2AConfig.Auth: added` plus the new `CredentialScope`, no incompatible changes. Coverage on the package went 89.2% to 91.0%.

agent/remoteagent/v2/auth.go
```diff
@@ -0,0 +1,495 @@
...
+	return fmt.Errorf("remoteagent: cannot send %T over a2a, where the agent card decides placement; "+
+		"only auth.APIKeyCredential, auth.BearerCredential and auth.OAuth2Credential are supported", c)
+}
+
+// schemeAccepts reports whether the card's security scheme named name can carry
+// a credential written as p. It default-denies: an absent card, or a name the
+// card does not declare, yields false, so the interceptor moves on without the
+// adapter having minted anything for a scheme that cannot be satisfied.
+func schemeAccepts(card *a2a.AgentCard, name a2a.SecuritySchemeName, p placement) bool {
+	if card == nil {
```
karolpiotrowicz, line 229: The nil-card default-deny here is the fail-safe that stops the adapter handing the raw secret to a scheme name it knows nothing about, and inverting it to `return true` leaves the whole suite green. It is also the uncovered branch behind this function's 85.7%. Reachability is thin on the mainline — the credential-bearing cancel gets the real card re-attached first — so this is test debt on a fail-closed guard rather than a live hole. A row in `TestCredentialsServiceGetMatchesScheme` that calls `Get` with no card on the context and expects `ErrCredentialNotFound` would cover it.

agent/remoteagent/v2/auth.go
```diff
@@ -0,0 +1,495 @@
...
+// [a2aclient.SessionIDFrom]. agentName is the remote agent's A2AConfig.Name.
+//
+// It is exported for the case A2AConfig.Auth cannot serve: a caller combining a
+// custom ClientProvider with their own [a2aclient.AuthInterceptor] attaches
+// this themselves, and gets a key that cannot collide across users, sessions or
+// remote agents.
+//
+// The parts are percent-encoded and joined with "/". A provider that wants them
+// back splits on "/" and calls [url.QueryUnescape] on each.
+func CredentialScope(s session.Session, agentName string) a2aclient.SessionID {
```
karolpiotrowicz, line 50: `CredentialScope` dereferences `s` with no guard, so `CredentialScope(nil, name)` panics. That matters more than the usual nil-argument case because the doc directly invites external callers to use this for the custom-`ClientProvider` path, and it is the one exported function this change adds.

agent/remoteagent/v2/auth.go
```diff
@@ -0,0 +1,495 @@
...
+func (e *redactedError) Unwrap() error { return e.cause }
+
+// redactTokenError strips the token endpoint's verbatim response body from an
+// [oauth2.RetrieveError]. The a2a interceptor logs whatever this adapter
+// returns at ERROR level, and what an identity provider echoes into a non-2xx
+// body is outside our control — some reflect the request back.
+//
+// Only the branch that carries the body is rewritten: RetrieveError.Error()
+// prints the body exactly when the response was not a well-formed OAuth2 error,
+// which is also when its content is least predictable. A response that did
```
karolpiotrowicz, line 471: "A response that did name an error code already prints only that" understates it. At x/oauth2 v0.36.0 the `ErrorCode != ""` branch prints the code, `ErrorDescription` and `ErrorURI`. The redaction itself is correct — that branch never touches `Body`, so nothing verbatim from the response escapes — but `ErrorDescription` is still free text chosen by the token endpoint, and the comment reads as though it were just the code. `TestRedactTokenError` leaves both fields empty, so it cannot catch the difference.

agent/remoteagent/v2/auth.go
```diff
@@ -0,0 +1,495 @@
...
+		cause: err,
+	}
+}
+
+// isTypedNil reports whether v is a nil pointer, func or map held in a non-nil
+// interface — auth.ProviderFunc(nil), say. Such a value passes an ordinary
+// != nil check and then panics on the first call.
+func isTypedNil(v any) bool {
+	switch rv := reflect.ValueOf(v); rv.Kind() {
+	case reflect.Chan, reflect.Func, reflect.Map, reflect.Pointer, reflect.Slice, reflect.UnsafePointer:
```
karolpiotrowicz, line 490: The doc two lines up says "a nil pointer, func or map", but the switch also covers `Slice`, `Chan` and `UnsafePointer`. A provider implemented on a nil named map or slice with a value receiver is callable and safe, and `NewA2A` now rejects it. `Chan` is arguably right to reject, since a receive on a nil channel blocks forever. Either way the comment and the code should say the same thing.

agent/remoteagent/v2/auth.go
```diff
@@ -0,0 +1,495 @@
...
+	defer cancel()
+
+	type result struct {
+		tok *oauth2.Token
+		err error
+	}
+	// Buffered, so a mint nobody is waiting for can still complete and let its
+	// goroutine exit rather than blocking forever on the send.
+	done := make(chan result, 1)
+	go func() {
```
karolpiotrowicz, line 342: Worth considering a single-flight here. Bounding the wait releases the caller, so with a hung token endpoint the next invocation starts another mint while this goroutine is still parked, where the previous synchronous call throttled itself by holding the caller. `auth/providers.go`'s own `lazyTokenSource` already carries a single-flight for the same situation. The parked goroutine also runs third-party `Token()` with no `recover()`, so a panicking source takes the process down.

agent/remoteagent/v2/a2a_agent.go
```diff
@@ -517,7 +622,10 @@ func cleanupRemoteTask(ctx context.Context, cfg A2AConfig, card *a2a.AgentCard,
 		return
 	}
 
-	ctx = context.WithoutCancel(ctx)
+	// WithoutCancel returns its own type, which is no longer an
+	// agent.InvocationContext; re-wrap so a credential provider can still
+	// recover the ADK context here, exactly as it can on the send path.
+	ctx = reattachInvocation(ctx, context.WithoutCancel(ctx))
```
karolpiotrowicz, line 628: `reattachInvocation` runs whether or not `Auth` is set, because it keys only on the context already being an `agent.InvocationContext`, which it is on the unset path. So a caller who never opted in now gets an `authContext` in `RemoteTaskCleanupCallback` and in a custom `CancelTask` where they previously got what `context.WithoutCancel` returns, and `ctx.(agent.InvocationContext)` flips from false to true. Every method delegates, so semantics are identical and I could not construct a consumer this breaks — but it contradicts the claim at auth.go:107-109 that an opted-out caller sees no change to the context's dynamic type, and the callback's own doc at line 383. Gating on `cfg.Auth != nil` costs nothing here since `cleanupRemoteTask` already takes `cfg`.

karolpiotrowicz (changes requested):
Round 1's blocker is properly closed and I can show it: reverting only the scope change turns `TestRemoteAgent_AuthScopeIsPerUser` and `TestRemoteAgent_AuthConcurrentInvocations` red, and a generated run over 400,000 identity tuples containing `/`, `%`, `%2F` and NUL produced no colliding key. The scheme-placement check, the redirect policy and the context-type preservation on cleanup all hold up under deliberate attempts to break them. What I would fix before this merges is one behaviour and four guards that no test can currently fail on.

### A card with an empty requirement object goes out unauthenticated, silently

The warning at [a2a_agent.go:466](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/agent/remoteagent/v2/a2a_agent.go#L466) tests `len(card.SecurityRequirements) == 0`. A card that declares a scheme and one **empty** requirement object has length 1, so the warning does not fire, and the a2a interceptor's inner `for schemeName := range requirement` has nothing to iterate, so `Get` is never called. The request leaves with no credential and nothing is logged anywhere. `security: [{}]` is the OpenAPI spelling of "authentication optional", so a real card can carry it.

<details><summary>Reproduction — drop into <code>agent/remoteagent/v2</code> and run <code>go test -run TestEmptyRequirementObject ./agent/remoteagent/v2/</code></summary>

```go
func TestEmptyRequirementObject(t *testing.T) {
	var mu sync.Mutex
	var authz []string
	srv := serveFreshA2A(t, "ok", func(r *http.Request) {
		mu.Lock()
		authz = append(authz, r.Header.Get("Authorization"))
		mu.Unlock()
	})
	card := newSecureCard(srv.URL,
		a2a.NamedSecuritySchemes{"bearer": a2a.HTTPAuthSecurityScheme{Scheme: "Bearer"}},
		a2a.SecurityRequirementsOptions{{}}, // one EMPTY requirement object
	)
	called := 0
	prov := auth.ProviderFunc(func(ctx context.Context) (auth.Credential, error) {
		called++
		return auth.BearerCredential{Token: "secret-token"}, nil
	})
	a, err := NewA2A(A2AConfig{Name: "a2a", AgentCard: card, Auth: prov})
	if err != nil {
		t.Fatalf("NewA2A: %v", err)
	}
	warns := &countingHandler{match: "a2a auth"}
	ctx := newInvocationContextFor(t, t.Name(), "u", "s")
	scoped := ctx.WithContext(log.AttachLogger(ctx, slog.New(warns)))
	if _, err := runAndCollect(scoped, a); err != nil {
		t.Fatalf("run: %v", err)
	}
	mu.Lock()
	defer mu.Unlock()
	t.Logf("provider called %d times, Authorization=%q, warnings=%d", called, authz, warns.count.Load())
}

// observed: provider called 0 times, Authorization=[""], warnings=0
```
</details>

The invariant worth holding is that every path which reaches the wire without a credential emits exactly one signal, rather than that this particular shape is special-cased.

### Four guards no test can fail on

I checked these by neutering each one and re-running `./agent/remoteagent/... ./server/adka2a/... ./internal/agent/remoteagent/...`. Twenty-one other guards in this change do fail their tests when broken, which is why these four stand out rather than reading as ordinary coverage debt.

**[a2a_agent.go:420](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/agent/remoteagent/v2/a2a_agent.go#L420) — `OwnsAuthScope: cfg.Auth != nil`.** Changing it to `false` leaves the entire suite green. It is the only line joining the user-facing `Auth` field to the server-side cancel path, so if it were ever wrong, every `Auth` user's adka2a-issued cancel for an abandoned child task would go out unauthenticated — the exact failure [executor.go:350](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/server/adka2a/v2/executor.go#L350) was added to fix. The executor tests cannot call `NewA2A` because of the import cycle and hand-set the field instead, so nothing anywhere asserts the derivation.

**[auth.go:51](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/agent/remoteagent/v2/auth.go#L51) — the exported `CredentialScope` wrapper.** Swapping `s.UserID()` and `s.ID()` survives, and so does dropping `agentName` entirely. [auth_test.go:274](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/agent/remoteagent/v2/auth_test.go#L274) pins the internal four-string function against literals, but every test that touches the exported wrapper computes both the expectation and the actual value from it. The two mutations matter for different reasons: the argument swap desynchronizes this path from [executor.go:336](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/server/adka2a/v2/executor.go#L336), which builds the key from the internal function directly, so a provider caching per scope would resolve the cancel under a different key than the send. Dropping `agentName` reinstates the cross-agent token leak that component was added to prevent. One row driving the exported wrapper with three distinct literals would close both.

**[executor.go:350-351](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/server/adka2a/v2/executor.go#L350-L351) — the card re-attachment.** Deleting it keeps both executor tests green, because [cancel_auth_test.go:101](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/server/adka2a/v2/cancel_auth_test.go#L101) installs `a2aclient.NewInMemoryCredentialsStore()`, which resolves on (scope, scheme name) and never looks at the card. Production always uses this package's own `credentialsService`, which does read it — with no card on the context `schemeAccepts` returns false, the cancel goes out unauthenticated and the remote task leaks. Separately, the config those tests build sets `OwnsAuthScope: true` alongside a caller-supplied `ClientProvider`, and [a2a_agent.go:397](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/agent/remoteagent/v2/a2a_agent.go#L397) makes that combination unreachable, so the fixture is a shape `NewA2A` cannot produce.

**[a2a_agent.go:466-476](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/agent/remoteagent/v2/a2a_agent.go#L466-L476) — `warnNoRequirement`.** Deleting the whole block leaves the suite green and no test matches its message, unlike its sibling `warnMismatch`. This is the same block as the finding above.

### The `ClientProvider` documentation describes a state that cannot occur

[a2a_agent.go:315-318](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/agent/remoteagent/v2/a2a_agent.go#L315-L318) tells the reader that the context a `ClientProvider` receives carries the credential scope and is an `agent.InvocationContext` wrapper. But [a2a_agent.go:397](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/agent/remoteagent/v2/a2a_agent.go#L397) rejects `Auth` together with `ClientProvider`, so whenever a caller-supplied provider runs, `Auth` is nil, and [auth.go:114](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/agent/remoteagent/v2/auth.go#L114) returns the invocation context untouched — no scope and no wrapper. `TestRemoteAgent_AuthUnsetLeavesTheContextAlone` asserts precisely that. Someone who trusts this paragraph writes `a2aclient.SessionIDFrom(ctx)` in their provider and gets `ok == false` on every call, which is the opposite of what the error message at line 397 sends them to do. The wrapper half of the sentence does become true on one call, the cleanup `CancelTask`, which makes the paragraph harder to spot as wrong rather than easier.

### Already true before this change

Worth knowing, not for this PR to fix. An `AgentCardProvider` returning `(nil, nil)` panics on a nil dereference — I checked, and `main` panics identically, one frame further in, because `a2aclient.Factory.CreateFromCard` dereferences the card on its first line. The adka2a identity derivation at [metadata.go:62](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/server/adka2a/v2/metadata.go#L62) and the unbounded serial cancel loop are likewise base behaviour, though the loop now carries credential resolution and so has a larger worst case than it did.

I also chased the worry that giving the client an explicit `http.Client` newly caps SSE streams at three minutes. It does not: in a2a-go v2.5.0 the JSON-RPC and REST transports each build `&http.Client{Timeout: defaultRequestTimeout}` when handed no client, which is the same application site, so the comment at [auth.go:365-368](https://github.com/google/adk-go/blob/2ff5dae228b91eb0d4ae462f7b6e51378d87b163/agent/remoteagent/v2/auth.go#L365-L368) is exactly right.

### What I would want before merge

A test that fails if `OwnsAuthScope` stops tracking `Auth`, one that drives the exported `CredentialScope` with three distinct literals, one that exercises the executor cancel through this package's own credentials adapter rather than the in-memory store, and a decision on the empty-requirement case — either widen the check or say in the field doc that a card shaped that way is treated as needing no credential. The documentation correction on `ClientProvider` matters as much as the tests, since that paragraph actively sends a reader the wrong way. Everything under the inline comments is optional.

PR #1151, opened 2026-07-13
fix(sequentialagent): correct New godoc placement and copy-pasted error string

dmora:
The `// New creates a SequentialAgent.` doc comment is attached to the `seqAgent` type — the live-mode changes inserted the type declaration between the comment and `func New`, leaving `New` with only the orphaned "Use the SequentialAgent when..." fragment. This PR reunites the full doc comment with `New` and gives `seqAgent` its own comment.

It also fixes the custom-Run guard error string, which reports `LoopAgent doesn't allow custom Run implementations` from inside `sequentialagent.New` (copy-paste from loopagent — parallelagent names itself correctly).

No behavior change beyond the error text. `go build ./agent/workflowagents/...`, `go vet`, and `go test ./agent/workflowagents/...` pass; no test asserts the old string.

Found while integrating v1.5.0 into our fork.


wolo-lab (approved):
LGTM, thanks!

PR #1155, opened 2026-07-15
feat: support multi-module Go development in CI and update developer documentation

hanorik:
This PR transitions the repository to a multi-module workspace to prevent heavy/optional dependencies from bloating the core ADK module. 

**Key Changes:**
* **Dynamic CI Matrix**: Upgraded `.github/workflows` to dynamically discover all `go.mod` files and run tests, linters, and `govulncheck` independently via matrix strategies.
* **Anti-Bloat Guardrail**: Added a custom CI job preventing the core `go.mod` from accidentally importing internal submodules.
* **Documentation**: Added "Multi-Module Development" guidelines to `CONTRIBUTING.md` and `AGENTS.md`.

**Testing Plan**:
* Verified CI matrix discovery and bash guardrail scripts locally.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -19,33 +19,87 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test:
+  discover:
     runs-on: ubuntu-latest
+    outputs:
+      modules: ${{ steps.set-modules.outputs.modules }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - id: set-modules
+        run: |
+          MODULES=$(find . -not -path '*/\.git/*' -name go.mod -exec dirname {} \; | sort | jq -R -s -c 'split("\n")[:-1]')
+          echo "modules=$MODULES" >> $GITHUB_OUTPUT
+
+  test:
+    needs: discover
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        module: ${{ fromJson(needs.discover.outputs.modules) }}
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
     
     - name: Setup
       uses: ./.github/actions/setup
 
-    - name: Tidy
+    - name: Initialize workspace
+      run: go work init && go work use -r .
+
+    - name: Tidy Check
+      working-directory: ${{ matrix.module }}
       run: go mod tidy -diff
       
     - name: Build
+      working-directory: ${{ matrix.module }}
       run: go build -mod=readonly -v ./...
 
     - name: Test
+      working-directory: ${{ matrix.module }}
       run: go test -race -mod=readonly -v -count=1 -shuffle=on ./...
   
   lint:
+    needs: discover
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        module: ${{ fromJson(needs.discover.outputs.modules) }}
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
     
     - name: Setup
       uses: ./.github/actions/setup
+
+    - name: Initialize workspace
+      run: go work init && go work use -r .
     
     - name: Lint
       uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20  # v9.2.0
       with:
+        working-directory: ${{ matrix.module }}
         install-mode: goinstall
         version: 5256574b81bcedfbcae9099f745f6aee9335da10 # v2.3.1
+
+  guardrail:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+    
+    - name: Setup
+      uses: ./.github/actions/setup
+
+    - name: Check root go.mod against internal modules
+      run: |
+        ROOT_REQS=$(go mod edit -json | jq -r '.Require // [] | .[] | .Path')
+        
+        FAIL=0
+        while IFS= read -r -d '' dir; do
+          MOD_PATH=$(cd "$dir" && go mod edit -json | jq -r '.Module.Path')
+          if [ -n "$MOD_PATH" ] && echo "$ROOT_REQS" | grep -F -x -q "$MOD_PATH"; then
+            echo "ERROR: Root go.mod requires internal submodule $MOD_PATH"
+            FAIL=1
+          fi
+        done < <(find . -mindepth 2 -not -path '*/\.git/*' -name go.mod -printf '%h\0')
+        
+        exit $FAIL
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -7,19 +7,42 @@ on:
   workflow_dispatch:
 
 jobs:
-  test:
+  discover:
     runs-on: ubuntu-latest
+    outputs:
+      modules: ${{ steps.set-modules.outputs.modules }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - id: set-modules
+        run: |
+          MODULES=$(find . -not -path '*/\.git/*' -name go.mod -exec dirname {} \; | sort | jq -R -s -c 'split("\n")[:-1]')
+          echo "modules=$MODULES" >> $GITHUB_OUTPUT
+
+  test:
+    needs: discover
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        module: ${{ fromJson(needs.discover.outputs.modules) }}
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
     
     - name: Setup
       uses: ./.github/actions/setup
 
+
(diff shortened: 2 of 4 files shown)


.github/workflows/nightly.yml
```diff
@@ -7,19 +7,44 @@ on:
...
+      fail-fast: false
+      matrix:
+        module: ${{ fromJson(needs.discover.outputs.modules) }}
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
     
     - name: Setup
       uses: ./.github/actions/setup
 
+    - name: Initialize workspace
```
wolo-lab: Maybe we should remove it?
Why: once we have a submodule that imports the
core module, a workspace resolves that import from the local checkout instead of
the version pinned in the submodule's `go.mod`. So CI stays green even if the
submodule's `go.mod` is missing the `require` on core, or its `go.sum` is
incomplete - exactly the failures a real `go get <submodule>@<tag>` consumer would
hit. Building in isolation exercises the same resolution path the consumer sees.

wolo-lab (dismissed):
LGTM. One comment from my side.

Please wait for the approval from Hana.

.github/workflows/nightly.yml
hanorik: Verified locally and it seems relevant to remove it. Good catch!

wolo-lab (dismissed):

AGENTS.md
```diff
@@ -104,6 +106,28 @@ See `examples/quickstart` for a full runnable program.
 - **Add cross-cutting behavior:** register a `plugin.New(plugin.Config{...})`
   hook (`Before*`/`After*` for run/agent/model/tool) instead of editing the loop.
 
+## Multi-Module Development
+
+**Policy**: New integrations with heavy or optional dependencies must be created as separate Go modules.
+
+**Steps to Add a New Module (e.g., `plugin/myplugin`)**:
+1. Navigate into the directory: `cd <module_directory_path>`
+2. Initialize the module: `go mod init google.golang.org/adk/v2/<module_directory_path>`
```
hyangah: Should it be `google.golang.org/adk/<module_directory_path>` or `google.golang.org/adk/v2/<module_directory_path>`?

If it goes with `google.golang.org/adk/v2/<module_directory_path>`, can we check if the tag should be `v2/<module_directory_path>/v0.1.0` or `<module_directory_path>/v0.1.0`? (I am not sure either. Need testing.)

AGENTS.md
```diff
@@ -104,6 +106,28 @@ See `examples/quickstart` for a full runnable program.
...
+3. Add your Go code, dependencies, and tests.
+4. Tidy the module: `go mod tidy`
+5. Return to the repo root.
+6. Tidy the root module: `go mod tidy`
+7. Verify everything builds and tests from the root: `go build ./... && go test ./...`. The CI will automatically pick up the new module on the PR.
```
hyangah: `./...` doesn't cross the module boundary.  Maybe suggest `go build work`?

https://github.com/golang/go/issues/71294

AGENTS.md
hanorik: `google.golang.org/adk/v2` as it's `module google.golang.org/adk/v2` in the root module rn. Will verify the tag, thanks!

AGENTS.md
hanorik: Good point, changed!

CONTRIBUTING.md
```diff
@@ -43,6 +44,30 @@ To work on a 1.x fix, base your branch on `v1`:
...
+**Local Development**: Contributors should use `go work init && go work use -r .` to set up their local workspaces.
+
+**Steps to Add a New Module (e.g., `plugin/myplugin`)**:
+1. Navigate into the directory: `cd <module_directory_path>`
+2. Initialize the module: `go mod init google.golang.org/adk/v2/<module_directory_path>`
+3. Add your Go code, dependencies, and tests.
+4. Tidy the module: `go mod tidy`
+5. Return to the repo root.
+6. Tidy the root module: `go mod tidy`
+7. Verify everything builds and tests from the root: `go build ./... && go test ./...`. The CI will automatically pick up the new module on the PR.
```
karolpiotrowicz: It looks like the `./...` → `work` fix from the review landed in `AGENTS.md` but this duplicated copy didn't get updated. Run from the repo root, `go build ./...` / `go test ./...` stop at any nested `go.mod`, so they'd silently skip the new submodule — the `work` pattern (Go 1.25+) is what actually covers it.

```suggestion
7. Verify everything builds and tests from the root: `go build work && go test work`. The CI will automatically pick up the new module on the PR.
```

.github/workflows/go.yml
```diff
@@ -19,33 +19,87 @@ concurrency:
...
       uses: ./.github/actions/setup
 
-    - name: Tidy
-      run: go mod tidy -diff
+    - name: Initialize workspace
+      run: go work init && go work use -r .
+
+    - name: Tidy Check
+      working-directory: ${{ matrix.module }}
+      run: go mod tidy && git diff --exit-code
```
karolpiotrowicz: `go mod tidy && git diff --exit-code` only inspects tracked files, so if a new module produces an **untracked** `go.sum`, the check passes silently. `go mod tidy -diff` is non-mutating, catches `go.sum` changes too, and drops the extra `git diff` step:

```suggestion
      run: go mod tidy -diff
```

(If you'd rather keep the mutate-then-verify form, `git status --porcelain` instead of `git diff --exit-code` would also catch untracked files.)

.github/workflows/go.yml
```diff
@@ -19,33 +19,87 @@ concurrency:
...
+        
+        FAIL=0
+        for dir in $SUBMODULES; do
+          MOD_PATH=$(cd "$dir" && go mod edit -json | jq -r '.Module.Path')
+          if echo "$ROOT_REQS" | grep -q -x "$MOD_PATH"; then
+            echo "ERROR: Root go.mod requires internal submodule $MOD_PATH"
+            FAIL=1
+          fi
+        done
+        exit $FAIL
```
karolpiotrowicz: A few robustness edge cases in the guardrail loop (none bite today with a single module, but worth hardening before the first submodule lands):

- `for dir in $SUBMODULES` is unquoted, so it word-splits on any path containing a space.
- `grep -q -x "$MOD_PATH"` matches the module path as a **regex** — the `.`s match any character; `grep -Fxq` compares it as a literal string.
- If `MOD_PATH` is empty and the root has no requires, `grep -qx ""` matches an empty line and reports a false positive.

Switching the loop to a null-delimited `while IFS= read -r -d ''` (fed by `find … -printf '%h\0'`), plus `grep -Fxq` and a non-empty guard on `MOD_PATH`, covers all three.

karolpiotrowicz (commented):
The "Multi-Module Development" section is duplicated verbatim between `AGENTS.md` and `CONTRIBUTING.md`, and the two copies have already drifted (step 7 differs: `go build work` in `AGENTS.md` vs `go build ./...` in `CONTRIBUTING.md`). To keep them from diverging further, could we keep a single source of truth — e.g. `CONTRIBUTING.md` owns the section and `AGENTS.md` links to it (or vice-versa)? That way there's only one place to update when the workflow changes.

A couple of specific inline notes below.

AGENTS.md
hanorik: I verified and in the end it should be `google.golang.org/adk/<module_directory_path>` so the tag should be `<module_directory_path>/v0.1.0`.

CONTRIBUTING.md
hanorik: Yes, forgot to change it there as well. Fixed.

.github/workflows/go.yml
hanorik: Fixed.

.github/workflows/go.yml
hanorik: Applied corresponding fixes, good catch!

AGENTS.md
```diff
@@ -104,6 +106,28 @@ See `examples/quickstart` for a full runnable program.
 - **Add cross-cutting behavior:** register a `plugin.New(plugin.Config{...})`
   hook (`Before*`/`After*` for run/agent/model/tool) instead of editing the loop.
 
+## Multi-Module Development
+
+**Policy**: New integrations with heavy or optional dependencies must be created as separate Go modules.
+
+**Steps to Add a New Module (e.g., `plugin/myplugin`)**:
```
wolo-lab: Instead of duplication we should maybe just point the section from CONTRIBUTING.md? Currently there's a risk of getting out of sync in the next update.

CONTRIBUTING.md
```diff
@@ -43,6 +44,30 @@ To work on a 1.x fix, base your branch on `v1`:
...
+
+**Local Development**: Contributors should use `go work init && go work use -r .` to set up their local workspaces.
+
+**Steps to Add a New Module (e.g., `plugin/myplugin`)**:
+1. Navigate into the directory: `cd <module_directory_path>`
+2. Initialize the module: `go mod init google.golang.org/adk/<module_directory_path>`
+3. Add your Go code, dependencies, and tests.
+4. Tidy the module: `go mod tidy`
+5. Return to the repo root.
+6. Tidy the root module: `go mod tidy`
```
wolo-lab, line 59: Should we have a step between 6 and 7: " Add the module to your workspace: `go work use ./<module_directory_path>`"?

wolo-lab (dismissed):

AGENTS.md
hanorik: SG, changed.

CONTRIBUTING.md
hanorik: You are right, added.

wolo-lab (approved):

hyangah (approved):

PR #1159, opened 2026-07-16
feat(session): add GetUserState for parity with adk-python

### LINKED ISSUE #1158: session.Service is missing GetUserState (parity with adk-python)
### Description

adk-python's `BaseSessionService` defines `get_user_state(*, app_name, user_id) -> dict[str, Any]` (`adk-python/src/google/adk/sessions/base_session_service.py:114-152`), which returns the raw (un-prefixed) user-scoped state for an `(app_name, user_id)` pair **without** requiring an active `session_id`. Per its docstring, this exists so callers can bootstrap context before `create_session`, instead of doing an expensive `list_sessions` + `get_session` just to read user-scoped data.

It's implemented concretely in:
- `in_memory_session_service.py:315-318` — trivially returns `dict(self.user_state.get(app_name, {}).get(user_id, {}))`
- `sqlite_session_service.py:364-369`
- `database_session_service.py:725+`
- `vertex_ai_session_service.py:364-380` — explicitly raises `NotImplementedError` (not supported by that backend)

### Problem in adk-go

`session.Service` (`session/service.go:25-32`) has no equivalent method — only `Create/Get/List/Delete/AppendEvent`. Notably, the storage already exists for this in the in-memory backend: `inMemoryService.userState map[string]map[string]stateMap` (`session/inmemory.go:42`) is populated by `updateUserState` but has no standalone read accessor. Similarly, `session/database/service.go` already has `fetchStorageUserState(tx, appName, userID)` (line 450) used internally by `Create`/`Get`/`List`, which could back a `GetUserState` directly.

Today, a Go caller who wants to read a user's persisted state before creating a session has to call `List` then `Get` on every returned session just to access merged user-scoped state — an unnecessarily expensive pattern the Python API explicitly documents avoiding.

### Suggested fix

- Add `GetUserState(ctx context.Context, req *GetUserStateRequest) (*GetUserStateResponse, error)` to `session.Service`.
- Implement in `session/inmemory.go` using the existing `userState` map.
- Implement in `session/database/service.go` using the existing `fetchStorageUserState` helper.
- In `session/vertexai/vertexai.go`, return an explicit "not supported" error, mirroring adk-python's Vertex AI backend.
- Update `server/adkrest/internal/fakes/testsessionservice.go` fake to satisfy the interface.

### Environment

- adk-go: main @ cbe2604
- Found by comparing against adk-python (`base_session_service.py`, `in_memory_session_service.py`, `database_session_service.py`)

mdabydeen:

- Closes: #1158

**2. Problem / Solution:**

**Problem:** adk-python's `BaseSessionService` exposes `get_user_state(app_name, user_id)` to read user-scoped state without requiring an active session — useful for bootstrapping context before `create_session` without an expensive `list_sessions` + `get_session` round trip. `session.Service` in adk-go has no equivalent, even though the in-memory backend already stores this data internally (`inMemoryService.userState`) and the database backend already has a helper (`fetchStorageUserState`) that could back it directly.

**Solution:**
- Add `GetUserState(ctx, *GetUserStateRequest) (*GetUserStateResponse, error)` to the `session.Service` interface.
- Implement it in `session/inmemory.go` using the existing `userState` map.
- Implement it in `session/database/service.go` using the existing `fetchStorageUserState` helper.
- In `session/vertexai/vertexai.go`, return an explicit "not supported" error, mirroring adk-python's `VertexAiSessionService.get_user_state`, which also doesn't support this.
- Update the `FakeSessionService` test double in `server/adkrest/internal/fakes` to satisfy the interface.

### Testing Plan

**Unit Tests:**

Added a shared `GetUserState` battery to `session/sessiontestsuite`, gated behind a new `SuiteOptions.SupportsGetUserState` flag, covering: empty state for unknown users, reading un-prefixed keys set both at session-create time and via a later `AppendEvent` state delta, and per-user isolation. Opted in for the in-memory and database backends (both of which now pass it). Added a standalone unit test asserting the Vertex AI backend returns an error, since that backend's contract-test run is driven by recorded gRPC fixtures I can't regenerate without live GCP credentials.

```
ok  	google.golang.org/adk/v2/session	...
ok  	google.golang.org/adk/v2/session/database	...
ok  	google.golang.org/adk/v2/session/vertexai	...
... (all other packages ok)
```

`go build`, `go vet`, and `go mod tidy -diff` are all clean.

N/A — internal service API addition, covered by unit/contract tests above.

### Checklist

- [ ] I have manually tested my changes end-to-end. (N/A, see above)
- [x] Any dependent changes have been merged and published in downstream modules. (N/A)


karolpiotrowicz:
@wolo-lab this needs a compatibility decision before a code review is worth doing.

The change adds `GetUserState` to the exported `session.Service` interface. Go has no default interface method, so every out-of-tree implementation stops compiling. Nothing in the repository breaks, because both in-tree implementations are updated here — the entire break lands outside. The alternatives are an optional extension interface with a type assertion at the call site, or an embeddable default. Which of those we want is a project call.

PR #1161, opened 2026-07-16
feat(memory): add AddEventsToMemory for incremental ingestion

### LINKED ISSUE #1160: memory.Service is missing AddEventsToMemory (incremental ingestion, parity with adk-python)
### Description

adk-python's `BaseMemoryService` defines `add_events_to_memory(*, app_name, user_id, events, session_id=None, custom_metadata=None)` (`adk-python/src/google/adk/memory/base_memory_service.py:64-100`) to let callers persist a specific subset of events (e.g. only the latest turn) as an incremental delta, instead of re-ingesting the whole session via `add_session_to_memory`.

`in_memory_memory_service.py:75-101` implements it: it appends only events with non-empty content, deduping by `event.id` against whatever's already stored for that `(app_name, user_id, session_id)` key — so calling it repeatedly with overlapping events is idempotent.

`vertex_ai_memory_bank_service.py:235+` also implements it against the real Memory Bank API.

### Problem in adk-go

`memory.Service` (`memory/service.go:31-37`) only exposes `AddSessionToMemory` and `SearchMemory` — there's no way to ingest a subset of events without re-adding/re-scanning the whole session, and `memory/inmemory.go`'s `AddSessionToMemory` (line 60) doesn't have any per-event dedup logic at all (it just wholesale-replaces the stored slice for that session ID each call: `v[sid] = values`).

### Suggested fix

- Add `AddEventsToMemory(ctx context.Context, req *AddEventsToMemoryRequest) error` to `memory.Service`, with fields mirroring Python's kwargs (`AppName`, `UserID`, `Events []*session.Event`, `SessionID` optional, `CustomMetadata` optional).
- Implement in `memory/inmemory.go`: append only events with non-empty `LLMResponse.Content`, dedup by `event.ID` against what's already stored for that session-scope key (mirroring `in_memory_memory_service.py:87-101`).
- In `memory/vertexai/vertexai.go`, either port the real ingestion call (bigger) or return an explicit "not supported" error for now, to be filled in as a follow-up — whichever the PR author judges best scoped.

### Environment

- adk-go: main @ cbe2604
- Found by comparing against adk-python (`base_memory_service.py`, `in_memory_memory_service.py`)

mdabydeen:

- Closes: #1160

**2. Problem / Solution:**

**Problem:** adk-python's `BaseMemoryService.add_events_to_memory(*, app_name, user_id, events, session_id=None, custom_metadata=None)` lets callers persist a subset of events (e.g. just the latest turn) as an incremental delta, instead of re-ingesting the whole session via `add_session_to_memory`. `memory.Service` in adk-go has no equivalent — `AddSessionToMemory` is the only ingestion path, and it wholesale-replaces the stored events for a session ID on every call (no per-event dedup at all).

**Solution:**
- Add `AddEventsToMemory(ctx, *AddEventsToMemoryRequest) error` to `memory.Service`, with fields mirroring Python's kwargs (`AppName`, `UserID`, `Events`, optional `SessionID`, optional `CustomMetadata`).
- Implement in `memory/inmemory.go`: extracted the event→`value` conversion (skip events with no text content) into a shared `eventToValue` helper reused by both `AddSessionToMemory` and the new method, and dedup incoming events by `event.ID` against whatever's already stored for that `(app, user, session)` scope — mirroring `in_memory_memory_service.py`'s dedup behavior, so repeated calls with overlapping events are idempotent.
- `memory/vertexai/vertexai.go` returns an explicit "not supported" error for now. adk-python's Vertex AI Memory Bank backend does implement this against the real ingestion API, but porting that is a distinctly larger, separate effort from this interface addition — happy to follow up separately if useful.

### Testing Plan

**Unit Tests:**

Added `Test_inMemoryService_AddEventsToMemory` covering: events become searchable, repeated calls with overlapping events are deduped by ID, isolation across sessions/users, and that contentless events are ignored (matching existing `AddSessionToMemory` semantics). Added a standalone unit test asserting the Vertex AI backend's explicit unsupported error.

```
ok  	google.golang.org/adk/v2/memory	...
ok  	google.golang.org/adk/v2/memory/vertexai	...
... (all other packages ok)
```

`go build`, `go vet`, and `go mod tidy -diff` are all clean.

N/A — internal service API addition, covered by unit tests above.

### Checklist

- [ ] I have manually tested my changes end-to-end. (N/A, see above)
- [x] Any dependent changes have been merged and published in downstream modules. (N/A)


memory/inmemory.go
```diff
@@ -106,6 +84,72 @@ func (s *inMemoryService) AddSessionToMemory(ctx context.Context, curSession ses
...
+	}
+	existing := sessions[sid]
+
+	existingIDs := make(map[string]struct{}, len(existing))
+	for _, v := range existing {
+		existingIDs[v.id] = struct{}{}
+	}
+
+	for _, event := range req.Events {
+		if _, ok := existingIDs[event.ID]; ok {
```
karolpiotrowicz, line 138: `event.ID` is the dedup key with no guard for the empty string, so the first ID-less event claims `""` and every later one in this batch — and in every later call for this `(app, user, session)` scope — hits the `continue` below. The method still returns `nil`. Details and a reproduction are in the review body.

memory/inmemory.go
```diff
@@ -106,6 +84,72 @@ func (s *inMemoryService) AddSessionToMemory(ctx context.Context, curSession ses
...
+			continue
+		}
+		v, ok := eventToValue(event)
```
karolpiotrowicz, line 141: `eventToValue` runs here, inside the write lock taken at line 122, while `AddSessionToMemory` calls the same helper at line 63 — before its lock at line 73. `s.mu` guards the whole service across every app and user, and `SearchMemory` holds the read lock for its entire scan, so tokenizing caller text under the exclusive lock blocks unrelated tenants for the duration.

Non-blocking, and the net effect is probably still an improvement, since this tokenizes one turn where `AddSessionToMemory` re-tokenizes the whole session. I did not measure contention, so treat this as matching the sibling method's shape rather than as a demonstrated slowdown. Building the values before taking the lock costs nothing.

memory/inmemory_test.go
```diff
@@ -163,6 +163,105 @@ func Test_inMemoryService_SearchMemory(t *testing.T) {
...
+		req := &memory.AddEventsToMemoryRequest{
+			AppName:   "app1",
+			UserID:    "user1",
+			SessionID: "sess1",
+			Events:    []*session.Event{newEvent("event1", "The quick brown fox")},
+		}
+		if err := s.AddEventsToMemory(t.Context(), req); err != nil {
+			t.Fatalf("AddEventsToMemory() [1] error = %v", err)
+		}
+		req.Events = []*session.Event{newEvent("event1", "The quick brown fox"), newEvent("event2", "jumps over the lazy dog")}
```
karolpiotrowicz, line 209: This second call re-sends `event1` alongside `event2`, which is the one input where append-and-dedupe and a plain wholesale overwrite produce the same answer — so this subtest passes on an implementation that simply replaces the stored slice and keeps none of the incremental behaviour. Two calls carrying disjoint events is the assertion that pins it. The review body has one that passes here and fails on the overwrite version.

memory/inmemory_test.go
```diff
@@ -163,6 +163,105 @@ func Test_inMemoryService_SearchMemory(t *testing.T) {
...
+		}
+
+		got, err := s.SearchMemory(t.Context(), &memory.SearchRequest{AppName: "app1", UserID: "user2", Query: "unique-marker-word"})
+		if err != nil {
+			t.Fatalf("SearchMemory() error = %v", err)
+		}
+		if len(got.Memories) != 0 {
+			t.Errorf("SearchMemory() leaked across users, got %+v", got.Memories)
+		}
+	})
```
karolpiotrowicz, line 241: The name says "other sessions or users" but only `UserID` varies — there is no second `SessionID` and no second `AppName` anywhere in the new subtests. Replacing `sid := sessionID(req.SessionID)` with `sid := sessionID("")` leaves the whole suite green, because `SearchMemory` scans every session bucket for the `(app, user)` key and never observes which one an event landed in.

The case that catches it: the same event ID under `sess1` and `sess2` for one user, asserting both are searchable. Per-session dedup scope is the thing the name is promising.

memory/inmemory_test.go
```diff
@@ -163,6 +163,105 @@ func Test_inMemoryService_SearchMemory(t *testing.T) {
...
+		}
+
+		got, err := s.SearchMemory(t.Context(), &memory.SearchRequest{AppName: "app1", UserID: "user1", Query: "anything"})
+		if err != nil {
+			t.Fatalf("SearchMemory() error = %v", err)
+		}
+		if len(got.Memories) != 0 {
+			t.Errorf("SearchMemory() = %+v, want no memories for a contentless event", got.Memories)
+		}
+	})
```
karolpiotrowicz, line 262: Two more cases worth adding while you are here, both of which currently pass on a broken implementation.

A contentless event with `ID: "event1"` followed, in the same call, by a real event with `ID: "event1"` — dropping the `if !ok { continue }` guard makes the contentless one register its ID and the real one gets discarded, and nothing here notices.

The same event ID twice in one batch, asserting a single stored memory — removing `existingIDs[event.ID] = struct{}{}` leaves the suite green, since every duplicate the current tests send arrives on a separate call.

karolpiotrowicz (commented):
The refactor underneath this is clean — pulling the event-to-value conversion into `eventToValue` leaves `AddSessionToMemory` behaviourally identical, and the new method takes the same write lock as the existing writer, which I checked by driving all three methods concurrently on one `(app, user)` under the race detector. Two things need resolving before it lands, and one of them is a question for the maintainers rather than for you.

## Events without an ID are silently dropped

[`memory/inmemory.go#L137-L147`](https://github.com/google/adk-go/blob/2fec6af41ec4944e14e6ce9e71b8483539a0ab8e/memory/inmemory.go#L137-L147) dedupes on `event.ID` with no guard for the empty string. The first event with `ID == ""` claims that key, and every later ID-less event — in the same batch and in every subsequent call for that `(app, user, session)` scope — is skipped. The method returns `nil`, so the caller is never told.

`session.Event.ID` is a plain string commented "Set by storage" and nothing fills it unless the caller went through `session.NewEvent`. Events built by hand therefore arrive with an empty ID, and building events by hand is exactly what this API invites when its doc says to persist "only a subset of events". The memory package's own existing fixtures use ID-less events, and `AddSessionToMemory` stores them today.

This is the one place where the parity argument runs the other way. adk-python can dedupe on `event.id` safely because [`Event.model_post_init`](https://github.com/google/adk-python/blob/c506ddf3bc34a6312ffc81899221bfa3f2da3b1d/src/google/adk/events/event.py#L282-L286) fills the id on every construction, so an empty id is unreachable there. Go has no equivalent, and the repo says so in its own words at [`internal/llminternal/functions_test.go#L213-L215`](https://github.com/google/adk-go/blob/2fec6af41ec4944e14e6ce9e71b8483539a0ab8e/internal/llminternal/functions_test.go#L213-L215): "A raw `&session.Event{}` literal leaves ID as `""` which breaks features that rely on event IDs."

The reproduction below drops the model's entire turn. The same two events through `AddSessionToMemory` keep both, which is what isolates it to the new path.

<details><summary>Reproduction — <code>go test -run 'TestEmptyID' ./memory/</code></summary>

```go
package memory_test

import (
	"testing"

	"google.golang.org/genai"

	"google.golang.org/adk/v2/memory"
)

func evNoID(text string, role genai.Role) *session.Event {
	return &session.Event{
		Author:      string(role),
		LLMResponse: model.LLMResponse{Content: genai.NewContentFromText(text, role)},
	}
}

// FAILS: SearchMemory("help") returns 0 memories. The model's turn is gone.
func TestEmptyID_CollapsesOneBatch(t *testing.T) {
	s := memory.InMemoryService()
	if err := s.AddEventsToMemory(t.Context(), &memory.AddEventsToMemoryRequest{
		AppName: "app1", UserID: "user1", SessionID: "sess1",
		Events: []*session.Event{
			evNoID("hello world", genai.RoleUser),
			evNoID("how can I help you", genai.RoleModel),
		},
	}); err != nil {
		t.Fatalf("AddEventsToMemory() error = %v", err)
	}
	got, err := s.SearchMemory(t.Context(), &memory.SearchRequest{AppName: "app1", UserID: "user1", Query: "help"})
	if err != nil {
		t.Fatalf("SearchMemory() error = %v", err)
	}
	if len(got.Memories) != 1 {
		t.Errorf("SearchMemory(%q) = %d memories, want 1", "help", len(got.Memories))
	}
}

// PASSES on the same two events: the control that isolates this to the new path.
func TestEmptyID_ControlAddSessionKeepsBoth(t *testing.T) {
	s := memory.InMemoryService()
	if err := s.AddSessionToMemory(t.Context(), makeSession(t, "app1", "user1", "sess1", []*session.Event{
		evNoID("hello world", genai.RoleUser),
		evNoID("how can I help you", genai.RoleModel),
	})); err != nil {
		t.Fatalf("AddSessionToMemory() error = %v", err)
	}
	got, err := s.SearchMemory(t.Context(), &memory.SearchRequest{AppName: "app1", UserID: "user1", Query: "help"})
	if err != nil {
		t.Fatalf("SearchMemory() error = %v", err)
	}
	if len(got.Memories) != 1 {
		t.Errorf("CONTROL: SearchMemory(%q) = %d memories, want 1", "help", len(got.Memories))
	}
}
```

A second shape is worse: ingest one ID-less event through `AddSessionToMemory`, and every ID-less event sent to `AddEventsToMemory` for that scope afterwards is dropped for the lifetime of the process.

</details>

## Whether this belongs on the `Service` interface — a question for the maintainers

Adding the method to [`memory.Service`](https://github.com/google/adk-go/blob/2fec6af41ec4944e14e6ce9e71b8483539a0ab8e/memory/service.go#L43) is source-breaking for anyone with a custom memory backend. `apidiff` reports it as the only incompatible change in the module, and a type implementing just `AddSessionToMemory` and `SearchMemory` compiles against the base and fails against this branch with "missing method AddEventsToMemory". Nothing in-tree breaks, since both in-repo implementations are updated here — the whole break lands out of tree, and `runner.Config.MemoryService`, `adkrest.Handler.MemoryService` and `cmd/launcher` all invite outside implementations.

Worth surfacing because the parity argument cuts against this shape too: adk-python deliberately made [`add_events_to_memory`](https://github.com/google/adk-python/blob/c506ddf3bc34a6312ffc81899221bfa3f2da3b1d/src/google/adk/memory/base_memory_service.py#L64-L95) a concrete method raising `NotImplementedError`, not an `@abstractmethod` like its two neighbours, precisely so existing subclasses keep working. Go has no default method on an interface, so the equivalents are an optional extension interface with a type assertion at the call site, or an embeddable default of the kind `agent.StrictContextMock` already provides. Which of those the project wants is not yours to decide alone, and I would rather that call were made explicitly than arrived at by merge. If the answer is to take the break, `README-v2.md` is where the project records these and this change has no entry there yet.

## The new tests pass on an implementation that does the opposite

Replacing the whole append-and-dedupe body with a wholesale overwrite — that is, deleting the incremental behaviour this PR exists to add — leaves the suite green. So does ignoring `req.SessionID` entirely, dropping the contentless-event guard, and dropping the within-batch dedup line.

The gap is that only one of the four subtests makes two calls at all, and [it re-sends the full prior set on the second call](https://github.com/google/adk-go/blob/2fec6af41ec4944e14e6ce9e71b8483539a0ab8e/memory/inmemory_test.go#L198-L221) — the one input on which append-and-dedupe and wholesale-replace agree. Two calls carrying *disjoint* events is the assertion that pins the feature, and I confirmed it passes here and fails on the overwrite version:

```go
func TestDisjointCallsAccumulate(t *testing.T) {
	s := memory.InMemoryService()
	add := func(id, text string) {
		t.Helper()
		if err := s.AddEventsToMemory(t.Context(), &memory.AddEventsToMemoryRequest{
			AppName: "app1", UserID: "user1", SessionID: "sess1",
			Events: []*session.Event{{ID: id, LLMResponse: model.LLMResponse{
				Content: genai.NewContentFromText(text, genai.RoleUser)}}},
		}); err != nil {
			t.Fatalf("AddEventsToMemory(%s) error = %v", id, err)
		}
	}
	add("e1", "the quick brown fox")
	add("e2", "jumps over the lazy dog")
	got, err := s.SearchMemory(t.Context(), &memory.SearchRequest{AppName: "app1", UserID: "user1", Query: "fox dog"})
	if err != nil {
		t.Fatalf("SearchMemory() error = %v", err)
	}
	if len(got.Memories) != 2 {
		t.Errorf("SearchMemory() = %d memories, want 2", len(got.Memories))
	}
}
```

Three smaller cases are worth a test each, and I have left the details on the relevant lines: the same event ID under two different sessions of one user, a contentless event followed by a real event with the same ID in one batch, and the same ID twice in one batch.

## Smaller things

These are all on the diff as line comments: the Vertex "not supported" error being unmatchable by `errors.Is`, the absent concurrency coverage for the new method, `eventToValue` running inside the write lock where the sibling method runs it outside, and two documentation lines that promise slightly more than the implementation provides.

Two things I checked and am not raising, so you do not have to defend them: the request-level `CustomMetadata` being dropped by the in-memory service matches adk-python exactly, and the duplicate produced by omitting the optional `SessionID` happens in adk-python too — its `__unknown_session_id__` sentinel puts an absent id in a different bucket and the search reads every bucket, so it double-counts there as well. A nil element in the event slice panics, but `AddSessionToMemory` panics the same way on `main` today, so that one predates this change.

Stubbing the Vertex path rather than porting the real ingestion call was the right call for this PR's scope, and the issue offered it explicitly. Worth an issue for the follow-up, along with reaching the capability from the agent-facing surface — adk-python exposes it on the callback context, and right now nothing outside `memory/` can call this method at all.

memory/inmemory_test.go
```diff
@@ -163,6 +163,105 @@ func Test_inMemoryService_SearchMemory(t *testing.T) {
...
+}
```
karolpiotrowicz, line 263: Nothing drives `AddEventsToMemory` concurrently. `Test_inMemoryService_SearchMemory_Concurrent` further down exercises `AddSessionToMemory` and `SearchMemory` only, so `go test -race` says nothing about the new method — a later refactor that dropped the lock here would go unnoticed by CI.

The locking is correct as written: I ran eight writers on `AddEventsToMemory` against concurrent `AddSessionToMemory` and `SearchMemory`, all on one `(app, user)` and one `SessionID` so they genuinely contend, and it is clean under the detector. Worth adding a case so it stays that way. Note that the existing concurrent test seeds every session with `nil` events and gives each writer a distinct session id, so it never actually contends for a slot either.

memory/service.go
```diff
@@ -33,11 +33,35 @@ type Service interface {
 	//
 	// A session can be added multiple times during its lifetime.
 	AddSessionToMemory(ctx context.Context, s session.Session) error
+	// AddEventsToMemory adds an explicit list of events to the memory service.
+	//
+	// This is intended for callers who want to persist only a subset of
+	// events (e.g. the latest turn) rather than re-ingesting the full
+	// session via AddSessionToMemory. Implementations should treat Events
+	// as an incremental update and must not assume it represents the full
```
karolpiotrowicz, line 41: This tells implementers to treat `Events` as an incremental update, which is right, but the reverse direction is destructive and undocumented: `AddSessionToMemory` still assigns `v[sid] = values` wholesale, so a later full-session flush discards anything added incrementally that is not in that session's event list. I confirmed it — add an event here, then call `AddSessionToMemory` for the same session without it, and it is gone with no error.

adk-python behaves identically, so this is not drift. It is a sentence missing from a doc comment on the one API whose whole purpose is to be mixed with the other path.

memory/service.go
```diff
@@ -33,11 +33,35 @@ type Service interface {
...
+// AddEventsToMemoryRequest represents a request for [Service.AddEventsToMemory].
+type AddEventsToMemoryRequest struct {
+	AppName string
+	UserID  string
+	Events  []*session.Event
+
+	// Below are optional fields.
+
+	// SessionID scopes the events to a session. Implementations may ignore
+	// it if not applicable.
```
karolpiotrowicz, line 58: "Scopes the events to a session" reads as an isolation boundary, and it is not one. `SearchMemory` iterates every session bucket under the `(app, user)` key and `SearchRequest` has no session field at all, so events stored under one `SessionID` surface in searches for every other session of that user. `SessionID` scopes deduplication and nothing else.

Worth saying so explicitly here, since an integrator reading this line could reasonably put sensitive events under a separate session expecting them to stay there.

memory/vertexai/vertexai.go
```diff
@@ -89,6 +89,12 @@ func (v *vertexAIService) AddSessionToMemory(ctx context.Context, s session.Sess
 	return err
 }
 
+// AddEventsToMemory is not yet supported by the Vertex AI Memory Bank
+// service; use AddSessionToMemory to ingest a full session instead.
+func (v *vertexAIService) AddEventsToMemory(ctx context.Context, req *memory.AddEventsToMemoryRequest) error {
+	return fmt.Errorf("vertexAIService does not support AddEventsToMemory")
```
karolpiotrowicz, line 95: A caller cannot tell this apart from a transient failure without matching the string. Backend-agnostic code that wants "try the incremental path, fall back to `AddSessionToMemory` when the backend cannot do it" has to `strings.Contains` the message, and will then also fall back on genuine errors — or, worse, read a real failure as "unsupported" and skip ingestion silently.

This matters more once the method is required on the interface, because every generic caller has to handle the unsupported case. An exported sentinel checked with `errors.Is` is the pattern the repo already uses elsewhere (`workflow.ErrParallelHITLUnsupported`, `workflow.ErrUnsupportedFanIn`), and `errors.ErrUnsupported` exists in the standard library for exactly this.

Separately: this is the only `fmt.Errorf` in the file with no format verb and no wrapping — the other five all carry verbs and four wrap with `%w`. `errors.New` would do if the sentinel is out of scope.

PR #1163, opened 2026-07-16
fix(llmagent): avoid shared-state races on parallel single_turn

### LINKED ISSUE #1137: Data race: parallel single_turn dispatches of the same sub-agent write shared agent state
## 🔴 Required Information

**Describe the Bug:**

Dispatching the same `single_turn` sub-agent more than once via parallel function calls in a single model turn is a data race. `handleFunctionCalls` runs each function call on its own goroutine (`internal/llminternal/base_flow.go:1064`), each call's `SingleTurnTool.Run` constructs and runs a workflow node for the sub-agent, and `RunLLMAgentAsNode` then writes to the sub-agent's shared internal state on every dispatch:

- `agent/llmagent/llm_agent_wrapper.go:86` — `state.IncludeContents = "none"` (unconditional per single_turn dispatch; write-write between concurrent dispatches, and write-read against the contents processor reading `IncludeContents` for the sibling dispatch's LLM request)
- `agent/llmagent/llm_agent_wrapper.go:74-76` and `workflow/agent_node.go:53-58` — the lazy `ModeUnset → ModeSingleTurn` defaulting, the same unsynchronized read-then-write pattern for agents whose mode is resolved at node construction/dispatch time (reachable e.g. through direct `workflow.NewAgentNode` use, since `NewSingleTurnTool` constructs a fresh node per call)

The same lazy-defaulting pattern also exists at `runner/runner.go:194-197` (root agents default `ModeUnset → ModeChat` on first `Runner.Run`), which races for servers that share one root agent instance across concurrent `Run` calls.

Since parallel fan-out to sub-agents is a core multi-agent pattern, any program doing it trips the race detector, and per the Go memory model the writes are undefined behavior even though the racing writes happen to store equal values.

**Steps to Reproduce:**

1. `go mod init adkrace && go get google.golang.org/adk/v2@v2.0.0`
2. Save the test below as `race_test.go`
3. `go test -race .`
4. The run fails with `WARNING: DATA RACE` at `agent/llmagent/llm_agent_wrapper.go:86` (report attached under Logs)

**Expected Behavior:**

Dispatching a sub-agent on several branches concurrently — the documented parallel fan-out pattern — is race-free, and `go test -race` passes.

**Observed Behavior:**

`go test -race` reports a write-write data race on the sub-agent's internal state between the two dispatch goroutines and fails the test. Race report attached under Logs.

**Environment Details:**

- ADK Library Version: v2.0.0 (also reproduces at current `main`, v2.0.1-0.20260709115221-d0111d0dc15c)
- OS: macOS (darwin/arm64)
- Go Version: go1.26.1

**Model Information:**

- Which model is being used: N/A — reproduces with a scripted `model.LLM` test double; the race is model-independent.

---

## 🟡 Optional Information

**Regression:**

No — v2.0.0 is the first release with the v2 node runtime these code paths belong to.

**Logs:**

```text
WARNING: DATA RACE
Write at 0x00c00018e5a0 by goroutine 12:
  google.golang.org/adk/v2/agent/llmagent.(*llmAgent).RunNode.RunLLMAgentAsNode.func1()
      google.golang.org/adk/v2@v2.0.0/agent/llmagent/llm_agent_wrapper.go:86 +0x360
  google.golang.org/adk/v2/workflow.(*AgentNode).Run.func1()
      google.golang.org/adk/v2@v2.0.0/workflow/agent_node.go:136 +0x750
  google.golang.org/adk/v2/workflow.(*dynamicSubScheduler).runNode()
      google.golang.org/adk/v2@v2.0.0/workflow/dynamic_scheduler.go:295 +0x1134
  google.golang.org/adk/v2/workflow.(*dynamicSubScheduler).RunNode()
      google.golang.org/adk/v2@v2.0.0/workflow/dynamic_scheduler.go:86 +0x1ec
  google.golang.org/adk/v2/workflow.RunNode[go.shape.interface {}]()
      google.golang.org/adk/v2@v2.0.0/workflow/run_node.go:139 +0x194
  google.golang.org/adk/v2/internal/workflowinternal.(*SingleTurnTool).Run()
      google.golang.org/adk/v2@v2.0.0/internal/workflowinternal/single_turn_tool.go:79 +0x384
  google.golang.org/adk/v2/internal/llminternal.(*Flow).callTool()
      google.golang.org/adk/v2@v2.0.0/internal/llminternal/base_flow.go:1246 +0x164
  google.golang.org/adk/v2/internal/llminternal.(*Flow).handleFunctionCalls.func2()
      google.golang.org/adk/v2@v2.0.0/internal/llminternal/base_flow.go:1161 +0x1940
  google.golang.org/adk/v2/internal/llminternal.(*Flow).handleFunctionCalls.gowrap1()
      google.golang.org/adk/v2@v2.0.0/internal/llminternal/base_flow.go:1213 +0x48

Previous write at 0x00c00018e5a0 by goroutine 11:
  google.golang.org/adk/v2/agent/llmagent.(*llmAgent).RunNode.RunLLMAgentAsNode.func1()
      google.golang.org/adk/v2@v2.0.0/agent/llmagent/llm_agent_wrapper.go:86 +0x360
  [... identical stack for the second parallel dispatch ...]

Goroutine 12 (running) created at:
  google.golang.org/adk/v2/internal/llminternal.(*Flow).handleFunctionCalls()
      google.golang.org/adk/v2@v2.0.0/internal/llminternal/base_flow.go:1064 +0x584
  [...]
```

**Screenshots / Video:**

N/A

**Additional Context:**

The racing writes at `llm_agent_wrapper.go:86` store the same value (`"none"`), so the practical blast radius today is mostly that any application-level `go test -race` covering parallel fan-out fails, and that the Go memory model gives no guarantee the sibling dispatch's contents processor observes a consistent value. The `Mode` variants of the pattern (`workflow/agent_node.go:56`, `llm_agent_wrapper.go:75`, `runner/runner.go:196`) have the same shape: a lazy read-then-write default applied to shared agent state on a per-run code path.

```go
package adkrace

import (
	"context"
	"iter"
	"sync"
	"testing"

	"google.golang.org/genai"
)

// scriptedLLM implements model.LLM by yielding canned responses in call order
// (the last response repeats once the script is exhausted). Safe for
// concurrent calls.
type scriptedLLM struct {
	mu        sync.Mutex
	responses []*model.LLMResponse
	callIndex int
}

func (*scriptedLLM) Name() string { return "scripted-llm" }

func (l *scriptedLLM) GenerateContent(
	_ context.Context, _ *model.LLMRequest, _ bool,
) iter.Seq2[*model.LLMResponse, error] {
	l.mu.Lock()
	i := min(l.callIndex, len(l.responses)-1)
	l.callIndex++
	response := l.responses[i]
	l.mu.Unlock()
		yield(response, nil)
	}
}

// staticLLM implements model.LLM by always replying with the same text. Safe
// for concurrent calls.
type staticLLM struct{ text string }

func (*staticLLM) Name() string { return "static-llm" }

func (l *staticLLM) GenerateContent(
	_ context.Context, _ *model.LLMRequest, _ bool,
) iter.Seq2[*model.LLMResponse, error] {
		yield(&model.LLMResponse{
			Content: genai.NewContentFromText(l.text, genai.RoleModel),
		}, nil)
	}
}

// singleTurnCall builds a FunctionCall part invoking the worker sub-agent's
// auto-installed single_turn tool.
func singleTurnCall(id, request string) *genai.Part {
	return &genai.Part{FunctionCall: &genai.FunctionCall{
		ID:   id,
		Name: "worker",
		Args: map[string]any{"request": request},
	}}
}

// TestParallelSingleTurnDispatch dispatches the same single_turn sub-agent
// twice via parallel function calls in one model turn. Under -race this
// fails: both dispatches write the shared agent's internal state.
func TestParallelSingleTurnDispatch(t *testing.T) {
	ctx := context.Background()

	worker, err := llmagent.New(llmagent.Config{
		Name:        "worker",
		Description: "Handles one delegated task.",
		Model:       &staticLLM{text: "task done"},
		Mode:        llmagent.ModeSingleTurn,
	})
	if err != nil {
		t.Fatal(err)
	}

	coordinatorModel := &scriptedLLM{responses: []*model.LLMResponse{
		{Content: &genai.Content{Role: genai.RoleModel, Parts: []*genai.Part{
			singleTurnCall("call-a", "do task A"),
			singleTurnCall("call-b", "do task B"),
		}}},
		{Content: genai.NewContentFromText("all done", genai.RoleModel)},
	}}

	coordinator, err := llmagent.New(llmagent.Config{
		Name:      "coordinator",
		Model:     coordinatorModel,
		SubAgents: []agent.Agent{worker},
	})
	if err != nil {
		t.Fatal(err)
	}

		AppName:        "race-repro",
		Agent:          coordinator,
	})
	if err != nil {
		t.Fatal(err)
	}

	createResp, err := sessionService.Create(ctx, &session.CreateRequest{
		AppName: "race-repro",
		UserID:  "user",
	})
	if err != nil {
		t.Fatal(err)
	}

	for _, err := range agentRunner.Run(
		ctx, "user", createResp.Session.ID(),
		genai.NewContentFromText("fan out", genai.RoleUser),
		agent.RunConfig{},
	) {
		if err != nil {
			t.Fatal(err)
		}
	}
}
```

- Always (100%) — the reproduction above trips the race detector deterministically.

deepakraog:
## Summary
- Fixes [#1137](https://github.com/google/adk-go/issues/1137): parallel `single_turn` dispatches of the same sub-agent no longer race on shared LLM agent `State`.
- Stop writing `IncludeContents` / `Mode` on every node dispatch; derive single-turn current-turn contents from `Mode` at read time, and resolve `ModeUnset` once via `ResolveMode` (`sync.Once`).
- Adds a `-race` regression test that fans out two parallel `single_turn` tool calls through a chat coordinator.

## Testing plan
- [x] Reproduced the issue on latest `main` with `go test -race` (DATA RACE at `llm_agent_wrapper.go` IncludeContents write)
- [x] `go test -race ./agent/llmagent` (includes new `TestRunLLMAgentAsNode_ParallelSingleTurnDispatch_NoRace`)
- [x] `go test -race ./internal/llminternal`
- [x] `go test ./workflow ./runner ./internal/workflowinternal`
- [x] Delegation httprr tests still pass (`TestDelegation_03_ChatToSingleTurn`, `TestDelegation_04_ChatToTwoSingleTurnParallel`, `TestDelegation_11_ChatToSingleTurnChainedTools`)

## Notes
- Google CLA signed for `deepakraog`.
- Comment requesting assignment posted on #1137.


wolo-lab:
We're consolidating the #1137 work onto #1267, which the team is carrying, so this one won't be merged as-is. The reasoning is worth sharing, because part of your approach is what we ended up wanting.

Deriving the include-contents policy at read time rather than writing it onto the agent is the right call, and `EffectiveIncludeContents` gets a detail that two of the other candidate fixes miss: it preserves an explicit `IncludeContents` instead of forcing `none` for every single_turn agent. That matches adk-python, where `_llm_agent_wrapper.py` gates the override on `include_contents` being absent from `model_fields_set`. #1267 arrives at the same behavior.

The part that doesn't hold is moving the mode default to construction time. `SingleTurnTool.Run` builds a fresh node on every dispatch — [`single_turn_tool.go:74`](https://github.com/google/adk-go/blob/78439a6e4008251796410774c6b34aabbfe51eb0/internal/workflowinternal/single_turn_tool.go#L74) at commit [78439a6](https://github.com/google/adk-go/commit/78439a6e4008251796410774c6b34aabbfe51eb0) — so `NewAgentNode` is on the hot path, not in setup. Wrapping one undeclared agent concurrently still trips the race detector with this branch applied. The construction-time write in `runner.New` and `installTaskTools` also keeps the first-placement-wins behavior, so an agent used both as a graph node and as a chat sub-agent still resolves differently depending on which object was built first.

#1267 removes the write entirely and resolves the mode per invocation instead, which covers both the dispatch path and the ordering problem. Happy to have your eyes on it — the include-contents handling there is the same idea as yours.

PR #1164, opened 2026-07-17
feat(artifact): support artifact:// reference parts in the in-memory service

### LINKED ISSUE #1157: Artifact save/load has no scope validation for artifact:// reference URIs (cross-tenant escape)
### Description

adk-python's artifact service supports `Part`s that are references to other artifacts, using URIs of the form:

```
artifact://apps/{app_name}/users/{user_id}/sessions/{session_id}/artifacts/{filename}/versions/{version}
artifact://apps/{app_name}/users/{user_id}/artifacts/{filename}/versions/{version}
```

See `adk-python/src/google/adk/artifacts/artifact_util.py`:
- `parse_artifact_uri()` parses these URIs.
- `is_artifact_ref(part)` checks `part.file_data.file_uri.startswith("artifact://")`.
- `validate_artifact_reference_scope(...)` raises `InputValidationError` if the URI's embedded `app_name`/`user_id` doesn't match the caller's, or if a session-scoped ref's `session_id` doesn't match the caller's session.

This is wired into both save and load in `in_memory_artifact_service.py` (e.g. lines ~133-141 for save, ~186-195 for load), and covered by tests in `tests/unittests/artifacts/test_artifact_service.py` (`test_artifact_reference_rejects_cross_user_on_save`, cross-app and cross-session variants) and `tests/unittests/artifacts/test_artifact_util.py`.

### Problem in adk-go

`artifact/` has no equivalent concept at all — there is no `artifact://` URI parsing, no `is_artifact_ref` check, and no scope validation anywhere in `Save`/`Load` (confirmed: zero occurrences of `artifact://` anywhere under `artifact/`, in either `inmemory.go` or `gcsartifact/service.go`).

Concretely, this means a `genai.Part` whose `FileData.FileURI` is an `artifact://...` reference pointing at another app/user/session's artifact is not rejected — there's simply no code path that inspects it. In a multi-tenant deployment, if a model or tool ever emits (or a client submits) a `FileData` part with such a URI, nothing in the Go artifact service stack would reject a cross-tenant reference the way Python does.

### Suggested fix

Port the equivalent of `artifact_util.go`:
- A URI parser matching the two Python regexes.
- An `IsArtifactRef(part *genai.Part) bool` helper.
- A `ValidateArtifactReferenceScope(...)` check, wired into the save/load paths of `artifact/inmemory.go` and `artifact/gcsartifact/service.go`, returning an error (consistent with existing `Validate()` error patterns in `artifact/service.go`) on cross-app/user/session escape.

### Environment

- adk-go: main @ cbe2604
- Found by comparing against adk-python (`artifact_util.py`, `in_memory_artifact_service.py`)

mdabydeen:
## Summary

adk-python's artifact service supports `Part`s that are references to other
artifacts, using `artifact://apps/{app}/users/{user}/[sessions/{session}/]artifacts/{file}/versions/{version}`
URIs (see `artifact_util.py` / `in_memory_artifact_service.py`). adk-go had
no equivalent — `SaveRequest.Validate()` unconditionally rejects any `Part`
that isn't `Text` or `InlineData`, so a `FileData`-only reference part could
never reach storage.

This PR ports the reference feature to the **in-memory** artifact service:

- New `artifact/artifact_util.go`: `ParseArtifactURI`, `BuildArtifactURI`,
  `IsArtifactRef`, `ValidateArtifactReferenceScope`.
- `SaveRequest.Validate()` now accepts `FileData`-only parts when the URI is
  a well-formed `artifact://` reference.
- `inMemoryService.Save` validates the reference's scope (app/user/session)
  before storing it.
- `inMemoryService.Load` resolves references recursively (bounded to depth
  10 to guard against cycles), re-validating scope at each hop so a
  reference can never escape the app/user/session it was saved under.

## Scope

Kept intentionally small, per `AGENTS.md`'s "keep PRs small and focused —
one concern per PR":

- **Not included:** `artifact/gcsartifact/service.go` wiring. Python's GCS
  service supports refs too, but via blob metadata — a separate, mechanically
  different port. Happy to follow up with that PR next.
- **Not included:** a Go equivalent of `validate_path_segment`. adk-go's
  in-memory service uses an ordered key encoding rather than raw filesystem
  paths, so this doesn't apply the same way; flagging it here rather than
  silently dropping it.

Fixes #1157

## Testing Plan

- `golangci-lint run` (v2.3.1) — 0 issues
- `go mod tidy -diff` — clean, no new deps
- New tests: `artifact/artifact_util_test.go` (URI parsing/building, scope
  validation), plus save/load round-trip and cross-app/user/session
  rejection cases added to `artifact/inmemory_test.go`, plus new
  `SaveRequest.Validate()` cases in `artifact/request_validation_test.go`.


karolpiotrowicz:
@wolo-lab a direction call before review.

The change exports five new identifiers from the public `artifact` package and relaxes an existing public contract. adk-python exports no equivalents, so this would be new surface unique to Go rather than parity. Worth deciding whether we want that surface before reviewing how well it is built.

PR #1166, opened 2026-07-17
fix(llmagent): resolve agent mode per call instead of writing shared state

### LINKED ISSUE #1137: Data race: parallel single_turn dispatches of the same sub-agent write shared agent state

chethanuk:

- Closes: #1137

**Problem:**

`llminternal.State` is shared: `Reveal` hands every caller a pointer to the one struct an agent carries for its whole lifetime. Three run-time paths wrote it. On current main they are `IncludeContents="none"` on every `single_turn` dispatch (`agent/llmagent/llm_agent_wrapper.go:85`) and the lazy `ModeUnset` defaults in the node wrapper (`agent/llmagent/llm_agent_wrapper.go:73-74`) and in `Runner.Run` (`runner/runner.go:213-215`). `handleFunctionCalls` runs one goroutine per function call, so parallel calls to the same agent race on all three. The `runner/runner.go:194-199` @mathewpang quoted is that same write in v2.0.0; main has since moved it down, so the two line numbers are one bug, not two.

The `Mode` write is `"" -> "chat"`, which changes the string header's data pointer and its length non-atomically, so a racing reader can pair the new length with the old nil pointer and segfault in the comparison that follows. @mathewpang reported that crash from production and showed it needs no shared-root-agent server: `agentTool.Run` builds a fresh `Runner` over the one shared sub-agent on every call, inside the per-call goroutine (`tool/agenttool/agent_tool.go:143`). The REST server has the same shape, since `getRunner` builds a `Runner` per request over the loader's single shared agent (`server/adkrest/controllers/runtime.go:213`, `agent/loader.go:53`).

**Solution:**

Stop storing both values.

A `single_turn` agent gets the current turn only, which the wrapper used to arrange by overwriting `IncludeContents`. The contents processor now derives that from the mode instead, ten lines from where it already reads it.

A mode default depends on the agent's role in one particular call, not on the agent, so the caller that knows the role declares it on the context for that one agent (`llminternal.WithDefaultMode`) and readers resolve `State.Mode` against it (`State.ResolveMode`) without writing back. The runner declares chat for its root. The node wrapper defaults to single_turn, as `Config.Mode` documents, and declares what it resolved so the request processors below it read the same value. That holds whether the wrapper is reached through `workflow.AgentNode` or called directly. The declaration names its agent, so agents dispatched below it keep their own behavior. `State` gains no lock and stays copyable. Resolving per call is also what closes the `agenttool` path: an earlier revision of this PR moved the write from `Runner.Run` into `runner.New`, which only relocated it into the per-call goroutine, where it raced identically.

**Why not `ResolveMode` behind a `sync.Once`?** The `Once` is first-caller-wins, and the call sites pass different defaults: `ModeChat` from the runner, `ModeSingleTurn` from the node wrapper. Whichever role wins the race decides the other role's behavior for the rest of the process, so an agent can no longer be a chat root in one goroutine and a single_turn node in another. A `Once` also establishes happens-before only against other `Do` callers, and four readers take `state.Mode` directly, synchronizing with nothing: `internal/llminternal/basic_processor.go:53`, `internal/llminternal/outputschema_processor.go:105`, `workflow/validation.go:248`, `workflow/agent_node.go:132`.

**Behavior change, with a one-line remedy.** `workflow.NewAgentNode` no longer mutates the agent handed to it, so a wrapped agent's `State.Mode` stays `ModeUnset`. Before, wrapping a sub-agent as a node *before* building its parent left that sub-agent at `single_turn` and the parent then installed a `SingleTurnTool` for it; wrapping it afterwards did not, so the two construction orders disagreed. Now both orders leave the sub-agent at `llmagent.New`'s own default, chat, and a chat sub-agent gets no delegation tool, so a parent that used to receive one stops receiving it. Setting `Mode: llmagent.ModeSingleTurn` in the sub-agent's config restores the tool and pins it under either order. `TestSubAgentMode_ConstructionOrder` covers all four combinations.

The four run-time readers that distinguish single_turn resolve the mode from the context instead (the contents, identity and agent-transfer request processors, plus the wrapper). Readers that only test for `ModeTask` are unaffected, because a declared default is never task. Exported `llmagent.PrepareLLMAgentInput` follows the same rule and now returns a seed event for an unset-mode agent where it used to return nil.

The change buys something back, too: an agent whose mode is left unset can be a runner root and a workflow node, in either order and at the same time. Before, whichever role wrote first won, and a node-wrapped agent handed to `runner.New` failed with "must be a chat LlmAgent, but has mode single_turn".

Two writes to shared agent state stay: `llmagent.New` defaults a sub-agent's mode in place (`agent/llmagent/llmagent.go:157`), which happens at construction rather than per call, and `sequentialAgent.RunLive` appends a tool and an instruction suffix to each sub-agent on every call (`agent/workflowagents/sequentialagent/agent.go:158`). The `RunLive` one touches different fields (`state.Tools` and `state.Instruction`) on the Live path only, and undoing an append is not the same fix as dropping a default, so it needs its own change rather than a wider version of this one.

If you would rather resolve the default some other way, threading it through `agent.InvocationContext` instead of a context value for instance, I am glad to redo the mechanism; the diagnosis and the tests hold either way.

### Testing Plan

**Unit Tests:**

Regression guards, each checked against the unfixed code:

- `TestAgentTool_ParallelDispatch_SharedStateNotWritten` (`tool/agenttool`) is the reported path: two and eight parallel calls to one agent tool in a single model turn. On the pre-fix code it fails even without `-race` (`worker Mode after run = "chat", want unchanged ""`), and under `-race` it reports the write reached from `agent_tool.go:143`.
- `TestRunner_ConcurrentRootRuns` (`runner`) is the server shape: eight concurrent `Run` calls on distinct sessions over one unset-mode root. With the old `Runner.Run` write restored it reports `WARNING: DATA RACE` between the read at `runner.go:223` and the write at `runner.go:225`, and the Mode assertion fails.
- `TestParallelSingleTurnDispatch_SharedStateNotWritten` (`agent/llmagent`) does the same for the `IncludeContents` write, on a parallel `single_turn` fan-out.
- `TestRunner_SameAgentAsRootAndNodeConcurrently` (`runner`) runs one unset-mode agent in both roles at once and asserts each call sees what its role should: the chat root gets the session history, the node gets the current turn only. Make the node role resolve to chat and it fails with `single_turn node request carried 3 contents, want the current turn only (1)`.
- `TestSubAgentMode_ConstructionOrder` (`agent/llmagent`) pins the sub-agent's mode and the parent's installed tools for both construction orders. Restore the `NewAgentNode` stamp and its wrap-first row fails on both assertions.

The rest pin the resolution rule, and each fails on main. `TestRunner_RootMode` (`runner`) covers the five root modes including the node-wrapped root main rejects. `TestModeDefaultIsResolvedPerCall` and `TestUnsetModeRoot_RunsAsChatWithoutWriting` (`agent/llmagent`) cover precedence and a chat root's history. The `contents`, `identity` and `agent_transfer` tables gain rows for a caller-declared single_turn, including one where the declaration belongs to a different agent and must not apply. The agent-transfer row compares the whole request against an explicitly single_turn agent's, because the processor appends the transfer tool's declaration whatever the mode and only the instruction text differs, so walking `SystemInstruction` alone would not prove much.

Delegation is the behavior most exposed to this change, so these three recorded-model tests were re-run and pass unchanged: `TestDelegation_03_ChatToSingleTurn`, `TestDelegation_04_ChatToTwoSingleTurnParallel`, `TestDelegation_11_ChatToSingleTurnChainedTools`.

```
go build -mod=readonly ./...                             ok, both modules
go test -race -mod=readonly -count=1 -shuffle=on ./...   ok, 71 packages; plugin/agentanalytics ok separately
go test -race -count=5 <the new tests>                   ok
go vet ./...                                             ok
golangci-lint run                                        0 issues, both modules
go mod tidy -diff                                        clean, both modules
```

The reporter's repro is `TestAgentTool_ParallelDispatch_SharedStateNotWritten` rather than a separate manual run: it drives a real coordinator, agent tool, sub-agent runner and session through `runner.Run`, which is the crash path from the issue.

### Checklist


chethanuk:
@hanorik @baptmont Please review

chethanuk:
@hanorik @baptmont Please review this as well

wolo-lab:
We're landing the #1137 fix through #1267, which the team is carrying. That's a close call rather than an obvious one, and you should know why, because your PR and #1267 reached the same design independently.

`WithDefaultMode` plus a name-scoped `ResolveMode(ctx, agentName, fallback)` is the right shape: resolve per invocation, carry it on the context, never write to state shared across concurrent runs. #1267 does the same thing under different names. Your version also merges cleanly with main and its own suite is green, which is more than the other candidates manage.

The one behavioral difference is `include_contents`. This PR makes a single_turn agent hide history unconditionally, and documents that as intended. adk-python treats an explicitly set `include_contents` as beating the placement — `_llm_agent_wrapper.py` only forces `'none'` when the field is absent from `model_fields_set` — so a caller who explicitly asks for history keeps it there. #1267 follows Python on that point.

There's a smaller second difference. #1267 separates the two questions: whether to hide history follows the placement alone, while whether to shape the turn as single_turn also honors the agent's own declaration. The practical effect is that a declared single_turn agent reached outside a placement keeps its conversation.

Neither of those makes this PR wrong, and if you'd like to review #1267 your read on the resolution semantics would be a genuinely useful one to have on it.

PR #1167, opened 2026-07-17
fix(llminternal): guard RunLive errChan sends so abandoned producers exit on reconnect

### LINKED ISSUE #1152: Flow.RunLive leaks one goroutine per reconnect: unbuffered errChan abandoned on resumable errors
### Description

In `internal/llminternal/base_flow.go`, `Flow.RunLive` creates a fresh unbuffered `errChan` per connection attempt (main @ 0c88126, line 373; present since v1.3.0 and unchanged at v1.5.0). The reader and sender goroutines send to it without a cancellation guard:

- reader: `errChan <- err` (line 389)
- sender: `errChan <- err` (lines 430, 439)

The consumer reads a single error in the main `select` (line 521). When that error `isResumable`, the loop sets `reconnect = true`, breaks out, and starts a new attempt with a **new** `errChan` — abandoning the previous one. If the *other* goroutine of the old attempt subsequently fails (e.g. the sender hits the closed connection after the reader already reported), it blocks forever on a send to the abandoned unbuffered channel: one leaked goroutine (plus its stack and captured connection state) per reconnect cycle. Long-lived live sessions on flaky networks accumulate these.

### Suggested fix

Wrap each `errChan` send in a `select` with `case <-connCtx.Done():` (mirroring the existing `eventsChan` pattern in the same function), and ensure the per-attempt `cancelConn` is invoked on the reconnect path before starting the next attempt — today `cleanup()` runs on terminal paths but not when `reconnect = true` breaks the select.

Alternatively, `errChan := make(chan error, 2)` (one slot per producer goroutine) also unblocks the producers, at the cost of masking rather than modeling the lifecycle.

### Environment

Observed by code inspection at v1.5.0 and confirmed unchanged on main (0c88126). Found during an integration review of v1.5.0 for a fork we run in a pre-production live voice deployment.

chethanuk:
Fixes google/adk-go#1152

Long-lived `/run_live` servers leak one goroutine (plus its captured connection state) every time a live session reconnects — and, it turns out, on ordinary teardown too. The producers inside `Flow.RunLive` do bare sends on the unbuffered, per-attempt `errChan`; once the consumer loop has stopped reading that channel, the producer blocks on the send forever.

## Root cause

`internal/llminternal/base_flow.go` spawns two producer goroutines per connection attempt (a reader and a sender) that report failures via `errChan <- err` at three sites (`:389`, `:430`, `:439`). The consumer reads `errChan` at most once per attempt: on a resumable error it sets `reconnect = true`, breaks out, and `cleanup()` cancels `connCtx` and closes the connection. The other producer's in-flight `Recv`/`Send*` call now fails against the closed connection, and its bare send on the abandoned channel blocks permanently:

```mermaid
sequenceDiagram
    participant R as reader
    participant C as consumer
    participant S as sender
    R->>C: errChan <- resumable err
    C->>C: reconnect = true, break
    C->>C: cleanup: cancelConn + Close conn
    Note over C: next attempt gets fresh errChan
    S--xS: SendContent fails (conn closed)
    S--xS: errChan <- err on abandoned channel
    Note over S: blocks forever - leaked
```

The same mechanics fire without any reconnect: cancel the invocation context while the reader is parked in `Recv` (the everyday session-teardown path), and the consumer exits via `ctx.Done()` → `cleanup()` closes the connection → the reader's `Recv` fails → its bare send strands it. So every normal live-session teardown leaks the reader goroutine, which makes this slightly broader than the reconnect-only symptom in the issue.

One small correction to the issue text, respectfully: `cleanup()` *does* run on the reconnect path — it is called unconditionally after the consumer loop (`:541`), before the `reconnect` check. Only the send-site guards are missing, which makes the fix smaller than proposed.

## The fix

Guard all three `errChan` sends with the exact idiom the same function already uses for `eventsChan` (`:409-413`):

```go
select {
case errChan <- err:
case <-connCtx.Done():
}
return
```

Guarding on **`connCtx`**, not `ctx`, is load-bearing: `cleanup()` always cancels `connCtx` — before the next attempt and before every return — so a guarded send is *guaranteed* to unblock an abandoned producer. `ctx` outlives reconnects and would leave the producer blocked across attempts.

Deliberately untouched: the `eventsChan` send (already guarded), the consumer's double-break flow at `:521-537`, and `cleanup()` placement.

**Why not a buffered `make(chan error, 2)`?** It also fixes the current leak, but as discussed on the issue it silently caps the number of background workers and masks the lifecycle instead of modeling it; the `select` guard extends the function's existing convention. An `errgroup` restructure would be a rewrite, out of proportion for this bug.

## Tests

`TestRunLiveNoGoroutineLeak` (new `base_flow_live_test.go`) fakes the Gemini Live wire protocol with an `httptest` + `gorilla/websocket` server (the same approach as go-genai's own `live_test.go`), pointed at via `HTTPOptions.BaseURL` with a `ws://` scheme — no new dependencies, no production test hooks.

| case | server script | asserts |
|---|---|---|
| reader resumable error reconnects without leak | conn 1: hard close after setup (client sees resumable 1006); conn 2: serves a turn | turn received after reconnect, 2 connections, no leak |
| sender error after connection loss does not leak | conn 1: hard close; conn 2: echoes a turn per client message | retried `Send` succeeds on conn 2, no leak |
| non-resumable error terminates | close frame code 1002 | error surfaces through the iterator, exactly 1 connection, no leak |
| parent ctx cancel exits cleanly | server stays silent; test cancels ctx | RunLive unwinds fully, no leak |

The leak assertion is dependency-free (goleak is not in go.mod): it snapshots `runtime.Stack(all=true)`, counts goroutines still inside `llminternal.(*Flow).RunLive`, and polls until the count returns to the subtest's baseline or a 5s deadline expires, failing with the captured stacks.

## Testing Plan

```
go build -mod=readonly ./...
go test -race -run TestRunLiveNoGoroutineLeak ./internal/llminternal/ -count=20   # 100/100 pass
go test -race -mod=readonly -count=1 -shuffle=on ./...                            # 2408 passed, 151 packages
golangci-lint run internal/llminternal/...                                        # no issues
go mod tidy -diff                                                                 # clean (no new deps)
```

Behavior-change evidence — the new test on the **unfixed** code fails deterministically, with the leaked producers parked in `chan send` at the exact bare-send site:

```
--- FAIL: TestRunLiveNoGoroutineLeak/reader_resumable_error_reconnects_without_leak
    base_flow_live_test.go:316: leaked 1 RunLive goroutines (baseline 0):
    goroutine 38 [chan send]:
    google.golang.org/adk/v2/internal/llminternal.(*Flow).RunLive.func1.2(...)
        internal/llminternal/base_flow.go:389 +0x6b5
    created by google.golang.org/adk/v2/internal/llminternal.(*Flow).RunLive.func1 in goroutine 22
--- FAIL: TestRunLiveNoGoroutineLeak/sender_error_after_connection_loss_does_not_leak
    base_flow_live_test.go:316: leaked 2 RunLive goroutines (baseline 1): ...
--- FAIL: TestRunLiveNoGoroutineLeak/parent_ctx_cancel_exits_cleanly
    base_flow_live_test.go:316: leaked 1 RunLive goroutines (baseline 3): ...
```

With the fix applied, all four cases pass 100/100 under `-race -count=20`.


internal/llminternal/base_flow.go
```diff
@@ -436,7 +445,10 @@ func (f *Flow) RunLive(ctx agent.InvocationContext) (agent.LiveSession, iter.Seq
 								sess.audioMgr.CacheInput(ctx, blob.Data, blob.MIMEType)
 							}
 							if err := liveConn.SendRealtime(connCtx, req.RealtimeInput); err != nil {
-								errChan <- err
+								select {
```
hanorik, line 448: Reverting just this guard still passes `-count=3`; the other two get caught. Nothing sets `RealtimeInput`, so this site is uncovered.

internal/llminternal/base_flow_live_test.go
```diff
@@ -0,0 +1,322 @@
...
+		if _, _, err := conn.ReadMessage(); err != nil {
+			return
+		}
+	}
+}
+
+// runLiveStacks returns the goroutine stacks still parked inside RunLive's
+// producer/consumer closures, and their count.
+func runLiveStacks() (int, string) {
+	buf := make([]byte, 1<<20)
```
hanorik, line 111: `runtime.Stack` truncates silently and `n == len(buf)` isn't checked, so a large dump under-counts and a real leak passes. Suggest growing until `n < len(buf)`.

internal/llminternal/base_flow_live_test.go
```diff
@@ -0,0 +1,322 @@
...
+				if !sawTurn {
+					t.Error("never received a model turn for the retried send")
+				}
+			},
+			wantConns: 2,
+		},
+		{
+			// A close frame with a non-resumable code must surface as an error
+			// to the caller and terminate the flow after a single connection.
+			name: "non-resumable error terminates",
```
hanorik: Passes 5/5 against unfixed code — the sender exits via `connCtx.Done()` at `base_flow.go:428`. Good behaviour test, but doesn't guard the leak regression; maybe rename.

hanorik (commented):
Hi, thanks for taking this! Fix looks right — all my comments are about the test, not `base_flow.go`.

### Nits (none blocking)

1. **One guard is uncovered.** Revert *only* the `SendRealtime` guard (`base_flow.go:448-451`) and the test still passes — reverting either of the other two gets caught. Nothing populates `LiveRequest.RealtimeInput`. A case sending a `*genai.Blob` across a connection loss would cover it.

2. **The detector can fail open.** `base_flow_live_test.go:108` uses a fixed 1 MiB buffer and doesn't check `n == len(buf)`. `runtime.Stack` truncates silently, so a big dump makes it *under*-count and a real leak passes. Latent today (~1.2 KB), but wrong failure direction. Two-line fix: grow until `n < len(buf)`.

3. **`non-resumable error terminates` passes without the fix** (5/5 on base) — the sender exits via the existing `connCtx.Done()` case at `:428`, so nothing is stranded. Fine as a behaviour test, just not a leak test; worth renaming.

### Outside the diff

`base_flow.go:383` — it seems that the `SendHistory` failure path is the only `return` between `context.WithCancel` (`:352`) and the end of the attempt loop that skips `cleanup()`, so it might leaks the context *and* the socket (`vet`'s `lostcancel` misses it; the closure defeats it). Same bug class, eight lines above your first hunk. Feel free to verify and file it here or separately.

baptmont (approved):

PR #1168, opened 2026-07-17
feat(mcptoolset): surface elicitation and tool result _meta

### LINKED ISSUE #1165: mcptoolset: tool result `_meta` is discarded, leaving no path for MCP server auth challenges (e.g. SEP-1036 URL elicitation)

`mcptoolset` flattens every `CallToolResult` to `map[string]any{"output": <text or structuredContent>}` and discards the result's `_meta` field entirely (`tool/mcptoolset/tool.go`, `Run`). The MCP spec reserves `_meta` precisely so servers can attach metadata to responses, and a growing class of servers uses it to signal out-of-band actions on tool results.

The concrete case: MCP gateways that front multiple backends with per-user OAuth. When the user is not authenticated to a backend, the gateway returns a tool result carrying an authentication challenge (backend name, login URL) in `_meta`, alongside human-readable text. With adk-go today, an agent or its host application can never see that challenge: by the time the tool result reaches any extensible layer (callbacks, flow, A2A conversion), only the flattened text remains. Host applications are reduced to regexing login URLs out of model output.

Relatedly, `Run` has a `// TODO: add auth` at the `CallTool` site, and MCP has since standardized URL-mode elicitation (SEP-1036, spec 2025-11-25: `elicitation/create` with `mode: "url"`, error `-32042`) as the native "user must visit this URL" mechanism. There is currently no way for that to surface through `mcptoolset` either.

Note this is adjacent to but distinct from #574: that issue is about ADK-native tool auth (`adk_request_credential` parity with adk-python). This one is about auth signals originating *in the MCP server*, which need to reach the embedding application even once #574 lands.

### Proposed Solution

Two independent pieces, smallest first:

1. **Stop dropping `_meta`.** Either include it in the returned function response map under a reserved key (e.g. `"_meta"`, matching what a raw MCP result would serialize to), or add an optional hook to `mcptoolset.Config` (e.g. `OnToolResult func(ctx, *mcp.CallToolParams, *mcp.CallToolResult)`) so embedders can observe the raw result before flattening. The hook variant is fully backward compatible.

2. **Surface MCP auth signals.** When the transport/client receives a SEP-1036 URL elicitation (or the server returns the URL-elicitation-required error) during a tool call, expose it to the embedding application rather than failing opaquely, so hosts can present the URL to the user and resume. This could build on the same hook, or on the go-sdk's elicitation handler plumbed through `mcptoolset.Config`.

### Impact on your work

We run Go agents (via kagent) against an MCP gateway with per-user OAuth to backends. Because `_meta` is dropped, we currently work around this by passing a custom `*mcp.Client` via `mcptoolset.Config.Client` with sending middleware that intercepts `tools/call` results before the toolset flattens them. It works, but it re-implements observation below the toolset and every adk-go embedder with the same need must rediscover the trick. First-class support would let hosts map server auth challenges to their protocol of choice (in our case the A2A `auth-required` task state, which the A2A spec section 7.6 defines for exactly this).

### Alternatives Considered

1. **Client middleware via `Config.Client`** (our current workaround): works through public API, but couples embedders to go-sdk middleware internals and only covers `_meta`, not elicitation.
2. **Put the challenge in tool result text and parse it downstream**: fragile, breaks on any wording change; this is what we're trying to eliminate.
3. **Wrapping `tool.Toolset`**: not viable, the flattening happens inside `mcpTool.Run` before any wrapper sees the result.

QuentinBisson:

- Closes: #1165

**Problem:**

`mcptoolset` flattens every `CallToolResult` to `{"output": ...}` and drops the result's `_meta` field, and there is no way to handle elicitation requests: servers using URL-mode elicitation (SEP-1036, spec 2025-11-25) for out-of-band flows such as per-user OAuth challenges fail opaquely inside the toolset (`client does not support "url" elicitation`). go-sdk already ships the full client surface (`ClientOptions.ElicitationHandler`, `ElicitationCompleteHandler`, `URLElicitationCapabilities`), so this is purely mcptoolset plumbing.

**Solution:**

Two independent pieces, as proposed in #1165:

1. `Config.ElicitationHandler` and `Config.ElicitationCompleteHandler`, applied to the default MCP client. Setting a handler declares both form and URL elicitation capabilities (the SDK's inferred capability covers form mode only; `RootsV2` preserves the default roots capability that setting `Capabilities` would otherwise disable). Combining the handlers with a custom `Config.Client` returns an error from `New`, since handlers must be configured in that client's own `mcp.ClientOptions`.
2. Result `_meta` passthrough: when a tool result carries `Meta`, it is included in the returned function response map under the `"_meta"` key, mirroring the raw MCP serialization. Results without `Meta` are unchanged.

The `_meta` passthrough drops the keys the MCP protocol reserves for itself (the `modelcontextprotocol`/`mcp` prefixes, plus the unprefixed `progressToken`, `traceparent`, `tracestate` and `baggage`), so only the server's own metadata reaches the model. A result that is marked as an error carries the same metadata on the exported `ToolError`, which an `llmagent.Config.OnToolErrorCallbacks` entry reads with `errors.As`. A result with no content at all keeps its `_meta` with an empty output; a result whose content the toolset cannot render still fails, with or without `_meta`.

3. `Tools` drops a server tool that advertises a name the framework dispatches itself, and logs it. `IsReservedToolName` covers `transfer_to_agent`, `adk_request_credential`, `adk_request_confirmation`, `stop_streaming` (served by the live flow before the tool lookup) and `task_completed` (injected by `sequentialagent`). Dropping rather than erroring is what keeps the rest of the toolset usable for every caller: `tool.FilterToolset` returns the inner `Tools()` error, so a caller following the `Config.ToolFilter` deprecation has no escape hatch.

I went with the reserved-key shape for `_meta` rather than an `OnToolResult` hook (the other option floated in #1165) because it is data-only and involves no new callback surface; happy to rework to the hook shape if preferred.

### Testing Plan

**Unit Tests:**

New tests in `tool/mcptoolset`, all running real MCP protocol traffic over in-memory transports against a real `mcp.Server`:

- `TestCallToolMeta`: text, structured and content-free results with `_meta` preserved; no `_meta` key when the result has none; a non-text result still fails whether or not it carries `_meta`.
- `TestCallToolErrorCarriesServerMeta`: an error result reaches the caller as a `*ToolError` carrying the metadata.
- `TestIsReservedMetaKey` / `TestServerMeta`: reserved-key filtering, prefixed and unprefixed.
- `TestElicitationHandler`: server raises a URL-mode elicitation mid tool call; the configured handler receives the URL and the call completes.
- `TestElicitationCompleteHandler`: full URL-mode sequence. The server answers the `tools/call` only after it reads the client's reply to the elicitation, and that reply arrives only once the completion handler releases the elicitation handler, so the test fails if `ElicitationCompleteHandler` is not wired. Its server is a raw JSON-RPC peer, because the SDK exports no server API for `notifications/elicitation/complete`.
- `TestToolsDropsReservedToolName`: each reserved name is dropped and the rest of the toolset is returned.
- `TestNewRejectsElicitationHandlerWithCustomClient` and `TestNewRejectsElicitationCompleteHandlerWithoutElicitationHandler`: configuration errors from `New`.

```
$ go test -race -count=1 -shuffle=on ./tool/mcptoolset/
ok  	google.golang.org/adk/v2/tool/mcptoolset
```

The elicitation test drives the full path end-to-end (real `mcp.Server` → `elicitation/create` with `mode: "url"` → client handler → tool result), which is the same wire sequence an MCP gateway produces. We (Giant Swarm) also plan to run this against our MCP gateway (per-user OAuth to backends, challenges mapped to the A2A `auth-required` task state) and can report back on the PR.

### Checklist

### Additional context

Python twin (elicitation callback only; `meta` already survives there): google/adk-python#6422 / google/adk-python#6423.

### DIFF
--- a/tool/mcptoolset/meta_test.go
+++ b/tool/mcptoolset/meta_test.go
@@ -40,6 +40,13 @@ func TestIsReservedMetaKey(t *testing.T) {
 		{key: "com.giantswarm.muster/authChallenge", want: false},
 		{key: "mcp/key", want: false},
 		{key: "bare", want: false},
+		// Unprefixed keys the protocol reserves as an exception to the prefix rule.
+		{key: "progressToken", want: true},
+		{key: "traceparent", want: true},
+		{key: "tracestate", want: true},
+		{key: "baggage", want: true},
+		// A key name holds no slash, so the prefix ends at the first one.
+		{key: "io.modelcontextprotocol/a/b", want: true},
 	}
 
 	for _, tc := range tests {
@@ -69,6 +76,8 @@ func TestServerMeta(t *testing.T) {
 			name: "reserved keys dropped alongside server keys",
 			meta: mcp.Meta{
 				mcp.MetaKeyServerInfo:  &mcp.Implementation{Name: "s"},
+				"progressToken":        "p1",
+				"traceparent":          "00-trace-span-01",
 				"com.example/auth":     map[string]any{"url": "https://idp.example.com/login"},
 				"com.example.mcp/note": "kept",
 			},
--- a/tool/mcptoolset/meta_test.go
+++ b/tool/mcptoolset/meta_test.go
@@ -1,0 +1,89 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mcptoolset
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+func TestIsReservedMetaKey(t *testing.T) {
+	tests := []struct {
+		key  string
+		want bool
+	}{
+		{key: mcp.MetaKeyServerInfo, want: true},
+		{key: mcp.MetaKeyClientInfo, want: true},
+		{key: mcp.MetaKeyProtocolVersion, want: true},
+		{key: mcp.MetaKeyClientCapabilities, want: true},
+		{key: mcp.MetaKeySubscriptionID, want: true},
+		{key: "io.modelcontextprotocol/anythingElse", want: true},
+		{key: "dev.mcp/key", want: true},
+		{key: "org.modelcontextprotocol.api/key", want: true},
+		{key: "com.mcp.tools/key", want: true},
+		{key: "com.example.mcp/key", want: false},
+		{key: "com.example/auth", want: false},
+		{key: "com.giantswarm.muster/authChallenge", want: false},
+		{key: "mcp/key", want: false},
+		{key: "bare", want: false},
+	}
+
+	for _, tc := range tests {
+		if got := isReservedMetaKey(tc.key); got != tc.want {
+			t.Errorf("isReservedMetaKey(%q) = %v, want %v", tc.key, got, tc.want)
+		}
+	}
+}
+
+func TestServerMeta(t *testing.T) {
+	tests := []struct {
+		name string
+		meta mcp.Meta
+		want map[string]any
+	}{
+		{
+			name: "nil meta",
+			meta: nil,
+			want: nil,
+		},
+		{
+			name: "only reserved keys",
+			meta: mcp.Meta{mcp.MetaKeyServerInfo: &mcp.Implementation{Name: "s"}},
+			want: nil,
+		},
+		{
+			name: "reserved keys dropped alongside server keys",
+			meta: mcp.Meta{
+				mcp.MetaKeyServerInfo:  &mcp.Implementation{Name: "s"},
+				"com.example/auth":     map[string]any{"url": "https://idp.example.com/login"},
+				"com.example.mcp/note": "kept",
+			},
+			want: map[string]any{
+				"com.example/auth":     map[string]any{"url": "https://idp.example.com/login"},
+				"com.example.mcp/note": "kept",
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			if diff := cmp.Diff(tc.want, serverMeta(tc.meta)); diff != "" {
+				t.Errorf("serverMeta() mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
--- a/tool/mcptoolset/meta_test.go
+++ b/tool/mcptoolset/meta_test.go
@@ -36,6 +36,12 @@ func TestIsReservedMetaKey(t *testing.T) {
 		{key: "org.modelcontextprotocol.api/key", want: true},
 		{key: "com.mcp.tools/key", want: true},
 		{key: "com.example.mcp/key", want: false},
+		// Only the second label carries the marker, so a marker at any other
+		// position leaves the key to the server.
+		{key: "modelcontextprotocol.io/key", want: false},
+		{key: "mcp.dev/key", want: false},
+		{key: "a.b.mcp.d/key", want: false},
+		{key: "tools.api.mcp.com/key", want: false},
 		{key: "com.example/auth", want: false},
 		{key: "com.giantswarm.muster/authChallenge", want: false},
 		{key: "mcp/key", want: false},
--- a/tool/mcptoolset/set.go
+++ b/tool/mcptoolset/set.go
@@ -59,6 +59,9 @@ func New(cfg Config) (tool.Toolset, error) {
 		if cfg.Client != nil {
 			return nil, fmt.Errorf("mcptoolset: ElicitationHandler and ElicitationCompleteHandler cannot be combined with a custom Client; set them in the client's mcp.ClientOptions instead")
 		}
+		if cfg.ElicitationHandler == nil {
+			return nil, fmt.Errorf("mcptoolset: ElicitationCompleteHandler requires ElicitationHandler to be set; the client cannot service an elicitation without it")
+		}
 		clientOptions = &mcp.ClientOptions{
 			ElicitationHandler:         cfg.ElicitationHandler,
 			ElicitationCompleteHandler: cfg.ElicitationCompleteHandler,
@@ -160,7 +163,9 @@ type Config struct {
 
 	// ElicitationCompleteHandler handles notifications/elicitation/complete
 	// notifications, which servers send when an out-of-band (URL-mode)
-	// elicitation has been completed.
+	// elicitation has been completed. It requires ElicitationHandler to also
+	// be set, since a completion notification cannot arrive unless an
+	// elicitation was created first.
 	// It can only be set when Client is nil; for a custom Client, set the
 	// handler in the client's mcp.ClientOptions instead.
 	ElicitationCompleteHandler func(context.Context, *mcp.ElicitationCompleteNotificationRequest)
--- a/tool/mcptoolset/set.go
+++ b/tool/mcptoolset/set.go
@@ -16,6 +16,7 @@
 package mcptoolset
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 
@@ -53,8 +54,29 @@ func New(cfg Config) (tool.Toolset, error) {
 	if err != nil {
 		return nil, err
 	}
+	var clientOptions *mcp.ClientOptions
+	if cfg.ElicitationHandler != nil || cfg.ElicitationCompleteHandler != nil {
+		if cfg.Client != nil {
+			return nil, fmt.Errorf("mcptoolset: ElicitationHandler and ElicitationCompleteHandler cannot be combined with a custom Client; set them in the client's mcp.ClientOptions instead")
+		}
+		clientOptions = &mcp.ClientOptions{
+			ElicitationHandler:         cfg.ElicitationHandler,
+			ElicitationCompleteHandler: cfg.ElicitationCompleteHandler,
+			// The capability inferred from ElicitationHandler alone covers only
+			// form mode; URL mode must be declared explicitly. RootsV2 preserves
+			// the default roots capability, which setting Capabilities would
+			// otherwise disable.
+			Capabilities: &mcp.ClientCapabilities{
+				Elicitation: &mcp.ElicitationCapabilities{
+					Form: &mcp.FormElicitationCapabilities{},
+					URL:  &mcp.URLElicitationCapabilities{},
+				},
+				RootsV2: &mcp.RootCapabilities{ListChanged: true},
+			},
+		}
+	}
 	return &set{
-		mcpClient:                   newConnectionRefresher(cfg.Client, transport),
+		mcpClient:                   newConnectionRefresher(cfg.Client, transport, clientOptions),
 		toolFilter:                  cfg.ToolFilter,
 		requireConfirmation:         cfg.RequireConfirmation,
 		requireConfirmationProvider: cfg.RequireConfirmationProvider,
@@ -127,6 +149,21 @@ type Config struct {
 	// before execution. If set to true, the ADK framework will automatically initiate
 	// a Human-in-the-Loop (HITL) confirmation request when a tool is invoked.
 	RequireConfirmation bool
+
+	// ElicitationHandler handles elicitation/create requests from the MCP
+	// server, including URL-mode elicitations that servers use for
+	// out-of-band interactions such as auth challenges. Setting it makes the
+	// client advertise the elicitation capability.
+	// It can only be set when Client is nil; for a custom Client, set the
+	// handler in the client's mcp.ClientOptions instead.
+	ElicitationHandler func(context.Context, *mcp.ElicitRequest) (*mcp.ElicitResult, error)
+
+	// ElicitationCompleteHandler handles notifications/elicitation/complete
+	// notifications, which servers send when an out-of-band (URL-mode)
+	// elicitation has been completed.
+	// It can only be set when Client is nil; for a custom Client, set the
+	// handler in the client's mcp.ClientOptions instead.
+	ElicitationCompleteHandler func(context.Context, *mcp.ElicitationCompleteNotificationRequest)
 
 	// RequireConfirmationProvider allows for dynamic determination of whether
 	// user confirmation is needed. This field is a function called at runtime to decide if
--- a/tool/mcptoolset/set.go
+++ b/tool/mcptoolset/set.go
@@ -24,6 +24,7 @@ import (
 
 	"google.golang.org/adk/v2/agent"
 	"google.golang.org/adk/v2/auth"
+	"google.golang.org/adk/v2/internal/llminternal"
 	"google.golang.org/adk/v2/tool"
 )
 
@@ -156,7 +157,17 @@ type Config struct {
 	// ElicitationHandler handles elicitation/create requests from the MCP
 	// server, including URL-mode elicitations that servers use for
 	// out-of-band interactions such as auth challenges. Setting it makes the
-	// client advertise the elicitation capability.
+	// client advertise the elicitation capability for both form and URL mode.
+	//
+	// For URL mode the handler must not return until the out-of-band
+	// interaction has completed: nothing waits for the server's
+	// notifications/elicitation/complete notification on its behalf, and the
+	// call is retried as soon as the handler returns. A handler that returns
+	// early makes the server re-request input until the retry budget is spent.
+	// A server that reports the URL through the CodeURLElicitationRequired
+	// (-32042) error code rather than through an elicitation request is not
+	// handled at all.
+	//
 	// It can only be set when Client is nil; for a custom Client, set the
 	// handler in the client's mcp.ClientOptions instead.
 	ElicitationHandler func(context.Context, *mcp.ElicitRequest) (*mcp.ElicitResult, error)
@@ -211,6 +222,10 @@ func (s *set) Tools(ctx agent.ReadonlyContext) ([]tool.Tool, error) {
 
 	var adkTools []tool.Tool
 	for _, mcpTool := range mcpTools {
+		if llminternal.IsReservedToolName(mcpTool.Name) {
+			return nil, fmt.Errorf("MCP server advertises tool %q, a name reserved by the framework", mcpTool.Name)
+		}
+
 		t, err := convertTool(mcpTool, s.mcpClient, s.requireConfirmation, s.requireConfirmationProvider)
 		if err != nil {
 			return nil, fmt.Errorf("failed to convert MCP tool %q to adk tool: %w", mcpTool.Name, err)
--- a/tool/mcptoolset/set.go
+++ b/tool/mcptoolset/set.go
@@ -69,7 +69,9 @@ func New(cfg Config) (tool.Toolset, error) {
 			// The capability inferred from ElicitationHandler alone covers only
 			// form mode; URL mode must be declared explicitly. RootsV2 preserves
 			// the default roots capability, which setting Capabilities would
-			// otherwise disable.
+			// otherwise disable. RootsV2 is deprecated as of the 2026-07-28
+			// revision (SEP-2577) but remains the only field that carries
+			// ListChanged, so it stays until the SDK offers a replacement.
 			Capabilities: &mcp.ClientCapabilities{
 				Elicitation: &mcp.ElicitationCapabilities{
 					Form: &mcp.FormElicitationCapabilities{},
@@ -168,6 +170,10 @@ type Config struct {
 	// (-32042) error code rather than through an elicitation request is not
 	// handled at all.
 	//
+	// One handler serves every tool of the toolset, and the tool calls of a
+	// single turn run on separate goroutines, so the handler must be safe for
+	// concurrent use.
+	//
 	// It can only be set when Client is nil; for a custom Client, set the
 	// handler in the client's mcp.ClientOptions instead.
 	ElicitationHandler func(context.Context, *mcp.ElicitRequest) (*mcp.ElicitResult, error)
@@ -176,7 +182,9 @@ type Config struct {
 	// notifications, which servers send when an out-of-band (URL-mode)
 	// elicitation has been completed. It requires ElicitationHandler to also
 	// be set, since a completion notification cannot arrive unless an
-	// elicitation was created first.
+	// elicitation was created first. Like ElicitationHandler, it must be safe
+	// for concurrent use.
+	//
 	// It can only be set when Client is nil; for a custom Client, set the
 	// handler in the client's mcp.ClientOptions instead.
 	ElicitationCompleteHandler func(context.Context, *mcp.ElicitationCompleteNotificationRequest)
@@ -222,10 +230,6 @@ func (s *set) Tools(ctx agent.ReadonlyContext) ([]tool.Tool, error) {
 
 	var adkTools []tool.Tool
 	for _, mcpTool := range mcpTools {
-		if llminternal.IsReservedToolName(mcpTool.Name) {
-			return nil, fmt.Errorf("MCP server advertises tool %q, a name reserved by the framework", mcpTool.Name)
-		}
-
 		t, err := convertTool(mcpTool, s.mcpClient, s.requireConfirmation, s.requireConfirmationProvider)
 		if err != nil {
 			return nil, fmt.Errorf("failed to convert MCP tool %q to adk tool: %w", mcpTool.Name, err)
@@ -233,6 +237,12 @@ func (s *set) Tools(ctx agent.ReadonlyContext) ([]tool.Tool, error) {
 
 		if s.toolFilter != nil && !s.toolFilter(ctx, t) {
 			continue
+		}
+
+		// Checked after the filter so that excluding the tool by name remains a
+		// way to keep the rest of the toolset usable.
+		if llminternal.IsReservedToolName(mcpTool.Name) {
+			return nil, fmt.Errorf("MCP server advertises tool %q, a name reserved by the framework", mcpTool.Name)
 		}
 
 		adkTools = append(adkTools, t)
--- a/tool/mcptoolset/set.go
+++ b/tool/mcptoolset/set.go
@@ -124,6 +124,14 @@ func authHTTPClient(base *http.Client, provider auth.CredentialProvider) *http.C
 }
 
 // Config provides initial configuration for the MCP ToolSet.
+//
+// A server tool whose name the framework dispatches itself is dropped from the
+// toolset and logged, because a call to that name never reaches the tool. The
+// drop is unconditional: transfer_to_agent is registered only for an agent
+// that has transfer targets, and task_completed only by sequentialagent during
+// RunLive, so an agent that registers neither still loses a server tool that
+// carries one of those names. ToolFilter runs first and cannot keep such a
+// tool.
 type Config struct {
 	// Client is an optional custom MCP client to use. If nil, a default client will be created.
 	Client *mcp.Client
--- a/tool/mcptoolset/set.go
+++ b/tool/mcptoolset/set.go
@@ -18,6 +18,7 @@ package mcptoolset
 import (
 	"context"
 	"fmt"
+	"log"
 	"net/http"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
@@ -160,6 +161,10 @@ type Config struct {
 	// server, including URL-mode elicitations that servers use for
 	// out-of-band interactions such as auth challenges. Setting it makes the
 	// client advertise the elicitation capability for both form and URL mode.
+	// There is no way to advertise one mode alone, so the handler must service
+	// both: a handler that rejects URL mode makes the server re-request input
+	// until the retry budget is spent. Set the handler in a custom Client's
+	// mcp.ClientOptions to declare a narrower capability.
 	//
 	// For URL mode the handler must not return until the out-of-band
 	// interaction has completed: nothing waits for the server's
@@ -170,9 +175,13 @@ type Config struct {
 	// (-32042) error code rather than through an elicitation request is not
 	// handled at all.
 	//
-	// One handler serves every tool of the toolset, and the tool calls of a
-	// single turn run on separate goroutines, so the handler must be safe for
-	// concurrent use.
+	// ElicitRequest.Params.URL arrives unprompted from the server and carries
+	// no scheme restriction, so the handler must validate it before it opens
+	// the URL or shows it to a user.
+	//
+	// One handler serves every tool of the toolset for the whole lifetime of
+	// the toolset, across every invocation, and concurrent tool calls reach it
+	// on separate goroutines, so it must be safe for concurrent use.
 	//
 	// It can only be set when Client is nil; for a custom Client, set the
 	// handler in the client's mcp.ClientOptions instead.
@@ -239,10 +248,11 @@ func (s *set) Tools(ctx agent.ReadonlyContext) ([]tool.Tool, error) {
 			continue
 		}
 
-		// Checked after the filter so that excluding the tool by name remains a
-		// way to keep the rest of the toolset usable.
+		// The framework serves a call to a reserved name itself, so the tool is
+		// unreachable. Dropping it keeps the rest of the toolset usable.
 		if llminternal.IsReservedToolName(mcpTool.Name) {
-			return nil, fmt.Errorf("MCP server advertises tool %q, a name reserved by the framework", mcpTool.Name)
+			log.Printf("adk: mcptoolset: MCP server advertises tool %q, a name reserved by the framework; dropping it", mcpTool.Name)
+			continue
 		}
 
 		adkTools = append(adkTools, t)
--- a/tool/mcptoolset/set_test.go
+++ b/tool/mcptoolset/set_test.go
@@ -16,6 +16,7 @@ package mcptoolset_test
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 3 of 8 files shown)


tool/mcptoolset/set.go
```diff
@@ -47,8 +48,29 @@ import (
 //		},
 //	})
 func New(cfg Config) (tool.Toolset, error) {
+	var clientOptions *mcp.ClientOptions
+	if cfg.ElicitationHandler != nil || cfg.ElicitationCompleteHandler != nil {
+		if cfg.Client != nil {
+			return nil, fmt.Errorf("mcptoolset: ElicitationHandler and ElicitationCompleteHandler cannot be combined with a custom Client; set them in the client's mcp.ClientOptions instead")
+		}
```
karolpiotrowicz, line 62: The capability branch is gated on either handler being set, but only `ElicitationHandler` can actually service an incoming `elicitation/create`. If a caller sets only `ElicitationCompleteHandler`, the client advertises form+URL elicitation yet the SDK rejects the request at runtime with `"client does not support elicitation"` — and a completion notification can't arrive without a prior elicitation anyway, so this config can't work.

Simplest fix is to fail fast in `New`, e.g. right after the custom-client check:

```suggestion
		}
		if cfg.ElicitationHandler == nil {
			return nil, fmt.Errorf("mcptoolset: ElicitationCompleteHandler requires ElicitationHandler to be set")
		}
```

(Alternatively, gate just the `Capabilities.Elicitation` advertisement on `cfg.ElicitationHandler != nil` — but since the complete-handler-only case is inert, erroring is clearer.)

tool/mcptoolset/tool.go
```diff
@@ -169,9 +167,22 @@ func (t *mcpTool) Run(ctx agent.Context, args any) (map[string]any, error) {
...
-		"output": textResponse.String(),
-	}, nil
+	return functionResponse(res, textResponse.String()), nil
+}
+
+// functionResponse builds the function response map for a tool result.
+// The result's _meta field is preserved under the "_meta" key, mirroring the
+// raw MCP serialization, so that metadata attached by the server (e.g. auth
+// challenges from MCP gateways) reaches callbacks and the embedding
+// application instead of being silently dropped.
```
karolpiotrowicz: Small doc clarification: this map is the function response returned to the model, so `_meta` reaches the LLM (and is persisted to session/traces), not only "callbacks and the embedding application." Could you extend the comment to say so? It's the intended behavior per #1165, but the model being a consumer is the important part for anyone reading this later.

karolpiotrowicz (commented):
Thanks for this, @QuentinBisson! I know this is still a draft, but since it's linked to #1165 I took an early look — and it's already in really good shape, so I wanted to share some feedback now rather than make you wait. Appreciate the work you've put in here, especially the clean split into the two independent pieces (`_meta` passthrough + elicitation plumbing) and the in-memory-transport tests running real protocol traffic against a live server.

I pulled the branch and it holds up well: the full test suite passes, `-race` is clean on the package, and it's backward compatible (`_meta` is only added when present, and the two new `Config` fields are purely additive).

None of this is urgent given the draft status — just flagging it early so it's easy to fold in. One thing I'd want to resolve before it comes out of draft, plus a couple of doc clarifications and an optional design question:

**1. `ElicitationCompleteHandler` set on its own advertises a capability the client can't service.** `New` enters the capability branch on `ElicitationHandler != nil || ElicitationCompleteHandler != nil`, so setting only `ElicitationCompleteHandler` advertises form+URL elicitation while `ElicitationHandler` stays `nil` — and the SDK then rejects any incoming `elicitation/create` with `"client does not support elicitation"`. Since a completion notification can't arrive unless an elicitation was created first, this combination is effectively unusable. I'd suggest failing fast in `New` (details inline).

**2. Doc accuracy on `_meta`.** The `functionResponse` comment says the metadata reaches "callbacks and the embedding application" — worth adding that it's also included in the function response returned to the model (and therefore persisted in session/traces). That's the intended behavior per #1165, but it's the most consequential consumer, so it's good to spell out.

**3. (Optional / your call) Opt-in vs. always-on for `_meta`.** Because the metadata now flows to the model for *every* server that emits it, users of meta-emitting servers will start seeing it surface without opting in. The reserved-key shape you chose is faithful to #1165 — just flagging the always-on aspect in case you'd prefer a flag or the `OnToolResult` hook that was also floated.

A few optional test/robustness follow-ups can wait until it's marked ready. Thanks again for driving this — happy to review again whenever you flip it out of draft.

QuentinBisson:
Thanks for the early look, @karolpiotrowicz, and for pulling the branch and running `-race` — really appreciate the thoroughness while it's still draft.

**1. `ElicitationCompleteHandler` without `ElicitationHandler`** — good catch, fixed. `New` now fails fast right after the custom-client check: an `ElicitationCompleteHandler` with no `ElicitationHandler` returns an error, since the client can't service an `elicitation/create` and a completion can't arrive without a prior elicitation. Added a field-doc note and a regression test (`TestNewRejectsElicitationCompleteHandlerWithoutElicitationHandler`). I went with erroring over silently gating the capability advertisement, per your suggestion — the inert config should be loud.

**2. `_meta` doc accuracy** — fixed. The `functionResponse` comment now spells out that the map is the function response returned to the model, so `_meta` reaches the LLM and is persisted to session/traces, on top of callbacks and the embedding application.

**3. Always-on vs opt-in for `_meta`** — I'd like to keep it always-on. The Python twin (google/adk-python) already does this: `_run_async_impl` returns `response.model_dump(exclude_none=True)`, i.e. the full `CallToolResult` including `_meta`, unconditionally whenever the server emits it — no flag, no hook. Making Go opt-in would diverge from the established twin for no correctness gain, and the reserved-key shape is what #1165 settled on. If the always-on surface turns out to be a problem in practice, an `OnToolResult` hook is a clean additive follow-up rather than something that needs to block this. Flagging the always-on behavior explicitly here so it's on record as a conscious decision.

Both fixes are pushed. Marking ready for review.

karolpiotrowicz (commented):
Both threads from the earlier round are properly resolved. The `New` guard is exactly the right shape and [`set_test.go:1069`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/tool/mcptoolset/set_test.go#L1069) pins it, the `functionResponse` doc now says plainly that the map reaches the model and is persisted to session and traces, and the `_meta` filtering added since is a real improvement over dropping the metadata wholesale. The second-label rule in [`isReservedMetaKey`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/tool/mcptoolset/tool.go#L227-L241) matches the 2026-07-28 spec exactly, `com.example.mcp/` counter-example included, which is the part that is easy to get wrong.

One thing needs fixing before this goes in, and it comes from the reserved-tool-name check rather than from the `_meta` work.

**A server-advertised tool name can take the whole agent down, and the operator cannot prevent it.** The check at [`set.go:225-227`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/tool/mcptoolset/set.go#L225-L227) runs before the tool filter at [`set.go:234`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/tool/mcptoolset/set.go#L234), so a `ToolFilter` that explicitly excludes the offending tool does not help. The error is not scoped to the toolset either: [`tools_processor.go:42-44`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/internal/llminternal/tools_processor.go#L42-L44) yields it and returns, which fails the whole invocation. So a remote MCP server that happens to expose a tool called `transfer_to_agent` disables every tool on the agent, with no configuration available to work around it.

```go
// tool/mcptoolset/set_test.go
func TestReservedNameDefeatsToolFilter(t *testing.T) {
	ts, err := mcptoolset.New(mcptoolset.Config{
		Transport:  srv(t, "get_weather", "transfer_to_agent"), // real mcp.Server, in-memory transport
		ToolFilter: tool.StringPredicate([]string{"get_weather"}),
	})
	if err != nil {
		t.Fatalf("New: %v", err)
	}
	inv := icontext.NewInvocationContext(t.Context(), icontext.InvocationContextParams{})
	tools, err := ts.Tools(icontext.NewReadonlyContext(inv))
	t.Logf("tools=%d err=%v", len(tools), err)
	if err != nil {
		t.Fatal("get_weather is unreachable even though transfer_to_agent was filtered out")
	}
}
```

```
$ go test -run TestReservedNameDefeatsToolFilter ./tool/mcptoolset/
    tools=0 err=MCP server advertises tool "transfer_to_agent", a name reserved by the framework
    get_weather is unreachable even though transfer_to_agent was filtered out
--- FAIL: TestReservedNameDefeatsToolFilter (0.00s)
```

Worth knowing while you decide how to handle it: on an agent that already has transfer targets, a server tool named `transfer_to_agent` was rejected before this change too, by the duplicate check in [`toolutils.go:47-49`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/tool/toolutils/toolutils.go#L47-L49). The newly-broken cases are a single agent with no transfer targets meeting that name, and any agent meeting `adk_request_credential` or `adk_request_confirmation`. The first of those is the ordinary configuration, and it is also the one where there is nothing to hijack, so failing closed there costs availability without buying protection. Guarding the framework's dispatch names is worth doing, so this is about where the check sits rather than whether it should exist.

**Two reserved-key cases slip through the `_meta` filter.** The spec reserves four unprefixed keys — `progressToken`, `traceparent`, `tracestate` and `baggage` — as an explicit exception to the prefix rule, and [`tool.go:228-231`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/tool/mcptoolset/tool.go#L228-L231) returns early for any key with no slash, so none of them are dropped. These are request-direction keys, so this bites when a server or gateway echoes them onto a result, and the effect is that protocol metadata lands in the function response and from there in the model's context and the session. Separately, [`tool.go:228`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/tool/mcptoolset/tool.go#L228) uses `strings.LastIndex`, but the spec terminates the prefix at the first slash and a key name may not contain one, so a key like `io.modelcontextprotocol/a/b` parses to a second label of `modelcontextprotocol/a` and is kept. That second one only admits malformed keys, so it matters much less than the first.

```
$ go test -run TestUnprefixedReservedKeys ./tool/mcptoolset/
    isReservedMetaKey("progressToken")                     = false
    isReservedMetaKey("traceparent")                       = false
    isReservedMetaKey("tracestate")                        = false
    isReservedMetaKey("baggage")                           = false
    isReservedMetaKey("io.modelcontextprotocol/a/b")       = false
    isReservedMetaKey("io.modelcontextprotocol/serverInfo") = true
```

The table at [`meta_test.go:29-42`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/tool/mcptoolset/meta_test.go#L29-L42) is otherwise a good regression net, and adding those cases to it would have caught both.

**A successful result carrying only `_meta` still loses it.** When the server returns no text content and no structured content, [`tool.go:160-162`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/tool/mcptoolset/tool.go#L160-L162) returns a bare error, so the metadata reaches neither the response nor a `ToolError`. That line predates this PR, so it is not something the change broke, but a metadata-only auth challenge is close enough to the motivating case that it is worth deciding deliberately rather than leaving it as the one path that drops `_meta`.

A few smaller things, none of which need to block:

- **The `ToolError` doc could name the hook that can actually read `Meta`.** [`tool.go:167-169`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/tool/mcptoolset/tool.go#L167-L169) is accurate as written, and `errors.As` does work in an error callback. It is just not obvious from the comment that the reachable caller is `OnToolErrorCallbacks` on [`llmagent.go:333`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/agent/llmagent/llmagent.go#L333), or that [`base_flow.go:1295-1297`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/internal/llminternal/base_flow.go#L1295-L1297) flattens the error to its message for the model, so `Meta` is deliberately not model-visible the way success-path `_meta` is.
- **The elicitation handler fields do not mention concurrency.** One handler closure is shared by every tool on the toolset, and tool calls in a turn run on separate goroutines, so a handler can be entered concurrently. The docs at [`set.go:157-182`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/tool/mcptoolset/set.go#L157-L182) are detailed about the URL-mode blocking obligation, which makes the silence on concurrency read as "no obligation".
- **`RootsV2` at [`set.go:78`](https://github.com/google/adk-go/blob/e3e14c6dbbd81c3da77e8b12695770be8c3afe38/tool/mcptoolset/set.go#L78) is deprecated as of 2026-07-28 (SEP-2577).** The reason for setting it is sound and the field stays functional through the deprecation window, so this is only worth a comment noting why it is still used.
- **`ElicitationCompleteHandler` has no happy-path test.** The rejection case is covered, but nothing exercises both handlers set with a completion notification actually delivered, which is the behavior the field exists for.
- **The description is now missing some of the change.** It does not mention the exported `ToolError`, or that `Tools` can fail on a reserved tool name, both of which are things a user of the package would want to find there.

The in-memory-transport tests running real protocol traffic against a live server continue to be the best part of this, and they are what made the `_meta` behavior straightforward to check.

QuentinBisson:
Thank @karolpiotrowicz those are now addressed :)

karolpiotrowicz (dismissed):
All four items from the last round are properly fixed, and the fixes are the right shape rather than the minimum that would pass. The reserved-name check moved below the filter, the four unprefixed protocol keys and the first-slash prefix parse are both handled, the metadata-only result now surfaces instead of erroring, and the new test cases in [`meta_test.go`](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/tool/mcptoolset/meta_test.go#L40-L50) are exactly the ones that would have caught the two `_meta` bugs. Standing up a raw JSON-RPC peer to exercise `notifications/elicitation/complete` because the SDK exposes no server API for it is more effort than this needed.

One new correctness bug, and two things worth knowing before merge.

**The metadata-only branch keys on the wrong emptiness test.** [`tool.go:160-167`](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/tool/mcptoolset/tool.go#L160-L167) enters the new branch when `textResponse.Len() == 0`, which is also true for a result whose content is non-text. So an image-only result that happens to carry server `_meta` now returns an empty output and reports success, while the identical result without `_meta` still returns an error. The image is dropped either way, so nothing new is lost, but the caller loses the signal that anything went wrong, and `{"output": ""}` is indistinguishable from a genuine empty string. Keying the branch on `len(res.Content) == 0` restores the distinction.

```
$ go test -run TestMetaOnlyDivergence ./tool/mcptoolset/
    image + unrelated _meta -> res=map[string]interface {}{"_meta":map[string]interface {}{"com.example/x":1}, "output":""}  err=<nil>
    image, no _meta         -> res=map[string]interface {}(nil)                                                             err=no text content in tool response
```

<details><summary>The test that produces it</summary>

```go
func TestMetaOnlyDivergence(t *testing.T) {
	img := &mcp.ImageContent{Data: []byte{1, 2, 3}, MIMEType: "image/png"}

	withMeta, err1 := runTool(t, func(ctx context.Context, r *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
		return &mcp.CallToolResult{Content: []mcp.Content{img}, Meta: mcp.Meta{"com.example/x": 1}}, nil
	})
	t.Logf("image + unrelated _meta -> res=%#v err=%v", withMeta, err1)

	noMeta, err2 := runTool(t, func(ctx context.Context, r *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
		return &mcp.CallToolResult{Content: []mcp.Content{img}}, nil
	})
	t.Logf("image, no _meta          -> res=%#v err=%v", noMeta, err2)

	if err1 == nil && err2 != nil {
		t.Errorf("an unrelated _meta key turns a hard error into a silent empty success")
	}
}
```
</details>

**The new elicitation test passes with the feature disconnected.** In [`elicitation_complete_test.go`](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/tool/mcptoolset/elicitation_complete_test.go#L188-L204) the raw server writes the elicitation request, then the completion notification, then the `tools/call` response, without ever reading the client's reply to the elicitation. The tool call therefore completes from the already-written response whether or not the completion handler ran. Deleting the `ElicitationCompleteHandler` line from the client options in `set.go` leaves `go test -run TestElicitationCompleteHandler` reporting `ok`, so the file does not currently guard the thing it was written for. Having the server block on reading the client's response to `elicitID` before answering the `tools/call` would make it real.

Two smaller things in the same file: the [`select` on `elicited`](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/tool/mcptoolset/elicitation_complete_test.go#L85-L89) has a `default` branch with no happens-before edge against the handler goroutine, so it can report "handler was never called" spuriously under load, and [`srv.elicitedURL()`](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/tool/mcptoolset/elicitation_complete_test.go#L93-L95) compares the server's own `loginURL` constant against itself while its message says it is checking what the handler received. Capturing `req.Params.URL` inside the handler, as `TestElicitationHandler` already does, would make that assertion mean what it says.

**The deprecation notice points at an API that cannot express the exclusion.** The comment at [`set.go:242`](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/tool/mcptoolset/set.go#L241-L244) says excluding the tool by name keeps the rest of the toolset usable, and that is true for `Config.ToolFilter`. But that field is marked [`// Deprecated: use tool.FilterToolset instead`](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/tool/mcptoolset/set.go#L148), and [`filteredToolset.Tools`](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/tool/tool.go#L105-L109) calls the inner `Tools()` first and returns its error, so a caller who follows the deprecation gets no escape hatch. `agentregistry` constructs the toolset with no filter at all. This is much less serious than it looks, because `set.Tools` already fails the whole call on any `ListTools` error and always has, so a misbehaving server could do this before the change. It is worth resolving as an API-consistency matter rather than a vulnerability, and skipping the offending tool rather than erroring would close it for every caller at once.

A few smaller items, none of which need to block:

- **`IsReservedToolName` is missing two names its own doc covers.** It says "a function call name that the flow dispatches itself", but [`stop_streaming`](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/internal/llminternal/base_flow.go#L1138) is intercepted before the tool lookup and [`task_completed`](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/internal/llminternal/base_flow.go#L550) ends a sequential-agent step, and neither is in [the list](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/internal/llminternal/functions.go#L30-L36).
- **There is no way to advertise form-only elicitation.** [`set.go:75-79`](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/tool/mcptoolset/set.go#L75-L79) always declares both modes, so a caller with a form-only handler will receive URL requests it cannot service and hit the retry behaviour the field doc describes. Worth a sentence in the doc if it is a deliberate simplification.
- **The concurrency note understates the sharing.** [`set.go:173-175`](https://github.com/google/adk-go/blob/40e8218cb966bcea7308bc09b2f3247afc49c406/tool/mcptoolset/set.go#L173-L175) scopes it to "a single turn", but one handler closure and one client session serve the toolset for the agent's whole lifetime, across every invocation.
- **The elicitation docs never say the URL is untrusted.** `req.Params.URL` arrives unprompted from the server and carries no scheme restriction, which matters because the natural implementation hands it to a browser.
- `serverMeta` is computed twice on the metadata-only path, and `unprefixedReservedMetaKeys` would read more naturally as a `switch` inside `isReservedMetaKey` than as a package-level map.

The `_meta` filtering itself holds up well. It matches the second-label rule in the 2026-07-28 spec in both directions, including the `com.example.mcp/` case that is easy to get backwards, and the malformed-key edges I could find (`.mcp/x` over-filtered, `a..mcp/k` under-filtered) are both unreachable for a conforming server and harmless either way, since a server wanting to reach the model can just use its own vendor prefix.

QuentinBisson (commented):
Thanks, all three of the substantial items were real. Fixed in 898423a.

**Metadata-only branch keyed on the wrong emptiness test.** Correct, and the divergence was exactly as you reproduced it. The branch now keys on `len(res.Content) == 0`, so a non-text result fails with or without `_meta`. `TestCallToolMeta` gained a `non-text result with server meta` case and a `non-text result without server meta` case; reverting the fix makes the first one fail. `serverMeta` is now computed once, ahead of the error branch, and `functionResponse` takes the filtered map rather than the result.

**The completion test passed with the feature disconnected.** Also correct. The raw server now keeps the client's responses in a channel the serve loop does not consume, and `callTool` waits for the reply to the `elicitation/create` request before it answers the `tools/call` (with a bounded wait that answers a JSON-RPC error instead of hanging). Deleting `ElicitationCompleteHandler` from the client options now fails the test with `the client never answered the elicitation`, since the elicitation handler never returns. The `select`/`default` on the handler channel is gone: the handler pushes `req.Params.URL` and the assertion blocks on it, so it checks what the handler received rather than the server's own constant. `srv.elicitedURL()` is deleted.

**Reserved names error out with no escape hatch.** Taken as you suggest: `Tools` now logs and skips the offending tool instead of failing the whole call, which closes it for `tool.FilterToolset` and for `agentregistry` at the same time. `TestToolsRejectsReservedToolName` becomes `TestToolsDropsReservedToolName`, asserting the reserved tool is absent and the rest of the toolset is returned; the separate filter test is redundant now and is gone.

Smaller items:

- `IsReservedToolName` covers `stop_streaming` and `task_completed`, both behind new named constants that `base_flow.go` now uses in place of the literals.
- `Config.ElicitationHandler` doc now says both modes are always advertised, that a form-only handler must go through a custom `Client`'s `mcp.ClientOptions`, that `Params.URL` is untrusted and unrestricted in scheme, and that one handler and one session serve the toolset for its whole lifetime across every invocation, not for a single turn.
- `unprefixedReservedMetaKeys` is now a `switch` inside `isReservedMetaKey`.

`go test -race -count=1 -shuffle=on ./tool/mcptoolset/ ./internal/llminternal/... ./agent/workflowagents/...` is green, and `golangci-lint run` reports nothing new on the touched packages.

QuentinBisson:
Rebased onto main. One semantic note from the rebase: #1352 made an empty text result a success unless non-text blocks were dropped, so I took that guard and removed the separate metadata-only branch. A metadata-only result still surfaces with its `_meta`, a non-text result still errors with or without `_meta`, and the tests from the last round pass unchanged. The earlier behaviour where empty content with no `_meta` returned an error is gone, as it is on main. @karolpiotrowicz ready for another look when you have time.

karolpiotrowicz (dismissed):
Round 1's findings are all closed. The reserved-tool-name check now runs after the operator's `ToolFilter` and drops just that tool instead of aborting the invocation, the four unprefixed reserved keys are matched, the prefix scan uses the first slash, and an empty-text success keeps its `_meta`. Thanks for recutting onto current `main` as well.

One new problem, in the rule that is the point of the feature. Two things need to be fixed before merge: the reserved-key rule has to match every label of the prefix except the last, and `TestIsReservedMetaKey` needs at least one row that fails under the rule as it stands.

`isReservedMetaKey` treats a key as reserved when the **second** label of its prefix is `modelcontextprotocol` or `mcp`. The specification's rule is that the marker may sit at any position that is not the last: *"Any prefix beginning with zero or more valid labels, followed by `modelcontextprotocol` or `mcp`, followed by any valid label, is reserved for MCP use. For example: `modelcontextprotocol.io/`, `mcp.dev/`, `api.modelcontextprotocol.org/`, and `tools.mcp.com/` are all reserved."*

Two of those four examples fall through. Executed against the head:

```
modelcontextprotocol.io/x        got=false want=true   spec example 1
mcp.dev/x                        got=false want=true   spec example 2
api.modelcontextprotocol.org/x   got=true  want=true   spec example 3
tools.mcp.com/x                  got=true  want=true   spec example 4
a.b.modelcontextprotocol.c/x     got=false want=true   deeper prefix
a.b.mcp.d/x                      got=false want=true   deeper prefix
```

A key that falls through is kept by `serverMeta`, set as `response["_meta"]`, and travels out as the `genai.FunctionResponse.Response` — so it reaches the model and is persisted to the session event, which is what the filter exists to prevent.

What bounds this today: the Go SDK spells all five of its own reserved keys in reverse-DNS (`io.modelcontextprotocol/serverInfo` and siblings), and every one of those puts the marker at index 1, so nothing the SDK emits leaks. The exposure is to servers that use the specification's literal forms, which are equally valid.

Worth saying that you read the spec correctly in the harder direction: `mcp/key` and `com.example.mcp/key` both return false, and they should, because the marker has no following label. The rule is the mirror image of the right one rather than a careless one. Whatever shape the fix takes, it needs to keep that second property — a marker in the final position is still not reserved.

Two of the tests bless the defect rather than catching it, which is why it survived:

- Every `want: true` prefix row in `TestIsReservedMetaKey` puts the marker at index 1, so the table is satisfied by the current rule and would stay green after a correct fix. `tools.api.mcp.com/x`, `modelcontextprotocol.io/x` and `mcp.dev/x` would each discriminate.
- `TestToolsDropsReservedToolName` configures no `ToolFilter`, so moving the reserved check back in front of the filter — undoing the round-1 fix — leaves the suite green. A case with a filter that records the names it is offered would pin the ordering.

Smaller things:

- The "only non-text content" path returns a bare `fmt.Errorf` and discards the `meta` computed further up, which is asymmetric with the `IsError` path you deliberately changed to carry it. A server attaching an auth challenge under its own prefix loses it exactly when the result is unrenderable.
- The reserved-name drop is unconditional, but `transfer_to_agent` is registered only when the agent has transfer targets and `task_completed` only by `sequentialagent` during `RunLive`. A lone `llmagent` with no sub-agents whose MCP server exposes a tool called `transfer_to_agent` loses a tool that works on `main`, with only a log line. Either scope the check to names the framework actually registers for that invocation, or document the unconditional drop on `Config`.
- `_meta` forwarding is on by default with no opt-out, so a server's own bookkeeping keys now reach the model and the session on every call. That is what the feature asks for, so I mention it rather than object.

CI has never run on this head — three check suites are sitting at `action_required` waiting on maintainer approval, so everything above comes from a local run.

QuentinBisson:
Three of the four are fixed. The blocking one rests on a superseded revision.

The text you quote is **2025-06-18**. It was replaced in 2025-11-25 and reads the same in **2026-07-28**, the current revision:

> Any prefix where the **second label** is `modelcontextprotocol` or `mcp` is **reserved** for MCP use.
> For example: `io.modelcontextprotocol/`, `dev.mcp/`, `org.modelcontextprotocol.api/`, and `com.mcp.tools/` are all reserved.
> However, `com.example.mcp/` is NOT reserved, as the second label is `example`.

That is what `isReservedMetaKey` implements, and the test table is those five keys. go-sdk v1.7.0 puts `2026-07-28` first in `supportedProtocolVersions`.

"Any label except the last" would un-reserve every key the SDK emits, since all six spell the marker last:

```
io.modelcontextprotocol/protocolVersion       second-label=true   any-but-last=false
io.modelcontextprotocol/clientInfo            second-label=true   any-but-last=false
io.modelcontextprotocol/clientCapabilities    second-label=true   any-but-last=false
io.modelcontextprotocol/logLevel              second-label=true   any-but-last=false
io.modelcontextprotocol/subscriptionId        second-label=true   any-but-last=false
io.modelcontextprotocol/serverInfo            second-label=true   any-but-last=false
```

Your three rows are `want: false` under the current rule (second label `io`, `dev`, `api`). I added them plus `a.b.mcp.d/` as `false` rows, so the "any but last" rule now fails the table on all six SDK keys. Happy to cover the 2025-06-18 forms too as a superset rule if you want servers still on that revision handled.

Fixed:

- The non-text-only path returns `UnsupportedContentError{ToolName, Meta}` rather than a bare `fmt.Errorf`, so metadata survives an unrenderable result. Error string unchanged.
- `TestToolsDropsReservedToolName` installs a `ToolFilter` that records the names it is offered; moving the check ahead of the filter now fails.
- `Config` documents the unconditional drop. Scoping it would need the toolset to know the agent's transfer targets at `Tools()` time, which it does not. Say the word if you want that plumbing instead.

tool/mcptoolset/tool.go
```diff
@@ -176,12 +170,116 @@ func (t *mcpTool) Run(ctx agent.Context, args any) (map[string]any, error) {
...
+func (e *ToolError) Error() string {
+	if e.Details == "" {
+		return "Tool execution failed."
+	}
+	return "Tool execution failed. Details: " + e.Details
+}
+
+// UnsupportedContentError reports a tool result the toolset cannot render as a
+// function response, because every content block is of a type it does not
+// convert. Callers reach it with errors.As to read the metadata the server
```
karolpiotrowicz, line 206: This says the error means "every content block is of a type it does not convert", but the guard is `textResponse.Len() == 0 && droppedNonText`. A result of `[TextContent{Text: ""}, ImageContent{...}]` triggers it even though the text block *was* converted — it just contributed no text. The inline comment at lines 163-167 states the real rule correctly, so it is only this summary that is falsifiable.

No test covers that shape either: every row in the non-text table pairs a non-empty caption with the non-text block.

karolpiotrowicz (commented):
Nothing in the code blocks this, but the branch cannot be merged as it stands, and the reason needs a maintainer's decision rather than a rebase.

**First, a correction I owe you.** My last round called the reserved-`_meta`-prefix rule a bug, on the grounds that it missed `modelcontextprotocol.io/` and `mcp.dev/`. That was wrong, and the four rows you added in [this round's commit](https://github.com/google/adk-go/commit/35acf0b8683c85c3f2c9088ca878ee9a7f4a44ff) are right. The specification changed the rule between revisions: 2025-06-18 used a positional rule whose examples were the forward-DNS spellings, and both 2025-11-25 and 2026-07-28 replaced it with "any prefix where the second label is `modelcontextprotocol` or `mcp`", giving `io.modelcontextprotocol/`, `dev.mcp/`, `org.modelcontextprotocol.api/` and `com.mcp.tools/` as reserved and `com.example.mcp/` as explicitly not. I was quoting the superseded revision. Your implementation matches the current one exactly, the pre-existing table rows are that revision's own example list, and the SDK you pin targets 2026-07-28. Sorry for the detour — the rows you added are a good regression guard and should stay.

## The branch no longer merges, and neither obvious resolution compiles

[#1401](https://github.com/google/adk-go/commit/d56d1a808c46d70b3f2a12d7aa6b4ff9421e45c8) landed on main 42 minutes after you pushed this round, so this is timing rather than anything you did. It replaced the "only non-text content" error with a renderer that formats every content type. On current main a tool returning only an image now yields `{"output":"[MCP image: mimeType=\"image/png\", size=4 bytes]"}` and no error, and `droppedNonText` no longer exists anywhere in the package.

The textual conflict in `tool.go` understates the problem, because main also added `tool/mcptoolset/tool_test.go`, which merges cleanly and so raises no conflict marker. I merged your head into main and tried both resolutions:

```
# keep this PR's tool.go
$ go vet ./tool/mcptoolset/
tool/mcptoolset/tool_test.go:713:14: undefined: hasMIMECharsetParameter

# keep main's tool.go
$ go vet ./tool/mcptoolset/
tool/mcptoolset/meta_test.go:59:13: undefined: isReservedMetaKey
tool/mcptoolset/meta_test.go:99:33: undefined: serverMeta
```

So neither side can simply win, and reconciling them is a design question rather than a merge chore: should a result carrying only non-text content fail loudly, as this PR does, or render a placeholder and report success, as main now does? Main's behaviour reports success while discarding the payload, which is exactly what your comment at [tool.go:163-167](https://github.com/google/adk-go/blob/35acf0b8683c85c3f2c9088ca878ee9a7f4a44ff/tool/mcptoolset/tool.go#L163-L167) says it is there to prevent. Your own [TODO(#1401)](https://github.com/google/adk-go/blob/35acf0b8683c85c3f2c9088ca878ee9a7f4a44ff/tool/mcptoolset/tool.go#L169) anticipated this, which is why I would rather you got an answer than guessed.

I would hold off reworking until a maintainer says which behaviour survives. If main's renderer stands, `UnsupportedContentError` and `TestUnsupportedContentCarriesServerMeta` have no code path left. If failing loudly stands, the renderer needs the guard back. Either way the `_meta` threading — `serverMeta`, `ToolError`, `isReservedMetaKey` — is orthogonal and survives, since main has none of it.

## A scope question that is above both of us

This adds four exported symbols to `tool/mcptoolset`, and the reserved-`_meta` filter has no counterpart in adk-python: at [its current main](https://github.com/google/adk-python/commit/6de43b05a07f8ca89909664b3a21049646d7336c) it returns the whole dumped result to the model ([mcp_tool.py:636](https://github.com/google/adk-python/blob/6de43b05a07f8ca89909664b3a21049646d7336c/src/google/adk/tools/mcp_tool/mcp_tool.py#L636)) and deliberately treats `_meta` as opaque server-owned JSON ([mcp_tool.py:81-86](https://github.com/google/adk-python/blob/6de43b05a07f8ca89909664b3a21049646d7336c/src/google/adk/tools/mcp_tool/mcp_tool.py#L81-L86)). adk-go would be the only implementation that filters it.

I think filtering is the right instinct — a careless server can otherwise put arbitrary keys into the model's context and the persisted event. It is still a divergence from every sibling, so it wants a maintainer's sign-off before you spend more time on it. The elicitation half is straightforward parity, so that part is not in question ([mcp_toolset.py:172](https://github.com/google/adk-python/blob/6de43b05a07f8ca89909664b3a21049646d7336c/src/google/adk/tools/mcp_tool/mcp_toolset.py#L172)).

## Worth fixing whenever the rework happens

The godoc on the new type at [tool.go:204-206](https://github.com/google/adk-go/blob/35acf0b8683c85c3f2c9088ca878ee9a7f4a44ff/tool/mcptoolset/tool.go#L204-L206) says the error means "every content block is of a type it does not convert", but the guard is `textResponse.Len() == 0 && droppedNonText`, so an empty `TextContent` alongside an `ImageContent` also triggers it even though that text block was converted. Your inline comment two dozen lines up already states the real rule correctly.

The causal clause in the `Config` doc at [set.go:128-129](https://github.com/google/adk-go/blob/35acf0b8683c85c3f2c9088ca878ee9a7f4a44ff/tool/mcptoolset/set.go#L128-L129) — "because a call to that name never reaches the tool" — holds only for `stop_streaming`, which the flow intercepts before the tool lookup. The other four reserved names resolve through `toolsDict`, so a server tool carrying one of them would have been reached. Dropping them is still right, and the sentence that follows already discloses the over-drop, so it is only the reason that needs rewording.

Three smaller ones, none blocking. The `Meta` field doc says keys are dropped when they sit "in prefixes the MCP protocol reserves", but four unprefixed keys are dropped too. The new type could carry `ToolError`'s caveat that `Meta` never reaches the model, since the same is true of it. And one extra table row like `{key: "net.mcp/hint", want: true}` would pin the second-label rule itself rather than the four prefixes the spec happens to use as examples — I replaced the whole function body with a hardcoded list of those four prefixes and the package still passed.

Everything else is in good shape: build, vet, gofmt and the full suite are green, the API change is purely additive, and the filter-ordering assertion you added genuinely holds — inverting the order fails all five subtests.

tool/mcptoolset/tool.go
```diff
@@ -176,12 +170,116 @@ func (t *mcpTool) Run(ctx agent.Context, args any) (map[string]any, error) {
...
+// attached to the result, which a plain error message cannot carry.
+type UnsupportedContentError struct {
+	// ToolName is the name of the tool that returned the result.
+	ToolName string
+
+	// Meta holds the metadata the server attached to the result, without keys
+	// in prefixes the MCP protocol reserves for itself. It is nil when the
+	// server attached no metadata of its own.
```
karolpiotrowicz, line 214: "without keys in prefixes the MCP protocol reserves for itself" leaves out the four unprefixed keys `isReservedMetaKey` also drops — `progressToken`, `traceparent`, `tracestate` and `baggage`. A caller reading this would expect a bare `traceparent` to survive into `Meta`.

The same wording is on `ToolError.Meta`, so this is a two-site fix rather than something this commit introduced.

tool/mcptoolset/tool.go
```diff
@@ -176,12 +170,116 @@ func (t *mcpTool) Run(ctx agent.Context, args any) (map[string]any, error) {
...
+	if e.Details == "" {
+		return "Tool execution failed."
+	}
+	return "Tool execution failed. Details: " + e.Details
+}
+
+// UnsupportedContentError reports a tool result the toolset cannot render as a
+// function response, because every content block is of a type it does not
+// convert. Callers reach it with errors.As to read the metadata the server
+// attached to the result, which a plain error message cannot carry.
```
karolpiotrowicz, line 207: `ToolError`'s doc just above spells out that the flow renders the error for the model as its message alone, so `Meta` never reaches the model. That is equally true here — both are returned as errors from `Run`, so neither goes through `functionResponse` — but only the older type says so. Worth copying the caveat across, since the natural reading of "callers reach it with errors.As to read the metadata" is that the metadata gets somewhere.

tool/mcptoolset/set.go
```diff
@@ -95,6 +124,14 @@ func authHTTPClient(base *http.Client, provider auth.CredentialProvider) *http.C
 }
 
 // Config provides initial configuration for the MCP ToolSet.
+//
+// A server tool whose name the framework dispatches itself is dropped from the
+// toolset and logged, because a call to that name never reaches the tool. The
```
karolpiotrowicz, line 129: "because a call to that name never reaches the tool" holds only for `stop_streaming`, which `base_flow.go` intercepts before the `toolsDict` lookup. `transfer_to_agent`, `task_completed`, `adk_request_credential` and `adk_request_confirmation` all resolve through that lookup, so with none of them registered a server tool carrying one of those names would have been reached and now is not.

The drop is still the right call and your next sentence already discloses the over-drop — it is the stated reason that is wrong, for four of the five names.

tool/mcptoolset/meta_test.go
```diff
@@ -0,0 +1,104 @@
...
+		{key: "dev.mcp/key", want: true},
+		{key: "org.modelcontextprotocol.api/key", want: true},
+		{key: "com.mcp.tools/key", want: true},
+		{key: "com.example.mcp/key", want: false},
+		// Only the second label carries the marker, so a marker at any other
+		// position leaves the key to the server.
+		{key: "modelcontextprotocol.io/key", want: false},
+		{key: "mcp.dev/key", want: false},
+		{key: "a.b.mcp.d/key", want: false},
+		{key: "tools.api.mcp.com/key", want: false},
```
karolpiotrowicz, line 44: These four rows are correct and worth keeping — they pin the current second-label rule against the superseded 2025-06-18 positional one, which is exactly the reading I got wrong last round.

One gap: every `want: true` prefix in the table is one of `io.modelcontextprotocol/`, `dev.mcp/`, `org.modelcontextprotocol.api/` or `com.mcp.tools/`. I replaced the whole body of `isReservedMetaKey` with a hardcoded list of those four and the package still passed, even though the substitute wrongly forwards spec-reserved keys like `net.mcp/hint` and `uk.modelcontextprotocol/x`. A single row such as `{key: "net.mcp/hint", want: true}` closes it.

tool/mcptoolset/set_test.go
```diff
@@ -965,3 +966,404 @@ func TestMCPTool_NonTextOnlyResponse(t *testing.T) {
...
+				})
+			}
+			if _, err := server.Connect(t.Context(), serverTransport, nil); err != nil {
+				t.Fatal(err)
+			}
+
+			var offered []string
+			ts, err := mcptoolset.New(mcptoolset.Config{
+				Transport: clientTransport,
+				ToolFilter: func(ctx agent.ReadonlyContext, t tool.Tool) bool {
```
karolpiotrowicz, line 1260: The closure parameter `t tool.Tool` shadows the subtest's `t *testing.T` from a few lines up. It compiles and does the right thing today because both types have a `Name()` method and you want the tool's — but a `t.Errorf` added inside this closure later would either target the wrong `t` or stop compiling. Renaming the parameter avoids that.

karolpiotrowicz:
@wolo-lab — flagging this one for a direction call before the author puts in more work.

The elicitation half of this PR is parity with adk-python. The other half adds a reserved-key filter for `_meta` that adk-python does not have, where the whole dumped result is returned and `_meta` is treated as opaque. The PR also adds four exported symbols to `tool/mcptoolset`. That combination is a scope question rather than a code question, so your view on whether we want this shape seems more useful than another round on the implementation.

One correction for the record. An earlier review here objected that the reserved-prefix rule missed the specification's own examples. That objection rested on the superseded 2025-06-18 revision of the MCP specification. The current revision reserves a prefix when its second label is `modelcontextprotocol` or `mcp`, and rules that `com.example.mcp/` is not reserved because its second label is `example` — which is exactly what this implementation does. The author was right, and I have dismissed both of my change-requests.

Separately, the branch no longer merges. #1401 landed 42 minutes after this was pushed and the two touch the same file. Neither mechanical resolution compiles, so reconciling them is a design call as well.

QuentinBisson:
@karolpiotrowicz thanks for the correction on the `_meta` prefix rule, and for merging both sides locally rather than just flagging the conflict marker.

Rather than wait on one direction call that covers two unrelated changes, would you prefer I split this into two PRs?

**1. Elicitation only** — `Config.ElicitationHandler`, `Config.ElicitationCompleteHandler`, the capability declaration, the `New` guards and their tests. This touches neither `tool.go`'s content rendering nor anything #1401 changed, so it merges independently.

On the scope question, this half is not a divergence from the siblings — it follows one. The same change shipped in adk-python as [`48246195`](https://github.com/google/adk-python/commit/48246195ac) in **v2.7.0** (imported from google/adk-python#6423), which is the `elicitation_callback` at [`mcp_toolset.py:172`](https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_toolset.py#L172) you linked as parity. The API shape has been in a released adk-python for six weeks.

**2. `_meta` passthrough and content rendering** — `serverMeta`, `isReservedMetaKey`, `ToolError` / `UnsupportedContentError`, and the reserved-tool-name drop. Both open questions live entirely here: whether a non-text-only result fails loudly or renders a placeholder after #1401, and whether adk-go should filter `_meta` when adk-python treats it as opaque. That second one is a real divergence and deserves the sign-off you asked for — it just shouldn't gate the first half.

Your own note that the `_meta` threading is "orthogonal and survives" either resolution is what makes the split clean rather than a rework.

PR #1169, opened 2026-07-17
fix(llminternal): resume confirmed tool calls in request order

DABH:
## Problem

`RequestConfirmationRequestProcessor` re-dispatches a batch of confirmed tool calls by ranging `toolsToResumeByFunctionCallID` (a map). Go randomizes map iteration order, so when a user resumes two or more tool confirmations in one message, the re-queued function calls are built in a different order run-to-run for identical input.

That makes three things nondeterministic:

- the order the confirmed function calls are re-dispatched,
- the order of the assembled `FunctionResponse` parts, and
- `StateDelta` last-writer-wins precedence: `mergeEventActions` is applied in slice order, so for confirmed tools that write overlapping state keys the winning writer was effectively random.

This breaks integrations that need deterministic re-dispatch and response ordering when replaying agent pipelines with ADK — e.g. durable-execution engines such as Temporal, where a workflow can install a deterministic task runner but cannot recover a re-dispatch order that was already randomized *upstream* of it.

`generateRequestConfirmationEvent` already fixed the identical hazard on the emit side (iterating the ordered `Content.Parts` instead of ranging `RequestedToolConfirmations`, mirroring adk-python), but the resume path was missed.

## Fix

Record insertion order (`resumeOrder`) while collecting confirmed calls from the ordered `calls` slice, then use that order to re-dispatch. This makes the re-dispatch order, the assembled response order, and the resulting `StateDelta` merge precedence deterministic.

Scope note: this fixes *ordering*, not concurrency. Tools still execute via `platform.RunTasks`, so this PR does not by itself claim fully deterministic replays — it makes the re-dispatch and response order (and the state-merge precedence that follows from it) deterministic.

## Testing plan

- `TestRequestConfirmationResumeOrderIsStable`: six confirmations resumed in one message, request order deliberately unsorted and the user responses assembled in the reversed order, asserted stable across 10 runs. The six-tools / ten-iteration constants are load-bearing against Go's small-map rotation behavior (documented in a comment).
- `TestRequestConfirmationResumeSkipsAlreadyAnsweredCalls`: covers the already-answered skip branch — a later event answers a strict subset of the confirmed IDs; the remaining calls still come out in request order.


karolpiotrowicz (commented):
The ordering fix looks correct to me. `resumeOrder` is scoped per-event alongside the map, it can't accumulate duplicates (the only other mutation of `toolsToResumeByFunctionCallID` is the `delete` pass, which runs after collection has finished), and every surviving entry is emitted exactly once — so the new `cap` hint on `parts` is exact. Asserting on the returned `FunctionResponse` IDs is a valid proxy for the `FunctionCall` order being built, since `handleFunctionCalls` writes results positionally into `fnResponseEvents[i]` and the merge preserves slice order.

Two things I'd like to see before this merges.

**1. The new skip branch isn't covered.**

```go
cc, ok := toolsToResumeByFunctionCallID[callID]
if !ok {
    // Don't resume calls that have already answered (see delete calls above).
    continue
}
```

A coverage run over `./internal/llminternal/` shows this `continue` is never executed by any test in the package. The deletion pass at lines 147-157 doesn't run in any existing case either — `TestParallelFunctionCallsWithHITL` and `TestParallelFunctionCallsWithPartialHITL` both leave `confirmationEventIndex == len(events)-1`, so that loop body is skipped entirely. Since the branch is new here, could you pin it? An event after the user confirmation that answers a strict subset of the confirmed IDs, asserting the remainder still come out in request order, should do it.

**2. The new test can't distinguish request order from response order.**

`confirmationParts` and `responseParts` are both built by ranging `wantOrder`, so an implementation that ordered by the *user's response* order would pass identically. Building `responseParts` from a reversed copy would close that gap, and it's safe today since `confirmationResponses` is a map and already discards response order.

Two notes on the description rather than the code:

- "deterministic replays" claims more than this buys. Tools still execute concurrently via `platform.RunTasks`, and `session.NewEvent` allocates its UUID and timestamp inside each task, so side-effect ordering and event identity stay nondeterministic. "Deterministic re-dispatch and response order" is the accurate framing.
- Conversely, the PR undersells one effect: `mergeEventActions` is applied in slice order, so `StateDelta` last-writer-wins precedence also becomes deterministic. For anyone whose confirmed tools write overlapping state keys, that's arguably the more consequential half of the fix.

One observation worth capturing in a comment: the choice of six tools is load-bearing. Go's small-map iteration randomises a start offset within the group, so the reachable orders are rotations of insertion order rather than arbitrary permutations. At six entries the identity order comes up 3/8 of the time, which makes the 10-iteration loop a roughly 1-in-18,000 blind spot — but at two tools it would be 7/8 per iteration, and the same 10-iteration test would pass on the unfixed code about 26% of the time. Worth a note so nobody later trims it to two calls or a single iteration.

Minor and optional: `gotEvents[0].LLMResponse.Content.Parts` will panic rather than fail cleanly if the merged event is ever nil (`mergeParallelFunctionResponseEvents` returns `nil, nil` when every entry is nil), and the neighbouring tests are table-driven, which would fit naturally if you add the cases above.

DABH:
Thanks — great review. Addressed all of it in a follow-up commit (8826fbd):

**1. Skip branch.** Added `TestRequestConfirmationResumeSkipsAlreadyAnsweredCalls`: three confirmed calls where a later, agent-authored event already answers one of them, so the delete pass removes it and the `!ok` branch runs; the test asserts the remaining two still come out in request order.

**2. Request vs. response order.** `responseParts` in `TestRequestConfirmationResumeOrderIsStable` is now built from `wantOrder` reversed, so an implementation ordering by the user's response order fails. (Safe because `confirmationResponses` is a map and already discards response order, as you noted.)

**3. "deterministic replays" framing.** Agreed it overclaimed — reworded the description to "deterministic re-dispatch and response order," with an explicit scope note that tools still execute via `platform.RunTasks` (this fixes ordering, not concurrency/event-identity). I kept the durable-execution motivation but scoped to ordering: the reason it matters for engines like Temporal is that the map order is *upstream* of the injectable task runner, so a deterministic runner can't recover an order that was already randomized before re-dispatch.

**4. StateDelta.** Called this out in both the description and a code comment: stable slice order also makes `mergeEventActions`' last-writer-wins precedence deterministic for tools that write overlapping state keys.

**5. Six-tools-is-load-bearing.** Added a comment on the iteration loop capturing the small-map rotation math (identity ~3/8 at six entries → ~1-in-18,000 over 10 runs; ~7/8 at two tools → ~26% false-pass on unfixed code), so nobody trims it to two calls or a single iteration.

**6. nil-safety / table-driven.** Added nil/`Content` guards before dereferencing the merged event's parts (handles the `nil, nil` return from `mergeParallelFunctionResponseEvents`). On table-driven: I kept these as two focused tests rather than folding into the neighbouring table — they exercise materially different paths (full-set ordering vs. subset-skip) and the load-bearing rotation rationale reads more clearly inline than as a table field. Happy to convert if you'd prefer consistency.

karolpiotrowicz:
The six points from the last round are all addressed, and the two new tests hold up.

- **Skip branch.** `TestRequestConfirmationResumeSkipsAlreadyAnsweredCalls` does reach the `!ok` branch, and the `Author: "testAgent"` event is a neat way to drive the delete pass without the event being mistaken for the confirmation response. Disabling the guard makes the test fail, so it's pinning the behaviour rather than just walking the line.
- **Request vs. response order.** Reversing `responseOrder` closes the gap — an implementation that ordered by response arrival now fails.
- **Description.** "Deterministic re-dispatch and response order," with the explicit `platform.RunTasks` scope note, is accurate.
- **StateDelta.** Good call surfacing this in both the description and the code comment. `mergeEventActions` applies in slice order, so for tools writing overlapping keys the precedence really was the more consequential half.
- **Six-tools rationale.** The numbers check out: identity order comes up ≈0.375 at six entries and ≈0.875 at two, giving ≈1-in-18,000 over ten runs versus ≈26% at two calls. Worth noting the comment's non-obvious claim is also correct — the reachable orders are rotations of insertion order, not arbitrary permutations, which is exactly why the fan-out is load-bearing.
- **Two focused tests rather than the table.** Agreed, no need to convert; the rotation rationale reads better inline.

On behaviour: only ordering changes here. The set of dispatched calls, the per-ID execution counts and the yield count are all unchanged from the merge-base, so this isn't a semantic change for existing users — and the previous order genuinely varied run to run, so there was no ordering anyone could have depended on. It also lines up with `functions.go:47-51`, which already applied the same ordered-slice-instead-of-map pattern to the emit side; this finishes that on the resume side.

One thing I'd like before this merges.

**The `seen` guard is the one new invariant, and nothing tests it.** Delete lines 135-138 and the suite stays green — but two confirmation requests in a single event resolving to the same `originalFunctionCall.ID` then land in `resumeOrder` twice, and the tool is dispatched twice within one re-dispatch. A test with a duplicate original ID in one confirmation event, asserting a single dispatch, would pin it. That also indirectly protects something else: `resumeOrder` has to stay declared inside the outer loop, and hoisting it likewise leaves the suite green while adding an extra execution.

Optional, and skip it if you add that test: the comment at 175-177 ("re-dispatching them would double-run the tool") reads as a general anti-double-run guarantee, but it's scoped to responses already visible in the event snapshot. Narrowing the wording would avoid setting up a later reader.

Nothing else blocking. There are a couple of pre-existing issues in this processor that this PR correctly leaves alone — the outer loop at 110 yields per event against a static snapshot, so confirmations spread across two events come back in reverse order, and a repeated original ID across events can execute twice, where adk-python accumulates across all events and dispatches once. I'll file those separately rather than widening this PR.

DABH:
Thanks again — addressed the blocking item and the optional cleanup (rebased into the single commit).

**`seen` guard now pinned.** Added `TestRequestConfirmationResumeDedupesDuplicateOriginalID`: one confirmation event with two distinct confirmation requests (`confirmation_a`, `confirmation_b`) that resolve to the same `originalFunctionCall.ID`, both approved, asserting the tool is dispatched exactly once. Deleting the guard (appending to `resumeOrder` unconditionally) makes it fail with `["call_dup", "call_dup"]`, so it pins the dedup invariant — and, as you noted, keeping `resumeOrder` declared inside the outer loop, since a hoist would show up here too. Also added a comment at the guard explaining why the ID is recorded only once.

**Comment narrowed.** Reworded the skip-branch comment so it no longer reads as a general anti-double-run guarantee — it's now scoped to "calls whose response is already present in this event snapshot ... would run its tool a second time within this resume."

Left the two pre-existing issues you flagged (per-event yield reversing cross-event confirmations; repeated original ID across events) alone for the separate follow-ups.

karolpiotrowicz (approved):

PR #1171, opened 2026-07-17
fix(session/vertexai): quote userId in AIP-160 list filter

petrmarinec:

**Problem**

`listSessions` in `session/vertexai/vertexai_client.go` builds the Vertex AI Agent Engine list filter by interpolating the raw `UserID` directly into a quoted [AIP-160](https://google.aip.dev/160) string literal:

```go
rpcReq.Filter = fmt.Sprintf("userId=\"%s\"", req.UserID)
```

A double-quote character in `UserID` breaks out of the quoted value and lets a caller append arbitrary AIP-160 predicates. For example a `UserID` of `attacker" OR userId!="` produces the filter:

```
userId="attacker" OR userId!=""
```

Since `userId!=""` matches every session with a non-empty user id, this returns sessions belonging to **all** users of the same Agent Engine deployment (cross-user session enumeration). Each returned session exposes its id, user id, last update time, and state.

This is the Go counterpart of an issue previously reported and fixed in adk-python — see google/adk-python#5270 / PR google/adk-python#5273 (merged as `bdece00`), where the same unescaped-`user_id` interpolation was fixed with a filter-literal quoting helper. I noticed the same pattern here while reviewing the sibling port and am proposing the equivalent fix.

**Solution**

Add a small `quoteFilterLiteral` helper that escapes backslashes first, then double-quotes, so the whole value stays inside the AIP-160 string literal regardless of what is passed in, and use it when building the filter:

```go
rpcReq.Filter = "userId=" + quoteFilterLiteral(req.UserID)
```

## Testing Plan

**Unit tests**

Added `session/vertexai/filter_test.go` with table-driven coverage for `quoteFilterLiteral`:

- plain value: `alice` -> `"alice"`
- quote injection: `attacker" OR userId!="` -> `"attacker\" OR userId!=\""`
- backslash-only: `\` -> `"\\"`
- empty string: `` -> `""`

**Manual validation (filter-string transformation)**

- On `main`, a `UserID` of `attacker" OR userId!="` produces the filter `userId="attacker" OR userId!=""` — the injected `OR userId!=""` predicate is live and returns every user's sessions.
- With this change, the same input produces `userId="attacker\" OR userId!=\""` — the metacharacters stay inside the quoted literal and the filter matches only the literal (non-existent) user.

## Additional context

Small, focused fix that keeps the current filter-construction approach but ensures `UserID` remains data instead of altering the filter expression. Mirrors the escaping order used in the merged adk-python fix (backslashes first, then double quotes).


wolo-lab (approved):
LGTM, thanks!

PR #1172, opened 2026-07-18
fix(gcsartifact): prevent artifact version race in concurrent Save

wolo-lab:
## Problem
`gcsService.Save` assigned version numbers non-atomically: it listed the
existing versions, computed `max+1`, and wrote the blob at that version. Two
concurrent `Save` calls for the same artifact read the same version set, pick
the same next version, and the second upload **silently overwrites the first** —
callers get duplicate version numbers and one object is lost. The code carried a
`TODO` acknowledging the race.

Note: adk-python's `GcsArtifactService` has the same unguarded
`max(versions) + 1` logic, so this is Go-specific hardening rather than a port.

## Summary
- Write the new blob with a "does not exist" generation precondition
  (`ifGenerationMatch=0`) via
  `ObjectHandle.If(storage.Conditions{DoesNotExist: true})`. GCS has no
  transaction spanning the list and the write, so a colliding writer now fails
  the precondition with HTTP 412 instead of clobbering the existing object.
- On a 412, `Save` re-reads the versions and retries with a fresh number,
  bounded by `maxSaveAttempts` (returns `ErrVersionConflict` if exhausted). N
  concurrent saves therefore yield exactly versions `1..N` with no lost writes.
- Extracted `writeArtifact` / `isPreconditionFailed` helpers and added an
  `ifNotExist()` method to the internal `gcsObject` seam and its wrapper.
- **Public API:** one additive, backward-compatible change — the new exported
  sentinel `gcsartifact.ErrVersionConflict`, so callers can distinguish a lost
  version race from a real write failure and retry. `apidiff` reports no
  incompatible change. `artifact.Service.Save`'s doc now notes that an
  optimistic implementation may transiently fail to claim a version.
- The in-memory GCS fake was rewritten so it actually models the precondition,
  and it now tracks object existence explicitly — a zero-byte artifact is a
  real, listable object, as it is in GCS. A separate `httptest`-backed test
  drives the production `ObjectHandle` wrapper offline and asserts the
  precondition really reaches the wire.

### DIFF
--- a/artifact/gcsartifact/gcs_client.go
+++ b/artifact/gcsartifact/gcs_client.go
@@ -39,6 +39,9 @@ type gcsObject interface {
 	newReader(ctx context.Context) (io.ReadCloser, error)
 	delete(ctx context.Context) error
 	attrs(ctx context.Context) (*storage.ObjectAttrs, error)
+	// ifNotExist returns a handle whose write succeeds only if the object does
+	// not already exist; writing over an existing object fails with HTTP 412.
+	ifNotExist() gcsObject
 }
 
 // gcsObjectIterator
@@ -95,6 +98,11 @@ type gcsObjectWrapper struct {
 // NewWriter implements the gcsObject interface for gcsObjectWrapper.
 func (w *gcsObjectWrapper) newWriter(ctx context.Context) gcsWriter {
 	return &gcsWriterWrapper{w: w.object.NewWriter(ctx)}
+}
+
+// ifNotExist implements the gcsObject interface for gcsObjectWrapper.
+func (w *gcsObjectWrapper) ifNotExist() gcsObject {
+	return &gcsObjectWrapper{object: w.object.If(storage.Conditions{DoesNotExist: true})}
 }
 
 // NewReader implements the gcsObject interface for gcsObjectWrapper.
--- a/artifact/gcsartifact/gcs_test.go
+++ b/artifact/gcsartifact/gcs_test.go
@@ -17,29 +17,44 @@ package gcsartifact
 import (
 	"bytes"
 	"context"
+	"errors"
+	"fmt"
 	"io"
 	"io/fs"
+	"net/http"
+	"slices"
 	"strings"
 	"sync"
 	"testing"
 	"time"
 
 	"cloud.google.com/go/storage"
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
+	"google.golang.org/genai"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"google.golang.org/adk/v2/artifact"
 	"google.golang.org/adk/v2/internal/artifact/tests"
 )
 
-// newGCSArtifactServiceForTesting creates a gcsService for the specified bucket using a mocked inmemory client
-func newGCSArtifactServiceForTesting(bucketName string) (artifact.Service, error) {
+// newGCSServiceForTesting creates a *gcsService backed by the in-memory fake. It
+// stubs out the retry backoff so retry-heavy tests stay fast.
+func newGCSServiceForTesting(bucketName string) *gcsService {
 	client := newFakeClient()
-	s := &gcsService{
+	return &gcsService{
 		bucketName:    bucketName,
 		storageClient: client,
 		bucket:        client.bucket(bucketName),
+		sleep:         func(ctx context.Context, _ time.Duration) error { return ctx.Err() },
 	}
-	return s, nil
+}
+
+// newGCSArtifactServiceForTesting creates a gcsService for the specified bucket using a mocked inmemory client
+func newGCSArtifactServiceForTesting(bucketName string) (artifact.Service, error) {
+	return newGCSServiceForTesting(bucketName), nil
 }
 
 func TestGCSArtifactService(t *testing.T) {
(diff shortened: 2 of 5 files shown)


artifact/gcsartifact/gcs_client.go
```diff
@@ -97,6 +100,11 @@ func (w *gcsObjectWrapper) newWriter(ctx context.Context) gcsWriter {
 	return &gcsWriterWrapper{w: w.object.NewWriter(ctx)}
 }
 
+// ifNotExist implements the gcsObject interface for gcsObjectWrapper.
+func (w *gcsObjectWrapper) ifNotExist() gcsObject {
+	return &gcsObjectWrapper{object: w.object.If(storage.Conditions{DoesNotExist: true})}
```
karolpiotrowicz, line 105: This is the only line in the repository that applies the real precondition, and no test reaches it — the package's tests all go through the fake's own `ifNotExist`, and `gcsObjectWrapper` is only constructed by `NewService`, which nothing in-repo calls. I checked what that costs: rewriting this to `return &gcsObjectWrapper{object: w.object}` — i.e. deleting the fix from the production path entirely — leaves `go test ./...` fully green. The compile-time assertion at the bottom of the file pins the method's signature, not its body.

Not a blocker, and I realise the whole adapter layer has always been uncovered. But this line is now load-bearing in a way the others are not. `storage.NewClient` speaks HTTP, so an `httptest.Server` plus `option.WithEndpoint(...)` and `option.WithoutAuthentication()` would let one `Save` run through the real wrapper offline and assert the upload carries `x-goog-if-generation-match: 0`. If that is more than you want in this PR, even asserting that `ifNotExist()` returns a *different* `*storage.ObjectHandle` than the receiver would catch the no-op regression. Happy for it to be a follow-up.

artifact/gcsartifact/gcs_test.go
```diff
@@ -70,90 +238,108 @@ func (c *fakeClient) bucket(name string) gcsBucket {
...
-	for name, obj := range f.objectsMap {
+	var attrs []*storage.ObjectAttrs
+	for name, b := range f.blobs {
 		if q != nil && q.Prefix != "" && !strings.HasPrefix(name, q.Prefix) {
 			continue
 		}
-		if !obj.deleted {
-			matchingObjects = append(matchingObjects, obj)
+		b.mu.Lock()
+		if !b.deleted && b.data != nil {
```
karolpiotrowicz: Using `b.data != nil` as the existence test means a zero-byte artifact is modelled as not existing, which makes it invisible here and makes the `mustNotExist` check below never fire for it.

That is reachable through the public API: `SaveRequest.Validate` only requires `Part.InlineData != nil`, so `genai.NewPartFromBytes([]byte{}, "application/octet-stream")` is a valid request, and `bytes.Buffer.Bytes()` on an empty buffer returns nil. Saving that artifact twice in a row on this branch returns version 1 both times — the second overwrites the first — and `Versions()` and `Load()` then report it as not found. On the merge-base the same sequence returns 1 then 2. Real GCS does create and list a zero-byte object, so production is unaffected; it is the fake that diverges, which means this input class has no test coverage.

I can see why the guard is there — `object(name)` creates the blob eagerly, so without it every losing attempt would leak a phantom version into the listing. An explicit `exists bool` on `fakeBlob`, set in `Close`, would give you both: no phantoms, and zero-byte artifacts that behave like real ones. The same substitution applies at `attrs` and `newReader`, and at the `mustNotExist` check in `fakeWriter.Close`.

karolpiotrowicz (commented):
The race is real and this closes it. I reproduced the original bug against the merge-base and it fails hard: 10 concurrent `Save` calls for one artifact returned duplicate version numbers and silently dropped writes in 18 of 20 runs (e.g. returned `[1 2 2 3 3 4 4 4 5 6]`, stored `[1 2 3 4 5 6]`). The same reproduction passes 50/50 on this branch, and 30/30 under `-race` with no data races. I also wrote a deterministic control that commits a competing blob at the same object name just before the conditional write commits: `Save` correctly retried, returned version 2, and left the competitor's bytes at version 1 intact. Build, `go vet`, `gofmt`, `staticcheck`, `golangci-lint` and `go mod tidy -diff` are all clean, and `apidiff` reports one compatible addition and no incompatible change.

Worth calling out as a bonus: the fake rewrite fixes a fidelity bug of its own. On the merge-base, merely taking an object handle registered a listable entry, so a failed `Load` of a missing version poisoned the version listing (`Versions()` returned `[1 7 8]` after one save plus one failed load; it now correctly returns `[1 2]`).

Two small things, then a couple of questions. Neither blocks.

**Test coverage of the actual precondition.** See the inline notes on `artifact/gcsartifact/gcs_client.go` and `artifact/gcsartifact/gcs_test.go`.

**PR description.** The description says "No public API change", but `ErrVersionConflict` is a new exported symbol in a package whose only prior export was `NewService`. It is purely additive and backward compatible (`apidiff` agrees), so this is just a wording fix — but since the module's exported surface is permanent, it is worth being explicit about in the description so the addition gets a deliberate look.

**Questions, not change requests:**

- Is a fixed *attempt* budget the shape you want, rather than a deadline? A count bounds a closed batch of writers, but under a steady arrival stream a writer that backs off can be lapped by newcomers that start with no delay. Modelling GCS list latency (a bucket wrapper that makes the listing stale by a fixed delay) with the real jittered backoff, I saw 0 conflicts at 50 and 60 concurrent savers and 1/100 at 100, but 52/200 with 200 savers on a single artifact name and a 40 ms stale listing. That is well beyond anything the in-tree callers produce — they all build per-invocation or per-millisecond file names — and failing closed is strictly better than the old silent overwrite, so this may well be fine as is. If it is, would it be worth noting the failure mode on `artifact.Service.Save` as well? Callers program against the interface, and none of the current ones inspect or retry the error.

- The Python implementation still uses the unguarded `max(versions) + 1`. Since a precondition-based scheme only holds when every writer uses it, a bucket written by both implementations would still lose a version. Might be worth a line in the doc comment, and an issue on the Python side.

baptmont (approved):

artifact/gcsartifact/gcs_client.go
wolo-lab: Confirmed and fixed in this PR rather than a follow-up — the test turned out to
be cheap enough that deferring it wasn't worth it.

artifact/gcsartifact/gcs_test.go
wolo-lab: Good catch, reproduced exactly as described:

    versions returned: 1, 1
    Versions() = artifact not found: file does not exist
    Load()     = artifact not found

`SaveRequest.Validate` only requires `Part.InlineData != nil`, and
`bytes.Buffer.Bytes()` on an empty buffer returns nil, so this is reachable from
the public API and the fake diverged from GCS, which does create and list a
zero-byte object.

Fixed, with one deviation from the suggestion: rather than adding `exists`
alongside `deleted`, I replaced `deleted` with it. Two flags for one property
invites them to disagree, and `delete` setting `exists = false` covers the same
ground — `objects`, `attrs`, `newReader` and the `mustNotExist` check in
`fakeWriter.Close` all collapse to a single condition.

Added `TestSaveZeroByteArtifact` as a regression test through the public API
(two saves -> versions 1 and 2, both listable, `Load` returns empty data). It
fails with `Save() version = 1, want 2` if the existence test is reverted to
`data != nil`.

wolo-lab:
> The race is real and this closes it. I reproduced the original bug against the merge-base and it fails hard: 10 concurrent `Save` calls for one artifact returned duplicate version numbers and silently dropped writes in 18 of 20 runs (e.g. returned `[1 2 2 3 3 4 4 4 5 6]`, stored `[1 2 3 4 5 6]`). The same reproduction passes 50/50 on this branch, and 30/30 under `-race` with no data races. I also wrote a deterministic control that commits a competing blob at the same object name just before the conditional write commits: `Save` correctly retried, returned version 2, and left the competitor's bytes at version 1 intact. Build, `go vet`, `gofmt`, `staticcheck`, `golangci-lint` and `go mod tidy -diff` are all clean, and `apidiff` reports one compatible addition and no incompatible change.
> 
> Worth calling out as a bonus: the fake rewrite fixes a fidelity bug of its own. On the merge-base, merely taking an object handle registered a listable entry, so a failed `Load` of a missing version poisoned the version listing (`Versions()` returned `[1 7 8]` after one save plus one failed load; it now correctly returns `[1 2]`).
> 
> Two small things, then a couple of questions. Neither blocks.
> 
> **Test coverage of the actual precondition.** See the inline notes on `artifact/gcsartifact/gcs_client.go` and `artifact/gcsartifact/gcs_test.go`.
> 
> **PR description.** The description says "No public API change", but `ErrVersionConflict` is a new exported symbol in a package whose only prior export was `NewService`. It is purely additive and backward compatible (`apidiff` agrees), so this is just a wording fix — but since the module's exported surface is permanent, it is worth being explicit about in the description so the addition gets a deliberate look.
> 
> **Questions, not change requests:**
> 
> * Is a fixed _attempt_ budget the shape you want, rather than a deadline? A count bounds a closed batch of writers, but under a steady arrival stream a writer that backs off can be lapped by newcomers that start with no delay. Modelling GCS list latency (a bucket wrapper that makes the listing stale by a fixed delay) with the real jittered backoff, I saw 0 conflicts at 50 and 60 concurrent savers and 1/100 at 100, but 52/200 with 200 savers on a single artifact name and a 40 ms stale listing. That is well beyond anything the in-tree callers produce — they all build per-invocation or per-millisecond file names — and failing closed is strictly better than the old silent overwrite, so this may well be fine as is. If it is, would it be worth noting the failure mode on `artifact.Service.Save` as well? Callers program against the interface, and none of the current ones inspect or retry the error.
> * The Python implementation still uses the unguarded `max(versions) + 1`. Since a precondition-based scheme only holds when every writer uses it, a bucket written by both implementations would still lose a version. Might be worth a line in the doc comment, and an issue on the Python side.

Thanks — the reproduction numbers and the mutation check made this straight
forward to act on. All four points addressed:

**Test coverage of the precondition** and **the fake's zero-byte handling** —
both fixed in this PR; details in the inline threads.

**PR description** — you're right, and it was wrong twice over: it also
described the precondition as an `x-goog-if-generation-match` header, which is
not what goes on the wire. Rewritten to call out `ErrVersionConflict` explicitly
as the one additive public API change.

**Attempt budget vs deadline.** I'd like to keep the count, because the context
already supplies the deadline: `sleepContext` selects on `ctx.Done()`, so a
caller with a timeout bounds the whole loop regardless of the budget. The
uncapped worst case is the sum of the jittered caps, ~9.3s of backoff across 15
sleeps (expected ~4.6s with full jitter), which a caller can cut short at any
point. Your starvation scenario is real, but it needs sustained arrivals on one
artifact name, and failing closed beats the previous silent overwrite. What I
did change: the context error from the backoff was previously returned bare, so
a caller got a naked `context.DeadlineExceeded` with no indication of which save
failed — it's now wrapped with the filename.

On documenting it on `artifact.Service.Save`: agreed that callers program
against the interface, so I added a sentence there — but deliberately a generic
one ("implementations that assign versions optimistically may fail to claim one
while other writers save the same artifact; that failure is transient and the
call can be retried") rather than naming `ErrVersionConflict`. A GCS-specific
sentinel in the interface doc would oblige every other implementation to honour
it, and a caller that wants to branch on it can import the package. Happy to
promote it to `artifact` if you'd rather callers be able to detect it without
depending on the GCS implementation.

**Interoperability.** The `Save` doc now records the constraint in general
terms — the scheme only holds if every writer uses the precondition, so a bucket
also written by a client that assigns versions unconditionally can still lose
one. I kept it implementation-agnostic on purpose: a claim about another
codebase's current behaviour is not something this repo can keep in sync. I'll
raise the concrete case separately with the Python side.

Also noted in the doc that each attempt costs one listing plus one upload — I
kept the re-list rather than incrementing locally on 412, since `max+1` from a
fresh listing jumps past every committed version instead of probing linearly.

PR #1173, opened 2026-07-18
feat(auth/gcp): add GCP credential provider

wolo-lab:
## Problem

`auth/gcp` has the credentials-service REST client but no `auth.CredentialProvider` that resolves the acting user's credential from the invocation context, so there is no way to wire per-user GCP credentials (Agent Identity / IAM Connector) into a tool call. adk-python has this via its GCP auth provider.

## Summary

`gcp.NewProvider(ctx, gcp.ProviderConfig)` returns an `auth.CredentialProvider` that resolves a per-user credential and maps it to an `auth.Credential`. The acting user is read from the ADK context at resolve time, so the provider is long-lived and shared, and never needs per-user configuration. `ctx` roots the default client built lazily when `ProviderConfig.Client` is nil — values only, cancellation is not honored, because that client outlives any one request.

The lazy default client bounds its waiters, not its attempt. `google.FindDefaultCredentials` reads the credentials file with `os.ReadFile` and probes with the context-free `metadata.OnGCE()`, so no context can interrupt it. Each caller waits on the earlier of its own context and 30s. A failed attempt is not cached and the next call retries. A hung one is kept rather than abandoned — abandoning it starts a fresh lookup every 30s, each parked in a syscall pinning an OS thread — and if it eventually lands, its client is published to everyone after it.

`NewProvider` validates at wiring time what would otherwise fail on every request from inside a transport: the resource name must be one of the two shapes `Client` routes on, and a `Client` that did not come from `NewClient` is rejected. The shape check is stricter than `Client` and than adk-python, both of which send any non-connector name to Agent Identity.

New sentinels `gcp.ErrNoActingUser` and `gcp.ErrClientUnavailable` make both new failure modes matchable. Existing sentinels, `*APIError` and `*auth.ConsentRequiredError` all stay reachable with `errors.Is`/`errors.As` through the provider's wrap.

A provider error names the resource and nothing else. It becomes the tool's error, which is fed back to the model and persisted as a session event, and every other identifier available at that point is caller-supplied. This is a deliberate divergence from adk-python, which names the user.

### Requirements this puts on the transport

Neither can be enforced from this package, and both are stated on `NewProvider`.

- Every authenticated request must descend from the invoking user's context. A transport that shares one connection across invocations does not qualify, `mcptoolset` included: its per-call POSTs are per-user, but the MCP session and the standalone server-to-client stream are opened on the context of whichever user connected first and stay bound to it.
[...]

PR #1174, opened 2026-07-18
feat(auth): add CredentialStore and cache GCP credentials

wolo-lab:
## Problem

The GCP provider (#1173) does a credential-service round trip — plus up to a ~10s pending poll — on every outbound request. There is no caching. adk-python has an `InMemoryCredentialService`, but it is not on this provider's path at all, so caching GCP credentials is new work rather than a port.

A credential cache is also exactly where cross-user leaks live, so most of the design here is about what the key must cover.

## Summary

- Adds `auth.CredentialStore` and `InMemoryCredentialStore`: concurrency-safe, per-entry expiry, a small clock-skew margin exported as `auth.ExpirySkew`, and a paced sweep so a principal that resolves once and never returns does not keep its credential for the life of the process.
- The GCP provider caches through the store (a private in-memory one by default, or `ProviderConfig.Store`). `Client.RetrieveCredential` returns a `*Retrieval` carrying the service's `expireTime`, so entries expire when the credential does.
- **The cache key covers everything that decides which credential comes back**: the app, the end user, the resource, the scopes, the continue URI, and the `*Client` — which fixes both the service asked and the identity the ask is authenticated as. Components are length-prefixed before hashing, so no delimiter inside a scope or URI can collide two schemes. Entries are shared exactly among providers that share a `Client`, which is what makes one store safe behind several providers.
- **Nothing is cached without a usable lifetime.** An absent, unparseable, already-past, or too-short expiry is declined, and the far end is clamped to an hour so a wrong or injected `expireTime` cannot pin an entry. That hour is also the bound on how long a credential revoked before its expiry keeps being served; `Client.CacheKey` names the entry to `Delete` sooner.
- Caching never fails auth: a store read error, a write error, or a malformed hit all fall through to a fresh retrieval.

Two things worth flagging for review rather than reading out of the diff:

- `Client.RetrieveCredential`'s signature changes, which `apidiff` reports as the only incompatible change in either package. `auth/gcp` is absent from v2.0.0, v2.1.0 and v2.2.0, so this reaches only someone tracking `main`.
- Concurrent cold misses on one key are not coalesced — N misses make N round trips. This is never worse than the per-request round trip it replaces, and `singleflight` would need a context detached from any one caller to avoid one cancellation failing everybody who joined. Left as a follow-up rather than folded into this change.

### DIFF
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -182,10 +182,18 @@ type Request struct {
 	ContinueURI string
 }
 
+// Retrieval is the result of [Client.RetrieveCredential].
+type Retrieval struct {
+	Credential auth.Credential
+	// ExpiresAt is the credential's expiry, or the zero time when the service
+	// reports no lifetime.
+	ExpiresAt time.Time
+}
+
 // RetrieveCredential retrieves a credential for req, polling while the service
 // reports a non-interactive pending state (up to the configured poll timeout).
 // If interactive consent is required it returns an [auth.ConsentRequiredError].
-func (c *Client) RetrieveCredential(ctx context.Context, req Request) (auth.Credential, error) {
+func (c *Client) RetrieveCredential(ctx context.Context, req Request) (*Retrieval, error) {
 	if req.Resource == "" {
 		return nil, errors.New("gcp: RetrieveCredential requires a Resource")
 	}
@@ -210,7 +218,11 @@ func (c *Client) RetrieveCredential(ctx context.Context, req Request) (auth.Cred
 		}
 		switch o := res.(type) {
 		case credOutcome:
-			return mapCredential(o.header, o.token)
+			cred, err := mapCredential(o.header, o.token)
+			if err != nil {
+				return nil, err
+			}
+			return &Retrieval{Credential: cred, ExpiresAt: o.expiresAt}, nil
 		case consentOutcome:
 			return nil, &auth.ConsentRequiredError{AuthURI: o.authURI, Nonce: o.nonce}
 		case rejectedOutcome:
@@ -238,8 +250,12 @@ func (c *Client) RetrieveCredential(ctx context.Context, req Request) (auth.Cred
 type outcome interface{ isOutcome() }
 
 type (
-	// credOutcome carries a successfully retrieved {header, token} credential.
-	credOutcome struct{ header, token string }
+	// credOutcome carries a successfully retrieved {header, token} credential and
+	// its expiry (zero when the service does not report one).
+	credOutcome struct {
+		header, token string
+		expiresAt     time.Time
+	}
 	// pendingOutcome means retrieval is still pending; poll again.
 	pendingOutcome struct{}
 	// consentOutcome means interactive consent is required at authURI.
@@ -256,11 +272,28 @@ func (pendingOutcome) isOutcome()  {}
 func (consentOutcome) isOutcome()  {}
 func (rejectedOutcome) isOutcome() {}
 
-// credentialPayload is the {header, token} success shape shared by both services
-// (under "success" for Agent Identity, "response" for the IAM Connector operation).
+// credentialPayload is the success shape shared by both services (under
+// "success" for Agent Identity, "response" for the IAM Connector operation): the
+// {header, token} pair plus the token's expiry. An empty expireTime means the
+// service can't say when the token expires (possibly permanent), so callers must
+// not treat it as "expires now".
 type credentialPayload struct {
-	Token  string `json:"token"`
-	Header string `json:"header"`
+	Token      string `json:"token"`
+	Header     string `json:"header"`
+	ExpireTime string `json:"expireTime"` // RFC 3339; empty when the service reports no expiry
+}
+
+// parseExpireTime parses a proto Timestamp (RFC 3339) into a time.Time, or
+// returns the zero time when empty or malformed.
+func parseExpireTime(s string) time.Time {
+	if s == "" {
+		return time.Time{}
+	}
+	t, err := time.Parse(time.RFC3339, s)
+	if err != nil {
+		return time.Time{}
+	}
+	return t
 }
 
 // retrieveRequest is the JSON body for both services' credentials:retrieve RPC
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -17,6 +17,8 @@ package gcp
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -170,20 +172,40 @@ type Client struct {
 	connectorURL     string
 	pollTimeout      time.Duration
 	initialBackoff   time.Duration
-	// cacheSlot separates clients that can mint different credentials, so the
-	// provider's cache key can include it. What the services return depends on
-	// the endpoint asked and on the identity the request is authenticated as, and
-	// a caller can supply both — so two providers sharing one store must not
-	// share an entry unless their clients agree on both.
+	// cacheSlot names this Client for the credential cache. What the services
+	// return depends on the endpoint asked and on the identity the ask is
+	// authenticated as, so two providers sharing one store must not share an
+	// entry unless their Clients agree on both.
 	//
-	// A caller-supplied HTTPClient is opaque, so its slot is unique per client.
-	// Clients left to Application Default Credentials all authenticate as the one
-	// process principal, so they share a slot and keep sharing cache entries.
+	// It is one value per Client, not one per identity, because this package
+	// cannot observe the identity a Client authenticates as. A caller-supplied
+	// HTTPClient is opaque by construction. So is Application Default
+	// Credentials: NewClient resolves it afresh on every call, so two ADC clients
+	// in one process need not be the same principal — the credentials file can be
+	// rewritten between them, and oauth2.HTTPClient in the context supplies the
+	// transport underneath the ADC one, which this package's own PollTimeout doc
+	// invites a caller to use. Erring towards a false miss costs a round trip;
+	// erring the other way discloses one principal's token to another.
+	//
+	// The value is unique per process as well as within one, so a store that
+	// outlives the process — or is shared by two — misses rather than serving an
+	// entry written by a Client that no longer exists and cannot be identified.
 	cacheSlot string
 }
 
-// callerSeq numbers clients whose identity this package cannot inspect.
-var callerSeq atomic.Uint64
+// clientSeq numbers Clients within this process, and clientNonce separates one
+// process's numbering from another's. See the cacheSlot field.
+var (
+	clientSeq   atomic.Uint64
+	clientNonce = newClientNonce()
+)
+
+func newClientNonce() string {
+	var b [16]byte
+	// Documented never to fail, and it panics rather than returning short.
+	_, _ = rand.Read(b[:])
+	return hex.EncodeToString(b[:])
+}
 
 // Config configures a [Client]. A nil *Config, or any zero-valued field, uses
 // the corresponding default.
@@ -192,6 +214,10 @@ type Config struct {
 	// from Application Default Credentials (cloud-platform scope). If set, it is
 	// used verbatim and ADC is not applied, so it must carry its own credentials
 	// and should refuse redirects for the reason [NewClient] describes.
+	//
+	// Two Clients are two cache dimensions even when built from one HTTPClient,
+	// so build a Client once and share it. One per request resolves nothing from
+	// cache and leaves an entry behind per request.
 	HTTPClient *http.Client
 	// AgentIdentityEndpoint overrides the Agent Identity base URL (scheme+host).
 	// It is used as given, not parsed: an http:// value would send the ADC token
@@ -259,11 +285,20 @@ func NewClient(ctx context.Context, cfg *Config) (*Client, error) {
 		}
 		c.httpClient = hc
 	}
-	c.cacheSlot = joinFields("adc", c.agentIdentityURL, c.connectorURL)
-	if cfg.HTTPClient != nil {
-		c.cacheSlot = joinFields("caller", strconv.FormatUint(callerSeq.Add(1), 10))
-	}
+	c.cacheSlot = joinFields(clientNonce, strconv.FormatUint(clientSeq.Add(1), 10))
 	return c, nil
+}
+
+// CacheKey returns the key a credential for scheme, retrieved through c on
+// behalf of userID under appName, is cached under — the key to hand
+// [auth.CredentialStore.Delete] to invalidate that credential ahead of its
+// expiry, on a consent revocation or a logout.
+//
+// It is meaningful only to the process that produced c, and only for as long as
+// c is alive: a Client is one cache dimension, so rebuilding it strands whatever
+// the old one cached.
+func (c *Client) CacheKey(scheme ProviderScheme, appName, userID string) auth.CredentialKey {
+	return auth.CredentialKey{AppName: appName, UserID: userID, Key: cacheSlot(c, scheme)}
 }
 
 // joinFields encodes fields as one unambiguous string, each prefixed with its
@@ -314,9 +349,12 @@ type Request struct {
 
 // Retrieval is the result of [Client.RetrieveCredential].
 type Retrieval struct {
+	// Credential authenticates outbound requests as the end user.
 	Credential auth.Credential
 	// ExpiresAt is the credential's expiry, or the zero time when the service
-	// reports no lifetime.
+	// reports no lifetime — which it does when the token may be permanent, or
+	// when it cannot say. Either way the lifetime is unknown, so a caller must
+	// not treat the zero value as "expires now" and must not cache it.
 	ExpiresAt time.Time
 }
 
@@ -455,9 +493,14 @@ func (rejectedOutcome) isOutcome() {}
 // service can't say when the token expires (possibly permanent), so callers must
 // not treat it as "expires now".
 type credentialPayload struct {
-	Token      string `json:"token"`
-	Header     string `json:"header"`
-	ExpireTime string `json:"expireTime"` // RFC 3339; empty when the service reports no expiry
+	Token  string `json:"token"`
+	Header string `json:"header"`
+	// ExpireTime is RFC 3339, and empty when the service reports no expiry. The
+	// name is the one in the published API surface
+	// (https://agentidentitycredentials.googleapis.com/$discovery/rest?version=v1,
+	// Success.expireTime), which also warns that the token may be revoked, or
+	// expire slightly early through clock skew, before the time it names.
+	ExpireTime string `json:"expireTime"`
 }
 
 // parseExpireTime parses a proto Timestamp (RFC 3339) into a time.Time, or
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -503,8 +503,11 @@ type credentialPayload struct {
 	ExpireTime string `json:"expireTime"`
 }
 
-// parseExpireTime parses a proto Timestamp (RFC 3339) into a time.Time, or
-// returns the zero time when empty or malformed.
+// parseExpireTime parses the service's expiry into a time.Time, collapsing an
+// absent one and an unparseable one to the same zero time. Callers read that as
+// "lifetime unknown" and decline to cache, which is the safe reading of both:
+// the service omits the field when the token may be permanent or when it cannot
+// say, and a value it sent but we cannot read tells us no more than silence.
 func parseExpireTime(s string) time.Time {
 	if s == "" {
 		return time.Time{}
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -23,7 +23,9 @@ import (
 	"io"
 	"net/http"
 	"regexp"
+	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"golang.org/x/oauth2"
@@ -168,7 +170,20 @@ type Client struct {
 	connectorURL     string
 	pollTimeout      time.Duration
 	initialBackoff   time.Duration
+	// cacheSlot separates clients that can mint different credentials, so the
+	// provider's cache key can include it. What the services return depends on
+	// the endpoint asked and on the identity the request is authenticated as, and
+	// a caller can supply both — so two providers sharing one store must not
+	// share an entry unless their clients agree on both.
+	//
+	// A caller-supplied HTTPClient is opaque, so its slot is unique per client.
+	// Clients left to Application Default Credentials all authenticate as the one
+	// process principal, so they share a slot and keep sharing cache entries.
+	cacheSlot string
 }
+
+// callerSeq numbers clients whose identity this package cannot inspect.
+var callerSeq atomic.Uint64
 
 // Config configures a [Client]. A nil *Config, or any zero-valued field, uses
 // the corresponding default.
@@ -244,7 +259,24 @@ func NewClient(ctx context.Context, cfg *Config) (*Client, error) {
 		}
 		c.httpClient = hc
 	}
+	c.cacheSlot = joinFields("adc", c.agentIdentityURL, c.connectorURL)
+	if cfg.HTTPClient != nil {
+		c.cacheSlot = joinFields("caller", strconv.FormatUint(callerSeq.Add(1), 10))
+	}
 	return c, nil
+}
+
+// joinFields encodes fields as one unambiguous string, each prefixed with its
+// byte length. No delimiter occurring inside a field can then make two different
+// field lists encode alike, which joining on a separator alone allowed.
+func joinFields(fields ...string) string {
+	var b strings.Builder
+	for _, f := range fields {
+		b.WriteString(strconv.Itoa(len(f)))
+		b.WriteByte(':')
+		b.WriteString(f)
+	}
+	return b.String()
 }
 
 // Request identifies the resource and acting user for a credential retrieval.
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -280,6 +280,14 @@ type Request struct {
 	ContinueURI string
 }
 
+// Retrieval is the result of [Client.RetrieveCredential].
+type Retrieval struct {
+	Credential auth.Credential
+	// ExpiresAt is the credential's expiry, or the zero time when the service
+	// reports no lifetime.
+	ExpiresAt time.Time
+}
+
 // RetrieveCredential retrieves a credential for req, polling while the service
 // reports a non-interactive pending state (up to the configured poll timeout).
 // If interactive consent is required it returns an [auth.ConsentRequiredError].
@@ -296,7 +304,7 @@ type Request struct {
 // was true in v2.3.0. A decode failure is now [ErrMalformedResponse] and no
 // longer carries the [encoding/json] error behind it, so [errors.As] against
 // *json.SyntaxError stops finding one.
-func (c *Client) RetrieveCredential(ctx context.Context, req Request) (_ auth.Credential, err error) {
+func (c *Client) RetrieveCredential(ctx context.Context, req Request) (_ *Retrieval, err error) {
 	if req.Resource == "" {
 		return nil, errors.New("gcp: RetrieveCredential requires a Resource")
 	}
@@ -351,7 +359,11 @@ func (c *Client) RetrieveCredential(ctx context.Context, req Request) (_ auth.Cr
 		}
 		switch o := res.(type) {
 		case credOutcome:
-			return mapCredential(o.header, o.token, req.UserID, req.ContinueURI)
+			cred, err := mapCredential(o.header, o.token, req.UserID, req.ContinueURI)
+			if err != nil {
+				return nil, err
+			}
+			return &Retrieval{Credential: cred, ExpiresAt: o.expiresAt}, nil
 		case consentOutcome:
 			return nil, &auth.ConsentRequiredError{AuthURI: o.authURI, Nonce: o.nonce}
 		case rejectedOutcome:
@@ -383,8 +395,12 @@ func (c *Client) RetrieveCredential(ctx context.Context, req Request) (_ auth.Cr
 type outcome interface{ isOutcome() }
 
 type (
-	// credOutcome carries a successfully retrieved {header, token} credential.
-	credOutcome struct{ header, token string }
+	// credOutcome carries a successfully retrieved {header, token} credential and
+	// its expiry (zero when the service does not report one).
+	credOutcome struct {
+		header, token string
+		expiresAt     time.Time
+	}
 	// pendingOutcome means retrieval is still pending; poll again.
 	pendingOutcome struct{}
 	// consentOutcome means interactive consent is required at authURI.
@@ -401,11 +417,28 @@ func (pendingOutcome) isOutcome()  {}
 func (consentOutcome) isOutcome()  {}
 func (rejectedOutcome) isOutcome() {}
 
-// credentialPayload is the {header, token} success shape shared by both services
-// (under "success" for Agent Identity, "response" for the IAM Connector operation).
+// credentialPayload is the success shape shared by both services (under
+// "success" for Agent Identity, "response" for the IAM Connector operation): the
+// {header, token} pair plus the token's expiry. An empty expireTime means the
+// service can't say when the token expires (possibly permanent), so callers must
+// not treat it as "expires now".
 type credentialPayload struct {
-	Token  string `json:"token"`
-	Header string `json:"header"`
+	Token      string `json:"token"`
+	Header     string `json:"header"`
+	ExpireTime string `json:"expireTime"` // RFC 3339; empty when the service reports no expiry
+}
+
+// parseExpireTime parses a proto Timestamp (RFC 3339) into a time.Time, or
+// returns the zero time when empty or malformed.
+func parseExpireTime(s string) time.Time {
+	if s == "" {
+		return time.Time{}
+	}
+	t, err := time.Parse(time.RFC3339, s)
+	if err != nil {
+		return time.Time{}
+	}
+	return t
 }
 
 // retrieveRequest is the JSON body for both services' credentials:retrieve RPC
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -521,9 +521,23 @@ type credentialPayload struct {
 	// Success.expireTime), which also warns that the token may be revoked, or
 	// expire slightly early through clock skew, before the time it names. The IAM
 	// Connector service publishes no discovery document to anonymous callers, so
-	// its field is taken to be the same name; if it is not, a connector
-	// credential simply reports no lifetime and is never cached.
-	ExpireTime string `json:"expireTime"`
+	// its field is assumed to be the same name and shape; if it is not, a
+	// connector credential reports no lifetime and is never cached.
+	ExpireTime lenientTime `json:"expireTime"`
+}
+
+// lenientTime is a JSON string that declines to fail. Only the cache reads this
+// field, and the IAM Connector's shape for it is an assumption, so a value of an
+// unexpected type must cost a cache entry rather than the whole retrieval — the
+// credential itself is perfectly usable without an expiry.
+type lenientTime string
+
+func (t *lenientTime) UnmarshalJSON(b []byte) error {
+	var s string
+	if json.Unmarshal(b, &s) == nil {
+		*t = lenientTime(s)
+	}
+	return nil
 }
 
 // parseExpireTime parses the service's expiry into a time.Time, collapsing an
@@ -531,11 +545,11 @@ type credentialPayload struct {
 // "lifetime unknown" and decline to cache, which is the safe reading of both:
 // the service omits the field when the token may be permanent or when it cannot
 // say, and a value it sent but we cannot read tells us no more than silence.
-func parseExpireTime(s string) time.Time {
-	if s == "" {
+func parseExpireTime(v lenientTime) time.Time {
+	if v == "" {
 		return time.Time{}
 	}
-	t, err := time.Parse(time.RFC3339, s)
+	t, err := time.Parse(time.RFC3339, string(v))
 	if err != nil {
 		return time.Time{}
 	}
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -187,6 +187,9 @@ type Client struct {
 	// invites a caller to use. Erring towards a false miss costs a round trip;
 	// erring the other way discloses one principal's token to another.
 	//
+	// The converse is a precondition this package cannot enforce and states on
+	// Config.HTTPClient: one Client must not authenticate as several identities.
+	//
 	// The value is unique per process as well as within one, so a store that
 	// outlives the process — or is shared by two — misses rather than serving an
 	// entry written by a Client that no longer exists and cannot be identified.
@@ -219,9 +222,16 @@ type Config struct {
 	// used verbatim and ADC is not applied, so it must carry its own credentials
 	// and should refuse redirects for the reason [NewClient] describes.
 	//
+	// It must authenticate as one identity for the Client's whole life. A Client
+	// is a credential-cache dimension precisely because it is assumed to fix who
+	// the credential service mints on behalf of, and a transport that picks its
+	// credentials out of the request context breaks that assumption without this
+	// package being able to see it: every tenant would share one cache entry.
+	// Build a Client per identity instead.
+	//
 	// Two Clients are two cache dimensions even when built from one HTTPClient,
-	// so build a Client once and share it. One per request resolves nothing from
-	// cache and leaves an entry behind per request.
+	// so build a Client once per identity and share it. One per request resolves
+	// nothing from cache and leaves an entry behind per request.
 	HTTPClient *http.Client
 	// AgentIdentityEndpoint overrides the Agent Identity base URL (scheme+host).
 	// It is used as given, not parsed: an http:// value would send the ADC token
--- a/auth/gcp/client.go
+++ b/auth/gcp/client.go
@@ -200,8 +200,12 @@ var (
 	clientNonce = newClientNonce()
 )
 
+// nonceBytes is the width of clientNonce. 128 bits makes a collision between two
+// processes not worth reasoning about.
+const nonceBytes = 16
+
 func newClientNonce() string {
-	var b [16]byte
+	var b [nonceBytes]byte
 	// Documented never to fail, and it panics rather than returning short.
 	_, _ = rand.Read(b[:])
 	return hex.EncodeToString(b[:])
@@ -294,6 +298,10 @@ func NewClient(ctx context.Context, cfg *Config) (*Client, error) {
 // [auth.CredentialStore.Delete] to invalidate that credential ahead of its
 // expiry, on a consent revocation or a logout.
 //
+// Pass the same [ProviderScheme] value the provider was built with. A scheme
+// differing in any field names a different entry, and Delete of a key nothing
+// is filed under succeeds silently, so a mistake here reads as "already gone".
+//
 // It is meaningful only to the process that produced c, and only for as long as
 // c is alive: a Client is one cache dimension, so rebuilding it strands whatever
 // the old one cached.
@@ -351,10 +359,11 @@ type Request struct {
 type Retrieval struct {
 	// Credential authenticates outbound requests as the end user.
 	Credential auth.Credential
-	// ExpiresAt is the credential's expiry, or the zero time when the service
-	// reports no lifetime — which it does when the token may be permanent, or
-	// when it cannot say. Either way the lifetime is unknown, so a caller must
-	// not treat the zero value as "expires now" and must not cache it.
+	// ExpiresAt is the credential's expiry, or the zero time when the lifetime is
+	// unknown: the service reports none when the token may be permanent or when
+	// it cannot say, and one it reports that cannot be parsed arrives here the
+	// same way. A caller must not read the zero value as "expires now", and must
+	// not cache a credential carrying it.
 	ExpiresAt time.Time
 }
 
@@ -495,11 +504,15 @@ func (rejectedOutcome) isOutcome() {}
 type credentialPayload struct {
 	Token  string `json:"token"`
 	Header string `json:"header"`
-	// ExpireTime is RFC 3339, and empty when the service reports no expiry. The
-	// name is the one in the published API surface
+	// ExpireTime is RFC 3339, and empty when the service reports no expiry.
+	//
+	// For Agent Identity the name is the one in the published API surface
 	// (https://agentidentitycredentials.googleapis.com/$discovery/rest?version=v1,
 	// Success.expireTime), which also warns that the token may be revoked, or
-	// expire slightly early through clock skew, before the time it names.
+	// expire slightly early through clock skew, before the time it names. The IAM
+	// Connector service publishes no discovery document to anonymous callers, so
+	// its field is taken to be the same name; if it is not, a connector
+	// credential simply reports no lifetime and is never cached.
 	ExpireTime string `json:"expireTime"`
 }
 
--- a/auth/gcp/provider.go
+++ b/auth/gcp/provider.go
@@ -47,6 +47,11 @@ type ProviderConfig struct {
 	// Client reaches the credential services. When nil, a default client (backed
 	// by Application Default Credentials) is created lazily on first use.
 	Client *Client
+	// Store caches resolved credentials across requests (keyed by app, user, and
+	// resource). When nil, an in-memory store is used. Caching matters here
+	// because each miss is a network round-trip (and up to a ~10s pending poll)
+	// to the credential service.
+	Store auth.CredentialStore
 }
 
 // ErrNoActingUser means the provider could not determine the acting end user,
@@ -73,13 +78,19 @@ func NewProvider(scheme Scheme, cfg *ProviderConfig) (auth.CredentialProvider, e
 	if cfg.Client != nil && cfg.Client.httpClient == nil {
 		return nil, errors.New("gcp: ProviderConfig.Client must come from NewClient")
 	}
-	// Defensive copy: the provider outlives this call and re-reads Scopes per request, so it must not alias a caller-mutable slice.
+	// Defensive copy: the provider outlives this call and reads Scopes on every
+	// request, so it must not alias a slice the caller can mutate later.
 	scheme.Scopes = slices.Clone(scheme.Scopes)
-	return &provider{scheme: scheme, client: cfg.Client}, nil
+	store := cfg.Store
+	if store == nil {
+		store = auth.NewInMemoryCredentialStore()
+	}
+	return &provider{scheme: scheme, store: store, client: cfg.Client}, nil
 }
 
 type provider struct {
 	scheme Scheme
+	store  auth.CredentialStore
 
 	mu         sync.Mutex
 	client     *Client
@@ -98,16 +109,33 @@ func (p *provider) Credential(ctx context.Context) (auth.Credential, error) {
 		return nil, fmt.Errorf("%w: invocation for app %q session %q has an empty UserID", ErrNoActingUser, id.AppName, id.SessionID)
 	}
 
+	key := auth.CredentialKey{AppName: id.AppName, UserID: id.UserID, Key: p.scheme.Name}
+	// A store read error is non-fatal: fall through and fetch a fresh credential.
+	if cred, ok, err := p.store.Get(ctx, key); err == nil && ok {
+		return cred, nil
+	}
+
 	client, err := p.resolveClient(ctx)
 	if err != nil {
 		return nil, err
 	}
-	return client.RetrieveCredential(ctx, Request{
+	r, err := client.RetrieveCredential(ctx, Request{
 		Resource:    p.scheme.Name,
 		UserID:      id.UserID,
 		Scopes:      p.scheme.Scopes,
 		ContinueURI: p.scheme.ContinueURI,
 	})
+	if err != nil {
+		return nil, err
+	}
+	// Cache only when the service reported an expiry: a zero time means "never
+	// expires" to the store, and the GCP services omit it only when the lifetime
+	// is unknown — caching that would risk serving a stale credential forever.
+	// Best-effort: a store write failure must not fail auth.
+	if !r.ExpiresAt.IsZero() {
+		_ = p.store.Set(ctx, key, r.Credential, r.ExpiresAt)
+	}
+	return r.Credential, nil
 }
 
 // resolveClient returns the configured client, creating a default one (backed by
--- a/auth/gcp/provider.go
+++ b/auth/gcp/provider.go
@@ -19,12 +19,15 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"strings"
 	"sync"
+	"time"
 
 	"golang.org/x/sync/singleflight"
 
 	"google.golang.org/adk/v2/agent"
 	"google.golang.org/adk/v2/auth"
+	"google.golang.org/adk/v2/platform"
 )
 
 // Scheme identifies a GCP auth resource and the access it requests. It mirrors
@@ -85,11 +88,25 @@ func NewProvider(scheme Scheme, cfg *ProviderConfig) (auth.CredentialProvider, e
 	if store == nil {
 		store = auth.NewInMemoryCredentialStore()
 	}
-	return &provider{scheme: scheme, store: store, client: cfg.Client}, nil
+	return &provider{scheme: scheme, slot: schemeSlot(scheme), store: store, client: cfg.Client}, nil
+}
+
+// maxCachedLifetime caps how long a credential is cached, whatever expiry the
+// service reports, so a bad or injected expireTime cannot pin one indefinitely.
+const maxCachedLifetime = time.Hour
+
+// schemeSlot is the store slot for a scheme. Scopes and the continue URI shape
+// what the minted token authorizes, so credentials that differ in them must not
+// share an entry; the sort makes the slot independent of the caller's ordering.
+func schemeSlot(s Scheme) string {
+	scopes := slices.Clone(s.Scopes)
+	slices.Sort(scopes)
+	return s.Name + "|" + strings.Join(scopes, ",") + "|" + s.ContinueURI
 }
 
 type provider struct {
 	scheme Scheme
+	slot   string
 	store  auth.CredentialStore
 
 	mu         sync.Mutex
@@ -109,9 +126,13 @@ func (p *provider) Credential(ctx context.Context) (auth.Credential, error) {
 		return nil, fmt.Errorf("%w: invocation for app %q session %q has an empty UserID", ErrNoActingUser, id.AppName, id.SessionID)
 	}
 
-	key := auth.CredentialKey{AppName: id.AppName, UserID: id.UserID, Key: p.scheme.Name}
+	// The slot covers everything that shapes the minted token, not just the
+	// resource: a store shared by a broad and a narrow provider would otherwise
+	// serve the broad token to the narrow one.
+	key := auth.CredentialKey{AppName: id.AppName, UserID: id.UserID, Key: p.slot}
 	// A store read error is non-fatal: fall through and fetch a fresh credential.
-	if cred, ok, err := p.store.Get(ctx, key); err == nil && ok {
+	// A hit carrying no credential is treated as a miss; the interface allows it.
+	if cred, ok, err := p.store.Get(ctx, key); err == nil && ok && cred != nil {
 		return cred, nil
 	}
 
@@ -128,12 +149,15 @@ func (p *provider) Credential(ctx context.Context) (auth.Credential, error) {
 	if err != nil {
 		return nil, err
 	}
-	// Cache only when the service reported an expiry: a zero time means "never
-	// expires" to the store, and the GCP services omit it only when the lifetime
-	// is unknown — caching that would risk serving a stale credential forever.
-	// Best-effort: a store write failure must not fail auth.
+	// Cache only when the service reported an expiry; it omits one when the
+	// lifetime is unknown, and a credential with no known lifetime cannot be
+	// vouched for later. Best-effort: a store write failure must not fail auth.
 	if !r.ExpiresAt.IsZero() {
-		_ = p.store.Set(ctx, key, r.Credential, r.ExpiresAt)
+		expiresAt := r.ExpiresAt
+		if capped := platform.Now(ctx).Add(maxCachedLifetime); expiresAt.After(capped) {
+			expiresAt = capped
+		}
+		_ = p.store.Set(ctx, key, r.Credential, expiresAt)
 	}
 	return r.Credential, nil
 }
--- a/auth/gcp/provider.go
+++ b/auth/gcp/provider.go
@@ -70,13 +70,15 @@ type ProviderConfig struct {
 	//
 	// One store can safely back several providers: entries are keyed by the app,
 	// the end user, and everything that decides what the service returns — the
-	// resource, the scopes, the continue URI, and the Client. Providers whose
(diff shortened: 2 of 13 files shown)


auth/gcp/provider.go
```diff
@@ -84,16 +95,33 @@ func (p *provider) Credential(ctx context.Context) (auth.Credential, error) {
 		return nil, errors.New("gcp: no acting user in ADK context; provider must run within an agent invocation")
 	}
 
+	key := auth.CredentialKey{AppName: id.AppName, UserID: id.UserID, Key: p.scheme.Name}
```
karolpiotrowicz: The key covers app, user and resource, but not `Scopes` or `ContinueURI` — and both are sent on the wire (`retrieveRequest`), so they determine what the minted token authorizes.

Because `ProviderConfig.Store` is exported so one store can back several providers, two providers for the same resource that differ only in scopes collide. I measured this with a shared store: a `drive` provider and a `drive.readonly` provider, same user — the read-only provider was served the **broad** `drive` token, and its own request never reached the service (1 service call total). With the default per-provider store there's no collision (2 calls, correct tokens each).

Direction matters: broad-then-narrow hands out more authority than was asked for; narrow-then-broad produces a confusing 403. For reference, `adk-python` derives its credential key from a digest of the whole scheme, scopes included.

Suggestion — widen the key while the package is still unreleased:

```go
// in NewProvider, after the Scopes clone:
slices.Sort(scheme.Scopes)

// in Credential:
key := auth.CredentialKey{
	AppName: id.AppName,
	UserID:  id.UserID,
	Key:     p.scheme.Name + "|" + strings.Join(p.scheme.Scopes, ",") + "|" + p.scheme.ContinueURI,
}
```

If sharing a store across differently-scoped providers isn't a configuration you want to support, documenting that on `ProviderConfig.Store` would be a reasonable alternative.

auth/gcp/provider.go
```diff
@@ -84,16 +95,33 @@ func (p *provider) Credential(ctx context.Context) (auth.Credential, error) {
...
+	// A store read error is non-fatal: fall through and fetch a fresh credential.
+	if cred, ok, err := p.store.Get(ctx, key); err == nil && ok {
```
karolpiotrowicz: Minor robustness point: `Get` returns `(Credential, bool, error)`, so `(nil, true, nil)` is representable, and this returns it straight through — `Credential()` then hands the caller `(nil, nil)`. I confirmed that with a store implementation returning a nil credential on a hit.

`auth.Transport` fails closed on a nil credential, so nothing in-repo breaks, but `CredentialProvider` is exported and third-party callers may not check. Treating it as a miss keeps the failure local:

```go
if cred, ok, err := p.store.Get(ctx, key); err == nil && ok && cred != nil {
	return cred, nil
}
```

Rejecting a nil credential in `Set` would close the other half.

auth/gcp/provider.go
```diff
@@ -47,6 +47,11 @@ type ProviderConfig struct {
 	// Client reaches the credential services. When nil, a default client (backed
 	// by Application Default Credentials) is created lazily on first use.
 	Client *Client
+	// Store caches resolved credentials across requests (keyed by app, user, and
+	// resource). When nil, an in-memory store is used. Caching matters here
+	// because each miss is a network round-trip (and up to a ~10s pending poll)
+	// to the credential service.
+	Store auth.CredentialStore
```
karolpiotrowicz, line 85: No test sets this field, so the store plumbing is unasserted — including the best-effort property described at line 120 ("a store write failure must not fail auth"), which is currently stated only in a comment. A fake store that records calls and one that always errors would pin both, and the two-user test suggested on line 98 could take the store as a parameter and cover this at no extra cost.

auth/gcp/provider_test.go
```diff
@@ -87,6 +88,68 @@ func TestNewProviderValidatesScheme(t *testing.T) {
 	}
 }
 
+func TestProviderCachesCredential(t *testing.T) {
```
karolpiotrowicz: This test asserts the call count but never that two principals stay separate, and that turns out to be the only thing standing between the cache and a cross-user leak.

Concretely: replacing the key construction at `provider.go:98` with `key := auth.CredentialKey{}` — one global entry shared by every user in the process — leaves the **whole** test suite green (70 packages, 0 failures), and that mutant genuinely leaks: with it, a second user receives the first user's token. The bar is reasonable, though — the control mutant (removing the `!r.ExpiresAt.IsZero()` guard) *is* caught, by `TestProviderSkipsCacheWithoutExpiry` on line 132. The gap is specific to the key.

A test along these lines would close it:

```go
func TestProviderCacheIsolatesPrincipals(t *testing.T) {
	// server echoes the requesting userId back in the token
	// resolve for user-1 and user-2 (and for the same user under two app names)
	// assert each receives its own token, on the cold read and on the cached read
}
```

auth/store.go
```diff
@@ -0,0 +1,94 @@
...
+// CredentialStore caches resolved credentials across calls, keyed by
+// [CredentialKey]. It exists so network-backed providers (e.g. auth/gcp) avoid a
+// credential-service round-trip on every request; self-caching providers
+// (oauth2.TokenSource) do not need it. Implementations must be safe for
+// concurrent use.
+type CredentialStore interface {
+	// Get returns the cached, unexpired credential for key, if present.
+	Get(ctx context.Context, key CredentialKey) (Credential, bool, error)
+	// Set stores cred for key. A zero expiresAt means the entry does not expire.
+	Set(ctx context.Context, key CredentialKey, cred Credential, expiresAt time.Time) error
```
karolpiotrowicz, line 122: `CredentialStore` is a bare exported interface with no unexported method, so its method set freezes the moment this ships: any later addition breaks every third-party implementer. Two things are worth settling now rather than after release.

**No invalidation.** With only `{Get, Set}` there's no way to purge an entry ahead of its expiry — not on consent revocation, not on logout, not on a downstream 401. A `Delete(ctx context.Context, key CredentialKey) error` costs little today.

**The accepted lifetime is entirely service-chosen.** A response claiming `expireTime: 9999-12-31T23:59:59Z` is honoured verbatim — I checked, and three resolves then produce a single service call. Clamping to a sane maximum in the provider before calling `Set` would bound the exposure of any bad or injected expiry.

auth/store.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// InMemoryCredentialStore is a concurrency-safe, process-local [CredentialStore]
+// (per app+user+key, across sessions). It mirrors adk-python's
+// InMemoryCredentialService.
+type InMemoryCredentialStore struct {
+	mu      sync.Mutex
+	entries map[CredentialKey]cacheEntry
```
karolpiotrowicz: The only removal path is `Get` deleting the key it was asked for (line 80), so nothing ever sweeps: a principal that resolves once and never returns keeps its entry — and its credential — for the lifetime of the process. Measured: 200k already-expired entries survive 10k `Set`/`Get` on a different live key, ~27 MiB retained.

In fairness the shape is benign-ish — growth is O(distinct principals ever seen) at a few hundred bytes each, not O(requests); re-caching the same keys millions of times adds nothing. So whether this needs an LRU or a size cap depends on your expected tenant count. But an amortised sweep on `Set`, or a cap, is easy to add, and a `Delete` on the interface (see line 49) is worth having either way.

auth/store.go
```diff
@@ -0,0 +1,94 @@
...
+// Get implements [CredentialStore]. Expiry is evaluated against the context's
+// clock ([platform.Now]), so tests can drive it deterministically.
+func (s *InMemoryCredentialStore) Get(ctx context.Context, key CredentialKey) (Credential, bool, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	e, ok := s.entries[key]
+	if !ok {
+		return nil, false, nil
+	}
+	if !e.expiresAt.IsZero() && platform.Now(ctx).Add(expirySkew).After(e.expiresAt) {
```
karolpiotrowicz: `platform.Now(ctx)` is caller-supplied code, and it's invoked while `s.mu` is held. `sync.Mutex` isn't reentrant, so a `TimeProvider` that touches the store — directly, or transitively through an instrumented clock — deadlocks the whole cache. I confirmed this: the `Get` never returns and keeps the lock.

Hoisting it out is a one-liner:

```go
now := platform.Now(ctx)
s.mu.Lock()
defer s.mu.Unlock()
...
if !e.expiresAt.IsZero() && now.Add(expirySkew).After(e.expiresAt) {
```

Worth a comment on the mutex too: `Get` mutates the map via `delete`, so `Mutex` here is a correctness constraint, not just a simpler choice — a future switch to `RWMutex` with `RLock` would be unsafe.

auth/store.go
```diff
@@ -0,0 +1,94 @@
...
+
+// CredentialStore caches resolved credentials across calls, keyed by
+// [CredentialKey]. It exists so network-backed providers (e.g. auth/gcp) avoid a
+// credential-service round-trip on every request; self-caching providers
+// (oauth2.TokenSource) do not need it. Implementations must be safe for
+// concurrent use.
+type CredentialStore interface {
+	// Get returns the cached, unexpired credential for key, if present.
+	Get(ctx context.Context, key CredentialKey) (Credential, bool, error)
+	// Set stores cred for key. A zero expiresAt means the entry does not expire.
```
karolpiotrowicz: Making a zero `expiresAt` mean "never expires" puts the fail-open case on the Go zero value, in brand-new public API. The GCP provider handles it correctly (it only calls `Set` when the service reported an expiry), but that guard lives in the consumer rather than in the store, so a second implementer or a future caller gets a permanent entry by omission.

Either invert it (zero = already expired) or, if the current semantics are deliberate, make it explicit here — something like "a zero `expiresAt` means the entry never expires; callers that cannot establish a lifetime should not call `Set`".

auth/store.go
```diff
@@ -0,0 +1,94 @@
...
+	Set(ctx context.Context, key CredentialKey, cred Credential, expiresAt time.Time) error
+}
+
+// InMemoryCredentialStore is a concurrency-safe, process-local [CredentialStore]
+// (per app+user+key, across sessions). It mirrors adk-python's
```
karolpiotrowicz: The comparison is a bit off on the details. The app→user→key bucketing genuinely does match `InMemoryCredentialService`, but: that service has no expiry at all (this one adds per-entry expiry, which is an improvement); its credential key is a digest over the whole auth scheme, scopes included; and for `GcpAuthProviderScheme` — a `CustomAuthScheme` — `CredentialManager` returns the provider's credential directly and never consults the credential service.

Something like "serves the same role as adk-python's `InMemoryCredentialService` (app+user+key bucketing), and adds per-entry expiry" would be accurate.

auth/store.go
```diff
@@ -0,0 +1,94 @@
...
+// InMemoryCredentialService.
+type InMemoryCredentialStore struct {
```
karolpiotrowicz, line 169: A zero-value `InMemoryCredentialStore` is asymmetric: `Get` is a clean miss, but `Set` panics with `assignment to entry in nil map`. Since the type is exported, `var s auth.InMemoryCredentialStore` is a natural thing to write.

Either initialise lazily in `Set` under the lock (`if s.entries == nil { s.entries = make(...) }`) or state on the type that `NewInMemoryCredentialStore` is required.

auth/gcp/client.go
```diff
@@ -212,11 +228,28 @@ func (pendingOutcome) isOutcome()  {}
...
+}
+
+// parseExpireTime parses a proto Timestamp (RFC 3339) into a time.Time, or
+// returns the zero time when empty or malformed.
+func parseExpireTime(s string) time.Time {
+	if s == "" {
+		return time.Time{}
+	}
+	t, err := time.Parse(time.RFC3339, s)
+	if err != nil {
```
karolpiotrowicz, line 434: This error branch is the only uncovered block in the new code. The behaviour is right — I checked malformed, empty, absent, past and `0001-01-01T00:00:00Z` inputs, and all five correctly skip the cache (three resolves → three service calls), so it fails closed. It's just not pinned, on a path that parses input from a network service.

One row in the `TestRetrieveCredential` table with `"expireTime":"not-a-time"`, asserting the credential still comes back with a zero `ExpiresAt`, would cover it.

Worth noting in passing that `0001-01-01T00:00:00Z` parses to exactly `time.Time{}`, so "the service sent proto-zero" and "the service sent nothing" are indistinguishable downstream. Both mean "don't cache" today, so it's harmless — just a little surprising if the zero ever grows a second meaning.

auth/store_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	// A different key is a miss.
+	if _, ok, _ := s.Get(ctx, auth.CredentialKey{AppName: "app", UserID: "other", Key: "res"}); ok {
+		t.Error("Get() for a different user = hit, want miss")
+	}
+}
+
+func TestInMemoryCredentialStoreExpiry(t *testing.T) {
+	ctx := t.Context()
```
karolpiotrowicz, line 48: This case asserts the clock-skew window against the real wall clock, while `TestInMemoryCredentialStoreGetHonorsClock` just below demonstrates the deterministic `platform.WithTimeProvider` seam. The 10s skew gives plenty of headroom so it's unlikely to flake in practice, but using the seam here too would make the intent ("an expiry inside the skew window is treated as expired") explicit rather than timing-dependent.

karolpiotrowicz (commented):
Adding the expiry to the payload *and* the cache in the same change is the right order of operations, and it directly answers the "no caching" concern from the provider PR. I reviewed only the diff against `wolo/auth-gcp-provider`; the rest of `auth/gcp` belongs to the earlier PRs in the stack.

**The headline is a clean result.** Because a credential cache is exactly where cross-user leaks live, I attacked the cache key rather than reading it, and it holds up:

- `CredentialKey` is a **struct used directly as a map key**, so there is no string concatenation and no delimiter/canonicalization surface. I tried 9 adversarial pairs (`/`, `|`, `:`, NUL delimiter shifting, empty components, case, Turkish dotless ı, NFC/NFD) — user `a` + key `b/c` vs user `a/b` + key `c` and friends — **0 collisions**, with a negative control confirming the harness detects a real hit.
- Two users on one provider get their own tokens on both the cold and cached read; the same user id under two different app names does **not** share an entry.
- Excluding `SessionID` from the key is the right call, and I want to call it out explicitly since it's the kind of decision that looks like an omission later.

So I found no cross-principal disclosure. The findings below are about the dimensions the key *doesn't* cover, and about what protects it.

---

### Main points

**1. The cache key omits the scope dimension.** `Scopes` and `ContinueURI` go on the wire and determine what the minted token authorizes, but they aren't in the key. Since `ProviderConfig.Store` is exported specifically so one store can back several providers, two providers for the same resource that differ only in scopes will collide. I measured it: a `drive` provider and a `drive.readonly` provider sharing one store, same user — the read-only provider was served the **broad** `drive` token and its own request never reached the service (1 service call total). The default per-provider store is unaffected (2 calls, correct tokens). Worth noting `adk-python` derives its credential key from a digest of the *whole* scheme, scopes included.

**2. Nothing protects the cache key.** Replacing `provider.go:98` with `key := auth.CredentialKey{}` — one global entry shared by every user in the process — leaves the **entire** test suite green (70 packages, 0 failures), and that mutant is a genuine leak (bob receives alice's token). This isn't an unreasonable bar: the control mutant, removing the `!r.ExpiresAt.IsZero()` guard, *is* caught by `TestProviderSkipsCacheWithoutExpiry`. The gap is specific to the key. A two-user (and two-app) provider test asserting each principal gets its own token would close it, and it's the one thing I'd most like to see before this lands.

**3. `CredentialStore` freezes as `{Get, Set}` with no way to invalidate.** There's no `Delete`/`Invalidate`, so nothing can purge an entry ahead of its expiry — not a revocation, not a logout, not a downstream 401 — and the accepted lifetime is entirely service-chosen: a response claiming `expireTime: 9999-12-31` is honoured verbatim. Because this is a bare exported interface with no unexported method, adding a method after release breaks every third-party implementer. Adding `Delete` (and clamping the accepted expiry) is cheap now and awkward later. The same argument applies to widening `CredentialKey` for point 1.

**4. Eviction.** The only removal is `Get` deleting the key it was asked for, so a principal that never returns retains its entry forever. Measured: 200k already-expired entries survive 10k `Set`/`Get` on a different live key, ~27 MiB retained. To be fair to the design, growth is O(distinct principals) at a few hundred bytes each, not O(requests) — re-caching the same keys millions of times adds nothing. Whether that needs an LRU/cap now depends on your expected tenant count, but a `Delete` on the interface is worth having regardless.

### Smaller items

- `auth/store.go:79` — `platform.Now(ctx)` runs caller-supplied code while `s.mu` is held, and `sync.Mutex` isn't reentrant: a `TimeProvider` that touches the store deadlocks it (confirmed). Hoisting the call above the `Lock` fixes it.
- `auth/gcp/client.go:249-251` — the `time.Parse` error branch is the only uncovered block in the new code. The behaviour is safe (a malformed, empty, absent, past or proto-zero `expireTime` all correctly skip the cache — I checked all five), it's just unpinned; one table row would do it.
- `auth/gcp/provider.go:54` — `ProviderConfig.Store` isn't set by any test, so the plumbing (and the documented "a store write failure must not fail auth" property) is unasserted. The same test that fixes point 2 can take a store and cover this at no extra cost.
- `auth/store.go:100` — a store returning `(nil, true, nil)` makes `Credential()` return `(nil, nil)`. `auth.Transport` fails closed on that, so it's minor, but treating a nil credential as a miss would be more robust.
- `auth/store.go:52-54` — the "mirrors adk-python's `InMemoryCredentialService`" comment is a little off: the app→user→key bucketing does match, but the Python service has no expiry, its key includes scopes, and for `GcpAuthProviderScheme` (a `CustomAuthScheme`) it isn't in the path at all. Maybe "serves the same role as" plus a note that this adds per-entry expiry.
- `auth/store.go:48` — a zero `expiresAt` meaning "never expires" makes the Go zero value the fail-open case. The provider guards it correctly; a second implementer might not.
- `auth/store.go:55` — a zero-value `InMemoryCredentialStore` panics on `Set` (nil map) while `Get` is a clean miss. A lazy init, or a doc line requiring the constructor, would remove the asymmetry.
- `auth/store_test.go:49` — the skew-window case uses the real wall clock while the very next test demonstrates the `platform.WithTimeProvider` seam.

### Questions rather than findings

- Do the credential services actually mint scope-differentiated tokens for a single `authProviders/*` resource? If they don't, point 1 is only a documentation note — I couldn't verify this from outside.
- Is sharing one `Store` across providers a configuration you intend to support? Nothing in-repo does it today, and if it isn't intended, documenting that on the field would be an alternative to widening the key.
- Concurrent cold misses for one identity aren't coalesced (32 goroutines → 32 round trips). This is **not** a regression — I compared against pre-cache behaviour and the change is never worse, and far better in steady state (32 sequential resolves → 1 call vs 32) — but `singleflight` is already imported in the file if the cold-start burst matters against a per-user quota.

One cross-reference: the redirect issue noted on the earlier client PR becomes more durable with caching in place. An injected credential is now stored with an attacker-chosen expiry and kept being served after the source is unreachable, where previously each resolve refetched. Nothing to change here — it's an argument for fixing it upstream, and for clamping the expiry this PR accepts.

Build, `go vet`, `gofmt`, `golangci-lint`, `staticcheck`, `go mod tidy -diff` and `go test -race` are all clean, and `govulncheck`'s two findings are identical at the base commit (zero delta). `TestProviderCachesCredential` correctly fails at the base commit, so the new coverage is real.

karolpiotrowicz (dismissed):
Reviewed at [2e12ee85](https://github.com/google/adk-go/commit/2e12ee85cb1343e2380fe151da1903244b657afa), against the merge base [7e641247](https://github.com/google/adk-go/commit/7e64124704602b2f4daef46240c8da9a3a7a20e6). The store's expiry logic is careful — the skew is applied in the conservative direction, the service's expiry is distrusted and capped, and the clock is resolved outside the lock with a comment explaining why. Three things need fixing before this lands — two cache-key defects and one revocation gap — and two more are worth settling in the same pass: a pair of key dimensions no test can fail on, and the store's growth behavior.

First, though: this branch does not merge into its own base. There are content conflicts in `auth/gcp/provider.go` and `auth/gcp/provider_test.go`, and the branch is 8 commits behind `wolo/auth-gcp-provider`, whose head moved to [6dbe3401](https://github.com/google/adk-go/commit/6dbe3401174444341ed1d10357df2b8221b6b4ea) while this review was running. Those are the two files most of what follows is about, so it is worth rebasing before acting on any of it — some line references will move.

## A shared store serves one caller identity's token to another

The cache key is `{AppName, UserID, slot}` at [provider.go#L132](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/gcp/provider.go#L132), and the slot covers only `Name`, `Scopes` and `ContinueURI` ([provider.go#L101-L105](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/gcp/provider.go#L101-L105)). The `*Client` is not in it, and the `*Client` is what fixes the minting identity — `Config.HTTPClient` is documented at [client.go#L96-L100](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/gcp/client.go#L96-L100) as "used verbatim and ADC is not applied, so it must carry its own credentials". Two providers built with different `HTTPClient` identities and one shared `Store` therefore share a cache entry, and sharing a store across providers is this PR's own tested idiom.

The comment at [provider.go#L129-L131](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/gcp/provider.go#L129-L131) says the slot "covers everything that shapes the minted token". The caller identity shapes it and is not covered.

<details><summary>Reproduction — one endpoint, two caller identities, one shared store</summary>

```go
// in auth/gcp, package gcp_test. Run: go test -run TestClientIdentityNotInKey ./auth/gcp/
type identityRT struct {
	base http.RoundTripper
	who  string
}

func (t identityRT) RoundTrip(r *http.Request) (*http.Response, error) {
	r = r.Clone(r.Context())
	r.Header.Set("X-Caller-Identity", t.who)
	return t.base.RoundTrip(r)
}

func TestClientIdentityNotInKey(t *testing.T) {
	var seen []string
		who := r.Header.Get("X-Caller-Identity")
		seen = append(seen, who)
		// the service mints a token scoped to the CALLING service account
		_, _ = io.WriteString(w, `{"success":{"token":"tok-minted-for-`+who+
			`","header":"Authorization: Bearer","expireTime":"2999-01-01T00:00:00Z"}}`)
	}))
	defer srv.Close()

	store := auth.NewInMemoryCredentialStore()
	scheme := gcp.Scheme{Name: "projects/p/locations/l/authProviders/ap", Scopes: []string{"drive"}}
	mk := func(who string) auth.CredentialProvider {
		hc := &http.Client{Transport: identityRT{base: srv.Client().Transport, who: who}}
		c, _ := gcp.NewClient(t.Context(), &gcp.Config{HTTPClient: hc, AgentIdentityEndpoint: srv.URL})
		p, _ := gcp.NewProvider(scheme, &gcp.ProviderConfig{Client: c, Store: store})
		return p
	}

	_, _ = mk("sa-alpha").Credential(adkContext(t, "user-1"))
	cred, _ := mk("sa-beta").Credential(adkContext(t, "user-1"))
	bc := cred.(auth.BearerCredential)
	if bc.Token != "tok-minted-for-sa-beta" {
		t.Errorf("sa-beta provider served %q; service saw callers %v", bc.Token, seen)
	}
}
```

Observed: the sa-beta provider is served `tok-minted-for-sa-alpha`, and the service records only `[sa-alpha]` as a caller.

</details>

## The slot encoding is not injective

[provider.go#L104](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/gcp/provider.go#L104) joins with `|` and `,`, neither of which is escaped or excluded from the components. `resourceNameRE` does bar `|` from `Name`, but that check runs inside `RetrieveCredential` *after* the store read, and it is never applied to `Scopes` or `ContinueURI` — which is where both collisions live.

```
Scopes:["a,b"]  and  Scopes:["a","b"]                          -> both slot to  "R|a,b|"
{Scopes:["x"], ContinueURI:"y|z"}  and  {Scopes:["x|y"], ContinueURI:"z"}  -> both slot to  "R|x|y|z"
```

Driven through `NewProvider(...).Credential(ctx)` with a shared store, each pair produces **one** service call instead of two, and the second provider is handed a token minted for the first one's scopes. Since the encoding is the thing at fault rather than any one component, length-prefixing or hashing a struct-encoded key would close the whole class rather than the two instances above — I have not run that change, so treat it as a direction and not a verified patch.

## A revoked credential keeps working for up to an hour

[store.go#L53-L56](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/store.go#L53-L56) documents `Delete` as the hook for "consent revocation, logout, or a downstream 401". Nothing outside tests calls it, and `auth.Transport.RoundTrip` returns a downstream 401 to the caller without invalidating anything.

Before this change every `Credential()` was a live round trip, so a revocation took effect on the very next request. After it, warming the cache and then making the service reject returns the cached token with a nil error and never re-consults the service, for `min(service expiry, now+1h)`. That may well be the trade you intend for a credential cache — but right now the interface doc promises an invalidation path that no code path reaches, so either the wiring or the promise should change.

## Two key dimensions have no test that can fail

`AppName` and `ContinueURI` are both in the key by construction and neither is covered. Deleting `AppName` from the key at [provider.go#L132](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/gcp/provider.go#L132) leaves `go test ./auth/...` fully green, and so does dropping `ContinueURI` from the slot. The cause for the first is [provider_test.go#L208](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/gcp/provider_test.go#L208), which hardcodes `AppName: "app"` at the only place a test context is built.

This is worth separating from ordinary coverage talk: `schemeSlot` is at 100% statement coverage and the `ContinueURI` deletion still survives. The `UserID`, `Scopes` and skew dimensions all *do* have tests that die when you break them, so the gap is specific rather than general.

A second, separate test problem sits next to it. `TestProviderClampsServiceExpiry` at [provider_test.go#L388-L391](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/gcp/provider_test.go#L388-L391) hand-writes the slot string and asserts only that no entry exists, with no positive control that one existed beforehand. Any change to the slot format turns it into a silent no-op that also stops guarding `maxCachedLifetime`.

## Deleting the "don't cache without an expiry" guard also leaves the suite green

Removing `if !r.ExpiresAt.IsZero()` at [provider.go#L155-L160](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/gcp/provider.go#L155-L160) and calling `Set` unconditionally breaks no test, because `TestProviderSkipsCacheWithoutExpiry` passes no `Store` and so measures the default store's own rejection rather than the provider's guard. Since `ProviderConfig.Store` is exported, the guard is exactly the thing that protects a third-party store from being asked to cache a credential of unknown lifetime.

The same shape applies to the store's locking: removing every mutex from `auth/store.go` leaves `go test -race ./auth/...` green, because no test drives the store from two goroutines at once. The locking is correct as written — the green race run just is not evidence of it.

## Growth

[store.go#L119-L126](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/store.go#L119-L126) sweeps only on every 256th `Set`, `Get` evicts only the key it looked up, and there is no size cap. 255 principals that each resolve once and never return leave 255 expired entries resident indefinitely, which is the case [store.go#L59-L60](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/store.go#L59-L60) already names. That is bearer-token material staying in the heap well past its usefulness.

## Smaller things

- **`Get`'s three-value return is unspecified.** [store.go#L47-L48](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/store.go#L47-L48) says nothing about `(nil, true, nil)` or about a credential returned alongside a non-fatal error. The provider requires `err == nil && ok && cred != nil`, so a store reporting a degraded backend has its credential silently dropped and the cache becomes permanently ineffective. The comment at [provider.go#L133-L134](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/gcp/provider.go#L133-L134) says "the interface allows it", but the interface does not say so, and this is new exported surface so it is cheapest to pin now.
- **A past-but-well-formed `expireTime` is written to the store.** Only the upper bound is clamped. `InMemoryCredentialStore` shrugs it off on the next `Get`, but a TTL-based implementation computing `expiresAt.Sub(now)` gets a negative TTL.
- **Concurrent cold callers are not coalesced.** [provider.go#L135-L160](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/gcp/provider.go#L135-L160) is an unsynchronized check-then-act, so N concurrent misses on one key can each make their own round trip and each poll for up to the 10s timeout. `singleflight` is already imported and used for the much cheaper client init at [provider.go#L187](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/gcp/provider.go#L187). This is still a strict improvement over the previous per-request round trip, so it is a follow-up rather than a blocker.
- **`RetrieveCredential`'s signature change is source-breaking**, confirmed by `apidiff` as the only incompatible change in either package. It has never appeared in a tagged v2 release — `auth/gcp/client.go` is absent from v2.0.0, v2.1.0 and v2.2.0 — so this only reaches anyone tracking `main`. Worth a line in the PR body rather than any code change.
- **`expirySkew` is unexported**, so a third-party store cannot honor the tolerance the provider assumes, and the provider re-checks nothing after a hit.
- **`ConsentRequiredError.Key` is a `string`** documented as "the credential-store key", while the store's key is now a struct. That field predates this PR and is never populated, but the two now describe the same concept incompatibly.
- **`%+v` on a `CredentialKey`** at [store.go#L110](https://github.com/google/adk-go/blob/2e12ee85cb1343e2380fe151da1903244b657afa/auth/store.go#L110) puts the app name and end-user ID into an error string. No token material, so this is log hygiene rather than credential exposure.
- **An empty `AppName` is accepted** where an empty `UserID` is rejected, so two apps that both omit it share cache entries.

One claim I want to flag as *not* a problem, in case it comes up: `parseExpireTime` using `time.RFC3339` is fine. Go's parser accepts fractional seconds after a seconds field regardless of the layout — `.000Z`, `.123456Z` and `.123456789Z` all parse cleanly on go1.26.6, so protobuf JSON timestamps are handled.

auth/store.go
```diff
@@ -0,0 +1,271 @@
...
+	// differ only in scope would share an entry, and the narrower one would be
+	// served the broader token.
+	//
+	// Derive it from a collision-resistant digest of those inputs rather than by
+	// joining them, so that no delimiter appearing inside one of them can make
+	// two different sets encode alike. auth/gcp is the reference construction.
+	Key string
+}
+
+// The same warning applies to a store that flattens a CredentialKey into one
```
karolpiotrowicz, line 62: This paragraph is separated from `CredentialKey` by the blank line above it and from `CredentialStore`'s doc by the blank line below, so it is attached to no declaration and `go doc -all ./auth` renders it nowhere. I checked: the string `flattens a CredentialKey` appears once in the source and zero times in the rendered package docs.

That matters more here than an orphaned comment usually would. `CredentialStore` carries no unexported method specifically so third parties can implement it, and this is the only place that tells such an implementer not to join the three key fields on a separator. The rule about the *slot* does render, at L56-58. Its extension to a store that flattens all three fields does not, and that is the half with the cross-user consequence.

Deleting the blank line above attaches it to `CredentialKey`. Lint cannot catch this — `revive`'s `exported` and `package-comments` do not look at comment attachment.

auth/gcp/client.go
```diff
@@ -190,9 +245,39 @@ func NewClient(ctx context.Context, cfg *Config) (*Client, error) {
...
+// expiry, on a consent revocation or a logout.
+//
+// Pass the same [ProviderScheme] value the provider was built with. A scheme
+// differing in any field names a different entry, and Delete of a key nothing
+// is filed under succeeds silently, so a mistake here reads as "already gone".
+//
+// It is meaningful only to the process that produced c, and only for as long as
+// c is alive: a Client is one cache dimension, so rebuilding it strands whatever
+// the old one cached.
+func (c *Client) CacheKey(scheme ProviderScheme, appName, userID string) auth.CredentialKey {
```
karolpiotrowicz, line 318: `CacheKey` derives from the caller's live `ProviderScheme`, while `NewProvider` keyed on `slices.Clone(cfg.Scheme.Scopes)` taken at construction. A caller who reuses the slice it passed — which `TestProviderCredential` shows callers do — gets a key that names nothing, and `Delete` returns nil, so the revocation reads as "already gone" while the credential keeps being served for the rest of the hour.

```go
scheme := gcp.ProviderScheme{Name: R, Scopes: []string{"drive"}}
p, _ := gcp.NewProvider(ctx, gcp.ProviderConfig{Scheme: scheme, Client: c, Store: s})
_, _ = p.Credential(adkCtx)
scheme.Scopes[0] = "drive.readonly"          // caller reuses its own slice
_, ok, _ := s.Get(ctx, c.CacheKey(scheme, "app", "user-1"))
// ok == false
```

I ran that against the head and it reproduces. The doc at L257-259 warns that a *different* scheme names a different entry, which is not the same warning — the caller here believes it is passing the same value.

auth/gcp/provider.go
```diff
@@ -185,14 +251,28 @@ func (p *provider) Credential(ctx context.Context) (auth.Credential, error) {
...
 		return nil, err
 	}
-	cred, err := client.RetrieveCredential(ctx, Request{
+
+	key := auth.CredentialKey{AppName: id.AppName, UserID: id.UserID, Key: cacheSlot(client, p.scheme)}
+	// A store read error is non-fatal: fall through and fetch a fresh credential.
+	// A hit carrying no credential is a miss, though the interface forbids one:
+	// a third-party store is not worth failing closed over, and returning a nil
+	// credential to auth.Transport would fail the request anyway.
+	if cred, ok, err := p.store.Get(ctx, key); err == nil && ok && cred != nil {
```
karolpiotrowicz, line 303: The `err == nil` arm is the only implementation of `CredentialStore.Get`'s "callers discard the credential whenever the error is non-nil", and nothing fails if it goes. Rewrite this as `if cred, ok, _ := p.store.Get(ctx, key); ok && cred != nil` and the whole suite stays green, because `recordingStore.Get` returns `(nil, false, getErr)` when `getErr` is set, so `ok == false` masks the arm in every case that exercises it.

A fourth case on `recordingStore` returning `(cred, true, err)`, asserting the provider refetches, would close it.

auth/gcp/provider.go
```diff
@@ -125,9 +148,14 @@ func NewProvider(ctx context.Context, cfg ProviderConfig) (auth.CredentialProvid
 	if cfg.Client != nil && cfg.Client.httpClient == nil {
 		return nil, errors.New("gcp: ProviderConfig.Client must come from NewClient")
 	}
+	store := cfg.Store
+	if store == nil {
+		store = auth.NewInMemoryCredentialStore()
```
karolpiotrowicz, line 173: `ProviderConfig.Store`'s doc promises a *private* in-memory store on the nil path, and no test can fail on that word. Replace this line with a package-level `var defaultStore = auth.NewInMemoryCredentialStore()` and the suite stays green, because every test on the default path builds a fresh `Client` per provider, so no cross-provider hit is observable through a shared map anyway.

Two providers with `Store` unset, one `Client`, resolving the same app/user/scheme, asserting two service calls, would pin it.

auth/store.go
```diff
@@ -0,0 +1,271 @@
...
+
+// Set implements [CredentialStore].
+func (s *InMemoryCredentialStore) Set(_ context.Context, key CredentialKey, cred Credential, expiresAt time.Time) error {
+	if cred == nil {
+		return errors.New("auth: Set requires a credential")
+	}
+	// The key names the app and the end user, so it stays out of the message: an
+	// error text is logged far more freely than the store's contents.
+	if expiresAt.IsZero() {
+		return errors.New("auth: Set requires an expiry")
```
karolpiotrowicz, line 217: The comment three lines up says the key stays out of the message because error text is logged more freely than the store's contents, and nothing enforces it. Change this to `fmt.Errorf("auth: Set requires an expiry for %+v", key)` and the `auth` suite stays green — `TestInMemoryCredentialStoreSetRequiresCredentialAndExpiry` only checks nil-ness.

`auth/gcp` tests the same house rule twice, in `TestProviderErrorAttribution` and `TestProviderNoActingUser`, both asserting caller-supplied ids are absent from the text. Asserting the returned error contains neither `key.AppName` nor `key.UserID` would bring this in line.

auth/gcp/provider.go
```diff
@@ -185,14 +251,28 @@ func (p *provider) Credential(ctx context.Context) (auth.Credential, error) {
...
 	client, err := p.resolveClient(ctx)
 	if err != nil {
 		// Not attributed: a client-init failure is about this process's own
 		// credentials, not about the resource, and every provider in the process
 		// fails it identically. Attributing it would also stack the package prefix.
 		return nil, err
 	}
-	cred, err := client.RetrieveCredential(ctx, Request{
+
+	key := auth.CredentialKey{AppName: id.AppName, UserID: id.UserID, Key: cacheSlot(client, p.scheme)}
```
karolpiotrowicz, line 298: `AppName` is not sent to the credential service, so an empty one can only merge buckets that would have been minted the same credential — no cross-user exposure. Still, `id.UserID` is rejected empty a few lines up and `id.AppName` is not, and now that it is a cache dimension the asymmetry has a visible consequence: two apps that both omit it share entries. The line that checks `UserID` predates this change, but the consequence is new.

auth/gcp/client.go
```diff
@@ -190,9 +245,39 @@ func NewClient(ctx context.Context, cfg *Config) (*Client, error) {
...
+	return auth.CredentialKey{AppName: appName, UserID: userID, Key: cacheSlot(c, scheme)}
```
karolpiotrowicz, line 319: Two things about this signature, both cheap now and permanent later.

`appName` and `userID` are adjacent untyped strings, so `c.CacheKey(scheme, userID, appName)` compiles, returns a well-formed key, and `Delete` on it succeeds silently. The doc's "reads as 'already gone'" warning covers a mismatched scheme but not a transposition.

Separately, a zero `gcp.Client` has an empty `cacheSlot` and this happily digests it. `NewProvider` rejects a `Client` that did not come from `NewClient`, and `CacheKey` has no equivalent guard and cannot fail.

auth/store.go
```diff
@@ -0,0 +1,271 @@
...
+	defer s.mu.Unlock()
+	if s.entries == nil {
+		s.entries = make(map[CredentialKey]cacheEntry)
+	}
+	s.sweep(time.Now())
+	// Round(0) strips the monotonic reading a caller derived expiresAt from its
+	// own clock would carry. Comparing two entries in different clock domains
+	// diverges across a suspend, when the monotonic clock stops but the wall
+	// clock does not.
+	s.entries[key] = cacheEntry{cred: cred, expiresAt: expiresAt.Round(0)}
```
karolpiotrowicz, line 230: `Set` rejects a zero expiry but accepts one already in the past, while `expired` below is documented as the exact complement of the test a producer applies. The provider does apply the floor, so this is unreachable from ADK, and the internal tests deliberately depend on `Set` accepting a spent expiry to build fixtures — so it reads as a deliberate choice. Worth a word on `Set`, since a third-party producer reading the interface doc would take the floor for a store-side guarantee.

auth/gcp/provider.go
```diff
@@ -142,8 +170,46 @@ func NewProvider(ctx context.Context, cfg ProviderConfig) (auth.CredentialProvid
...
+// on this provider's path at all: GcpAuthProviderScheme is a CustomAuthScheme,
+// and CredentialManager returns the provider's credential directly without ever
+// loading or saving one (credential_manager.py). Caching GCP credentials is a Go
+// addition, and this slot is its own design. The nearest analogue is
+// AuthConfig.get_credential_key (auth_tool.py), which joins two digests of
+// canonical JSON — one of the auth scheme, one of the credential used to obtain
+// it. Go cannot produce the second, a Client's credentials being opaque to this
+// package, so it names the Client instead.
+func cacheSlot(c *Client, s ProviderScheme) string {
+	scopes := slices.Clone(s.Scopes)
```
karolpiotrowicz, line 221: Scopes are sorted but not deduplicated, so `["a","a"]` and `["a"]` are different entries. That is a miss rather than a wrong credential, which is the right direction, and deduplicating would assume the service treats the two identically. Noting it only because the sort is here and the compact is not, which reads as an oversight rather than a decision.

auth/store.go
```diff
@@ -0,0 +1,271 @@
...
+// another's credential. Length-prefix or digest them, as auth/gcp does for the
+// slot itself.
+
+// CredentialStore caches resolved credentials across calls, keyed by
+// [CredentialKey]. It exists so network-backed providers (e.g. auth/gcp) avoid a
+// credential-service round-trip on every request; self-caching providers
+// (oauth2.TokenSource) do not need it. Implementations must be safe for
+// concurrent use.
+//
+// It carries no unexported method, so anyone can implement it and its method set
```
karolpiotrowicz, line 76: This paragraph reasons carefully about what freezes when the interface has no unexported method, and stops at the method set. `InMemoryCredentialStore` uses `CredentialKey` directly as a map key, which only works because all three fields are comparable, and every third-party in-memory store will do the same — but nothing says that is a property `CredentialKey` intends to keep. Adding a `[]string` field later would fail to compile every such store, and `apidiff` reports a field addition as compatible.

auth/gcp/provider.go
```diff
@@ -201,7 +281,38 @@ func (p *provider) Credential(ctx context.Context) (auth.Credential, error) {
...
+func (p *provider) cache(ctx context.Context, key auth.CredentialKey, r *Retrieval) {
+	now := time.Now()
+	if !r.ExpiresAt.After(now.Add(auth.ExpirySkew)) {
+		return
+	}
+	expiresAt := r.ExpiresAt
+	if capped := now.Add(maxCachedLifetime); expiresAt.After(capped) {
+		expiresAt = capped
+	}
+	_ = p.store.Set(ctx, key, r.Credential, expiresAt)
```
karolpiotrowicz, line 347: The cache write is best-effort and its error is discarded, but it rides the request context. `InMemoryCredentialStore.Set` ignores its context so the default path is unaffected, and the interface doc explicitly invites a store that persists outside the process — for one of those, a context cancelled between the service response and this call means the entry is never written. The credential has already been paid for by then.

auth/store.go
```diff
@@ -0,0 +1,271 @@
...
+func (s *InMemoryCredentialStore) sweep(now time.Time) {
+	every := s.sweepEvery
+	if every == 0 {
+		every = sweepInterval
+	}
+	if !s.lastSweep.IsZero() && now.Sub(s.lastSweep) < every {
+		return
+	}
+	s.lastSweep = now
+	for k, e := range s.entries {
```
karolpiotrowicz, line 255: The sweep is O(entries) and now runs under the exclusive mutex on the `Get` path too, so once per `sweepInterval` one reader pays a full traversal while every other credential resolution in the process blocks behind it. The pacing is the right fix for what I raised last time and is strictly cheaper than scanning on a fixed fraction of `Set`s, so I am not asking for it back — but the pacing rationale above frames the cost as amortizable CPU, and the part that is not amortizable is the lock hold. The resident count the type's own doc describes (distinct keys within one credential lifetime) is what sizes the stall. I did not measure it.

karolpiotrowicz (changes requested):
Reviewed at [ae357b16](https://github.com/google/adk-go/commit/ae357b1665ee2a916a0e0dd86c0a71b7fd3fc01c), against the merge base [d29b4e86](https://github.com/google/adk-go/commit/d29b4e868107573123eccd7eaa4f712986570061). All three blocking findings from the last round are closed, and two of them are closed in a way I could demonstrate rather than read. The cache key is a length-prefixed digest that now covers the `*Client`, the tests discriminate on every key dimension, and the expiry policy is asserted on what the store was actually handed instead of on absence. Every mutation I could think of against the key now fails a test that names the property it broke.

One new thing blocks it, and it is not something you could have known about on 26 August: `v2.3.0` shipped in the meantime and it carries `auth/gcp`.

## `RetrieveCredential`'s signature change now breaks a released version

At [v2.3.0](https://github.com/google/adk-go/commit/cf61116c6f088f94ae307ec78b8a9424be2f29eb), tagged 31 August, [`client.go#L188`](https://github.com/google/adk-go/blob/cf61116c6f088f94ae307ec78b8a9424be2f29eb/auth/gcp/client.go#L188) is `func (c *Client) RetrieveCredential(ctx context.Context, req Request) (auth.Credential, error)`. This PR changes it to return `*Retrieval` at [`client.go#L311`](https://github.com/google/adk-go/blob/ae357b1665ee2a916a0e0dd86c0a71b7fd3fc01c/auth/gcp/client.go#L311). `auth/gcp` reached `main` when #1149 merged on 19 August, and `go list -m -versions google.golang.org/adk/v2` returns `v2.0.0 v2.1.0 v2.2.0 v2.3.0`, so the old signature is on the module proxy and that tag is immutable.

The PR body still says `auth/gcp` "is absent from v2.0.0, v2.1.0 and v2.2.0, so this reaches only someone tracking `main`". That was right when you wrote it and is no longer right.

The break belongs to this PR rather than to its base. Comparing exported API against the v2.3.0 tree: the merge base is compatible-only over it, adding `ErrClientUnavailable`, `ErrNoActingUser`, `NewProvider`, `ProviderConfig` and `ProviderScheme`, while this head reports exactly one incompatible change, the `RetrieveCredential` signature. The `auth` package itself stays additive-only at both revisions.

The repository's own check agrees and is currently red — the [apidiff job](https://github.com/google/adk-go/actions/runs/32917318742/job/98023635515) reports the same single incompatible change and exits 1, with every other check green. That workflow also prints the way out: label the PR `breaking-change` and the comparison still records the break but stops failing. The PR carries no labels today. AGENTS.md puts breaking the public API under **Never** and any breaking change under **Ask first**, so applying that label is a maintainer decision to make explicitly rather than something a reviewer can infer. Making the change additive instead — leaving `RetrieveCredential` alone and putting the expiry on a new method — is the other way, and `Retrieval` is already a compatible addition either way.

I could not find any basis for treating `auth/gcp` as exempt: `doc.go` calls the *services* preview, not the Go API, and there is no `internal/` segment, build tag, or per-package versioning carve-out in CONTRIBUTING.md.

## The three earlier blockers

**The `*Client` is in the key.** [`cacheSlot`](https://github.com/google/adk-go/blob/ae357b1665ee2a916a0e0dd86c0a71b7fd3fc01c/auth/gcp/provider.go#L200-L208) puts `c.cacheSlot` first among the hashed fields. Removing it fails `TestProviderCacheDimensions/caller_identity` and `TestProviderCacheSeparatesClientInstances`, so the dimension is pinned rather than merely present.

**The encoding is injective.** [`joinFields`](https://github.com/google/adk-go/blob/ae357b1665ee2a916a0e0dd86c0a71b7fd3fc01c/auth/gcp/client.go#L271-L279) length-prefixes every field before the digest. Reverting it to a plain `|` join reproduces both of the collision pairs I gave last time, inside `TestProviderCacheDimensions`, and also fails `TestJoinFieldsIsInjective`. The exhaustive test over an alphabet containing `:`, `|` and `,` is the right shape for this.

**Revocation is answered, differently from how I framed it.** I asked for either the wiring or the promise to change, and you changed the promise: [`store.go#L123-L139`](https://github.com/google/adk-go/blob/ae357b1665ee2a916a0e0dd86c0a71b7fd3fc01c/auth/store.go#L123-L139) now states that ADK does not call `Delete` and bounds the staleness, and `Client.CacheKey` gives a caller a real hook, which `TestClientCacheKeyMatchesWhatTheProviderWrote` drives end to end. That answers it. What remains is narrower and is already written down at [`provider.go#L78-L84`](https://github.com/google/adk-go/blob/ae357b1665ee2a916a0e0dd86c0a71b7fd3fc01c/auth/gcp/provider.go#L78-L84): a provider built without an explicit `Client` never exposes the one it builds, so on the documented default path there is no key to hand `Delete` at all. I confirmed that by construction rather than by reading. Worth deciding whether the default path should be able to invalidate, or whether the doc should say plainly that it cannot.

## One thing the `Store` doc does not tell an operator

[`ProviderConfig.Store`](https://github.com/google/adk-go/blob/ae357b1665ee2a916a0e0dd86c0a71b7fd3fc01c/auth/gcp/provider.go#L66-L85) reads as an invitation to share one store across providers, and the natural next step for a multi-replica deployment is to implement `auth.CredentialStore` over something shared. That store will never serve a hit, because the `Client` dimension is a per-process random nonce. The consequence is written down — on [`Client.CacheKey`](https://github.com/google/adk-go/blob/ae357b1665ee2a916a0e0dd86c0a71b7fd3fc01c/auth/gcp/client.go#L261-L263), which is not where someone configuring a store is reading. The behaviour itself is the right call and the reasoning behind it convinced me. It costs a round trip in the wrong case and discloses nothing, which is the correct direction. It just needs to be said at the field.

The rest is on the diff. Two pre-existing items I am not charging to this PR, in case they are of interest: the `resolveClient` error arm inside `Credential` is executed by no test at either revision, and the panic and nil-client arms of `runInit` do not wrap `ErrClientUnavailable`, both unchanged from the base. Coverage on `auth` went from 82.7% to 87.2% and on `auth/gcp` from 94.6% to 95.2%.

One process note: GitHub currently reports this branch as conflicting, against a cached base sha that is no longer the branch head. A local three-way merge into the current `wolo/auth-gcp-provider` head is clean, so a push may be all that is needed to clear it.

auth/store.go
```diff
@@ -0,0 +1,271 @@
...
+	Key string
+}
+
+// The same warning applies to a store that flattens a CredentialKey into one
+// identifier of its own — a row key, a filename, a cache line. AppName and
+// UserID come off the request and ADK does not authenticate either, so joining
+// the three fields on a separator lets {app "acme", user "bob|X"} and
+// {app "acme|bob", user "X"} name one entry, and one end user is then served
+// another's credential. Length-prefix or digest them, as auth/gcp does for the
+// slot itself.
```
karolpiotrowicz, line 68: This paragraph has a blank line on both sides, so it attaches to no declaration and `go doc -all ./auth` renders none of it. I checked at this head — none of "flattens a CredentialKey", "row key, a filename" or "Length-prefix or digest" appears in the output.

The digest-not-join rule does survive on the `Key` field just above, so the core advice reaches a reader. What is lost is its extension to a store that flattens all three fields into one identifier, which is the case this text was written for and the one where an unauthenticated `AppName` and `UserID` can collide two principals.

Folding it into the `Key` field doc, or into the `CredentialStore` doc, would put it where an implementer is actually reading.

auth/providers.go
```diff
@@ -165,8 +167,11 @@ func ServiceAccount(cfg ServiceAccountConfig) CredentialProvider {
 			return ts, nil
 		}
 		if len(cfg.JSONKey) > 0 {
-			// Stricter than adk-python (scopes optional there): an explicit-key
-			// access token is scope-bound, so no scopes = unusable — fail fast.
+			// An explicit-key access token is scope-bound, so no scopes = unusable.
+			// Fail at wiring time rather than at request time, which is also where
```
karolpiotrowicz, line 171: This says the check fails "at wiring time rather than at request time", and it fails at request time. The `if len(cfg.Scopes) == 0` below sits inside the closure handed to `lazyTokenSource`, which runs it on the first `Credential(ctx)` call rather than in `ServiceAccount`.

The check itself is unchanged from the base, so only the claim is new. Worth noting `ServiceAccount` returns a bare `CredentialProvider` with no error, so it could not fail at construction even if you wanted it to — which is probably the distinction the comment was reaching for, and it is worth spelling out that way.

auth/gcp/provider.go
```diff
@@ -233,7 +313,38 @@ func (p *provider) Credential(ctx context.Context) (auth.Credential, error) {
...
+//
+// The far end is clamped rather than rejected, so a wildly distant expiry —
+// wrong, or injected — shortens to the cap instead of pinning the entry.
+//
+// Wall clock, not platform.Now: what is being decided is when a real credential
+// stops working, which no simulated clock changes. [auth.CredentialStore.Set]
+// says the same of the value written here.
+func (p *provider) cache(ctx context.Context, key auth.CredentialKey, r *Retrieval) {
+	now := time.Now()
+	if !r.ExpiresAt.After(now.Add(auth.ExpirySkew)) {
```
karolpiotrowicz, line 340: This floor has no deterministic boundary test, while its store-side complement does.

`store.go` states the two are exact complements — the provider stores exactly what the store will serve and nothing it won't. `TestExpiredBoundary` pins the store side at a fixed clock, equality case included. On this side, the only exercise is `TestProviderCachedExpiry`'s "exactly the store's margin" case, which sets `expireTime` to `issued + 10s` and is then read at a strictly later `now`, so it lands inside the margin rather than at equality. I measured about 11µs of drift on the store-side equivalent, and this path additionally spans a test server build and a live round trip.

An equality-inverted floor here would survive the suite. The consequence is small — one guaranteed-dead write, self-healing on the next read — but the documented complement relationship is only half-guarded.

karolpiotrowicz (commented):
Reviewed at [4f9b60fc](https://github.com/google/adk-go/commit/4f9b60fceabd69cde65dba9b960075c74abdd32a), against the merge base [a7826f5f](https://github.com/google/adk-go/commit/a7826f5f04e139c637eaccfd93b9464138a62e58). The branch has been rebased since I last looked, so my inline threads from 4 September point at a revision that no longer exists. The two things that block are here in the body, and the smaller items are on their own lines.

Worth saying up front what the rebase did and did not change. Nine of the ten commits are identical in content, and the tenth is net-zero inside the review range: `func (p *provider) attribute` exists at neither the merge base nor the head, so it was added and removed within the branch. The diff against the new base contains no code the previous one did not. That means the `RetrieveCredential` break is still open for exactly the reason it was open before, and it also means I went looking harder this round rather than re-reading the same thing. One new item came out of that, and it is the substantive part of this comment.

## `RetrieveCredential` still breaks the published v2.3.0

Unchanged from my last review, so I will not restate the argument. The state today: `v2.3.0` is still the newest published version, [`client.go#L188`](https://github.com/google/adk-go/blob/cf61116c6f088f94ae307ec78b8a9424be2f29eb/auth/gcp/client.go#L188) there still returns `auth.Credential`, [`client.go#L342`](https://github.com/google/adk-go/blob/4f9b60fceabd69cde65dba9b960075c74abdd32a/auth/gcp/client.go#L342) here returns `*Retrieval`, the `apidiff` check is red on this head, and the PR carries no labels.

I did re-check attribution against the new base, since the base moved. Comparing exported API against the v2.3.0 tree, the merge base is compatible-only over it with six additions, while this head reports the same single incompatible change. So the break still belongs to this PR rather than to #1173.

One thing I checked and want to record because it looks like an out: the workflow file triggers on base `main` and `v1`, and this PR's base is neither, so the check should not have applied to a stacked branch. It ran anyway. The failing run is attributed to this PR with base `wolo/auth-gcp-provider`, and it exits 1 with the "label the pull request `breaking-change`" message.

**This is what has to change before merge.** Either keep `RetrieveCredential` at its released signature and put the expiry on a second method, or apply the `breaking-change` label so the break is recorded deliberately. The second is a maintainer's call to make explicitly rather than something I can infer, which is why I am asking rather than picking.

## Nothing pins which credential a cache hit serves

This is the new one, and it is the other thing I would want fixed before this merges.

The three tests that deliberately exercise a cache hit all discard the credential and assert only how many times the service was called: [`provider_test.go#L828-L836`](https://github.com/google/adk-go/blob/4f9b60fceabd69cde65dba9b960075c74abdd32a/auth/gcp/provider_test.go#L828-L836), [`#L931-L938`](https://github.com/google/adk-go/blob/4f9b60fceabd69cde65dba9b960075c74abdd32a/auth/gcp/provider_test.go#L931-L938) and [`#L947-L954`](https://github.com/google/adk-go/blob/4f9b60fceabd69cde65dba9b960075c74abdd32a/auth/gcp/provider_test.go#L947-L954). Each is `if _, err := p.Credential(...)` followed by a check on `e.calls()`.

The tests that do assert token values go the other way. `TestProviderCacheDimensions` and `TestProviderStoreWritesTheKeyItRead` both assert `calls == 2`, so every resolve in them is a miss by construction. Tests that reach the hit branch never look at what came back, and tests that look never reach it.

So the key derivation is pinned hard — genuinely, and that work holds up — while the step that hands the entry to the caller is not pinned at all. Here is the check, which takes about a minute:

```
# in auth/gcp/provider.go, replace the cache-hit branch
-	if cred, ok, err := p.store.Get(ctx, key); err == nil && ok && cred != nil {
-		return cred, nil
-	}
+	if _, ok, err := p.store.Get(ctx, key); err == nil && ok {
+		return auth.BearerCredential{Token: "MUTANT-WRONG-TOKEN"}, nil
+	}

go test ./auth/...
```

Every cache hit now serves a bearer token belonging to nobody, and the suite passed 2 of the 3 runs I gave it. Per test, over 5 runs each: `TestProviderCachesCredential` 5 pass, `TestProviderCacheHitsAcrossProviders` 5 pass, `TestProviderCacheIgnoresScopeOrder` 5 pass. The only test that ever catches it is `TestProviderCredentialConcurrent`, which does assert the exact token, and which caught it 3 times in 5 — its hits are a race between 32 goroutines and a cold cache, so in CI, where the suite runs once, this ships about two times in three.

I checked the branch is actually reached rather than skipped: an atomic counter on it records 4 executions in a full package run, one each from the three tests above and the rest from the concurrent one.

The strictly true version of this needs no mutant at all — a test that binds the credential to `_` cannot fail on its value, whatever the coverage says. Coverage on `auth/gcp` rose to 97.0% across exactly this gap, because statement coverage measures reach and not what was asserted.

The cheapest fix is in `TestProviderCachesCredential`, which already has an `echoServer` producing a token that names the request it answered — capture the second `Credential()` result and compare it to the first.

## Smaller things

Five of these are on the lines themselves rather than here: an orphaned doc paragraph in `auth/store.go` that renders nowhere, a new comment in `auth/providers.go` that says "wiring time" where the code means request time, the caching floor in `auth/gcp/provider.go` having no deterministic boundary test where its store-side complement does, `CacheKey`'s two adjacent untyped strings, and `Get` reading the clock outside the lock that `Set` and `Delete` read it inside.

## Things I looked at and am not raising

The unbounded store size and the O(entries) sweep, the absent single-flight on a cold burst, the `Delete`-versus-in-flight-retrieval window, and the default path's inability to name its own cache entries are all declared in the code with their bounds and their remedies, so I am treating them as costs you chose rather than things you missed. I took no load or residency profile, so I have nothing to say about the first two beyond what the comments already say.

Two process notes. This branch conflicts with #1465, which also touches `auth/gcp/client_test.go`, and whichever lands second will need the rebase. And I have two `CHANGES_REQUESTED` reviews stacked on this PR — the one from 25 August is stale, since all three of its blockers are closed, so I will clear it.

auth/gcp/client.go
```diff
@@ -215,9 +270,39 @@ func NewClient(ctx context.Context, cfg *Config) (*Client, error) {
...
+// expiry, on a consent revocation or a logout.
+//
+// Pass the same [ProviderScheme] value the provider was built with. A scheme
+// differing in any field names a different entry, and Delete of a key nothing
+// is filed under succeeds silently, so a mistake here reads as "already gone".
+//
+// It is meaningful only to the process that produced c, and only for as long as
+// c is alive: a Client is one cache dimension, so rebuilding it strands whatever
+// the old one cached.
+func (c *Client) CacheKey(scheme ProviderScheme, appName, userID string) auth.CredentialKey {
```
karolpiotrowicz, line 318: `appName` and `userID` are adjacent untyped strings, so a transposition at the call site compiles and produces a well-formed key. Combined with `Delete` returning nil for a key nothing is filed under, a revocation handler that swaps them reports success while the real entry keeps being served for the rest of its cached lifetime.

The doc below warns that a mismatched `ProviderScheme` reads as "already gone", which is the same hazard, but not about the argument order — and the two bare strings are the easier slip of the two. A named struct, or a doc line, would close it.

auth/store.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// NewInMemoryCredentialStore returns an empty [InMemoryCredentialStore].
+func NewInMemoryCredentialStore() *InMemoryCredentialStore {
+	return &InMemoryCredentialStore{entries: make(map[CredentialKey]cacheEntry)}
+}
+
+// Get implements [CredentialStore].
+func (s *InMemoryCredentialStore) Get(_ context.Context, key CredentialKey) (Credential, bool, error) {
+	now := time.Now()
```
karolpiotrowicz, line 193: `Get` reads the clock here, before taking the lock on the next line, where `Set` and `Delete` both read it after. Under contention the reading is stale by the lock wait by the time `expired` uses it.

I tried to turn this into a served-but-spent entry under a 300ms concurrent write storm and could not — the reader still reported a miss. The sweep cannot be backdated either: a negative delta is below the interval, so the guard returns before `lastSweep` is written. So this is worth aligning for consistency with the other two methods rather than for any consequence I could produce.

karolpiotrowicz:
@wolo-lab — the compatibility note in the description undercounts the releases this reaches, and I would like your call on the direction before this lands rather than after.

The description says `auth/gcp` is absent from v2.0.0, v2.1.0 and v2.2.0, so the signature change reaches only someone tracking `main`. That holds for those three, but the list stops two releases short. `auth/gcp` shipped in **v2.3.0** and **v2.4.0**, and both carry the current signature. I checked the published module zips rather than the tags, so this is what a user actually downloaded:

```
$ curl -s https://proxy.golang.org/google.golang.org/adk/v2/@v/list
v2.0.0  v2.1.0  v2.2.0  v2.3.0  v2.4.0

$ unzip -p v2.3.0.zip 'google.golang.org/adk/v2@v2.3.0/auth/gcp/client.go' | grep RetrieveCredential
func (c *Client) RetrieveCredential(ctx context.Context, req Request) (auth.Credential, error) {

$ unzip -p v2.4.0.zip 'google.golang.org/adk/v2@v2.4.0/auth/gcp/client.go' | grep RetrieveCredential
func (c *Client) RetrieveCredential(ctx context.Context, req Request) (_ auth.Credential, err error) {
```

So this is a removal from a released v2 surface, not a change confined to `main`. Anyone who wrote `cred, err := client.RetrieveCredential(ctx, req)` against either release stops compiling on the next one, at [`client.go:292`](https://github.com/google/adk-go/blob/v2.4.0/auth/gcp/client.go#L292) as released against [`client.go:396`](https://github.com/google/adk-go/blob/52bfd81cdad5cdb887d9c923f6f667c39cc33d7c/auth/gcp/client.go#L396) here. That is also why apidiff is red rather than merely reporting.

Two things that seem worth weighing, neither of which is my call:

- **Nothing inside the repo needs the break.** On [`main`](https://github.com/google/adk-go/blob/4e57df40cbd35f055ba654b03ffb0e509fdd754a/auth/gcp/client.go#L292) there is not one non-test caller of `RetrieveCredential`, and in this PR the only caller is [`provider.go:307`](https://github.com/google/adk-go/blob/52bfd81cdad5cdb887d9c923f6f667c39cc33d7c/auth/gcp/provider.go#L307). The cost falls entirely on external callers and the benefit is entirely internal, so the trade is not the usual one where a break buys a cleaner call site for everybody.
- **The expiry genuinely has to come from somewhere.** `auth.Credential` is a one-method interface with no lifetime on it, so the cache cannot key expiry off the existing return type — [`Retrieval`](https://github.com/google/adk-go/blob/52bfd81cdad5cdb887d9c923f6f667c39cc33d7c/auth/gcp/client.go#L369) is answering a real need, and I am not arguing it away.

What I would like confirmed is which way you want to go: take the break and label the PR `breaking-change` so it is on the record and the gate stops failing, or keep `RetrieveCredential` as released and put the expiry on a second entry point. I have not implemented the second option, so I am naming it as an alternative rather than recommending it — the version-policy question is yours either way, and it is the part I do not want decided implicitly by a red check.

PR #1175, opened 2026-07-19
fix(memory/vertexai): scope MemoryBank by app_name in addition to user_id

petrmarinec:
## Description of Change

`memory/vertexai` builds the Vertex AI MemoryBank scope from **only** `user_id`, on both the write and read paths:

```go
func createUserScope(userID string) map[string]string {
    return map[string]string{"user_id": userID}
}
```

- `addSession` (GenerateMemories) stores memories with `Scope: createUserScope(s.UserID())`.
- `searchMemory` (RetrieveMemories) queries with `Scope: createUserScope(req.UserID)`.

`app_name` is never part of the scope, so when more than one application shares a MemoryBank (agent engine) for the same user population, both emit a byte-identical scope for the same user: application B calling `SearchMemory` for user `U` retrieves memories that application A generated for user `U`. Retrieved memories are facts extracted from conversations.

To be precise about what this is: on the REST path the app name comes from the request body and the handler does not authenticate the caller, so `app_name` is a **namespacing key, not an authenticated confidentiality boundary**. Adding it partitions memories per application (matching the rest of the framework) and removes the effortless cross-read that exists today, but it is a namespacing convention rather than a tenant-isolation guarantee.

This is inconsistent with the rest of the framework, which partitions memory by `(app_name, user_id)`:

- The in-memory memory service keys its store on `key{appName, userID}` (`memory/inmemory.go`).
- `memory.SearchRequest` already carries `AppName`, and sessions expose `AppName()` — the value is available at both call sites but unused here.
- The adk-python MemoryBank implementation includes `app_name` in the scope on both the generate and retrieve calls.

## Solution

Include `app_name` in the scope on both paths, matching the in-memory service and adk-python:

```go
func createScope(appName, userID string) map[string]string {
    return map[string]string{"app_name": appName, "user_id": userID}
}
```

`addSession` now uses `createScope(s.AppName(), s.UserID())` and `searchMemory` uses `createScope(req.AppName, req.UserID)`.

## Compatibility (breaking change — needs a maintainer posture call)

Memory Bank matches the entire scope map exactly on retrieval (`RetrieveMemoriesRequest.Scope`: a memory must have exactly the same scope to be retrieved). Memories written before this change carry the one-key `{user_id}` scope, so after upgrading they are **not** returned to the new `{app_name, user_id}` retrieval. This package has shipped since v1.3.0, so everything those deployments wrote becomes unreachable on upgrade — silently: `searchMemory` returns an empty response with a nil error, and this client exposes no list/delete call, so the memories are orphaned rather than deleted.

This is unavoidable for any correct fix, because the partition key was never recorded on the old writes. I have deliberately **not** added an unconditional `{user_id}`-only fallback on empty reads: that would reinstate the original shared cross-read for all pre-upgrade data. If a legacy-read path is wanted it must be opt-in and off by default. Whether to accept the break with a release note or add an opt-in migration is a compatibility-posture decision for the maintainers rather than something to settle in this diff.

## Testing Plan

- `TestScopeIncludesAppNameOnBothCallSites` stands up an in-process `MemoryBankService` fake (bufconn, mirroring `session/vertexai`) that records the scope of each request, and pins both production call sites: `GenerateMemories` emits a different scope for two applications sharing one user, and `RetrieveMemories` queries the same `(app_name, user_id)` scope the writer used. Reverting either call site to a `user_id`-only scope fails this test (the previous helper-only tests did not).
- `TestCreateScope` retains the direct check that the scope carries exactly `app_name` and `user_id`. The earlier `TestCreateScopeDistinguishesApps` is removed as subsumed by it.


karolpiotrowicz (changes requested):
The scoping bug is real and this is the right fix for it. I drove the public path — `runner.Run` → `InvocationContext.Memory().SearchMemory` — against a local fake MemoryBank that records what goes on the wire, and on `main` two applications sharing one agent engine emit a byte-identical scope for the same user, so one application's search returns the other's memories. That result needs no assumption about how Memory Bank matches scopes: the two requests are literally the same bytes. With this PR applied the two applications emit distinct scopes and the cross-read stops. The `{app_name, user_id}` pair also matches what [`memory/inmemory.go`](https://github.com/google/adk-go/blob/058933d583478dcae7dd5d870f366b51e3b00ae1/memory/inmemory.go#L36-L38) already keys on and what adk-python does at every one of its six scope sites, including [retrieve](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/memory/vertex_ai_memory_bank_service.py#L550-L553).

Two things need settling before this can merge, and the first one is really a question for the maintainers rather than something to fix in the diff.

## Existing memories stop being readable, silently

Memory Bank retrieval matches the entire scope map exactly. The generated client this repo depends on says so on `RetrieveMemoriesRequest.Scope` in `cloud.google.com/go/aiplatform@v1.121.0` — "A memory must have exactly the same scope (`Memory.scope`) as the scope provided here to be retrieved (same keys and values)" — and the [Memory Bank documentation](https://docs.cloud.google.com/gemini-enterprise-agent-platform/scale/memory-bank/fetch-memories) repeats it: "Only memories that have the exact same scope (independent of order) as the retrieval request are returned."

So a memory written as `{user_id: U}` is not returned to a request for `{app_name: A, user_id: U}`. This package has shipped in every release since v1.3.0 — nine releases over about four months — and everything those deployments wrote carries the one-key scope. After an upgrade, none of it comes back. The failure is invisible from the outside: [`searchMemory`](https://github.com/google/adk-go/blob/e1e302d98c6723311f1d0b9f2e6b0ece6c96f481/memory/vertexai/vertexai_client.go#L116) returns an empty `SearchResponse` with a nil error, the package logs nothing, and `ServiceConfig` has no way to ask for the old behavior. An operator sees an agent that has forgotten its users and cannot distinguish that from a genuine miss. The memories are not deleted, but this client has no list or delete call, so they are unreachable.

I want to be clear that this is not a mistake in the change. Any correct fix orphans data written before the partition key existed, because the key was never captured. What the PR is missing is the acknowledgement — and the closing line of the description currently says the opposite: "Existing single-application behavior is unchanged (the added scope key only narrows retrieval)." That sentence is about exactly the population whose memories go away, and it is the reason this could land unnoticed. Please drop or correct it.

One thing worth ruling out early, because it looks like the obvious escape hatch and is not: an unconditional fallback that retries `{user_id}` when the scoped read comes back empty would return the legacy shared bucket to every application, which reinstates the original cross-read for all pre-upgrade data. If a legacy read is wanted it has to be opt-in and off by default. Whether to do that, or accept the break with a release note, is a call about the project's compatibility posture rather than something to decide in this diff — which is why I think a maintainer needs to weigh in before it merges.

## The tests do not pin either call site

`TestCreateScope` and `TestCreateScopeDistinguishesApps` both call `createScope` directly, so they assert that a two-entry map literal contains its own arguments. The two lines that actually changed are untested: change [line 116](https://github.com/google/adk-go/blob/e1e302d98c6723311f1d0b9f2e6b0ece6c96f481/memory/vertexai/vertexai_client.go#L116) to `createScope("", req.UserID)` — reintroducing the read-path half of the bug — and `go test ./memory/vertexai/` still passes. Same for [line 86](https://github.com/google/adk-go/blob/e1e302d98c6723311f1d0b9f2e6b0ece6c96f481/memory/vertexai/vertexai_client.go#L86) on the write path. Coverage agrees: `createScope` is at 100%, `addSession` and `searchMemory` are at 0%.

This is worth doing properly rather than dropping, because the isolation guarantee lives at those two lines and nothing currently guards it. It is achievable without touching any production code. `vertexai_client_test.go` is already in `package vertexai`, so `&vertexAIClient{...}` can be constructed directly against an in-process `grpc.Server` registering a stub `MemoryBankServiceServer` that records `req.Scope`, dialled with `option.WithEndpoint` plus `option.WithoutAuthentication` and insecure transport credentials. `aiplatform`, `option` and `grpc` are all already dependencies. I wrote one that way while reviewing this and it works: the fake server and the client wiring came to 77 lines, plus the assertions. The assertion to make is that two sessions differing only in `AppName` produce two different recorded scopes on `GenerateMemories`, and that the `RetrieveMemories` scope equals the writer's — which fails on `main` and passes with this change.

## Smaller things

`TestCreateScopeDistinguishesApps` is subsumed by `TestCreateScope`: any change to `createScope` that makes the second one fail already fails the first one's exact-value assertion. It could go, in favour of the call-site test.

The new comment says the change "matches the (app_name, user_id) keying used by … the adk-python / adk-js MemoryBank implementations". adk-python checks out exactly. I have not verified the adk-js half, so either cite it or drop that clause.

The same comment describes the change as stopping applications from reading each other's memories, which reads as a security boundary. On the REST path the app name comes from the request body and the handler does not authenticate the caller, so it is closer to a namespacing convention — a real and worthwhile one, and strictly better than `main`, where the cross-read needs no effort at all. Wording it as isolation rather than as a confidentiality guarantee would set the right expectation.

Two things I ran into that belong in their own issues rather than here, noted only so they do not get lost.

- **App names are not canonicalized.** `session/vertexai` accepts an app name as either a bare numeric engine ID or a full `projects/…/reasoningEngines/<id>` resource name and stores whichever the caller passed. Once the app name is part of the memory scope, the same engine addressed both ways writes into two partitions.
- **The launchers derive `AppName` inconsistently.** The console launcher hard-codes `"console_app"` with a comment saying it is not important, a2a uses the root agent's name, and the REST controller uses the request field. That already partitions memory differently per front-end with the in-memory service today.

Finally, on process rather than code: nothing in the tracker asks for this change and there is no linked issue, so no maintainer has yet formed a view on the compatibility question above. Given the change is right and has been sitting for a while, it would be good to get that decision made so it can land.

karolpiotrowicz:
@wolo-lab flagging this one for a posture call rather than a code review.

The tenant leak it fixes is real and reproducible — two apps sharing an agent engine currently emit an identical memory scope for the same user, so one app can read the other's memories. The problem is the upgrade path: Memory Bank matches the whole scope map exactly, so adding `app_name` makes every memory written by the nine releases this package has shipped in unreachable, silently and with no opt-out. That is unavoidable for any correct fix, not something the author chose. What needs deciding is the compatibility posture and whether we want a migration or a documented break.

petrmarinec:
addressed the review:

- fixed the description: dropped the "behavior is unchanged" line, added a compatibility section (pre-fix {user_id} memories won't be returned after the change; no {user_id} fallback since it'd reinstate the cross-read — break vs migration is your call).
- added a bufconn MemoryBankService fake that pins both call sites (reverting either to user_id-only now fails); dropped the subsumed scope test.
- reworded createScope as namespacing, not a confidentiality boundary; dropped the adk-js citation.

the red "GitHub Actions Scan" was zizmor on .github/workflows/* (untouched here), already fixed on main — updated the branch with main, which clears it.

PR #1179, opened 2026-07-20
test(auth): offline end-to-end auth for mcptoolset and remoteagent

wolo-lab:
Config.Auth (mcptoolset) and A2AConfig.Auth (remoteagent) apply per-request credentials via a transport RoundTripper / client interceptor, but had no end-to-end coverage proving the credential actually reaches the wire for the security schemes a server or agent card can declare.

Add fully offline demonstrations (no API key, no external services):

- mcptoolset: a local streamable-HTTP MCP server requires a credential on every request (tool listing and calling alike). Verifies Config.Auth attaches it for a bearer token, a header API key, and 2-legged (client_credentials) OAuth; without a provider the server rejects with 401. Mirrors adk-python's contributing/samples/mcp/mcp_toolset_auth/oauth_mcp_server.py.
- remoteagent: a local a2asrv records request headers; asserts A2AConfig.Auth attaches the credential per the card's declared scheme for both HTTP/OAuth2 (Authorization: Bearer) and API-key (named header) schemes.

### DIFF
--- a/agent/remoteagent/v2/auth_test.go
+++ b/agent/remoteagent/v2/auth_test.go
@@ -94,36 +94,73 @@ func TestNewA2AAuthWithClientProviderIsError(t *testing.T) {
 }
 
 func TestRemoteAgent_AuthAttachesBearerHeader(t *testing.T) {
+	// The interceptor only attaches auth when the card declares a matching
+	// security requirement, so the card must carry one.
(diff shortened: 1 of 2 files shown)


wolo-lab:
Auth related work is suspended. Reopen once it's done.

PR #1180, opened 2026-07-20
docs(examples): add mcpauth example for auth providers with MCP toolset

wolo-lab:
There was no runnable example showing how to wire the ADK auth subsystem into an MCP toolset, or how the different built-in credential providers map onto mcptoolset.Config.Auth.

examples/mcpauth is a buildable template that selects a provider via AUTH_MODE: GCP (per-user; supports interactive 3-legged consent), header API key, static bearer, 2-legged OAuth (client credentials), ADC, and service account. Comments walk through both the non-interactive path (applied transparently by the transport RoundTripper) and the interactive consent flow (adk_request_credential, resumed once the user consents). The live consent flow requires a real MCP server and GCP auth resource.

### DIFF
--- a/examples/mcpauth/main.go
+++ b/examples/mcpauth/main.go
@@ -1,0 +1,182 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


wolo-lab:
To reopen once auth is ready.

PR #767, opened 2026-04-28
fix: preserve unrelated function events

gustafvh:
### Link to Issue or Description of Change

- Closes: google/adk-go#761

### Problem

`rearrangeEventsForLatestFunctionResponse` rebuilds history when the latest event is a function response. After finding the matching function-call event, it collects related response events to merge, then returns only the history up to that matching call plus the merged response.

That can drop unrelated function-call or function-response events that happened between the matching call and the latest response. For long-running tool flows, this loses valid tool history that should still be replayed.

### Solution

When scanning intermediate events between the matching function-call event and the latest response:

- preserve intermediate function-call events in the result history
- preserve unrelated intermediate function-response events in the result history
- merge only response events related to the matching call event

This keeps unrelated tool history intact while still consolidating the response sequence that triggered rearrangement.

### Testing Plan

- [x] Added `Rearrangement preserves unrelated function events` covering an interleaved long-running response sequence with an unrelated tool call/response before the final response.
- [x] Ran focused unit tests locally.

```sh
go test ./internal/llminternal
```

### Alignment with adk-python

This preserves the shared history invariant that function-response rearrangement should repair pairing for the target long-running response without dropping unrelated valid events from the conversation history.

PR #1182, opened 2026-07-20
fix: keep latest function response as the final content when rearranging history

### LINKED ISSUE #1181: Late long-running tool completion is buried mid-history by response rearrangement — models answer the stale exchange left at the tail
## 🔴 Required Information

**Describe the Bug:**

When a long-running (async) tool's final `FunctionResponse` arrives as the *latest* session event — after other, unrelated tool exchanges have happened — the request contents sent to the model end on a **stale** exchange instead of the completion.

`buildContentsDefault` runs two rearrangement passes back-to-back:

1. `rearrangeEventsForLatestFunctionResponse` handles the late response correctly: it finds the matching call, merges intermediate + final responses, **preserves** the intervening unrelated tool events (the deliberate fix from #767 for #761), and appends the merged response **last**. At this point the contents are right.
2. `rearrangeEventsForFunctionResponsesInHistory` then rebuilds the list attaching *every* response event immediately after its matching call — **including the merged response pass 1 just placed at the tail**. The completion is dragged back next to its (much earlier) call, and whatever unrelated exchange happened last is left as the final content.

Models act on the final contents, so they answer the stale exchange and never see the completion.

**Steps to Reproduce:**

1. Build a session history where a long-running tool's final response is the latest event, with an unrelated tool exchange in between (this is exactly the event sequence of the existing test case `"Rearrangement preserves unrelated function events"` in `contents_processor_test.go`):
   ```
   1. user:  text "Run long process and search"
   2. model: FunctionCall  A (long_running_tool, id=long_call_123)
   3. user:  FunctionResponse A-intermediate ("processing")
   4. model: FunctionCall  B (search_tool, id=call_123)
   5. user:  FunctionResponse B ("results...")
   6. user:  FunctionResponse A-final ("completed")   ← latest event
   ```
2. Run the contents pipeline (`ContentsRequestProcessor` / `buildContentsDefault`) over it.
3. Inspect the resulting `LLMRequest.Contents` ordering.

**Expected Behavior:**

The latest function response remains the **final** content. The unrelated B exchange stays preserved (the property #767 established) — *before* the completion rather than after it:

```
user text · B-call · B-resp · A-call · A-final(merged)
```

**Observed Behavior:**

The completion is buried mid-history and the stale B exchange is the tail:

```
user text · A-call · A-final(merged) · B-call · B-resp   ← model answers B-resp
```

Unit-level demonstration on current `main` (cmp diff, `-want +got`, from a regression test we can contribute):

```
  []*genai.Content{
  	&{... "Plan how to add feature Q" ..., Role: "user"},
+ 	&{Parts: {&{FunctionCall: plan_tool ...}}, Role: "model"},     ← completion buried here
+ 	&{Parts: {&{FunctionResponse: plan_tool completed ...}}, Role: "user"},
  	&{Parts: {&{FunctionCall: status_tool ...}}, Role: "model"},
  	&{Parts: {&{FunctionResponse: status "running, 21s" ...}}, Role: "user"},
- 	&{Parts: {&{FunctionCall: plan_tool ...}}, Role: "model"},     ← where it must be
- 	&{Parts: {&{FunctionResponse: plan_tool completed ...}}, Role: "user"},
  }
```

Real-world impact: in our production multi-agent supervisor this made **three different models** consistently blind to long-running tool completions — each re-summarized the stale status exchange at the tail; one literally repeated a "21 seconds elapsed, still running" status while the completed 5.5 KB result sat merged mid-history, immediately after a call issued minutes earlier.

**Environment Details:**

 - ADK Library Version: v1.5.0, v2.0.0, and current `main` (the three functions involved are byte-identical across all three)
 - OS: macOS (logic bug — OS-independent)
 - Go Version: go 1.25

**Model Information:**
 - Which model is being used: gemini-3.1-pro (native path); also reproduced with DeepSeek V3.2 and Kimi K2 through an OpenAI-compatible path — the contents are already wrong before any provider adapter runs, so the bug is model-independent.

---

## 🟡 Optional Information

**Regression:**

Partially — the *burying* behavior arrived with #767 (2026-05-28). Before #767, the same flow **dropped** the unrelated intervening events entirely (that data loss was #761). So no prior version handled this history correctly: pre-#767 lost events, post-#767 buries the completion. The fix proposed below is the first state with both properties (nothing dropped *and* the completion stays last).

**Logs:**

```text
(see the cmp diff under Observed Behavior; happy to attach full pre/post contents dumps)
```

Add this row to `TestContentsRequestProcessor_Rearrange` in `internal/llminternal/contents_processor_test.go` on unpatched `main` — it fails with the diff shown above (fixtures follow the file's existing style):

```go
{
	name: "Late async completion stays the final content",
	events: []*session.Event{
		{Author: "user", LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("Plan how to add feature Q", "user")}},
		{Author: agentName, LLMResponse: model.LLMResponse{Content: NewContentFromFunctionCall(fcAsyncPlan, "model")}},
		{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frAsyncPlanAck, "user")}},
		{Author: agentName, LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("I dispatched the planning task.", "model")}},
		{Author: "user", LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("What is the status?", "user")}},
		{Author: agentName, LLMResponse: model.LLMResponse{Content: NewContentFromFunctionCall(fcAsyncStatus, "model")}},
		{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frAsyncStatus, "user")}},
		{Author: agentName, LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("Still running.", "model")}},
		{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frAsyncPlanFinal, "user")}},
	},
	want: []*genai.Content{
		genai.NewContentFromText("Plan how to add feature Q", "user"),
		NewContentFromFunctionCall(fcAsyncStatus, "model"),
		NewContentFromFunctionResponse(frAsyncStatus, "user"),
		NewContentFromFunctionCall(fcAsyncPlan, "model"),
		NewContentFromFunctionResponse(frAsyncPlanFinal, "user"),
	},
},
```

 - Always (100%) — deterministic whenever a function response arrives as the latest event with an unrelated tool exchange between it and its call.

**Additional Context:**

*Relationship to existing issues:*
- #761 / #767 — #767 correctly made pass 1 preserve intervening tool events; those preserved events are exactly what pass 2 then promotes over the fresh tail. This report disputes only the final placement, not the preservation.
- #760 (and PRs #766 / #880) — the orphan-response error path in pass 1. Disjoint code path; the fix below does not touch it.

*adk-python parity:* adk-python cannot reach this state — its `_rearrange_events_for_latest_function_response` truncates history (`result_events = events[:function_call_event_idx + 1]` plus the merged response), so its tail is *always* the merged latest response, at the cost of dropping the unrelated exchange (the very behavior #767 fixed in Go). On the observable property ("what is the final content when a long-running tool completes late?") Go currently *disagrees* with adk-python; the proposed fix re-aligns the observable tail while keeping Go's superior no-drop behavior.

*Proposed fix (PR to follow):* keep pass 2's adjacency invariant (every call event immediately followed by its consolidated response) and add one rule — **the call/response pair answered by the final event is emitted last**. One-function change in `rearrangeEventsForFunctionResponsesInHistory`; provably a no-op whenever the last event is not a function response; one existing test expectation updates (the #767 case, with the reasoning above), four regression cases added. Live-verified end-to-end with gemini-3.1-pro against the patched module.

dmora:

- Closes: #1181
- Related: #767, #761

**Problem:**

When a long-running tool's final response arrives as the latest session event,
`rearrangeEventsForLatestFunctionResponse` correctly consolidates it as the last content — but
`rearrangeEventsForFunctionResponsesInHistory` then re-attaches it immediately after its (much
earlier) call, leaving a stale unrelated exchange as the final content. Models act on the final
contents: in a production multi-agent supervisor, three different models (gemini-3.1-pro native;
DeepSeek V3.2 and Kimi K2 via an OpenAI-compatible path) were consistently blind to long-running
completions, re-answering the stale status exchange left at the tail instead. Full mechanism and
minimal repro in the linked issue.

Adjacency — the invariant pass 2 enforces — is necessary but not sufficient: the **latest
response must also remain the final content**.

**Solution:**

Pass 2 keeps its full adjacency invariant and adds one rule: **the call/response pair answered by
the final event is emitted last.** While rebuilding, if a call event's response set includes the
final event, the pair (call + consolidated response) is routed to a `tailPair` slice appended at
the end. ~25-line one-function change, no new imports.

Three guarantees, all test-covered:

1. **No events are dropped** — #761/#767's preservation property is intact; the unrelated exchange
   moves before the completion instead of after it.
2. **Every call is still immediately followed by its consolidated response** (parallel calls move
   with their call event as one unit, text parts included) — no new turn-structure shape is
   introduced.
3. **Pure no-op whenever the last event is not a function response**: the relocation gate keys on
   the final event's index appearing in a response set, which only ever holds response-event
   indices. All non-async histories produce byte-identical output (both "…in history (non-final
   event)" cases and all other existing rows pass unchanged).

One existing expectation updates: `"Rearrangement preserves unrelated function events"` (added by
#767). Its purpose — the unrelated `fcBasic` exchange must survive — still holds and is still
asserted; the previous `want` additionally encoded the incidental mid-history placement of the
merged completion, which is the bug this PR fixes. The updated `want` keeps the preserved exchange
*before* the completed pair so the freshest response is never buried behind a stale one (this also
matches adk-python's observable tail for the same flow — python truncates intervening events in
its pass 1, so its final content is always the merged latest response).

### Testing Plan

**Unit Tests:**

4 new rows in the existing `TestContentsRequestProcessor_Rearrange` table: `Late async completion
stays the final content` (the minimal repro), `Late async completion stays final with mixed text
and call parts` (production Gemini shape — text parts ride the call events; the tail pair moves as
a unit), `Adjacent final response after earlier exchange keeps order` (identity guard), `Late
parallel completion stays final past unrelated exchange` (parallel pair moves as one unit). Before
the fix, the three async rows + the updated #767 row fail with the completion buried mid-history;
after the fix all 17 rows pass.

```
go build -mod=readonly ./...                                  # ok
go test -race -mod=readonly -count=1 -shuffle=on ./...        # ok — 69 packages, 0 failures
go mod tidy -diff                                             # clean
golangci-lint run   (v2.3.1, repo config)                     # 0 issues
golangci-lint fmt --diff                                      # clean
```

Ran our production multi-agent supervisor against this branch (`go.mod` replace onto the patched
module), with tool completions delivered through the FunctionResponse path. Setup: dispatch a
long-running plan tool asynchronously; while it runs, ask "What's the status?" so a status/wait
exchange lands after the dispatch (the exact shape that triggers the bug); let the completion
arrive as a late `FunctionResponse` reusing the original call id.

Observed with gemini-3.1-pro, exercising the vulnerable shape twice in one session (the plan
completion, then a follow-up inspect tool completing the same way):

- Every completion-turn request was accepted by the Gemini API — no `INVALID_ARGUMENT` /
  turn-structure rejections (guarantee 2 verified live, not just analytically).
- The model acted on the fresh tail both times: it summarized the completed result and
  autonomously continued the pipeline off it — exactly the behavior the bug suppressed (on
  unpatched adk-go, the same scenario ends with the model re-answering the stale status).

Event-sequence excerpt (late completion as final event → next model turn acts on it):

```text
user|RESP:plan id=adk-e6beb946        ← late completion, original call id, last event
model|TEXT: "I've completed the plan and have the result. …"   ← model sees it
model|RESP:inspect (DISPATCHED)       ← pipeline continues off the result
```

### Checklist

- [x] Any dependent changes have been merged and published in downstream modules. (N/A — no dependent changes)

### Additional context

- The merged-responses branch writing to `tailPair` is defensive consistency: through the standard
  two-pass pipeline it is unreachable (`rearrangeEventsForLatestFunctionResponse` pre-merges the
  final pair's responses into a single event, so the tail pair always takes the single-response
  branch); it matters only if this function is ever driven standalone.
- Fallback considered and rejected: gating the relocation on `LongRunningToolIDs` — pass 2 has
  never read that field, the #767 fixture doesn't set it (the suite would assert opposite outcomes
  for identical event shapes), and unmarked async flows (A2A, session resume, injected events)
  would stay broken. Can switch to the gated variant if you prefer the narrower blast radius.
- Same-file neighbors: #766 / #880 (competing fixes for #760's orphan-response error path in
  pass 1) and #1044 (compaction filtering) — function-disjoint from this change; test-file
  conflicts are additive. Happy to rebase in any order.
- The bug exists identically on the `v1` line (the three functions are byte-identical
  v1.5.0 ↔ main). Happy to open a `v1` backport if you still cut 1.x releases — maintainers' call.

### DIFF
--- a/internal/llminternal/contents_processor.go
+++ b/internal/llminternal/contents_processor.go
@@ -355,6 +355,11 @@ SearchLoop: // A label to allow breaking out of the nested loop
 // responses may not have originally been consecutive. It preserves all
 // non-tool-call events (like user messages) in their original order.
 //
+// The call/response pair answered by the final event in the history is
+// emitted last: the latest function response is the newest information in
+// the session, and models act on the final contents, so it must not be
+// re-attached mid-history behind older (now stale) exchanges.
+//
 // It returns a new, correctly ordered slice of events or an error if the
 // history is malformed (e.g., a response is found without a corresponding call).
 func rearrangeEventsForFunctionResponsesInHistory(events []*session.Event) ([]*session.Event, error) {
@@ -374,8 +379,16 @@ func rearrangeEventsForFunctionResponsesInHistory(events []*session.Event) ([]*s
 		}
 	}
 
+	// The last event is the model's freshest information. When it is a
+	// function response (e.g. a long running tool completed after later,
+	// unrelated exchanges), its call/response pair must remain the last
+	// exchange in the rebuilt history.
+	lastEventIndex := len(events) - 1
+
 	// Rebuild the event list
 	var resultEvents []*session.Event
+	// Call/response pair answered by the final event; appended last.
+	var tailPair []*session.Event
 
 	for _, event := range events {
 		// If the event contains responses, skip it. It will be handled
@@ -389,9 +402,6 @@ func rearrangeEventsForFunctionResponsesInHistory(events []*session.Event) ([]*s
 			// This is a regular event (e.g., user message). Just append it.
 			resultEvents = append(resultEvents, event)
 		} else {
-			// This is a function call event, append it and search for responses
-			resultEvents = append(resultEvents, event)
-
 			// Find the unique indices of all corresponding response events.
 			// Using a map[int]struct{} as a set.
 			responseEventIndicesSet := make(map[int]struct{})
@@ -401,6 +411,15 @@ func rearrangeEventsForFunctionResponsesInHistory(events []*session.Event) ([]*s
 				}
 			}
 
+			// This is a function call event: append it, then its responses.
+			// If the pair is answered by the final event, route the whole
+			// pair (call + consolidated response) to the tail instead.
+			dst := &resultEvents
+			if _, found := responseEventIndicesSet[lastEventIndex]; found {
+				dst = &tailPair
+			}
+			*dst = append(*dst, event)
+
 			// If no responses were found for any calls in this event, continue.
 			if len(responseEventIndicesSet) == 0 {
 				continue
@@ -409,7 +428,7 @@ func rearrangeEventsForFunctionResponsesInHistory(events []*session.Event) ([]*s
 			// If there's only one unique response event, append it directly.
 			if len(responseEventIndicesSet) == 1 {
 				for index := range responseEventIndicesSet { // A trick to get the single key
-					resultEvents = append(resultEvents, events[index])
+					*dst = append(*dst, events[index])
 				}
 			} else {
 				// Multiple response events exist for that function call so we merge them.
@@ -431,12 +450,12 @@ func rearrangeEventsForFunctionResponsesInHistory(events []*session.Event) ([]*s
 				if err != nil {
 					return nil, fmt.Errorf("failed to merge response events: %w", err)
 				}
-				resultEvents = append(resultEvents, mergedEvent)
+				*dst = append(*dst, mergedEvent)
 			}
 		}
 	}
 
-	return resultEvents, nil
+	return append(resultEvents, tailPair...), nil
 }
 
 // mergeFunctionResponseEvents merges a list of function response events into one.
--- a/internal/llminternal/contents_processor_test.go
+++ b/internal/llminternal/contents_processor_test.go
@@ -1096,6 +1096,44 @@ func TestContentsRequestProcessor_Rearrange(t *testing.T) {
 		Response: map[string]any{"confirmed": false},
 	}
 
+	// Async Call/Responses: a long-running call whose final response arrives
+	// after later, unrelated exchanges (the async / late-completion flow).
+	fcAsyncPlan := &genai.FunctionCall{
+		ID:   "async_plan_call",
+		Name: "plan_tool",
+		Args: map[string]any{"task": "feature"},
+	}
+	frAsyncPlanAck := &genai.FunctionResponse{
+		ID:       "async_plan_call",
+		Name:     "plan_tool",
+		Response: map[string]any{"status": "dispatched"},
+	}
+	frAsyncPlanFinal := &genai.FunctionResponse{
+		ID:       "async_plan_call",
+		Name:     "plan_tool",
+		Response: map[string]any{"status": "completed", "result": "plan ready"},
+	}
+	fcAsyncStatus := &genai.FunctionCall{
+		ID:   "async_status_call",
+		Name: "status_tool",
+		Args: map[string]any{"target": "plan"},
+	}
+	frAsyncStatus := &genai.FunctionResponse{
+		ID:       "async_status_call",
+		Name:     "status_tool",
+		Response: map[string]any{"status": "running", "elapsed": "21s"},
+	}
+	fcAsyncWait := &genai.FunctionCall{
+		ID:   "async_wait_call",
+		Name: "wait_for",
+		Args: map[string]any{"target": "plan"},
+	}
+	frAsyncWait := &genai.FunctionResponse{
+		ID:       "async_wait_call",
+		Name:     "wait_for",
+		Response: map[string]any{"status": "timeout"},
+	}
+
 	// Error Call/Response
 	frOrphaned := &genai.FunctionResponse{
 		ID:       "no_matching_call",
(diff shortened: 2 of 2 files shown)


karolpiotrowicz:
The reordering holds up under differential testing against the pre-change implementation: the no-op property when the final event isn't a function response, the no-drop/no-duplicate property, and call/response adjacency all survive. The tail also lines up with adk-python on the #767 fixture — python reaches the same tail by truncating in pass 1 (`contents.py:282`) where adk-go preserves, so the pass-2 divergence here is what keeps the two observably aligned. Worth saying that explicitly in the description, since AGENTS.md makes python the behavioral reference.

One thing worth adding before merge: **a test row for a call that never gets a response.** None of the 17 rows in `TestContentsRequestProcessor_Rearrange` has one, and the shape is reachable — `base_flow.go:1164-1170` returns before creating the response event for a deferring `ResponseDeferrer` or an `IsLongRunning()` tool, which `handle_function_calls_async_test.go:282,290,298` pins with `wantFREventIndex: -1`. On that shape the relocation changes the output:

```
events: user "do X" | model CALL:lro | user RESP:lro(pending)
        | model CALL:pending   (never answered)
        | user RESP:lro(completed)    <- final

before: user "do X" | CALL:lro | RESP:lro | CALL:pending | "Continue processing…"
after:  user "do X" | CALL:pending | CALL:lro | RESP:lro
```

Two consecutive `model` contents, and the continuation turn at `contents_processor.go:66-70` no longer fires. Both look fine to me — the API accepts consecutive model turns, and ending on a real tool result is exactly when that nudge shouldn't fire (adk-python emits none here either). But it's a behavior change on an untested reachable shape, and a row would make it a decision rather than a side effect.

Nit: `contents_processor.go:358-361` says "The call/response pair" — if two call events are answered by the same final event, both route to `tailPair`. Suggest "pair(s)". The duplicate emission itself is pre-existing.

On the "unreachable" note for the merged-responses branch: it is reachable through pass 1's early return at `:245-253`, but only when a call ID repeats across two call events, so for unique-ID histories the claim holds. Narrowing the wording would make it airtight.

jjsasha63:
Closing this as stale — there has been no response to the review feedback from
27 July, and no activity from the author since 20 July.

The change itself looked sound and CI was green. What was outstanding:

1. A test row for a call that never receives a response — the relocation
   changes the output on that shape and no existing row covers it.
2. The "The call/response pair" comment at `contents_processor.go:358-361`
   should read "pair(s)", since two calls answered by the same final event both
   route to `tailPair`.
3. Narrowing the "unreachable" note on the merged-responses branch, which is
   reachable when a call ID repeats across two call events.

@dmora if you would like to pick this back up, address those three points and
reopen — or comment here and we will reopen it for you. Thanks for the very
thorough report and repro in #1181.

PR #1183, opened 2026-07-21
fix(workflow): resume LlmAgent nodes paused on a long-running tool request

wolo-lab:
## Problem

An LlmAgent running as an `AgentNode` inside a graph can pause on a tool's long-running request — interactive OAuth consent (`adk_request_credential`) or tool confirmation (`adk_request_confirmation`) — exactly as it does as a root agent. The pause already works: the scheduler parks the node on the generic `Event.LongRunningToolIDs` signal. Resuming did not.

- `workflowagent.detectResume` only recognised `adk_request_input` responses, matching the `FunctionResponse` by function name, so a consent or confirmation reply fell through to a fresh `Workflow.Run` instead of a targeted `Resume` — which, in a multi-node graph, re-runs already-completed upstream nodes.
- Even when resumed, an `AgentNode` did not opt into re-entry, so `Workflow.Resume` took the handoff path — mark the node completed, treat the raw response as its output — and never re-ran the agent to finish the pending tool call.
- On re-entry the node re-fed its original input, which makes a single_turn or task LlmAgent re-call the still-pending tool and pause again, a loop.

Parity note: adk-python's `build_node` defaults an LlmAgent node to `rerun_on_resume=True`, and an LlmAgent node there resumes on the same `LongRunningToolIDs` signal, re-running from history. It sets the single_turn placement default on a `clone()` of the agent rather than on the agent itself, which is the same separation `AgentNode.Run` reaches by binding the resolved mode to the invocation.

## Summary

- `detectResume` matches `FunctionResponse`s to open interrupts by **ID**, not by function name — the engine's pause is already name-agnostic, riding on `Event.LongRunningToolIDs`. A stale or mistargeted reply still reconstructs state and surfaces `ErrNothingToResume` rather than silently starting a fresh `Run`.
- `AgentNode` defaults `RerunOnResume=true` **for LlmAgent nodes**, gated on `isLlmAgent` to match `runner.newAgentNode` and adk-python's `build_node`. Other agent kinds keep the engine default of handoff, and an explicit caller value always wins. The single_turn placement default is left to `AgentNode.Run`, which resolves it per activation and binds it to the invocation, so wrapping a shared agent in a node never mutates it.
- `AgentNode.Run` drops the node input on resume so the agent continues from history. Resume detection needs **two** signals, because neither is sufficient alone:
  - the scheduler's per-activation resume flag (`IsResumeActivation`), since the engine injects resolved interrupt responses only on the `Resume` path. A history scan cannot express this — history never un-answers an interrupt, so it latches and starves every later activation of the node in the same invocation, whether from a loop-back edge, a `RetryConfig` retry or a `ParallelWorker` item.
  - proof the pause was raised **under this node**. The flag alone over-fires, because a dynamic child inherits the resume context of the ancestor that paused, so a freshly delegated child would lose its input.

  Interrupts are attributed the way rehydration attributes them, by `NodeInfo.Path` with `Author` as the fallback, so a pause raised by a sub-agent the node's agent delegated to counts as the node's own — matching the engine, which parks and re-enters the node on exactly those interrupts.
- Node config defaults are consolidated into `applyAgentNodeDefaults`, mirroring `applyDynamicDefaults`.
- Every resume payload decodes through one `utils.UnwrapResponse`. The fresh-reply decoder was duplicated in `runner` and `workflowagent` and disagreed with the history decoder, so the same reply decoded differently depending on whether it was taken from the inbound turn or read back from history. It also unwrapped `"payload"` out of a multi-key map, silently dropping the flag from a tool confirmation's `{"confirmed": b, "payload": v}`, and did not understand `{"result": v}` — the shape adk-python's `_wrap_response` and the web frontend emit.
  The unification also widens where a string value is JSON-parsed. The old inbound decoders did it only under `response`, so `{"payload": "true"}` and `{"result": "123"}` now yield a bool and a number where the runner and workflowagent paths previously yielded the string. The history decoder already parsed all three, and making the two agree is the point of the change, but it is a change to the wire contract for anything that sent a JSON-looking string under `payload` or `result`.
- Replaying an answer a node has already acted on no longer re-runs it. Rehydration marks such a resume consumed on both arms that can produce one, including a node that still holds a second interrupt open, and `Resume` skips it. Answering the still-open interrupt continues to re-run the node.
- A settled reply echoed back alongside new user text no longer fails the turn. History never un-answers an interrupt, so a re-entry node's settled ID stays recognisable for the rest of the session. `detectResume` now distinguishes an ID the run can still act on from one it has already acted on, and a turn whose only matches are spent falls through to a fresh `Run` when it carries anything else. A bare replay with nothing else in it still reaches `Resume`, so its diagnostic is unchanged.
- A handoff node's genuine first resume stays in the filter. Such a node resolves its interrupt from history and rehydrates as `NodeCompleted`, which a status-only filter excluded, so a credential or confirmation reply to a non-LlmAgent `AgentNode` restarted the workflow instead of resuming it. The rule is handed to the dispatching packages through `internal/workflowstate` rather than widening the workflow package's API.
- A handoff node that raised several interrupts at once and has been answered on only some of them stays waiting instead of completing. Completing it dropped the unanswered requests and handed successors an output standing in for a decision nobody had made — for a rejected confirmation, one gating the very work it rejected. Mirrors adk-python's replay interceptor, which bubbles the unresolved IDs rather than completing the node.
- An answer counts as acted upon only once the node settles on it, by producing an output or pausing again on a new interrupt. Treating any emitted event as proof wedged the run when an activation persisted a tool result and then lost its model call: history never un-answers an interrupt, so every retry was skipped as a replay and the approved work could never complete.
- The YAML surface honours `rerun_on_resume` on an agent node, so a configured workflow can keep the handoff behaviour it had before the default changed.

### DIFF
--- a/agent/workflowagent/workflow.go
+++ b/agent/workflowagent/workflow.go
@@ -142,11 +142,25 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 	// tool's long-running request (adk_request_credential /
 	// adk_request_confirmation), not just the workflow's own
 	// adk_request_input, and the engine's pause is name-agnostic
-	// (Event.LongRunningToolIDs). A stale response matches nothing, so
-	// Workflow.Resume yields ErrNothingToResume instead of a silent fresh Run.
+	// (Event.LongRunningToolIDs).
+	//
+	// A response only routes this turn to Resume when it is actually aimed at
+	// this run: either its ID is an interrupt the run knows about, or it is
+	// explicitly an answer to the workflow's own input request. Otherwise the
+	// turn falls through to a fresh Run, as it did before ID-keyed matching —
+	// a turn may legitimately carry both text and an unrelated tool reply, and
+	// routing that to Resume would fail it with ErrNothingToResume and discard
+	// the text. Mirrors runner.buildResumeResponses, which filters the same way.
+	//
+	// A wrong-but-deliberate answer (right name, unknown ID) still reaches
+	// Resume, so the caller keeps the ErrNothingToResume diagnostic.
+	known := knownInterruptIDs(state)
 	responses := map[string]any{}
 	for _, fr := range frs {
 		if fr == nil || fr.ID == "" {
+			continue
+		}
+		if _, ok := known[fr.ID]; !ok && fr.Name != workflow.WorkflowInputFunctionCallName {
 			continue
 		}
 		responses[fr.ID] = utils.UnwrapResponse(fr.Response)
@@ -156,4 +170,32 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 	}
 
 	return responses, state, true, nil
+}
+
+// knownInterruptIDs collects the interrupt IDs of nodes the rehydrated run can
+// still act on — those waiting for an answer, and those a re-entry node is
+// about to be re-run with. A node that already settled is excluded, so a reply
+// to an interrupt this run has finished with no longer counts as a resume.
+//
+// ResumedInputs is included, not just Interrupts: the runner appends the reply
+// to the session before the agent runs, so on a genuine first resume the
+// interrupt is already resolved and rehydration hands it back on the node.
+func knownInterruptIDs(state *workflow.RunState) map[string]struct{} {
+	ids := map[string]struct{}{}
+	for _, ns := range state.Nodes {
+		if ns == nil || (ns.Status != workflow.NodeWaiting && ns.Status != workflow.NodePending) {
+			continue
+		}
+		for _, id := range ns.Interrupts {
+			if id != "" {
+				ids[id] = struct{}{}
+			}
+		}
+		for id := range ns.ResumedInputs {
+			if id != "" {
+				ids[id] = struct{}{}
+			}
+		}
+	}
+	return ids
 }
--- a/agent/workflowagent/workflow.go
+++ b/agent/workflowagent/workflow.go
@@ -15,11 +15,8 @@
 package workflowagent
 
 import (
-	"encoding/json"
 	"fmt"
 	"iter"
-
-	"google.golang.org/genai"
 
 	"google.golang.org/adk/v2/agent"
 	agentinternal "google.golang.org/adk/v2/internal/agent"
@@ -152,42 +149,11 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 		if fr == nil || fr.ID == "" {
 			continue
 		}
-		responses[fr.ID] = decodeResumeResponse(fr)
+		responses[fr.ID] = utils.UnwrapResponse(fr.Response)
 	}
 	if len(responses) == 0 {
 		return nil, nil, false, nil
 	}
 
 	return responses, state, true, nil
-}
-
-// decodeResumeResponse extracts the user-supplied payload from a
-// FunctionResponse that answers a paused node's interrupt — a workflow input
-// (adk_request_input), tool consent (adk_request_credential), or confirmation
-// (adk_request_confirmation). detectResume matches by interrupt ID, so this is
-// no longer specific to workflow input.
-//
-// Three accepted shapes, in priority order:
-//
-//  1. {"response": <value>}  — when value is a string, it is
-//     parsed as JSON and the result returned; if the string is
-//     not valid JSON it is returned verbatim. When value is any
-//     other type, it is returned as-is.
-//  2. {"payload": <any>}     — value returned verbatim.
-//  3. anything else           — the whole Response map is returned.
-func decodeResumeResponse(fr *genai.FunctionResponse) any {
-	if raw, ok := fr.Response["response"]; ok {
-		if s, isStr := raw.(string); isStr {
-			var decoded any
-			if err := json.Unmarshal([]byte(s), &decoded); err == nil {
-				return decoded
-			}
-			return s
-		}
-		return raw
-	}
-	if payload, ok := fr.Response["payload"]; ok {
-		return payload
-	}
-	return fr.Response
 }
--- a/agent/workflowagent/workflow.go
+++ b/agent/workflowagent/workflow.go
@@ -78,18 +78,17 @@ func New(cfg Config) (agent.Agent, error) {
 // workflowAgent is the wrapper that dispatches between
 // Workflow.Run (fresh turn) and Workflow.Resume (resume turn).
 // The dispatch decision is made by inspecting ctx.UserContent for
-// a FunctionResponse targeting a previously-emitted RequestInput.
-// The workflow's RunState lives in session.State, not on this
-// struct, so a single *workflowAgent safely services many
-// concurrent sessions.
+// a FunctionResponse that answers an interrupt this run can still
+// act on — see detectResume. The workflow's RunState lives in
+// session.State, not on this struct, so a single *workflowAgent
+// safely services many concurrent sessions.
 type workflowAgent struct {
 	workflow *workflow.Workflow
 }
 
 // run is the agent.Config.Run callback. It dispatches between
-// Workflow.Resume (when the inbound user content carries a
-// FunctionResponse to a previously-emitted RequestInput) and
-// Workflow.Run (every other turn).
+// Workflow.Resume (when the inbound user content answers one of
+// this run's open interrupts) and Workflow.Run (every other turn).
 func (a *workflowAgent) run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
 	return func(yield func(*session.Event, error) bool) {
 		responses, state, ok, err := a.detectResume(ctx)
--- a/agent/workflowagent/workflow.go
+++ b/agent/workflowagent/workflow.go
@@ -128,15 +128,6 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 		return nil, nil, false, nil
 	}
 
-	// A resume answers a paused node by matching a FunctionResponse to an
-	// open interrupt by ID. Match by ID, not by function name: besides the
-	// workflow's own adk_request_input, an LlmAgent node can pause on a
-	// tool's long-running request (adk_request_credential /
-	// adk_request_confirmation), and the engine's pause is name-agnostic
-	// (it rides on Event.LongRunningToolIDs). A stale or mistargeted
-	// response still reconstructs state; Workflow.Resume then yields
-	// ErrNothingToResume rather than silently starting a fresh Run.
-	//
 	// Scope rehydration to this run's invocation (the runner reuses the
 	// paused run's ID on resume) so a prior completed run in the same
 	// session does not leak in.
@@ -150,12 +141,18 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 		return nil, nil, false, nil
 	}
 
+	// Key by interrupt ID, not function name: an agent node pauses on a
+	// tool's long-running request (adk_request_credential /
+	// adk_request_confirmation), not just the workflow's own
+	// adk_request_input, and the engine's pause is name-agnostic
+	// (Event.LongRunningToolIDs). A stale response matches nothing, so
+	// Workflow.Resume yields ErrNothingToResume instead of a silent fresh Run.
 	responses := map[string]any{}
 	for _, fr := range frs {
 		if fr == nil || fr.ID == "" {
 			continue
 		}
-		responses[fr.ID] = decodeWorkflowInputResponse(fr)
+		responses[fr.ID] = decodeResumeResponse(fr)
 	}
 	if len(responses) == 0 {
 		return nil, nil, false, nil
@@ -164,8 +161,11 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 	return responses, state, true, nil
 }
 
-// decodeWorkflowInputResponse extracts the user-supplied payload
-// from a FunctionResponse targeting a workflow input request.
+// decodeResumeResponse extracts the user-supplied payload from a
+// FunctionResponse that answers a paused node's interrupt — a workflow input
+// (adk_request_input), tool consent (adk_request_credential), or confirmation
+// (adk_request_confirmation). detectResume matches by interrupt ID, so this is
+// no longer specific to workflow input.
 //
 // Three accepted shapes, in priority order:
 //
@@ -175,7 +175,7 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 //     other type, it is returned as-is.
 //  2. {"payload": <any>}     — value returned verbatim.
 //  3. anything else           — the whole Response map is returned.
-func decodeWorkflowInputResponse(fr *genai.FunctionResponse) any {
+func decodeResumeResponse(fr *genai.FunctionResponse) any {
 	if raw, ok := fr.Response["response"]; ok {
 		if s, isStr := raw.(string); isStr {
 			var decoded any
--- a/agent/workflowagent/workflow.go
+++ b/agent/workflowagent/workflow.go
@@ -118,27 +118,25 @@ func (a *workflowAgent) run(ctx agent.InvocationContext) iter.Seq2[*session.Even
 }
 
 // detectResume inspects the inbound user message for FunctionResponses
-// targeting a previously-emitted RequestInput. Returns the
-// responses map keyed by InterruptID (suitable for
-// Workflow.Resume), the RunState loaded from session, and true if
-// this turn is a resume; (nil, nil, false) for a fresh turn.
+// that answer a paused node's long-running interrupt. Returns the
+// responses map keyed by InterruptID (suitable for Workflow.Resume),
+// the RunState loaded from session, and true if this turn is a resume;
+// (nil, nil, false) for a fresh turn.
 func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]any, *workflow.RunState, bool, error) {
 	frs := utils.FunctionResponses(ctx.UserContent())
 	if len(frs) == 0 {
 		return nil, nil, false, nil
 	}
 
-	responses := map[string]any{}
-	for _, fr := range frs {
-		if fr.Name != workflow.WorkflowInputFunctionCallName {
-			continue
-		}
-		responses[fr.ID] = decodeWorkflowInputResponse(fr)
-	}
-	if len(responses) == 0 {
-		return nil, nil, false, nil
-	}
-
+	// A resume answers a paused node by matching a FunctionResponse to an
+	// open interrupt by ID. Match by ID, not by function name: besides the
+	// workflow's own adk_request_input, an LlmAgent node can pause on a
+	// tool's long-running request (adk_request_credential /
+	// adk_request_confirmation), and the engine's pause is name-agnostic
+	// (it rides on Event.LongRunningToolIDs). A stale or mistargeted
+	// response still reconstructs state; Workflow.Resume then yields
+	// ErrNothingToResume rather than silently starting a fresh Run.
+	//
 	// Scope rehydration to this run's invocation (the runner reuses the
 	// paused run's ID on resume) so a prior completed run in the same
 	// session does not leak in.
@@ -149,6 +147,17 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 		return nil, nil, false, err
 	}
 	if state == nil {
+		return nil, nil, false, nil
+	}
+
+	responses := map[string]any{}
+	for _, fr := range frs {
+		if fr == nil || fr.ID == "" {
+			continue
+		}
+		responses[fr.ID] = decodeWorkflowInputResponse(fr)
+	}
+	if len(responses) == 0 {
 		return nil, nil, false, nil
 	}
 
--- a/agent/workflowagent/workflow.go
+++ b/agent/workflowagent/workflow.go
@@ -24,6 +24,7 @@ import (
 	"google.golang.org/adk/v2/agent"
 	agentinternal "google.golang.org/adk/v2/internal/agent"
 	"google.golang.org/adk/v2/internal/utils"
+	"google.golang.org/adk/v2/internal/workflowstate"
 	"google.golang.org/adk/v2/session"
 	"google.golang.org/adk/v2/workflow"
 )
@@ -156,7 +157,7 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 	//
 	// A wrong-but-deliberate answer (right name, unknown ID) still reaches
 	// Resume, so the caller keeps the ErrNothingToResume diagnostic.
-	known := knownInterruptIDs(state)
+	known := workflowstate.ActionableInterruptIDs(state)
 	responses := map[string]any{}
 	for _, fr := range frs {
 		if fr == nil || fr.ID == "" {
@@ -172,32 +173,4 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 	}
 
 	return responses, state, true, nil
-}
-
-// knownInterruptIDs collects the interrupt IDs of nodes the rehydrated run can
-// still act on — those waiting for an answer, and those a re-entry node is
-// about to be re-run with. A node that already settled is excluded, so a reply
-// to an interrupt this run has finished with no longer counts as a resume.
-//
-// ResumedInputs is included, not just Interrupts: the runner appends the reply
-// to the session before the agent runs, so on a genuine first resume the
-// interrupt is already resolved and rehydration hands it back on the node.
-func knownInterruptIDs(state *workflow.RunState) map[string]struct{} {
-	ids := map[string]struct{}{}
-	for _, ns := range state.Nodes {
-		if ns == nil || (ns.Status != workflow.NodeWaiting && ns.Status != workflow.NodePending) {
-			continue
-		}
-		for _, id := range ns.Interrupts {
-			if id != "" {
-				ids[id] = struct{}{}
-			}
-		}
-		for id := range ns.ResumedInputs {
-			if id != "" {
-				ids[id] = struct{}{}
-			}
-		}
-	}
-	return ids
 }
--- a/agent/workflowagent/workflow.go
+++ b/agent/workflowagent/workflow.go
@@ -145,11 +145,25 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 	// tool's long-running request (adk_request_credential /
 	// adk_request_confirmation), not just the workflow's own
 	// adk_request_input, and the engine's pause is name-agnostic
-	// (Event.LongRunningToolIDs). A stale response matches nothing, so
-	// Workflow.Resume yields ErrNothingToResume instead of a silent fresh Run.
+	// (Event.LongRunningToolIDs).
+	//
+	// A response only routes this turn to Resume when it is actually aimed at
+	// this run: either its ID is an interrupt the run knows about, or it is
+	// explicitly an answer to the workflow's own input request. Otherwise the
+	// turn falls through to a fresh Run, as it did before ID-keyed matching —
+	// a turn may legitimately carry both text and an unrelated tool reply, and
+	// routing that to Resume would fail it with ErrNothingToResume and discard
+	// the text. Mirrors runner.buildResumeResponses, which filters the same way.
+	//
+	// A wrong-but-deliberate answer (right name, unknown ID) still reaches
+	// Resume, so the caller keeps the ErrNothingToResume diagnostic.
+	known := knownInterruptIDs(state)
 	responses := map[string]any{}
 	for _, fr := range frs {
 		if fr == nil || fr.ID == "" {
+			continue
+		}
+		if _, ok := known[fr.ID]; !ok && fr.Name != workflow.WorkflowInputFunctionCallName {
 			continue
 		}
 		responses[fr.ID] = utils.UnwrapResponse(fr.Response)
@@ -159,4 +173,32 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 	}
 
 	return responses, state, true, nil
+}
+
+// knownInterruptIDs collects the interrupt IDs of nodes the rehydrated run can
+// still act on — those waiting for an answer, and those a re-entry node is
+// about to be re-run with. A node that already settled is excluded, so a reply
+// to an interrupt this run has finished with no longer counts as a resume.
+//
+// ResumedInputs is included, not just Interrupts: the runner appends the reply
+// to the session before the agent runs, so on a genuine first resume the
+// interrupt is already resolved and rehydration hands it back on the node.
+func knownInterruptIDs(state *workflow.RunState) map[string]struct{} {
+	ids := map[string]struct{}{}
+	for _, ns := range state.Nodes {
+		if ns == nil || (ns.Status != workflow.NodeWaiting && ns.Status != workflow.NodePending) {
+			continue
+		}
+		for _, id := range ns.Interrupts {
+			if id != "" {
+				ids[id] = struct{}{}
+			}
+		}
+		for id := range ns.ResumedInputs {
+			if id != "" {
+				ids[id] = struct{}{}
+			}
+		}
+	}
+	return ids
 }
--- a/agent/workflowagent/workflow.go
+++ b/agent/workflowagent/workflow.go
@@ -131,15 +131,6 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 		return nil, nil, false, nil
 	}
 
-	// A resume answers a paused node by matching a FunctionResponse to an
-	// open interrupt by ID. Match by ID, not by function name: besides the
-	// workflow's own adk_request_input, an LlmAgent node can pause on a
-	// tool's long-running request (adk_request_credential /
-	// adk_request_confirmation), and the engine's pause is name-agnostic
-	// (it rides on Event.LongRunningToolIDs). A stale or mistargeted
-	// response still reconstructs state; Workflow.Resume then yields
-	// ErrNothingToResume rather than silently starting a fresh Run.
-	//
 	// Scope rehydration to this run's invocation (the runner reuses the
 	// paused run's ID on resume) so a prior completed run in the same
 	// session does not leak in.
@@ -153,12 +144,18 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 		return nil, nil, false, nil
 	}
 
+	// Key by interrupt ID, not function name: an agent node pauses on a
+	// tool's long-running request (adk_request_credential /
+	// adk_request_confirmation), not just the workflow's own
+	// adk_request_input, and the engine's pause is name-agnostic
+	// (Event.LongRunningToolIDs). A stale response matches nothing, so
+	// Workflow.Resume yields ErrNothingToResume instead of a silent fresh Run.
 	responses := map[string]any{}
 	for _, fr := range frs {
 		if fr == nil || fr.ID == "" {
 			continue
 		}
-		responses[fr.ID] = decodeWorkflowInputResponse(fr)
+		responses[fr.ID] = decodeResumeResponse(fr)
 	}
 	if len(responses) == 0 {
 		return nil, nil, false, nil
@@ -167,8 +164,11 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 	return responses, state, true, nil
 }
 
-// decodeWorkflowInputResponse extracts the user-supplied payload
-// from a FunctionResponse targeting a workflow input request.
+// decodeResumeResponse extracts the user-supplied payload from a
+// FunctionResponse that answers a paused node's interrupt — a workflow input
+// (adk_request_input), tool consent (adk_request_credential), or confirmation
+// (adk_request_confirmation). detectResume matches by interrupt ID, so this is
+// no longer specific to workflow input.
 //
 // Three accepted shapes, in priority order:
 //
@@ -178,7 +178,7 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 //     other type, it is returned as-is.
 //  2. {"payload": <any>}     — value returned verbatim.
 //  3. anything else           — the whole Response map is returned.
-func decodeWorkflowInputResponse(fr *genai.FunctionResponse) any {
+func decodeResumeResponse(fr *genai.FunctionResponse) any {
 	if raw, ok := fr.Response["response"]; ok {
 		if s, isStr := raw.(string); isStr {
 			var decoded any
--- a/agent/workflowagent/workflow.go
+++ b/agent/workflowagent/workflow.go
@@ -81,18 +81,17 @@ func New(cfg Config) (agent.Agent, error) {
 // workflowAgent is the wrapper that dispatches between
 // Workflow.Run (fresh turn) and Workflow.Resume (resume turn).
 // The dispatch decision is made by inspecting ctx.UserContent for
-// a FunctionResponse targeting a previously-emitted RequestInput.
-// The workflow's RunState lives in session.State, not on this
-// struct, so a single *workflowAgent safely services many
-// concurrent sessions.
+// a FunctionResponse that answers an interrupt this run can still
+// act on — see detectResume. The workflow's RunState lives in
+// session.State, not on this struct, so a single *workflowAgent
+// safely services many concurrent sessions.
 type workflowAgent struct {
 	workflow *workflow.Workflow
 }
 
 // run is the agent.Config.Run callback. It dispatches between
-// Workflow.Resume (when the inbound user content carries a
-// FunctionResponse to a previously-emitted RequestInput) and
-// Workflow.Run (every other turn).
+// Workflow.Resume (when the inbound user content answers one of
+// this run's open interrupts) and Workflow.Run (every other turn).
 func (a *workflowAgent) run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
 	return func(yield func(*session.Event, error) bool) {
 		responses, state, ok, err := a.detectResume(ctx)
--- a/agent/workflowagent/workflow.go
+++ b/agent/workflowagent/workflow.go
@@ -121,27 +121,25 @@ func (a *workflowAgent) run(ctx agent.InvocationContext) iter.Seq2[*session.Even
 }
 
 // detectResume inspects the inbound user message for FunctionResponses
-// targeting a previously-emitted RequestInput. Returns the
-// responses map keyed by InterruptID (suitable for
-// Workflow.Resume), the RunState loaded from session, and true if
-// this turn is a resume; (nil, nil, false) for a fresh turn.
+// that answer a paused node's long-running interrupt. Returns the
+// responses map keyed by InterruptID (suitable for Workflow.Resume),
+// the RunState loaded from session, and true if this turn is a resume;
+// (nil, nil, false) for a fresh turn.
 func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]any, *workflow.RunState, bool, error) {
 	frs := utils.FunctionResponses(ctx.UserContent())
 	if len(frs) == 0 {
 		return nil, nil, false, nil
 	}
 
-	responses := map[string]any{}
-	for _, fr := range frs {
-		if fr.Name != workflow.WorkflowInputFunctionCallName {
-			continue
-		}
-		responses[fr.ID] = decodeWorkflowInputResponse(fr)
-	}
-	if len(responses) == 0 {
-		return nil, nil, false, nil
-	}
-
+	// A resume answers a paused node by matching a FunctionResponse to an
+	// open interrupt by ID. Match by ID, not by function name: besides the
+	// workflow's own adk_request_input, an LlmAgent node can pause on a
+	// tool's long-running request (adk_request_credential /
+	// adk_request_confirmation), and the engine's pause is name-agnostic
+	// (it rides on Event.LongRunningToolIDs). A stale or mistargeted
+	// response still reconstructs state; Workflow.Resume then yields
+	// ErrNothingToResume rather than silently starting a fresh Run.
+	//
 	// Scope rehydration to this run's invocation (the runner reuses the
 	// paused run's ID on resume) so a prior completed run in the same
 	// session does not leak in.
@@ -152,6 +150,17 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 		return nil, nil, false, err
 	}
 	if state == nil {
+		return nil, nil, false, nil
+	}
+
+	responses := map[string]any{}
+	for _, fr := range frs {
+		if fr == nil || fr.ID == "" {
+			continue
+		}
+		responses[fr.ID] = decodeWorkflowInputResponse(fr)
+	}
+	if len(responses) == 0 {
 		return nil, nil, false, nil
 	}
 
--- a/agent/workflowagent/workflow.go
+++ b/agent/workflowagent/workflow.go
@@ -18,11 +18,8 @@
 package workflowagent
 
 import (
-	"encoding/json"
 	"fmt"
 	"iter"
-
-	"google.golang.org/genai"
 
 	"google.golang.org/adk/v2/agent"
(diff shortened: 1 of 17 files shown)


workflow/agent_node.go
```diff
@@ -90,6 +104,12 @@ func NewAgentNode(a agent.Agent, cfg NodeConfig) (*AgentNode, error) {
 // Run implements the Node interface.
 func (n *AgentNode) Run(ctx agent.Context, input any) iter.Seq2[*session.Event, error] {
 	return func(yield func(*session.Event, error) bool) {
+		// On resume, re-feeding the input would make a single_turn/task
+		// LlmAgent re-call the still-pending tool and pause again; drop it
+		// so the agent continues from history. Mirrors runner.runAgentNodeBody.
+		if nodeIsResuming(ctx.Session(), ctx.InvocationID(), n.agent.Name()) {
```
karolpiotrowicz: This drops the input on every activation of the node once the invocation contains an answered interrupt for it, not only on the resume activation. `nodeIsResuming` is re-derived from history, and history never "un-answers", so the predicate latches for the rest of the invocation.

That matters because a node can legitimately be activated again in the same invocation: a conditional loop-back edge (`{From: n, To: n, Route: StringRoute("loop")}` — the shape `TestScheduler_RetryLoop` already exercises; only *unconditional* cycles are rejected), a `RetryConfig` retry, and a `ParallelWorker`'s per-item activations. Each of those then runs with `input = nil`, silently — no error, and after `ValidateInput` has already accepted the input that is about to be discarded.

Measured on a self-loop graph whose agent has an answered interrupt in the current invocation: on this branch both activations receive `nil` user content; on the base commit both receive their input.

The engine already carries an exact per-activation signal for this: `startNode` injects `ResumeInputs` into the node context, non-nil only for `scheduleResumedNode` (the `Resume` path), nil for a normal `scheduleNode`, and `handleCompletion` clears `ns.ResumedInputs` on completion with the comment "a future activation (if any, e.g. via loop-back routing) starts a fresh lifecycle". Gating the drop on that signal instead of on a history scan would remove the latch — and, as a side benefit, avoid the O(history) rescan this does on every activation of every `AgentNode`.

workflow/agent_node.go
```diff
@@ -163,6 +183,57 @@ func (n *AgentNode) Run(ctx agent.Context, input any) iter.Seq2[*session.Event,
...
+	answered := map[string]struct{}{}
+	for i := 0; i < events.Len(); i++ {
+		ev := events.At(i)
+		if ev == nil {
+			continue
+		}
+		if invocationID != "" && ev.InvocationID != invocationID {
+			continue
+		}
+		if ev.Author == agentName {
```
karolpiotrowicz: The doc comment above says an interrupt raised under a different author "rides on the child node's own Run" — that holds when each agent is its own node, but not for an `AgentNode` wrapping a composite agent, where the sub-agent is not a node at all. This file already forwards distinct-author sub-agent events (see the comment on `isOwnAgentEvent` and the synthesize guard below), and the scheduler calls `trackInterrupts` *before* its descendant short-circuit and stamps `NodeInfo.Path`, so the engine does attribute the sub-agent's interrupt to this node, parks it, and — with the new default — re-enters it.

So on resume the node re-enters but this check returns false, and the original input is re-fed: exactly the "re-call the still-pending tool and pause again" loop the PR is fixing, still present for delegating agents. Reproduced end-to-end with an `AgentNode` over an agent whose pause event is authored by a sub-agent: the node re-enters (2 activations) and receives its node input on the resume activation.

Consider attributing with the same rule rehydration uses — `NodeInfo.Path` via `eventNodeName`, with `Author` as the fallback — so the node's view of "my interrupt" matches the engine's.

workflow/agent_node.go
```diff
@@ -163,6 +183,57 @@ func (n *AgentNode) Run(ctx agent.Context, input any) iter.Seq2[*session.Event,
...
+			for _, id := range ev.LongRunningToolIDs {
+				if id != "" {
+					open[id] = struct{}{}
+				}
+			}
+		}
+		for _, fr := range utils.FunctionResponses(utils.Content(ev)) {
```
karolpiotrowicz: Two things about this loop:

1. **No `Author == "user"` gate.** The equivalent scan in `persistence.go` gates on it and spells out why: "A user FunctionResponse resolves an interrupt — not the tool's own initial 'pending' response (authored by the node)". When a long-running tool returns a non-nil result the flow emits a `FunctionResponse` carrying the same ID as the pausing call, authored by the agent. Both halves of the predicate can therefore be satisfied by the agent's own events with no human reply at all — I get `nodeIsResuming == true` from a history containing only the raise plus the agent-authored pending response. Combined with the latching above, that means a non-resume activation can lose its input with nobody in the loop.

2. **`utils.FunctionResponses` does not nil-check parts** (unlike `frPart` in `persistence.go`), and this now runs over all history on every activation, so a single event whose content holds a `nil` part panics the node. Reproduced with `Parts: []*genai.Part{nil}`. The scheduler recovers node panics, so the blast radius is small, but using `frPart`-style iteration here would match the sibling scanner and cost nothing.

agent/workflowagent/workflow.go
```diff
@@ -152,11 +141,31 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 		return nil, nil, false, nil
 	}
 
+	// Key by interrupt ID, not function name: an agent node pauses on a
+	// tool's long-running request (adk_request_credential /
+	// adk_request_confirmation), not just the workflow's own
+	// adk_request_input, and the engine's pause is name-agnostic
+	// (Event.LongRunningToolIDs). A stale response matches nothing, so
+	// Workflow.Resume yields ErrNothingToResume instead of a silent fresh Run.
+	responses := map[string]any{}
```
karolpiotrowicz, line 157: Keying by interrupt ID is the right fix for dispatch, but dropping the name filter also lets consent/confirmation replies reach the *handoff* branch of `Workflow.Resume`, and nothing on that path validates them. `schemaFromEvent` only records a schema for the workflow's own input requests, so `validateResolved` is a pass-through for `adk_request_credential` / `adk_request_confirmation`; the handoff branch then marks the node completed and forwards the raw reply as its output.

Concretely, for a node in handoff mode — the default for a non-LlmAgent `AgentNode`, which includes remote-agent nodes that carry long-running IDs reconstructed from the remote — replying to a paused confirmation with `{"confirmed": false}` completes the node and runs its unconditional successors exactly as an approval would, with the raw payload (including anything else in the map) as their input. On the base commit the same turn ran nothing, because consent replies never reached `Resume`. For an `AgentNode` successor that payload would also be JSON-marshalled into the successor's user content by `nodeInputToContent`, which is the sort of content `contents_processor.go` deliberately keeps out of model history.

This is narrower than it sounds — an LlmAgent node now defaults to re-entry, where the tool layer still interprets the reply — but it would be good to make it structural rather than incidental: e.g. resolve `fr.ID` back to the call that raised it and either require the documented shape for consent/confirmation kinds, or force re-entry for them so the reply is interpreted by the tool layer rather than by the graph.

karolpiotrowicz (commented):
The resume dispatch is well aimed and the reported bug is real. I checked it out: with the change reverted, the follow-up consent turn falls through to a fresh `Workflow.Run` that replays the completed upstream node and re-feeds the paused node its input; with the change applied, the paused node resumes in place. Build, the full test suite, `go vet`, `golangci-lint`, `-race` on both touched packages and an exported-API compatibility check are all clean, and no exported signature changes.

Three things I think are worth resolving before this lands, all of which I was able to reproduce; details are in the inline comments.

**1. `nodeIsResuming` latches for the whole invocation.** It is derived from session history, and history never "un-answers", so once this node's interrupt is answered anywhere in the invocation the predicate stays true for *every* later activation of the node — a conditional loop-back edge, a `RetryConfig` retry, or a parallel-worker per-item activation. Those activations then silently lose their input. Measured on a self-loop graph: on this branch both activations receive `nil` user content, on the base commit both receive their input.

**2. The composite/sub-agent case is still broken.** When the pause is raised by a delegated sub-agent, the scheduler attributes the interrupt to this node and re-enters it, but the `Author == agentName` test does not recognise it, so the node is re-fed its original input — the "re-call the pending tool and pause again" loop this PR sets out to remove. Measured end-to-end: the node re-enters (good) but with its input re-fed.

**3. Consent replies now reach the handoff branch unvalidated.** Removing the name filter is the right call for dispatch, but it also routes `adk_request_credential` / `adk_request_confirmation` replies into the handoff path, which has no schema for those interrupt kinds and treats any payload as the answer. A node in handoff mode (the default for a non-LlmAgent `AgentNode`) completes on `{"confirmed": false}` and runs its unconditional successors, with the raw payload as their input.

Smaller points, also inline: the whole-history scan can be answered by the agent's own pending tool response rather than a human reply; it uses a part helper that does not nil-check parts; the LlmAgent default is currently asserted by no test; and `decodeResumeResponse` and the history-side `unwrapResponse` unwrap the same reply differently.

Two questions rather than claims:

- A user turn whose `FunctionResponse` matches no open interrupt now fails the whole turn with `ErrNothingToResume` (zero events) where it previously started a fresh `Workflow.Run`. For a genuinely *stale* reply that seems clearly better. Is it also intended for a reply that was never a workflow interrupt at all — a client-side function result arriving while some node happens to be paused?
- For agent nodes declared in YAML, `internal/configurable/configurable_workflow.go:334` constructs `workflow.NodeConfig{}` and there is no agent-node `rerun_on_resume` key, so such an LlmAgent node always takes the new default with no way to opt out. Is that acceptable for now, or worth a follow-up so "an explicit value always wins" also holds there?

One note on the test suite rather than the code: deleting the `RerunOnResume` default block leaves the entire repository's tests green, so the headline behaviour of this PR is currently unasserted. (Making the same assignment unconditional *is* caught, by `agent/remoteagent/v2`, so the tri-state itself is covered.)

agent/workflowagent/workflow.go
```diff
@@ -166,7 +175,7 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
 //     other type, it is returned as-is.
 //  2. {"payload": <any>}     — value returned verbatim.
 //  3. anything else           — the whole Response map is returned.
-func decodeWorkflowInputResponse(fr *genai.FunctionResponse) any {
+func decodeResumeResponse(fr *genai.FunctionResponse) any {
```
karolpiotrowicz: Minor, but worth noting now that this function decodes consent and confirmation replies too: it and the history-side `unwrapResponse` in `persistence.go` disagree on the same payload. `unwrapResponse` unwraps only when the map has exactly one key and also understands `"result"`; this one unwraps `"response"`/`"payload"` regardless of the map's other keys and never looks at `"result"`. So `{"result": v}` or `{"response": v, "other": w}` decode differently depending on whether the reply is taken from the inbound turn or rehydrated from history — which can now differ across a process restart for the same logical reply. A shared helper used by both would remove the divergence.

workflow/agent_node_test.go
```diff
@@ -831,3 +831,115 @@ func TestAgentNode_AutomaticOutputExtraction(t *testing.T) {
...
+			if got := nodeIsResuming(sess, tc.invocationID, agentName); got != tc.want {
+				t.Errorf("nodeIsResuming(_, %q, %q) = %v, want %v", tc.invocationID, agentName, got, tc.want)
+			}
+		})
+	}
+}
+
+// TestNewAgentNode_RerunOnResumeDefault pins the isLlmAgent-gated default:
+// a non-LlmAgent node gets no default, and an explicit value always wins.
+func TestNewAgentNode_RerunOnResumeDefault(t *testing.T) {
```
karolpiotrowicz: This pins the two negative cases but not the positive one: all three subtests use `nonLLM` (built with `agent.New`), which does not satisfy `llminternal.Agent`, so `applyAgentNodeDefaults` returns at the early `if !ok` and the default block is never reached. Deleting that block leaves the whole repository's test suite green, so the PR's headline behaviour — an LlmAgent node defaulting to re-entry — is currently asserted nowhere. (The tri-state itself is covered: making the assignment unconditional does fail a test elsewhere.)

A fourth case built from a real `llmagent.New` agent asserting `RerunOnResume != nil && *… == true` would close it; `agent_node_modes_test.go` already has the construction pattern.

workflow/agent_node.go
wolo-lab: Confirmed and fixed — the sharpest of the six.

Reproduced on a self-loop graph (`{From: n, To: n, Route: StringRoute("loop")}`)
whose agent has an answered interrupt in the current invocation: **both**
activations received `nil` user content on this branch; both received their
input on the base commit. Exactly the latch you describe.

I took your suggestion — gate on the per-activation signal — but the pure form
introduced a different bug, so the fix carries **two** gates:

1. `IsResumeActivation()` on the node context (`resumeInputs != nil`), read
   through an optional-interface assertion. `startNode` always supplies the
   delta, so a fresh activation resets it to false; only `scheduleResumedNode`
   makes it true. No latch, and the `O(history)` rescan is off the normal path.
2. Proof the pause was raised **under this node**.

The second gate exists because the flag alone over-fires: `dynamic_scheduler.go:234`
builds the child delta without `ResumeInputs`, so a dynamic child *inherits* the
resuming ancestor's context. Measured on the pure-flag version — an orchestrator
that resumes and then delegates to a fresh `AgentNode` child (the
`llm_agent_wrapper.go:394` / `single_turn_tool.go:74` shape) — the child got
`userContent = <nil>`.

Each gate is mutation-checked: removing gate 1 fails
`TestAgentNode_LoopBackKeepsInput`, removing gate 2 fails
`TestAgentNode_FreshDynamicChildKeepsInput`.

workflow/agent_node.go
wolo-lab: Confirmed and fixed. Reproduced end-to-end: an `AgentNode` over an agent whose
pause event is authored by a sub-agent — rehydration attributed it to the node
(`Status:PENDING ResumedInputs:map[X:yes]`), the node re-entered, and the resume
activation was re-fed `task-input`.

Adopted your rule: attribution is now `NodeInfo.Path` with `Author` as the
fallback. One wrinkle worth noting — it needs the same *segment* walk
`eventNodeName` does (split on `/`, strip `@run`), not a plain `staticNodeName`
compare: the scheduler stamps `expectedPath = nr.nodePath`, i.e. `"coord@1"`, so
comparing the whole first segment misses. My first attempt did exactly that and
this case stayed broken.

Pinned by `TestAgentNode_ResumesOnSubAgentInterrupt`; mutation-checked
(author-only attribution fails it).

workflow/agent_node.go
wolo-lab: Both halves confirmed. Both are moot now — the replacement never scans FunctionResponses, it only reads LongRunningToolIDs, which is nil-safe — but for the record:

* No Author == "user" gate. Measured nodeIsResuming == true from a history containing only the raise plus the agent's own pending FunctionResponse, with no human reply. Exactly as you described.
* Nil part. Parts: []*genai.Part{nil} panics — utils.FunctionResponses has no nil check, unlike frPart. Reproduced.

utils.FunctionResponses (and FunctionCalls) still lack the nil check for their other callers. Left alone as out of scope here — say the word and I'll file a separate issue.

agent/workflowagent/workflow.go
wolo-lab: Reproduced, with two corrections.

Baseline. On 020d9416 that turn doesn't run nothing — it falls through to a fresh Workflow.Run and re-executes completed upstream nodes (this PR's own test fails there with upstream runs after resume = 2, want 1). Handoff forwarding is narrower than what it replaced.

Parity. It matches adk-python exactly, so forcing re-entry would be a deliberate divergence:

agent/workflowagent/workflow.go
wolo-lab: Fixed: one utils.UnwrapResponse used by all three call sites, both copies deleted. Semantics are the history/adk-python rule, so a reply read back from history decodes identically to the same reply taken from the inbound tur

workflow/agent_node_test.go
wolo-lab: Fixed by moving the table to workflow_test (the internal package can't import llmagent — import cycle) and adding the LlmAgent case. The mutation now fails with RerunOnResume = nil, want true.

karolpiotrowicz (dismissed):
The two-signal `isResuming` is the right shape and it holds up under pressure — I re-ran the loop-back, fresh-dynamic-child and re-delegation cases and each gate is genuinely load-bearing, and attributing by `NodeInfo.Path` with `Author` as the fallback does match what the engine parks and re-enters on. Deleting the LlmAgent default block now fails a test, which closes the last round's gap. Collapsing the three decoders into one is a clear improvement, and preserving the `confirmed` flag on a multi-key map fixes a real bug.

Three behaviours are worse on this branch than on `main`. Two of them I'd want resolved before this lands.

### A rejected confirmation completes the node and runs its guarded successors

Build a handoff-mode `AgentNode` (a non-LlmAgent, so `RerunOnResume` stays nil) whose agent raises two `adk_request_confirmation` interrupts in one event — the shape `generateRequestConfirmationEvent` emits when a model produces two tool calls that each need confirmation. Give it one unconditional successor. On the follow-up turn, reply `{"confirmed": false}` to the first interrupt only and leave the second unanswered.

- **On `main`:** the successor runs 0 times, the node stays paused, and both confirmations are re-raised.
- **On this branch:** the successor runs once, the node is no longer paused, and the second confirmation is silently discarded.

The chain is three steps. [`knownInterruptIDs`](https://github.com/google/adk-go/blob/993614c777377cd9a4859af3409da5a2f5cfa5a1/agent/workflowagent/workflow.go#L193) admits an ID that is *already answered and in history* via `ResumedInputs`, so the new ID-keyed filter lets a tool-named reply through where the name filter rejected it outright. In `Resume`, `freshMatched` then stays empty, because [the loop](https://github.com/google/adk-go/blob/993614c777377cd9a4859af3409da5a2f5cfa5a1/workflow/resume.go#L124) iterates `ns.Interrupts` — the *unresolved* IDs — and the answered one is not among them. Control reaches the handoff branch, which [completes the node and clears its interrupts](https://github.com/google/adk-go/blob/993614c777377cd9a4859af3409da5a2f5cfa5a1/workflow/resume.go#L181-L183) unconditionally, without reading the payload or checking that unresolved interrupts remain.

Worth noting alongside it: `schemaFromEvent` records a schema only for the workflow's own input request, so `ns.interruptSchemas[id]` is always nil for confirmation and credential interrupts and `validateResolved` is a pass-through for exactly the payloads this change newly routes into `Resume`. The successors also receive `nil` rather than the reply, since `out := ns.Output` is nil for a `NodeWaiting` node, so they cannot inspect the decision they are gated on.

This affects every `AgentNode` over a non-LlmAgent, including remote-agent nodes. The new re-entry default deliberately does not cover that population.

### A resume that emits and then fails leaves the run permanently unresumable

Take a re-entry node whose resume activation emits one persisted event and then returns an error — an LlmAgent that emits a function-response event and then has its model call fail is the ordinary shape. Retry the same approval.

- **On `main`:** the retry re-runs the node and the work completes.
- **On this branch:** every retry returns `ErrNothingToResume` with zero events, and the node never runs again. The approved side effect never settles.
- **Control:** the identical failure with *no* event emitted first recovers normally on this branch, which isolates the cause to the sweep below.

[The consumed sweep](https://github.com/google/adk-go/blob/993614c777377cd9a4859af3409da5a2f5cfa5a1/workflow/persistence.go#L192-L193) marks every answer already in history as consumed the moment the node authors *any* event, so a node that emitted and then died is indistinguishable from one that settled. Rehydration then sets [`reentryConsumed`](https://github.com/google/adk-go/blob/993614c777377cd9a4859af3409da5a2f5cfa5a1/workflow/persistence.go#L374) and `Resume` [skips it](https://github.com/google/adk-go/blob/993614c777377cd9a4859af3409da5a2f5cfa5a1/workflow/resume.go#L153) on every subsequent turn. Nothing recovers it, because history never un-answers. Tying `consumed` to the node actually settling — an `Output`, or a new interrupt in `LongRunningToolIDs` — rather than to any emitted event would close it, though you may see a cleaner cut.

For what it's worth, streaming chunks are safe here: the runner gates `AppendEvent` on `!event.LLMResponse.Partial`, so a partial never reaches history. It takes a complete event to trigger this.

### A replayed approval carrying new user text destroys the whole turn

After a genuine re-entry resume the node stays `NodePending` with `ResumedInputs` populated, and `knownInterruptIDs` [gates on Status alone](https://github.com/google/adk-go/blob/993614c777377cd9a4859af3409da5a2f5cfa5a1/agent/workflowagent/workflow.go#L185), never on `reentryConsumed` — so that settled ID stays "known" forever. Any later turn that echoes the reply alongside real user text is routed to `Resume`, hits the `reentryConsumed` skip, schedules nothing, and fails with `ErrNothingToResume`. The user's text is discarded. Sending "now book me a flight" plus an echoed approval yields 0 events and an error on this branch, against 1 event and no error on `main`.

This is the outcome the comment directly above the new filter says must not happen. `TestRunner_WorkflowHITL_SettledResponseDoesNotHijackTurn` covers the handoff shape, where the node rehydrates `NodeCompleted` and is correctly excluded. The re-entry shape — now the LlmAgent default, so the population this PR is written for — is not covered.

### Three more, non-blocking but worth a look

- **Partial resume is not idempotent.** The partial-resume branch at [`persistence.go:352`](https://github.com/google/adk-go/blob/993614c777377cd9a4859af3409da5a2f5cfa5a1/workflow/persistence.go#L352-L358) never sets `reentryConsumed`, so a replayed answer re-runs a re-entry node whenever a second interrupt is still open. Same on `main`, so not a regression, but it is a reachable replay the new guard does not cover. A resume activation that performs a side effect and emits nothing has the same hole, for the same reason as the wedge above.
- **`samePathActivation` matches in both directions.** [The suffix test](https://github.com/google/adk-go/blob/993614c777377cd9a4859af3409da5a2f5cfa5a1/workflow/agent_node.go#L276) returns true for `("worker@1", "orch@1/worker@1")`, so if the same node name exists both at root level and as a dynamic child, the child drops input it was explicitly handed, with no error. The reverse arm exists because `Run` stamps the workflow prefix and `Resume` rebuilds the path without it — making those two agree would let the match be one-directional.
- **A new session read from a node goroutine.** [`raisedInterrupt`](https://github.com/google/adk-go/blob/993614c777377cd9a4859af3409da5a2f5cfa5a1/workflow/agent_node.go#L234) calls `sess.Events()` from inside a node's goroutine. [`session/database`](https://github.com/google/adk-go/blob/993614c777377cd9a4859af3409da5a2f5cfa5a1/session/database/session.go#L60-L62) and `session/vertexai` return `events(s.events)` without taking the read lock their `AppendEvent` holds for writing, unlike `session/inmemory`. That is their bug rather than this PR's, but this adds a reader on a concurrent path and `-race` reports it through the runner over sqlite.

### Smaller points

- YAML-declared agent nodes take the new re-entry default with no way out. [`configurable_workflow.go:335`](https://github.com/google/adk-go/blob/993614c777377cd9a4859af3409da5a2f5cfa5a1/internal/configurable/configurable_workflow.go#L335) still passes an empty `NodeConfig{}`, while the FunctionNode and ToolNode paths both read `cfg.RerunOnResume`. This is the second of the two questions from last time.
- The unified decoder widens JSON-parsing of string values to `payload` and `result`, where the old inbound decoders only did it under `response`. So `{"payload": "true"}` now yields a bool where it used to yield the string. I could not find a producer in the repo that hits it, and the console HITL helper parses first so it cannot, but it is a real change in the wire contract and worth a line in the PR description.
- A handoff node paused on a tool interrupt still falls through to a fresh `Workflow.Run` and replays completed upstream nodes, because `knownInterruptIDs` excludes `NodeCompleted`, which is what such a node rehydrates to. Unchanged from `main`, but the doc comment on the new dispatch reads as though name-agnostic resume now works everywhere, and it works for re-entry nodes only.

The nil-part crash in `utils.FunctionResponses` is still reachable from `detectResume`, outside the scheduler's recover. Please do file that separately as you offered.

One question before you start on the first item, since it changes the shape of the fix: are handoff-mode nodes in scope for tool interrupts at all? If they are, the handoff branch needs to refuse to complete a node that still has unresolved interrupts, and confirmation and credential kinds need a schema. If they are not, the cleaner cut is to keep tool-named replies out of `Resume` for handoff nodes and force re-entry for those kinds, which also answers the point from last round. Your call — I have no strong view on which, only that the current middle ground is the one that bites.

wolo-lab:
All three of the behaviours you called worse than `main` reproduce, and all three are fixed. Two things you flagged as smaller points were already fixed in commits that were sitting unpushed on my end when you reviewed, so you were reading a stale head for those — sorry.

## A rejected confirmation completing the node

Fixed, and your question answers itself against adk-python. `_replay_interceptor`'s Case 2 is exactly this shape:

```python
if unresolved:
    if node.rerun_on_resume and recovered.resolved_ids:
        should_run = True
        resume_inputs = recovered.resolved_responses
    else:
        interrupts = unresolved
```

A node that does not rerun and still has unresolved interrupts stays waiting and re-bubbles them. Case 4 then keeps the handoff semantic for the all-resolved case, so handoff nodes are in scope and the answer is your first option rather than the second.

So [`Resume`'s handoff branch](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/resume.go#L174-L188) now refuses to complete a node with unanswered interrupts left. Rejecting one of two confirmations keeps the node waiting with the second still open and schedules no successors. The all-answered path is unchanged, and there is a control test for it so the two do not drift.

One thing that fell out of it: the turn used to end in `ErrNothingToResume`, because nothing was scheduled. That is wrong for the same reason it was wrong in your third item — the answer did arrive and was recorded. `answeredThisTurn` now moves above the switch in `inferNodeState` so every arm computes it, and a node left waiting on its other interrupts counts as progress.

I did not do the schema half. `schemaFromEvent` still records nothing for confirmation and credential interrupts, so `validateResolved` is still a pass-through for them. Happy to take it here if you want it in scope, otherwise I will file it.

## A resume that emits and then fails

Fixed with the cut you suggested, and I owe you a correction on it. My first attempt looked like it did not work: I probed with an agent that emitted a final *text* event and then died, and the retry stayed wedged. That is `synthesizeAgentOutput` doing its job — it stamps an `Output` on a final model text event, so that event genuinely does carry the node's output and the node had in fact settled. Your shape, a function-response event, carries no output and no interrupt, and the cut catches it.

[The sweep](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/persistence.go#L188-L200) now marks answers consumed only on an event with an `Output` or new `LongRunningToolIDs`. Retrying an approval after a mid-flight failure re-runs the node.

The boundary worth naming, since it is not obvious: an agent whose final text lands and which then fails is still treated as settled. That looks right to me — the output exists — but it is a real edge and I would rather you knew where the line is than discover it later.

## A replayed approval carrying new user text

Fixed. A settled ID stays recognisable, but `ActionableInterruptIDs` now reports liveness per ID rather than bare membership, and a turn whose only matches are spent falls through to a fresh `Run` when it carries anything else. A bare replay with nothing else in it still reaches `Resume`, so the diagnostic you get from `TestWorkflowAgent_Resume_ReentryIsIdempotent` is unchanged.

Your reading of the mirror comment was right and I have dropped the claim. `buildResumeResponses` builds its pending set from `waitingInterruptIDs`, which needs `NodeWaiting`, union `openLongRunningCallIDs`, which deletes every answered ID — and the runner appends the inbound message before the agent runs. A settled ID is in neither set, so the two filters never did filter the same way.

## The rest

**Partial resume is not idempotent** — fixed. The partial-resume arm sets `reentryConsumed` on the same terms as the all-resolved one. There is a positive control asserting that answering the still-open interrupt continues to re-run the node, since that is the thing a replay guard is most likely to swallow.

**YAML agent nodes** — fixed, in a commit from 12 August that had not been pushed when you looked. The agent-node branch reads `cfg.RerunOnResume` like the FunctionNode and ToolNode paths.

**A handoff node falling through to a fresh `Workflow.Run`** — also fixed in an unpushed commit from the same day. A node that settled its interrupt on this very turn now stays in the filter, handed to the dispatching packages through `internal/workflowstate` rather than by widening the workflow package's API.

**The decoder widening** — now called out in the PR description, with `{"payload": "true"}` as the example.

**`samePathActivation`** — confirmed at the unit level, `("worker@1", "orch@1/worker@1")` does return true. I could not build a reachable case: my attempt parked the graph at an upstream asker so the same-named static node never paused, and the precondition never fired. Your fix is the right shape and I would rather do it as a follow-up than land a path change I cannot demonstrate. Say the word if you want it here.

**The unlocked `Events()`** — verified in both backends and filed as #1553. `LastUpdateTime` two lines below takes `RLock` in each, so it reads as an omission rather than a decision.

**The nil-part crash** — filed as #1552. It is all three `Parts` accessors in `internal/utils`, not just `FunctionResponses`.

The branch is rebased onto `main` and no longer conflicts. The mode mutation is gone from `applyAgentNodeDefaults` — `Run`'s per-placement binding already covers it — along with the godoc that promised it and the adk-python parity claim that went with it, which was wrong: `build_node` sets the mode on a `clone()`, not on the agent it is handed.

workflow/resume.go
```diff
@@ -164,6 +171,21 @@ func (w *Workflow) Resume(
...
+			} else if remaining := unansweredInterrupts(ns, freshMatched); len(remaining) > 0 {
+				// Handoff, but the node raised interrupts nobody has
+				// answered yet. It is still waiting: completing it here
+				// would drop the unanswered ones and hand successors an
+				// output standing in for a decision that was never made
+				// — for a rejected confirmation, one gating the very
+				// work it rejected. Mirrors adk-python's replay
+				// interceptor, which keeps such a node waiting and
+				// re-bubbles the unresolved IDs.
+				ns.Interrupts = remaining
```
karolpiotrowicz, line 183: This assignment is never read. Replacing it with a no-op leaves every test in `workflow`, `agent/workflowagent`, `internal/configurable` and `runner` green — pass 2 reads only `Branch` and `state.completed`, and `finalize` reads only `Status`.

The comment above it says the branch "re-bubbles the unresolved IDs", mirroring adk-python's replay interceptor. It does not. With nothing scheduled, this turn yields no event and no error, so a client that just rejected one of two confirmations gets an entirely empty response with no fresh signal that the second is still open. The first half of the claim is real and is what `TestWorkflowAgent_Handoff_PartialAnswerKeepsNodeWaiting` pins — the node stays waiting and successors do not run. It is the re-bubbling half that is not there.

workflow/resume.go
```diff
@@ -164,6 +171,21 @@ func (w *Workflow) Resume(
 				ns.Status = NodePending
 				s.scheduleResumedNode(node, ns.Input, ns.TriggeredBy, ns.Branch, ns.ResumedInputs)
 				scheduled++
+			} else if remaining := unansweredInterrupts(ns, freshMatched); len(remaining) > 0 {
```
karolpiotrowicz, line 174: Unlike the re-entry branch just above, this one never folds `freshMatched` into `ns.ResumedInputs`. A caller driving `ReconstructRunState` and `Resume` across two calls, supplying answers that are not yet in history, loses the first batch: the later call computes `out = resumeOutput(freshMatched)` over only its own answers.

I could not build a realistic caller for it. Both in-tree dispatchers append the user message to the session before rehydrating, so `freshMatched` is always empty on those paths, and `interruptSchemas`, `answeredThisTurn` and `reentryConsumed` are unexported, so `ReconstructRunState` is the only source of a usable `RunState`. Not blocking. It is still a change from `c6a514b8`, where the same call consumed `freshMatched` through the else arm.

agent/workflowagent/workflow.go
```diff
@@ -155,33 +143,72 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
...
+			continue
+		}
+		if fr := p.FunctionResponse; fr != nil {
+			if _, ok := matched[fr.ID]; !ok {
+				return true
 			}
-			return s
+			continue
+		}
+		if p.Text != "" || p.FunctionCall != nil || p.InlineData != nil || p.FileData != nil {
```
karolpiotrowicz, line 209: Deleting `p.InlineData != nil || p.FileData != nil` from this line leaves every test green, so nothing echoes a settled approval alongside an attachment.

Beyond the coverage gap, the allow-list shape means any part kind not named here — `ExecutableCode`, `CodeExecutionResult`, or whatever `genai.Part` grows next — is read as "nothing else in this message", so the turn routes to `Resume` and fails with `ErrNothingToResume`, dropping the part. Inverting the test, so that anything which is not a matched `FunctionResponse` counts as other content, would make new part kinds default to "has work".

agent/workflowagent/workflow.go
```diff
@@ -155,33 +143,72 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
...
+	// adk_request_input, and the engine's pause is name-agnostic
+	// (Event.LongRunningToolIDs).
+	//
+	// A response only routes this turn to Resume when it is actually aimed at
+	// this run: either its ID is an interrupt the run knows about, or it is
+	// explicitly an answer to the workflow's own input request. Otherwise the
+	// turn falls through to a fresh Run, as it did before ID-keyed matching —
+	// a turn may legitimately carry both text and an unrelated tool reply, and
+	// routing that to Resume would fail it with ErrNothingToResume and discard
+	// the text. Mirrors runner.buildResumeResponses, which filters the same way.
```
karolpiotrowicz, line 158: The two no longer filter the same way. This one keys on `ActionableInterruptIDs` with a live/spent bit and adds the other-content fallthrough below. `buildResumeResponses` in `runner/run_node.go` keys on waiting-only IDs unioned with the open long-running calls in history, has no notion of spent, and has no fallthrough. A reader trusting this sentence will assume a fix applied here is mirrored there.

agent/workflowagent/workflow.go
```diff
@@ -155,33 +143,72 @@ func (a *workflowAgent) detectResume(ctx agent.InvocationContext) (map[string]an
...
+	//
+	// A wrong-but-deliberate answer (right name, unknown ID) still reaches
```
karolpiotrowicz, line 160: This no longer holds when the message carries anything else. A right-name/unknown-ID answer now gives `ok == false`, leaves `live` false, and — with text alongside — falls through to a fresh run at the check below rather than reaching `Resume`. The behaviour is what I would want, since failing that turn would discard the text. It is the sentence that needs the qualifier.

internal/configurable/configurable_workflow.go
```diff
@@ -331,8 +341,13 @@ func resolveNodeFromYAML(ctx context.Context, parentPath, ref, absPath string) (
 		*collector = append(*collector, ag)
 	}
 
+	var nodeCfg agentNodeYAMLConfig
+	if err := yaml.Unmarshal(data, &nodeCfg); err != nil {
+		return nil, fmt.Errorf("failed to parse agent node config %q: %w", absPath, err)
```
karolpiotrowicz, line 346: This is the one line the increment adds that no test executes. It fires only when `rerun_on_resume` is present with a non-boolean value, and the new table covers `true`, `false` and absent.

Hard-failing the load is the right call — it matches what the FunctionNode and ToolNode branches already do with a malformed node config, and the alternative is the silent-default bug this commit exists to fix. Worth a case pinning it, and worth one that goes through the public loader rather than calling `resolveNodeFromYAML` directly, since the direct call skips the node cache the new type's doc comment reasons about.

karolpiotrowicz (changes requested):
Two things need fixing before this lands, and one of them is my fault from last round. Everything else here is small.

Before merge:

1. A replayed approval must not re-run an `LlmAgent` node that has already settled. It does today, and the reproduction below runs through the real runner.
2. A turn that carries new user text must not be swallowed because it also echoes an answer the run has already acted on. Today it is, in two shapes — one fails the turn with `ErrNothingToResume`, the other produces no events and no error at all.

All three behaviours I called worse than `main` last time are fixed, and each fix is pinned by a test that fails without this round's commits and passes with them ([8d7ca93f](https://github.com/google/adk-go/commit/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043)) — I ran all six individually rather than trusting the suite. The partly-answered handoff node stays waiting, a resume that emits and then fails is retryable again, and a replayed approval carrying new text no longer destroys the turn. The YAML knob, the partial-resume replay guard and the `internal/workflowstate` cut are all good, and I like that the last one keeps the workflow package's public surface closed. Build, the full suite, `go vet`, `golangci-lint`, `-race` on the touched packages and an exported-API compatibility check are clean on the merge result, with a positive control made to fire for each.

## The consumed sweep never fires for an LlmAgent node, so a replayed approval charges twice

This is the cut I asked for last round, and I gave you the wrong field.

[`persistence.go:200`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/persistence.go#L200) keys on `ev.Output != nil`. An `AgentNode` over an `LlmAgent` settles by emitting model text: [`agent_node.go:331-347`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/agent_node.go#L331-L347) stamps `Output` and `NodeInfo.MessageAsOutput` on that event, and then [`runner.go:740-745`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/runner/runner.go#L740-L745) and [`run_node.go:212-218`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/runner/run_node.go#L212-L218) clone it with `clone.Output = nil` before `AppendEvent`. What lands in history has `Output == nil` and no `LongRunningToolIDs`, so the answer is never marked consumed, `reentryConsumed` stays false, the guard at [`resume.go:153`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/resume.go#L153) does not fire, and the node runs again.

That is the default HITL configuration — `RerunOnResume` defaults to re-entry for exactly this node kind — and it is the shape `TestWorkflowAgent_Resume_ReentryIsIdempotent` is meant to protect. The reason the test does not catch it is that its fixtures set `Output` explicitly and `drainAgent` appends events verbatim, so neither the fixture nor the harness ever exercises the field the runner strips.

Measured through the real runner: the approved side effect runs twice with this head applied ([8d7ca93f](https://github.com/google/adk-go/commit/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043)), once at the previous round's head, and twice on `main`. So this is not a regression against `main` — `main` has no replay guard at all — but the guard this PR adds does not cover the population the PR is written for.

The property a fix needs to keep: an answer is consumed exactly when the node settled, decided on the event shape the runner actually writes. [`agent_node.go:383-393`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/agent_node.go#L383-L393) is the recovery path for that, and [`persistence.go:252`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/persistence.go#L252) already uses it two screens away for the same reason. I did run one variant before suggesting it — swapping the condition for `if _, hasOut := childEventOutput(ev); hasOut || len(ev.LongRunningToolIDs) > 0` makes the reproduction pass, keeps `TestWorkflowAgent_Resume_FailedActivationStaysResumable` and the four touched packages green, and every mutation I could make of that new condition is caught by your existing tests. Take it or cut it differently, but the field is the thing to change.

<details>
<summary>Reproduction — <code>go test -run TestReplayedApprovalReRunsTextCompletingNode ./runner/</code></summary>

```go
// runner package. The agent settles on a plain final model-text event and sets no
// explicit Output, which is what an LlmAgent does.
func reproTextCompletingAgent(t *testing.T, name, fcID string, runs *atomic.Int32) agent.Agent {
		Name: name,
				if !reproAnsweredInHistory(ctx.Session(), fcID) {
					req := session.NewEvent(ctx, ctx.InvocationID())
					req.Author = name
					req.LongRunningToolIDs = []string{fcID}
					req.LLMResponse = model.LLMResponse{Content: &genai.Content{
						Role: genai.RoleModel,
						Parts: []*genai.Part{{FunctionCall: &genai.FunctionCall{
							ID: fcID, Name: "adk_request_credential",
						}}},
					}}
					yield(req, nil)
					return
				}
				runs.Add(1) // the side effect the human authorised, once
				done := session.NewEvent(ctx, ctx.InvocationID())
				done.Author = name
				done.LLMResponse = model.LLMResponse{Content: &genai.Content{
					Role:  genai.RoleModel,
					Parts: []*genai.Part{{Text: "charged"}},
				}}
				yield(done, nil)
			}
		},
	})
	if err != nil {
		t.Fatalf("agent.New: %v", err)
	}
	return a
}

func TestReplayedApprovalReRunsTextCompletingNode(t *testing.T) {
	ctx := t.Context()
	const fcID = "cred-1"
	var runs atomic.Int32

	rerun := true
	node, err := workflow.NewAgentNode(
		reproTextCompletingAgent(t, "worker", fcID, &runs),
		workflow.NodeConfig{RerunOnResume: &rerun},
	)
	if err != nil {
		t.Fatalf("NewAgentNode: %v", err)
	}
	r := newWorkflowRunner(t, workflow.Chain(workflow.Start, node))

	turn1 := drainRunner(t, r.Run(ctx, nodeTestUser, nodeTestSession, userText("go"), agent.RunConfig{}))
	if id, _ := findLongRunningInterrupt(turn1); id != fcID {
		t.Fatalf("turn 1 did not pause on %q", fcID)
	}

	drainRunner(t, r.Run(ctx, nodeTestUser, nodeTestSession,
		resumeContent(fcID, "adk_request_credential", "approved"), agent.RunConfig{}))
	if got := runs.Load(); got != 1 {
		t.Fatalf("side effect ran %d time(s) after the approval, want 1", got)
	}

	// The client re-submits the identical approval: a double submit, a retry, an echo.
	for range r.Run(ctx, nodeTestUser, nodeTestSession,
		resumeContent(fcID, "adk_request_credential", "approved"), agent.RunConfig{}) {
	}
	if got := runs.Load(); got != 1 {
		t.Errorf("side effect ran %d time(s) after a replayed approval, want 1", got)
	}
}
```

```
--- FAIL: TestReplayedApprovalReRunsTextCompletingNode
    side effect ran 2 time(s) after a replayed approval, want 1
```
</details>

## A spent answer reads as live on a handoff node, and the turn carrying it is lost

`inferNodeState` sets `reentryConsumed` in its two re-entry arms, [`persistence.go:385`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/persistence.go#L385) and [`persistence.go:401`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/persistence.go#L401), but not in the `NodeWaiting` arm at [`persistence.go:386-392`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/persistence.go#L386-L392) — which is the arm your new partly-answered handoff node lands in. [`state.go:221`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/state.go#L221) reads that field for every node regardless of mode, so a spent answer on such a node is published as live.

Two shapes fall out of it, and the second is the one I would fix first.

**Partly answered.** Node raised `confirm-a` and `confirm-b`, `confirm-a` was rejected on turn 2. On turn 3 a client echoes that rejection alongside "actually, cancel everything". Because the ID reads live, the `!live && carriesOtherContent` escape at [`workflow.go:185`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/agent/workflowagent/workflow.go#L185) is skipped, the turn routes to `Resume`, which finds nothing to schedule and fails with `ErrNothingToResume`. The instruction is gone. This is the same class as `TestWorkflowAgent_SettledReplyDoesNotDiscardNewText`, on the node shape that test does not cover.

**Fully answered, one ID echoed back.** Worse, because nothing is reported at all. The sibling ID that was answered and is not re-echoed keeps `scan.resolvedCount[id] == 1` forever, so `answeredThisTurn` at [`persistence.go:367-372`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/persistence.go#L367-L372) never goes false again, so the completed node passes the gate at [`state.go:213`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/state.go#L213) on every subsequent turn and its IDs read live. The turn routes to `Resume`, schedules nothing, and yields **zero events and no error**. Echoing *both* settled IDs in the same message makes it work, which is what isolates the cause.

<details>
<summary>Reproduction — both shapes, <code>go test -run 'TestPartlyAnswered|TestFullySettled' ./agent/workflowagent/</code></summary>

Built on your own `newTwoConfirmAgent` fixture, in package `workflowagent`.

```go
// Shape 1: one of two confirmations answered on turn 2, echoed with new text on turn 3.
func TestPartlyAnsweredHandoffEchoPlusTextIsLost(t *testing.T) {
	const idA, idB = "confirm-a", "confirm-b"
	var successorRuns atomic.Int32
	worker, err := workflow.NewAgentNode(
		newTwoConfirmAgent(t, "approve_payment", idA, idB), workflow.NodeConfig{})
	if err != nil {
		t.Fatalf("NewAgentNode: %v", err)
	}
	guarded := workflow.NewFunctionNode("execute_payment",
		func(_ agent.Context, _ any) (string, error) {
			successorRuns.Add(1)
			return "paid", nil
		}, workflow.NodeConfig{})
	a := makeAgent(t, workflow.Chain(workflow.Start, worker, guarded))
	sess := newFakeSession()
	runFreshTurn(t, sess, a, "pay 500")

	reject := &genai.Content{Role: genai.RoleUser, Parts: []*genai.Part{{
		FunctionResponse: &genai.FunctionResponse{
			ID: idA, Name: "adk_request_confirmation",
			Response: map[string]any{"confirmed": false},
		},
	}}}
	drainAgent(t, sess, a.Run(newMockCtx(sess, a, reject)), nil)

	echoPlusText := &genai.Content{Role: genai.RoleUser, Parts: []*genai.Part{
		{Text: "actually, cancel everything"},
		{FunctionResponse: &genai.FunctionResponse{
			ID: idA, Name: "adk_request_confirmation",
			Response: map[string]any{"confirmed": false},
		}},
	}}
	for _, err := range a.Run(newMockCtx(sess, a, echoPlusText)) {
		if err != nil {
			t.Errorf("turn carrying new text failed with %v; the text was discarded", err)
			break
		}
	}
}
```

```
--- FAIL: TestPartlyAnsweredHandoffEchoPlusTextIsLost
    turn carrying new text failed with workflow: no waiting node matched the
    supplied responses; the text was discarded
```

Shape 2 is the same test with both confirmations answered on turn 2. Counting events on the third turn:

```
echo ONE settled id + text  -> 0 events, err=<nil>, successor ran 1x   (the instruction vanished)
echo BOTH settled ids + text -> 1 event,  err=<nil>, successor ran 1x   (the turn runs)
```

A single-interrupt handoff node behaves the same with this head applied, at the previous round's head, and on `main` — a fresh run, 2 events — so this is specific to a node that raised more than one interrupt.
</details>

The property a fix needs to keep: an ID is live only when `Resume` can act on it this turn. That is per-ID, and `answeredThisTurn` is a per-node flag that never goes false again once any one of the node's answers has appeared exactly once in history, so it cannot express it.

## Things I looked at and am not asking you to change

The settled-reply fallthrough at [`workflow.go:185`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/agent/workflowagent/workflow.go#L185) restarts the graph from `Start`, so an upstream node with a side effect runs a second time — I measured it at 2x with a counting node upstream of the paused one. But `TestWorkflowAgent_UnmatchedFunctionResponseRunsFresh` already pins upstream re-running on a fresh non-resume turn as intended, so this is the engine's existing semantics for the path you are routing into rather than something the change introduces. Worth knowing the boundary is there.

The nil-by-default hook var in `internal/workflowstate` is safe: `agent/workflowagent` imports `workflow`, so Go completes `workflow`'s `init` first and the var cannot be observed nil. Only the doc sentence promising a nil return meaning "cannot tell" describes something no caller can reach.

File the schema half separately, as you offered. Same for `samePathActivation` — I agree with not landing a path change you cannot demonstrate. Thank you for filing #1552 and #1553, and for naming the final-text boundary in `synthesizeAgentOutput` — an agent whose text lands and which then fails counts as settled, and I would rather read that in the thread than find it later.

One note on the `-race` green, since it is easy to over-read on this diff: [`state.go:197-224`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/state.go#L197-L224) has no test reference anywhere in the tree and the new branch at [`resume.go:174-188`](https://github.com/google/adk-go/blob/8d7ca93f5caaff9aece169a9bc3ba1baf4cf9043/workflow/resume.go#L174-L188) starts no goroutine, so the detector reached almost none of the new code. It is not evidence either way here.

PR #1185, opened 2026-07-22
fix(session): maintain app and user state update times

### LINKED ISSUE #1177: session/database: persisting app/user state fails on MySQL (strict mode) — zero-value update_time
## 🔴 Required Information

**Describe the Bug:**

The GORM-backed session service (`session/database`) fails on MySQL whenever it
persists app-scoped or user-scoped state (`app:` / `user:` prefixed state keys).
The `update_time` column of `app_states` / `user_states` is never assigned, so Go's
zero `time.Time` is written explicitly, which MySQL rejects under its default strict
mode (`NO_ZERO_DATE` / `STRICT_TRANS_TABLES`). SQLite, PostgreSQL and Spanner accept
the zero value, so the bug only surfaces on MySQL — which is one of the dialects
explicitly supported by `gorm_datatypes.go` (`stateMap` maps to `LONGTEXT` for MySQL).

**Steps to Reproduce:**

1. Start a MySQL 8.x/9.x server with default settings (strict mode on).
2. `svc, _ := database.NewSessionService(mysql.Open(dsn))` + `database.AutoMigrate(svc)`.
3. Either create a session with `app:`/`user:` scoped initial state, **or** append any
   event whose actions carry an `app:`/`user:` state delta (e.g. a tool doing
   `ctx.State().Set("user:pref", "x")` in a runner configured with this service).
4. The write fails with Error 1292 (see Observed Behavior).

**Expected Behavior:**

App/user state is persisted, with `update_time` maintained automatically — matching
adk-python, where the same tables define ORM-level auto timestamps
(`sessions/schemas/v1.py`):

```python
class StorageAppState(Base):
    update_time: Mapped[datetime] = mapped_column(
        PreciseTimestamp, default=func.now(), onupdate=func.now()
    )
```

**Observed Behavior:**

```
failed to save app state: Error 1292 (22007): Incorrect datetime value: '0000-00-00' for column 'update_time' at row 1
```

Generated SQL (from GORM logs):

```sql
INSERT INTO `app_states` (`app_name`,`state`,`update_time`)
VALUES ('myapp','{"config":"v1"}','0000-00-00 00:00:00')
ON DUPLICATE KEY UPDATE `state`=VALUES(`state`),`update_time`=VALUES(`update_time`)
```

**Environment Details:**

- ADK Library Version: v2, `main` @ `020d941` (2026-07-17); gorm.io/gorm v1.31.0, gorm.io/driver/mysql v1.6.0
- OS: macOS (reproduces on Linux as well — server side is MySQL 9.4 in Docker, also reproduces on MySQL 8.x)
- Go Version: 1.25

**Model Information:**

- N/A — model-independent; fails at the session persistence layer.

---

## 🟡 Optional Information

**Regression:**

No — present since the service was introduced (#177). The field was only touched once
since (#414, added `precision:6`), never assigned.

**Logs:**

```text
2026/07/19 17:33:48 .../session/database/service.go:400
Error 1292 (22007): Incorrect datetime value: '0000-00-00' for column 'update_time' at row 1
[0.999ms] [rows:0] INSERT INTO `app_states` (`app_name`,`state`,`update_time`)
VALUES ('dataagent','{"total_output_tokens":71,"total_prompt_tokens":1574}','0000-00-00 00:00:00')
ON DUPLICATE KEY UPDATE `state`=VALUES(`state`),`update_time`=VALUES(`update_time`)
```

```go
svc, _ := database.NewSessionService(mysql.Open(dsn))
_ = database.AutoMigrate(svc)

// Fails:
_, err := svc.Create(ctx, &session.CreateRequest{
    AppName: "myapp", UserID: "u1",
    State: map[string]any{"app:config": "v1"},
})
// err: failed to save app state: Error 1292 (22007): Incorrect datetime value:
//      '0000-00-00' for column 'update_time' at row 1
```

- Always (100%) — deterministic on MySQL with default `sql_mode`.

**Additional Context:**

Root cause: `storageAppState` / `storageUserState`
(`session/database/storage_session.go:366-384`) declare `UpdateTime` with no
auto-update tag:

```go
type storageAppState struct {
    AppName    string `gorm:"primaryKey;"`
    State      stateMap
    UpdateTime time.Time `gorm:"precision:6"`
}
```

and neither write path assigns it before `tx.Save(...)`:

- `Create` → `service.go:98-118` (initial `app:`/`user:` state)
- `AppendEvent` → `service.go:385-407` (state deltas from events)

A DDL-level `DEFAULT CURRENT_TIMESTAMP` would not help: the column is explicitly
present in the generated statement, so the explicit (invalid) value always wins over
any column default. For comparison, `storageSession.UpdateTime` *is* maintained
(`service.go:57` on create, `service.go:424` on append), which is why only the two
state tables are affected. The field is also never read anywhere in the Go
implementation, and the database test suite runs on in-memory SQLite only, which is
why the MySQL failure went unnoticed.

This looks like a port gap rather than intended behavior — adk-python auto-maintains
the same column via `default=func.now(), onupdate=func.now()` (see Expected Behavior),
which is exactly what GORM's `autoUpdateTime` provides.

Suggested fix — either:

1. Tag the fields so GORM fills them on insert/update, mirroring adk-python:

   ```go
   UpdateTime time.Time `gorm:"precision:6;autoUpdateTime:micro"`
   ```

2. Or assign explicitly before `Save`, mirroring what `AppendEvent` already does for
   the session row (`storageSess.UpdateTime = event.Timestamp`), keeping the three
   tables' timestamps consistent within one transaction.

Workaround: relax the server's `sql_mode` (drop `NO_ZERO_DATE`/strict) — usable for
dev, not great for shared/production MySQL instances.

Related: #409 fixed a different MySQL timestamp issue in the same service
(sub-millisecond precision on the `sessions` table causing false stale-session
errors); #699 touches the same file area but is a table-naming feature request.

Happy to send a PR for either variant if useful.

ZYZ666-RGB:

- Closes: #1177

**Problem:**

The GORM-backed session service writes a zero `time.Time` to `app_states.update_time` and `user_states.update_time`. MySQL rejects that explicit zero date in its default strict mode, so persisting `app:` or `user:` state fails.

**Solution:**

Enable GORM-managed update timestamps for `storageAppState` and `storageUserState`, while retaining microsecond column precision.

The regression test covers both:

- initial app/user state persistence through `Create`
- subsequent app/user state updates through `AppendEvent`

No public API or dependency changes.

### Testing Plan

- [x] Added an offline regression test.
- [x] Verified the test fails on the previous code with zero timestamps.
- [x] All tests pass locally.

```sh
go build -mod=readonly ./...
go test -race -mod=readonly -count=1 ./session/database
go test -race -mod=readonly -count=1 -shuffle=on ./...
golangci-lint run
go mod tidy -diff
```

`golangci-lint` reported `0 issues`; `go mod tidy -diff` produced no output.

### Checklist

- [x] I have performed a self-review.
- [x] I have added tests that prove the fix works.
- [x] New and existing tests pass locally.

### DIFF
--- a/session/database/service_test.go
+++ b/session/database/service_test.go
@@ -51,6 +51,63 @@ func Test_databaseService_CreateUsesProviders(t *testing.T) {
 	}
 }
 
+func Test_databaseService_AppAndUserStateUpdateTimes(t *testing.T) {
+	dbNow := time.Date(2026, time.July, 22, 1, 2, 3, 456000000, time.UTC)
+	s := emptyService(t)
+	s.db.NowFunc = func() time.Time { return dbNow }
+
+	created, err := s.Create(t.Context(), &session.CreateRequest{
+		AppName: "app",
+		UserID:  "user",
+		State: map[string]any{
+			"app:setting":     "initial",
+			"user:preference": "initial",
+		},
+	})
+	if err != nil {
+		t.Fatalf("Create() error = %v", err)
+	}
+
+	assertStateUpdateTimes(t, s, "app", "user", dbNow)
+
+	dbNow = dbNow.Add(time.Minute)
+	event := &session.Event{
+		ID:        "state-update",
+		Timestamp: created.Session.LastUpdateTime().Add(time.Second),
+		Actions: session.EventActions{
+			StateDelta: map[string]any{
+				"app:setting":     "updated",
+				"user:preference": "updated",
+			},
+		},
+	}
+	if err := s.AppendEvent(t.Context(), created.Session, event); err != nil {
+		t.Fatalf("AppendEvent() error = %v", err)
+	}
+
+	assertStateUpdateTimes(t, s, "app", "user", dbNow)
+}
+
+func assertStateUpdateTimes(t *testing.T, s *databaseService, appName, userID string, want time.Time) {
+	t.Helper()
+
+	var appState storageAppState
+	if err := s.db.First(&appState, "app_name = ?", appName).Error; err != nil {
+		t.Fatalf("fetch app state: %v", err)
+	}
+	if !appState.UpdateTime.Equal(want) {
+		t.Errorf("app state UpdateTime = %v, want %v", appState.UpdateTime, want)
+	}
+
+	var userState storageUserState
+	if err := s.db.First(&userState, "app_name = ? AND user_id = ?", appName, userID).Error; err != nil {
+		t.Fatalf("fetch user state: %v", err)
+	}
+	if !userState.UpdateTime.Equal(want) {
+		t.Errorf("user state UpdateTime = %v, want %v", userState.UpdateTime, want)
+	}
+}
+
 // TestDatabaseService_AppendEvent_WorkflowFieldsRoundTrip guards that
 // the storage layer serializes/deserializes the workflow event fields
 // (NodeInfo, RequestedInput, Routes, IsolationScope); dropping them
--- a/session/database/storage_session.go
+++ b/session/database/storage_session.go
@@ -366,7 +366,7 @@ func createEventFromStorageEvent(se *storageEvent) (*session.Event, error) {
 type storageAppState struct {
 	AppName    string `gorm:"primaryKey;"`
 	State      stateMap
-	UpdateTime time.Time `gorm:"precision:6"`
+	UpdateTime time.Time `gorm:"precision:6;autoUpdateTime"`
 }
 
 // TableName explicitly sets the table name for the AppState struct.
@@ -379,7 +379,7 @@ type storageUserState struct {
 	AppName    string `gorm:"primaryKey;"`
 	UserID     string `gorm:"primaryKey;"`
 	State      stateMap
-	UpdateTime time.Time `gorm:"precision:6"`
+	UpdateTime time.Time `gorm:"precision:6;autoUpdateTime"`
 }
 
 // TableName explicitly sets the table name for the UserState struct.


2evl1u:
Verified this PR against the original repro from #1177. Environment: MySQL 9.4 in
Docker with default `sql_mode` (`STRICT_TRANS_TABLES,NO_ZERO_DATE,...`), Go 1.25, macOS.

- **Control on `main` @ `020d941`** (same MySQL): both failure paths still reproduce —
  `Create` with `app:`/`user:` scoped initial state, and `AppendEvent` carrying an
  `app:`/`user:` state delta both fail with
  `Error 1292 (22007): Incorrect datetime value: '0000-00-00' for column 'update_time'`.
- **With this PR (`7ed78c4`)**: both paths succeed under strict mode, and both tables
  now get real timestamps:

  ```
  app_states:  pr1185check | 2026-07-23 10:11:47.521000
  user_states: pr1185check | u1 | 2026-07-23 10:11:47.521000
  ```

- `go test ./session/database/` passes locally, including the new regression test.
- Also ran a full agent app (multi-turn, tools, app-state writes from a plugin)
  end-to-end against MySQL strict mode on this branch — clean.
- One expected note: rows written before the fix keep their zero `update_time`
  (forward-only, no migration) — seems fine since nothing reads the column yet.

Thanks for the quick fix @ZYZ666-RGB!

ZYZ666-RGB:
Hi maintainers, just following up on this PR. The independent verification above reproduced both failure paths on main with MySQL strict mode and confirmed that both succeed with this change. Please let me know if any additional changes or testing are needed. Thanks!

wolo-lab:
Thanks for the fix. It's solved by https://github.com/google/adk-go/pull/1204.

PR #1186, opened 2026-07-22
fix(parallelagent): await sub-agent teardown before returning on early stop

wolo-lab:
## Problem
When `parallelagent` runs its sub-agents and the consumer stops iterating early — a sibling produced a terminal result, or the invocation was cancelled / deadline-exceeded — the returned iterator closed `doneChan` and returned immediately, **without joining the sub-agent goroutines** it spawned via `errgroup`.
A sub-agent's deferred teardown therefore keeps running after the parallel agent (and the runner above it) have already returned. The concrete case: a remote A2A sub-agent (`remoteagent/v2`) runs `cleanupRemoteTask` in a `defer`, which issues a `CancelTask` RPC (up to 5s) on `context.WithoutCancel(ctx)`. That RPC outlives the run and still reads request-scoped context values.
In google3 the a2a-go server executes each task under a detached context that invalidates `.Value()` lookups once its execution callback returns. The leaked cleanup RPC then reads `census` from that already-ended context and panics (`detach: Value called for key census after callback passed to detach.Go has already returned`), which cascades into a wave of `DEADLINE_EXCEEDED` under load.
`parallelagent` is the only goroutine-spawning primitive in `agent/` and `runner/`, so this is the single place a sub-agent's teardown can escape the run's lifetime.

## Summary
- The iterator now drains `resultsChan` in its deferred cleanup. The funnel goroutine closes `resultsChan` only after `errGroup.Wait()`, so draining blocks until every sub-agent goroutine — and its deferred teardown — has returned.
- Deadlock-free: on the normal path the outer loop already exited because the funnel closed the channel (the drain returns at once); on early stop, `close(doneChan)` unblocks the sub-agents, they run teardown and return, the funnel closes the channel, and the drain returns.
- Behavioral note: an early stop now **blocks until sub-agent teardown completes**, bounded by whatever the sub-agent bounds itself to (e.g. remoteagent's 5s cancel timeout).

### DIFF
--- a/agent/workflowagents/parallelagent/agent.go
+++ b/agent/workflowagents/parallelagent/agent.go
@@ -68,8 +68,9 @@ func New(cfg Config) (agent.Agent, error) {
 func run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
 	curAgent := ctx.Agent()
 
-	// Cancelable so an early consumer stop aborts the sub-agents promptly instead
-	// of blocking until they next yield (see the iterator's deferred cleanup).
+	// Cancelable so an early consumer stop aborts each sub-agent's in-flight Run
+	// promptly. Teardown still runs to completion and the iterator waits for it
+	// (see the deferred cleanup), so a sub-agent that blocks in teardown blocks run.
 	subAgentsCtx, cancelSubAgents := context.WithCancel(ctx)
 
 	var (
@@ -116,9 +117,9 @@ func run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
 
 	return func(yield func(*session.Event, error) bool) {
 		// Await sub-agent goroutines (incl. their deferred teardown) before
-		// returning, even on early stop: cancel them so they abort promptly, then
-		// drain resultsChan — the funnel closes it only after errGroup.Wait().
-		// Otherwise teardown can outlive the run and touch an already-ended context.
+		// returning, even on early stop: cancel them, then drain resultsChan, which
+		// the funnel closes only after errGroup.Wait(). Otherwise teardown can
+		// outlive the run and touch an already-ended context.
 		defer func() {
 			cancelSubAgents()
 			close(doneChan)
--- a/agent/workflowagents/parallelagent/agent.go
+++ b/agent/workflowagents/parallelagent/agent.go
@@ -16,6 +16,7 @@
 package parallelagent
 
 import (
+	"context"
 	"fmt"
 	"iter"
 
@@ -67,8 +68,12 @@ func New(cfg Config) (agent.Agent, error) {
 func run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
 	curAgent := ctx.Agent()
 
+	// Cancelable so an early consumer stop aborts the sub-agents promptly instead
+	// of blocking until they next yield (see the iterator's deferred cleanup).
+	subAgentsCtx, cancelSubAgents := context.WithCancel(ctx)
+
 	var (
-		errGroup, errGroupCtx = errgroup.WithContext(ctx)
+		errGroup, errGroupCtx = errgroup.WithContext(subAgentsCtx)
 		doneChan              = make(chan bool)
 		resultsChan           = make(chan result)
 	)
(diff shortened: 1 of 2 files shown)


karolpiotrowicz:
Thanks for tracking this down — awaiting sub-agent teardown before the iterator returns is the right call, and the deterministic channel-gated test is really nice.

One behavioral tradeoff worth flagging (I was able to reproduce it): the fix changes an early consumer stop from *return-immediately* into *block until every sub-agent goroutine (including its deferred teardown) has returned*. The drain in the iterator's deferred cleanup —

```go
for range resultsChan { // drain until the funnel closes resultsChan
}
```

— has no `ctx`/timeout escape, and `runSubAgent` only observes `done`/`ctx.Done()` at its `select` points (i.e. at yield boundaries). So if a sub-agent's `Run` blocks between yields while ignoring cancellation, **or** its deferred teardown blocks, `errGroup.Wait()` never returns, `resultsChan` is never closed, and the drain (hence `run`, hence the runner) blocks indefinitely. Before this change that same sub-agent leaked a goroutine in the background but let the caller return — so this trades a background leak for a foreground hang on the misbehaving path.

Minimal repro (hangs after this change; returns immediately on `main`):

```go
func TestParallelAgent_EarlyStop_BlockingTeardown(t *testing.T) {
	blockTeardown := make(chan struct{}) // never closed
	sub := must(agent.New(agent.Config{
		Name: "leaky-teardown",
		Run: func(agent.InvocationContext) iter.Seq2[*session.Event, error] {
				defer func() { <-blockTeardown }() // teardown ignores ctx, blocks forever
				for {
					if !yield(&session.Event{LLMResponse: model.LLMResponse{
						Content: genai.NewContentFromText("ev", genai.RoleModel)}}, nil) {
						return
					}
				}
			}
		},
	}))
	pa := must(parallelagent.New(parallelagent.Config{AgentConfig: agent.Config{
		Name: "p", SubAgents: []agent.Agent{sub}}}))
	r, err := runner.New(runner.Config{AppName: "a", Agent: pa,
		SessionService: session.InMemoryService(), AutoCreateSession: true})
	if err != nil {
		t.Fatal(err)
	}

	done := make(chan struct{})
	go func() {
		defer close(done)
		for _, e := range r.Run(t.Context(), "u", "s",
			genai.NewContentFromText("hi", genai.RoleUser), agent.RunConfig{}) {
			_ = e
			break // early stop after first event
		}
	}()
	select {
	case <-done: // main: fires immediately. With this change: never fires.
		t.Fatal("Run() did not return after early stop")
	}
}
```

Notably this is reachable with the shipped `remoteagent/v2`: `cleanupRemoteTask` runs on `context.WithoutCancel(ctx)`, so the new `cancelSubAgents()` is a no-op for it; its default `CancelTask` path is 5s-bounded (an uncancellable stall on early stop), but the `RemoteTaskCleanupCallback` path has no timeout and `client.Destroy()` takes no context — either can block the drain unboundedly.

I don't think there's a clean drop-in fix: bounding the drain with a timeout/watchdog would let teardown outlive the iterator again, which reintroduces exactly the escaping-teardown case this PR fixes — so the unconditional await is arguably the correct primitive, and any hard bound belongs in the sub-agent's teardown. Given that, my suggestions are lightweight and non-blocking:

- Add a doc comment on `ParallelAgent`/`run` noting that an early stop now blocks until every sub-agent's teardown completes, bounded only by each sub-agent's own teardown bounds (a sub-agent that blocks in teardown or ignores `ctx` will block `Run`). The `// … aborts the sub-agents promptly …` comment is accurate for a sub-agent's active `Run`, but not for teardown that ignores cancellation — worth scoping it to avoid surprise.
- Possibly a follow-up to give `remoteagent/v2` teardown a caller-bounded budget (`client.Destroy()` / the cleanup callback), since that's what turns the bounded wait into a potential hang.
- Optional: a test covering the early-stop-with-in-flight-error case would be a nice complement to the teardown test.

Overall this LGTM as a fix; just wanted to surface the new blocking-on-teardown behavior so it's a documented, intentional contract.

karolpiotrowicz (dismissed):
Approving — this is a solid fix for a real teardown-escape bug, and the deterministic test is great. I left one note above about the new *block-until-teardown* behavior on early stop (reproduced): it's a reasonable, arguably-correct tradeoff, so treating it as non-blocking — just worth a doc comment scoping the "aborts promptly" wording to a sub-agent's active Run, plus a possible follow-up to bound remoteagent/v2 teardown. Nice work.

hanorik (approved):
Restoring previous approval as it become outdated after comment adjustment.

Issue #1076, opened 2026-06-22
Remote A2A task is orphaned when parent task is cancelled before the first remote event


merkulovlad:
**Describe the Bug:**

When an ADK agent delegates execution to a remote A2A agent, cancelling the
parent task before the first remote task event is received can leave the remote
task running.

The remote A2A server creates the child task and assigns its task ID, but the
remote agent proxy learns that ID only from the first task-bearing event.

If the parent task is cancelled before that event arrives:

1. The parent invocation context is cancelled.
2. `SendStreamingMessage` stops reading the remote stream.
3. `lastEvent` remains `nil`.
4. `cleanupRemoteTask` returns without sending `CancelTask`.
5. The child task remains running on the remote server without an active parent.

This creates an orphan remote task and may waste resources indefinitely.

**Steps to Reproduce:**

1. Start A2A Server B with an executor that reports when a child task has been
   created and delays its first task-bearing event.
2. Create a remote A2A agent pointing to Server B.
3. Configure the remote agent as a subagent of the root agent on A2A Server A.
4. Send a streaming message to Server A.
5. Wait until Server B confirms that the child task has been created.
6. Cancel the parent task on Server A before allowing Server B to emit its
   first task-bearing event.
7. Allow Server B to emit the delayed event containing the child task ID.
8. Observe that Server B does not receive `CancelTask` for the child task.

**Expected Behavior:**

Remote task cleanup should remain possible when the parent task is cancelled
before the first remote event arrives.

After the remote task ID becomes available, Server A should send
`CancelTask(childTaskID)` to Server B. Cleanup should be bounded so that an
unresponsive remote server cannot block cancellation indefinitely.

**Observed Behavior:**

The remote task is not cancelled.

`agent/remoteagent/v2/a2a_agent.go` stores the remote task identity in
`lastEvent`. When cancellation happens before the first event,
`cleanupRemoteTask` returns immediately:

```go
if lastEvent == nil {
	return
}
```

Therefore no remote `CancelTask` request is sent.

The deterministic regression test added in my fork fails consistently:

```text
--- FAIL: TestRemoteAgent_CancelsRemoteTaskWhenContextCanceledBeforeFirstEvent
    a2a_agent_test.go: timeout waiting for remote task to be cancelled on remote side
FAIL
```

Test:

https://github.com/merkulovlad/adk-go/blob/fix/a2a-remote-task-cancellation/agent/remoteagent/v2/a2a_agent_test.go#L1212

**Environment Details:**

[...]

PR #1187, opened 2026-07-22
fix(remoteagent): skip best-effort remote cancel on cancel/deadline teardown

wolo-lab:
## Problem
`cleanupRemoteTask` issues a best-effort `CancelTask` RPC for a non-terminal remote A2A task, using `ctx = context.WithoutCancel(ctx)` so the RPC can still go out after the run ended. `WithoutCancel` keeps the parent value chain.
When the invocation was torn down by cancellation or a deadline, that value chain can reference a context whose lifetime has already ended. In google3 the a2a-go server runs task execution under a detached context that invalidates `.Value()` after its callback returns; the in-flight `CancelTask` then reads request-scoped values (`census`) from that dead context → panic → `DEADLINE_EXCEEDED` cascade in production.

Remote-task cancellation is best-effort and adk-go-specific — adk-python does not cancel remote tasks at all.

## Summary
- In the default cleanup path, skip the `CancelTask` RPC when `cause` is `context.Canceled` or `context.DeadlineExceeded` (via `errors.Is`, so wrapped causes are caught too).
- A custom `RemoteTaskCleanupCallback` is left untouched — it still receives `cause` and decides for itself.
- Trade-off: on the cancel/deadline path the remote task is no longer proactively cancelled and will expire on its own (via the remote server's timeout). This is the best-effort contract; it's the same early-cancel window discussed in #1076.
- This is a targeted safety net; the structural fix for the underlying goroutine-lifetime race is #1186. They compose as defense-in-depth.

### DIFF
--- a/agent/remoteagent/v2/a2a_agent.go
+++ b/agent/remoteagent/v2/a2a_agent.go
@@ -337,7 +337,7 @@ func cleanupRemoteTask(ctx context.Context, cfg A2AConfig, card *a2a.AgentCard, 
 	// torn down by cancellation or a deadline. The context lifetime is unreliable
 	// during forced teardown (a detached parent may already have ended), and
 	// adk-python does not cancel remote tasks at all; the remote task expires on its own.
-	if cause != nil && (errors.Is(cause, context.Canceled) || errors.Is(cause, context.DeadlineExceeded)) {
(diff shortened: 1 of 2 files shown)


wolo-lab:
it's not a right solution.

PR #1202, opened 2026-07-25
fix(session/database): format stale-session timestamps with UnixMicro

### LINKED ISSUE #1170: session/database: stale-session error formats timestamps as 1970-01-21 (UnixMicro passed to time.Unix nsec arg)
### Summary

The error message for `stale session error` in `session/database/service.go` renders both timestamps as ~1970-01-21 instead of the actual wall-clock time, because it passes `time.Time.UnixMicro()` (microseconds since epoch) into `time.Unix(0, X)` whose second argument is nanoseconds. Off by 1000×.

### Reproduction

Any code path that triggers optimistic-concurrency rejection on a session update in the current year. Observed on `v1.2.0` and confirmed present in `v1.5.0`. Example operator-facing log line:

```
failed to append event to sessionService: stale session error:
  last update time from request (1970-01-21T15:38:42.372056389Z)
  is older than in database    (1970-01-21T15:38:42.372500103Z)
```

### The math

- Current unix time (2026-07-17): ~1,784,322,372 seconds since epoch → ~1,784,322,372,000,000 microseconds
- `time.Unix(0, 1_784_322_372_000_000)` treats the second arg as nanoseconds
- 1,784,322,372,000,000 nanoseconds = 1,784,322 seconds = **1970-01-21T15:38:42** ✓ (matches observed)

The comparison itself on the line above (`if storageUpdateTime > sessionUpdateTime`) is fine — both sides are in the same unit, so stale detection works correctly. **Only the error message formatting is broken.** Impact: operators debugging concurrency issues can't correlate the reported timestamps with anything else in their logs.

### Location

`session/database/service.go`, lines ~378-381 (same in v1.2.0, v1.5.0):

```go
storageUpdateTime := storageSess.UpdateTime.UnixMicro()  // microseconds
sessionUpdateTime := session.updatedAt.UnixMicro()       // microseconds
    return fmt.Errorf(
        "stale session error: last update time from request (%s) is older than in database (%s)",
        time.Unix(0, sessionUpdateTime).Format(time.RFC3339Nano),  // ❌ nsec arg gets microseconds
        time.Unix(0, storageUpdateTime).Format(time.RFC3339Nano),  // ❌ same
    )
}
```

### Suggested fix

Use `time.UnixMicro` which correctly interprets microseconds since epoch:

```go
time.UnixMicro(sessionUpdateTime).Format(time.RFC3339Nano),
time.UnixMicro(storageUpdateTime).Format(time.RFC3339Nano),
```

Or convert to nanoseconds explicitly if the file convention prefers `time.Unix`:

```go
time.Unix(0, sessionUpdateTime*1000).Format(time.RFC3339Nano),
time.Unix(0, storageUpdateTime*1000).Format(time.RFC3339Nano),
```

`time.UnixMicro` is more legible; both produce the correct output.

### Additional context

The nearby comment "We use `UnixMicro()` for microsecond-level precision, matching the Python code" makes the *comparison* unit choice explicit — the intent is unambiguous. Only the error-formatting call site missed the unit-mismatch when constructing display strings.

Happy to send a PR if useful — the fix is a two-line change.

PratikDhanave:
## Summary

Fixes #1170.

The stale-session error in `session/database/service.go` rendered both timestamps via `time.Unix(0, micros)`, whose second argument is **nanoseconds**. A `UnixMicro()` value was therefore interpreted 1000× too small and printed as `~1970-01-21`, e.g.:

```
stale session error: last update time from request (1970-01-21T15:38:42...Z)
  is older than in database (1970-01-21T15:38:42...Z)
```

The optimistic-concurrency comparison itself was always correct — both sides use `UnixMicro()` — so stale detection worked; only the operator-facing message was wrong, which made concurrency errors impossible to correlate with other logs.

## Change

- Use `time.UnixMicro(...)` for the two formatted timestamps, which interprets the value as microseconds since the epoch and names the unit at the call site (consistent with the adjacent `UnixMicro()` comparison).

## Testing Plan

- Added `TestDatabaseService_AppendEvent_StaleSessionErrorFormatsTimestamps`, which triggers the stale path via two handles to the same row (pinning `Create`'s clock with `platform.WithTimeProvider`) and asserts the message renders the real wall-clock date and does **not** contain `1970`.
- Verified the test is a genuine regression guard: it **passes** with the fix and **fails** against the pre-fix code, reproducing the exact `1970-01-21...` output.
- `go test -race -count=1 ./session/database/` — pass
- `go vet ./session/database/` — pass
- `golangci-lint run ./session/database/` — 0 issues
- `gofmt -l` — clean

### DIFF
--- a/session/database/service.go
+++ b/session/database/service.go
@@ -376,8 +376,8 @@ func (s *databaseService) applyEvent(ctx context.Context, session *localSession,
 		if storageUpdateTime > sessionUpdateTime {
 			return fmt.Errorf(
 				"stale session error: last update time from request (%s) is older than in database (%s)",
-				time.Unix(0, sessionUpdateTime).Format(time.RFC3339Nano),
-				time.Unix(0, storageUpdateTime).Format(time.RFC3339Nano),
+				time.UnixMicro(sessionUpdateTime).Format(time.RFC3339Nano),
+				time.UnixMicro(storageUpdateTime).Format(time.RFC3339Nano),
 			)
 		}
 
--- a/session/database/service_test.go
+++ b/session/database/service_test.go
@@ -15,6 +15,7 @@
 package database
 
 import (
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


PratikDhanave:
@googlebot I signed it!

wolo-lab:
Duplicate of #1237

PR #1205, opened 2026-07-25
fix(openaimodel): replay assistant turns with output_text content

### LINKED ISSUE #1197: openaimodel: assistant messages sent with `input_text` content type → 400 on multi-turn (should be `output_text`)
**Describe the Bug:**

The OpenAI model connector (`google.golang.org/adk/v2/model/openaimodel`) serializes every message's content with type `input_text`, regardless of role. When a prior assistant turn is replayed in a multi-turn conversation, the OpenAI Responses API rejects the request because assistant message content must use type `output_text`. As a result, every conversation fails starting from the **second** message — on any model, reasoning or not.

**Steps to Reproduce:**

1. Create the minimal agent under "Minimal Reproduction Code" with a valid OpenAI API key.
2. Run `OPENAI_API_KEY=sk-... go run . console`.
3. Send a first message, e.g. `hi` — the assistant replies successfully.
4. Send a second message, e.g. `can you code` — the request fails with HTTP 400.

**Expected Behavior:**

The second message succeeds: the prior assistant turn is replayed with content type `output_text`, and the conversation continues.

**Observed Behavior:**

The first request succeeds, but the second request fails with HTTP 400. The second `input` item (the assistant turn from message #1) is sent with `input_text`, which OpenAI rejects:

```text
AGENT_ERROR: POST "https://api.openai.com/v1/responses": 400 Bad Request {
    "message": "Invalid value: 'input_text'. Supported values are: 'output_text' and 'refusal'.",
    "type": "invalid_request_error",
    "param": "input[1].content[0]",
    "code": "invalid_value"
  }
```

`input[1]` is the assistant turn. The rejection is role/content-type based and is independent of the model.

**Environment Details:**

- ADK Library Version: `google.golang.org/adk/v2 v2.1.0` (latest; `v2.0.0` and `v2.1.0` are the only published v2 releases)
- OS: Windows 11 (OS-independent)
- Go Version: 1.26.5

**Model Information:**

- Which model is being used: Observed with `gpt-5.4-nano`; also reproduces with non-reasoning models such as `gpt-4o-mini`. The bug is in request construction, not model-specific.

---

**Regression:**

Not tested on `v2.0.0`. `v2.1.0` is the latest release; the hardcoded `input_text` content type appears to have been present since the OpenAI connector was introduced.

**Logs:**

```text
Agent -> Hi there! 👋 How can I help you today?
User -> can you code
Agent -> 
AGENT_ERROR: POST "https://api.openai.com/v1/responses": 400 Bad Request {
    "message": "Invalid value: 'input_text'. Supported values are: 'output_text' and 'refusal'.",
    "type": "invalid_request_error",
    "param": "input[1].content[0]",
    "code": "invalid_value"
  }
```

**Screenshots / Video:**

N/A

**Additional Context:**

Root cause is in `model/openaimodel/request.go` (v2.1.0):

- `convertContents` routes every role through `newMessage` and wraps it as `ResponseInputItemUnionParam{OfMessage: msg}` (line 103), i.e. an *easy input* message.
- `newMessage` hardcodes the content item type as `input_text` (line 169):

```go
textParam := responses.ResponseInputTextParam{
    Text: txt,
    Type: constant.InputText("input_text"), // applied to assistant turns too
}
```

`EasyInputMessageParam`'s content union (`ResponseInputContentUnionParam`) only exposes `OfInputText`/`OfInputImage`/`OfInputFile` — it cannot represent `output_text`. Assistant turns must instead be sent as an output message item whose content uses `output_text` (openai-go's `ResponseInputItemUnionParam.OfOutputMessage` → `ResponseOutputMessageParam` with `OfOutputText` content).

Suggested fix (~15 lines): branch on role — emit `ResponseInputItemUnionParam{OfOutputMessage: ...}` with `output_text` content for `assistant`/`model` role, and keep `input_text` for `user`/`system`/`developer`:

```go
// inside flushText in convertContents
switch curRole {
case genai.RoleModel:
    out, err := newOutputMessage(textParts)
    if err != nil {
        return err
    }
    if out != nil {
        items = append(items, responses.ResponseInputItemUnionParam{OfOutputMessage: out})
    }
default:
    msg, err := newMessage(curRole, textParts)
    if err != nil {
        return err
    }
    if msg != nil {
        items = append(items, responses.ResponseInputItemUnionParam{OfMessage: msg})
    }
}

func newOutputMessage(texts []string) (*responses.ResponseOutputMessageParam, error) {
    content := make([]responses.ResponseOutputMessageContentUnionParam, 0, len(texts))
    for _, txt := range texts {
        if strings.TrimSpace(txt) == "" {
            continue
        }
        content = append(content, responses.ResponseOutputMessageContentUnionParam{
            OfOutputText: &responses.ResponseOutputTextParam{Text: txt}, // Type defaults to "output_text"
        })
    }
    if len(content) == 0 {
        return nil, nil
    }
    // Role defaults to "assistant", Type defaults to "message".
    return &responses.ResponseOutputMessageParam{Content: content}, nil
}
```

Separately, reasoning output items (`encrypted_content`/`phase`) are currently dropped (flattened into `Thought` text parts). That is a distinct concern and does not need to block this fix, but it should eventually be addressed for correct multi-turn with reasoning models.

```go
package main

import (
	"context"
	"log"
	"os"

	openaimodel "google.golang.org/adk/v2/model/openaimodel"
)

func main() {
	ctx := context.Background()
	m, err := openaimodel.NewModel(ctx, "gpt-4o-mini", &openaimodel.ClientConfig{
		APIKey: os.Getenv("OPENAI_API_KEY"),
	})
	if err != nil {
		log.Fatal(err)
	}
		Name:        "repro_agent",
		Model:       m,
		Instruction: "You are a helpful assistant.",
	})
	if err != nil {
		log.Fatal(err)
	}
	l := full.NewLauncher()
	if err := l.Execute(ctx, &launcher.Config{AgentLoader: agent.NewSingleLoader(a)}, os.Args[1:]); err != nil {
		log.Fatal(err)
	}
}
```

Run with `OPENAI_API_KEY=sk-... go run . console`, then send two messages.

- Always (100%)

PratikDhanave:
## Summary

Fixes #1197.

`convertContents` (`model/openaimodel/request.go`) routed **every** role through `newMessage`, which hardcodes the content item type as `input_text`. When a prior assistant turn is replayed in a multi-turn conversation, the OpenAI Responses API rejects it — assistant content must use `output_text` — so every conversation fails from the **second** message onward, on any model:

```
400 Bad Request: Invalid value: 'input_text'. Supported values are:
'output_text' and 'refusal'. (param: input[1].content[0])
```

`EasyInputMessageParam`'s content union (`ResponseInputContentUnionParam`) can only represent `input_text`/`input_image`/`input_file`, so an assistant turn must instead be sent as an **output message** item.

## Change

Branch `flushText` on role:
- `genai.RoleModel` → new `newOutputMessage` helper builds a `ResponseOutputMessageParam` with `output_text` content (`Status: completed`; `ID` left unset, since OpenAI mints message IDs on output and a replayed turn has none to echo), appended as `OfOutputMessage`.
- `user` / `system` / `developer` → unchanged easy input message with `input_text`.

## Testing Plan

- Added `TestBuildOpenAIParams_MultiTurnAssistantUsesOutputText`: a user → assistant → user conversation. Asserts the assistant turn becomes an `OfOutputMessage` with `output_text` content, user turns stay `OfMessage` with `input_text`, and — crucially — that the assistant item **marshals to JSON** containing `"output_text"` and never `"input_text"` (the actual wire format sent to OpenAI).
- Verified it's a genuine regression guard: **passes** with the fix, **fails** against the pre-fix code (assistant turn was an `OfMessage`, `OfOutputMessage == nil`).
- Confirmed the marshaled request body:
  ```json
  { "content": [{ "text": "hello there", "type": "output_text" }],
    "status": "completed", "role": "assistant", "type": "message" }
  ```
- `go test -race -count=1 ./model/openaimodel/` — pass (full package)

**Note on live verification:** I don't have an OpenAI API key, so I could not exercise the real multi-turn round-trip end-to-end. Verification is offline: the emitted request now matches the `output_text` shape the API documents and that the 400 error demands. @-the-reporter (who has a repro) confirming the live multi-turn flow would be appreciated.

### DIFF
--- a/model/openaimodel/request.go
+++ b/model/openaimodel/request.go
@@ -95,12 +95,26 @@ func convertContents(contents []*genai.Content) (responses.ResponseInputParam, e
 			if len(textParts) == 0 {
 				return nil
 			}
-			msg, err := newMessage(curRole, textParts)
-			if err != nil {
-				return err
-			}
-			if msg != nil {
-				items = append(items, responses.ResponseInputItemUnionParam{OfMessage: msg})
+			// Assistant turns must be replayed as output messages whose content
+			// uses "output_text"; the OpenAI Responses API rejects "input_text"
+			// for the assistant role (EasyInputMessageParam can only emit
+			// input_text). Every other role stays an easy input message.
+			if curRole == genai.RoleModel {
+				msg, err := newOutputMessage(textParts)
+				if err != nil {
+					return err
+				}
+				if msg != nil {
+					items = append(items, responses.ResponseInputItemUnionParam{OfOutputMessage: msg})
+				}
+			} else {
+				msg, err := newMessage(curRole, textParts)
+				if err != nil {
+					return err
+				}
+				if msg != nil {
+					items = append(items, responses.ResponseInputItemUnionParam{OfMessage: msg})
+				}
 			}
 			textParts = textParts[:0]
 			return nil
@@ -181,6 +195,35 @@ func newMessage(role genai.Role, texts []string) (*responses.EasyInputMessagePar
 		Content: responses.EasyInputMessageContentUnionParam{
 			OfInputItemContentList: contentList,
 		},
+	}, nil
+}
+
+// newOutputMessage builds an assistant output message whose content uses the
+// "output_text" type, as required when replaying a prior assistant turn to the
+// OpenAI Responses API. The message ID is left unset: OpenAI assigns message
+// IDs on output, and a replayed turn has none to echo back.
+func newOutputMessage(texts []string) (*responses.ResponseOutputMessageParam, error) {
+	if len(texts) == 0 {
+		return nil, nil
+	}
+	contentList := make([]responses.ResponseOutputMessageContentUnionParam, 0, len(texts))
+	for _, txt := range texts {
+		if strings.TrimSpace(txt) == "" {
+			continue
+		}
+		contentList = append(contentList, responses.ResponseOutputMessageContentUnionParam{
+			OfOutputText: &responses.ResponseOutputTextParam{
+				Text: txt,
+				Type: constant.OutputText("output_text"),
+			},
+		})
+	}
+	if len(contentList) == 0 {
+		return nil, nil
+	}
+	return &responses.ResponseOutputMessageParam{
+		Content: contentList,
+		Status:  responses.ResponseOutputMessageStatusCompleted,
 	}, nil
 }
 
--- a/model/openaimodel/request_test.go
+++ b/model/openaimodel/request_test.go
@@ -15,6 +15,7 @@
 package openaimodel
 
 import (
+	"encoding/json"
 	"errors"
 	"reflect"
 	"strings"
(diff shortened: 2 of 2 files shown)


rwrz:
I can confirm that this fix the issue with gpt models; Please, review it; Current version is broken with OpenAI models

hanorik:
Thanks for this — the diagnosis in #1197 and the fix here are both correct, and the PR description is good. There are some nits I'd like addressed before merge:

- The change introduces a third item kind that participates in item ordering, and nothing pins that ordering. I deleted the `flushText()` call before the function-call append (`request.go:137`) and the entire package suite stayed green, even though the mutation genuinely reverses the emitted order. A case with a model turn containing text and a function call, asserting `output_text` item then `function_call` item, would close it.
- `request.go:100-101` — the comment says `EasyInputMessageParam` "can only emit input_text". That isn't quite right: its content union is `string | ResponseInputMessageContentListParam` and `assistant` is a valid role, so a plain-string assistant message is legal. The approach you chose is fine and arguably more explicit; the justification just overstates. Worth a sentence on why `ResponseOutputMessageParam` was preferred.
- The test never asserts `id` is absent — which is the load-bearing assumption of the whole design. It still passes with `ID: "bogus id with spaces"`, the exact value class that 400s per `openai/openai-go#334`. A `!strings.Contains(raw, "id")` assertion, or asserting the full expected JSON, would guard it.
- `normalizeRole`'s `genai.RoleModel` branch (`request.go:234-235`) is now unreachable from production, but `request_test.go:441` still asserts it — a green signal for a dead path. Worth a comment at minimum, since anyone later adding case `assistant`: there would silently reintroduce this bug.
- `newOutputMessage` returns an error that is structurally always nil; the signature could just return the pointer.
- Prefer table-driven tests; this is a single shape.

hanorik (changes requested):
Thanks for this — the diagnosis in https://github.com/google/adk-go/issues/1197 and the fix here are both correct, and the PR description is good. There are some nits I'd like addressed before merge:

The change introduces a third item kind that participates in item ordering, and nothing pins that ordering. I deleted the flushText() call before the function-call append (request.go:137) and the entire package suite stayed green, even though the mutation genuinely reverses the emitted order. A case with a model turn containing text and a function call, asserting output_text item then function_call item, would close it.
request.go:100-101 — the comment says EasyInputMessageParam "can only emit input_text". That isn't quite right: its content union is string | ResponseInputMessageContentListParam and assistant is a valid role, so a plain-string assistant message is legal. The approach you chose is fine and arguably more explicit; the justification just overstates. Worth a sentence on why ResponseOutputMessageParam was preferred.
The test never asserts id is absent — which is the load-bearing assumption of the whole design. It still passes with ID: "bogus id with spaces", the exact value class that 400s per openai/openai-go#334. A !strings.Contains(raw, "id") assertion, or asserting the full expected JSON, would guard it.
normalizeRole's genai.RoleModel branch (request.go:234-235) is now unreachable from production, but request_test.go:441 still asserts it — a green signal for a dead path. Worth a comment at minimum, since anyone later adding case assistant: there would silently reintroduce this bug.
newOutputMessage returns an error that is structurally always nil; the signature could just return the pointer.
Prefer table-driven tests; this is a single shape.

cruvie:
REVIEW THIS PLS.

wojtas000:
Since there's been no activity since the review, I've opened https://github.com/google/adk-go/pull/1427 to carry it forward. Your commit is cherry-picked unchanged and keeps you as an author. Follow-up commits address @hanorik review

wojtas000:
The fix has been merged in https://github.com/google/adk-go/pull/1427. Closing

PR #1206, opened 2026-07-25
fix(remoteagent/v2): run after-callbacks on aggregated artifact events

### LINKED ISSUE #948: remoteagent/v2: AfterA2ARequestCallbacks are not invoked on the aggregated event synthesized from partial artifact chunks

PratikDhanave:
## Summary

Fixes #948.

`AfterA2ARequestCallback`s registered on a `remoteagent.NewA2A` agent were invoked on each *incoming* A2A event, but **not** on the non-partial event the run processor *synthesizes* when it reassembles a partial-chunked artifact.

In `agent/remoteagent/v2/a2a_agent.go`, `processEvent` ran `runAfterA2ARequestCallbacks` on the converted incoming event, then called `aggregatePartial` and yielded its output directly. When an artifact is streamed as `Append` chunks ending with `LastChunk`, `aggregatePartial` builds a brand-new aggregated event via `buildNonPartialAggregation` and that event was yielded without ever passing through the callbacks. Because chunked artifacts arrive entirely as `Partial` events, a callback that (as is typical) only acts on non-partial events skipped every chunk *and* never saw the one non-partial event carrying the reassembled artifact — it silently no-op'd for any chunked artifact.

This affected **both** synthesis sites: the last-chunk path (`a2a_agent_run_processor.go:116`) and the terminal-status flush (`:71-78`).

## Change

Run the after-callbacks on the events returned by `aggregatePartial` before yielding. The pass-through `event` already ran callbacks, so it's skipped by pointer identity (it's the same pointer in every returned slice), avoiding double-invocation; only the synthesized aggregation events — which are freshly allocated — get callbacks. This covers both synthesis sites without touching the aggregator.

## Testing Plan

- Added `TestRemoteAgent_AfterCallbackRunsOnAggregatedArtifact`: streams an artifact as an initial chunk + an `Append`/`LastChunk` chunk (routing through `buildNonPartialAggregation`) and registers a callback that records only **non-partial** events. Asserts the callback saw the reassembled `"Hello, world!"` event.
  - The callback deliberately ignores partial chunks, so its effect can't leak into the aggregated event via chunk-metadata aggregation — only a direct call on the synthesized event can satisfy the assertion. (The pre-existing "after invoked for every event" case doesn't exercise this: it sets metadata on partials, which propagates through aggregation, so it passes with or without the fix.)
- Verified it's a genuine regression guard: **passes** with the fix, **fails** without it (non-partial events seen = `[]`).
- `go test -race -count=1 ./agent/remoteagent/v2/` — pass (full package)


jjsasha63 (approved):
Approving. This is the one we are taking of the four PRs opened against #948, and sorry for the wait.

The production change is identical in all four. #1206 is the only one written after the agent.Context rename in #1109, so it is the only one whose test still compiles against main. It also applies cleanly despite the "behind" label.

Verified: reverting the production hunk makes the new test fail with "non-partial events seen = []", so it is a real regression guard rather than a self-confirming test, and ./agent/remoteagent/... is green with the fix applied. To check locally:

    go test ./agent/remoteagent/v2/ -run TestRemoteAgent_AfterCallbackRunsOnAggregatedArtifact

Three notes, none blocking, all true of the other three PRs equally:

1. The test proves the callback is invoked, not that a callback returning a replacement event has that replacement used. Dropping `toEmit = cbResp` leaves the test passing. It matches the existing call site above, so I am not asking for a change here.
2. Two sites synthesize events. This covers the LastChunk reassembly; the terminal-status flush is not covered. I have asked for that case as a follow-up on #1022, where the test came from.
3. In the terminal-flush path the aggregated events are yielded before `event`, but `event`'s callback ran first. So a callback that tracks order sees a different sequence than the consumer does. Pre-existing shape, worth knowing.

Closing #973, #1022 and #1035 as duplicates.

PR #1207, opened 2026-07-25
fix(adkgo/deploy): derive Agent Engine builder Go version from go.mod

### LINKED ISSUE #1196: adkgo v2.1.0 uses incompatible Go builder version
## 🔴 Required Information

**Describe the Bug:**
`adkgo v2.1.0 deploy agentengine` generates an InlineSource Dockerfile using `golang:1.25`, but `google.golang.org/adk/v2 v2.1.0` declares Go `1.26.5` in its `go.mod`.

The generated Dockerfile is defined here:

https://github.com/google/adk-go/blob/f4c76700f4013fefe6c1ba01495188ec5775c689/cmd/adkgo/internal/deploy/agentengine/agentengine.go#L187

As a result, a Go ADK v2.1.0 application that builds successfully with Go 1.26.5 fails during the managed Agent Engine source-image build.

**Steps to Reproduce:**
1. Create a Go agent using `google.golang.org/adk/v2 v2.1.0`.
2. Ensure its `go.mod` contains:
   ```go
   go 1.26.5

   require google.golang.org/adk/v2 v2.1.0
3. Confirm the project builds locally:
go mod tidy
go test ./...
go vet ./...
go build ./...
4. Deploy it using the pinned v2.1.0 CLI:
go run google.golang.org/adk/v2/cmd/adkgo@v2.1.0 deploy agentengine \
  --entry_point_path=./main.go \
  --name=adk-go-v2-inline-test \
  --project_name="$GOOGLE_CLOUD_PROJECT" \
  --region=europe-west1 \
  --source_dir=.
5. The source archive uploads successfully, but the managed build fails:
Sending CreateReasoningEngine request...
Waiting for operation to complete...

Error: operation failed: rpc error: code = InvalidArgument desc = Build failed.
The issue might be caused by incorrect code, requirements.txt file or other dependencies.
6. Inspect the Dockerfile generated by adkgo. It contains:
FROM golang:1.25 as builder
WORKDIR /app
COPY . .
RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ...
Expected Behavior:
adkgo v2.1.0 deploy agentengine should generate a builder image compatible with the Go version required by ADK v2.1.0, allowing a locally valid ADK Go v2.1.0 application to deploy successfully.
For example:
FROM golang:1.26.5 AS builder
Alternatively, the builder Go version could be derived from the application or ADK module requirements rather than being hardcoded.
Observed Behavior:
The generated Dockerfile hardcodes:
FROM golang:1.25 as builder
The managed Agent Engine image build subsequently fails with:
Error: operation failed: rpc error: code = InvalidArgument desc = Build failed.
The issue might be caused by incorrect code, requirements.txt file or other dependencies.
The same source passes go test, go vet, and go build with Go 1.26.5.
Environment Details:
- ADK Library Version: google.golang.org/adk/v2 v2.1.0
- OS: Linux, arm64
- Go Version: go1.26.5 linux/arm64
- Deployment region: europe-west1
- Deployment command: go run google.golang.org/adk/v2/cmd/adkgo@v2.1.0 deploy agentengine

PratikDhanave:
## Summary

Fixes #1196.

`adkgo deploy agentengine` generated an InlineSource Dockerfile hardcoding `FROM golang:1.25 as builder`, but the ADK v2 module declares `go 1.26.5`. An app that builds fine locally with Go 1.26.5 then failed the managed Agent Engine source-image build, because the builder image used an older toolchain than the code requires.

## Change

Derive the builder image tag from the `go` directive in the application's `go.mod` (`f.source.sourceDir`), so the managed build always uses a toolchain that satisfies what the app declares. Falls back to the toolchain running `adkgo`, then to a constant. This is self-maintaining — it won't re-drift on the next Go release.

The `go` directive is parsed with a small dependency-free helper (a digit-prefix guard avoids matching module paths in a `require` block); I avoided promoting the already-indirect `golang.org/x/mod` to a direct dependency to keep `go.mod` untouched.

## Testing Plan

- `TestGoVersionFromModFile` — table covering patch (`1.26.5`) and minor (`1.26`) directives, and non-matching lines (`require` block modules, `toolchain go1.26.5`, no directive).
- `TestPrepareDockerfile_UsesGoModVersion` — writes a temp `go.mod` with `go 1.26.5`, runs `prepareDockerfile`, and asserts the generated Dockerfile contains `FROM golang:1.26.5 AS builder` and no longer hardcodes `golang:1.25`.
- Verified it's a genuine regression guard: **passes** with the fix, **fails** against the pre-fix hardcoded line.
- `go test -race -count=1 ./cmd/adkgo/internal/deploy/agentengine/` — pass


baptmont (commented):
Nice fix — deriving the tag from the app's own `go.mod` is the right shape, and it's self-maintaining in the way the issue asked for. Two details I want to call out as good calls: reading `sourceDir/go.mod` directly rather than walking up the filesystem is correct (only `sourceDir`'s contents end up in the tarball, so an enclosing module wouldn't be visible to the remote build anyway), and avoiding the promotion of `golang.org/x/mod` to a direct dependency keeps `go.mod` untouched, which is worth it here.

Three things I hit while testing this against the failure in #1196.

### 1. A trailing comment on the `go` directive produces an invalid Dockerfile

`isGoVersion` only checks that the first byte is a digit, and nothing strips `//`, so a comment on the `go` line is carried into the `FROM` instruction:

```
go 1.26.5 // pinned by platform team
```

generates

```dockerfile
FROM golang:1.26.5 // pinned by platform team AS builder
```

which Docker rejects outright:

```
ERROR: dockerfile parse error on line 1: FROM requires either one or three arguments
```

Comments on the `go` line are reasonably common in the wild. The same weakness lets `go 1abc` through as a version (`golang:1abc` is a 404 on Docker Hub).

Stripping anything from `//` onward before parsing, and validating the shape rather than just the first byte, covers both.

### 2. Without `GOTOOLCHAIN=auto` there is no margin for error

The underlying reason `golang:1.25` fails isn't only that the tag is old — it's that the official images pin `ENV GOTOOLCHAIN=local` (docker-library/golang#472, *"don't auto-upgrade the gotoolchain"*), so the `go` command is forbidden from fetching the toolchain `go.mod` asks for:

```
go: go.mod requires go >= 1.26.5 (running go 1.25.12; GOTOOLCHAIN=local)
```

That's also why the reporter's app builds locally but not remotely — their local `GOTOOLCHAIN=auto` silently papers over the same mismatch.

Deriving the tag makes the version correct, but leaves it a single point of failure: any residual mismatch is still fatal. Adding one line to the builder stage

```dockerfile
ENV GOTOOLCHAIN=auto
```

makes the two mechanisms cover each other — the derived tag keeps the fast path (no toolchain download, works when the build has no proxy egress), and `auto` rescues the cases the tag can't: a Docker Hub publishing lag right after a Go release, or a transitive dependency requiring a newer toolchain than the main module declares.

I verified the safety net works in isolation, using a deliberately stale base image:

```dockerfile
FROM golang:1.25 AS builder
ENV GOTOOLCHAIN=auto
```

```
#8 1.739 go: downloading go1.26.5 (linux/amd64)
#8 14.82 go version go1.26.5 linux/amd64
#8 DONE 23.2s
```

### 3. `defaultBuilderGoVersion = "1.26"` will re-rot

The last-resort constant is the same species of hardcoded version as the `golang:1.25` this PR removes — it resets the clock rather than stopping it. Since that branch is only reached when the required toolchain is entirely unknown, a rolling tag (`"latest"`) can't be too old and can't go stale. Worth a small test asserting the default doesn't pin a version, so it doesn't quietly get pinned again later.

---

Happy to send any of this as a patch if that's easier than reworking it yourself — just say the word.

baptmont (changes requested):

PratikDhanave:
Thanks for the thorough review, @baptmont — all three addressed in the latest push:

1. **Trailing comment / malformed version** — `goVersionFromModFile` now drops anything from `//` onward before parsing, and `isGoVersion` validates the full shape via regexp instead of only the leading byte, so `go 1.26.5 // pinned` yields `1.26.5` and `go 1abc` is rejected (no more `golang:1abc` 404). Added both cases to `TestGoVersionFromModFile`.
2. **`GOTOOLCHAIN=auto`** — emitted in the builder stage so the derived tag and the toolchain-download path cover each other. Asserted in the Dockerfile test. (Amusingly, my local build hit the exact `GOTOOLCHAIN=local` mismatch you described until I set this.)
3. **`defaultBuilderGoVersion`** — now the rolling `"latest"` tag rather than a pinned `1.26`, with `TestDefaultBuilderGoVersionIsNotPinned` guarding it from being re-pinned.

`go build`/`vet`/`test`/`golangci-lint` all green. PTAL.

jjsasha63 (approved):
Reviewed at `87b977d`. I went back over the three points from the earlier review and checked each one against the code rather than the summary, then verified the underlying premise with real container builds. All three are properly addressed, so I think this is clear to go in. Two small test-only suggestions below, neither blocking.

## The three earlier points

**1. Trailing comment and malformed version — fixed.** I ran the old and new parsers side by side:

| `go.mod` line | Before | Now |
|---|---|---|
| `go 1.26.5 // pinned by platform team` | `FROM golang:1.26.5 // pinned by platform team` | `FROM golang:1.26.5` |
| `go 1abc` | `FROM golang:1abc` | rejected, falls back |
| `go 1.26.5//pinned` | `FROM golang:1.26.5//pinned` | `FROM golang:1.26.5` |
| `go 1.26.5 // use 1.99 in prod` | invalid | `FROM golang:1.26.5` |

The last two were not in the original comment. Validating the whole shape rather than the leading byte covers them as a side effect, which is the right kind of fix.

**2. `GOTOOLCHAIN=auto` — fixed, and the reasoning behind it holds.** I checked the premise first:

```
$ docker run --rm golang:1.25 sh -c 'go env GOTOOLCHAIN; go version'
local
go version go1.25.14 linux/amd64
```

Then built three images against a module declaring `go 1.26.5`:

| Builder stage | Result |
|---|---|
| `golang:1.25`, no `ENV` (what `main` generates today) | fails: `go: go.mod requires go >= 1.26.5 (running go 1.25.14; GOTOOLCHAIN=local)` |
| `golang:1.25` + `ENV GOTOOLCHAIN=auto` | succeeds, `go: downloading go1.26.5 (linux/amd64)` |
| `golang:1.26.5` + `ENV GOTOOLCHAIN=auto` (what this PR generates) | succeeds, no toolchain download |

The first row reproduces #1196 with the reporter's exact symptom, which is useful confirmation that this PR is aimed at the right thing. The second shows the safety net works alone; the third shows the derived tag keeps the fast path so that download does not normally happen. The `ENV` is in the builder stage ahead of the `RUN`, which is the placement that works.

**3. Rolling default — implemented.** The constant is `"latest"` and `TestDefaultBuilderGoVersionIsNotPinned` does catch a naive re-pin; I confirmed by setting it back to `"1.26"` and watching the test fail.

## Two test-only suggestions

**The main test cannot currently tell the two code paths apart.** I mutation-tested each new test by breaking the matching production code. Four of five mutations were caught. The one that survives is the PR's core behaviour: replacing the body of the `go.mod` branch in `builderGoVersion` with a discard,

```go
if data, err := os.ReadFile(filepath.Join(dir, "go.mod")); err == nil {
	_ = goVersionFromModFile(data)
}
```

leaves the suite green. The fixture writes `go 1.26.5`, and a runner on a 1.26.5 toolchain reaches `golang:1.26.5` through the `runtime.Version()` fallback too, so the assertion passes either way. A fixture version nobody is running fixes it:

```go
// agentengine_test.go:61
[]byte("module example.com/agent\n\ngo 1.21.9\n")
// agentengine_test.go:79
if !strings.Contains(string(content), "FROM golang:1.21.9 AS builder") {
```

With that change the same mutation fails, as it should.

**The "not pinned" guard is narrower than it reads.** It relies on `isGoVersion` returning false, which is also true for pinned tags carrying a variant suffix:

| Future value | Guard |
|---|---|
| `1.26` | rejects, correct |
| `1.26-bookworm` | allows |
| `1.26-alpine` | allows |
| `1.25.3-trixie` | allows |

Those are all pinned and all plausible things to write later. Comparing against `"latest"` directly, or asserting the value contains no digit, closes it.

## Other things I checked and found fine

Every tag the pattern admits is a real image: I queried Docker Hub for `1.26`, `1.26.5`, `1.26.0`, `1.26.6`, `1.26rc1` and `latest`, all HTTP 200, so no realistic `go` directive derives a 404 tag.

`builderGoVersion`'s `sourceDir`-then-`os.Getwd()` logic matches `createArchive`, so the file it parses is the one that lands at the root of the tarball. Easy to get subtly wrong; it is right.

The derived string goes straight into a `FROM` line that a remote builder executes, so I tried to break out of it: embedded newlines and carriage returns, `1.26.5;RUN ...`, `1.26.5 AS b` followed by a newline, `1.26.5-alpine`, surrounding whitespace, an overlong numeric version. All rejected. The line split plus the fully anchored pattern closes it, and Go's `$` is `\z` rather than multiline so a trailing newline cannot slip past. A CRLF `go.mod` still parses.

One optional thought, take it or leave it: when `go.mod` is missing or unparseable the fallback is silent, and nothing in the output says which builder version was chosen. `prepareDockerfile` already has the printer one line above. Since the bug being fixed here was an opaque remote build failure, a single `p("Builder Go version:", ...)` might save the next person the same dig.

`gofmt` and `go vet` clean; the package tests pass on this head.

jjsasha63 (approved):

baptmont (approved):
All three addressed, and I checked them by running the code rather than reading the diff.

**Version parsing.** The comment is stripped before parsing and `isGoVersion` validates the whole shape via `goVersionPattern`. Probing the cases I raised plus a few neighbours:

```
"go 1.26.5 // pinned by platform team"   -> "1.26.5"
"go 1abc"                                -> ""
"go 1.26.5//nospace"                     -> "1.26.5"
"\tgo 1.23.1\t// tabbed"                 -> "1.23.1"
"go // only a comment"                   -> ""
"go 1.26rc1"                             -> "1.26rc1"
```

The pre-release case still working matters — `^[0-9]+(\.[0-9]+){0,2}([a-z]+[0-9]+)?$` is tight enough to reject `1abc` without rejecting `1.26rc1`, which a stricter pattern would have.

**`GOTOOLCHAIN=auto`.** In the builder stage immediately after `FROM`, with the reasoning in a comment and an assertion in `TestPrepareDockerfile_UsesGoModVersion`. The derived tag and the toolchain download now cover each other, which was the point.

**`defaultBuilderGoVersion`.** `"latest"`, and `TestDefaultBuilderGoVersionIsNotPinned` guards it with `isGoVersion(defaultBuilderGoVersion)` — reusing the validator as the thing that fails if someone re-pins it. That is a nicer guard than the one I suggested, since it cannot drift away from what counts as a pinned version.

Also read the parts I had not commented on. The fallback chain reads correctly: app `go.mod`, then the toolchain running `adkgo`, then the rolling tag, each guarded so an unreadable or malformed input falls through rather than producing a bad tag. `runtime.Version()` goes through the same validator, so a devel toolchain string cannot become a `FROM` either.

`go build` and the package tests are green locally, and CI is green.

Clearing my earlier request. One housekeeping note: the branch is five commits behind `main` and will need an update before it can merge.

PR #1221, opened 2026-07-29
test: cover artifact version service contract

### LINKED ISSUE #681: Missing contract test coverage for GetArtifactVersion()
The new `func (s *gcsService) GetArtifactVersion(ctx context.Context, req *artifact.GetArtifactVersionRequest)` method does not appear to be covered by the contract tests yet.

https://github.com/google/adk-go/blob/b8eb8c5f07cf43889a02bce2eac2d104a81fb927/artifact/request_validation_test.go#L329

It would be useful for the contract tests to define which ArtifactVersion fields are required across backends, which are optional, and whether backend-specific omissions are allowed.

vivekjm:
## Summary

Add the missing shared service-contract coverage for `GetArtifactVersion` so every artifact backend is checked against the same behavior.

The tests cover:

- resolving the latest and a specific artifact version
- the metadata fields shared by all backends (`Version` and `MimeType`)
- text artifacts, which use the default `text/plain` MIME type
- missing artifacts in both empty and previously populated services

Backend-specific metadata such as canonical URI, creation time, and custom metadata is intentionally not required by the shared contract.

Fixes #681

## Testing plan

- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.3.1 run`
- `go mod tidy -diff`

### DIFF
--- a/internal/artifact/tests/service_suite.go
+++ b/internal/artifact/tests/service_suite.go
@@ -143,6 +143,45 @@ func testArtifactService(ctx context.Context, t *testing.T, srv artifact.Service
 		}
 	})
 
+	t.Run(fmt.Sprintf("GetArtifactVersion_%s", testSuffix), func(t *testing.T) {
+		for _, tc := range []struct {
+			name     string
+			fileName string
+			version  int64
+			want     *artifact.ArtifactVersion
+		}{
+			{
+				name:     "latest",
+				fileName: "file1",
+				version:  0,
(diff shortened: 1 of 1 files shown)


vivekjm:
Thanks for taking a look. Since this PR was opened, #1156 landed on main with the same shared GetArtifactVersion contract coverage, including latest and explicit versions plus missing-version and missing-file cases. Updating this branch would duplicate that suite, so I’m closing this PR rather than keeping a redundant change open.

PR #1223, opened 2026-07-29
llminternal: cancel active streaming tools on live session Close

0xloophole:
liveSessionImpl.Close() closes the done channel but never inspects or
cancels tools registered in activeTools. A StreamingFunctionTool
goroutine spawned by handleFunctionCalls therefore keeps running
RunStream after the session has ended, with no way to observe
termination.

Close() now cancels every registered task using the same locking
already used by RegisterStreamingTool/UnregisterStreamingTool/
CancelAllStreamingTools. The streaming goroutine is restructured so a
call blocked mid-iteration inside RunStream no longer blocks reaction
to cancellation: a dedicated inner goroutine is the sole caller of
RunStream, forwarding chunks over a channel that the outer goroutine
selects against alongside cancelCtx.Done().

cleanup() in RunLive is intentionally left untouched beyond a comment:
it runs on every reconnect attempt where session resumption is still
intended, so activeTools cancellation stays exclusive to Close(), the
sync.Once-guarded true termination path.

### DIFF
--- a/internal/llminternal/base_flow.go
+++ b/internal/llminternal/base_flow.go
@@ -247,6 +247,19 @@ func (s *liveSessionImpl) recvIter() iter.Seq2[*session.Event, error] {
 
 func (s *liveSessionImpl) Close() error {
 	s.closeOnce.Do(func() {
+		// Cancel all in-flight streaming tools BEFORE closing s.done.
+		// Tools that respect ctx.Done() get a clean, deterministic signal
+		// here; Send() only becomes a backstop for tools that are still
+		// mid-flight when this fires.
+		s.mu.Lock()
+		for toolName, tasks := range s.activeTools {
+			for _, t := range tasks {
+				t.cancel()
+			}
+			delete(s.activeTools, toolName)
+		}
+		s.mu.Unlock()
+
 		close(s.done)
 	})
 	return nil
@@ -365,6 +378,11 @@ func (f *Flow) RunLive(ctx agent.InvocationContext) (agent.LiveSession, iter.Seq
 			cleanup := func() {
 				cancelConn()
 				_ = liveConn.Close()
+				// Deliberately does NOT cancel s.activeTools. cleanup() runs on every
+				// reconnect attempt during transient network failures, where session
+				// resumption -- and continuity of in-flight streaming tools -- is
+				// still intended. activeTools cancellation belongs exclusively to
+				// liveSessionImpl.Close(), the sync.Once-guarded true termination path.
 			}
 
 			eventsChan := make(chan *session.Event)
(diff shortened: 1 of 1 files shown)


karolpiotrowicz (changes requested):
The diagnosis here is correct and the `Close()` change is the right fix. A streaming tool that blocks *between* yields — waiting on a timer or a network read, so it never reaches `Send` — had no stop signal at all: `Close()` closed `done` but never called the `cancel` functions held in `activeTools`, so `ctx.Done()` never fired for it. The `Send` → `io.EOF` backstop only ever covered chatty tools. Cancelling the registered tasks closes that gap, and doing it under `s.mu` before `close(s.done)` is the right ordering. Worth noting this puts the Go implementation ahead of adk-python on this point, which still has the equivalent leak.

My suggestion is to split the PR, because the two halves have very different risk profiles.

**The 12-line `Close()` hunk is sufficient on its own.** I applied only that hunk to the base commit, left the original single-goroutine loop untouched, and the repro passes under `-race`: a tool that blocks without yielding observes cancellation after `Close()`. The ~68-line goroutine restructuring isn't needed to fix the reported bug, and as written it introduces two regressions.

### 1. The drain goroutine is unnecessary, and it leaks

```go
case <-cancelCtx.Done():
    go func() { for range chunkCh {} }()
    return
```

The comment justifies this as ensuring the inner goroutine "is never left with no reader", but the inner goroutine's send is itself a `select` against the same `cancelCtx.Done()` (`base_flow.go:1142-1146`). The outer goroutine only reaches this branch when `cancelCtx` is *already* cancelled, and a closed channel is permanently ready — so the inner `select` always has a runnable case and can never block permanently.

Meanwhile `for range chunkCh` only terminates when `chunkCh` is closed, and `chunkCh` is closed only by the inner goroutine's `defer`. If a tool wedges inside `RunStream` and ignores its context, that defer never runs and the drainer parks forever. With such a tool I count **2 goroutines surviving `Close()` on this branch vs 1 on the base commit** — the two being the inner driver (`base_flow.go:1141`) and the drainer (`base_flow.go:1183`).

Deleting it and returning directly restores parity; I ran 300 close cycles with a chatty tool and a deliberately slow reader with zero survivors and no race.

```diff
 case <-cancelCtx.Done():
-    go func() { for range chunkCh {} }()
     return
```

### 2. Error chunks now signal "keep going" to the tool

This one is subtle and specific to `range`-over-func. Previously the loop body did `return` on an error, which makes the current `yield` call return `false` — the documented stop signal. Now the inner goroutine successfully sends the error chunk on `chunkCh`, the loop body completes normally, and `yield` returns **`true`**, while the outer goroutine receives that error and abandons the stream.

So at the moment the framework gives up on the tool, the tool is told the chunk was accepted and it should continue. Measured against a tool that records what `yield` returned: base gives `false` with 0 further yields, this branch gives `true` with 1 further yield.

```diff
 case chunkCh <- chunkResult{chunk, err}:
+    if err != nil {
+        return
+    }
 case <-cancelCtx.Done():
     return
```

More generally, putting a channel between the producer and the consumer erases the backpressure that `range`-over-func encodes in `yield`'s return value — the send succeeds whether or not the consumer still cares. That's the main argument for keeping the single-goroutine loop.

### 3. Tests

`liveSessionImpl.Close` is at 0.0% statement coverage, so the green `-race` run isn't evidence that the new concurrency is correct — three of the restructured blocks (`:1159-1162`, `:1171-1174`, and the drain's loop body) don't execute at all in the suite. Two tests would cover the change and are about a dozen lines each:

- a tool that blocks without yielding, selecting only on `ctx.Done()`, asserted to observe cancellation after `Close()` — this fails on the base commit and passes here, so it directly guards the bug you're fixing;
- a goroutine-count check after `Close()` with a context-ignoring tool, which would have caught the drainer issue above.

### Suggested path

Ship the `Close()` hunk on its own — it's a clean, well-scoped fix — plus the first test. If you'd like to keep the restructuring, it's probably better as a follow-up with the two fixes above and the leak-count test, so the bug fix isn't held up by it.

One thing to be aware of for a follow-up, not something to fix here: `Close()` is `sync.Once`-guarded and `RegisterStreamingTool` has no closed-check, so a tool registered after `Close()` has drained the map is never cancelled and nothing sweeps it again. That's reachable, since `Close()` runs on the transport's teardown path while `RunLive`'s loop dispatches function calls on another goroutine. It's pre-existing rather than introduced here, but it does bound the invariant this PR is establishing — a `closed` flag set under `s.mu`, with `Register` cancelling immediately when it's set, would close it.

PR #1225, opened 2026-07-29
docs(llminternal): correct comment on synthetic user-turn injection

### LINKED ISSUE #1218: Incorrect comment: contents_processor.go claims the Gemini API requires role alternation
### Summary

`internal/llminternal/contents_processor.go:64-65` states a constraint that does not appear to hold:

```go
// Gemini API requires role alternation (model turns cannot be consecutive).
// If the conversation history concludes on a model turn, inject a synthetic user continuation turn.
```

The API does accept consecutive `model` contents. Sending a history with two adjacent `model` contents mid-conversation (both plain text, and with function calls) returned HTTP 200 on `gemini-flash-latest` and `gemini-3.5-flash`.

The **code** at `:66-70` is fine and should stay — it just isn't doing what the comment says. Its real justification is the one given for the identical repair in `model/gemini/gemini.go:163-164` ("appends a user content, so that model can continue to output") and in adk-python's `base_llm.py` ("Insert a user content to preserve user intent and to avoid empty model response"). That is a generation-quality nudge, not an API requirement. Notably, adk-python contains no role-alternation enforcement anywhere.

### Why it's worth fixing

The comment reads as a specification, and it is load-bearing in review: it is the stated basis for treating any mid-history `model`/`model` adjacency as a defect. It's also inaccurate in a second way — the pipeline already produces consecutive `model` contents on ordinary paths (two adjacent model events, e.g. a thought-only turn followed by an answer), and the repair at `:66-70` only inspects the last content, so it never protected against the mid-history case to begin with.

### Observed behaviour

Consecutive `model` contents were accepted in every shape tested. The one case that was rejected turned out to be unrelated to role ordering:

| shape | result |
|---|---|
| consecutive `model` text contents mid-history | 200 |
| consecutive `model` contents containing function calls, with `thoughtSignature` present | 200 |
| alternating roles, function calls without `thoughtSignature` | 200 |
| consecutive `model` contents with function calls and **no** `thoughtSignature` anywhere | 400 — `Function call is missing a thought_signature in functionCall parts` |

Adding a signature to either call in that last shape returned it to 200. So the rejection tracks `thoughtSignature` presence, not role alternation. This shouldn't affect normal operation, since function calls emitted by Gemini carry signatures and they are round-tripped, but it may be worth being aware of for replayed, resumed, or synthetically constructed histories.

### Suggested change

Reword the comment to describe what the code actually does and why, for example:

```go
// If the conversation history concludes on a model turn, inject a synthetic user
// continuation turn so the model keeps producing output rather than returning an
// empty response. (Mirrors maybeAppendUserContent in model/gemini and
// adk-python's _maybe_append_user_content.)
```

No behaviour change intended — comment only.

Environment: verified against `generativelanguage.googleapis.com/v1beta`, models `gemini-flash-latest` and `gemini-3.5-flash`. Not re-verified against Vertex.

AloysJehwin:
## Summary

Fixes #1218

The comment at `internal/llminternal/contents_processor.go:64-65` stated that the Gemini API requires role alternation and rejects consecutive model turns. As the issue documents, this is incorrect — the API accepts consecutive `model` contents (HTTP 200 in all tested shapes).

The code itself is correct and should stay. The real justification for injecting a synthetic user continuation turn is a **generation-quality nudge**: without it the model may return an empty response or fail to summarise completed work. This matches the rationale in:
- `model/gemini/gemini.go:163-164` — "appends a user content, so that model can continue to output"
- adk-python's `base_llm.py` — "Insert a user content to preserve user intent and to avoid empty model response"

## Changes

- Updated the two-line comment to accurately describe why the synthetic user turn is injected.

## Test results

```
ok  google.golang.org/adk/v2/internal/llminternal  0.608s
ok  google.golang.org/adk/v2/internal/llminternal/googlellm  0.007s
```

## AI Usage

Implemented with Claude Code assistance. All changed lines reviewed manually.

### DIFF
--- a/internal/llminternal/contents_processor.go
+++ b/internal/llminternal/contents_processor.go
@@ -61,8 +61,9 @@ func ContentsRequestProcessor(ctx agent.InvocationContext, req *model.LLMRequest
 			return
 		}
 		req.Contents = append(req.Contents, contents...)
-		// Gemini API requires role alternation (model turns cannot be consecutive).
-		// If the conversation history concludes on a model turn, inject a synthetic user continuation turn.
+		// If the conversation history concludes on a non-user turn, inject a synthetic
+		// user continuation turn to prompt the model to continue processing the
+		// previous request or provide a final summary if the work is complete.
 		if len(req.Contents) > 0 {
 			if last := req.Contents[len(req.Contents)-1]; last != nil && last.Role != "user" {
 				req.Contents = append(req.Contents, genai.NewContentFromText("Continue processing previous requests as instructed. Exit or provide a summary if no more outputs are needed.", "user"))


AloysJehwin:
Closing — will look for a more substantial code logic fix instead.

PR #1226, opened 2026-07-31
Fix unsafe A2A peer-supplied transferToAgent metadata

prasanna8585:
A remote A2A peer could set `transfer_to_agent` metadata on any message/task/status-update/artifact-update it sends back, and `toEventActions()` in `server/adka2a/v2/events.go` restored it verbatim into the local `session.EventActions`, which `base_flow.go`'s `agentToRun` then honors to select the next agent to run -- letting a malicious or compromised remote peer redirect the local orchestrator's control flow within its own configured multi-agent tree.

This mirrors the fix already merged in `google/adk-python` (0ba7d3cba7004c0fcc0a05f1f4cfb6ea78e38f91, "fix: ignore unsafe A2A peer-supplied event actions metadata"), which explicitly documents `transfer_to_agent` as unsafe peer-controllable state that drives the caller's own control flow and must never be rebuilt from peer-supplied metadata.

Note: `server/adka2a/v2/events_test.go`'s `TestToSessionEvent` (case `"message"`) currently asserts the old (vulnerable) behavior as correct -- happy to update that expectation in this PR if maintainers confirm the intended fix direction; left it as-is pending review so the failing test itself demonstrates the bug.

Related: filed a companion report/PR for the identical gap in `google/adk-js`; `google/adk-java` was checked and does not have this class of bug (no A2A actions-restoration code exists there yet).


prasanna8585:
Hi Team,
Is there any update required or code changes from my side?

yarolegovich:
This makes sense if a remoteagent is used to connect to an untrusted agent. We can't however simply remove this behaviour, because implementations can be relying on it in their closed trusted deployments. We should introduce a flag like `remoteagent.A2AConfig.AllowTransferToAgent` and redact the transfer only when it's false.

prasanna8585:
Updated per your feedback -- added `A2AConfig.AllowTransferToAgent` (defaults to `false`, so untrusted-by-default connections stay safe) and made the `TransferToAgent` restoration conditional on it instead of removing it unconditionally.

Threaded through the full call chain (`A2AConfig` -> the run processor -> `ToSessionEventWithParts` -> the per-event-type helpers -> `processA2AMeta` -> `toEventActions`). `ToSessionEvent`, the simple public wrapper, always passes `false`, so existing callers of that API get the secure behavior automatically without any change on their end.

Updated the existing `"message"` test case that asserted the old behavior, and added `TestToSessionEventWithParts_AllowTransferToAgent` covering both the default-redacted and explicit-opt-in paths.

One honest caveat: I verified this with `gofmt` (syntax/formatting) across two independent fresh pulls of this branch, but wasn't able to run a full `go build`/`go test` in the environment I used -- it requires Go 1.26.5 and I didn't have network access to fetch that toolchain. Happy to address anything CI turns up.

server/adka2a/v2/events.go
```diff
@@ -72,23 +72,32 @@ func EventToMessage(event *session.Event) (*a2a.Message, error) {
...
 }
 
 // ToSessionEventWithParts converts the provided a2a event to session event with custom part converter.
-func ToSessionEventWithParts(ctx agent.InvocationContext, event a2a.Event, partConverter A2APartConverter) (*session.Event, error) {
+//
+// allowTransferToAgent controls whether a transfer_to_agent value set by the
+// remote peer in its response metadata is honored; see
+// [remoteagent.A2AConfig.AllowTransferToAgent] for why this defaults to
+// false in most callers.
+func ToSessionEventWithParts(ctx agent.InvocationContext, event a2a.Event, partConverter A2APartConverter, allowTransferToAgent bool) (*session.Event, error) {
```
yarolegovich: now we make a breaking API change. it's probably better to have remoteagent use the flag it gets in the config to restore transferToAgent from metadata after converter runs.

prasanna8585: Updated per your feedback -- this replaces the previous commit rather than building on top of it, since the point was to avoid the breaking change in the first place.

server/adka2a/v2/events.go and metadata.go are back to their original signatures from the first commit on this branch -- no new parameter on ToSessionEventWithParts or anywhere else. toEventActions still never restores TransferToAgent from peer metadata, unconditionally.

Added one small, purely additive exported function instead: TransferToAgentFromMeta(meta) (string, bool). A2AConfig.AllowTransferToAgent now drives a post-processing step in a2a_agent.go's processEvent, applied after either the default converter or a custom cfg.Converter runs -- so it works uniformly regardless of which converter path was used, and adka2a's conversion API doesn't need to know about this trust decision at all.

Updated the round-trip metadata test (its outgoing/incoming directions are now legitimately asymmetric for this one field) and replaced the test that relied on the removed parameter with a smaller one plus a direct test for the new helper.

Same honest caveat as before: verified with gofmt across two independent copies, but wasn't able to run a full go build/go test -- no network access to fetch the Go 1.26.5 toolchain this repo requires in the environment I used.

prasanna8585:
@yarolegovich is there anything more needed form my side or are we ready to merge this PR?

yarolegovich:
@prasanna8585 thanks for the ping. Please cleanup excessive comments in the code, they don't add any value. Exported symbols must have comments, in other cases comments are only needed if they explain something not obvious from the code. For example:

> // TransferToAgent intentionally NOT restored from peer metadata.

is a good comment. 

> 		// wantAfterProcessA2AMeta is the event processA2AMeta should produce
		// from a2aMeta, if different from event. processA2AMeta never
		// restores TransferToAgent from peer-supplied metadata (see
		// TransferToAgentFromMeta), so it's asymmetric with setActionsMeta
		// (the outgoing direction, tested against event/a2aMeta directly)
		// whenever TransferToAgent is set.

Is not.

prasanna8585:
@yarolegovich , thanks for the review i have made the changes caan you review it?

yarolegovich (dismissed):

agent/remoteagent/v2/a2a_agent.go
```diff
@@ -129,6 +129,22 @@ type A2AConfig struct {
...
+	// configured multi-agent tree, so accepting it from a remote peer lets
+	// that peer redirect the local orchestrator's control flow. This
+	// defaults to false (the peer-supplied value is redacted) so that
+	// connecting to an agent you do not fully trust is safe by default. Set
+	// this to true only for closed, trusted deployments that rely on the
+	// remote peer being able to select the next local agent.
+	//
+	// This is applied uniformly after conversion, whether the default
+	// converter or a custom Converter is used.
+	AllowTransferToAgent bool
```
yarolegovich, line 188: one last minor change and good to merge, can you please modify this config creation in the deprecated package: https://github.com/google/adk-go/blob/main/agent/remoteagent/a2a_agent.go#L135 to set `AllowTransferToAgent: true`

yarolegovich (approved):

PR #1232, opened 2026-07-31
feat(compaction): summarize completed invocations as the conversation grows

baptmont:
Second of five. **Re-cut**: this PR previously showed code that later PRs
corrected, and telemetry used to be a separate PR (#1233, now closed). It now
stands on its own.

## What this adds

The compaction library, the engine behind it, and the sliding-window strategy
the runner drives once an invocation finishes. After this PR, compaction works
end to end.

`compaction.Config` is the public surface: which strategy to run, and a
`Summarizer` to run it with.

## A Summarizer returns a summary, not an event

It returns content plus token usage. Returning a whole event would let
third-party code set the authorship, a state delta, an agent transfer and its
own covered range, none of which is summarizing. The framework builds the event
and derives the range from the events it handed over.

The default `LLMSummarizer` carries a timeout, an allow-list of generation
settings, and a check that the generation was not truncated or filtered. A
truncated summary stored as a real one silently loses the turns it covers, and
a `StopSequences` hit reports `STOP`, which is indistinguishable from finishing
normally.

## One definition of coverage

Inside the recorded range and not named as a hole. Every predicate that could
disagree about whether an event is covered is a way to delete conversation, so
there is exactly one.

Prompt assembly substitutes a summary for the events it covers and leaves the
rest raw. An unanswered tool call is never separated from its response, and a
window is never summarized across a branch or isolation-scope boundary.

## What third-party code cannot do

A summarizer sees copies built field by field from what a summarizer is for,
not the session's live events. Copying the struct and severing pointers
afterwards does not hold: `session.Event` and `genai.Part` between them reach
sixteen pointers, and a field added upstream would be silently shared.

A race guard re-reads the session before and after the plugin pass and discards
the summary if anything landed inside its range while it was being produced.
Once is not enough, because a plugin is arbitrary code and runs in between.

## Why telemetry is here

The compactor calls it directly. The span opens before summarization and stays
open until the caller reports what became of the summary, which is what stopped
the five discard paths reporting success for an event that reached no session.
That makes telemetry inseparable from the engine, which is why #1233 is closed.

## Testing plan

`go build`, `go test -race -count=1 -shuffle=on`, `golangci-lint run` and
`go mod tidy -diff` clean in both modules, verified at this slice specifically.
Known-unrelated failure: `TestConfigureExporters` in `./telemetry`.

Note: the tail-retention tests in `runner/compaction_test.go` and
`telemetry_test.go` arrive with #1234, since they exercise a strategy this PR
does not add.

[...]

PR #1231, opened 2026-07-31
feat(session): carry a context-compaction record on a session event

baptmont:
First of five in the context-compaction stack. **Re-cut**: this PR previously
showed code that later PRs corrected. It now stands on its own.

## What this adds

`session.EventCompaction` and `session.EventRef`: the stored shape a compaction
summary takes, and the guarantee that every backend keeps it.

A summary is an ordinary event whose `Actions.Compaction` names the timestamp
range it stands in for, the events inside that range it does *not* stand in for,
and the summary content. Nothing here produces one. This is the record as stored
data, so the storage contract is settled before anything writes to it.

## Why the exclusions exist

A range alone is not enough. Choosing a window filters events out of the middle
of its own span, by branch, by isolation scope, and by what a retained tail
holds back. An interval covering the ends also covers the gaps, and an event in
a gap is dropped from every later prompt having been summarized by nothing.

Imprecision here is not symmetric, and the code says so. A reference matching
too much leaves an extra event raw beside a summary of it, which is visible and
recoverable. A reference that fails to match does not fall back to anything:
coverage is the range minus the exclusions, so the event it protected becomes
covered by a summary that never described it.

## The record is the framework's

A tool handler, an agent callback and a workflow tool node each hold an
`EventActions` that lands on the persisted event, so each clears the field, and
the REST mapper drops an inbound one. Anything able to plant a record can erase
history and substitute content of its own.

## New obligations on session.Service

Both enforced by the shared conformance suite, so an implementation that misses
either goes red rather than failing quietly:

- An event arriving without an ID is assigned one **in place**. A stored event
  that cannot be named cannot be referred to by a record.
- `Actions.Compaction` survives the round trip. A summary carries its content
  only there, with no content and no deltas, so a backend deciding what to
  persist by looking at content or deltas drops it silently.

The conformance cases use nanosecond timestamps so a backend that rounds is
visible, and one case checks a hole still names its event after a round trip.

## Testing plan

`go build`, `go test -race -count=1 -shuffle=on`, `golangci-lint run` and
`go mod tidy -diff` clean in both modules. Verified at this slice specifically,
not only at the stack tip. Known-unrelated failure in this tree:
`TestConfigureExporters` in `./telemetry`, green on main and in CI.

## Changed since the last review

**The Vertex AI gap is closed.** All three compaction conformance cases were
skipping there for want of a recording, so the suite reported PASS while
asserting nothing about the backend the questions were about. They are recorded
now and they answer both: a compaction record survives the round trip, and a
[...]

Issue #298, opened 2025-11-16
ADR-010: Native SDK Session History Compaction


raphaelmansuy:
# ADR-010: Native SDK Session History Compaction

**Status:** Proposed  
**Date:** 2025-01-16  
**Authors:** Raphaël MANSUY
**Deciders:** Google ADK SDK Architecture Team  

---

## Executive Summary

This ADR specifies the **native implementation** of session history compaction directly within the **Google ADK Go SDK** (`google.golang.org/adk`), matching the proven design from `google.adk` (Python) while leveraging Go's type safety and performance characteristics.

**Critical Design Principle - No Database Changes:**
- ✅ **No schema migration required** - `EventActions` already serialized as flexible JSON/bytes field
- ✅ **No new tables/columns** - Compaction stored as regular event with `actions.compaction` populated
- ✅ **Backward compatible** - Existing databases work without modification
- ✅ **Matches Python ADK** - Python uses pickled actions blob, Go uses JSON bytes (equivalent)

**Strategic Value:**
- ✅ **API Parity**: Matches Python ADK exactly, ensuring consistent behavior across SDKs
- ✅ **Zero Overhead**: No wrapper layers, native JSON serialization handles compaction field
- ✅ **Type Safety**: Compile-time guarantees for compaction metadata structure
- ✅ **Performance**: 60-80% context reduction, application-layer filtering
- ✅ **Developer Experience**: Simple API with sensible defaults, auto-triggered compaction
- ✅ **Minimal Implementation**: Just add struct fields, no database/storage changes

---

### Current State Analysis

| Aspect | Python ADK | Go ADK (Current) | This ADR |
|--------|------------|------------------|----------|
| Compaction Support | ✅ Native (`EventActions.compaction`) | ❌ None | ✅ Native (identical API) |
| Event Filtering | ✅ Automatic | ❌ Manual | ✅ Automatic |
| Token Management | ✅ Auto-trigger on threshold | ❌ Unbounded growth | ✅ Auto-trigger |
| Storage Schema | ✅ `EventCompaction` table | ❌ N/A | ✅ Native GORM support |
| Configuration | ✅ `EventsCompactionConfig` | ❌ N/A | ✅ `CompactionConfig` |

### Problem

Without compaction, Go ADK sessions suffer from:
1. **Exponential Token Growth**: Context doubles every turn (Turn 1: 100 tokens → Turn 3: 450 tokens)
[...]

PR #1233, opened 2026-07-31
feat(telemetry): trace context compaction

baptmont:
Compaction previously left no trace, so there was no way to tell from
production whether it was firing, how much it was shrinking, or whether
the summarizer was failing.

Each summarization now runs inside a "compact_events <trigger>" span
carrying the trigger, session, summarizer type, event count, configured
thresholds and the resulting range.

Three states are distinguishable in a trace:

  - no span: the trigger was evaluated and declined
  - span with no result attributes: the summarizer ran and declined
  - span with error status: the summarizer failed

The span name and the gen_ai.compaction.* attribute keys are part of
ADK's cross-language telemetry contract, so dashboards written against
one implementation work against the others.

## Testing Plan

- `go build`, `go test -race -count=1 -shuffle=on ./...`, `golangci-lint run` and `go mod tidy -diff` are clean on this commit alone.
- Span assertions via an in-memory OTel exporter: span name, attribute values, omission of the strategy that did not fire, error status with a recorded exception, and the distinction between "ran and declined" and "never ran".
## Stack

PR 3 of 6 for context compaction. Targets `baptmont/compaction-02-sliding-window` rather than `main`, so the feature lands on `main` as one pre-reviewed merge from `baptmont/context-compaction`.

| | PR | |
|---|---|---|
| 1 | #1231 | event model + compaction library |
| 2 | #1232 | wire up sliding-window compaction |
| 3 | #1233 | telemetry span |
| 4 | #1234 | token-threshold tail retention |
| 5 | #1235 | reachable from every serving surface |
| 6 | #1236 | end-to-end test against a real model |

PR 3 of 6 is this one.

Every commit in the stack builds and passes `go test -race -count=1 -shuffle=on ./...` on its own.

Part of #298.

### DIFF
--- a/internal/compactioninternal/compactor.go
+++ b/internal/compactioninternal/compactor.go
@@ -21,6 +21,8 @@ import (
 
 	"go.opentelemetry.io/otel/codes"
 
+	"google.golang.org/genai"
+
 	"google.golang.org/adk/v2/internal/telemetry"
 	"google.golang.org/adk/v2/platform"
 	"google.golang.org/adk/v2/session"
@@ -88,6 +90,7 @@ func summarizeTraced(ctx context.Context, cfg *compaction.Config, sess session.S
 		Trigger:            trigger,
 		SessionID:          sessionID,
 		SummarizerType:     summarizerTypeName(cfg.Summarizer),
+		Backend:            summarizerBackend(cfg.Summarizer),
 		EventCount:         len(window),
 		CompactionInterval: cfg.CompactionInterval,
 		OverlapSize:        cfg.OverlapSize,
@@ -190,4 +193,17 @@ func summarizerTypeName(s compaction.Summarizer) string {
 		return name
 	}
 	return t.String()
+}
+
+// summarizerBackend reports which Google backend a Summarizer's model talks to.
+//
+// It is an optional interface rather than a field, matching how the rest of the
+// framework distinguishes Vertex AI from the Gemini API: a third-party
+// Summarizer that has no model, or does not care to say, simply leaves the
+// span's gen_ai.system unset rather than being forced to invent one.
+func summarizerBackend(s compaction.Summarizer) genai.Backend {
+	if v, ok := s.(interface{ GetGoogleLLMVariant() genai.Backend }); ok {
+		return v.GetGoogleLLMVariant()
+	}
+	return genai.BackendUnspecified
 }
--- a/internal/compactioninternal/compactor.go
+++ b/internal/compactioninternal/compactor.go
@@ -1,0 +1,105 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package compactioninternal
+
+import (
+	"context"
+	"fmt"
+
+	"google.golang.org/adk/v2/platform"
+	"google.golang.org/adk/v2/session"
+	"google.golang.org/adk/v2/session/compaction"
+)
+
+// HasSlidingWindow reports whether sliding-window compaction is enabled.
+//
+// This lives here rather than as a method on compaction.Config because nothing
+// outside the framework needs to ask, and keeping it off the public type leaves
+// users with just the fields they set.
+func HasSlidingWindow(cfg *compaction.Config) bool {
+	return cfg != nil && cfg.CompactionInterval > 0
+}
+
+// SlidingWindow summarizes a window of completed invocations once enough of
+// them have accumulated, and returns the resulting compaction event, ready for
+// the caller to append to the session.
+//
+// It returns a nil event, and no error, whenever there is nothing to do: fewer
+// than cfg.CompactionInterval invocations since the last compaction, a window
+// with no self-contained prefix, or a summarizer that declined to produce a
+// summary. Callers treat all three the same way, by leaving history untouched.
+//
+// The runner calls this after an invocation finishes and all of its events have
+// been persisted; compacting mid-invocation is the tail-retention strategy's
+// job.
+func SlidingWindow(ctx context.Context, cfg *compaction.Config, sess session.Session) (*session.Event, error) {
+	if !HasSlidingWindow(cfg) {
+		return nil, nil
+	}
+	if cfg.Summarizer == nil {
+		return nil, fmt.Errorf("no Summarizer configured")
+	}
+	if sess == nil {
+		return nil, nil
+	}
+
+	events := collect(sess)
+	window := selectSlidingWindow(events, cfg.CompactionInterval, cfg.OverlapSize)
+	if len(window) == 0 {
+		return nil, nil
+	}
+
+	summary, err := cfg.Summarizer.SummarizeEvents(ctx, window)
+	if err != nil {
+		return nil, fmt.Errorf("sliding-window summarization failed: %w", err)
+	}
+	return stamp(ctx, summary), nil
+}
+
+// stamp fills in the identity fields a [Summarizer] leaves blank, so the
+// returned event is ready to append.
+//
+// The invocation ID is deliberately fresh rather than borrowed from the covered
+// turns: sliding-window selection counts invocations, and reusing a covered one
+// would skew the next window. Both the ID and the timestamp come from
+// [platform], so a test that installs providers keeps deterministic output.
+func stamp(ctx context.Context, ev *session.Event) *session.Event {
+	if ev == nil {
+		return nil
+	}
+	if ev.ID == "" {
+		ev.ID = platform.NewUUID(ctx)
+	}
+	if ev.InvocationID == "" {
+		ev.InvocationID = "e-" + platform.NewUUID(ctx)
+	}
+	if ev.Timestamp.IsZero() {
+		ev.Timestamp = platform.Now(ctx)
+	}
+	return ev
+}
+
+// collect materializes a session's events into a slice.
+func collect(sess session.Session) []*session.Event {
+	all := sess.Events()
+	if all == nil {
+		return nil
+	}
+	events := make([]*session.Event, 0, all.Len())
+	for ev := range all.All() {
+		events = append(events, ev)
+	}
+	return events
+}
--- a/internal/compactioninternal/compactor.go
+++ b/internal/compactioninternal/compactor.go
@@ -86,9 +86,18 @@ func summarizeTraced(ctx context.Context, cfg *compaction.Config, sess session.S
 	if sess != nil {
 		sessionID = sess.ID()
 	}
+	// The turn that triggered this compaction, taken from the newest event in
+	// the session. The span is not a child of that turn's span, so without an
+	// attribute there is no way to ask which turn a compaction belonged to.
+	//
+	// The newest event rather than the newest one in the window: the window is
+	// what is being summarized, which for tail retention deliberately excludes
+	// the turn in progress, and it is that turn we want to name.
+	invocationID := latestInvocationID(sess)
 	ctx, span := telemetry.StartCompactEventsSpan(ctx, telemetry.StartCompactEventsSpanParams{
 		Trigger:            trigger,
 		SessionID:          sessionID,
+		InvocationID:       invocationID,
 		SummarizerType:     summarizerTypeName(cfg.Summarizer),
 		Backend:            summarizerBackend(cfg.Summarizer),
 		EventCount:         len(window),
@@ -206,4 +215,15 @@ func summarizerBackend(s compaction.Summarizer) genai.Backend {
 		return v.GetGoogleLLMVariant()
 	}
 	return genai.BackendUnspecified
+}
+
+// latestInvocationID returns the invocation of the newest event in sess.
+func latestInvocationID(sess session.Session) string {
+	events := collect(sess)
+	for i := len(events) - 1; i >= 0; i-- {
+		if id := events[i].InvocationID; id != "" {
+			return id
+		}
+	}
+	return ""
 }
--- a/internal/compactioninternal/compactor.go
+++ b/internal/compactioninternal/compactor.go
@@ -65,6 +65,15 @@ func SlidingWindow(ctx context.Context, cfg *compaction.Config, sess session.Ses
 	if err != nil {
 		return nil, fmt.Errorf("sliding-window summarization failed: %w", err)
 	}
+	if summary == nil {
+		return nil, nil
+	}
+	// A Summarizer is third-party code. One that returns an ordinary event
+	// instead of a compaction record would otherwise be appended verbatim,
+	// adding a conversational turn while compacting nothing.
+	if !compaction.IsCompactionEvent(summary) {
+		return nil, fmt.Errorf("summarizer returned an event carrying no compaction record")
+	}
 	return stamp(ctx, summary), nil
 }
 
--- a/internal/compactioninternal/compactor.go
+++ b/internal/compactioninternal/compactor.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 
+	"google.golang.org/adk/v2/internal/telemetry"
 	"google.golang.org/adk/v2/platform"
 	"google.golang.org/adk/v2/session"
 	"google.golang.org/adk/v2/session/compaction"
@@ -61,20 +62,55 @@ func SlidingWindow(ctx context.Context, cfg *compaction.Config, sess session.Ses
 		return nil, nil
 	}
 
-	summary, err := cfg.Summarizer.SummarizeEvents(ctx, window)
+	summary, err := summarizeTraced(ctx, cfg, sess, telemetry.CompactionTriggerSlidingWindow, window)
 	if err != nil {
 		return nil, fmt.Errorf("sliding-window summarization failed: %w", err)
 	}
-	if summary == nil {
-		return nil, nil
+	return summary, nil
+}
+
+// summarizeTraced runs the configured summarizer inside a compact_events span,
+// validates what comes back, and stamps it.
+//
+// Stamping happens before the result is recorded so the span carries a real
+// event ID rather than an empty one. The span covers an actual summarization
+// only, so its presence in a trace means compaction really ran. A trigger that
+// was evaluated and declined produces nothing, which keeps the signal useful.
+func summarizeTraced(ctx context.Context, cfg *compaction.Config, sess session.Session, trigger string, window []*session.Event) (*session.Event, error) {
+	sessionID := ""
+	if sess != nil {
+		sessionID = sess.ID()
 	}
+	ctx, span := telemetry.StartCompactEventsSpan(ctx, telemetry.StartCompactEventsSpanParams{
+		Trigger:            trigger,
+		SessionID:          sessionID,
+		SummarizerType:     fmt.Sprintf("%T", cfg.Summarizer),
+		EventCount:         len(window),
+		CompactionInterval: cfg.CompactionInterval,
+		OverlapSize:        cfg.OverlapSize,
+		TokenThreshold:     cfg.TokenThreshold,
+		EventRetentionSize: cfg.EventRetentionSize,
+	})
+	defer span.End()
+
+	summary, err := cfg.Summarizer.SummarizeEvents(ctx, window)
 	// A Summarizer is third-party code. One that returns an ordinary event
 	// instead of a compaction record would otherwise be appended verbatim,
-	// adding a conversational turn while compacting nothing.
-	if !compaction.IsCompactionEvent(summary) {
-		return nil, fmt.Errorf("summarizer returned an event carrying no compaction record")
+	// adding a conversational turn while compacting nothing. Checked before the
+	// result is recorded so the span shows the failure.
+	if err == nil && summary != nil && !compaction.IsCompactionEvent(summary) {
+		err = fmt.Errorf("summarizer returned an event carrying no compaction record")
+		summary = nil
 	}
-	return stamp(ctx, summary), nil
+	summary = stamp(ctx, summary)
+	telemetry.TraceCompactionResult(span, telemetry.TraceCompactionResultParams{
+		ResultEvent: summary,
+		Error:       err,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return summary, nil
 }
 
 // stamp fills in the identity fields a [Summarizer] leaves blank, so the
--- a/internal/compactioninternal/compactor.go
+++ b/internal/compactioninternal/compactor.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"fmt"
 
+	"go.opentelemetry.io/otel/codes"
+
 	"google.golang.org/adk/v2/internal/telemetry"
 	"google.golang.org/adk/v2/platform"
 	"google.golang.org/adk/v2/session"
@@ -91,7 +93,18 @@ func summarizeTraced(ctx context.Context, cfg *compaction.Config, sess session.S
 		TokenThreshold:     cfg.TokenThreshold,
 		EventRetentionSize: cfg.EventRetentionSize,
 	})
-	defer span.End()
+	// A Summarizer is third-party code and may panic. The OTel SDK records an
+	// exception event on the way out but leaves the status Unset, which reads
+	// as success, so a panicking summarizer would look like a healthy one that
+	// happened to produce nothing. Mark it and let the panic continue.
+	defer func() {
+		if r := recover(); r != nil {
+			span.SetStatus(codes.Error, fmt.Sprintf("summarizer panicked: %v", r))
+			span.End()
+			panic(r)
+		}
+		span.End()
+	}()
 
 	summary, err := cfg.Summarizer.SummarizeEvents(ctx, window)
 	// A Summarizer is third-party code. One that returns an ordinary event
@@ -102,7 +115,15 @@ func summarizeTraced(ctx context.Context, cfg *compaction.Config, sess session.S
 		err = fmt.Errorf("summarizer returned an event carrying no compaction record")
 		summary = nil
 	}
-	summary = stamp(ctx, summary)
+	// Stamped only once the result is known to be usable. A summarizer can
+	// return an event alongside an error, and that event is discarded, so
+	// stamping first spent a UUID on it and handed telemetry the identity of
+	// something that never reached the session.
+	if err != nil {
+		summary = nil
+	} else {
+		summary = stamp(ctx, summary)
+	}
 	telemetry.TraceCompactionResult(span, telemetry.TraceCompactionResultParams{
 		ResultEvent: summary,
 		Error:       err,
--- a/internal/compactioninternal/compactor.go
+++ b/internal/compactioninternal/compactor.go
@@ -17,6 +17,7 @@ package compactioninternal
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"go.opentelemetry.io/otel/codes"
 
@@ -86,7 +87,7 @@ func summarizeTraced(ctx context.Context, cfg *compaction.Config, sess session.S
 	ctx, span := telemetry.StartCompactEventsSpan(ctx, telemetry.StartCompactEventsSpanParams{
 		Trigger:            trigger,
 		SessionID:          sessionID,
-		SummarizerType:     fmt.Sprintf("%T", cfg.Summarizer),
+		SummarizerType:     summarizerTypeName(cfg.Summarizer),
 		EventCount:         len(window),
 		CompactionInterval: cfg.CompactionInterval,
 		OverlapSize:        cfg.OverlapSize,
@@ -168,4 +169,25 @@ func collect(sess session.Session) []*session.Event {
 		events = append(events, ev)
 	}
 	return events
+}
+
+// summarizerTypeName is the bare type name of a Summarizer, without package
+// qualifier or pointer marker.
+//
+// The reference implementation puts type(summarizer).__name__ on this span, so
+// "LLMSummarizer" is what a consumer joining traces across implementations
+// expects to match against. Sprintf("%T") would emit
+// "*compaction.LLMSummarizer", which names a Go type rather than a summarizer.
+func summarizerTypeName(s compaction.Summarizer) string {
+	if s == nil {
+		return ""
+	}
+	t := reflect.TypeOf(s)
+	for t.Kind() == reflect.Pointer {
+		t = t.Elem()
+	}
+	if name := t.Name(); name != "" {
+		return name
+	}
+	return t.String()
 }
--- a/internal/compactioninternal/telemetry_test.go
+++ b/internal/compactioninternal/telemetry_test.go
@@ -82,7 +82,7 @@ func TestSlidingWindowEmitsSpan(t *testing.T) {
 		"gen_ai.operation.name":             "compact_events",
 		"gen_ai.conversation.id":            "sess",
 		"gen_ai.compaction.trigger":         "sliding_window",
-		"gen_ai.compaction.summarizer_type": "*compactioninternal.fakeSummarizer",
+		"gen_ai.compaction.summarizer_type": "fakeSummarizer",
 		"gen_ai.compaction.result_event_id": got.ID,
 	} {
 		if a[key].AsString() != want {
@@ -106,13 +106,16 @@ func TestSlidingWindowEmitsSpan(t *testing.T) {
 	// and it must be the right range in the right layout. Asserting only that
 	// the attributes are non-empty left the layout, the bound each one is
 	// sourced from, and the timestamps themselves all unprotected.
-	wantStart := at(1).Format(time.RFC3339Nano)
-	wantEnd := at(4).Format(time.RFC3339Nano)
-	if got := a["gen_ai.compaction.start_timestamp"].AsString(); got != wantStart {
-		t.Errorf("start_timestamp = %q, want %q", got, wantStart)
+	// Epoch seconds as a float, matching the reference implementation. The type
+	// is asserted as well as the value, because emitting these as strings is the
+	// defect this pins and a string attribute reads back as zero here.
+	wantStart := float64(at(1).UnixNano()) / float64(time.Second)
+	wantEnd := float64(at(4).UnixNano()) / float64(time.Second)
+	if got := a["gen_ai.compaction.start_timestamp"]; got.Type() != attribute.FLOAT64 || got.AsFloat64() != wantStart {
+		t.Errorf("start_timestamp = %v (%v), want %v (FLOAT64)", got.Emit(), got.Type(), wantStart)
 	}
-	if got := a["gen_ai.compaction.end_timestamp"].AsString(); got != wantEnd {
-		t.Errorf("end_timestamp = %q, want %q", got, wantEnd)
+	if got := a["gen_ai.compaction.end_timestamp"]; got.Type() != attribute.FLOAT64 || got.AsFloat64() != wantEnd {
+		t.Errorf("end_timestamp = %v (%v), want %v (FLOAT64)", got.Emit(), got.Type(), wantEnd)
 	}
 	if got := a["gen_ai.compaction.result_event_id"].AsString(); got == "" {
 		t.Error("result_event_id is empty, so a trace cannot be joined to the stored summary")
--- a/internal/compactioninternal/telemetry_test.go
+++ b/internal/compactioninternal/telemetry_test.go
@@ -1,0 +1,180 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package compactioninternal
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+
+	"google.golang.org/adk/v2/internal/telemetry"
+	"google.golang.org/adk/v2/session"
+	"google.golang.org/adk/v2/session/compaction"
+)
+
+// spanRecorder installs an in-memory tracer for the calling test.
+func spanRecorder(t *testing.T) *tracetest.InMemoryExporter {
+	t.Helper()
+	exp := tracetest.NewInMemoryExporter()
+	telemetry.OverrideTracerForTesting(t, sdktrace.NewTracerProvider(sdktrace.WithSyncer(exp)))
+	return exp
+}
+
+// attrs flattens a span's attributes for lookup by key.
+func attrs(kvs []attribute.KeyValue) map[string]attribute.Value {
+	out := make(map[string]attribute.Value, len(kvs))
+	for _, kv := range kvs {
+		out[string(kv.Key)] = kv.Value
+	}
+	return out
+}
+
+func TestSlidingWindowEmitsSpan(t *testing.T) {
+	exp := spanRecorder(t)
+
+	events := []*session.Event{
+		textEvent("a", "inv1", 1, "q1"), modelTextEvent("b", "inv1", 2, "a1"),
+		textEvent("c", "inv2", 3, "q2"), modelTextEvent("d", "inv2", 4, "a2"),
+	}
+	cfg := &compaction.Config{CompactionInterval: 2, OverlapSize: 1, Summarizer: &fakeSummarizer{summary: "sum"}}
+
+	got, err := SlidingWindow(context.Background(), cfg, &staticSession{events: events})
+	if err != nil {
+		t.Fatalf("SlidingWindow() error = %v", err)
+	}
+	if got == nil {
+		t.Fatal("SlidingWindow() produced no summary")
+	}
+
+	spans := exp.GetSpans()
+	if len(spans) != 1 {
+		t.Fatalf("got %d spans, want 1", len(spans))
+	}
+	span := spans[0]
+	if want := "compact_events sliding_window"; span.Name != want {
+		t.Errorf("span name = %q, want %q", span.Name, want)
+	}
+	if span.Status.Code == codes.Error {
+		t.Errorf("span status = error, want unset: %v", span.Status)
+	}
+
+	a := attrs(span.Attributes)
+	for key, want := range map[string]string{
+		"gen_ai.operation.name":             "compact_events",
+		"gen_ai.conversation.id":            "sess",
+		"gen_ai.compaction.trigger":         "sliding_window",
+		"gen_ai.compaction.summarizer_type": "*compactioninternal.fakeSummarizer",
+		"gen_ai.compaction.result_event_id": got.ID,
+	} {
+		if a[key].AsString() != want {
+			t.Errorf("attribute %s = %q, want %q", key, a[key].AsString(), want)
+		}
+	}
+	if a["gen_ai.compaction.event_count"].AsInt64() != 4 {
+		t.Errorf("event_count = %d, want 4", a["gen_ai.compaction.event_count"].AsInt64())
+	}
+	if a["gen_ai.compaction.compaction_interval"].AsInt64() != 2 {
+		t.Errorf("compaction_interval = %d, want 2", a["gen_ai.compaction.compaction_interval"].AsInt64())
+	}
+	if a["gen_ai.compaction.overlap_size"].AsInt64() != 1 {
+		t.Errorf("overlap_size = %d, want 1", a["gen_ai.compaction.overlap_size"].AsInt64())
+	}
+	// Only the strategy that fired should be described.
+	if _, ok := a["gen_ai.compaction.token_threshold"]; ok {
+		t.Error("token_threshold attribute is present on a sliding-window span, want it omitted")
+	}
+	// The range must be recorded so a trace shows what the summary replaced.
+	if a["gen_ai.compaction.start_timestamp"].AsString() == "" {
+		t.Error("start_timestamp attribute is empty")
+	}
+	if a["gen_ai.compaction.end_timestamp"].AsString() == "" {
+		t.Error("end_timestamp attribute is empty")
+	}
+}
+
+func TestCompactionSpanRecordsFailure(t *testing.T) {
+	exp := spanRecorder(t)
+
+	events := []*session.Event{
+		textEvent("a", "inv1", 1, "q1"), modelTextEvent("b", "inv1", 2, "a1"),
+		textEvent("c", "inv2", 3, "q2"), modelTextEvent("d", "inv2", 4, "a2"),
+	}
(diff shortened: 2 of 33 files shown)


internal/telemetry/compaction.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"google.golang.org/adk/v2/session"
+)
+
+const compactEventsName = "compact_events"
+
+// timestampLayout renders compaction range bounds. Event timestamps are
+// time.Time, and RFC 3339 with nanoseconds preserves their full precision while
+// staying readable in a trace viewer.
+const timestampLayout = time.RFC3339Nano
```
karolpiotrowicz: [blocker] `start_timestamp` and `end_timestamp` are emitted as RFC3339Nano **strings**, while adk-python emits the same two attribute keys as **floats of epoch seconds** — `adk-python/src/google/adk/events/event_actions.py:68-72` declares them `start_timestamp: float` / `end_timestamp: float`, and `adk-python/src/google/adk/telemetry/tracing.py:507,509` assigns them straight into the attribute dict. OTLP carries a typed union, so a backend stores one as a numeric measure and the other as a string label: the obvious panel — `end - start`, i.e. how much wall-clock history each summary swallowed — works against Python and silently returns nothing for Go, with no error. That makes the PR description's claim that "dashboards written against one implementation work against the others" false for these two attributes, even though all ten `gen_ai.compaction.*` keys and the span name do match Python exactly. Two smaller defects ride along in the same encoding choice: `RFC3339Nano` strips trailing fractional zeros, so lexicographic order disagrees with time order (`2026-01-01T00:00:00Z` sorts *after* `2026-01-01T00:00:00.5Z`), and `Format` is called without `.UTC()` at `:136-137`, so the host timezone reaches the wire (a Warsaw-located event timestamp is emitted as `2026-01-01T01:00:00+01:00`). Suggest matching Python — `genAICompactionStartTimestamp.Float64(float64(ts.UnixNano()) / 1e9)` — and, if a human-readable form is wanted in a trace viewer, adding it under a separate Go-only key so the contract key stays numeric.

internal/telemetry/compaction.go
```diff
@@ -0,0 +1,139 @@
...
+	// Error is the summarization failure, if any.
+	Error error
+}
+
+// TraceCompactionResult records the outcome of a compaction on span.
+//
+// A nil ResultEvent with a nil Error is a summarizer that declined; the span is
+// left successful with no result attributes, which distinguishes "ran and
+// produced nothing" from "ran and failed".
+func TraceCompactionResult(span trace.Span, params TraceCompactionResultParams) {
```
karolpiotrowicz, line 165: [major] The result attributes are set whenever `ResultEvent != nil && ResultEvent.Actions.Compaction != nil`, regardless of `params.Error`: `:128` records the error and then `:131` falls through to `:134-138` instead of returning. A `Summarizer` that returns `(partialEvent, err)` therefore produces a single span that is simultaneously `codes.Error` and carrying a `result_event_id` — for an event that `internal/compactioninternal/compactor.go:103` discards and `runner/runner.go:235-241` never appends, so the ID exists in no session and a drill-down from it finds nothing. Driving that summarizer through the real `SlidingWindow` path yields `status=Error desc="boom"` alongside `gen_ai.compaction.result_event_id`, `.start_timestamp` and `.end_timestamp` on the same span. adk-python cannot produce this shape — an exception from the summarizer at `adk-python/src/google/adk/apps/compaction.py:61` skips `span.set_attributes(_build_compaction_result_attributes(...))` entirely — and it falsifies the disjoint three-state model in the PR description. One line fixes it: `if params.Error != nil { recordErrorAndStatus(span, params.Error); return }`. The shipped `LLMSummarizer` always returns `(nil, err)`, so this is bounded to user-supplied summarizers.

internal/compactioninternal/telemetry_test.go
```diff
@@ -0,0 +1,180 @@
...
+	if a["gen_ai.compaction.event_count"].AsInt64() != 4 {
+		t.Errorf("event_count = %d, want 4", a["gen_ai.compaction.event_count"].AsInt64())
+	}
+	if a["gen_ai.compaction.compaction_interval"].AsInt64() != 2 {
+		t.Errorf("compaction_interval = %d, want 2", a["gen_ai.compaction.compaction_interval"].AsInt64())
+	}
+	if a["gen_ai.compaction.overlap_size"].AsInt64() != 1 {
+		t.Errorf("overlap_size = %d, want 1", a["gen_ai.compaction.overlap_size"].AsInt64())
+	}
+	// Only the strategy that fired should be described.
```
karolpiotrowicz: [major] These assertions pin that the attributes *exist*, not what they contain, which leaves the contract keys unprotected at exactly the points where they have already drifted. Changing `timestampLayout` to `time.RFC1123`, or sourcing `start_timestamp` from `EndTimestamp` so both keys report the end instant, leaves both packages green; so does renaming `gen_ai.compaction.token_threshold` or `gen_ai.compaction.event_retention_size`, because the fixture leaves `TokenThreshold` at 0 and the branch at `internal/telemetry/compaction.go:105-109` is never taken. Renaming `gen_ai.compaction.start_timestamp` or `.event_count` *does* fail the suite, so the keys these assertions name are pinned — their values are not, and two of the ten contract keys have no positive assertion at all. Suggested strengthening: install a deterministic time provider and assert the exact timestamp literals rather than non-emptiness; add a positive case that configures `TokenThreshold`/`EventRetentionSize` and asserts both keys, both `INT64` values and the `token_threshold` span name; and assert the *full* attribute key set with `cmp.Diff` against a sorted expected slice, so any added attribute — beneficial or a content leak — shows up.

internal/compactioninternal/compactor.go
```diff
@@ -61,11 +62,47 @@ func SlidingWindow(ctx context.Context, cfg *compaction.Config, sess session.Ses
...
+// Stamping happens before the result is recorded so the span carries a real
+// event ID rather than an empty one. The span covers an actual summarization
+// only, so its presence in a trace means compaction really ran. A trigger that
+// was evaluated and declined produces nothing, which keeps the signal useful.
+func summarizeTraced(ctx context.Context, cfg *compaction.Config, sess session.Session, trigger string, window []*session.Event) (*session.Event, error) {
+	sessionID := ""
+	if sess != nil {
+		sessionID = sess.ID()
+	}
+	ctx, span := telemetry.StartCompactEventsSpan(ctx, telemetry.StartCompactEventsSpanParams{
```
karolpiotrowicz, line 97: [major] The span is started from the context `compactAfterInvocation` was handed, which is the *outer* context: `runner/runner.go:405` has already drained `rootAgent.Run(ctx)` by then, and a Go context is a value, so the `invoke_agent` span created inside `Run` cannot propagate back out. Run end-to-end through the real `Runner`, the compaction span comes out as a **root span in its own trace ID** (three distinct trace IDs across a two-turn run); with an ambient caller span it becomes a **sibling of every `invoke_agent`** under that parent rather than a child of the turn that triggered it. There is no attribute fallback either — the span omits `gcp.vertex.agent.invocation_id`, which both `invoke_agent` (`internal/telemetry/telemetry.go:96`) and the node spans (`internal/telemetry/node_tracing.go:126`) set. So whether compaction fired for *this* turn, and what it cost that turn, is not answerable from the emitted data, which is one of the three questions the PR description says the span now answers; being a root span also means it takes an independent head-sampling decision from the turn. Setting `gcp.vertex.agent.invocation_id` here restores correlation without restructuring the runner; `trace.WithLinks(trace.LinkFromContext(...))` would additionally let a viewer draw the edge. `runner/run_node.go:198` has the identical shape.

internal/telemetry/compaction.go
```diff
@@ -0,0 +1,139 @@
...
+// The span name and the gen_ai.compaction.* attribute keys are part of ADK's
+// cross-language telemetry contract, so a dashboard or query written against
+// one ADK implementation works against the others. Renaming them here breaks
+// that, so treat them as fixed.
+//
+// The span wraps only the summarizer call, so its presence in a trace means
+// compaction really ran. A run that evaluated a trigger and declined produces
+// no span, which keeps the signal meaningful.
+func StartCompactEventsSpan(ctx context.Context, params StartCompactEventsSpanParams) (context.Context, trace.Span) {
+	attrs := []attribute.KeyValue{
```
karolpiotrowicz, line 114: [major] `gen_ai.system` is absent from this attribute set, while adk-python sets it on every compaction span (`adk-python/src/google/adk/telemetry/tracing.py:474`: `GEN_AI_SYSTEM: _guess_gemini_system_name()`, resolving to `gemini` or `vertex_ai`). The emitted Go key set is exactly the ten `gen_ai.compaction.*` keys plus `gen_ai.conversation.id` and `gen_ai.operation.name`, so a dashboard whose global filter is `gen_ai.system = "gemini"` — the shape the ADK observability docs push you toward, where `gen_ai.system` is a key attribute on every documented span — drops the Go compaction span from every panel. Two mitigations are worth weighing: this is a repo-wide gap rather than one this change introduces (no non-test Go source in the repo sets the attribute), and the OTel semantic-conventions package now annotates `gen_ai.system` as deprecated in favour of `gen_ai.provider.name`. Either add `semconv.GenAISystemKey.String(...)` here *and* track the repo-wide gap, or narrow the contract comment at `:82-85` to cover keys rather than the full attribute set — adding it here alone should not be treated as satisfying the contract.

internal/telemetry/compaction.go
```diff
@@ -0,0 +1,139 @@
...
+	// ResultEvent is the compaction event produced, or nil when the summarizer
+	// declined. Its identity fields must already be stamped.
+	ResultEvent *session.Event
+	// Error is the summarization failure, if any.
+	Error error
+}
+
+// TraceCompactionResult records the outcome of a compaction on span.
+//
+// A nil ResultEvent with a nil Error is a summarizer that declined; the span is
```
karolpiotrowicz, line 162: [major] When `ResultEvent == nil && Error == nil` the span is left `Unset` with no result attributes, so the only signal for that state is an attribute being *absent* — which cannot be counted, graphed or alerted on. For the shipped `LLMSummarizer` that state is not the benign "ran and declined" the PR description describes: `SlidingWindow` only calls the summarizer with a non-empty window (`internal/compactioninternal/compactor.go:61-63`), so the sole remaining way to reach `(nil, nil)` is the model returning no content at all — safety block, empty candidates, quota-degraded response (`session/compaction/llm_summarizer.go:134-137` deliberately maps that to `nil, nil`). A summarizer failing 100% of the time therefore reports a 0% error rate while session history silently never shrinks, and the two causes of absence (healthy decline vs. empty model response) are conflated. adk-python produces the identical shape (`_build_compaction_result_attributes(None)` returns `{}`), so making the outcome first-class — `gen_ai.compaction.outcome = summarized | empty_response | error`, or at minimum `span.AddEvent("summarizer_returned_no_content")` — is worth coordinating with adk-python rather than adding Go-only.

karolpiotrowicz (commented):
The span is a genuine improvement over no telemetry at all, and the three-state
design is the right shape. Two things undercut it.

**1. The cross-language claim is not true as written.** The PR says dashboards
written against one implementation work against the others, and that sentence
also ships as a doc comment at `compaction.go:82-85`. But
`gen_ai.compaction.start_timestamp` / `end_timestamp` are emitted here as
RFC3339Nano **strings** (`internal/telemetry/compaction.go:34,:136-137`), while
adk-python puts a **float of epoch seconds** under the same keys
(`event_actions.py:68-72`, assigned raw in `tracing.py:505-509`). Same key,
incompatible OTLP value type — a strict-typed backend drops one or the other, so
a shared dashboard cannot work. Two smaller issues ride along: the timestamps
are not `.UTC()`-normalised, so the host timezone reaches the wire (observed
`2026-01-01T01:00:00+01:00`), and RFC3339Nano sorts lexicographically in the
opposite order to time for mixed offsets. Emitting epoch seconds matches Python
and fixes all three.

**2. The three states are not actually distinguishable.** Result attributes are
recorded on the error path too, so one span reports both "the summarizer failed"
and "it produced summary X" — and `result_event_id` there names an event that was
discarded. Separately, "ran and declined" is signalled only by the *absence* of
attributes, and for the shipped `LLMSummarizer` that absence is exactly what a
degraded model response produces — so a summarizer failing every call looks like
a 0% error rate. A panic leaves the span status `Unset`.

Two follow-ups worth considering while the file is open: the span is emitted as
its own trace root with no `invocation_id` (3 distinct trace ids in a single
run), so a failure cannot be traced back to the turn that caused it; and the
summarizer's `UsageMetadata` is on the very event handed to the recorder but is
dropped, which is why "how much it was shrinking" is still unanswerable. The
tests pin attribute names but not values — an RFC1123 format change ships green,
along with 7 other mutations.

---

### Additional findings

[minor] The summarizer's own model call is untraced: `session/compaction/llm_summarizer.go:122` calls `s.model.GenerateContent(ctx, req, false)` directly, bypassing the repo's only instrumentation site (`StartGenerateContentSpan` is declared at `internal/telemetry/telemetry.go:92` and has exactly one caller, `internal/llminternal/base_flow.go:844`). That call is billable and appears in no `generate_content` span, so every cost and latency dashboard built on those spans under-counts by exactly the compaction traffic — end-to-end with the default summarizer, a run makes 3 model calls but produces 2 `generate_content` spans and 1 `compact_events` span, and the summarizer's model, finish reason and per-call latency are recorded nowhere. `llm_summarizer.go` is not touched by this change, which is why this is minor rather than a defect of the diff, but it is a gap a change whose stated purpose is making compaction observable is well placed to close. At minimum, put `gen_ai.request.model` and the token counts on the `compact_events` span; better, route the summarizer's call through the same instrumentation so it becomes a child `generate_content` span.

internal/telemetry/compaction.go
```diff
@@ -0,0 +1,139 @@
...
+// left successful with no result attributes, which distinguishes "ran and
+// produced nothing" from "ran and failed".
+func TraceCompactionResult(span trace.Span, params TraceCompactionResultParams) {
+	recordErrorAndStatus(span, params.Error)
+
+	ev := params.ResultEvent
+	if ev == nil || ev.Actions.Compaction == nil {
+		return
+	}
+	span.SetAttributes(
```
karolpiotrowicz, line 190: [major] This reads `ev.ID` and `ev.Actions.Compaction.{Start,End}Timestamp` and drops `ev.LLMResponse.UsageMetadata`, which is populated on the very event handed to the recorder: `session/compaction/summary_event.go:50-63` builds it with `LLMResponse: model.LLMResponse{UsageMetadata: usage}` from the live response. The complete emitted set is ten attributes, of which `event_count` is how many events went *in* and the rest are config knobs — there is no before-size, after-size, token count or byte count, so compression ratio and tokens saved are uncomputable and "how much it was shrinking", the PR description's second motivating question, remains unanswerable. The sibling `TraceGenerateContentResult` (`internal/telemetry/telemetry.go:119-129`) already shows the pattern, and mirroring it here is about three lines using data the function already holds: `if u := ev.LLMResponse.UsageMetadata; u != nil { span.SetAttributes(semconv.GenAIUsageInputTokens(...), semconv.GenAIUsageOutputTokens(...)) }`. adk-python emits no token attributes either, so this would be a Go-side extension worth coordinating across implementations.

internal/telemetry/compaction.go
```diff
@@ -0,0 +1,139 @@
...
+// no span, which keeps the signal meaningful.
+func StartCompactEventsSpan(ctx context.Context, params StartCompactEventsSpanParams) (context.Context, trace.Span) {
+	attrs := []attribute.KeyValue{
+		semconv.GenAIOperationNameKey.String(compactEventsName),
+		semconv.GenAIConversationID(params.SessionID),
+		genAICompactionTrigger.String(params.Trigger),
+		genAICompactionSummarizerType.String(params.SummarizerType),
+		genAICompactionEventCount.Int(params.EventCount),
+	}
+	// Omit a threshold that is not configured, so a span shows only the
```
karolpiotrowicz: [minor] The comment says the threshold belonging to the strategy that did not fire is omitted, but the branches gate on `params.CompactionInterval > 0` and `params.TokenThreshold > 0` — that is, on what is *configured*, not on what fired. Both strategies together is a legal configuration (`session/compaction/compaction.go:98-124` rejects negatives and the two half-set combinations but explicitly permits both), and with both set a `sliding_window` span carries `token_threshold` and `event_retention_size` as well, even though tail retention did not fire. The behaviour matches adk-python (`apps/compaction.py:48-57` passes all four config values through regardless of the trigger, and each non-`None` one is emitted), so keying the gating off `params.Trigger` would create a *new* divergence — the fix is the wording, not the code. Suggest correcting this comment, the matching comment at `internal/compactioninternal/telemetry_test.go:99`, and the corresponding line in the PR description to say *configured*; the test currently passes only because its fixture leaves `TokenThreshold` at 0.

internal/compactioninternal/compactor.go
```diff
@@ -61,11 +62,47 @@ func SlidingWindow(ctx context.Context, cfg *compaction.Config, sess session.Ses
...
+		Trigger:            trigger,
+		SessionID:          sessionID,
+		SummarizerType:     fmt.Sprintf("%T", cfg.Summarizer),
+		EventCount:         len(window),
+		CompactionInterval: cfg.CompactionInterval,
+		OverlapSize:        cfg.OverlapSize,
+		TokenThreshold:     cfg.TokenThreshold,
+		EventRetentionSize: cfg.EventRetentionSize,
+	})
+	defer span.End()
```
karolpiotrowicz: [minor] `defer span.End()` with no recover means a panicking summarizer ends the span with `Status.Code = Unset`: the OTel SDK does add an `exception` event, but nothing sets `codes.Error`, so status-based failure counting reads the crash as a success. Reproduced with a panicking summarizer: `span ended=true status=Unset events=1`, the single event being `exception`. The span itself is not leaked — `defer` sits immediately after the start — but note that `internal/compactioninternal/telemetry_test.go:132-134` asserts only that the span has events, which a panic also satisfies, so the PR's own failure test cannot distinguish a panic from a returned error. A small shared helper in `internal/telemetry` would cover it: recover, `span.SetStatus(codes.Error, fmt.Sprint(r))`, `span.End()`, then re-panic.

internal/compactioninternal/compactor.go
```diff
@@ -61,11 +62,47 @@ func SlidingWindow(ctx context.Context, cfg *compaction.Config, sess session.Ses
...
+// was evaluated and declined produces nothing, which keeps the signal useful.
+func summarizeTraced(ctx context.Context, cfg *compaction.Config, sess session.Session, trigger string, window []*session.Event) (*session.Event, error) {
+	sessionID := ""
+	if sess != nil {
+		sessionID = sess.ID()
+	}
+	ctx, span := telemetry.StartCompactEventsSpan(ctx, telemetry.StartCompactEventsSpanParams{
+		Trigger:            trigger,
+		SessionID:          sessionID,
+		SummarizerType:     fmt.Sprintf("%T", cfg.Summarizer),
```
karolpiotrowicz: [minor] `fmt.Sprintf("%T", cfg.Summarizer)` puts a Go type *expression* on a contract key where adk-python puts a bare class name (`adk-python/src/google/adk/apps/compaction.py:51`: `summarizer_type=type(config.summarizer).__name__`). Observed Go values are `*compactioninternal.fakeSummarizer` and `compactioninternal.partialSummarizer`; in production the default is `*compaction.LLMSummarizer` against Python's `LlmEventSummarizer`. A `group by gen_ai.compaction.summarizer_type` panel therefore renders disjoint series for the same conceptual summarizer, and a `= "LlmEventSummarizer"` filter excludes Go entirely. Exact value parity is not achievable — the class names genuinely differ across languages — so the realistic options are shape parity (`reflect.TypeOf(...).Elem().Name()` for a bare name) or stating in the doc comment that `summarizer_type` is language-local and outside the shared contract.

internal/telemetry/compaction.go
```diff
@@ -0,0 +1,139 @@
...
+// TraceCompactionResult records the outcome of a compaction on span.
+//
+// A nil ResultEvent with a nil Error is a summarizer that declined; the span is
+// left successful with no result attributes, which distinguishes "ran and
+// produced nothing" from "ran and failed".
+func TraceCompactionResult(span trace.Span, params TraceCompactionResultParams) {
+	recordErrorAndStatus(span, params.Error)
+
+	ev := params.ResultEvent
+	if ev == nil || ev.Actions.Compaction == nil {
```
karolpiotrowicz, line 176: [minor] This treats `ev.Actions.Compaction == nil` as "no result", while the runner's test is `summary == nil` (`runner/runner.go:235-240`, which then calls `AppendEvent`). The two disagree, so for an event in that shape the session gains an event while the span reports the declined shape — base attributes only, no `result_event_id`, `status=Unset` — i.e. telemetry states the opposite of what happened to the session. The shape is an in-tree concept rather than hypothetical: `session/compaction/compaction.go:147-149` (`IsCompactionEvent`) and its doc explicitly distinguish declaring a compaction from that compaction having content. It is not reachable with the shipped summarizer, since `NewSummaryEvent` always sets `Actions.Compaction`, so this only affects custom summarizers. Consider keying the declined branch on `ev == nil` alone and recording `result_event_id` plus an explicit marker when the action is missing, or rejecting such an event upstream in `summarizeTraced`.

internal/compactioninternal/compactor.go
```diff
@@ -61,11 +62,47 @@ func SlidingWindow(ctx context.Context, cfg *compaction.Config, sess session.Ses
...
+		EventCount:         len(window),
+		CompactionInterval: cfg.CompactionInterval,
+		OverlapSize:        cfg.OverlapSize,
+		TokenThreshold:     cfg.TokenThreshold,
+		EventRetentionSize: cfg.EventRetentionSize,
+	})
+	defer span.End()
+
+	summary, err := cfg.Summarizer.SummarizeEvents(ctx, window)
+	summary = stamp(ctx, summary)
```
karolpiotrowicz: [minor] `stamp(ctx, summary)` now runs *before* the `if err != nil` check at `:102`, whereas previously stamping was reached only after the error check. On a failed summarization that returned a partial event, the event is therefore mutated in place (ID, InvocationID, Timestamp) and two `platform.NewUUID(ctx)` draws plus one `platform.Now(ctx)` now happen where there were none — measured with a counting provider: 2 UUID draws on that path, 0 before — which shifts every subsequent generated ID in a fixed-sequence or record-replay fixture, on an object this code then discards. With the conventional `(nil, err)` return `stamp` returns immediately at `:116-118` and draws nothing, so this is bounded to partial-event summarizers. The doc comment at `:74-76` justifies stamping before recording so the span carries a real event ID, and that is still satisfied by stamping only on the success path — checking `err` first also fixes the error-path result attributes flagged at `internal/telemetry/compaction.go:127`.

internal/telemetry/compaction.go
```diff
@@ -0,0 +1,139 @@
...
+		attrs = append(attrs,
+			genAICompactionInterval.Int(params.CompactionInterval),
+			genAICompactionOverlapSize.Int(params.OverlapSize))
+	}
+	if params.TokenThreshold > 0 {
+		attrs = append(attrs,
+			genAICompactionTokenThreshold.Int(params.TokenThreshold),
+			genAICompactionEventRetention.Int(params.EventRetentionSize))
+	}
+	return tracer.Start(ctx, fmt.Sprintf("%s %s", compactEventsName, params.Trigger), trace.WithAttributes(attrs...))
```
karolpiotrowicz, line 148: [minor] This span is the only channel the change adds, and the decline paths (`internal/compactioninternal/compactor.go:49-63`) emit nothing at all — so the three questions the PR description raises (is it firing, is the summarizer failing, how much is it shrinking) are all *rate* questions being answered from head-sampled spans with no denominator. There is no unsampled aggregate channel to fall back on: the repo has zero metric instruments — a search across non-test Go sources for `Int64Counter`, `Float64Counter`, the histogram constructors, `otel.Meter` and `MeterProvider` matches only a doc comment at `telemetry/telemetry.go:62`. Compounding it, the compaction span is usually a root span (see the note on `internal/compactioninternal/compactor.go:84`), so it samples independently of the turn. Standing up a `MeterProvider` is fairly clearly out of scope for this change — but it is worth either saying so in the PR description, or emitting an OTel log record per compaction through the existing `otelLogger` (`internal/telemetry/logger.go:56-60`), which is unsampled and needs no new infrastructure.

internal/compactioninternal/compactor.go
```diff
@@ -61,11 +62,47 @@ func SlidingWindow(ctx context.Context, cfg *compaction.Config, sess session.Ses
 		return nil, nil
```
karolpiotrowicz, line 67: [minor] The span does not start until `:84`, after `collect(sess)` has walked every event in the session and `selectSlidingWindow` has scanned it, and `defer span.End()` at `:94` fires before the append at `runner/runner.go:238`. Its duration is therefore summarizer latency rather than compaction latency, and on a large session the untraced prefix can dominate. Because `compactAfterInvocation` has no span of its own, and compaction runs inside the `Run` iterator on the caller's critical path (`runner/runner.go:449`, `runner/run_node.go:198`), that time appears nowhere in the turn's trace — so "turn p99 moved after the compaction rollout, where did it go?" is not answerable from it. Moving this span earlier would cost a span per declined trigger, which the change deliberately avoids, so the better shape is a parent span in `compactAfterInvocation` covering selection, summarization and append; that would also address the correlation gap noted at `:84`.

internal/telemetry/compaction.go
```diff
@@ -0,0 +1,139 @@
...
+	CompactionInterval int
+	OverlapSize        int
+	TokenThreshold     int
+	EventRetentionSize int
+}
+
+// StartCompactEventsSpan starts a span covering one context-compaction
+// summarization, named "compact_events <trigger>".
+//
+// The span name and the gen_ai.compaction.* attribute keys are part of ADK's
```
karolpiotrowicz: [minor] The cross-language contract this comment asserts has exactly one counterpart implementation and no written artifact. adk-java has compaction (`Compaction.java`, `EventCompaction.java`, `TailRetentionEventCompactor.java`) and zero compaction telemetry — searching that repository for `compact_events` or `gen_ai.compaction` returns no hits; adk-docs has no compaction row in its trace-schema table (`docs/observability/traces.md:24-29`); and adk-python carries no reciprocal statement. Nothing enforces the contract on either side, which is how it drifted on the timestamp value type, `gen_ai.system` and `summarizer_type` before review. Suggest narrowing this comment and the PR description to name adk-python specifically, and adding the compaction span to the adk-docs trace schema with the keys **and their value types** — which would also give those three divergences one place to be settled for all implementations.

internal/telemetry/compaction.go
baptmont: Already closed upstream, by `8e8b1ddf` on #1231. `compactor.go:101-104` now turns a summarizer result that is not an `IsCompactionEvent` into an error and nils the event, so `SlidingWindow` errors and the runner never appends. Since `IsCompactionEvent` requires both a record and content, no surviving event can reach the `ev.Actions.Compaction == nil` branch this comment is about.

internal/telemetry/compaction.go
baptmont: The harm you describe here is gone, though not from this PR. `llm_summarizer.go` now returns an *error* for a no-usable-content response rather than `(nil, nil)`, so a summarizer failing every call no longer reports a 0% error rate, and the safety, recitation and token-limit stops surface instead of being reported as "nothing to compact". `(nil, nil)` now means only an empty window, which the compactor never passes.

What survives is the narrower design point you make: "ran and declined" is still signalled only by an absent attribute, and it now takes a custom summarizer to reach. That is on the telemetry list with the rest, not fixed here.

baptmont:
Triaged all fifteen against the current code rather than replying from the summary. One is already fixed, one is partly overtaken by upstream changes, and thirteen are open. Nothing in this PR's own code has been revised yet, so that is expected.

Two replies posted inline where the ground has moved under a comment (`3701643756`, `3701643740`).

The blocker (`3701643730`, timestamps as RFC3339Nano strings rather than epoch floats, and no `.UTC()`) I have deliberately not touched yet, because it is a contract decision and the parity half of it cannot be checked from this repo. I can confirm the Go-side facts: the host zone does reach the wire, and stripped fractional zeros make lexicographic order disagree with time order. If you can confirm what adk-python emits for those two keys I will match it exactly.

Same for `3701643738` (`gen_ai.system`) and `3701643752` (`%T` versus a bare class name): both are parity claims I can implement immediately once confirmed, and both are one-liners.

`3701643731` (error status and result attributes on the same span) and `3701643758` (stamping before the error check) are the same defect from two sides and I will fix them together with the one-line early return. `3701643733` is a real test gap and I reproduced your mutation result.

internal/telemetry/compaction.go
baptmont: Fixed. `TraceCompactionResult` now returns immediately once the error has been recorded, so a failed compaction never carries `result_event_id` or the range. The producing side is fixed too: `summarizeTraced` no longer stamps an event that came back alongside an error, since the caller discards it (that is `3701643758`, the same defect from the other side, and both go away with the one guard).

Regression test: `TestCompactionSpanOmitsResultWhenSummarizerAlsoErrors`, using a summarizer that returns a valid compaction event *and* an error, which is your scenario. Against the previous code it reports all three attributes on an error span:

```
gen_ai.compaction.result_event_id = "fa7e5f7a-..."
gen_ai.compaction.start_timestamp = "2026-01-01T00:00:01Z"
gen_ai.compaction.end_timestamp   = "2026-01-01T00:00:04Z"
```

Worth noting the existing failure test could not have caught this: its fake returns `(nil, err)`, so there was never a result to leak.

internal/compactioninternal/compactor.go
baptmont: Fixed together with `3701643731`, which is the same defect seen from the producing side. `stamp` now runs only when `err == nil`, and the event is dropped otherwise, so a discarded result no longer costs a UUID or reaches telemetry.

Your counting probe was the useful detail here: two UUID draws on that path is what made it obvious the event was being prepared and then thrown away.

internal/compactioninternal/compactor.go
baptmont: Fixed. The span now ends through a deferred `recover` that sets `codes.Error` with the panic value before re-panicking, so a panicking summarizer no longer produces a span that reads as success. The panic itself still propagates unchanged.

Regression test: `TestCompactionSpanMarksAPanic`, which asserts both halves, the `Error` status and that the panic still escapes. Against the previous code it reports `span status = Unset`.

You were right that the existing assertion could not catch it: `len(Events) != 0` is satisfied by the SDK's own exception event, which a panic produces too.

internal/telemetry/compaction.go
baptmont: Fixed, by correcting the comment rather than the gating. Both strategies can legitimately be configured at once, so "the strategy that actually fired" was never what the branches expressed. The comment now says the span carries only the knobs in play, and points at `Trigger` as the attribute that names which strategy produced the span.

internal/compactioninternal/telemetry_test.go
baptmont: Fixed, and I reproduced your mutation results first to be sure the assertions were as weak as you described.

`TestSlidingWindowEmitsSpan` now asserts the exact rendered values rather than non-emptiness, so both mutations you named are caught:

- `RFC3339Nano` to `RFC1123` now fails with `start_timestamp = "Thu, 01 Jan 2026 00:00:01 UTC", want "2026-01-01T00:00:01Z"`.
- Sourcing `start_timestamp` from `EndTimestamp` now fails with `start_timestamp = "2026-01-01T00:00:04Z", want "2026-01-01T00:00:01Z"`.

It also asserts `result_event_id` is non-empty, so a trace can be joined to the stored summary.

The two tail-retention keys are now covered as well, but in #1234 rather than here: `TailRetention` does not exist in this commit, so a test for them cannot compile against this branch. I wrote it here first, found that out, and moved it to the first commit that has a producer. `TestCompactionSpanRecordsTailRetentionThresholds` asserts both values and that the sliding-window knob stays off the span; renaming `token_threshold` now fails it with `token_threshold = 0, want 10`.

internal/telemetry/compaction.go
baptmont: Confirmed against the source and fixed. I could not verify the parity half earlier, so I went and read adk-python rather than leave it open.

`EventCompaction.start_timestamp` is declared `float` ("in seconds") in `src/google/adk/events/event_actions.py`, and `_build_compaction_result_attributes` in `src/google/adk/telemetry/tracing.py` puts that float straight on the span under the same key. So the divergence is exactly as you described: same key, different type.

Both bounds are now emitted as epoch seconds via `Float64`. That removes the zone offset and the sort-order problem with it, since the two side defects you found were both consequences of rendering a string.

The assertion pins the attribute *type* as well as the value, so reverting to a string fails with `start_timestamp = 2026-01-01T00:00:01Z (STRING), want 1.767225601e+09 (FLOAT64)` rather than passing silently on a coincidence.

internal/compactioninternal/compactor.go
baptmont: Confirmed and fixed. adk-python passes `summarizer_type=type(config.summarizer).__name__` (`src/google/adk/apps/compaction.py`), so the contract value is a bare class name.

Go now emits the bare type name, dereferencing pointers, so the default summarizer reports `LLMSummarizer` rather than `*compaction.LLMSummarizer`. Reverting to `%T` fails the assertion.

internal/telemetry/compaction.go
baptmont: Confirmed missing, but I have not added it, because reading both implementations turned up an ambiguity I would rather you settle than guess at.

You are right that adk-python sets it: `_build_compaction_attributes` starts with `GEN_AI_SYSTEM: _guess_gemini_system_name()`. The complication is the value. Python's helper returns `GenAiSystemValues.VERTEX_AI.name.lower()` or `GEMINI.name.lower()`, so `"vertex_ai"` or `"gemini"`. This repo is on semconv v1.36.0, where `GenAISystemGCPVertexAI` and `GenAISystemGCPGemini` are `"gcp.vertex_ai"` and `"gcp.gemini"`. I verified those two values rather than assuming them.

So matching the key is easy and matching the value is a choice between the two semconv generations, which is not compaction's to make: no Go span sets `gen_ai.system` today, only `logger.go` does and only on log records. Adding it to the compaction span alone would make it the only span in the Go implementation carrying it.

There is a second problem: Python decides from a global enterprise-mode check, while the Go compaction path has no `genai.Backend` to inspect, so there is nothing local to derive it from.

Happy to do whichever you prefer, including plumbing the backend through: (a) Go's semconv value, consistent within Go, (b) Python's literal value, consistent across implementations, or (c) leave it until `gen_ai.system` is set repo-wide.

baptmont: Now done, after settling the value question I raised above.

The span carries `gen_ai.system`, sourced from the summarizer through the same optional interface the rest of the framework uses to tell Vertex AI from the Gemini API, rather than a parameter threaded down from the runner. A third-party `Summarizer` with no model leaves the attribute unset instead of being made to invent one.

I went with this repo's semconv values, so `gcp.gemini` and `gcp.vertex_ai`, not adk-python's `gemini` and `vertex_ai`. Mixing semconv generations inside one implementation seemed worse than differing from another implementation on a value, and the gap is repo-wide rather than compaction's to close. It is recorded in the cross-language notes as a deliberate divergence, so whichever side moves first can align the other.

Table-driven test covers both backends and a summarizer that reports neither.

baptmont:

## Landed

| Decision | Where | Note |
|---|---|---|

## Four things worth flagging

## Still open

internal/compactioninternal/compactor.go
baptmont: Not done.

Defect stands. `StartCompactEventsSpanParams` (`internal/telemetry/compaction.go:69-89`) has no invocation-ID field; no `gcp.vertex.agent.invocation_id` and no `trace.WithLinks` anywhere in `compaction.go` or `compactor.go`. Span still starts from the outer ctx at `internal/compactioninternal/compactor.go:89`, called from `runner/runner.go:257`.

You are right and it is still open. Recorded in the audit of this review rather than quietly dropped; leaving the thread unresolved.

internal/telemetry/compaction.go
baptmont: Not done.

Defect stands. No `UsageMetadata` or `GenAIUsage*` reference anywhere in `internal/telemetry/compaction.go`; the result branch at `:165-169` sets three attributes only.

You are right and it is still open. Recorded in the audit of this review rather than quietly dropped; leaving the thread unresolved.

internal/telemetry/compaction.go
baptmont: Not done.

Defect stands. A repo-wide grep for `Int64Counter`, `Float64Counter`, `Int64Histogram`, `otel.Meter`, `MeterProvider` in non-test Go matches only the doc comment at `telemetry/telemetry.go:62`. No log record is emitted per compaction either.

You are right and it is still open. Recorded in the audit of this review rather than quietly dropped; leaving the thread unresolved.

internal/compactioninternal/compactor.go
baptmont: Not done.

Defect stands. `collect` and `selectSlidingWindow` run at `internal/compactioninternal/compactor.go:64-65`, before the span starts at `:89`; the span ends at `:110` while the append happens later at `runner/runner.go:282`. No parent span wraps `compactAfterInvocation` (`runner/runner.go:234`).

You are right and it is still open. Recorded in the audit of this review rather than quietly dropped; leaving the thread unresolved.

internal/telemetry/compaction.go
baptmont: Not done.

Defect stands. `internal/telemetry/compaction.go:94-97` still says a query "written against one ADK implementation works against the others", unnarrowed. No docs artifact was added; the only in-repo record of the divergences is the inline comment at `:110-114`.

You are right and it is still open. Recorded in the audit of this review rather than quietly dropped; leaving the thread unresolved.

baptmont:

## Thread state

| PR | Resolved | Left open |
|---|---:|---:|
| #1231 | 16 | 4 |
| #1232 | 13 | 6 |
| #1233 | 8 | 7 |
| #1234 | 8 | 13 |
| #1235 | 3 | 5 |
| #1236 | 4 | 3 |

## State of the code

internal/telemetry/compaction.go
baptmont: Done. The twin comment in the test said "fired" too and I missed it the first time; both now say the span carries only the knobs of the configured strategy, and point at the trigger attribute as the thing that names which one ran.

internal/telemetry/compaction.go
baptmont: Done. The span carries `gen_ai.usage.input_tokens` and `gen_ai.usage.output_tokens` when the summarizer reports them.

Your framing is the reason it matters: compaction spends a model call in order to save tokens later, so a span that does not record what it spent cannot show whether it paid for itself.

internal/compactioninternal/compactor.go
baptmont: Partly done, and the useful half is in.

The span now carries `gcp.vertex.agent.invocation_id`, so "did compaction fire for this turn" is answerable. It is taken from the newest event in the session rather than the newest in the window, deliberately: the window is what is being summarized, and tail retention excludes the turn in progress from it, which is exactly the turn worth naming.

Not done: the span is still started from the outer context, so it remains a root span rather than a child of the turn. That is a change to how the runner threads its context and I would rather not fold it into this.

baptmont: Closed, though not the way I first tried, and the attempt is worth reporting.

I made the hook start the span from the invocation's context so it would nest under the turn. It made no difference, and a test showed why: the outer context already carries whatever span the caller had active, so the compaction span already joins the caller's trace. I reverted it.

Making it a child of `invoke_agent` specifically is not right either. That span has ended by the time compaction runs, since compaction happens after the invocation completes. A span that starts after its parent ends is a worse description of what happened than a sibling.

What was genuinely missing is the ability to join them, and that is now there: the span carries `gcp.vertex.agent.invocation_id`. A test pins both properties, that compaction stays in the caller's trace and that it names the invocation it followed.

internal/compactioninternal/telemetry_test.go
baptmont: Now fully closed. The whole attribute key set is compared against an expected list, so adding, renaming or dropping a key fails rather than passing unnoticed. The individual assertions only ever checked the keys they named.

internal/telemetry/compaction.go
baptmont: Accepted as is.

You are right that rates cannot be computed from sampled spans. Adding metric instruments is a repo-wide observability decision rather than a compaction one: there are no meters anywhere in this codebase today, and compaction should not be the first thing to introduce them on its own terms.

For what it is worth, adk-python emits spans only too, and adk-kotlin emits neither. Recorded in the notes.

internal/compactioninternal/compactor.go
baptmont: Looked at this and decided against it, so here is the reasoning rather than silence.

Starting the span before selection would emit a span for every evaluation that declines, which breaks the property that a span in a trace means compaction actually ran. That property is pinned by a test and is the more useful signal.

What the span would gain is the time for an in-memory scan and window selection, microseconds against a model call. The doc comment now says explicitly what the span covers and why, instead of leaving the scope unexplained.

internal/telemetry/compaction.go
baptmont: Narrowed to what is true. The parity with adk-python is real, and I verified it from source rather than assuming: the ten attribute keys, the span name and the operation name are identical there. But nothing enforces the agreement, and adk-kotlin has no compaction telemetry at all, so "cross-language contract" meant two implementations rather than three. The comment says that now, and the key set has a test.

karolpiotrowicz (commented):
Reviewed at `f838737`. The doc-only commits pushed since then do not change any line cited below.

The attribute set and span name do match adk-python, and the claim in the doc comment that they were read from the Python source rather than assumed checks out. Four things stop this delivering what the description promises.

**1. A compaction that was discarded emits a span that says it succeeded.** The span is ended inside [`summarizeTraced`](https://github.com/google/adk-go/blob/f838737b6292316cfd834f99e735f68b132fd003/internal/compactioninternal/compactor.go#L97-L120) , before the five paths in `compactAfterInvocation` that can throw the summary away. With a plugin that rejects compaction events, the run returned `context compaction failed: plugin rejected the summary event`, nothing was stored, and the span was `status=Unset` carrying `result_event_id`, `start_timestamp` and `end_timestamp` — an ID that exists in no session. The other four paths (cancelled context, reload failure, `RangeRaced`, append failure) have the same shape.

**2. The span is a root in its own trace by default.** Started three turns with no ambient caller span and got six distinct trace IDs, with the compaction span's parent invalid every time; with a caller span it becomes a sibling of `invoke_agent`, never a child. Under head sampling the two take independent decisions, so a sampled turn usually loses its compaction span, and compaction spans turn up as orphan single-span traces with nothing to join them to. `gcp.vertex.agent.invocation_id` is the only fallback correlator, and it is inferred from the newest session event rather than passed down — with two invocations in flight on one session, both compaction spans can carry the same invocation ID, so at least one of them names a turn that did not cause it. The authoritative ID is already in hand at the call site that invokes the compactor.

**3. Two attributes diverge from adk-python in presence and in meaning.** `gen_ai.system` is set [only for the two Google backends](https://github.com/google/adk-go/blob/f838737b6292316cfd834f99e735f68b132fd003/internal/telemetry/compaction.go#L129-L134) , while `adk-python` sets it unconditionally in the dict literal — over a non-Google model the attribute is simply absent here, and a test pins that absence. And [`gen_ai.usage.output_tokens`](https://github.com/google/adk-go/blob/f838737b6292316cfd834f99e735f68b132fd003/internal/telemetry/compaction.go#L182-L189) is candidates-only, where `TraceGenerateContentResult` in the same package sets the same key as candidates + thoughts with a semconv citation. Two spans in one trace mean different things by one key. The `> 0` gating also makes a genuine zero absent rather than zero.

**4. A zero timestamp is emitted as the year 1754.** [`epochSeconds`](https://github.com/google/adk-go/blob/f838737b6292316cfd834f99e735f68b132fd003/internal/telemetry/compaction.go#L40-L42) on a zero `time.Time` gives `-6.795e+09`. This is reachable through the shipped path, not only a hand-built record: `NewSummaryEvent` takes `events[0].Timestamp` verbatim and the in-memory service does not stamp a missing timestamp, so I got `end - start ≈ 271 years` for three seconds of history, on a span reporting success. adk-python omits the key when the bound is absent; matching that is better than substituting 1970.

On the tests: the key-set assertion pins the eleven attributes the minimal fixture happens to emit, and a legal both-strategies config emits fourteen. Renaming `gen_ai.compaction.token_threshold`, renaming `event_retention_size`, and flipping `token_threshold` from `INT64` to a string all leave the suite green — a type flip on a shared key is exactly the silent cross-language break the description warns about. Adding an attribute that carries conversation text on any conditional branch is green too, so the no-content-leak property has nothing enforcing it. The sub-second timestamp path is unpinned because the fixture lands on whole seconds, and `TestCompactionSpanJoinsTheCallersTrace` manufactures the parent it then asserts exists, and never reads the correlation attribute.

Two smaller notes. The panic path sets an error status but records no exception event, so an alert keyed on `exception.type` misses every panicking summarizer. And the summarizer's own model call emits no span at all — `compact_events` is a childless leaf wrapping an untraced LLM call, so when a compaction fails after spending tokens there is no child span to recover the cost from.

One thing not to change: the "lossless" `epochSeconds` rewrite (`float64(t.Unix()) + float64(t.Nanosecond())/1e9`) is not lossless. Measured across 2000 consecutive nanoseconds at current epoch magnitudes it improves the worst-case error from 244 ns to 119 ns and leaves ~200 ns granularity, because 61 bits do not fit in a 53-bit mantissa however the arithmetic is arranged. Ordering is preserved in both forms. Float seconds is the cross-language contract, so this is a cost of the contract rather than a defect.

baptmont:
Fixes are in #1316.

**Finding 1.** Fixed, and it needed more than moving the `span.End()`. Summarization is not over when the summarizer returns, so the span now stays open until the caller says what became of the summary. A discarded one is recorded on the same key as a decline, because to anything reading the trace they mean the same thing: compaction was wanted, a model call was spent, and the prompt did not shrink. All five paths report now, and the one that used to publish a `result_event_id` for an event no session holds no longer names a result at all.

**Finding 2.** Half fixed, and I want to be straight about which half.

The correlation attribute is fixed: the caller passes the invocation ID instead of the compactor reading the newest event in the session. Your concurrency case was exactly right, both compactions read the same event and at least one named a turn that did not cause it.

The parenting is not fixed. I tried attaching the invocation's span context in the runner and it was a no-op, because the agent derives `spanCtx` internally in `agent.Run` and passes it only to its own children, so the runner never sees it. Reverted rather than left in as dead code. The test says this is a known gap now instead of starting a caller span and then asserting a parent exists, which was fair criticism.

**Finding 3.** Output tokens fixed: candidates plus thoughts, matching `TraceGenerateContentResult` in the same package. You are right that a thinking model's summary was under-reported and that two spans in one trace disagreed on one key.

`gen_ai.system` I left as it is, and I think that is the right call. The rest of this repo's telemetry omits it for a backend it cannot name, and naming a provider we have not identified seemed worse than saying nothing. What I did do is delete the second copy of the mapping so compaction uses the same one as everything else, and move the reasoning to the test that pins it, so a repo-wide change has to update this rather than discover it in a dashboard.

**Finding 4.** Fixed. A zero bound is omitted rather than published as 1754. Reachable through the shipped path, as you said, since nothing on the append path stamped a missing timestamp.

**On the tests.** The three rename and type-flip cases were already covered by the time I got here. The content-leak branch and the sub-second path are still not pinned.

**The two notes.** The panic path records an exception event now. The summarizer's own model call still emits no span.

**epochSeconds.** Left alone. I had my agent check your numbers with `big.Float` and they hold: worst case 242ns to 119ns, roughly 222ns granularity either way, no order inversions in either form. Thanks for pre-empting that one, it is exactly the change someone would have made later on a reading of the comment.

baptmont:
Closing. Telemetry is no longer a separable slice of this stack.

Review of the stack changed the compactor so that it calls telemetry directly: the span is opened before summarization and stays open until the caller reports what became of the summary, which is what stopped the five discard paths reporting success for an event that reached no session. That means `compactor.go` cannot compile without the telemetry package, so telemetry can no longer come *after* the engine in the chain.

Keeping this PR would have meant hand-splitting `compactor.go` into a version with the spans removed and re-adding them here, which is a boundary the code no longer has.

The telemetry work is unchanged and now lands in #1232 alongside the engine that calls it. Nothing was dropped: the whole stack tip is byte-identical to what it was before the re-cut.

The review threads here stay readable. Sorry for the churn.

PR #1234, opened 2026-07-31
feat(compaction): compact mid-invocation once the prompt crosses a threshold

baptmont:
Third of five. **Re-cut**: this PR previously showed code that later PRs
corrected, and its base moved from `compaction-03-telemetry` (now closed) to
`compaction-02-sliding-window`.

## What this adds

Tail retention: the strategy that actually bounds prompt size.

The sliding window in #1232 replaces each group of invocations with one summary,
and summaries are never re-summarized, so it reduces prompt size by a constant
factor rather than bounding it. Tail retention seeds each new window with the
previous summary, so history stays one rolling summary plus a raw tail however
long the conversation runs.

It runs inside an invocation, before a model call, once the prompt passes
`TokenThreshold`. Running mid-turn is what lets it catch a single long
tool-calling turn that inflates the prompt on its own, which a post-invocation
strategy cannot see until the turn is over.

## Two things worth reviewing closely

**The live turn's question is held back.** Summarizing it means summarizing the
instruction currently being carried out, and `EventRetentionSize` cannot protect
it, because it counts events and a turn is not a fixed number of them: one tool
round costs two. Only that one event is held back, so a long tool loop can still
compact its own traffic.

**A blocked head does not stall it.** One unanswered call among parallel
siblings used to stop the strategy for the rest of the session, silently, since
"no window" and "nothing to do yet" are both nil. The scan now resumes wherever
the set of open obligations changed, not only where it grew.

## Enable this or the sliding window, not both

They share a candidate rule. Tail retention summarizes the events no compaction
already covers, and the sliding window covers everything it reaches, so with
both enabled tail retention never finds enough uncovered events to fire and the
ceiling never applies. Measured over 160 turns with a 520-character summary:
tail retention alone holds the prompt flat at about 550 characters, the two
together grow it to 41,000.

This is documented rather than rejected in `Validate`. adk-python starves its
own token-threshold strategy the same way, so it is the shared design, and
AGENTS.md makes adk-python the source of truth for behaviour.

## Testing plan

`go build`, `go test -race -count=1 -shuffle=on`, `golangci-lint run` and
`go mod tidy -diff` clean in both modules, verified at this slice specifically.
Known-unrelated failure: `TestConfigureExporters` in `./telemetry`.

## Changed since the last review

**Your first finding reproduces, and I was wrong to say otherwise.** I reported
that I could not make the rolling summary fail end to end and asked for your
scenario. My probes never built it: the live-head exclusion only lands inside
[...]

PR #1235, opened 2026-07-31
feat(server): make context compaction reachable from every serving surface

baptmont:
Fourth of five. **Re-cut**: this PR previously showed code that later PRs
corrected. It now stands on its own.

## What this adds

The `Compaction` config threaded through the REST server, the Pub/Sub and
Eventarc triggers, the launcher, A2A and Agent Engine, so an application gets
the same compaction behaviour whichever way it is served.

## Fail at startup, not on every request

Every surface refuses a config it cannot serve, by dry-running `runner.New` per
application at construction. The config is validated inside `runner.New` and a
runner is built per request, so without this a process starts cleanly, reports
healthy, and returns 500 on 100% of traffic with an error naming nothing an
operator can act on. `&compaction.Config{}` is the natural "enable it with
defaults" spelling and is exactly the shape that fails.

## A compaction failure does not fail the turn

Compaction runs after the agent has answered and after its events are persisted,
so a failure means only that a later prompt will be larger. Each surface
recognises `ErrCompaction`, logs it and carries on, rather than returning a 500
that discards the answer, streaming an error to a client that already received
one, or NACKing a Pub/Sub message that was handled.

## Two limitations, documented rather than hidden

**Triggers.** Each delivery gets a session of its own, so history never
accumulates and a sliding-window interval is never reached. Configuring one
there is logged at construction, saying which strategy does work instead.

**Scope.** The setting is process-wide. One server serves many applications
through its agent loader and they all get the same config and the same
`Summarizer`, so the same model and the same quota. Applications needing
different compaction should run separately. adk-python scopes this per `App`;
mirroring that means threading it through the loader, which is a larger change
than this stack should carry.

## API compatibility

The three exported controller constructors keep the signatures they were
released with, and options live on `...WithOptions` siblings. A trailing
variadic is source-compatible only for plain calls: any use of the identifier as
a value breaks. Each constructor is pinned by assignment to an explicit function
type, so re-adding a variadic fails to compile.

## Testing plan

`go mod tidy -diff` clean in both modules, verified at this slice specifically.
Known-unrelated failure: `TestConfigureExporters` in `./telemetry`.

## Changed since the last review

**Agent Engine and A2A run the startup dry run now**, through
`internal/compactionvalidate` rather than a third and fourth copy of the loop,
so all four surfaces give one answer to "can this config serve these apps".
Those two serve only `loader.RootAgent()`, so they use a root-only variant: the
shared version would refuse to start over an application they never run.

**The wiring has tests that fail when it is removed.** I measured which lines
were actually protected by deleting each one and running the suite: the REST
pair was caught, both Agent Engine sites were not. Both now count summarizer
calls, which is what distinguishes "compaction ran and its failure was
tolerated" from "compaction never ran", and I checked each fails when its own
line is deleted.

**The console launcher tolerates `ErrCompaction`**, so a compaction failure no
longer prints `AGENT_ERROR` under an answer the user has already read. It was
the last surface without that branch.

**The stale "prompts stay small as a conversation grows" wording is gone**, here
and in `runner.Config`. There are no copies left.

Not done: the `console`, `web/api` and `a2a` launcher wiring lines still have no
behavioural test. They need a launcher with an interactive loop to drive, and a
test that passes whether or not the wiring is there would be worse than none.

**Review this exact commit: `6f01f04a5fa10533eba93d04f61bf65ed4904f5e`.** Check out the SHA, not a branch
name. Several branches in this repository have similar names and hold older
versions of this same code: `baptmont/compaction-03-telemetry` (`ff5b84bb`)
and `baptmont/compaction-07-review-fixes` (`888ee5c1`) belong to closed PRs
from an earlier shape of the stack, and the pre-re-cut tips are still reachable.
Two agents auditing this stack have already spent a full run on the wrong tree
and reported bugs that no longer exist. If `internal/compactioninternal` has no
`coverIndex` and no `RepairAfterAppend`, you are on the wrong commit.

**Deliberate decisions, so they are not re-reported as defects.** Each is
documented at the code:

- `HasUsableSummary` is only a nil check. A record whose content holds no prose
  still covers its range and deletes those events. The stricter check was
  written and reverted: adk-python checks only `is None`, this stack tracks it,
  and nothing here can produce such a record.
- Tail-retention summaries do not run the plugin pipeline. adk-python runs no
  plugin hook on either compaction path, so this matches it.
- The repair after the append is one round. A straggler landing during the
  repair would need another, and each round strictly shrinks what the record
  covers on both axes -- the timestamp range and the stream position. It shrinks
  only the range if the positional guard is taken from the surviving record, and
  that was a real defect, fixed in round 6 by `foldCorrections`.
- `session/database` orders ties by `id`, which is stable but arbitrary. True
  insertion order needs a sequence column, which Vertex AI would not carry.
- The sliding window and tail retention do not compose; enabling both is
  documented as leaving the prompt unbounded, and adk-python behaves the same.

**Known flakes**, both unrelated to this stack and both passing on re-run:
`TestDelegation_04_ChatToTwoSingleTurnParallel` (issue #1313) and an occasional
golangci-lint runner error reporting `no export data for github.com/google/uuid`.

**Verification:** build, `go test -race -count=1 -shuffle=on`,
`golangci-lint run` and `go mod tidy -diff` are clean at this commit in both
modules, and CI runs on this stack now.

### Decisions taken since round 3, so they are not re-reported

Each is documented at the code. Raise them if you disagree, but they are choices
rather than oversights:

- **Parallel fan-out costs 8x on tail retention.** Eight children each pass the
  gate and seven summaries are discarded. The output and the prompt are correct;
  it is model spend. Making siblings share a gate walks back the per-agent
  scoping that fixed the loop-agent case, so it stands.
- **`session/database` breaks a timestamp tie on `id DESC`.** Two random UUIDs,
  which is not conversational order. adk-python orders identically, and after
  the repair fix nothing correctness-critical depends on tie order.
- **The sliding-window path runs the plugin pipeline and tail retention does
  not.** adk-python runs no plugin hook on either, so the first is a deliberate
  divergence in the safe direction and the second matches the reference. The gap
  is documented on the exported surface, at `compaction.Config.TokenThreshold`.
- **Two records over an identical range are resolved by stream position.** This
  cannot tell a re-summarization from two independent summaries that happen to
  span one instant, because the record stores holes rather than membership.
  Preferring the later one keeps the documented property that a re-summarized
  range never appears twice, at the cost of the rarer case. Holes are unioned
  across records that share a range and the same summary text, which is what a
  correction is, so a correction cannot lose its straggler to an ordering tie.
- **`TestNoEventIsDroppedWithoutHavingBeenSummarized` is breadth, not a
  substitute for the targeted tests.** It found a real unsoundness during
  development, and it does not currently generate the shapes that reproduce the
  tie-safe cut, the hole union or the content check. Its comment says so.
- **The two `raw_event` Vertex recordings pin a whole `session.Event` JSON dump**
  and are brittle to any new field on that struct. Out of scope here: not
  compaction code.

## Round 6 — self-review before re-requesting review

Four independent reviewers were run over the stack before asking for another
round. Every fix below is mutation-tested: the mutation that reproduces the
original defect is named in the test's doc comment.

**Two defects that deleted conversation.**

- `copyAny` returned any payload that was not `map[string]any`/`[]any`
  untouched. A tool payload here is not decoded JSON -- it is whatever a Go
  handler returned, stored as-is -- so `loadmemorytool`, which stores
  `[]memory.Entry` holding a `*genai.Content`, handed a summarizer live pointers
  into stored history. Now a depth-bounded reflective deep copy.
- The repair pass covered an event it never saw. A correction shrinks coverage
  on the timestamp axis and moves forward on the stream axis; the positional
  guard used the survivor's position, so an event appended between a record and
  its correction was outside the original's reach and inside the survivor's.
  Holes were already folded across the correction group and position was not;
  `foldCorrections` now folds both, taking the group's earliest position.

**Four more.** `newCoverIndex` indexed contentless records that prompt assembly
skips, leaving events permanently raw and permanently uncompactable at once;
tail retention drew its window from the whole session while the threshold that
triggered it was scope-filtered, so a sub-agent could spend its one compaction on
a sibling branch; token usage was dropped on the error path, so the most
expensive failure mode reported zero spend; and the corrective write ran on the
caller's context, so a cancellation between the two appends left a record
claiming a range nothing had summarized.

**Four tests that could not fail** were found and strengthened, including the
alias guard itself, which compared only top-level `genai.Part` pointers and so
was blind to a fresh `Blob` holding the original's byte slice.

**Known to remain.** When the only uncovered candidates ahead of the retained
tail sit inside the previous range -- what a repair manufactures -- the rolling
seed can produce a range strictly narrower than the record it replaces, so
neither subsumes the other and both summaries materialize in one prompt. It is
self-limiting and costs a wasted compaction rather than data. Left deliberately:
both candidate fixes carry hole-inheritance subtleties whose failure mode is
deleting conversation, so it wants its own change and its own review.


server/adkrest/controllers/runtime.go
```diff
@@ -42,11 +43,33 @@ type RuntimeAPIController struct {
...
+		c.eventsCompactionConfig = cfg
+	}
 }
 
 // NewRuntimeAPIController creates the controller for the Runtime API.
-func NewRuntimeAPIController(sessionService session.Service, memoryService memory.Service, agentLoader agent.Loader, artifactService artifact.Service, sseTimeout time.Duration, pluginConfig runner.PluginConfig, autoCreateSession bool) *RuntimeAPIController {
-	return &RuntimeAPIController{sessionService: sessionService, memoryService: memoryService, agentLoader: agentLoader, artifactService: artifactService, sseTimeout: sseTimeout, pluginConfig: pluginConfig, autoCreateSession: autoCreateSession}
+func NewRuntimeAPIController(sessionService session.Service, memoryService memory.Service, agentLoader agent.Loader, artifactService artifact.Service, sseTimeout time.Duration, pluginConfig runner.PluginConfig, autoCreateSession bool, opts ...RuntimeAPIOption) *RuntimeAPIController {
+	c := &RuntimeAPIController{sessionService: sessionService, memoryService: memoryService, agentLoader: agentLoader, artifactService: artifactService, sseTimeout: sseTimeout, pluginConfig: pluginConfig, autoCreateSession: autoCreateSession}
+	for _, opt := range opts {
```
karolpiotrowicz: [blocker] The trailing `opts ...RuntimeAPIOption` changes the *type* of `NewRuntimeAPIController`, not just what a call site is allowed to pass, so this is a breaking change to the exported API — and the same applies to `NewPubSubController` (`server/adkrest/controllers/triggers/pubsub.go:39`) and `NewEventarcController` (`server/adkrest/controllers/triggers/eventarc.go:40`). A trailing variadic is source-compatible only for plain calls; every use of the identifier as a *value* breaks:

```go
var _ func(session.Service, memory.Service, agent.Loader, artifact.Service,
    time.Duration, runner.PluginConfig, bool) *controllers.RuntimeAPIController = controllers.NewRuntimeAPIController
```

Code of that shape — variable declaration, func-typed argument, struct field, map value — compiles cleanly against the base branch and fails against this one with 6 errors across the three constructors, and an API-compatibility check reports all three as incompatible. `reflect`-driven wiring degrades to a run-time break instead of a compile error: `NumIn` goes 7→8 and `IsVariadic` false→true. Nothing in-repo broke because every in-repo caller is a plain call, but that says nothing about downstream: the REST routers live under `server/adkrest/internal/`, so calling these constructors directly is the only supported way for a third party to mount the controllers on their own mux, and `google.golang.org/adk/v2` carries no pre-release marker. The PR description's "Every existing caller compiles unchanged, which tests pin" is not accurate — `server/adkrest/controllers/runtime_test.go:282` and `server/adkrest/controllers/triggers/options_test.go:28` are plain call sites, which a trailing variadic can never break, so they pass identically with and without this change and pin nothing. Suggest carrying the config on a config struct field (or a separate `NewRuntimeAPIControllerWithOptions`) instead, and adding an API-compatibility check to CI so the next one is caught mechanically.

server/adkrest/controllers/triggers/triggers.go
```diff
@@ -42,6 +43,25 @@ type RetriableRunner struct {
...
+// WithEventsCompactionConfig enables context compaction for the runners a
+// trigger controller creates, so older session events are summarized and
+// prompts stay small as a conversation grows. See [compaction.Config].
+func WithEventsCompactionConfig(cfg *compaction.Config) ControllerOption {
+	return func(r *RetriableRunner) {
+		r.eventsCompactionConfig = cfg
+	}
 }
 
 func (r *RetriableRunner) RunAgent(ctx context.Context, appName, userID, messageContent string) ([]*session.Event, error) {
```
karolpiotrowicz, line 135: [major] Every trigger delivery creates a fresh session, so a sliding window that needs at least `CompactionInterval` distinct invocations can never be satisfied on the Pub/Sub and Eventarc surfaces. Driving `RetriableRunner.RunAgent` — the function both handlers call — six times gives `sessionsCreated=6` with `compactionEvents=0` and `summarizerCalls=0` at interval 2 and interval 3; at interval 1 it is 6 and 6, i.e. one summarizer call per delivery over a single-turn session that is then discarded, which is pure cost for no benefit. Tail retention does not rescue the surface either: `TokenThreshold: 1, EventRetentionSize: 1` on a single delivery still produced `summarizerCalls=0` on that workload, though a long multi-step tool-calling delivery is untested, so tail retention is unproven rather than refuted. This contradicts the PR description's surface list — two of the listed surfaces accept the config and do nothing with it, with no warning at any point. Suggest either rejecting `CompactionInterval > 0` at construction on a per-delivery-session surface, or giving triggers a durable session keyed by the message correlation ID so a window can accumulate; at minimum, document the limitation.

cmd/launcher/console/console.go
```diff
@@ -104,12 +104,13 @@ func (l *consoleLauncher) Run(ctx context.Context, config *launcher.Config) erro
 	sess := resp.Session
 
 	r, err := runner.New(runner.Config{
```
karolpiotrowicz, line 107: [major] This is the one surface that fails fast on the compaction config. Neither `adkrest.NewServer` nor `agentengine.NewHandler` (`server/agentengine/handler.go:39`) calls `EventsCompactionConfig.Validate()`, so an invalid config is accepted at startup and instead turns into a 500 on every subsequent request: a config that fails its own `Validate()` ("no compaction strategy is enabled…") still gets `err=nil` from `NewServer`, after which the server binds, reports healthy, and returns `500 failed to create runner: invalid EventsCompactionConfig` on 100% of traffic. `&compaction.Config{}` is the natural "enable it with defaults" spelling and is exactly the shape guaranteed to fail, and `runner/runner.go:146-147` documents fail-fast as the intent — so the REST and Agent Engine paths are the inconsistent ones, not this one. Calling `Validate()` in `adkrest.NewServer`, `agentengine.NewHandler` and `launcher.Execute` and returning the error would make the failure surface once, at startup, which is where a rollout can act on it.

cmd/launcher/launcher.go
```diff
@@ -63,4 +64,10 @@ type Config struct {
 	A2AOptions       []a2asrv.RequestHandlerOption
 	PluginConfig     runner.PluginConfig
 	TelemetryOptions []telemetry.Option
+
+	// EventsCompactionConfig enables context compaction for the sessions the
```
karolpiotrowicz: [major] `EventsCompactionConfig` is scoped to the server, but one server serves many apps through an `agent.Loader`, so the feature cannot be enabled for one app without applying to all of them. Two measured consequences follow from that single root cause: a *valid* config combined with a workflow-agent app returns `500 failed to create runner: EventsCompactionConfig needs a Summarizer: root agent "wfapp" is not an LLM agent` for that app while its LLM sibling serves fine — enabling a feature for app A breaks app B, at request time rather than at startup; and one `Summarizer` instance (one model, one credential, one quota) receives the conversation content of every tenant. Multi-app is a first-class deployment shape here: `agent.NewMultiLoader` is exported and the REST route is `/apps/{app}/…`. adk-python scopes this per app — `src/google/adk/apps/app.py:84` carries `events_compaction_config` on the `App` — so mirroring that with the server-level value as a fallback, and skipping rather than failing when a root agent cannot supply a default summarizer, would make incremental adoption possible.

server/adkrest/controllers/triggers/options_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"google.golang.org/adk/v2/runner"
+	"google.golang.org/adk/v2/session/compaction"
+)
+
+// TestControllerOptionsAreBackwardCompatible pins that the trigger controller
+// constructors still accept their original argument list. The compaction option
+// was added variadically precisely so existing callers keep compiling; a switch
+// to a required parameter would break this file.
+func TestControllerOptionsAreBackwardCompatible(t *testing.T) {
```
karolpiotrowicz: [major] The new tests assert that the field is *set*, not that it has any *effect*, and most of the plumbing this PR exists to add is consequently unprotected: deleting the `EventsCompactionConfig:` assignment at 8 of the 10 propagation sites leaves the entire repository test suite green. The survivors are `cmd/launcher/console/console.go:112`, `cmd/launcher/web/a2a/a2a.go:130`, `cmd/launcher/web/api/api.go:85`, `cmd/launcher/web/triggers/pubsub/pubsub.go:122`, `cmd/launcher/web/triggers/eventarc/eventarc.go:122`, `server/adkrest/controllers/triggers/triggers.go:97`, `server/agentengine/controllers/method/stream_query.go:200` and `.../streaming_agent_run_with_events.go:239`; only `server/adkrest/controllers/runtime.go:244` is caught by a real assertion. The two REST tests also survive targeted mutation: truncating the prompt outright after `internal/llminternal/contents_processor.go:122` leaves `TestRESTCompaction_DisabledByDefault` passing, because it pins "0 compaction events" rather than "behaviour is unchanged"; and making `isCovered` (`internal/compactioninternal/apply.go:129`) return false for `Author == "model"` leaves `TestRESTCompaction_EnabledViaServerConfig` passing, because it asserts `q1`/`q2` are gone but never checks `answer 1`/`answer 2`. Suggest asserting effect instead of existence — compare the full prompt against a no-compaction baseline in the disabled test, add the model turns to the "gone" list in the enabled test, and add one behavioural test per surface. The trigger surface being inert is precisely the class of defect this gap allowed through.

examples/compaction/main.go
```diff
@@ -0,0 +1,103 @@
...
+//
+// Compaction keeps an agent's prompt small as its conversation grows: older
+// turns are summarized into a single event, and later prompts carry that
+// summary instead of the raw turns. Two triggers are available, and this
+// example arms both:
+//
+//   - Sliding window fires after every CompactionInterval completed turns.
+//   - Tail retention fires mid-turn, once a prompt reaches TokenThreshold.
+//
+// Setting EventsCompactionConfig on [launcher.Config] enables compaction for
```
karolpiotrowicz: [major] Both halves of this sentence are inaccurate, and the command the doc prints at `:33` does not run. `full.NewLauncher()` (`cmd/launcher/full/full.go:32`) is `universal(console, web(webui, a2a, pubsub, eventarc, api))`: it *does* serve pubsub, eventarc and api — the surfaces where compaction currently has no effect — and it does *not* serve Agent Engine at all, so the doc names a surface that is absent while omitting two where the feature is inert. The documented invocation exits immediately:

```
$ compaction-example web
Run failed: no active sublaunchers found - please specify them in the command line.
  Possible values: [webui a2a pubsub eventarc api]
```

`web webui` is the working form. The example type-checks, which is why a runtime-only argument error survived; a smoke test that actually executes the binary with a dry-run or `--help` would keep the documented command honest.

server/adkrest/controllers/runtime.go
karolpiotrowicz: [minor] The option loop calls `opt(c)` with no nil guard, so a nil entry panics at construction; `server/adkrest/controllers/triggers/pubsub.go:48` and `.../eventarc.go:49` have the same shape, and all three were confirmed to panic with a bare `invalid memory address or nil pointer dereference`. Passing nil is largely a caller error, but a dynamically assembled `[]RuntimeAPIOption` with one conditionally-unset entry is ordinary code, and the resulting server-startup panic names nothing that would help diagnose it. `for _, opt := range opts { if opt != nil { opt(c) } }` in all three constructors, plus a test, turns it into a no-op.

server/adkrest/controllers/runtime.go
```diff
@@ -42,11 +43,33 @@ type RuntimeAPIController struct {
...
+
+	eventsCompactionConfig *compaction.Config
+}
+
+// RuntimeAPIOption configures optional [RuntimeAPIController] behaviour.
+//
+// The constructor takes its required dependencies positionally; anything
+// optional is supplied here instead, so new capabilities do not keep widening
+// an already long signature or break existing callers.
+type RuntimeAPIOption func(*RuntimeAPIController)
```
karolpiotrowicz: [minor] The option pattern is inconsistent across the parent/child package pair: two parallel option types carry two identically-named `WithEventsCompactionConfig` functions, and one of them — `ControllerOption` (`server/adkrest/controllers/triggers/triggers.go:56-61`) — is typed over `*RetriableRunner` rather than a controller, which mis-names it and pins the runner as an exported extension point whose fields third parties cannot reach. There is also a latent ordering trap: `pubsub.go:53` and `eventarc.go:54` size the semaphore from the positional `triggerConfig` *before* the option loop runs, so an option mutating `MaxConcurrentRuns` is silently ignored — setting it to 8 via an option still leaves `cap(semaphore) == 1`. No shipped option can trigger this today, since `WithEventsCompactionConfig` is the only `ControllerOption` in the package and nothing sets `MaxConcurrentRuns`, so it is latent rather than a live defect. Applying options before deriving state from the config, with a comment recording the requirement, would close it for the next option added.

karolpiotrowicz (commented):
Getting compaction to every surface is the right goal and the launcher/REST
plumbing is mostly mechanical, but three things need resolving, and one of them
is a compatibility break the PR description says does not exist.

**1. The three constructors are a breaking change.** `NewRuntimeAPIController`,
`NewEventarcController` and `NewPubSubController` gained a trailing variadic, and
`apidiff` reports all three as Incompatible. A trailing variadic is
source-compatible for ordinary call sites, which is why the in-repo tests still
build — but not for code that takes the function *value*, passes it as a
func-typed argument, or assigns it to a func type. A probe of that shape compiles
against the base and fails against this PR. The description's "Every existing
caller compiles unchanged, which tests pin" does not hold: the cited tests are
plain call sites and pin nothing. An options struct, or a new `…WithOptions`
constructor alongside the existing one, avoids it.

**2. Compaction cannot work on Agent Engine.** `session/vertexai` persists
`Actions.Compaction` nowhere — `eventNeedsRawEvent` (`vertexai_client.go:310`)
omits it and `createAiplatformpbEventActions` (`:433`) returns nil — so every
summary is discarded after being paid for, and the next turn re-summarizes the
same range. Running the session-service conformance suite across the three
backends shows the split cleanly: inmemory and database pass, vertexai comes back
with `Compaction: nil`. Since this PR is what makes the surface supported, it is
worth either fixing the client or not advertising the surface yet.

**3. The REST create path accepts a client-supplied `actions.compaction`.**
`sessions.go:75-80` copies it straight from the request body. Plain
`author:"model"` is already defended by `ConvertForeignEvent`, but the compaction
path is exempted from that sanitizer and force-sets `Author="model"` — so this is
a bypass of an existing defense rather than a new instance of a known one. A
far-future nested `EndTimestamp` also pins the injected text as the last turn of
every subsequent prompt. Dropping `Actions.Compaction` on ingress is the
straightforward fix.

A few more worth a look: the compaction config is process-global, so on a
multi-tenant server it applies to apps that never asked for it and routes every
tenant's content through one summarizer; an invalid config is accepted at startup
and then 500s every request; a summarizer failure turns a successful turn into a
500 that discards the agent's answer and includes provider/project detail; and
the Pub/Sub and Eventarc paths create a new session per delivery, so compaction
can never fire at an interval of 2 or more. Finally, `examples/compaction`'s
documented `web` command does not run, and its package doc names Agent Engine,
which `full.NewLauncher` does not serve.

---

### Additional findings

**[blocker] `session/vertexai` drops every compaction record, making compaction a paid no-op on Agent Engine.** In `session/vertexai/vertexai_client.go`, `eventNeedsRawEvent` (`:310`) omits `Actions.Compaction` from the "needs a raw-event blob" predicate; `createAiplatformpbEventActions` (`:433`) returns `nil, nil` when `StateDelta` and `ArtifactDelta` are both empty, which is exactly the shape of a compaction event; and `aiplatformToSessionEventActions` (`:458`) restores only those two fields. Adding a `Compaction` field to the existing `with_all_fields` fixture in `session/sessiontestsuite/service_suite.go` — which had zero compaction coverage, so no backend was ever forced to prove the round-trip — makes the in-memory and database services pass and vertexai fail with `Compaction: nil`, and the recorded-RPC fixture still matches, so the field never reaches the wire at all rather than being mis-read. The consequence is not only lost auditability: over 8 turns at `CompactionInterval: 2`, a session service that drops the record produces 7 summarizer calls instead of 4, each re-summarizing the whole history (windows grow 4→22), and yields a *larger* final prompt than the faithful backend. The canonical Agent Engine deployment (`examples/agentengine/main.go:140`) uses `vertexai.NewSessionService`, so the surface this PR advertises is the one where the feature costs money and does nothing, silently. The mapping gap predates this PR — it arrived with the earlier compaction change in this stack — but this is the PR that makes Agent Engine a supported compaction surface and asserts it works (also raised on #1231, which introduces the record; the fix belongs there, but it blocks the surface this PR advertises). adk-python persists the record deliberately (`src/google/adk/sessions/vertex_ai_session_service.py:47`, `_COMPACTION_CUSTOM_METADATA_KEY = '_compaction'`; write path `:432-441`, read path `:576-588`). Suggest mirroring that in `custom_metadata`, or at minimum adding `Actions.Compaction != nil` to `eventNeedsRawEvent`, and landing the conformance-suite fixture so every backend has to prove the round-trip.

**[blocker] A client-supplied `actions.compaction` in a REST create body erases the conversation and injects attacker text as a model turn.** `createSession` (`server/adkrest/controllers/sessions.go:75-80`) appends client-supplied events verbatim, and `models.ToSessionEvent` (`server/adkrest/internal/models/event.go:35` and `:101`) copies `Actions.Compaction` straight off the wire with no provenance check, so one create body carrying a compaction record spanning `0001-01-01` to `9999-12-31` suppresses every genuine summary. Driven through the real handler over HTTP, the clean session compacts to 2 summaries plus recent turns while the poisoned one retains all 6 raw turns, records 4 compaction events instead of 3 — the server is billed *more*, for summaries it then discards — and places the attacker's content as the last model message before the continuation prompt, the highest-leverage position in the prompt for an injection. The primitive predates this PR and does not depend on compaction being enabled, since `compactioninternal.Apply` runs unconditionally at `internal/llminternal/contents_processor.go:122` (tracked on #1232); what this PR adds is the REST surface that accepts the field, which is what makes the ungated apply remotely reachable, plus an operator-facing switch that the injection specifically neutralizes while keeping the bill. `Actions.Compaction` is server-produced bookkeeping and should be stripped in `models.ToSessionEvent` — or the request rejected outright when it is present — with a regression test asserting that a create body carrying `actions.compaction` is refused or sanitized.

**[major] A summarizer failure converts a successful turn into a 500 and discards the agent's answer** (`server/adkrest/controllers/runtime.go:112-118`; `runner/runner.go:463`). Post-invocation compaction runs after the agent has already answered, and its error propagates: turn 2 returns `500 failed to run agent: post-invocation context compaction failed: sliding-window summarization failed: …` while 4 events including the answer are persisted in the session. The client never receives the reply and retries, re-billing an agent turn against a session that already contains the answer — and a summarizer 429 is the most likely trigger, since the summarizer shares quota with the agent. The response body also echoes the upstream error unredacted, disclosing the project number, project ID, region, model name and endpoint to the caller. Surfacing the failure is correct, but a log line or metric satisfies that without destroying completed work: return the collected events with 200, record the compaction failure, and let the next turn retry. Separately, map upstream errors to a sanitized client-facing message and keep the detail server-side.

**[major] A summarizer 429 amplifies one trigger delivery into N full agent re-runs** (`server/adkrest/controllers/triggers/triggers.go:151`, `isResourceExhausted`, with the retry loop at `:90-97`). The check matches `"429"` anywhere in the error string, and the compaction error wraps the upstream text verbatim, so a summarizer throttle raised *after* the agent turn already succeeded is misread as the agent being throttled. One Pub/Sub delivery with `MaxRetries: 3` produced 4 agent model calls and 4 summarizer attempts; the agent turn succeeded every time and only the bookkeeping failed. Retrying re-executes the agent against the very quota that is exhausted, so it makes the condition worse, and the eventual 500 then invites Pub/Sub redelivery for a further multiplier — all under exactly the quota pressure compaction is meant to relieve. Suggest classifying by origin rather than substring: a typed or `status.Code` check, and no retry when the failure came from post-invocation compaction, which pairs naturally with treating that failure as non-fatal.

**[minor] `/run_live` accepts the compaction config and never compacts, and `/run_sse` is entirely unexercised** (`server/adkrest/controllers/runtime.go:313`, `runner/runner.go:555-558`). `getRunner` feeds `EventsCompactionConfig` to all three handlers, but `Runner.RunLive` deliberately bypasses `compactionctx`, so the runtime behaviour is intended; the gap is that neither the `adkrest.ServerConfig` nor the `launcher.Config` doc comment records the exception, while `runner.Config`'s comment does. A user enabling compaction on a bidi/live deployment therefore gets nothing and has no way to learn that from the configuration they are setting. `RunSSEHandler` and `RunLiveHandler` are also at 0.0% coverage, so the streaming assembly path — where a summary meets partial or streamed events — is untested. Suggest noting the live-run exception in both server-level doc comments and adding an SSE test asserting compaction behaviour on the streaming path.

**[minor] Compaction has no prose documentation anywhere in the repository.** A case-insensitive search for `compact` across every markdown file in the worktree returns nothing — not `README.md`, not `README-v2.md`, not `examples/README.md`. A user-facing feature spanning six PRs ships with godoc comments and one example and no prose entry point, so nothing tells a user that the feature exists, when to use it, or what it costs; given the per-surface caveats above, cost is exactly what a user needs to know. A per-example README is not the convention here — `examples/quickstart` has none either — so the gap is the top-level documentation: a short section in `README.md` / `README-v2.md` linking to the example and stating the per-surface limitations would cover it.

**[minor] Compaction emits consecutive model-role summaries at the head of the prompt, breaking strict role alternation** (`internal/compactioninternal/apply.go`, surfaced by the example's `CompactionInterval: 2`). A clean, unattacked session's prompt opens with `[model] SUMMARY` twice in a row before the first `[user]` turn, and the count grows by one per compacted window — a longer run showed three consecutive summaries. Gemini tolerates consecutive same-role turns and nothing failed because of it, so this is a portability risk for stricter model APIs rather than a live break, and the substitution mechanism itself predates this PR; what this PR contributes is the example configuration that makes the shape the default a user sees. Merging adjacent summaries into a single summary event during substitution would keep the prompt head to one `model` turn regardless of how many windows have been compacted.

**[minor] The example teaches aggressive, uncosted tuning values** (`examples/compaction/main.go`, around line 94). `CompactionInterval: 2`, a 32,000-token threshold and a 10-event retention size appear as bare literals with no rationale and no reference to library defaults. Small values are a legitimate way to make the effect visible in a demo, but the example is silent about the trade-off while being the only documentation that exists, so the numbers read as recommendations rather than demo settings: `CompactionInterval: 2` works out to roughly one extra model call per two user turns (3 summarizer calls over 6 turns, 4 over 8 in the runs above). A comment on each constant giving its cost implication, plus a line stating that these are demo values and not recommended defaults, would be enough.

baptmont:
Triaged all eight. All eight stand.

On the blocker: you are right, and it is worse than the comment claims. I verified it by compilation rather than inspection, assigning all three constructors to their pre-PR function types in a downstream file. That file builds against the base commit and fails with three errors against this one: `NewRuntimeAPIController` goes 7 to 8 parameters, `NewPubSubController` and `NewEventarcController` 6 to 7, all three now variadic.

The detail that settles it: `v2.2.0` was tagged shortly before this commit, so these are in a shipped API, on a module with no pre-release marker and no `apidiff` gate in CI. And because `server/adkrest/internal/routers` is internal, calling these constructors directly and mounting the controller yourself is the only supported third-party path, which is exactly what `examples/bidi/main.go` does in-repo. The PR's own "backward compatible" tests are plain call expressions, the one form a trailing variadic can never catch.

I plan to keep each signature byte for byte and add an additive `...WithOptions` sibling, roughly eight lines each, switching the in-repo call sites to those. Tell me if you would rather accept the break and document it in `README-v2.md` instead, since that is a judgement about your release policy rather than about the code.

I also reproduced `3701657680`: the documented `web` command exits with "no active sublaunchers found", and `full.NewLauncher()` does not include Agent Engine while the doc says it does. Both halves wrong, as you said.

cmd/launcher/console/console.go
baptmont: Fixed. `adkrest.NewServer` and `agentengine.NewHandler` now call `Config.Validate()` and refuse to construct on an invalid compaction config, naming the field.

You had the failure mode exactly right: the config is validated inside `runner.New`, which both surfaces call per request, so without a check at construction an invalid config produced a server that started cleanly and then 500'd on every request. The operator saw a broken deployment rather than a refused start.

Regression test: `TestNewServerRejectsInvalidCompactionConfig`, using `OverlapSize` without an interval. Without the check, `NewServer` accepts it.

server/adkrest/controllers/runtime.go
baptmont: Fixed in all three constructors. A nil option is skipped rather than dereferenced, since options are commonly built by a helper that returns nil when it has nothing to apply, and a variadic parameter makes passing one easy. Panicking during construction is a poor way to report that.

Regression test: `TestControllerOptionsToleratesNil`, which also checks that a nil option does not stop a later real option from applying. Without the guard it panics.

Note this code may move depending on how the signature question in your other comment is resolved. The guard travels with the loop either way.

examples/compaction/main.go
baptmont: Both halves confirmed and fixed. I reproduced the command rather than reading the code for it:

```
$ go run ./examples/compaction web
Run failed: no active sublaunchers found - please specify them in the command line.
Possible values: [webui a2a pubsub eventarc api]
```

The doc now shows `web webui` and says explicitly that the bare form exits that way.

The surface list was wrong in both directions, as you said. `full.NewLauncher()` is console plus web with webui, a2a, pubsub, eventarc and api: Agent Engine is not among them, and the three that were unmentioned are. The doc now names what the launcher actually serves, and notes separately that Agent Engine reads the same config field but is served by its own handler rather than by this launcher.

server/adkrest/controllers/runtime.go
baptmont: Fixed, and you were right about the severity.

Each signature is restored byte for byte, and a `...WithOptions` sibling carries the options. In-repo call sites use the sibling.

Two things I checked rather than assumed. `v2.2.0` was tagged shortly before the commit that changed these, so they are in a released API, not an unreleased one. And the test that claimed to pin backward compatibility could never have caught this: it was an ordinary call expression, which is exactly the form a trailing variadic does not break. It now assigns each constructor to an explicit function type, and re-adding the variadic fails it.

Separately, nothing in CI would have caught this either, so I opened #1297 to add an `apidiff` job. It reports these three constructors against the commit that introduced them and passes against the fix. Independent of this stack, so it can land on its own timeline.

baptmont:

## Landed

| Decision | Where | Note |
|---|---|---|

## Four things worth flagging

## Still open

server/adkrest/controllers/triggers/triggers.go
baptmont: Not done.

Still present. `triggers/triggers.go:68-75` creates a fresh session inside every `RunAgent`. No guard rejects `CompactionInterval > 0`, no durable session, no doc note. Only compaction mentions in the package are the plumbing itself (`triggers.go:47,58-63,97`).

You are right and it is still open. Recorded in the audit of this review rather than quietly dropped; leaving the thread unresolved.

cmd/launcher/launcher.go
baptmont: Not done.

Still present. The field lives only on the server/launcher config (`cmd/launcher/launcher.go:72`, `server/adkrest/handler.go:85`). No per-app carrier exists. `runner/runner.go:162-165` still returns an error, not a skip, when a root agent is not an LLM agent.

You are right and it is still open. Recorded in the audit of this review rather than quietly dropped; leaving the thread unresolved.

server/adkrest/controllers/triggers/options_test.go
baptmont: Not done.

Still present. The disabled test only counts compaction events (`compaction_integration_test.go:118-124`), with no baseline prompt comparison. The enabled test's "gone" list is still just `q1`/`q2` (`compaction_integration_test.go:103`); the model turns are not checked. No behavioural test for a2a, api, triggers or Agent Engine — only `adkrest` has one.

You are right and it is still open. Recorded in the audit of this review rather than quietly dropped; leaving the thread unresolved.

server/adkrest/controllers/runtime.go
baptmont: Not done.

Still present. `ControllerOption` is still `func(*RetriableRunner)` (`triggers/triggers.go:55`), and two same-named `WithEventsCompactionConfig` still exist (`runtime.go:60` and `triggers.go:61`). The semaphore is still sized from the positional parameter at `pubsub.go:68` and `eventarc.go:69`, so an option that changes `retriable.triggerConfig.MaxConcurrentRuns` is still ignored. No comment records the requirement.

You are right and it is still open. Recorded in the audit of this review rather than quietly dropped; leaving the thread unresolved.

baptmont:

## Thread state

| PR | Resolved | Left open |
|---|---:|---:|
| #1231 | 16 | 4 |
| #1232 | 13 | 6 |
| #1233 | 8 | 7 |
| #1234 | 8 | 13 |
| #1235 | 3 | 5 |
| #1236 | 4 | 3 |

## State of the code

cmd/launcher/console/console.go
baptmont: Now done for the launcher too. `launcher.Config` gains a `Validate`, called from each `Execute` entry point, so `web a2a`, `web pubsub` and `web eventarc` fail at startup naming the field rather than starting healthy and failing every request.

Table-driven test covers the nil case, a valid config and the three rejections.

server/adkrest/controllers/runtime.go
baptmont: Accepted as is for the typing, and worth noting what did change.

The two option types remain distinct: `ControllerOption` applies to the retriable runner and `RuntimeAPIOption` to the runtime controller, and merging them would mean one type that can only be applied to one of two unrelated structs. Both now skip nil options rather than panicking, and both constructors keep their released signatures with `WithOptions` siblings, which was the part that actually broke callers.

The semaphore is still sized from the constructor parameter rather than from anything an option could change, so the ordering trap you describe is real but currently unreachable: no option touches `MaxConcurrentRuns`. Worth revisiting if one ever does.

server/adkrest/controllers/triggers/options_test.go
baptmont: Partly addressed, and I want to be accurate about how much.

There is now a behavioural test on the trigger surface: it asserts the warning fires for a config that cannot work there and stays quiet for one that can. And the launcher has a table-driven test that a bad config stops startup.

What is still true is that the per-surface tests mostly assert the field is plumbed rather than that compaction changes a prompt on that surface. The strongest coverage of the effect lives in the runner and end-to-end tests instead. Leaving this open rather than claiming otherwise.

cmd/launcher/launcher.go
baptmont: Documented rather than changed.

The config is server-wide, and one server can serve many applications through its agent loader, so they all get the same setting and the same `Summarizer` instance, which means the same model. The field now says that, and says that applications needing different compaction, or which must not share a summarizer, should run on separate servers.

Scoping it per application means threading it through the loader, which is a bigger change than this stack should carry. Happy to do it separately if you would rather have it.

server/adkrest/controllers/triggers/triggers.go
baptmont: Fixed as a warning, since the underlying model is not something to change here.

You are right that a sliding window can never fire on these surfaces: each delivery runs in its own session, so history never accumulates and the interval is never reached. Configuring one is now logged at construction, saying which strategy does work instead. Tail retention is unaffected, because it measures the prompt inside a single run.

The option's documentation explains the same thing, so the failure mode is visible before anyone runs it.

karolpiotrowicz (commented):
Reviewed at `bc66a9e`. The doc-only commits pushed since then do not change any line cited below.

The `…WithOptions` approach holds up under the check that matters: I compiled a downstream module against this revision and its parent using all three constructors as values — variable of func type, func-typed argument, struct field, map value, and `reflect` — and it compiles on both, identically. Two smaller source breaks remain that an API-compatibility check does not model: a type embedding `launcher.Config` alongside anything else with a `Validate()` method now fails with `ambiguous selector`, and unkeyed composite literals of `ServerConfig` and `launcher.Config` break (though `go vet` already flagged those on the parent, so it is the usual accepted cost). Both are worth a line in the description.

What blocks is that the fail-fast check does not achieve what its own comment says.

**1. Startup validation is shape-only, so an invalid setup still 500s every request.** [`NewServer`](https://github.com/google/adk-go/blob/bc66a9ee95f34f53de0d64df47b14d310693f8ed/server/adkrest/handler.go#L38-L45) says it validates here so a bad config cannot "start cleanly and then fail every request with a 500 that names nothing the operator can act on" — and that is what happens. `&compaction.Config{CompactionInterval: 2}` with no `Summarizer` passes `Validate()`, the server starts, and a request to a non-LLM root agent returns `500 failed to create runner: EventsCompactionConfig needs a Summarizer: root agent "wf_app" is not an LLM agent`.

This is not a multi-tenancy problem, which matters for the fix: I reproduced the identical 500 with a single-app `NewSingleLoader`, where there is one root agent, known at construction time. Resolving the config against the loader's agents in `NewServer` covers that case and the multi-app one; per-app scoping is a separate, larger design question.

**2. The trigger constructors never validate, and on Pub/Sub that is a redelivery loop.** [`NewPubSubControllerWithOptions`](https://github.com/google/adk-go/blob/bc66a9ee95f34f53de0d64df47b14d310693f8ed/server/adkrest/controllers/triggers/pubsub.go#L47-L49) and its Eventarc twin return only a controller, so they have no way to refuse a bad config. `&compaction.Config{}` constructs fine and then every delivery returns `500 … invalid EventsCompactionConfig: no compaction strategy is enabled`, which Pub/Sub push reads as a NACK. These constructors are new in this PR, so giving them an `error` return costs nothing in compatibility.

**3. The new warning is wrong in two directions.** The [gate](https://github.com/google/adk-go/blob/bc66a9ee95f34f53de0d64df47b14d310693f8ed/server/adkrest/controllers/triggers/triggers.go#L71-L80) prints "it will never fire" at `CompactionInterval: 1`, where I measured six deliveries producing six compaction events and six summarizer calls — one extra model call per message on a single-turn session that is then discarded. It is also false across retries: the session is created outside `runAgentWithRetry`, so one throttled delivery accumulates invocations in one session and the window does fire. And because it requires `TokenThreshold == 0`, the configuration in [`examples/compaction/main.go`](https://github.com/google/adk-go/blob/bc66a9ee95f34f53de0d64df47b14d310693f8ed/examples/compaction/main.go) gets no warning at all.

**4. A summarizer failure discards an answer the agent already produced.** On the trigger path the runner is built with [the compaction config](https://github.com/google/adk-go/blob/bc66a9ee95f34f53de0d64df47b14d310693f8ed/server/adkrest/controllers/triggers/triggers.go#L106-L116) and the caller does not distinguish `compaction.ErrCompaction`, so a failed summary turns a delivery in which the agent answered into an HTTP 500 with the events thrown away. [the package doc](https://github.com/google/adk-go/blob/bc66a9ee95f34f53de0d64df47b14d310693f8ed/session/compaction/compaction.go#L57-L63) states the opposite intent — a compaction failure "costs a smaller prompt later, not the user's answer".

The reason none of these were caught: the new tests assert the field was *stored*, not that it has an effect. Deleting the line that hands the config to the runner on the trigger path leaves the whole suite green, and the same is true of the console, A2A, api, Agent Engine and launcher-validation sites. [`NewRuntimeAPIController`](https://github.com/google/adk-go/blob/bc66a9ee95f34f53de0d64df47b14d310693f8ed/server/adkrest/controllers/runtime.go#L66-L74) also has no signature pin, so the change its doc comment warns against can be reintroduced invisibly — the sibling package's [`options_test.go`](https://github.com/google/adk-go/blob/bc66a9ee95f34f53de0d64df47b14d310693f8ed/server/adkrest/controllers/triggers/options_test.go#L56-L65) shows the one-line type-alias idiom that catches it at compile time.

Smaller: two of the three new `Validate()` call sites are unreachable (`console.NewLauncher` and `web.NewLauncher` return `SubLauncher`, which has no `Execute`); `agentengine.NewHandler` re-implements `launcher.Config.Validate()` inline rather than calling it, so any future check added there silently skips Agent Engine; the multi-tenant caveat is documented on `ServerConfig` but not on `launcher.Config`, which is the field operators actually set and which reaches more surfaces; `/run_live` accepts the config and cannot act on it, with no warning; and the example's documented command (`web webui`) serves the UI with no REST backend — `/api/list-apps` is a 404 until you add `api`.

baptmont:
Fixes are in #1316. Thanks for compiling a downstream module against both revisions to check the `…WithOptions` approach; that is the check that mattered and I would not have thought to do it that thoroughly.

**Finding 1.** Fixed. Validation now resolves the config against the loader's agents rather than checking its shape, so the case you hit, no `Summarizer` over a non-LLM root agent, is refused at construction. It is a dry run of `runner.New` per app rather than a second copy of its checks, so it cannot drift from the path a request takes.

One wrinkle worth knowing: it runs twice, once without the compaction config and once with, and reports only a failure the compaction config introduced. Without that it refused perfectly good setups whose session service was not wired yet at construction time.

**Finding 2.** Fixed. Both `WithOptions` constructors return an error now. You are right that it costs nothing, they are new in this stack, and the two older constructors keep their signatures.

**Finding 3.** Fixed, and it was wrong in all three ways you describe. At interval 1 it fires on every delivery, so that now warns for the opposite reason: the summary goes into a session that is discarded when the delivery ends. At a larger interval it warns that a delivery handled in one attempt will not reach the interval, which is the accurate version of "never". And it no longer goes quiet when `TokenThreshold` is set, since the sliding window is just as inert next to tail retention.

While in there I fixed the `Each retry = new session` comment on `RunAgent`, which says the opposite of what the code does, and is what makes your point about retries sharing a session true.

**Finding 4.** Fixed with the rest of the `ErrCompaction` classification on #1232.

**On the vacuity.** Fair, and the same was true of most of what this review found. The new tests assert effect: the startup one builds a server that cannot serve its only app and expects a refusal, and the trigger one expects the delivery to succeed while the summarizer fails.

The `NewRuntimeAPIController` signature pin is not in yet.

**The smaller items.** Not fixed: the two unreachable `Validate()` call sites, `agentengine.NewHandler` re-implementing `launcher.Config.Validate()` inline, the multi-tenant caveat missing from `launcher.Config`, `/run_live` accepting a config it cannot act on, and the example's documented command. None looked worth holding the stack for, but say if you disagree on any of them, particularly the Agent Engine one, since a future check there really would skip silently.

karolpiotrowicz (commented):
**Two of the five surfaces skip the startup check that this PR is built around, so a bad config starts healthy and then fails every request.** The design here is to reject a config the server cannot actually serve at construction time, by building a throwaway runner per application. REST does that at [`validateCompactionAgainstAgents`](https://github.com/google/adk-go/blob/6f4ee681dcb22cb0ab35bbba8af68a76cc928e5c/server/adkrest/handler.go#L44), and both trigger controllers do the same.

Agent Engine and A2A do not. [`server/agentengine/handler.go:47`](https://github.com/google/adk-go/blob/6f4ee681dcb22cb0ab35bbba8af68a76cc928e5c/server/agentengine/handler.go#L47) calls `config.Validate()`, which only checks that the config's fields are internally consistent — it never involves the agent. [`cmd/launcher/web/a2a/a2a.go:122`](https://github.com/google/adk-go/blob/6f4ee681dcb22cb0ab35bbba8af68a76cc928e5c/cmd/launcher/web/a2a/a2a.go#L122) checks nothing at all before handing the config to `adka2a.NewExecutor`.

So both accept the exact case your own comment in `adkrest/handler.go` calls the one operators actually hit: no `Summarizer` supplied, over a root agent that has no model to default to. The process then starts cleanly, reports healthy, and returns 500 on 100% of traffic. Running the same config and the same loader through all of them side by side, adkrest and the triggers refuse it and `agentengine.NewHandler` returns a nil error.

**Most of the wiring this PR adds has no test that would notice if it were removed.** Taking each line that puts `EventsCompactionConfig` into a `runner.Config` and deleting it, then running `go test ./cmd/... ./server/...`, leaves the suite green for 8 of the 10 surfaces. Only the REST pair is caught, and both by the same test. Four of the six new `errors.Is(ErrCompaction)` tolerance branches are unexercised too, including both REST ones.

The trigger and Agent Engine cases are worth calling out specifically. On those, removing the wiring leaves their compaction tests green, while separately disabling the tolerance branch makes them fail. A test with that pair of behaviours cannot distinguish "compaction ran and its failure was tolerated" from "compaction never ran at all", which is the property it is supposed to be establishing.

Two smaller ones:

- The console launcher is the only surface in this PR without an `ErrCompaction` branch, so a compaction failure prints `AGENT_ERROR` directly underneath the answer the user has already received.
- [`server/adkrest/controllers/runtime.go:59-60`](https://github.com/google/adk-go/blob/6f4ee681dcb22cb0ab35bbba8af68a76cc928e5c/server/adkrest/controllers/runtime.go#L59-L60) carries the "prompts stay small as a conversation grows" wording you corrected elsewhere after the last round. This is the second surviving copy — the other is on #1232, where I have flagged it.

baptmont:
Both fixed.

**Agent Engine and A2A now run the same dry run as the others.** Rather than a third and fourth copy of the loop, it moved to `internal/compactionvalidate.AgainstAgents` and all four surfaces call it, so there is one answer to "can this config serve these apps" instead of four that can drift. Test drives the exact case your comment names, no `Summarizer` over a workflow root agent, and asserts the refusal names the app.

**The trigger surface now has a test that fails when the wiring is deleted.** You were precise about why the old one could not: it asserted a compaction failure was tolerated, which a surface that never runs compaction satisfies equally well. The new one counts summarizer calls, so it distinguishes "compaction ran and its failure was tolerated" from "compaction never ran". I verified it by deleting the `EventsCompactionConfig` assignment in `triggers.go` and watching it fail.

I did that for the trigger surface, not for all eight you listed. The same shape applies to the rest and I would rather add them deliberately than generate six near-copies at the end of a long run.

The console launcher's missing `ErrCompaction` branch and the stale doc line on `runtime.go` are still open.

Separately, on your point that no CI had ever run on this stack: fixed. `.github/workflows/go.yml` only triggered for PRs targeting `main` and `v1`, so a stack targeting its own branches was never built, tested or linted. It now includes `baptmont/**`, and the jobs are running on these PRs for the first time.

karolpiotrowicz (approved):
The extraction into `internal/compactionvalidate` holds up, and it is stronger than what it replaced rather than merely equivalent.

I drove the failing configuration — compaction enabled, no `Summarizer`, a non-LLM root agent — through the real public constructor of every serving surface that can build a runner with compaction: adkrest, the Pub/Sub and Eventarc trigger controllers, Agent Engine, the A2A launcher, and the console and web launchers. All of them now refuse it at construction. Last round Agent Engine and A2A accepted it, started healthy, and returned 500 on every request, so that hole is closed.

The shared `againstAgents` also picks up a guard the REST original did not have: a nil `SessionService` used to make the old inline check `continue` past the very case it existed to catch, which turned it into a no-op that accepted anything. Root-only versus every-app is correctly matched to what each surface actually serves.

Two smaller things, neither blocking. Five of the ten config-wiring lines are now pinned by a test that fails when you delete them, up from two — the five survivors are all under `cmd/launcher/**`, and a single console test would close both a wiring gap and a tolerance-branch gap at once. And `server/adkrest/controllers/runtime.go:59-60` still carried the "prompts stay small as a conversation grows" wording I flagged, which you have since replaced at all five sites.

baptmont:
The REST constructors now take config structs instead of functional options, plus the rebase.

**`RuntimeAPIOption`, `ControllerOption` and both `WithEventsCompactionConfig` helpers are gone**, replaced by `RuntimeAPIControllerConfig` and a shared `ControllerConfig`. Options solved only the compaction field and left the six or seven positional parameters in place, so every later addition would have needed a third constructor. A struct absorbs both, and the repo already leans this way — 30 config-taking constructors against 9 option types.

Removing them costs nothing: those `...WithOptions` constructors were added in this stack and have never been released.

The three original positional constructors are untouched and now carry `// Deprecated:` markers pointing at the new ones.

**This also closes the finding you left open on the last round** — that nothing pinned the constructor signatures against a re-break. That is now structural rather than test-enforced: a config struct cannot be widened in a breaking way, so there is no signature left to accidentally change.

The trigger warnings about `CompactionInterval` on a per-delivery session moved intact into `newRetriableRunner`. `options_test.go` became `config_test.go`, and the nil-option tolerance test is deleted along with the mechanism it tested.

**Shared context for this round.** The whole stack was rebased onto current `main` (`3f69c505`), absorbing 55 commits. Four conflicts, all from #1252 "give Event a consistent JSON encoding":

- `session/session.go` — main's new JSON tags, `MarshalJSON` and `nilOrRef` merged with our `Compaction` field and the `EventCompaction`/`EventRef` types. One hunk ended mid-function and truncated `nilOrRef`; `gofmt` caught it.
- `internal/telemetry/logger.go` — the only semantic one. Main changed `variantToGenAISystem` to return `*attribute.KeyValue`; ours returned `*log.KeyValue` via a `GenAISystemAttr` helper we had extracted. Resolved by keeping **main's signature** and implementing it through our helper, so there is one source of truth rather than two.
- Imports only in `session/database/service_test.go` and `session/vertexai/service_test.go`.

Every slice was rebuilt from the rebased tree and verified to build, lint and test on its own. The full suite is clean.

karolpiotrowicz (approved):
Re-approving at 506124c2. The prior approval dated from 24 August and pointed at f47f4d6c, two rounds of change back.

Since the revision I validated at d6cefddb, five files moved: the config field rename from EventsCompactionConfig to Compaction reaching the web API wiring, and four test files. The rename is complete and consistent, with the old name gone from production code entirely and the new one declared at all five sites. Nothing released carried the old name.

baptmont:
Rebased again onto current `main` (`9db1ff12`), which had moved another 15 commits. Content is unchanged from the round you approved, **except one fix that this rebase exposed** — worth a look, since it was not in what you reviewed.

**Startup validation walked the agent loader even when compaction was disabled.** `compactionvalidate.AgainstAgents` enumerated every application before checking whether there was any config to dry-run against. `againstAgents` already returned nil for a nil config, so the walk was pointless work on every server start — and it made a server that never configured compaction depend on its loader being listable.

Main's new `TestNewServerDebugAPIGate` (from #1413) constructs a server with a loader that is a nil pointer behind a non-nil interface. `loader == nil` does not catch that shape, so `ListAgents` panicked and `NewServer` took the process down during construction:

```
agent.(*singleLoader).ListAgents(0x0?)
compactionvalidate.AgainstAgents(0x0, ...)
adkrest.NewServer(...)
```

Fixed by hoisting the nil-config check to the top of both `AgainstAgents` and `AgainstRootAgent`, before either touches the loader or the root agent. `TestValidationDoesNotTouchTheLoaderWhenCompactionIsOff` pins it with a loader that panics if asked to list; removing the guard makes it panic.

One conflict in this rebase, in `server/adkrest/handler.go`: #1413 added a `DebugAPIConfig` field to `ServerConfig` while this stack added `Compaction`. Both additive, both kept.

Every slice was rebuilt from the rebased tree and verified to build, lint and test on its own. Full suite clean.

karolpiotrowicz (dismissed):
Re-approving at ab9a132c after the rebase onto main 9db1ff12. Content identical to what I approved at 506124c2 — the one file the range-diff flags, stream_query.go, differs only in context position and its added and removed lines are unchanged.

karolpiotrowicz (approved):

baptmont:
### Round update: rebased onto main, pinned SHA moved

**#1234 has merged** (`664cec63` on `main`), so the branch below this one is now
part of `main`. This branch has had `main` merged into it — the repository
requires a branch to be up to date before it can merge, and `BEHIND` is a hard
block here.

**The pinned commit changed, and the old one is genuinely misleading now.**

- Review `6f01f04a5fa10533eba93d04f61bf65ed4904f5e`.
- The previously pinned `ab9a132c` predates the merge of `main` and differs from
  the current head in `runner/runner.go` and `runner/runner_test.go` (+52 lines).
  Those lines arrived from `main` via #1234; they are not part of this PR. An
  agent reviewing `ab9a132c` would be reading this slice on top of a `main` that
  does not yet contain #1234.

**Nothing in this slice's own code changed this round.** The only difference
between the old pin and the new head is `main` arriving. The diff against `main`
is now exactly the serving-surfaces change — `cmd/launcher/**`,
`server/adkrest/**`, `server/adka2a/**`, `server/agentengine/**` — with no
`session/` or `internal/compactioninternal/` files, because those are in `main`
already.

One stale phrase in the description is also fixed: the config field is
`Compaction`, not `EventsCompactionConfig`; it was renamed during review.

PR #1236, opened 2026-07-31
test(llmagent): drive context compaction against a recorded real model

baptmont:
Last of five. **Re-cut**: this PR previously showed code that a later PR
corrected. It now stands on its own.

## What this adds

The end-to-end test and the worked example.

Every other test in the stack uses a fake summarizer, which is right for pinning
behaviour but says nothing about whether a real model, handed a real transcript
through the real prompt-assembly path, produces something usable. This one runs
the full chain against recorded traffic: turns accumulate, the threshold trips,
a summary is written, and the next prompt carries the summary in place of the
turns it covers while the raw tail stays intact.

The assertions are load-bearing rather than structural. A compaction that
deletes its covered events without substituting anything fails here, and so does
one whose summary never reaches the prompt.

## The example arms one strategy

The sliding window, with the tail-retention pair commented out and what to
delete in order to swap. Arming both is what the package documentation warns
against: the sliding window consumes the events tail retention would summarize,
so the ceiling never applies. The example previously armed both and called the
two triggers independent.

The documented command is `web webui`, not `web`. The bare form exits with
"no active sublaunchers found".

## Recording note

`agent/llmagent` carries a broad `//go:generate go test -httprecord=Test`
directive, and `-httprecord` matches cassette *file paths* rather than test
names, so the documented package-wide command re-records all 20 cassettes in the
package rather than the one you meant. Filed separately as #1330. To refresh
only this cassette:

    go test ./agent/llmagent/ -httprecord='TestCompactionE2E\.httprr' -run TestCompactionE2E

## Testing plan

`go mod tidy -diff` clean in both modules. This is the stack tip, and its tree
is byte-identical to the pre-re-cut tip, `734b2d03`. Known-unrelated failure:
`TestConfigureExporters` in `./telemetry`.

## Changed since the last review

**The cassette is re-recorded**, which was the blocker on the thought-signature
finding you raised. That fix is in #1232 now: a signature is a handle on one
model's reasoning within one exchange, and a summary goes to a different model
on a different call, where it means nothing and only costs tokens. The traffic
shows it: the first request after a compaction went from 5,208 bytes to 1,883,
and the whole cassette from 45,757 to 37,149. The signatures the agent replays
for its own earlier turns are untouched, because those do go back to the model
that issued them.

Re-scanned for credentials after recording: no API key, no authorization or
bearer token, no quota-project header, and the surviving header set matches the
other cassettes in the directory.

**The re-record command works now.** `compactionModel` memoises per test so the
agent and the summarizer share one instance, rather than opening two recorders
that truncate one trace.

**The size claim is corrected, and not the way I expected.** I added the
assertion the doc promised and it failed: 782 characters before the compaction,
835 after. Not the signature, which is not text and never appeared in that
measurement. Three short turns get replaced by one model-written summary, and a
summary of two one-line turns is longer than the two lines. Compaction pays off
against real history and this fixture is not that, so the doc now says what the
test establishes. A loose bound stays, so something non-prose riding along would
still be caught.

**Review this exact commit: `fb50537fd2d83cc97cafad047a49b7d2f4179475`.** Check out the SHA, not a branch
name. Several branches in this repository have similar names and hold older
versions of this same code: `baptmont/compaction-03-telemetry` (`ff5b84bb`)
and `baptmont/compaction-07-review-fixes` (`888ee5c1`) belong to closed PRs
from an earlier shape of the stack, and the pre-re-cut tips are still reachable.
Two agents auditing this stack have already spent a full run on the wrong tree
and reported bugs that no longer exist. If `internal/compactioninternal` has no
`coverIndex` and no `RepairAfterAppend`, you are on the wrong commit.

**Deliberate decisions, so they are not re-reported as defects.** Each is
documented at the code:

- `HasUsableSummary` is only a nil check. A record whose content holds no prose
  still covers its range and deletes those events. The stricter check was
  written and reverted: adk-python checks only `is None`, this stack tracks it,
  and nothing here can produce such a record.
- Tail-retention summaries do not run the plugin pipeline. adk-python runs no
  plugin hook on either compaction path, so this matches it.
- The repair after the append is one round. A straggler landing during the
  repair would need another, and each round strictly shrinks what the record
  covers on both axes -- the timestamp range and the stream position. It shrinks
  only the range if the positional guard is taken from the surviving record, and
  that was a real defect, fixed in round 6 by `foldCorrections`.
- `session/database` orders ties by `id`, which is stable but arbitrary. True
  insertion order needs a sequence column, which Vertex AI would not carry.
- The sliding window and tail retention do not compose; enabling both is
  documented as leaving the prompt unbounded, and adk-python behaves the same.

**Known flakes**, both unrelated to this stack and both passing on re-run:
`TestDelegation_04_ChatToTwoSingleTurnParallel` (issue #1313) and an occasional
golangci-lint runner error reporting `no export data for github.com/google/uuid`.

**Verification:** build, `go test -race -count=1 -shuffle=on`,
`golangci-lint run` and `go mod tidy -diff` are clean at this commit in both
modules, and CI runs on this stack now.

### Decisions taken since round 3, so they are not re-reported

Each is documented at the code. Raise them if you disagree, but they are choices
rather than oversights:

- **Parallel fan-out costs 8x on tail retention.** Eight children each pass the
  gate and seven summaries are discarded. The output and the prompt are correct;
  it is model spend. Making siblings share a gate walks back the per-agent
  scoping that fixed the loop-agent case, so it stands.
- **`session/database` breaks a timestamp tie on `id DESC`.** Two random UUIDs,
  which is not conversational order. adk-python orders identically, and after
  the repair fix nothing correctness-critical depends on tie order.
- **The sliding-window path runs the plugin pipeline and tail retention does
  not.** adk-python runs no plugin hook on either, so the first is a deliberate
  divergence in the safe direction and the second matches the reference. The gap
  is documented on the exported surface, at `compaction.Config.TokenThreshold`.
- **Two records over an identical range are resolved by stream position.** This
  cannot tell a re-summarization from two independent summaries that happen to
  span one instant, because the record stores holes rather than membership.
  Preferring the later one keeps the documented property that a re-summarized
  range never appears twice, at the cost of the rarer case. Holes are unioned
  across records that share a range and the same summary text, which is what a
  correction is, so a correction cannot lose its straggler to an ordering tie.
- **`TestNoEventIsDroppedWithoutHavingBeenSummarized` is breadth, not a
  substitute for the targeted tests.** It found a real unsoundness during
  development, and it does not currently generate the shapes that reproduce the
  tie-safe cut, the hole union or the content check. Its comment says so.
- **The two `raw_event` Vertex recordings pin a whole `session.Event` JSON dump**
  and are brittle to any new field on that struct. Out of scope here: not
  compaction code.

## Round 6 — self-review before re-requesting review

Four independent reviewers were run over the stack before asking for another
round. Every fix below is mutation-tested: the mutation that reproduces the
original defect is named in the test's doc comment.

**Two defects that deleted conversation.**

- `copyAny` returned any payload that was not `map[string]any`/`[]any`
  untouched. A tool payload here is not decoded JSON -- it is whatever a Go
  handler returned, stored as-is -- so `loadmemorytool`, which stores
  `[]memory.Entry` holding a `*genai.Content`, handed a summarizer live pointers
  into stored history. Now a depth-bounded reflective deep copy.
- The repair pass covered an event it never saw. A correction shrinks coverage
  on the timestamp axis and moves forward on the stream axis; the positional
  guard used the survivor's position, so an event appended between a record and
  its correction was outside the original's reach and inside the survivor's.
  Holes were already folded across the correction group and position was not;
  `foldCorrections` now folds both, taking the group's earliest position.

**Four more.** `newCoverIndex` indexed contentless records that prompt assembly
skips, leaving events permanently raw and permanently uncompactable at once;
tail retention drew its window from the whole session while the threshold that
triggered it was scope-filtered, so a sub-agent could spend its one compaction on
a sibling branch; token usage was dropped on the error path, so the most
expensive failure mode reported zero spend; and the corrective write ran on the
caller's context, so a cancellation between the two appends left a record
claiming a range nothing had summarized.

**Four tests that could not fail** were found and strengthened, including the
alias guard itself, which compared only top-level `genai.Part` pointers and so
was blind to a fresh `Blob` holding the original's byte slice.

**Known to remain.** When the only uncovered candidates ahead of the retained
tail sit inside the previous range -- what a repair manufactures -- the rolling
seed can produce a range strictly narrower than the record it replaces, so
neither subsumes the other and both summaries materialize in one prompt. It is
self-limiting and costs a wasted compaction rather than data. Left deliberately:
both candidate fixes carry hole-inheritance subtleties whose failure mode is
deleting conversation, so it wants its own change and its own review.


agent/llmagent/llmagent_compaction_test.go
```diff
@@ -0,0 +1,246 @@
...
+	"google.golang.org/adk/v2/model"
+	"google.golang.org/adk/v2/session"
+	"google.golang.org/adk/v2/session/compaction"
+	"google.golang.org/adk/v2/tool"
+	"google.golang.org/adk/v2/tool/functiontool"
+)
+
+//go:generate go test -httprecord=TestCompaction
+
+// TestCompactionE2E drives a real model through enough turns to trigger a
```
karolpiotrowicz, line 45: [major] The comment below states the test's reason to exist — that only a real request can show the substituted history still satisfies the API (role alternation, no orphaned function response, recovered call ordering). Replay cannot establish that. `httprr` keys on the full request wire form (request line, headers, body) and does a single exact map lookup that hard-errors on a miss (`internal/httprr/rr.go:276-308`, `:334-338`), so the pass/fail signal is byte equality of the prompt with the recording, which is orthogonal to well-formedness. Concretely, the test cannot distinguish a cosmetic one-space edit to the agent instruction from a summary re-authored with role `user` — two consecutive `user` turns, a genuine role-alternation break: both fail at `llmagent_compaction_test.go:156` with the same `cached HTTP response not found` message. Of 13 deliberate mutations, 5 survived undetected (widening the summary range, disabling `recoverCompactedFunctionCalls`, disabling `longestSelfContainedPrefix`, physically deleting covered events, blanking their content), and all 8 that were caught failed at `:156` — the turn loop, which runs ahead of every assertion — so none of the assertions at `:168`, `:173`, `:188`, `:191` or `:194` did the catching.

Suggestion: keep the cassette for what it genuinely is, a regression lock on prompt bytes, and document it as such; then add the well-formedness checks offline against the `prompts [][]*genai.Content` the test already captures — assert role alternation over `prompts[len-1]`, assert every `FunctionResponse.ID` has an earlier matching `FunctionCall.ID`, and assert the covered timestamp range equals the events replaced. Those need no model, would catch four of the five survivors, and keep working after a re-record.

agent/llmagent/llmagent_compaction_test.go
```diff
@@ -0,0 +1,246 @@
...
+// sliding-window compaction, then checks that the next prompt is both smaller
+// and still accepted.
+//
+// Everything else about compaction is covered offline with fake summarizers and
+// fake models, which is faster and far less brittle. The one thing those cannot
+// establish is that a compacted prompt is well formed, because a fake model
```
karolpiotrowicz: [major] The property named here — no function response left without its call, and recovered call ordering surviving — is never exercised by this recording. Decoding the committed cassette, the only prompt assembled after a summary exists (exchange 5, the one `:177-180` calls the interesting one) is two contents, both plain text: the summary and `What was my favourite colour again?`, with zero `functionCall` and zero `functionResponse` parts. With `CompactionInterval: 2` over three turns the compaction window swallows the entire tool exchange, so the case these guards exist for — a summary boundary landing between a call and its response — is not in the recording, and disabling either `recoverCompactedFunctionCalls` or `longestSelfContainedPrefix` leaves the test green. Extending the scenario with a tool call issued after the compaction point (a fourth turn such as "and in Geneva?") would put a summary and live function traffic in the same prompt, which could then be asserted offline.

karolpiotrowicz (commented):
An end-to-end test against a recorded model is the right instinct for this
feature, and the cassette is clean. But the test does not currently establish
the property it was written for.

**The assertion is a golden-bytes lock.** httprr keys replay on the full request
body, and the `t.Fatalf` on a replay miss sits at `:156`, ahead of every
assertion in the file. So *any* change to the assembled prompt fails, and they
all fail identically — a one-space edit to an instruction and a genuine
role-alternation break produce the same message at the same line. Running 13
mutations against it: 5 survive, including the wrong-covered-range mutation, both
function-call/response pairing guards, and two store violations. All 8 kills
report at `:156`, so none of them came from an assertion in the test. The
structural reason is worth stating plainly: with a byte-exact replay gate ahead
of the assertions, the test can only detect "the prompt changed", never "the
prompt is malformed".

**The scenario also never exercises the property.** Exchange 5 of the cassette is
`model(text) | user(text)` with zero function parts, so the
call/response-pairing invariant is untested even at record time.

Splitting this in two would give both halves real teeth: an offline test that
asserts structurally on the assembled prompt — role alternation, and every
function response having its matching call — with no cassette in the path; plus
the recorded test extended to include function traffic across a compaction
boundary.

Two things this PR inherits but is the first to compose, so they are worth
flagging here: with compaction **disabled**, a single planted compaction record
still replaces the whole history, because `Apply` at
`contents_processor.go:122` never consults the config — so the "off by default"
guarantee holds only for sessions that contain no compaction record. And
summaries are never re-summarized: at 40 turns with the shipped config the
prompt carries 19 of them.

---

### Additional findings

[major] **`session/vertexai` silently drops `Actions.Compaction`** (`session/vertexai/vertexai_client.go`, in `createAiplatformpbEventActions` / `eventNeedsRawEvent` / `aiplatformToSessionEventActions`). A compaction summary event round-tripped through the Vertex AI session service loses its `Actions.Compaction` with no error: `createAiplatformpbEventActions` returns nil for it, `eventNeedsRawEvent` returns false so the `raw_event` fallback is never taken, and the value read back is nil. Compaction on this backend is therefore write-only — the summary is stored as a contentless event and every later prompt is assembled as if no compaction had happened, so the prompt keeps growing on the backend most likely to host long sessions and the failure surfaces far from its cause. This is not a platform limit: serialising the same `Actions` through the existing `raw_event` path yields `{"compaction":{"compactedContent":…,"startTimestamp":…,"endTimestamp":…}}` correctly, and adk-python persists the equivalent through a `_compaction` custom-metadata key. Adding `Compaction != nil` to `eventNeedsRawEvent` is a one-line fix that reuses the working path; `session/sessiontestsuite/service_suite.go` has no `Compaction` case at all, so no backend is currently held to this. Raised at stack level — not introduced by this PR.

[major] **Summaries accumulate without bound** (`internal/compactioninternal/apply.go`, `internal/compactioninternal/window.go`). Sliding-window compaction appends one summary per window and never subsumes earlier ones, so the number of summaries in the prompt grows linearly with conversation length. Driving 40 turns with the interval and overlap that `examples/compaction/main.go` ships puts 19 summaries in a single prompt, and consecutive summaries share a turn, so the same content is re-sent repeatedly. The feature lowers the constant factor but leaves the prompt O(turns), which does not match the premise that compaction keeps the token count from climbing on long sessions. Subsuming a summary whose range is contained in a newer summary's range, or re-summarising accumulated summaries, would bound it; at minimum the growth characteristic deserves to be documented.

[major] **Partially overlapping summaries are both materialized** (`internal/compactioninternal/apply.go`). When two summaries overlap partially, `Apply` emits both, so the events in the overlap are represented twice in the assembled prompt — summaries covering `e1..e3` and `e2..e5` both materialize, duplicating `e2` and `e3`. This is not an edge case under the shipped configuration: with `OverlapSize > 0` — used by `examples/compaction/main.go` and by this test at `:145` — every consecutive pair of summaries overlaps by construction, so it compounds the accumulation above. Suggest dropping any summary whose covered range intersects a later summary's range, or merging the ranges, so each event is represented exactly once.

[major] **`EventRetentionSize: 0` passes `Validate()` and swallows the in-flight user question** (`internal/compactioninternal/tail_retention.go`; `compaction.Config.Validate()`). Setting `TokenThreshold` while leaving `EventRetentionSize` at its zero value validates cleanly, and the prompt that results contains the summary but not the question that triggered it — the model is asked to answer something it was never shown, with no error anywhere. The same scenario with `EventRetentionSize: 2` keeps the question, isolating the zero value as the cause. Because this is the Go zero value, anyone configuring only `TokenThreshold` lands on it and `Validate()` endorses the result. Suggest rejecting `EventRetentionSize == 0` when `TokenThreshold > 0` (or defaulting it to a sane minimum), and unconditionally exempting the current invocation's events from tail retention.

[major] **Three exported constructors are API-incompatible** — `NewRuntimeAPIController`, `NewEventarcController` and `NewPubSubController` in `server/.../controllers`. Each gained a trailing variadic options parameter, which `apidiff` classifies as Incompatible against `main`. Ordinary call sites still compile, but function-value and interface-satisfaction uses do not: assigning `controllers.NewRuntimeAPIController` to a variable of its previous function type is now a compile error. The change arrived in #1235 rather than here (the diff between #1235 and #1236 is API-clean), so this is raised at stack level: either keep the old constructors and add `NewXWithOptions` variants, or record the exception explicitly in the description with a release-note entry.

[major] **No CI runs on any PR in this stack** (`.github/workflows/go.yml:7-13`). The Go workflow triggers only on pull requests targeting `main` or `v1`; every PR here targets the previous branch in the stack — this one targets `baptmont/compaction-05-serving-surfaces` — so the workflow never fires, and `gh pr checks 1236` returns `cla/google` alone. `lint (.)` and `test (.)` are required on `main`, so they will first execute when the whole stack merges, on the combined result. That leaves the Testing Plan as an unverified self-report at review time and makes this PR's "so CI stays hermetic" argument unobservable on the PR arguing for it. Adding the stack's branch prefix to `pull_request.branches` (or using `branches: ['**']`) would verify each PR where it is reviewed.

[minor] **The feature ships with no user-facing documentation.** A new public `session/compaction` package and a `runner.Config.EventsCompactionConfig` field land with no prose anywhere in the repository — `grep -ril compaction --include='*.md' .` returns nothing, and there is no `docs/` or `website/` tree. The package doc also carries a dangling link: `session/compaction/compaction.go:21` references `[Apply]`, which lives in `internal/compactioninternal` and so cannot resolve from an exported package. Package-level prose covering the two strategies, the config fields and the ordering guarantees would make the feature learnable without reading the source; the `[Apply]` reference should name an exported symbol or drop the brackets.

[minor] **`examples/compaction/main.go:25-27` names serving surfaces the launcher does not serve.** The doc comment says the launcher serves the console, the web UI, A2A and Agent Engine, but `cmd/launcher/full/full.go:32-33` composes console, webui, a2a, pubsub, eventarc and api: `agentengine` is absent even though `cmd/launcher/agentengine/` exists, and the two trigger surfaces that are composed go unmentioned — the least obvious ones, and the ones most worth an example. Given there is no other documentation for the feature, this is the artifact readers will follow. Correct the comment to match `full.go`, or compose `agentengine` if the example's claim is the intended behaviour. Introduced by #1235.

[minor] **The summarizer's own prompt size pollutes the tail-retention threshold** (`internal/compactioninternal/tail_retention.go`). Token accounting reads the newest response's `promptTokenCount`, but immediately after a compaction the newest response is the summarizer's, whose prompt is unrelated to the agent's: a 50000-token summarizer prompt is read as the agent's prompt size when the agent's real figure is around 100, so a 1000-token threshold fires when it should not. End to end this moves the summarizer call count in the wrong direction — `TokenThreshold=0` gives 3 calls, `TokenThreshold=100000` gives 5. It self-corrects once a later response lands, hence minor, but it happens on the turn immediately after every compaction. Skipping compaction-authored events when locating the newest usage record would fix it.

agent/llmagent/llmagent_compaction_test.go
```diff
@@ -0,0 +1,246 @@
...
+// The cassette is sensitive to anything that changes prompt bytes, including the
+// summarizer prompt template, the transcript line format, tool-argument
+// rendering and truncation behaviour. Any of those changes requires re-recording.
+func TestCompactionE2E(t *testing.T) {
+	// Matches llmagent_delegation_test.go, which is the most recently recorded
+	// suite in this package. Change it only alongside a re-record, since the
+	// model name is part of the request URL the cassette keys on.
+	const compactionModelName = "gemini-3.5-flash"
+
+	// Skip until the cassette exists, so an unrecorded checkout still has a
```
karolpiotrowicz: [minor] Skipping when the cassette is absent turns a lost or renamed cassette into a silent pass: removing `testdata/TestCompactionE2E.httprr` yields `--- SKIP`, `PASS`, exit 0. This is the only skip-on-missing-cassette in the repository — every other cassette test goes through `NewGeminiTestClientConfig`, where a failing `httprr.Open` is fatal (`internal/testutil/genai.go:31-38, :63-69`). The stated rationale, that an unrecorded checkout keeps the suite green, is moot because the cassette is committed, so the branch is only reachable by deleting or renaming the file — which is precisely the outcome a package-wide record run produces (see the note on `:38`). Suggest dropping the skip and letting `t.Fatal` fire like the other cassette tests, or gating it on an explicit opt-in environment variable so it cannot trigger by accident.

agent/llmagent/llmagent_compaction_test.go
```diff
@@ -0,0 +1,246 @@
...
+	"google.golang.org/adk/v2/internal/httprr"
+	"google.golang.org/adk/v2/internal/testutil"
+	"google.golang.org/adk/v2/model"
+	"google.golang.org/adk/v2/session"
+	"google.golang.org/adk/v2/session/compaction"
+	"google.golang.org/adk/v2/tool"
+	"google.golang.org/adk/v2/tool/functiontool"
+)
+
+//go:generate go test -httprecord=TestCompaction
```
karolpiotrowicz: [minor] This adds a third `//go:generate` directive to a package whose existing directive already matches every cassette in it: `llmagent_test.go:48` runs `go test -httprecord=Test`, which is unanchored and matches all 20 files under `testdata/`. In record mode `httprr.Open` creates — and therefore truncates — the trace file before any request is attempted (`internal/httprr/rr.go:107-152`); measured on a copy of this cassette with no network involved, it went from 23450 bytes to 16 (just the `httprr trace v1` header). So one `go generate ./agent/llmagent/...` empties all 20 cassettes, and `AGENTS.md:143-145` documents package-wide `go generate` as the recording procedure, contradicting the warning at `:69-74`. The hazard predates this PR, which is why this is minor, but it is worth anchoring the record patterns per file and having `httprr` write to a temp file and rename on success rather than truncating in place.

agent/llmagent/llmagent_compaction_test.go
```diff
@@ -0,0 +1,246 @@
...
+		Tools:                    []tool.Tool{weather},
+		BeforeModelCallbacks:     []llmagent.BeforeModelCallback{capture},
+		DisallowTransferToParent: true,
+		DisallowTransferToPeers:  true,
+	})
+	if err != nil {
+		t.Fatalf("llmagent.New() error = %v", err)
+	}
+
+	// Interval 2 keeps the recording short. Overlap 1 exercises the seam logic,
```
karolpiotrowicz: [minor] The justification for `OverlapSize: 1` does not hold for this recording: there is no second window. The cassette contains exactly one summarizer prompt — exchange 4, the only request with no `systemInstruction` — hence one compaction and one window, consistent with `summaries := make([]*session.Event, 0, 1)` at `:162` and the `len(summaries) == 0` check at `:168`. No seam is ever formed, so nothing reaches back into an already-summarized turn and the overlap setting has no observable effect beyond shifting which events the single window selects. Either extend the recording so a second window actually forms (which would also help the gap noted at `:46`), or reword the comment to say overlap is set only to fix the window selection.

agent/llmagent/llmagent_compaction_test.go
```diff
@@ -0,0 +1,246 @@
...
+		OverlapSize:        1,
+	})
+
+	const sessionID = "compaction_session"
+	turns := []string{
+		"What is the weather in Zurich?",
+		"My favourite colour is teal, remember that.",
+		"What was my favourite colour again?",
+	}
+	for i, turn := range turns {
```
karolpiotrowicz, line 225: [minor] The recorded final answer is the memory-recall evidence the PR body claims — the model still answering correctly from the summary alone — and the test discards it: the first return value of `testutil.CollectTextParts` is `_`, and nothing else in the test inspects model text. The cassette's fifth response is `Your favorite color is teal.`, produced from a prompt containing only the summary and the question, so the property genuinely held at record time; a regression producing a summary the model could not answer from would still pass. A substring check is not wording-pinning — the test already does one at `:188` — so capturing the last turn's text and asserting `strings.Contains(strings.ToLower(last), "teal")` recovers the most valuable signal in the recording at negligible cost.

agent/llmagent/llmagent_compaction_test.go
```diff
@@ -0,0 +1,246 @@
...
+//
+// Everything else about compaction is covered offline with fake summarizers and
+// fake models, which is faster and far less brittle. The one thing those cannot
+// establish is that a compacted prompt is well formed, because a fake model
+// accepts anything handed to it. Only a real request can show that the
+// substituted history still satisfies the API: role alternation holds, no
+// function response is left without its call, and the recovered call ordering
+// survives. That is the whole reason this test talks to a model.
+//
+// Deliberately not asserted: the wording of the summary. It is model output, and
```
karolpiotrowicz: [minor] This says the summary's wording is deliberately not asserted, but `:188` asserts the prompt contains the recorded summary verbatim and `:191` asserts the summary does not quote `turns[0]` verbatim — both are properties of model wording. `:191` passes only because the recorded summary paraphrases ("The user initially asked about the weather in Zurich."); had the summarizer quoted the question, which is ordinary behaviour, the assertion would fail with nothing actually wrong, and it can only ever fire on the run that re-records — exactly when it is hardest to tell a regression from a rewording. Note also that `turns[1]` is absent from the final prompt and never asserted on, so the substitution property is spot-checked on only one of the two covered turns. A structural form is more robust: derive the covered event set from `summaries[…].Actions.Compaction.StartTimestamp`/`EndTimestamp` and assert none of those events' contents appear in `prompts[len-1]`.

baptmont:
Triaged all seven against the current code and the committed cassette. All seven stand, and two of them I can now state more strongly than the comment did.

On `3701666664`: I decoded the cassette. It holds five exchanges, and exchange 5, the only post-summary prompt, is exactly `[model: summary text, user: question]` with zero `functionCall` or `functionResponse` parts. So `recoverCompactedFunctionCalls` and `longestSelfContainedPrefix` are provable no-ops against this recording, and the call-pairing property the test exists to demonstrate is not exercised at all.

On `3701666676`: stronger than "there is no second window". The first compaction has `firstNew == 0`, so `max(0, firstNew-overlap)` is `order[0]` regardless of `OverlapSize`. The setting is inert here by construction, not just unexercised.

Two of these cannot be fixed without recording new live traffic, which needs credentials, so I will bring them back to the maintainer rather than guess. The rest, including the misleading doc comments at `:44-50` and `:52-54`, the silent skip on a missing cassette, and the redundant `go:generate`, are offline fixes and are on the list.

agent/llmagent/llmagent_compaction_test.go
baptmont: Fixed by making the claim true rather than by deleting it.

You are right that replay cannot establish this. It keys on exact request bytes, so a structural defect introduced later surfaces as a cassette miss and aborts the run on the turn loop, ahead of every assertion. The doc now says exactly that, and the structural property is asserted here directly and offline instead, over the prompts the agent actually sent rather than inferred from the fact that a recorded model once accepted the bytes.

`assertNoOrphanFunctionResponses` runs over every captured prompt and checks that each function response is preceded by the call it answers, keyed on call ID where the model supplies one and on name otherwise. It is load-bearing: stubbing out the call-recording line makes prompts 1 and 2 fail with `carries a function response "get_weather" with no preceding call`.

That last detail also sharpens your other comment. Pairing is exercised, on the two pre-compaction prompts. What is not exercised is pairing across a summary, which is the specific gap, and I have written it up as such.

agent/llmagent/llmagent_compaction_test.go
baptmont: Confirmed, and I verified it independently rather than taking it on trust: the final prompt is two contents, a `model` summary and a `user` question, with zero function parts. So `recoverCompactedFunctionCalls` and `longestSelfContainedPrefix` are inert against this recording, exactly as you say.

I have not re-recorded, because that needs live credentials and a fourth turn calling a tool after the compaction point, which is a maintainer decision rather than mine. What I have done is stop the test claiming the property. The doc now states the gap in the terms you set: the recorded conversation issues no tool call after the compaction point, the two paths are therefore not exercised here, they are covered offline in `internal/compactioninternal`, and closing the gap needs a re-record with a fourth turn.

The structural assertion added for `3701666662` does now check call pairing on the two pre-compaction prompts, so the uncovered case is specifically pairing *across* a summary rather than pairing in general.

agent/llmagent/llmagent_compaction_test.go
baptmont: Fixed. A missing cassette is now `t.Fatalf`, matching every other cassette test in the package. The cassette is committed, so its absence means it was lost or renamed rather than never recorded, and skipping turned that into a silent pass. The message says as much and carries the re-record command.

Recording mode still proceeds when the file is absent, since that is the run that creates it.

agent/llmagent/llmagent_compaction_test.go
baptmont: Fixed, directive removed. The package-level `-httprecord=Test` at `llmagent_test.go:48` is unanchored and already matches every cassette here, so the third directive added nothing except another way to re-record all of them by accident. The comment now says why the absence is deliberate, so nobody helpfully adds it back.

agent/llmagent/llmagent_compaction_test.go
baptmont: Fixed, and your stronger version is the one I used. It is not merely that this recording has no second window: the first compaction of a session starts at the first invocation whatever the overlap is, because `firstNew` is zero and `max(0, firstNew-overlap)` is `order[0]` regardless. The setting was inert by construction, not just unexercised.

So `OverlapSize` is gone from the config rather than re-justified, and the comment says overlap needs a second window and therefore belongs in the offline tests where windows are cheap.

Removing it changed no request bytes, as expected: the cassette still replays without a re-record.

agent/llmagent/llmagent_compaction_test.go
baptmont: Fixed. The final answer is captured and asserted to mention the colour, so the memory-recall property the PR body claims is now actually checked rather than discarded into `_`.

It is load-bearing: asserting on a different colour fails with `the model could not recall the colour from the summary alone; answer was "your favorite color is teal."`, which also shows the assertion is reading real recorded output.

agent/llmagent/llmagent_compaction_test.go
baptmont: Fixed. The two statements now say different things, which is what was missing. The doc says no particular wording is asserted, because that is model output; what is asserted is that whatever the summarizer produced is what reaches the next prompt, which is what the containment check at the bottom actually tests.

Your other observation is fixed too: `turns[1]` is inside the compacted range and was never asserted absent, so half the range went unchecked. The test now asserts both covered turns are gone. Inverting that check confirms it bites, since turn 2 really is absent from the final prompt.

agent/llmagent/llmagent_compaction_test.go
baptmont: Closed properly rather than documented away: the cassette is re-recorded with a fourth turn that calls a tool after the compaction point.

The recording now contains a prompt with six contents carrying a summary, a function call and its response together, which is the arrangement pairing has to survive and the thing this test is better placed to check than an offline test.

Two assertions had to move with the extra turn, both because they were reading the wrong thing rather than because a property changed: the recall check now reads the answer to the turn that asks about the colour rather than whichever turn is last, and the summary check uses the first stored summary rather than the last, since with four turns the last compaction is written after the final model call and cannot appear in any recorded prompt.

The new arrangement is asserted, not assumed: at least one prompt must carry a summary and function traffic together, so a later re-record that stops calling tools after the compaction point fails rather than quietly losing the coverage again.

Recorded against `gemini-3.5-flash`. No credentials in the cassette: the request scrubber drops the API key header and I checked the tree for key material afterwards.

baptmont:

## Landed

| Decision | Where | Note |
|---|---|---|

## Four things worth flagging

## Still open

baptmont:

## Thread state

| PR | Resolved | Left open |
|---|---:|---:|
| #1231 | 16 | 4 |
| #1232 | 13 | 6 |
| #1233 | 8 | 7 |
| #1234 | 8 | 13 |
| #1235 | 3 | 5 |
| #1236 | 4 | 3 |

## State of the code

agent/llmagent/llmagent_compaction_test.go
baptmont: Strengthened further. Beyond the call-pairing check, the test now asserts the compacted range directly: every event inside the first compaction's range must be absent from the final prompt. That is the contract, where the two named turns were only a sample of it.

Role alternation I have deliberately left out. Adjacent same-role turns are normal in this API, as you noted yourself on a related thread, so asserting strict alternation would pin something that is not actually required.

agent/llmagent/llmagent_compaction_test.go
baptmont: The documentation is now consistent with the behaviour. Two paragraphs still described the skip, including the recording instructions telling the reader the test skips until a cassette exists. Both corrected.

agent/llmagent/llmagent_compaction_test.go
baptmont: Accepted as is.

The doc contradiction is fixed and turn 2 is now asserted absent, which were the substantive parts. The containment checks do compare the recorded summary text, so a re-record changes what they match against, but that is inherent to a replayed test asserting that the stored summary is what reaches the prompt: comparing anything looser would stop pinning the property.

karolpiotrowicz (commented):
Reviewed at `30f5061`. The doc-only commits pushed since then do not change any line cited below.

The [missing-cassette guard](https://github.com/google/adk-go/blob/30f50610f24d16b19c038c2f6b339109bca9d9d4/agent/llmagent/llmagent_compaction_test.go#L98-L104) is a real improvement — deleting the file now fails with a re-record hint instead of skipping. The problem is that the assertions it protects cannot run when they matter.

**1. Every structural assertion sits behind a `t.Fatalf` that fires first.** The turn loop [aborts on any replay miss](https://github.com/google/adk-go/blob/30f50610f24d16b19c038c2f6b339109bca9d9d4/agent/llmagent/llmagent_compaction_test.go#L176-L182) , and a compaction defect changes the prompt bytes, so it always misses. Concretely: I made the compaction path orphan a function response in the final prompt only, so turns 1–3 replay identically. Shipped code dies at line 179 with `cached HTTP response not found`. The same defect with that `t.Fatalf` downgraded to `t.Logf` produces `prompt 5 carries a function response "get_weather" with no preceding call` — the assertion is correct and unreachable. This is mechanical rather than incidental: `capture` runs in a `BeforeModelCallback` and nothing mutates `req.Contents` before the request goes out, so any structural defect the offline checks could see is also a byte change, and the byte change aborts first. Moving the structural sweep inside the turn loop and checking the run error after it is a small change that makes the assertions load-bearing.

**2. A compaction that deletes the conversation passes.** I patched the append path to drop the covered events from the store instead of summarizing them; the session went from 14 events to 2, every raw turn unrecoverable, and the test exited 0 with no diagnostic. "Compacting rather than truncating" is the property the test comment names, and nothing distinguishes "absent from the prompt" from "absent from the store".

**3. The one assertion that does fire goes vacuous under the defect it exists for.** The [covered-range loop](https://github.com/google/adk-go/blob/30f50610f24d16b19c038c2f6b339109bca9d9d4/agent/llmagent/llmagent_compaction_test.go#L249-L261) skips parts with no text, so blanking the covered events' `Parts` leaves it making zero comparisons — it asserts four in the baseline — and the test still passes. The same guard means the compacted function call and its response are never checked for absence: six events are in range and only four are inspected, and the two skipped ones are exactly the tool traffic. A minimum comparison count, and matching the call and response by ID, would close both. The loop also nil-derefs on a content-less event in range and takes the package binary down with a SIGSEGV; the filter uses `compaction.IsCompactionEvent`, which the package doc explicitly describes as answering "is there a usable summary" rather than "is this bookkeeping".

**4. Package-wide `go generate` destroys every cassette in the package.** The [unanchored directive](https://github.com/google/adk-go/blob/30f50610f24d16b19c038c2f6b339109bca9d9d4/agent/llmagent/llmagent_test.go#L48) matches every file under `testdata/`, and `httprr` truncates the trace before any request is attempted — on copies, 20 cassettes went from 325,771 bytes to 320. This predates the PR and reproduces on the parent branch, so it is not yours to fix; but this change adds the twentieth and largest file to the blast radius, the warning comment names only one of three destructive spellings, and [`AGENTS.md`](https://github.com/google/adk-go/blob/30f50610f24d16b19c038c2f6b339109bca9d9d4/AGENTS.md#L140-L146) still documents the package-wide command as the recording procedure. Anchoring the record patterns per file would settle it.

Two smaller ones. The absence assertions compare summary text against covered turn text, and the recorded summary says "The current weather in Zurich is sunny." while a covered event says "The weather in Zurich is currently sunny." — one word's ordering from a false failure on any re-record, at exactly the moment a maintainer can least tell a rewording from a regression. And the guard accepts a 16-byte header-only stub left behind by a failed re-record, then dies 80 lines later with a message that never mentions the cassette.

Worth keeping in the file: stripped of what it cannot establish, this test does prove three things nothing else here proves — a real model accepted a compacted prompt at record time, a summary and live tool traffic coexist in one accepted prompt, and one bookkeeping-range defect is caught offline. That is a narrower claim than the header comment makes, and saying so in the comment is better than dropping the test.

baptmont:
This PR is back to the three files it should be. The fixes that were briefly sitting on top of it are in #1316; that was my doing and it made this one unreadable.

Findings 1, 2, 3 and both smaller notes are fixed here.

**Finding 2 is the one that bothered me most.** You are right that it is worse than a gap: the test's header named "compacting rather than truncating" as the property and then could not tell the two apart. It asserts against the session store now, and I checked it catches your patched append path.

**Finding 3.** The sweep counts its comparisons and fails below a floor, so blanking the covered parts reports `0 comparisons` instead of passing quietly. It filters on the compaction field rather than `IsCompactionEvent`, which is what left the tool traffic unexamined, and the call and response are matched by ID rather than by text since neither carries any. The nil dereference is guarded.

I also stopped the absence checks searching the summary itself. Your Zurich example is a good catch and would have bitten whoever re-recorded next, at the worst possible moment.

**Finding 1.** The offline checks run before any run failure is reported now, so a defect gives a diagnosis rather than `cached HTTP response not found`. Your caveat stands and I put it in the comment: the framework's own pairing guard rejects an orphaned response before a prompt is sent, so `assertNoOrphanFunctionResponses` catches a defect only in a narrow window. Running it is still the difference between naming the property and naming the bytes.

**Finding 4.** Left alone deliberately. It predates this stack, it is a foot-gun for the whole package rather than for this test, and `AGENTS.md` documents the destructive command as the procedure. It wants its own change. Happy to file it if you have not.

**On the header comment.** Taken. It now says plainly what a passing run does and does not establish, which is a narrower claim than before and closer to what you described.

karolpiotrowicz (commented):
Reviewed at `06f453cc`.

Three things that were not load-bearing last round now are, and I checked each
one bites: a compaction that drops the covered events from the store instead of
summarizing them fails at
[`:287`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L286-L294)
naming the deletion, a range that claims more than it summarized fails at
[`:346`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L339-L348)
naming the event, and a 16-byte cassette stub now dies at
[`:123`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L119-L126)
with the size and the re-record command instead of eighty lines later. That is
the substance of the last two rounds, and it landed.

**The diagnosability claim is still not true, and the assertion that moved cannot
fire.** The doc at
[`:58-62`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L58-L62)
says the offline assertions run before the miss is reported, so the failure says
which property broke. One of the six assertion groups moved ahead of the
[`t.Fatalf` at `:226-228`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L219-L228),
and it is the one that cannot report anything. Make
`compactioninternal.Apply` return its input unchanged and the run dies at `:227`
with `cached HTTP response not found` and nothing else. Downgrade that one
`t.Fatalf` to `t.Logf` and the same defect reports five named failures — `:263`,
`:266`, `:269`, `:274`, `:292` — so the diagnosis the comment promises is already
written and is being discarded one line before it runs.

The assertion that did move is
[`assertNoOrphanFunctionResponses`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L414-L444),
and an orphan cannot reach it.
[`preprocess`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/internal/llminternal/base_flow.go#L566-L570)
runs to completion and returns on the first error before
[`BeforeModelCallbacks` fire](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/internal/llminternal/base_flow.go#L770-L772),
and `ContentsRequestProcessor` runs inside it — so an unpaired response either
aborts the turn at
[`contents_processor.go:325`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/internal/llminternal/contents_processor.go#L325-L330)
or is
[dropped at `:409`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/internal/llminternal/contents_processor.go#L408-L413)
before `capture` sees the request. Strip the function-call parts out of the
assembled prompt and the failure is the framework's message, never the test's.
The
[note at `:215-218`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L213-L218)
already suspects this and describes a window where the prompt is assembled but
not yet validated. There is no such window. Moving the rest of the checks ahead
of the fatal fixes both halves at once, and it is the change that makes the last
round's work visible when it matters.

**The `checked` floor is set below what the sweep actually does, and the slack is
exactly the tool traffic.** The floor is
[4](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L366-L373)
and the run makes six comparisons: four text parts plus the covered call and its
response. Blank `FunctionCall.ID` and `FunctionResponse.ID` on the covered events
and `checked` lands on exactly 4, the floor passes, the test is green, and
[the tool half at `:352-362`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L352-L362)
has made zero comparisons — the "goes quiet rather than failing" mode the floor
exists to prevent, on the pair the comment above it calls the most likely to
break. The `fc.ID != ""` guards are the mechanism: an ID-less model, the case
[`callKey`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L446-L453)
was written for, silences the same half. Counting comparisons per covered event
and requiring each to contribute one would hold whatever a re-record changes. The
two comments also disagree on the number —
[`:349-351`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L349-L351)
implies six and
[`:368-369`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L366-L370)
says four.

**`outsideSummary` is applied to one of the two places that need it.** The
reasoning at
[`:315-321`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L315-L321)
is right, and the near-miss it cites is real, but
[`:265` and `:273`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L265-L275)
still search `final` with the summary in it. The summarizer's own template
[instructs the model to "identify and reiterate the user request"](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/session/compaction/llm_summarizer.go#L38-L41),
so a re-record whose summary quotes turn 1 rather than paraphrasing it fails both
of those against a correct implementation. Searching `outsideSummary` in both
places is the whole fix.

**Turn 4 exercises call pairing, not call recovery.**
[`:190-193`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L189-L195)
and
[`:64-67`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L64-L67)
name both paths.
[`recoverCompactedFunctionCalls` recovers only long-running calls](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/internal/compactioninternal/apply.go#L238-L242),
and `get_weather` is not one, so the recovery branch is never entered. Neither
pair straddles the boundary either: Zurich's call and response are both inside
the summary and Oslo's are both outside it. What turn 4 genuinely buys is a
prompt carrying a summary and self-contained tool traffic together, which is
worth having and is
[asserted honestly at `:375-388`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L375-L388).
Narrowing the two comments to that costs nothing. Making the claim true needs a
long-running tool and a re-record, which is a bigger ask than this PR should
carry.

- **The summary needle and the prompt haystack use different renderers.**
  [`textOf`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L470-L481)
  concatenates parts bare while
  [`promptTextOf`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L483-L502)
  prefixes each with `[role] ` and a newline. A summary with two text parts makes
  `:263` fail, turns the `:321` strip into a no-op, and drives
  `withSummaryAndTools` to zero — three unrelated-looking failures from one
  cause. Today's recording has one part, so this is latent, but
  [`isProse`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/session/compaction/summary_event.go#L142-L153)
  omits the `Thought` check that
  [`hasText`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/session/compaction/llm_summarizer.go#L233-L250)
  has, so a thinking model can produce the two-part shape.
- **The re-record hint is scoped to the block that rarely fires.** `reRecord` at
  [`:113`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L111-L118)
  is local to the missing-cassette guard, so the failure a maintainer actually
  hits at `:227` mentions neither the cassette nor `git restore`, which is the
  action someone who just truncated it without credentials needs.
- **`NewTestAgentRunnerWithCompaction(t, a, nil)` returns a runner with
  compaction off**, because
  [`resolveCompactionConfig` short-circuits on nil](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/runner/runner.go#L149-L152).
  Today's only caller passes a literal, so this is a trap for the next one rather
  than a live bug.
- **Two comments the recording contradicts.**
  [`:439-442`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L438-L443)
  says the recording has no tool traffic after the compaction point, which turn 4
  was added to create, and
  [`:44-46`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L44-L46)
  says the test checks that the next prompt is smaller, which nothing measures.
- **`sessionEventsFor` uses `context.Background()`** at
  [`:457`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/agent/llmagent/llmagent_compaction_test.go#L455-L462)
  where the rest of the helpers use `t.Context()`, and re-hardcodes `test_app`
  and `test_user`. A `SessionEvents(t, sessionID)` accessor on the runner would
  remove both.

On the cassette hazard from last round: it still reproduces at this head, and
running the package-level record pattern without credentials takes
`agent/llmagent/testdata/` from 20 cassettes to 20 header-only stubs, because
`httprr` truncates before the credential check.
[`AGENTS.md:144`](https://github.com/google/adk-go/blob/06f453cc7e70254eee1ce687b87d0b183fa74872/AGENTS.md#L143-L145)
still names the package-wide command as the procedure. Not this PR's to fix, and
this PR adds the only guard in the package that catches the result — worth saying
here only because it now adds the largest file to that blast radius.

The first point is the one I would like fixed before this merges. The second is
cheap and closes the last gap the covered-range sweep has. The rest can ride
along or not.

karolpiotrowicz (commented):
No credentials reached the recording. `agent/llmagent/testdata/TestCompactionE2E.httprr` is a cassette — 45,757 bytes of real HTTP traffic captured once against the live Gemini endpoint and replayed on every later run, so the test needs no network and no key. Because it is a verbatim capture of an authenticated call committed to a public repository, the thing to check is whether the key that authorised it came along. It did not: no `x-goog-api-key`, no `Authorization` or bearer token, no `AIza…` key, no quota-project header, no cookie, no service-account identifier. The header names that survive are the same set as every other cassette in that directory.

That holds by construction rather than by luck, which is the part that matters for review. [`internal/testutil/genai.go:41-46`](https://github.com/google/adk-go/blob/4fe3602a69891c9f4ecbfd683f62b92288bc6eb5/internal/testutil/genai.go#L41-L46) registers a scrubber that deletes those headers from the copy httprr writes to disk, in both capitalisations, so a future re-record is stripped the same way. Two things about the test itself.

**The re-record command documented in this file cannot produce a valid cassette.** `newGeminiModel` is called twice, at [`:163`](https://github.com/google/adk-go/blob/4fe3602a69891c9f4ecbfd683f62b92288bc6eb5/agent/llmagent/llmagent_compaction_test.go#L163) for the agent and [`:188`](https://github.com/google/adk-go/blob/4fe3602a69891c9f4ecbfd683f62b92288bc6eb5/agent/llmagent/llmagent_compaction_test.go#L188) for the summarizer, and each call opens its own recorder on the same trace path, which is derived from the test name.

In replay that is harmless, which is why every check passes today. In record mode each is an `os.Create` — truncating, with its own independent write offset — so the two writers overwrite each other. Four requests produce a trace that stops parsing after one record, and it lands above the 1024-byte floor this file checks, so `minCassetteBytes` does not catch it. [`llmagent_delegation_test.go:72-79`](https://github.com/google/adk-go/blob/4fe3602a69891c9f4ecbfd683f62b92288bc6eb5/agent/llmagent/llmagent_delegation_test.go#L72-L79) documents this exact hazard and memoises the model per test and file to avoid it. One corroborating detail: the committed cassette is a clean contiguous 8-record stream, which two competing writers cannot produce, so whatever generated it was not the documented command.

**The doc comment promises a size check that does not exist, and the recording shows the opposite.** [`:45-47`](https://github.com/google/adk-go/blob/4fe3602a69891c9f4ecbfd683f62b92288bc6eb5/agent/llmagent/llmagent_compaction_test.go#L45-L47) says the test checks the next prompt is both smaller and still accepted. There is no size assertion anywhere in the file, and the committed cassette records `promptTokenCount` going 303 → 869 across the compaction.

A large part of that growth is a `thoughtSignature` copied from the summarizer's own response into the stored summary — 3,464 bytes of one record's 4,975 — which `IsProsePart` does not filter out. That last part is a real defect in the summary path rather than a documentation problem, and it is the more interesting half of this finding.

baptmont:
Thank you for auditing the cassette for credentials. Two of the three are fixed and the third needs something I do not have.

**The re-record path is fixed.** You were right that two recorders on one trace path cannot produce a valid cassette, and right that the committed one's clean 8-record stream proves it was not made by the documented command. `compactionModel` memoises per test so the agent and the summarizer share one instance, the way the delegation tests do for exactly this reason.

**The size claim is corrected, and the correction is not the one I expected.** I added the assertion the doc promised and it failed: 782 characters before the compaction, 835 after. Not the thought signature, which `promptTextOf` does not count. It is simpler than that. Three short turns get replaced by one model-written summary, and a summary of two one-line turns is longer than the two lines. The doc was wrong to claim the prompt shrinks here, so it now says what the test establishes, that the summary stands in for the turns it covers and the model accepts it. There is a loose bound, a doubling, so something non-prose riding along would still be caught.

**The thought signature is not fixed here, and I want to be straight about why.** You are right that it is the more interesting half: 3,464 bytes of one record, and it is not prose. The fix is two lines, stripping `ThoughtSignature` when the summary content is built, with a test. A signature is a handle on one model's reasoning within one exchange, handed back to that model later in the same exchange. A summary is not going back to the summarizer, it becomes the agent's context read by a different model, so the handle means nothing there.

Making that change alters the request this test sends, so the cassette no longer matches and the test fails on a replay miss. Re-recording needs live credentials, which I do not have. I could have edited the cassette so the stored request matches, and given that you specifically checked its provenance I was not willing to doctor it quietly. So the change is set aside as a patch rather than applied, and it wants a re-record to land.

Requests 2 to 7 in the cassette also carry signatures, but those are the agent replaying its own prior turns to the same model, which is correct and unaffected.

karolpiotrowicz (commented):
The double-open is fixed — a counter on `httprr.Open` goes from 2 to 1, and the committed cassette is 8 contiguous records to EOF. The `ThoughtSignature` passthrough is fixed and properly pinned: deleting the strip fails `TestNewSummaryEventDropsTheThoughtSignature`. The re-recorded cassette is clean of credentials. One thing left, and it has a trap in it.

The new size assertion cannot catch the defect it was added for. [`llmagent_compaction_test.go:305`](https://github.com/google/adk-go/blob/4c0abe833ec9de214d095094b2130d14d1208042/agent/llmagent/llmagent_compaction_test.go#L305) compares `prompts[len-2]` with `prompts[len-1]`, which on this six-prompt recording is prompts[4] against prompts[5] — both of them after the compaction boundary. And [`promptTextOf`](https://github.com/google/adk-go/blob/4c0abe833ec9de214d095094b2130d14d1208042/agent/llmagent/llmagent_compaction_test.go#L517) counts text and function names only, so a `ThoughtSignature`, being a `[]byte` field, is worth zero characters to it. Applied verbatim to the previous recording — the one carrying the 3,464-byte signature — it does not fire.

The trap is that moving it to the boundary its own comment describes makes it worse, not better. The real boundary is prompts[2] against prompts[3], and there the fixed recording measures 183 to 870, so an assertion relocated there would fail on correct behaviour. Growth across a compaction is expected here: the summary replaces several short turns with one long paragraph. Whatever this assertion should be, it is not a ratio bound on adjacent prompts, and I would rather see it dropped than left where it cannot fail.

karolpiotrowicz (approved):
The size assertion is gone and the reasoning you left in its place is better than the assertion was.

[`llmagent_compaction_test.go:295-306`](https://github.com/google/adk-go/blob/49f31fd88efb35392360feeb1286557663b3332c/agent/llmagent/llmagent_compaction_test.go#L295-L306) now records why there is none: at the real compaction boundary this recording goes 183 characters to 870, because a summary of several short turns is longer than the turns, so an assertion placed there fails on working code — and placed anywhere else it compares two prompts on the same side of the boundary and cannot fail at all. That is exactly the trap, and removing the assertion is the right call rather than relocating it. The doc comment at the top of the test carries the same numbers now instead of the claim it used to make.

`TestCompactionE2E` passes here, and the load-bearing part of it — that the covered turns stop being sent and the summary reaches the prompt — is untouched.

Nothing further from me on this one.

baptmont:
Rebase, and one fix the rebase surfaced.

**`TestCompactionE2E.httprr` was claimed by no `-httprecord` directive.** This file carried a comment asserting the package-level directive matched every cassette in the package — true when written, false now: main has since narrowed it to the FunctionTool, LLMAgent and ToolCallback cassettes. So this cassette could not be re-recorded by `go generate`, and `TestHTTPRecordDirectivesPartitionCassettes` failed after the rebase.

Added a directive scoped to this one cassette and rewrote the comment to say why it exists. The cassette itself is unchanged — the model traffic did not need re-recording, only the directive that claims it.

**Shared context for this round.** The whole stack was rebased onto current `main` (`3f69c505`), absorbing 55 commits. Four conflicts, all from #1252 "give Event a consistent JSON encoding":

- `session/session.go` — main's new JSON tags, `MarshalJSON` and `nilOrRef` merged with our `Compaction` field and the `EventCompaction`/`EventRef` types. One hunk ended mid-function and truncated `nilOrRef`; `gofmt` caught it.
- `internal/telemetry/logger.go` — the only semantic one. Main changed `variantToGenAISystem` to return `*attribute.KeyValue`; ours returned `*log.KeyValue` via a `GenAISystemAttr` helper we had extracted. Resolved by keeping **main's signature** and implementing it through our helper, so there is one source of truth rather than two.
- Imports only in `session/database/service_test.go` and `session/vertexai/service_test.go`.

Every slice was rebuilt from the rebased tree and verified to build, lint and test on its own. The full suite is clean.

karolpiotrowicz (approved):
Re-approving at 6b9000dd to anchor the approval to the current head. The prior one named 49f31fd8, and this branch does not dismiss stale reviews on push, so the badge stayed green while the SHA moved underneath it.

Nothing to re-review: this commit's own content is identical between the two revisions, with no file showing a real change. Everything that moved came from rebasing onto the PRs below it, each of which is approved at its own current head.

baptmont:
Rebased again onto current `main` (`9db1ff12`), which had moved another 15 commits. Content is unchanged from the round you approved, **except one fix that this rebase exposed** — worth a look, since it was not in what you reviewed.

**Startup validation walked the agent loader even when compaction was disabled.** `compactionvalidate.AgainstAgents` enumerated every application before checking whether there was any config to dry-run against. `againstAgents` already returned nil for a nil config, so the walk was pointless work on every server start — and it made a server that never configured compaction depend on its loader being listable.

Main's new `TestNewServerDebugAPIGate` (from #1413) constructs a server with a loader that is a nil pointer behind a non-nil interface. `loader == nil` does not catch that shape, so `ListAgents` panicked and `NewServer` took the process down during construction:

```
agent.(*singleLoader).ListAgents(0x0?)
compactionvalidate.AgainstAgents(0x0, ...)
adkrest.NewServer(...)
```

Fixed by hoisting the nil-config check to the top of both `AgainstAgents` and `AgainstRootAgent`, before either touches the loader or the root agent. `TestValidationDoesNotTouchTheLoaderWhenCompactionIsOff` pins it with a loader that panics if asked to list; removing the guard makes it panic.

One conflict in this rebase, in `server/adkrest/handler.go`: #1413 added a `DebugAPIConfig` field to `ServerConfig` while this stack added `Compaction`. Both additive, both kept.

Every slice was rebuilt from the rebased tree and verified to build, lint and test on its own. Full suite clean.

karolpiotrowicz (dismissed):
Re-approving at fb50537f after the rebase onto main 9db1ff12. Content identical to what I approved at 6b9000dd, with no file carrying a real change.

karolpiotrowicz (approved):

PR #1237, opened 2026-07-31
fix(session/database): format stale-session timestamps as UnixMicro

### LINKED ISSUE #1170: session/database: stale-session error formats timestamps as 1970-01-21 (UnixMicro passed to time.Unix nsec arg)

PratikDhanave:
## Summary

The stale-session error in `session/database/service.go` compares `UnixMicro()` values (microseconds since epoch) but formats them for display with `time.Unix(0, x)`, whose second argument is **nanoseconds**. Being off by 1000×, every real timestamp collapses to `1970-01-2x`, so operators debugging optimistic-concurrency rejections see:

```
stale session error: last update time from request
  (1970-01-21T15:38:42.372056389Z) is older than in database
  (1970-01-21T15:38:42.372500103Z)
```

instead of the actual wall-clock times, making the log lines impossible to correlate with anything else.

The comparison itself is correct — both sides share the same unit, so stale detection still works. **Only the display strings were wrong.**

The fix uses `time.UnixMicro`, which interprets the value as microseconds since epoch.

Fixes #1170

## Testing

- Added `TestDatabaseService_AppendEvent_StaleErrorFormatsTimestamps`, which drives a real optimistic-concurrency conflict (two handles on the same session; one advances the stored `UpdateTime`, the other then appends and is rejected) and asserts the rendered error contains the true `2026-07-17` timestamps and never `1970`.
- Verified the test **fails on the pre-fix code** (renders `1970-01-21T21:08:...`) and **passes with the fix** — a genuine regression guard.
- `go vet ./session/database/` and `go test ./session/database/` pass.


wolo-lab (dismissed):
Thank you. Please resolve conflicts and we will merge your fix.

wolo-lab (approved):

PR #1239, opened 2026-08-02
fix(workflow): make ParallelWorker fail-fast actually stop new dispatch under maxConcurrency

### LINKED ISSUE #1238: ParallelWorker: fail-fast cancellation doesn't stop new items from being dispatched under maxConcurrency
`ParallelWorker.Run` documents that a non-retryable error from one item makes the
workflow "fail fast, cancel other in-flight workers" (`workflow/parallel_worker.go:63-64`).
Under `maxConcurrency > 0` this does not hold: once one item fails, every other item
still waiting on the concurrency semaphore gets dispatched anyway, running its full
work before the error surfaces.

## Repro

`go test -race -run TestParallelWorker_MaxConcurrencyFailFast ./workflow/`, with this
test added locally:

```go
func TestParallelWorker_MaxConcurrencyFailFast(t *testing.T) {
	const nItems = 5
	var started int32

	wrapped := NewFunctionNode("slow_or_fail", func(ctx agent.Context, input int) (int, error) {
		if input == 0 {
			return 0, errors.New("error 0")
		}
		atomic.AddInt32(&started, 1)
		time.Sleep(50 * time.Millisecond)
		return input, nil
	}, defaultNodeConfig)

	pw, _ := NewParallelWorker("parallel", wrapped, 1, defaultNodeConfig) // maxConcurrency=1
	exCtx := agent.NewContext(newMockCtx(t))

	for range pw.Run(exCtx, []any{0, 1, 2, 3, 4}) {
	}
	// started == 4 on current main: every item after the failing one still runs.
}
```

On current `main` (`a8e4606`), `started` is deterministically `4` (all of them),
even though item 0 fails immediately.

## Root cause, as far as I've traced it

Two separate things compound:

1. The dispatch loop's cancellation check (`workflow/parallel_worker.go:113-118`) reads
   `<-ctx.Done()`, where `ctx` is `Run`'s original parameter. The cancel-on-failure path
   calls `cancelFunc()`, which was obtained from `ctx.WithAgentCancel()` and only cancels
   the derived `cancelCtx`/`workerCtx`, not `ctx` itself. So that `select` case can never
   fire from our own fail-fast cancellation, only from an external cancellation of the
   caller's context.
2. Even checking the right context, the result-collection loop that calls `cancelFunc()`
   (`workflow/parallel_worker.go:135-149`) only starts after the dispatch `for` loop above
   it has finished issuing every item, so under `maxConcurrency` the cancellation can't
   reach an item still blocked on the semaphore.

Happy to send a PR restoring the correct fail-fast behavior for the semaphore-gated path.

AmirF194:
Fixes #1238

`ParallelWorker.Run`'s doc comment promises that a non-retryable error makes the
workflow "fail fast, cancel other in-flight workers". Under `maxConcurrency > 0`
that didn't hold: every item still waiting on the concurrency semaphore was
dispatched anyway and ran to completion before the error surfaced, so a hard
failure on item 0 still paid for every remaining item's real work (e.g. an LLM
call) instead of bounding spend.

Two things compounded:

- The dispatch loop's cancellation check read `ctx.Done()`, `Run`'s original
  parameter. `cancelFunc` (from `ctx.WithAgentCancel()`) only cancels the
  derived `cancelCtx`/`workerCtx`, so that `select` case could never fire from
  our own fail-fast path, only from an externally cancelled caller context.
- Even against the right context, `cancelFunc` was only called from the
  result-collection loop, which only starts once the dispatch loop above it
  has finished issuing every item. So cancellation couldn't reach an item
  still blocked on the semaphore no matter which context it checked.

Fix: check `workerCtx.Done()` (the context `cancelFunc` actually cancels), and
have the failing worker call it directly and synchronously, before its
deferred semaphore release runs. That guarantees a slot freed by a failing
item is never reachable by the dispatch loop until cancellation has already
taken effect. Right after acquiring a slot, a second `select` with a `default`
case re-checks `workerCtx.Done()`: without it, the two cases becoming ready at
the same instant (slot freed vs. context cancelled) would leave a plain
two-case `select` to pick between them at random.

## Note on PR #1090

`competing-prs.sh` flags draft PR #1090 ("initial Split") as touching this same
file; it's a large, one-commit, untouched-since-open context-API refactor.
Independently of this fix, it also rewrites the dispatch loop's cancellation
check from `ctx.Done()` to `cancelCtx.Done()` as part of that larger change,
which agrees with the root cause described above. Flagging it here since it's
real overlapping work, not implying anything about how the two should be
sequenced.

## Verification

- `TestParallelWorker_MaxConcurrencyFailFast` (new): `maxConcurrency=1`, item 0
  fails immediately, items 1-4 do real work (a sleep) if dispatched. Fails on
  unmodified `main` (`started == 4`, i.e. every remaining item still ran) and
  passes on this branch (`started` well under that), both confirmed
  deterministic over 20 runs with `-race`. Clean `golang:1.26-bookworm`
  container both sides.
- `go test -race -mod=readonly -count=1 -shuffle=on ./...` at the repo root:
  all packages pass, including `workflow`.
- `go build -mod=readonly ./...`, `go mod tidy -diff` (empty), `gofmt -l` on
  the changed files (empty), and `golangci-lint run ./workflow/...` (0
  issues) all clean.
- Not checked: behavior under the unbounded-concurrency path
  (`maxConcurrency <= 0`) is unchanged and untested here, since all
  `nItems` goroutines are launched immediately in that mode regardless of
  cancellation timing, matching the existing `TestParallelWorker_FailFast`
  coverage.


indurireddy-TF:
Hi @AmirF194 Thanks for the fix ,I verified the fix locally. The fail-fast regression test now passes consistently across 100 runs, and the previously reproduced behavior is no longer occurring.
image

The fix is working as expected from my testing. @baptmont Can you please take a look at this Pr ?

AmirF194:
Thanks for testing this so thoroughly, glad the 100-run check held up.

baptmont:
Hey @AmirF194 you are not currently covered under a CLA, please visit https://cla.developers.google.com/, or check the failed invocation above for more info.

AmirF194:
#1292 fixes the same race (dispatch outrunning fail-fast cancellation under maxConcurrency), opened a bit after this one. The two differ in one detail: this PR moves the failure report ahead of the semaphore release in runWorker (via reportFailure, a sync.Once around cancelFunc) so a newly dispatched worker can't slip in between an error being recorded and cancellation firing, and it adds the same re-check on the no-semaphore path. #1292's re-check covers only the semaphore branch.

This PR is still blocked on signing the Google CLA on my end (see the check above), so happy to stand down if a maintainer would rather move forward with #1292 first.

jjsasha63:
I checked the dispatch-loop change against the merge base: revert workflow/parallel_worker.go
and TestParallelWorker_MaxConcurrencyFailFast fails with started = 4, so the new test does
reproduce #1238.

The collection-loop change drops one error path. With `if res.err != nil` gone,
`reportFailure` is the only thing that sets `firstErr`, and runWorker's retry-delay arm at
parallel_worker.go:214 never calls it. It sends `workerResult{err: ctx.Err()}` straight to
`resCh`. So cancelling a ParallelWorker that has a RetryConfig, while one of its workers is
parked in a retry delay, yields a success event with nil outputs instead of context.Canceled.

My repro: one item, MaxAttempts 5, InitialDelay 2s, a wrapped node that always fails, and the
caller context cancelled 300ms in. This branch gives err=nil and one event, 3 runs out of 3.
main gives context.Canceled and no event.

Routing that arm through reportFailure covers it:

```go
case <-ctx.Done():
    reportFailure(ctx.Err())
    resCh <- workerResult{index: idx, err: ctx.Err()}
    return
```

With that one line on top of your branch, the repro passes 5 out of 5,
TestParallelWorker_FailFast is clean over 2000 runs with -race,
TestParallelWorker_MaxConcurrencyFailFast passes 100 out of 100, and
go test -race -shuffle=on ./workflow/ is green.

Nothing covers cancellation during a retry delay today, which is why CI stayed green here.
It would be worth adding a test for it alongside the fix.

AmirF194:
Good catch. Fixed at 9d85dbd with the one-line reportFailure(ctx.Err()) call you proposed: the exhausted-attempts path already reported before releasing sem, the retry-delay cancellation path just never did.

Reproduced your scenario first (one item, MaxAttempts 5, InitialDelay 2s, always-failing node, context cancelled 300ms in), confirmed the added test fails against the pre-fix code with the same nil-error outcome you described, then applied the fix and confirmed it passes, including 50 shuffled -race reruns of the full ParallelWorker suite.

jjsasha63:
I checked 9d85dbd: my probe passes 5 out of 5, TestParallelWorker_FailFast is clean over
2000 runs with -race, your new test passes 100 out of 100, and go test -race -shuffle=on
./workflow/ is green.

I also deleted the `reportFailure(ctx.Err())` line and reran
TestParallelWorker_CancelDuringRetryDelay on its own. It fails with "expected error on
cancellation during retry delay, got nil", so the test pins the fix rather than passing
alongside it.

One small thing on the comment above that test:

```go
// jjsasha63's finding on #1239: runWorker's retry-delay select only ...
```

A GitHub handle in source outlives the context that makes it meaningful, and the PR thread
already carries the attribution. The mechanism explanation underneath it is the part worth
keeping. Happy either way.

Nothing else outstanding from me. The branch is behind main, so it will want a rebase before
merge.

jjsasha63 (commented):
LGTM

jjsasha63 (approved):
LGTM

PR #1240, opened 2026-08-02
fix: validate deploy CLI values embedded in generated Dockerfile content

herdiyana256:
Values from `--entry_point_path`, `--a2a_agent_url`, and the derived executable name are interpolated directly into generated Dockerfile content via string concatenation, with no escaping.

In `cloudrun.go` these values only reach `COPY` paths and a `CMD` JSON-array string, so rejecting quotes/backticks/newlines/NUL is enough. In `agentengine.go`, `--entry_point_path` and the executable name are also embedded into a `RUN` instruction, which always runs in shell form (`/bin/sh -c "..."`), so shell metacharacters like `;`, `|`, and `&` are dangerous there even without any quote or newline breakout.

Adds `ValidateDockerfileSafe` (blocklist, for COPY/CMD-array contexts) and a stricter `ValidateShellArgSafe` (allowlist, for values reaching a RUN line), and applies the appropriate one at each call site. `execFile` is upgraded to the stricter check in both deploy paths for consistency, since it's always meant to be a bare filename.

PR #1241, opened 2026-08-02
fix: add optional OIDC verification for Pub/Sub and Eventarc HTTP triggers

herdiyana256:
`PubSubTriggerHandler` and `EventarcTriggerHandler` (`server/adkrest/controllers/triggers`) run an agent against the request body with no authentication of their own. Anyone who can reach `/apps/{app_name}/trigger/pubsub` or `/apps/{app_name}/trigger/eventarc` can trigger an arbitrary agent run with attacker-controlled content. This relies entirely on the deployment platform (for example Cloud Run configured to require IAM authentication) to restrict access, which the library itself has no way to verify and which doesn't hold for every context these launchers can be run in.

## What this adds

One new opt-in field on `TriggerConfig`, `OIDC *OIDCConfig`, holding two settings, with a flag for each on both the `pubsub` and `eventarc` sublaunchers:

- **`ExpectedAudience`** / `-trigger_oidc_audience`. Requests must carry a Google-signed OIDC bearer token (the kind Pub/Sub push subscriptions and Eventarc triggers already attach when configured with a service account) whose audience matches and whose issuer is Google, verified via `idtoken.Validate`. Requests without one are rejected with 401 before the agent is invoked.
- **`AllowedServiceAccounts`** / `-trigger_oidc_service_accounts`. The verified token must additionally carry a verified `email` claim matching one of the listed service accounts, the identity your subscription or trigger actually delivers as. Unlisted principals get 403.

The second one is why the first isn't sufficient on its own. `idtoken.Validate` checks the Google signature, the expiry and the audience string, and deliberately leaves every identity claim to the caller. The audience is chosen freely by whoever mints the token, so any principal that can call `iam.serviceAccounts.getOpenIdToken` on a service account of its own can obtain a Google-signed token for your audience. Verifying the audience alone establishes "somebody holds a Google-signed token for this string", not "my subscription is calling". Tokens carrying no `email` claim at all are rejected rather than skipped, since `getOpenIdToken` defaults `includeEmail` to false.

This mirrors what landed in adk-python as `GoogleOidcVerifier(audience, allowed_emails)` (google/adk-python#6629, commit `73d9fe0`), so the two implementations stay aligned; the flag names match its `--trigger_oidc_audience` and `--trigger_oidc_service_accounts`. The Go side additionally checks the issuer, and rejects a non-boolean `email_verified` rather than accepting it by truthiness.

`adkgo deploy cloud-run` forwards both settings to whichever sublaunchers are enabled: `--pubsub_oidc_audience`, `--pubsub_oidc_service_accounts`, `--eventarc_oidc_audience`, `--eventarc_oidc_service_accounts`. That path already deploys with `--no-allow-unauthenticated`, so this is the layer underneath it rather than the only one.

## Compatibility

Left unset, behavior is unchanged and the struct change is API-compatible: `OIDC` is a pointer, so `TriggerConfig` stays comparable and `apidiff` against `main` reports no incompatible changes. Both launchers print a startup warning when no audience is set, and a second one when an audience is set without an allow-list, so the remaining gap is visible rather than implied. Those warnings read the `TriggerConfig` that was actually built, not the flag strings.

## Notes

- Verification is switched off by leaving `TriggerConfig.OIDC` nil. A non-nil block with an empty `ExpectedAudience` reads as "pin these principals" while verifying nothing, so both `*WithConfig` constructors reject it, and the handler denies on it directly rather than falling through to the unverified path.
- The 401/403 body is one fixed string; the underlying verification error is logged server-side, so an anonymous caller can't distinguish expired from audience-mismatch from bad-signature, nor a rejected principal from an unverifiable token. 401 responses carry `WWW-Authenticate`.
- The `Bearer` scheme is matched case-insensitively, per RFC 9110. More than one `Authorization` header is rejected rather than resolved, so a proxy in front that forwards the last value can't disagree with this handler about which credential arrived.
- The `OIDCConfig` and its allow-list are copied at construction, so a caller that mutates its own copy afterwards doesn't change what a running handler enforces.
- Token verification is a field on `RetriableRunner` (defaulted to `idtoken.Validate`) rather than a package-level variable, so tests can install a fake without a live network call to Google's certificate endpoint and without racing on a global.
- The audience a Pub/Sub push subscription sends is whatever `--push-auth-token-audience` was set to, defaulting to the full push endpoint URL. The flag help says so, since setting the bare service root instead is an easy way to 401 every delivery.

## Open question for a maintainer

With only `-trigger_oidc_audience` set, the endpoint still accepts any principal that can mint a Google-signed token for that audience. Whether the allow-list should be *required* rather than recommended is a policy call I've left open: `getOpenIdToken` defaults `includeEmail` to false, so mandating the pin would turn a documented weak mode into a 403 on every legitimate delivery for anyone whose token carries no email claim — and on Pub/Sub push a 403 is a NACK, so that's a redelivery loop rather than a clean refusal. Audience-only is still worth something on its own, since it stops a token minted for another service being spent here.

### DIFF
--- a/cmd/launcher/web/triggers/pubsub/pubsub.go
+++ b/cmd/launcher/web/triggers/pubsub/pubsub.go
@@ -36,6 +36,7 @@ type pubsubConfig struct {
 	triggerBaseDelay  time.Duration
 	triggerMaxDelay   time.Duration
 	triggerMaxRuns    int
+	oidcAudience      string
 }
 
 type pubsubLauncher struct {
@@ -53,6 +54,9 @@ func NewLauncher() web.Sublauncher {
 	fs.DurationVar(&config.triggerBaseDelay, "trigger_base_delay", 1*time.Second, "Base delay for trigger retry exponential backoff")
 	fs.DurationVar(&config.triggerMaxDelay, "trigger_max_delay", 10*time.Second, "Maximum delay for trigger retry exponential backoff")
 	fs.IntVar(&config.triggerMaxRuns, "trigger_max_concurrent_runs", 100, "Maximum concurrent trigger runs")
+	fs.StringVar(&config.oidcAudience, "oidc_audience", "", "Expected audience of the Google-signed OIDC bearer token attached by Pub/Sub push subscriptions. "+
+		"If set, requests without a valid token for this audience are rejected with 401. If unset, this endpoint performs no authentication of its own and "+
+		"relies entirely on the deployment platform (for example Cloud Run configured to require IAM authentication) to restrict who can reach it.")
 
 	return &pubsubLauncher{
 		config: config,
@@ -110,6 +114,7 @@ func (p *pubsubLauncher) SetupSubrouters(router *mux.Router, config *launcher.Co
 		BaseDelay:         p.config.triggerBaseDelay,
 		MaxDelay:          p.config.triggerMaxDelay,
 		MaxConcurrentRuns: p.config.triggerMaxRuns,
+		ExpectedAudience:  p.config.oidcAudience,
 	}
 
 	controller := triggers.NewPubSubController(
@@ -133,4 +138,7 @@ func (p *pubsubLauncher) SetupSubrouters(router *mux.Router, config *launcher.Co
 // UserMessage implements web.Sublauncher.
 func (p *pubsubLauncher) UserMessage(webURL string, printer func(v ...any)) {
 	printer(fmt.Sprintf("       pubsub:  PubSub trigger endpoint is available at %s%s/apps/{app_name}/trigger/pubsub", webURL, p.config.pathPrefix))
+	if p.config.oidcAudience == "" {
+		printer("       pubsub:  WARNING: -oidc_audience is not set; this endpoint accepts unauthenticated requests unless the deployment platform restricts access on its own.")
+	}
 }
--- a/server/adkrest/controllers/triggers/config.go
+++ b/server/adkrest/controllers/triggers/config.go
@@ -26,4 +26,16 @@ type TriggerConfig struct {
 	MaxDelay time.Duration
 	// MaxConcurrentRuns is the maximum number of concurrent runs.
 	MaxConcurrentRuns int
+	// ExpectedAudience, when non-empty, requires every request to carry a
+	// Google-signed OIDC bearer token (as attached by Pub/Sub push
+	// subscriptions and Eventarc triggers configured with a service account)
+	// whose audience claim matches this value; requests without one are
+	// rejected with 401 before the agent is invoked.
+	//
+	// This endpoint accepts arbitrary attacker-controlled content as agent
+	// input and otherwise has no authentication of its own: on platforms
+	// that don't already gate the endpoint with their own IAM check (or if
+	// that gate is ever misconfigured), leaving this unset means anyone who
+	// can reach the endpoint can trigger arbitrary agent runs.
+	ExpectedAudience string
 }
--- a/server/adkrest/controllers/triggers/triggers.go
+++ b/server/adkrest/controllers/triggers/triggers.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/gorilla/mux"
+	"google.golang.org/api/idtoken"
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/v2/agent"
@@ -137,6 +138,38 @@ func calculateBackoff(attempt int, base, maxDelay time.Duration) time.Duration {
 	delay := min(time.Duration(backoff), maxDelay)
 	jitter := time.Duration(rand.Float64() * float64(delay) * 0.5)
 	return delay + jitter
+}
+
+// validateIDToken verifies a Google-signed OIDC ID token against an expected
+// audience. It is a package variable (rather than a direct call to
+// idtoken.Validate) so tests in this package can substitute a fake verifier
+// without making a real network call to Google's certificate endpoint.
+var validateIDToken = idtoken.Validate
+
+// verifyPushRequestAuth requires a valid Google-signed OIDC bearer token when
+// expectedAudience is non-empty; it is a no-op when expectedAudience is empty,
+// preserving prior behavior for deployments that rely entirely on
+// platform-level access control (for example Cloud Run configured to require
+// IAM authentication) in front of this endpoint.
+//
+// Pub/Sub push subscriptions and Eventarc Cloud Run triggers configured with
+// a service account both attach exactly this kind of token, so this does not
+// require any change to how the trigger is configured on the Google Cloud
+// side — only that ExpectedAudience is set to match.
+func verifyPushRequestAuth(r *http.Request, expectedAudience string) error {
+	if expectedAudience == "" {
+		return nil
+	}
+	const bearerPrefix = "Bearer "
+	authHeader := r.Header.Get("Authorization")
+	if !strings.HasPrefix(authHeader, bearerPrefix) {
+		return fmt.Errorf("missing bearer token")
+	}
+	token := strings.TrimPrefix(authHeader, bearerPrefix)
+	if _, err := validateIDToken(r.Context(), token, expectedAudience); err != nil {
+		return fmt.Errorf("invalid identity token: %w", err)
+	}
+	return nil
 }
 
 // Resolve the target app name from the request.
--- a/cmd/launcher/web/triggers/eventarc/eventarc.go
+++ b/cmd/launcher/web/triggers/eventarc/eventarc.go
@@ -36,6 +36,7 @@ type eventarcConfig struct {
 	triggerBaseDelay  time.Duration
 	triggerMaxDelay   time.Duration
 	triggerMaxRuns    int
+	oidcAudience      string
 }
 
 type eventarcLauncher struct {
@@ -53,6 +54,9 @@ func NewLauncher() web.Sublauncher {
 	fs.DurationVar(&config.triggerBaseDelay, "trigger_base_delay", 1*time.Second, "Base delay for trigger retry exponential backoff")
 	fs.DurationVar(&config.triggerMaxDelay, "trigger_max_delay", 10*time.Second, "Maximum delay for trigger retry exponential backoff")
 	fs.IntVar(&config.triggerMaxRuns, "trigger_max_concurrent_runs", 100, "Maximum concurrent trigger runs")
+	fs.StringVar(&config.oidcAudience, "oidc_audience", "", "Expected audience of the Google-signed OIDC bearer token attached by Eventarc triggers configured with a service account. "+
+		"If set, requests without a valid token for this audience are rejected with 401. If unset, this endpoint performs no authentication of its own and "+
+		"relies entirely on the deployment platform (for example Cloud Run configured to require IAM authentication) to restrict who can reach it.")
 
 	return &eventarcLauncher{
 		config: config,
@@ -110,6 +114,7 @@ func (e *eventarcLauncher) SetupSubrouters(router *mux.Router, config *launcher.
 		BaseDelay:         e.config.triggerBaseDelay,
 		MaxDelay:          e.config.triggerMaxDelay,
 		MaxConcurrentRuns: e.config.triggerMaxRuns,
+		ExpectedAudience:  e.config.oidcAudience,
 	}
 
 	controller := triggers.NewEventarcController(
@@ -133,4 +138,7 @@ func (e *eventarcLauncher) SetupSubrouters(router *mux.Router, config *launcher.
 // UserMessage implements web.Sublauncher.
 func (e *eventarcLauncher) UserMessage(webURL string, printer func(v ...any)) {
 	printer(fmt.Sprintf("       eventarc: Eventarc trigger endpoint is available at %s%s/apps/{app_name}/trigger/eventarc", webURL, e.config.pathPrefix))
+	if e.config.oidcAudience == "" {
+		printer("       eventarc: WARNING: -oidc_audience is not set; this endpoint accepts unauthenticated requests unless the deployment platform restricts access on its own.")
+	}
 }
--- a/server/adkrest/controllers/triggers/eventarc.go
+++ b/server/adkrest/controllers/triggers/eventarc.go
@@ -53,6 +53,11 @@ func NewEventarcController(sessionService session.Service, agentLoader agent.Loa
 
 // EventarcTriggerHandler handles the Eventarc trigger endpoint.
 func (c *EventarcController) EventarcTriggerHandler(w http.ResponseWriter, r *http.Request) {
+	if err := verifyPushRequestAuth(r, c.runner.triggerConfig.ExpectedAudience); err != nil {
+		respondError(w, http.StatusUnauthorized, fmt.Sprintf("authentication failed: %v", err))
+		return
+	}
+
 	var event models.EventarcTriggerRequest
 	contentType := r.Header.Get("Content-Type")
 	// The HTTP Content-Type header MUST be set to the media type of an event format for structured mode.
--- a/server/adkrest/controllers/triggers/eventarc_test.go
+++ b/server/adkrest/controllers/triggers/eventarc_test.go
@@ -202,3 +202,46 @@ func TestEventarcTriggerHandler(t *testing.T) {
 		})
 	}
 }
+
+// TestEventarcTriggerHandler_RequiresAuthWhenAudienceConfigured mirrors
+// TestPubSubTriggerHandler_RequiresAuthWhenAudienceConfigured: once an
+// operator opts in to OIDC verification via ExpectedAudience, a request with
+// no bearer token must be rejected before the agent is invoked.
+func TestEventarcTriggerHandler_RequiresAuthWhenAudienceConfigured(t *testing.T) {
+	mockAgentRunCount := 0
+	testAgent, err := agent.New(agent.Config{
+		Name: "test-agent",
+		Run: func(ctx agent.InvocationContext) iter.Seq2[*session.Event, error] {
+			return func(yield func(*session.Event, error) bool) {
+				mockAgentRunCount++
+				yield(&session.Event{ID: "success-event"}, nil)
+			}
+		},
+	})
+	if err != nil {
+		t.Fatalf("agent.New failed: %v", err)
+	}
+
+	sessionService := &fakes.FakeSessionService{Sessions: make(map[fakes.SessionKey]fakes.TestSession)}
+	agentLoader := agent.NewSingleLoader(testAgent)
+	config := defaultTriggerConfig
+	config.ExpectedAudience = "https://example-agent.example.com"
+	controller := triggers.NewEventarcController(sessionService, agentLoader, nil, nil, runner.PluginConfig{}, config)
+
+	req, err := http.NewRequest(http.MethodPost, "/apps/test-agent/triggers/eventarc", bytes.NewBuffer([]byte(`{}`)))
+	if err != nil {
+		t.Fatalf("new request: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/cloudevents+json")
+	req = mux.SetURLVars(req, map[string]string{"app_name": "test-agent"})
+	rr := httptest.NewRecorder()
+
+	controller.EventarcTriggerHandler(rr, req)
+
+	if rr.Code != http.StatusUnauthorized {
+		t.Errorf("expected status %d for a request with no bearer token, got %d. Body: %s", http.StatusUnauthorized, rr.Code, rr.Body.String())
+	}
+	if mockAgentRunCount != 0 {
+		t.Errorf("expected the agent not to run for an unauthenticated request, got %d run attempts", mockAgentRunCount)
+	}
+}
--- a/server/adkrest/controllers/triggers/pubsub.go
+++ b/server/adkrest/controllers/triggers/pubsub.go
@@ -52,6 +52,11 @@ func NewPubSubController(sessionService session.Service, agentLoader agent.Loade
 
 // PubSubTriggerHandler handles the PubSub trigger endpoint.
 func (c *PubSubController) PubSubTriggerHandler(w http.ResponseWriter, r *http.Request) {
+	if err := verifyPushRequestAuth(r, c.runner.triggerConfig.ExpectedAudience); err != nil {
+		respondError(w, http.StatusUnauthorized, fmt.Sprintf("authentication failed: %v", err))
+		return
+	}
+
 	// Parse the request to the request model.
 	var req models.PubSubTriggerRequest
 	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
--- a/server/adkrest/controllers/triggers/pubsub_test.go
+++ b/server/adkrest/controllers/triggers/pubsub_test.go
@@ -131,6 +131,49 @@ func setupTest(t *testing.T, a agent.Agent) *triggers.PubSubController {
 	return triggers.NewPubSubController(sessionService, agentLoader, nil, nil, runner.PluginConfig{}, defaultTriggerConfig)
 }
 
+// TestPubSubTriggerHandler_RequiresAuthWhenAudienceConfigured guards against
+// the endpoint accepting unauthenticated requests once an operator has opted
+// in to OIDC verification via ExpectedAudience: without a bearer token at all
+// (the realistic "reachable from the open internet" case) the request must be
+// rejected before the agent is ever invoked.
+func TestPubSubTriggerHandler_RequiresAuthWhenAudienceConfigured(t *testing.T) {
+	mockAgentRunCount := 0
+	testAgent := createMockAgent(t, nil, &mockAgentRunCount, nil)
+
+	sessionService := &fakes.FakeSessionService{Sessions: make(map[fakes.SessionKey]fakes.TestSession)}
+	agentLoader := agent.NewSingleLoader(testAgent)
+	config := defaultTriggerConfig
+	config.ExpectedAudience = "https://example-agent.example.com"
+	apiController := triggers.NewPubSubController(sessionService, agentLoader, nil, nil, runner.PluginConfig{}, config)
+
+	reqObj := models.PubSubTriggerRequest{
+		Message: models.PubSubMessage{
+			Data: []byte(base64.StdEncoding.EncodeToString([]byte("Hello agent"))),
+		},
+		Subscription: "test-sub",
+	}
+	reqBytes, err := json.Marshal(reqObj)
+	if err != nil {
+		t.Fatalf("marshal request: %v", err)
+	}
+
+	req, err := http.NewRequest(http.MethodPost, "/apps/test-agent/triggers/pubsub", bytes.NewBuffer(reqBytes))
+	if err != nil {
+		t.Fatalf("new request: %v", err)
+	}
+	req = mux.SetURLVars(req, map[string]string{"app_name": "test-agent"})
+	rr := httptest.NewRecorder()
+
+	apiController.PubSubTriggerHandler(rr, req)
+
+	if rr.Code != http.StatusUnauthorized {
+		t.Errorf("expected status %d for a request with no bearer token, got %d. Body: %s", http.StatusUnauthorized, rr.Code, rr.Body.String())
+	}
+	if mockAgentRunCount != 0 {
+		t.Errorf("expected the agent not to run for an unauthenticated request, got %d run attempts", mockAgentRunCount)
+	}
+}
+
 func createMockAgent(t *testing.T, results []error, runCount *int, expectedAttributes map[string]string) agent.Agent {
 	t.Helper()
 	testAgent, err := agent.New(agent.Config{
--- a/server/adkrest/controllers/triggers/triggers_internal_test.go
+++ b/server/adkrest/controllers/triggers/triggers_internal_test.go
@@ -1,0 +1,102 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package triggers
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"google.golang.org/api/idtoken"
+)
+
+func TestVerifyPushRequestAuth(t *testing.T) {
+	const audience = "https://example-agent.example.com"
+
+	tests := []struct {
+		name             string
+		expectedAudience string
+		authHeader       string
+		validate         func(ctx context.Context, idToken, audience string) (*idtoken.Payload, error)
+		wantErr          bool
+	}{
+		{
+			name:             "audience not configured: no-op regardless of header",
+			expectedAudience: "",
+			authHeader:       "",
+			wantErr:          false,
+		},
+		{
+			name:             "audience configured, no Authorization header",
+			expectedAudience: audience,
+			authHeader:       "",
+			wantErr:          true,
+		},
+		{
+			name:             "audience configured, non-Bearer scheme",
+			expectedAudience: audience,
+			authHeader:       "Basic dXNlcjpwYXNz",
+			wantErr:          true,
+		},
+		{
+			name:             "audience configured, valid bearer token",
+			expectedAudience: audience,
+			authHeader:       "Bearer valid-token",
+			validate: func(ctx context.Context, idToken, aud string) (*idtoken.Payload, error) {
+				if idToken != "valid-token" || aud != audience {
+					t.Errorf("validate called with (%q, %q), want (%q, %q)", idToken, aud, "valid-token", audience)
+				}
+				return &idtoken.Payload{Audience: aud}, nil
+			},
+			wantErr: false,
+		},
+		{
+			name:             "audience configured, token fails verification",
+			expectedAudience: audience,
+			authHeader:       "Bearer forged-token",
+			validate: func(ctx context.Context, idToken, aud string) (*idtoken.Payload, error) {
+				return nil, errors.New("idtoken: invalid token")
+			},
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			original := validateIDToken
+			defer func() { validateIDToken = original }()
+			if tt.validate != nil {
+				validateIDToken = tt.validate
+			} else {
+				validateIDToken = func(context.Context, string, string) (*idtoken.Payload, error) {
+					t.Fatal("validateIDToken should not be called")
+					return nil, nil
+				}
+			}
+
+			req := httptest.NewRequest(http.MethodPost, "/apps/test-agent/trigger/pubsub", nil)
+			if tt.authHeader != "" {
+				req.Header.Set("Authorization", tt.authHeader)
+			}
+
+			err := verifyPushRequestAuth(req, tt.expectedAudience)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("verifyPushRequestAuth() error = %v, wantErr %v", err, tt.wantErr)
+			}
+		})
+	}
+}


herdiyana256:
Following up on this one. The equivalent fix for adk-python is open at google/adk-python#6629 and is also still waiting, and I'd rather the two didn't drift apart before either of them lands.

@anFatum you built both subrouters, so you'll have a better read than me on whether `-oidc_audience` sits at the right level here, or whether audience config belongs further up in the launcher rather than on the handler.

@baptmont @kdroste-google adding you in case `server/adkrest/controllers/triggers` falls to whoever is on rotation.

One thing worth stressing since it's a behavior change on a request path: it's off unless you set it. Anyone not passing the flag gets exactly what they get today, plus one warning line at startup.

karolpiotrowicz (dismissed):
The direction here is right and the endpoints do need this. But as written the check does not authenticate the caller, and the flag help and struct comment tell an operator that it does. That combination is what needs fixing before this lands, and the fix is small.

## The check verifies the audience and then discards the identity

[`triggers.go:169`](https://github.com/google/adk-go/blob/80283525af949f834b92d7a6abe82156a1d7c856/server/adkrest/controllers/triggers/triggers.go#L169) is `if _, err := validateIDToken(r.Context(), token, expectedAudience); err != nil`. The `*idtoken.Payload` goes to the blank identifier. In `google.golang.org/api@v0.279.0`, `idtoken.Validate` checks three things: `aud` string equality, `exp`, and the Google signature. It parses `iss`, `sub`, `email` and `email_verified` into the payload and deliberately leaves them to the caller — the doc comment says so in as many words. Nothing in this change looks at any of them.

So the property enforced is "somebody holds an unexpired Google-signed token whose `aud` equals this string". The audience is chosen by whoever mints the token, not by the recipient: `generateIdToken` takes `audience` as a required free-form string and authorizes on `iam.serviceAccounts.getOpenIdToken` against the *caller's own* service account. Your own [`auth/providers.go:132-147`](https://github.com/google/adk-go/blob/80283525af949f834b92d7a6abe82156a1d7c856/auth/providers.go#L132-L147) mints arbitrary-audience tokens with the same library.

I drove this end to end rather than reasoning about it: a real `gorilla/mux` router registered as [`pubsub.go:134`](https://github.com/google/adk-go/blob/80283525af949f834b92d7a6abe82156a1d7c856/cmd/launcher/web/triggers/pubsub/pubsub.go#L134) does, served over a real socket, `ExpectedAudience` set, real `idtoken.Validate` with a live certificate fetch, and an `Authorization` header carrying a genuine Google-signed token from an ordinary personal Google account unrelated to any Pub/Sub subscription:

```
status=200  body={"status":"success"}  agent runs=1
```

The agent ran on the caller's payload. Google's Pub/Sub push-authentication guidance covers exactly this: its Go sample, using this library, checks `payload.Issuer` and then says "IMPORTANT: you should validate claim details not covered by signature and audience verification above ... Ensure that `payload.Claims["email"]` is equal to the expected service account ... Ensure that `payload.Claims["email_verified"]` is set to true."

One limit, because it bounds the urgency. I showed a token from an unrelated principal is accepted. I did **not** show the other half — that an outsider can get a Google signature over *your* chosen audience — since I had no service account to mint with and three attempts failed on that. That half rests on the API reference, not on a run I can show you. One service account and one `generateIdToken` call would settle it, and it is worth doing before this is discussed publicly.

The fix is about ten lines and I ran it: capture the payload, require `iss` to be `accounts.google.com` or `https://accounts.google.com`, require `email_verified`, and require `email` to match a configured allow-list of the service accounts the trigger uses. The same request then gets 401 with the agent not running, and the package and launcher suites stay green. Two things only surfaced by executing it. `triggers_internal_test.go` stops compiling against the new signature, and its success fixture returns a payload with no `Issuer` or `email`, so it needs updating in the same change or the fix goes red. And `generateIdToken`'s `includeEmail` defaults to false, so a token can legitimately carry no `email` claim — the check has to reject a token lacking it, not just compare when present.

## The documentation asserts the stronger guarantee

This is the half I would fix even if the code stayed as it is, because it turns a partial control into a misleading one.

[`config.go:29-33`](https://github.com/google/adk-go/blob/80283525af949f834b92d7a6abe82156a1d7c856/server/adkrest/controllers/triggers/config.go#L29-L33) describes the required token as the one "attached by Pub/Sub push subscriptions and Eventarc triggers configured with a service account". [`triggers.go:155-158`](https://github.com/google/adk-go/blob/80283525af949f834b92d7a6abe82156a1d7c856/server/adkrest/controllers/triggers/triggers.go#L155-L158) goes further: "this does not require any change to how the trigger is configured on the Google Cloud side — only that ExpectedAudience is set to match." Both flag help strings say the same thing. An operator reads that as "only my subscription gets in", and the sentence about no further configuration being needed is what stops them asking which identity is checked. The `config.go` comment is candid about the *unset* case, which makes the silence about the set case stand out more, not less.

## Smaller things

- No handler-level test sends an `Authorization` header, so the accept path through the public handler is untested. I measured what that costs: making the handler return 401 for *every* request whenever the feature is enabled leaves the whole suite green, as does an early `return nil` for an empty token after [`triggers.go:168`](https://github.com/google/adk-go/blob/80283525af949f834b92d7a6abe82156a1d7c856/server/adkrest/controllers/triggers/triggers.go#L168). A change that bricks the feature for every real user is currently invisible.
- Deleting the `ExpectedAudience:` line from either launcher's `TriggerConfig` literal — [`pubsub.go:117`](https://github.com/google/adk-go/blob/80283525af949f834b92d7a6abe82156a1d7c856/cmd/launcher/web/triggers/pubsub/pubsub.go#L117) or [`eventarc.go:117`](https://github.com/google/adk-go/blob/80283525af949f834b92d7a6abe82156a1d7c856/cmd/launcher/web/triggers/eventarc/eventarc.go#L117) — also leaves the suite green. Not really yours: the existing `TestSetupSubrouters` asserts a route match and no config field, so the same deletion on `MaxRetries` survives identically. Worth knowing because the startup warning reads the flag field rather than what the controller received, so broken wiring would suppress the warning and look correct.
- [`pubsub.go:56`](https://github.com/google/adk-go/blob/80283525af949f834b92d7a6abe82156a1d7c856/server/adkrest/controllers/triggers/pubsub.go#L56) and [`eventarc.go:57`](https://github.com/google/adk-go/blob/80283525af949f834b92d7a6abe82156a1d7c856/server/adkrest/controllers/triggers/eventarc.go#L57) put the wrapped `idtoken` error in the 401 body, so an anonymous caller can tell expired from audience-mismatch from bad-signature. A fixed body plus a server-side log would close that.
- A transport failure reaching Google's certificate endpoint also surfaces as 401 rather than 503, so an egress problem rejects legitimate deliveries and reads as an auth failure in the logs.
- [`triggers.go:165`](https://github.com/google/adk-go/blob/80283525af949f834b92d7a6abe82156a1d7c856/server/adkrest/controllers/triggers/triggers.go#L165) matches `"Bearer "` byte-exactly, and RFC 9110 makes the scheme case-insensitive. It fails closed so nothing gets in, but a lowercase `bearer` is rejected with "missing bearer token", which misdescribes what happened.
- The new 401 carries no `WWW-Authenticate` header. These endpoints never returned 401 before, so that is new here.
- Neither the help text nor the comment says what value the audience should be. A Pub/Sub push subscription's `aud` is whatever `--push-auth-token-audience` was set to, and defaults to the full push endpoint URL — while both test fixtures use a bare service root. Someone will set the service URL and get 401 on every delivery.
- [`triggers.go:147`](https://github.com/google/adk-go/blob/80283525af949f834b92d7a6abe82156a1d7c856/server/adkrest/controllers/triggers/triggers.go#L147) is a writable package-level variable on the auth path. It is not a race today — nothing in the package starts a goroutine or calls `t.Parallel()` — but the first test that adds `t.Parallel()` makes it one, and the loser would silently run with the other test's fake verifier. A field on `RetriableRunner`, defaulted in the constructors, gives the same test seam without the global.

## What held up

The ordering is right and should not change: the check is the first statement in both handlers, ahead of the body read, the route lookup, the semaphore and the agent, and both new tests assert the agent did not run rather than only checking the status. With the flag unset the function returns before touching a header, so existing deployments are unaffected, the struct change is API-compatible, and both production construction sites are updated. I looked for a way past the check by omitting or malforming the header and found none — absent, empty, `Basic`, bare `Bearer` and duplicate headers all fail closed. Nor can a caller steer verification to a key they control: the certificate endpoints are hardcoded Google URLs and no `jku`/`jwk`/`x5u` header is honoured.

Two notes on the branch rather than the code. No Go build, test or lint job ran on this commit — only `check-changes` and the CLA check — so CI has not exercised any of this. And the branch is behind `main`, though the merge is clean and I built and tested the merge result.

karolpiotrowicz:

It adds optional OIDC verification for the Pub/Sub and Eventarc trigger endpoints, which brings a new exported field on `TriggerConfig` and a dependency on `google.golang.org/api/idtoken`. adk-python has the same flag but pairs it with a caller-identity allow-list this PR omits. Two things to decide: whether we take the field and the dependency, and whether audience-only verification is the bar we want.

herdiyana256:
Thanks. The identity point is correct and I've fixed it rather than argued it. Pushed as `877a63c`, rebased onto current `main`.

## The check now pins the caller

`verifyPushRequestAuth` keeps the `*idtoken.Payload` instead of discarding it, and:

- requires `iss` to be `accounts.google.com` or `https://accounts.google.com`, whenever verification is enabled;
- requires `email_verified` and an `email` matching a new `AllowedServiceAccounts` allow-list, when one is configured (403, not 401, since the credential verified but the principal isn't one this endpoint accepts);
- rejects a token carrying **no** `email` claim rather than skipping the comparison. You were right that `generateIdToken` defaults `includeEmail` to false, so "absent" is a case that actually occurs.

New `-oidc_service_accounts` flag on both sublaunchers (comma-separated), and `Parse` rejects it without `-oidc_audience`.

On the scope question @karolpiotrowicz raised for @wolo-lab: this is now the same shape as adk-python, which is not a coincidence. `GoogleOidcVerifier(audience, allowed_emails)` landed there in `73d9fe0` (via #6629), checking `email_verified` and an allow-list. `-oidc_service_accounts` mirrors `--trigger_oidc_service_accounts`. So the two repos no longer drift, and the API question is one that's already been answered once. The Go side is slightly stricter in that it also checks the issuer.

I did **not** attempt the half you flagged as unverified: minting a Google signature over a chosen audience from an unrelated service account. I have no service account to test it with either. The fix doesn't depend on that half being true: the allow-list closes it regardless of whether the audience can be borrowed.

## The documentation asserted the stronger guarantee

Fixed, and I agree this was the worse half. The "no change to how the trigger is configured on the Google Cloud side" sentence is gone, because it isn't true once you pin identity, since the token has to carry an email claim. `TriggerConfig` and both flag help strings now say plainly that the audience is chosen by whoever mints the token and that verifying it alone does not identify the caller. Both launchers warn when an audience is set without an allow-list, in addition to the existing unset warning.

The flag help now also says what the audience should be: the subscription's `--push-auth-token-audience`, defaulting to the full push endpoint URL. That one would have cost someone a debugging afternoon.

## The measurements

Your two mutants are the reason the tests changed shape, so I ran them as tests rather than trusting the assertion:

| mutation | before | now |
|---|---|---|
| handler returns 401 for every request when enabled | green | **FAIL** |
| `return nil` for an empty token | green | **FAIL** |
| allow-list check removed | n/a | **FAIL** |
| issuer check removed | n/a | **FAIL** |
| drop `ExpectedAudience:` from `pubsub.go` `TriggerConfig` literal | green | **FAIL** |
| drop it from `eventarc.go` | green | **FAIL** |

The handler tests now go through a real `mux.Router` with an `Authorization` header and assert both the status and the agent run count, so the accept path is covered. The launcher test sends an unauthenticated request after `SetupSubrouters` and expects 401. That catches broken wiring end-to-end without asserting on a config field, which is why the same test now exists for `eventarc` too (it had no test file at all).

`triggers_internal_test.go` was updated in the same commit, and its success fixture now returns a realistic payload with an issuer and a verified email. It did need both, as you said.

## Smaller items

- **Error detail in the body**: fixed. The 401/403 body is now fixed text; the wrapped `idtoken` error goes to the server log only.
- **`WWW-Authenticate`**: added on 401.
- **Case-sensitive `Bearer`**: fixed, matched case-insensitively per RFC 9110.
- **Package-level `validateIDToken`**: now a field on `RetriableRunner`, defaulted to `idtoken.Validate` when nil. The new tests use `t.Parallel()` throughout and pass under `-race`, which wasn't possible before.

Two I deliberately left:

- **Transport failure to Google's cert endpoint surfaces as 401, not 503.** Real, and worth doing, but distinguishing it means classifying errors out of `idtoken.Validate`, which doesn't expose a typed transport error, so it'd be string matching on the wrapped error. That seems like it belongs in a follow-up rather than smuggled in here.
- **`TestSetupSubrouters` asserting a route match and no config.** Left as you found it, since the same gap covers `MaxRetries` and isn't specific to this change; the new wiring test covers the security-relevant field.

## CI

Still only `check-changes` and the CLA check on this branch, with no Go build/test/lint job, same as you observed. That's not something I can trigger from the branch. Locally on the rebased commit: `go build ./...`, `go vet`, and `go test -race ./server/adkrest/... ./cmd/launcher/...` are all green. If a maintainer can approve the workflow run, that would confirm it in CI.

server/adkrest/controllers/triggers/triggers.go
```diff
@@ -237,6 +244,112 @@ func calculateBackoff(attempt int, base, maxDelay time.Duration) time.Duration {
...
+
+// bearerToken extracts the credential from an Authorization header. RFC 9110
+// makes the scheme case-insensitive.
+func bearerToken(authHeader string) (string, error) {
+	scheme, token, found := strings.Cut(authHeader, " ")
+	if !found || !strings.EqualFold(scheme, "Bearer") {
+		return "", fmt.Errorf("Authorization header is not a Bearer credential")
+	}
+	if token = strings.TrimSpace(token); token == "" {
+		return "", fmt.Errorf("Bearer credential is empty")
```
karolpiotrowicz: `golangci-lint` v2.3.1, the version CI pins, flags both of these under ST1005 — error strings shouldn't be capitalized. Line 329 has the same. `.golangci.yml` disables ST1000/1003/1016/1020/1021/1022 but leaves ST1005 on, and the same run against `main` is clean, so these two are the only new ones.

server/adkrest/controllers/triggers/auth_handler_internal_test.go
```diff
@@ -0,0 +1,241 @@
...
+	"testing"
+	"time"
+
+	"github.com/gorilla/mux"
+
+	"google.golang.org/adk/v2/agent"
+	"google.golang.org/adk/v2/server/adkrest/internal/fakes"
+	"google.golang.org/adk/v2/server/adkrest/internal/models"
+	"google.golang.org/adk/v2/session"
+	"google.golang.org/api/idtoken"
```
karolpiotrowicz: goimports wants `google.golang.org/api/idtoken` in the third-party group with `github.com/gorilla/mux`, not in the `google.golang.org/adk/v2` group — `.golangci.yml` sets `local-prefixes: google.golang.org/adk/v2`. This is the third of the three lint findings on CI's pinned version.

server/adkrest/controllers/triggers/config.go
```diff
@@ -26,4 +26,33 @@ type TriggerConfig struct {
...
+	// AllowedServiceAccounts, when non-empty, additionally requires the
+	// verified token to carry a verified email claim matching one of these
+	// service account addresses, the identity the Pub/Sub push subscription
+	// or Eventarc trigger was configured with. Requests presenting a valid
+	// but unlisted principal are rejected with 403.
+	//
+	// Requires ExpectedAudience; it is ignored on its own. Tokens minted
+	// without an email claim (getOpenIdToken defaults includeEmail to false)
+	// are rejected, so the subscription must be configured to include it.
+	AllowedServiceAccounts []string
```
karolpiotrowicz, line 75: Adding a `[]string` here makes `TriggerConfig` non-comparable — the four existing fields are `int` and `time.Duration`, so `cfg1 == cfg2` compiles today and won't after this. The repo's own `.github/scripts/apidiff.sh` against `main` reports it as an incompatible change and exits 1. Nothing in this repo compares two of these, so it's a downstream-only break; the script's message says the route for a deliberate one is the `breaking-change` label.

cmd/launcher/web/triggers/pubsub/pubsub.go
```diff
@@ -110,6 +124,9 @@ func (p *pubsubLauncher) SetupSubrouters(router *mux.Router, config *launcher.Co
 		BaseDelay:         p.config.triggerBaseDelay,
 		MaxDelay:          p.config.triggerMaxDelay,
 		MaxConcurrentRuns: p.config.triggerMaxRuns,
+		ExpectedAudience:  p.config.oidcAudience,
+
+		AllowedServiceAccounts: splitList(p.config.oidcSvcAccounts),
```
karolpiotrowicz: Delete this line and its Eventarc twin and every test still passes. The audience half is covered — deleting the `ExpectedAudience:` line above does fail `TestSetupSubroutersWiresOIDCConfig` in both packages — but nothing pins the allow-list reaching the controller.

The reason it matters more than an ordinary coverage gap is the warning at lines 168-173: it switches on `p.config.oidcSvcAccounts`, the launcher's own flag string, not on the `TriggerConfig` that was built. So if this line went away the operator would see no warning while the endpoint accepted any holder of a Google-signed token for the audience. Passing `-oidc_service_accounts` in the existing test and asserting a 403 for an unlisted principal would cover both.

server/adkrest/controllers/triggers/triggers.go
```diff
@@ -237,6 +244,112 @@ func calculateBackoff(attempt int, base, maxDelay time.Duration) time.Duration {
...
+// subscription's audience is whatever --push-auth-token-audience was set to,
+// defaulting to the full push endpoint URL, and that pinning the caller
+// identity via AllowedServiceAccounts requires the token to carry an email
+// claim, so unlike audience verification alone, that part does depend on how
+// the trigger is configured on the Google Cloud side.
+func (r *RetriableRunner) verifyPushRequestAuth(req *http.Request) error {
+	expectedAudience := r.triggerConfig.ExpectedAudience
+	if expectedAudience == "" {
+		return nil
+	}
```
karolpiotrowicz, line 319: This returns before the allow-list is read, so `AllowedServiceAccounts` set with no `ExpectedAudience` disables verification entirely rather than pinning anything. Both launchers reject that pair at parse time and the field's doc comment says it needs an audience, so it only reaches someone constructing `TriggerConfig` directly — and it lands them on the pre-change behavior rather than something worse. Worth noting that both `*WithConfig` constructors already return an error and already reject an unusable compaction config, so there's somewhere to put the guard.

server/adkrest/controllers/triggers/triggers.go
```diff
@@ -237,6 +244,112 @@ func calculateBackoff(attempt int, base, maxDelay time.Duration) time.Duration {
...
+	token, err := bearerToken(req.Header.Get("Authorization"))
```
karolpiotrowicz: `Header.Get` returns only the first `Authorization` value, so a duplicated header is ignored rather than rejected. Nothing gets past the check — the first header still has to validate — but a proxy in front that reads the last value would disagree with this handler about which credential was presented. `req.Header.Values("Authorization")` with a length check would settle it.

server/adkrest/controllers/triggers/triggers.go
```diff
@@ -237,6 +244,112 @@ func calculateBackoff(attempt int, base, maxDelay time.Duration) time.Duration {
...
+	// allow-list the only property established is that somebody holds a
+	// Google-signed token for this audience, which any principal able to mint
+	// one can satisfy.
+	if len(r.triggerConfig.AllowedServiceAccounts) == 0 {
+		return nil
+	}
+	email, _ := payload.Claims["email"].(string)
+	verified, _ := payload.Claims["email_verified"].(bool)
+	if email == "" || !verified || !slices.Contains(r.triggerConfig.AllowedServiceAccounts, email) {
+		return &authError{status: http.StatusForbidden, public: "untrusted token principal", err: fmt.Errorf("principal %q (verified=%t) is not an allowed service account", email, verified)}
```
karolpiotrowicz: This body differs from the 401 body, so an anonymous caller who receives `untrusted token principal` rather than `invalid identity token` learns that the audience guess was right, the token verified, and an allow-list is configured. The 401-vs-403 status split already carries that and matches adk-python, so this is small — but one constant body with the distinction kept in the log would close it. `TestRespondAuthErrorHidesDetail` only checks that the string "expired" is absent from one 401 body, so it wouldn't catch a regression here.

server/adkrest/controllers/triggers/triggers.go
```diff
@@ -237,6 +244,112 @@ func calculateBackoff(attempt int, base, maxDelay time.Duration) time.Duration {
...
+	// idtoken.Validate checks the signature, the expiry and the audience
+	// string, and leaves every identity claim to the caller. Without an
+	// allow-list the only property established is that somebody holds a
+	// Google-signed token for this audience, which any principal able to mint
+	// one can satisfy.
+	if len(r.triggerConfig.AllowedServiceAccounts) == 0 {
+		return nil
+	}
+	email, _ := payload.Claims["email"].(string)
+	verified, _ := payload.Claims["email_verified"].(bool)
```
karolpiotrowicz, line 357: The `.(bool)` assertion failing closed on a non-boolean `email_verified` is the right call and is stricter than adk-python, which uses Python truthiness and would accept the string `"true"`. No test pins it though — the table covers the absent claim and the literal `false`, so a later change to a friendlier coercion would keep the suite green. A case with `Claims["email_verified"] = "true"` asserting 403 would lock in the behavior you chose.

karolpiotrowicz (dismissed):
Both things I flagged last time are fixed, and the fix is better than what I asked for. The identity is now pinned rather than discarded, and the documentation describes the weaker property instead of the stronger one. I re-ran the whole battery against the new head and drove the handler end to end again.

What settles the first one: with `-oidc_service_accounts` set, the same probe that previously let an unrelated principal through now returns 403 with the agent not running.

```
audience only, foreign principal:      status=200  agentRuns=1
audience + allow-list, same token:     status=403  agentRuns=0
```

The tests behind it are real, not decorative. I deleted the principal pin, deleted the issuer check, made the handler accept any non-empty bearer token, and made it reject every request — all four are caught by the suite. Last time three equivalent mutations survived. `auth_handler_internal_test.go` driving both handlers through a real router with a token, and asserting the agent-run count rather than just the status, is what made the difference.

Two things need to happen before this can merge, and neither is a defect in the logic.

## The apidiff check fails

Adding `AllowedServiceAccounts []string` at [`config.go:57`](https://github.com/google/adk-go/blob/bf0b1a9781f8286b39dfb485199f8de2df8604f1/server/adkrest/controllers/triggers/config.go#L57) makes `TriggerConfig` non-comparable — its four existing fields are all `int` and `time.Duration`, so today `cfg1 == cfg2` compiles and after this it does not. Running the repo's own `.github/scripts/apidiff.sh` against `main` gives:

```
Incompatible changes:
- ./server/adkrest/controllers/triggers.TriggerConfig: old is comparable, new is not
```

Nothing in this repo compares two `TriggerConfig` values, so this only bites a downstream caller. The script's own message says the route for a deliberate break is to label the PR `breaking-change`, which it doesn't currently carry. Either that, or hold the allow-list in a comparable form.

## Three lint findings, on the version CI pins

`golangci-lint` v2.3.1 on the two changed packages, against a clean baseline on `main`:

```
auth_handler_internal_test.go:29:1: File is not properly formatted (goimports)
triggers.go:329:14: ST1005: error strings should not be capitalized
triggers.go:332:14: ST1005: error strings should not be capitalized
```

The import one is `google.golang.org/api/idtoken` sitting in the `google.golang.org/adk/v2` group instead of the third-party group above it.

Worth knowing: the Go build, test, lint and apidiff jobs have never run on this branch. Only `check-changes` and the CLA check have, so everything above is from a local run rather than from a red check you can see.

## The allow-list wiring is the one thing the tests still can't see

[`pubsub.go:129`](https://github.com/google/adk-go/blob/bf0b1a9781f8286b39dfb485199f8de2df8604f1/cmd/launcher/web/triggers/pubsub/pubsub.go#L129) and its Eventarc twin pass `AllowedServiceAccounts` through to the controller. Delete both of those lines and every test still passes. `TestSetupSubroutersWiresOIDCConfig` covers the audience half well — deleting the `ExpectedAudience:` line does fail it in both packages, so the test is doing its job — but the allow-list half has nothing.

What makes this worth fixing rather than noting is the interaction with the startup warning. [`pubsub.go:168-173`](https://github.com/google/adk-go/blob/bf0b1a9781f8286b39dfb485199f8de2df8604f1/cmd/launcher/web/triggers/pubsub/pubsub.go#L168-L173) switches on `p.config.oidcSvcAccounts`, the launcher's own flag string, not on the `TriggerConfig` that was actually built. So if that wiring line ever goes away, the operator sees no warning while the endpoint accepts anyone holding a Google-signed token for the audience — the 200 case above. Extending the existing test to pass `-oidc_service_accounts` and assert a 403 for an unlisted principal closes it.

## Smaller things

- [`triggers.go:284-287`](https://github.com/google/adk-go/blob/bf0b1a9781f8286b39dfb485199f8de2df8604f1/server/adkrest/controllers/triggers/triggers.go#L284-L287) returns before the allow-list is read, so `AllowedServiceAccounts` with no `ExpectedAudience` turns verification off entirely. Both launchers reject that combination at parse time and the field's own doc says it needs an audience, so this only reaches someone building `TriggerConfig` directly — and it lands them on the pre-change behavior rather than something worse. Both `*WithConfig` constructors already return an error and already reject an unusable compaction config.
- The 403 body still differs from the 401 body. A caller who gets `untrusted token principal` rather than `invalid identity token` learns the audience guess was right and an allow-list is configured. The status codes already carry that, and they match adk-python, so this is small — but one constant body would close it.
- [`triggers.go:288`](https://github.com/google/adk-go/blob/bf0b1a9781f8286b39dfb485199f8de2df8604f1/server/adkrest/controllers/triggers/triggers.go#L288) uses `Header.Get`, which returns only the first `Authorization` header. A duplicate is ignored rather than rejected. No way past the check — the first header still has to validate — but a proxy in front that reads the last value would disagree with this handler about which credential arrived.
- `adkgo deploy cloud-run` can't turn any of this on: `cloudrun.go` forwards all four `trigger_*` flags to each sublauncher and has no counterpart for the new pair. That path passes `--no-allow-unauthenticated`, so it isn't exposed — what's missing is the defence-in-depth layer.
- The flags are `-oidc_audience` and `-oidc_service_accounts`, while adk-python uses `--trigger_oidc_audience` and `--trigger_oidc_service_accounts` and the other four flags on this same flagset are all `trigger_`-prefixed. They're unreleased, so renaming is free now.
- `AllowedServiceAccounts` is stored by slice header at [`triggers.go:127`](https://github.com/google/adk-go/blob/bf0b1a9781f8286b39dfb485199f8de2df8604f1/server/adkrest/controllers/triggers/triggers.go#L127), so a caller that keeps its slice and mutates it later changes the live allow-list. A `slices.Clone` and a line in the field doc would settle it.
- `email_verified` correctly fails closed on a non-boolean, which is better than adk-python's truthiness check — but no test pins it, so a later "friendlier" coercion would stay green.

## One question rather than a finding

With only `-oidc_audience` set, the endpoint still accepts any principal that can mint a Google-signed token for that audience. You now say this in four places and warn about it at every boot, which is exactly what I asked for, so I'm not treating it as a defect.

The open question is whether the allow-list should be required rather than recommended. There's a real cost to requiring it: `getOpenIdToken` defaults `includeEmail` to false, so mandating the pin would turn a documented weak mode into a 403 on every legitimate delivery for anyone whose token carries no email claim — and on Pub/Sub push a 403 is a NACK, so that's a redelivery loop rather than a clean refusal. Audience-only is still worth something on its own, since it stops a token minted for another service being spent here. I'd like a maintainer's call on it rather than deciding it in review.

## What held up

The ordering is right and the tests prove it: the check is the first statement in both handlers, ahead of the body read, the route lookup, the semaphore and the agent, and the new tests assert the agent didn't run rather than only checking the status. `idtoken.Validate` in `v0.293.0` checks `aud`, `exp` and the signature and nothing else — [`triggers.go:313-320`](https://github.com/google/adk-go/blob/bf0b1a9781f8286b39dfb485199f8de2df8604f1/server/adkrest/controllers/triggers/triggers.go#L313-L320) closes the identity gap and the issuer check above it closes the ES256/IAP key-set path, which would otherwise verify a token issued by `https://cloud.google.com/iap`. Claim handling fails closed everywhere I pushed on it: nil payload, nil claims map, `email` as a non-string, `email_verified` as a non-boolean, an empty string in the allow-list. With the audience unset the function returns before touching a header, so existing deployments are unaffected, and both production construction sites are updated.

Two things I couldn't check, so you shouldn't read them as cleared. I never ran the real `idtoken.Validate` — the sandbox has no network and every test installs a stub, so the issuer behavior above is from reading the pinned library source. And `-race` passes but proves little here: no test drives two concurrent requests through one controller, and the launcher packages have no parallel tests at all.

server/adkrest/controllers/triggers/config.go
herdiyana256: Fixed by keeping it comparable rather than taking the label. The two settings now live behind `OIDC *OIDCConfig`, so the pointer field leaves `TriggerConfig` comparable and `cfg1 == cfg2` still compiles. Both fields are unreleased, so the move cost nothing, and it groups a pair that only works together.

`.github/scripts/apidiff.sh origin/main` on the new head:

```
Target origin/main, comparing against merge base ffddb2a6d7ed
Collecting the API at the base revision
  google.golang.org/adk/v2
  google.golang.org/adk/plugin/agentanalytics
Comparing HEAD against it
No incompatible changes.
```

server/adkrest/controllers/triggers/auth_handler_internal_test.go
herdiyana256: Fixed: `google.golang.org/api/idtoken` now sits with `github.com/gorilla/mux` in the third-party group.

server/adkrest/controllers/triggers/triggers.go
herdiyana256: Fixed, both lowercased, and they are `errors.New` now since neither formats anything. `golangci-lint` v2.3.1 (I installed the pinned commit rather than trusting my newer local one) reports `0 issues` across the three changed packages.

cmd/launcher/web/triggers/pubsub/pubsub.go
herdiyana256: Fixed at the source rather than only in the test. Both launchers now build their `TriggerConfig` in one place, `triggerConfig()`, and `SetupSubrouters` and `UserMessage` both read it — so the warning can no longer describe a pin the controller doesn't have. `TestTriggerConfigWiresOIDC` asserts both OIDC fields on what that function returns, and `TestUserMessageWarnsFromBuiltConfig` asserts the warning that goes with it. Deleting the `AllowedServiceAccounts:` line fails both, in both packages.

What I could not do is the end-to-end 403 you suggested. Reaching it needs a token that verifies, which means either a real call to Google's certificate endpoint or the `validateIDToken` seam — and that seam is unexported inside `triggers`, which is the shape you asked for last round. So the audience half stays covered end-to-end by the 401, and the allow-list half is pinned at the config boundary.

server/adkrest/controllers/triggers/triggers.go
herdiyana256: Fixed. Verification is switched off by leaving `TriggerConfig.OIDC` nil, so a non-nil block with no audience is unambiguously a mistake, and it's now rejected in three places:

- both `*WithConfig` constructors return an error, alongside the compaction check you pointed at;
- the deprecated constructors have no error to return, so they log and build a controller that denies every request rather than handing back a nil pointer that panics on the first delivery;
- `verifyPushRequestAuth` denies (500) on the same condition, so a controller built some other way can't end up on the unverified path.

server/adkrest/controllers/triggers/triggers.go
herdiyana256: Fixed: `req.Header.Values("Authorization")` with a length check, so two credentials are ambiguous rather than silently resolved to the first. `TestBearerToken` covers two headers and two headers where the second isn't a Bearer credential; `TestVerifyPushRequestAuth` covers the duplicate at the handler level. Reverting to `Header.Get` semantics fails both.

server/adkrest/controllers/triggers/triggers.go
herdiyana256: Fixed. There's one `authFailedMessage` constant for every rejection now, and `authError` no longer carries a public string at all — only the status and the log-side error. `TestRespondAuthErrorBodyDoesNotVaryWithStatus` compares the 401 and 403 bodies directly, so the wording can't drift apart again.

server/adkrest/controllers/triggers/triggers.go
herdiyana256: Pinned: a case with `Claims["email_verified"] = "true"` asserting 403. Changing the assertion to a truthiness check (`payload.Claims["email_verified"] != nil`) now fails it.

herdiyana256:
Thanks every finding is fixed, pushed as `4160f59`. Replies are on the individual threads; this is the summary and the measurements.

## The two things that had to happen before merge

**apidiff.** Fixed by keeping `TriggerConfig` comparable rather than by asking for the `breaking-change` label. The two settings now sit behind `OIDC *OIDCConfig`, a pointer field, so `cfg1 == cfg2` still compiles. Both fields are unreleased, so the move was free, and it groups a pair that only works together — which also gave the `AllowedServiceAccounts`-without-audience case somewhere to be rejected. `.github/scripts/apidiff.sh origin/main` on the new head:

```
Comparing HEAD against it
No incompatible changes.
```

**Lint.** All three fixed. I installed the commit CI pins rather than trusting my newer local binary `golangci-lint has version 2.3.1` — and it reports `0 issues` across `server/adkrest/controllers/triggers`, both launcher packages and `cmd/adkgo/internal/deploy/cloudrun`.

## The allow-list wiring
Fixed at the source rather than only in the test, because you were right that the coverage gap wasn't the real problem. Both launchers build their `TriggerConfig` in one function now, and both `SetupSubrouters` and `UserMessage` read it. So the case you described — the wiring line disappears, the operator still sees no warning can't happen: the warning follows the config that was built.

`TestTriggerConfigWiresOIDC` pins both OIDC fields to what that function returns, in both packages. The end-to-end 403 you suggested is the one thing I couldn't do: it needs a token that verifies, which means either a real call to Google's certificate endpoint or the `validateIDToken` seam — and that seam is unexported inside `triggers`, which is the shape you asked for last round. So the audience half stays covered end-to-end by the 401 and the allow-list half is pinned at the config boundary.

## Everything else from the review
- **`AllowedServiceAccounts` with no audience.** Verification is now switched off by leaving `OIDC` nil, so a non-nil block with no audience is unambiguously a mistake. Both `*WithConfig` constructors reject it; the deprecated constructors log and build a controller that denies every request rather than returning a nil pointer that panics on the first delivery; and `verifyPushRequestAuth` denies on the same condition, so a controller built some other way can't land on the unverified path.
- **403 body.** One constant for every rejection. `authError` no longer carries a public string at all — status and log-side error only.
- **Duplicate `Authorization`.** `Header.Values` with a length check: two credentials are ambiguous rather than resolved to the first.
- **`adkgo deploy cloud-run`.** `--pubsub_oidc_audience`, `--pubsub_oidc_service_accounts` and the eventarc pair, forwarded to whichever sublauncher is enabled. Two things while I was in there: the values are JSON-encoded into the `CMD` array rather than pasted between quotes, since they're operator-supplied strings landing in a JSON literal, and the allow-list without an audience is rejected at deploy time instead of producing a crash-looping revision.
- **Flag names.** Now `-trigger_oidc_audience` and `-trigger_oidc_service_accounts`, matching adk-python and the four `trigger_`-prefixed flags already on that flagset.
- **Slice aliasing.** The `OIDCConfig` and its allow-list are copied at construction.
- **`email_verified`.** Pinned with a `"true"` string case asserting 403.

## Measurements
Every fix above, mutated and re-run. All twelve are caught; the last one wasn't at first, which is the useful part.

| mutation | result |
|---|---|
| drop `AllowedServiceAccounts` wiring (pubsub) | **FAIL** - `TestTriggerConfigWiresOIDC`, `TestUserMessageWarnsFromBuiltConfig` |
| drop `AllowedServiceAccounts` wiring (eventarc) | **FAIL** - same two |
| drop `ExpectedAudience` wiring | **FAIL** - `TestSetupSubroutersWiresOIDCConfig`, `TestTriggerConfigWiresOIDC` |
| accept duplicate `Authorization` headers | **FAIL** - `TestBearerToken`, `TestVerifyPushRequestAuth` |
| coerce `email_verified` truthily | **FAIL** - `TestVerifyPushRequestAuth` |
| put the error detail back in the 403 body | **FAIL** - `TestRespondAuthErrorBodyDoesNotVaryWithStatus` |
| drop the misconfiguration denial | **FAIL** - `TestDeprecatedConstructorsFailClosedOnOIDCWithoutAudience` |
| drop the constructor guard | **FAIL** - `TestControllersRejectOIDCWithoutAudience` |
| drop OIDC forwarding in `cloudrun.go` | **FAIL** - `TestWriteTriggerArgs` |
| paste the deploy flag values unescaped | **FAIL** - `TestWriteTriggerArgsEscapesFlagValues` |
| stop cloning the allow-list | **FAIL** - `TestOIDCConfigIsCopiedAtConstruction` |

That last row survived on the first attempt: my test mutated the caller's slice with `append`, which reallocates, so it would have passed with or without the `slices.Clone`. Mutating element 0 in place is the case that actually aliases, and the test does that now.

`go build ./...`, `go vet`, and `go test -race -count=1 ./server/adkrest/... ./cmd/launcher/... ./cmd/adkgo/...` are green locally. CI on this branch still runs only `check-changes` and the CLA check everything above is a local run, not a green check you can see, and the Go build/test/lint/apidiff jobs would still need a maintainer to approve the workflow.

## Still yours to call
The open question is unchanged and I haven't decided it in code: with only an audience set, the endpoint accepts any principal that can mint a Google-signed token for it. Requiring the allow-list would close that, at the cost of 403ing every delivery whose token carries no email claim and on Pub/Sub push a 403 is a NACK, so a redelivery loop rather than a clean refusal. It's now stated in the PR description as an explicit open item so it doesn't get lost. @wolo-lab, that one and the scope question @karolpiotrowicz raised are the same call.

herdiyana256:
I'll rebase onto main to clear the cloudrun.go conflict.

Everything from the last review was addressed in 4160f59, with replies on each thread, so the CHANGES_REQUESTED state predates those fixes.

@wolo-lab this is still assigned to you for the scope call, which is the only part left that isn't mine to decide: whether the idtoken dependency and the new TriggerConfig.OIDC field are acceptable, and whether the allow-list should be required rather than recommended.

One other thing: lint (.) and test (.) are Required but have never run on this branch, the workflow still needs a maintainer to approve it.

cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
```diff
@@ -229,23 +253,45 @@ CMD ["/app/` + f.build.execFile + `", "web", "-port", "` + strconv.Itoa(flags.cl
 			}
 			if flags.cloudRun.pubsub {
 				b.WriteString(`, "pubsub"`)
-				fmt.Fprintf(&b, `, "--trigger_max_retries", "%d"`, flags.cloudRun.pubsubTrigger.maxRetries)
-				fmt.Fprintf(&b, `, "--trigger_base_delay", "%s"`, flags.cloudRun.pubsubTrigger.baseDelay.String())
-				fmt.Fprintf(&b, `, "--trigger_max_delay", "%s"`, flags.cloudRun.pubsubTrigger.maxDelay.String())
-				fmt.Fprintf(&b, `, "--trigger_max_concurrent_runs", "%d"`, flags.cloudRun.pubsubTrigger.maxRuns)
+				writeTriggerArgs(&b, flags.cloudRun.pubsubTrigger)
```
karolpiotrowicz: This call site and its eventarc twin on line 260 are the only place the per-trigger wiring is decided, and nothing exercises them. Swapping this to `flags.cloudRun.eventarcTrigger` leaves `go test ./cmd/adkgo/internal/deploy/cloudrun/...` green, because both new tests call `writeTriggerArgs` directly rather than going through `prepareDockerfile`, which has no test of its own.

The swap is not a cosmetic one. With pubsub's audience set and eventarc's empty, the pubsub endpoint deploys with no `--trigger_oidc_audience` at all and runs unverified, which is the state this PR exists to prevent.

A test over `prepareDockerfile` asserting each trigger's own audience lands in its own section of the emitted CMD would close it. That function currently reads the package global rather than the receiver, so it may be easier after the rebase, which moves those reads onto `f`.

cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
```diff
@@ -131,6 +146,15 @@ func (f *deployCloudRunFlags) computeFlags() error {
 			if f.cloudRun.debugAPI && !f.cloudRun.api {
 				return fmt.Errorf("cannot enable Debug API without having enabled API")
 			}
+			// The sublauncher rejects this pair too, but that happens inside
+			// the deployed container, where the operator sees a crash-looping
+			// revision rather than an error.
+			if f.cloudRun.pubsubTrigger.oidcServiceAccounts != "" && f.cloudRun.pubsubTrigger.oidcAudience == "" {
+				return fmt.Errorf("--pubsub_oidc_service_accounts requires --pubsub_oidc_audience")
```
karolpiotrowicz: This check and the eventarc one below it are exercised by nothing — deleting both leaves the package's tests green. The failure mode is the benign direction, an operator getting a crash-looping revision instead of a local error, which is exactly what the comment above says these lines exist to prevent.

They are cheap to pin: both run before any filesystem work, so a table test over `computeFlags` with each half of each pair set would do it.

cmd/adkgo/internal/deploy/cloudrun/cloudrun_internal_test.go
```diff
@@ -0,0 +1,120 @@
...
+					t.Errorf("writeTriggerArgs() = %q, want it not to contain %q", got, notWant)
+				}
+			}
+		})
+	}
+}
+
+// The CMD line is a JSON array, so a flag value carrying a quote must not be
+// able to close the string it sits in and add arguments of its own.
+func TestWriteTriggerArgsEscapesFlagValues(t *testing.T) {
```
karolpiotrowicz, line 101: This only injects a hostile value into `oidcAudience`. The service-accounts value goes through the identical `jsonString` call on line 281 of `cloudrun.go` and has no equivalent case, so removing the encoding from that one line alone keeps the suite green.

Both values are correctly encoded today, so this is coverage rather than a defect. Tabling this test over both fields, or setting both in one case and round-tripping the whole array through `json.Unmarshal`, covers it.

cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
```diff
@@ -229,23 +253,45 @@ CMD ["/app/` + f.build.execFile + `", "web", "-port", "` + strconv.Itoa(flags.cl
...
+		fmt.Fprintf(b, `, "--trigger_oidc_audience", %s`, jsonString(cfg.oidcAudience))
+	}
+	if cfg.oidcServiceAccounts != "" {
+		fmt.Fprintf(b, `, "--trigger_oidc_service_accounts", %s`, jsonString(cfg.oidcServiceAccounts))
+	}
+}
+
+// jsonString renders v as a JSON string literal. Marshaling a string cannot
+// fail, so the error path is only there to keep the output well-formed.
+func jsonString(v string) string {
```
karolpiotrowicz, line 350: The comment says marshaling a string cannot fail, which is true, but it skips the case that does change the value: `json.Marshal` does not reject invalid UTF-8, it substitutes U+FFFD per bad byte.

`--pubsub_oidc_audience=$'https://x\xff'` therefore deploys a container verifying an audience the operator never typed, and every push delivery is rejected with nothing pointing back at the deploy step. This fails closed, so it is a diagnosability problem rather than a security one.

Worth noting that `util.ValidateDockerfileSafe` on main refuses exactly this input for the values it guards. After the rebase that helper is available here.

cmd/launcher/web/triggers/pubsub/pubsub_test.go
```diff
@@ -95,3 +98,194 @@ func TestSetupSubrouters(t *testing.T) {
 		t.Errorf("SetupSubrouters() did not register expected route")
 	}
 }
+
+// TestSetupSubroutersWiresOIDCConfig checks that -trigger_oidc_audience actually
+// reaches the controller, not just the launcher's own config struct. The
+// startup warning reads the flag field, so config that never gets threaded
```
karolpiotrowicz: This sentence is no longer true, and this push is what changed it: `UserMessage` now reads `p.triggerConfig().OIDC` rather than the flag field. The failure it describes is also inverted now — reading the built config is precisely what stops a setting that never reaches the controller from still being announced as enforced.

The line directly above got the flag rename, so this one was just missed.

cmd/launcher/web/triggers/eventarc/eventarc_test.go
```diff
@@ -0,0 +1,219 @@
...
+	"testing"
+
+	"github.com/gorilla/mux"
+
+	"google.golang.org/adk/v2/cmd/launcher"
+)
+
+// TestSetupSubroutersWiresOIDCConfig checks that -trigger_oidc_audience actually
+// reaches the controller, not just the launcher's own config struct. The
+// startup warning reads the flag field, so config that never gets threaded
```
karolpiotrowicz: Same stale sentence as in the pubsub test: `UserMessage` reads the built config now, not the flag field.

karolpiotrowicz (changes requested):
Two things to fix before this merges, both about what the tests can catch rather than what the code does. Everything the last round asked for is resolved, and I checked each one rather than taking the description's word for it.

## Resolved since the last round

The comparability fix works. `TriggerConfig` stays comparable with `OIDC` as a pointer, the repo's apidiff check passes against `main`, and I confirmed the check still has teeth by adding a `[]string` field directly to the struct and watching it fail with the same "old is comparable, new is not" it produced last time. **No `breaking-change` label is needed.** The three lint findings are gone at the version CI pins.

The rest of the list closed too, and each one is now held by a test that fails if you take the behaviour away: the audience-without-allow-list case denies instead of disabling, the 401 and 403 bodies are one constant, a second `Authorization` header is rejected rather than ignored, the allow-list is cloned rather than aliased, the deploy command can turn the feature on, and the flag names now match adk-python.

## The allow-list wiring is pinned closer to the source, but the production path still is not

Last round's point was that deleting the allow-list from the launcher left both suites green. That exact deletion now fails `TestTriggerConfigWiresOIDC`, so the fix is real. One seam further along is still open.

[`TestTriggerConfigWiresOIDC`](https://github.com/google/adk-go/blob/4160f59448d8e3d265718b75c855f81d63bb1650/cmd/launcher/web/triggers/pubsub/pubsub_test.go#L218) reads `l.triggerConfig().OIDC` directly and never calls `SetupSubrouters`, which is the path that actually hands the controller its config at [`pubsub.go:128`](https://github.com/google/adk-go/blob/4160f59448d8e3d265718b75c855f81d63bb1650/cmd/launcher/web/triggers/pubsub/pubsub.go#L128) and [`eventarc.go:128`](https://github.com/google/adk-go/blob/4160f59448d8e3d265718b75c855f81d63bb1650/cmd/launcher/web/triggers/eventarc/eventarc.go#L128). Rewrite `SetupSubrouters` to build the config inline and keep only the audience, and the whole suite stays green while the deployed endpoint accepts any caller holding a Google-signed token for that audience.

```
# in cmd/launcher/web/triggers/pubsub/pubsub.go, replace
#     TriggerConfig:   p.triggerConfig(),
# with an inline triggers.TriggerConfig that sets ExpectedAudience and omits AllowedServiceAccounts
go test ./cmd/launcher/web/triggers/... ./server/adkrest/controllers/triggers/...
# ok — all four packages pass
```

The audience half does not have this gap: `TestSetupSubroutersWiresOIDCConfig` drives a real router and observes a 401, so dropping the audience from `SetupSubrouters` is caught. It is the allow-list specifically that no test follows to the controller.

**The property to hold, rather than the one case above:** every OIDC setting the launcher parses must be observable on the config `SetupSubrouters` passes to the controller, not merely on the helper that builds it. Otherwise the same gap reopens wherever a future change routes around `triggerConfig()`.

The comment on that test ([`pubsub_test.go:170`](https://github.com/google/adk-go/blob/4160f59448d8e3d265718b75c855f81d63bb1650/cmd/launcher/web/triggers/pubsub/pubsub_test.go#L170), and the same text in the eventarc test) says it pins "the TriggerConfig that `SetupSubrouters` hands the controller, which is the only place the launcher can get either of them wrong". That is the sentence to make true, or to correct.

I am deliberately not prescribing how. I considered two approaches and executed neither, so I would rather you pick.

## `prepareDockerfile` has no test at all

Details inline on [`cloudrun.go:256`](https://github.com/google/adk-go/blob/4160f59448d8e3d265718b75c855f81d63bb1650/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go#L256). The short version: the new tests exercise `writeTriggerArgs` directly, so swapping which trigger's flags each call site passes goes unnoticed, and the failure that swap produces is an endpoint deployed without the audience it needed.

## The rebase has one resolution worth getting right

The conflict is in `prepareDockerfile`. Your branch calls `writeTriggerArgs(&b, flags.cloudRun.pubsubTrigger)` off the package global, and [#1240](https://github.com/google/adk-go/pull/1240) moved every read in that function onto the receiver `f`, with a comment explaining that a global read there lets the guard and the line it guards come apart for any caller that is not the cobra `RunE`. Resolving in favour of your side reintroduces exactly that. The resolution that keeps both changes is `writeTriggerArgs(&b, f.cloudRun.pubsubTrigger)`, and the same for eventarc.

Worth knowing that [#1399](https://github.com/google/adk-go/pull/1399) also adds Cloud Run deploy flags in the same file, so whichever of you lands second will rebase again.

## Two unrelated things I found while testing, which are not yours

Both predate this branch and neither should hold up this PR. The Eventarc handler hangs forever if `MaxConcurrentRuns` is left at zero, because the resulting channel is unbuffered rather than nil, and the Pub/Sub controller never assigns its semaphore at all, so that limit is silently unenforced there. I can open an issue for both if that is useful.

The scope question raised on 2 September is still with @wolo-lab and is independent of the code review. One correction to what I wrote then: this is not a new module dependency — `go.mod` is untouched and `google.golang.org/api` was already required — and adk-python has the whole feature, including both flag names, so the parity gap is narrower than I first described.

herdiyana256:
Both gaps are closed and the rebase conflict is resolved the way you called it. Pushed as a rebase onto current main, so the branch is clean against `prepareDockerfile` again.

The allow-list now reaches the controller through a seam the tests can see. Each launcher builds its controller through a `newController` field defaulted to the real constructor, the same field shape as the `validateIDToken` seam you preferred last round. `TestSetupSubroutersPassesConfigToController` installs a stub that captures the `ControllerConfig` that `SetupSubrouters` actually passes and asserts the allow-list on it, rather than reading `triggerConfig()` directly. The mutation you described, rewriting `SetupSubrouters` to build the config inline and keep only the audience, now fails that test in both packages instead of staying green. The comment on the old test that claimed it pinned "the only place the launcher can get either of them wrong" is gone, since that sentence was the thing that was not true.

`prepareDockerfile` has a test now. `TestPrepareDockerfileWiresEachTriggerAudience` drives it end to end, reads back the generated Dockerfile, parses the CMD array, and asserts each trigger's own audience lands in its own section. Swapping `writeTriggerArgs(&b, f.cloudRun.pubsubTrigger)` to the eventarc trigger fails it. The conflict resolution keeps the `f.cloudRun` receiver reads from #1240 rather than reintroducing the package global.

The parse-time guards are pinned and the UTF-8 case is closed. The service-account-without-audience checks and the Dockerfile-safety checks moved into `validateTriggerOIDCFlags`, a pure method with no filesystem work, covered by `TestValidateTriggerOIDCFlags` over each half of each pair. On the U+FFFD point you raised: the OIDC values now go through `util.ValidateDockerfileSafe` up front, the same as `--a2a_agent_url`, so an invalid-UTF-8 audience is rejected at the deploy step rather than deploying a container that verifies an audience the operator never typed. The `jsonString` comment now says this instead of claiming marshaling cannot change the value. The escape test sets both hostile values and round-trips the whole CMD array through `json.Unmarshal`, so removing the encoding from the service-accounts line alone no longer stays green.

The two stale comments in the launcher tests that still described `UserMessage` reading the flag field are corrected; it reads the built config now.

Measurements, each mutation run rather than asserted:

    mutation                                              result
    swap eventarc trigger for pubsub in prepareDockerfile FAIL - TestPrepareDockerfileWiresEachTriggerAudience
    drop the pubsub oidc-requires-audience guard          FAIL - TestValidateTriggerOIDCFlags
    drop service-accounts jsonString escaping             FAIL - TestWriteTriggerArgsEscapesFlagValues
    build SetupSubrouters config inline, audience only    FAIL - TestSetupSubroutersPassesConfigToController

Locally on the rebased head: `go build`, `go vet`, `go test -race -count=1`, `gofmt`, `golangci-lint` at the config in the repo (0 issues), and `.github/scripts/apidiff.sh origin/main` (no incompatible changes) all pass across the four packages. CI still needs a maintainer to approve the workflow run.

On the two things you found that are not mine: the Eventarc zero-`MaxConcurrentRuns` hang and the Pub/Sub semaphore never being assigned both predate this branch, so a separate issue for each sounds right rather than folding them in here. Please do open them.

@wolo-lab the scope call is still yours and unchanged by this push: whether to take the `TriggerConfig.OIDC` field, and whether the allow-list should be required rather than recommended. karol's correction stands, `go.mod` is untouched so this is not a new dependency.

wolo-lab:
Thank you for staying with it this long.

The scope call is mine and I owe you an answer. My one open concern is how this fits with [#1561](https://github.com/google/adk-go/pull/1561), which landed while this PR was in review and gave adk-go a pluggable authentication seam that did not exist when you designed this. I want to agree the shape with its author before I answer, rather than send you a direction and then walk it back.

One thing you can have now, because it may save you a week. The direction I expect to land on is that this verification becomes an authn.Authenticator provider living in server/authn next to the four already there, rather than a new field on the released TriggerConfig. That piece looks independent of everything I still have to settle: it is your existing verification logic as the body of Authenticate, plus its tests, touching nothing else. The wiring onto the routes, the flags, and one question about status codes are the parts that depend on the conversation I still owe you.

I would not start on it before I confirm, and I will confirm early next week. I would rather you knew where this is heading than hear nothing for a week.

herdiyana256:
Thanks for the update, and for saying where it is heading rather than going quiet for a week.

You said not to start before you confirm. I built the provider anyway, locally, on a branch off `main`. It is not pushed and it is not on this PR, so there is nothing here you are committed to. I am raising it because writing it settled one of the questions you listed as still open, and that is more useful to you before the conversation with #1561's author than after.

It is two files in `server/authn`, `googleoidc.go` and its test, nothing else touched. `go.mod` and `go.sum` are unchanged, since `idtoken` comes from `google.golang.org/api` which is already required. `NewGoogleOIDC(audience string, allowedServiceAccounts []string)` sits next to the other four constructors, and `Authenticate` is the verification logic from this PR: `idtoken.Validate` against the audience, the issuer check in both spellings, then the verified `email` claim against the allow-list. On success the `Caller` carries `sub` as `UserID`, and `email`, `email_verified` and `aud` as claims.

On the status codes: the seam has no 403.

```go
// server/authn/authn.go
func writeAuthError(w http.ResponseWriter, err error) {
	if !errors.Is(err, ErrUnauthenticated) {
		http.Error(w, "authentication failed", http.StatusInternalServerError)
		return
	}
	http.Error(w, "unauthorized", http.StatusUnauthorized)
}
```

`ErrUnauthenticated` is 401 and anything else is 500, so a token that verifies but names a principal outside the allow-list has two options. Keep the allow-list in `Authenticate` and wrap `ErrUnauthenticated`, which is what I did, and the 401/403 split in this PR becomes 401 only. Or move the allow-list out to an `authz.Authorizer` and get a real 403, at the cost of admitting an unlisted principal as authenticated first.

What does not work is returning a bare error, because it lands in the 500 branch. On Pub/Sub push a 5xx is a NACK, so a forged token becomes a redelivery loop rather than a refusal, the same shape as the 403-on-every-delivery problem in the open question at the bottom of the description. It is about six lines either way, so whichever you prefer is fine.

One thing I guessed at. `NewGoogleOIDC` returns `(Authenticator, error)` and rejects an empty audience, where the other four return a bare `Authenticator`. An audience-less verifier accepts any Google-signed token while looking configured, so I did not want it constructible. If you would rather match the siblings, it can return an authenticator that denies everything instead.

Separately, `Middleware`'s 401 sets no `WWW-Authenticate` header, and the handler in this PR does. Small, but it belongs with the wiring.

The tests carry over as they are: a valid token from an unlisted principal, a token with no `email` claim at all since `getOpenIdToken` defaults `includeEmail` to false, `email_verified` arriving as the string `"true"`, and two `Authorization` headers on one request. Green under `-race`, vet and golangci-lint clean.

Tell me which way the allow-list goes when you have it and I will open it as its own PR against `main`, or fold it in here.

wolo-lab:
### Review Summary & Next Steps

**1. `idtoken` dependency: approved**
Already required by the module, so no `go.mod` change.

**2. `TriggerConfig` field: not in this shape**
#1561 landed the `authn.Authenticator` seam while this PR was in review. Rather than adding a field to the released `TriggerConfig`, please move the verification into an `authn.Authenticator` and gate the trigger handlers with `authn.Middleware`. The released struct stays untouched, the verifier becomes reusable, and the caller constructs it, which is what lets an operator supply an HTTP client to `idtoken.Validate` — it fetches Google's signing certificates on the request path. adk-python has the same structure, with the verifier injected into the trigger router as a callable.

**3. Allow-list: mandatory**
Audience verification alone does not identify the caller. `aud` is chosen by whoever mints the token, so any principal with a service account can obtain a Google-signed one for it.

`includeEmail=false` reaches neither delivery path. It belongs to `iamcredentials.generateIdToken`, while a push subscription's OIDC config has only [`audience` and `serviceAccountEmail`](https://pkg.go.dev/google.golang.org/api/pubsub/v1#OidcToken) and no switch for the claim. Eventarc delivers through the same machinery: a trigger's [`transport.pubsub.subscription`](https://pkg.go.dev/google.golang.org/api/eventarc/v1#Pubsub) is "the name of the Pub/Sub subscription created and managed by Eventarc as a transport for the event delivery". Worth a line in the flag help that a hand-minted token for manual testing needs `includeEmail: true`.

**4. How to land it: split into two PRs**

*   **PR 1 (the provider):** `authn.NewGoogleOIDC` in `server/authn`, with the allow-list inside it rather than in an `authz.Authorizer` — `CanActAsUser` needs a `{user_id}` in the path, and trigger paths carry `{app_name}`. Add an `authn.ErrForbidden` sentinel and a third branch in `writeAuthError` so an unlisted principal gets 403. We are deliberately stricter than adk-python here, so please carry one sentence saying why in the provider's doc comment and in the PR description.
*   **PR 2 (the wiring):** trigger routes, CLI flags, Cloud Run forwarding. Give `pubsub` and `eventarc` a config struct each, carrying an `Authenticator`, so the two can grow independently later — and no `Authorizer`, since trigger paths have no user to authorize. Don't read `launcher.Config.Authenticator`: it covers the REST API, and widening it would gate Pub/Sub deliveries with an authenticator chosen for browser traffic. When both a config authenticator and the OIDC flags are set, the config wins.

*(One thing to expect rather than fix: the `WWW-Authenticate` header your handler sets on a 401 has no equivalent in `authn.Middleware`, so it will be missing after the move. That gap is ours and we are tracking it separately.)*

**5. CI & bugs**

*   I will open separate issues for the two bugs you found (Eventarc hang & Pub/Sub semaphore).
*   We will raise the flag divergence with adk-python ourselves. Their `--trigger_oidc_service_accounts` stays optional, and that is a conversation for their maintainers rather than something to land on you.

*Apologies for the late architectural shift caused by the #1561 merge. If you prefer not to do this refactor, let us know and we will gladly carry your branch forward.*

herdiyana256:
I've done the refactor from the review and split it into two PRs as suggested (point 4).

PR 1 (the provider) is up: #1636 -- authn.NewGoogleOIDC in server/authn with the allow-list inside it, an ErrForbidden sentinel and a third arm in writeAuthError so an unlisted principal gets 403. The released TriggerConfig is untouched, and go.mod is unchanged.

PR 2 (the wiring -- trigger routes, CLI flags, Cloud Run forwarding) is ready and I'll open it once #1636 lands, so it reviews as a clean diff on top. It gives pubsub and eventarc a Config each carrying an Authenticator, keeps the allow-list mandatory, and does not read launcher.Config.Authenticator.

I'll close this PR in favour of those two unless you'd prefer otherwise.

wolo-lab:
Thank you so much for your flexibility and for moving so quickly on this refactoring! Closing this PR in favor of the new ones sounds like a great plan. Someone from our team will take a look at #1636 shortly. I really appreciate your excellent collaboration and patience with the late architectural shift!

Issue #1248, opened 2026-08-05
streamingResponseAggregator drops real UsageMetadata when a later chunk in the same turn carries none


AmirF194:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

## 🔴 Required Information

**Describe the Bug:**
`streamingResponseAggregator.aggregateResponse` in `internal/llminternal/stream_aggregator.go` overwrites `s.usageMetadata` unconditionally on every chunk:

```go
s.response = llmResponse
s.usageMetadata = llmResponse.UsageMetadata
if llmResponse.GroundingMetadata != nil {
    s.groundingMetadata = llmResponse.GroundingMetadata
}
if llmResponse.CitationMetadata != nil {
    s.citationMetadata = llmResponse.CitationMetadata
}
```

`GroundingMetadata` and `CitationMetadata`, two lines below, are both sticky (assigned only when non-nil, i.e. keep-last-non-nil). `UsageMetadata` has no such guard.

Real streams legitimately carry `UsageMetadata` on a non-final chunk and leave it absent on the true final chunk. The repo's own `TestMetadataVertexAISSEStream` fixture models exactly this shape (a leading metadata-only chunk for `gemini-3-flash-preview` + `googleSearch` grounding, followed by a contentless-usage final chunk) but never asserts on `UsageMetadata`. When that shape occurs, the final chunk's nil `UsageMetadata` silently wipes out the real usage an earlier chunk reported, and `Close()` hands the caller a final aggregated `LLMResponse` with `UsageMetadata == nil` despite tokens genuinely having been consumed.

Any `AfterModelCallback`-based cost tracking, budget enforcement, or usage logging reads zero/missing tokens for a turn that actually consumed them, which is worse than over-counting: a budget guard keyed off zero usage fails open silently.

**Steps to Reproduce:**
1. Construct a `streamingResponseAggregator` via `NewStreamingResponseAggregator()`.
2. Feed it a chunk with `UsageMetadata: &genai.GenerateContentResponseUsageMetadata{TotalTokenCount: 150}` and no `Candidates`.
3. Feed it a second, final chunk with real `Candidates` content and no `UsageMetadata` (mirrors the shape already covered in `TestMetadataVertexAISSEStream`).
4. Call `Close()`.

**Expected Behavior:**
[...]

PR #1249, opened 2026-08-05
test(llminternal): pin keep-last semantics for streaming usage metadata

SHAI-shivansh-sharma:
## Description

This PR adds unit test coverage for streaming `UsageMetadata` aggregation in `internal/llminternal`.

The production fix for preserving `UsageMetadata` across chunks, originally reported in #1248, was merged in #1257.

This PR now focuses on the remaining test gap by pinning the **keep-last-non-nil** behavior when later streaming chunks contain updated `UsageMetadata`.

Because Gemini usage metadata in streaming responses is cumulative, the latest non-nil `UsageMetadata` should overwrite earlier metadata values.

## Changes

- Added `TestUsageMetadataUpdatedByLaterChunkWithValue`.
- Removed the redundant test for later nil metadata.
- No production code changes are included.

Related: #1248

## Testing

Ran locally:

```bash
go test -count=1 -run UsageMetadata ./internal/llminternal/
go test ./internal/llminternal/...

### CODE REVIEW

**Comment by SHAI-shivansh-sharma** [comment]
Thanks for reviewing. I have signed the Google CLA and the check is now passing.

The fix preserves UsageMetadata across streaming chunks when later chunks don't include usage metadata. I have also added a regression test covering the dropped usage scenario.

@precision @appleboy @rspier @levigross

**Comment by SHAI-shivansh-sharma** [comment]
Hi, could one of the maintainers/contributors please review this PR when you get a chance? The changes are ready for review. Thanks!
@karolpiotrowicz

**Comment by SHAI-shivansh-sharma** [comment]
Hi @baptmont 
Can you please review the changes ?

**Comment by baptmont** [comment]
Thanks for chasing this one down — you filed #1248 *and* the fix, and the
report was accurate. **#1257 landed the same guard on 7 Aug**, two days after
you opened this, and closed #1248. Once `main` merged in, your production
change became a no-op and dropped out.

The diff is now a single file:

```
internal/llminternal/stream_aggregator_test.go  +86/-0
```

So the description no longer describes the change. It says *"This applies the
same nil-guard (keep-last-non-nil) pattern already used for GroundingMetadata
and CitationMetadata"* and *"Fixes the bug described in the issue"* — but
nothing is applied, and both new tests pass against unmodified `main`. Anyone
reviewing from the description would approve believing they were merging a fix.

That is worth correcting rather than closing, because one of the two tests
turns out to be the only thing in the suite holding a real property.

## Which of the two tests earns its place

I mutation-tested rather than reasoning from the names.

**Mutation 1 — drop the guard entirely** (`s.usageMetadata = llmResponse.UsageMetadata`):

| test | catches it |
| --- | --- |
| `TestUsageMetadataRetainedWhenLaterChunkHasNoUsageMetadata` (existing, from #1257) | yes |
| `TestUsageMetadataNotDroppedByLaterChunkWithNone` (new) | yes |
| `TestUsageMetadataUpdatedByLaterChunkWithValue` (new) | no |

**Mutation 2 — keep-*first* instead of keep-last**
(`if llmResponse.UsageMetadata != nil && s.usageMetadata == nil`):

| test | catches it |
| --- | --- |
| existing test | no |
| `TestUsageMetadataNotDroppedByLaterChunkWithNone` | no |
| **`TestUsageMetadataUpdatedByLaterChunkWithValue`** | **yes — the only test in the package** |

The value is the opposite way round from what the names suggest.

- **`TestUsageMetadataUpdatedByLaterChunkWithValue` is worth keeping.** Nothing
  else pins keep-*last*. Today `aggregateResponse` takes the newest non-nil
  usage, which is right for Gemini's cumulative counts, and if someone
  "hardened" that into keep-first to protect the first chunk, this is the only
  test that would go red.
- **`TestUsageMetadataNotDroppedByLaterChunkWithNone` is largely redundant.** It
  catches exactly the mutation the existing test already catches, and it sits
  directly beneath it under a near-identical name
  (`NotDroppedByLaterChunkWithNone` vs `RetainedWhenLaterChunkHasNoUsageMetadata`).
  It is not worthless — it is the only case where the usage-carrying chunk has
  **no candidates at all**, which is the Vertex-plus-grounding shape from #1248
  — but as written, two tests with the same name shape and the same coverage
  will read as an accident to the next person in this file.

Reproduce with:

```bash
# mutation 1
go test -count=1 -run UsageMetadata ./internal/llminternal/
# then re-apply the guard as `&& s.usageMetadata == nil` for mutation 2
```

## Suggested shape

- Retitle to something like
  `test(llminternal): pin keep-last semantics for streaming usage metadata`.
  The repo squash-merges, so a `fix:` subject with no production change lands on
  `main` and misreports the history.
- Rewrite the description: say #1257 landed the fix, and that this adds the
  coverage it did not.
- Keep `TestUsageMetadataUpdatedByLaterChunkWithValue` as-is.
- Either drop `TestUsageMetadataNotDroppedByLaterChunkWithNone`, or fold its
  no-candidates chunk into the existing `...RetainedWhenLaterChunkHasNo...` test
  as a second case, so the distinct shape survives without a duplicate test.

## Smaller things

- **CI has never actually run here.** `Go` and `API diff` are both sitting at
  `action_required` — the first-time-contributor approval gate. The only green
  checks are `check-changes` and the CLA, so "all unit tests pass locally" is
  currently the only evidence. (I ran the package locally: it is green, and the
  new file is `gofmt` clean.)
- **No linked issue.** The template placeholders are still literal
  (`Closes: #_issue_number_`), and everything below your written description is
  unedited boilerplate. #1248 is the issue, though it is already closed by #1257.
- **The checklist is ticked all the way down**, including "manually tested
  end-to-end" and "dependent changes merged and published in downstream
  modules". Neither applies to a test-only change; leaving them unticked with a
  one-line reason reads better than ticking them.
- **The branch is behind `main`** and has two `Merge branch 'main'` commits.
  A rebase would keep the history readable.

None of this is a reason to lose the work — the keep-last test is a genuine gap
you have closed. It just needs to say what it is.

SHAI-shivansh-sharma:
@baptmont Thanks for the detailed review and mutation testing. I’ve updated the PR to be test-only since #1257 already landed the production fix. I removed the production-code diff and the redundant test, and kept TestUsageMetadataUpdatedByLaterChunkWithValue to pin the keep-last semantics for streaming UsageMetadata. I also rebased onto main, updated the PR title/description, and verified the relevant tests locally.

SHAI-shivansh-sharma:
Hi @baptmont
Can you please review the changes ?

PR #1252, opened 2026-08-05
fix(session): give Event a consistent JSON encoding

### LINKED ISSUE #1251: session.Event JSON encoding is inconsistent and silently corrupts events exchanged with adk-python

## 🔴 Required Information
Describe the Bug:

session.Event has no coherent JSON encoding. Only two of its fields carry a json tag (IsolationScope, NodeInfo); neither model.LLMResponse nor session.EventActions carries any. Marshaling a minimal Event therefore produces 27 top-level keys in mixed casing — almost all PascalCase — of which 12 are null.

This became load-bearing when session/vertexai started writing a whole Event into the Vertex raw_event field, which is shared with adk-python. The two runtimes disagree about that payload in both directions:

Go → Python fails silently. adk-python's Event binds camelCase aliases, ignores unknown keys, and defaults every field. Every PascalCase key is dropped and model_validate_json still succeeds, returning a fabricated event — new random id, empty author/invocation_id, timestamp of "now", empty actions. No error anywhere.
Python → Go fails partially. adk-python writes timestamp as a float epoch, which time.Time rejects. The error aborts the decode part-way through, so every field ordered after the timestamp is left unset even though it is present in the payload.
Steps to Reproduce:

git clone https://github.com/google/adk-go && cd adk-go (at 046ae95)
Save the program under Minimal Reproduction Code as repro/main.go
go run ./repro
Observe the key casing/count, and the partially-populated event in step 2 of the output
Expected Behavior:

Marshaling an Event produces camelCase keys matching adk-python's dump (the contract server/adkrest/internal/models/event.go already defines), with empty fields omitted.
An event written by Go round-trips through adk-python with its id, author, invocation_id and timestamp intact.
An event written by adk-python decodes fully in Go, including fields after timestamp.
Observed Behavior:

1. Go writes 27 top-level keys, 12 of them null
   keys: [Actions Author AvgLogprobs Branch CitationMetadata Content CustomMetadata
          ErrorCode ErrorMessage FinishReason GroundingMetadata ID InputTranscription
          Interrupted InvocationID LogprobsResult LongRunningToolIDs ModelVersion Output
          OutputTranscription Partial RequestedInput Routes SessionResumptionHandle
          Timestamp TurnComplete UsageMetadata]

2. Reading an adk-python event: err = Time.UnmarshalJSON: input is not a JSON string
   ID="e-2" InvocationID="inv-2" Author="agent" Branch="" StateDelta=map[]
Note in (2) that Branch and StateDelta are empty despite both being present in the input JSON — the decode aborted at timestamp.

Feeding the Go output from (1) verbatim to adk-python Event.model_validate_json:

> VALIDATION SUCCEEDED. Resulting event:
>   id            = '53d607ea-a361-432e-9d58-4b73c1512b84'   # not 'e-1'
>   invocation_id = ''                                        # not 'inv-1'
>   author        = ''                                        # not 'agent'
>   timestamp  = 1785933555.1635234                        # "now", not the event's
>   branch        = None
>   content       = None
>   actions       = EventActions(state_delta={}, artifact_delta={}, ...)

Environment Details:

ADK Library Version: google.golang.org/adk/v2 @ main (046ae95); also present in v2.1.0
OS: Linux
Go Version: 1.26.5
Model Information: N/A — no model involved; this is pure serialization.

🟡 Optional Information
Regression: No. The encoding has been inconsistent since the fields were introduced; it only became reachable when whole-Event serialization arrived with session/vertexai in the 2.x line.

Minimal Reproduction Code:

```
package main

import (
	"encoding/json"
	"fmt"
	"sort"
	"time"

)

func main() {
	// 1. What Go writes.
	ev := &session.Event{
		ID: "e-1", InvocationID: "inv-1", Author: "agent",
		Timestamp: time.Unix(1733040000, 0).UTC(),
	}
	b, _ := json.Marshal(ev)
	var m map[string]json.RawMessage
	json.Unmarshal(b, &m)
	keys, nulls := make([]string, 0, len(m)), 0
	for k, v := range m {
		keys = append(keys, k)
		if string(v) == "null" {
			nulls++
		}
	}
	sort.Strings(keys)
	fmt.Printf("1. Go writes %d top-level keys, %d of them null\n   keys: %v\n\n",
		len(keys), nulls, keys)

	// 2. What adk-python writes, read back by Go.
	pythonEvent := `{
	  "id": "e-2", "invocationId": "inv-2", "author": "agent",
	  "timestamp": 1733040000.5, "branch": "main",
	  "actions": {"stateDelta": {"user:k": "v"}, "artifactDelta": {}}
	}`
	got := &session.Event{}
	err := json.Unmarshal([]byte(pythonEvent), got)
	fmt.Printf("2. Reading an adk-python event: err = %v\n", err)
	fmt.Printf("   ID=%q InvocationID=%q Author=%q Branch=%q StateDelta=%v\n",
		got.ID, got.InvocationID, got.Author, got.Branch, got.Actions.StateDelta)
}
```
How often has this issue occurred?: Always (100%) — deterministic.

Additional Context:

Fixing this changes the shape of what Go writes (camelCase, empties dropped: 27 keys → 5 for an empty event), so it touches the high-fan-in session and model packages and is a wire-format change for anyone reading persisted events by key name. plugin/agentanalytics, which marshals LLMResponse into BigQuery, is the known in-repo consumer.

Reading stays backward compatible — encoding/json matches keys case-insensitively, so previously written events still decode.

jjsasha63:
Link to Issue or Description of Change
1. Link to an existing issue (if applicable):

Closes: #1251 

Problem:

session.Event has no coherent JSON encoding — only 2 of its fields carry a json tag, and neither model.LLMResponse nor session.EventActions carries any. A minimal Event marshals to 27 mostly-PascalCase keys, 12 of them null.

Since session/vertexai writes a whole Event into the Vertex raw_event field, and that field is shared with adk-python, this silently corrupts data in both directions:

Go → Python: every PascalCase key is dropped by adk-python's extra='ignore', and model_validate_json still succeeds, returning a fabricated event (new random id, empty author/invocation_id, timestamp of "now").
Python → Go: the float epoch timestamp aborts the decode part-way, so every field after it is silently left unset.

Solution:

Tag Event, EventActions and LLMResponse with the camelCase contract that server/adkrest/internal/models/event.go already defines, and accept both timestamp dialects on read.

Three points where the obvious approach doesn't work:

stateDelta/artifactDelta need a custom MarshalJSON, not a tag. omitempty drops nil and allocated-empty alike, and NewEvent allocates both maps — an allocated-empty map decoding back as nil panics on the first write. A bare tag keeps them distinct but writes null, which adk-python rejects (both are non-Optional dict). Omitting the key only when nil satisfies both runtimes; omitempty on a pointer keys off the pointer, not the length.
Numeric timestamps resolve to microseconds, matching Python's datetime. float64 can't hold epoch seconds to ns precision. Out-of-range values are rejected rather than narrowed, since an out-of-range float64 saturates silently and loses its sign.
An all-zero NodeInfo decodes back to nil, because adk-python dumps a nodeInfo object on every event and readers here test the pointer, not the contents.

Reading stays backward compatible — encoding/json matches keys case-insensitively, so previously written events still decode.

Writing changes shape: camelCase keys, empties dropped, 27 keys → 5 for an empty event. plugin/agentanalytics (marshals LLMResponse into BigQuery) is the known in-repo consumer.
Testing Plan
Unit Tests:

I have added or updated unit tests for my change.
All unit tests pass locally.

New: session/event_json_test.go (385 lines) and session/vertexai/raw_event_crossruntime_test.go (181 lines), covering the golden key set, both timestamp dialects, out-of-range and boundary timestamps, NodeInfo nil-ness, and the delta-map encoding (nil / allocated-empty / populated / mixed).

Full workspace run, matching CI (go work init && go work use -r .):

> $ go build -mod=readonly work
> $ go test -race -mod=readonly -count=1 -shuffle=on work
> ok  google.golang.org/adk/v2/session          0.071s
> ok  google.golang.org/adk/v2/session/database 0.405s
> ok  google.golang.org/adk/v2/session/vertexai 1.475s
> ok  google.golang.org/adk/v2/model            ...
> ... all modules ok
> 
> $ golangci-lint run                       # root module
> 0 issues.
> $ cd plugin/agentanalytics && golangci-lint run
> 0 issues.
> 
> $ go mod tidy -diff                       # both modules, no output

Every behavioural change was checked with a negative control — revert the fix, confirm the new test fails with the expected diff, restore. Two examples:

Dropping EventActions.MarshalJSON → actions.stateDelta = null.
Replacing it with plain omitempty → StateDelta is nil after a round trip; writing to it would panic.

Manual End-to-End (E2E) Tests:

Per CONTRIBUTING.md ("we lean on adk-python for being the source of truth"), the encoding was validated against real adk-python rather than against my reading of its models. The JSON this branch emits was captured verbatim and fed to pydantic:

> Input (verbatim Go output)
> EventActions.model_validate_json
> {} (both maps nil)
> ✅ accepted, both default to {}
> {"stateDelta":{},"artifactDelta":{}}
> ✅ accepted
> {"stateDelta":{"user:k":"v"},"artifactDelta":{"f.png":7}}
> ✅ values preserved
> every field populated, incl. requestedToolConfirmations
> ✅ all values round-trip
> {"stateDelta":null} (what a bare tag would write)
> ❌ Input should be an object [type=dict_type]
> {"StateDelta":{}} (what main writes today)
> ❌ Extra inputs are not permitted [type=extra_forbidden]

Before/after on the headline case — the same Event marshaled by Go, read by adk-python:

> BEFORE (main):  id='53d607ea-…' (regenerated)  author=''  invocation_id=''
>                 timestamp=<now>                 actions=<all defaults>
>                 → validation SUCCEEDS, event silently blank
> 
> AFTER  (this):  id='e-1'  author='agent'  invocation_id='inv-1'
>                 → values preserved

Also verified there are no duplicate keys in the output (the MarshalJSON relies on encoding/json's depth-0-shadows-depth-1 rule) by parsing the raw bytes with a object_pairs_hook that rejects repeats.

Additional context
This touches the high-fan-in session and model packages and changes a persisted wire format, so per AGENTS.md Boundaries I filed #1251  first — happy to adjust the approach before review time is spent on the diff.

One note for reviewers: telemetry fails under -race in workspace mode on any commit before 046ae95 (conflicting Schema URL: …1.41.0 and …1.40.0, from dependency unification across the two modules). It is unrelated to this change and is fixed by the dependency bump in #1242; this branch is rebased onto it.

### DIFF
--- a/model/llm.go
+++ b/model/llm.go
@@ -40,29 +40,29 @@ type LLMRequest struct {
 // LLMResponse is the raw LLM response.
 // It provides the first candidate response from the model if available.
 type LLMResponse struct {
-	Content             *genai.Content
-	CitationMetadata    *genai.CitationMetadata
-	GroundingMetadata   *genai.GroundingMetadata
-	UsageMetadata       *genai.GenerateContentResponseUsageMetadata
-	CustomMetadata      map[string]any
-	LogprobsResult      *genai.LogprobsResult
-	InputTranscription  *genai.Transcription
-	OutputTranscription *genai.Transcription
-	ModelVersion        string
+	Content             *genai.Content                              `json:"content,omitempty"`
+	CitationMetadata    *genai.CitationMetadata                     `json:"citationMetadata,omitempty"`
+	GroundingMetadata   *genai.GroundingMetadata                    `json:"groundingMetadata,omitempty"`
+	UsageMetadata       *genai.GenerateContentResponseUsageMetadata `json:"usageMetadata,omitempty"`
+	CustomMetadata      map[string]any                              `json:"customMetadata,omitempty"`
+	LogprobsResult      *genai.LogprobsResult                       `json:"logprobsResult,omitempty"`
+	InputTranscription  *genai.Transcription                        `json:"inputTranscription,omitempty"`
+	OutputTranscription *genai.Transcription                        `json:"outputTranscription,omitempty"`
+	ModelVersion        string                                      `json:"modelVersion,omitempty"`
 	// Partial indicates whether the content is part of a unfinished content stream.
 	// Only used for streaming mode and when the content is plain text.
 	// The Runner fully processes only the final non-partial event, partial
 	// events are simply forwarded downstream (eg. to UI for display).
-	Partial bool
+	Partial bool `json:"partial,omitempty"`
 	// Indicates whether the response from the model is complete.
 	// Only used for streaming mode.
-	TurnComplete bool
+	TurnComplete bool `json:"turnComplete,omitempty"`
 	// Flag indicating that LLM was interrupted when generating the content.
 	// Usually it is due to user interruption during a bidi streaming.
-	Interrupted             bool
-	SessionResumptionHandle string
-	ErrorCode               string
-	ErrorMessage            string
-	FinishReason            genai.FinishReason
-	AvgLogprobs             float64
+	Interrupted             bool               `json:"interrupted,omitempty"`
+	SessionResumptionHandle string             `json:"sessionResumptionHandle,omitempty"`
+	ErrorCode               string             `json:"errorCode,omitempty"`
+	ErrorMessage            string             `json:"errorMessage,omitempty"`
+	FinishReason            genai.FinishReason `json:"finishReason,omitempty"`
+	AvgLogprobs             float64            `json:"avgLogprobs,omitempty"`
 }
--- a/session/event_json_test.go
+++ b/session/event_json_test.go
@@ -1,0 +1,480 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package session_test
+
+import (
+	"context"
+	"encoding/json"
+	"sort"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/genai"
+
+	"google.golang.org/adk/v2/model"
+	"google.golang.org/adk/v2/session"
+)
+
+func fullyPopulatedEvent() *session.Event {
+	return &session.Event{
+		LLMResponse: model.LLMResponse{
+			Content:      genai.NewContentFromText("hello", genai.RoleModel),
+			ModelVersion: "gemini-3-pro",
+			Partial:      true,
+			TurnComplete: true,
+			Interrupted:  true,
+			ErrorCode:    "429",
+			ErrorMessage: "rate limited",
+			AvgLogprobs:  -0.25,
+		},
+		ID:                 "evt-1",
+		Timestamp:          time.Unix(1733040000, 0).UTC(),
+		InvocationID:       "inv-1",
+		Branch:             "root.planner",
+		IsolationScope:     "approval_subflow",
+		Author:             "planner",
+		LongRunningToolIDs: []string{"tool-1", "tool-2"},
+		Routes:             []string{"reviewer", "writer"},
+		RequestedInput: &session.RequestInput{
+			InterruptID: "approval-1",
+			Message:     "Approve?",
+			Payload:     map[string]any{"doc": "contract.pdf"},
+		},
+		Output:   map[string]any{"summary": "ok", "risk_items": float64(3)},
+		NodeInfo: &session.NodeInfo{Path: "wf/planner", MessageAsOutput: true, OutputFor: []string{"wf"}},
+		Actions: session.EventActions{
+			StateDelta:        map[string]any{"stage": "drafted"},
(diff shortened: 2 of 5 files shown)


odsod:
@jjsasha63 this is a pain point for us as well - have you seen this? https://github.com/google/adk-go/blob/main/server/adkrest/internal/models/event.go - we currently copy the code from here into our application - would it make sense for your PR to consolidate the JSON conversions?

jjsasha63:
Thanks — hadn't looked at `adkrest/internal/models` closely. Agreed it's the same problem solved twice.

I compared the wire formats: every field in `models.Event` exists on `session.Event` with an identical JSON tag name, so it's a strict subset. `session.Event` emits five more — `timestamp`, `citationMetadata`, `customMetadata`, `logprobsResult`, `sessionResumptionHandle` — so consolidating would be additive for adkrest clients rather than breaking. `timestamp` being absent today looks like an oversight.

I'd rather not fold it into this PR: it changes the shape of a public REST response, and that feels like a call for the adkrest owners rather than something to land inside a session-encoding fix. If an owner is happy for those five fields to appear, I'll send the follow-up straight after this — it's deleting `models.Event`/`ToSessionEvent`/`FromSessionEvent` and updating 4 call sites in `controllers/`.

For your immediate case you shouldn't need to wait for any of that. Once this lands, `session.Event` round-trips on its own, so you can marshal it directly and drop the copied code.

odsod:
@jjsasha63 Sounds good to me! Your PR solves our problem and I'm all for getting this shipped sooner rather than later :relaxed:

baptmont (approved):

PR #1256, opened 2026-08-05
fix(session): refresh stale database handles

### LINKED ISSUE #1229: session/database: no way for a second writer to append without invalidating the runner's session handle (write-lease OCC)
## Summary

`session/database`'s `AppendEvent` enforces optimistic concurrency on `last_update_time` (`service.go` ~L376 in v2.1.0): an append through a handle older than the row fails with `stale session error`. That makes every session handle an implicit exclusive write lease — any out-of-band append invalidates every other holder's handle, including a runner mid-turn, whose next streamed event append then fails and kills the turn.

The check itself is reasonable OCC. The problem is there's no affordance for a *legitimate* second writer, and the semantic is undocumented (discovered empirically — see also #539, which asks for exactly this kind of internals documentation).

## Why this matters

Two independent downstream projects hit this in production-shaped code and converged on the same workaround:

- **[go-steer/core-agent](https://github.com/go-steer/core-agent)**: a subagent runner writing into the parent's session row invalidated the parent runner's handle mid-stream. Documented in [`docs/eventlog-decisions.md`](https://github.com/go-steer/core-agent/blob/main/docs/eventlog-decisions.md) ("Subagent runs in a derived session row, not the parent's"); fixed with derived session IDs (`<parent>:sub:<branch>`).
- **[go-steer/mast](https://github.com/go-steer/mast)**: a daemon appending operator/audit markers (abort, shutdown-interruption) to a session with a live turn killed that turn; worse, the dying turn's cleanup path then erased the marker. Fixed the same way — a companion row (`<sid>:mast-ops`) — in [go-steer/mast#49](https://github.com/go-steer/mast/pull/49); constraint written up in [`docs/durable-execution-design.md`](https://github.com/go-steer/mast/blob/main/docs/durable-execution-design.md).

The derived-row workaround works but has real costs: one logical session fragments across rows, listing surfaces phantom rows unless callers know the suffix scheme, queries need two lookups, and the suffix conventions are project-specific — which erodes the cross-runtime session compatibility story (a Python ADK reader has no idea `:sub:` or `:mast-ops` rows belong to their primaries).

## Repro

```go
svc, _ := database.NewSessionService(sqlite.Open("s.db")) // AutoMigrate'd
a, _ := svc.Get(ctx, &session.GetRequest{AppName: app, UserID: u, SessionID: sid})
b, _ := svc.Get(ctx, &session.GetRequest{AppName: app, UserID: u, SessionID: sid})
_ = svc.AppendEvent(ctx, b.Session, evB)      // ok — advances last_update_time
err := svc.AppendEvent(ctx, a.Session, evA)   // "stale session error: ..."
```

Note the in-memory service has no such check, so `Service` implementations diverge behaviorally — tests that pass in-memory fail on the database service, which is how both projects shipped the bug before catching it.

## Request

Any one of these would remove the need for derived-row conventions (in rough preference order):

1. **Opt-in refetch-and-retry on stale** — e.g. an `AppendEvent` option or service-level option that, on the stale check, refetches the row and retries the append. Events are append-only and `StateDelta` is last-write-wins, so the retry semantics are well-defined.
2. **A cheap handle-refresh API** — `Refresh(ctx, session)` (or `Get` documented as the sanctioned revalidation) so long-lived holders can revalidate after a known out-of-band write, and runners could retry-once on stale.
3. **At minimum, document the write-lease semantic** on `Service.AppendEvent` — including the in-memory/database divergence — so downstreams design for it up front instead of discovering it as a production incident.

Related: #539 (documenting session internals), #1170 (the stale error's timestamp formatting, which makes the failure harder to diagnose when it does fire).

Soundcreates:
Fixes #1229

Refresh and retry once when optimistic concurrency detects a stale session handle. Local handle state is only mutated after commit and newest session timestamps are preserved. Added a two writer regression test covering event and state delta persistence.

Tests:
* go test ./session/database
* go test ./session/...
* git diff --check


AmirF194 (commented):
Good catch on the OCC retry gap, and nice touch keeping the local handle update strictly post-commit.

One question on `refreshSession` (session/database/service.go:372-388): it replaces `sess.state` and `sess.events` with the pointers from the freshly-fetched `current` session, rather than merging into the existing maps the way every other write path here does (`updateSessionState` uses `maps.Copy(sess.state, ...)` at session.go:188, in place).

`localSession.State()` (session.go:53-58) hands out a `*state{mu: &s.mu, state: s.state}` that captures the map reference at call time. If a caller holds onto that handle across an `AppendEvent` (a normal pattern: grab `Session().State()` once, do a few `Get`/`Set` calls over the course of an operation), a refresh triggered by a concurrent writer swaps `sess.state` to a new map object underneath it. The held handle keeps pointing at the old map: further `Set` calls land there and are invisible to `Close`/persistence, and further `Get` calls never see the second writer's data even though `mu` is shared and there's no data race.

I checked every in-repo caller (agent/common_context.go, workflow/function_node.go, the plugin/conformance packages) and they all call `Session().State()` fresh immediately before each access, so nothing here hits it today. But `session.Session` is a public interface and this is the one write path in the file that doesn't preserve the "old handles stay live" contract the rest of the type relies on. Was swapping the pointer instead of merging deliberate (e.g. to also drop entries the local session added but the other writer never saw), or would `maps.Copy` here also close the gap?

karolpiotrowicz (changes requested):
The retry-once-on-stale idea is the right shape for [#1229](https://github.com/google/adk-go/issues/1229), and two of the incidental changes here are real improvements: guarding `storageSess.UpdateTime` so an out-of-order event can no longer rewind the row's clock, and moving the local append after the commit so a failed persist no longer leaves a phantom event on the handle. Two things need fixing before this can land, though, and there is an ordering question with another open PR.

Checked out at [67927684](https://github.com/google/adk-go/commit/67927684caba45e176ce72a550ca18932df7103a) against `main` at [160a3976](https://github.com/google/adk-go/commit/160a39763e901047505abcc1acc32707edc0fab5). Worth knowing up front: the branch no longer merges — both changed files conflict — and the green checks on this head are from 15 August, which predates every commit now on `main` in this package.

## `temp:` state stops reaching the live session handle

[`service.go:343`](https://github.com/google/adk-go/blob/67927684caba45e176ce72a550ca18932df7103a/session/database/service.go#L343) now rebinds `event` to the trimmed copy before [`service.go:352`](https://github.com/google/adk-go/blob/67927684caba45e176ce72a550ca18932df7103a/session/database/service.go#L352) passes it to `sess.appendEvent`, so `updateSessionState` never sees the `temp:` keys and they never land in the in-memory session state. Previously `appendEvent` received the untrimmed event, and the other two backends still do it that way deliberately — [`session/inmemory.go:360-363`](https://github.com/google/adk-go/blob/67927684caba45e176ce72a550ca18932df7103a/session/inmemory.go#L360-L363) and [`session/vertexai/session.go:79-82`](https://github.com/google/adk-go/blob/67927684caba45e176ce72a550ca18932df7103a/session/vertexai/session.go#L79-L82) both call `updateSessionState` with the raw event and trim only the copy they store.

This breaks a documented handoff. `temp:` is an accepted instruction-template prefix, and [`instruction_processor.go:150`](https://github.com/google/adk-go/blob/67927684caba45e176ce72a550ca18932df7103a/internal/llminternal/instruction_processor.go#L150) resolves `{temp:x}` against the raw session state rather than the callback delta, so a non-optional placeholder returns an error when the key is absent. The shipped sequential-agents example relies on exactly that: [`main.go:84`](https://github.com/google/adk-go/blob/67927684caba45e176ce72a550ca18932df7103a/examples/workflowagents/sequentialCode/main.go#L84) writes `OutputKey: "temp:review_comments"` and [`main.go:103`](https://github.com/google/adk-go/blob/67927684caba45e176ce72a550ca18932df7103a/examples/workflowagents/sequentialCode/main.go#L103) reads `{temp:review_comments}` in the next agent's instruction. On the database backend that read goes from working to hard-failing.

`TestDatabaseService_AppendEvent_PreservesInputEventTempState` does not catch it, because it asserts on the caller's input event and on the stored event, never on `Session().State()`.

<details><summary>Repro — passes on the merge-base, fails on this head</summary>

```go
// session/database/service_test.go
func TestTempStateReachesLocalHandle(t *testing.T) {
	t0 := time.Date(2026, time.November, 30, 0, 0, 0, 0, time.UTC)
	ctx := platform.WithTimeProvider(t.Context(), func() time.Time { return t0 })
	s := emptyService(t)
	created, err := s.Create(ctx, &session.CreateRequest{AppName: "app", UserID: "user", SessionID: "sess"})
	if err != nil {
		t.Fatalf("Create: %v", err)
	}
	ev := &session.Event{ID: "e1", Timestamp: t0.Add(time.Second),
		Actions: session.EventActions{StateDelta: map[string]any{"temp:k": "v", "plain": "p"}}}
	if err := s.AppendEvent(ctx, created.Session, ev); err != nil {
		t.Fatalf("AppendEvent: %v", err)
	}
	if _, err := created.Session.State().Get("temp:k"); err != nil {
		t.Errorf("temp: key not visible on the live handle: %v", err)
	}
}
```

```
$ go test -run TestTempStateReachesLocalHandle ./session/database/
merge-base:  --- PASS
this head:   --- FAIL
  temp: key not visible on the live handle: state key does not exist
```

</details>

Keeping a separate variable for the trimmed event and passing only that to `applyEvent` fixes it. I ran that change and the probe above plus the whole `session/database` suite pass with it, including your new test.

## The retry changes a contract `main` has a test for

`TestDatabaseService_AppendEvent_StaleErrorFormatsTimestamps` on `main` asserts that an append through a stale handle returns an error ([`service_test.go:132-139`](https://github.com/google/adk-go/blob/160a39763e901047505abcc1acc32707edc0fab5/session/database/service_test.go#L132-L139)). With this change it returns `nil`, because the refresh sets the handle's `updatedAt` to the stored value and the retry then passes the check. Resolving the conflicts as favourably as I could — keeping `main`'s `time.UnixMicro` formatting fix and your `%w` wrapping, and keeping every test from both sides — the package is red:

```
--- PASS: TestDatabaseService_AppendEvent_RefreshesStaleHandle
--- FAIL: TestDatabaseService_AppendEvent_StaleErrorFormatsTimestamps
    service_test.go:135: AppendEvent through stale handle: got nil error, want stale session error
FAIL    google.golang.org/adk/v2/session/database
```

There is a second, independent break in the same test: `errStaleSession` is `errors.New("stale session")`, so the message no longer contains the string `"stale session error"` that line 138 checks for. Fixing the `nil` assertion alone still leaves it red.

This is worth more than a rebase, because #1229 asked for the retry to be **opt-in** — an `AppendEvent` option or a service option — precisely so the existing rejection stays the default. Making it unconditional silently converts a detected conflict into a last-writer-wins merge for every caller, and a caller that wants to recompute its event against the newer state now has no way to find out that it should. Either gate the retry behind an option, or change that test deliberately in this PR with the reasoning written down.

## Overlap with #1244

As it stands, `refreshSession` replaces the session's whole state map at [`service.go:388`](https://github.com/google/adk-go/blob/67927684caba45e176ce72a550ca18932df7103a/session/database/service.go#L388), while `State()` captures that map when it is called. A `session.State` obtained before a refresh is therefore detached afterwards: it cannot see refreshed keys and writes through it go into the abandoned map. Under `-race`, that swap and the matching `sess.events` swap on the next line each race the unlocked reads in `State()` and `Events()`.

[#1244](https://github.com/google/adk-go/pull/1244) already restructures exactly this: it makes `state` hold the session rather than the map, and takes the read lock in `Events()`. I applied its `session.go` changes on top of this branch and both problems disappear — the detached-view repro passes and the race count drops to zero, with no change needed here. So this is a sequencing question rather than work for you: if #1244 lands first, this PR is unaffected, and if this one lands first the swap needs to become a `clear` plus `maps.Copy` so outstanding views stay attached. The two PRs touch the same functions and will conflict either way.

## Smaller things

- **Event order on the handle after a retry.** The refresh installs the database's chronological list and the retried event is then appended at the tail regardless of its timestamp. Using your own test's fixture, the handle ends up `[event-b, event-a]` while the database returns `[event-a, event-b]`. Readers that walk `Events()` in order see the inversion until the next `Get`. Your new test covers this scenario but asserts only `Events().Len() == 2`.
- **The stale error is dropped when the refresh fails.** [`service.go:365-367`](https://github.com/google/adk-go/blob/67927684caba45e176ce72a550ca18932df7103a/session/database/service.go#L365-L367) wraps the `Get` error only, so the two disagreeing timestamps that made the append fail are lost from the message.
- **`return nil // unreachable`** at [`service.go:369`](https://github.com/google/adk-go/blob/67927684caba45e176ce72a550ca18932df7103a/session/database/service.go#L369) reports success. It is genuinely unreachable today, but if the attempt bound ever changes it would report success for an append that never committed.

karolpiotrowicz:
@wolo-lab flagging this one for a posture call rather than a code review.

adk-python raises `StaleSessionError` when a session handle goes stale and leaves reconciliation to the caller. This PR silently refreshes and retries instead, so a write can land against a state the caller never read — a different contract for `AppendEvent` than the sibling implementation offers. Worth deciding which posture ADK Go wants before the author reworks it.

Soundcreates:
Thanks for digging this deep into it, this is a lot to work through. Understood on all four points:

1. the temp: trimming happening before appendEvent/updateSessionState instead of after, so those keys never reach the live session state
2. the retry needs to be opt in (not the default), since main already has a test asserting stale appends fail, and the errStaleSession message needs to actually contain "stale session error"
3. noted on the #1244 overlap, sounds like that's more of a sequencing thing than something I need to fix here
4. the smaller stuff, event ordering after a retry, the dropped stale error details on refresh failure, and the unreachable return nil

Going to work through these and also rebase since the branch doesn't merge cleanly anymore. Will comment again once it's done.

Soundcreates:
Alright, made the changes.

1. temp: state - the trim was happening on the shared event variable before it got passed down to sess.appendEvent, so updateSessionState never saw those keys. Now the original event still goes to sess.appendEvent so the live handle gets the full state, and only a separate trimmed copy goes to applyEvent for persisting. Added your repro test (temp: key reaching Session().State() after AppendEvent).

2. retry as opt in - added EnableStaleRetry(service) following the same pattern as AutoMigrate, a type assert plus a flag on databaseService. Default is off, so AppendEvent on a stale handle still returns errStaleSession like main expects, TestDatabaseService_AppendEvent_StaleErrorFormatsTimestamps passes untouched. Also fixed the error text back to "stale session error" and kept main's time.UnixMicro formatting (the branch had drifted back to the old time.Unix bug during the merge). Added a test that checks the retry only kicks in once EnableStaleRetry is called.

3. left the #1244 overlap alone since you said that's a sequencing thing, not something for this PR.

4. smaller stuff - wrapped the original stale error together with the refresh failure instead of dropping it, and the retried event now gets inserted into the local handle's event list by timestamp instead of just appended at the tail (fixes the ordering you flagged, used sort.Search for the insertion point). Left the "return nil unreachable" thing but made it actually return the last error now rather than faking success, in case the attempt bound ever changes.

Rebased on main too since it wasn't merging cleanly anymore, and go test ./... is green including -race on session/database.

Let me know if the EnableStaleRetry shape works for you or if you had something else in mind, happy to adjust. Would appreciate another pass when you have time.

session/database/session.go
```diff
@@ -81,7 +82,19 @@ func (s *localSession) appendEvent(event *session.Event) error {
...
+	// retried event's timestamp can sort earlier than the tail the refresh just
+	// installed. sort.Search finds the first index whose timestamp is strictly
+	// after processedEvent's, so it inserts there rather than blindly at the
+	// tail -- and for ties it lands after every existing equal-timestamp event,
+	// which keeps a plain append's tie-breaking behavior unchanged.
+	idx := sort.Search(len(s.events), func(i int) bool {
+		return s.events[i].Timestamp.After(processedEvent.Timestamp)
+	})
+	s.events = append(s.events, nil)
+	copy(s.events[idx+1:], s.events[idx:])
```
karolpiotrowicz, line 96: This `copy` shifts elements inside the backing array, and `Events()` hands that same array out without copying it. So an `events` value a caller already holds can be rewritten underneath them by a later out-of-order append, whenever the slice has spare capacity. A plain append only ever wrote past the old length, so a returned view used to be stable for its lifetime.

Single-threaded reproduction, no concurrency needed. Seed three events so the slice reaches `len=3, cap=4`, hold `evs := sess.Events()`, then append an event whose timestamp sorts between the first and second:

```
retained view before: [e0 e1 e2]
retained view after:  [e0 mid e1]
live handle:          [e0 mid e1 e2]
```

The caller's view lost `e2` and gained an event it never asked about. On the merge-base the same probe leaves the retained view unchanged.

Building the new slice and publishing it in one assignment avoids this, and also means the field is never observable in a half-built state:

```go
next := make([]*session.Event, len(s.events)+1)
copy(next, s.events[:idx])
next[idx] = processedEvent
copy(next[idx+1:], s.events[idx:])
s.events = next
```

session/database/session.go
```diff
@@ -81,7 +82,19 @@ func (s *localSession) appendEvent(event *session.Event) error {
...
 	processedEvent := trimTempDeltaState(event)
-	s.events = append(s.events, processedEvent)
+	// s.events is chronological. A plain append is correct for the common case
+	// (each new event's timestamp is >= the last), but after an OCC refresh the
+	// retried event's timestamp can sort earlier than the tail the refresh just
+	// installed. sort.Search finds the first index whose timestamp is strictly
+	// after processedEvent's, so it inserts there rather than blindly at the
+	// tail -- and for ties it lands after every existing equal-timestamp event,
+	// which keeps a plain append's tie-breaking behavior unchanged.
+	idx := sort.Search(len(s.events), func(i int) bool {
```
karolpiotrowicz, line 92: Worth knowing that this insert changes ordering for callers who never opt into the retry, not just on the refresh path.

`AppendEvent` fills in a missing `ID` but not a missing `Timestamp` — it only truncates, and a truncated zero is still zero. So a struct-literal event built without a timestamp makes `sort.Search` return 0 and lands at the head of the conversation. On the merge-base it went to the tail.

To be fair to the change, the new placement is the more correct one. `Get` returns events ordered by timestamp, so I measured what each side does with such an event:

```
this branch:  handle = [no-timestamp real0 real1]   database = [no-timestamp real0 real1]
merge-base:   handle = [real0 real1 no-timestamp]   database = [no-timestamp real0 real1]
```

The live handle now agrees with what a later `Get` returns, which it did not before. So this is a fix rather than a regression — but it is a visible default-path behavior change that the description does not mention, and a caller relying on arrival order would notice. Rejecting or defaulting a zero `Timestamp` in `AppendEvent` would also remove the surprise at source.

session/database/service.go
```diff
@@ -375,21 +400,64 @@ func (s *databaseService) AppendEvent(ctx context.Context, curSession session.Se
...
+}
 
-	// Trim temp state before persisting
-	event = trimTempDeltaState(event)
-	// applyChanges and persist them
-	err := s.applyEvent(ctx, sess, event)
+// refreshSession replaces the mutable contents of a local handle with the
+// current database snapshot after another writer advances its OCC lease.
+func (s *databaseService) refreshSession(ctx context.Context, sess *localSession) error {
+	resp, err := s.Get(ctx, &session.GetRequest{
```
karolpiotrowicz, line 447: This `Get` drops the filters the handle was created with. `GetRequest` carries `After` and `NumRecentEvents`, and `Get` honors both, so a handle deliberately scoped to a small window comes back from the refresh holding everything.

Measured: a handle obtained with `NumRecentEvents: 2` held 8 events after a single refresh. Nothing is lost or corrupted, but a caller who bounded the window to keep a prompt small silently stops being bounded, and the events feed straight into the model request.

Carrying the originating request's `After` and `NumRecentEvents` onto the handle and replaying them here would keep the refresh faithful to what the caller asked for.

session/database/service.go
```diff
@@ -82,6 +91,22 @@ func AutoMigrate(service session.Service) error {
...
+//
+// NOTE: This function relies on a type assertion to the concrete
+// *databaseService implementation. It will return an error if the provided
+// session.Service is a different implementation.
+func EnableStaleRetry(service session.Service) error {
+	dbservice, ok := service.(*databaseService)
+	if !ok {
+		return fmt.Errorf("invalid session service type")
+	}
+	dbservice.retryOnStale = true
```
karolpiotrowicz, line 106: Two rough edges on this as a public API, neither blocking.

This is a plain write to a field that every `AppendEvent` reads unsynchronized at [`service.go:410`](https://github.com/google/adk-go/blob/a31818532354b355cdad46c633e5c6e8a7fa6bd2/session/database/service.go#L410). Nothing in the signature or the doc comment tells a caller it must happen before the service starts serving, and because the parameter is `session.Service` the caller cannot even see that it is a field write. One sentence in the doc comment saying to call it before the service is shared would close that.

The type assertion also means a wrapped or decorated `session.Service` — telemetry, auth, a test double — cannot use this at all. I note `AutoMigrate` has the same shape and the same caveat, so this is consistent with the package rather than novel. A constructor option would sidestep both points, if you would rather not add another free function that only works on the concrete type.

karolpiotrowicz (commented):
The two things flagged last round are both fixed on this head, and I could verify each rather than take it on trust. The retry is genuinely opt-in — the default path still fails with the same stale error it always did — and `temp:` state now reaches the live handle again, with a regression test guarding it. Merging the branch against current `main` also leaves `session/database` green, including `main`'s own `TestDatabaseService_AppendEvent_StaleErrorFormatsTimestamps`. Thank you for turning both around.

One new problem needs fixing before this goes in, and it is on the refresh path itself.

## The refresh throws away state the database never had

[`service.go:458`](https://github.com/google/adk-go/blob/a31818532354b355cdad46c633e5c6e8a7fa6bd2/session/database/service.go#L458) replaces the handle's whole state map with a fresh snapshot from the database:

```go
sess.state = current.state
```

`trimTempDeltaState` strips `temp:`-prefixed keys before anything is persisted, so the handle is the only place those keys ever live and a database snapshot can never carry them back. Every `temp:` key from an earlier event in the session is dropped by the refresh. The same assignment also detaches any `session.State` the caller is already holding, because [`State()`](https://github.com/google/adk-go/blob/a31818532354b355cdad46c633e5c6e8a7fa6bd2/session/database/session.go#L54-L59) captures the map reference at call time — that caller then writes into an orphaned map, and `Set` returns `nil`, so nothing signals the loss.

This matters because it can cost the caller exactly the turn the retry was rescuing. `replaceMatch` in `internal/llminternal/instruction_processor.go` resolves `{temp:x}` placeholders through `ctx.Session().State().Get`, and for a placeholder without the `?` suffix a missing key returns an error that propagates out and fails the turn. `examples/workflowagents/sequentialCode/main.go` ships that exact shape, with `OutputKey: "temp:review_comments"` feeding a `{temp:review_comments}` instruction. So a runner that opts in to survive a second writer can instead die on a placeholder it can no longer resolve.

It reproduces deterministically, with no concurrency involved:

<details>
<summary>Reproduction — drop into <code>session/database</code> and run <code>go test -run TestTempStateLostOnRefresh ./session/database/</code></summary>

```go
func TestTempStateLostOnRefresh(t *testing.T) {
	t0 := time.Date(2026, time.November, 30, 0, 0, 0, 0, time.UTC)
	ctx := platform.WithTimeProvider(t.Context(), func() time.Time { return t0 })
	s := emptyService(t)
	if err := EnableStaleRetry(s); err != nil {
		t.Fatalf("EnableStaleRetry: %v", err)
	}
	if _, err := s.Create(ctx, &session.CreateRequest{AppName: "app", UserID: "u", SessionID: "sid"}); err != nil {
		t.Fatalf("Create: %v", err)
	}
	get := func() session.Session {
		r, err := s.Get(ctx, &session.GetRequest{AppName: "app", UserID: "u", SessionID: "sid"})
		if err != nil {
			t.Fatalf("Get: %v", err)
		}
		return r.Session
	}
	base := time.Date(2026, time.December, 1, 0, 0, 0, 0, time.UTC)
	a := get()

	// The runner's handle picks up a temp: key from one of its own events.
	if err := s.AppendEvent(ctx, a, &session.Event{
		ID: "a1", Timestamp: base.Add(time.Second),
		Actions: session.EventActions{StateDelta: map[string]any{"temp:user_name": "ada"}},
	}); err != nil {
		t.Fatalf("AppendEvent(a1): %v", err)
	}
	if got, err := a.State().Get("temp:user_name"); err != nil || got != "ada" {
		t.Fatalf("precondition: got %v, err %v", got, err)
	}

	// A legitimate second writer advances the row — the scenario in #1229.
	if err := s.AppendEvent(ctx, get(), &session.Event{ID: "b1", Timestamp: base.Add(2 * time.Second)}); err != nil {
		t.Fatalf("AppendEvent(b1): %v", err)
	}

	// The handle is stale, so this append refreshes and retries. It succeeds...
	if err := s.AppendEvent(ctx, a, &session.Event{ID: "a2", Timestamp: base.Add(3 * time.Second)}); err != nil {
		t.Fatalf("AppendEvent(a2): %v", err)
	}
	// ...and the temp: key is gone.
	if _, err := a.State().Get("temp:user_name"); err != nil {
		t.Errorf("temp:user_name lost after the refresh: %v", err)
	}
}
```

</details>

Refreshing the map in place instead of rebinding it fixes both halves at once. I ran this and it flips the reproduction above to passing, keeps the map identity stable so held `State()` values stay attached, and leaves the package suite green:

```go
maps.DeleteFunc(sess.state, func(k string, _ any) bool {
	return !strings.HasPrefix(k, session.KeyPrefixTemp)
})
maps.Copy(sess.state, current.state)
```

One case that fix deliberately does not settle, because I do not think there is one right answer and it is your call: a value written through `State().Set` that no event has carried yet is still discarded by the refresh. Treating the row as authoritative and dropping it is defensible, and so is preserving it. Worth a decision either way, since today it happens silently.

## Does this close #1229?

Yes, on the letter of it. The issue offers three options and says any one of them would do, and option 1 is here and works — I confirmed the issue's own reproduction fails on the merge-base, still fails on this branch by default, and succeeds once `EnableStaleRetry` is called. The missing refresh API and the undocumented write-lease semantics are follow-ups rather than gaps in this change. Fixing line 458 is what turns "closes it nominally" into "closes it cleanly".

## Smaller things

These are on the diff as line comments: the ordered insert can rewrite an `Events()` value a caller already holds, an event with no `Timestamp` now sorts to the front of the conversation, the refresh drops the `NumRecentEvents` window the handle was created with, and `EnableStaleRetry` has a couple of rough edges as an API.

## Two notes that are not code

The branch conflicts with `main` and needs a rebase. It is a single hunk — `main` renamed `applyEvent`'s `session` parameter to `sess` and kept the `sess.updatedAt = storageSess.UpdateTime` line this change removes. Resolved in your favour it compiles and passes.

The PR description reads as though the retry is now the default: "Refresh and retry once when optimistic concurrency detects a stale session handle." The change is opt-in, which is the part reviewers most need to see. Worth saying so in the description.

Separately, and not something to fix here: the optimistic-concurrency check in `applyEvent` reads the row, compares timestamps in Go, and then saves, which is not atomic against a concurrent transaction under READ COMMITTED. That pattern predates this change and I could not exercise it — the in-memory SQLite used by the tests rejects concurrent writers outright — so I mention it only as something worth its own issue.

Soundcreates:
Thanks for the careful follow-up review. I went through the new comments and pushed another update.

The merge conflict with current `main` is fixed now, and the PR is mergeable again. I also fixed the refresh issue you called out: the existing state map is updated in place, so held `State()` values stay attached, and local `temp:` keys survive the refresh. I added a regression check for that case.

I also kept the event ordering and opt-in retry changes from the previous round. The database package passes `go test -race ./session/database`.

I’ve noted the remaining non-blocking points about the retained event slice, zero timestamps, request filters, and the `EnableStaleRetry` API. Thanks for catching those. I’ll leave the broader API and concurrency questions for follow-up unless you think any of them should block this PR.

Soundcreates:
Just to make this explicit, the branch has been updated with the requested changes and rebased onto `main` to clear the merge conflicts.

The refresh path now keeps the existing state map attached, preserves local `temp:` values, and has a regression test for it. The focused database race test is passing too. Thanks again for the detailed review.

PR #1257, opened 2026-08-05
fix(llm): retain streaming usage metadata

### LINKED ISSUE #1248: streamingResponseAggregator drops real UsageMetadata when a later chunk in the same turn carries none

Soundcreates:
Fixes #1248

Retain non nil UsageMetadata across streaming chunks so a later chunk without usage cannot erase usage reported earlier. Added regression coverage.

Tests:
* go test ./internal/llminternal/...

### DIFF
--- a/internal/llminternal/stream_aggregator.go
+++ b/internal/llminternal/stream_aggregator.go
@@ -78,7 +78,9 @@ func (s *streamingResponseAggregator) ProcessResponse(ctx context.Context, genRe
 
 func (s *streamingResponseAggregator) aggregateResponse(llmResponse *model.LLMResponse) *model.LLMResponse {
 	s.response = llmResponse
-	s.usageMetadata = llmResponse.UsageMetadata
+	if llmResponse.UsageMetadata != nil {
+		s.usageMetadata = llmResponse.UsageMetadata
+	}
 	if llmResponse.GroundingMetadata != nil {
 		s.groundingMetadata = llmResponse.GroundingMetadata
 	}
--- a/internal/llminternal/stream_aggregator_test.go
+++ b/internal/llminternal/stream_aggregator_test.go
@@ -1104,6 +1104,50 @@ func TestMetadataOnlyChunkDoesNotAbortStream(t *testing.T) {
 	}
 }
 
+func TestUsageMetadataRetainedWhenLaterChunkHasNoUsageMetadata(t *testing.T) {
+	aggregator := llminternal.NewStreamingResponseAggregator()
+	ctx := t.Context()
+
+	usageMetadata := &genai.GenerateContentResponseUsageMetadata{
+		PromptTokenCount:     3,
+		CandidatesTokenCount: 5,
+		TotalTokenCount:      8,
+	}
+	chunks := []*genai.GenerateContentResponse{
+		{
+			Candidates:    []*genai.Candidate{{Content: genai.NewContentFromText("Hello", "model")}},
+			UsageMetadata: usageMetadata,
+		},
+		{
+			Candidates: []*genai.Candidate{{
+				Content:      genai.NewContentFromText(" world", "model"),
+				FinishReason: genai.FinishReasonStop,
+			}},
+		},
+	}
+
+	for _, chunk := range chunks {
+		for _, err := range aggregator.ProcessResponse(ctx, chunk) {
+			if err != nil {
+				t.Fatalf("unexpected error processing chunk: %v", err)
+			}
+		}
+	}
+
+	finalResponse := aggregator.Close()
+	if finalResponse == nil {
+		t.Fatal("expected a final aggregated response, got nil")
+	}
+	if finalResponse.UsageMetadata == nil {
+		t.Fatal("expected usage metadata in the final aggregated response, got nil")
+	}
+	if finalResponse.UsageMetadata.PromptTokenCount != usageMetadata.PromptTokenCount ||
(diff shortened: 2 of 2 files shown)


indurireddy-TF:
Hi @baptmont ,I tested the changes locally, and the fix is working as expected. I also ran the tests, and everything is passing.
image

Could you please review the PR when you have a chance? Thanks!

baptmont (approved):

PR #1258, opened 2026-08-05
fix(workflow): propagate external context cancellation

### LINKED ISSUE #1253: workflow: external context cancellation is misclassified as sibling cancellation, so Workflow.Run ends as a clean success
**Describe the bug**

`scheduler.handleCompletion` classifies every node completion whose error is `context.Canceled` as a benign sibling cancellation and drops it:

```go
if errors.Is(it.err, context.Canceled) {
    ns.Status = NodeCancelled
    return nil // sibling cancellation; not the original error
}
```

(workflow/scheduler.go, in `handleCompletion`)

That classification is only correct when the scheduler itself cancelled the node — `cancelAll` cancelling the siblings of a failed node, whose own error surfaces separately. But per-node contexts inherit from the workflow's invocation context (`startNode` derives them from `parentCtx`), so an **external** cancellation — the caller cancelling its context, an HTTP request timeout, a client disconnect — completes every node with the same `context.Canceled`. In that case there is no failing sibling whose error will surface: `NodeCancelled` is never inspected again, the consumer loop drains, and `Workflow.Run`'s iterator simply ends with no error.

The impact is worst on the runner path: `Runner.Run` drives a root `LlmAgent` through a synthetic single-node workflow (`runner.runNode`), so a request timeout that kills an in-flight agent invocation ends `Runner.Run`'s iterator cleanly. The caller has no way to distinguish "the run was killed mid-flight" from "the agent finished" — a server recording invocation outcomes marks the killed run as a success, and nothing anywhere carries an error. (This is how we found it: a long agent invocation was killed by its platform's request timeout after ~56 minutes of work, and the server recorded a clean completion with no user-visible trace.)

`TestNestedWorkflow_Cancellation` currently pins the swallow (`expected no error on cancellation`), which suggests the conflation was unintentional rather than a design choice — see the Python comparison below.

**To Reproduce**

A node blocks until its context dies, mirroring an in-flight LLM call; the caller cancels the workflow's own context. `Workflow.Run` yields no error:

```go
type blockUntilCancelledNode struct {
	workflow.BaseNode
	started chan struct{}
}

func (n *blockUntilCancelledNode) Run(ctx agent.Context, _ any) iter.Seq2[*session.Event, error] {
		close(n.started)
		<-ctx.Done()
	}
}

func TestExternalCancellationIsSwallowed(t *testing.T) {
	ctx, cancel := context.WithCancel(context.Background())
	// invocation context wrapping ctx, e.g. the package's test MockInvocationContext
	mockCtx := &MockInvocationContext{Context: ctx}

	n := &blockUntilCancelledNode{
		BaseNode: workflow.NewBaseNode("n", "", workflow.NodeConfig{}),
		started:  make(chan struct{}),
	}
	w, _ := workflow.New("", []workflow.Edge{{From: workflow.Start, To: n}})

	errCh := make(chan error, 1)
	go func() {
		var firstErr error
		for _, err := range w.Run(mockCtx) {
			if err != nil && firstErr == nil {
				firstErr = err
			}
		}
		errCh <- firstErr
	}()

	<-n.started
	cancel()

	if err := <-errCh; err == nil {
		t.Fatal("Workflow.Run ended cleanly for an externally-cancelled run; want an error")
	}
}
```

The same holds end-to-end through `Runner.Run` with a root `LlmAgent` whose model call dies on `ctx.Done()`: the iterator ends with no error yielded.

**Expected behavior**

An external cancellation of the workflow's invocation context should fail the run — `Workflow.Run` should yield the cancellation (ideally `context.Cause`) — while scheduler-initiated sibling cancellation keeps surfacing only the failing node's own error.

That is what adk-python does. In its workflow engine, an external `asyncio.CancelledError` re-raises through the scheduler to the caller (`_dynamic_node_scheduler.py` catches `CancelledError` only to clean up its run table, then `raise`), and `NodeStatus.CANCELLED` is assigned only to tasks the engine itself cancelled during shutdown cleanup (`_workflow.py` `_cleanup_all_tasks`, via `task.cancelled()` on tasks it called `.cancel()` on). The Go port approximated "did I cancel this task" with `errors.Is(err, context.Canceled)`, which conflates the caller's cancellation with the scheduler's.

Two adjacent notes if this gets fixed:

- The externally-cancelled completion must be classified **before** the retry block: `ShouldRetry`'s default predicate doesn't exclude cancellation, and retrying a node under a dead context just re-fails.
- There's an ordering where no `context.Canceled` completion is left to classify (a node finishes just before observing the cancel, and the consumer's `parentCtx.Done()` branch has already stopped successor scheduling). The `Done()` branch itself needs to record the cancellation, or that ordering still ends cleanly with part of the graph unexecuted.

- OS: macOS 15 / Linux
- Go version: 1.26
- ADK version: v2.1.0 (also present on main)

Soundcreates:
Fixes #1253

External cancellation of a workflow invocation is now propagated instead of being classified as benign sibling cancellation. The scheduler records cancellation in the parent Done path and uses context.Cause while preserving sibling cancellation behavior.

Tests:
* go test ./workflow -count=1
* go test -race ./workflow -run TestScheduler_ExternalCancellationFailsRun_or_FailedSiblingsCancelled -count=1

### DIFF
--- a/workflow/scheduler.go
+++ b/workflow/scheduler.go
@@ -498,6 +498,23 @@ func runNode(
 	}
 }
 
+// echoesCancellation reports whether err is the dying invocation
+// context reflected back by a node rather than a failure the node
+// determined for itself. runNode sets completion.err to ctx.Err() when
+// a node returns because its context died, so the echo takes three
+// shapes: the parent's own error (context.DeadlineExceeded for a
+// request timeout), its cause (set via context.WithCancelCause), or a
+// plain context.Canceled from the per-node cancel that cancelAll fires
+// in response — a node can observe that before the parent's deadline.
+//
+// Only meaningful once s.parentCtx has failed; context.Cause returns
+// nil for a live context and errors.Is(err, nil) is never true.
+func (s *scheduler) echoesCancellation(err error) bool {
+	return errors.Is(err, context.Canceled) ||
+		errors.Is(err, s.parentCtx.Err()) ||
+		errors.Is(err, context.Cause(s.parentCtx))
+}
+
 // cancelAll cancels every running task. Idempotent: cancelled
 // goroutines may still push events that already left the producer
 // before observing ctx.Done(); the consumer continues draining
@@ -535,6 +552,7 @@ func (s *scheduler) cancelAll() {
 // node-side accumulators.
 func (s *scheduler) run(yield func(*session.Event, error) bool) {
 	var pendingErr error  // first non-nil node error; surfaced after drain
+	var cancelErr error   // cause of an external cancellation; surfaced only when no node reported an error
 	draining := false     // true once cancelAll has run; remaining queue items are drained without yielding or scheduling new successors
 	consumerGone := false // true once the caller broke the range loop; no further yield is allowed
 
@@ -545,12 +563,24 @@ func (s *scheduler) run(yield func(*session.Event, error) bool) {
 		select {
 		case item = <-s.eventQueue:
 		case <-doneChan:
+			doneChan = nil // Disable this case so we don't busy loop
+		}
+
+		// Cancellation of the invocation context is external to the
+		// scheduler: cancelAll only cancels the per-node contexts, so
+		// parentCtx stays live through sibling cancellation. Record the
+		// cause and start draining here rather than in the doneChan case
+		// alone, because a ready doneChan does not have to win the select
+		// — a queued completion can be picked ahead of it, and the loop
+		// can exit before it is ever selected (e.g. cancelAll stops the
+		// last pending retry timer). Either way the run must not report
+		// success.
+		if cancelErr == nil && s.parentCtx.Err() != nil {
+			cancelErr = context.Cause(s.parentCtx)
 			if !draining {
 				draining = true
-				pendingErr = context.Cause(s.parentCtx)
 				s.cancelAll()
 			}
-			doneChan = nil // Disable this case so we don't busy loop
 		}
 
 		switch it := item.(type) {
@@ -595,8 +625,16 @@ func (s *scheduler) run(yield func(*session.Event, error) bool) {
 	// Surface the first error to the caller — but not if the caller
 	// already broke the range loop, since yielding after a false return
 	// panics the iterator.
-	if pendingErr != nil && !consumerGone {
-		yield(nil, pendingErr)
+	//
+	// A node's own error outranks the cancellation cause: when a caller
+	// cancels because a node failed, "context canceled" alone would hide
+	// the reason the run stopped.
+	runErr := pendingErr
+	if runErr == nil {
+		runErr = cancelErr
+	}
+	if runErr != nil && !consumerGone {
+		yield(nil, runErr)
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


indurireddy-TF:
Hi @baptmont 
I tested the changes locally, and the fix is working as expected. I also ran the tests, and everything is passing.
Could you please take a look at the PR? Thanks!

image

baptmont:
I will take a look @indurireddy-TF!
Also @Soundcreates you are not currently covered under a CLA, please visit https://cla.developers.google.com/, or check the failed invocation above for more info.

Soundcreates:
> I will take a look @indurireddy-TF!
> Also @Soundcreates you are not currently covered under a CLA, please visit https://cla.developers.google.com/, or check the failed invocation above for more info.

Hi there, have done that.

baptmont:
Thanks @Soundcreates — this is a good fix, and notably you handled *both* of the footnotes in #1253, not just the headline: the classification sits before the retry block, and the `doneChan` branch covers the ordering where no cancelled completion is left to inspect. That's the part that's easy to miss.

I've pushed `e40432b` on top rather than sending you round another review cycle. Since it changes behaviour beyond what you signed up for, here's the full reasoning — push back on any of it.

### Why I touched it at all: neither hunk was pinned by a test

I mutation-tested the two production changes against your test suite:

| Mutation | Result |
|---|---|
| Delete the `handleCompletion` hunk, keep `pendingErr = context.Cause(...)` | **all tests pass** |
| Delete `pendingErr = context.Cause(...)`, keep `handleCompletion` | **all tests pass** |

Either mechanism alone satisfies every assertion, so the suite couldn't tell them apart — and the second ordering from the issue (the one that motivated the `doneChan` line) had no coverage at all. That ordering is reachable: if the only thing keeping the loop alive is a pending retry timer, `cancelAll` stops it and empties `retryTimers`, so the loop exits without another trip through `handleCompletion`.

So the two are now folded into one check that runs *after* the select instead of inside the `doneChan` case. A ready `doneChan` doesn't have to win a `select`, and the loop can exit before it's ever selected; checking where the loop can actually observe it covers both orderings, and as a side effect stops `tryDispatchPending` from starting a fresh node under a dead context.

### Two behaviour changes on top

**1. A node's own error now outranks the cancellation cause.** Measured on your branch, 40 runs, a node failing with its own error while the context is cancelled:

```
main:        real domain error preserved   40/40
your branch: masked as context.Canceled    40/40
```

Both of your hunks cause this independently, so it wasn't a slip — but it's the same information loss #1253 is about, pointed the other way: a caller who cancels *because* a node failed now gets a bare `context.Canceled` and no trace of why. `context.Cause` is now the fallback, used only when no node reported an error of its own.

**2. An externally cancelled node is marked `NodeCancelled`.** The early return skipped `ns.Status`, leaving the node at `NodeRunning` — documented as "has a live task in flight… must be re-scheduled". adk-python marks every task it reaps during shutdown `CANCELLED` regardless of who cancelled it (`_workflow.py` `_cleanup_all_tasks`, run from a `finally`) and distinguishes external from engine-initiated cancellation only by whether the run raises — never by the status. Matched that, and widened the `NodeCancelled` doc comment accordingly.

### The new tests caught a bug in my own change

Worth calling out, because it's exactly what the missing coverage was hiding. My first attempt guarded with `!errors.Is(it.err, context.Canceled)` — but `runNode` sets `completion.err = ctx.Err()` verbatim, which is `context.DeadlineExceeded` on a request timeout. A timed-out node was therefore classified as "failed on its own terms" and marked `NodeFailed`. Hence `echoesCancellation`, which also covers `context.Cause` (for `WithCancelCause`) and a plain `context.Canceled` that a node can observe from its per-node cancel before the parent's deadline lands.

### Coverage

Four new tests, each verified against a reverted implementation:

| Reverted | Test that fails |
| --- | --- |
| the cancellation recording | `..._DuringPendingRetry` |
| the node-error preference | `..._PrefersNodeError` |
| `ns.Status = NodeCancelled` | `..._MarksNodeCancelled` (cancel + deadline) |
| `context.Cause` → `context.Canceled` | `..._DeadlineSurfacesCause` |
| the `DeadlineExceeded` arm of the echo check | both of the above |

`TestScheduler_FailedSiblingsCancelled` and `TestScheduler_CallerBreakStopsScheduling` still pass, so the fix doesn't over-trigger on the sibling path.

Verified locally: full workspace `go test -race -count=1 -shuffle=on work`, `golangci-lint run` clean in both modules, `go mod tidy -diff` clean, and `./workflow/...` at `-count=10 -shuffle=on` with no flakes.

baptmont (approved):

PR #1259, opened 2026-08-05
fix(remoteagent): respect isolation scope

### LINKED ISSUE #1220: remoteagent/v2: outbound A2A message ignores IsolationScope — a scoped node dispatch sends the whole shared-session history and inherits a prior dispatch's contextID
## 🔴 Required Information

**Describe the Bug:**

When a `remoteagent/v2` A2A agent is dispatched as an isolation-scoped workflow node inside the calling agent's session (`workflow.RunNode` with `WithUseSubBranch()` + `WithIsolationScopeFromNodePath()`), the message sent to the remote server is built from the session's raw event list instead of the invocation's seeded input:

- `newMessage` → `toMissingRemoteSessionParts` (`agent/remoteagent/v2/{a2a_agent.go,utils.go}`) iterates `ctx.Session().Events()` with no isolation-scope (or branch) filtering, walking backward until it finds an event authored by the remote agent itself. On the first dispatch in a session there is no such event, so the **entire conversation** is converted into message parts — the user's original message verbatim, plus every coordinator event rewritten by `presentAsUserMessage` as `For context: [<agent>] said / called tool / tool returned ...` prose (including full tool-result payloads). The node input (`RunNode`'s seeded `UserContent`) is never consulted; it only reaches the remote server embedded in the trailing `[<coordinator>] called tool <remote> with parameters: map[request:...]` history line.
- The same walk also returns the contextID recorded on the last remote-agent-authored event in the session, so **every dispatch after the first reuses the previous dispatch's A2A contextID**, even though each scoped `RunNode` dispatch is an independent task.

This looks like a bug rather than intended history-sync because:

- `session.Event.IsolationScope`'s doc states the visibility contract: *"an event is visible only when event.IsolationScope equals the agent's isolation scope (exact match; empty sees only empty)"*. The LLM flow's contents processor enforces exactly that when building prompt history for scoped LlmAgents; `remoteagent/v2` bypasses it, so a scoped remote child reads unscoped events a scoped LLM child cannot see.
- For LLM children the framework goes further: `RunLLMAgentAsNode` forces `IncludeContents="none"` on ModeSingleTurn children so a dispatched child sees only its seeded request. A remote-agent child has no equivalent, so an A2A `SubAgents` entry with `Mode: ModeSingleTurn` (or any RunNode dispatch) replays the coordinator's whole history to the remote server.
- Consequences in practice: the full conversation — including other tools' result payloads, which can be large and sensitive — is shipped to an external A2A server on every dispatch; and when the remote server keys per-context state by contextID (natural for `a2asrv.AgentExecutor` implementations), two dispatches sharing an inherited contextID collide — concurrent dispatches can tear down each other's per-context resources.

**Steps to Reproduce:**

1. Create an A2A remote agent fronting any `a2asrv` executor (e.g. a local httptest server), with a `BeforeRequestCallbacks` entry that logs the outgoing request (see Minimal Reproduction Code).
2. Register a custom tool on a coordinator LlmAgent whose `Run` dispatches the remote agent as an isolation-scoped node — the same dispatch shape as the built-in single_turn tool, plus the isolation scope.
3. Run the coordinator via `runner.Run` with an in-memory session: user message `"the user's original question"`, then have the model emit two sequential function calls with `request="fetch the first file"` / `"fetch the second file"`.
4. Inspect the logged outbound `a2a.SendMessageRequest` for each dispatch.

**Expected Behavior:**

An isolation-scoped dispatch sends only its seeded input (`"fetch the first file"`), with no contextID inherited from events outside its scope — matching the documented `IsolationScope` visibility contract and the isolation `RunLLMAgentAsNode` applies to LLM children.

**Observed Behavior:**

First dispatch — parts:

```text
["the user's original question",
 "For context:",
 "[coordinator] called tool remoteChild with parameters: map[request:fetch the first file]"]
```

Second dispatch — the parts additionally replay the first dispatch's rewritten result, and `ContextID` equals the contextID minted for the first dispatch's task:

```text
["For context:",
 "[coordinator] remoteChild tool returned result: map[result:...]",
 "For context:",
 "[coordinator] called tool remoteChild with parameters: map[request:fetch the second file]"]
```

**Environment Details:**

 - ADK Library Version: v2.1.0 (behavior also present on `main` at time of filing)
 - OS: macOS (not OS-specific)
 - Go Version: 1.26.5

**Model Information:**
 - Which model is being used: N/A — the bug is in outbound A2A message construction, independent of the model.

---

## 🟡 Optional Information

**Regression:**
No — the v2 `remoteagent` has built the message this way since the workflow machinery was introduced; `agent/remoteagent` (v1) constructs its outbound message the same way and appears to have the same behavior.

**Additional Context:**
N/A

```go
// 1. The remote agent, with a callback that logs what would be sent:
	Name:      "remoteChild",
	AgentCard: card, // points at a local a2asrv httptest server
	BeforeRequestCallbacks: []remoteagent.BeforeA2ARequestCallback{
		func(ctx agent.Context, req *a2a.SendMessageRequest) (*session.Event, error) {
			log.Printf("outbound message: contextID=%q parts=%v", req.Message.ContextID, req.Message.Parts)
			return nil, nil
		},
	},
})

// 2. A coordinator tool that dispatches it as an isolation-scoped node:
func (t *dispatchTool) Run(toolCtx agent.Context, args any) (map[string]any, error) {
	node, _ := workflow.NewAgentNode(remoteAgent, workflow.NodeConfig{})
	out, err := workflow.RunNode[any](toolCtx, node, args.(map[string]any)["request"],
		workflow.WithUseSubBranch(),
		workflow.WithIsolationScopeFromNodePath(),
	)
	return map[string]any{"result": out}, err
}
```

 - Always (100%)

Soundcreates:
Fixes #1220

Scoped remoteagent v2 invocations now filter shared session history to events within the same isolation scope (in `utils.go`), so no contextID or content is inherited from events outside the scope. `getUserFunctionCallAt` is scope-filtered too now, so a function call/response pair from a sibling scope can no longer leak its TaskID, contextID, or content into a different scope.

Seeded `UserContent` is used only as a fallback, for the first dispatch in a scope that has no history of its own yet (checked directly, not by an empty parts count, so a re-dispatch with nothing new in scope produces an empty message instead of resending the original input). Subsequent dispatches in the same scope reuse that scope's history and contextID, including input-required resume.

v1 has the same unscoped shape and is untouched, deferred.

Tests:
* go test ./agent/remoteagent/v2/...
* go test ./agent/... ./workflow/... ./server/adka2a/...
* go test -race ./agent/remoteagent/v2/...
* git diff --check


jjsasha63 (changes requested):
Thanks for taking this on. The `utils.go` half is the right fix, but the early return in `newMessage` is broader than the issue asks for and regresses three behaviours that work on `main` today.

**The two changes overlap.** `toMissingRemoteSessionParts` is only ever called from `newMessage`. Because the early return fires whenever `ctx.IsolationScope() != ""`, the new scope filter in `utils.go` is unreachable for scoped contexts — the only case #1220 is about. It only runs for unscoped invocations, where it newly excludes scoped events from history. That is arguably correct per the `IsolationScope` contract, but it is an unstated, untested behaviour change.

**It is untested.** I removed both `utils.go` hunks and `go test ./agent/remoteagent/v2/...` still passed, including `TestRemoteAgent_ScopedMessageUsesSeededInput`. That test only exercises the `newMessage` early return.

**Three regressions.** I ran the probes below against this branch and against `origin/main`. All three pass on `main` and fail here.

| Case | `main` | this PR |
|---|---|---|
| Second dispatch in the **same** scope | reuses that scope's contextID, sends the unsent turn | mints a new contextID, resends stale `UserContent` |
| Input-required resume (user answers the remote task) | keeps TaskID and ContextID, sends the answer | drops both, resends the original node input — the remote sees a new task and the answer is lost |
| Scoped invocation with nil `UserContent` | sends the in-scope turn | 0 parts, so `run()` short-circuits and never contacts the server |

The second one is the one I would worry about most. `IsolationScope` is inherited, not per-node: `base_flow.go` runs a transferred sub-agent on the *same* `InvocationContext`, so a remote agent reached by `transfer_to_agent` under a scoped ancestor sees a non-empty scope and the ancestor's stale seeded content. It never sees the coordinator's actual request.

The issue asks for no contextID inherited from events *outside* the scope. Dropping same-scope continuity as well goes further than that.

**Why some seeding is still needed.** The scope filter alone is not enough: on the first dispatch there are no in-scope events yet, because `AgentNode.Run` seeds `UserContent` without appending an event. So the filtered history is empty and the message would be empty. Seeding from `UserContent` is the right instinct — it just needs to be a fallback rather than a replacement.

**Suggested shape.** Drop the early return, keep the `utils.go` filter, and seed only when the scope has no history of its own:

```go
parts, contextID := toMissingRemoteSessionParts(ctx, events, cfg)
// A scoped node dispatch receives its input through UserContent, which is
// not an event in the shared session. Seed from it when the scope has no
// history of its own yet.
if len(parts) == 0 && ctx.IsolationScope() != "" {
    if uc := ctx.UserContent(); uc != nil && len(uc.Parts) > 0 {
        event := session.NewEvent(ctx, ctx.InvocationID())
        event.Author = "user"
        event.Content = uc
        seeded, err := convertParts(ctx, cfg, event)
        if err != nil {
            return nil, fmt.Errorf("event part conversion failed: %w", err)
        }
        parts = seeded
    }
}
msg := a2a.NewMessage(a2a.MessageRoleUser, parts...)
msg.ContextID = contextID
return msg, nil
```

With this, all three probes pass, `TestRemoteAgent_ScopedMessageUsesSeededInput` still passes unchanged, and `go test ./agent/... ./workflow/... ./server/adka2a/...` is green.

**Remaining gap either way:** `getUserFunctionCallAt` is not scope-filtered, so a function response from another scope could still be picked up. Worth a follow-up.

**Smaller notes**

- Please add a same-scope positive control and a contextID-isolation case, so the `utils.go` filter is actually covered.
- The description says "History fallback enforces exact isolation scope matching", but that fallback cannot be reached from a scoped context as written.
- The branch is behind `main` and needs a rebase; the CLA check is still failing.
- v1 has the same shape and is untouched. Fine to defer, but worth saying so in the description.

<details>
<summary>Probe tests (drop into <code>agent/remoteagent/v2</code>, uses the package's existing <code>newEventFromParts</code> helper)</summary>

```go
package remoteagent

import (
	"testing"

	"github.com/a2aproject/a2a-go/v2/a2a"
	"google.golang.org/genai"

	adka2a "google.golang.org/adk/v2/server/adka2a/v2"
)

func newScopedCtx(t *testing.T, agentName, scope string, userContent *genai.Content, events ...*session.Event) agent.InvocationContext {
	t.Helper()
	ctx := t.Context()
	store := session.InMemoryService()
	resp, err := store.Create(ctx, &session.CreateRequest{AppName: "test", UserID: "u"})
	if err != nil {
		t.Fatalf("store.Create() error = %v", err)
	}
	for _, e := range events {
		if err := store.AppendEvent(ctx, resp.Session, e); err != nil {
			t.Fatalf("store.AppendEvent() error = %v", err)
		}
	}
	a, err := agent.New(agent.Config{Name: agentName})
	if err != nil {
		t.Fatalf("agent.New() error = %v", err)
	}
	return icontext.NewInvocationContext(ctx, icontext.InvocationContextParams{
		Agent: a, Session: resp.Session, IsolationScope: scope, UserContent: userContent,
	})
}

// Within one isolation scope, a second dispatch should keep the contextID
// minted by the first and replay the same-scope turns the remote has not seen.
func TestProbe_SameScopeContinuity(t *testing.T) {
	const scope, remoteName = "workflow/node@1", "remote-agent"

	first := newEventFromParts("user", genai.NewPartFromText("first request"))
	first.IsolationScope = scope

	remoteReply := &session.Event{
		Author: remoteName,
			Content:        genai.NewContentFromParts([]*genai.Part{{Text: "first answer"}}, genai.RoleModel),
			CustomMetadata: adka2a.ToCustomMetadata(a2a.NewTaskID(), "ctx-same-scope"),
		},
		Actions:        session.EventActions{StateDelta: map[string]any{}, ArtifactDelta: map[string]int64{}},
		IsolationScope: scope,
	}

	follow := newEventFromParts("user", genai.NewPartFromText("follow-up in the same scope"))
	follow.IsolationScope = scope

	ictx := newScopedCtx(t, remoteName, scope, genai.NewContentFromText("first request", genai.RoleUser), first, remoteReply, follow)
	msg, err := newMessage(ictx, A2AConfig{})
	if err != nil {
		t.Fatalf("newMessage() error = %v", err)
	}
	if msg.ContextID != "ctx-same-scope" {
		t.Errorf("ContextID = %q, want same-scope contextID %q", msg.ContextID, "ctx-same-scope")
	}
	if len(msg.Parts) != 1 || msg.Parts[0].Text() != "follow-up in the same scope" {
		t.Errorf("parts = %v, want the unsent same-scope turn", msg.Parts)
	}
}

// Input-required resume: the reply must go back on the same TaskID/ContextID.
func TestProbe_ScopedInputRequiredResume(t *testing.T) {
	const scope, remoteName = "workflow/node@1", "remote-agent"
	taskID := a2a.NewTaskID()

	call := newEventFromParts(remoteName, &genai.Part{FunctionCall: &genai.FunctionCall{ID: "fc-1", Name: "ask_user"}})
	call.CustomMetadata = adka2a.ToCustomMetadata(taskID, "ctx-resume")
	call.IsolationScope = scope

	answer := newEventFromParts("user", &genai.Part{FunctionResponse: &genai.FunctionResponse{
		ID: "fc-1", Name: "ask_user", Response: map[string]any{"answer": "blue"},
	}})
	answer.IsolationScope = scope

	ictx := newScopedCtx(t, remoteName, scope, genai.NewContentFromText("original node input", genai.RoleUser), call, answer)
	msg, err := newMessage(ictx, A2AConfig{})
	if err != nil {
		t.Fatalf("newMessage() error = %v", err)
	}
	if msg.TaskID != taskID {
		t.Errorf("TaskID = %q, want %q", msg.TaskID, taskID)
	}
	if msg.ContextID != "ctx-resume" {
		t.Errorf("ContextID = %q, want %q", msg.ContextID, "ctx-resume")
	}
}

// A scoped invocation with no seeded UserContent produces an empty message;
// run() short-circuits on len(msg.Parts)==0 and never calls the server.
func TestProbe_ScopedNoUserContent(t *testing.T) {
	const scope, remoteName = "workflow/node@1", "remote-agent"

	turn := newEventFromParts("user", genai.NewPartFromText("in-scope question"))
	turn.IsolationScope = scope

	ictx := newScopedCtx(t, remoteName, scope, nil, turn)
	msg, err := newMessage(ictx, A2AConfig{})
	if err != nil {
		t.Fatalf("newMessage() error = %v", err)
	}
	if len(msg.Parts) == 0 {
		t.Errorf("scoped invocation with nil UserContent produced an empty message; the in-scope turn was dropped")
	}
}
```
</details>

Soundcreates:
Thanks for the detailed review  this makes sense. Dropping the early return in newMessage and keeping the utils.go scope filter as the source of truth, with UserContent only as a fallback seed when the scope has no history yet, is the right shape. I'll apply the suggested diff, add the probe tests plus positive/isolation coverage for the utils.go filter, rebase onto main, and push.

Soundcreates:
Hey, I applied the changes you suggested. Dropped the early return in newMessage, kept the utils.go isolation filter as the actual source of truth, and only fall back to seeding from UserContent when the scope doesn't have any history of its own yet, pretty much matching the diff you posted.

Also added the three probe tests you included (same scope continuity, input required resume, and the nil UserContent case), plus a same scope positive control and a contextID isolation case for the utils.go filter like you asked. TestRemoteAgent_ScopedMessageUsesSeededInput still passes without changes.

Ran go test ./agent/... ./workflow/... ./server/adka2a/... locally and everything's green. Rebased on main too, CLA check is passing now.

Left getUserFunctionCallAt and the v1 package alone since you flagged those as follow ups, not blocking this one.

Would appreciate another look when you get a chance, thanks again for the thorough review.

jjsasha63 (commented):
Thanks — this is the shape I suggested, and the three regressions are gone. I
re-ran the probes against this branch and all three pass. Restoring the commit-1
early return on top of these tests fails four of them, so the tests hold the
behaviour down.

Coverage checks out too. I removed each production hunk in turn:

| Removed | Tests that fail |
|---|---|
| backward-loop filter (`utils.go:99`) | `ScopedMessageUsesSeededInput`, `ScopedMessage_ContextIDIsolation` |
| forward-loop filter (`utils.go:115`) | those two, plus `ScopedMessage_SameScopePositiveControl` |
| `UserContent` seeding (`a2a_agent.go:553`) | `ScopedMessageUsesSeededInput`, `ScopedMessage_ContextIDIsolation` |

Rebase and CLA are sorted, CI is green.

One test gap left. The filter's unscoped side is uncovered. If I change both
filters to `if ctx.IsolationScope() != "" && event.IsolationScope != ctx.IsolationScope()`,
every test still passes, so "an unscoped invocation no longer replays scoped
events" is unverified. It is the right behaviour — `session.Event.IsolationScope`
documents "exact match; empty sees only empty", and
`internal/llminternal/contents_processor.go:119` uses the same unconditional
predicate — but it deserves one case.

The guard is looser than its comment. The comment says "when the scope has no
history of its own yet"; the code checks `len(parts) == 0`, which is also true
once the scope's history has all been sent:

```go
first := newEventFromParts("user", genai.NewPartFromText("original input"))
first.IsolationScope = scope
reply := // remote-agent event, IsolationScope: scope, contextID "ctx-a"

ictx := newScopedCtx(t, remoteName, scope,
    genai.NewContentFromText("original input", genai.RoleUser), first, reply)
msg, _ := newMessage(ictx, A2AConfig{})
// got:  1 part, "original input", ContextID "ctx-a"
// main: 0 parts, so run() short-circuits at a2a_agent.go:401
```

A re-dispatch with nothing new in the scope re-sends the original input instead
of doing nothing. The window is narrow, since a transfer or a new turn puts an
in-scope event in front of it. Either tighten the guard to "no in-scope events
at all", or reword the comment to match the code.

Follow-up, and the description overclaims slightly. `getUserFunctionCallAt` at
`a2a_agent.go:540` runs before any filtering, so it leaks content, not just a
contextID:

```go
call := newEventFromParts(remoteName, &genai.Part{FunctionCall: /* ID: "fc-x" */})
call.CustomMetadata = adka2a.ToCustomMetadata(foreignTask, "ctx-foreign")
call.IsolationScope = "workflow/other@1"
answer := newEventFromParts("user", &genai.Part{FunctionResponse: /* ID: "fc-x" */})
answer.IsolationScope = "workflow/other@1"

// A scope "workflow/node@1" invocation sends TaskID foreignTask,
// ContextID "ctx-foreign", and the sibling scope's response content.
```

This is pre-existing on `main`, so it is not a regression and I am happy to
defer it. But the description says "no contextID or content is inherited from
events outside the scope", and this path contradicts that. Either fix it here —
passing `ctx.IsolationScope()` into `getUserFunctionCallAt` and skipping
non-matching events is about four lines, plus the one existing call site at
`utils_test.go:144` — or narrow the wording.

None of this blocks. I will approve once the unscoped test is in, or as-is if
you would rather take all three as follow-ups.

Soundcreates:
Thanks for verifying it against the probes and checking coverage that carefully. Got it on all three points:

1. the unscoped side of the filter is untested, will add a case where an unscoped invocation should not replay scoped events
2. the guard is looser than the comment says, len(parts) == 0 also triggers after the scope's history has already been fully sent, so a re-dispatch with nothing new resends the original input instead of doing nothing. Will tighten it to check for no in-scope events at all rather than the seeded-parts count
3. getUserFunctionCallAt leaking content from other scopes, not just contextID, pre-existing on main but worth fixing here since it's a small change. Will scope-filter it and update the one existing call site

Also noted the description overclaiming, will fix the wording once the above lands.

Soundcreates:
Made the changes.

1. added TestRemoteAgent_ScopedMessage_UnscopedDoesNotReplayScopedEvents, an unscoped context with one scoped event and one unscoped event in the session, checks only the unscoped one shows up in the message. Confirmed it would actually fail if the filter got loosened to only apply for scoped contexts.

2. swapped the len(parts) == 0 check for a new hasIsolationScopeHistory helper that walks the session events and checks if any of them are in the current scope, regardless of whether they've all already been sent. Added TestRemoteAgent_ScopedMessage_ReDispatchWithNoNewHistoryIsEmpty using your exact repro, first turn plus remote reply both in scope, re-dispatch now produces an empty message and keeps the existing contextID instead of resending the original input.

3. getUserFunctionCallAt now takes the scope and requires both the response and the matching call to be in it, updated the one call site in a2a_agent.go and the existing test in utils_test.go to pass "" for the unscoped cases. Added TestRemoteAgent_ScopedMessage_FunctionCallScopeIsolation covering your repro, foreign scope task/contextID/content no longer leak in, falls back to the seeded UserContent instead.

Updated the PR description too so it doesn't overclaim anymore.

go test ./agent/remoteagent/v2/... and go test ./agent/... ./workflow/... ./server/adka2a/... are green, also ran the v2 package with -race.

Thanks again for the thorough pass, let me know if anything's off.

jjsasha63 (commented):
The citation in my last review was the wrong file. It does not change your code,
but it should be on the record before someone else follows it.

I pointed at `session.Event.IsolationScope` and
`internal/llminternal/contents_processor.go:119` to argue the filter should be
unconditional. Those track
`flows/llm_flows/contents.py::_should_include_event_in_context`, the LLM
prompt-history filter. The counterpart of `newMessage` and
`toMissingRemoteSessionParts` is
`agents/remote_a2a_agent.py::_construct_message_parts_from_session`, and the two
Python filters do not agree with each other.

## What adk-python 2.8.0 does

The remote agent gates the scope filter on task mode:

```python
task_scope = ctx.isolation_scope if self.mode == "task" else None   # :1154
```

`contents.py` has no such gate: `ev_iso != isolation_scope` returns False
outright, and its docstring says an unscoped coordinator "sees only unscoped
events". So Python filters unconditionally for the LLM prompt and conditionally
for the A2A message.

The gate rarely fires, though. `workflow/_workflow.py:553-559` gives a node a
workflow-path scope only when `node.mode == "task"` and returns None otherwise,
so a non-None scope and `mode == "task"` arrive together. They come apart only
for a non-task-mode remote agent running under an inherited scope.

Your unconditional filter therefore matches Python everywhere the two systems
overlap, and it matches the contract in `session/session.go:119` ("exact match;
empty sees only empty"). I am not asking you to change it.

What is missing is the reasoning.
`TestRemoteAgent_ScopedMessage_UnscopedDoesNotReplayScopedEvents` is the test I
asked for, and if I gate both filters on `ctx.IsolationScope() != ""` it is the
only test in the package that fails. So the unscoped side rests on that test
alone, with a wrong citation behind it. Could you add a line at the filter in
`utils.go`? Something like:

```go
// Exact match, so an unscoped invocation replays only unscoped events
// (session.Event.IsolationScope). adk-python's remote agent gates this on
// task mode instead; we follow the Go contract, which its own prompt-history
// filter already uses.
if event.IsolationScope != ctx.IsolationScope() {
```

## Two parity notes, neither for this PR

- Python's remote agent refuses workflow-path scopes in task mode. `:1216-1221`
  raises "Workflow path scopes are not supported", pinned by
  `test_run_async_impl_task_mode_rejects_missing_trigger_fc` with
  `isolation_scope = "workflow_path/node@run1"`, the exact shape you handle. Its
  task-mode seed is the coordinator's triggering `FunctionCall` (`:1175-1189`),
  with `user_content` as the fallback for a "workflow-node task with no
  originating FC" (`contents.py:283-296`). Seeding from `UserContent` alone is
  the consistent choice for Go.
- Python restricts the matching call to `event.author == self.name` and converts
  a function response that answers nobody's call into text, so the receiving
  runner does not reject the message (`:1227-1235`, `:1267-1280`). Go's
  `isFunctionCallEvent` does neither. Worth a follow-up issue.

## Last round's two fixes

Both are covered. Removing the scope checks in `getUserFunctionCallAt` fails
`TestRemoteAgent_ScopedMessage_FunctionCallScopeIsolation`, and putting the seed
guard back to `len(parts) == 0` fails
`TestRemoteAgent_ScopedMessage_ReDispatchWithNoNewHistoryIsEmpty`. CI is green
and the branch is one commit behind `main`.

Add that comment and I will approve.

Soundcreates:
Added the requested comment at the exact-match isolation filter, documenting that unscoped invocations replay only unscoped events and tying it to the session contract/prompt-history behavior. Pushed as 90aab85 after rebasing onto the latest PR branch tip. The focused, broader, race, and diff-check validations remain green.

jjsasha63 (commented):
Nearly there — the comment landed on the reasoning I withdrew rather than the
correction.

"matches the prompt-history filter's treatment of isolation scopes" is the
citation I retracted last round. The prompt-history filter is not the
counterpart of this code;
`remote_a2a_agent.py::_construct_message_parts_from_session` is, and it gates
the filter on task mode. As written, the comment points the next reader at the
wrong file.

Here is the text, to paste as-is:

```go
// Isolation scopes require an exact match, so an unscoped invocation replays
// only unscoped events, per session.Event.IsolationScope. adk-python's remote
// agent gates this on task mode instead
// (remote_a2a_agent.py::_construct_message_parts_from_session); we follow the
// Go contract, which its own prompt-history filter already uses.
if event.IsolationScope != ctx.IsolationScope() {
```

Could you put something on the second loop too (`utils.go:121`)? That is the one
deciding what actually gets sent, and it has no comment at all. A
`// same exact-match rule as above` is plenty.

One correction of my own. I said last round that removing the scope checks in
`getUserFunctionCallAt` fails
`TestRemoteAgent_ScopedMessage_FunctionCallScopeIsolation`. That holds for both
checks together. Removing either one alone passes the whole package, so neither
is independently covered — the test puts both the call and the response out of
scope, which either check blocks by itself. I probed the two mixed cases,
response in scope with the call outside it and the reverse, and your code
handles both correctly. So it is a test gap, not a defect. Worth a case each,
but I am not asking for it.

The rest is green: the seed guard and the unscoped filter are both still killed
by their tests, and all 10 CI checks pass. Push the comment and I will approve.

Soundcreates:
Thanks again for the careful review. I checked the branch and the earlier updates are in place, including the rebase that cleared the merge conflicts.

The scoped history filtering, fallback seeding, same-scope continuity, and function-call isolation changes are all on the branch, along with the extra coverage you asked for. GitHub is showing the PR as mergeable, and the focused and broader tests are passing.

I left the v1 and the remaining follow-up behaviour alone as discussed. Appreciate you catching the edge cases here.

jjsasha63 (changes requested):
The comment did not land. Since my last review the branch has one new commit,
1044f2c, and it is a merge from main:

    git diff 92ee539..1044f2c -- agent/remoteagent/

is empty. utils.go:102-104 still reads "matches the prompt-history filter's
treatment of isolation scopes", which is the citation I withdrew, and utils.go:121
still has no comment.

The text again, for utils.go:102:

```go
// Isolation scopes require an exact match, so an unscoped invocation replays
// only unscoped events, per session.Event.IsolationScope. adk-python's remote
// agent gates this on task mode instead
// (remote_a2a_agent.py::_construct_message_parts_from_session); we follow the
// Go contract, which its own prompt-history filter already uses.
if event.IsolationScope != ctx.IsolationScope() {
```

and at utils.go:121:

```go
// Same exact-match rule as above.
if event.IsolationScope != ctx.IsolationScope() {
```

## Something I got wrong four rounds ago

I asked you to scope-filter both events in `getUserFunctionCallAt`. The check on
the response is the wrong half, and it drops a human's answer. I ran the same input
against this branch and against origin/main. The call is stamped `wf/A@1`, the
answer `wf/B@1`, and the invocation runs in `wf/A@1`.

| | main | this PR |
|---|---|---|
| resume | on the call's TaskID, 1 part | TaskID empty, 0 parts |

The two events are stamped by different writers. workflow/agent_node.go:152 stamps
the agent's call with the node scope. The answer is stamped at runner/runner.go:1092
from `findActiveTaskIsolationScope`, which returns the most recent scope with no
successful `finish_task` — by recency, not by which call is being answered. Two
scoped nodes waiting on input at once, and the answer carries the other one's scope.

The call's scope is authoritative and the response's is a guess, so checking the
call alone is enough. Drop the response check at utils.go:46:

```go
if candidate.Author != "user" {
```

and keep utils.go:55 as it is. I ran the three cases with only the call-side check:

    call A, answer B, invocation A  ->  resumes, ctx-A, 1 part
    both other, invocation node     ->  no resume, seeded input  (#1220 still fixed)
    all A                           ->  resumes, ctx-A, 1 part

The whole package passes. `TestRemoteAgent_ScopedMessage_FunctionCallScopeIsolation`
keeps passing, because it puts both events out of scope and the call check blocks it
on its own.

The current failure is quiet, whichever way you go. a2a_agent.go:401
yields an empty event with no log. The `cleanupRemoteTask` defer is registered at
:418, after that return, so the peer's input-required task is never released. adk-python logs and releases control at the same point
(remote_a2a_agent.py:1577-1582).

## The test gap I mentioned, now measured

Dropping `|| candidate.IsolationScope != scope` passes the package, and so does
dropping `request.IsolationScope != scope ||`. If you take the fix above there is
one check left, and a case with the call out of scope and the response in it covers
it.

## Parity

Read against src/google/adk/agents/remote_a2a_agent.py and
flows/llm_flows/contents.py at adk-python 2.8.0.

- Diverges, accepted: Go filters history unconditionally, Python gates on task mode
  at :1165. Go matches its own contract at session/session.go:117-121 and
  internal/llminternal/contents_processor.go:119.
- Diverges, new: Python does not scope the resume path at all —
  `find_matching_function_call(ctx.session.events)` at :1091 walks the whole
  session. The check I asked for has no Python precedent.
- Not ported: NewRemoteAgentEvent (server/adka2a/v2/events.go:30) sets Author and
  Branch but not IsolationScope; Python sets `isolation_scope=ctx.isolation_scope`
  on the events it builds (:424, :468). workflow/agent_node.go:183 stamps them
  after the fact, so it is latent. Follow-up, not this PR.

CI is green and the branch is behind main again. The seed guard and the unscoped
filter are both still killed by their tests.

Soundcreates:
Sorry about that, and thanks for catching it. I added the missing comments in utils.go, including the exact scope matching explanation and the note on the second filter. I also removed the response scope check so a user answer can resume the task based on the matching call scope. The focused tests pass, and I pushed the update as b85c444. Appreciate the careful review.

jjsasha63 (approved):
All three asks from last round are in, and I could not break any of them. Approving.

The comment text at utils.go:101-105 landed verbatim, and the second loop at :122 now carries `// Same exact-match rule as above.`

The response-side check is gone and the call-side one stays (utils.go:44 and :54). I re-ran the mutation I could not kill last round: deleting `request.IsolationScope != scope ||` now fails `TestRemoteAgent_ScopedMessage_FunctionCallScopeIsolation`. So the remaining check is covered, and the test gap I raised closes without a new test.

I also re-checked the reasoning behind the ask, since I had it wrong two rounds before that. `findActiveTaskIsolationScope` (runner/runner.go:1116-1147) returns the most recent scope with no successful `finish_task`, chosen by recency, and it never reads which call is being answered. The call is stamped from the node's own scope at workflow/agent_node.go:182. The call's stamp is authoritative and the response's is a guess, so the call side is the right half to check.

Mutation results across the whole change:

| check | site | result |
|---|---|---|
| call-side scope check | utils.go:54 | killed |
| backward-scan filter | utils.go:106 | killed |
| forward-scan filter | utils.go:123 | killed |
| `!hasIsolationScopeHistory` gate | a2a_agent.go:556 | killed |

`go test -race ./agent/remoteagent/v2/...` passes, `go test ./agent/... ./workflow/... ./server/adka2a/...` passes, and CI is green.

PR #1266, opened 2026-08-07
docs(examples/agentregistry): add discovery and capability-binding samples

wolo-lab:
## Summary

Adds `examples/agentregistry/`: a README index plus two samples, following the
`examples/workflow` layout (concept, goal, Mermaid diagram, config table, run
instructions, example session). The pair mirrors the package's own split —
discovery methods and binding helpers.

- `discover` — lists agents, MCP servers and endpoints through the `All*`
  iterators, with an optional server-side filter. No LLM, no resource names
  needed.
- `bind` — builds an `LlmAgent` from a **capability** (`REGISTRY_TOOL`, default
  `deploy_service_from_image`) rather than a resource name: it scans the
  catalog's declared-tool metadata, picks the MCP server that provides it, and
  binds that one. Binding by resource name would have been interchangeable with
  a hardcoded endpoint URL; binding by capability is not, because the answer
  depends on what is registered in the project at that moment. You ask for a
  capability and get a provider, so the agent ends up with everything that
  server serves.
- Both handle the two real-catalog cases: no provider at all, and several
  providers for one tool name (generic names like `list_services` are declared
  by both `run.googleapis.com` and `agentregistry.googleapis.com`, meaning
  different things). The sample takes the first and names the rest; choosing
  between providers is application policy, so it is deliberately not a knob.
- Both unwrap `*agentregistry.APIError` in an `explain` helper, so a denial
  prints one actionable line naming the missing permission and resource instead
  of the API's JSON envelope. The helper is duplicated rather than shared:
  `examples/` has no cross-sample package today, and keeping each `main.go`
  self-contained seemed worth ~20 repeated lines. 

Verified against a live Agent Registry rather than only compiled: catalog
listing, filtered listing, capability binding with and without ambiguity, a
real MCP tool call over an auto-authenticated `*.googleapis.com` endpoint, and
the not-found and permission-denied paths. The example sessions in the READMEs
are captured output.

The READMEs also record what that exercise turned up: resources are registered
as a `Service` and read back under generated `agentregistry-<uuid>` names;
enabling a Google Cloud API auto-registers its MCP server, so the samples
usually run with nothing registered by hand; `--agent-spec-type=a2a-agent-card`
is required for `RemoteAgent` to resolve an agent, since `no-spec` produces a
`CUSTOM` protocol that never resolves; and a filter value in single quotes is
accepted and silently matches nothing, so string values need double quotes.


examples/agentregistry/bind/main.go
```diff
@@ -0,0 +1,172 @@
...
+	// WithMCPHeaders when it needs credentials. This resolves the endpoint only:
+	// the MCP session opens on the first model turn.
+	toolset, err := client.MCPToolset(ctx, server.Name)
+	if err != nil {
+		log.Fatalf("Failed to build a toolset for MCP server %q: %s", server.Name, explain(err))
+	}
+
+	// The empty config resolves the backend from the environment: a Gemini API
+	// key, or Vertex AI when GOOGLE_GENAI_USE_VERTEXAI is "1" or "true".
+	m, err := gemini.NewModel(ctx, "gemini-flash-latest", &genai.ClientConfig{})
```
kdroste-google: Will gemini-flash-latest work with VertextAI?

examples/agentregistry/bind/README.md
```diff
@@ -0,0 +1,124 @@
...
+Two facts about real catalogs fall out of that, and the sample handles both:
+
+- **A capability may have no provider.** Then there is nothing to bind and the sample says so.
+- **A capability may have several.** Tool names are not unique across servers — in a stock project both `run.googleapis.com` and `agentregistry.googleapis.com` declare `list_services`, meaning entirely different things. The sample takes the first and names the others rather than guessing silently. Choosing between them is policy (a trusted publisher, an allowlist), so it belongs in your application, not here — and a generic tool name is the real smell.
+
+Egress auth stays free along the way: an MCP server on `*.googleapis.com` inherits the credentials the registry client already holds, so nothing in this sample builds an OAuth client.
+
+## Workflow
+
+```mermaid
```
kdroste-google, line 22: To be considered: maybe a sequence diagram will fit better here

examples/agentregistry/bind/README.md
```diff
@@ -0,0 +1,124 @@
...
+| `GOOGLE_CLOUD_REGION` | no | Vertex AI region; genai reads it, the registry does not |
+| `GOOGLE_CLOUD_LOCATION` | no | Registry location, defaults to `global` — **genai reads it too**, and it wins over `GOOGLE_CLOUD_REGION`, so a regional value moves the catalog lookup as well |
+| `REGISTRY_TOOL` | no | Capability to look for, defaults to `list_log_names` |
+
+`console` is the default; the launcher also serves `web api`, `web a2a` and `web webui`. An unrecognised argument prints the full command-line syntax.
+
+## Example session
+
+Real output. The default capability resolves to one provider, and the tools it brings are then callable:
+
```
kdroste-google, line 85: It's not reproductible by the user, maybe it would be better to skip it

examples/agentregistry/bind/README.md
```diff
@@ -0,0 +1,124 @@
...
+$ REGISTRY_TOOL=send_carrier_pigeon go run ./examples/agentregistry/bind/ console
+Failed to find a provider for "send_carrier_pigeon" in projects/my-project/locations/global:
+no registered MCP server declares this tool; run the discover sample to see what is available
+```
+
+At no point does the sample contain an endpoint URL, a tool schema, or a token.
+
+## Notes
+
+- **Declared tools are metadata, not a contract.** `MCPServer.Tools` is what was uploaded with the server spec, so it can lag the server. It is the right input for *choosing* a server and the wrong one for *calling* it — the agent only ever calls what the live MCP connection reports.
```
kdroste-google: I don't understand the part about "the wrong one"

examples/agentregistry/discover/main.go
```diff
@@ -0,0 +1,198 @@
...
+	names := make([]string, 0, len(tools))
+	for _, t := range tools {
+		names = append(names, t.Name)
+	}
+	return names
+}
+
+func endpointURL(e *agentregistry.Endpoint) string {
+	for _, i := range e.Interfaces {
+		if i.URL != "" {
```
kdroste-google, line 149: Returns the first one, is that intended?

kdroste-google (changes requested):

examples/agentregistry/bind/README.md
wolo-lab: Right, replaced.

examples/agentregistry/bind/main.go
wolo-lab: No, it doesn't :( 
Changing to 3.5 and adding a note.

examples/agentregistry/bind/README.md
wolo-lab: Let's leave it as discussed: this are basic logs which should be visible for everyone.

examples/agentregistry/bind/README.md
wolo-lab: right, simplifying this point.

examples/agentregistry/discover/main.go
wolo-lab: Right, updated to display all.

kdroste-google (approved):
LGTM, thank you!

PR #1268, opened 2026-08-07
fix(workflow): build route edges in a deterministic order

wolo-lab:
## Problem
`EdgeBuilder.AddRoutes` ranges a Go map, so the edges it appends come out in a
fresh random order on every run. The YAML config path has the same bug
independently: `parseEdges` ranges the map it decoded each route map into.
Edge order is observable, so this is not cosmetic:
- `newGraph` (workflow/graph.go:30) keeps each source node's successor list in
  input order, `findSuccessors` (workflow/scheduler.go:986) walks it in that
  order, and both dispatch sites start nodes in the order they come back. When
  several routes match the same event — which they can, since `Event.Routes` is
  a `[]string` (session/session.go:129) — edge order decides whose goroutine is
  launched first.
- Under `WithMaxConcurrency` it stops being a head-start bias and becomes fully
  determined: successors past the limit land in `pendingQueue` in that order and
  drain strictly FIFO. `TestMaxConcurrency_PendingDispatchedFIFO`
  (workflow/max_concurrency_test.go:110) already pins that FIFO.
The practical cost is unreproducible runs and fixtures that can only be asserted
on order-insensitively. `TestEdgeBuilder`'s assertion loop was an
order-insensitive scan that existed for exactly this reason.

## Summary
- Both paths now iterate the route keys sorted.
- On the Go builder path sorting is the only option available: `map[string]Node`
  has already destroyed the caller's order by the time `AddRoutes` is entered.
- On the YAML path it is a choice. `item.Content` survives `item.Decode`, so the
  author's declaration order is still reachable there; honouring it costs
  ~15-20 lines once alias keys (whose `Content` node carries the anchor name,
  not the resolved key) and merge-injected keys (which have no declared position
  at all) are handled. Sorting gives every key a defined position in six lines
  and leaves yaml.v3's mapping decode untouched, so merge keys, anchors,
  duplicate-key detection with line numbers and every type error behave exactly
  as before. Declaration order is worth revisiting on its own: the YAML route
  order is `internal/` behaviour with no documented contract and `AddRoutes` has
  no non-test callers, so neither path is expensive to reverse.
- It also leaves the Go and YAML paths agreeing on one ordering rule instead of
  two opposite ones.
- `TestEdgeBuilder`'s scan now compares edges in order, and its map literal was
  reordered so declaration order contradicts sorted order — it detected the bug
  3 times in 40 runs before, 35 now. `TestEdgeBuilder_AddRoutesSortsByRoute` and
  `TestParseEdges_RouteOrderIsDeterministic` use enough routes to push the map
  past Go's single-bucket layout, where iteration is only a random rotation and
  matched sorted order by luck too often; both are red 40/40 against the unfixed
  code.
- `TestEdgeBuilder_AddRoutesSortsBytewise` pins *which* order. Its keys and
  target names are chosen so byte order differs from numeric, case-folded and
  by-target-name order — each of those otherwise passes the entire suite, which
  would leave the documented `"10"` before `"2"` unguarded.
- `AddRoute`'s doc comment is corrected in passing, since the new `AddRoutes`
  comment now points readers at it: it described the parameter as `any` (it is
  `Route`) and omitted `Default` from the routes it accepts.

Not a breaking change: the previous order was random, and Go re-randomizes map
iteration per `range` rather than once per process, so no passing program could
have depended on it.


karolpiotrowicz (commented):
The ordering fix itself holds up. Both paths use the same `slices.Sorted(maps.Keys(...))` primitive, the bytewise property is genuinely pinned on the Go path, and `Default`'s position in the edge slice turns out not to matter at dispatch — `findSuccessors` matches it by identity, stashes it, and appends it after the loop only when no concrete route fired, so no ordering of the edge slice can shadow a concrete route. Build, the full suite, vet, gofmt, `-race` and govulncheck are all green, and the only static-analysis findings in the tree are pre-existing on main.

What I'd change is in the tests and the prose around them.

**The `AddRoutes` case in `TestEdgeBuilder` catches the regression about 89% of the time, not always.** Its map has two keys ([`edgebuilder_test.go:83-86`](https://github.com/google/adk-go/blob/0d95f9b4831c15a1c6dc7de2f3b8b13bb73d2a5a/workflow/edgebuilder_test.go#L83-L86)), and a two-entry map iterates as one of two rotations, one of which equals the asserted order. I reverted the sort to a bare `range` and ran that subtest 300 times: 268 failed, 32 passed. So it does mostly work, but the comment above it — "Literal order is the reverse of sorted order, so this case can tell the two apart" ([`edgebuilder_test.go:81-82`](https://github.com/google/adk-go/blob/0d95f9b4831c15a1c6dc7de2f3b8b13bb73d2a5a/workflow/edgebuilder_test.go#L81-L82)) — promises something stronger than a 1-in-9 escape rate. A third key whose sorted position is not a rotation of literal order would make it deterministic, which is what the two dedicated tests already do.

**`TestParseEdges_RouteOrderIsDeterministic` does not pin bytewise order.** Every key is alphabetic ([`configurable_workflow_test.go:607-620`](https://github.com/google/adk-go/blob/0d95f9b4831c15a1c6dc7de2f3b8b13bb73d2a5a/internal/configurable/configurable_workflow_test.go#L607-L620)), so there is no `"10"`/`"2"` pair to separate bytewise from numeric. I swapped the YAML comparator for a numeric-aware one and the test stayed green. Bytewise is currently guaranteed on the YAML path only by both paths happening to call the same stdlib primitive, which is exactly the coupling `TestEdgeBuilder_AddRoutesSortsBytewise` exists to stop relying on. Adding a digit pair to the fixture mirrors that test at basically no cost.

**The same test also cannot distinguish the implementation from one that pins `default` to the end.** All eight named keys are uppercase, so lowercase `default` sorts last by construction. I added a mutation that explicitly forces the default key last while keeping the sort, and the test still passed. That one is cosmetic given the dispatch behavior above, so treat it as lower priority than the other two — one lowercase-keyed case would close it.

**Nothing connects either builder to `Workflow.Run`.** `AddRoutes` appears only in `edgebuilder.go` and its own test, with no production caller anywhere in the repo, and no test feeds either path's output into the scheduler. The new doc comment now promises that edge order "drives the order successors are started in, and the pending queue under `WithMaxConcurrency`" ([`edgebuilder.go:64-66`](https://github.com/google/adk-go/blob/0d95f9b4831c15a1c6dc7de2f3b8b13bb73d2a5a/workflow/edgebuilder.go#L64-L66)). That statement is accurate as written, but it is the one behavior in this change with no test behind it, and it is the reason the change matters.

**The PR description's explanation of why the Go-path tests work is wrong, though the tests are fine.** It says they "use enough routes to push the map past Go's single-bucket layout". Both maps have four keys, comfortably inside the eight-slot group — I measured a four-key map over 2000 iterations and got exactly four orders, all rotations of insertion order, biased 62% toward insertion order. The tests are reliably red anyway, for a better reason: the expected sequence is not any rotation of the literal order. That is the property doing the work, and it is worth writing down where the comments are, because reordering either literal into a rotation of sorted order would quietly turn the test into a coin flip. The equivalent comment on the YAML test ([`configurable_workflow_test.go:604-606`](https://github.com/google/adk-go/blob/0d95f9b4831c15a1c6dc7de2f3b8b13bb73d2a5a/internal/configurable/configurable_workflow_test.go#L604-L606)) is correct, since nine keys really do exceed one group.

**Minor, on the corrected `AddRoute` comment.** It now reads as a closed set — "route may be `StringRoute`, `IntRoute`, `BoolRoute`, `MultiRoute`, or `Default`" ([`edgebuilder.go:38-40`](https://github.com/google/adk-go/blob/0d95f9b4831c15a1c6dc7de2f3b8b13bb73d2a5a/workflow/edgebuilder.go#L38-L40)) — but `Route` is an exported open interface, and `nil` is also accepted and means something specific: `findSuccessors` treats a nil route as an unconditional edge. Fixing the old `any` claim was right, and "such as" plus a note about nil would finish the job.

One thing I ran into that is not yours, so I'll file it separately rather than pile it on here: `resolveNodeLike` treats any `.yaml`/`.yml` route target as a path, explicitly permits absolute refs, and reads the file with no containment check, where the sibling helper for agent references does guard exactly that. It sits directly under the loop this PR rewrote, which is how I hit it, but it is present unchanged on the merge base and no line in this diff touches it.

PR #1269, opened 2026-08-07
fix(workflow): report every graph validation violation at once

wolo-lab:
## Problem

`workflow.New` validates a graph with 13 checks, grouped into three
sequential phases (`validateNodes` → `validateSubWorkflowNames` →
`validateWorkflow`). Every check returned the *first* violation it found, and
every phase returned the first failing check. A caller with three mistakes in
their edge list therefore needed three fix-recompile cycles to see all three —
each run revealed exactly one of them.

Which violation surfaced was also non-deterministic: six of the checks iterate
`graph.successors` / `predecessors` / `allNodes`, so on a graph with more than
one violation of the same kind, Go's randomized map iteration decided what the
user was told to fix.

## Summary

- Each phase now runs all of its checks, and each check collects all of its own
  findings; they come back combined with `errors.Join`. Phases still
  short-circuit each other — a later phase assumes the earlier ones hold, and
  there is no point checking routing on a graph whose node names collide.
- Not a breaking change: `errors.Is` matches through joined errors, so code
  keying off `ErrDuplicateNodeName`, `ErrMultipleDefaultRoutes`, … keeps
  working. A lone violation is returned unjoined, so the single-violation case
  is byte-identical to before (including direct `==` against a sentinel).
- Checks that walked a map now iterate nodes in name order, so the reported
  order — and with aggregation, the whole error string — is stable across runs.
- `validateCycles` reports one finding per node that closes a cycle rather than
  one per back-edge. On a complete 400-node graph that is 399 lines / 14 KB
  instead of 79 800 lines / 3 MB.
- `joinViolations` drops repeated identical messages, so two edges breaking one
  rule the same way read as one problem.
- `ErrSubWorkflowNameCollision` now names the offending node; without it,
  multiple collisions produced indistinguishable duplicate lines.

Example, a graph with three mistakes:

```
node has more than one default route: "A" nodes not reachable from start node: "D, E" unconditional cycle detected: "D"
```

### DIFF
--- a/workflow/graph.go
+++ b/workflow/graph.go
@@ -14,6 +14,11 @@
 
 package workflow
 
+import (
+	"slices"
+	"strings"
+)
+
 // graph is the precomputed structural view of a workflow's edges.
 // Built once at workflow construction; queried by the engine at
 // dispatch time.
@@ -37,13 +42,23 @@ func newGraph(edges []Edge) *graph {
 	return &graph{successors: succ, predecessors: pred}
 }
 
-// allEdges returns all edges in the graph.
+// allEdges returns all edges in the graph, grouped by source node in
+// name order.
 func (g *graph) allEdges() []Edge {
 	var edges []Edge
-	for _, succs := range g.successors {
-		edges = append(edges, succs...)
+	for _, n := range g.sortedNodes() {
+		edges = append(edges, g.successors[n]...)
 	}
 	return edges
+}
+
+// sortedNodes returns all nodes ordered by name. Callers that report
+// findings to the user (graph validation) use it so the output does not
+// inherit Go's randomized map iteration order.
+func (g *graph) sortedNodes() []Node {
+	nodes := g.allNodes()
+	slices.SortFunc(nodes, func(a, b Node) int { return strings.Compare(a.Name(), b.Name()) })
+	return nodes
 }
 
 // successorsOf returns the outgoing edges for a node.
--- a/workflow/validation.go
+++ b/workflow/validation.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"iter"
 	"slices"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


karolpiotrowicz (commented):
The phase structure holds up. I diffed each rewritten check against its
fail-fast predecessor and the set of violations detected is unchanged — the
switch from ranging `successors`/`predecessors` to `sortedNodes()` walks a
superset whose extra entries index to nil slices, `allEdges()` yields the same
multiset, and the cycle DFS still reports at least one node per cycle. Two
things I'd like resolved before this merges, then some smaller ones.

Line references are pinned to `949e9d4`.

## The `(node %q)` suffix contradicts the guarantee three lines above it

[`checkSubWorkflowName`](https://github.com/google/adk-go/blob/949e9d481f8725926f3a626ef1f262a4ca1efe1f/workflow/validation.go#L137)
now formats `"%w: %q (node %q)"`. That is the only message this check can emit,
so its violation is always the lone-violation path — the one
[`joinViolations`](https://github.com/google/adk-go/blob/949e9d481f8725926f3a626ef1f262a4ca1efe1f/workflow/validation.go#L81-L85)
documents as keeping "the exact identity and message it had before validation
started aggregating", and the one the description calls byte-identical
including `==` against a sentinel.

The suffix also can't distinguish anything through the public API. `subWorkflow`
is unexported and only
[`NewWorkflowNode`](https://github.com/google/adk-go/blob/949e9d481f8725926f3a626ef1f262a4ca1efe1f/workflow/workflow_node.go#L34-L43)
sets it, from the same `name` it hands to `New`, so a node's name always equals
its sub-workflow's name and the collision only fires when both equal the
parent's:

```
sub-workflow name collision: "wf" (node "wf")
```

The multi-collision case the description gives as the motivation is unreachable
for the same reason. Two colliding sub-workflows would both have to be named
after the parent, so `validateUniqueNames` rejects them a phase earlier and the
collision message never renders at all:

```
New("wf", []Edge{{Start, wn1}, {wn1, wn2}})  ->  duplicate node name: wf
```

[`TestValidateSubWorkflowNames`](https://github.com/google/adk-go/blob/949e9d481f8725926f3a626ef1f262a4ca1efe1f/workflow/validation_test.go#L654-L666)
matches with `strings.Contains` on the old prefix, so it passes either way.

## A duplicated edge now recommends a JoinNode that wouldn't help

[`validateFanIn`](https://github.com/google/adk-go/blob/949e9d481f8725926f3a626ef1f262a4ca1efe1f/workflow/validation.go#L385-L400)
counts predecessor *edges*, not distinct predecessors. That was safe while
[`validateUniqueEdges`](https://github.com/google/adk-go/blob/949e9d481f8725926f3a626ef1f262a4ca1efe1f/workflow/validation.go#L264-L277)
returned first. Now both run, and `New("wf", []Edge{{Start, A}, {A, B}, {A, B}})`
reports:

```
duplicate edge: from "A" to "B"
non-JoinNode fan-in is not yet supported: node "B" has 2 unconditional incoming edges; use a JoinNode to converge branches
```

B has one predecessor, listed twice. The second line is false and the change it
asks for wouldn't resolve the graph. Phases short-circuit each other on exactly
this reasoning, and it applies between checks inside `validateWorkflow` too
wherever one check's violation invalidates a later check's input.

## Smaller things

- **The cycle test guards nothing.**
  [`TestValidateCycles_OneFindingPerNode`](https://github.com/google/adk-go/blob/949e9d481f8725926f3a626ef1f262a4ca1efe1f/workflow/validation_test.go#L526-L545)
  asserts `got == 0 || got > n`. The fail-fast code returns 1 finding for that
  20-node complete graph and this one returns 19, so both satisfy the assertion
  and the test passes unmodified against the old code. Its comment — and the
  description's "79 800 lines / 3 MB" — describe a naive implementation rather
  than the one being replaced, which returned a single error.

- **The dedup ships without a test.** Nothing in the package reaches the
  `seen[err.Error()]` branch in
  [`joinViolations`](https://github.com/google/adk-go/blob/949e9d481f8725926f3a626ef1f262a4ca1efe1f/workflow/validation.go#L86-L100).
  Statement coverage reads 100% because the nil check and the dedup check share
  one `continue`. Two edges breaking one rule identically would cover the
  behaviour the description advertises.

- **The ordering guarantee is tested at one call site.**
  `TestValidateWorkflow_StableViolationOrder` drives `validateDefaultRoute`
  only, so dropping the sort from the other five `sortedNodes()` callers leaves
  the suite green.

- **`err == ErrNoStartNode` stops matching** once a graph has a second
  violation, because
  [`errors.Join`](https://github.com/google/adk-go/blob/949e9d481f8725926f3a626ef1f262a4ca1efe1f/workflow/validation.go#L99)
  wraps. It's the only sentinel ever returned bare, so it's the only one
  affected, and `errors.Is` is unaffected throughout. Still worth a release
  note.

- **Removing the early returns widened where existing panics are reachable.** A
  nil `Edge.To`, or a `Node` implemented as a value type with a slice field,
  used to come back as a validation error when an earlier check caught
  something first. Now
  [`distinctNodes`](https://github.com/google/adk-go/blob/949e9d481f8725926f3a626ef1f262a4ca1efe1f/workflow/validation.go#L107)
  hashes every node in the very first check and panics. Those inputs panic on
  the old code too once they get past the first violation, so this widens an
  existing hole rather than opening one — but a returned error would beat a
  panic in a constructor.

- **`validateStateSchemaConsistency` repeats the whole declared-field list** in
  every finding at
  [validation.go:487](https://github.com/google/adk-go/blob/949e9d481f8725926f3a626ef1f262a4ca1efe1f/workflow/validation.go#L487),
  so 40 unknown fields against a 40-field schema goes from 859 bytes to 34 KB.
  Naming it once would read better as well as being smaller.

- **`sortedNodes()` is rebuilt six times per `validateWorkflow`**, each time
  walking both maps and re-sorting. Computing it once and passing it down would
  make the ordering guarantee harder to drop by accident.

PR #1271, opened 2026-08-08
docs(examples/agentregistry): add an A2A publish-and-resolve sample

wolo-lab:
## Problem

Stacked on #1266.

That PR covers discovery and `MCPToolset`, but leaves `RemoteAgent` documented
and never exercised. The reason is that the only A2A agent in a stock registry
is Google-managed, and calling it needs that product's own OAuth scopes — an
ADC token scoped to `cloud-platform` comes back with
`403 PERMISSION_DENIED: Request had insufficient authentication scopes`. So the
half of the client that resolves agents had no runnable example, and the setup
that would make one work was undocumented.

## Summary

Adds `examples/agentregistry/a2a/`, which registers an agent of your own and
then reaches it back through the catalog.

The sample turns on one property of the service: **the registry stores the card
and never calls the URL inside it.** So the card can point at
`http://localhost:8765` and the whole loop closes with nothing deployed — no
Cloud Run service, no tunnel, no public endpoint. The published agent is a
canned echo, so the sample needs no model credentials either; ADC is enough.

- One process plays both roles — it serves the agent over A2A and consumes it
  through the registry — so the loop runs in one terminal. The consuming half
  only ever knows a resource name; the URL, the transport and the agent's
  identity all come out of the card.
- The echo replies with the request text, so a green run covers both directions
  rather than just proving something came back.
- `ask` fails loudly on `LLMResponse.ErrorMessage`. A rejected A2A call arrives
  as an event with an error and no content, which otherwise reads as the agent
  simply having nothing to say.
- The README carries the one-time `gcloud services create` recipe, the
  `services describe --format='value(registryResource)'` step for the projected
  agent name, and a delete command to clean up.
- The index's registration section shrinks to the `Service` → projection
  concept and defers to the sample for the working card, so the traps are
  documented once instead of twice.

Verified against a live Agent Registry: the exact card JSON in the README was
registered, the sample run against it, and the pasted session is its real
output. Both the project-number form that `gcloud` prints and the project-ID
form resolve. The throwaway `Service` was deleted afterwards and the catalog
confirmed clean.

### DIFF
--- a/examples/agentregistry/README.md
+++ b/examples/agentregistry/README.md
@@ -20,7 +20,7 @@ export GOOGLE_CLOUD_PROJECT=your-project
 
 You probably don't have to register anything to try the samples: enabling a supported Google Cloud API auto-registers its **remote MCP server** in your registry. A fresh project typically already lists `run.googleapis.com`, `logging.googleapis.com`, `compute.googleapis.com`, and friends — all of which `bind` can consume as-is.
 
-The client authenticates with ADC and bills the quota project (`GOOGLE_CLOUD_QUOTA_PROJECT`, then the credentials' `quota_project_id` on the ADC path, then `Config.ProjectID` — which these samples take from `GOOGLE_CLOUD_PROJECT`). `gcloud` picks its quota project separately, so pass `--billing-project` to it or an unrelated default can deny the call with `USER_PROJECT_DENIED`.
+The client authenticates with ADC and bills the quota project (`GOOGLE_CLOUD_QUOTA_PROJECT`, the credentials' `quota_project_id`, then `GOOGLE_CLOUD_PROJECT`, in that order). `gcloud` picks its quota project separately, so pass `--billing-project` to it or an unrelated default can deny the call with `USER_PROJECT_DENIED`.
 
 To see what is registered without running any Go code:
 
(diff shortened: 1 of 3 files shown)


examples/agentregistry/README.md
```diff
@@ -76,25 +77,9 @@ The dotted line is the point of the whole thing: **the registry is a catalog, ne
 
 ## Registering your own agent
 
-The registry's write surface is the `Service` resource. You create a `Service`; the registry *projects* read-only `Agent` and `McpServer` resources from it — those are what the client reads, and they carry generated IDs like `agentregistry-00000000-0000-0000-630f-070a9d06e171`, not the ID you chose. Get the projected name from `services describe`:
+The registry's write surface is the `Service` resource. You create a `Service`; the registry *projects* read-only `Agent` and `McpServer` resources from it — those are what the client reads, and they carry generated IDs like `agentregistry-00000000-0000-0000-630f-070a9d06e171`, not the ID you chose, so read the projected name back with `services describe --format='value(registryResource)'`.
 
-```bash
-gcloud agent-registry services create my-agent \
```
kdroste-google, line 82: Is this piece of document deleted on purpose?

kdroste-google (approved):
LGTM, just see one comment, thank you!

examples/agentregistry/README.md
wolo-lab: Yes, it was moved into a2a/README.md which has a card that was actually registered against a live registry. 
Below there's a link to the mentioned previously README.

Issue #1273, opened 2026-08-08
[docs]: Go is not rendering the OpenAPI documentation, resulting in a discrepancy with the live documentation endpoint.


shuqingzai:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

## 🔴 Required Information

The ADK documentation (`https://adk.dev/api-reference/rest/openapi.json`) is inconsistent with the actual API endpoints exposed by the Go implementation.

Example: 

- go uses /api as the prefix by default
- go /health /apps/{app_name}/app-info is missing
- go has `/run_live`, but it's not available in the production environment.
- Go does not expose a `/docs` endpoint to render OpenAPI.  -- https://adk.dev/runtime/api-server/#interactive-api-docs

**Steps to Reproduce:**
Please provide a numbered list of steps to reproduce the behavior:
1. Install '...'
2. Run '....'
3. Open '....'
4. Provide error or stacktrace

**Expected Behavior:**

**Observed Behavior:**
What actually happened? Include error messages or crash stack traces here.

**Environment Details:**

 - ADK Library Version: main
 - OS: [e.g., macOS, Linux, Windows]
 - Go Version: v1.26.5

**Model Information:**
 - Which model is being used: (e.g., gemini-2.5-pro)

---

## 🟡 Optional Information
*Providing this information greatly speeds up the resolution process.*

**Regression:**
Did this work in a previous version of ADK? (Yes/No) If so, which one?

**Logs:**
Please attach relevant logs. Wrap them in code blocks (```) or attach a
text file.
```text
// Paste logs here
```

**Screenshots / Video:**
If applicable, add screenshots or screen recordings to help explain
your problem.

**Additional Context:**

**Minimal Reproduction Code:**
Please provide a code snippet or a link to a Gist/repo that isolates the issue.
```go
// Code snippet here
```

**How often has this issue occurred?:**

 - Always (100%)
 - Often (50%+)
 - Intermittently (<50%)
 - Once / Rare

vivekjm:
I’m working on the focused `/health` parity gap from this report. I’ll keep the change limited to the Go REST server route and regression coverage, matching the Python server’s `{"status":"ok"}` response; the broader OpenAPI/docs discrepancies can remain separate.

PR #1274, opened 2026-08-09
fix(adkrest): add health endpoint

vivekjm:
## Summary

- add a lightweight `GET /health` route to the ADK REST server
- return the same `{"status":"ok"}` response as ADK Python
- cover the route status, content type, and response body

This addresses the missing health endpoint called out in google/adk-go#1273. I kept the change focused on that concrete parity gap; the broader OpenAPI and documentation differences remain separate.

## Testing plan

- `go test -race -count=1 ./server/adkrest/...`
- `golangci-lint run` in the root module and `plugin/agentanalytics` (v2.3.1, 0 issues)
- `go mod tidy -diff` in the root module and `plugin/agentanalytics`


kdroste-google (approved):
@vivekjm, it seems fine, thank you!

PR #1277, opened 2026-08-10
fix(configurable): prevent path traversal in AgentTool config_path resolution (#878)

wolo-lab:
## Problem

`ResolveAgentReference` accepts absolute `refPath` values unconditionally and joins
relative paths with `filepath.Join` without any boundary enforcement. An
attacker-controlled `config_path` in an agent YAML can therefore read any file the
server process can reach. `v1` carries the identical resolution logic, so the
vulnerability is present here too.

## Summary

Backports #878 from `main`.

- Reject absolute `refPath` values.
- After resolving, verify the cleaned path stays inside the parent agent's directory.

The same fix landed in adk-python (google/adk-python#5826) and adk-java
(google/adk-java#1218); the pattern is identical across the three SDKs.

### DIFF
--- a/internal/configurable/configurable_utils.go
+++ b/internal/configurable/configurable_utils.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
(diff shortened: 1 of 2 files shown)


karolpiotrowicz (commented):
The change matches #878 exactly: after the cherry-pick, ResolveAgentReference is
byte-for-byte the same function as on main.

One thing to add before this lands. The commit message carries "BREAKING: an
absolute config_path is no longer accepted", but the PR description doesn't.
That distinction mattered less on main, where it shipped inside a major version;
on v1 it's a behavioral break in a minor release, and anyone currently passing an
absolute config_path starts getting an error on upgrade. Worth surfacing in the
description and in the v1.6.0 release notes.

The backport is clearly warranted — on current v1 the absolute-path case resolves
and loads the target, so /etc/passwd is read and parsed as agent YAML.

PR #793, opened 2026-05-05
fix: update built-in load_memory tool for compatibility with VertexAiSessionService

nicholasbreckwoldt:
- Closes: #792

**Problem:**
Built-in ADK `load_memory` tool throws an error when using VertexAiSessionService to manage session event persistence. Downstream, the function call response is passed to structpb.NewStruct() when used with this custom session service implementation. structpb is strictly incompatible with native Go structs. Passing a struct directly causes a runtime error / failure.

Screenshot 2026-05-05 at 10 47 35

**Solution:**
This PR updates the built-in `load_memory` tool logic to convert the [memory.searchResponse] struct into a generic map[string]any before returning. This is done using a JSON round-trip (marshall / unmarshal) to ensure more general compatability, including with VertexAiSessionService.

Screenshot 2026-05-05 at 10 48 35

### Testing Plan
**Unit Tests:**

- [x] I have added or updated unit tests for my change.
- [x] All unit tests pass locally.

The following local tests were run and passed successfully. Note minor adjustment to the `TestLoadMemoryTool_Run` test, i.e. marshalling tool responses into the expected [memory.SearchResponse] type to make handling more robust and type safe.

```go
go test ./tool/loadmemorytool
```

Test results below.

Screenshot 2026-05-05 at 10 33 13

**Manual End-to-End (E2E) Tests:**

Manually tested with ADK agent configured with VertexAISessionService and PR-proposed changes. Code snippet and screenshot showing E2E testing below:

```go
...
  // Initialise Vertex AI Engine Session Service
  sessionService, err := vertexai.NewSessionService(ctx, vertexai.VertexAIServiceConfig{
  	ProjectID: "PROJECT",
  	Location:  "LOCATION",
      ReasoningEngine: "REASONING_ENGINE",
  })
  if err != nil {
      log.Fatalf("failed to create session service")
  }

// Create new agent
  adkAgent, err := llmagent.New(llmagent.Config{
      Name:        "agent",
      Model:       model,
      Description: "Description",
      Instruction: "You are a helpful assistant..",
      Tools: []tool.Tool{
          loadmemorytool.New(),
      },
  })
  if err != nil {
      log.Fatalf("failed to create agent: %v", err)
  }  

  // Configure launcher
  config := &launcher.Config{
      AgentLoader: agent.NewSingleLoader(adkAgent),
      SessionService: sessionService,
      MemoryService: memory.InMemoryService(),
  }

...
```

Screenshot 2026-05-05 at 10 54 17

### Checklist

[...]

PR #1281, opened 2026-08-10
fix: backport load_artifacts, streaming usage and artifact-delta fixes (#670, #1257, #904)

wolo-lab:
## Problem
Three defects present on `v1`:
- `load_artifacts` silently loads nothing once a session has been persisted and
  reloaded. Session storage carries the function response through JSON, so
  `Response.AsMap()` hands back `[]any` where the tool wrote `[]string`, and
  `processLoadArtifactsFunctionCall` asserted the concrete type:
  `invalid artifact names type: []interface {}, expected []string`.
- Streaming `UsageMetadata` is overwritten by later chunks, so a chunk without usage
  erases usage reported earlier.
- The Vertex AI session client drops `Actions.ArtifactDelta` when appending events and
  does not restore it when listing them; artifact versions that exceed the `int32` wire
  field are silently truncated.

## Summary
Backports #670, #1257 and #904 from `main`, plus the `GetArtifactVersion` coverage from
#1156.
- Accept both `[]string` and `[]any` for `artifact_names` on the read side, and report a
  non-string element by index instead of the type of the whole slice. `Run` keeps
  returning `[]string`: the write path no longer calls `structpb.NewStruct`, `toStructPB`
  round-trips through JSON, and the `[]string` branch has to stay for in-process callers
  anyway.
- Retain non-nil `UsageMetadata` across streaming chunks.
- Include and restore `Actions.ArtifactDelta`, reject out-of-range artifact versions
  instead of truncating, and keep nil `Actions` from panicking while preserving empty
  action messages on the wire.

### DIFF
--- a/internal/artifact/tests/service_suite.go
+++ b/internal/artifact/tests/service_suite.go
@@ -56,6 +56,15 @@ func TestArtifactService(t *testing.T, name string, factory func(t *testing.T) (
 		}
 		testArtifactService_UserScoped(ctx, t, srv, name)
 	})
+	t.Run(fmt.Sprintf("Test%sArtifactService_GetArtifactVersion", name), func(t *testing.T) {
+		ctx := t.Context()
+		// Create the service using the factory for this sub-test.
+		srv, err := factory(t)
+		if err != nil {
+			t.Fatalf("Failed to set up service: %v", err)
+		}
+		testArtifactService_GetArtifactVersion(ctx, t, srv, name)
+	})
 }
 
 func testArtifactService(ctx context.Context, t *testing.T, srv artifact.Service, testSuffix string) {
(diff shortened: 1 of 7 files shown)


karolpiotrowicz (commented):
The four cherry-picks match their originals, but the description describes a
revision of #670 that was explicitly abandoned upstream.

It says Run is converted to []any and that structpb.NewStruct fails when
persisting the response. Neither is true of what shipped: the final revision of
#670 reverted Run to []string and kept only the read-side type switch, and the
write path stopped calling structpb.NewStruct in #793. v1 already carries that
toStructPB JSON round-trip — identical to main's — so the write-path failure the
description opens with doesn't exist on this branch either. Worth rewriting the
first bullet of both Problem and Summary to cover the read path only, otherwise
the v1.6.0 notes will claim a fix that isn't in the diff.

wolo-lab:
> The four cherry-picks match their originals, but the description describes a revision of #670 that was explicitly abandoned upstream.
> 
> It says Run is converted to []any and that structpb.NewStruct fails when persisting the response. Neither is true of what shipped: the final revision of #670 reverted Run to []string and kept only the read-side type switch, and the write path stopped calling structpb.NewStruct in #793. v1 already carries that toStructPB JSON round-trip — identical to main's — so the write-path failure the description opens with doesn't exist on this branch either. Worth rewriting the first bullet of both Problem and Summary to cover the read path only, otherwise the v1.6.0 notes will claim a fix that isn't in the diff.

Thanks, updated.

karolpiotrowicz (approved):

PR #786, opened 2026-05-01
docs(llmagent): correct OutputSchema comment — tools still work via set_model_response

nuthalapativarun:
### Description of Change

**Problem:**
The comment for `OutputSchema` in `agent/llmagent/llmagent.go` incorrectly states that setting `OutputSchema` prevents the agent from using any tools. This is inaccurate and misleads users into thinking `OutputSchema` and tools are mutually exclusive.

**Solution:**
Updated the comment to reflect actual behavior: the framework transparently injects a `set_model_response` tool when `OutputSchema` is set, so all tool types (function tools, RAG, agent transfer, etc.) continue to work alongside structured output. See `internal/llminternal/outputschema_processor.go` for the implementation.

Fixes #756

### Testing Plan

**Unit Tests:**

- [ ] I have added or updated unit tests for my change.
- [x] All unit tests pass locally.

_Comment-only change; no behavior changes._

```
$ go test ./agent/llmagent/...
ok  	google.golang.org/adk/agent/llmagent	0.239s
```

**Manual End-to-End (E2E) Tests:**

Not applicable — documentation fix only.

### Checklist

- [x] I have read the [CONTRIBUTING.md](./CONTRIBUTING.md) document.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have manually tested my changes end-to-end.
- [x] Any dependent changes have been merged and published in downstream modules.

PR #1282, opened 2026-08-10
fix: backport web launcher h2c, deploy and documentation fixes (#826, #919, #786, #1212, #1162)

wolo-lab:
## Problem

A group of small user-visible defects still present on `v1`:

- The web launcher listener speaks HTTP/1 only, so a custom `web.Sublauncher` cannot
  mount an HTTP/2-aware handler (a native gRPC server, for instance) without standing up
  a second server.
- `adkgo deploy agentengine` hardcodes a `GOOGLE_API_KEY` secret env var, so the deploy
  fails for anyone who has not created that secret.
- The `load_skill` tool documents its parameter as `skill_name`; the actual parameter is
  `name`.
- The `OutputSchema` doc comment claims tools stop working, which is wrong — they still
  work via `set_model_response`.
- The README does not list the Kotlin and TypeScript ADK implementations.
- The `examples/agentengine` sample still pins `gemini-2.5-flash`, behind the rest of the
  v1 examples (`gemini-3.1-flash-lite`).
- The web launcher listener serves HTTP/1 only, so a custom `web.Sublauncher` cannot mount
  an HTTP/2-aware handler (a native gRPC/Connect server) without a second listener.

## Summary

Backports #826, #1162, #919, #786 and #1212 from `main`.

- Enable h2c on the web launcher listener so HTTP/1 and cleartext HTTP/2 share it, with
  no change to the public launcher APIs.
- Drop the hardcoded `GOOGLE_API_KEY` secret from the Agent Engine deploy path.
- Correct the `load_skill` parameter name and the `OutputSchema` comment.
- List the Kotlin and TypeScript implementations in the README.
- Drop the hardcoded `GOOGLE_API_KEY` secret from the Agent Engine deploy path and bump the
  `examples/agentengine` model from `gemini-2.5-flash` to `gemini-3.5-flash`. On `main` the
  same commit went `gemini-flash-latest` → `gemini-3.5-flash`; on `v1` the base is older, so
  it is a two-generation jump for the example. The end state matches `main`.
- Add an opt-in `-h2c` flag (default false) enabling prior-knowledge cleartext HTTP/2
  alongside HTTP/1 on the web launcher listener. No Go API change; the CLI grows one flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@
       <a href="https://google.github.io/adk-docs/">Docs</a> &
       <a href="https://github.com/google/adk-go/tree/main/examples">Samples</a> &
       <a href="https://github.com/google/adk-python">Python ADK</a> &
-      <a href="https://github.com/google/adk-java">Java ADK</a> & 
+      <a href="https://github.com/google/adk-java">Java ADK</a> &
+      <a href="https://github.com/google/adk-kotlin">Kotlin ADK</a> &
+      <a href="https://github.com/google/adk-js">TypeScript ADK</a> &
       <a href="https://github.com/google/adk-web">ADK Web</a>.
     </h3>
 </html>
--- a/agent/llmagent/llmagent.go
+++ b/agent/llmagent/llmagent.go
@@ -254,8 +254,10 @@ type Config struct {
 	InputSchema *genai.Schema
 	// The output schema when agent replies.
 	//
-	// NOTE: when this is set, agent can only reply and cannot use any tools,
-	// such as function tools, RAGs, agent transfer, etc.
+	// When OutputSchema is set, the framework transparently injects a
+	// set_model_response tool so the model can still invoke other tools
+	// (function tools, RAG, agent transfer, etc.) while producing structured
+	// output. See internal/llminternal/outputschema_processor.go for details.
 	OutputSchema *genai.Schema
 
 	// Callbacks are executed in the order they are provided.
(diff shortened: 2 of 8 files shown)


karolpiotrowicz (commented):
The five cherry-picks match their originals. Two notes on the description.

#1162 also bumps the example model, and the description mentions only the secret.
On main that was gemini-flash-latest → gemini-3.5-flash; on v1 the base is
gemini-2.5-flash, so the same commit is a two-generation jump here. The end state
matches main either way, but it's a user-visible change to the example and should
be listed.

#826 is a feature rather than a fix — it adds a -h2c flag to the web launcher.
It's opt-in and defaults to false, so the risk is low, but whether features go
into the v1 line at this point is a call for you rather than something the diff
can answer.

wolo-lab:
> The five cherry-picks match their originals. Two notes on the description.
> 
> #1162 also bumps the example model, and the description mentions only the secret. On main that was gemini-flash-latest → gemini-3.5-flash; on v1 the base is gemini-2.5-flash, so the same commit is a two-generation jump here. The end state matches main either way, but it's a user-visible change to the example and should be listed.
> 
> #826 is a feature rather than a fix — it adds a -h2c flag to the web launcher. It's opt-in and defaults to false, so the risk is low, but whether features go into the v1 line at this point is a call for you rather than something the diff can answer.

Thx, the description has been updated.

karolpiotrowicz (approved):

PR #1290, opened 2026-08-10
fix(llminternal): bound consecutive thought-only turns in the base flow

jjsasha63:
A thought-only ("thinking") turn reports as a final response but carries no answer, so since v2 Flow.Run deliberately does not stop on it and calls the model again. Nothing bounded that retry: a model that keeps returning only thoughts makes the run continue indefinitely, appending an event per turn and re-sending an ever-growing history on each call. A Go codebase migrating from v1.5 to v2 hit this as an integration test that hung rather than failed.

This is a v2 regression. v1 had no thought handling at all and stopped on any final response, so the loop could not spin; 893e4a4 (#1109) added the retry without a bound.

Cap the run at maxConsecutiveThoughtOnlyTurns consecutive thought-only turns, leaving the last thinking event as the result. Any turn that is not a final response resets the count, so only genuinely consecutive thought-only turns are bounded. Giving up is logged, because the bound is a guess about model behaviour: if a future model legitimately needs more thinking turns, that should surface in logs rather than silently truncating answers.

The bound is deliberately not configurable. It is a safety net against a degenerate model rather than a tuning knob, and a caller has no basis for choosing a value. It is a var only so tests can exercise the boundary.

This bounds the thought-only retry specifically. It does not bound the pre-existing tool-call loop, which is unbounded independently of thoughts; a general per-run LLM call limit, as in adk-python's RunConfig.max_llm_calls, would be a separate change.

Testing Plan:
- TestRun_ThoughtOnlyTurnsTerminate drives the flow with a model that only ever returns thoughts. Reverting the cap makes it yield >50 events and call the model 51 times before the test's safety limit trips; with the cap it stops at the bound.
- TestRun_ThoughtOnlyTurnCountResets interleaves a function-call turn between thought-only turns and asserts the count resets. Deleting the reset branch makes it fail, pinning the consecutive semantics.
- go build -mod=readonly work
- go test -race -mod=readonly -count=1 -shuffle=on work
- golangci-lint run (v2.3.1): 0 issues
- go mod tidy -diff: clean in both modules


internal/llminternal/base_flow.go
```diff
@@ -100,8 +100,22 @@ var (
...
+// at most maxConsecutiveThoughtOnlyTurns-1 times waiting for an answer. A model
+// that keeps emitting thoughts without ever surfacing an answer would otherwise
+// spin forever, appending an event per turn and re-sending an ever-growing
+// history. Any turn that is not a final response resets the count.
+//
+// This is a safety net against a degenerate model, not a tuning knob: the bound
+// is set well above the number of consecutive thinking turns a model is expected
+// to need, so reaching it should be rare enough to warrant the log below. It is
+// a var rather than a const only so tests can exercise the boundary.
+var maxConsecutiveThoughtOnlyTurns = 3
```
wolo-lab: I would change it to const (and remove the last sentence) as it is not modified in tests.

internal/llminternal/base_flow.go
```diff
@@ -100,8 +100,22 @@ var (
...
 
+// maxConsecutiveThoughtOnlyTurns bounds how many thought-only ("thinking")
+// turns in a row the flow accepts before giving up, so the model is re-called
+// at most maxConsecutiveThoughtOnlyTurns-1 times waiting for an answer. A model
+// that keeps emitting thoughts without ever surfacing an answer would otherwise
+// spin forever, appending an event per turn and re-sending an ever-growing
+// history. Any turn that is not a final response resets the count.
+//
+// This is a safety net against a degenerate model, not a tuning knob: the bound
+// is set well above the number of consecutive thinking turns a model is expected
```
wolo-lab: nit: "well above" - I think it was related to the previous version.

internal/llminternal/base_flow_test.go
```diff
@@ -771,3 +774,139 @@ func TestIsThoughtOnlyTurn(t *testing.T) {
...
+	if events != maxConsecutiveThoughtOnlyTurns {
+		t.Errorf("Run() yielded %d events, want %d (one per thought-only turn)", events, maxConsecutiveThoughtOnlyTurns)
+	}
+}
+
+// scriptedThinkingModel replays a fixed sequence of turns and then keeps
+// repeating the final one, so a test can interleave thought-only turns with a
+// turn that is not a final response.
+type scriptedThinkingModel struct {
+	model.LLM
```
wolo-lab: It seems to me that we can remove it, right?

internal/llminternal/base_flow_test.go
wolo-lab, line 1: We promise that in case exceeding the limit, we will return the last thinking event. Let's add an assertion for this.

internal/llminternal/base_flow.go
```diff
@@ -115,11 +129,25 @@ func (f *Flow) Run(ctx agent.InvocationContext) iter.Seq2[*session.Event, error]
...
+				// A thought-only ("thinking") turn reports as final but has no
+				// answer; don't stop on it — call the model again. Give up once
+				// the model has produced only thoughts too many times in a row,
+				// leaving the last thinking event as the result.
+				if !isThoughtOnlyTurn(lastEvent) {
+					return
+				}
+				thoughtOnlyTurns++
+				if thoughtOnlyTurns >= maxConsecutiveThoughtOnlyTurns {
+					log.Printf("adk: model produced %d consecutive thought-only turns without an answer; giving up and returning the last thinking event\n", thoughtOnlyTurns)
```
wolo-lab: I think we don't need this: "\n".

wolo-lab: It would not be very useful while debugging. I would add more details like: model name, invocation id, agent name.

wolo-lab (changes requested):

karolpiotrowicz (commented):
The hang is real and this stops it. On `main`, a model that only ever emits
thoughts runs until the consumer gives up. Two things about *how* it stops, on
top of the review already on this PR.

**A model that legitimately needs three consecutive thinking turns loses its
answer.** The cap tolerates two. Scripting a model as thought, thought, thought,
answer:

```
events=3  modelCalls=3  surfaced answer text=""
```

`Run` returns `err == nil` and the answer is never requested. That is a quieter
failure than the hang it replaces, and the last commit lowering the bound from 10
to 3 made it considerably easier to reach.

**Downstream can't distinguish a truncated run from an empty one.** The give-up
event carries empty `ErrorCode`, `ErrorMessage`, `FinishReason` and `Escalate`.
`workflow.synthesizeAgentOutput` skips `Thought` parts when building
`messageText`, so it caches `Output=""` with `MessageAsOutput=true` — the state
its own doc comment warns produces a bogus completion on resume. Seven lines
below the new code, the `Partial` branch surfaces an error for a comparable
situation.

**Nothing pins the bound's value.** Both new tests assert against
`maxConsecutiveThoughtOnlyTurns` itself
([L824](https://github.com/google/adk-go/blob/a266ad55f465a45d1dd239b9ae78ec00e4507a68/internal/llminternal/base_flow_test.go#L824),
[L908](https://github.com/google/adk-go/blob/a266ad55f465a45d1dd239b9ae78ec00e4507a68/internal/llminternal/base_flow_test.go#L908)),
so changing it changes behaviour without failing anything — the 10→3 commit
needed no test edit. One case with a literal would cover it.

On the review already here: the `const` change, the "well above" wording at
[L111](https://github.com/google/adk-go/blob/a266ad55f465a45d1dd239b9ae78ec00e4507a68/internal/llminternal/base_flow.go#L111),
the `model.LLM` embed at
[L836](https://github.com/google/adk-go/blob/a266ad55f465a45d1dd239b9ae78ec00e4507a68/internal/llminternal/base_flow_test.go#L836)
and the missing assertion that the last thinking event is returned all hold up.
The variable is only ever read, so `const` compiles as-is. The trailing `\n` is
the one I'd leave alone — `log.Output` adds a newline only when one is absent, so
there's no double newline, and every other `log.Printf` in the file carries it.

Line references are against `a266ad55f465a45d1dd239b9ae78ec00e4507a68`.

jjsasha63:
maxConsecutiveThoughtOnlyTurns is now a const, and the sentence explaining why it was a var is gone. It stays unexported: nothing outside internal/llminternal can read it, and the tests are in the same package.
Dropped the stale "well above" wording.
Removed the model.LLM embed from scriptedThinkingModel. 
The log line now names the model, the agent and the invocation ID.
Dropped the trailing \n.
Added the assertion that the last thinking event is returned, described below.

A model needing three thinking turns lost its answer
Confirmed, with the numbers you gave. Scripting the model as thought, thought, thought, answer against the previous commit:
The bound thinking turns is back to 10. 

Nothing pinned the bound's value
Fixed by the same assertion that answers wolo-lab's point, so one change covers both. The fake model now numbers each thought, and TestRun_ThoughtOnlyTurnsTerminate asserts the last yielded event carries exactly thinking 10.

I checked the assertion bites by mutating the source twice:
>= to > in the give-up condition fails with - "thinking 10" / + "thinking 11"
the bound from 10 to 9 fails with - "thinking 10" / + "thinking 9", and fails only on that assertion, because every other assertion compares against the constant
The second is exactly the gap you found — under the old tests, the 10 to 3 commit needed no test edit.

Downstream cannot distinguish a truncated run from an empty one
Not taken in this PR. You are right about the state: the give-up event carries empty ErrorCode, ErrorMessage and FinishReason, so workflow.synthesizeAgentOutput caches Output="" with MessageAsOutput=true, the case its own doc comment warns about.

The fix that matches the Partial branch you pointed at is to yield an error behind a sentinel:

var ErrThoughtOnlyLimit = errors.New("model produced only thoughts without an answer")

It changes the contract. Today the run ends with err == nil and the thinking event as the result, which is what this PR promises. Turning that into an error is a larger change than the hang fix and deserves its own review rather than riding along here.
The sentinel would not reach external callers. llminternal is internal and nothing re-exports its errors; ErrModelNotConfigured has the same problem today. For errors.Is to be useful to users of the module, the sentinel needs to live in a public package, and that is an API decision I would rather not make unilaterally.
Happy to fold it in here if you would prefer not to split it. Otherwise I will open a follow-up, which would pair naturally with LiveRunConfig.MaxLLMCalls in agent/live.go — it is declared and currently read nowhere, so a general per-run call limit has a home waiting.

wolo-lab (approved):
Thanks!

PR #1291, opened 2026-08-10
fix(openaimodel): serialize assistant messages with output_text type (#1197)

### LINKED ISSUE #1197: openaimodel: assistant messages sent with `input_text` content type → 400 on multi-turn (should be `output_text`)

omen18:
Fixes #1197

### Summary
The OpenAI model connector () previously serialized every message's content with type `input_text` regardless of role. When a prior assistant turn was replayed in a multi-turn conversation, OpenAI API rejected the request with HTTP 400 because assistant content must use type `output_text`.

### Solution
- Updated `flushText` in `model/openaimodel/request.go` to branch on `genai.RoleModel` and emit `ResponseInputItemUnionParam{OfOutputMessage: ...}` with `output_text` content using a helper `newOutputMessage`.
- Added unit test `TestBuildOpenAIParams_AssistantOutputText` verifying assistant turns are serialized with `output_text` content type.

### DIFF
--- a/model/openaimodel/request.go
+++ b/model/openaimodel/request.go
@@ -95,12 +95,23 @@ func convertContents(contents []*genai.Content) (responses.ResponseInputParam, e
 			if len(textParts) == 0 {
 				return nil
 			}
-			msg, err := newMessage(curRole, textParts)
-			if err != nil {
-				return err
-			}
-			if msg != nil {
-				items = append(items, responses.ResponseInputItemUnionParam{OfMessage: msg})
+			switch curRole {
+			case genai.RoleModel:
+				out, err := newOutputMessage(textParts)
+				if err != nil {
+					return err
+				}
+				if out != nil {
+					items = append(items, responses.ResponseInputItemUnionParam{OfOutputMessage: out})
+				}
+			default:
+				msg, err := newMessage(curRole, textParts)
+				if err != nil {
+					return err
+				}
+				if msg != nil {
+					items = append(items, responses.ResponseInputItemUnionParam{OfMessage: msg})
+				}
 			}
 			textParts = textParts[:0]
 			return nil
@@ -182,6 +193,22 @@ func newMessage(role genai.Role, texts []string) (*responses.EasyInputMessagePar
 			OfInputItemContentList: contentList,
 		},
 	}, nil
+}
+
+func newOutputMessage(texts []string) (*responses.ResponseOutputMessageParam, error) {
+	content := make([]responses.ResponseOutputMessageContentUnionParam, 0, len(texts))
+	for _, txt := range texts {
+		if strings.TrimSpace(txt) == "" {
+			continue
+		}
+		content = append(content, responses.ResponseOutputMessageContentUnionParam{
+			OfOutputText: &responses.ResponseOutputTextParam{Text: txt},
+		})
+	}
+	if len(content) == 0 {
+		return nil, nil
+	}
+	return &responses.ResponseOutputMessageParam{Content: content}, nil
 }
 
 func normalizeRole(role genai.Role) (responses.EasyInputMessageRole, error) {
--- a/model/openaimodel/request_test.go
+++ b/model/openaimodel/request_test.go
@@ -616,3 +616,40 @@ func TestNewJSONSchemaFormat(t *testing.T) {
 		})
 	}
 }
+
+func TestBuildOpenAIParams_AssistantOutputText(t *testing.T) {
+	req := &model.LLMRequest{
+		Contents: []*genai.Content{
+			genai.NewContentFromText("hello", genai.RoleUser),
+			genai.NewContentFromText("hi there!", genai.RoleModel),
+			genai.NewContentFromText("how are you?", genai.RoleUser),
+		},
+	}
+	params, err := buildOpenAIParams("fallback", req)
+	if err != nil {
+		t.Fatalf("buildOpenAIParams() err = %v", err)
+	}
+	items := params.Input.OfInputItemList
+	if len(items) != 3 {
+		t.Fatalf("expected 3 items, got %d", len(items))
+	}
+	// Item 0: user (message)
+	if items[0].OfMessage == nil {
+		t.Fatalf("item 0 should be OfMessage")
+	}
+	// Item 1: assistant (output message with output_text)
+	if items[1].OfOutputMessage == nil {
+		t.Fatalf("item 1 (assistant turn) should be OfOutputMessage, got %+v", items[1])
+	}
+	outParts := items[1].OfOutputMessage.Content
+	if len(outParts) != 1 || outParts[0].OfOutputText == nil {
+		t.Fatalf("unexpected assistant output content: %+v", outParts)
+	}
+	if got, want := outParts[0].OfOutputText.Text, "hi there!"; got != want {
+		t.Fatalf("assistant text mismatch got=%q want=%q", got, want)
+	}
+	// Item 2: user (message)
+	if items[2].OfMessage == nil {
+		t.Fatalf("item 2 should be OfMessage")
+	}
+}


hanorik (changes requested):
Thanks for this, and sorry for the slow response.

Heads up: **#1205 fixes #1197 with essentially the same approach** and is
awaiting requested changes — credit to @PratikDhanave for getting there first.
Whichever PR lands the points below is the one I'll merge.

**Blocker:** the CLA check is failing; nothing can merge until it's signed.

## Test gaps (I mutated your branch to confirm each)

1. **Ordering isn't pinned.** Deleting the `flushText()` before the
   function-call append (`request.go:133`) leaves the whole package green,
   though it genuinely reverses the emitted order.
   `TestBuildOpenAIParams_FunctionCall` doesn't cover it — its model turn has no
   text parts. Add a model turn with text *and* a function call, asserting
   `OfOutputMessage` then `OfFunctionCall`.

2. **`id` absence isn't asserted**, and it's the load-bearing assumption here:
   OpenAI mints message IDs on output, so a replayed turn must omit it. Adding
   `ID: "bogus id with spaces"` at `request.go:211` keeps the suite green — the
   exact value class that 400s (openai/openai-go#334).

3. **Assert the wire format, not the structs.** #1205 marshals the assistant
   item and checks it contains `"output_text"` and never `"input_text"`; that's
   what actually proves the 400 is fixed. Also assert user turns keep
   `input_text` — you only check `OfMessage != nil`.

## Code

4. **`Status` is never set** — the one real wire difference between the two PRs:

   ```text
   #1291: {"content":[...],"role":"assistant","type":"message"}
   #1205: {"content":[...],"status":"completed","role":"assistant","type":"message"}
   ```

   It's `json:"status,omitzero" api:"required"` in openai-go v3.50.0, so it
   drops out of the payload entirely. Set
   `responses.ResponseOutputMessageStatusCompleted`, or comment why not.

5. `normalizeRole`'s `genai.RoleModel` branch (`request.go:218-219`) is now
   unreachable from production, but `request_test.go:379` still asserts it — a
   green signal for a dead path. Worth a comment: anyone later adding `case
   assistant:` there would silently reintroduce this bug.

6. `newOutputMessage` returns an always-nil error — just return the pointer. It
   also needs a doc comment stating the "ID deliberately unset" rationale that
   (2) depends on.

7. Prefer table-driven tests; this is a single shape.

## Don't copy this bit from #1205

It sets `Type: constant.OutputText("output_text")` explicitly. That's redundant
— the field is tagged `default:"output_text"` and both spellings marshal
identically. Your version is fine as-is.

Its comment also claims `EasyInputMessageParam` "can only emit input_text".
Not quite: a plain-string assistant message is legal
(`{"content":"...","role":"assistant","type":"message"}`, which is what
adk-python emits). The output-message approach is fine, just don't justify it
that way.

wojtas000:
Since there's been no activity since the review, I've opened https://github.com/google/adk-go/pull/1427 to carry it forward. 
And your CLA check is failing, which blocks merge regardless of the code (worth signing if you plan to contribute in the future)

wojtas000:
The fix has been merged in https://github.com/google/adk-go/pull/1427. Closing

PR #1292, opened 2026-08-10
fix(workflow): make ParallelWorker fail-fast stop new dispatch under maxConcurrency (#1238)

### LINKED ISSUE #1238: ParallelWorker: fail-fast cancellation doesn't stop new items from being dispatched under maxConcurrency

omen18:
Fixes #1238

### Summary
`ParallelWorker.Run` documents that a non-retryable error from one item makes the workflow fail fast and cancel other in-flight workers. Under `maxConcurrency > 0` this did not hold: once one item failed, every other item still waiting on the concurrency semaphore got dispatched anyway and ran to completion before the error surfaced.

### Solution
- Checked `workerCtx.Done()` in dispatch loop `select` instead of the root `ctx.Done()`.
- Added a non-blocking check on `workerCtx.Done()` after acquiring the semaphore slot to avoid race conditions.
- Updated `runWorker` to invoke `cancelFunc()` synchronously upon non-retryable worker failure before releasing its semaphore slot.
- Added unit test `TestParallelWorker_MaxConcurrencyFailFast` verifying that items after a fail-fast failure are not dispatched under `maxConcurrency`.

### DIFF
--- a/workflow/parallel_worker.go
+++ b/workflow/parallel_worker.go
@@ -112,16 +112,23 @@ func (n *ParallelWorker) Run(ctx agent.Context, input any) iter.Seq2[*session.Ev
 			if sem != nil {
 				select {
 				case sem <- struct{}{}:
-				case <-ctx.Done():
+				case <-workerCtx.Done():
 					wg.Done()
 					continue
+				}
+				select {
+				case <-workerCtx.Done():
+					<-sem
+					wg.Done()
+					continue
+				default:
 				}
 			}
 
 			itemBranch := deriveSubBranch(parentBranch, wrappedName+"@"+strconv.Itoa(i+1))
 
 			itemCtx := workerCtx.WithDelta(&agent.CommonContextDelta{InvocationContextDelta: &agent.InvocationContextDelta{Branch: &itemBranch}})
-			go n.runWorker(itemCtx, i, item, sem, resCh, &wg)
+			go n.runWorker(itemCtx, i, item, sem, resCh, &wg, cancelFunc)
 		}
 
 		// Goroutine to close channel when all workers are done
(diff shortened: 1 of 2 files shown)


karolpiotrowicz:
This makes the same change as #1239 — both switch the dispatch-loop select to `workerCtx.Done()` and re-check cancellation after acquiring the semaphore slot. #1239 also guards the unbounded path where `maxConcurrency` is 0, which this PR leaves open, so we are consolidating there.

Closing this one to keep the review in a single place. The fix here was correct and the credit stands.

PR #1297, opened 2026-08-10
ci: guard the exported API against accidental breaking changes

baptmont:
## What

Adds an `apidiff` check that fails a pull request on incompatible changes to the module's exported API. Adding to the API is always allowed.

## Why

The build and the tests do not catch every break. Adding a trailing variadic parameter to an exported function changes that function's *type*: every ordinary call site still compiles, so nothing goes red, while any caller that held the function as a value stops compiling.

That is not hypothetical. It happened to `NewRuntimeAPIController`, `NewPubSubController` and `NewEventarcController`, and reached a release. A test written specifically to pin backward compatibility did not catch it either, because a plain call expression is exactly the form a trailing variadic can never break.

## Two details worth reviewing

- **It compares against the merge base, not the target tip.** My first draft compared against `origin/main` and produced two convincing false positives on a branch that had been cut before #1252 landed, reporting it as having *removed* the methods that PR added. A check that cries wolf gets ignored, so this seemed worth getting right.
- **`internal/` packages are excluded**, since they are not public API.

## Testing Plan

Verified end to end against real commits rather than a synthetic case:

- Against the commit that introduced the variadics, it reports both constructors and exits 1.
- Against the commit that restores the signatures, it exits 0 with no findings.
- The pinned `apidiff` version resolves and installs from a clean module.
- Workflow YAML parses; the script passes `bash -n`.

Local run on the breaking commit:

```
=== google.golang.org/adk/v2/server/adkrest/controllers/triggers ===
Incompatible changes:
- NewPubSubController: changed from func(..., TriggerConfig) *PubSubController
  to func(..., TriggerConfig, ...ControllerOption) *PubSubController
```

## Notes

This is independent of the context-compaction stack (#1231 to #1236), though that is where the motivating break was found and fixed.


baptmont:
Pushed 3c1d2e21, which changes how the comparison is driven. The description above still describes the per-package version, so, in short:

**One `apidiff -m` per module instead of one per exported package.** The per-package loop took **11m16s** on a warm cache and over 15 minutes cold — 87 invocations, each reloading and type-checking the whole dependency graph. Module mode reports the same findings in **22s**.

**It now covers `plugin/agentanalytics`.** `go list ./...` only ever saw the root module, so the submodule — which is public API — was going unchecked. Module mode also skips `internal/` by itself, so the grep filtering is gone.

**A module that cannot be read now fails the check.** Every `apidiff` call carried a trailing `|| true`. A base revision that failed to build wrote no snapshots, every package was then skipped for want of one, and the job went green having compared nothing — the one outcome worse than not having the check. `apidiff` exits non-zero only when it could not do its job (reporting a break is a *successful* run), so the two cases separate cleanly. The `breaking-change` label deliberately does not cover this one: a comparison that did not happen is not a break anyone decided to make.

Re-verified the same cases: clean tree exits 0; a trailing variadic on `runner.New` and on `NewBatchProcessor` is reported and exits 1; the label still downgrades a break to a report; a module that does not compile fails even with the label set.

Backport to `v1` is #1304 — both files there are byte-identical to these.

Generated with CloudCode, session `ses_00f468febffe7dsyc1CNVAgbg7`.

kdroste-google (approved):
LGTM

PR #1301, opened 2026-08-11
fix(skilltoolset): accept scalar allowed-tools frontmatter

### LINKED ISSUE #1296: skill.Frontmatter.AllowedTools does not follow the spec
## 🔴 Required Information

**Describe the Bug:**
`skill.Frontmatter` declares `AllowedTools []string`, so `skill.Parse` only accepts allowed-tools written as a YAML list. The Agent Skills specification, which frontmatter.go links to in its own doc comment, defines the field as a space-separated string:

> | allowed-tools | No | Space-separated string of pre-approved tools the skill may use. (Experimental) |

The example below  does not parse in ADK Go. The type accepts only a YAML list, which the specification never describes, and rejects the scalar form it does describe.

```yaml
allowed-tools: Bash(git:*) Bash(jq:*) Read
```

For reference, ADK Python models the same field as a string and matches the spec([source](https://github.com/google/adk-python/blob/0477e5743bdf5e0cce5a698951c7b70a07baa80a/src/google/adk/skills/models.py#L71)):

```python
allowed_tools: Optional[str] = Field(
    None,
    alias="allowed-tools",
    serialization_alias="allowed-tools",
)
```

**Steps to Reproduce:**
1. Create some test files with dummy skills inside a `skills` dir

skills/spec-skill/SKILL.md - allowed-tools as the spec documents it: a space-separated string.
```
---
name: spec-skill
description: Uses the allowed-tools form the spec documents.
allowed-tools: Bash(git:*) Bash(jq:*) Read
---
Instructions.
SPEC
```

skills/list-skill/SKILL.md - allowed-tools as a YAML list, which the spec does not describe.
```
---
name: list-skill
description: Uses a YAML list, which the spec does not describe.
allowed-tools:
  - Bash(git:*)
  - Bash(jq:*)
  - Read
---
Instructions.
LIST
```

skills/other-skill/SKILL.md - An unrelated, valid skill with no allowed-tools at all.
```
---
name: other-skill
description: An unrelated, perfectly valid skill.
---
Instructions.
OTHER
```

2. Init and create a small program to load them:

main.go
```go
package main

import (
	"context"
	"fmt"
	"os"

	"google.golang.org/adk/v2/tool/skilltoolset/skill"
)

func main() {
	ctx := context.Background()
	src := skill.NewFileSystemSource(os.DirFS("skills"))

	for _, name := range []string{"spec-skill", "list-skill", "other-skill"} {
		fm, err := src.LoadFrontmatter(ctx, name)
		if err != nil {
			fmt.Printf("LoadFrontmatter(%s) -> REJECTED: %v\n", name, err)
			continue
		}
		fmt.Printf("LoadFrontmatter(%s) -> accepted: %q\n", name, fm.AllowedTools)
	}

	// One unloadable skill fails the whole listing, hiding the other two.
	fms, err := src.ListFrontmatters(ctx)
	fmt.Printf("\nListFrontmatters -> returned %d of 3 skills\n", len(fms))
	fmt.Printf("ListFrontmatters -> err: %v\n", err)
```

3. Run `go mod tidy && go run .`

**Expected Behavior:**

`allowed-tools: Bash(git:*) Bash(jq:*) Read` parses, per the specification the package links to, and `ListFrontmatters` returns all three skills.

**Observed Behavior:**

The spec form is rejected, the form the spec does not describe is the only one accepted, and the two unrelated valid skills are lost along with it:

```text
LoadFrontmatter(spec-skill) -> REJECTED: invalid frontmatter: parse frontmatter: parse frontmatter: yaml: unmarshal errors:
  line 3: cannot unmarshal !!str `Bash(gi...` into []string
LoadFrontmatter(list-skill) -> accepted: ["Bash(git:*)" "Bash(jq:*)" "Read"]
LoadFrontmatter(other-skill) -> accepted: []

ListFrontmatters -> returned 0 of 3 skills
ListFrontmatters -> err: invalid frontmatter: parse frontmatter: parse frontmatter: yaml: unmarshal errors:
  line 3: cannot unmarshal !!str `Bash(gi...` into []string
```

**Environment Details:**

- ADK Library Version: v2.2.0 (also reproduced on v2.1.0; `main` carries the same declaration at the time of writing)
- OS: macOS
- Go Version: go1.26.5 darwin/arm64

**Model Information:**

Which model is being used: N/A — this is in frontmatter parsing, before any model is involved.

---

## 🟡 Optional Information

**Regression:**

No

**Additional Context:**

In the wild we see two main forms for this property, neither of which loads today:

- `allowed-tools: Bash(git:*) Bash(jq:*) Read` — the spec's space-separated string.
- `allowed-tools: Read, Bash(kubectl get:*)` — the comma-separated form Claude Code skills commonly use. Not in the spec, but common enough in existing skill files to be worth tolerating.

Note that a tool entry may itself contain spaces (`Bash(git add:*)`), so a plain `strings.Fields` split would break those; splitting on separators outside the parentheses handles both spellings.

A possible fix that keeps existing callers working: keep `AllowedTools` a `[]string` for consumers, and give `Frontmatter` a custom `UnmarshalYAML` that accepts either a sequence or a scalar, splitting the scalar into items. That stays compatible with skills already written as a YAML list against the current behaviour, while accepting the spec form. If instead the field is changed to a plain `string` to match ADK Python exactly, I think that is a breaking change for anyone relying on the current list decoding.

See the script under "Steps to Reproduce".

 - Always (100%)

ktsoator:

Closes: #1296

`skill.Frontmatter.AllowedTools` previously accepted only YAML sequences, while the Agent Skills specification defines `allowed-tools` as a scalar string.

This change:
- accepts the specification scalar form;
- preserves compatibility with YAML sequences;
- keeps the public `AllowedTools []string` API unchanged;
- avoids splitting whitespace and commas inside tool parentheses.

### Testing Plan

- `golangci-lint run` in the root and `plugin/agentanalytics` modules


ktsoator:
CLA details have been updated. @googlebot rescan

tool/skilltoolset/skill/frontmatter.go
```diff
@@ -43,6 +44,87 @@ type Frontmatter struct {
...
+			return fmt.Errorf("allowed-tools must be a string or a sequence of strings")
+		}
+		*a = splitAllowedTools(node.Value)
+		return nil
+	default:
+		return fmt.Errorf("allowed-tools must be a string or a sequence of strings")
+	}
+}
+
+func splitAllowedTools(value string) []string {
```
kdroste-google: How about adding an error to returned values? 
We should emit an error on unexpected input - like unbalanced '(', ')' 

It would make sense provide some information (in both the code and the error message) about the accepted format - whitespace /comma separated list of items like Tool|Tool(WellBalancedExpression)

tool/skilltoolset/skill/frontmatter.go
```diff
@@ -43,6 +44,87 @@ type Frontmatter struct {
...
+
+	for _, r := range value {
+		switch {
+		case r == '(':
+			depth++
+			current.WriteRune(r)
+		case r == ')':
+			if depth > 0 {
+				depth--
+			}
```
kdroste-google, line 113: We should emit an Error on negative depth

tool/skilltoolset/skill/frontmatter.go
```diff
@@ -43,6 +44,87 @@ type Frontmatter struct {
...
+			current.WriteRune(r)
+		case depth == 0 && (unicode.IsSpace(r) || r == ','):
```
kdroste-google, line 116: It accepts ",,,,  , ,," but I think it's OK, the resulting list should be empty anyway

tool/skilltoolset/skill/frontmatter_test.go
```diff
@@ -122,6 +122,26 @@ func TestParse_Valid(t *testing.T) {
...
+			input: "---\nname: skill\ndescription: skill\nallowed-tools: Bash(git:*) Bash(jq:*) Read\n---\nMarkdown Body",
+			want: &Frontmatter{
+				Name:         "skill",
+				Description:  "skill",
+				AllowedTools: []string{"Bash(git:*)", "Bash(jq:*)", "Read"},
+			},
+			wantInstruction: "Markdown Body",
+		},
+		{
+			name:  "allowed tools scalar with spaces inside parentheses",
```
kdroste-google: allowed comma separated tools ... ?

kdroste-google (changes requested):
Hi, @ktsoator, thank you for this contribution! 
It all makes perfect sense, just a few remarks from my side.
At the moment I don't see a need for allowing unbalanced / weird ordered '(', ')' in patterns. WDYT?

ktsoator:
Hi @kdroste-google,

Thank you for taking the time to review this PR and for the thoughtful feedback.

I have been using ADK Go for a while, and my team has also used it to deliver a real project at my company. The framework has been genuinely useful to us, so I wanted to contribute some fixes and improvements back to the repository where I can.

Your point about unbalanced parentheses makes sense. Since the parser already treats parentheses structurally when preserving spaces and commas inside tool expressions, silently accepting malformed expressions would be surprising.

I have updated the implementation to:

- return an error for an unexpected `)`;
- return an error when an opening `(` is left unclosed;
- document the accepted whitespace/comma-separated `Tool` and `Tool(...)` formats in the code and error message;
- add tests for unbalanced and incorrectly ordered parentheses;
- rename the existing comma-separated test to make that coverage explicit.

Separator-only input such as `",,,,  , ,"` remains accepted as an empty list, as discussed.

The build, race tests, lint, and module tidy checks all pass locally after the changes. Thanks again for helping improve the implementation.

tool/skilltoolset/skill/frontmatter.go
ktsoator: Agreed. I kept this behavior unchanged: separator-only input is accepted and produces an empty list.

tool/skilltoolset/skill/frontmatter.go
ktsoator: Thank you. Addressed in `12f3503`: `splitAllowedTools` now returns `([]string, error)`, and both the code comment and error message describe the accepted whitespace/comma-separated `Tool` or `Tool(...)` format.

tool/skilltoolset/skill/frontmatter.go
ktsoator: Addressed in `12f3503`. An unexpected closing `)` at depth zero now returns an error, and tests cover both an unexpected closing parenthesis and a closing parenthesis before an opening one.

tool/skilltoolset/skill/frontmatter_test.go
ktsoator: The comma-separated form is covered by the following test using `Read, Bash(kubectl get:*)`. I renamed it to `allowed tools comma-separated with spaces inside parentheses` in `12f3503` to make that coverage explicit.

kdroste-google (approved):
@ktsoator, thank you for those fixes!
I'm glad to hear that you find ADK-GO useful - waiting for your next contributions!

VDVsx:
Thanks both for the fast turnaround on this one!

PR #1311, opened 2026-08-12
docs: replace the nonexistent TempStatePrefix in state delta comments

Trainingcqy:
## Summary

Two doc comments reference `TempStatePrefix`, an identifier that does not exist in the repository:

- `internal/sessionutils/utils.go:30`
- `session/database/service.go:482`

Changed to the constants they actually use:

- `internal/sessionutils` uses the unexported `tempPrefix` from the same package
- `session/database` uses `session.KeyPrefixTemp`

Comment only, no behavior change.


wolo-lab (approved):
LGTM, thank you!

PR #1312, opened 2026-08-12
fix: implement reconnection policy with bounded retries and backoff for live sessions

hanorik:
`Flow.RunLive`'s connect loop treats a set of errors as resumable (`io.EOF`, `broken pipe`, `connection reset`, `1008`, `GoAway`) and reconnects on them, but it does so with no backoff, no jitter, no attempt cap and no time budget. Against an endpoint that completes the websocket handshake and then immediately hangs up — a GoAway, a 1008 policy close, an LB draining a backend — this becomes a hot loop that redials as fast as it can dial, measured at ~1200 attempts/sec. A single live session will burn a core, issue hundreds of TLS handshakes per second at the Live endpoint, and emit two log lines per attempt; with no jitter, every client retries in lockstep, so a transient regional blip can be amplified into a sustained outage. Worse, resumable errors are swallowed rather than pushed to the event stream, so the loop never terminates and never reports failure: the caller's iterator simply goes silent forever.

This adds a bounded, paced reconnect policy. Retries now back off exponentially (250ms base, ×2, capped at 5s) with ±20% jitter to desynchronise clients, over a budget of 8 retries — roughly a 24s window worst case, chosen so the fix does not simply trade a hot loop for dropping every session that hits a brief network blip. A connection that stays up for 30s counts as progress and restores the budget, so a long-lived healthy session is not killed by occasional GoAways, while a backend that accepts and dies immediately never earns the budget back and is bounded as intended. When the budget is exhausted the caller now receives an error instead of silence. Failed redials and resumable `SendHistory` failures are charged to the same budget — without that the backoff would be unreachable during a real outage, since the first failed dial would end the session — though a first-ever connect failure still fails fast, as that indicates bad credentials or an unknown model rather than a blip. The backoff is interruptible, so `Close()` and context cancellation stay prompt, and no socket is dialled after teardown.


karolpiotrowicz:
Against a fake Live endpoint that completes the websocket handshake and then hangs up, `main` redials 157,138 times in 90 seconds and the caller's iterator never returns. On this branch the same endpoint gives 9 dials and terminates after 21.09 s with `giving up after 8 consecutive reconnect attempts` reaching the caller. The bound works, and the interruptible backoff holds up under both `Close()` and context cancellation. Four things I'd want addressed, ranked by how much they'd cost a caller.

**`Send` now blocks for the entire reconnect window, and the caller has no way to bound it.** [`inputCh`](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L254) is unbuffered, and its only reader is the per-connection writer goroutine [started at L526](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L526), which `cleanup()` kills before the loop backs off. [`Send`](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L261-L268) selects only on `inputCh` and `done`, takes no context and has no deadline. On the path where redials keep failing ([L436-L441](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L436-L441)) no writer goroutine is created for the whole budget, so a single `Send` issued during an outage waits out all of it. Measured on this branch with the shipped defaults: **19.9 s**, returning `io.EOF`. Before the change the same event unblocked `Send` in roughly one dial round-trip. I don't think there's an obvious right answer here — failing fast would drop a caller's audio during a 250 ms reconnect the policy exists to ride out — but a caller that gets no signal and can pass no deadline can't make that trade itself.

<details>
<summary>Repro — drop into <code>internal/llminternal</code>, then <code>go test -run TestSendStallsForTheWholeReconnectBudget ./internal/llminternal/</code></summary>

```go
package llminternal

import (
	"iter"
	"testing"
	"time"

	"github.com/gorilla/websocket"
	"google.golang.org/genai"

)

// Shipped defaults: no reconnect policy is set, exactly as
// agent/llmagent/llmagent.go builds the Flow for RunLive.
func TestSendStallsForTheWholeReconnectBudget(t *testing.T) {
	// Dial 1 connects and dies; every later dial is refused, so no writer
	// goroutine exists for the whole budget.
	client, dials := startRefusingLiveServer(t, 1, func(conn *websocket.Conn) {})

	f := &Flow{
		Model:             &fakeLiveModel{client: client},
		RequestProcessors: []func(ctx agent.InvocationContext, req *model.LLMRequest, f *Flow) iter.Seq2[*session.Event, error]{liveConfigProcessor},
	}
	ctx, cancel := newLiveInvocationContext(t)
	defer cancel()

	sess, seq, err := f.RunLive(ctx)
	if err != nil {
		t.Fatalf("RunLive failed: %v", err)
	}
	drained := drainAsync(seq)
	waitForConns(t, dials, 2) // dial 1 accepted, dial 2 refused: we are in the budget

	start := time.Now()
	sendErr := sess.Send(agent.LiveRequest{
		Content: &genai.Content{Role: "user", Parts: []*genai.Part{{Text: "ping"}}},
	})
	blocked := time.Since(start)
	_ = awaitDrain(t, drained)

	if blocked > time.Second {
		t.Errorf("Send blocked for %v (err=%v) after %d dials; a caller pumping realtime audio has no way to bound this",
			blocked, sendErr, dials.Load())
	}
}
```

```
--- FAIL: TestSendStallsForTheWholeReconnectBudget (20.16s)
    zz_send_stall_test.go:45: Send blocked for 19.879310899s (err=EOF) after 9 dials; a caller pumping realtime audio has no way to bound this
```

</details>

**The budget re-arm has no ceiling, so two backend shapes still reconnect forever.** [`madeProgress`](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/live_reconnect.go#L74-L76) is lifetime-only, and when it fires [L641](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L638-L644) sets `attempt = 0`, after which [`if attempt > 0`](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L392) skips the budget check and the backoff together. Nothing anywhere counts total reconnects or bounds the session's lifetime. One shape is a backend that holds each connection just past `resetAfter` before dropping it. The other is any backend at all when the iterator's consumer is slow, because [`connectedAt`](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L450) is stamped at dial time while the elapsed value is read [at L638](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L638-L644) only when the main goroutine dequeues from the unbuffered [`errChan`](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L462), so the consumer's own latency is charged to the connection as uptime. Against a server that hangs up instantly on every connection, with the shipped 30 s threshold and a consumer slower than that, I measured 3 dials in 150 s and no error ever delivered. Neither shape is a regression, and both run at least four orders of magnitude slower than the loop this replaces, so I read it as the fix being incomplete rather than wrong. Worth saying which one you intend, though: as written the guarantee is a bound on rate, not on count, and `live_reconnect.go` doesn't say so.

**The loop-top teardown guard isn't exercised by anything, and the comment above its test says it is.** [`TestRunLiveNoDialAfterTeardownOnResetPath`](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/live_reconnect_test.go#L450) [closes the session while connection 1 is still up](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/live_reconnect_test.go#L479-L481), so the flow is parked in the inner select and leaves through the pre-existing [`case <-sess.done`](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L651). The outer loop is never re-entered, so [the guard](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L386) is never evaluated. Deleting that guard outright leaves the whole package green and leaves that test green 30 runs out of 30, and `go test -covermode=atomic -coverprofile=cover.out ./internal/llminternal/` reports the guard body and both `return true` arms of `tornDown` as never executed. [The comment at L446-L449](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/live_reconnect_test.go#L446-L449) says "Without the loop-top check a closed session still gets dialled" — that's exactly the property worth pinning, and it's the one the test doesn't reach.

**The resumable-`SendHistory` retry has no test at all.** [L473-L481](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L473-L481) is one of the three behaviours the description advertises. The only test on that path, `TestRunLiveClosesConnectionOnSendHistoryFailure`, injects [`math.NaN()`](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow_live_test.go#L405-L417), so the failure is a client-side marshal error and [`isResumable`](https://github.com/google/adk-go/blob/7753f82e0612d32946eb8ca8cc256a6daa0e3923/internal/llminternal/base_flow.go#L356-L369) matches none of `io.EOF`, `broken pipe`, `connection reset`, `EOF`, `1008` or `GoAway` against it. The branch reports zero executions. It's worth more than an ordinary coverage gap because `attempt++` lives inside it — lose that one line and the branch becomes a bare `continue` with `attempt == 0`, which skips both the budget check and the backoff and restores the loop this change exists to remove, with the suite still green.

None of these block merge from my side. The change is a large net improvement on what's there now, and the `Send` stall is the only one I'd want a decision on before it lands rather than after.

wolo-lab:
Closing as stale, feel free to reopen.

PR #1314, opened 2026-08-12
fix: implement reconnection policy with bounded retries and backoff for live sessions

hanorik:
This PR duplicates #1312 for `v1` branch.

### DIFF
--- a/internal/llminternal/base_flow.go
+++ b/internal/llminternal/base_flow.go
@@ -93,6 +93,41 @@ type Flow struct {
 	BeforeToolCallbacks   []BeforeToolCallback
 	AfterToolCallbacks    []AfterToolCallback
 	OnToolErrorCallbacks  []OnToolErrorCallback
+
+	// reconnect bounds RunLive's reconnect attempts. Nil means the defaults;
+	// only tests set it, to shrink the delays.
+	reconnect *reconnectPolicy
+}
+
+// tornDown reports whether the session was closed or the invocation cancelled.
+func tornDown(ctx context.Context, sess *liveSessionImpl) bool {
+	select {
+	case <-sess.done:
+		return true
+	case <-ctx.Done():
+		return true
+	default:
+		return false
+	}
+}
+
+// waitBeforeReconnect paces a reconnect, reporting whether it should go ahead.
+// Teardown during the wait cuts it short, so Close stays prompt.
+//
+// The re-check after the timer matters: select picks uniformly among ready
+// cases, so a teardown landing as the timer fires would otherwise report
+// "proceed" half the time and dial a socket nobody reads.
+func waitBeforeReconnect(ctx context.Context, sess *liveSessionImpl, d time.Duration) bool {
+	t := time.NewTimer(d)
+	defer t.Stop()
+	select {
+	case <-t.C:
+		return !tornDown(ctx, sess)
+	case <-sess.done:
+		return false
+	case <-ctx.Done():
+		return false
+	}
 }
 
 var (
@@ -333,7 +368,44 @@ func (f *Flow) RunLive(ctx agent.InvocationContext) (agent.LiveSession, iter.Seq
 
 		iCtx, isIContext := ctx.(*icontext.InvocationContext)
 
+		policy := f.reconnect
+		if policy == nil {
+			policy = defaultReconnectPolicy()
+		}
+		// attempt counts consecutive reconnects that made no progress; lastErr
+		// is what triggered the last one.
+		attempt := 0
+		var lastErr error
+		everConnected := false
+
 		for {
+			// Every iteration, not just when backing off: a reconnect that
+			// skips the delay must not dial after teardown either.
+			if tornDown(ctx, sess) {
+				if err := ctx.Err(); err != nil {
+					sess.pushError(err)
+				}
+				return
+			}
+			if attempt > 0 {
+				if attempt > policy.maxRetries {
+					// Resumable errors are swallowed while retrying, so
+					// without this the caller's stream just goes quiet.
+					sess.pushError(fmt.Errorf("live session: giving up after %d consecutive reconnect attempts: %w", policy.maxRetries, lastErr))
+					return
+				}
+				delay := policy.delay(attempt)
+				log.Printf("live session: reconnect attempt %d/%d in %v", attempt, policy.maxRetries, delay)
+				if !waitBeforeReconnect(ctx, sess, delay) {
+					// Cancellation reports itself, matching the consumer
+					// loop's ctx.Done arm; Close is a clean teardown.
+					if err := ctx.Err(); err != nil {
+						sess.pushError(err)
+					}
+					return
+				}
+			}
+
 			if isIContext {
 				handle := iCtx.LiveSessionResumptionHandle()
 				if handle != "" {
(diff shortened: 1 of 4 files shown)


karolpiotrowicz:
The reconnect loop does what it says on a dead endpoint — against a fake Live server that completes the handshake and hangs up, `main` before this change redials about 1,700 times a second and never reports anything to the caller, and with this change it stops after 9 dials in 20 seconds and surfaces the exhausted budget. The backoff maths, the interruptible wait, and the first-dial-is-fatal rule all hold up. Four things below, one of which I think is worth fixing before this lands.

**The retry budget re-arms on a slow consumer, which can keep a dead session alive indefinitely.** [`connectedAt`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/base_flow.go#L448) is stamped at the dial, but the duration handed to `madeProgress` is only computed when the consumer loop dequeues the error at [`base_flow.go:636`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/base_flow.go#L636). `errChan` is unbuffered, so the reader goroutine parks on [the send](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/base_flow.go#L492) until the consumer comes back round — and the consumer can be blocked arbitrarily long in [`pushEvent`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/base_flow.go#L562), which waits on the caller, or in [`handleFunctionCalls`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/base_flow.go#L598), which runs tools inline. That stall is credited as connection uptime. Once it exceeds `resetAfter`, [`attempt = 0`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/base_flow.go#L639) and the [`attempt > 0`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/base_flow.go#L390) guard at the loop top is then skipped, so the redial is unpaced as well as uncharged.

At the shipped defaults I instrumented the reset line and ran a consumer that takes 31s per event: it fired three times in 200 seconds, and the last connection scored as progress had been credited 1m2s of uptime against a real lifetime of milliseconds. The budget never accumulates, so the give-up path is never reached and the caller's stream stays silent. Both of the blocking sites above are reachable in ordinary use — the `adkrest` handler relays each event with a blocking `ws.WriteJSON` and sets no write deadline, and any tool that runs longer than 30s blocks the same loop with no network pathology at all.

The fix is small: capture the death time where the error is produced rather than where it is consumed — carry `struct{err error; at time.Time}` on `errChan` with `at` taken in the reader before the blocking send, then score `at.Sub(connectedAt)`.

<details>
<summary>Reproduction — <code>go test -run TestReconnectBudgetSurvivesSlowConsumer ./internal/llminternal/</code></summary>

Uses only helpers already in the package. Shrunk thresholds so it runs in 12s, the mechanism is the same at the shipped values.

```go
func TestReconnectBudgetSurvivesSlowConsumer(t *testing.T) {
	// Every connection: handshake, one frame, then hard close. The real
	// lifetime of each connection is milliseconds.
	client, connCount := startFakeLiveServer(t, func(connNum int, conn *websocket.Conn) {
		_ = conn.WriteMessage(websocket.TextMessage, []byte(serverContentPong))
	})

	p := testReconnectPolicy(time.Millisecond, 4)
	p.resetAfter = 200 * time.Millisecond
	f := newReconnectFlow(client, p)
	ctx, cancel := newLiveInvocationContext(t)
	defer cancel()

	sess, seq, err := f.RunLive(ctx)
	if err != nil {
		t.Fatalf("RunLive failed: %v", err)
	}
	defer func() { _ = sess.Close() }()

	var done atomic.Bool
	go func() {
		for range seq {
			time.Sleep(400 * time.Millisecond) // the only unusual thing here
		}
		done.Store(true)
	}()

	time.Sleep(12 * time.Second)
	// 1 initial dial + 4 retries. Anything past that means the budget re-armed.
	if got := connCount.Load(); got > 5 || !done.Load() {
		t.Fatalf("dialled %d times in 12s, terminated=%v; want <= 5 dials and a reported error", got, done.Load())
	}
}
```

```
dialled 16 times in 12s, terminated=false; want <= 5 dials and a reported error
```

Sampling the dial count every 1.2s gives 2, 4, 5, 7, 8, 10, 11, 13, 14, 16 — climbing linearly rather than stopping at the bound.
</details>

**`workflow.CalculateDelay` does not exist on this branch.** [`live_reconnect.go:42`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/live_reconnect.go#L42) says the delay maths duplicates it "so the two cannot drift", and [`live_reconnect_test.go:655-656`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/live_reconnect_test.go#L655-L656) adds "which cannot be called from here (workflow imports this package)". There is no `workflow` package anywhere on `v1` and no `workflow/` directory, so the invariant those comments ask a future maintainer to preserve cannot be checked here, and the parenthetical about the import cycle is not true on this branch. [`live_reconnect_test.go:765`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/live_reconnect_test.go#L765) carries the same reference. Worth either dropping the cross-reference on this branch or pointing it at something that exists here.

**The loop-top teardown guard is never exercised, and the test named for it exits through a different path.** [`TestRunLiveNoDialAfterTeardownOnResetPath`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/live_reconnect_test.go#L450) calls [`sess.Close()`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/live_reconnect_test.go#L480) before [releasing the server](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/live_reconnect_test.go#L481), so `sess.done` closes while the connection is still up and the consumer wakes on the pre-existing [`case <-sess.done:`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/base_flow.go#L649) arm and returns without re-entering the loop top. Its `connCount == 1` assertion holds either way. Delete [the guard](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/base_flow.go#L384-L389) outright and `go test ./internal/llminternal/` still passes, and so does that test 30 times out of 30. In the same run both [`return true` arms of `tornDown`](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/base_flow.go#L105-L108) never execute — across the whole package suite `tornDown` only ever returns through its `default`. A two-line direct test of `tornDown` with a closed session and with a cancelled context would cover both arms deterministically, which is cheaper than reordering the integration test.

**`defaultReconnectResetAfter` is pinned only to being positive, and it is the constant that guards the bug this PR fixes.** The [only assertion on it](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/live_reconnect_test.go#L792-L793) fires for `resetAfter <= 0`, and its own message names the hazard — "a crash loop restores the budget on every attempt". Change [30s](https://github.com/google/adk-go/blob/549f76026de6b325ed173f54b899b475f37c475b/internal/llminternal/live_reconnect.go#L31) to 100ms and the whole package suite still passes, because the one integration test that runs the default policy kills its first connection in about a millisecond, well under either threshold. Against a backend that accepts and drops every 150ms or so, that value would re-arm the budget on every drop and restore the unbounded loop, green. A magnitude band rather than a sign check would catch it.

I have a few smaller notes on test coverage of the shipped constants and the `SendHistory` retry path — I'll send those separately so this one stays readable.

wolo-lab:
Let's reopen once https://github.com/google/adk-go/pull/1312 is ready.

PR #1320, opened 2026-08-12
fix(loadartifactstool): support load_artifacts when combined with other tool calls

### LINKED ISSUE #1319: loadartifactstool: load_artifacts ignored when preceded by other tool responses in multi-part content
## 🔴 Required Information

**Describe the Bug:**
In `tool/loadartifactstool`, `processLoadArtifactsFunctionCall` only inspects `req.Contents[last].Parts[0]` when processing function responses. When an LLM executes multiple tool calls in a single turn (e.g. `[other_tool, load_artifacts]`), the function response for `load_artifacts` is placed at `Parts[1]` or later. Because only `Parts[0]` is checked, `processLoadArtifactsFunctionCall` returns early without processing `load_artifacts`, causing the requested artifact contents to be silently skipped.

**Steps to Reproduce:**
1. Save an artifact (e.g. `doc1.txt`) to the session artifact store.
2. Construct an `LLMRequest` whose last `Content` contains a multi-part function response where `Parts[0]` is a response to another function (e.g. `other_function`) and `Parts[1]` is a response to `load_artifacts`.
3. Pass the request to `loadartifactstool.ProcessRequest`.
4. Observe that `load_artifacts` is ignored and `doc1.txt` is not appended to `LLMRequest.Contents`.

**Expected Behavior:**
`loadartifactstool` should scan all parts in `req.Contents[last].Parts` to find and process any `load_artifacts` function response, regardless of its position in the `Parts` slice.

**Observed Behavior:**
`processLoadArtifactsFunctionCall` inspects only `lastContent.Parts[0]`. If `Parts[0]` does not match `"load_artifacts"`, it returns `nil` and skips loading the artifact.

**Environment Details:**
 - ADK Library Version: v2.x (main)
 - OS: Linux
 - Go Version: 1.26.5

**Model Information:**
 - Which model is being used: gemini-2.5-flash / gemini-2.5-pro

---

## 🟡 Optional Information

 - Always (100%) when `load_artifacts` is preceded by another tool response in a multi-part turn.

chill-czar:
Closes #1319

## Summary

- Add multi-part `FunctionResponse` traversal in `processLoadArtifactsFunctionCall` (`tool/loadartifactstool/load_artifacts_tool.go`) to find `load_artifacts` responses across all parts in `req.Contents[last].Parts`
- Add unit test coverage `TestLoadArtifactsTool_ProcessRequest_Artifacts_MultiPartFunctionResponse` in `tool/loadartifactstool/load_artifacts_tool_test.go` asserting multi-part tool execution correctly loads artifact content

## Why

When an LLM executes multiple tool calls in a single turn (e.g. `[other_tool, load_artifacts]`), the function response for `load_artifacts` is placed at `Parts[1]` or later in `req.Contents[last].Parts`. Previously, `processLoadArtifactsFunctionCall` checked only `Parts[0]` (`firstPart := lastContent.Parts[0]`), returning early if `Parts[0]` was a response to a different function. As a result, `load_artifacts` was silently ignored and requested artifact contents were never loaded into context. Iterating over all parts in `lastContent.Parts` locates `load_artifacts` wherever it appears in the response array.

## Test Plan

- [x] Unit tests pass: `go test -race -v ./tool/loadartifactstool/...`
- [x] Linter & Mod Tidy check pass: `go mod tidy -diff`
- [x] Empirical verification: Confirmed `TestLoadArtifactsTool_ProcessRequest_Artifacts_MultiPartFunctionResponse` fails on `main` without the fix (`Expected 2 contents, but got: 1`) and passes cleanly with this fix
- [x] No regressions: All existing `loadartifactstool` test cases pass cleanly

/cc @hyangah @kdroste-google @dpasiukevich @karolpiotrowicz @wolo-lab

### DIFF
--- a/tool/loadartifactstool/load_artifacts_tool.go
+++ b/tool/loadartifactstool/load_artifacts_tool.go
@@ -162,14 +162,18 @@ func (t *artifactsTool) processLoadArtifactsFunctionCall(ctx agent.Context, req 
 	if lastContent == nil || len(lastContent.Parts) == 0 {
 		return nil
 	}
-	firstPart := lastContent.Parts[0]
-	if firstPart.FunctionResponse == nil {
-		return nil
+	var functionResponse *genai.FunctionResponse
+	// Iterate over all parts in the last content turn to find load_artifacts responses.
+	// Note: adk-python only checks parts[0]; scanning all parts is intentional in Go to
+	// support parallel/multi-tool turns where load_artifacts may not be the first part.
+	for _, part := range lastContent.Parts {
+		if part != nil && part.FunctionResponse != nil && part.FunctionResponse.Name == "load_artifacts" {
+			functionResponse = part.FunctionResponse
+			// Keep only the first load_artifacts response if multiple exist in one turn.
+			break
+		}
 	}
-
-	functionResponse := firstPart.FunctionResponse
-
-	if functionResponse.Name != "load_artifacts" {
+	if functionResponse == nil {
 		return nil
 	}
 	artifactNamesRaw, ok := functionResponse.Response["artifact_names"]
--- a/tool/loadartifactstool/load_artifacts_tool.go
+++ b/tool/loadartifactstool/load_artifacts_tool.go
@@ -159,14 +159,14 @@ func (t *artifactsTool) processLoadArtifactsFunctionCall(ctx agent.Context, req 
 	if lastContent == nil || len(lastContent.Parts) == 0 {
 		return nil
 	}
-	firstPart := lastContent.Parts[0]
-	if firstPart.FunctionResponse == nil {
-		return nil
+	var functionResponse *genai.FunctionResponse
+	for _, part := range lastContent.Parts {
+		if part != nil && part.FunctionResponse != nil && part.FunctionResponse.Name == "load_artifacts" {
+			functionResponse = part.FunctionResponse
+			break
+		}
 	}
-
-	functionResponse := firstPart.FunctionResponse
-
-	if functionResponse.Name != "load_artifacts" {
+	if functionResponse == nil {
 		return nil
 	}
 	artifactNamesRaw, ok := functionResponse.Response["artifact_names"]
--- a/tool/loadartifactstool/load_artifacts_tool_test.go
+++ b/tool/loadartifactstool/load_artifacts_tool_test.go
@@ -408,6 +408,95 @@ func TestLoadArtifactsTool_ProcessRequest_NilArtifacts_WithFunctionCall(t *testi
 	}
 }
 
+func TestLoadArtifactsTool_ProcessRequest_Artifacts_MultiPartFunctionResponse(t *testing.T) {
+	otherFR := &genai.FunctionResponse{
+		Name: "other_function",
+		Response: map[string]any{
+			"status": "ok",
+		},
+	}
+	loadArtifactsFR := &genai.FunctionResponse{
+		Name: "load_artifacts",
+		Response: map[string]any{
+			"artifact_names": []string{"doc1.txt"},
+		},
+	}
+
+	tests := []struct {
+		name  string
+		parts []*genai.Part
+	}{
+		{
+			name: "load_artifacts after other function",
+			parts: []*genai.Part{
+				genai.NewPartFromFunctionResponse(otherFR.Name, otherFR.Response),
+				genai.NewPartFromFunctionResponse(loadArtifactsFR.Name, loadArtifactsFR.Response),
+			},
+		},
+		{
+			name: "load_artifacts before other function",
+			parts: []*genai.Part{
+				genai.NewPartFromFunctionResponse(loadArtifactsFR.Name, loadArtifactsFR.Response),
+				genai.NewPartFromFunctionResponse(otherFR.Name, otherFR.Response),
+			},
+		},
+		{
+			name: "nil part alongside load_artifacts",
+			parts: []*genai.Part{
+				nil,
+				genai.NewPartFromFunctionResponse(loadArtifactsFR.Name, loadArtifactsFR.Response),
+			},
+		},
+	}
+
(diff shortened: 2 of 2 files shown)


chill-czar:
@cla/google check

chill-czar:
Hi @hyangah @kdroste-google @dpasiukevich @karolpiotrowicz - could you please review this PR when you get a chance? Thank you!

indurireddy-TF:
Thanks @chill-czar for putting together this fix, I verified locally that the fix is working as expected, and the test results look good.
Could you please take a look when you get a chance? @baptmont @kdroste-google 
image

wolo-lab (dismissed):
The fix is correct and the reasoning behind it holds up.

Worth stating explicitly, because it is the strongest thing about this change: the new loop's first iteration carries the same predicate the old `Parts[0]` read did, so any function response the old code selected is selected identically, and everything downstream of the selection is untouched. There is no input on which this delivers fewer artifacts than `main` does today.

Two things before it can land.

## Needs a rebase

Please rebase and resolve conflicts.

## One behavior change that is worth a follow-up

Widening the scan means a `load_artifacts` response at a non-zero index is now acted on, and so are its failures. For a turn shaped `[other_tool, load_artifacts]` where an artifact name does not resolve, the old code returned early and the turn completed. Now the load fails, the error leaves `ProcessRequest`, and `toolPreprocess` propagates it until the whole agent turn ends.

<details><summary>Reproduction — drop into <code>load_artifacts_tool_test.go</code> and run <code>go test -run Repro ./tool/loadartifactstool/</code></summary>

```go
func TestRepro_MissingArtifactAtNonZeroIndex(t *testing.T) {
	tl := loadartifactstool.New()
	tc := createToolContext(t)
	if _, err := tc.Artifacts().Save(t.Context(), "real.txt", &genai.Part{Text: "x"}); err != nil {
		t.Fatalf("save: %v", err)
	}
		Role: "user",
		Parts: []*genai.Part{
			genai.NewPartFromFunctionResponse("other_tool", map[string]any{"status": "ok"}),
			genai.NewPartFromFunctionResponse("load_artifacts", map[string]any{"artifact_names": []string{"ghost.txt"}}),
		},
	}}}
	// Before this change: nil, and the turn completes.
	// After: failed to load artifact ghost.txt: failed to load artifact ghost.txt: artifact not found: file does not exist
	t.Log(tl.(toolinternal.RequestProcessor).ProcessRequest(tc, req))
}
```
</details>

I do not think this should block the change. The same hard failure already happens on `main` whenever `load_artifacts` is the first part, the old behavior on this input was itself wrong (a silent no-op while the function response told the model the load had succeeded), and any correct fix to the positional bug inherits the failure mode. It is also not limited to unresolvable names — the two type errors below [`group.Wait()`](https://github.com/google/adk-go/blob/daa9ae5d61f170b20398a4785930ce809eabfc40/tool/loadartifactstool/load_artifacts_tool.go#L212) end the turn the same way. adk-python skips a missing artifact with a warning and carries on, and additionally retries the name with a `user:` prefix, so closing this would also close a parity gap. A separate issue would be the right home for it.

## Smaller points, none blocking

- **The [`break`](https://github.com/google/adk-go/blob/daa9ae5d61f170b20398a4785930ce809eabfc40/tool/loadartifactstool/load_artifacts_tool.go#L166) keeps only the first `load_artifacts` response.** If the model issues two `load_artifacts` calls in one turn, `mergeParallelFunctionResponseEvents` concatenates both responses into one content with no deduplication by name, and the second is dropped while its function response reports success. `main` behaves the same way, so this is a retained limitation rather than something you introduced. Collecting every match instead of breaking is a few lines, and if first-match is deliberate a one-line comment saying so would stop someone "cleaning up" the `break` later.
- **The new test pins the last index rather than position independence.** Its content is the only multi-part one in the suite, so an implementation that simply took `Parts[len-1]` would pass all eight tests, as would one that dropped the `break`. Adding the mirror case, `[load_artifacts, other_function]` expecting two contents, would close that.
- **The `part != nil` check quietly fixes a real nil dereference** that the old `Parts[0]` read would have hit, and nothing exercises it. A case with a nil element in `Parts` would be worth having.
- **`Role: "model"` on the fixture** at [line 371](https://github.com/google/adk-go/blob/daa9ae5d61f170b20398a4785930ce809eabfc40/tool/loadartifactstool/load_artifacts_tool_test.go#L371) is not a shape the runtime produces — function response contents are built with `"user"`, and a non-user last content would have had a synthetic user turn appended ahead of it before the tool ever ran. Nothing here depends on `Role`, and the neighbouring tests do the same thing, so this is only worth changing if you are touching them anyway.
- **A note on the divergence from adk-python would help.** Python still indexes `parts[0]`, so after this lands the two implementations disagree on the same session history. Go's behavior is the right one, but without a marker a future parity sweep could revert it.

chill-czar:
Thanks @wolo-lab for the thorough review and approval, and thanks @indurireddy-TF for testing and verifying!

I have rebased the branch onto latest `upstream/main` (resolving the conflict with #1390) and addressed all review points:
- **Inline Documentation**: Added notes documenting the intentional divergence from `adk-python` (scanning all parts rather than indexing `parts[0]`) and the rationale for the `break` (retaining first-match behavior).
- **Position Independence Tests**: Expanded tests into a table-driven suite verifying both positions (`[other_tool, load_artifacts]` and `[load_artifacts, other_tool]`).
- **Nil Element Test**: Added test coverage for a `nil` part in `Parts` (`[nil, load_artifacts]`).
- **Fixture Role**: Updated test fixtures to use `Role: genai.RoleUser`.

Regarding the follow-up on missing artifact error handling and `adk-python` parity (warning/skipping missing artifacts and `user:` prefix retry), I agree that is best tracked in a dedicated follow-up issue once this lands.

All workspace tests, linters, and tidy checks pass cleanly.

wolo-lab (approved):
Thank you for addressing optional comments! Approved.

PR #1322, opened 2026-08-12
fix(session/vertexai): treat nil function args as an empty struct

### LINKED ISSUE #1321: session/vertexai: events with a zero-argument function call are never persisted — nil Args is rejected by toStructPB
## 🔴 Required Information

**Describe the Bug:**

`session/vertexai` fails to persist any event containing a function call whose `Args` are nil, or a function response whose `Response` is nil. This is the normal shape for a tool that declares no parameters (`get_time`, `list_items`, …) or a tool that returns no payload.

`createAiplatformpbContent` converts both fields with `toStructPB`:

```go
// session/vertexai/vertexai_client.go:639
args, err := toStructPB(part.FunctionCall.Args)
if err != nil {
    return nil, fmt.Errorf("failed to convert function call to structpb: %w", err)
}
```

`toStructPB` marshals its input to JSON and hands the result to `structpb.Struct.UnmarshalJSON`. A nil map marshals to the JSON literal `null`, which `structpb` rejects:

```go
// session/vertexai/vertexai_client.go:922
func toStructPB(value any) (*structpb.Struct, error) {
	data, err := json.Marshal(value)   // map[string]any(nil) -> "null"
	...
	if err := res.UnmarshalJSON(data); err != nil {   // rejects "null"
		return nil, fmt.Errorf("failed to unmarshal JSON data to structpb: %w", err)
	}
```

The error propagates out of `AppendEvent` (`vertexai_client.go:273-276`), so the event is never written to the session.

The other three `toStructPB` call sites (lines 74, 458, 686) are guarded by a `len(...) > 0` or `!= nil` check. The two function-call sites at lines 639 and 652 are not.

**Steps to Reproduce:**

1. Register a tool that takes no parameters, e.g. a `get_time` function with an empty parameter schema.
2. Configure the runner with a Vertex AI session service (`session/vertexai`).
3. Send a message that causes the agent to call that tool.
4. Observe the failure when the function call event is persisted.

This does not depend on the tool implementation — it only requires the model to emit a function call with no `args` field, which is the standard wire form for a zero-parameter tool. `genai` leaves `Args` nil in that case:

```go
var p genai.Part
json.Unmarshal([]byte(`{"functionCall":{"id":"c1","name":"get_time"}}`), &p)
// p.FunctionCall.Args == nil
```

**Expected Behavior:**

A function call with no arguments persists successfully, storing an empty `args` struct. Same for a function response with no payload.

**Observed Behavior:**

```
failed to add event to session: failed to append event: error creating content: failed to convert function call to structpb: failed to unmarshal JSON data to structpb: proto: syntax error (line 1:1): unexpected token null
```

**Environment Details:**

- ADK Library Version: v2.2.0 (also reproduces on `main` at `cead71c`)
- OS: macOS (Darwin 25.3.0)
- Go Version: 1.26.5

**Model Information:**

- N/A — the bug is in event serialization, not model-specific.

---

## 🟡 Optional Information

**Regression:**

Yes. The last release without this behaviour is **v1.4.0**. It is present in **v1.5.0, v1.5.1, v1.6.0, v2.0.0, v2.1.0 and v2.2.0**.

These conversions previously went through `structpb.NewStruct`, which returns an empty `Struct` for a nil map. The switch to the JSON-based helper was needed to support values `structpb.NewStruct` cannot handle (typed slices such as `[]string`, see #668), but it also changed the nil case from "empty struct" to "error":

```go
structpb.NewStruct(map[string]any(nil))  // -> &structpb.Struct{}, nil
toStructPB(map[string]any(nil))          // -> nil, "unexpected token null"
```

```go
package main

import (
	"encoding/json"
	"fmt"

	"google.golang.org/protobuf/types/known/structpb"
)

func main() {
	// What toStructPB does with FunctionCall.Args for a zero-parameter tool:
	data, _ := json.Marshal(map[string]any(nil))
	fmt.Printf("marshaled: %s\n", data) // null

	res := &structpb.Struct{}
	fmt.Println(res.UnmarshalJSON(data))
	// proto: syntax error (line 1:1): unexpected token null

	// The previous helper accepted it:
	s, err := structpb.NewStruct(map[string]any(nil))
	fmt.Printf("NewStruct: %v, err=%v\n", s, err) // , err=<nil>
}
```

**Suggested Fix:**

Treat a `null` marshaling as an empty `Struct` inside `toStructPB`, which restores the previous behaviour for all five call sites rather than guarding the two affected ones:

```go
if string(data) == "null" {
	return &structpb.Struct{}, nil
}
```

Happy to send a PR with this fix and regression tests if that direction looks right.

ravyg:

- Fixes: #1321

### Problem

`session/vertexai` fails to persist any event containing a function call with nil `Args`, or a function response with nil `Response`. That is the normal shape for a tool declaring no parameters (`get_time`, `list_items`, …) or a tool returning no payload — `genai` leaves the field nil when the model emits a function call with no `args`.

`createAiplatformpbContent` converts both fields with `toStructPB`, which marshals to JSON and hands the result to `structpb.Struct.UnmarshalJSON`. A nil map marshals to the literal `null`, which `structpb` rejects. The error propagates out of `AppendEvent`, so the event is never written:

```
failed to add event to session: failed to append event: error creating content:
failed to convert function call to structpb: failed to unmarshal JSON data to structpb:
proto: syntax error (line 1:1): unexpected token null
```

These conversions previously used `structpb.NewStruct`, which returns an empty `Struct` for a nil map. The move to the JSON-based helper was needed to support values `structpb.NewStruct` cannot handle (typed slices such as `[]string`, #668), but it also turned the nil case from "empty struct" into an error:

```go
structpb.NewStruct(map[string]any(nil))  // -> &structpb.Struct{}, nil
toStructPB(map[string]any(nil))          // -> nil, "unexpected token null"
```

Last unaffected release is v1.4.0; present in v1.5.0, v1.5.1, v1.6.0, v2.0.0, v2.1.0 and v2.2.0.

### Solution

Treat a `null` marshaling as an empty `Struct` inside `toStructPB`.

Fixing the helper rather than guarding the two call sites keeps all five callers consistent and restores the exact prior `structpb.NewStruct` semantics. The three other call sites (lines 74, 458, 686) already guard with `len(...) > 0` or `!= nil`, so this is a no-op for them; only the two function-call sites were unguarded.

### Testing Plan

- [x] `TestToStructPB`: added `nil map converts to an empty struct` and `untyped nil converts to an empty struct`.
- [x] `TestCreateAiplatformpbContent`: added `function call with nil args` and `function response with nil response`, covering the failure at the level users hit it.
- [x] Confirmed all four new cases fail without the fix and pass with it.
- [x] Existing tests in the package unchanged and green.

```sh
go build -mod=readonly ./...
go test -race -count=1 -shuffle=on ./session/...
gofmt -l session/
go mod tidy -diff
```

### Alignment with adk-python

A tool with no parameters is a first-class case across ADK implementations; persisting its call event should not depend on whether the model emitted an empty `args` object or omitted the field.

### DIFF
--- a/session/vertexai/vertexai_client.go
+++ b/session/vertexai/vertexai_client.go
@@ -919,10 +919,18 @@ func createGroundingMetadata(metadata *aiplatformpb.GroundingMetadata) *genai.Gr
 // It uses JSON marshaling as an intermediary step to safely serialize
 // the input data before constructing the *structpb.Struct.
 // Returns an error if any part of the JSON round-trip or conversion fails.
+//
+// A nil value marshals to the JSON literal "null", which structpb rejects.
+// Such a value is treated as an empty Struct, matching structpb.NewStruct(nil)
+// and keeping callers that pass an absent map (for example a function call
+// that takes no arguments) from failing.
 func toStructPB(value any) (*structpb.Struct, error) {
 	data, err := json.Marshal(value)
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal value: %w", err)
+	}
+	if string(data) == "null" {
+		return &structpb.Struct{}, nil
 	}
 	res := &structpb.Struct{}
 	if err := res.UnmarshalJSON(data); err != nil {
--- a/session/vertexai/vertexai_test.go
+++ b/session/vertexai/vertexai_test.go
@@ -477,6 +477,29 @@ func TestToStructPB(t *testing.T) {
 			expectError: true,
 		},
 		{
+			// A tool that takes no arguments leaves FunctionCall.Args nil, which
+			// marshals to the JSON literal "null". That must convert to an empty
+			// Struct rather than an error, as structpb.NewStruct(nil) did.
+			name:        "nil map converts to an empty struct",
+			input:       map[string]any(nil),
+			expectError: false,
+			validate: func(t *testing.T, s *structpb.Struct) {
+				if got := len(s.GetFields()); got != 0 {
+					t.Errorf("len(fields) = %d, want 0", got)
+				}
+			},
+		},
+		{
+			name:        "untyped nil converts to an empty struct",
+			input:       nil,
+			expectError: false,
+			validate: func(t *testing.T, s *structpb.Struct) {
+				if got := len(s.GetFields()); got != 0 {
+					t.Errorf("len(fields) = %d, want 0", got)
+				}
+			},
+		},
+		{
 			name: "custom struct representing possible state delta",
 			input: struct {
 				StringValue string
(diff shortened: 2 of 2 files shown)


kdroste-google (approved):
Hi, @ravyg , thank you for spotting this issue and preparing the fix!

PR #1324, opened 2026-08-13
fix(remoteagent): cancel A2A task after early context cancellation

### LINKED ISSUE #1076: Remote A2A task is orphaned when parent task is cancelled before the first remote event

ktsoator:

Closes #1076.

When an invocation is canceled before the first A2A task-bearing event arrives, the remote task may already be running. Closing the stream immediately prevents the client from learning its task ID and can leave it running.

This change keeps the stream open briefly after cancellation to discover a cleanup target, without emitting further session events or invoking response callbacks during that drain.

- The task-information wait is limited to five seconds and bounded by the caller's deadline. If that deadline expires before task information arrives, no further drain is possible.
- A `Task` or `TaskStatusUpdateEvent` with an ID supplies the cleanup state. An artifact task ID is retained as a fallback when no state-bearing target is known and stops post-cancellation draining immediately. Previously observed terminal state prevents cancellation.
- Any final `Message`, including one with a `TaskID`, stops the drain and clears any earlier cleanup target. Its task association alone does not establish a cancellable remote task.
- The default `CancelTask` RPC reuses the shared cleanup budget when at least 100 ms remain. Otherwise, it receives a detached five-second timeout. This fallback can extend total cleanup time beyond the shared budget and the caller's deadline; 100 ms is a best-effort threshold, not a guarantee of network completion.
- Non-streaming responses retain their cleanup target. Working tasks are canceled, while completed tasks and input-required tasks without an error are preserved.

No public API or dependency changes are introduced.

### Testing Plan

Automated coverage includes cancellation before the first task event, caller-deadline propagation, callback suppression during draining, shared-budget reuse, artifact fallback, terminal-state precedence, final Messages, and non-streaming cleanup.

Direct context tests cover an elapsed deadline while `Err()` is still nil. Run-level tests pin the cleanup policy at 1, 99, 100 and 101 ms remaining. The direct test fails with the old Err()-only predicate, and the non-streaming test fails if cleanup-target recording is removed.

The following checks passed after syncing with the PR's latest main merge:

```text
go build -mod=readonly work
golangci-lint run
go mod tidy -diff
```

Lint and tidy checks passed in both the root module and `plugin/agentanalytics`.

Manual verification used real local A2A JSON-RPC servers and clients:

- The original two-server E2E test confirmed cancellation of Server B's child task after canceling Server A before the child's first task-bearing event. The [execution log](https://github.com/google/adk-go/pull/1324#issuecomment-5325171307) is attached in the discussion.
- A near-deadline probe with `-race` delivered 100/100 cancellation requests across five timing/concurrency cases. An identical probe on `4ae95ba` delivered 0/20 requests in each of the two 1 ms cases. These measurements were taken before the latest main merge; the [results table and timing adjustment](https://github.com/google/adk-go/pull/1324#issuecomment-5569707196) are documented in the discussion.

### Checklist

- [x] I have read `CONTRIBUTING.md`.
- [x] I have performed a self-review.
- [x] I have documented the cleanup timing and its limitations.
- [x] I have added regression and real-transport E2E coverage.
- [x] Build, race tests, lint, and tidy checks pass locally.
- [x] No dependent changes are required.


ktsoator:
Hi maintainers, thank you for your time. This PR addresses #1076 with a bounded, task-ID-driven cleanup path rather than a fixed delay, and includes deterministic regression coverage for cancellation before the first remote event. All required checks are passing. When you have a chance, I would appreciate a review.

ktsoator:
Manual live A2A E2E verification completed against commit `10670d1` on macOS arm64 with Go 1.26.6.

The standalone verification harness started Server A and Server B as live local A2A JSON-RPC services on separate TCP listeners, then used an independent A2A client to exercise this sequence:

1. Server B created the child task and paused before emitting its first task-bearing event.
2. The client observed Server A's parent task ID and canceled the parent while the child event was still paused.
3. After Server A emitted `TASK_STATE_CANCELED`, Server B released its first event containing the child task ID.
4. Server B received `CancelTask` for that child task.
5. A final `GetTask` call to Server B returned `TASK_STATE_CANCELED`.

Relevant log:

```text
[000004ms] [server-b] child task created; first event is paused
[000004ms] [client] precondition satisfied: parent and child task IDs observed, but server-b has emitted no event
[000005ms] [client] server-a parent state=TASK_STATE_CANCELED
[000005ms] [client] releasing server-b first event
[000005ms] [server-b] first task-bearing event state=TASK_STATE_SUBMITTED
[000005ms] [server-b] CancelTask received for the child task
[000007ms] [client] queried server-b child state=TASK_STATE_CANCELED
[000007ms] [result] PASS
```

karolpiotrowicz (changes requested):
The orphaned-task bug is real and this closes it for the common case. I extracted your regression test and ran it against `main` as it stood during the review ([49dfa68](https://github.com/google/adk-go/commit/49dfa683fc04b90b1a0adc6b197f3a916a2ceb10)) rather than against the branch base: it fails there with `timeout waiting for remote task cancellation` and passes with the change applied. The branch is 55 commits behind and `a2a-go` moved to `/v2` in the interim, so I ran everything on the merge result — it automerges cleanly and the suite passes on it.

Two things I'd like resolved before this goes in, and a test-adequacy point I'd rather fix in the same round than leave behind. One of the two is a single line.

## The 5-second wait is unconditional but the benefit is not

The wait costs the caller time in cases where it can never pay off, and it is not configurable. Measuring `cancel()` to `Run()` returning, on the merge result against `main`:

- Remote yields nothing yet: `0 ms` → `5002 ms`.
- Remote streams `a2a.Message` events with no `TaskID`: `0 ms` → `5000 ms`, and `CancelTask` is still never sent — a message can't be cleaned up, so the full wait buys nothing.
- Task ID already seen before the cancel: `0 ms` on both. The stall is confined to the pre-first-event window.

The case that concerns me most is a caller-set deadline. With an invocation carrying a 300 ms timeout, the remote stream stayed open for `5304 ms`, and the client was handed a context reporting no deadline at all — `context.WithoutCancel` at [a2a_agent.go:212](https://github.com/google/adk-go/blob/2f65b2a2be352fc583ccfc773b4170428eaf2919/agent/remoteagent/v2/a2a_agent.go#L212) strips the deadline along with the cancellation, so a transport that sizes its own read timeout from `ctx.Deadline()` no longer gets one. On `main` the same invocation ends the stream at exactly 300 ms.

The two budgets also compose rather than sharing a deadline. Releasing the task-bearing event 4 seconds in, with a `CancelTask` that then hangs, `Run()` took `9002 ms` to return — the task-ID wait plus the pre-existing 5-second cancel budget, both on the caller's critical path.

I don't think the tradeoff is wrong, but it's currently silent: [`remoteTaskIDWaitTimeout`](https://github.com/google/adk-go/blob/2f65b2a2be352fc583ccfc773b4170428eaf2919/agent/remoteagent/v2/a2a_agent.go#L199) is an unexported constant, and the bound is stated in the PR description and nowhere in the code or the package docs.

## The task-ID predicate accepts events cleanup will refuse

[a2a_agent.go:350](https://github.com/google/adk-go/blob/2f65b2a2be352fc583ccfc773b4170428eaf2919/agent/remoteagent/v2/a2a_agent.go#L350) marks the task ID received for any event with a non-empty `TaskInfo().TaskID`. `cleanupRemoteTask` accepts a narrower set: it refuses `*a2a.Message` outright at [a2a_agent.go:376](https://github.com/google/adk-go/blob/2f65b2a2be352fc583ccfc773b4170428eaf2919/agent/remoteagent/v2/a2a_agent.go#L376), and it derives the task state only for `*a2a.Task` and `*a2a.TaskStatusUpdateEvent` at [a2a_agent.go:379-387](https://github.com/google/adk-go/blob/2f65b2a2be352fc583ccfc773b4170428eaf2919/agent/remoteagent/v2/a2a_agent.go#L379-L387). Any other type leaves `state` at the zero value, for which `Terminal()` is false. So ending the wait and ending it *usefully* are different conditions, and only the first is checked. Two shapes fall through:

A post-cancellation stream of `Message{TaskID}` then `Task{working}` ends the wait on the message, cancels the stream, and cleanup then discards it — no `CancelTask`, and the `Task` event that cleanup would have acted on is never delivered at all. This one is no worse than `main`, which orphans the same shape, and I could not find any adk-go server that emits a bare task-scoped message in a response stream, so it's incompleteness rather than a regression.

The artifact-event shape is the one that is reachable here. `TaskArtifactUpdateEvent` carries a `TaskID` and no state, and adk-go's own server streams them — [a2a_e2e_test.go:487](https://github.com/google/adk-go/blob/160a39763e901047505abcc1acc32707edc0fab5/agent/remoteagent/v2/a2a_e2e_test.go#L487) does exactly that in a loop, pinned at the current main ([160a3976](https://github.com/google/adk-go/commit/160a39763e901047505abcc1acc32707edc0fab5)). Given `Artifact{TaskID}` followed by `StatusUpdate{completed}` after cancellation, the loop returns on the artifact event, never sees the terminal update, and cleanup sends `CancelTask` against a task that had already finished. `main` sends nothing. A conformant server rejects it — `a2asrv` returns `ErrTaskNotCancelable` for a terminal task — so the cost is a wasted round trip and a warning log rather than anything worse. Both shapes go away together if line 350 tests the same condition cleanup will accept.

## The tests pin the fix but not its guardrails

I checked this by mutation rather than by reading. Reverting `newRemoteTaskStreamContext` to a plain `context.WithCancel(parent)` — a full revert of the detach — is caught by `TestRemoteAgent_CancelsRemoteTaskWhenContextCanceledBeforeFirstEvent`, so the core of the fix is genuinely protected. But deleting the whole post-cancellation block at [a2a_agent.go:355-360](https://github.com/google/adk-go/blob/2f65b2a2be352fc583ccfc773b4170428eaf2919/agent/remoteagent/v2/a2a_agent.go#L355-L360) leaves that test, `TestRemoteTaskStreamContext_BoundsTaskIDWait` and `TestRemoteAgent_CleanupCallback` all green.

The reason is worth knowing, because it makes two assertions look stronger than they are: the fixture task at [a2a_agent_test.go:1417-1421](https://github.com/google/adk-go/blob/2f65b2a2be352fc583ccfc773b4170428eaf2919/agent/remoteagent/v2/a2a_agent_test.go#L1417-L1421) has no parts and a non-terminal status, so it converts to a nil session event, and `runAfterA2ARequestCallbacks` returns early when both the event and the error are nil. The `t.Error` guarding the callback at [line 1451](https://github.com/google/adk-go/blob/2f65b2a2be352fc583ccfc773b4170428eaf2919/agent/remoteagent/v2/a2a_agent_test.go#L1451) and the `emitted` check at [line 1494](https://github.com/google/adk-go/blob/2f65b2a2be352fc583ccfc773b4170428eaf2919/agent/remoteagent/v2/a2a_agent_test.go#L1494) can therefore never fire against any implementation. A fixture that converts to a non-nil event would make both real.

`TestRemoteTaskStreamContext_BoundsTaskIDWait` also passes under that same full revert, since it only asserts `Done()` fires *within* one second — an implementation that never waits satisfies it. A lower-bound assertion would pin the behavior the name promises.

Coverage agrees with the mutants: the `continue` at [a2a_agent.go:359](https://github.com/google/adk-go/blob/2f65b2a2be352fc583ccfc773b4170428eaf2919/agent/remoteagent/v2/a2a_agent.go#L359) and the `return` in the new pre-send guard at [a2a_agent.go:341](https://github.com/google/adk-go/blob/2f65b2a2be352fc583ccfc773b4170428eaf2919/agent/remoteagent/v2/a2a_agent.go#L341) are executed by no test in the package.

## Smaller things

The comment at [a2a_e2e_test.go:551](https://github.com/google/adk-go/blob/160a39763e901047505abcc1acc32707edc0fab5/agent/remoteagent/v2/a2a_e2e_test.go#L551) reads "Cancel only after the subagent's output reaches the client: before that the parent doesn't know the subagent task ID, so cancellation can't propagate." That is the limitation this PR removes, so it now describes the old behavior. That test is also the only real-transport cancellation coverage in the package, and extending it to cancel before the first event would exercise the new path end to end.

`markTaskIDReceived` closes the channel with no guard of its own. The only thing preventing a double close is the loop-local `taskIDReceived` bool at the call site. Nothing reaches it today, since it would need a client that calls `yield` from several goroutines, which breaks the iterator contract anyway.

Non-blocking, and worth saying since it's easy to miss in a diff this size: the new `ctx.Err()` check before the stream is created is a genuine improvement over `main`, which sent unconditionally on an already-cancelled context. The concurrency also holds up: `-race` is clean, and I confirmed the detector actually reaches the new goroutine rather than passing vacuously by injecting a deliberate unsynchronized write between the `AfterFunc` goroutine and `markTaskIDReceived`, which it caught. Goroutine counts are flat across 30 cancelled invocations, and `apidiff`, `govulncheck`, `gofmt`, `go vet` and `golangci-lint` are all clean.

ktsoator:
Thank you for the thorough review, especially the mutation checks and timing
measurements. I updated the branch to the current main and addressed the
points you raised:

- The detached stream context now preserves the caller's deadline. Waiting
  for cleanup-eligible task information and the default CancelTask request
  share one five-second budget. A direct message without a task ID stops the
  wait immediately. I kept the timeout internal, but documented the fixed
  budget and the caller-deadline upper bound on RemoteTaskCleanupCallback.
- Cleanup eligibility now matches the cleanup path: only non-nil Task and
  TaskStatusUpdateEvent values with a task ID are retained. Task-scoped
  messages and artifact updates neither end the wait nor overwrite the last
  eligible cleanup target.
- The delayed fixture now produces a real session event. The tests cover the
  wait lower bound, deadline propagation, the shared budget, the pre-send
  guard, the post-cancellation continue path, task-scoped messages followed
  by tasks, and artifacts followed by terminal status updates.
- TestA2ACleanupPropagation now covers both cancellation before Server B's
  first task-bearing event and cancellation after a remote artifact. The
  early-cancellation case confirms that the remote-agent invocation context
  is canceled before releasing the first event, then verifies that Server B
  receives CancelTask for the correct child task and stores it as canceled.

The full workspace build and race-enabled test suite pass. Lint reports zero
issues in both modules, and both go mod tidy -diff checks are clean. The
focused E2E test also passed 50 consecutive race-enabled runs.

Thank you again for the detailed feedback. When you have time, could you
please take another look?

agent/remoteagent/v2/a2a_agent.go
```diff
@@ -483,55 +588,108 @@ func (a *a2aAgent) run(ctx agent.InvocationContext, cfg A2AConfig) iter.Seq2[*se
...
+		}
+		target.taskInfo = event.TaskInfo()
+		target.state = event.Status.State
+	case *a2a.TaskStatusUpdateEvent:
+		if event == nil {
+			return target, false
+		}
+		target.taskInfo = event.TaskInfo()
+		target.state = event.Status.State
+	default:
```
karolpiotrowicz, line 670: Narrowing eligibility to `*a2a.Task` and `*a2a.TaskStatusUpdateEvent` is the right call for the completed-task-then-artifact case, and it does fix it. It also drops the one shape where an artifact update was the only carrier of the task ID. A stream that yields nothing but `TaskArtifactUpdateEvent` and is then cancelled loses the ID entirely: `lastCleanupTarget` stays nil, the drain runs its full five seconds, and no `CancelTask` goes out. The merge-base read the ID off the artifact event and cancelled in 0 ms, so this is worse on both latency and correctness for that stream shape.

I am not treating it as blocking, because adk-go's own server cannot produce it. `executor.go` yields either a submitted task at [:210-215](https://github.com/google/adk-go/blob/aa8d077573c21eb705e197a049d2bfea11d82480/server/adka2a/v2/executor.go#L210-L215) or a `Working` status at [:227-231](https://github.com/google/adk-go/blob/aa8d077573c21eb705e197a049d2bfea11d82480/server/adka2a/v2/executor.go#L227-L231) before anything reaches `task_artifact.go`, so a third-party server is needed to hit it. Keeping the artifact's `TaskInfo` as a fallback target, used only while `lastCleanupTarget` is still nil, would cover it without reopening the case you just closed.

agent/remoteagent/v2/a2a_agent_test.go
```diff
@@ -1480,6 +1481,618 @@ func TestRemoteAgent_CleanupCallback(t *testing.T) {
...
+
+		<-streamStarted
+		start := time.Now()
+		cancel()
+		synctest.Wait()
+		time.Sleep(4 * time.Second)
+		close(releaseTask)
+		<-runDone
+
+		wantDeadline := start.Add(remoteTaskCleanupTimeout)
```
karolpiotrowicz: `wantDeadline` here and the elapsed check on line 1860 both derive their expected value from `remoteTaskCleanupTimeout`, so neither one pins it. Changing the constant to `30 * time.Second` leaves this test green and the rest of the package green with it. `TestRemoteTaskStreamContext_BoundsCleanupWait` declares its own local `timeout` and passes it in as a parameter, so it constrains the mechanism rather than the number.

That mattered less on `main`, where the five seconds bounded only an RPC nobody was waiting on. Here the same constant also bounds a stall the caller feels, so one assertion against a literal `5 * time.Second` is worth having.

karolpiotrowicz (changes requested):
This rework closes almost everything from the last round. The deadline handling it adds to satisfy one of those points has gone a step too far, though: an invocation that ends because its own deadline expired now orphans a remote task it already knows about, where `main` cancelled it. That is the one thing blocking. Two smaller points are on the lines they concern.

## Deadline expiry no longer cancels a known remote task

`newRemoteTaskCleanupBudget` derives the budget's base from the caller's deadline at [a2a_agent.go:389-395](https://github.com/google/adk-go/blob/aa8d077573c21eb705e197a049d2bfea11d82480/agent/remoteagent/v2/a2a_agent.go#L389-L395), so `base` dies at the same instant the invocation does. `start()` layers `WithTimeout(base, 5s)` on top of it, which cannot outlive `base`, and the deferred cleanup then returns at [a2a_agent.go:687-689](https://github.com/google/adk-go/blob/aa8d077573c21eb705e197a049d2bfea11d82480/agent/remoteagent/v2/a2a_agent.go#L687-L689) without sending `CancelTask`.

So a caller who wraps `runner.Run` in a `context.WithTimeout` — the ordinary way to bound an agent run — gets exactly the orphan this PR exists to close, and gets it silently, because that early return logs nothing while the failure path two lines below it does. Measured with the same test source compiled into both trees and driving `Run`: a 300 ms invocation deadline, `Task{working}` delivered at t≈0 so the task ID is known, then the deadline fires. The merge-base ([ffddb2a6](https://github.com/google/adk-go/commit/ffddb2a6d7ed97974e82815d00c95eb148038f2c)) sends `CancelTask` at 301 ms. This branch sends nothing. No test reaches those three lines — `TestRemoteAgent_DefaultCleanupHonorsInvocationDeadline` does not, because its fake returns while the invocation is still alive, so cleanup runs before the deadline rather than after it.

The two budgets are worth separating rather than the two contexts. Clamping the *wait* to the caller's deadline is right and is what I asked for last round. Clamping the *cancel RPC* to it is what breaks, and the rest of the repo already does the opposite for cleanup work: [`RepairContext`](https://github.com/google/adk-go/blob/aa8d077573c21eb705e197a049d2bfea11d82480/internal/compactioninternal/apply.go#L61-L74) and [auth/providers.go:196-199](https://github.com/google/adk-go/blob/aa8d077573c21eb705e197a049d2bfea11d82480/auth/providers.go#L196-L199) both run detached on `WithTimeout(context.WithoutCancel(ctx), …)`, and so did `cleanupRemoteTask` itself at [a2a_agent.go:520 and :530](https://github.com/google/adk-go/blob/ffddb2a6d7ed97974e82815d00c95eb148038f2c/agent/remoteagent/v2/a2a_agent.go#L520-L531) before this change. Falling back to a detached budget when the shared one is already spent is three lines, restores the merge-base behaviour, and leaves every test in this PR green — I ran it.

One related wording point: the doc text at [a2a_agent.go:320-323](https://github.com/google/adk-go/blob/aa8d077573c21eb705e197a049d2bfea11d82480/agent/remoteagent/v2/a2a_agent.go#L320-L323) says the caller's deadline is "an upper bound" for the wait and for the default cancel RPC. On a deadline that has already passed it is not a bound on the RPC, it removes the RPC, and a reader would not expect that from the sentence.

<details><summary>Failing test — drops into <code>a2a_agent_test.go</code> unchanged, it uses your own <code>newRemoteAgentWithClient</code></summary>

```go
type deadlineProbeClient struct {
	canceled chan a2a.TaskID
}

func (c *deadlineProbeClient) SendMessage(context.Context, *a2a.SendMessageRequest) (a2a.SendMessageResult, error) {
	return nil, fmt.Errorf("not implemented")
}

func (c *deadlineProbeClient) SendStreamingMessage(ctx context.Context, _ *a2a.SendMessageRequest) iter.Seq2[a2a.Event, error] {
	return func(yield func(a2a.Event, error) bool) {
		yield(&a2a.Task{
			ID:        "remote-task",
			ContextID: "remote-context",
			Status:    a2a.TaskStatus{State: a2a.TaskStateWorking},
		}, nil)
		<-ctx.Done() // stream stays open on the context it was handed
	}
}

func (c *deadlineProbeClient) CancelTask(_ context.Context, req *a2a.CancelTaskRequest) (*a2a.Task, error) {
	c.canceled <- req.ID
	return &a2a.Task{ID: req.ID, Status: a2a.TaskStatus{State: a2a.TaskStateCanceled}}, nil
}

func (c *deadlineProbeClient) Destroy() error { return nil }

// The invocation carries a 300ms deadline. A non-terminal Task arrives at once,
// so the remote task ID is known. The deadline then expires.
func TestRemoteAgent_CancelsRemoteTaskWhenInvocationDeadlineExpires(t *testing.T) {
	client := &deadlineProbeClient{canceled: make(chan a2a.TaskID, 1)}
	remoteAgent := newRemoteAgentWithClient(t, client)

	ctx, cancel := context.WithTimeout(context.Background(), 300*time.Millisecond)
	defer cancel()
	sess := prepareSession(t, ctx, []*session.Event{newUserHello()})
	ictx := icontext.NewInvocationContext(ctx, icontext.InvocationContextParams{
		Session:   sess,
		RunConfig: &agent.RunConfig{StreamingMode: agent.StreamingModeSSE},
	})

	done := make(chan struct{})
	go func() {
		defer close(done)
		for range remoteAgent.Run(ictx) {
		}
	}()
	<-done

	select {
	case id := <-client.canceled:
		if id != "remote-task" {
			t.Fatalf("CancelTask() task ID = %q, want %q", id, "remote-task")
		}
	case <-time.After(time.Second):
		t.Fatal("CancelTask() was not sent for a known non-terminal remote task")
	}
}
```

```
go test ./agent/remoteagent/v2/ -run TestRemoteAgent_CancelsRemoteTaskWhenInvocationDeadlineExpires
```

</details>

## What this round resolves

Everything else I raised last time is closed, and I checked each of them by breaking the new code rather than by reading the diff.

- **The task-ID predicate and cleanup now share one function**, so the wait ends only on an event cleanup will act on. Both the task-scoped-message shape and the completed-task-then-artifact shape are gone, and putting the old permissive predicate back fails the suite.
- **The two budgets no longer compose.** With a cleanup-eligible event arriving 4.9 s after a cancel, `CancelTask` goes out with 97 ms of the single budget left, where the previous shape spent 5 s waiting and then started a fresh 5 s.
- **The stream context now carries the caller's deadline**, so an invocation with a 300 ms timeout returns at 300 ms rather than the 5304 ms I measured last round, and the transport can size its own timeouts again.
- **Deleting the post-cancellation drain, or reverting the detach, now fails tests** — including `TestA2ACleanupPropagation`, which covers both cancel orderings over a real HTTP transport. The comment describing the limitation this PR removes is gone with it.
- **The concurrency holds up.** `-race` is clean on the package, and it does reach the new goroutine rather than passing vacuously: an unsynchronized write shared between the `AfterFunc` body and `stopWaiting` is reported. Goroutine counts are flat over 30 cancelled invocations and over 5 that burn the full wait. `apidiff`, `gofmt`, `go vet`, `golangci-lint`, `staticcheck` and `govulncheck` are clean or byte-identical to the merge-base, and the suite result matches the merge-base package for package.

agent/remoteagent/v2/a2a_agent_test.go
ktsoator: Addressed. The shared-budget test now compares both the CancelTask deadline and the elapsed duration against a literal 5 * time.Second, so changing remoteTaskCleanupTimeout causes the test to fail.

agent/remoteagent/v2/a2a_agent.go
ktsoator: Thank you for catching this. TaskArtifactUpdateEvent task information is now retained only as a fallback when no state-bearing cleanup target has been observed. Artifacts received during post-cancellation draining do not stop the drain, and a later Task or TaskStatusUpdateEvent replaces the fallback. Added coverage verifies artifact-only streams, drain-time discovery, and terminal status precedence.

ktsoator:
Thank you for the follow-up review and for providing the focused regression
test. I addressed the remaining deadline regression and both inline comments.

When the shared cleanup context has already expired, the default cleanup now
falls back to a fresh detached five-second context before calling CancelTask.
The new regression test verifies that a known non-terminal task is canceled
after the invocation deadline expires and that the RPC receives an active
context.

Artifact task information is now retained only as a fallback when no
state-bearing target is available. It does not overwrite a terminal target or
prematurely stop post-cancellation draining before a later terminal status.
The shared-budget test also pins the five-second policy with literal expected
values, and the public documentation now distinguishes the caller-bounded
task-information wait from the detached fallback cancel RPC.

After merging the latest main branch, the full workspace build and
race-enabled test suite pass. Lint reports zero issues in both modules, and
both go mod tidy -diff checks are clean.

Thank you again for the detailed review. Could you please take another look
when convenient?

agent/remoteagent/v2/a2a_agent.go
```diff
@@ -414,20 +506,47 @@ func (a *a2aAgent) run(ctx agent.InvocationContext, cfg A2AConfig) iter.Seq2[*se
...
+				return false
+			}
+			taskInfo := artifact.TaskInfo()
+			if taskInfo.TaskID == "" {
+				return false
+			}
+			lastCleanupTarget = &remoteTaskCleanupTarget{taskInfo: taskInfo}
+
+			// During post-cancellation draining, keep waiting for an event carrying task state.
+			return ctx.Err() == nil
```
karolpiotrowicz: Returning `ctx.Err() == nil` here means the drain never stops early once the target came from an artifact, because during a drain this is always false. The caller of this closure only sets `cleanupTargetKnown` on a `true` return, so `stopWaiting()` is not reached and the wait runs to the full five seconds with the task ID already in hand.

Measured against a remote that answers `CancelTask` promptly, artifact arriving 200 ms after cancellation, three runs each: `Run` returns 5.004 s after the cancel on this branch and 200 ms on `main`. Against a remote that is also slow to answer, the drain budget and the RPC budget add rather than share — 10.0 s against 5.2 s.

Either treat a known task ID as sufficient to stop the drain when no state-bearing event follows, or compute one absolute deadline at budget start so the drain and the RPC together can never exceed `remoteTaskCleanupTimeout`.

agent/remoteagent/v2/a2a_agent.go
```diff
@@ -365,6 +369,94 @@ type a2aAgent struct {
...
+	startOnce  sync.Once
+	ctx        context.Context
+	cancel     context.CancelFunc
+}
+
+func newRemoteTaskCleanupBudget(parent context.Context, timeout time.Duration) *remoteTaskCleanupBudget {
+	base := context.WithoutCancel(parent)
+	var cancelBase context.CancelFunc
+	if deadline, ok := parent.Deadline(); ok {
+		base, cancelBase = context.WithDeadline(base, deadline)
```
karolpiotrowicz, line 398: Putting the caller's deadline on the detached base means the drain is exactly zero whenever the invocation ends *because* that deadline fired: the stream context derives from this same base, so both are already done at that instant and the `AfterFunc` select falls straight through.

So a deadline-bounded invocation whose task ID has not yet appeared on the stream still orphans the remote task. Measured with a 100 ms deadline and the remote withholding the task object until 300 ms, 20 runs at `GOMAXPROCS=1`: 0 of 20 delivered on this branch and 0 of 20 on `main`, `CancelTask` never invoked on either. Not a regression — the same shape with `context.WithCancel` goes 0 of 20 to 20 of 20, which is the fix working. Worth a test so the gap does not get rediscovered as a bug later.

agent/remoteagent/v2/a2a_agent.go
```diff
@@ -414,20 +506,47 @@ func (a *a2aAgent) run(ctx agent.InvocationContext, cfg A2AConfig) iter.Seq2[*se
...
+		rememberCleanupTarget := func(event a2a.Event) bool {
+			target, ok := remoteTaskCleanupTargetFromEvent(event)
+			if ok {
+				lastCleanupTarget = &target
+				return true
+			}
+			if lastCleanupTarget != nil {
+				return false
+			}
+			artifact, ok := event.(*a2a.TaskArtifactUpdateEvent)
```
karolpiotrowicz, line 528: The fallback accepts only `*a2a.TaskArtifactUpdateEvent`, so a task-scoped `*a2a.Message` — which carries a non-empty `TaskID` just as an artifact does — is neither a cleanup target here nor a wait-stopper at the direct-response check below. A stream whose only post-cancellation output is such messages therefore burns the whole budget and then cancels nothing, despite having held the ID the entire time.

`main` refused `*a2a.Message` too, so this is a gap in the new fallback rather than something this PR broke, and I can see that continuing to wait past one is deliberate since it carries no task state. Accepting any event with a non-empty `TaskID` into the `lastCleanupTarget == nil` fallback, while keeping the same "keep draining for state" return, would cover it without changing that.

agent/remoteagent/v2/a2a_agent_test.go
```diff
@@ -1480,6 +1481,813 @@ func TestRemoteAgent_CleanupCallback(t *testing.T) {
...
+		}()
+		<-runDone
+
+		select {
+		case got := <-cancelTaskCalls:
+			if got.taskID != "remote-task" {
+				t.Fatalf("CancelTask() task ID = %q, want %q", got.taskID, "remote-task")
+			}
+			if got.contextErr != nil {
+				t.Fatalf("CancelTask() context error = %v, want nil", got.contextErr)
```
karolpiotrowicz, line 2138: This assertion does not discriminate. A context whose deadline has already passed but whose `Err()` has not been set yet satisfies `contextErr == nil`, which is exactly the state the cleanup path hands `CancelTask` on the deadline-expiry path, so the test passes over the defect described in the review.

Recording `ctx.Deadline()` in the same fake and failing when under a millisecond remains makes this test fail 7 runs in 20 with `remaining=0s`, and pass 20 in 20 once the predicate is corrected. Note that is a sound regression guard after a fix rather than a reliable detector before one.

The three other run-level cleanup tests in this file take the context as `_`, so they cannot see it either.

karolpiotrowicz (changes requested):
The detached fallback you added is the right shape, and it does fire — but the test that guards it asks the wrong question, so at the exact moment it is needed it usually does not fire. A deadline-bounded invocation still orphans its remote task, and I can now reproduce that on ordinary multi-core hardware, not only on one CPU. That is the one thing still blocking, and closing it means the predicate must ask whether the budget has usable time left rather than whether it has already errored — treating a context whose deadline has already passed as spent is what takes the measurement below back to 100/100. Everything else below is non-blocking, and the two smaller code points are on the lines they concern.

## The cancel RPC still goes out on a context with no time left

[`cleanupRemoteTask`](https://github.com/google/adk-go/blob/d9cf1ea4a4927164b593e7aa6200807869f81ec6/agent/remoteagent/v2/a2a_agent.go#L696-L701) asks `cleanupCtx.Err() != nil`, which tests whether the budget has *already errored* rather than whether it has time an RPC could use. Those two differ at exactly the instant that matters. [`newRemoteTaskCleanupBudget`](https://github.com/google/adk-go/blob/d9cf1ea4a4927164b593e7aa6200807869f81ec6/agent/remoteagent/v2/a2a_agent.go#L388-L395) re-imposes the caller's deadline on the detached base and [`start()`](https://github.com/google/adk-go/blob/d9cf1ea4a4927164b593e7aa6200807869f81ec6/agent/remoteagent/v2/a2a_agent.go#L403-L408) clamps to it, so when the invocation ends *because* its deadline fired, the base's deadline timer and the `context.AfterFunc` that starts the budget are two independent timers due at the same instant. In the common ordering the base's cancellation has not propagated yet, `Err()` is still nil on a context whose deadline is already in the past, the fallback is skipped, and `CancelTask` runs on it and dies before the request leaves the process.

The measurement is a real `httptest` A2A JSON-RPC server behind `a2asrv.NewJSONRPCHandler` with a real `a2aclient`, a 400 ms invocation deadline, a non-terminal task delivered at t≈0 so the ID is known, and the executor still working when the deadline fires. Counted: how often the server's cancel handler actually sees the RPC.

| | `GOMAXPROCS=1` | `GOMAXPROCS=2`, 300 runs | `GOMAXPROCS=2` and `64`, pooled |
|---|---|---|---|
| `main` ([a6114727](https://github.com/google/adk-go/commit/a61147276167f6b1b636d863f23a030c0a6babdf)) | 100/100 | 300/300 | 400/400 |
| this branch ([d9cf1ea4](https://github.com/google/adk-go/commit/d9cf1ea4a4927164b593e7aa6200807869f81ec6)) | 1/100 | 297/300 | 592/600 |

Eight orphans in 600 runs at two or more CPUs, against none in 400 on `main` — one-sided Fisher p = 0.016, so this is not confined to single-CPU hosts, though that is where it is close to total. Three controls, because a timing number is worth nothing without them. The same harness delivers 0/10 at the head I blocked last time ([aa8d0775](https://github.com/google/adk-go/commit/aa8d077573c21eb705e197a049d2bfea11d82480)), so it can produce a negative. The identical probe with `context.WithCancel` instead of `context.WithTimeout` delivers 20/20 on **both** trees, so the caller's deadline is the only variable. And treating a context whose deadline has already passed as spent, in addition to one that has already errored, takes the 100-run measurement back to 100/100 and leaves `go test ./agent/remoteagent/...` green — I ran it to confirm the diagnosis rather than to prescribe the fix.

The cheapest way to see it is inside your own test. [`TestRemoteAgent_CancelsRemoteTaskWhenInvocationDeadlineExpires`](https://github.com/google/adk-go/blob/d9cf1ea4a4927164b593e7aa6200807869f81ec6/agent/remoteagent/v2/a2a_agent_test.go#L1956-L1962) asserts `got.contextErr != nil` fails, and an already-past-deadline context satisfies that too, because its `Err()` has not been set yet. Recording `ctx.Deadline()` in the same fake and failing when under a millisecond remains makes that test fail 7 runs in 20 with `remaining=0s`, and pass 20 in 20 once the predicate is corrected. Worth adding either way, though note it is a sound regression guard after the fix rather than a reliable detector before one.

<details><summary>Reproduction — drops into <code>a2a_agent_test.go</code> unchanged, it uses your own <code>startA2AServer</code> and <code>mockA2AExecutor</code></summary>

```go
const noDeadline = time.Duration(1) << 62

type budgetProbeClient struct {
	inner     *a2aclient.Client
	budgetOut chan time.Duration
}

func (c *budgetProbeClient) SendMessage(ctx context.Context, req *a2a.SendMessageRequest) (a2a.SendMessageResult, error) {
	return c.inner.SendMessage(ctx, req)
}

func (c *budgetProbeClient) SendStreamingMessage(ctx context.Context, req *a2a.SendMessageRequest) iter.Seq2[a2a.Event, error] {
	return c.inner.SendStreamingMessage(ctx, req)
}

func (c *budgetProbeClient) CancelTask(ctx context.Context, req *a2a.CancelTaskRequest) (*a2a.Task, error) {
	rem := noDeadline
	if dl, ok := ctx.Deadline(); ok {
		rem = time.Until(dl)
	}
	select {
	case c.budgetOut <- rem:
	default:
	}
	return c.inner.CancelTask(ctx, req)
}

func (c *budgetProbeClient) Destroy() error { return c.inner.Destroy() }

// Real A2A JSON-RPC server, real client. The invocation carries a 400ms
// deadline, a non-terminal task arrives at once, and the deadline then fires
// while the remote is still working.
func TestRemoteAgent_RealTransportCancelReachesServerOnDeadlineExpiry(t *testing.T) {
	const runs = 100
	delivered, subMilli := 0, 0
	for i := 0; i < runs; i++ {
		cancelReceived := make(chan a2a.TaskID, 1)
		release := make(chan struct{})
		var releaseOnce sync.Once
		var wg sync.WaitGroup

		server := startA2AServer(&mockA2AExecutor{
			cancelFn: func(_ context.Context, reqCtx *a2asrv.ExecutorContext) iter.Seq2[a2a.Event, error] {
				return func(yield func(a2a.Event, error) bool) {
					select {
					case cancelReceived <- reqCtx.TaskID:
					default:
					}
					yield(a2a.NewStatusUpdateEvent(reqCtx, a2a.TaskStateCanceled, nil), nil)
				}
			},
			executeFn: func(ctx context.Context, reqCtx *a2asrv.ExecutorContext) iter.Seq2[a2a.Event, error] {
				return func(yield func(a2a.Event, error) bool) {
					wg.Add(1)
					defer wg.Done()
					if !yield(a2a.NewSubmittedTask(reqCtx, reqCtx.Message), nil) {
						return
					}
					select {
					case <-release:
					case <-ctx.Done():
					case <-time.After(10 * time.Second):
					}
				}
			},
		})

		card := &a2a.AgentCard{
			SupportedInterfaces: []*a2a.AgentInterface{a2a.NewAgentInterface(server.URL, a2a.TransportProtocolJSONRPC)},
			Capabilities:        a2a.AgentCapabilities{Streaming: true},
		}
		budgetOut := make(chan time.Duration, 1)
		remoteAgent := utils.Must(NewA2A(A2AConfig{
			Name:      "remote-agent",
			AgentCard: card,
			ClientProvider: func(ctx context.Context, c *a2a.AgentCard) (A2AClient, error) {
				inner, err := a2aclient.NewFromCard(ctx, c)
				if err != nil {
					return nil, err
				}
				return &budgetProbeClient{inner: inner, budgetOut: budgetOut}, nil
			},
		}))

		base := context.Background()
		sess := prepareSession(t, base, []*session.Event{newUserHello()})
		ctx, cancel := context.WithTimeout(base, 400*time.Millisecond)
		ictx := icontext.NewInvocationContext(ctx, icontext.InvocationContextParams{
			Session:   sess,
			RunConfig: &agent.RunConfig{StreamingMode: agent.StreamingModeSSE},
		})

		done := make(chan struct{})
		go func() {
			defer close(done)
			for range remoteAgent.Run(ictx) {
			}
		}()
		<-done

		if rem := <-budgetOut; rem != noDeadline && rem < time.Millisecond {
			subMilli++
		}
		select {
		case <-cancelReceived:
			delivered++
		}

		releaseOnce.Do(func() { close(release) })
		wg.Wait()
		server.Close()
		cancel()
	}
	t.Logf("server received CancelTask %d/%d; RPC budget under 1ms %d/%d", delivered, runs, subMilli, runs)
	if delivered != runs {
		t.Fatalf("the remote task was orphaned in %d of %d runs", runs-delivered, runs)
	}
}
```

```
GOMAXPROCS=1 go test ./agent/remoteagent/v2/ -run TestRemoteAgent_RealTransportCancelReachesServerOnDeadlineExpiry -v
```

</details>

## A caller that cancels now waits five seconds where it waited a fifth of one

This one is not blocking, but it is the largest behavior change in the PR that nothing in the suite measures. When the first event carrying the task ID is a `TaskArtifactUpdateEvent` and the remote keeps the stream open, [`rememberCleanupTarget`](https://github.com/google/adk-go/blob/d9cf1ea4a4927164b593e7aa6200807869f81ec6/agent/remoteagent/v2/a2a_agent.go#L519-L533) records the target and then returns `ctx.Err() == nil`, which is false during a drain — so `stopWaiting()` is never reached and the drain runs to the full budget with the ID already in hand. Measured against a remote that answers `CancelTask` promptly, with the artifact arriving 200 ms after cancellation: `Run` returns 5.004 s after the cancel on this branch and 200 ms on `main`, three runs each, and the cancel RPC goes out 4.8 s later than it needs to. Against a remote that is also slow to answer the cancel, the drain budget and the RPC budget then add rather than share — 10.0 s against 5.2 s. I have not established how often a stream leads with an artifact rather than a `Task`, so treat the frequency as unknown. The cost when it does happen is measured.

## What this round resolves

I checked each of these by breaking the new code rather than by reading the diff.

- **`CancelTask` is now always attempted after a deadline expires.** The detached fallback closes the "no RPC at all" half of what I raised last round.
- **The post-cancellation drain genuinely fixes the case it was written for.** With the remote withholding the task object until 300 ms and the caller cancelling at 100 ms, `main` delivers 0 of 20 cancels and never invokes `CancelTask` on the client at all. This branch delivers 20 of 20.
- **The concurrency holds up, and the clean `-race` result is load-bearing rather than vacuous.** Planting an unsynchronized counter shared between the `AfterFunc` body and `stopWaiting` produces three race reports naming exactly those two sites, from three different tests including the real-transport end-to-end one — so the package's tests do drive that goroutine concurrently.
- **The rest of the battery is clean.** `gofmt`, `go vet`, `go mod tidy -diff` and `golangci-lint` are clean on both modules, `apidiff` reports no exported API change, `staticcheck`'s two findings are byte-identical to `main` and both sit in the deprecated v1 package, and the suite passes on 164 of 165 root packages on both trees plus `plugin/agentanalytics`. The one package neither arm exercised is `session/vertexai`, which hangs identically on `main` in my environment.

## Smaller things

- **The deadline path gets no drain at all**, so the fix does not reach it. When the invocation ends by deadline expiry before the task ID has ever appeared on the stream, the budget's base carries the same deadline, so the wait is structurally zero and the task is orphaned — 0 of 20 on this branch and 0 of 20 on `main`, so it is not a regression, just an unfixed half. The doc sentence is accurate about it, since "bounded by the caller's deadline" does mean zero when the deadline is what ended the run, but a test pinning that would stop it being rediscovered.
- **The v1 doc now describes behavior v1 no longer has.** `agent/remoteagent` delegates to `v2.NewA2A` at [a2a_agent.go:296](https://github.com/google/adk-go/blob/d9cf1ea4a4927164b593e7aa6200807869f81ec6/agent/remoteagent/a2a_agent.go#L296), so v1 embedders inherit the drain and the clamp, while [a2a_agent.go:124](https://github.com/google/adk-go/blob/d9cf1ea4a4927164b593e7aa6200807869f81ec6/agent/remoteagent/a2a_agent.go#L124) still says the default is "a cancel RPC request with 5 second timeout" — the sentence this PR corrected in v2.
- **`AfterA2ARequestCallback`'s doc says the callback fires for every streamed response**, at [a2a_agent.go:52](https://github.com/google/adk-go/blob/d9cf1ea4a4927164b593e7aa6200807869f81ec6/agent/remoteagent/v2/a2a_agent.go#L50-L52), but conversion now stops once the invocation is cancelled. In practice that is the terminal error event per cancelled invocation rather than a stream of them, since there was no drain window before this change for anything else to arrive in.
- **A task-scoped `*a2a.Message` is neither a cleanup target nor a wait-stopper**, so a stream whose only post-cancellation output is such messages burns the whole budget and then cancels nothing, despite having held the ID throughout. `main` refused those too, so this is a gap in the new fallback rather than something the PR broke, and waiting past them is deliberate since they carry no task state.

ktsoator:
Thank you for the follow-up review and for the deadline probe. I pushed commit 4ae95ba with the remaining changes.

Cleanup now treats a context with an already-passed deadline as unusable, even if Err() has not been set yet. CancelTask falls back to a fresh detached five-second timeout in that case.

An artifact carrying a task ID now stops the drain immediately, and CancelTask reuses the remaining shared budget. I also added coverage for the case where the invocation deadline expires before any task information arrives, and updated the v1 and callback documentation.

For task-scoped messages, I checked the behavior in a2a-go v2.5 before changing it. Its taskupdate.IsFinal logic treats every Message as a final response; a Message TaskID associates the response with a task but does not by itself mean there is a cancellable stored task. Based on that, the stream now stops draining for any Message without calling CancelTask. A final Message also clears an earlier non-terminal cleanup target, covering the Task{working} -> Message{final} sequence.

Does this match your understanding of the intended Message semantics, or would you still prefer retaining the Message TaskID as a cleanup fallback?

The full workspace build and race-enabled test suite pass. Lint reports zero issues in both modules, both go mod tidy -diff checks are clean, and the GitHub checks are green.

Could you please take another look when convenient?

agent/remoteagent/v2/a2a_agent_test.go
ktsoator: Fixed. The test now records the `CancelTask` context deadline and verifies that the RPC receives a usable detached budget instead of an already-expired context.

agent/remoteagent/v2/a2a_agent.go
ktsoator: Added a test for this case and documented the limitation: if the invocation deadline expires before any task information arrives, there is no remaining drain window.

agent/remoteagent/v2/a2a_agent.go
ktsoator: I checked the behavior in `a2a-go v2.5`. Since every `Message` is treated as final and its `TaskID` does not necessarily identify a cancellable stored task, the stream now stops draining without retaining it as a cleanup fallback. Do you think the `TaskID` should still be retained here?

agent/remoteagent/v2/a2a_agent.go
ktsoator: Fixed. An artifact task ID now stops the drain, and `CancelTask` uses the remaining shared budget instead of starting another five-second window.

agent/remoteagent/v2/a2a_agent.go
```diff
@@ -483,55 +602,114 @@ func (a *a2aAgent) run(ctx agent.InvocationContext, cfg A2AConfig) iter.Seq2[*se
...
+	return target, true
+}
+
+func contextHasUsableTime(ctx context.Context) bool {
+	if ctx == nil || ctx.Err() != nil {
+		return false
 	}
-	if state.Terminal() {
+	deadline, ok := ctx.Deadline()
+	return !ok || time.Until(deadline) > 0
```
karolpiotrowicz: This is the line the blocking point in the review body turns on. `time.Until(deadline) > 0` is true for a remainder of one microsecond, which is as unusable to an RPC as a remainder of zero, so the detached fallback below is skipped exactly when it is most needed. A minimum remainder here, with the fallback taking over below it, put every row of the measurement back to 20 of 20.

agent/remoteagent/v2/a2a_agent.go
```diff
@@ -483,55 +602,114 @@ func (a *a2aAgent) run(ctx agent.InvocationContext, cfg A2AConfig) iter.Seq2[*se
...
-		state = t.Status.State
+	return target, true
+}
+
+func contextHasUsableTime(ctx context.Context) bool {
+	if ctx == nil || ctx.Err() != nil {
+		return false
 	}
-	if state.Terminal() {
+	deadline, ok := ctx.Deadline()
```
karolpiotrowicz, line 683: Minor, and I would leave the line as it is: the `!ok` arm cannot be reached from the one call site, because `cleanupCtx` is either nil (already handled above) or the value of `cleanupBudget.start()`, which is always a `context.WithTimeout` and so always carries a deadline. Worth knowing rather than changing, since it means a test that exercises the deadline-less case is testing an input production cannot produce.

agent/remoteagent/v2/a2a_agent.go
```diff
@@ -483,55 +602,114 @@ func (a *a2aAgent) run(ctx agent.InvocationContext, cfg A2AConfig) iter.Seq2[*se
 
 		if ctx.RunConfig().StreamingMode == agent.StreamingModeNone {
 			a2aEvent, a2aErr := sender.SendMessage(ctx, req)
+			rememberCleanupTarget(a2aEvent)
```
karolpiotrowicz, line 610: This call site is executed by `TestRemoteAgent_ADK2ADK`, which drives `StreamingModeNone` against a real server, but nothing asserts its effect: deleting the line entirely leaves `go test ./agent/remoteagent/...` green. The gap is that the only `StreamingModeNone` test runs to a terminal task, so `cleanupRemoteTask` returns before reaching the cancel. A case where the non-streaming response is a non-terminal task would pin it.

agent/remoteagent/v2/a2a_agent.go
```diff
@@ -483,55 +602,114 @@ func (a *a2aAgent) run(ctx agent.InvocationContext, cfg A2AConfig) iter.Seq2[*se
...
+			if a2aEvent != nil {
+				if rememberCleanupTarget(a2aEvent) {
+					if !cleanupTargetKnown {
+						cleanupTargetKnown = true
+						streamCtx.stopWaiting()
+					}
+				}
+				if msg, ok := a2aEvent.(*a2a.Message); ok && msg != nil {
+					lastCleanupTarget = nil
+					directResponseReceived = true
```
karolpiotrowicz: Naming nit now that the `msg.TaskID == ""` guard is gone: this flag is set by a task-scoped message too, and such a message is not a direct response. `finalMessageReceived` would describe the condition it now holds, and it is the flag that short-circuits the drain a few lines below.

agent/remoteagent/v2/a2a_agent_test.go
```diff
@@ -1480,6 +1481,909 @@ func TestRemoteAgent_CleanupCallback(t *testing.T) {
...
+			}
+			if !got.hasDeadline {
+				t.Fatal("CancelTask() context has no deadline")
+			}
+			if got.deadlineRemaining < time.Millisecond {
```
karolpiotrowicz: This assertion states the standard the code does not enforce: the test requires at least a millisecond of budget, while `contextHasUsableTime` accepts any positive remainder. Worth reconciling the two in whichever direction you settle on. Note also that inside a `synctest` bubble this assertion cannot fail against the previous predicate, because the bubble fires all timers due at one fake instant before the observing goroutine resumes, so the budget has already errored by the time cleanup looks.

karolpiotrowicz (commented):
The predicate change does what it needed to do. `contextHasUsableTime` closes the case I blocked on last round: with a 400 ms invocation deadline and a non-terminal task known at t≈0, the server's cancel handler now runs 100 times in 100 at `GOMAXPROCS=1` and 100 in 100 at `GOMAXPROCS=64`, against 1 in 100 at the previous head. The cancel RPC never received a sub-millisecond budget in any of those 200 runs. Same harness, same controls as last round: it still returns 0 of 20 at the head I first blocked ([aa8d0775](https://github.com/google/adk-go/commit/aa8d077573c21eb705e197a049d2bfea11d82480)), so it can still produce a negative, and swapping `context.WithTimeout` for `context.WithCancel` gives 20 of 20 on both trees.

One thing still needs fixing before this can merge, and it is the other half of the point I raised in the round-2 review rather than something new: the cleanup budget check at `contextHasUsableTime` has to reject a remainder too small to carry a round trip, not only a remainder of zero. Everything else below is non-blocking, and the small code points are on the lines they concern.

## The fallback triggers on "no time left" when it needs to trigger on "not enough time left"

[`contextHasUsableTime`](https://github.com/google/adk-go/blob/4ae95baa5c264bebcbb80c9ae0171fe6c128eb64/agent/remoteagent/v2/a2a_agent.go#L674-L680) returns true for any strictly positive remainder, so [`cleanupRemoteTask`](https://github.com/google/adk-go/blob/4ae95baa5c264bebcbb80c9ae0171fe6c128eb64/agent/remoteagent/v2/a2a_agent.go#L705) skips the detached fallback and issues `CancelTask` on whatever is left. Because [`newRemoteTaskCleanupBudget`](https://github.com/google/adk-go/blob/4ae95baa5c264bebcbb80c9ae0171fe6c128eb64/agent/remoteagent/v2/a2a_agent.go#L389-L402) re-imposes the caller's deadline on the detached base and [`start()`](https://github.com/google/adk-go/blob/4ae95baa5c264bebcbb80c9ae0171fe6c128eb64/agent/remoteagent/v2/a2a_agent.go#L404-L409) clamps to it, "whatever is left" is the time between the invocation ending and the caller's deadline. When those are close together the RPC gets a budget it cannot complete in, and dies before the request leaves the process.

The property, rather than the one failing case: **the cleanup budget has to be large enough to carry a round trip, not merely larger than zero.** A remainder of zero and a remainder of one microsecond are the same thing from the RPC's point of view, and only the first currently takes the fallback.

Measured against the same real `httptest` A2A server, with a 500 ms invocation deadline, a non-terminal task known at t≈0, the remote still working, and the caller cancelling at `deadline - ε`. Counted: how often the server's cancel handler actually runs.

| ε | `GOMAXPROCS` | this branch ([4ae95ba](https://github.com/google/adk-go/commit/4ae95baa5c264bebcbb80c9ae0171fe6c128eb64)) | smallest budget seen | `main` ([a611472](https://github.com/google/adk-go/commit/a61147276167f6b1b636d863f23a030c0a6babdf)) |
|---|---|---|---|---|
| 20 ms | 8 | 20/20 | 18.76 ms | 20/20 |
| 2 ms | 8 | 20/20 | 473.7 µs | 20/20 |
| **1 ms** | **8** | **15/20** | **1.4 µs** | **20/20** |
| **1 ms** | **1** | **10/20** | **60.2 µs** | **20/20** |
| 0 ms | 1 | 20/20, fallback fires | 5.000 s | — |

`CancelTask` was entered 20 times out of 20 in every row on this branch, so the RPC is attempted every time and simply does not arrive in the failing runs. `main` gives it a detached five seconds unconditionally and lands 20 of 20 in every row it was run. That makes this a case `main` handles and the branch does not.

Two things worth saying plainly about how far that generalises. It is measured on loopback, which is the most forgiving transport there is, so the window on a real network is proportionally wider than one millisecond rather than narrower. And it needs the invocation to end within about a round trip of the caller's own deadline, which I have not shown a real caller doing — I would not want to claim the shape is common, only that when it happens the task is orphaned and nothing says so beyond a `log.Warn`.

Your own test already states the standard the code does not enforce: [`a2a_agent_test.go:1973`](https://github.com/google/adk-go/blob/4ae95baa5c264bebcbb80c9ae0171fe6c128eb64/agent/remoteagent/v2/a2a_agent_test.go#L1973-L1974) fails when the `CancelTask` context has less than a millisecond left.

I ran a fix rather than only suggesting one. Requiring a minimum remainder before using the shared budget, with the fallback taking over below it, puts every row above at 20 of 20 including ε = 1 ms at both `GOMAXPROCS` settings, keeps the deadline-expiry case at 50 of 50, and leaves `go test ./agent/remoteagent/...` green.

I used 100 ms as the minimum to check that, and I have no strong view on the number — the choice is yours. Detaching the cancel RPC from the caller's deadline altogether would close it too, and that is what `RepairContext` in [`internal/compactioninternal/apply.go`](https://github.com/google/adk-go/blob/4ae95baa5c264bebcbb80c9ae0171fe6c128eb64/internal/compactioninternal/apply.go#L57-L74) and [`auth/providers.go`](https://github.com/google/adk-go/blob/4ae95baa5c264bebcbb80c9ae0171fe6c128eb64/auth/providers.go#L196-L199) already do for their own cleanup work.

<details><summary>Reproduction — drops into <code>a2a_agent_test.go</code> unchanged, it uses your own <code>startA2AServer</code> and <code>mockA2AExecutor</code></summary>

```go
func TestRemoteAgent_CancelReachesServerWhenCancelledNearDeadline(t *testing.T) {
	const runs = 20
	const deadline = 500 * time.Millisecond
	const eps = time.Millisecond

	delivered := 0
	for i := 0; i < runs; i++ {
		cancelReceived := make(chan a2a.TaskID, 1)
		release := make(chan struct{})
		var releaseOnce sync.Once
		var wg sync.WaitGroup

		server := startA2AServer(&mockA2AExecutor{
			cancelFn: func(_ context.Context, reqCtx *a2asrv.ExecutorContext) iter.Seq2[a2a.Event, error] {
				return func(yield func(a2a.Event, error) bool) {
					select {
					case cancelReceived <- reqCtx.TaskID:
					default:
					}
					yield(a2a.NewStatusUpdateEvent(reqCtx, a2a.TaskStateCanceled, nil), nil)
				}
			},
			executeFn: func(ctx context.Context, reqCtx *a2asrv.ExecutorContext) iter.Seq2[a2a.Event, error] {
				return func(yield func(a2a.Event, error) bool) {
					wg.Add(1)
					defer wg.Done()
					if !yield(a2a.NewSubmittedTask(reqCtx, reqCtx.Message), nil) {
						return
					}
					select {
					case <-release:
					case <-ctx.Done():
					case <-time.After(10 * time.Second):
					}
				}
			},
		})

		card := &a2a.AgentCard{
			SupportedInterfaces: []*a2a.AgentInterface{a2a.NewAgentInterface(server.URL, a2a.TransportProtocolJSONRPC)},
			Capabilities:        a2a.AgentCapabilities{Streaming: true},
		}
		remoteAgent := utils.Must(NewA2A(A2AConfig{Name: "remote-agent", AgentCard: card}))

		base := context.Background()
		sess := prepareSession(t, base, []*session.Event{newUserHello()})
		ctx, cancel := context.WithTimeout(base, deadline)
		ictx := icontext.NewInvocationContext(ctx, icontext.InvocationContextParams{
			Session:   sess,
			RunConfig: &agent.RunConfig{StreamingMode: agent.StreamingModeSSE},
		})

		done := make(chan struct{})
		go func() {
			defer close(done)
			for range remoteAgent.Run(ictx) {
			}
		}()
		// Cancel while the caller's own deadline is still a millisecond away.
		timer := time.AfterFunc(deadline-eps, cancel)
		<-done
		timer.Stop()

		select {
		case <-cancelReceived:
			delivered++
		case <-time.After(2 * time.Second):
		}

		releaseOnce.Do(func() { close(release) })
		wg.Wait()
		server.Close()
		cancel()
	}
	if delivered != runs {
		t.Fatalf("the remote task was orphaned in %d of %d runs", runs-delivered, runs)
	}
}
```

```
GOMAXPROCS=8 go test ./agent/remoteagent/v2/ -run TestRemoteAgent_CancelReachesServerWhenCancelledNearDeadline -v
```

</details>

## Task-scoped messages: your reading is right, keep the change

You asked directly, so a direct answer. I checked `a2a-go v2.5.0` rather than reasoning from the type name, and your conclusion holds. `internal/taskupdate/final.go` returns true for every `*a2a.Message`, and `internal/taskexec/subscription.go` returns from the subscription as soon as it yields a final event, so an `a2asrv`-backed server cannot continue a task stream past a Message at all. The A2A specification agrees: the streaming section lists only `Task`, `TaskStatusUpdateEvent` and `TaskArtifactUpdateEvent` as task-stream events, and puts intermediate agent text inside `TaskStatusUpdateEvent.status.message`.

I also measured it end to end, because the shape I raised last round was worth settling rather than arguing. With `Task{submitted}` at t≈0, a task-scoped `a2a.NewMessageForTask` mid-stream while the invocation is still live, and the caller cancelling afterwards, no cancel reaches the server on this branch — and none reaches it on `main` either, 0 of 20 on both, at both `GOMAXPROCS` settings and also when a further `working` status update follows the message. Issuing `CancelTask` by hand from a separate client afterwards fails 10 times in 10 with `task in non-cancelable state TASK_STATE_FAILED`.

So there is nothing to preserve here and no regression to undo. Retaining the message's `TaskID` as a cleanup fallback, which was my suggestion in an earlier round, would buy a refused RPC. Treat that earlier point as withdrawn.

## The guard on the deadline fix is unreliable

This is not blocking, but it is why the same defect has now cost three rounds. The new assertion in `TestRemoteAgent_CancelsRemoteTaskWhenInvocationDeadlineExpires` is inside a `synctest` bubble, and a bubble fires all timers due at one fake instant before the observing goroutine resumes. The budget's `Err()` is therefore always already set by the time cleanup looks, so the `time.Until(deadline) > 0` clause never gets to decide anything and the test passes whether or not the clause is there.

Concretely: reverting `contextHasUsableTime` to the previous `Err()`-only predicate leaves `go test ./agent/remoteagent/v2/` green in 9 runs out of 9. Widening to `go test ./agent/remoteagent/...` catches it about a third of the time, and whether it catches it depends on how busy the machine is rather than on the code. The unchanged tree never went red in any of those runs, so this is not ordinary flakiness — the suite simply cannot see this class of change reliably.

A direct table test of `contextHasUsableTime`, outside any bubble, fixes that. Feeding it a context that reports a deadline in the past while `Err()` is still nil catches the reverted predicate 3 times in 3, under the narrow selector where the current suite catches it 0 times in 9. A second case with a deadline a millisecond in the future would pin the floor from the first section at the same time.

<details><summary>The table test I ran</summary>

```go
// passedDeadlineContext reports a deadline that is already in the past while
// Err() is still nil. That is the state a caller's context is in between its
// deadline elapsing and its timer goroutine running, which is exactly when
// cleanup decides whether the shared budget is still usable.
type passedDeadlineContext struct{ context.Context }

func (passedDeadlineContext) Deadline() (time.Time, bool) {
	return time.Now().Add(-time.Second), true
}

func (passedDeadlineContext) Err() error { return nil }

func TestContextHasUsableTime(t *testing.T) {
	canceled, cancel := context.WithCancel(context.Background())
	cancel()
	future, cancelFuture := context.WithTimeout(context.Background(), time.Hour)
	defer cancelFuture()

	tests := []struct {
		name string
		ctx  context.Context
		want bool
	}{
		{name: "nil context", ctx: nil, want: false},
		{name: "no deadline, not canceled", ctx: context.Background(), want: true},
		{name: "no deadline, already canceled", ctx: canceled, want: false},
		{name: "deadline far in the future", ctx: future, want: true},
		{
			name: "deadline already passed but Err not yet set",
			ctx:  passedDeadlineContext{Context: context.Background()},
			want: false,
		},
	}
	for _, tt := range tests {
			if got := contextHasUsableTime(tt.ctx); got != tt.want {
				t.Errorf("contextHasUsableTime() = %v, want %v", got, tt.want)
			}
		})
	}
}
```

</details>

## What this round resolves

I checked each of these by breaking the new code rather than by reading the diff.

- **The deadline-expiry orphan is closed.** 100 of 100 at both `GOMAXPROCS=1` and `GOMAXPROCS=64`, with the cancel RPC always receiving a real budget.
- **The artifact drain no longer runs to the full five seconds** once the task ID is in hand, and the cancel goes out on the shared budget rather than a second one.
- **The v1 documentation and the `AfterA2ARequestCallback` documentation now match the code.** The stale comment in `a2a_e2e_test.go` about cancellation not propagating is gone too.
- **The rest of the battery is clean on both trees.** `gofmt`, `go vet`, `go mod tidy -diff` and `golangci-lint` are clean on both modules, `apidiff` reports no exported API change, `staticcheck`'s two findings are byte-identical to `main` and both sit in the deprecated v1 package, and the suite passes on 164 of 165 root packages on both trees plus `plugin/agentanalytics`. The one package neither tree exercised is `session/vertexai`, which hangs identically on `main` in my environment.

One caveat on the clean `-race` result, since I relied on it last round and cannot this time: every deadline-carrying test in the package now runs inside a `synctest` bubble, and the real-transport measurements above ran without `-race`, so it is evidence that nothing new surfaced rather than evidence that the concurrent path was exercised.

ktsoator:
Thanks for the detailed measurements and for checking the Message semantics. I have the changes ready locally and have verified them; they are not pushed yet.

I used the 100 ms minimum you tested. If less time remains, `CancelTask` gets a detached five-second timeout. Both API docs now explain that this fallback can extend total cleanup time.

I added direct context tests that catch the Err()-only predicate, plus run-level tests covering 1, 99, 100 and 101 ms remaining. The new non-streaming cleanup test also fails if the cleanup-target recording is removed. The flag is now named `finalMessageReceived`.

Real HTTP results with `-race`:

| Cancellation timing | GOMAXPROCS | Before (`4ae95ba`) | After (local changes) |
| --- | ---: | ---: | ---: |
| 20 ms before deadline | 8 | Not run | 20/20 |
| 2 ms before deadline | 8 | Not run | 20/20 |
| 1 ms before deadline | 8 | 0/20 | 20/20 |
| 1 ms before deadline | 1 | 0/20 | 20/20 |
| Deadline expiry | 1 | Not run | 20/20 |

Counts are cancellations received by the server. The fixed version gave every RPC close to five seconds, with no race reports.

The original timer often fired late on my machine, so for this comparison I woke it early and yielded until the target instant. Both versions used the same adjustment, and every scheduled cancellation occurred before the deadline. These are local loopback results with race instrumentation, so the failure rates are not directly comparable to your original measurements.

Build, full workspace race tests, lint and tidy passed before the final rename; the remoteagent race tests and the real HTTP probe passed afterwards.

agent/remoteagent/v2/a2a_agent.go
ktsoator: Updated locally to require at least 100 ms before reusing the budget; otherwise, `CancelTask` gets a detached five-second timeout.

agent/remoteagent/v2/a2a_agent.go
ktsoator: Thanks, I've kept the `!ok` check as it is.

agent/remoteagent/v2/a2a_agent.go
ktsoator: Added a local regression test for a non-streaming working task. Removing this line makes it fail with zero `CancelTask` calls.

agent/remoteagent/v2/a2a_agent.go
ktsoator: Renamed to `finalMessageReceived` in the local changes.

agent/remoteagent/v2/a2a_agent_test.go
ktsoator: The local test now expects the full five-second fallback budget. I also added a direct context test outside `synctest` that fails with the Err()-only predicate.

ktsoator:
Pushed in 01511cd and updated the PR description. All checks are green. Does the 100 ms minimum and detached fallback look reasonable to you?
:)

karolpiotrowicz:
@wolo-lab — flagging the number here for you rather than answering it myself, because it is a policy call about how the SDK behaves on a user's behalf rather than a correctness question.

The shape is settled and it is the one I measured last round: when the caller's remaining budget is too small to carry a cancel round trip, `cleanupRemoteTask` stops reusing it and takes a detached timeout instead. What is left is two constants, and both encode a product position rather than a fact about the code.

- [`remoteTaskCleanupMinRemaining = 100ms`](https://github.com/google/adk-go/blob/01511cdac0d47d3cd1fd6cb3395b32ba13059038/agent/remoteagent/v2/a2a_agent.go#L380) — the floor below which the caller's own budget is abandoned. Set it too low and a cancel RPC is issued that cannot arrive, which is the orphaned-task case this PR exists to close. Set it too high and we walk away from a budget that would have worked, spending the detached five seconds instead.
- [`remoteTaskCleanupTimeout = 5s`](https://github.com/google/adk-go/blob/01511cdac0d47d3cd1fd6cb3395b32ba13059038/agent/remoteagent/v2/a2a_agent.go#L376) — how long we keep working after the caller has given up. This is the part users notice, since a `Run` that returns can still hold a goroutine talking to the remote for another five seconds.

100 ms is the value I happened to use to demonstrate the fix, not a recommendation. It puts every row of my measurement at 20 of 20 including the one-millisecond case, but that was measured on loopback, so a floor tuned to it is tuned to the most forgiving transport there is.

There is a third option worth a moment before settling on a number, which is to detach the cancel RPC from the caller's deadline unconditionally and drop the threshold entirely. That is what [`RepairContext`](https://github.com/google/adk-go/blob/01511cdac0d47d3cd1fd6cb3395b32ba13059038/internal/compactioninternal/apply.go#L57-L74) and [`auth/providers.go`](https://github.com/google/adk-go/blob/01511cdac0d47d3cd1fd6cb3395b32ba13059038/auth/providers.go#L196-L199) already do for their own cleanup work, and it is what `main` does today. It trades a tunable for a uniform five seconds of post-cancellation work, and it means one fewer constant whose right value nobody can derive.

@ktsoator the code re-review is separate and I will come back to it — this is only about who owns the number.

ktsoator:
Thanks for clarifying. I understand that 100 ms was an experimental value, not a recommended default. I'll wait for @wolo-lab's preference on the cleanup policy before changing it further.

PR #1328, opened 2026-08-13
fix(remoteagent): reject card sources that are neither http(s) URLs nor paths

jjsasha63:
`NewAgentCardProvider` chose between fetching and reading with
`strings.HasPrefix` on `http://` or `https://`. Every other source fell through
to `os.ReadFile`, which took the whole string as a literal path.

Nothing was read that shouldn't have been — `os.ReadFile("file:///opt/card.json")`
looks for a file of that exact name and never reaches `/opt/card.json`. Two
things were wrong all the same:

- the failure named a path the caller never wrote. `file:///opt/card.json` and a
  typo like `htp://example.com` both reported `no such file or directory`
  against the whole string, which says nothing about what was actually wrong
- the prefix test was case sensitive, so `HTTP://example.com` was read off disk
  instead of fetched. That one is a plain bug

Classify on the scheme instead: `http` and `https` are fetched, any other scheme
is an error that names it, everything else is a path.

### What counts as a scheme

A colon alone does not make one. It is an ordinary character in a POSIX
filename, so `cards:v2/agent.json` and `card.json:stream` are readable files
this provider has always loaded, and they stay paths. A source is a URL only
once it carries the two slashes as well.

What may sit in front of `://` is left to `net/url`: the prefix is a scheme
exactly when `url.Parse` gives it back unchanged, case aside. That keeps the rule
RFC 3986's rather than a second opinion about it, and it settles the awkward
sources without a special case — `dir/sub://x` and `notes:/a://b` each hold a
character no scheme may hold, so both are paths.

### Behaviour change

- a source carrying a scheme other than `http` or `https` is now rejected by
  name, where it used to reach `os.ReadFile` whole and fail as a missing path
- `HTTP://example.com` now fetches
- `ErrUnsupportedCardSource` is exported, so a caller can separate this
  permanent failure from a fetch worth retrying. This is the only API addition
- paths are untouched, relative ones containing a colon included
- one Windows form does change: `C://cards/agent.json` is rejected as scheme
  `c`, since nothing about the length of a scheme makes it a path. `C:/` and
  `C:\` forms are unaffected, and the doubled slash only ever resolved through
  Windows path normalization

### Testing plan

`go build work`, `go test -race -count=1 -shuffle=on work`, `go vet`,
`golangci-lint run` and `go mod tidy -diff` in both modules: all clean.

Tests are a classification table plus provider-level cases for the uppercase
scheme, a local file, rejected schemes, and four relative paths containing a
colon written to disk and read back. Each was checked by mutation rather than
watched to pass: matching on the last `://` instead of the first, widening the
scheme test to a length check, and restoring the old prefix test each fail the
cases written for them.


karolpiotrowicz (commented):
`NewAgentCardProvider` is exported and shipped in v2.2.0, and this change rejects
relative paths that used to load. Any first path segment that starts with a
letter and contains a colon now parses as a scheme:

```
weird:name.json      -> unsupported agent card source: scheme "weird" is not supported
cards:v2/agent.json  -> scheme "cards" is not supported
card.json:stream     -> scheme "card.json" is not supported
```

Each of those is a readable file that `os.ReadFile` opens without complaint,
which is what the old code did with it. Absolute paths are unaffected because the
leading `/` stops any scheme being parsed, and that is why the [13-case
table](https://github.com/google/adk-go/blob/db550d64d321e82c835686e30cae3114cd7b6491/agent/remoteagent/v2/a2a_agent_test.go#L1509)
doesn't catch it — every path case in it is either absolute or has no colon.

Two claims in the description depend on this not happening:

- "those reads failed anyway, with a worse message" — for the inputs above they
  succeeded.
- The `file:///opt/card.json` motivation. `os.ReadFile` treats the whole string
  as a literal path, so it returns ENOENT and never reaches the target. Worth
  trying against a real file: a `file://` source pointing at a readable card
  returns `no such file or directory` on `main`. What this change delivers is a
  clearer error and case-insensitive scheme matching, both worth having, but the
  description reads as though a file-read hole is being closed.

Checking for `://` after the scheme rather than just the scheme text would
address the regression and the next item together.

The rest is smaller:

- **`len(scheme) == 1` isn't gated on GOOS.** On Linux `a:evil.json` is
  classified as a path and read, which is the reasonable thing to do with a
  filename someone passed. It does falsify the doc line at
  [L101](https://github.com/google/adk-go/blob/db550d64d321e82c835686e30cae3114cd7b6491/agent/remoteagent/v2/a2a_agent.go#L101),
  "Any other scheme is rejected".
- **The swallowed parse error at
  [L84](https://github.com/google/adk-go/blob/db550d64d321e82c835686e30cae3114cd7b6491/agent/remoteagent/v2/a2a_agent.go#L84)**
  means a scheme containing a tab or a space (`file\t:/etc/hosts`) falls through
  as a path. Nothing is read, since the open fails, but it classifies
  inconsistently with the neighbouring cases.
- **`http:card.json` and a bare `https:` now reach the network** where they were
  local reads before, because the `//` requirement went away.
- **`errUnsupportedCardSource` at
  [L66](https://github.com/google/adk-go/blob/db550d64d321e82c835686e30cae3114cd7b6491/agent/remoteagent/v2/a2a_agent.go#L66)
  is unexported**, so a caller outside the package can't `errors.Is` a new
  permanent failure mode on public API.
- **Test gaps:** the table has no two-character scheme and no source with two
  colons, so widening to `len(scheme) <= 2` or matching on the last colon instead
  of the first would both pass it unchanged.

Line references are against `db550d64d321e82c835686e30cae3114cd7b6491`.

agent/remoteagent/v2/a2a_agent.go
kdroste-google, line 116: I think we should not have os.ReadFile here. Can we switch to Root.FS instead?

kdroste-google (changes requested):

kdroste-google (approved):
LGTM, thank you!

PR #1331, opened 2026-08-13
docs(agents): add .agents/skills and a skill for authoring examples

wolo-lab:
## Problem

Writing an example for this repo means reverse-engineering the conventions from
a dozen READMEs under `examples/workflow` and `examples/agentregistry`: the
section order, the `Concept` / `Needs LLM?` block, the diagram, the environment
table, the transcript. The conventions are consistent, but they live only in the
examples themselves, so each new one rediscovers them — and misses whatever no
single sample makes visible.

The parts that get missed are the ones that cost a review round. Recent reviews
of the `agentregistry` samples turned up a model alias (`gemini-flash-latest`)
in a sample whose README documents the Vertex AI path, which Vertex does not
serve; a transcript whose framing promised more reproducibility than the output
had; and a flow chart that hid the fact that resolution happens once, before the
agent exists. None of that is discoverable from the API docs.

adk-python already has a home for this kind of knowledge — `.agents/skills/`,
currently 11 skills, including `adk-sample-creator`. adk-go has no equivalent,
so there is nowhere to write any of it down.

## Summary

- Adds `.agents/skills/`, the same location adk-python uses, so compatible agent
  tools pick skills up without extra configuration.
- First skill, `adk-sample-creator`: directory layout and the parent index row,
  `main.go` anatomy, the README shape, the transcript rules, and the checks to
  run before opening a PR.
- Templates ship as files (`templates/main.go.tmpl`, `templates/README.md.tmpl`)
  rather than fenced blocks, so an example starts as a copy instead of a
  transcription and `SKILL.md` stays short enough to read end to end.
- The README template defaults to a `sequenceDiagram` — a sample is a sequence
  of calls, and ordering is usually the point — and has a `Requirements` section
  covering enabled APIs, IAM roles and credentials, not just authentication.
- Existing READMEs keep their `Authentication` and `Workflow` headings; the new
  names apply to what gets written from now on.
- `CONTRIBUTING.md` describes the folder and lists what is in it; `AGENTS.md`
  links there and keeps the one behavioural line, the way it already defers to
  `CONTRIBUTING.md` for multi-module policy and the PR process.

Depends on #1266, which adds `examples/agentregistry/`. The skill cites those
samples as the standard for an example that talks to a live service, so that one
should land first.


.agents/skills/adk-sample-creator/SKILL.md
```diff
@@ -0,0 +1,139 @@
...
+```
+examples/<area>/<name>/
+├── main.go
+└── README.md
+```
+
+`lower_snake` directory names (`hitl_simple`), nested by theme once a family
+grows (`routing/llm`, `dynamic/hitl`). When the area keeps an index table —
+`examples/workflow/README.md` and `examples/agentregistry/README.md` do — add a
+row with the same columns as its neighbours, including `LLM?`. Most areas keep
```
kdroste-google, line 35: I don't understand this rows, columns and "including 'LLM?'". Could you elaborate?

kdroste-google (changes requested):

.agents/skills/adk-sample-creator/SKILL.md
wolo-lab: "including LLM" means "token will be consumed" so the user will pay for running this sample and results might depend on the model used.

kdroste-google (approved):
LGTM, thank you

baptmont:
Removing `v1-needed`. The backport attempt is closed in #1532.

The skill files apply cleanly to `v1`, but the skill describes `main`'s tree and four of its instructions do not hold there: `examples/workflow/` and `examples/agentregistry/` do not exist on `v1` (there is no `agentregistry` package at all), and `v1` is single-module so the workspace setup does not apply. Backporting it would tell an agent to model new examples on directories it cannot find.

Adapting it would mean maintaining a `v1`-specific fork of a doc that re-drifts whenever `main`'s examples change, on a branch that takes only fixes that must ship to 1.x. Not worth it for an example-authoring skill. If `v1` ever wants agent guidance, it wants writing against its own tree.

PR #1332, opened 2026-08-13
fix(artifact): reject path separators in appName/userID/sessionID

wojtas000:
## Problem
Path to the GCS object is built by concatenating fields appName/userID/sessionID/fileName/version to one string: `fmt.Sprintf("%s/%s/%s/%s/%d", appName, userID, sessionID, fileName, version)`. Because the key is a single flat string, a `/` in any segment moves the boundaries. Example:
For GCS artifact service, if one app has appName = `app/u1`, userID = `u2` and another has appName = `app`, userID = `u1/u2`, then for the same sessionID both resolve to the same session prefix: `app/u1/u2/<sessionID>`.
Identical prefixes allow artifacts to leak across apps/users (read, list, overwrite, delete).

Examples:
- one artifact overwritten with another one coming from different app / user (so e.g. latest version of the artifact is artifact from another app / user) 
- session which can see artifacts from different apps

Important caveat: REST router in the adk server can't feed a decoded "/" into a path segment, but one could use the Gcs Artifact Service independently (e.g. as an agent tool) where it would be doable.

## Solution
Add segment validation (for appName, userID, sessionID) similarily to existing `validateFileName`.

## Testing plan
- `golangci-lint run` - 0 issues
- `go build -mod=readonly work` passed
- `go test -race -mod=readonly -count=1 -shuffle=on work` green

## Potential improvements
- refactor: it could be possible to merge `validatePathSegments` with already existing `validateFileName` into one function
- all `Validate` methods accross `artifact/service.go` have very similar structure, and we might introduce helper function which could be reused across all `Validate` implementations to not duplicate redundant code

### DIFF
--- a/artifact/request_validation_test.go
+++ b/artifact/request_validation_test.go
@@ -122,6 +122,42 @@ func TestSaveRequest_Validate(t *testing.T) {
 			wantErr:    true,
 			wantErrMsg: "invalid name: filename cannot contain path separators",
 		},
+		{
+			name: "AppName with path separator",
+			req: &SaveRequest{
+				AppName:   "My/App",
+				UserID:    "user-123",
+				SessionID: "sess-abc",
+				FileName:  "file.txt",
+				Part:      genai.NewPartFromText("data"),
+			},
+			wantErr:    true,
+			wantErrMsg: "invalid AppName: cannot contain path separators",
+		},
+		{
+			name: "UserID with path separator",
+			req: &SaveRequest{
+				AppName:   "MyApp",
+				UserID:    "user/123",
+				SessionID: "sess-abc",
+				FileName:  "file.txt",
+				Part:      genai.NewPartFromText("data"),
+			},
+			wantErr:    true,
+			wantErrMsg: "invalid UserID: cannot contain path separators",
+		},
+		{
+			name: "SessionID with backslash separator",
+			req: &SaveRequest{
+				AppName:   "MyApp",
+				UserID:    "user-123",
+				SessionID: `sess\abc`,
+				FileName:  "file.txt",
+				Part:      genai.NewPartFromText("data"),
+			},
+			wantErr:    true,
+			wantErrMsg: "invalid SessionID: cannot contain path separators",
+		},
 	}
 	executeValidatorTestCases(t, "SaveRequest", testCases)
 }
@@ -174,6 +210,39 @@ func TestLoadRequest_Validate(t *testing.T) {
 			},
 			wantErr:    true,
 			wantErrMsg: "invalid name: filename cannot contain path separators",
+		},
+		{
+			name: "AppName with path separator",
+			req: &LoadRequest{
+				AppName:   "My/App",
+				UserID:    "user-123",
+				SessionID: "sess-abc",
+				FileName:  "file.txt",
+			},
+			wantErr:    true,
+			wantErrMsg: "invalid AppName: cannot contain path separators",
+		},
+		{
+			name: "UserID with path separator",
+			req: &LoadRequest{
+				AppName:   "MyApp",
+				UserID:    "user/123",
+				SessionID: "sess-abc",
+				FileName:  "file.txt",
+			},
+			wantErr:    true,
+			wantErrMsg: "invalid UserID: cannot contain path separators",
+		},
+		{
+			name: "SessionID with backslash separator",
+			req: &LoadRequest{
+				AppName:   "MyApp",
+				UserID:    "user-123",
+				SessionID: `sess\abc`,
+				FileName:  "file.txt",
+			},
+			wantErr:    true,
+			wantErrMsg: "invalid SessionID: cannot contain path separators",
 		},
 	}
 	executeValidatorTestCases(t, "LoadRequest", testCases)
@@ -228,6 +297,39 @@ func TestDeleteRequest_Validate(t *testing.T) {
 			wantErr:    true,
 			wantErrMsg: "invalid name: filename cannot contain path separators",
 		},
+		{
+			name: "AppName with path separator",
+			req: &DeleteRequest{
+				AppName:   "My/App",
+				UserID:    "user-123",
+				SessionID: "sess-abc",
+				FileName:  "file.txt",
+			},
+			wantErr:    true,
+			wantErrMsg: "invalid AppName: cannot contain path separators",
+		},
+		{
+			name: "UserID with path separator",
+			req: &DeleteRequest{
+				AppName:   "MyApp",
+				UserID:    "user/123",
+				SessionID: "sess-abc",
+				FileName:  "file.txt",
+			},
+			wantErr:    true,
+			wantErrMsg: "invalid UserID: cannot contain path separators",
+		},
+		{
+			name: "SessionID with backslash separator",
+			req: &DeleteRequest{
+				AppName:   "MyApp",
+				UserID:    "user-123",
+				SessionID: `sess\abc`,
+				FileName:  "file.txt",
+			},
+			wantErr:    true,
+			wantErrMsg: "invalid SessionID: cannot contain path separators",
+		},
 	}
 	executeValidatorTestCases(t, "DeleteRequest", testCases)
 }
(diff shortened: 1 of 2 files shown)


baptmont (approved):

karolpiotrowicz (changes requested):
The collision is real and reproduces: with `appName="app/u1", userID="u2"` and
`appName="app", userID="u1/u2"`, both tuples resolve to the prefix
`app/u1/u2/<sessionID>/`, and before this change both requests validate. Guarding
the segments is the right call, and rejecting at the boundary is a better answer
than escaping the key, which would silently orphan every stored object whose
segments contain a space or a non-ASCII character.

Two things need to change before this can land, both consequences of where the
check is applied rather than of the check itself.

**A Vertex AI ReasoningEngine resource name is a supported `AppName`, and it is
now rejected.** `getReasoningEngineID` accepts the full resource name, and its
own error text names that form as valid
([vertexai_client.go#L550](https://github.com/google/adk-go/blob/d601e47f0d0dc981e4b45a0111a9dff6eb4bbde0/session/vertexai/vertexai_client.go#L550),
[#L569](https://github.com/google/adk-go/blob/d601e47f0d0dc981e4b45a0111a9dff6eb4bbde0/session/vertexai/vertexai_client.go#L569)).
It is asserted valid in-tree with `ReasoningEngine` unset
([vertexai_test.go#L194-L199](https://github.com/google/adk-go/blob/d601e47f0d0dc981e4b45a0111a9dff6eb4bbde0/session/vertexai/vertexai_test.go#L194-L199)).
The runner copies the session's app name straight into every artifact request
([runner.go#L260-L265](https://github.com/google/adk-go/blob/d601e47f0d0dc981e4b45a0111a9dff6eb4bbde0/runner/runner.go#L260-L265)
→ [artifacts.go#L34-L41](https://github.com/google/adk-go/blob/d601e47f0d0dc981e4b45a0111a9dff6eb4bbde0/internal/artifact/artifacts.go#L34-L41)),
so after this change every Save, Load, List, Delete, Versions and
GetArtifactVersion fails for a deployment configured that way:

```go
a := &artifact.SaveRequest{
    AppName:   "projects/my-proj/locations/us-central1/reasoningEngines/1234567890",
    UserID:    "u1", SessionID: "s1", FileName: "f.txt",
    Part:      genai.NewPartFromText("data"),
}
a.Validate() // before: nil.  after: invalid AppName: cannot contain path separators
```

The in-tree examples set `ReasoningEngine` explicitly and the Agent Engine server
passes the bare numeric ID, so nothing shipped in the repo hits this — but the
resource-name form is supported and tested, and the two subsystems now disagree
about what an `AppName` is. Either narrow the constraint, normalise the resource
name to the engine ID before it reaches the artifact layer, or drop support for
that form deliberately and update `session/vertexai` alongside.

**Artifacts already in a bucket become unreadable, and also undeletable.** The
guard runs on the read and cleanup paths too, not only on Save — see the
`DeleteRequest` call site at
[service.go#L219-L222](https://github.com/google/adk-go/blob/d601e47f0d0dc981e4b45a0111a9dff6eb4bbde0/artifact/service.go#L219-L222).
An object written by an earlier binary at `app/u1/u2/s1/f.txt/1` loads and deletes
cleanly today and returns `invalid UserID: cannot contain path separators` after
this change, so an operator cannot drain or clean up the very data the fix is
meant to address without dropping to `gsutil`. Worth deciding explicitly whether
Save-only rejection for a release, with reads and deletes still permitted, is the
better shape.

The backslash half of the guard
([service.go#L137](https://github.com/google/adk-go/blob/d601e47f0d0dc981e4b45a0111a9dff6eb4bbde0/artifact/service.go#L137))
sharpens that: `\` is an ordinary character in a GCS object name and never a key
separator, so it cannot produce the collision, but banning it strands
already-stored artifacts for any deployment whose user IDs look like
`DOMAIN\user`. If it is there to protect a future filesystem-backed
implementation, that is a reasonable argument to make in the comment.

- **The constraint is not documented anywhere a caller would look.** The only
  description of it sits on the unexported helper
  ([service.go#L129-L142](https://github.com/google/adk-go/blob/d601e47f0d0dc981e4b45a0111a9dff6eb4bbde0/artifact/service.go#L129-L142)).
  The exported `AppName, UserID, SessionID` fields on the six request structs, the
  package doc, and `runner.Config.AppName`
  ([runner.go#L46](https://github.com/google/adk-go/blob/d601e47f0d0dc981e4b45a0111a9dff6eb4bbde0/runner/runner.go#L46))
  all say nothing. Since the signatures are unchanged, documentation and a release
  note are the only warning users get.
- **The description's note about the REST router does not hold for `/run`.**
  `appName`, `userId` and `sessionId` arrive in the JSON body there, not as path
  segments
  ([runtime.go#L23-L29](https://github.com/google/adk-go/blob/d601e47f0d0dc981e4b45a0111a9dff6eb4bbde0/server/adkrest/internal/models/runtime.go#L23-L29)),
  so a `/` does reach them. The practical effect is that the session is created and
  the run starts before the first artifact call fails, which is worth a sentence in
  the description and may argue for validating the tuple at session creation.
- **The new table covers one diagonal of the matrix.** `AppName` and `UserID` are
  tested with `/` only, and `SessionID` with `\` only, so a validator that skipped
  the `/` check on `SessionID` would still pass. Driving the cases from a
  `for _, sep := range []string{"/", "\\"}` loop over all three fields closes it,
  and a case asserting rejection through a `Service` method would pin that the
  guard is actually reached.
- **`validatePathSegments` accepts empty values**, which is safe only because
  every caller runs the required-field check first. Worth a line in the doc comment
  if it stays that way.

Two behaviours adjacent to this change are worth knowing about, though neither is
introduced here and neither should hold up the PR: a session literally named
`user` produces the same listing prefix as the user scope, and `List` and
`Versions` do not check segment depth, so objects already written with a
separator still surface under a valid tuple. Happy to open separate issues for
both if that is useful.

wojtas000:
## Proposal: don't land the strict segment rejection — adk-python already tried and reverted it

After looking at how adk-python handled the same finding, I think rejecting `/` in `userID`/`sessionID` (and `appName`) is the wrong fix and we shouldn't merge it as-is. Evidence and an alternative below.

### What this PR does
Rejects path separators (`/`, `\`) in the artifact-key identity segments so that distinct identities can't collide to the same flat GCS object key (`appName/userID/sessionID/fileName/version`).

### adk-python tried exactly this and rolled it back
Public commit history in `google/adk-python`:

- **Introduced** strict validation rejecting `/` in app/user/session: [`8718aef`](https://github.com/google/adk-python/commit/8718aefff01d3ad37a4f9de73ca6abfdb3bb452e) (closed [google/adk-python#6115](https://github.com/google/adk-python/issues/6115) and community PR [google/adk-python#6116](https://github.com/google/adk-python/pull/6116)).
- **Rolled back the same day:** [`961f3e8`](https://github.com/google/adk-python/commit/961f3e885c68385d67823885897f1c5690ff5b4c).
- **Re-landed, deliberately permitting interior slashes:** [`45a77dc`](https://github.com/google/adk-python/commit/45a77dc514f1e151c08e32f7c9a94b0a222eea36) — the current `validate_path_segment` only rejects empty values, NUL bytes, a leading `/` or `\`, drive prefixes, and `..` traversal segments. Interior slashes are allowed on purpose.

The rollback happened because rejecting interior slashes broke deployments that pass IDs which legitimately contain slashes (namespaced user IDs). So this is a **confirmed breaking change**, not a hypothetical one — and adk-python, which we treat as the source of truth for cross-SDK behavior, deliberately settled on allowing interior slashes. Landing the strict rule in Go would break the same class of users and diverge from Python.

### This finding is really two separate defects
1. **Version-scan conflation** — when a segment can contain `/`, listing versions of artifact `foo` also picks up the distinct nested artifact `foo/bar`. Python fixed this in [`08d21cc`](https://github.com/google/adk-python/commit/08d21cc35569939d033fdac296e349be7c5d2268) (a `_parse_version` helper that only counts a blob whose suffix after the prefix is a single numeric segment). This is the facet that actually closed the canonical finding.
2. **Segment collision / prefix shift** — `userID="a/b" + sessionID="c"` vs `userID="a" + sessionID="b/c"` resolve to the same prefix. This is what this PR targets, and it's the one that can't be fixed by rejection without the breakage above.

(In GCS, `..` traversal is a non-issue — it's a flat namespace, so `app/../x` is a literal object, not `x`. Traversal guards mostly matter for filesystem-backed stores, which Go doesn't have.)

### Recommendation
1. **Close this PR** — don't land the strict rejection (confirmed breaking, and it diverges from adk-python).
2. **Treat the collision as a known, cross-SDK-consistent limitation** and document it. Real isolation for slash-in-ID deployments has to come from the deployment (per-tenant buckets, or IAM prefix conditions) plus callers keeping IDs collision-free — a flat key can't guarantee isolation while permitting `/` in the segments.
3. **Version-scan conflation** - Python's version-scan guard ([`08d21cc`](https://github.com/google/adk-python/commit/08d21cc35569939d033fdac296e349be7c5d2268)) does not need to be implemented in ADK Go as we already forbid `/` in `fileName` ([`validateFileName`](https://github.com/google/adk-go/blob/2325d1cf4f0c1e097f13403d84493cea23660f03/artifact/service.go#L118-L123)), so the filename-nesting version conflation can't occur today

Would appreciate thoughts on whether we want strict cross-SDK parity (allow interior slashes, match Python) vs. a stricter Go-only stance before we close.

karolpiotrowicz:
Let's close here and align on the shared approach between ADK's.

PR #1336, opened 2026-08-15
fix(llminternal): read streaming mode from RunConfig to avoid nil deref

rootkiller6788:

- Closes: #_issue_number_
- Related: #_issue_number_

possible._

**Problem:**
_A clear and concise description of what the problem is._

**Solution:**
_A clear and concise description of what you want to happen and why you choose
this solution._

### Testing Plan

**Unit Tests:**

- [ ] I have added or updated unit tests for my change.

### Checklist

- [ ] I have read the [CONTRIBUTING.md](./CONTRIBUTING.md) document.
- [ ] I have performed a self-review of my own code.
- [ ] New and existing unit tests pass locally with my changes.

### Additional context


baptmont:
The fix is right, and the guard is load-bearing rather than defensive padding. I reverted `base_flow.go` while keeping your new test and got exactly the panic you describe:

```
--- FAIL: TestCallLLMStreamingModeWithoutContextRunConfig
```

`internal/llminternal` and `runner` are green with the change.

One thing I wanted to confirm before signing off, and it is worth recording here: `ctx.RunConfig()` and `runconfig.FromContext(ctx)` are **not** the same object, or even the same type.

```go
// internal/agent/runconfig
type RunConfig struct {
    StreamingMode StreamingMode        // "none" | "sse" | "bidi"
    Live          *agent.LiveRunConfig
}

// agent
type RunConfig struct {
    StreamingMode StreamingMode        // "none" | "sse"
    SaveInputBlobsAsArtifacts bool
}
```

They also differ in how they propagate: `runconfig.FromContext` reads a Go `context.Value`, which every child context inherits automatically, whereas `ctx.RunConfig()` reads a field that has to be copied by hand at each `NewInvocationContext` call. So the substitution is only safe if every construction site carries it. I checked all seven:

| site | RunConfig |
| --- | --- |
| `runner.go:281` (Run) | `&cfg` |
| `run_node.go:240` | `&cfg` |
| `llmagent.go:471` | `ctx.RunConfig()` |
| `llm_agent_wrapper.go:110` | `ctx.RunConfig()` |
| `parallelagent/agent.go:89` | `ctx.RunConfig()` |
| `workflow/agent_node.go:104` | `ctx.RunConfig()` |
| `runner.go:472` (RunLive) | **not set → nil** |

In the normal path the runner fills both from the same variable a few lines apart — `runconfig.StreamingMode(cfg.StreamingMode)` at `runner.go:257` and `RunConfig: &cfg` at `runner.go:287` — so `StreamingMode` is literally the same string on both sides and the two expressions agree.

`RunLive` is the only divergence, and it happens to be benign for this comparison: it sets the context copy to a hardcoded `Bidi` and leaves the invocation context's `RunConfig` nil, so the old code compared `"bidi" != "sse"` and the new code hits your nil guard. Both yield `false`.

Worth noting the enums are not interchangeable in general: `runconfig.StreamingMode` has a third value (`Bidi`) that `agent.StreamingMode` does not. The swap works because you are testing for `SSE` specifically.

Two description-only asks before this merges:

1. The PR body is still the unfilled template — no problem statement, no testing plan, and `Closes: #_issue_number_`. The code comment cites issue #586; please link it properly so the fix is traceable.
2. Please add a sentence covering the `RunLive` case. It is the obvious "does this change behavior?" question, and answering it in the description saves the next reader the walk through all seven call sites.

Nothing blocking in the code itself.

rootkiller6788:
Closes: #586

  Problem

  runconfig.FromContext(ctx) returns nil when no RunConfig is stored in the Go context — which happens whenever
  agent.Run() is invoked directly (bypassing the runner). internal/llminternal/base_flow.go then dereferences it
  unconditionally:

  causing a nil pointer dereference on the first LLM call. The runner (runner/runner.go, runner/run_node.go) is the only
  place that calls runconfig.ToContext, so any direct agent.Run() caller panics.

  Solution

  Read the streaming mode from the invocation context's RunConfig() instead of the runconfig context value, guarded
  against nil:

  useStream := false
  if rc := ctx.RunConfig(); rc != nil {
      useStream = rc.StreamingMode == agent.StreamingModeSSE
  }

  RunConfig() is populated on every construction path (runner.go, run_node.go, and the
  agent/wrapper/parallelagent/workflow sites), so it is the canonical source. This also matches the existing pattern in
  agent/remoteagent/v2/a2a_agent.go.

  Note on RunLive: internal/agent/runconfig.RunConfig and agent.RunConfig are distinct types — the former carries a
  third StreamingModeBidi value and a Live field. The live path (runner.go:472) sets only the context copy (to Bidi) and
  leaves the invocation context's RunConfig nil. That divergence is benign here: the old expression evaluated "bidi" !=
  "sse" → false, and the new one falls through the nil guard to false. Both yield non-streaming, and RunLive does not
  reach callLLM anyway.

  Testing Plan

  Unit tests:
  - Added TestCallLLMStreamingModeWithoutContextRunConfig, which constructs an invocation context with no runconfig
  context value (simulating a direct agent.Run() call) but an SSE RunConfig, and asserts the model receives stream=true
  without panicking. Reverting the fix reproduces the exact nil-pointer panic.
  - go test ./internal/llminternal/ passes locally.

baptmont:
Thanks — merging this. The fix is correct and I verified it: reverting `base_flow.go` while keeping your test reproduces the exact nil deref, and `internal/llminternal` and `runner` are green with the change.

Two corrections to the write-up, so the record is right for whoever reads this next.

**`RunConfig()` is not populated on every construction path.** `runner.go:472` (RunLive) leaves it out:

```go
iCtx := icontext.NewInvocationContext(ctx, icontext.InvocationContextParams{
    Artifacts: artifacts, Memory: memoryImpl,
    Session: storedSession, Agent: agentToRun, UserContent: nil,
})
```

So `ctx.RunConfig()` is nil there by construction. That does not weaken the change — it is precisely why your `if rc != nil` guard is needed, and the behaviour is unchanged in that path either way (the old code compared `"bidi" != "sse"`, the new code takes the nil branch, both yield `false`). But the sentence as written implies the guard is redundant, which is the opposite of the truth and is the kind of thing that invites someone to "simplify" it later.

**`a2a_agent.go` is not a precedent for this pattern — it is the same bug.** Line 292 is an unguarded dereference:

```go
```

A repo-wide sweep finds exactly one unguarded `ctx.RunConfig().` deref and that is it, reachable the same way #586 is: `agent.Run()` invoked directly, or the RunLive path above. I will file it as a follow-up; it should get the same treatment as this one.

Two follow-ups, neither blocking:

1. `agent/remoteagent/v2/a2a_agent.go:292` — same crash, still open.
2. A test for the guard itself. I removed the `if rc != nil` and the full suite still passed, so nothing currently pins it — and the branch it protects is the live path, where the config is nil by construction. A small table over `{SSE → true, None → false, nil → false}` would cover it.

baptmont (approved):

PR #1351, opened 2026-08-17
fix(cmd/launcher/web): initialize telemetry before starting HTTP server (#1350)

### LINKED ISSUE #1350: cmd/launcher/web: HTTP server listener and goroutine leak if telemetry initialization fails
## 🔴 Required Information

**Describe the Bug:**
In `cmd/launcher/web/web.go:188-199`, `webLauncher.Run` starts `srv.ListenAndServe()` in a background goroutine before calling `telemetry.InitAndSetGlobalOtelProviders`.
If telemetry initialization fails (e.g. invalid cloud credentials, bad config, or exporter error), `Run` returns immediately with an error without shutting down or closing `srv`.
This leaves an orphan TCP socket actively listening on the target port and a leaked background goroutine.

**Steps to Reproduce:**
1. Run the web launcher with telemetry configured to fail (e.g. `--cloud-telemetry` without Google Cloud credentials or with an invalid telemetry exporter).
2. `webLauncher.Run` returns an error (`telemetry initialization failed: ...`).
3. Notice that the TCP port remains bound and the HTTP server goroutine continues running in the background.

**Expected Behavior:**
Telemetry is initialized before starting the network listener, or any active HTTP server is cleanly closed when telemetry setup fails.

**Observed Behavior:**
The HTTP server continues running in the background while `Run` has returned an error, leaking the listener socket and goroutine.

**Environment Details:**
- ADK Library Version: v2.x (main at 3a474df)
- OS: Linux
- Go Version: 1.26.6

**Model Information:**
- N/A

---

## 🟡 Optional Information

- Always (100%) when telemetry initialization fails.

chill-czar:

- Closes: #1350

### Problem

In `webLauncher.Run`, `srv.ListenAndServe()` was previously spawned in a background goroutine prior to initializing telemetry providers. If telemetry setup failed, `Run` returned immediately without shutting down `srv`, leaking the listening socket and background goroutine.

### Solution

Reorder `telemetry.InitAndSetGlobalOtelProviders` to run before starting the `srv.ListenAndServe()` background goroutine, ensuring that no socket is bound or background goroutine spawned if telemetry initialization fails.

### Testing Plan

- [x] Tested web launcher and subrouters locally.
- [x] All repository verification commands pass:
  ```text
  go test -v -race -count=1 ./cmd/launcher/web/...
  go build -mod=readonly work
  golangci-lint run
  go mod tidy -diff
  ```

### Checklist

### DIFF
--- a/cmd/launcher/web/web.go
+++ b/cmd/launcher/web/web.go
@@ -186,6 +186,11 @@ func (w *webLauncher) Run(ctx context.Context, config *launcher.Config) error {
 	}
 	log.Println()
 
+	telemetryService, err := telemetry.InitAndSetGlobalOtelProviders(ctx, config, w.config.otelToCloud)
+	if err != nil {
+		return fmt.Errorf("telemetry initialization failed: %v", err)
+	}
+
 	srv := w.buildHTTPServer(router)
 
 	errChan := make(chan error, 1)
@@ -195,11 +200,6 @@ func (w *webLauncher) Run(ctx context.Context, config *launcher.Config) error {
 		}
 		close(errChan)
 	}()
-
-	telemetryService, err := telemetry.InitAndSetGlobalOtelProviders(ctx, config, w.config.otelToCloud)
-	if err != nil {
-		return fmt.Errorf("telemetry initialization failed: %v", err)
-	}
 
 	select {
 	case <-ctx.Done():
--- a/cmd/launcher/web/web.go
+++ b/cmd/launcher/web/web.go
@@ -183,6 +183,11 @@ func (w *webLauncher) Run(ctx context.Context, config *launcher.Config) error {
 	}
 	log.Println()
 
+	telemetryService, err := telemetry.InitAndSetGlobalOtelProviders(ctx, config, w.config.otelToCloud)
+	if err != nil {
+		return fmt.Errorf("telemetry initialization failed: %v", err)
+	}
+
 	srv := w.buildHTTPServer(router)
 
 	errChan := make(chan error, 1)
@@ -192,11 +197,6 @@ func (w *webLauncher) Run(ctx context.Context, config *launcher.Config) error {
 		}
 		close(errChan)
 	}()
-
-	telemetryService, err := telemetry.InitAndSetGlobalOtelProviders(ctx, config, w.config.otelToCloud)
-	if err != nil {
-		return fmt.Errorf("telemetry initialization failed: %v", err)
-	}
 
 	select {
 	case <-ctx.Done():
--- a/cmd/launcher/web/web_test.go
+++ b/cmd/launcher/web/web_test.go
@@ -15,10 +15,19 @@
 package web
 
 import (
+	"context"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
 	"testing"
+	"time"
+
+	"github.com/gorilla/mux"
+	"go.opentelemetry.io/otel/sdk/resource"
+
+	"google.golang.org/adk/v2/cmd/launcher"
+	"google.golang.org/adk/v2/telemetry"
 )
 
 func TestH2CFlag(t *testing.T) {
@@ -118,5 +127,57 @@ func assertProtocol(t *testing.T, client *http.Client, url string, wantMajor int
 	}
 	if resp.ProtoMajor != wantMajor {
 		t.Errorf("response protocol = %q, want HTTP/%d", resp.Proto, wantMajor)
+	}
+}
+
+type telemetryFailSublauncher struct{}
+
+func (telemetryFailSublauncher) Keyword() string { return "repro" }
+
+func (telemetryFailSublauncher) Parse(args []string) ([]string, error)             { return args, nil }
+func (telemetryFailSublauncher) CommandLineSyntax() string                         { return "" }
+func (telemetryFailSublauncher) SimpleDescription() string                         { return "" }
+func (telemetryFailSublauncher) UserMessage(webURL string, printer func(v ...any)) {}
+func (telemetryFailSublauncher) SetupSubrouters(r *mux.Router, c *launcher.Config) error {
+	return nil
+}
+
+// TestRunDoesNotLeakListenerWhenTelemetryInitFails covers issue #1350: when
+// telemetry initialization fails, Run must not leave an HTTP listener bound.
+func TestRunDoesNotLeakListenerWhenTelemetryInitFails(t *testing.T) {
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatalf("net.Listen() failed: %v", err)
+	}
+	port := ln.Addr().(*net.TCPAddr).Port
+	if err := ln.Close(); err != nil {
+		t.Fatalf("listener Close() failed: %v", err)
+	}
+
+	l := NewLauncher(telemetryFailSublauncher{}).(*webLauncher)
+	if _, err := l.Parse([]string{"--port", fmt.Sprint(port), "repro"}); err != nil {
+		t.Fatalf("Parse() failed: %v", err)
+	}
+
+	// A resource whose schema URL conflicts with resource.Default()'s makes
+	// resource.Merge fail, so telemetry init fails without touching the network.
+	bad := resource.NewWithAttributes("https://conflicting.invalid/schema/v1")
+	config := &launcher.Config{
+		TelemetryOptions: []telemetry.Option{telemetry.WithResource(bad)},
+	}
+
+	if err := l.Run(context.Background(), config); err == nil {
+		t.Fatalf("Run() succeeded, want telemetry initialization failure")
+	}
+
+	addr := fmt.Sprintf("127.0.0.1:%d", port)
+	deadline := time.Now().Add(3 * time.Second)
+	for time.Now().Before(deadline) {
+		conn, err := net.DialTimeout("tcp", addr, 100*time.Millisecond)
+		if err == nil {
+			_ = conn.Close()
+			t.Fatalf("port %d still bound after Run() returned an error: listener leaked", port)
+		}
+		time.Sleep(10 * time.Millisecond)
 	}
 }


chill-czar:
Hi @indurireddy-TF, could you please review this PR when you have a chance? It ensures telemetry is initialized prior to starting the background HTTP server goroutine in `webLauncher.Run` (closes #1350). All unit and race tests pass locally. Thank you!

indurireddy-TF:
Hi @chill-czar Thanks for raising this PR I verified the proposed fix, and it is working as expected. The reproduction test now passes, and all existing test cases are passing as well.Could you please take a look when you get a chance @baptmont ?

karolpiotrowicz (commented):
The reordering is correct and I was able to confirm it fixes the reported leak. Driving the real `webLauncher.Run` with a telemetry configuration that fails deterministically, the listener is still bound after `Run` returns an error on `main`, and is never bound with this PR applied. Build, `go vet`, `staticcheck`, `golangci-lint`, `govulncheck`, `-race` and `apidiff` are all clean on the merge result, and the two failing packages in a full `go test ./...` fail identically on `main`, so neither is yours.

One thing worth adding before this merges, one thing worth knowing about the new ordering, and a couple of notes that are not yours to fix.

## The fix is unguarded

The branch this PR creates — [`web.go` L186-L189, PR head 2994462](https://github.com/google/adk-go/blob/29944624df36a4a353364e77d0761345ee2d5f76/cmd/launcher/web/web.go#L186-L189) — is executed by no test in the repository, so reverting the reordering leaves the whole suite green and the bug comes back silently. `CONTRIBUTING.md` asks for a test with each change ([`CONTRIBUTING.md` L146-L151, main at 160a397](https://github.com/google/adk-go/blob/160a39763e901047505abcc1acc32707edc0fab5/CONTRIBUTING.md#L146-L151)), and the PR checklist has "I have added tests that prove my fix is effective" ticked, but the diff is `cmd/launcher/web/web.go` alone.

A test is straightforward to write offline, which was my main worry before trying it. Giving `telemetry.WithResource` a resource whose schema URL conflicts with the one on `resource.Default()` makes `resource.Merge` fail inside telemetry setup, with no network involved. I ran exactly the code below against current `main` and against `main` with this PR applied: it fails three times out of three without the change and passes three out of three with it, and the rest of `./cmd/launcher/...` stays green.

<details><summary>Reproduction test (drop into <code>cmd/launcher/web/web_test.go</code>)</summary>

```go
type telemetryFailSublauncher struct{}

func (telemetryFailSublauncher) Keyword() string { return "repro" }

func (telemetryFailSublauncher) Parse(args []string) ([]string, error)             { return args, nil }
func (telemetryFailSublauncher) CommandLineSyntax() string                         { return "" }
func (telemetryFailSublauncher) SimpleDescription() string                         { return "" }
func (telemetryFailSublauncher) UserMessage(webURL string, printer func(v ...any)) {}
func (telemetryFailSublauncher) SetupSubrouters(r *mux.Router, c *launcher.Config) error {
	return nil
}

// TestRunDoesNotLeakListenerWhenTelemetryInitFails covers issue #1350: when
// telemetry initialization fails, Run must not leave an HTTP listener bound.
func TestRunDoesNotLeakListenerWhenTelemetryInitFails(t *testing.T) {
	ln, err := net.Listen("tcp", "127.0.0.1:0")
	if err != nil {
		t.Fatalf("net.Listen() failed: %v", err)
	}
	port := ln.Addr().(*net.TCPAddr).Port
	if err := ln.Close(); err != nil {
		t.Fatalf("listener Close() failed: %v", err)
	}

	l := NewLauncher(telemetryFailSublauncher{}).(*webLauncher)
	if _, err := l.Parse([]string{"--port", fmt.Sprint(port), "repro"}); err != nil {
		t.Fatalf("Parse() failed: %v", err)
	}

	// A resource whose schema URL conflicts with resource.Default()'s makes
	// resource.Merge fail, so telemetry init fails without touching the network.
	bad := resource.NewWithAttributes("https://conflicting.invalid/schema/v1")
	config := &launcher.Config{
		TelemetryOptions: []telemetry.Option{telemetry.WithResource(bad)},
	}

	if err := l.Run(context.Background(), config); err == nil {
		t.Fatalf("Run() succeeded, want telemetry initialization failure")
	}

	addr := fmt.Sprintf("127.0.0.1:%d", port)
	deadline := time.Now().Add(3 * time.Second)
	for time.Now().Before(deadline) {
		conn, err := net.DialTimeout("tcp", addr, 100*time.Millisecond)
		if err == nil {
			conn.Close()
			t.Fatalf("port %d still bound after Run() returned an error: listener leaked", port)
		}
		time.Sleep(10 * time.Millisecond)
	}
}
```

Run it with `go test ./cmd/launcher/web/ -run TestRunDoesNotLeakListenerWhenTelemetryInitFails`. On top of what `web_test.go` already imports it needs `context`, `fmt`, `time`, `github.com/gorilla/mux`, `go.opentelemetry.io/otel/sdk/resource`, `google.golang.org/adk/v2/cmd/launcher` and `google.golang.org/adk/v2/telemetry`.
</details>

## Telemetry init is now on the critical path to binding the port

This is not a change request, and I do not think it should hold the PR up, but it is worth recording since the ordering is the whole point of the change. Because init now runs before the listener, its duration is added to the time before anything is bound. With default flags that is unmeasurable. With `--otel_to_cloud` and a healthy host it is about five milliseconds. Where the GCE metadata endpoint is unreachable it is much longer: I measured 4.3 s when the endpoint refuses the connection, and about 40 s when packets are dropped, against roughly 2 ms on `main` in the same conditions.

The worst case is bounded at about 95 s, and I checked that it really is a ceiling rather than something that scales — a metadata stub delaying 8 s per request and one delaying 20 s both produced 95.1 s, so the limit is the resource detector's own budget. In every one of those slow cases telemetry initialization then failed, and since almost every example in the repository calls `log.Fatalf` on the launcher error, the process was going to die either way. What the old ordering added during that window was a listener serving traffic from a doomed process, which is the leak you are fixing, so I do not read this as a regression.

Two things I tried that do not work, so nobody else spends time on them. Wrapping the init in a `context.WithTimeout` does not bound it — the detector ignores the caller's deadline, and a two-second timeout still returned after 41 seconds. Binding the listener first and initializing telemetry afterwards is worse than either ordering: the socket accepts a TCP connection in 1.5 ms while HTTP requests hang until the client gives up, so it passes a readiness probe and serves nothing.

## Smaller notes, none blocking

`Run` still returns without shutting telemetry down when the server itself fails to start — [`web.go` L209-L213, PR head 2994462](https://github.com/google/adk-go/blob/29944624df36a4a353364e77d0761345ee2d5f76/cmd/launcher/web/web.go#L209-L213) returns on both `errChan` outcomes, and the only call to `telemetryService.Shutdown` sits in the `ctx.Done()` arm. I measured zero shutdown calls on a bind failure both with and without your change, so this predates it and is not yours to fix here. The console launcher already has the shape that would fix it, initializing telemetry and immediately deferring the shutdown at [`console.go` L75-L85, main at 160a397](https://github.com/google/adk-go/blob/160a39763e901047505abcc1acc32707edc0fab5/cmd/launcher/console/console.go#L75-L85). Worth a separate issue rather than growing this PR.

The branch is 59 commits behind `main` and will need a rebase before it can merge. `main` has changed this file in the meantime, adding a `config.Validate()` call to `Execute`, but nothing that touches `Run`, so the merge is clean today and the reordered function comes through unchanged.

chill-czar:
Thanks for the thorough review and for providing the reproduction test, @karolpiotrowicz!

1. **Regression test added**: Added `TestRunDoesNotLeakListenerWhenTelemetryInitFails` to `cmd/launcher/web/web_test.go` using the conflicting schema resource approach. Verified offline with `go test -race` and `golangci-lint`.
2. **Rebased**: Rebased onto latest `main`.
3. **Telemetry init on critical path**: Appreciate the deep dive into the metadata timeout behaviors and the confirmation that failing before binding the listener prevents serving traffic from an unhealthy process.
4. **Deferred telemetry shutdown**: Agreed that handling deferred telemetry shutdown on server startup failure should be addressed in a follow-up issue rather than expanding this PR. I will open an issue tracking this once this PR is merged.

jjsasha63 (approved):
This looks right to me and I would merge it as is.

I checked the ordering on main. Nothing between the two points depends on
it: buildHTTPServer only reads w.config, and the router is built earlier at
web.go:160.

One thing the reorder introduces. The "Web servers starts on %s" log at
web.go:183 now prints before telemetry.New runs, so the gap between showing
the user a URL and actually binding the port is wider than it was. Moving
those log lines below the telemetry call closes it.

Your checklist ticks "I have added tests that prove my fix is effective" and
the diff has none. If you want to add one, config.TelemetryOptions is not the
seam: telemetry.Option is a sealed interface and every exported With* returns
nil, so no option can make telemetry.New fail. web_test.go is in package web
and already does NewLauncher().(*webLauncher), so set the field directly:

    t.Setenv("GOOGLE_APPLICATION_CREDENTIALS", "/nonexistent/creds.json")
    w.config.otelToCloud = true

That fails in configure() on every run, with no network and no ADC. Then call
Run, expect the error, and assert the port is free. Worth having, but I would
not hold the fix for it.

Separately, and pre-existing rather than yours: on the
`case err, ok := <-errChan` path, telemetryService.Shutdown is never called,
so a server failure leaks the OTel providers the same way. That one belongs in
its own issue.

chill-czar:
Thanks for the review and approval, @jjsasha63!

Just to clarify on the test: commit `33f45c5` already added `TestRunDoesNotLeakListenerWhenTelemetryInitFails` to `cmd/launcher/web/web_test.go`. It uses `telemetry.WithResource(bad)` with a conflicting schema URL, which causes `resource.Merge` inside `setup_otel.go` to fail deterministically offline without touching credentials or the network.

To keep this approved PR clean and ready to merge without resetting approvals, I won't push further changes to this branch. I will open a separate follow-up issue/PR right after this merges to:
1. Move the startup log below `telemetry.InitAndSetGlobalOtelProviders`.
2. Defer `telemetryService.Shutdown` on the server startup/bind failure path (matching `console.go`).

jjsasha63:
Correcting my last comment: I was wrong about config.TelemetryOptions.

I checked whether an option's apply() can return an error, found that all ten
exported With* return nil, and concluded from that no option can make
telemetry.New fail. That was the wrong test. The failure does not have to come
from apply() — it can come from the value the option sets, further down in
configure().

@karolpiotrowicz's test above does work, and it is the better approach.
WithResource with a conflicting schema URL fails in resolveResource before any
network call:

    failed to resolve resource: failed to merge with config resource:
    conflicting Schema URL: https://opentelemetry.io/schemas/1.43.0 and
    https://conflicting.invalid/schema/v1

Use that, not the GOOGLE_APPLICATION_CREDENTIALS approach I suggested. It
stays inside the public API and leaves the process environment alone.

On holding for the test: I said I would not. Now that a drop-in test exists in
this thread, that reason is gone, so I have no objection to landing it with
the test.

PR #1356, opened 2026-08-18
fix(session/inmemory): strip temp: keys from the canonical event record

### LINKED ISSUE #1354: session.InMemoryService().AppendEvent: canonical session record skips temp: key stripping (v1.6.0 regression)
## Title
`session.InMemoryService().AppendEvent`: canonical session record skips `temp:` key stripping (v1.6.0 regression)

## Summary
`temp:`-prefixed `StateDelta` keys are supposed to be discarded once an event is persisted (`KeyPrefixTemp` doc: "discarded after the invocation completes"). Since v1.6.0, they leak into the session record returned by later `Get()`/`ListEvents()` calls. Confirmed clean on v1.5.1, leaking on v1.6.0.

## Root cause
`inMemoryService.AppendEvent` (session/inmemory.go) computes the temp-stripped `StateDelta` twice, inconsistently:

1. `sess.appendEvent(event)` → `trimTempDeltaState(event)` correctly strips `temp:` keys, but the result only lands on the **caller's local session copy** (`Create()`/`Get()` never hand callers a pointer to the canonical record — see `copySessionWithoutStateAndEvents`).
2. Separately, the canonical record (`stored_session`, looked up fresh from `s.sessions`) gets its own `eventCopy` built via `maps.Clone(event.Actions.StateDelta)` — cloned straight from the **original, untrimmed** `event`, never touching the `processedEvent` from step 1.

Until v1.6.0 this was masked by an aliasing side effect: `trimTempDeltaState` used to mutate the shared `event` pointer in place, so by the time step 2 ran, `event.Actions.StateDelta` was already clean. #1128 ("prevent mutation of input event during temporary state delta stripping", backported in `8b8c07a3`/v1.6.0) fixed that aliasing bug by returning a copy instead of mutating in place — correct in isolation, but it silently broke step 2's implicit reliance on the mutation.

Net effect: `AppendEvent`'s public contract (`temp:` keys don't survive past the current invocation) is broken for any caller that later reads state back via `Get()`/`ListEvents()`.

## Reproduction
```go
package main

import (
	"context"
	"fmt"

)

func main() {
	ctx := context.Background()

	created, _ := svc.Create(ctx, &session.CreateRequest{AppName: "demo", UserID: "user-1"})
	sess := created.Session

	evt := session.NewEvent("inv-1")
	evt.Actions.StateDelta = map[string]any{
		"temp:scratch":   "ephemeral",
		"persistent_key": "keep",
	}
	if err := svc.AppendEvent(ctx, sess, evt); err != nil {
		panic(err)
	}

	got, _ := svc.Get(ctx, &session.GetRequest{AppName: "demo", UserID: "user-1", SessionID: sess.ID()})
	for e := range got.Session.Events().All() {
		fmt.Printf("StateDelta after Get(): %#v\n", e.Actions.StateDelta)
	}
}
```

- `go.mod` pinned to `google.golang.org/adk v1.5.1` → `StateDelta after Get(): map[string]interface {}{"persistent_key":"keep"}` (correct)
- `go.mod` pinned to `google.golang.org/adk v1.6.0` → `StateDelta after Get(): map[string]interface {}{"persistent_key":"keep", "temp:scratch":"ephemeral"}` (leak)

## Suggested fix
In `AppendEvent`, build `eventCopy.Actions.StateDelta` from the `processedEvent` returned by `sess.appendEvent(event)` (or call `trimTempDeltaState` once and reuse its result for both the local and canonical copies), rather than re-deriving it from the original `event`.

## Environment
- `google.golang.org/adk` v1.6.0 (regressed), v1.5.1 (last known good)
- Go 1.23+

jordigilh:
Fixes #1354.

## Affected versions

| Version | Affected? | Evidence |
|---|---|---|
| `v1.5.1` | No | Confirmed clean in #1354's repro |
| `v1.6.0` | **Yes** | Confirmed leaking in #1354's repro (this is the version the issue was originally filed against) |
| `v2.0.0` | No | Tagged before #1128 (the mutation-fix that exposed this bug) merged into `main`; `git compare b09bad9...v2.0.0` shows `behind` -- the regression-causing commit isn't in this tag |
| `v2.1.0` | **Yes** | Tagged after #1128 merged; `git compare b09bad9...v2.1.0` shows `ahead` -- contains the regression |
| `v2.2.0` | **Yes** | Same as `v2.1.0`; `git compare b09bad9...v2.2.0` shows `ahead` |

Since `main` is the branch `v2.x` releases are cut from, this fix (once merged) will land in the next `v2.x` tag automatically. `v1.6.0` is on the separate, frozen `v1` branch and needs its own backport PR (flagging this so maintainers can prioritize -- happy to open the `v1` backport myself once this is reviewed/merged).

`AppendEvent` built `eventCopy.Actions.StateDelta` by cloning `event` (the original, caller-provided event), not the `trimTempDeltaState` result already computed two lines earlier by `sess.appendEvent`. Before #1128, `trimTempDeltaState` mutated `event` in place, so this happened to work. Once #1128 made it return a copy instead (to stop mutating the caller's event, a legitimate fix), `eventCopy` started cloning the untrimmed original — so `temp:`-prefixed `StateDelta` keys now survive into the canonical session record and reappear on a later `Get()`/`ListEvents()`, instead of being discarded after the invocation as documented.

`sess.appendEvent` now returns the already-trimmed event so `AppendEvent` can reuse it for `eventCopy` instead of re-deriving `StateDelta` from `event`. `database` and `vertexai` backends were checked and don't have this issue — both already reuse the trimmed result for persistence.

This landed in the current `v1` line as `v1.6.0`; may need a backport once merged (see affected-versions table above).

## Testing plan
- `TestInMemoryService_AppendEvent_StripsTempKeysFromCanonicalRecord`: reproduces the leak via a fresh `Get()` call (not the `Session` handle `AppendEvent` was called with — that handle is a separate local copy that `sess.appendEvent` already trims correctly, which is why the existing `TestInMemoryService_AppendEvent_PreservesInputEventTempState` test didn't catch this).
- `TestInMemoryService_AppendEvent_MultipleEventsAllStripped`: same check across multiple appended events.
- `go test ./session/...`: full suite passes.

---

## Maintainer update

Review feedback landed on this branch directly, so two things above are now out of date.

**Implementation.** `sess.appendEvent` keeps its `error`-only signature, matching the `database` and `vertexai` backends. Reusing its return value meant cloning a map already published into `sess.events` while holding only the service lock, which raced any goroutine writing into the newest stored event. The trim now happens at the service level, before that publication:

```go
trimmedDelta := maps.Clone(trimTempDeltaState(event).Actions.StateDelta)
```

Same shape `session/database` already uses, and race-neutral against `main`.

**Tests**, in addition to the two listed above:
- `TestInMemoryService_AppendEvent_AllTempKeysStrippedKeepsEmptyDelta` — every key stripped leaves the stored delta non-nil and empty, so the JSON round trip with adk-python holds.
- `TestInMemoryService_AppendEvent_CanonicalRecordDoesNotAliasLiveDelta` — the canonical record shares neither the live handle's map nor the caller's, across both delta shapes.

**Out of scope, tracked separately: #1560.** `AppendEvent` also clones `ArtifactDelta` and `RequestedToolConfirmations` without the session lock, and `trimTempDeltaState`'s shallow copy means the published event shares both with the caller's event. Both throw 5/5 on `main` today with no PR applied, as do the unlocked reads at `ExtractStateDeltas` and `updateSessionState`. None of it is a regression from this change and none of it is touched here.


jordigilh:
@baptmont @hanorik -- tagging you both since git history shows you're the most recently active on this exact code path:

- **@baptmont** (João Westerberg): authored the most recent commits to `session/inmemory.go` and approved #1128, the PR whose (correct, in isolation) fix exposed this regression.
- **@hanorik** (Mikalai Senkevich): merged #1128, and has the most recent commit (2026-07-08) to `session/inmemory.go`.

I don't have write/triage access on this repo to formally request review, so flagging here instead. Given this affects the currently-shipping `v2.1.0`/`v2.2.0` releases (see the affected-versions table in the PR description) as well as `v1.6.0`, wanted to make sure it gets eyes sooner rather than later. Happy to open the `v1` backport once this is reviewed.

karolpiotrowicz (dismissed):
The fix is right and the tests pin the thing that was actually invisible. I checked it by running both trees: with a delta of `{"temp:secret":"...", "keep":"yes"}`, a fresh `Get()` on main returns `map[keep:yes temp:secret:SCRATCH]` and with this change returns `map[keep:yes]`. In-invocation reads still work — `temp:secret` is still readable off the live handle on both sides — and the caller's own `*Event` is left untouched. That last part matters more than it looks: keeping `updateSessionState` on the untrimmed event at [`session/inmemory.go:364`](https://github.com/google/adk-go/blob/caef30051b4a002e0399130f0b08cfd5035cf9e8/session/inmemory.go#L364) before trimming at [`:367`](https://github.com/google/adk-go/blob/caef30051b4a002e0399130f0b08cfd5035cf9e8/session/inmemory.go#L367) is the easiest thing to get wrong here, and it was not gotten wrong. This also brings Go to parity with adk-python, which reaches the same end state by mutating the event in place at [`base_session_service.py:214-218`](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/sessions/base_session_service.py#L214-L218) — returning a copy, as you have done, is the better choice, because it preserves the caller's event where Python does not.

## One thing to change before merge

[`session/inmemory.go:238`](https://github.com/google/adk-go/blob/caef30051b4a002e0399130f0b08cfd5035cf9e8/session/inmemory.go#L238) now clones a map that `(*session).appendEvent` published into `sess.events` a moment earlier. `appendEvent` takes and releases `sess.mu` internally, so by the time `maps.Clone` runs, the only lock held is the service lock `s.mu` — a different mutex from the one guarding the map. Anything holding the live handle and writing into the newest stored event's `StateDelta` then races the clone, and the result is an unrecoverable runtime throw rather than a `-race` report, so it kills the process in ordinary builds too.

<details><summary>Reproduction — throws in 0.32s with this change applied, passes twice at 25s each without it</summary>

```go
// session/zz_alias_test.go
func TestAliasRace(t *testing.T) {
	cr, _ := svc.Create(t.Context(), &session.CreateRequest{AppName: "app", UserID: "u"})
	live := cr.Session
	stop := make(chan struct{})
	go func() {
		for {
			select {
			case <-stop:
				return
			default:
			}
			evs := live.Events()
			if n := evs.Len(); n > 0 {
				if e := evs.At(n - 1); e != nil && e.Actions.StateDelta != nil {
					e.Actions.StateDelta["injected"] = 1
				}
			}
		}
	}()
	for i := range 20000 {
		ev := &session.Event{
			ID:        fmt.Sprintf("e%d", i),
			Timestamp: time.Now(),
			Actions:   session.EventActions{StateDelta: map[string]any{"temp:t": i, "k": i}},
		}
		if err := svc.AppendEvent(t.Context(), live, ev); err != nil {
			t.Fatal(err)
		}
	}
	close(stop)
}
```

`go test -race -count=2 ./session/ -run TestAliasRace`

```
fatal error: concurrent map clone and map write
maps.Clone[...](...)
google.golang.org/adk/v2/session.(*inMemoryService).AppendEvent(...)
	session/inmemory.go:238
```
</details>

I want to be accurate about how much of this is yours, because most of it is not. The underlying pattern — reading the caller's delta map without the lock that guards it — is already on main at [`session/inmemory.go:257`](https://github.com/google/adk-go/blob/caef30051b4a002e0399130f0b08cfd5035cf9e8/session/inmemory.go#L257), where `ExtractStateDeltas` iterates it, and at `updateSessionState`. I ran four combinations of delta shape and writer, and three of them already throw on main. Exactly one flipped from safe to unsafe: a `temp:`-bearing delta with a writer that mutates a stored event. That one happens to be the path this change is about, which is why it is worth closing here even though the class predates you.

The fix I verified is to clone inside the lock — have `appendEvent` return `map[string]any` instead of `*Event`, ending with `return maps.Clone(processedEvent.Actions.StateDelta), nil`, and assign that directly at line 238. With that applied the reproduction above passes, and so do all nine tests in the file, under `-race`. Worth knowing before you decide how far to go: it closes the combination this change opened, and it does **not** close the pre-existing ones — the tree still throws at the `ExtractStateDeltas` line under a caller that mutates its own delta. That broader hazard belongs in its own issue rather than in this PR.

## Smaller things

The comment at [`session/inmemory.go:227-229`](https://github.com/google/adk-go/blob/caef30051b4a002e0399130f0b08cfd5035cf9e8/session/inmemory.go#L227-L229) says `eventCopy` is built from `processedEvent` rather than `event`, and that the canonical record therefore matches what `appendEvent` persisted locally. Only `StateDelta` comes from `processedEvent` — the other twelve fields come from `event` — and the two records demonstrably do not match, because the canonical copy clones `Compaction`, `ArtifactDelta` and `RequestedToolConfirmations` while the local record aliases them. That asymmetry is exactly what `TestInMemoryService_AppendEvent_CopiesCompaction` on main exists to pin. Scoping the sentence to `StateDelta` would make it true.

Both new tests always carry a non-temp key alongside the `temp:` one, so neither reaches the case where **every** key is stripped — the one input class where this change alters the shape of the stored map rather than its contents. It goes from a populated map to an allocated-but-empty one, which is correct and is what `EventActions.MarshalJSON` wants, but nothing pins it. A third case with `StateDelta: map[string]any{"temp:a": 1}` asserting the stored delta is non-nil with `len == 0` would stop a later tidy-up from returning `nil` there and silently breaking the round-trip with adk-python.

## On rebasing

The branch conflicts with main. The conflict is confined to `session/inmemory_test.go`, where #1231 appended `TestInMemoryService_AppendEvent_CopiesCompaction` at the same offset as your two tests — `session/inmemory.go` merges with no choice to make, and all nine tests pass together once both sides are kept. One thing to catch while you are in there: your comment at [`session/inmemory_test.go:240`](https://github.com/google/adk-go/blob/caef30051b4a002e0399130f0b08cfd5035cf9e8/session/inmemory_test.go#L240) says "the test above", and after the rebase the test above will be `CopiesCompaction` rather than `PreservesInputEventTempState`. Naming the test explicitly would survive the move.

## Separately, not for this PR

Three `temp:` containment gaps sit next to this code and are unchanged by it, so please do not take them on here. `Create()` at [`session/inmemory.go:81`](https://github.com/google/adk-go/blob/caef30051b4a002e0399130f0b08cfd5035cf9e8/session/inmemory.go#L81) discards the session-scoped return of `ExtractStateDeltas` and merges the raw state, so a literal `temp:` key passed at creation is persisted permanently — the database backend gets this right. `session/vertexai` sends the untrimmed event to the remote API before trimming its local copy, which makes it the last backend still leaking after your change lands. And the shared conformance suite asserts only on session state, never on stored event deltas, which is why none of this was caught by the suite the three backends share. I will file those.

wojtas000:
I take this PR over as we would like to have this fixed. Thanks @jordigilh for your contribution!

wojtas000:
Took this over to land the review feedback. @jordigilh's commit is intact, and `main` is merged in, so the conflict is gone.

**The race.** The clone was reading a map `appendEvent` had already published into `sess.events`, while holding only the service lock — I reproduced the throw in 3 of 5 runs. The trim and clone now happen at the service level, before that publication, so the map isn't reachable by anyone else yet: 0 of 5 after. `appendEvent` keeps its `error`-only signature, matching the database and Vertex AI backends, and this is the same shape `session/database` already uses.

**Smaller things.** The comment at the `eventCopy` site is scoped to `StateDelta`, and the canonical-record test names the test it contrasts with rather than saying "the test above".

**Two tests.** One for the case where every key is stripped, where the stored delta has to stay non-nil and empty. One for aliasing: the canonical record shares neither the live handle's map nor the caller's. It runs both delta shapes, because `trimTempDeltaState` returns the event untouched when it strips nothing — only the no-`temp:` case reaches the arrangement where the live session publishes the caller's own map. Deleting the clone fails it either way.

Deliberately untouched, now filed as #1560: the unlocked reads at `ExtractStateDeltas` and `updateSessionState`, and the clones of `ArtifactDelta` and `RequestedToolConfirmations` — `trimTempDeltaState` copies the event shallowly, so the published event shares both of those maps with the caller's. All of it throws 5/5 on `main` with no PR applied, so none of it is a regression from this change. Separate issue as @karolpiotrowicz suggested, with the measurements in it.

`./session/...` is green under `-race -shuffle=on`; lint, gofmt and `go mod tidy -diff` clean.

kdroste-google (approved):
LGTM, thank you!

PR #1357, opened 2026-08-18
fix: reject tool confirmations resumed from non-agent-authored events

prasanna8585:
RequestConfirmationRequestProcessor resumes a pending tool call by scanning session events for a `request_confirmation` FunctionCall whose ID matches an incoming confirmation response, but it never checked who authored that FunctionCall event.

Function call parts also reach the session from places other than this agent's own model turn — notably an A2A peer response converted into a model-role event. Without an author check, a peer could inject such an event carrying an attacker-chosen tool name and arguments, and have it resumed as a genuine confirmed call once any confirmation response with a matching ID arrived.

This mirrors the check already present in adk-java (`RequestConfirmationLlmRequestProcessor.isResumableFunctionCall`), which explicitly guards against the same A2A scenario.

Also updates four existing test fixtures that set `Author` to the literal string "agent" instead of the real agent name ("testAgent") — they were passing only because no author check previously existed to catch the mismatch.

Verified locally: full `RequestConfirmation*` suite passes, and a PoC reproducing the forged-event resume (attacker-controlled tool + args dispatched via a non-agent-authored event) now correctly rejects the call after this fix.


baptmont:
Hi @prasanna8585, Thank you for the PR but it is currently needing some lint fixes

prasanna8585:
Hi @baptmont ,
Can you please review the changes ?

baptmont (approved):

PR #1359, opened 2026-08-18
fix(model/openaimodel): surface server-side response failures

hanorik:
A failed response comes back as HTTP 200 with status "failed", so the blocking path could not tell one from a normal turn: convertResponse read only IncompleteDetails.Reason. A failure with no output was reported as ErrNoOutputItems and one with partial output as a clean STOP, and in both cases resp.Error.Message — the only account of what went wrong — was dropped. The streaming path already errored out on "response.failed", so the two disagreed on identical server behaviour.

Fail the conversion on status "failed", carrying the server's message and error code, and have the streamed event build its error the same way. A status that is neither "completed" nor absent no longer reads as STOP when no incomplete reason is given.

## Behaviour changes callers can see

**A failed response now fails the call, on both paths.** `status: "failed"` arrives as HTTP 200. Previously a failure with no output surfaced as `ErrNoOutputItems` and one with partial output as a clean `STOP` turn; both now return `ErrResponseFailed`, and any partial text is discarded on the blocking path. A caller that matched on `ErrNoOutputItems` to detect this, or that consumed the partial answer, needs updating. Streaming keeps deltas it has already delivered, then ends the turn with the error.

**`finishReason` returns `Other` where it used to return `Stop`** — for `cancelled`, `queued`, `in_progress`, a bare `incomplete` carrying no `incomplete_details.reason`, and any status a later SDK adds. This is deliberate: calling a cancelled generation a clean stop is what let a truncated answer pass as a finished one.

The consequence is real where a non-`Stop` reason is treated as fatal. `session/compaction/llm_summarizer.go:263` turns any non-`Stop` finish reason into a hard error, so a summarization that previously succeeded against one of those statuses now fails instead. Both are hard to reach against `api.openai.com` — a real cancelled response carries no output items and so errors earlier, and a real `incomplete` always carries a reason — but an OpenAI-compatible endpoint may well produce them.

**Error text changed.** A failure now reads `openai: response failed (id "resp_123", code "server_error"): upstream exploded`, wrapping the new `ErrResponseFailed` sentinel. Anything matching the old streamed text (`openai response failed: <msg>`) should use `errors.Is` instead.

**A body with an `error` object and no `status` at all now counts as failed.** `status` is not in the API's required set while `error` is, so such a body is conforming; it previously read as a clean `STOP` with partial text. An explicit status is never overridden by it.

## Testing Plan

Repo gates, both modules (root and `plugin/agentanalytics`), all green:

```
go build -mod=readonly work                        OK
go test -race -mod=readonly -count=1 -shuffle=on work   ok (all packages)
golangci-lint run                                  0 issues.
go mod tidy -diff                                  (no output)
```

`model/openaimodel` has no httprr cassettes, so its unit tests assert this
package's serialization against its own expectation — nothing in CI proves the
provider accepts a payload. Three things stand in for that here.

**Mutation testing.** Each assertion added by this PR was checked by breaking
the code it guards and confirming the suite goes red:

| Mutation | Result |
|---|---|
| drop the pre-marker `strings.TrimSpace` in `clipServerText` | 4 tests fail |
| boundary `n == maxServerTextRunes-1` | `exactly the cap` fails |
| boundary `n == maxServerTextRunes+1` | `one past the cap` fails |
| reword `openai stream error: %q` | `a short message is quoted…` fails |
| reword the bare `openai stream error` | `an error event with no message…` fails |

**Fuzzing.** `FuzzClipServerText` ran 45s / ~920k execs with no crasher.

**Live e2e** (private `adk-openai-e2e` suite, since this package has no
cassettes):

```
adk-go under test: adk-wt/pr1359
branch / commit:   pr1359 / 6d753fc5
openai-go:         v3.54.0
model:             gpt-5.4-nano
TOTAL              54 calls   5175 in   713 out   $0.001925
```

Scenarios 01–15 pass, including 14 (error paths) and 15 (unsupported config),
which are the ones this change's defect class lives in. Scenario 16 fails, and
fails identically against pristine `main` at a6114727 — it targets the
`ThinkingConfig`/`service_tier` support from the unmerged #1392, so it is
unrelated to this PR.

## Review loop

Three rounds of the six-lens panel earlier in the PR's life; the third found
three self-inflicted defects (a false fuzz bound, an unpinned seed, an uncovered
fallback), all fixed.

The latest round answered three review findings — a fuzz assertion that fired on
correct output, a table comment naming an operator the code no longer used, and
two unpinned error strings — and re-ran a narrowed panel (testing, correctness,
idioms) over the delta. All three lenses returned CLEAR. Two non-blocking nits
were taken (comment wording, subtest ordering); one was declined: the new gate
stays a nested `if` rather than an early `return`, because a `return` in a fuzz
body silently skips any assertion added after it.


baptmont (dismissed):

karolpiotrowicz (commented):
The blocking fix is right. Replaying live Responses API traffic — normal and truncated turns, streamed and not, plus a stream dropped before its terminal event — gives identical text, finish reason and usage on `main` and on this branch, so the change is inert on conforming traffic.

One thing is worth closing first, and it is invisible from this branch because it only appears after a rebase onto `main`.

**The guard does not run on the streaming path once a delta has arrived.** `convertResponse` is the only place `status == "failed"` is checked ([response.go#L32-L34](https://github.com/google/adk-go/blob/85838feb0c46f8a80ab61bc7b713b93c0670159a/model/openaimodel/response.go#L32-L34)), and since #1355 `generateStream` calls it only inside `if final == nil` ([openai.go#L168-L186](https://github.com/google/adk-go/blob/2325d1cf4f0c1e097f13403d84493cea23660f03/model/openaimodel/openai.go#L168-L186)). A stream that carried text skips it entirely.

```
created → output_text.delta → completed, whose response object is
  {"status":"failed","error":{"code":"server_error","message":"upstream exploded"},"output":[…]}

  streaming: err=nil, TurnComplete=true, FinishReason=OTHER, partial text delivered
  blocking:  openai: response failed: upstream exploded (server_error)
```

Before this change both paths returned that as a clean turn, so they agreed. Now they disagree, which is the failure mode the PR exists to remove. A real `response.failed` event after deltas is handled correctly, so reaching this needs a server that reports the failure in the terminal body instead — not `api.openai.com`, but the package is scoped to compatible endpoints and I have no measurement of what Ollama, LM Studio or vLLM emit here. Four lines before `finalizeStreamResponse` is cheaper than finding out:

```go
if openaiResp != nil && openaiResp.Status == responses.ResponseStatusFailed {
    yield(nil, failedResponseError(openaiResp))
    return
}
```

**Nothing pins the guard's narrowness.** `"failed"` is the only status any test feeds through `convertResponse`, so widening it to also reject `cancelled`, `incomplete` and `queued` keeps the suite green — while turning every `max_output_tokens` truncation into a hard error. A `convertResponse` case asserting that a `completed` turn and a reason-carrying `incomplete` turn still succeed would close it. Real bodies always carry `status` and the package's stream fixtures omit it ([openai_test.go#L139-L149](https://github.com/google/adk-go/blob/2325d1cf4f0c1e097f13403d84493cea23660f03/model/openaimodel/openai_test.go#L139-L149)), which is part of why this is easy to miss.

**Two comments outrun the code.** `"failed" never reaches here` ([response.go#L204-L206](https://github.com/google/adk-go/blob/85838feb0c46f8a80ab61bc7b713b93c0670159a/model/openaimodel/response.go#L204-L206)) is false on `main`, and `TestModel_Generate_FailedStatus` is documented as pinning the blocking path to the streamed one ([openai_test.go#L73-L75](https://github.com/google/adk-go/blob/85838feb0c46f8a80ab61bc7b713b93c0670159a/model/openaimodel/openai_test.go#L73-L75)) but passes `false`. Closing the gap above makes the first one true again.

Non-blocking: a whitespace-only server message renders a dangling separator, the `(code)` parenthetical duplicates a code the message already names, a failed turn drops billable usage and the response id, and the absent-status subtest is named `/`.

wojtas000:
Follow-up commit addressing issues in the review:
- Fixed blocking gap: `status: "failed"` check ran only when `final == nil` - otherwise the failure went unnoticed and the turn came back as normal finished answer (inconsistent with `generate()` non-streaming path)
- Added `Response ID` to the error text
- narrowness is pinned in tests
- Added beyond the review: event with `error` and no `status` also counts as failed

model/openaimodel/response.go
```diff
@@ -42,6 +46,50 @@ func convertResponse(resp *responses.Response) (*genai.GenerateContentResponse,
...
+// else is left to quote back to the provider.
+func failedResponseError(resp *responses.Response) error {
+	// Trimmed, so a message of nothing but spaces leaves no dangling separator.
+	msg := strings.TrimSpace(resp.Error.Message)
+	var details []string
+	if id := strings.TrimSpace(resp.ID); id != "" {
+		details = append(details, id)
+	}
+	// A code the message already spells out is not worth repeating.
+	if code := strings.TrimSpace(string(resp.Error.Code)); code != "" && !strings.Contains(msg, code) {
```
karolpiotrowicz: The code de-duplication is an unanchored substring test, so a code that merely appears inside a longer token in the message is dropped entirely. `code="server_error"` with `message="Downstream returned server_error_5xx; retry later."` renders `openai: response failed (resp_1): Downstream returned server_error_5xx; retry later.` — the actual code never appears as a field. Comparing on a token boundary, or matching a prefix, would keep it.

Separately, this `TrimSpace` and the one on the ID above are never exercised as trims: no fixture supplies a whitespace-only code or ID. Only the `TrimSpace` on the message has a test behind it, the "blank message" row.

model/openaimodel/response.go
```diff
@@ -42,6 +46,50 @@ func convertResponse(resp *responses.Response) (*genai.GenerateContentResponse,
...
+// failedResponseError renders a failure as an error quoting the server: the
+// message as the text, the response ID and error code as a parenthetical. The
+// ID is there because the response is discarded with the failure, so nothing
+// else is left to quote back to the provider.
+func failedResponseError(resp *responses.Response) error {
+	// Trimmed, so a message of nothing but spaces leaves no dangling separator.
+	msg := strings.TrimSpace(resp.Error.Message)
+	var details []string
+	if id := strings.TrimSpace(resp.ID); id != "" {
+		details = append(details, id)
```
karolpiotrowicz: The ID gets no equivalent of the de-duplication applied to the code just below, so `ID="resp_123"` with a message that already names `resp_123` renders `openai: response failed (resp_123): resp_123 could not be completed`.

model/openaimodel/response.go
```diff
@@ -42,6 +46,50 @@ func convertResponse(resp *responses.Response) (*genai.GenerateContentResponse,
...
+	}
+	// A code the message already spells out is not worth repeating.
+	if code := strings.TrimSpace(string(resp.Error.Code)); code != "" && !strings.Contains(msg, code) {
+		details = append(details, code)
+	}
+	switch joined := strings.Join(details, ", "); {
```
karolpiotrowicz, line 114: The parenthetical is an unlabelled comma-joined list, so a server-chosen ID containing `", "` is indistinguishable from an `id, code` pair: `ID="resp_123, invalid_prompt"` renders `openai: response failed (resp_123, invalid_prompt)`, exactly what an ID plus that error code would render. Labelling the fields would remove the ambiguity.

karolpiotrowicz (commented):
The streaming gap from the last round is fixed, and it is pinned rather than merely present: deleting the new check in `generateStream` makes `TestModel_FailedStatus` and `TestModel_FailedStatus_PathsAgree` fail. Replaying real Responses API traffic through both trees — a completed body, a truncated one, and a genuinely cancelled one obtained by creating a background response and cancelling it — gives identical text, finish reason and usage on `main` and on this branch, so the change is inert on conforming traffic. Build, `go test ./...` (79 packages), vet, gofmt, staticcheck, govulncheck and apidiff are all clean, and it still merges and passes on current `main`.

Four things worth your time. A few smaller notes are on the lines themselves.

## The `case "":` branch has no test that distinguishes its two halves

Changing `||` to `&&` at [`response.go:59`](https://github.com/google/adk-go/blob/5713a398a45031df95ff8d6c47104d38caf134f5/model/openaimodel/response.go#L54-L62) leaves the entire repo suite green. Nothing anywhere feeds a status-less body with exactly one of message or code: `bodyFailedNoStatus` sets both, `evCompletedNoStatus` sets neither, the "absent" row of `TestConvertResponse_StatusOtherThanFailed` is deliberately given no error object, and the message-only and code-only rows in `stream_test.go` hard-code `"status":"failed"` so they never reach the branch.

That is the realistic half that is untested. The branch exists for OpenAI-compatible endpoints, and a gateway sending `message` without `code` is exactly what you would expect from one. Two rows on the existing table close it.

## A test comment claims a guarantee the table does not provide

[`response_test.go:195-198`](https://github.com/google/adk-go/blob/5713a398a45031df95ff8d6c47104d38caf134f5/model/openaimodel/response_test.go#L195-L198) says widening the check "breaks this". Adding `responses.ResponseStatusQueued` to `reportsFailure` leaves the whole suite green — the table covers completed, incomplete three ways, cancelled, `"moderated"` and absent, but has no `queued` row, and the only `queued` in the package calls `finishReason` directly so it cannot see a `reportsFailure` change. `queued` is one of the statuses your own doc comment on the guard names.

Either add the row or soften the comment. As it stands, someone who widens the check, reads that comment and sees a green suite will conclude the widening was safe.

## `finishReason` changes what callers see, and nothing outside a code comment says so

The new status switch at [`response.go:227-237`](https://github.com/google/adk-go/blob/5713a398a45031df95ff8d6c47104d38caf134f5/model/openaimodel/response.go#L227-L237) returns `Other` where `main` returned `Stop`, for `cancelled`, `queued`, `in_progress`, a bare `incomplete`, and any status a later SDK adds. I think the new value is the right one — calling a cancelled generation a clean stop is what let a truncated summary get stored — and I could not reach it on real traffic: a real cancelled response comes back with zero output items (3/3, cancelling after 3s, 6s and 10s) so it errors at `ErrNoOutputItems` first, and a real `incomplete` always carries `incomplete_details.reason` (3/3) so the reason branch wins.

The consequence is real for anyone who does hit it. [`llm_summarizer.go:263`](https://github.com/google/adk-go/blob/5713a398a45031df95ff8d6c47104d38caf134f5/session/compaction/llm_summarizer.go#L263-L264) turns any non-`Stop` finish reason into a hard error, so a summarization that previously succeeded now fails. That belongs in the PR description, not only in the comment at `response.go:233`.

Worth checking against #1373 too, which rewrites `finishReason` with a different signature on the same package.

## Server-controlled text goes into the error uncapped and unescaped

[`response.go:82`](https://github.com/google/adk-go/blob/5713a398a45031df95ff8d6c47104d38caf134f5/model/openaimodel/response.go#L82) interpolates the provider's `Error.Message`, `Error.Code` and the response ID with bare `%s`. `TrimSpace` strips the ends only, so an interior `\r\n` forges a second log line, and a 1 MiB message produces a 1,048,601-byte error. On the blocking path that reaches [`triggers.go:230`](https://github.com/google/adk-go/blob/5713a398a45031df95ff8d6c47104d38caf134f5/server/adkrest/controllers/triggers/triggers.go#L230), which decides "throttled" by `strings.Contains(err.Error(), "429")` — a provider message containing `429` anywhere drives the retry loop and eventually a NACK.

I am flagging this at low priority deliberately. The pattern predates this change: `main` already did it on the streaming path, and [`stream.go:91`](https://github.com/google/adk-go/blob/5713a398a45031df95ff8d6c47104d38caf134f5/model/openaimodel/stream.go#L91) still does and is untouched here. If it is worth fixing it is worth fixing all three sites together, not just the new one, and it is a fair follow-up rather than something to hold this PR for.

## What I could not check

No OpenAI-compatible gateway was exercised, only `api.openai.com`. That is the population the `case "":` branch was written for, and it is the one thing that would settle whether the branch can misfire: if such a server omits `status` and sets any error field on a successful turn, every call fails. Every status-less case I ran was the real API's bytes with the field stripped in transit, which is not the same evidence. One `ollama serve` and a single call through `examples/openai/main.go`, checking whether `error` is absent or null on a good response, would answer it.

I also never saw a real HTTP-200 `status:"failed"` body — it is a generation-time incident and not summonable, so the central fix is verified against synthetic bodies and the SDK schema rather than against a live failure.

model/openaimodel/response.go
wojtas000: Dropped the de-duplication rather than anchoring it: losing the code the caller needs is worse than printing it twice, and both trims are now covered by padded and whitespace-only fixtures.

model/openaimodel/response.go
wojtas000: The ID gets no de-duplication either now, for the same reason — symmetric with the code, and pinned by a row where the message names the ID.

model/openaimodel/response.go
wojtas000: Labels alone only moved the collision — an ID of `resp_123, code invalid_prompt` still rendered identically to an id/code pair — so both fields are now labelled *and* quoted, with a row pinning exactly that forgery.

wojtas000:
All four, plus the three line notes, are addressed in `5789635` and `8dad2f3`.

**1. The `case "":` branch has no test that distinguishes its two halves** — added two rows, a status-less body with only a message and one with only a code, so flipping the `||` to `&&` now fails both.

**2. A test comment claims a guarantee the table does not provide** — added `queued` and `in_progress` rows, so widening `reportsFailure` with any status the guard's comment names now fails a row; the comment's own enumeration was stale and is updated.

**3. `finishReason` changes what callers see** — now in the PR description, naming the `llm_summarizer.go:263` consequence explicitly, alongside the other three visible changes (failed responses erroring on both paths, the new error text, the absent-status rule).

**4. Server-controlled text goes into the error uncapped and unescaped** — fixed at all three sites together as you framed it, in its own commit: the failed response body, the ID and code, and the generic stream error event. Every server-chosen string is capped at 256 runes with the truncation marked, and its control characters rendered visibly, so an interior CRLF cannot forge a log line. Removing either the cap or the escaping fails a test.

**On the gateway you could not exercise** — I could not run one either, but Ollama answers it at source: `ToResponse` sets `Status: "completed"` and `Error: nil, // Only populated on failure`, and the streaming `buildResponseObject` hard-codes `"error": nil`, so `status` is never absent there and the branch is unreachable. vLLM requires `status` and has its `error` field commented out entirely. `error: null` decodes to a zero struct, so the null case is inert. That leaves a gateway that both omits `status` and sets a non-empty error field on a successful turn — source reading, not a live run, and I have not seen a real HTTP-200 `status:"failed"` body either.

**On #1373** — it changes `finishReason`'s signature and rewrites the same `case "":` arm, and its `truncated()` default arm is a superset of this status switch. Whichever lands first, the other should rebase rather than merge; the two rules should end up in one function, not two.

model/openaimodel/response.go
```diff
@@ -42,6 +47,99 @@ func convertResponse(resp *responses.Response) (*genai.GenerateContentResponse,
...
+	b.Grow(len(s))
+	for _, r := range s {
+		switch {
+		case r == '\n':
+			b.WriteString(`\n`)
+		case r == '\r':
+			b.WriteString(`\r`)
+		case r == '\t':
+			b.WriteString(`\t`)
+		case unicode.IsControl(r):
```
karolpiotrowicz: This arm is the one that covers ESC, DEL and the C1 block, and it is the only arm with no test behind it — the coverage profile shows an execution count of zero, and deleting it leaves the package green. Every control character the suite feeds through `escapeControl` is `\r`, `\n` or `\t`, each of which is handled above.

A fuzz target would be the cheapest fix and would cover more than a row would: `clipServerText` and `escapeControl` are pure, total, string-in/string-out functions with two checkable invariants — no output rune satisfies `unicode.IsControl`, and the rune count never exceeds the cap plus the marker. That one file also closes the byte-versus-rune and boundary gaps below.

karolpiotrowicz (commented):
All four are addressed, and the two that had a measured prior are measurably closed: flipping `||` to `&&` and widening `reportsFailure` to `queued` both now fail, where before they left the whole suite green. Widening to `in_progress` fails too. Deleting the streaming fix still fails, so the original guard is still pinned. Gates are green on both trees — build, `go test ./...` (79 packages), vet, gofmt, race, staticcheck, govulncheck, apidiff compatible, and the nested `plugin/agentanalytics` module. Replaying real Responses API traffic through both trees, including a genuinely cancelled response, gives identical results.

The Ollama and vLLM source reading answers the question I could not close last round, and it checks out: `ToResponse` sets `Status: "completed"` with `Error: nil`, and vLLM makes `status` required with the error field commented out. Worth being precise about what it settles, though. Both reads cover the *success* serialisers at an unpinned `main`, so they say nothing about those projects' error paths, about other gateways, or about the versions people actually run. It narrows the `case "":` question a lot. It does not close it.

Two things worth your time, both in the new code. Everything else is small and batched on the lines.

## The generic control-character arm is never exercised

Deleting `case unicode.IsControl(r)` from [`escapeControl`](https://github.com/google/adk-go/blob/8dad2f3f67d4f23dcd6646d63325242a58ab80e3/model/openaimodel/response.go#L69-L88) leaves the package green. The coverage profile says the same thing without any mutation: that arm has an execution count of zero. Every control character any test feeds through it is `\r`, `\n` or `\t`, and each of those has its own dedicated arm above it.

So the branch that actually defends against ESC, DEL and the C1 block is the one branch with no test behind it. A message of `"\x1b[2J\x1b[1;1H"` reaching a terminal that is tailing logs clears the screen. The code is right — I am not claiming otherwise — but nothing would tell you if it stopped being right.

A related one, same cause: deleting `case r == '\r'` also survives, because the assertion that looks like it pins CR is satisfied by `%q` on the ID rather than by `escapeControl` on the message.

## `reportsFailure` and `failedResponseError` disagree about whitespace

[`reportsFailure`](https://github.com/google/adk-go/blob/8dad2f3f67d4f23dcd6646d63325242a58ab80e3/model/openaimodel/response.go#L97-L106) tests the raw strings; [`failedResponseError`](https://github.com/google/adk-go/blob/8dad2f3f67d4f23dcd6646d63325242a58ab80e3/model/openaimodel/response.go#L118-L141) tests the clipped ones. A body with no `status`, `error.message` of `"  "`, and real output items therefore takes the failure path on the raw value, then renders to nothing:

```
reportsFailure  -> true
convertResponse -> "openai: response failed"
```

The output items are discarded and the error names nothing the server said. The renderer's own comment says a value of nothing but spaces should contribute nothing — the predicate twenty lines above says the same value means the server reported a failure. One of the two is wrong, and trimming in the predicate settles it.

I could not reach this from a real provider, and the gateway reading above makes it unlikely. It is one line in code this PR introduces, so I would rather raise it than not.

## One suggestion, and one trap if you take the other route

Four of the smaller items — the `\u2028` gap, the message imitating the parenthetical, and the escaping not being invertible — are all one decision: the message is interpolated bare while the ID and code get `%q`. `%q` on the message dissolves all three at once. The cost is the "left exactly as the server wrote it" property you currently pin in a test, so it is a real trade-off rather than an obvious win. `auth/gcp/client.go` already made the same call in the other direction, with a comment explaining why, if you want a precedent to point at.

If you instead widen `escapeControl` to cover `Zl`, `Zp` and `Cf`, note that `\x%02x` renders U+2028 as `\x2028` — a four-digit `\x` escape, which is ambiguous. That branch needs `\u%04x` above U+00FF.

## One thing that is not yours to fix

The `429` path from the last round is still open and cannot be closed here. `isResourceExhausted` in the triggers controller decides "throttled" by looking for `429` anywhere in `err.Error()`, so a short provider message containing those digits still drives the retry loop — capping and escaping do not touch it. Two small notes. Escaping can now *synthesise* the substring, since a `\x04` followed by `29` renders as `\x0429`. And there is a second substring-on-error classifier in `base_flow.go` that I did not check for reachability from this package. Both belong in an issue against those owners rather than in this PR.

model/openaimodel/response.go
```diff
@@ -42,6 +47,99 @@ func convertResponse(resp *responses.Response) (*genai.GenerateContentResponse,
...
+// maxServerTextRunes bounds any server-chosen string an error quotes.
+const maxServerTextRunes = 256
+
+// clipServerText trims a server-chosen string and caps its length, so that a
+// pathological one — a megabyte of message — cannot become the error a caller
+// logs. Truncation is marked, so a clipped value is never mistaken for what the
+// server actually said.
+func clipServerText(s string) string {
+	s = strings.TrimSpace(s)
+	if r := []rune(s); len(r) > maxServerTextRunes {
```
karolpiotrowicz: Two mutants survive here, so the arithmetic is unpinned in both directions. Slicing by bytes instead of runes leaves the suite green — it would sever a multi-byte rune and emit invalid UTF-8 — because every fixture that reaches the clip is pure ASCII. Changing `>` to `>=` also survives. At exactly 256 runes it appends a spurious `…` to text that was never truncated, which is the one thing the doc comment above says a clipped value must never be confused with. Two rows, 256 and 257 runes, plus one multi-byte row, pin all of it.

Separately: `[]rune(s)` materialises the whole input before the length test, so an 8 MiB message costs a 32 MiB transient allocation to produce at most a kilobyte. `utf8.RuneCountInString` avoids it. Error path only, so this is minor.

model/openaimodel/response_test.go
```diff
@@ -83,6 +83,317 @@ func TestConvertResponse_NoOutput(t *testing.T) {
...
+		}
+	})
+
+	t.Run("a huge message is capped", func(t *testing.T) {
+		err := failedResponseError(&responses.Response{
+			Status: responses.ResponseStatusFailed,
+			Error:  responses.ResponseError{Message: strings.Repeat("A", 1<<20)},
+		})
+		// The cap plus the sentinel, the ": " separator and the ellipsis.
+		if got, max := len(err.Error()), maxServerTextRunes+64; got > max {
```
karolpiotrowicz: This bound is true of its fixture rather than of the code. The cap counts runes and the assertion counts bytes, so it holds only because the payload is ASCII. Measured on the current implementation: 300 CJK runes render 796 bytes, 300 emoji 1052, and 300 C0 controls 1052, against the 320 asserted here. The id-and-code assertion below has the same shape — astral runes under `%q` reach about 5.2 KB against the 576 it allows.

The value is still bounded, so nothing is wrong with the code. The comment above states a guarantee the implementation does not make, and one non-ASCII row would keep it honest.

model/openaimodel/stream_test.go
```diff
@@ -137,3 +139,72 @@ func TestStreamTranslator_FunctionCall_MissingDoneName(t *testing.T) {
...
+	if err == nil {
+		t.Fatal("process() err = nil, want the stream error")
+	}
+	if strings.ContainsAny(err.Error(), "\r\n") {
+		t.Errorf("process() err = %q still contains a raw newline", err)
+	}
+
+	long := decodeEvent(t, `{"type":"error","message":"`+strings.Repeat("A", 1<<16)+`"}`)
+	_, err = newStreamTranslator().process(long)
+	if got, max := len(err.Error()), maxServerTextRunes+64; got > max {
```
karolpiotrowicz: `err` is dereferenced here without the nil guard the first half of the test has. If the `errorEvent` path ever regressed to returning a nil error, this panics and takes the rest of the package's results with it rather than failing one assertion.

model/openaimodel/response.go
wojtas000: Deleted `escapeControl` entirely and rendered every server-chosen string with `%q` instead, so the untested arm no longer exists — plus the two fuzz targets you suggested, which now kill the byte-vs-rune and boundary mutants as well.

model/openaimodel/response.go
wojtas000: Both mutants are dead: rows at exactly the cap and one past it pin the boundary, multi-byte and astral rows pin the slicing, and `[]rune(s)` is gone in favour of a ranging count that returns at the cap, so an 8 MiB message never materialises.

model/openaimodel/response_test.go
wojtas000: Every bound now counts runes, the unit the cap is written in, and the id/code row is deliberately non-ASCII; the justification comment was wrong too — those runes are printable, so `%q` emits them raw — and now says so.

model/openaimodel/stream_test.go
wojtas000: Guarded with `t.Fatal` before the dereference, and the surrounding subtests are now symmetric with the response path: same character classes rejected, and the cap fixture measured in runes.

wojtas000:
Both addressed in `acefcf4`, and I took the `%q` route you suggested.

**The generic control-character arm is never exercised** — deleted `escapeControl` outright rather than testing it: every server-chosen string now goes through `%q`, which closes the U+2028/U+2029 gap, the message imitating the parenthetical and the non-invertibility in one move, and follows the precedent you pointed at. Your trap is moot as a result — no `\x%02x` remains.

**`reportsFailure` and `failedResponseError` disagree about whitespace** — real bug, fixed: the predicate now clips too, and a test reproduces your exact body (no status, whitespace-only message and code, real output items) and asserts the turn converts instead of erroring.

**On the fuzz targets you suggested** — both added, and they earn their place: they kill the byte-versus-rune slicing and the boundary mutants that the tables missed. Their length ceiling is derived rather than guessed — `%q` renders an unprintable astral rune as the ten characters of `\U0010ffff`, so the worst case is 7727 runes against an asserted 7780. Both run as ordinary seed tests, and ~2M executions found nothing.

**On what the source reading settles** — agreed, and I have not overstated it in the PR description: it narrows the `case "":` question to a gateway that both omits `status` and sets a non-empty error field on a successful turn, which contradicts the field's only documented meaning. It does not close it, and I still have not seen a real HTTP-200 `status:"failed"` body.

**The 429 classifier** — still out of scope, and worth noting for the issue against those owners that `%q` slightly widens the synthesis surface you identified: alongside `\x04` + `29`, an unprintable astral rune whose code point contains those digits now renders `\U000e0429`.

model/openaimodel/servertext_test.go
```diff
@@ -0,0 +1,313 @@
...
+			t.Errorf("clipServerText(%q) returned %d runes, want at most %d", in, n, maxServerTextRunes+1)
+		}
+		if utf8.ValidString(in) && !utf8.ValidString(got) {
+			t.Errorf("clipServerText(%q) turned valid UTF-8 into %q", in, got)
+		}
+		if strings.TrimSpace(got) != got {
+			t.Errorf("clipServerText(%q) = %q, want no leading or trailing space", in, got)
+		}
+		// The marker sits at the end, so trailing space hides behind it.
+		if body, marked := strings.CutSuffix(got, "…"); marked && strings.TrimSpace(body) != body {
```
karolpiotrowicz: Gate this on the input having actually exceeded the cap, otherwise it fires on output that is correct:

```go
if utf8.RuneCountInString(strings.TrimSpace(in)) > maxServerTextRunes {
	if body, marked := strings.CutSuffix(got, "…"); marked && strings.TrimSpace(body) != body {
		t.Errorf("clipServerText(%q) = %q, want no space before the truncation marker", in, got)
	}
}
```

I ran it both directions. With the gate, `"rate limited, retrying …"` passes. Without the production code's pre-marker `TrimSpace` it still fails on a 250-`A` + 10-space + 10-`B` input, so the assertion keeps its teeth.

The alternative, if you would rather not compute a rune count in the assertion, is to give `clipServerText` a `(string, bool)` signature and have the test consult the real truncation flag.

model/openaimodel/servertext_test.go
```diff
@@ -0,0 +1,313 @@
...
+		name      string
+		in        string
+		wantRunes int
+		wantMark  bool
+	}{
+		{name: "short", in: "upstream exploded", wantRunes: 17},
+		{name: "blank", in: "  \n\t ", wantRunes: 0},
+		{
+			// Exactly at the cap: nothing was dropped, so nothing may claim it
+			// was. A >= in place of the > would break this and nothing else.
```
karolpiotrowicz: This comment names an operator the code no longer uses — `response.go:67` is `n == maxServerTextRunes` now, not `>`. The row is still the right row and still catches an off-by-one in either direction, so it is only the sentence that is stale.

model/openaimodel/stream.go
```diff
@@ -82,12 +82,15 @@ func (t *streamTranslator) process(evt responses.ResponseStreamEventUnion) (*gen
...
+		// failure reads the same however it arrived.
+		return nil, failedResponseError(&failed.Response)
 	case errorEvent:
 		// Generic stream errors are also returned.
-		if evt.Message != "" {
-			return nil, fmt.Errorf("openai stream error: %s", evt.Message)
+		// Same treatment as a failed response body: the text is the server's,
+		// so it is capped, and quoted rather than interpolated bare.
+		if msg := clipServerText(evt.Message); msg != "" {
+			return nil, fmt.Errorf("openai stream error: %q", msg)
```
karolpiotrowicz, line 100: This is the one error string the PR changes that no test pins to a literal. Changing the prefix to `openai: stream error:` leaves the whole package green — `TestStreamTranslator_ErrorEvent_ServerText` asserts properties (no control character, still errors with no message, a rune ceiling) rather than the text.

One row asserting `err.Error() == "openai stream error: \"boom\""`, mirroring what `stream_test.go:152` does for the response-failed path, closes it.

karolpiotrowicz (dismissed):
Everything I raised last round is closed except the `429` path, which is not yours to fix and is at the bottom. I measured each of the rest rather than reading the diff for them. U+2028, U+2029, U+202E, U+200B, NEL, ESC, DEL and the C1 block are all escaped now. `reportsFailure({no status, message "  \n ", code " \t "})` returns false and the turn converts with its output intact. The boundary sits at 255 unmarked, 256 unmarked, 257 marked, and 257 CJK runes come out as 771 valid bytes. Clipping 8 MiB of `A` costs 1259 ns and one allocation, against the 32 MiB transient it used to cost. Every statement the diff adds is executed at least once, which is the thing that was not true last round, when the generic control-character arm I flagged had an execution count of zero.

Deleting `escapeControl` outright was the better call. `%q` is stdlib, it covers Cf and Zl/Zp which the hand-written escaper never did, and the fuzz target asserts on the composed error text, so a field that stops being quoted fails a test rather than shipping. Build, tests, vet, `golangci-lint`, `staticcheck`, `go mod tidy -diff`, race and `govulncheck` are green on both trees, and `apidiff` reports only `ErrResponseFailed: added`.

Three things worth your time. None of them blocks.

## The new fuzz target fails on correct output

[`FuzzClipServerText`'s fourth assertion](https://github.com/google/adk-go/blob/6d753fc5060719696c440eb861f4152008386bc0/model/openaimodel/servertext_test.go#L119-L122) treats a trailing `…` as proof that we truncated. An input of 256 runes or fewer is returned verbatim, so a server message that ends in an ellipsis of its own trips it:

```
$ cat model/openaimodel/testdata/fuzz/FuzzClipServerText/repro
go test fuzz v1
string("rate limited, retrying …")

$ go test -run FuzzClipServerText ./model/openaimodel/
--- FAIL: FuzzClipServerText/repro
    servertext_test.go:121: clipServerText("rate limited, retrying …") =
        "rate limited, retrying …", want no space before the truncation marker
```

The shipped code is right — `CutSuffix` just cannot tell a marker [`clipServerText`](https://github.com/google/adk-go/blob/6d753fc5060719696c440eb861f4152008386bc0/model/openaimodel/response.go#L60-L72) appended from one the input already carried. This cannot turn CI red on its own: there is no `-fuzz` job in `.github/`, no `testdata/fuzz` in the tree, and a 60-second run over a million inputs did not find it, because the function does not branch on whether its input ends in `…` so the fuzzer gets no gradient toward one. But whoever runs `-fuzz` on the target you just added is the person who hits it, and Go writes the input to `testdata/fuzz/` for them. I left the fix on the line.

## A second consumer of the finish-reason change

The description names [the summarizer's non-`Stop` guard](https://github.com/google/adk-go/blob/6d753fc5060719696c440eb861f4152008386bc0/session/compaction/llm_summarizer.go#L263-L264), and there is another consumer. [`response.go:262-272`](https://github.com/google/adk-go/blob/6d753fc5060719696c440eb861f4152008386bc0/model/openaimodel/response.go#L262-L272) now returns `Other` for `cancelled`, `queued`, `in_progress` and a bare `incomplete`, and [`genaimessages.go:368-374`](https://github.com/google/adk-go/blob/6d753fc5060719696c440eb861f4152008386bc0/internal/telemetry/genaimessages.go#L368-L374) maps `FinishReasonOther` to `finishError`. So an ordinary truncation from a compatible endpoint now records as an error in telemetry, and a dashboard keyed on that attribute sees an error rate with no new failures behind it.

I think `Other` is the honest value and `Stop` was the bug, so I am not asking you to change the mapping. This is a decision about what you want the telemetry signal to be, and it deserves a line in the description next to the summarizer one. Nothing tests the pairing, and the import graph hides it — `examples/openai/main.go` is the only in-tree importer, and this consumer is reached by satisfying `model.LLM` at runtime instead.

## The description drifted from the code in the last commit

It still shows the message unquoted:

> A failure now reads `openai: response failed (id "resp_123", code "server_error"): upstream exploded`

[`acefcf44`](https://github.com/google/adk-go/commit/acefcf44f05925870661292a2ac921380cf9a2b5), the last commit, re-pinned every `wantErr` to the quoted form, so the real text now has the message inside `%q`. That paragraph is the one a maintainer updating a matcher would copy, and it is squash-merged into the commit message.

Two smaller things while you are in there. The behaviour-change list does not mention the [`errorEvent` path](https://github.com/google/adk-go/blob/6d753fc5060719696c440eb861f4152008386bc0/model/openaimodel/stream.go#L88-L95), where the message is now quoted, trimmed and capped at 256 runes — the prefix is unchanged so no matcher breaks, but a 300-rune diagnostic that used to arrive whole now arrives cut. And "a body with an `error` object and no `status` at all now counts as failed" is wider than the code: an empty or whitespace-only error object still converts as an ordinary turn, which is the right behaviour. "A non-empty `error` object" says it.

## One thing that is still not yours

The `429` path is unchanged. `isResourceExhausted` in the triggers controller still decides "throttled" by looking for `429` anywhere in `err.Error()`, and `%q` leaves digits alone, so a short provider message containing those digits still drives the retry loop. `%q` can also synthesise it, since `"\x04"` followed by `29` renders as `\x0429`. Separately, I did chase down the other classifier I flagged last round and it is a dead end in your favour: `isResumable` in `base_flow.go` lives inside `RunLive`, which bails unless the model supplies a live client, so no error from this package can reach it.

model/openaimodel/servertext_test.go
wojtas000: Fixed, with your gate — [it now runs only when the trimmed input exceeded the cap](https://github.com/google/adk-go/blob/c6fd1d82aa608f3832111d768e10216955f743b4/model/openaimodel/servertext_test.go#L125-L133), which is exactly `clipServerText`'s own truncation condition, so the assertion neither over- nor under-fires.

Kept it nested rather than an early `return`: this gate applies to one of four assertions in the fuzz body, and a `return` would silently skip anything added after it.

Also [seeded both directions](https://github.com/google/adk-go/blob/c6fd1d82aa608f3832111d768e10216955f743b4/model/openaimodel/servertext_test.go#L101-L111) — the "rate limited, retrying …" false positive, and a clip landing mid-whitespace. That second seed matters: before it, no seed reached the assertion at all, so it had no teeth under a plain `go test`. It does now — dropping the pre-marker `TrimSpace` from `clipServerText` fails four tests.

Passed on the `(string, bool)` signature: it would put a return value in production that no caller reads, only to spare the test a rune count.

model/openaimodel/servertext_test.go
wojtas000: Fixed — [the sentence no longer names an operator](https://github.com/google/adk-go/blob/c6fd1d82aa608f3832111d768e10216955f743b4/model/openaimodel/servertext_test.go#L39-L45), so it cannot go stale the next time the loop is rewritten; it says what the row and its neighbour pin between them instead.

Verified rather than asserted: mutating the boundary to `n == maxServerTextRunes-1` fails this row, and `n == maxServerTextRunes+1` fails the one below it.

model/openaimodel/stream.go
wojtas000: Fixed — confirmed your mutation first (rewording the prefix to `openai: stream error:` left the package green), then [added the row](https://github.com/google/adk-go/blob/c6fd1d82aa608f3832111d768e10216955f743b4/model/openaimodel/servertext_test.go#L328-L339).

Pinned the [bare no-message form](https://github.com/google/adk-go/blob/c6fd1d82aa608f3832111d768e10216955f743b4/model/openaimodel/servertext_test.go#L301-L310) too while I was there. That string predates the PR, but it is unpinned for the same reason and it is one line.

karolpiotrowicz (dismissed):
The fuzz assertion is fixed, and the fix keeps the real check live rather than switching it off. Gating on the input having actually exceeded the cap was the right move — dropping the pre-marker `strings.TrimSpace` in [`clipServerText`](https://github.com/google/adk-go/blob/305ae74e847dfff654a862096e4da0d3200f3c7e/model/openaimodel/response.go#L60-L72) still turns the package red, and on two rows independently:

```
--- FAIL: TestClipServerText/cut_inside_whitespace
    servertext_test.go:85: clipServerText() returned 257 runes, want 251
--- FAIL: FuzzClipServerText/seed#10
    servertext_test.go:131: clipServerText("AAA…AAA          BBBBBBBBBB") =
        "AAA…AAA      …", want no space before the truncation marker
```

The whitespace seed you added is what makes the second one fire under plain `go test`. Both stream-error strings are pinned now as well: rewording either goes red here and stayed green on the previous revision, so that gap is measurably closed.

Everything else in this round is the description, and the code is unchanged since my last pass.

**The worked example still shows the message unquoted.** The pinned expectation at [`response_test.go:112`](https://github.com/google/adk-go/blob/305ae74e847dfff654a862096e4da0d3200f3c7e/model/openaimodel/response_test.go#L112) is `openai: response failed (id "resp_123", code "server_error"): "the model failed to generate a response"` — message inside `%q`. The description still reads `…: upstream exploded`. The description squash-merges into the commit message, so a wrong example there is permanent, and that paragraph is what a maintainer updating a matcher would copy. While you are in there: "a body with an `error` object and no `status`" is wider than [the code](https://github.com/google/adk-go/blob/305ae74e847dfff654a862096e4da0d3200f3c7e/model/openaimodel/response.go#L82-L89), which needs a non-empty one.

**The finish-reason change still has a second consumer the description does not name.** [`response.go:262-272`](https://github.com/google/adk-go/blob/305ae74e847dfff654a862096e4da0d3200f3c7e/model/openaimodel/response.go#L262-L272) returns `Other` for `cancelled`, `queued`, `in_progress` and a bare `incomplete`, and [`genaimessages.go:368-374`](https://github.com/google/adk-go/blob/305ae74e847dfff654a862096e4da0d3200f3c7e/internal/telemetry/genaimessages.go#L368-L374) maps `Other` to `finishError`, so an ordinary truncation from a compatible endpoint records as an error in telemetry. I still think `Other` is the honest value, so this is a line in the description next to the [`llm_summarizer.go:263`](https://github.com/google/adk-go/blob/305ae74e847dfff654a862096e4da0d3200f3c7e/session/compaction/llm_summarizer.go#L263-L264) one rather than a change.

One thing for whoever sequences this: #1359 and #1373 both rewrite the `case "":` arm of `finishReason`, by different means — #1373 through a new `truncated(resp, incompleteEvent)` helper, this PR through a second switch on `resp.Status`. They conflict on `response.go`, `openai_test.go` and `response_test.go`, and in both orders equally, so landing one first does not make the other cheaper. Neither predicate contains the other, so a resolution that takes one side wholesale drops the other PR's fix. Worth a look from someone who has read both once the second one rebases.

Approving on the code. The two description edits are the only thing outstanding and I do not think they need another round.

karolpiotrowicz (approved):

PR #1361, opened 2026-08-19
fix: fence relayed agent content in remoteagent/v2's presentAsUserMessage

prasanna8585:
**Stacked on #1360** (`fix/fence-relayed-agent-content`) — this branch is built on top of that one and depends on it landing first. Until #1360 merges, this PR's diff will show both sets of changes; once #1360 merges and this branch is rebased onto `main`, the diff will narrow to just the change described below.

`presentAsUserMessage` (`agent/remoteagent/v2/utils.go`) wraps this agent's own output as a user-role message before sending it onward to a remote A2A peer that only accepts user-role messages. It duplicates the same unfenced pattern already fixed in `ConvertForeignEvent` (`internal/llminternal/contents_processor.go`, #1360) for local agent-to-agent handoffs: the relayed text is attacker-reachable (whoever talks to this agent steers what it says, and its tool results carry whatever a tool read), and without framing that content is indistinguishable from a genuine instruction to the remote peer's model.

This sweep was missed in the original PR because it only searched `internal/llminternal` for the pattern; a repo-wide search turned up this second, independent implementation in a different package (`agent/remoteagent/v2`).

Exports `QuoteUntrusted` and `ElideQuoteMarkers` from `fencing.go` (already package-private-but-otherwise-exported per that file's own stated rationale) so this cross-package call site can reuse them directly rather than duplicating the fencing logic a third time.

Verified: `agent/remoteagent/v2` and `internal/llminternal` test suites pass with fixtures updated to the new fenced format. `golangci-lint` clean.

### DIFF
--- a/agent/remoteagent/v2/a2a_agent_test.go
+++ b/agent/remoteagent/v2/a2a_agent_test.go
@@ -949,8 +949,8 @@ func TestRemoteAgent_RequestPayload(t *testing.T) {
 					Role: a2a.MessageRoleUser,
 					Parts: a2a.ContentParts{
 						a2a.NewTextPart("hello"),
-						a2a.NewTextPart("For context:"),
-						a2a.NewTextPart(fmt.Sprintf("[%s] said: hi", notRemoteAgentName)),
+						otherAgentPreambleA2APart(),
+						otherAgentA2APart("["+notRemoteAgentName+"] said:", "hi"),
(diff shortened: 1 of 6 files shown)


prasanna8585:
Closing this in favor of folding the same commit into #1360 directly, since this fix depends on fencing.go from that PR and there's no real benefit to tracking it separately while #1360 is still open. See #1360 for the combined change.

PR #1362, opened 2026-08-19
fix(llmagent): scope -httprecord directives so one cassette re-records alone

### LINKED ISSUE #1330: Recording one httprr cassette re-records every cassette in the package
### What happens

`AGENTS.md` documents the recording procedure as:

> To (re)record a package's traffic, supply real credentials (e.g. `GOOGLE_API_KEY`) and run `go generate ./<pkg>/...` (it runs `go test -httprecord=…`); commit the updated `testdata/*.httprr`.

`-httprecord` takes a **regexp matched against the cassette file path**, not a test name filter (`internal/httprr/rr.go:34`, "re-record traces for files matching `regexp`"). Three of the four directives in the repo are effectively "match everything":

| Directive | Location | Cassettes it matches |
|---|---|---|
| `-httprecord=Test` | `agent/llmagent/llmagent_test.go:48` | all 20 in the package (416K) |
| `-httprecord=TestDelegation` | `agent/llmagent/llmagent_delegation_test.go:66` | 10 of those 20 |
| `-httprecord=.*` | `agent/remoteagent/v2/a2a_e2e_test.go:17` | all 4 |
| `-httprecord=testdata/.*\.httprr` | `model/gemini/gemini_test.go:35` | all 4 |

So `go generate ./agent/llmagent/...` runs both directives in that package and re-records all 20 cassettes, with the 10 delegation ones recorded twice in one invocation.

### Why it matters

Someone who needs to refresh one cassette gets a diff touching every recorded interaction in the package, against live models whose responses differ run to run. The reviewer cannot tell the intended change from the churn, and an unrelated test can start failing because its cassette was silently replaced with a different model response. The largest package is the one with the broadest directive.

It also makes the documented procedure the wrong thing to run in almost every case. The safe invocation is the narrow one the docs do not mention:

    go test ./agent/llmagent/ -httprecord='TestFunctionTool\.httprr' -run TestFunctionTool

### Suggestion

Two parts, either useful on its own:

1. Narrow the directives so `go generate` on a package records what that package's directive names, rather than everything. `-httprecord=Test` in particular should either be scoped or removed, since the delegation file already carries its own.
2. Correct `AGENTS.md` to describe recording a single cassette as the normal case, and say that `-httprecord` matches file paths rather than test names. The current wording reads as if it were a `-run` filter.

Pre-existing, unrelated to any open PR. Found while reviewing the context-compaction stack, where adding one cassette to `agent/llmagent` put the other 19 in the blast radius of the documented command.

<sub>Filed from CloudCode session `ses_056950d6dffehtiYaTt2eD2VOa`.</sub>

chethanuk:

- Closes: #1330

**Problem:**

`-httprecord` is a regexp matched against the cassette's file path, not a `-run` test-name filter. `agent/llmagent` declares two of them: `-httprecord=Test` in `llmagent_test.go` matches all 20 cassettes in the package, and `-httprecord=TestDelegation` in `llmagent_delegation_test.go` matches 10 of those same 20. So the documented `go generate ./agent/llmagent/...` re-records the entire package and records the 10 delegation cassettes twice in one invocation.

Anyone refreshing one cassette gets a diff touching every recorded interaction in the package, against live models whose responses differ run to run. The intended change is unreadable next to the churn, and an unrelated test can start failing because its cassette was silently replaced with a different model response. `AGENTS.md` presents that whole-package command as the way to re-record, so the wrong thing to run is the documented thing.

**Solution:**

Scope each directive to the cassettes its own file owns, anchored on the `testdata/` prefix and the `.httprr` suffix so no name matches as a substring:

```
llmagent_test.go:            -httprecord=^testdata[/\\]Test(FunctionTool|LLMAgent|ToolCallback).*\.httprr$
llmagent_delegation_test.go: -httprecord=^testdata[/\\]TestDelegation_.*\.httprr$
```

Rewrite the recording section of `AGENTS.md`: re-recording a single cassette is the normal case and needs the name in both `-run` and `-httprecord`, `-httprecord` matches file paths rather than test names, and `go generate` on a package is safe because its directives no longer overlap.

Add `TestHTTPRecordDirectivesPartitionCassettes` (`internal/httprr_directives_test.go`) so the property is enforced rather than just fixed. It walks the module, groups every `//go:generate go test -httprecord=…` by package, and requires that each `testdata/*.httprr` is claimed by exactly one directive and each directive claims at least one cassette. The second half catches the quieter failure: a pattern that has drifted away from its cassettes, where `go generate` appears to succeed while recording nothing.

### Testing Plan

**Unit Tests:**

`go test -count=1 ./internal/ ./agent/llmagent/` passes; the new test reports 7 directives checked across 6 packages. Putting the two old patterns back makes it fail on each of the 10 doubly-claimed delegation cassettes, naming both directives that claim it.

Not run, and it should not be: recording needs live credentials and would rewrite the committed cassettes, which is the problem this fixes. The directives are `go:generate` comments, so nothing here can change replay behavior, and `agent/llmagent` still passes offline against the unchanged cassettes.

### Checklist

### Additional context

The single-directive packages (`agent/remoteagent/v2`, `model/gemini`) keep patterns that match all of their cassettes. For a package with one directive that already partitions, so the new test accepts them and this change leaves them alone; re-recording one cassette there still means naming it on the command line instead of running `go generate`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -143,9 +143,19 @@ See [Multi-Module Development](CONTRIBUTING.md#multi-module-development) in
 - Tests run **offline by default**: LLM HTTP traffic is replayed from
   `testdata/*.httprr` via `internal/httprr`. Never add live model or network
   calls to tests.
-- To (re)record a package's traffic, supply real credentials (e.g.
-  `GOOGLE_API_KEY`) and run `go generate ./<pkg>/...` (it runs
-  `go test -httprecord=…`); commit the updated `testdata/*.httprr`.
+- `-httprecord` takes a **regexp matched against the cassette's file path**,
+  not a `-run` test-name filter. Keep it as narrow as the set of cassettes you
+  mean to replace: recording against a live model produces a different response
+  every time, so a broad pattern rewrites unrelated cassettes and buries the
+  intended change.
+- To re-record **one** cassette — the normal case — supply real credentials
+  (e.g. `GOOGLE_API_KEY`) and name it in both flags:
+  `go test ./<pkg>/ -run TestFoo -httprecord='TestFoo\.httprr$'`.
+  Commit only that `testdata/*.httprr`.
+- To re-record a whole package, run `go generate ./<pkg>/...`; each package's
+  `//go:generate go test -httprecord=…` directives are scoped so that every
+  cassette is recorded exactly once (enforced by
+  `TestHTTPRecordDirectivesPartitionCassettes` in `internal`).
 - Prefer table-driven tests; shared helpers live in `internal/testutil`.
 
 ## Boundaries
--- a/agent/llmagent/llmagent_delegation_test.go
+++ b/agent/llmagent/llmagent_delegation_test.go
@@ -22,9 +22,14 @@
 // changing an Instruction, tool declaration, or agent shape, delete
 // the corresponding .httprr file and run:
 //
-//	GEMINI_API_KEY=... go test ./agent/llmagent/ -httprecord=TestDelegation_NN
+//	GEMINI_API_KEY=... go test ./agent/llmagent/ -run TestDelegation_NN \
+//	    -httprecord='TestDelegation_NN.*\.httprr$'
 //
-// To re-record everything: -httprecord=TestDelegation.
+// -httprecord is a regexp matched against the cassette's file path, not a
+// test-name filter, so keep it as narrow as the set of cassettes you mean to
+// replace. The go:generate directive below re-records every delegation
+// cassette and nothing else; `go generate ./agent/llmagent/...` also runs the
+// directive in llmagent_test.go, so it re-records the whole package.
 //
 // The tests generate *.events.yaml file to visualize delegation and overall
 // flow of execution.
@@ -63,7 +68,7 @@ import (
 
 const delegationModelName = "gemini-3.5-flash"
 
-//go:generate go test -httprecord=TestDelegation
+//go:generate go test -httprecord=^testdata[/\\]TestDelegation_.*\.httprr$
 
 // newDelegationModel returns a Gemini model whose HTTP transport is
 // the httprr record/replay wrapper, scoped to a trace file derived
--- a/agent/llmagent/llmagent_test.go
+++ b/agent/llmagent/llmagent_test.go
@@ -45,7 +45,7 @@ const (
 	message   = "Handle the requests as specified in the System Instruction."
 )
 
-//go:generate go test -httprecord=Test
+//go:generate go test -httprecord=^testdata[/\\]Test(FunctionTool|LLMAgent|ToolCallback).*\.httprr$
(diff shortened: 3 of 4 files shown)


baptmont:
Approving — this is right, and the guard test is the part that makes it worth more than a one-off patch. I verified the pieces rather than taking them on trust:

- `-httprecord` really is a path regexp, not a test filter: `Recording(file string)` compiles the flag and does `re.MatchString(file)` (`internal/httprr/rr.go:121-131`).
- The overlap is real: 19 cassettes in `agent/llmagent`, `-httprecord=Test` claims all 19, `-httprecord=TestDelegation` claims 10 of the same. (Body says 20; it is 19.)
- The new patterns partition exactly — 9 + 10 = 19, nothing unclaimed, nothing double-claimed.
- `TestHTTPRecordDirectivesPartitionCassettes` passes (7 directives, 6 packages) and, with the old patterns restored, fails naming each doubly-claimed cassette and both claimants.
- No runtime risk: these are `go:generate` comments, and `agent/llmagent` still replays green offline.

The dead-pattern half of the test is a nice catch — a directive that has drifted away from its cassettes would otherwise let `go generate` report success while recording nothing.

One scoping point, and since I filed #1330 I should be the one to raise it. The issue covers two distinct problems and this fixes one of them:

1. **Double-recording** — 10 cassettes recorded twice in a single `go generate`. Specific to `agent/llmagent`, fixed here, now guarded.
2. **Over-broad recording** — one directive claiming every cassette in its package. Still present, and the new test permits it by construction.

Five directives remain effectively "match everything":

| --- | --- | --- |
| `-httprecord=.*` | `tool/functiontool` | 1 |

`go generate ./tool/mcptoolset/...` still rewrites all 11 against live models, and the guard is happy: with one directive per package each cassette is claimed exactly once, and the test only rejects 0 or >1. So it enforces partition, not narrowness — which are easy to conflate.

I do not think that is a defect in the approach. A single-test-file package has nothing to partition, and the real remedy for the one-cassette case is the command you added to AGENTS.md, which works everywhere regardless of directives. But it does mean two small changes before merge:

1. Please switch `Closes: #1330` to `Related: #1330`. As written this auto-closes an issue whose headline complaint — refreshing one cassette rewrites the package — is still true for 23 cassettes across five packages. I would rather leave it open to track narrowing those, or at least close it deliberately with a note.
2. In AGENTS.md, "each package's `//go:generate` directives are **scoped** so that every cassette is recorded exactly once" is true but reads as "narrow", and `.*` is the opposite. Something like "do not overlap" describes what the test actually enforces.

Neither blocks the merge. Worth noting why this matters more than it looks: `TestDelegation_04` was flaking this week, and the tempting fix was to re-record its cassette. Under the old directives that would have rewritten all 19 against live models and buried the real bug (#1367) in churn.

baptmont (approved):

PR #1368, opened 2026-08-20
test(llminternal): cover the nil RunConfig branch in callLLM

baptmont:

- Related: #586, #1336, #881

**Problem:**

#1336 fixed the nil dereference in `callLLM` by reading the streaming mode from the invocation context behind a nil guard. Its test pins the *migration* — that the value comes from `ctx.RunConfig()` rather than `runconfig.FromContext(ctx)` — but not the *guard*. Deleting the guard leaves the whole suite green:

```go
// mutation applied to base_flow.go
useStream := ctx.RunConfig().StreamingMode == agent.StreamingModeSSE
```
```
```

That branch is reachable in production: `runner.RunLive` builds its invocation context without a `RunConfig`, so `ctx.RunConfig()` is nil there by construction. An unpinned guard is one refactor away from restoring the panic in #586.

**Solution:**

Replace the single-case test with a table over `SSE`, `none`, and `nil`, reusing the existing `recordStreamModel`. The old test is fully subsumed by the `SSE streams` case, so it is removed rather than left alongside.

Both halves of the expression are now load-bearing:

| mutation to `base_flow.go` | result |
| --- | --- |
| remove `if rc := ctx.RunConfig(); rc != nil` | **FAIL** — `nil run config does not stream` panics |
| replace comparison with `useStream = true` | **FAIL** — `none does not stream` |

**Credit:**

The underlying bug was diagnosed independently by @Haihan-Jiang in #881 three months before #1336, with the same fix. That PR went unreviewed and was superseded; they are credited as `Co-authored-by` here.

### Testing Plan

**Unit Tests:**

```
$ go test -race -count=1 ./internal/llminternal/...
ok      google.golang.org/adk/v2/internal/llminternal    2.302s
ok      google.golang.org/adk/v2/internal/llminternal/googlellm 1.178s

$ go build -mod=readonly work        # exit 0
$ go mod tidy -diff                  # no output
$ gofmt -l internal/llminternal/     # no output
$ go vet ./internal/llminternal/     # clean
```

Full workspace run was green apart from `TestCompat_A2ACleanupPropagation`, the known flake in `agent/remoteagent` (passes on re-run; this change touches one test file in an unrelated package).

This is test-only — no production code changes, no API surface change.


internal/llminternal/base_flow_test.go
```diff
@@ -872,6 +872,61 @@ func (m *alwaysThinkingModel) GenerateContent(ctx context.Context, req *model.LL
 	}
 }
 
+// TestCallLLMStreamingModeFromRunConfig pins how callLLM derives the streaming
+// flag, including the nil case (issue #586).
+//
+// None of these store a runconfig in the Go context, as when agent.Run() is
+// invoked directly and the runner is bypassed, so runconfig.FromContext would
+// return nil; the streaming mode is read from the invocation context instead.
+// The nil RunConfig case is not hypothetical: RunLive builds its invocation
```
karolpiotrowicz, line 881: This reachability claim does not hold. `callLLM` has exactly one non-test caller, `base_flow.go:633` inside `Flow.Run`, and `Flow.RunLive` never reaches it — it reads `runconfig.FromContext(ctx)` and returns `"live run config not found"` when that is nil. `runner.RunLive` does build an invocation context with no `RunConfig`, so half the sentence is observing something real, but that context flows into `Flow.RunLive`, not into `callLLM`.

The first paragraph already gives the accurate justification — an `agent.InvocationContext` built outside the runner and passed to `Run` directly, which is issue #586. The risk in leaving this as it stands is that someone checking the cited path finds it cannot reach `callLLM` and concludes the guard is dead code.

internal/llminternal/base_flow_test.go
```diff
@@ -872,6 +872,61 @@ func (m *alwaysThinkingModel) GenerateContent(ctx context.Context, req *model.LL
...
+		{
+			name:      "nil run config does not stream",
+			runConfig: nil,
+			want:      false,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			m := &recordStreamModel{}
```
karolpiotrowicz, line 909: Worth knowing while restoring this: `mockModelForTest` already lives in this package at [`base_flow_telemetry_test.go:39`](https://github.com/google/adk-go/blob/b84e6f72af70cd66560023c4f76cf983e9cbb8e8/internal/llminternal/base_flow_telemetry_test.go#L39) and carries a `generateContent func(ctx, req *model.LLMRequest, stream bool)` hook that receives exactly the value under test. Using it needs no new type, and a hook closure can record the call count that line 923 currently has no way to check.

One constraint if you do re-add a dedicated spy instead: it must not yield `(nil, nil)`, or `resp.Content` in `callLLM` dereferences the embedded nil `*model.LLMResponse`. Yielding nothing is safe.

internal/llminternal/base_flow_test.go
```diff
@@ -872,6 +872,61 @@ func (m *alwaysThinkingModel) GenerateContent(ctx context.Context, req *model.LL
...
+			})
+
+			req := &model.LLMRequest{}
+			for _, err := range f.callLLM(ctx, req, map[string]any{}, map[string]int64{}) {
+				if err != nil {
+					t.Fatalf("callLLM() error = %v, want nil", err)
+				}
+			}
+
+			if m.stream != tc.want {
```
karolpiotrowicz, line 923: This assertion cannot tell "the flow computed `false`" from "the flow never called the model", because the spy's `stream` field starts at `false` and nothing records the invocation. Both `want: false` rows are affected — only the SSE row proves the model was reached.

I confirmed it by making `callLLM` skip the model entirely: the SSE row fails, and the `none` and `nil` rows both pass. Seven other tests in the package do catch that particular regression, so the exposure is bounded, but the two rows are not load-bearing on their own terms. A `calls int` on the spy and a `m.calls == 1` check in every row would fix it.

internal/llminternal/base_flow_test.go
```diff
@@ -872,6 +872,61 @@ func (m *alwaysThinkingModel) GenerateContent(ctx context.Context, req *model.LL
...
+				t.Errorf("GenerateContent received stream=%v, want %v", m.stream, tc.want)
+			}
+		})
+  }
```
karolpiotrowicz, line 927: This brace is two spaces of indentation plus a trailing space, where gofmt wants a single tab. `gofmt -l internal/llminternal/` prints this file on the head and is silent on the merge base, so the change introduces it.

It matters beyond tidiness: the undefined-symbol failure is currently masking the formatters, so `lint (.)` will still be red after that is fixed unless this goes too.

karolpiotrowicz (changes requested):
The two red checks on this head, `lint (.)` and `test (.)`, are both caused by this change rather than inherited from a red base. I ran the gates on four trees — the merge base, this head, current `main`, and this head merged into current `main` — and the base and `main` are clean on every one.

## The build break

[`base_flow_test.go:909`](https://github.com/google/adk-go/blob/b84e6f72af70cd66560023c4f76cf983e9cbb8e8/internal/llminternal/base_flow_test.go#L909) calls `&recordStreamModel{}`, and that type does not exist anywhere in the tree. `go vet`, `go test`, `golangci-lint` and `staticcheck` all stop at the same line, so no test in `internal/llminternal` runs at all — not just the new one.

```
$ go vet ./...
# google.golang.org/adk/v2/internal/llminternal [google.golang.org/adk/v2/internal/llminternal.test]
internal/llminternal/base_flow_test.go:909:10: undefined: recordStreamModel
```

The helper arrived with #1336 ([0510fef](https://github.com/google/adk-go/commit/0510fef289ad7f7e5018f2f76977d7232e979987)) and was deleted five days before this head by #1290 ([11748f1](https://github.com/google/adk-go/commit/11748f1c4b5c480f9e2a0d484dc81e171b8258c0)), which removed both `recordStreamModel` and `TestCallLLMStreamingModeWithoutContextRunConfig`. The branch was rebased past that deletion, so it carries a pre-#1290 view of this file.

That same rebase is why the motivation below no longer holds — one cause, two symptoms.

## The stated motivation does not hold against this branch's base

The description says deleting the guard leaves the suite green. That was true when #1290 landed, and it is not true now, because #1290 also added `TestRun_ThoughtOnlyTurnsTerminate`, which builds its invocation context without a `RunConfig` and therefore reaches the guard incidentally.

I checked both readings of "deleting the guard" on the merge base, with no part of this PR applied. Dereferencing without the guard makes `TestRun_ThoughtOnlyTurnsTerminate` panic. Removing the `if` block and leaving `useStream := false`, which is nil-safe, turns `TestParallelFunctionCalls` red instead.

<details><summary>Guard deleted on the merge base, this PR not applied</summary>

```
$ cd <worktree at 9db1ff121df37250fae9e3ad1c76d3fb86661b4f>
# base_flow.go:836-839 replaced with:
#   useStream := ctx.RunConfig().StreamingMode == agent.StreamingModeSSE
$ go test -count=1 ./internal/llminternal/...
--- FAIL: TestRun_ThoughtOnlyTurnsTerminate (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered, repanicked]
FAIL	google.golang.org/adk/v2/internal/llminternal	0.061s
```
</details>

This does not make the change pointless. An incidental pin sitting inside a thought-only-turn test disappears the moment somebody gives that test a `RunConfig`, and a named test does not. But the claim in the description is worth correcting, since it is the argument for the PR.

## What the test does and does not pin

To answer this at all I had to re-add the missing helper locally, so everything below describes the test as it would behave once it compiles, not as submitted.

Both mutations the description names do work: removing the guard fails the nil row, and forcing `useStream = true` fails the `none` row. Flipping the comparison to `StreamingModeNone` fails two rows as well, which the description does not claim.

One mutation escapes the whole package suite, not just this test:

```go
// base_flow.go:838
useStream = rc.StreamingMode != agent.StreamingModeNone
```

All three rows pass, `go test ./internal/llminternal/...` passes, and an invocation carrying a plain `&agent.RunConfig{}` silently starts streaming. `StreamingMode` is a bare string with only `"none"` and `"sse"` named, so the zero value `""` is the ordinary case and the table never exercises it. A fourth row for `&agent.RunConfig{}` expecting `false` closes it, and it is the addition I would prioritise.

I have left the smaller points inline.

With the helper restored and `gofmt -w` applied, `go vet`, `golangci-lint` and `go test ./...` are all clean locally, 72 packages ok. Fixing the undefined symbol on its own is not enough — the typecheck failure is currently masking the formatter, so `lint (.)` stays red until the brace on line 927 is fixed too.

PR #449, opened 2025-12-30
fix: Only last state change for multiple tool calls is being committed to state

majiayu000:
Fixes #354

## Changes
- Use `maps.Copy` instead of assignment in `mergeEventActions` to properly merge state deltas from multiple tool calls
- Initialize `base.StateDelta` map before copying if it's nil
- Add comprehensive tests for `mergeEventActions` function

Issue #354, opened 2025-11-25
Only last state change for multiple tool calls is being committed to state


verdverm:
I have some custom `functiontool`s and the LLM is calling it in batch via message parts. In those `functiontool`s, I update the state, which is what the docs say to do: https://google.github.io/adk-docs/sessions/state/#a-warning-about-direct-state-modification

It looks like only the last tool call's state is being applied to the delta. I have confirmed the tools is called many times in one turn

### Example state modifying `functiontool`

```go
const CacheRemoveTool = "cache_remove"
const CacheRemoveDesc = `
adds or overwrites an entry in your working key/value cache and context
`

type CacheRemoveArgs struct {
	Key string `json:"key"` // path to a directory
}
type CacheRemoveResult struct {
	Key    string `json:"key"`    // path to a directory
	Status string `json:"status"` // "ok" or "error"
	Error  string `json:"error,omitempty"`
}

func NewCacheRemove() (tool.Tool, error) {
	handler := func(ctx tool.Context, input CacheRemoveArgs) (CacheRemoveResult, error) {
		k := fmt.Sprintf("cache:%s:%s", ctx.AgentName(), input.Key)
		fmt.Println("CACHE.REMOVE!", k)

		err := ctx.State().Set(k, nil)
		if err != nil {
			return CacheRemoveResult{Status: "error", Key: input.Key, Error: err.Error()}, err
		}
		return CacheRemoveResult{Status: "ok", Key: input.Key}, nil
	}
	return functiontool.New(functiontool.Config{
		Name:        CacheRemoveTool,
		Description: strings.TrimSpace(CacheRemoveDesc),
	}, handler)
}
```

### Event with bulk `functionCall`

```json
{
  "Content": {
    "parts": [
      {
        "text": "That is an insightful observation. Keeping the key with a marker like \"no relevant\" is a good way to maintain a memory of explored files without cluttering the context with content.\n\nI will now attempt to update a batch of keys using `cache_write` to see if it suffers from the same \"only the last command commits\" bug. I will use the marker value `\"no relevant\"` for the keys that currently show `<no value>` but have been examined.\n\nI will use the following keys for the test:\n1.  `package.json`\n2.  `src/main.tsx`\n3.  `src/App.tsx`\n4.  `src/vscodeApi.ts` (This will be the final command to see if it's the only one that persists)\n\n"
      },
      {
        "functionCall": {
          "id": "adk-3caaaeab-b818-4967-8a06-cf57908d0a72",
          "args": {
[...]

PR #1370, opened 2026-08-20
fix(llminternal): merge ArtifactDelta across parallel function responses

### LINKED ISSUE #493: Only first artifact delta for multiple tool calls is being committed to state
**Describe the bug**
When multiple function responses are returned in a single turn, the logic in `internal/llminternal/base_flow.go` `mergeEventActions` ignores artifact deltas. The result of this is that only the first non-nil artifact delta is preserved in the yielded event and all others are discarded.

**To Reproduce**
1. Make an LLM request which returns multiple function calls that produce artifact deltas.
2. The yielded `event.Actions.ArtifactDelta` will only include artifact deltas from the first function call

**Expected behavior**
The python adk uses deep merge, resulting in "last-write-wins" where the same artifact is updated multiple times. Given the intended use-case for artifact versioning, "highest-version-wins" seems more appropriate.

As an example if separate function calls produce updates to an artifact, `a`, with versions 1, 3, and 2, the python ADK will include `a: 2`, however returning `a: 3` preserves more information.

 - OS: macOS 26
 - Go version: 1.25
 - ADK version: 0.3.0

**Model Information:**
 - Reproduced with `gemini-3-flash-preview` and `gemini-3-pro-preview`

**Additional context**
Essentially the same as https://github.com/google/adk-go/issues/354 but for artifact rather than state deltas.

shoemoney:

- Closes: #493
- Related: #354 (the same defect for `StateDelta`, fixed in #449)

**Problem:**

`mergeEventActions` in `internal/llminternal/base_flow.go` merges `StateDelta`,
`SkipSummarization`, `TransferToAgent`, `Escalate` and `RequestedToolConfirmations`,
but never handled `ArtifactDelta` at all. When several tool calls in one turn each
save an artifact, only the first delta survives into the yielded event — every
subsequent one is silently discarded.

This is the artifact-side twin of #354, which was fixed for state deltas in #449.

**Solution:**

Merge `ArtifactDelta` by **highest version per artifact name**.

`ArtifactDelta` is `map[string]int64` mapping file name to version, and artifact
versions are monotonically increasing per file — `artifact/inmemory.go` `find()`
scans from `math.MaxInt64` downward and returns the first hit as "the latest one".
So the larger value is always the more recent write, and merging by max means a
parallel call that saved an older version cannot mask a newer one recorded by a
sibling call in the same turn.

**Deliberate divergence from the Python ADK, flagged for reviewers:** Python's
`merge_parallel_function_response_events` deep-merges into last-write-wins. For
versioned artifacts that loses information — given writes of `a:1`, `a:3`, `a:2`,
Python yields `a:2` while max-wins yields `a:3`. The issue reporter made this case
explicitly and I agree with it, but it *is* a behavioural difference from Python
parity. If the team prefers strict parity, changing `mergeArtifactDelta` to
unconditional assignment is a one-line change and I'm happy to switch it.

### Testing Plan

**Unit Tests:**

Two tests added to `internal/llminternal/base_flow_test.go`, alongside the existing
`TestMergeEventActions` introduced by #449:

- `TestMergeEventActionsArtifactDelta` — table test: nil base, disjoint names,
  same name with a lower incoming version, same name with a higher incoming
  version, and nil `other`.
- `TestMergeParallelFunctionResponseEventsArtifactDelta` — reproduces #493
  directly: three sequential merges must retain all three artifacts and keep the
  highest version of the one written twice.

Both were verified to **fail before the fix and pass after it**. Failing output on
unpatched `base_flow.go`, with the test file already in place:

```
--- FAIL: TestMergeEventActionsArtifactDelta (0.00s)
    base_flow_test.go:860: mergeEventActions() ArtifactDelta mismatch (-want +got):
          map[string]int64(
        -     {"a.txt": 1},
        +     nil,
    base_flow_test.go:860: mergeEventActions() ArtifactDelta mismatch (-want +got):
          map[string]int64{
              "a.txt": 1,
        -     "b.txt": 4,
--- FAIL: TestMergeParallelFunctionResponseEventsArtifactDelta (0.00s)
FAIL    google.golang.org/adk/v2/internal/llminternal   0.012s
```

After the fix:

```
--- PASS: TestMergeEventActionsArtifactDelta (0.00s)
--- PASS: TestMergeParallelFunctionResponseEventsArtifactDelta (0.00s)
ok      google.golang.org/adk/v2/internal/llminternal   0.011s
```

Full run across the affected trees, identical to the pre-change baseline on the
same commit (13 packages ok, 0 failures either side):

```
$ go test ./internal/llminternal/... ./session/... ./agent/...
ok  google.golang.org/adk/v2/internal/llminternal/googlellm
ok  google.golang.org/adk/v2/session
ok  google.golang.org/adk/v2/session/database
ok  google.golang.org/adk/v2/session/vertexai
... (13 ok, 0 FAIL)

$ go vet ./internal/llminternal/     # clean
$ gofmt -l internal/llminternal/     # clean
```

Not run — this is a pure in-memory merge function with no I/O, and the unit tests
exercise the exact code path the issue describes. Reproducing end-to-end requires a
live model emitting multiple parallel tool calls that each save an artifact; the
issue author reproduced it that way against `gemini-3-flash-preview` and
`gemini-3-pro-preview`.

### Checklist

- [ ] I have manually tested my changes end-to-end. — see note above

### Additional context

The diff is **20 additions, 0 deletions**. An earlier draft also removed the
`// TODO add similar logic for state` line directly below the `StateDelta` block,
on the assumption this change resolved it. I checked and that comment was added
*after* #449 merged, so it is live rather than stale and I could not prove it
refers to artifacts — it is left untouched. If it does refer to this, it is safe
to drop.

Not addressed here, deliberately: `mergeEventActions` still overwrites
`TransferToAgent` on a last-write-wins basis, which has the same
"parallel calls silently clobber each other" shape. That is a separate concern
with different semantics and belongs in its own issue and PR.

### DIFF
--- a/internal/llminternal/base_flow.go
+++ b/internal/llminternal/base_flow.go
@@ -1406,6 +1406,9 @@ func mergeEventActions(base, other *session.EventActions) *session.EventActions 
 	if other.StateDelta != nil {
 		base.StateDelta = deepMergeMap(base.StateDelta, other.StateDelta)
 	}
+	if other.ArtifactDelta != nil {
+		base.ArtifactDelta = mergeArtifactDelta(base.ArtifactDelta, other.ArtifactDelta)
+	}
 	// TODO add similar logic for state
 	if other.RequestedToolConfirmations != nil {
 		if base.RequestedToolConfirmations == nil {
@@ -1414,6 +1417,23 @@ func mergeEventActions(base, other *session.EventActions) *session.EventActions 
 		maps.Copy(base.RequestedToolConfirmations, other.RequestedToolConfirmations)
 	}
 	return base
+}
+
+// mergeArtifactDelta merges src into dst, keeping the highest version for each
+// artifact name. Artifact versions are monotonically increasing per file, so the
+// larger value is always the more recent write. Merging by max rather than
+// last-write-wins means a parallel tool call that saved an older version cannot
+// mask a newer one recorded by a sibling call in the same turn.
+func mergeArtifactDelta(dst, src map[string]int64) map[string]int64 {
+	if dst == nil {
+		dst = make(map[string]int64, len(src))
+	}
+	for name, version := range src {
+		if existing, ok := dst[name]; !ok || version > existing {
+			dst[name] = version
+		}
+	}
+	return dst
 }
 
 func deepMergeMap(dst, src map[string]any) map[string]any {
--- a/internal/llminternal/base_flow_test.go
+++ b/internal/llminternal/base_flow_test.go
@@ -813,3 +813,70 @@ func TestCallLLMStreamingModeWithoutContextRunConfig(t *testing.T) {
 		t.Errorf("GenerateContent received stream=%v, want true for SSE streaming mode", m.stream)
 	}
 }
+
+func TestMergeEventActionsArtifactDelta(t *testing.T) {
+	tests := []struct {
+		name  string
+		base  *session.EventActions
+		other *session.EventActions
+		want  map[string]int64
+	}{
+		{
+			name:  "base has no artifact delta",
+			base:  &session.EventActions{},
+			other: &session.EventActions{ArtifactDelta: map[string]int64{"a.txt": 1}},
+			want:  map[string]int64{"a.txt": 1},
+		},
+		{
+			name:  "disjoint artifacts are both preserved",
+			base:  &session.EventActions{ArtifactDelta: map[string]int64{"a.txt": 1}},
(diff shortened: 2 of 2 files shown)


baptmont:
Closing as a duplicate of #474, which is the canonical fix for #493 and is approved and green.
Nothing wrong with this patch, you diagnosed it correctly and independently landed on the same merge strategy #474 uses, highest-version-wins rather than adk-python's last-writer-wins.
Thanks for picking it up, an older open PR for the same issue is not easy to spot.

PR #1372, opened 2026-08-21
fix(openaimodel): serialize assistant history as text

Blinkuu:

**Problem:**

`openaimodel` serializes every text message as an `input_text` content part. On a follow-up turn, model history is normalized to the `assistant` role, but the OpenAI Responses API rejects `input_text` for assistant content:

```text
HTTP 400
code="invalid_value"
param="input[1].content[0]"
```

This makes the first model turn succeed and the second turn fail.

**Solution:**

Serialize assistant history through `EasyInputMessage` string content. User, system, and developer messages keep their existing typed input content. This uses the SDK's role-aware easy-message representation without inventing response item IDs or statuses.

### Testing Plan

**Unit Tests:**

- [x] Added `TestBuildOpenAIParams_AssistantHistory`, covering user → assistant → user history and asserting assistant string content.
- [x] `go test -race ./model/openaimodel -count=1`
- [x] `go test -race -mod=readonly -count=1 -shuffle=on work` — 73 packages passed; 83 packages had no tests.
- [x] `golangci-lint run` in the root module — 0 findings.
- [x] `go mod tidy -diff` in the root and `plugin/agentanalytics` modules — no diff.

A temporary local harness called `NewModel` against the OpenAI Responses API with user → assistant → user contents:

```text
go test ./model/openaimodel -run '^TestLiveAssistantHistory$' -count=1 -timeout=60s
ok google.golang.org/adk/v2/model/openaimodel 1.795s
```

The same two-turn flow was then exercised through a downstream live meeting agent: both model turns completed and the agent remained connected. No prompts, responses, credentials, or meeting content are included here.

### Checklist

- [x] I have commented my code where needed.
- [x] I have added tests that prove the fix is effective.
- [x] I have manually tested the change end-to-end.
- [x] No dependent downstream-module changes are required.

### Additional context

No matching open issue or pull request was found. The unrelated `plugin/agentanalytics` lint currently reports the existing `reflect.Ptr should be inlined` finding at `json_formatter.go:103`; this branch does not modify that module.

### DIFF
--- a/model/openaimodel/request.go
+++ b/model/openaimodel/request.go
@@ -159,6 +159,28 @@ func newMessage(role genai.Role, texts []string) (*responses.EasyInputMessagePar
 	if err != nil {
 		return nil, err
 	}
+	if msgRole == responses.EasyInputMessageRoleAssistant {
+		var content strings.Builder
+		for _, txt := range texts {
+			if strings.TrimSpace(txt) == "" {
+				continue
+			}
+			if content.Len() > 0 {
+				content.WriteByte('\n')
+			}
+			content.WriteString(txt)
+		}
+		if content.Len() == 0 {
+			return nil, nil
+		}
+		return &responses.EasyInputMessageParam{
+			Role: msgRole,
+			Type: responses.EasyInputMessageTypeMessage,
+			Content: responses.EasyInputMessageContentUnionParam{
+				OfString: param.NewOpt(content.String()),
+			},
+		}, nil
+	}
 	contentList := make(responses.ResponseInputMessageContentListParam, 0, len(texts))
 	for _, txt := range texts {
 		if strings.TrimSpace(txt) == "" {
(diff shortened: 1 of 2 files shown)


jjsasha63:
Thanks for this, and sorry it sat unreviewed.

This is the same bug as #1197, which is now fixed in main by #1427. That
one took the output-message route from #1205 and has been verified live
against gpt-5.4-nano, so I'm closing this as already fixed.

Two notes on your approach, because it was a reasonable call. The string
content form is valid, and it avoids the id/status question that made
#1205 look risky. That question turned out to be moot: the SDK tags id
omitzero, so an unset id is dropped from the request, and status is set
to completed. The one thing the string form loses is part boundaries —
multiple text parts get joined with \n instead of staying separate
content items, which main preserves.

If you hit anything else in openaimodel, please do open an issue first so
we can point you at work already in flight.

PR #1373, opened 2026-08-21
fix(model/openaimodel): correct finish reason, logprobs and tool calls when a stream ends

hanorik:
**Problem**

`model/openaimodel` reads the terminal response of a streamed turn — the
`response.created` / `.completed` / `.incomplete` event — to decide how that turn
ended. Five things about how it does that are wrong:

1. `response.incomplete` with no reason reports `STOP`. `IncompleteDetails.Reason`
   is not `api:"required"` in openai-go and `Status` is never read, so a turn the
   provider declared truncated closes as a clean stop and a caller that retries on
   non-STOP accepts a partial answer as final.
2. Logprobs come off the terminal event while content comes from the deltas, so a
   response can carry token probabilities for text it does not contain.
3. `{"type":"response.completed"}` with no response object decodes to a zero value
   and latches, ending the turn with empty `ModelVersion` and zero usage while
   locking out a well-formed event behind it.
4. The rebuild from the terminal event is gated on `final == nil`, which is false
   as soon as anything reaches the aggregator — including a function call it
   dropped. That turn closes empty while blocking returns the tool call.
5. A content-filtered turn reaches the caller with no block reason at all.
   `promptFeedback` is dead code: `Genai2LLMResponse` reads `PromptFeedback` only
   when there are no candidates, and `convertResponse` always emits one.

**Solution**

- `terminalEvent` replaces the two loose latch variables and records the event's
  own name, so a `response.incomplete` is known to be one even when its payload
  says nothing. `carriesResponse` gates latching on the response having actually
  decoded.
- `finishReason` consults `truncated()`: the event's name, then `Status`
  (enumerating the *finished* states, so a status added later defaults to
  truncated), then whether an `incomplete_details` object was sent at all. This is
  shared with the blocking path, which passes `incompleteEvent: false`.
- `logprobsFor()` returns the terminal event's logprobs only when the text they
  describe matches the answer carried — thoughts excluded, refusals included.
- The rebuild is gated on `carriesContent(final)`, and `adoptTerminalCalls()`
  makes the terminal event's tool calls the turn's own.
- `attachFinishSignal()` surfaces why a turn ended badly, on both paths.

**Pairing the two lists of tool calls**

`adoptTerminalCalls` has to say which streamed call each terminal item restates,
because an item may state a call without stating all of its fields — an item
naming a call but carrying no `arguments` says nothing about them, and taking the
blank would dispatch the tool with none. `streamedCounterpart` answers that with
three keys in order: the `call_id`, then the tool's name when exactly one streamed
call bears it, then position.

`name` and `call_id` are both `api:"required"` on a `function_call` item, so
against api.openai.com the first key always resolves and the rest never run. The
fallbacks exist for the OpenAI-compatible endpoints `ClientConfig.BaseURL` is
there to serve, which omit `call_id` on `response.output_item.added`, omit it on
the terminal item, or disagree between the two.

The rule the pairing keeps, and the reason it is worth stating: **arguments never
move from one named tool to a differently-named tool, and a streamed call restores
onto one call only.** `namesAgree` gates all three keys, so an event that crosses
two calls' IDs cannot hand each the other's arguments; and restoring once means
two items naming one tool do not both take its arguments, its ID, or its map.
One case stays out of reach and is documented rather than papered over: two calls
to the *same* tool, reordered, identified by neither key.

**One thing to look at closely.** The block reason does *not* go into `ErrorCode`
for a turn that still carries an answer. `ErrorCode` is not advisory downstream —
`tool/agenttool` aborts the run and discards the content, `adka2a` fails the A2A
task, and `converters.go:27` shows the framework rule that a candidate with content
never gets it set, so gemini leaves it empty for a filtered candidate that produced
text. Setting it here would turn a usable answer into a hard failure and split the
two model packages. Instead such a turn reports the provider's wording in
`CustomMetadata[FinishMessageKey]` beside a `FinishReason` that already says it was
cut short; only a turn with nothing to read uses the error fields.

`FinishMessageKey` is a new exported const (additive; apidiff allows additions).
No other exported identifier is added, removed, renamed or retyped, and `go.mod`
is untouched — the openai-go floor does not move, so this stays a patch-level fix.

### Behavioral changes

Enumerated by running a 34-case corpus of streamed and blocking response shapes
through the public API against `main` and against this branch, and diffing outcome
against outcome. Fifteen differ. None is a signature break; all are deliberate.

Two inputs that used to succeed now return an error:

| Input | `main` | Here |
|---|---|---|
| Terminal item with unparseable `arguments`, nothing streamed to pair it with | text returned, call silently dropped | `ErrFunctionCallArgs` |
| Terminal item with unparseable `arguments` beside a good call, counts unpairable | good call returned, bad one dropped | `ErrFunctionCallArgs` |

Both are the case where no readable copy of the call exists anywhere. Blocking
rejects the same bytes. Where the deltas *did* build the call usably, the turn
now succeeds and keeps that call — `main`'s behaviour — which is a change from
the first push of this PR.

Thirteen change a value or a shape:

| Input | `main` | Here |
|---|---|---|
| Terminal item states `arguments: "{}"` or `""` | streamed arguments | `{}` — the item states the call takes none |
| Terminal event lists the turn's calls in another order | streamed order | the event's order |
| Completed event states fewer calls than streamed | both calls | only the calls stated (an `incomplete` event is exempt) |
| Terminal event states a call the aggregator dropped | call lost | call recovered |
| Two items restate one streamed call | one call | two calls; the second takes no ID and no arguments |
| Terminal event carries no response object, or an empty one | `STOP`, zero usage | `FINISH_REASON_UNSPECIFIED`, usage left alone |
| Response carries an `error`, or a status that is not a finished one | `STOP` | `OTHER` + `CustomMetadata[FinishMessageKey]` |
| Content-filtered turn that still carries text (streaming and blocking) | `SAFETY`, no detail | `SAFETY` + `CustomMetadata[FinishMessageKey] = "content_filter"` |

Against api.openai.com none of the tool-call rows can occur: real terminal items
always carry `call_id` and echo the arguments they streamed. They reach the
compatible endpoints only.

### Testing Plan

**Unit Tests:**

```
$ go build -mod=readonly work
$ go test -race -mod=readonly -count=1 -shuffle=on work
ok    google.golang.org/adk/v2/model/openaimodel  1.814s
(all modules pass; no failures)

$ golangci-lint run                      # root and plugin/agentanalytics
0 issues.
$ go mod tidy -diff                      # root and plugin/agentanalytics
(no output)
```

Twenty-five new tests in the package, covering each defect above, the block reason
on both paths, and four paths that had no coverage: a trailing `response.created`,
a created-only stream, all four error sources, and the `iter.Seq2` contract when a
consumer breaks out of the range loop.

Every guard was then reverted individually and the suite re-run, to confirm each
test fails for its own reason rather than merely passing. That includes one
mutation per error source, and one for each key in the pairing: dropping the name
tier fails only `the tool's name pairs reordered calls the event does not
identify`; dropping `namesAgree` from the call-ID tier fails only `a call id the
names contradict does not pair`; dropping it from the position tier fails only
`position does not hand one tool another's arguments`; removing the lend seed
fails only `an id another item claims is not lent away`; restoring onto every item
fails only `a streamed call restores onto one item only`.

**Live:** the streaming path was exercised against the OpenAI API on
`gpt-5.4-nano` across single-tool, tool-selection, sequential-chain, tool-error,
streaming-with-tool-call, multi-agent-transfer and usage-metadata turns, on this
branch and on `main`. Every caller-visible result was identical, which is the
expected outcome — the shapes this PR changes are ones the real API does not
produce.

### Follow-ups

Not addressed here, deliberately:

- A terminal item stating neither a name nor a `call_id`, in a slot another item
  already took, can still pair to that same streamed call. Same family as the
  documented two-calls-to-one-tool case; it needs a name or an ID to do better.
- `model/gemini` drops `candidate.FinishMessage` for a candidate that carries
  content, the way this package did before `FinishMessageKey`. Worth the same
  treatment, in its own PR.


karolpiotrowicz:
Three behaviour changes against `main` that I don't think are intended. All three are in new code and all three reproduce.

**A batched answer is dropped silently.** A provider that puts the whole message on `response.completed` without an `id` produces zero yielded responses and no error. The consumer's `for … range GenerateContent(...)` body never executes, and it cannot distinguish that from the model having said nothing. On `main` the same bytes yield the answer.

**A content-filtered turn loses its label.** A `response.incomplete` carrying `status`, `incomplete_details.reason: "content_filter"` and `output`, but no `id`, reports `FINISH_REASON_UNSPECIFIED`. On `main` it reports `SAFETY`. Usage and logprobs go with it, since nothing downstream of the latch runs.

**A tool call is dispatched twice.** When the streamed call id came from the item-id fallback, the terminal event's copy of the same call is adopted as a second part. Wired up to an agent with a real tool, the handler is entered twice for one model intent — I ran it through `runner.Run` in SSE mode with a counter in the tool body, and got 2 on this branch against 1 on `main`. The two parts carry different IDs, so nothing downstream collapses them.

### The first two are one line

[`carriesResponse`](https://github.com/google/adk-go/blob/8f7350f4df425b473ee5314c99adc9455ee8b983/model/openaimodel/openai.go#L225-L227) returns `resp.JSON.ID.Valid()`, so when `id` is absent the [latch](https://github.com/google/adk-go/blob/8f7350f4df425b473ee5314c99adc9455ee8b983/model/openaimodel/openai.go#L137-L146) never fires and `term.seen` stays false. That skips the rebuild and falls into [the bare `return` at line 196](https://github.com/google/adk-go/blob/8f7350f4df425b473ee5314c99adc9455ee8b983/model/openaimodel/openai.go#L196-L199), which is the first symptom, and it takes [the `!term.seen` branch](https://github.com/google/adk-go/blob/8f7350f4df425b473ee5314c99adc9455ee8b983/model/openaimodel/openai.go#L294-L301) that returns before the finish reason is computed, which is the second.

The doc comment reasons that "ID is required of a response, so its presence stands for the object's". That holds for the zero-value object the guard is aimed at. It does not hold in the other direction: a response carrying `status`, `incomplete_details` and `output` has plainly decoded, and this rejects it anyway.

Gating on any field having decoded, rather than on `id` specifically, fixes both:

```go
func carriesResponse(resp *responses.Response) bool {
	j := &resp.JSON
	return j.ID.Valid() || j.Status.Valid() || j.Output.Valid() ||
		j.IncompleteDetails.Valid() || j.Error.Valid() || j.Model.Valid()
}
```

I ran that one: both cases pass and `go test ./...` stays green across the module, including `TestModel_GenerateStream_TerminalEventWithoutResponse` and `TestModel_GenerateStream_TerminalEventWithEmptyResponse`, which still reject their inputs because a bare `{}` leaves every field's raw value empty.

### The third is a key mismatch

[`adoptTerminalCalls` builds `held` from the streamed parts' `FunctionCall.ID`](https://github.com/google/adk-go/blob/8f7350f4df425b473ee5314c99adc9455ee8b983/model/openaimodel/openai.go#L251-L258) and [compares it against the terminal item's `CallID`](https://github.com/google/adk-go/blob/8f7350f4df425b473ee5314c99adc9455ee8b983/model/openaimodel/openai.go#L261). Those are only the same string when the translator had a mapping to use. [`itemToCallID` is written only when `response.output_item.added` carries a non-empty `call_id`](https://github.com/google/adk-go/blob/8f7350f4df425b473ee5314c99adc9455ee8b983/model/openaimodel/stream.go#L93-L100), and [`emitFunctionCall` otherwise falls back to `done.ItemID`](https://github.com/google/adk-go/blob/8f7350f4df425b473ee5314c99adc9455ee8b983/model/openaimodel/stream.go#L145-L148). When the fallback fires, `held` holds `fc_1\x00get_weather` while the terminal item keys `call_1\x00get_weather`, and the call is appended a second time.

Three shapes trigger it: `output_item.added` with a name but no `call_id`, `output_item.added` absent while the done event supplies the name, and `output_item.added` and the terminal item disagreeing on the `call_id`.

I'm not going to suggest a fix here, because the obvious one is wrong. Also matching on the terminal item's `id` does not close it, since a provider that omits `call_id` on `output_item.added` may omit `id` on the terminal item too, and then neither key matches. Whatever the fix, it has to keep the case the current code genuinely improves: when `output_item.added` is absent *and* the done event carries no name, the aggregator drops the call entirely and adoption is what recovers it. `main` returns nothing there and this branch correctly returns one call.

### Who this reaches

Narrower than it sounds, and worth stating. Real OpenAI always sends `id` on terminal events and always populates `call_id` on `response.output_item.added`, so none of the three can occur there. I ran the streaming path against the live API on `gpt-4o-mini` and `gpt-5-nano`, on this branch and on `main`, across normal, truncated, tool-calling, multi-tool and logprob-enabled turns, and every caller-visible result was identical. No OpenAI user is affected.

The exposure is the OpenAI-compatible endpoints that `ClientConfig.BaseURL` exists to serve. That is the same population several of the fixes in this PR are aimed at, which is why I don't think it can be waved off.

<details>
<summary>Reproduction — all three, drop into <code>model/openaimodel</code> and run <code>go test -run 'TestIdless|TestTerminalCall' ./model/openaimodel</code></summary>

```go
package openaimodel

import (
	"context"
	"fmt"
	"net"
	"net/http"
	"net/http/httptest"
	"testing"

	"google.golang.org/genai"

)

func streamOf(t *testing.T, events ...string) []*model.LLMResponse {
	t.Helper()
	srv := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Content-Type", "text/event-stream")
		for _, e := range append(events, "[DONE]") {
			fmt.Fprintf(w, "data: %s\n\n", e)
		}
	}))
	l, _ := net.Listen("tcp4", "127.0.0.1:0")
	srv.Listener = l
	srv.Start()
	defer srv.Close()

	m, _ := NewModel(context.Background(), "m", &ClientConfig{APIKey: "k", BaseURL: srv.URL})
	req := &model.LLMRequest{Contents: []*genai.Content{{Role: "user", Parts: []*genai.Part{{Text: "hi"}}}}}
	var out []*model.LLMResponse
	for resp, err := range m.GenerateContent(context.Background(), req, true) {
		if err != nil {
		}
		out = append(out, resp)
	}
	return out
}

func TestIdlessCompletedDropsTheAnswer(t *testing.T) {
	got := streamOf(t, `{"type":"response.completed","response":{"model":"m","status":"completed","output":[{"type":"message","content":[{"type":"output_text","text":"the whole answer"}]}]}}`)
	if len(got) == 0 {
		t.Fatalf("stream yielded nothing: the answer was dropped with no error")
	}
}

func TestIdlessIncompleteLosesSafety(t *testing.T) {
	got := streamOf(t,
		`{"type":"response.created","response":{"id":"resp_1","model":"m","status":"in_progress"}}`,
		`{"type":"response.output_text.delta","delta":"partial"}`,
		`{"type":"response.incomplete","response":{"model":"m","status":"incomplete","incomplete_details":{"reason":"content_filter"},"output":[{"type":"message","content":[{"type":"output_text","text":"partial"}]}]}}`,
	)
	if fr := got[len(got)-1].FinishReason; fr != genai.FinishReasonSafety {
		t.Fatalf("FinishReason = %q, want SAFETY", fr)
	}
}

func TestTerminalCallAdoptedTwice(t *testing.T) {
	got := streamOf(t,
		`{"type":"response.created","response":{"id":"resp_1","model":"m","status":"in_progress"}}`,
		`{"type":"response.output_item.added","item":{"id":"fc_1","type":"function_call","name":"get_weather"}}`,
		`{"type":"response.function_call_arguments.done","item_id":"fc_1","arguments":"{\"city\":\"Paris\"}"}`,
		`{"type":"response.completed","response":{"id":"resp_1","model":"m","status":"completed","output":[{"type":"function_call","name":"get_weather","call_id":"call_1","arguments":"{\"city\":\"Paris\"}"}]}}`,
	)
	var ids []string
	for _, p := range got[len(got)-1].Content.Parts {
		if p.FunctionCall != nil {
			ids = append(ids, p.FunctionCall.ID)
		}
	}
	if len(ids) != 1 {
		t.Fatalf("got %d function-call parts %v for one tool call, want 1", len(ids), ids)
	}
}
```

All three fail at `8f7350f` and all three pass with `model/openaimodel/{openai,response}.go` reverted to `0510fef`.

</details>

I have more from the same read, on the finish-signal routing and on a couple of smaller things. I'll follow up separately rather than pile it in here.

hanorik:
All three reproduce — thanks, the repro file made that quick.

**1 and 2 — fixed.** I've taken your `carriesResponse` rather than the widening I
offered in bug comment: yours subsumes it and also covers a response carrying
only `output`, which mine would still have dropped. A new test pins both halves —
the batched turn reaching the caller, and the cut-short turn reporting `SAFETY` —
and narrowing the guard back to `ID.Valid()` fails exactly that test and nothing
else. Workspace green under `-race -shuffle`; replaying live 4o-mini / 5-nano
captures gives identical caller-visible output either way.

Neither is from the last push, incidentally — that was a doc comment only, and both
reproduce at `51dc264`.

**3 — suggested fix, and I'd value your read.** Rather than fix the key, remove the
need for one: make the terminal event authoritative for tool calls. Replace the
aggregated function-call parts with its list instead of appending, keeping what
streamed when it carries none. That closes all three shapes plus the
`call_A`/`call_B` disagreement, keeps the nameless-call recovery, and gives
streaming the call IDs blocking already returns — it also stops the item-id
fallback reaching the wire as `call_id`.

One deliberate trade: if two calls stream and the terminal output lists one, the
second is dropped (today both survive). I read a completed response listing one call
as stating there was one, but I can gate replacement on the terminal list being no
shorter than the aggregate. Which would you prefer?

karolpiotrowicz:
Confirmed 1 and 2 are closed at `db4a53d`. The batched turn reaches the caller and the cut-short turn reports `SAFETY` with `content_filter` in the finish message. Your test discriminates too — narrowing [`carriesResponse`](https://github.com/google/adk-go/blob/db4a53d322a34a57c8ce036ef45947a04ca86888/model/openaimodel/openai.go#L227-L231) back to `ID.Valid()` fails `TestModel_GenerateStream_TerminalEventWithoutID`, both subtests, and nothing else across the module. Build, full suite, `-race`, `gofmt`, `vet` and `staticcheck` are all clean on the merge result. You are also right that neither came from the doc-comment push — that diff is four comment lines and no code, and both reproduce at `51dc264`.

## On making the terminal event authoritative

I built both of your options and a third, and ran them over ten shapes. Your instinct is right and I'd go further than the count gate.

The trade you named is real, and it has a second case you didn't mention that I think decides the question. Dropping the extra call is correct when the terminal event is a `response.completed`, because that response states the turn's whole output. It is wrong when the terminal event is a `response.incomplete`, because the turn was cut short and its output list may be shorter than what actually streamed — there you'd be discarding a call the model really did make.

So gate on the event type rather than on the count. You already track it, at [`terminalEvent.incomplete`](https://github.com/google/adk-go/blob/db4a53d322a34a57c8ce036ef45947a04ca86888/model/openaimodel/openai.go#L212-L218):

```go
// A completed response states the turn's whole output, so its call list is
// authoritative. An incomplete one was cut short and may list fewer than
// actually streamed, so keep the longer of the two.
if term.incomplete && len(fromTerminal) < heldCount {
	return nil
}
```

How the three behave, `replace` being your proposal and `gated` your count fallback:

| stream shape | today | replace | gated | completed-authoritative |
|---|---|---|---|---|
| `output_item.added` with a name, no `call_id` | **2 calls** | 1 | 1 | **1** |
| no `output_item.added`, name on the done event | **2 calls** | 1 | 1 | **1** |
| `added` says `call_A`, terminal says `call_B` | **2 calls** | 1 | 1 | **1** |
| `added` with a matching `call_id` | 1 | 1 | 1 | **1** |
| nameless call the aggregator dropped | 1 | 1 | 1 | **1** |
| two stream, terminal lists two | 2 | 2 | 2 | **2** |
| **completed** lists one, two streamed | 2 | 1 | 2 | **1** |
| **incomplete** lists one, two streamed | 2 | **1, call lost** | 2 | **2** |
| terminal lists no calls | kept | kept | kept | **kept** |
| text and a call together | order kept | order kept | order kept | **order kept** |

The count gate is safe but it also declines to replace on a completed response that lists fewer, which is the one case where the terminal genuinely is authoritative. Gating on `incomplete` gets both: your reading of a completed response as stating there was one call, and no silent loss on a truncated turn.

Module suite, `vet` and `gofmt` are clean under it, and `TestModel_GenerateStream_AdoptsTerminalToolCalls` passes unchanged. Replacing rather than appending also does what you said about the wire — the item-id fallback stops reaching `call_id`, since the parts that survive are the terminal event's.

One implementation note: filtering the aggregated function-call parts out and appending the terminal's list moves any call that preceded text to the end. It didn't matter in any shape I tried, and it may be worth preserving position if a call can precede text in practice.

## Two smaller things, neither blocking

**The `ErrorCode = "OTHER"` arm ships untested.** [The else branch](https://github.com/google/adk-go/blob/db4a53d322a34a57c8ce036ef45947a04ca86888/model/openaimodel/openai.go#L358-L365) is reachable — `response.created`, a `function_call_arguments.done` with no name, then a bare `response.incomplete` gets there. The aggregated turn has no parts, so [the `carriesContent` branch above it](https://github.com/google/adk-go/blob/db4a53d322a34a57c8ce036ef45947a04ca86888/model/openaimodel/openai.go#L349-L357) is skipped and the turn returns `ErrorCode "OTHER"` with `ErrorMessage "incomplete"`, where `main` returns `STOP` and neither. The behaviour looks right to me. The point is that setting `ErrorCode` is what makes `agenttool` abort and `adka2a` fail the task, and this is the one path in the change that does it with nothing covering it — `TestBlockedTurnSurfacesBlockReason` takes the `SAFETY` arm and `TestModel_GenerateStream_OutputlessTerminalEventKeepsTheReason` is `MAX_TOKENS`, which returns early.

**`finishMessage` can report "completed" as why a turn ended badly.** [It falls through to the payload's status](https://github.com/google/adk-go/blob/db4a53d322a34a57c8ce036ef45947a04ca86888/model/openaimodel/response.go#L230-L232), so a `response.incomplete` whose payload says `status: "completed"` gives `FinishReason OTHER` beside `openai_finish_message: "completed"`. Cosmetic, and `OTHER` is the right call. Worth a line only because it is the one place the change doesn't apply its own precedence rule — [`truncated` deliberately lets the event's name outrank a contradicting payload](https://github.com/google/adk-go/blob/db4a53d322a34a57c8ce036ef45947a04ca86888/model/openaimodel/response.go#L198-L201) and `finishMessage` doesn't do the same for `Status`. [`TestFinishMessage`](https://github.com/google/adk-go/blob/db4a53d322a34a57c8ce036ef45947a04ca86888/model/openaimodel/response_test.go#L546)'s six cases never pair a payload with an event name that contradicts it.

<details>
<summary>The shape matrix, if you want to run it — drop into <code>model/openaimodel</code></summary>

```go
package openaimodel

import (
	"context"
	"fmt"
	"net"
	"net/http"
	"net/http/httptest"
	"strings"
	"testing"

	"google.golang.org/genai"

)

func mStream(t *testing.T, events ...string) *model.LLMResponse {
	t.Helper()
	srv := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Content-Type", "text/event-stream")
		for _, e := range append(events, "[DONE]") {
			fmt.Fprintf(w, "data: %s\n\n", e)
		}
	}))
	l, _ := net.Listen("tcp4", "127.0.0.1:0")
	srv.Listener = l
	srv.Start()
	defer srv.Close()
	m, _ := NewModel(context.Background(), "m", &ClientConfig{APIKey: "k", BaseURL: srv.URL})
	req := &model.LLMRequest{Contents: []*genai.Content{{Role: "user", Parts: []*genai.Part{{Text: "hi"}}}}}
	var last *model.LLMResponse
	for resp, err := range m.GenerateContent(context.Background(), req, true) {
		if err != nil {
			t.Fatalf("err: %v", err)
		}
		last = resp
	}
	return last
}

func shape(r *model.LLMResponse) string {
	var s []string
	for _, p := range r.Content.Parts {
		if p.FunctionCall != nil {
			s = append(s, "fc:"+p.FunctionCall.ID)
		} else if p.Text != "" {
			s = append(s, "txt:"+p.Text)
		}
	}
	return strings.Join(s, " ")
}

const created = `{"type":"response.created","response":{"id":"resp_1","model":"m","status":"in_progress"}}`

func TestMatrix(t *testing.T) {
	cases := []struct {
		name string
		ev   []string
	}{
		{"A added-no-callid", []string{created,
			`{"type":"response.output_item.added","item":{"id":"fc_1","type":"function_call","name":"w"}}`,
			`{"type":"response.function_call_arguments.done","item_id":"fc_1","arguments":"{}"}`,
			`{"type":"response.completed","response":{"id":"r","model":"m","status":"completed","output":[{"type":"function_call","name":"w","call_id":"call_1","arguments":"{}"}]}}`}},
		{"B no-added, name on done", []string{created,
			`{"type":"response.function_call_arguments.done","item_id":"fc_1","name":"w","arguments":"{}"}`,
			`{"type":"response.completed","response":{"id":"r","model":"m","status":"completed","output":[{"type":"function_call","name":"w","call_id":"call_1","arguments":"{}"}]}}`}},
		{"C call_id disagrees", []string{created,
			`{"type":"response.output_item.added","item":{"id":"fc_1","type":"function_call","name":"w","call_id":"call_A"}}`,
			`{"type":"response.function_call_arguments.done","item_id":"fc_1","arguments":"{}"}`,
			`{"type":"response.completed","response":{"id":"r","model":"m","status":"completed","output":[{"type":"function_call","name":"w","call_id":"call_B","arguments":"{}"}]}}`}},
		{"completed lists 1, two streamed", []string{created,
			`{"type":"response.output_item.added","item":{"id":"f1","type":"function_call","name":"w","call_id":"c1"}}`,
			`{"type":"response.function_call_arguments.done","item_id":"f1","arguments":"{}"}`,
			`{"type":"response.output_item.added","item":{"id":"f2","type":"function_call","name":"w","call_id":"c2"}}`,
			`{"type":"response.function_call_arguments.done","item_id":"f2","arguments":"{}"}`,
			`{"type":"response.completed","response":{"id":"r","model":"m","status":"completed","output":[{"type":"function_call","name":"w","call_id":"c1","arguments":"{}"}]}}`}},
		{"incomplete lists 1, two streamed", []string{created,
			`{"type":"response.output_item.added","item":{"id":"f1","type":"function_call","name":"w","call_id":"c1"}}`,
			`{"type":"response.function_call_arguments.done","item_id":"f1","arguments":"{}"}`,
			`{"type":"response.output_item.added","item":{"id":"f2","type":"function_call","name":"w","call_id":"c2"}}`,
			`{"type":"response.function_call_arguments.done","item_id":"f2","arguments":"{}"}`,
			`{"type":"response.incomplete","response":{"id":"r","model":"m","status":"incomplete","incomplete_details":{"reason":"max_output_tokens"},"output":[{"type":"function_call","name":"w","call_id":"c1","arguments":"{}"}]}}`}},
	}
	for _, c := range cases {
		t.Run(c.name, func(t *testing.T) { t.Logf("%-34s => %s", c.name, shape(mStream(t, c.ev...))) })
	}
}
```

</details>

hanorik:
One behaviour change against `main` that the terminal-authoritative
change brings, since replacement converts every item the event states
rather than only the ones the turn lacked.

A terminal `function_call` item with malformed `arguments` now fails the
whole stream, even where the deltas built the same call usably. `main`
returns the streamed call with no error, and `db4a53d` did too — the
dedupe key matched, so the item was never converted.

Deliberate: blocking rejects the same body, and falling back to the
streamed copy would have the two paths disagree on identical bytes.
Pinned by `TestModel_GenerateStream_TerminalToolCallsAreAuthoritative/
an_unusable_item_fails_the_turn_even_beside_the_call_it_restates`. Say if
you'd rather the streamed call survive.

karolpiotrowicz (dismissed):
The design landed and I only have one thing left that I'd want changed before this merges.

The `incomplete` gate does what it needs to — a truncated turn listing one of two keeps both, a completed one listing one drops the second — and all three duplicating shapes are closed. You solved the ordering caveat rather than accepting it: `partsWithoutCalls` keeping the slot means a turn that interleaved text and calls comes back interleaved. Build, full suite, `-race`, `gofmt`, `vet` and `staticcheck` are clean on the merge result. `restoreUnstated` is a good catch I hadn't considered — an item naming a call but stating no arguments would otherwise dispatch the tool with none.

## The one blocking thing: reordered calls get each other's arguments

[`streamedCounterpart`](https://github.com/google/adk-go/blob/a4a6dc12c1ece65f6fe889552567d48811ccca50/model/openaimodel/openai.go#L328-L340) tries `CallID` and then falls back to position. Your doc comment names the hazard exactly — position "alone it would be wrong for an event listing the same calls in another order" — and the `CallID` guard is what should prevent it. But a terminal item carrying no `call_id` is the same provider shape that caused the duplication in the first place, so the guard is absent in precisely the case that needs it, and the fallback pairs each stated call with whichever one streamed in its slot.

[`restoreUnstated`](https://github.com/google/adk-go/blob/a4a6dc12c1ece65f6fe889552567d48811ccca50/model/openaimodel/openai.go#L369) then fills arguments from the wrong call:

```
streamed:   alpha {"x":1}   beta {"y":2}
delivered:  beta  {"x":1}   alpha {"y":2}
```

Both tools run, both with the other's arguments, and nothing surfaces it. That is worse than the duplicate dispatch this replaces — a duplicate at least carries the right arguments.

`Name` is available on both sides and unused for pairing. Trying an unambiguous name match before the position fallback closes it and leaves your suite green:

```go
if name := item.Name; name != "" {
	var match *genai.FunctionCall
	for _, call := range streamed {
		if call.Name != name {
			continue
		}
		if match != nil {
			match = nil // two calls to the same tool: name cannot tell them apart
			break
		}
		match = call
	}
	if match != nil {
		return match
	}
}
```

One case stays out of reach and wants a comment rather than a fix: two calls to the *same* tool, reordered, with no `call_id`. Nothing distinguishes them, so position is all there is and the arguments can still cross.

<details>
<summary>Reproduction — drop into <code>model/openaimodel</code> and run <code>go test -run TestTerminalReorderSwapsArguments ./model/openaimodel</code></summary>

```go
package openaimodel

import (
	"context"
	"encoding/json"
	"fmt"
	"net"
	"net/http"
	"net/http/httptest"
	"testing"

	"google.golang.org/genai"

)

func TestTerminalReorderSwapsArguments(t *testing.T) {
	srv := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Content-Type", "text/event-stream")
		for _, e := range []string{
			`{"type":"response.created","response":{"id":"r","model":"m","status":"in_progress"}}`,
			`{"type":"response.output_item.added","item":{"id":"f1","type":"function_call","name":"alpha","call_id":"cA"}}`,
			`{"type":"response.function_call_arguments.done","item_id":"f1","arguments":"{\"x\":1}"}`,
			`{"type":"response.output_item.added","item":{"id":"f2","type":"function_call","name":"beta","call_id":"cB"}}`,
			`{"type":"response.function_call_arguments.done","item_id":"f2","arguments":"{\"y\":2}"}`,
			`{"type":"response.completed","response":{"id":"r","model":"m","status":"completed","output":[` +
				`{"type":"function_call","name":"beta"},{"type":"function_call","name":"alpha"}]}}`,
			"[DONE]",
		} {
			fmt.Fprintf(w, "data: %s\n\n", e)
		}
	}))
	l, _ := net.Listen("tcp4", "127.0.0.1:0")
	srv.Listener = l
	srv.Start()
	defer srv.Close()

	m, _ := NewModel(context.Background(), "m", &ClientConfig{APIKey: "k", BaseURL: srv.URL})
	req := &model.LLMRequest{Contents: []*genai.Content{{Role: "user", Parts: []*genai.Part{{Text: "hi"}}}}}
	var last *model.LLMResponse
	for resp, err := range m.GenerateContent(context.Background(), req, true) {
		if err != nil {
			t.Fatalf("stream: %v", err)
		}
		last = resp
	}
	want := map[string]string{"alpha": `{"x":1}`, "beta": `{"y":2}`}
	for _, p := range last.Content.Parts {
		if p.FunctionCall == nil {
			continue
		}
		got, _ := json.Marshal(p.FunctionCall.Args)
		if string(got) != want[p.FunctionCall.Name] {
			t.Errorf("%s got args %s, want %s", p.FunctionCall.Name, got, want[p.FunctionCall.Name])
		}
	}
}
```

Fails at `a4a6dc1` with each call reporting the other's arguments, and passes with the name match above.

</details>

## Everything else is yours to take or leave

None of these is a merge condition, and I don't want them read as one.

- **Malformed arguments.** You asked, so: I'd let the streamed call survive, though not on the parity grounds. Blocking fails that body because the malformed item is all it has, while streaming has a usable call from the deltas — the two paths differ because their inputs differ, not because the rule differs. The reason I'd change it is your own design. `restoreUnstated` already settles that the event is authoritative for *which* calls a turn made rather than for every field of each, and falls back to the deltas when arguments are unstated. Arguments that are stated but unparseable are the same failure handled the opposite way. Either rule is fine — I'd just not keep them split, since a caller can't reason about the difference between the two payloads.
- **The `ErrorCode = "OTHER"` arm** still has no test. One case closes it.
- **`finishMessage` reporting "completed"** is still cosmetic.

## Where that leaves us

Fix the pairing and keep the gates green and I'll approve. The rest is discretionary and I'm happy for it to land as you see fit, or in a follow-up.

wojtas000:
Pairing is fixed, and the two discretionary points turned out to need evidence rather than a patch. Pushed on top of 4e4e5a17.

- **Reordered calls getting each other's arguments** — fixed. [`streamedCounterpart` now tries the name between the call ID and position](https://github.com/google/adk-go/blob/8cc376a85ea19f43d4bbf38e404cb3e22e67ed15/model/openaimodel/openai.go#L352-L398), close to what you wrote; your `TestTerminalReorderSwapsArguments` fails at 4e4e5a17 and passes here. Two calls to the same tool, reordered, with no `call_id` stay out of reach, and the doc comment says so rather than pretending otherwise.

  One thing your version left open, which I only found by running the crossed case: the `call_id` tier had no name check either, so an event stating `call_id: call_1` on an item named `beta` still took `alpha`'s arguments — the same swap through the key meant to prevent it. [`namesAgree`](https://github.com/google/adk-go/blob/8cc376a85ea19f43d4bbf38e404cb3e22e67ed15/model/openaimodel/openai.go#L400-L404) now gates all three tiers, so arguments never move between differently-named tools. That is a rule worth stating plainly, because it is the one this whole function exists to keep.

- **Malformed arguments** — taken, on your reasoning rather than mine. [An unusable item now keeps the call the deltas built](https://github.com/google/adk-go/blob/8cc376a85ea19f43d4bbf38e404cb3e22e67ed15/model/openaimodel/openai.go#L293-L304); `main` returns that call, so failing the turn broke a stream that works today, and the split you objected to is gone. An item nothing pairs still errors: with no readable copy, reporting the call would dispatch the tool with arguments the model did not write.

- **The `ErrorCode = "OTHER"` arm** — already covered, at 4e4e5a12 as well. [`TestUnfinishedTurnWithNothingToReadReportsErrorCode`](https://github.com/google/adk-go/blob/8cc376a85ea19f43d4bbf38e404cb3e22e67ed15/model/openaimodel/openai_test.go#L556-L583) drives it through a bare `response.incomplete`; replacing the assignment with `""` fails that test and nothing else.

- **`finishMessage` reporting "completed"** — not reachable. [The status arm is guarded on `!incompleteEvent || Status != completed`](https://github.com/google/adk-go/blob/8cc376a85ea19f43d4bbf38e404cb3e22e67ed15/model/openaimodel/response.go#L217-L242), and an exhaustive sweep of status × incomplete reason × `incomplete_details` presence × error message × event name — 256 combinations — produces no input where a turn surfacing a finish message reports "completed". Every combination that would returns "incomplete" instead. Happy to add that sweep as a test if you want it pinned.

Three more, all found by running a 34-case input corpus against `main` and this branch rather than by reading the diff, and all in the code above rather than yours:

- Restoring the streamed ID onto an item that stated none reported **two calls under one ID** when two items named one tool, and could hand away an ID a later item claimed for itself. [The ID is lent once, and the map is seeded with what the event states before any is lent](https://github.com/google/adk-go/blob/8cc376a85ea19f43d4bbf38e404cb3e22e67ed15/model/openaimodel/openai.go#L282-L321).
- Position paired an item to whatever streamed in its slot **regardless of tool name**, which is the transplant above by another route.
- Both items naming one tool took its arguments, which also made an event stating N items against one large streamed payload allocate N copies of it. [A streamed call now restores onto one call only](https://github.com/google/adk-go/blob/8cc376a85ea19f43d4bbf38e404cb3e22e67ed15/model/openaimodel/openai.go#L422-L458) — which of two items the model wrote those arguments for is unknowable, and restoring once removes the copy rather than bounding it.

One open question I would rather you rule on than decide myself. `namesAgree` means a provider that *renames* a call in the terminal event while keeping its `call_id` loses the streamed arguments and dispatches with none. I chose that because `restoreUnstated` already prefers no arguments over wrong ones, and because transplanting across names is the direction that does damage — but if you read a rename as a correction the provider is entitled to make, the ID tier is where you would drop the check.

karolpiotrowicz (dismissed):
The pairing is fixed and I have nothing left blocking. `TestTerminalReorderSwapsArguments` fails at [4e4e5a17](https://github.com/google/adk-go/commit/4e4e5a173863409743dfdc734c1c3f486ed743d7), the state before this push, and passes here. I also swept the general property behind it — two calls, the terminal event listing them in the reverse order, over all sixteen combinations of `call_id` present or absent on the two items — and no tool ever receives another tool's arguments. Fifteen of the sixteen cross before this push.

**Gating all three tiers on the name is more than I asked for, and it is the right shape.** What I prescribed left the `call_id` tier unchecked, so the crossed case you found still swapped, and it also dropped the IDs from reordered calls entirely — arguments correct, nothing for a caller to match a tool result back to. [`namesAgree`](https://github.com/google/adk-go/blob/8cc376a85ea19f43d4bbf38e404cb3e22e67ed15/model/openaimodel/openai.go#L402-L404) gating [all three tiers](https://github.com/google/adk-go/blob/8cc376a85ea19f43d4bbf38e404cb3e22e67ed15/model/openaimodel/openai.go#L369-L399) closes both.

**The `finishMessage` item was mine to withdraw, and I should have withdrawn it a round earlier.** The status guard was already in [a4a6dc12](https://github.com/google/adk-go/commit/a4a6dc12c1ece65f6fe889552567d48811ccca50) — the commit I was reviewing. I raised it correctly against [db4a53d](https://github.com/google/adk-go/commit/db4a53d322a34a57c8ce036ef45947a04ca86888), the revision at the time, then carried it forward without re-checking after the squash. Your reading is right, and an independent sweep of 360 combinations of event name, payload status, `incomplete_details` shape, error message and content presence agrees: no turn reports "completed" as why it ended. No test needed. The `ErrorCode` arm and the malformed-arguments change both hold up.

**On the rename question: keep the check where it is.** Three reasons. The two failure modes are not symmetric — an empty argument map is likely to be rejected by the tool's own validation, whereas a well-formed set written for a different tool is plausible input that fails silently. It is also the rule this design already settled: `restoreUnstated` prefers no arguments to wrong ones, and dropping the check on the ID tier would split one question across two adjacent paths again, which is the thing I objected to on malformed arguments. And a rename under a stable `call_id` is a provider contradicting its own `output_item.added`, not correcting it. The cost is real and worth naming in the doc comment: a caller cannot tell "this tool takes no arguments" from "the arguments were dropped", since both arrive as an empty map.

```
terminal item states call_id "cA" but names alpha_v2; "cA" streamed as alpha {"x":1}

before this push:  name="alpha_v2" id="cA" args={"x":1}
with this push:    name="alpha_v2" id="cA" args={}
```

**One thing to file away rather than act on.** The translator never reads `output_index`, though it is on the events it already handles. adk-python's Responses path keys streamed items on it before `item_id` and `call_id`, which would be an exact pairing and would retire the name-or-position heuristic. I have not checked how reliably OpenAI-compatible endpoints populate it, so treat that as a direction to look in, not a change to make here.

Two things on landing it. I am dismissing my earlier changes-requested review, which a comment does not clear on its own. The branch is behind `main` — the merge is clean and I ran everything on the merge result — and it conflicts with #1467, #1392 and #1359, all live in this package, so the merge order is worth choosing deliberately rather than discovering.

karolpiotrowicz (approved):

PR #1376, opened 2026-08-21
refactor(agent/llmagent): decouple workflow node prompt injection from session and eliminate wrappedSession

### LINKED ISSUE #1371: [Refactor]: Eliminate wrappedSession and decouple workflow node prompt injection from Session
### Problem Description

In `agent/llmagent/llm_agent_wrapper.go:642-660`, `single_turn` workflow nodes inject transient node inputs into the LLM prompt by wrapping the underlying session in synthetic `wrappedSession` and `wrappedEvents` structs (`PrepareLLMAgentInput`).

This has significant architectural drawbacks (noted by maintainers in `TODO` at `llm_agent_wrapper.go:642`):
1. **Session Pollution**: Wrapping the session pollutes the `session.Session` abstraction: any tool, plugin, callback, or telemetry hook inspecting `ctx.Session()` during execution sees synthetic phantom events that do not exist in persistent storage.
2. **Unnecessary Coupling**: Faking a session event just to feed a prompt to the LLM couples workflow graph node orchestration to session state manipulation.
3. **Index Drift Risk**: `wrappedEvents.At(i)` relies on manual index offset math (`insertAt`), which risks index misalignment during multi-round tool executions.

### Proposed Solution

As outlined in the codebase `TODO`:
1. Eliminate `wrappedSession`, `wrappedEvents`, and `PrepareLLMAgentInput` from `agent/llmagent/llm_agent_wrapper.go`.
2. Rely directly on `InvocationContext.UserContent()` (which already carries the rendered `nodeInput`).
3. Update `internal/llminternal/contents_processor.go` to directly read `ctx.UserContent()` and prepend it to `LLMRequest.Contents` for `single_turn` agents.
4. Keep the session history pure, properly decoupling workflow node inputs from session state.

pranav718:

- Closes: #1371

**Problem:**
Previously, `RunLLMAgentAsNode` in `agent/llmagent/llm_agent_wrapper.go` injected workflow node inputs by wrapping `session.Session` inside a synthetic wrapper (`wrappedSession` / `wrappedEvents`). This coupled prompt delivery to session history emulation and required maintaining virtual event indexing and slicing logic.

**Solution:**
1. Eliminated `wrappedSession`, `wrappedEvents`, and `newWrappedSession` from `agent/llmagent/llm_agent_wrapper.go`.
2. Passed `ctx.Session()` directly into `InvocationContext` and set `userContent = nodeInputToContent(nodeInput)`.
3. Decoupled prompt preparation in `internal/llminternal/contents_processor.go`:
   - Prioritizes `userContent` directly when provided.
   - Accurately identifies turn boundaries and branch events without mutating or polluting session history.
4. Added unit tests verifying real in-memory session history remains unpolluted.

---

### Testing Plan

**Unit Tests:**

**Summary of passed `go test` results:**
```text
ok  	google.golang.org/adk/plugin/agentanalytics	4.067s
ok  	google.golang.org/adk/v2/agent	4.552s
ok  	google.golang.org/adk/v2/agent/llmagent	5.308s
ok  	google.golang.org/adk/v2/internal/llminternal	10.920s
ok  	google.golang.org/adk/v2/internal/telemetry/functionaltest	10.518s
ok  	google.golang.org/adk/v2/workflow	7.405s
...
All tests green across all workspace modules.

### CODE REVIEW

**Review by karolpiotrowicz — CHANGES_REQUESTED** [review]
The refactor's goal is right and the direction is the one the removed TODO in `llm_agent_wrapper.go` itself prescribed: deliver the node input through `InvocationContext.UserContent()` and stop wrapping the session. Session purity is achieved by construction — `ctx.Session()` is now passed through untouched, and there is no longer any object that can report a synthetic event. The problem is the new delivery path: on the ordinary workflow-node path the node input no longer reaches the model at all.

## The node input is dropped whenever the session already holds a user turn

[`internal/llminternal/contents_processor.go#L196-L202`](https://github.com/google/adk-go/blob/b49c1e1dde249715c90b26af8413b355a7828d18/internal/llminternal/contents_processor.go#L196-L202) guards the prepend with `if len(contents) == 0 || contents[0].Role != genai.RoleUser`. That prepend is now the only route by which a node input reaches the request, so when the first surviving content is already user-role, the input is silently discarded and the model runs on whatever was there instead.

This is the normal case rather than an edge case. The runner appends the end user's turn to the session before the workflow starts ([`runner/runner.go#L655-L659`](https://github.com/google/adk-go/blob/b49c1e1dde249715c90b26af8413b355a7828d18/runner/runner.go#L655-L659)), so a user-authored event is present from the first turn onward. Since `RunLLMAgentAsNode` forces `IncludeContents="none"` for single_turn agents, the backward scan pivots on that user event, `buildContentsDefault` produces a user-role first content, and the guard suppresses the prepend.

I chained two single_turn nodes in a `workflow.Workflow` and ran it through `workflow.Workflow.Run`. On the merge-base the second node receives the first node's output. With this PR applied it receives the original user message instead, and the upstream node's output never appears in the request. The same happens for a delegated task agent whose first in-scope event is another agent's reply, because `ConvertForeignEvent` writes `Role: "user"` — there the `{"task": ...}` brief disappears and the model sees only `For context: [helper] said: ...`.

<details><summary>Reproduction — drop into <code>agent/llmagent/</code> and run <code>go test -run NodeInputReachesModel ./agent/llmagent/</code></summary>

```go
package llmagent_test

import (
	"context"
	"iter"
	"strings"
	"testing"

	"google.golang.org/genai"

)

type capturingLLM struct{ reqs []*model.LLMRequest }

func (c *capturingLLM) Name() string { return "capturing" }

func (c *capturingLLM) GenerateContent(_ context.Context, req *model.LLMRequest, _ bool) iter.Seq2[*model.LLMResponse, error] {
	c.reqs = append(c.reqs, req)
		yield(&model.LLMResponse{Content: &genai.Content{
			Role: "model", Parts: []*genai.Part{genai.NewPartFromText("done")},
		}}, nil)
	}
}

func TestNodeInputReachesModel(t *testing.T) {
	const nodeInput = "TRANSIENT-NODE-INPUT"

	created, err := svc.Create(t.Context(), &session.CreateRequest{AppName: "app", UserID: "u", SessionID: "s"})
	if err != nil {
		t.Fatalf("session.Create: %v", err)
	}
	// The runner appends the user turn before the workflow starts; mirror that.
	ev := session.NewEvent(t.Context(), "inv-existing")
	ev.Author = "user"
	ev.LLMResponse = model.LLMResponse{Content: &genai.Content{
		Role: genai.RoleUser, Parts: []*genai.Part{genai.NewPartFromText("PRIOR-USER-MESSAGE")},
	}}
	if err := svc.AppendEvent(t.Context(), created.Session, ev); err != nil {
		t.Fatalf("AppendEvent: %v", err)
	}
	got, err := svc.Get(t.Context(), &session.GetRequest{AppName: "app", UserID: "u", SessionID: "s"})
	if err != nil {
		t.Fatalf("session.Get: %v", err)
	}

	cap := &capturingLLM{}
	a, err := llmagent.New(llmagent.Config{Name: "st", Description: "d", Model: cap, Mode: llmagent.ModeSingleTurn})
	if err != nil {
		t.Fatalf("llmagent.New: %v", err)
	}
	ic := icontext.NewInvocationContext(t.Context(), icontext.InvocationContextParams{
		Agent: a, Session: got.Session, InvocationID: "inv-test",
	})
	for _, err := range llmagent.RunLLMAgentAsNode(a, agent.NewContext(ic), nodeInput) {
		if err != nil {
			t.Fatalf("RunLLMAgentAsNode: %v", err)
		}
	}
	if len(cap.reqs) == 0 {
		t.Fatal("model was never called")
	}
	var seen []string
	for _, c := range cap.reqs[0].Contents {
		for _, p := range c.Parts {
			if p == nil {
				continue
			}
			if strings.Contains(p.Text, nodeInput) {
				return
			}
			seen = append(seen, c.Role+":"+p.Text)
		}
	}
	t.Errorf("node input %q never reached the model; contents = %v", nodeInput, seen)
}
```

Passes on the merge-base, fails with this PR applied.

</details>

One caution on the fix, because I checked and the guard is not gratuitous. When `nodeInput` is nil, [`llm_agent_wrapper.go#L101-L105`](https://github.com/google/adk-go/blob/b49c1e1dde249715c90b26af8413b355a7828d18/agent/llmagent/llm_agent_wrapper.go#L101-L105) falls back to `ctx.UserContent()`, which on the runner path is the very same `*genai.Content` already appended as the last session event. An unconditional prepend would emit the user message twice. So the guard has a real case behind it — it just tests position and role where it needs to test identity against the content it is about to prepend. I deliberately have not proposed a specific predicate, because I have not run the duplicate-message case on either revision and I would rather not hand you a fix I cannot vouch for.

## Branched agents with `IncludeContents: "none"` end up with an empty request

[`contents_processor.go#L533-L537`](https://github.com/google/adk-go/blob/b49c1e1dde249715c90b26af8413b355a7828d18/internal/llminternal/contents_processor.go#L533-L537) adds a strict `event.Branch != branch` skip immediately above the existing `eventBelongsToBranch` call. The two disagree: `eventBelongsToBranch` deliberately treats `event.Branch == ""` as universally visible and accepts ancestor prefixes, and the new line rejects both.

The runner's user event never carries a branch — `session.NewEvent` does not set one and nothing in the `runner` package assigns it — while [`parallelagent/agent.go#L83-L95`](https://github.com/google/adk-go/blob/b49c1e1dde249715c90b26af8413b355a7828d18/agent/workflowagents/parallelagent/agent.go#L83-L95) gives every sub-agent `Branch = "<parent>.<sub>"`. So for a `ParallelAgent` child configured `IncludeContents: "none"`, the user event is skipped as a pivot, the fallback filters it out again, and `buildContentsDefault` returns nothing. The continuation-turn injection is itself guarded on a non-empty `Contents`, so nothing backfills and the model is called with an empty request. On the merge-base that same agent receives the user message.

There is a worse variant I could construct on paper but did not execute: with `UserContent` nil, the same path falls through to the full event list and delivers prior-turn history to an agent explicitly configured to receive none. Worth checking when you fix the pivot.

Once the strict skip is reconciled with `eventBelongsToBranch`, note that the `!eventBelongsToBranch(...)` check on the next line and its four-line rationale comment are currently unreachable — after the new skip, every surviving event satisfies it.

## The single-turn nudge is lost on the delegation path

[`contents_processor.go#L714-L719`](https://github.com/google/adk-go/blob/b49c1e1dde249715c90b26af8413b355a7828d18/internal/llminternal/contents_processor.go#L714-L719) returns `userContent` before reaching the `isolationScope != ""` branch that appends `SingleTurnNudge`. Because `dispatchTaskFC` always populates `UserContent`, that branch is now unreachable on the non-resume dispatch path, and an FC-delegated single_turn agent is no longer told it will get no follow-up replies. The brief text itself is unaffected — both revisions render the same JSON — so this is one missing sentence, not a lost task. Worth fixing anyway, since a single_turn agent that asks a clarifying question returns that question as the node's output. The doc comment just above the function still says the helper "falls back to userContent" when no matching FC is found, which is now the opposite of what it does.

## Lint

[`llmagent_saveoutput_test.go#L146-L149`](https://github.com/google/adk-go/blob/b49c1e1dde249715c90b26af8413b355a7828d18/agent/llmagent/llmagent_saveoutput_test.go#L146-L149) — deleting `TestWrappedSession_SeedNotPersisted` left `newSessionWithEvent` with no callers. Running the repository's pinned `golangci-lint` v2.3.1 goes from `0 issues` on the merge-base to `func newSessionWithEvent is unused (unused)` with this applied. Note that the CI checks on this PR have not actually run — only the CLA and change-detection jobs did — so there is no green signal here to rely on.

Rather than deleting the helper, it builds exactly the one-prior-user-event session that exposes the first issue above, so it may be more useful as the fixture for a real test.

## Smaller things

- The added test at [`llm_agent_wrapper_test.go#L382-L385`](https://github.com/google/adk-go/blob/b49c1e1dde249715c90b26af8413b355a7828d18/agent/llmagent/llm_agent_wrapper_test.go#L382-L385) asserts only that the session length is unchanged. It compiles and passes against the merge-base, and it still passes with the whole prepend block deleted — the model stub it uses discards the request, so it cannot observe delivery at all. Its comment says it asserts that node inputs "are passed via `UserContent`", which nothing in the body inspects. Running it on an empty session picks the one shape where the first issue is invisible.
- `PrepareLLMAgentInput` at [`llm_agent_wrapper.go#L126-L128`](https://github.com/google/adk-go/blob/b49c1e1dde249715c90b26af8413b355a7828d18/agent/llmagent/llm_agent_wrapper.go#L126-L128) now has no callers outside its own tests, and its doc still describes returning "the seeded user-role event". Keeping the exported symbol is the right call for compatibility, but the doc and the `RunLLMAgentAsNode` doc above it both still describe seeding, which no longer happens.
- [`contents_processor.go#L604-L617`](https://github.com/google/adk-go/blob/b49c1e1dde249715c90b26af8413b355a7828d18/internal/llminternal/contents_processor.go#L604-L617) — `isOtherAgentReply` gained an exemption for events carrying a `FunctionCall` named after the current agent. The predicate is per-part but the effect is per-event, so any text in that event rides along and is delivered verbatim as a `model`-role content with a live `FunctionCall` and no matching response, instead of being rewritten to `For context: [agent] said: ...`. I traced the delegation, `agenttool` and `transfer_to_agent` paths and each is blocked before this is reached, so I could not show it happening in practice — but an agent name colliding with a tool name would be enough, and nothing prevents that. Worth a comment stating the intended contract either way.
- The new `hasFunctionResponse` gate in `isTurnStartEvent` changes where the turn boundary falls for any `IncludeContents: "none"` agent whose history contains a user-authored `FunctionResponse`, and `synthesizeTaskFREvent` produces exactly those. It may well be an improvement — pivoting on a response would slice out the call it answers — but it is an undocumented behavior change in a change described as behavior-preserving, and no test covers it.
- The fallback at [`contents_processor.go#L576`](https://github.com/google/adk-go/blob/b49c1e1dde249715c90b26af8413b355a7828d18/internal/llminternal/contents_processor.go#L576) hands `buildContentsDefault` a slice from which out-of-scope events have already been removed, and that slice is what `buildTaskInputUserContent` searches for the originating delegation call. That call lives on an unscoped parent event, so the search cannot succeed there. It is inert today because the early return above short-circuits first, which means fixing the nudge issue without also fixing this would quietly kill the FC-args path.

## Before this can merge

The branch conflicts with `main` in `agent/llmagent/llm_agent_wrapper.go` — the compaction change moved that file after this branch was cut. A rebase is mechanical, but everything above was measured on the branch as it stands, not on a merged tree.

karolpiotrowicz:

It re-specifies how a workflow node's input reaches the model, moving that delivery into the shared contents processor and removing `wrappedSession` along the way. That construct is Go-only, so there is no sibling implementation to validate the new contract against, and the review found a blocker in exactly the rewritten logic. Worth deciding whether we want this re-specified now, before the author invests another round.

pranav718:
thanks for the detailed review @karolpiotrowicz, really appreciate the breakdown. 

makes total sense on the node input getting dropped. i'll hold off on fixes until the scope call with @wolo-lab is done. lmk what you guys decide!

PR #1377, opened 2026-08-21
feat: Add daily time partitioning to BigQuery event logging table

vasilyl:

- Closes: N/A

**Problem:**
Logging large volumes of events into a flat BigQuery table can lead to inefficient query scanning, resulting in slower analytical queries and higher query costs over time.

**Solution:**
Configure daily time partitioning during table creation inside `NewBigQueryAgentAnalyticsPluginWithClients`. The BigQuery table is created with `TimePartitioning` enabled on the `timestamp` column (via the shared `timestampField` constant) using `bq.DayPartitioningType`.

**Operational Notes:**
- **Newly Created Tables Only:** Time partitioning is applied only when the plugin creates a new table (`tableRef.Create`). Pre-existing unpartitioned tables are left untouched when `Metadata(ctx)` succeeds; migrating an existing unpartitioned table requires recreating or migrating it manually.
- **Partition Expiration Policy:** No explicit `TimePartitioning.Expiration` is set, so newly created partitioned tables inherit the dataset's `defaultPartitionExpirationMs`. Note that per BigQuery semantics, dataset partition expiration takes precedence over dataset default table expiration once a table is partitioned.
- **PR Stack Order:** This PR (#1377) is the base of the `agentanalytics` feature series (#1377–#1384) and is intended to merge first; dependent PRs in the series will be rebased onto `main` once this lands.

### Testing Plan

Added unit test `TestNewBigQueryAgentAnalyticsPlugin_CreateTable_WithPartitioning`, which unmarshals the captured JSON table-creation request payload and asserts exact values (`timePartitioning.field == "timestamp"` and `timePartitioning.type == "DAY"`).

### Checklist

> ⚠️ This content was generated by an AI-powered agentic tool. It may contain inaccuracies. Please verify before acting on it.

### DIFF
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go
@@ -15,6 +15,7 @@
 package agentanalytics
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"io"
@@ -340,5 +341,89 @@ func TestLogEvent_ExtractsTraceInfo(t *testing.T) {
 		}
 	case <-time.After(5 * time.Second):
 		t.Error("Timed out waiting for request")
+	}
+}
+
+func TestNewBigQueryAgentAnalyticsPlugin_CreateTable_WithPartitioning(t *testing.T) {
+	ctx := context.Background()
+	config := DefaultConfig()
+	config.Enabled = true
+	config.ProjectID = "test-project"
+	config.DatasetID = "test-dataset"
+	config.TableName = "test-table"
+
+	createCalled := false
+	var requestBody string
+
+	mockTransport := &mockTransport{
+		roundTrip: func(r *http.Request) (*http.Response, error) {
+			// Table metadata request: returns 404 Not Found to trigger creation
+			if r.Method == "GET" && strings.Contains(r.URL.Path, "/datasets/test-dataset/tables/test-table") {
+				return &http.Response{
+					StatusCode: http.StatusNotFound,
+					Body:       io.NopCloser(strings.NewReader(`{"error":{"code":404,"message":"Not found"}}`)),
+				}, nil
+			}
+			// Table creation request
+			if r.Method == "POST" && strings.Contains(r.URL.Path, "/datasets/test-dataset/tables") {
+				createCalled = true
+				bodyBytes, _ := io.ReadAll(r.Body)
+				requestBody = string(bodyBytes)
+				r.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+
+				return &http.Response{
+					StatusCode: http.StatusOK,
+					Body:       io.NopCloser(strings.NewReader("{}")),
+				}, nil
+			}
+			return &http.Response{
+				StatusCode: http.StatusOK,
+				Body:       io.NopCloser(strings.NewReader("{}")),
+			}, nil
+		},
+	}
+	httpClient := &http.Client{Transport: mockTransport}
+	bqClient, err := bq.NewClient(ctx, config.ProjectID, option.WithHTTPClient(httpClient))
+	if err != nil {
+		t.Fatalf("Failed to create bigquery client: %v", err)
+	}
+
+	lis, err := net.Listen("tcp", "localhost:0")
+	if err != nil {
+		t.Fatalf("failed to listen: %v", err)
+	}
+	gSrv := grpc.NewServer()
+	storagepb.RegisterBigQueryWriteServer(gSrv, &fakeBigQueryWriteServer{})
+	go func() { _ = gSrv.Serve(lis) }()
+	t.Cleanup(gSrv.Stop)
+
+	conn, err := grpc.NewClient(lis.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("failed to dial test server: %v", err)
+	}
+	t.Cleanup(func() { _ = conn.Close() })
+
+	writeClient, err := bqstorage.NewBigQueryWriteClient(ctx, option.WithGRPCConn(conn))
+	if err != nil {
+		t.Fatalf("Failed to create BigQuery write client: %v", err)
+	}
+
+	_, err = NewBigQueryAgentAnalyticsPluginWithClients(ctx, config, bqClient, writeClient)
+	if err != nil {
+		t.Fatalf("Plugin initialization error: %v", err)
+	}
+
+	if !createCalled {
+		t.Error("Expected table creation to be called")
+	}
+
+	if !strings.Contains(requestBody, "timePartitioning") {
+		t.Errorf("Expected request body to contain 'timePartitioning', got: %s", requestBody)
+	}
+	if !strings.Contains(requestBody, "DAY") {
+		t.Errorf("Expected partitioning type to be 'DAY', got request body: %s", requestBody)
+	}
+	if !strings.Contains(requestBody, "timestamp") {
+		t.Errorf("Expected partitioning field to be 'timestamp', got request body: %s", requestBody)
 	}
 }
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -102,6 +102,10 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		}
 		err = tableRef.Create(ctx, &bq.TableMetadata{
 			Schema: EventsSchema(),
+			TimePartitioning: &bq.TimePartitioning{
+				Field: "timestamp",
+				Type:  bq.DayPartitioningType,
+			},
 			Clustering: &bq.Clustering{
 				Fields: config.ClusteringFields,
 			},


karolpiotrowicz (commented):
The partitioning change itself is right, and it matches both reference implementations exactly: adk-python sets `TimePartitioning(type_=DAY, field="timestamp")` inside its `NotFound` branch and adk-java sets `TimePartitioning.newBuilder(Type.DAY).setField("timestamp")` inside `if (table == null)`, both at create time only. `timestamp` is a top-level `REQUIRED TIMESTAMP` in [`EventsSchema()`](https://github.com/google/adk-go/blob/a5bcaebd31295d4059fa463682972527b8051f23/plugin/agentanalytics/schema.go#L28), which is a legal partition column, and no `RequirePartitionFilter` is set anywhere in the module, so nobody querying the table today has to change a query. Worth noting too that your new test is the first thing in this package ever to execute the table-creation branch — I instrumented that branch with a panic and ran the full suite on your base commit, which stayed green, because every existing test goes through `setupTestPlugin` and its mock answers the metadata request with a 200.

## The eight PRs carry this same change

This is the thing worth deciding before the code. All eight of your open PRs against `plugin/agentanalytics` — #1377 through #1384 — contain the identical `TimePartitioning` block and a test function with the same name and the same assertions. None of them is branched on any other, so each one re-derives the change independently: #1377 and #1378 both sit on the [sequentialagent godoc fix](https://github.com/google/adk-go/commit/0da17d5183cc7affd4bdb7b4075f9e264bb598be), and #1379 through #1384 all sit on the older [adka2a handler-wiring docs commit](https://github.com/google/adk-go/commit/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e). #1377 merges onto `main` cleanly, and #1378 then conflicts on both files.

The practical consequence is that whichever lands first makes the partitioning content of the other seven redundant, and the rest need rebasing. If that layering is intentional it would help to say so in each description, naming the order you want them reviewed in. Otherwise #1377 is the smallest and most self-contained of the set and is the natural one to land first, with the others rebased onto it — but the same argument favors #1378, since it contains this change plus the create-error handling.

## The test passes with the wrong partitioning column

[The third assertion](https://github.com/google/adk-go/blob/a5bcaebd31295d4059fa463682972527b8051f23/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go#L462-L464) cannot verify what it says it verifies. The captured request body contains the serialized schema, in which `timestamp` appears both as `"name":"timestamp"` and inside that field's description, so `strings.Contains(requestBody, "timestamp")` is satisfied whatever `timePartitioning.field` holds. I measured it: change [`Field: "timestamp"`](https://github.com/google/adk-go/blob/a5bcaebd31295d4059fa463682972527b8051f23/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L105-L108) to a column that does not exist and the test still passes, and delete the `Field:` line entirely — which switches the table to ingestion-time partitioning, a different feature — and it still passes. Removing the whole block does fail the test, so the test is not inert, and the `DAY` assertion is load-bearing. It is specifically the column that goes unchecked.

Parsing the body instead of substring-matching it closes this. I ran the following against all four of those variants and it rejects every one, passes unmodified, and leaves `go test -race -count=1 -shuffle=on` green:

```go
var got struct {
	TimePartitioning *struct {
		Field string `json:"field"`
		Type  string `json:"type"`
	} `json:"timePartitioning"`
}
if err := json.Unmarshal([]byte(requestBody), &got); err != nil {
	t.Fatalf("Failed to parse create-table request body: %v", err)
}
if got.TimePartitioning == nil {
	t.Fatalf("Expected request body to contain timePartitioning, got: %s", requestBody)
}
if want := "timestamp"; got.TimePartitioning.Field != want {
	t.Errorf("timePartitioning.field = %q, want %q", got.TimePartitioning.Field, want)
}
if want := "DAY"; got.TimePartitioning.Type != want {
	t.Errorf("timePartitioning.type = %q, want %q", got.TimePartitioning.Type, want)
}
```

## Two things the description should say

Partitioning only reaches a table this code creates. [`Metadata(ctx)`](https://github.com/google/adk-go/blob/a5bcaebd31295d4059fa463682972527b8051f23/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L98-L99) succeeding skips the whole block, so anyone already running the plugin keeps their existing unpartitioned table and gets no log line about it. I confirmed that by driving the real constructor against a mock that answers the metadata request with a 200: `Create` is never called. That is the same behavior as the Python and Java plugins, so it is not a defect, but the description frames the change as fixing scan cost on large tables and the operators with large tables are exactly the ones it will not reach. A sentence saying it applies to newly created tables, and that migrating an existing one means recreating it, would save someone a confused afternoon.

The second is partition expiration. You set no `TimePartitioning.Expiration`, so a newly created table inherits the dataset's `defaultPartitionExpirationMs`. That property had no effect while the table was unpartitioned, and per the [BigQuery documentation](https://cloud.google.com/bigquery/docs/updating-datasets) it also overrides the dataset's default *table* expiration once the table is partitioned. For an operator whose `agent_analytics` dataset carries a retention policy, event rows will start aging out where they previously did not. Setting an expiration in code would be worse, since it would override their policy — this just wants a line in the description. Both siblings behave the same way, so if it is worth a config knob it belongs in adk-python first.

## Smaller things

[`if !createCalled`](https://github.com/google/adk-go/blob/a5bcaebd31295d4059fa463682972527b8051f23/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go#L452-L454) uses `t.Error`, so a genuine "create was never called" failure falls through to the three assertions below it and prints four messages, three of them echoing an empty body. `t.Fatalf` is the right call for a precondition.

The test repeats about 55 lines of [`setupTestPlugin`](https://github.com/google/adk-go/blob/a5bcaebd31295d4059fa463682972527b8051f23/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go#L202) — client, listener, gRPC server, connection, write client — differing only in the mock transport. Taking a `RoundTripper` as a parameter would let both share it, though with seven existing callers that refactor may be better done separately.

[`Field: "timestamp"`](https://github.com/google/adk-go/blob/a5bcaebd31295d4059fa463682972527b8051f23/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L106) is the fifth copy of that column name in the package with nothing tying it to [the schema](https://github.com/google/adk-go/blob/a5bcaebd31295d4059fa463682972527b8051f23/plugin/agentanalytics/schema.go#L28). Nothing is wrong today, and the siblings do the same, but a shared `const` would turn a future rename into a compile error rather than a create-time API error.

plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go
```diff
@@ -378,3 +379,87 @@ func TestLogEvent_ExtractsTraceInfo(t *testing.T) {
...
+		t.Error("Expected table creation to be called")
+	}
+
+	if !strings.Contains(requestBody, "timePartitioning") {
+		t.Errorf("Expected request body to contain 'timePartitioning', got: %s", requestBody)
+	}
+	if !strings.Contains(requestBody, "DAY") {
+		t.Errorf("Expected partitioning type to be 'DAY', got request body: %s", requestBody)
+	}
+	if !strings.Contains(requestBody, "timestamp") {
```
karolpiotrowicz, line 462: This assertion passes whether or not the table is partitioned, so it cannot fail on anything this PR changes.

The serialized create-table request always contains the substring `timestamp`, because [`schema.go:28`](https://github.com/google/adk-go/blob/ddde3d7a79deb697126af066382a0d73e4f437a5/plugin/agentanalytics/schema.go#L28) declares a column of that name. Deleting the whole `TimePartitioning` block and re-running the test shows it directly — the two assertions above fire, this one does not, and the captured body it prints contains `"name":"timestamp","type":"TIMESTAMP"` with no partitioning anywhere in it.

The consequence is that the partition *column* is unconstrained. Change [line 106](https://github.com/google/adk-go/blob/ddde3d7a79deb697126af066382a0d73e4f437a5/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L106) to `Field: "event_type"` — a `STRING` column, which BigQuery rejects for time partitioning — and the whole package suite still passes:

```
$ cd plugin/agentanalytics && go test -count=1 ./...
ok  	google.golang.org/adk/plugin/agentanalytics	2.355s
```

Against real BigQuery that create call would be rejected, and the error is only logged at [lines 113-117](https://github.com/google/adk-go/blob/ddde3d7a79deb697126af066382a0d73e4f437a5/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L113-L117), so the plugin would come up looking healthy and drop every row. That swallow predates this PR, but it is what turns an unguarded field into a silent failure.

Unmarshalling the body instead of substring-matching it fixes the assertion. I ran this version: it passes on your code and fails on the `event_type` mutation with `timePartitioning.field = "event_type", want "timestamp"`.

```go
var got struct {
	TimePartitioning *struct {
		Type  string `json:"type"`
		Field string `json:"field"`
	} `json:"timePartitioning"`
}
if err := json.Unmarshal([]byte(requestBody), &got); err != nil {
	t.Fatalf("create request body is not valid JSON: %v", err)
}
if got.TimePartitioning == nil {
	t.Fatalf("create request has no timePartitioning, got: %s", requestBody)
}
if got.TimePartitioning.Type != "DAY" {
	t.Errorf("timePartitioning.type = %q, want %q", got.TimePartitioning.Type, "DAY")
}
if got.TimePartitioning.Field != "timestamp" {
	t.Errorf("timePartitioning.field = %q, want %q", got.TimePartitioning.Field, "timestamp")
}
```

That also replaces the two assertions above it, since it checks type and field as fields of the same object rather than as three independent substring probes.

karolpiotrowicz (changes requested):
The four production lines are correct, and daily partitioning on `timestamp` is what both sibling implementations already do — adk-python at [`bigquery_agent_analytics_plugin.py:4878-4881`](https://github.com/google/adk-python/blob/a119dd7751082dbbd9a65f71e359abdc2be659cc/src/google/adk/plugins/bigquery_agent_analytics_plugin.py#L4878-L4881) and adk-java at [`BigQueryAgentAnalyticsPlugin.java:226-231`](https://github.com/google/adk-java/blob/4b058cd17346944ac047828eda6424e495204bf9/core/src/main/java/com/google/adk/plugins/agentanalytics/BigQueryAgentAnalyticsPlugin.java#L226-L231), both hardcoded to DAY on the same column, both with no expiry, both only on the create path. Leaving expiry unset is the right choice. The partitioning scheme is fixed when the table is created and cannot be changed afterwards, whereas expiry stays adjustable later through `ALTER TABLE ... SET OPTIONS`, so this change is irreversible in the one dimension where it agrees with the rest of ADK and reversible in the one where an operator might want to differ.

**One thing needs fixing before merge, and it is in the test rather than in the code.** I have left the detail and a reproduction inline on the line concerned. The short version: the third assertion in `TestNewBigQueryAgentAnalyticsPlugin_CreateTable_WithPartitioning` passes whether or not partitioning is configured at all, so nothing in the suite constrains which column the table ends up partitioned on. Changing [line 106](https://github.com/google/adk-go/blob/ddde3d7a79deb697126af066382a0d73e4f437a5/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L106) to a column BigQuery cannot partition on leaves `go test ./...` green. The other two assertions do their job — deleting the whole block fails the test — so this is one assertion to strengthen, not a rewrite.

**Worth adding to the PR description: existing tables are not migrated.** [`Create`](https://github.com/google/adk-go/blob/ddde3d7a79deb697126af066382a0d73e4f437a5/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L103-L112) only runs when the metadata lookup above it fails, so for any deployment whose events table already exists the new setting is never sent. The constructor returns no error and logs nothing on that path, and the plugin reports healthy while continuing to write to an unpartitioned table. That control flow predates this PR and both siblings share the same create-time-only limitation, so it is not something to fix here — but the people most likely to want this change are the ones with a large existing table, and they will get nothing from it without recreating the table. A sentence saying so would save someone a confusing afternoon.

Two smaller things, neither blocking:

- **A one-line comment on the partitioning block would help.** adk-java carries one right above the equivalent code explaining that DAY-on-`timestamp` is a deliberate cross-language parity choice. Without it the next reader cannot tell that `"timestamp"` is constrained by the other implementations rather than picked arbitrarily — which matters, because that literal has no compile-time link to the schema that declares the column.
- **The new test discards the plugin**, so its background flusher outlives the test. This is the existing pattern in `setupTestPlugin` rather than anything new here, so the useful fix is a `t.Cleanup` in that helper rather than a change to this test alone.

Finally, and unrelated to the code: this is one of eight open PRs you have against the same two files, and the other seven all conflict with this one. Whoever merges first sets the rebase order for the rest, so it would be worth saying which order you would like them taken in.

plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go
vasilyl: 1. Replaced all three `strings.Contains` substring probes in `TestNewBigQueryAgentAnalyticsPlugin_CreateTable_WithPartitioning` with structured `json.Unmarshal` into `got.TimePartitioning`, asserting exact values for `got.TimePartitioning.Field == "timestamp"` and `got.TimePartitioning.Type == "DAY"`.
2. Changed the `if !createCalled` guard to `t.Fatalf` so a missing create call fails immediately without falling through.
3. Added `const timestampField = "timestamp"` in `schema.go` so the partitioning field name and schema column definitions share a single compile-time source of truth.
4. Updated the PR description with operational notes covering newly-created-table scope, dataset partition expiration inheritance (`defaultPartitionExpirationMs`), and PR stack ordering.

PR #1378, opened 2026-08-21
feat: Implement automatic schema upgrades for BigQuery logger

vasilyl:

- Closes: N/A

**Problem:**
As the Go ADK analytics plugin updates its schema over time, users with pre-existing BigQuery tables will experience insertion errors due to missing columns, unless they manually migrate or recreate their tables.

**Solution:**
Implement automatic schema upgrades on initialization. During `NewBigQueryAgentAnalyticsPluginWithClients`, the plugin retrieves the existing table metadata, compares its schema against the latest `EventsSchema()`, and automatically updates the BigQuery table schema to add any missing fields or record sub-fields safely.

### Testing Plan

* `TestSchemaFieldsMatch` validates that the schema comparison algorithm correctly identifies top-level and nested structural differences.
* `TestNewBigQueryAgentAnalyticsPlugin_TableExists` verifies that existing tables are inspected and dynamically updated without throwing insertion errors.

### Checklist

### DIFF
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -102,6 +102,10 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		}
 		err = tableRef.Create(ctx, &bq.TableMetadata{
 			Schema: EventsSchema(),
+			TimePartitioning: &bq.TimePartitioning{
+				Field: "timestamp",
+				Type:  bq.DayPartitioningType,
+			},
 			Clustering: &bq.Clustering{
 				Fields: config.ClusteringFields,
 			},
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -321,7 +321,7 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 	return baseplugin.New(cfg)
 }
 
-func schemaFieldsMatch(existing, desired bq.Schema) (newFields bq.Schema, updatedRecords bq.Schema) {
+func schemaFieldsMatch(existing, desired bq.Schema) (newFields, updatedRecords bq.Schema) {
 	existingByName := make(map[string]*bq.FieldSchema)
 	for _, f := range existing {
 		existingByName[f.Name] = f
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -18,11 +18,14 @@ package agentanalytics
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"strings"
 	"time"
 
 	bq "cloud.google.com/go/bigquery"
 	bqstorage "cloud.google.com/go/bigquery/storage/apiv1"
 	"go.opentelemetry.io/otel/trace"
+	"google.golang.org/api/googleapi"
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/v2/agent"
@@ -30,6 +33,11 @@ import (
 	baseplugin "google.golang.org/adk/v2/plugin"
 	"google.golang.org/adk/v2/session"
 	"google.golang.org/adk/v2/tool"
+)
+
+const (
+	schemaVersion         = "1"
+	schemaVersionLabelKey = "adk_schema_version"
 )
 
 // NewBigQueryAgentAnalyticsPlugin creates a newly configured analytics plugin with default config.
@@ -95,24 +103,35 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 
 	// Ensure table exists (done once per instance)
 	tableRef := bqClient.Dataset(config.DatasetID).Table(config.TableName)
-	_, err := tableRef.Metadata(ctx)
+	meta, err := tableRef.Metadata(ctx)
 	if err != nil {
-		if config.Logger != nil {
-			config.Logger.Printf("Table %s not found. Creating it...", config.TableName)
-		}
-		err = tableRef.Create(ctx, &bq.TableMetadata{
-			Schema: EventsSchema(),
-			TimePartitioning: &bq.TimePartitioning{
-				Field: "timestamp",
-				Type:  bq.DayPartitioningType,
-			},
-			Clustering: &bq.Clustering{
-				Fields: config.ClusteringFields,
-			},
-		})
-		if err != nil {
+		if apiErr, ok := err.(*googleapi.Error); ok && apiErr.Code == http.StatusNotFound {
 			if config.Logger != nil {
-				config.Logger.Printf("Failed to create BigQuery table %v: %v", config.TableName, err)
+				config.Logger.Printf("Table %s not found. Creating it...", config.TableName)
+			}
+			err = tableRef.Create(ctx, &bq.TableMetadata{
+				Schema: EventsSchema(),
+				TimePartitioning: &bq.TimePartitioning{
+					Field: "timestamp",
+					Type:  bq.DayPartitioningType,
+				},
+				Clustering: &bq.Clustering{
+					Fields: config.ClusteringFields,
+				},
+				Labels: map[string]string{
+					schemaVersionLabelKey: schemaVersion,
+				},
+			})
+			if err != nil {
+				return nil, fmt.Errorf("failed to create BigQuery table: %w", err)
+			}
+		} else {
+			return nil, fmt.Errorf("failed to retrieve BigQuery table metadata: %w", err)
+		}
+	} else {
+		if config.AutoSchemaUpgrade {
+			if err := maybeUpgradeSchema(ctx, tableRef, meta, config.Logger); err != nil {
+				return nil, fmt.Errorf("failed to auto-upgrade BigQuery table schema: %w", err)
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


karolpiotrowicz (changes requested):
The schema-merge algorithm itself is correct, and I verified that rather than assuming it: driving a real upgrade against a table that carries a column ADK knows nothing about, an operator-edited column description, and a private sub-field inside `content_parts`, the emitted PATCH preserved all three and added the 13 missing columns. Nothing in the merge can drop, rename or retype an existing column. That is the hardest thing this change had to get right.

What blocks it is the surrounding policy: the constructor now treats routine table-metadata maintenance as fatal, and it does so by default. Below, the two failures that no restart clears, then the ones that do clear, then the version-label design, then tests.

## Startup failures that no restart clears

**The plugin now writes to the table even when there is nothing to upgrade, and a denied write stops the service from starting.** When the live schema already matches `EventsSchema()`, [`maybeUpgradeSchema`](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L432-L442) still issues `{"labels":{"adk_schema_version":"1"}}` just to stamp the label, and any error from that becomes a constructor failure at [line 134](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L131-L137). With [`AutoSchemaUpgrade: true`](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/config.go#L89-L93) in `DefaultConfig()`, that makes `bigquery.tables.update` a new hard requirement for every existing installation, including ones whose table is already perfectly correct. A deployment whose service account can append rows and read table metadata but deliberately cannot alter schemas works on `main` today and cannot start after this change. I ran the same mock against `main` to be sure it is this PR and not something pre-existing: `main` returns a working plugin and issues zero metadata writes. Three consecutive restarts against a 403 all fail identically, so there is no recovery.

**A non-404 error from `tables.get` is now fatal too, and `AutoSchemaUpgrade=false` does not switch it off.** [Line 129](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L128-L130) sits outside the `AutoSchemaUpgrade` guard, so a persistent 403 on metadata *read* — an append-only credential, which `main` tolerated by falling through — permanently prevents startup with no configuration that restores the old behaviour. I measured three restarts with the flag both on and off: all six fail. Worth separating from the policy question: the 404 discrimination this is part of is a genuine improvement, because `main` treated every error as "table missing" and then issued a create that could not succeed. It is the fatality, not the diagnosis, that is the problem.

There is a second reason this bites harder than it looks. [`NewBigQueryAgentAnalyticsPlugin`](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L44-L55), the simple four-argument entry point, hardcodes `DefaultConfig()` and takes no `Config`. A user hit by any of this cannot opt out without migrating to a different constructor.

<details><summary>Reproduction for the metadata-write case (mock transport, no live BigQuery)</summary>

```go
// Table exists and its schema already matches EventsSchema() exactly.
// The only outbound write is the version label, and it is denied.
mt := &mockTransport{roundTrip: func(r *http.Request) (*http.Response, error) {
    if r.Method == http.MethodGet && strings.Contains(r.URL.Path, "/tables/test-table") {
        return &http.Response{StatusCode: 200, Body: io.NopCloser(strings.NewReader(
            `{"tableReference":{...},"etag":"e1","schema":{"fields":<EventsSchema()>}}`))}, nil
    }
    if r.Method == http.MethodPatch {
        return &http.Response{StatusCode: 403, Body: io.NopCloser(strings.NewReader(
            `{"error":{"code":403,"message":"Permission bigquery.tables.update denied"}}`))}, nil
    }
    return &http.Response{StatusCode: 200, Body: io.NopCloser(strings.NewReader("{}"))}, nil
}}
_, err := NewBigQueryAgentAnalyticsPluginWithClients(ctx, DefaultConfig(), bqClient, writeClient)

// with the PR:  failed to auto-upgrade BigQuery table schema: failed to update
//               table labels: googleapi: Error 403: Permission ... denied
// on main:      err = nil, plugin non-nil, 0 metadata writes
```
</details>

## Concurrent startup: real, but bounded

Two more paths fail where `main` succeeded, and both are worth fixing even though they are less severe than the above, because they converge rather than repeat.

Replicas starting together all read the same ETag, so one wins the [`Update`](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L409-L411) and the rest get HTTP 412 and fail to construct, with no retry. On a first deployment against a dataset with no table yet, the same thing happens one step earlier: all replicas see 404, all call `Create`, and the losers get `409 Already Exists`, which [line 126](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L125-L127) turns into a constructor error. `main` logged the 409 and carried on, correctly, since the table does exist.

I checked whether these recover, and they do: a restarted replica re-reads, finds the label the winner wrote, and returns early, so the damage is bounded at one crash per losing replica rather than a loop. That is why I have not put them with the two above. It is still a crash-looping rollout on an ordinary concurrent deploy.

One question I could not settle without a live table, and it matters for how bad the 412 case is: does BigQuery bump a table's etag on streaming data writes? If it does, a busy table produces 412s with no concurrent replica involved at all, and the "it converges" argument does not hold.

## The version label decides more than it can know

[`maybeUpgradeSchema` returns immediately](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L375-L378) when the table's `adk_schema_version` label equals the hardcoded `schemaVersion` constant. Nothing couples that constant to `EventsSchema()` — no doc comment on either, no test, no CI check — so the next column added to `schema.go` without a manual bump is silently skipped on every table this release has already stamped. That is the exact failure the PR sets out to fix, re-created and made durable. Confirmed by test: a table labelled `"1"` while missing 15 of the 16 desired columns gets zero upgrade attempts and a successful constructor.

It compounds with how the comparison works. [`schemaFieldsMatch`](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L337) matches on field **name**, and compares `Type` only to decide whether to recurse into a RECORD, never for equality. So a table whose `content` column is STRING where `EventsSchema()` wants JSON reports no difference — and then takes the label-only branch and stamps the table as current, which the short-circuit above makes permanent. Detecting the drift would not let the plugin repair it, since BigQuery cannot retype a column through `tables.update`. The problem is writing down "this table is at schema v1" after verifying only that the column names line up.

A related consequence: the check is equality, not ordering, so a binary at version `"1"` reading a table labelled `"2"` writes it back down to `"1"`. I confirmed the PATCH in that case carries no schema, so nothing but the label churns, but mixed-version fleets will flap it.

## The feature has no test that exercises it

All four constructor tests the PR adds set [`config.AutoSchemaUpgrade = false`](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go#L515-L520), so none of them enters `maybeUpgradeSchema`. That includes `TestNewBigQueryAgentAnalyticsPlugin_TableExists`, which the PR description offers as verifying "that existing tables are inspected and dynamically updated" — it disables the upgrade on line 518 and asserts only that the constructor returns without error. `TestSchemaFieldsMatch` covers the pure comparison function directly, which is genuinely useful, but nothing covers the part that talks to BigQuery.

Three concrete consequences I measured rather than inferred. Deleting the entire `if config.AutoSchemaUpgrade { … }` call site leaves the module suite green. Changing the merge so it drops every pre-existing column — the one thing this feature must never do — also leaves it green, and instrumenting that loop with a counter and running the whole suite gives **zero** executions, so no existing test could catch a regression there however it was written. And the partitioning assertion at [test lines 456-464](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go#L456-L464) is `strings.Contains(requestBody, "timestamp")` against a body that embeds the whole serialized schema, in which `timestamp` is a column name — it passes for any partitioning field, so switching to `event_type` does not fail it either.

I wrote the test that establishes the merge is safe, in your existing `mockTransport` style, asserting on the parsed PATCH body. Happy to hand it over if that saves you time.

## Smaller things

The three new error returns at lines [126](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L125-L127), [129](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L128-L130) and [134](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L131-L137) do not close `writeClient`, where [the existing error path a few lines below](https://github.com/google/adk-go/blob/c5c674f8416e8d2a803fbc55e71a8668b28c1199/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L142-L148) does, and on `main` that function's single error return closed it too. Measured with a client that owns its gRPC connection: 30 failed constructions leak 120 goroutines, against 0 when the client is closed. One constraint on any fix — `...WithClients` is exported and receives both clients, and it never closes `bqClient` on any path, so closing a caller-supplied client there is not obviously right either. The clean hole is `...WithConfig`, which creates the write client itself and drops it if the call it delegates to fails.

`err.(*googleapi.Error)` on the metadata error is a bare type assertion. It works today, because a 404 is non-retryable and the client returns it unwrapped, so this is only hardening: the same client does wrap when a retryable error exhausts the context.

Missing fields are appended verbatim from `EventsSchema()`, `Required` flag included. Only `timestamp` is required today and a table lacking it could never be written to anyway, so nothing breaks now, but it is a trap for the first future column that is not nullable.

## On the stack

This PR contains a byte-identical copy of #1377's only commit, and #1379 carries both changes again on an older base. Each is opened against `main` as though independent, but they are not: merging #1377 first makes this one conflict in both changed files, and merging this one first makes #1379 conflict. Whichever order you pick, the rest need rebasing. It would be much easier to review as a real stack, or with the partitioning change landed on its own first.

karolpiotrowicz:
@wolo-lab flagging this one for a call on behaviour rather than a code review.

It adds automatic BigQuery schema upgrades, and the feature direction is settled by parity — adk-python has the identical feature, the same field name and the same default. What is not settled is how the Go version behaves when that default fires. `DefaultConfig()` sets `AutoSchemaUpgrade: true`, so the constructor issues a table-metadata write at startup and treats any error from it as fatal. A deployment whose credential can append rows and read table metadata but deliberately cannot alter schemas starts on `main` today and cannot start after this change, and a non-404 error from the metadata read is fatal whether the flag is on or off.

Four of the blockers in my review all flow from that one choice, so they are not really four independent defects — decide whether a startup-time metadata write should be on by default and whether its failure should stop the service, and the shape of the fix follows. The other half is the exported `AutoSchemaUpgrade` field itself, which lands new public configuration in `plugin/agentanalytics`, a separately released module.

Happy to re-review once there is a direction.

PR #1379, opened 2026-08-21
feat: Add option to automatically create per-event-type BigQuery views

vasilyl:

- Closes: N/A

**Problem:**
The BigQuery event log table consolidates diverse event types into a single schema where content payloads are serialized as raw JSON. Querying this raw table directly for specific event types is tedious and requires writing repetitive JSON extraction SQL queries.

**Solution:**
Add a configuration setting `CreateViews` (default `true`). When enabled, the plugin automatically and asynchronously creates simplified, typed BigQuery Views for each standard event type (e.g., `v_user_message`, `v_model_request`, `v_tool_call`). These views automatically extract JSON structures into clean, typed top-level columns (e.g., extracting prompts or tokens counts) for easy analytical querying.

### Testing Plan

Added unit tests validating view creation queries are formulated correctly and executed asynchronously without blocking the main thread initialization.

### Checklist

### DIFF
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -102,6 +102,10 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		}
 		err = tableRef.Create(ctx, &bq.TableMetadata{
 			Schema: EventsSchema(),
+			TimePartitioning: &bq.TimePartitioning{
+				Field: "timestamp",
+				Type:  bq.DayPartitioningType,
+			},
 			Clustering: &bq.Clustering{
 				Fields: config.ClusteringFields,
 			},
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -321,7 +321,7 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 	return baseplugin.New(cfg)
 }
 
-func schemaFieldsMatch(existing, desired bq.Schema) (newFields bq.Schema, updatedRecords bq.Schema) {
+func schemaFieldsMatch(existing, desired bq.Schema) (newFields, updatedRecords bq.Schema) {
 	existingByName := make(map[string]*bq.FieldSchema)
 	for _, f := range existing {
 		existingByName[f.Name] = f
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -543,7 +543,8 @@ func createAnalyticsViews(ctx context.Context, client *bq.Client, config Config)
 		allCols = append(allCols, extraCols...)
 		columnsStr := strings.Join(allCols, ",\n  ")
 
-		sql := fmt.Sprintf(viewSQLTemplate,
+		sql := fmt.Sprintf(
+			viewSQLTemplate,
 			config.ProjectID, config.DatasetID, viewName,
 			columnsStr,
 			config.ProjectID, config.DatasetID, config.TableName,
@@ -565,4 +566,3 @@ func createAnalyticsViews(ctx context.Context, client *bq.Client, config Config)
 	}
 	return nil
 }
-
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -38,7 +38,94 @@ import (
 const (
 	schemaVersion         = "1"
 	schemaVersionLabelKey = "adk_schema_version"
+	viewSQLTemplate       = "CREATE OR REPLACE VIEW `%s.%s.%s` AS\nSELECT\n  %s\nFROM\n  `%s.%s.%s`\nWHERE\n  event_type = '%s'"
 )
+
+var viewCommonColumns = []string{
+	"timestamp",
+	"event_type",
+	"agent",
+	"session_id",
+	"invocation_id",
+	"user_id",
+	"trace_id",
+	"span_id",
+	"parent_span_id",
+	"status",
+	"error_message",
+	"is_truncated",
+}
+
+var eventViewDefs = map[string][]string{
+	"USER_MESSAGE": {},
+	"MODEL_REQUEST": {
+		"JSON_VALUE(attributes, '$.model') AS model",
+		"content AS request_content",
+		"JSON_QUERY(attributes, '$.llm_config') AS llm_config",
+		"JSON_QUERY(attributes, '$.tools') AS tools",
+	},
+	"MODEL_RESPONSE": {
+		"JSON_QUERY(content, '$.response') AS response",
+		"CAST(JSON_VALUE(content, '$.usage.prompt') AS INT64) AS usage_prompt_tokens",
+		"CAST(JSON_VALUE(content, '$.usage.completion') AS INT64) AS usage_completion_tokens",
+		"CAST(JSON_VALUE(content, '$.usage.total') AS INT64) AS usage_total_tokens",
+		"CAST(JSON_VALUE(latency_ms, '$.total_ms') AS INT64) AS total_ms",
+		"CAST(JSON_VALUE(latency_ms, '$.time_to_first_token_ms') AS INT64) AS ttft_ms",
+		"JSON_VALUE(attributes, '$.model_version') AS model_version",
+		"JSON_QUERY(attributes, '$.usage_metadata') AS usage_metadata",
+	},
+	"MODEL_ERROR": {
+		"CAST(JSON_VALUE(latency_ms, '$.total_ms') AS INT64) AS total_ms",
+	},
+	"TOOL_START": {
+		"JSON_VALUE(content, '$.tool') AS tool_name",
+		"JSON_QUERY(content, '$.args') AS tool_args",
+		"JSON_VALUE(content, '$.tool_origin') AS tool_origin",
+	},
+	"TOOL_END": {
+		"JSON_VALUE(content, '$.tool') AS tool_name",
+		"JSON_QUERY(content, '$.result') AS tool_result",
+		"JSON_VALUE(content, '$.tool_origin') AS tool_origin",
+		"CAST(JSON_VALUE(latency_ms, '$.total_ms') AS INT64) AS total_ms",
+	},
+	"TOOL_ERROR": {
+		"JSON_VALUE(content, '$.tool') AS tool_name",
+		"JSON_QUERY(content, '$.args') AS tool_args",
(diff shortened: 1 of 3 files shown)


karolpiotrowicz (commented):
Reviewed at `35c5082b0b4286f09e70f86b341c5cba66b6392c`, building and testing inside `plugin/agentanalytics` rather than from the repo root — that module has its own `go.mod`, so a root `./...` run reaches none of it. Build, tests, vet, staticcheck, govulncheck and the API-compatibility check are all clean. Two things below want addressing before this leaves draft, and one finding cuts the other way: much of what looks broken is correct-but-early.

## The view expressions don't match what the plugin writes

Driving each callback and reading the rows that reach the write stream, 35 of the 38 extracted columns come back NULL — only `MODEL_REQUEST.model`, `MODEL_REQUEST.request_content` and `MODEL_RESPONSE.usage_metadata` resolve. The split is the useful part:

- **12 columns are correct as written and simply early.** The five event types nothing emits yet — `STATE_DELTA`, the three `HITL_*`, `A2A_INTERACTION` at [plugin.go#L106-L127](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L106-L127) — all arrive in #1381, and the shapes line up exactly. #1381 passes `map[string]any{"tool": …, "args": …}` as content for the HITL events and puts `state_delta` and `a2a_metadata` into attributes, which is precisely what these definitions read. No change needed.
- **6 columns are also correct and early.** Everything reading `latency_ms` lands once #1383 sets `attrs["latency_ms"]`, which [L306](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L306) routes into that column. Same for `parent_span_id` in the common column list.
- **17 columns are wrong today and stay wrong after the whole series**, because they read the wrong container or the wrong key. No sibling PR moves the tool name into `content` or changes the `TOOL_*` and `MODEL_*` payload shapes.

The last group is what needs fixing, and the tool views are the clearest case. [L81-L82](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L81-L82) reads `JSON_VALUE(content, '$.tool')` and `JSON_QUERY(content, '$.args')`, but [L387-L388](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L387-L388) puts the tool name into attributes and passes the args map as `content` itself:

```
row written for a call to tool "search" with args {"q":"cats"}:
  content    = {"q":"cats"}
  attributes = {"tool_name":"search"}
  latency_ms = NULL

v_tool_start then asks for:
  JSON_VALUE(content, '$.tool')         -> NULL   (name is attributes.tool_name)
  JSON_QUERY(content, '$.args')         -> NULL   (content IS the args map, so the path is '$')
  JSON_VALUE(content, '$.tool_origin')  -> NULL   (#1382 puts tool_origin in attributes)
```

The rest of the group, briefly:

- `v_tool_end.tool_result` reads `$.result` at [L87](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L87) while `content` is the result map.
- `v_agent_start.agent_instruction` reads `$.text_summary` at [L98](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L98) while [L356](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L356) passes `nil` content.
- `v_model_response` reads `$.response` and `$.usage.prompt|completion|total` while the serialized `LLMResponse` carries `$.content` and `$.usageMetadata.promptTokenCount`.
- `llm_config`, `tools` and `model_version` are not written anywhere in the series.

A test asserting that every JSON path in `eventViewDefs` is produced by at least one callback would have caught all 17, and would keep catching them as the payloads move.

## View creation blocks startup

[L226-L230](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L226-L230) calls `createAnalyticsViews` inline, and the loop does `q.Run(ctx)` followed by `job.Wait(ctx)` once per entry with no goroutine anywhere. The description says the views are created "asynchronously without blocking the main thread initialization", and as written every agent process start waits on 17 serialized BigQuery query jobs. Against a stub returning each job in 60 ms, construction goes from 1.8 ms with `CreateViews=false` to 1.036 s with it on, issuing 17 statements. I did not time a real `CREATE OR REPLACE VIEW` job, so treat 1.036 s as a floor set by the stub rather than as the cost you would see.

There is a second, silent consequence. `NewBatchProcessor` derives the processor's lifetime context from the constructor's context at [batch_processor.go#L52](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/batch_processor.go#L52). That coupling predates this PR, but the views loop is now long enough to exhaust a caller's startup deadline before the processor is built. With a 300 ms deadline around construction:

```
CreateViews=false -> 0 statements,  context healthy,   telemetry delivered
CreateViews=true  -> 17 statements, deadline exceeded, no telemetry at all
```

In both cases construction returns a plugin and no error, so the process looks healthy and writes nothing.

I'm not suggesting a shape for the fix. The obvious one — wrapping the call in a goroutine — needs the client's lifetime and the context moved with it, since [L161-L167](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L161-L167) closes `bqClient` as soon as the constructor returns and the caller's context may be cancelled straight after. Whatever the mechanism, the properties that matter are that startup isn't gated on the DDL and that failures still surface somewhere.

## View names collide across tables

[L539](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L539) builds the name from `ViewPrefix` and the event type only, while the `FROM` clause uses `TableName`. Two plugins in one dataset with `TableName=events_app_a` and `events_app_b` generate 17 identical fully-qualified names, and since the template at [L41](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L41) is `CREATE OR REPLACE VIEW`, whichever process starts last repoints all 17 at its own table with no error and no log line. The defaults steer toward one dataset with one table, so this needs a deliberate multi-table layout to hit, but nothing documents `ViewPrefix` as the way out of it.

The same statement also replaces a hand-authored view occupying one of those names, and with `CreateViews` defaulting to `true` at [config.go#L99](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/config.go#L99) that happens on the next restart after a version bump, with no opt-in. I haven't run that against real BigQuery, so I'm stating it as the documented behavior of the statement rather than as something I measured.

## v_user_message contains no user message

`viewCommonColumns` at [L44-L57](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L44-L57) has no `content` and no `attributes`, and `USER_MESSAGE` declares no extras at [L60](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L60). So the view projects twelve metadata columns and drops the message, though `content` is populated for that event. `v_event`, `v_invocation_start` and `v_invocation_end` are in the same position. Since `v_user_message` is the description's headline example, this looks like an oversight.

## Error handling

[L227](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L227) is `if err := createAnalyticsViews(...); err != nil && config.Logger != nil`. The error is never returned to the caller under any configuration — `Logger != nil` gates only whether it gets logged — so a caller cannot tell 17 views created from none. With a nil logger there is no trace at all: every DDL returning 403 against a nil logger still gave back a healthy plugin, no error and no log line. That needs a caller who starts from `DefaultConfig()` and nils `Logger`, since a hand-built `Config` leaves `CreateViews` false, but the `&&` conflating "did it fail" with "can I report it" is worth separating either way.

Separately, the loop ranges a map, so iteration order is randomized, and it returns on the first failure at [L557](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L557). A partial failure leaves a different subset of views on each start, and the operator hears about exactly one of them. A sorted key list and joined errors would make both the outcome and the report deterministic.

## Smaller points

- `ViewPrefix` goes straight into a backtick-quoted identifier at [L546-L552](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L546-L552) with no validation, as do `ProjectID`, `DatasetID` and `TableName`. I don't think this is a security issue — nothing in the module reads config from a file, flag or environment variable, so whoever sets these fields is writing the program that calls `client.Query` anyway. A character-class check still seems worth it, because a backtick or an empty prefix produces a broken or oddly-named object rather than an error. An empty prefix currently yields `` `project.dataset._model_request` ``.
- The doc comments at [config.go#L70-L74](https://github.com/google/adk-go/blob/35c5082b0b4286f09e70f86b341c5cba66b6392c/plugin/agentanalytics/config.go#L70-L74) don't mention that `CreateViews` defaults to on, or that it writes to the user's dataset on every start. Both seem worth saying on the field itself.
- The six test-file edits in this PR all set `config.CreateViews = false`, so nothing exercises `createAnalyticsViews`. The description's testing plan says tests were added validating that the view queries are formulated correctly and run without blocking initialization, and I couldn't find them.

## What I couldn't settle

Everything above about BigQuery's own behavior is inference from the documented semantics rather than something I ran: that `CREATE OR REPLACE VIEW` replaces an existing view, what 17 DDL statements per start cost against dataset metadata quotas when several replicas come up together, and whether requiring `bigquery.jobs.create` on the billing project is acceptable for a telemetry plugin. Each would take one run against a scratch dataset. The quota one is most likely to matter in practice.

One more sequencing note: these branch independently off `main` rather than stacking, so if this lands before #1381 and #1383 the five extra views and the latency columns sit empty until those follow.

wolo-lab:

PR #1380, opened 2026-08-21
feat: Add configuration options to filter logged events via allowlist and denylist

vasilyl:

- Closes: N/A

**Problem:**
In high-traffic production systems, logging every single event (such as every internal tool call or raw LLM trace) to BigQuery can result in massive amounts of log data, leading to excessive storage and ingestion costs.

**Solution:**
Introduce `EventAllowlist` and `EventDenylist` arrays in the configuration. When logging an event, the plugin checks these lists first. If an event is explicitly denied or not explicitly allowed (when an allowlist is defined), the event is ignored and not pushed to the batch processor.

### Testing Plan

Added `TestEventFiltering` which validates that events are allowed or denied correctly according to the configure filters, preventing unauthorized events from being pushed to the stream.

### Checklist

### DIFF
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -102,6 +102,10 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		}
 		err = tableRef.Create(ctx, &bq.TableMetadata{
 			Schema: EventsSchema(),
+			TimePartitioning: &bq.TimePartitioning{
+				Field: "timestamp",
+				Type:  bq.DayPartitioningType,
+			},
 			Clustering: &bq.Clustering{
 				Fields: config.ClusteringFields,
 			},
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -133,6 +133,26 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 
 	// Helper closure func to construct log events
 	logEvent := func(ctx context.Context, eventType string, content any, extraAttrs map[string]any) {
+		if len(config.EventDenylist) > 0 {
+			for _, d := range config.EventDenylist {
+				if d == eventType {
+					return
+				}
+			}
+		}
+		if len(config.EventAllowlist) > 0 {
+			allowed := false
+			for _, a := range config.EventAllowlist {
+				if a == eventType {
+					allowed = true
+					break
+				}
+			}
+			if !allowed {
+				return
+			}
+		}
+
 		row := make(map[string]any)
 		row["timestamp"] = time.Now()
 		row["event_type"] = eventType
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go
@@ -427,3 +427,124 @@ func TestNewBigQueryAgentAnalyticsPlugin_CreateTable_WithPartitioning(t *testing
 		t.Errorf("Expected partitioning field to be 'timestamp', got request body: %s", requestBody)
 	}
 }
+
+func TestEventFiltering(t *testing.T) {
+	ctx := context.Background()
+	config := DefaultConfig()
+	config.Enabled = true
+	config.ProjectID = "test-project"
+	config.DatasetID = "test-dataset"
+	config.TableName = "test-table"
+	config.EventAllowlist = []string{"USER_MESSAGE"}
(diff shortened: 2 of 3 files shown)


karolpiotrowicz (commented):
The filter is placed correctly: it sits at the top of `logEvent`, ahead of the row build and the `SmartTruncate`/`FormatContentParts` work, and the `processor.flush()` in `AfterRunCallback` is a separate statement, so filtering out `INVOCATION_END` does not strand queued rows. With neither list set, behavior is unchanged. Two things want fixing before this leaves draft, and the first is the one I would block on.

## A single typo in `EventAllowlist` silently discards every event

Setting `EventAllowlist` to `[]string{"USER_MESSSAGE"}` — one transposed letter — makes the allowlist non-empty, matches nothing, and drops all twelve event types at [`bigquery_agent_analytics_plugin.go#L143-L154`](https://github.com/google/adk-go/blob/b57204664fadd78d7943b9645f9fcf81c327b6b7/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L143-L154). Nothing is validated at construction and nothing is reported at drop time, so an injected `Logger` records nothing at all and the BigQuery table looks the same as it would for a healthy but quiet agent. A lowercase entry does the same thing, since matching is exact.

The package already handles its other discard path the opposite way — when the queue is full and an event is dropped, that is logged, at [`batch_processor.go#L110-L114`](https://github.com/google/adk-go/blob/b57204664fadd78d7943b9645f9fcf81c327b6b7/plugin/agentanalytics/batch_processor.go#L110-L114).

One constraint on whatever you pick here. Checking the lists against a fixed vocabulary at construction would reject configurations that become valid later — #1381 adds `STATE_DELTA` and six `HITL_*` event types, so a strict startup check written today would refuse a config authored against tomorrow's build.

## The new test pins neither the denylist nor the precedence rule

`TestEventFiltering` sets both lists at [`bigquery_agent_analytics_plugin_test.go#L438-L439`](https://github.com/google/adk-go/blob/b57204664fadd78d7943b9645f9fcf81c327b6b7/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go#L438-L439), and because `MODEL_REQUEST` is absent from the allowlist it is already dropped by the allowlist alone. The phase commented `// Test MODEL_REQUEST (in denylist)`, whose failure message reads "due to denylist" at [`#L526-L531`](https://github.com/google/adk-go/blob/b57204664fadd78d7943b9645f9fcf81c327b6b7/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go#L526-L531), is therefore satisfied without the denylist existing at all. Deleting the whole denylist block leaves the suite green:

```
# delete lines 136-142 of bigquery_agent_analytics_plugin.go, then:
$ go test -C plugin/agentanalytics -run TestEventFiltering -count=1 ./...
ok      google.golang.org/adk/plugin/agentanalytics     0.358s
```

The same disjoint pair leaves the precedence rule unpinned. Rewriting the filter so an allowlist hit skips the denylist — the reverse of today's deny-wins behavior, and observably different for an event type named in both lists — also leaves the test green.

One trap for the obvious fix, worth knowing before you hit it. `AfterRunCallback` emits `INVOCATION_END` itself, so a denylist-only case that reuses this test's setup will see that event arrive and its negative assertion will fail for a reason unrelated to the denylist. Denying `INVOCATION_END` too, or asserting on `event_type` rather than on mere arrival, avoids it.

## Documenting the two fields

`EventAllowlist` and `EventDenylist` are exported configuration on a published module, and [`config.go#L67-L69`](https://github.com/google/adk-go/blob/b57204664fadd78d7943b9645f9fcf81c327b6b7/plugin/agentanalytics/config.go#L67-L69) gives them one shared line. Four behaviors a caller currently has to guess at are worth stating: that the denylist wins when both lists name the same type, that matching is exact and case-sensitive, that a non-nil empty allowlist means "log everything" rather than "log nothing", and that a non-empty allowlist silently excludes any event type added in a later release. The last of those is not hypothetical, given #1381.

## Where the event-type vocabulary should live

The twelve legal values exist only as string literals at the twelve call sites, so a caller has no way to discover them except by reading the plugin source. #1379 already declares the same vocabulary as `eventViewDefs`. Worth deciding once where that list belongs, rather than landing a third copy of it.

## Smaller things

`TestEventFiltering` copies about 47 lines of `setupTestPlugin` instead of parameterising it, which is the natural vehicle for the denylist-only case above.

The explicit `EventAllowlist: nil` and `EventDenylist: nil` in `DefaultConfig` assign the zero value, though keeping them as a statement of intent is defensible.

The positive assertion uses a 1-second budget where the five sibling assertions in the same file use 5 seconds. I ran the test 50 times plain and 30 times under `-race` without a single timeout, so this is a consistency point rather than a live flake.

wolo-lab:

PR #1381, opened 2026-08-21
feat: Pass rich structured event types to OnEventCallback (StateDelta, HITL, A2A)

vasilyl:

- Closes: N/A

**Problem:**
Standard event logging only records basic string/json summaries of agent execution. Rich interactions like human-in-the-loop (HITL) approvals (credentials, confirmations, input), state delta modifications, and agent-to-agent interactions are not captured with structure, making analytical queries difficult.

**Solution:**
Refactor `OnEventCallback` in the analytics plugin to detect and parse structured event types: `STATE_DELTA` updates, `HITL` request events, and Agent-to-Agent (A2A) interaction statistics, serializing them into clear attributes and content blocks.

### Testing Plan

Unit tests verify structured schemas are built properly inside `bigquery_agent_analytics_plugin_test.go` when invoking these callbacks.

### Checklist

### DIFF
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -102,6 +102,10 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		}
 		err = tableRef.Create(ctx, &bq.TableMetadata{
 			Schema: EventsSchema(),
+			TimePartitioning: &bq.TimePartitioning{
+				Field: "timestamp",
+				Type:  bq.DayPartitioningType,
+			},
 			Clustering: &bq.Clustering{
 				Fields: config.ClusteringFields,
 			},
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -17,7 +17,9 @@ package agentanalytics
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	bq "cloud.google.com/go/bigquery"
@@ -236,6 +238,74 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		},
 		OnEventCallback: func(ctx agent.InvocationContext, ev *session.Event) (*session.Event, error) {
 			attrs := map[string]any{"event_author": ev.Author}
+
+			// 1. State Delta Logging
+			if len(ev.Actions.StateDelta) > 0 {
+				stateAttrs := map[string]any{
+					"state_delta": ev.Actions.StateDelta,
+				}
+				for k, v := range attrs {
+					stateAttrs[k] = v
+				}
+				logEvent(ctx, "STATE_DELTA", nil, stateAttrs)
+			}
+
+			// 2. HITL Logging (Injected synthetic tools calls & responses)
+			if ev.Content != nil {
+				for _, part := range ev.Content.Parts {
+					if part.FunctionCall != nil {
+						switch part.FunctionCall.Name {
+						case "adk_request_credential":
+							logEvent(ctx, "HITL_CREDENTIAL_REQUEST", map[string]any{"tool": part.FunctionCall.Name, "args": part.FunctionCall.Args}, attrs)
+						case "adk_request_confirmation":
+							logEvent(ctx, "HITL_CONFIRMATION_REQUEST", map[string]any{"tool": part.FunctionCall.Name, "args": part.FunctionCall.Args}, attrs)
+						case "adk_request_input":
+							logEvent(ctx, "HITL_INPUT_REQUEST", map[string]any{"tool": part.FunctionCall.Name, "args": part.FunctionCall.Args}, attrs)
+						}
+					}
+					if part.FunctionResponse != nil {
+						switch part.FunctionResponse.Name {
+						case "adk_request_credential":
+							logEvent(ctx, "HITL_CREDENTIAL_REQUEST_COMPLETED", map[string]any{"tool": part.FunctionResponse.Name, "result": part.FunctionResponse.Response}, attrs)
+						case "adk_request_confirmation":
+							logEvent(ctx, "HITL_CONFIRMATION_REQUEST_COMPLETED", map[string]any{"tool": part.FunctionResponse.Name, "result": part.FunctionResponse.Response}, attrs)
+						case "adk_request_input":
+							logEvent(ctx, "HITL_INPUT_REQUEST_COMPLETED", map[string]any{"tool": part.FunctionResponse.Name, "result": part.FunctionResponse.Response}, attrs)
+						}
+					}
+				}
+			}
+
+			// 3. A2A Interaction Logging
+			if ev.CustomMetadata != nil {
+				a2aKeys := make(map[string]any)
+				for k, v := range ev.CustomMetadata {
+					if strings.HasPrefix(k, "a2a:") {
+						a2aKeys[k] = v
+					}
+				}
+				if len(a2aKeys) > 0 {
+					a2aTruncatedRaw, _, _ := SmartTruncate(a2aKeys, config.MaxContentLen)
+					var a2aTruncated map[string]any
+					_ = json.Unmarshal(a2aTruncatedRaw, &a2aTruncated)
+
+					responsePayload := a2aKeys["a2a:response"]
+					var contentDict any
+					if responsePayload != nil {
+						contentTruncRaw, _, _ := SmartTruncate(responsePayload, config.MaxContentLen)
+						_ = json.Unmarshal(contentTruncRaw, &contentDict)
+					}
+
+					stateAttrs := map[string]any{
+						"a2a_metadata": a2aTruncated,
+					}
+					for k, v := range attrs {
+						stateAttrs[k] = v
+					}
+					logEvent(ctx, "A2A_INTERACTION", contentDict, stateAttrs)
+				}
+			}
+
 			logEvent(ctx, "EVENT", ev.Content, attrs)
 			return ev, nil
 		},
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go
@@ -15,6 +15,7 @@
 package agentanalytics
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"io"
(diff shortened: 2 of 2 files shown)


karolpiotrowicz (commented):
The port tracks adk-python closely — same event-type strings, same `_HITL_EVENT_MAP` names, same `a2a_metadata` shape — and it adds no public API surface at all, which I checked rather than assumed: `go doc -all` on the package is byte-identical against the merge base, and `apidiff` agrees. Worth noting because the title reads as though `OnEventCallback` now hands richer types to callback implementors. It doesn't. The type at [`plugin/plugin.go:192`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/plugin/plugin.go#L192) is untouched and everything new lives inside this plugin's own closure, so no existing implementation is affected.

Two things I'd want changed before this leaves draft, then the rest.

One scoping note first, since it silently invalidates a green build: `plugin/agentanalytics` is a nested module, and `go build ./...` / `go test ./...` at the repo root reaches none of these lines. Everything below was gated with the module directory as the working tree.

## Session state marked as ephemeral becomes permanent

[`bigquery_agent_analytics_plugin.go:243-251`](https://github.com/google/adk-go/blob/4216568d7805429f85089a71b626325c096149d3/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L243-L251) puts the raw `ev.Actions.StateDelta` into the row's `attributes` column with no filtering. That map still contains `temp:`-prefixed keys at this point. ADK documents those at [`session/session.go:420-424`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/session/session.go#L420-L424) as "Discarded after the invocation completes", and [`internal/sessionutils/utils.go:48`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/internal/sessionutils/utils.go#L48) is where every session backend strips them before persisting. So a value a tool parked in state precisely because it would not outlive the invocation now lands in a durable analytics table instead.

I measured it rather than inferring it — a delta of `{"temp:ephemeral_secret": "..."}` driven through the public `OnEventCallback()` accessor arrives verbatim in `attributes`.

adk-python decided this one already, and the other way: `_recursive_smart_truncate` redacts on exactly this prefix at [`bigquery_agent_analytics_plugin.py:819`](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/plugins/bigquery_agent_analytics_plugin.py#L819), writing `"[REDACTED]"` at [`:831`](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/plugins/bigquery_agent_analytics_plugin.py#L831), alongside a `_SENSITIVE_KEYS` set at [`:451`](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/plugins/bigquery_agent_analytics_plugin.py#L451) covering `access_token`, `password`, `client_secret` and seventeen more. Go's `SmartTruncate` at [`json_formatter.go:56`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/plugin/agentanalytics/json_formatter.go#L56) has no redaction of any kind. That absence predates this PR — the existing `EVENT` row already goes through the same unredacted path — so I'd treat the general redaction gap as its own issue. What is new here is that session state and `CustomMetadata` now flow through it, and `temp:` is the case where the framework has already told the developer the data will be dropped.

## The credential branch writes the credential

[`:266-275`](https://github.com/google/adk-go/blob/4216568d7805429f85089a71b626325c096149d3/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L266-L275) makes `part.FunctionResponse.Response` the `content` of a row whose `event_type` says what is in it. For `adk_request_credential` that response is the credential.

Being precise about what does and doesn't change, because I nearly overstated this: the bytes are already in the table. I instrumented `BatchProcessor.Append` and drove one event carrying a credential FunctionCall/FunctionResponse pair — on the merge base it produces one `EVENT` row whose `content` already contains `functionResponse.response.access_token` in full, and on this branch it produces five rows, one of which is a second copy under a self-describing type. Nobody who couldn't read it before can read it now.

What does change is that a retention or scrub rule scoped to `event_type = 'EVENT'` now leaves a live copy behind, and `DefaultConfig` clusters on `event_type`, so the copy is cheap to select. There's also no `Config` field that turns any of this off.

Reachability is mixed and worth knowing before you decide how much this matters. Nothing in adk-go produces an `adk_request_credential` FunctionCall or FunctionResponse today — the only non-test occurrence is the constant at [`internal/llminternal/contents_processor.go:636`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/internal/llminternal/contents_processor.go#L636), used to keep such events out of model history. But a remote A2A peer can produce one: [`server/adka2a/v2/parts.go:313-319`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/server/adka2a/v2/parts.go#L313-L319) turns a peer data part tagged `function_response` into a `genai.Part` carrying the peer's own `Name` and `Response`, and that event does reach this callback. So the branch is dormant against ADK's own flow and reachable from outside it.

## The three new tests pass with the whole change removed

Restore the production file to its state at the merge base ([0a51e15](https://github.com/google/adk-go/commit/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e)), keep the new tests, and `TestOnEventCallback_StateDelta` ([`:431`](https://github.com/google/adk-go/blob/4216568d7805429f85089a71b626325c096149d3/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go#L431)), `_HITL` ([`:463`](https://github.com/google/adk-go/blob/4216568d7805429f85089a71b626325c096149d3/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go#L463)) and `_A2A` ([`:508`](https://github.com/google/adk-go/blob/4216568d7805429f85089a71b626325c096149d3/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go#L508)) all still pass:

```
git checkout 0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e -- \
  plugin/agentanalytics/bigquery_agent_analytics_plugin.go
go test -C plugin/agentanalytics -count=1 -v \
  -run 'TestOnEventCallback_(StateDelta|HITL|A2A)' ./...
```

```
--- PASS: TestOnEventCallback_StateDelta (0.02s)
--- PASS: TestOnEventCallback_HITL (0.00s)
--- PASS: TestOnEventCallback_A2A (0.01s)
ok      google.golang.org/adk/plugin/agentanalytics     0.207s
```

Each asserts only that some `AppendRowsRequest` arrived within five seconds, and `AfterRunCallback` supplies one unconditionally by logging `INVOCATION_END` and flushing. So nothing checks the row count, the `event_type`, or the payload. The existing tests in this file share the pattern, so this isn't a house-style complaint — but `TestAfterModelCallback_PartialChunksDoNotEachLogUsage` in the same file already reads `GetRowCount()` off the request, and that's the shape these three need to catch a regression in four new row types.

## Smaller things, in rough order of how much they'd bite

- **`A2A_INTERACTION` fires once per streaming chunk.** There's no `ev.Partial` guard at [`:287-296`](https://github.com/google/adk-go/blob/4216568d7805429f85089a71b626325c096149d3/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L287-L296), and `RunOnEventCallback` at [`runner/runner.go:739`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/runner/runner.go#L739) runs for partial events. The client side already guards against exactly this for the request half at [`a2a_agent_run_processor.go:233-238`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/agent/remoteagent/v2/a2a_agent_run_processor.go#L233-L238), whose comment reads "only add request to the final event to avoid massive data duplication during streaming" — the response half is ungated, so it arrives on every chunk. `AfterModelCallback` in this same plugin returns early on `res.Partial`, which looks like the precedent to follow.
- **The gate is broader than Python's, which is why an interaction row can be empty.** Go emits whenever any `a2a:`-prefixed key is present, so metadata carrying only `a2a:task_id` and `a2a:context_id` yields a row with `content: null` — I measured that. Python requires `a2a:request` or `a2a:response` to be present at [`:6886-6889`](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/plugins/bigquery_agent_analytics_plugin.py#L6886-L6889).
- **The `_COMPLETED` rows can't be produced by a human's answer.** That answer arrives as the user message, built and appended in `appendMessageToSession` at [`runner/runner.go:1092`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/runner/runner.go#L1092), which only passes through `RunOnUserMessageCallback` at [`:1045`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/runner/runner.go#L1045) and never reaches this callback. The two REQUEST branches are fine — `adk_request_confirmation` is produced at [`internal/llminternal/functions.go:69`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/internal/llminternal/functions.go#L69) and `adk_request_input` at [`workflow/request_input.go:100`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/workflow/request_input.go#L100), both on agent-authored events.
- **A peer chooses the row type and can rewrite the correlation IDs.** The `switch` at [`:257`](https://github.com/google/adk-go/blob/4216568d7805429f85089a71b626325c096149d3/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L257) matches on `FunctionCall.Name` alone with no check that the framework synthesized it, and `maps.Copy(event.CustomMetadata, um)` at [`server/adka2a/v2/metadata.go:160`](https://github.com/google/adk-go/blob/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92/server/adka2a/v2/metadata.go#L160) lets peer-supplied metadata win over the IDs ADK derived from the transport. The payload was already logged either way, so what's at stake is whether a dashboard counting HITL events or joining on `a2a:task_id` can be trusted.
- **Integers above 2^53 lose precision on the A2A path only.** `SmartTruncate` marshals, [`:290`](https://github.com/google/adk-go/blob/4216568d7805429f85089a71b626325c096149d3/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L290) unmarshals into `any` so every number becomes `float64`, then `logEvent` marshals again. `int64(1234567890123456789)` comes out as `1234567890123456800`. Control from the same run: the identical value through `StateDelta`, which marshals once, is written exactly. Passing `a2aKeys` and the response straight to `logEvent` would remove the round trip. Whether production ever puts a raw `int64` there depends on `converters.ToMapStructure`, which I didn't trace.
- **The `EVENT` row is appended last** at [`:309`](https://github.com/google/adk-go/blob/4216568d7805429f85089a71b626325c096149d3/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L309), after up to N HITL rows, so it is the first one shed when the queue hits the drop arm in `batch_processor.go`. That inverts the priority you'd want.
- **Nothing joins the rows.** One event can now produce five, sharing only `invocation_id` and separate `time.Now()` stamps. `session.Event.ID` and `FunctionCall.ID` are both to hand, and without one of them two concurrent `adk_request_confirmation` calls give REQUEST and COMPLETED rows that can't be paired.
- **A state value `json.Marshal` rejects takes the batch with it.** `math.NaN()` makes `SmartTruncate` return nil bytes, `attributes` becomes `""`, and that column is typed `JSON`. The same door already exists via tool results, so this widens the input domain rather than opening it.
- **Four discarded errors** at [`:288-296`](https://github.com/google/adk-go/blob/4216568d7805429f85089a71b626325c096149d3/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L288-L296) mean a marshal failure silently produces a row with empty `a2a_metadata` and nothing recording that metadata was lost.
- **The tool names are bare literals.** `toolconfirmation.FunctionCallName` and `workflow.WorkflowInputFunctionCallName` are both exported and both equal to two of the three. The credential one is unexported and out of reach from this module, so this can only be two-thirds symbolic.
- **The A2A fixture uses a shape production doesn't produce.** [`:522`](https://github.com/google/adk-go/blob/4216568d7805429f85089a71b626325c096149d3/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go#L522) sets `a2a:response` to a Go string, where the real path supplies a `map[string]any`. The string form lands in `content` as a JSON scalar rather than an object, so `JSON_VALUE(content, '$.result')` behaves differently in the test than in production.

Three things I checked and found fine, so nobody re-derives them. Writing the `a2a:response` payload into both `attributes.a2a_metadata` and `content` looks redundant but matches adk-python exactly, so it's port fidelity. The double truncation on that path is idempotent and `is_truncated` still comes out true. And a nil `*genai.Part` in `ev.Content.Parts` would panic at [`:256`](https://github.com/google/adk-go/blob/4216568d7805429f85089a71b626325c096149d3/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L256), but `FormatContentParts` already dereferences the same element on the same call today, so that's not new.

Gates for the record, run against the merge of this branch into `main` at [2d8e76b2](https://github.com/google/adk-go/commit/2d8e76b2ae905beb237e2f3be03e5dbf8d996f92), scoped to the nested module: build, `go test -race -shuffle=on`, `gofmt`, `go vet` and `govulncheck` all clean, and the single `staticcheck` and `golangci-lint` findings are byte-identical on `main`. The branch merges cleanly.

wolo-lab:

PR #1382, opened 2026-08-21
feat: Implement tool provenance tracking for tool execution logging

vasilyl:

- Closes: N/A

**Problem:**
During complex multi-agent executions or nested tool invocations, it is difficult to trace the origin of a tool (e.g., whether it is a local function, an MCP tool, a sub-agent, or a transfer agent). This lack of provenance makes debugging and analytics on tool usage patterns highly challenging.

**Solution:**
Introduce a `getToolOrigin(t tool.Tool) string` helper inside `bigquery_agent_analytics_plugin.go` to inspect the tool's type and return its origin classification (`LOCAL`, `SUB_AGENT`, `MCP`, `A2A`, `TRANSFER_AGENT`, or `UNKNOWN`). Use this to populate a `tool_origin` attribute in telemetry logs for `TOOL_START`, `TOOL_END`, and `TOOL_ERROR` event types.

### Testing Plan

Added a unit test `TestToolProvenance` to verify that `getToolOrigin` correctly identifies `LOCAL` function tools and `SUB_AGENT` agent tools.

### Checklist

### DIFF
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -102,6 +102,10 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		}
 		err = tableRef.Create(ctx, &bq.TableMetadata{
 			Schema: EventsSchema(),
+			TimePartitioning: &bq.TimePartitioning{
+				Field: "timestamp",
+				Type:  bq.DayPartitioningType,
+			},
 			Clustering: &bq.Clustering{
 				Fields: config.ClusteringFields,
 			},
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -18,6 +18,8 @@ package agentanalytics
 import (
 	"context"
 	"fmt"
+	"reflect"
+	"strings"
 	"time"
 
 	bq "cloud.google.com/go/bigquery"
@@ -272,18 +274,25 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		},
 
 		BeforeToolCallback: func(ctx agent.Context, t tool.Tool, args map[string]any) (map[string]any, error) {
-			attrs := map[string]any{"tool_name": t.Name()}
+			attrs := map[string]any{
+				"tool_name":   t.Name(),
+				"tool_origin": getToolOrigin(t),
+			}
 			logEvent(ctx, "TOOL_START", args, attrs)
 			return nil, nil
 		},
 		AfterToolCallback: func(ctx agent.Context, t tool.Tool, args, res map[string]any, err error) (map[string]any, error) {
-			attrs := map[string]any{"tool_name": t.Name()}
+			attrs := map[string]any{
+				"tool_name":   t.Name(),
+				"tool_origin": getToolOrigin(t),
+			}
 			logEvent(ctx, "TOOL_END", res, attrs)
 			return nil, nil
 		},
 		OnToolErrorCallback: func(ctx agent.Context, t tool.Tool, args map[string]any, err error) (map[string]any, error) {
 			attrs := map[string]any{
 				"tool_name":     t.Name(),
+				"tool_origin":   getToolOrigin(t),
 				"error_message": err.Error(),
 			}
 			logEvent(ctx, "TOOL_ERROR", nil, attrs)
@@ -301,3 +310,34 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 
 	return baseplugin.New(cfg)
 }
+
+func getToolOrigin(t tool.Tool) string {
+	tType := reflect.TypeOf(t).String()
+	if strings.Contains(tType, "mcptoolset.mcpTool") {
+		return "MCP"
+	}
+	if strings.Contains(tType, "agenttool.agentTool") {
+		val := reflect.ValueOf(t)
+		if val.Kind() == reflect.Ptr {
+			val = val.Elem()
+		}
+		if val.Kind() == reflect.Struct {
+			agentField := val.FieldByName("agent")
+			if agentField.IsValid() {
+				agentType := agentField.Type().String()
+				if strings.Contains(agentType, "remoteagent.a2aAgent") || strings.Contains(agentType, "remoteagent/v2.a2aAgent") {
+					return "A2A"
+				}
+			}
+		}
+		return "SUB_AGENT"
+	}
+	if strings.Contains(tType, "functiontool.functionTool") {
+		return "LOCAL"
+	}
+	if strings.Contains(tType, "TransferToAgent") {
+		return "TRANSFER_AGENT"
+	}
+	return "UNKNOWN"
+}
+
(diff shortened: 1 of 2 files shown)


karolpiotrowicz (commented):
Reviewed this as a draft, so treat everything below as early feedback rather than a change request. The core mechanism has a problem worth knowing about before you build further on it, and the sibling implementations turn out to have already solved the same problem in a way that is worth borrowing.

One process note first: `plugin/agentanalytics` is a nested module, so `go build ./...` and `go test ./...` at the repo root reach none of this code — 155 packages, none in the module. Everything below was run with `-C plugin/agentanalytics`. Repo CI does cover it properly, so this only bites people reviewing locally.

## The A2A branch cannot fire

[`getToolOrigin`](https://github.com/google/adk-go/blob/0d9a26f81ca0b6b65436180f4179e12e739dff77/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L327-L330) reads `agentField.Type()`, which returns the field's *declared* type. `agentTool.agent` is declared as the interface `agent.Agent`, so that string is `"agent.Agent"` for every tool, and neither `"remoteagent.a2aAgent"` nor `"remoteagent/v2.a2aAgent"` can ever match it.

Reading the dynamic type instead would not rescue it. `remoteagent.NewA2A` builds an internal `a2aAgent` but returns `agent.New(...)` with that value captured in the `Run` closure, so an A2A agent's concrete type is `*agent.agent` — identical to any other custom agent. There is no reflect-visible difference at all.

Measured against published `adk/v2 v2.0.0`: a genuine remote A2A agent, wrapped with `agenttool.New`, classifies as `SUB_AGENT`. On the same two objects, the state the framework does record differs as you would want — `AgentType` reads `"RemoteAgent"` versus `"CustomAgent"` — so the information exists, it is just not in the Go type. That field lives in an `internal/` package, though, so reaching it deepens exactly the coupling in the next section.

```go
a2aAg, _ := remoteagent.NewA2A(remoteagent.A2AConfig{
    Name: "remote_helper", Description: "remote",
    AgentCard: &a2a.AgentCard{Name: "remote_helper", Description: "remote"},
})
custom, _ := agent.New(agent.Config{Name: "plain", Description: "custom", Run: noopRun})

fmt.Println(getToolOrigin(agenttool.New(a2aAg, nil)))   // SUB_AGENT  (want A2A)
fmt.Println(getToolOrigin(agenttool.New(custom, nil)))  // SUB_AGENT

v := reflect.ValueOf(agenttool.New(a2aAg, nil)).Elem().FieldByName("agent")
fmt.Println(v.Type().String())          // agent.Agent   <- static, always
fmt.Println(v.Elem().Type().String())   // *agent.agent  <- dynamic, still not a2a
```

Run with `go test -C plugin/agentanalytics ./...`.

</details>

## Both siblings already solve this, and neither uses type names

adk-python asks the real type: [`isinstance(tool.agent, RemoteA2aAgent)`](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/plugins/bigquery_agent_analytics_plugin.py#L367-L372). adk-java has the agent declare its own origin — [`BaseAgent.toolOrigin()`](https://github.com/google/adk-java/blob/5d8ef10812ceedbd9f3c23f3e345f3a43eb98981/core/src/main/java/com/google/adk/agents/BaseAgent.java#L266-L268) returns `BASE_AGENT` and [`RemoteA2AAgent` overrides it](https://github.com/google/adk-java/blob/5d8ef10812ceedbd9f3c23f3e345f3a43eb98981/a2a/src/main/java/com/google/adk/a2a/agent/RemoteA2AAgent.java#L605-L608) to return `A2A` — and the plugin just calls it.

That is the shape this branch is transliterating, and the transliteration is what fails: Python's `isinstance` reads the dynamic type, and the Go equivalent reads the static one. Since Go's `NewA2A` does not return a distinct type at all, the Java approach is the one that ports — which needs a small addition to `adk/v2` rather than more reflection here.

Two smaller parity points. Python declares a seventh value, `TRANSFER_A2A`, classified per-call by resolving the transfer target. The Go callbacks do receive `ctx` and `args`, so that is available if you want it, just not passed. Separately, in Python `SUB_AGENT` is defined as the else-branch of the A2A test, which is why emitting it for an A2A tool reads as a wrong value rather than a vaguer one.

## Renames upstream will degrade this silently

The classifier matches four unexported identifiers in a different module, and that module is pinned to published `v2.0.0` with no `replace`. Renaming any of them upstream is a legal, non-breaking change. I mutated each one and re-ran the suite:

| renamed | suite |
|---|---|
| `functiontool.functionTool` | fails — pinned by `TestToolProvenance` |
| `agenttool.agentTool` | fails — pinned |
| `mcptoolset.mcpTool` | **passes** |
| `TransferToAgent` | **passes** |
| the reflected `agent` field | **passes** |

`apidiff` cannot help here — it inspects exported API, and these are unexported and in a dependency — so nothing in CI would ever catch it. Two of the five are already pinned by your test, so extending that pattern to the rest is the cheap fix. All six branches are pinnable with locally declared shim types, including the ones that look untestable from this module.

## The new test cannot catch any of the above

[`TestToolProvenance`](https://github.com/google/adk-go/blob/0d9a26f81ca0b6b65436180f4179e12e739dff77/plugin/agentanalytics/bigquery_agent_analytics_plugin_test.go#L451-L455) wraps a nil agent, so the field's static type is `"agent.Agent"` and the assertion passes identically whether the A2A logic works or is deleted outright. Deleting the entire A2A block leaves the suite green. So does changing `UNKNOWN` to the empty string, and so does dropping the `tool_origin` key from the `TOOL_END` call site.

## Smaller things

Matching on `reflect.TypeOf(t).String()` compares the package *name*, not the import path, so a package called `mcptoolset` with an `mcpTool` type is recorded as `MCP` wherever it comes from. Anchoring on `PkgPath()` plus an exact `Name()` would close it. This needs code compiled into the binary, so it is a data-fidelity point rather than a security one — worth noting only because the column reads as attestation.

`strings.Contains(tType, "TransferToAgent")` has no package qualifier, unlike its three neighbours. A fully qualified literal costs nothing.

Of eight built-in tools I constructed, five return `UNKNOWN` — `geminitool.GoogleSearch`, `geminitool.New`, `loadartifactstool.New`, `loadmemorytool.New`, `preloadmemorytool.New` — while `exitlooptool.New` returns `LOCAL` purely because it is built on `functiontool`. That is defensible, but it means `LOCAL` currently means "implemented with functiontool" rather than "written by the application developer", and nothing says which is intended. A doc comment naming the six values and what each one means would settle that, and would also give an analyst something to write a query against.

[Line 321](https://github.com/google/adk-go/blob/0d9a26f81ca0b6b65436180f4179e12e739dff77/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L321) uses `reflect.Ptr` where the rest of the repo uses `reflect.Pointer` six times to two.

For what it is worth, the parts that are easy to get wrong are right. The unexported field is read with `Type()` rather than `Interface()`, which would panic. The pointer and validity guards are correct, the return values are six fixed literals so nothing caller-controlled reaches the row, and it costs 119 ns with zero allocations.

One thing I did not check: whether `MCP` and `TRANSFER_AGENT` ever fire for a real tool. The type names line up, but I only produced `MCP` from a package I wrote to mimic one, and I never exercised the transfer tool at all.

wolo-lab:

PR #1383, opened 2026-08-21
feat: Implement a contextual trace stack and manager to dynamically trace nested execution and spans

vasilyl:

- Closes: N/A

**Problem:**
In concurrent or asynchronous agent execution flows, standard flat context logs lack parent-child span structure, making it incredibly difficult to trace nested execution layers (e.g., which agent invocation triggered which tool or LLM call).

**Solution:**
Implement a thread-safe in-memory contextual trace stack (`spanStack` and `traceManager`) inside `bigquery_agent_analytics_plugin.go`. Push nested execution span contexts (`spanRecord`) onto the invocation's trace stack at execution boundaries (Runs, Models, Tools), linking children to parents via `parentSpanID`. Retrieve trace context when logging events, and ensure safe cleanup upon run completion.

### Testing Plan

Added a complete integration unit test `TestContextualTraceStacking` which triggers `BeforeRunCallback`, `BeforeModelCallback`, `AfterModelCallback`, and `AfterRunCallback` sequentially, verifying that:
1. A root span is correctly pushed.
2. Nested spans are linked to the root span as `parentSpanID`.
3. Stack sizes update correctly as spans are popped.
4. The invocation's trace stack is properly cleaned up and removed on execution end.

### Checklist

### DIFF
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -102,6 +102,10 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		}
 		err = tableRef.Create(ctx, &bq.TableMetadata{
 			Schema: EventsSchema(),
+			TimePartitioning: &bq.TimePartitioning{
+				Field: "timestamp",
+				Type:  bq.DayPartitioningType,
+			},
 			Clustering: &bq.Clustering{
 				Fields: config.ClusteringFields,
 			},
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -603,4 +603,3 @@ func (s *spanStack) getStartTime(spanID string) time.Time {
 	}
 	return time.Time{}
 }
-
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -18,7 +18,11 @@ package agentanalytics
 import (
 	"context"
 	"fmt"
+	"strings"
+	"sync"
 	"time"
+
+	"github.com/google/uuid"
 
 	bq "cloud.google.com/go/bigquery"
 	bqstorage "cloud.google.com/go/bigquery/storage/apiv1"
@@ -137,6 +141,7 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		row["timestamp"] = time.Now()
 		row["event_type"] = eventType
 
+		var invocationID string
 		if rCtx, ok := ctx.(agent.ReadonlyContext); ok {
 			if rCtx.AgentName() != "" {
 				row["agent"] = rCtx.AgentName()
@@ -145,6 +150,7 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 				row["session_id"] = rCtx.SessionID()
 			}
 			row["invocation_id"] = rCtx.InvocationID()
+			invocationID = rCtx.InvocationID()
 			row["user_id"] = rCtx.UserID()
 		} else if iCtx, ok := ctx.(agent.InvocationContext); ok {
 			if iCtx.Agent() != nil && iCtx.Agent().Name() != "" {
@@ -157,6 +163,7 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 				row["user_id"] = iCtx.Session().UserID()
 			}
 			row["invocation_id"] = iCtx.InvocationID()
+			invocationID = iCtx.InvocationID()
 		}
 
 		isTruncated := false
@@ -208,11 +215,30 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		}
 		row["is_truncated"] = isTruncated
 
-		// Add trace info dynamically resolving the context span
-		spanCtx := trace.SpanContextFromContext(ctx)
-		if spanCtx.IsValid() {
-			row["trace_id"] = spanCtx.TraceID().String()
-			row["span_id"] = spanCtx.SpanID().String()
+		stack := tm.getStack(invocationID)
+		if stack != nil {
+			tID, sID, pID := stack.getCurrentSpan(ctx)
+			if overrideSpanID, ok := extraAttrs["span_id_override"].(string); ok && overrideSpanID != "" {
+				sID = overrideSpanID
+			}
+			if overrideParentSpanID, ok := extraAttrs["parent_span_id_override"].(string); ok && overrideParentSpanID != "" {
+				pID = overrideParentSpanID
+			}
+
+			if tID != "" {
+				row["trace_id"] = tID
+				row["span_id"] = sID
+				if pID != "" {
+					row["parent_span_id"] = pID
+				}
+			}
+		} else {
+			// Fallback to ambient context
+			spanCtx := trace.SpanContextFromContext(ctx)
+			if spanCtx.IsValid() {
+				row["trace_id"] = spanCtx.TraceID().String()
+				row["span_id"] = spanCtx.SpanID().String()
+			}
 		}
 
 		processor.Append(row)
@@ -227,12 +253,24 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 			return msg, nil
 		},
 		BeforeRunCallback: func(ctx agent.InvocationContext) (*genai.Content, error) {
+			// Initialize stack and push root span
+			stack := tm.getStack(ctx.InvocationID())
+			stack.pushSpan(ctx, "invocation_root")
+
 			logEvent(ctx, "INVOCATION_START", nil, nil)
 			return nil, nil
 		},
 		AfterRunCallback: func(ctx agent.InvocationContext) {
 			logEvent(ctx, "INVOCATION_END", nil, nil)
+
+			// Pop the root span
+			stack := tm.getStack(ctx.InvocationID())
+			stack.popSpan()
+
 			processor.flush() // flush queue synchronously
+
+			// Clean up stack map to prevent memory leaks
+			tm.removeStack(ctx.InvocationID())
 		},
 		OnEventCallback: func(ctx agent.InvocationContext, ev *session.Event) (*session.Event, error) {
 			attrs := map[string]any{"event_author": ev.Author}
@@ -250,6 +288,9 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		},
 
 		BeforeModelCallback: func(ctx agent.Context, req *model.LLMRequest) (*model.LLMResponse, error) {
+			stack := tm.getStack(ctx.InvocationID())
+			stack.pushSpan(ctx, "llm_request")
+
 			attrs := map[string]any{"model": "unknown"}
 			if req.Model != "" {
 				attrs["model"] = req.Model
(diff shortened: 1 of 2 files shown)


karolpiotrowicz (commented):
Reviewed as a draft, so this is early feedback rather than a merge gate. The trace stack is the right shape for a sequential agent, but the framework calls these callbacks concurrently and in ways that do not pair push with pop, and that breaks span attribution on paths that are ordinary rather than exotic. Everything below was reproduced end to end through `runner.Run` with a scripted in-process model, and the row values come from decoding the Arrow batch the plugin actually sends.

One note on CI before anything else: `github.com/google/uuid` is imported directly at [`bigquery_agent_analytics_plugin.go:25`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L25) but is still `// indirect` at [`go.mod:33`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/go.mod#L33), so `go mod tidy -diff` fails and the `test (./plugin/agentanalytics)` job stops there. The job log has no `go build` and no `go test` step at all, so nothing green on this commit tells you anything about the module.

## Concurrent tool calls share one stack

`platform.RunTasks` gives every function call of a single model turn [its own goroutine](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/platform/exec.go#L88-L93), and [`base_flow.go:1242`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/internal/llminternal/base_flow.go#L1242) hands it the whole batch under one invocation ID. They therefore all reach the same `spanStack`. [`pushSpan`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L494-L531) takes its parent from whatever is on top, so with two tools in flight one concurrent sibling is recorded as the parent of the other — a nesting that never happened.

The pop side is worse, because [`popSpan`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L533-L547) returns the top of the stack rather than the caller's own span, and [`AfterToolCallback`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L390-L407) reads the parent and pops in two separate acquisitions of the mutex. A fast tool finishing alongside a slow one gets the slow one's `span_id` and the slow one's elapsed time under its own `tool_name`. Locking is not the issue — every access is correctly guarded, and the race detector stays silent on all of this. It is the data structure: a LIFO keyed on invocation ID cannot represent a fan-out.

The same shared-stack reach exists through [`workflow/scheduler.go:405`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/workflow/scheduler.go#L405), [`workflow/parallel_worker.go:124`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/workflow/parallel_worker.go#L124) and [`parallelagent`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/agent/workflowagents/parallelagent/agent.go#L88-L98), all of which run concurrently under one invocation ID, so a fix aimed only at `RunTasks` would be incomplete.

<details><summary>Reproduction — two tools in one turn, barrier-synchronised so both are in flight</summary>

```
--- FAIL: TestRepro_ConcurrentToolCalls_SiblingBecomesParent
    stack while both tools in flight: depth=3
      spans   = [2216a6ec35a84115 f253e92dcf724d82 fc307a0349da4f42]
      parents = [                  2216a6ec35a84115 f253e92dcf724d82]
    BUG: concurrent sibling recorded as parent. root=2216a6ec35a84115,
         tool span parents=[2216a6ec35a84115 f253e92dcf724d82] — the second
         tool's parent is the first tool's span, not the root.
```

A scripted model returns two `FunctionCall` parts. Both tools rendezvous on a barrier so neither returns until both have started. Run with `go test -race -run TestRepro_ConcurrentToolCalls ./...` in `plugin/agentanalytics`. It reports no data race.
</details>

## Push and pop do not pair on four paths

[`callTool`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/internal/llminternal/base_flow.go#L1276-L1296) calls `RunOnToolErrorCallback` and then calls `RunAfterToolCallback` unconditionally. The plugin pops in both, so any tool that returns an error consumes two spans for one push and takes the invocation root off the stack. From then on the rest of the invocation is pushed onto an empty stack, which mints a fresh trace ID at [`pushSpan`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L515-L520), so one invocation ends up written under two traces.

The same underflow happens with no error at all when the model names a tool that is not registered: [`base_flow.go:1120-1125`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/internal/llminternal/base_flow.go#L1120-L1125) calls the error callback with a `fakeTool` and never calls `BeforeTool`, so the pop has no matching push. One hallucinated tool name split a single invocation across four `trace_id` values in the reproduction, and after the underflow most rows carried no `trace_id`, `span_id` or `parent_span_id` at all, because the guard at [`:228`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L228) suppresses all three when the stack is empty.

It also fails in the other direction. A `BeforeModelCallback` that returns a cached response — [the documented use](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/agent/llmagent/llmagent.go#L225-L231) — or one that rejects a prompt by returning an error makes [`callLLM`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/internal/llminternal/base_flow.go#L788-L796) yield and return before `runAfterModelCallbacks`, so the span pushed at [`:292`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L290-L292) is never popped and every later row of that invocation, `INVOCATION_END` included, is attributed to it. Symmetrically, an `OnModelErrorCallback` that returns a recovery response lets the flow continue into `runAfterModelCallbacks`, giving one push two pops.

Worth flagging as a constraint on whichever fix you pick: simply deleting the pop in [`OnModelErrorCallback`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L362-L378) would leak the span on the ordinary model-error path, where the iterator returns before `AfterModel` ever runs.

<details><summary>Reproduction — one failing tool, then a second tool in the same invocation</summary>

```
--- FAIL: TestRepro_ToolError_DoublePop_SplitsTrace
    during the failing tool: depth=2 traces=[01e7ea6c09e3481aa158552d16cda999 ...]
    during a later tool   : depth=1 traces=[8f5e579d0ff1479daef10dc82f13310f]
    BUG: after the failing tool the stack lost the invocation root
    BUG: one invocation split across two trace IDs
```
</details>

## latency_ms is encoded twice

The four callbacks marshal `latencyMap` and store the resulting **string** into `attrs["latency_ms"]` — for example [`:401-403`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L401-L403). [`logEvent`'s `case "latency_ms"`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L201-L205) then runs `SmartTruncate` on that string, which marshals it a second time. Decoding the Arrow batch for a 5 ms tool call, the cell is literally `"{\"total_ms\":5}"`, outer quotes included, in a column declared [`bq.JSONFieldType`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/schema.go#L65). The contrast is `attributes` on the same row, which passes a live `map[string]any` at [`:212`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L211-L213) and comes out as a real object. I did not run this against real BigQuery, so I can tell you what the bytes are but not what a JSON-path query over them returns.

That same decode shows two other things. `span_id_override` and `parent_span_id_override` are read at [`:221-226`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L221-L226) but never removed, so the [`default` branch](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L206-L207) copies them into the user-facing `attributes` column — a decoded `TOOL_END` reads `{"parent_span_id_override":"761c…","span_id_override":"1132…","tool_name":"alpha"}`. And `MODEL_ERROR` rows come out with no `parent_span_id` while the `MODEL_RESPONSE` for the equivalent operation carries the root: the cause is that `OnModelErrorCallback` pops at [`:364`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L362-L365) before `logEvent` reads the current span, where its three siblings read the parent first. The `"parent_span_id_override": ""` on the next line is a red herring — `:224` ignores an empty override.

## Two trace IDs per invocation once OpenTelemetry is configured

This is the one I would most want a second opinion on, because it fires on a completely healthy invocation. Install a real `TracerProvider` and run one model call and one successful tool call, with no errors and no concurrency, and the rows come out under two traces:

```
INVOCATION_START trace=efe55835d8b2434ca8876dcae4d93735 span=c66f92a39f61402a
MODEL_REQUEST    trace=3c9a74c647a730a2d9c37104233266d0 span=fa8383ca4a294c88
MODEL_RESPONSE   trace=efe55835d8b2434ca8876dcae4d93735 span=fa8383ca4a294c88
TOOL_START       trace=3c9a74c647a730a2d9c37104233266d0 span=04f4fbec11c04b08
TOOL_END         trace=efe55835d8b2434ca8876dcae4d93735 span=04f4fbec11c04b08
INVOCATION_END   trace=efe55835d8b2434ca8876dcae4d93735 span=c66f92a39f61402a
```

`MODEL_REQUEST` and `MODEL_RESPONSE` are the same span with two different trace IDs. At `BeforeRunCallback` no ambient span exists yet, so [`pushSpan`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L508-L520) mints a synthetic UUID trace for the root. Later pushes see the real `generate_content` and `execute_tool` spans and adopt the real trace, and `logEvent` reports whichever record happens to be on top. Without an SDK there is no split, because every push inherits the root's synthetic trace — so this only shows up in exactly the deployment the feature is for.

Sitting underneath that is a decision worth writing down explicitly rather than leaving implicit in the diff. Before this change `span_id` was the ambient OpenTelemetry span ID and joined against a tracing backend. Now it is a UUID prefix, while `trace_id` still carries the real ambient value when there is one. The columns are documented as [OpenTelemetry IDs](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/schema.go#L34-L36) and those descriptions ship to BigQuery. Either the plugin owns synthetic IDs and the descriptions change, or it reuses the ambient span IDs and the stack only supplies parenting. The mixture is what does not work.

## Memory held after a run

`RunLive` creates the stack entry at [`runner.go:481`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/runner/runner.go#L479-L483) while the matching cleanup is deferred inside `wrappedIter` at [`:509`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/runner/runner.go#L507-L511), so a failed live connect, the early-exit branch, or a caller that never ranges the iterator all skip it. Twenty failed `RunLive` calls left twenty entries in the package-level map, which `CloseFunc` does not touch. `runner.Run` is fine — it registers the defer before calling `BeforeRun` — and I measured zero retention there over 100 invocations, so this is specific to the `RunLive` ordering. Since the runner asymmetry predates this change and only bites now that the plugin holds per-invocation state, it may be cleaner to make the plugin tolerate a missing `AfterRun` than to wait on a runner fix.

## Smaller things

[`getStack`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L463-L466) returns nil for an empty invocation ID and eight callback sites dereference it without a check, where `logEvent` at `:219` does check. No framework path produces an empty ID — I searched for one — but `agent.ContextMock` is exported from a non-test file and returns `""`, so a downstream unit test gets a panic instead of a no-op.

[`pushSpan`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L494) accepts a `spanName` and never uses it, so `"invocation_root"`, `"llm_request"` and `t.Name()` are all discarded and no span carries a name. [`spanRecord.ownsSpan`](https://github.com/google/adk-go/blob/2cc8a790eeeda6afc9b37738ce787c254bbfe76d/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L445) is written once and never read.

`TestNewBigQueryAgentAnalyticsPlugin_CreateTable_WithPartitioning` checks `strings.Contains(requestBody, "timestamp")` against the whole create body, which already contains `"name":"timestamp"` from the schema, so changing the partitioning field or dropping it entirely keeps the test green.

More broadly, no test in the package asserts any emitted row's `trace_id`, `span_id`, `parent_span_id` or `latency_ms`, and none invokes a tool callback at all — the new stacking test asserts on `stack.spans` internals. That is why none of the above showed up locally, and the race detector will not help either: with the mutex removed the suite still reports zero races and exits zero, because nothing drives the stack from two goroutines.

Finally, this branch conflicts with `main`, which now carries the fix that stops `AfterModelCallback` logging a row per partial chunk. Per-chunk logging was already the behaviour at this branch's base, so it is not something the PR introduced, but the resolution is not mechanical: the partial branch here is the only place a real stream's first-token time is captured, so taking `main`'s early return wholesale would remove `time_to_first_token_ms` for streaming calls.

wolo-lab:

PR #1384, opened 2026-08-21
feat: Implement GCS offloader to asynchronously offload large payloads

vasilyl:

- Closes: N/A

**Problem:**
Logging very large prompts, tool outputs, or media payloads (e.g., high-res images) directly in-line inside a relational BigQuery row is inefficient, expensive, and hit strict row limits. 

**Solution:**
Introduce a GCS-backed asynchronous offloading strategy inside `bigquery_agent_analytics_plugin.go` via `gcsOffloader` and `hybridContentParser`.
1. When the logged content (text or media) exceeds the configured `inlineTextLimit`, or is binary content, upload it to GCS using the `gcsOffloader`.
2. Record the object storage path as a `GCS_REFERENCE` in BigQuery, and store structured metadata under `object_ref` (including the GCS `uri`, bucket details, and an optional `authorizer` connection ID).
3. Fall back to standard smart truncation if the offloader is not configured or if the upload fails.

### Testing Plan

Added a robust unit test `TestHybridContentParser` with a mock httptest TLS GCS server verifying that:
1. Text payloads below the threshold are successfully kept inline and summarized.
2. Large text and binary media payloads are intercepted, correctly uploaded to the mock GCS bucket, and substituted in the database record with a GCS URI reference containing `object_ref` details.
3. If the GCS offloader is missing or fails, the parser gracefully falls back to local truncation.

### Checklist


karolpiotrowicz (commented):
I went through this at the pinned head [033651e8](https://github.com/google/adk-go/commit/033651e8d400876e77381399a2f7eb5978d14dfc) against its merge base [0a51e15](https://github.com/google/adk-go/commit/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e), running the module's suite in both trees and driving the plugin's callbacks through a fake BigQuery write server so I could decode the Arrow batch and look at the actual column values. The offloader design is sound and the `object_ref` shape lines up with what `schema.go` already declared. What I want to flag is mostly not about the offloading — it's about the move of content and attribute formatting out of the `logEvent` closure and into `writeBatch`. Three of the problems below fire with `GCSBucketName` empty, so they reach every current user of the plugin.

Everything here is reproducible on the branch as it stands. I know it's still a draft, so treat this as early input rather than a gate.

## Error rows lose their error message

[`batch_processor.go:189`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/batch_processor.go#L189) wraps the whole relocated block — including the `_extra_attrs` handling at [`:206`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/batch_processor.go#L206) — in `if rawContent, ok := row["_raw_content"]; ok && rawContent != nil`. Six callbacks pass a literal `nil` content, and two of them carry the attributes that matter: [`MODEL_ERROR`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L247) and [`TOOL_ERROR`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L266). For those rows the `error_message` column is now NULL, and `TOOL_ERROR` also loses `tool_name` from `attributes`. Before this change the attribute loop ran unconditionally in `logEvent`, so both were populated.

The same `if` carries a second condition, `len(extraAttrs) > 0`, which gates the `CustomTags` copy. Any event with no extra attributes — `USER_MESSAGE` and the four lifecycle events — now writes `attributes` NULL even when `Config.CustomTags` is set. One edit covers both.

<details><summary>Decoded from the AppendRowsRequest, base vs this branch</summary>

```
                                             base                              this branch
MODEL_ERROR   error_message                  "PROBE_SENTINEL_ERROR"            <nil>
TOOL_ERROR    error_message                  "PROBE_TOOL_ERROR"                <nil>
TOOL_ERROR    attributes                     {"tool_name":"probe_tool"}        <nil>
INVOCATION_START attributes (CustomTags set) {"deployment":"PROBE_TAG"}        <nil>
USER_MESSAGE     attributes (CustomTags set) (not run)                         <nil>
```

Each row decodes `event_type` correctly on both sides, so the decoder is reading the row under test. Driven through `p.OnModelErrorCallback()` / `p.OnToolErrorCallback()` / `p.BeforeRunCallback()` on a plugin built by `NewBigQueryAgentAnalyticsPluginWithClients`, then `p.AfterRunCallback()` to flush, decoding `req.GetArrowRows().GetRows().GetSerializedRecordBatch()` against the request's own writer schema.

</details>

## The plugin reads objects the runtime is still writing to

[`bigquery_agent_analytics_plugin.go:179`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L179) stores the caller's object into the row by pointer, and [`batch_processor.go:193`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/batch_processor.go#L193) parses it later on a different goroutine. The runtime keeps mutating those objects after the callback returns. `RunBeforeModelCallback` hands the plugin `req` at [`base_flow.go:781`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/internal/llminternal/base_flow.go#L781), the same `req` goes to the model at [`:811`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/internal/llminternal/base_flow.go#L811), and [`gemini.go:168`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/model/gemini/gemini.go#L168) then appends to `req.Contents`. Separately, [`base_flow.go:1266`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/internal/llminternal/base_flow.go#L1266) gives the plugin `fArgs` and [`:1273`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/internal/llminternal/base_flow.go#L1273) gives the same map to `tool.Run` — a tool that writes one key into its own args map would race the reflective walk in `recursiveSmartTruncate`, and a concurrent map read/write is a fatal throw, not a recoverable panic.

The reader is not hypothetical: [`Append`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/batch_processor.go#L109) starts `go b.flush()` from inside the callback once the queue reaches `BatchSize`, and the ticker goroutine flushes every second regardless. Whatever the fix, it needs the plugin to own what it queues — a snapshot taken while still on the caller's goroutine — rather than a pointer into live state. The one-level clone of `extraAttrs` at [`:181`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L181) already does this for attributes.

<details><summary>Race detector output</summary>

Driven by calling `BeforeModelCallback` with an `LLMRequest` whose last content has `Role: "model"` — the condition [`gemini.go:167`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/model/gemini/gemini.go#L167) requires before it appends — then performing that same append while a flush is in progress. `go test -race`:

```
WARNING: DATA RACE
Read at 0x00c00068c710 by goroutine 40:
  agentanalytics.(*hybridContentParser).parse()
      plugin/agentanalytics/bigquery_agent_analytics_plugin.go:473
  agentanalytics.(*BatchProcessor).writeBatch()
      plugin/agentanalytics/batch_processor.go:193
  agentanalytics.(*BatchProcessor).flush()
      plugin/agentanalytics/batch_processor.go:164
Previous write at 0x00c00068c710 by goroutine 33:
  [req.Contents = append(req.Contents, ...)]
```

Two further races are reported in the same run, at [`:474`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L474) (`c.Role`) and [`:340`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L340) (`content.Parts`).

</details>

## The `content` column changes shape for everyone, offloading or not

`parse` returns `map[string]any{"text_summary": …}` for a `*genai.Content` and `{"prompt": …, "system_prompt": …}` for a `*model.LLMRequest`, where the previous code wrote `SmartTruncate(content)` — the full object. `schema.go` types `content` as free-form JSON, so nothing on the BigQuery side rejects the new shape, and existing queries that reach into it just start returning NULL.

The bigger loss is that [`parseContentObject`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L359)'s branch chain has no arm for `FunctionResponse`, so a tool result falls through every branch and appears in neither `content` nor `content_parts`. `FunctionCall` keeps only the name, dropping `Args`. `LLMRequest.Config` and everything on an `LLMResponse` other than `Content` are dropped too, and when `res.Content` is nil the `content` column is NULL outright.

<details><summary>A USER_MESSAGE carrying a text part and a function response, with GCSBucketName empty</summary>

base:

```json
{"parts":[{"text":"hello world", ...},{"functionResponse":{"id":"","name":"get_weather","parts":[],"response":{"city":"Zurich","temp_c":21},"scheduling":"","willContinue":null}, ...}],"role":"user"}
```

this branch:

```json
{"text_summary":"hello world"}
```

</details>

## GCS object names are not unique

The object name is `<date>/<traceID>/<spanID>_p<idx><ext>` at [`:373`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L373) and [`:407`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L407), with no per-payload component, and `uploadContent` sets no `Conditions`, so a repeat write silently replaces the object. Three ways to collide, all reproduced by driving `OnUserMessageCallback` and `BeforeModelCallback` against a mock GCS endpoint:

- **No OpenTelemetry span in context.** [`logEvent`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L190) sets `trace_id`/`span_id` only when the span context is valid, so with tracing unconfigured both are empty and every offloaded payload for the day lands on `2026-09-02//_p0.txt`. Two events, two payloads, one object.
- **Two events sharing one span**, which is the normal case for a model request and its response: both wrote `2026-09-02/4bf92f3577b34da6a3ce929d0e0e4736/00f067aa0ba902b7_p0.txt`.
- **Two messages inside one `LLMRequest`**, because `idx` restarts at 0 for each element of `req.Contents`: both parts of a single row wrote the same name.

In every case both BigQuery rows end up pointing at one object holding the later payload, so a reader who resolves the first row's `uri` gets the second row's content. That is worse than losing it.

## Offloads are serial and land on the run-completion path

[`AfterRunCallback`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L212) calls `processor.flush()` synchronously, `flush` holds `flushMu` for the whole of `writeBatch`, and `writeBatch` uploads one part per iteration in a plain loop. With a mock GCS delaying each request 150 ms, one event with four large parts took **652 ms inside `AfterRunCallback`** — four round trips end to end, not four overlapping ones. `BatchSize` defaults to 100, so a busy flush is bounded only by how many offloadable parts it drained. While the lock is held the ticker's flush no-ops on `TryLock`, the queue keeps filling, and `Append` starts dropping. The comment at [`batch_processor.go:187`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/batch_processor.go#L187) describes this as background work, which it is relative to the callback, but not relative to the run. There is also no per-upload deadline — `flush` passes `b.ctx`, which has none — so a stalled endpoint parks the writer indefinitely. I did not reproduce that one.

## A failed upload is invisible

`hybridContentParser` holds no logger and never calls one, so nothing is written to `Config.Logger` when an upload fails. On the text path the [failure branch](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L429) falls back to inline truncation and produces a part byte-identical to one that was never eligible for offload — same `storage_mode`, same nil `uri`, same plain text. On the binary path the part gets `[UPLOAD FAILED]` in `text`, which is at least visible, but `storage_mode` stays `INLINE` and the bytes are stored nowhere. `newGCSOffloader` also doesn't check the bucket exists, so a typo in `GCSBucketName` produces this for every payload with nothing anywhere saying so.

## The exported constructor break

[`NewBatchProcessor`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/batch_processor.go#L52) gained a fifth parameter typed `*gcsOffloader`, which is unexported. That breaks every existing external call site, and the replacement is not usable from outside the package either — an out-of-module caller can only ever pass the untyped `nil`. This is what the red `apidiff` check is reporting. The check has a `breaking-change` label escape hatch, so it is a decision rather than a hard stop, but the unexported type in an exported signature is worth avoiding independently of the label.

## Smaller things

- **`mime_type` regresses for inline media when offloading is off.** The [no-offloader branch](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L396) sets only `text = "[BINARY DATA]"`, leaving the `"text/plain"` default, where `FormatContentParts` set the real type. An `image/png` part now records `text/plain`.
- **Two client leaks.** [`:128`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L128) returns on an offloader-construction failure without closing `writeClient`, unlike the adjacent path at [`:138`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L138). And [`:272`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L272) returns when `writeClient.Close()` errors, so `offloader.close()` below it never runs.
- **`content` is empty when everything is offloaded.** The [offload-success branch](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L409) never appends to `summaryText`, so an event whose every text part went to GCS writes `{"text_summary":""}`.
- **`object_ref.authorizer` can be written empty.** Nothing checks `ConnectionID` is set when `GCSBucketName` is.
- **`part_index` repeats within a row** for a multi-message `LLMRequest`, since the index restarts per `Contents` element and the lists are flattened.
- **`FormatContentParts` is now dead** outside its own test, and that test still asserts the inline-media mapping the live path no longer produces.
- **The truncation marker doubles up.** [`:414`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L414) calls `truncateString(part.Text, 200)`, which already appends `...[truncated]`, then `"... [OFFLOADED]"` is appended after it.
- **`inlineTextLimit` is hardcoded** to `32*1024` with no `Config` field, though the description calls it configured. Relatedly, [`:402`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L402) tests `p.maxLen != -1` while `truncateString` tests `maxLength < 0`, so a `MaxContentLen` of `-2` means "no truncation" to one and "offload everything" to the other.
- **`newGCSOffloader` takes a `projectID` it never uses**, and forwards `Config.ClientOptions` — set up for the BigQuery clients — straight to the GCS client.
- **A nil `*model.LLMResponse` reaching `AfterModelCallback` now crashes the process.** [`parse:501`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L501) dereferences `res` with no nil check, where the old path went through `recursiveSmartTruncate` and returned `"null"`. [`:474`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/bigquery_agent_analytics_plugin.go#L474) has the same shape, reading `c.Role` before `parseContentObject`'s own nil guard. The current flow can't deliver either input, since `base_flow.go:830` dereferences first — the reason it's worth mentioning is placement: the panic now lands on the goroutine started at [`batch_processor.go:80`](https://github.com/google/adk-go/blob/033651e8d400876e77381399a2f7eb5978d14dfc/plugin/agentanalytics/batch_processor.go#L80), which has no `recover`, so it takes the host process rather than unwinding through the agent's stack.

## Why the suite stays green through all of this

Worth calling out, because it's fixable in one place. No test in the module decodes a written column — the plugin tests assert only that an append request arrived — so the shape and NULL regressions above are invisible to `go test` on either side. `TestHybridContentParser` never constructs an `InlineData` part with an offloader present, so the media path in the description isn't exercised, and the mock GCS handler always returns 200, so neither failure branch runs. `-race` passes because nothing retains a callback-supplied object and mutates it afterwards. A single test that decodes the Arrow batch and asserts on two or three columns would have caught most of this.

wolo-lab:

PR #1385, opened 2026-08-21
feat: Add ConnectionID for BigQuery Object Tables secure external access configuration

vasilyl:

- Closes: N/A

**Problem:**
When offloading large text or media payloads to Google Cloud Storage (GCS), analytical queries on the logged telemetry database need a secure, unified way to link the external GCS references back to BigQuery. BigQuery Object Tables allow querying external GCS objects securely, but require an explicit BigQuery connection ID.

**Solution:**
Introduce a new configuration parameter `ConnectionID string` to the plugin `Config` struct (in `config.go`), and initialize it to an empty string by default in `DefaultConfig()`. This provides a clean, standard location to pass the BigQuery external connection resource path, preparing the analytics plugin to generate secure metadata references (`authorizer` field in external object refs).

### Testing Plan

Verified that the `ConnectionID` field is successfully defined on the `Config` struct, and that it defaults to an empty string when calling `DefaultConfig()` in testing.

### Checklist

### DIFF
--- a/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
+++ b/plugin/agentanalytics/bigquery_agent_analytics_plugin.go
@@ -102,6 +102,10 @@ func NewBigQueryAgentAnalyticsPluginWithClients(
 		}
 		err = tableRef.Create(ctx, &bq.TableMetadata{
 			Schema: EventsSchema(),
+			TimePartitioning: &bq.TimePartitioning{
+				Field: "timestamp",
+				Type:  bq.DayPartitioningType,
+			},
 			Clustering: &bq.Clustering{
 				Fields: config.ClusteringFields,
 			},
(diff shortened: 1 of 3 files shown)


jjsasha63:
Closing this as redundant: #1384 already adds the same ConnectionID field to
Config, and unlike this PR it actually consumes it — batch_processor.go passes
config.ConnectionID into newHybridContentParser, which uses it as the
"authorizer" on the external object refs. Landing this one first would leave
#1384 with a duplicate field declaration to unpick. On its own the field has no
reader anywhere in the tree, and a field on an exported struct is a
compatibility commitment we would then have to keep.

Two smaller things, which apply to #1384 too: `ConnectionID: ""` in
DefaultConfig is a no-op, since that is already the zero value for a string. And
the Testing Plan here describes a ConnectionID test that is not in the diff —
the test that shows up is the partitioning test from #1377, which this branch
sits on top of.

Thanks for splitting the work up, and sorry for the slow look at this one.

PR #1387, opened 2026-08-22
fix(agentengine): base64-encode []byte fields instead of failing thei…

### LINKED ISSUE #1386: []byte fields (e.g. genai.Part.ThoughtSignature) fail helper.ConvertSnake with "unsupported type: uint8", breaking Agent Engine event JSON
##  Problem Description

`helper.ConvertSnake` (in `server/agentengine/internal/helper/encode.go`) cannot convert `[]byte` struct fields. Its reflection-based type switch has no case for `reflect.Uint8`, and for slices it converts each element individually — so every byte in a `[]byte` hits the `default` branch and returns `unsupported type: uint8`, failing conversion of the entire containing struct.

The concrete trigger is `genai.Part.ThoughtSignature`, which thinking models (`gemini-2.5-flash` and later) routinely populate. The error is swallowed at the top level and the code falls back to returning the **original, unconverted** Go value:

```go
func ConvertSnake(o any) any {
	res, err := convertSnake("", "", o)
	if err != nil {
		log.Printf("Failed to convert: %+v of type %T: %v", o, o, err)
		// better to return an original version than nothing
		return o
	}
	return res
}
```

The affected event is then serialized with Go's native struct JSON tags (camelCase, `LLMResponse`-shaped) and mixed into a stream of otherwise correctly-converted snake_case events. This breaks the console Playground, breaks Gemini Enterprise's native Reasoning Engine integration, and can cascade into a fully dead conversation via GE's retry behavior (see Additional Context).

Go's standard `encoding/json.Marshal` already base64-encodes `[]byte` automatically; `convertSnake` is a hand-rolled converter (built to additionally rename fields to snake_case) whose type switch never accounted for byte slices.

**Steps to Reproduce:**

1. Deploy any ADK Go agent to Agent Engine via `sourceCodeSpec` or `containerSpec`, using a model that returns a non-empty thought signature (`gemini-2.5-flash` on a function-calling turn is sufficient; `gemini-3.6-flash` reproduces on *any* turn, including plain text with no tools).
2. Send it a request via `:streamQuery` with `"class_method": "async_stream_query"`, or simply open the agent in the console Playground.
3. Observe the response payload for the turn: it is camelCase / `LLMResponse`-shaped rather than snake_case, unlike every other event in the same stream.
4. Check Cloud Logging for the `log.Printf` emitted by `encode.go` — see the Logs section below.

Purely locally, without deploying: any struct containing a non-empty `[]byte` field passed through `helper.ConvertSnake` reproduces the same failure. Because `helper` is an `internal` package, the isolated repro has to live inside the module — see Minimal Reproduction Code.

**Expected Behavior:**

`[]byte` fields are base64-encoded into a JSON string, matching what `encoding/json.Marshal` already does and what the JSON convention for byte strings is. `thoughtSignature` should appear as `"thought_signature": "<base64>"` and the containing event should convert to snake_case like every other event.

**Observed Behavior:**

Conversion of the whole containing struct fails with `unsupported type: uint8`. The error is logged and swallowed, and the unconverted Go value is returned instead, producing a wrong-cased payload for that event. On the `async_stream_query` path the blast radius is larger: it wraps the **entire event batch** in one `StreamingAgentRunWithEventsResponse`, so a single bad `ThoughtSignature` anywhere in the batch breaks conversion of the whole response object rather than one event.

**Environment Details:**
- ADK Library Version: `google.golang.org/adk/v2` — confirmed on v2.0.0, v2.1.0 and v2.2.0 (the full published range at time of writing; not version-specific)
- OS: Agent Engine runtime
- Go Version: 1.26.5

**Model Information:**
- Which model is being used: `gemini-2.5-flash` and `gemini-3.6-flash` (any model that emits a non-empty `ThoughtSignature`)

---

## 🟡 Optional Information

**Regression:**

No. Reproduces identically on every published version of `google.golang.org/adk/v2` (v2.0.0 through v2.2.0), so there is no known-good earlier version to fall back to.

**Logs:**

```text
Failed to convert: {...} of type session.Event: failed to convert embedded
struct with name:LLMResponse o: {...} model.LLMResponse err: failed to
convert regular struct field with path: .LLMResponse.content err: ...
err: failed to convert regular struct field with path:
.LLMResponse.content*.parts.[]*.thoughtSignature err: failed to convert
slice element with path: .LLMResponse.content*.parts.[]*.thoughtSignature.[]
err: unsupported type: uint8
```

End-user-visible failure surfaced by Gemini Enterprise after its retries are exhausted:

```text
Agent returned an error (400): Agent failed with error: Reasoning Engine
stream closed cleanly without producing any events
(reasoning_engine=projects/<p>/locations/<l>/reasoningEngines/<id>, attempt=3/3).
```

**Additional Context:**

*Root cause, in code.* The type switch only handles signed integers:

```go
case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
	return v.Int(), nil

default:
	return nil, fmt.Errorf("unsupported type: %v", v.Kind())
```

and the slice branch descends per element, so each `uint8` lands in `default`:

```go
case reflect.Slice:
	res := []any{}
	for i := 0; i < v.Len(); i++ {
		elem, err := convertSnake(path+".[]", indent+"    ", v.Index(i).Interface())
		...
```

*Suggested fix.* Detect a byte slice before the generic per-element loop:

```go
case reflect.Slice:
	if v.Type().Elem().Kind() == reflect.Uint8 {
		return base64.StdEncoding.EncodeToString(v.Bytes()), nil
	}
	// existing per-element loop for non-byte slices
	...
```

This matches `encoding/json.Marshal`'s existing behavior and needs no schema or consumer changes.

*Trigger conditions (confirmed empirically).*
- `gemini-2.5-flash`: emits a non-empty `ThoughtSignature` on function-calling turns. Does **not** emit one when `GenerateContentConfig.ThinkingConfig.ThinkingBudget = 0`.
- `gemini-3.6-flash`: emits a non-empty `ThoughtSignature` on **every** turn observed — including plain text with no tool calls, no tools declared and no reasoning required — and this does **not** go away with `ThinkingBudget = 0`.
- Verified directly against `aiplatform.googleapis.com`'s `generateContent` REST API, not just through ADK, confirming the signature is a property of the model responses themselves rather than something ADK adds.

*Affected vs. unaffected code paths.* Only two files call `helper.ConvertSnake`, both under `server/agentengine/controllers/method/`:
- `stream_query.go` — backs the `query` and `stream_query` classMethods (`/api/reasoning_engine`, `/api/stream_reasoning_engine`; the console Playground and the Python SDK path).
- `streaming_agent_run_with_events.go` — backs `async_stream_query`, whose own auto-generated description states *"This method is primarily meant for invocation from Gemini Enterprise."* This is the path with whole-batch failure described above.

Confirmed **not** affected (no `ConvertSnake` usage): `server/adkrest`, which backs the local ADK Web UI / `agents-cli playground` dev server and uses plain `encoding/json`; and `server/adka2a`, the A2A protocol server. So a Go agent's local dev Playground works fine while the exact same agent deployed to Agent Engine breaks on any turn with a non-empty `ThoughtSignature` — which is a large part of why this is hard to catch before deployment.

*Observed real-world cascade (Gemini Enterprise).* This is not only a cosmetic display bug. From live Cloud Logging on a `gemini-3.6-flash` deployment, a single "hi"-level user turn produced 3 GE-initiated attempts within 5 seconds, all reusing the same session ID (IDs and resource paths redacted; preserved in an internal writeup available on request):

1. **Attempt 1** — session created, model responds, but `StreamingAgentRunWithEventsResponse` fails to convert (`unsupported type: uint8` on `.LLMResponse.content*.parts.[]*.thoughtSignature.[]`). GE appears to treat this as an unusable response and retries.
2. **Attempt 2** (~2s later, same session ID) — fails immediately with `rpc error: code = InvalidArgument desc = Session with user-provided ID '.../sessions/<id>' already exists.` Session creation is not idempotent against a retried request reusing the same session ID.
3. **Attempt 3** — model responds again, but appending the event to the session fails with `rpc error: code = Internal desc = Internal error encountered.`

The user-visible result is a completely dead conversation several retries and error types removed from the root cause, which made this very hard to diagnose without instrumenting the model response bytes directly.

*Workarounds considered, and why each is incomplete.*
- **`ThinkingConfig.ThinkingBudget = 0`** — fully avoids the bug on `gemini-2.5-flash` (confirmed: no signature emitted at all), but does **not** work on `gemini-3.6-flash` (confirmed: signature persists regardless of budget). Also a real quality trade-off, since it disables extended reasoning.
- **Stripping `ThoughtSignature` in an `AfterModelCallback`** — safe for plain-text turns (confirmed: Gemini does not require the signature to be echoed back for non-tool turns), but **unsafe** for function-calling turns. The Gemini API hard-requires the signature to be echoed back on `functionCall` parts in conversation history; omitting it returns `400 INVALID_ARGUMENT` with *"Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly"*. Stripping unconditionally would permanently break any session the first time a tool call happens, since the corrupted history is replayed on every subsequent turn.
- **Routing through A2A instead of the native Reasoning Engine integration** — viable and confirmed unaffected, but requires additional infrastructure (Agent Gateway for external reachability) beyond a code change.
- **Vendoring a patched copy of `google.golang.org/adk/v2`** — works, but means committing/hosting a ~15MB, 300+ file copy of the SDK and reapplying the patch on every upstream upgrade indefinitely.

`helper` is an `internal` package, so the isolated repro has to run from inside the module — e.g. as `server/agentengine/internal/helper/bytes_repro_test.go`:

```go
package helper

import "testing"

type payload struct {
	ThoughtSignature []byte
}

func TestConvertSnakeByteSlice(t *testing.T) {
	got := ConvertSnake(payload{ThoughtSignature: []byte{0x01, 0x02, 0x03}})

	// Want: map[string]any{"thought_signature": "AQID"}
	// Got:  helper.payload{ThoughtSignature:[]uint8{0x1, 0x2, 0x3}}
	//       plus a logged "unsupported type: uint8" and the unconverted fallback.
	t.Logf("%#v", got)
}
```

Any struct with a non-empty `[]byte` field reproduces it; nothing model- or ADK-agent-specific is required to hit the conversion failure itself.

- Always (100%) — deterministic whenever the response contains a non-empty `[]byte` field.

rizukijr:

- Closes: #1386

### Testing Plan

**Unit Tests:**

Added two tests to `encode_test.go`, following the existing file's conventions:

- `TestByteSlice` - a `session.Event` with a non-empty
  `Content.Parts[0].ThoughtSignature` now converts successfully, and the
  resulting `thought_signature` key holds the expected base64 string
  (`{0xDE, 0xAD, 0xBE, 0xEF}` -> `"3q2+7w=="`, matching what
  `encoding/json` produces for the same bytes).
- `TestByteSliceOmitEmpty` - confirms `omitempty` is unaffected: an empty
  `[]byte` is still correctly omitted, not encoded as an empty string.

Ran the full existing package suite alongside the new tests - zero regressions:

```
$ go test ./server/agentengine/...
ok      google.golang.org/adk/v2/server/agentengine/controllers/method  0.972s
ok      google.golang.org/adk/v2/server/agentengine/internal/helper     0.491s
$ go vet ./server/agentengine/...
(clean)
```

Ran the real `agentengine` launcher (`web agentengine` - the same launcher
composition `agentengine.NewLauncher` uses in production) locally against a
real deployed, Vertex AI Sessions-backed Reasoning Engine resource, using
`gemini-3.6-flash` (emits a non-empty `ThoughtSignature` on essentially
every turn, not just tool calls - the worst case observed). Sent the exact
`async_stream_query` request shape Gemini Enterprise uses:

```
curl -s -X POST -H "Content-Type: application/json" \
  "http://localhost:8080/api/stream_reasoning_engine" \
  -d '{"class_method": "async_stream_query", "input": {"user_id": "verify", "message": {"role": "user", "parts": [{"text": "hi"}]}}}'
```

- Before the fix: every signature-bearing event logged `Failed to convert:
  ... err: unsupported type: uint8` and shipped to the client in raw
  Go/camelCase field names (`thoughtSignature`, `finishReason`,
  `invocationId`) mixed into an otherwise snake_case stream.
- After the fix: zero `Failed to convert` log lines. The event now
  serializes correctly, e.g. `"thought_signature": "AY89a1..."`, with
  consistent snake_case throughout (`finish_reason`, `invocation_id`, etc).

Also confirmed this is not model- or version-specific: the same missing
`reflect.Uint8` case is present identically in all three published
`adk/v2` releases (v2.0.0, v2.1.0, v2.2.0), and the bug reproduces on
`gemini-2.5-flash` too on any function-calling turn (just not on every
turn, unlike `gemini-3.6-flash`). Full details, including the live Gemini
Enterprise retry-into-dead-session cascade this caused in production, are
in #1386.

### Checklist

### Additional context

`convertSnake`'s type switch only ever handled signed integers
(`Int`/`Int8`/`Int16`/`Int32`/`Int64`) - there was no case for `Uint8` or
any other `Uint*` kind. A `[]byte` field is a `reflect.Slice`, so it fell
into the generic per-element loop, and each individual byte (a bare
`uint8`) hit the `default: unsupported type` branch, failing conversion of
the *entire containing struct*. `encoding/json.Marshal` already solves
this correctly (base64-encodes `[]byte` automatically) - this fix just
applies the same, standard approach instead of reinventing a worse version
of it for one specific hand-rolled converter.

### DIFF
--- a/server/agentengine/internal/helper/encode.go
+++ b/server/agentengine/internal/helper/encode.go
@@ -15,6 +15,8 @@
 package helper
 
 import (
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"log"
 	"reflect"
@@ -120,6 +122,15 @@ func convertSnake(path, indent string, o any) (any, error) {
 		}
 		return m, nil
 	case reflect.Slice:
+		if rm, ok := o.(json.RawMessage); ok {
+			if len(rm) == 0 {
+				return nil, nil
+			}
+			return rm, nil
+		}
+		if v.Type().Elem().Kind() == reflect.Uint8 {
+			return base64.StdEncoding.EncodeToString(v.Bytes()), nil
+		}
 		res := []any{}
 		for i := 0; i < v.Len(); i++ {
 			elem, err := convertSnake(path+".[]", indent+"    ", v.Index(i).Interface())
--- a/server/agentengine/internal/helper/encode.go
+++ b/server/agentengine/internal/helper/encode.go
@@ -15,6 +15,7 @@
 package helper
 
 import (
+	"encoding/base64"
 	"fmt"
 	"log"
 	"reflect"
@@ -120,6 +121,9 @@ func convertSnake(path, indent string, o any) (any, error) {
 		}
 		return m, nil
 	case reflect.Slice:
+		if v.Type().Elem().Kind() == reflect.Uint8 {
+			return base64.StdEncoding.EncodeToString(v.Bytes()), nil
+		}
 		res := []any{}
 		for i := 0; i < v.Len(); i++ {
 			elem, err := convertSnake(path+".[]", indent+"    ", v.Index(i).Interface())
--- a/server/agentengine/internal/helper/encode_test.go
+++ b/server/agentengine/internal/helper/encode_test.go
@@ -163,6 +163,62 @@ func TestEventLogProbs(t *testing.T) {
 	}
 }
 
+func TestByteSlice(t *testing.T) {
+	event := session.Event{
+		ID: "1",
+		LLMResponse: model.LLMResponse{
+			Content: &genai.Content{
+				Parts: []*genai.Part{
+					{
+						Text:             "hi",
+						ThoughtSignature: []byte{0xDE, 0xAD, 0xBE, 0xEF},
+					},
+				},
+			},
+		},
+	}
+	o, err := convertSnake("", "", event)
+	if err != nil {
+		t.Fatalf("convertSnake() failed: %v", err)
+	}
+	m, ok := o.(map[string]any)
+	if !ok {
+		t.Fatalf("o is not a map: %T", o)
+	}
+	content, ok := m["content"].(map[string]any)
+	if !ok {
+		t.Fatalf("content is not a map: %T", m["content"])
+	}
+	parts, ok := content["parts"].([]any)
+	if !ok || len(parts) != 1 {
+		t.Fatalf("parts is not a 1-element slice: %v", content["parts"])
+	}
+	part, ok := parts[0].(map[string]any)
+	if !ok {
+		t.Fatalf("part is not a map: %T", parts[0])
+	}
+	want := "3q2+7w=="
+	if got, _ := part["thought_signature"].(string); got != want {
+		t.Errorf("thought_signature = %q, want %q", got, want)
+	}
+}
+
+func TestByteSliceOmitEmpty(t *testing.T) {
+	type A struct {
+		Sig []byte `json:"sig,omitempty"`
+	}
+	got, err := convertSnake("", "", A{})
+	if err != nil {
+		t.Fatalf("convertSnake() failed: %v", err)
+	}
+	m, ok := got.(map[string]any)
+	if !ok {
+		t.Fatalf("got is not a map: %T", got)
+	}
+	if _, present := m["sig"]; present {
+		t.Errorf("sig should be omitted for an empty []byte, got: %v", m["sig"])
+	}
+}
 func TestEmbedded(t *testing.T) {
 	type A struct {
 		anInteger int
--- a/server/agentengine/internal/helper/encode_test.go
+++ b/server/agentengine/internal/helper/encode_test.go
@@ -15,6 +15,7 @@
 package helper
 
 import (
+	"encoding/json"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -160,6 +161,99 @@ func TestEventLogProbs(t *testing.T) {
 		}
 	} else {
 		t.Errorf("o is not a map")
+	}
+}
+
+func TestByteSlice(t *testing.T) {
+	event := session.Event{
+		ID: "1",
+		LLMResponse: model.LLMResponse{
+			Content: &genai.Content{
+				Parts: []*genai.Part{
+					{
+						Text:             "hi",
+						ThoughtSignature: []byte{0xDE, 0xAD, 0xBE, 0xEF},
+					},
+				},
+			},
+		},
+	}
+	o, err := convertSnake("", "", event)
+	if err != nil {
+		t.Fatalf("convertSnake() failed: %v", err)
+	}
+	m, ok := o.(map[string]any)
+	if !ok {
+		t.Fatalf("o is not a map: %T", o)
+	}
+	content, ok := m["content"].(map[string]any)
+	if !ok {
+		t.Fatalf("content is not a map: %T", m["content"])
+	}
+	parts, ok := content["parts"].([]any)
+	if !ok || len(parts) != 1 {
+		t.Fatalf("parts is not a 1-element slice: %v", content["parts"])
+	}
+	part, ok := parts[0].(map[string]any)
+	if !ok {
+		t.Fatalf("part is not a map: %T", parts[0])
+	}
+	want := "3q2+7w=="
+	if got, _ := part["thought_signature"].(string); got != want {
+		t.Errorf("thought_signature = %q, want %q", got, want)
+	}
+}
+
+func TestByteSliceOmitEmpty(t *testing.T) {
+	type A struct {
+		Sig []byte `json:"sig,omitempty"`
+	}
+	got, err := convertSnake("", "", A{})
+	if err != nil {
+		t.Fatalf("convertSnake() failed: %v", err)
+	}
+	m, ok := got.(map[string]any)
+	if !ok {
+		t.Fatalf("got is not a map: %T", got)
+	}
+	if _, present := m["sig"]; present {
+		t.Errorf("sig should be omitted for an empty []byte, got: %v", m["sig"])
+	}
+}
+
+func TestRawMessage(t *testing.T) {
+	type A struct {
+		Default json.RawMessage `json:"default,omitempty"`
+	}
+	a := A{Default: json.RawMessage(`{"approved":false}`)}
+	got, err := convertSnake("", "", a)
+	if err != nil {
+		t.Fatalf("convertSnake() failed: %v", err)
+	}
+	b, err := json.Marshal(got)
+	if err != nil {
+		t.Fatalf("json.Marshal() failed: %v", err)
+	}
+	want := `{"default":{"approved":false}}`
+	if string(b) != want {
+		t.Errorf("marshaled output = %s, want %s", b, want)
+	}
+}
+
+func TestRawMessageOmitEmpty(t *testing.T) {
+	type A struct {
+		Default json.RawMessage `json:"default,omitempty"`
+	}
+	got, err := convertSnake("", "", A{})
+	if err != nil {
+		t.Fatalf("convertSnake() failed: %v", err)
+	}
+	m, ok := got.(map[string]any)
+	if !ok {
+		t.Fatalf("got is not a map: %T", got)
+	}
+	if _, present := m["default"]; present {
+		t.Errorf("default should be omitted for an empty json.RawMessage, got: %v", m["default"])
 	}
 }
 


karolpiotrowicz (dismissed):
The core of this is right, and the reasoning in the PR description is right with it. Go's `encoding/json` already base64-encodes `[]byte` with the standard alphabet, and because the pre-change error path fell back to `encoding/json`, the byte field was *already* going out as base64. I measured both trees and the emitted text for `{0xDE, 0xAD, 0xBE, 0xEF}` is `"3q2+7w=="` on each. So this does not change the wire representation of the bytes at all — it changes the envelope around them from the accidental camelCase fallback to the snake_case shape the converter is supposed to produce. That is the fix, and `TestByteSlice` is a real red-to-green test: it fails on the merge-base with `unsupported type: uint8` and passes with the change applied.

Two things need attention before this can land.

## The added test breaks the lint job

There is no blank line between the end of `TestByteSliceOmitEmpty` and `func TestEmbedded` at [encode_test.go:219-222](https://github.com/google/adk-go/blob/918be1955a6c423db28588ea421ac99f7c70dce4/server/agentengine/internal/helper/encode_test.go#L219-L222). The base file had that separator and the new hunk consumed it.

This is worth flagging rather than leaving for CI because CI has not run on this branch. The workflow is waiting on the first-time-contributor approval gate, so in 22 days nothing has built, tested or linted the commit. Running the version the repo pins at [.github/workflows/go.yml:88](https://github.com/google/adk-go/blob/918be1955a6c423db28588ea421ac99f7c70dce4/.github/workflows/go.yml#L88):

```
golangci-lint v2.3.1 run ./server/agentengine/internal/helper/...
  merge-base                                    exit 0, "0 issues."
  this branch                                   exit 1
     encode_test.go:221:1: File is not properly formatted (gofumpt)
  this branch + one blank line before TestEmbedded   exit 0, "0 issues."
```

Note that `gofmt -l` is clean on both, so a local `gofmt` check will not surface it — `.golangci.yml` enables `gofumpt` with `extra-rules: true`, and that is the rule firing.

## A `json.RawMessage` now goes out as base64

`json.RawMessage` is `[]byte`, so the new guard at [encode.go:123-126](https://github.com/google/adk-go/blob/918be1955a6c423db28588ea421ac99f7c70dce4/server/agentengine/internal/helper/encode.go#L123-L126) fires on it and base64-encodes JSON that was already JSON. `encoding/json` tests for `json.Marshaler` and `encoding.TextMarshaler` at the same decision point, which is what keeps it from doing this.

It is reachable through a declared field rather than a hand-built struct. [session/session.go:204](https://github.com/google/adk-go/blob/918be1955a6c423db28588ea421ac99f7c70dce4/session/session.go#L204) declares `RequestInput.ResponseSchema *jsonschema.Schema`, whose `Default` is a `json.RawMessage`, and the exported [workflow.NewRequestInputEvent](https://github.com/google/adk-go/blob/918be1955a6c423db28588ea421ac99f7c70dce4/workflow/request_input.go#L72-L95) sets both `RequestedInput` and `Content`, so the event clears the `Content == nil` skip at [stream_query.go:112-116](https://github.com/google/adk-go/blob/918be1955a6c423db28588ea421ac99f7c70dce4/server/agentengine/controllers/method/stream_query.go#L112-L116) and reaches `EmitJSON`. Driven through the public path from a real `session.Event`:

```
merge-base : "responseSchema":{"type":"object","description":"d","default":{"approved":false}}
this branch: "response_schema":{"default":"eyJhcHByb3ZlZCI6ZmFsc2V9","description":"d"}
```

I want to be fair about what that comparison shows. The merge-base line is the whole-event fallback, so on that side every key in the event was wrong, not just this one. A client reading the intended snake_case format could not use that event at all. What the change does is fix every key except this one, and turn this one from correct JSON into an opaque string. Net it is an improvement with a hole in it, not a regression.

I am not going to hand you a fix, because I tried the obvious one and it does not hold up. Adding `&& !v.Type().Implements(jsonMarshalerType) && !v.Type().Implements(textMarshalerType)` to the guard does make `response_schema` come out correct — but only because the event then errors out and reverts to the whole-event `encoding/json` fallback, camelCase envelope and all. It trades the hole for the bug you are fixing. Two constraints on any alternative, both measured:

- The converter has never honored `json.Marshaler`, so this is a pre-existing gap this change newly walks into rather than one it creates. `session.EventActions` has a hand-written `MarshalJSON` that emits `{}`, and the converter emits `{"artifact_delta":{},"state_delta":{}}` instead — identically on both trees.
- Fixing the byte branch alone will not make `response_schema` correct anyway. `jsonschema.Schema.Type` is tagged `json:"-"` and the `type` key is synthesized by that type's own `MarshalJSON`, so the converter drops `"type":"object"` on both trees. A schema carrying no `Default` at all already comes out identically before and after this change.

Given that, treating this as a follow-up on the converter rather than something to solve inside a four-line fix seems reasonable to me, as long as it is written down somewhere.

## Smaller, and genuinely minor

One more case changes behavior: a struct with an *anonymous* byte-slice field — `struct{ json.RawMessage }`, or an embedded named `[]byte` type — placed in `CustomMetadata`, `StateDelta`, `Output` or `Payload` now panics at the unchecked `embed.(map[string]any)` at [encode.go:82-91](https://github.com/google/adk-go/blob/918be1955a6c423db28588ea421ac99f7c70dce4/server/agentengine/internal/helper/encode.go#L82-L91), where before it produced valid JSON.

I checked how much of a hazard that really is before raising it, and the answer is: not much, because the assertion is already broken for most things. Embedded `time.Time`, an embedded named string, an embedded named non-byte slice, a nil embedded pointer, and even a nil or empty embedded byte slice all panic there on the merge-base too. Only the non-empty byte-slice case flips. No type in the repository has an anonymous byte-slice field, and the in-repo writers of `CustomMetadata` all store strings. The line is untouched by this PR, and it is the line that wants the `, ok` form — that is a separate change from this one.

## What I checked that holds up

Events that carry no bytes produce byte-identical output before and after, so the change is confined to the Kind it targets. Nothing persisted depends on the old behavior: there is no `testdata`, golden file or cassette anywhere under `server/agentengine`, and the only in-repo decode of this stream reads the `content` key alone. A 120-second fuzz of the round-trip through `EmitJSON` came back with 405,468 executions and no counterexample, though that only exercises the bytes in a concrete `[]byte` field and cannot reach either case above. `go vet`, `staticcheck` and `govulncheck` are clean and unchanged against the merge-base, and the two failing packages in the suite fail identically on the merge-base, so nothing in the suite is attributable to this change.

Worth saying explicitly: neither the race detector nor the API-compatibility check tells you anything here. There is no concurrency anywhere in this package or either of its importers, and `apidiff` prints `Ignoring internal package .../agentengine/internal/helper` and exits 0 without looking at it. I would not read either as a green light on this change.

rizukijr:
Thanks for the review @karolpiotrowicz 

please check the requested changes:

- **Fixed the `gofumpt` violation** — added the missing blank line before `TestEmbedded`, so the lint job now passes.
- **Fixed the `json.RawMessage` regression** — the `[]byte` guard now passes already-JSON bytes through verbatim instead of base64-encoding them, so `response_schema.default` stays valid JSON while the snake_case envelope is preserved. Added `TestRawMessage` and `TestRawMessageOmitEmpty` to cover it.
- **Acknowledged the remaining converter limitations as follow-ups** — marshalers (`json.Marshaler`/`encoding.TextMarshaler`) are not honored generally, and struct fields tagged `json:"-"` (such as `jsonschema.Schema`'s `type`) are dropped. Both are pre-existing on the merge-base, not regressions of this change.

karolpiotrowicz (commented):
Both are fixed, and your account of what you did is accurate on all three points. I re-ran the version the repo pins at [.github/workflows/go.yml:91](https://github.com/google/adk-go/blob/12bf9fcf3b7a97c41fd15fbb5d29346f8c066291/.github/workflows/go.yml#L91) and it exits 0 on this branch, including in the workspace mode CI actually uses, for both modules. The same invocation still exits 1 on the previous commit ([918be19](https://github.com/google/adk-go/commit/918be1955a6c423db28588ea421ac99f7c70dce4)), so that is a real pass rather than a check that stopped looking.

The `json.RawMessage` fix is better than the one I tried and discarded. Keying on the concrete type at [encode.go:125-130](https://github.com/google/adk-go/blob/12bf9fcf3b7a97c41fd15fbb5d29346f8c066291/server/agentengine/internal/helper/encode.go#L125-L130) avoids the trap that sank a `json.Marshaler` guard, which fixed `default` only by sending the whole event back to the camelCase fallback. Yours returns no error, so the fallback never engages. I drove nine sinks through `EmitJSON` from a real `session.Event` — the schema `Default`, nested `properties` and `$defs`, `Event.Output`, `RequestInput.Payload`, `StateDelta`, `CustomMetadata`, and the pointer, slice and map forms — and all nine come out inline with the snake_case envelope intact, including an event carrying a `ThoughtSignature` and a schema `Default` at once.

One thing worth your time. None of the four new tests cross `ConvertSnake` or `EmitJSON` — they all call the unexported `convertSnake` directly — and [TestRawMessage](https://github.com/google/adk-go/blob/12bf9fcf3b7a97c41fd15fbb5d29346f8c066291/server/agentengine/internal/helper/encode_test.go#L224-L241) declares a local `type A` rather than using `jsonschema.Schema`. So nothing in the suite pins the property this PR exists to establish, which is that the event stops falling back to the camelCase envelope, and a change to the real field's type upstream would not be caught. One test asserting a snake_case key on the output of `EmitJSON` for a `session.Event` carrying a schema would close both gaps.

Two smaller things, neither blocking. A `json.RawMessage` field with no `omitempty` now has its key dropped entirely, where the merge-base emitted `[]` and `encoding/json` emits `null` — the package goes the other way for ordinary slices, and `TestOmitEmpty` asserts that a mandatory nil slice keeps its key as `[]`. No field on the event path is affected, since every reachable one carries `omitempty`. And the embedded byte-slice panic from my last round is unchanged: `struct{ json.RawMessage }` and an embedded named `[]byte` still reach the unchecked assertion at [encode.go:89](https://github.com/google/adk-go/blob/12bf9fcf3b7a97c41fd15fbb5d29346f8c066291/server/agentengine/internal/helper/encode.go#L89) when non-empty. That line is yours to leave alone as far as I am concerned — no type in the repository has an anonymous byte-slice field, and the same assertion already crashes on the merge-base for embedded `time.Time`.

Worth knowing, since you wrote that the lint job now passes: it has not run. The Go workflow fired on this commit and returned `action_required`, so it is queued behind the first-time-contributor approval gate and has built, tested and linted nothing — at any revision of this PR, going back to the original submission. Everything above is from running the pinned tools locally. A maintainer approving the workflow would replace that with the real thing.

rizukijr:
Added a test that drives the public EmitJSON with a real session.Event carrying a jsonschema.Schema (Event.RequestedInput.ResponseSchema), asserting the snake_case envelope including nested Default and Properties. Mutation-tested: fails on pre-fix encode.go (base64-mangles Default, wrong camelCase key), passes on current fix.

PR #1389, opened 2026-08-22
fix(model/openaimodel): pin function tools to non-strict validation

hanorik:
**Problem:**

`convertFunctionDeclaration` never set `FunctionToolParam.Strict`. openai-go
tags that field `json:"strict,omitzero"`, so leaving it unset drops the key from
the request body entirely, and the Responses API reads an absent `strict` as
"attempt strict mode; if the schema cannot be made compatible, fall back to
non-strict, best-effort function calling". The validation mode a tool runs under
was therefore inferred by the server from the shape of whatever parameter schema
the caller happened to produce, rather than chosen by this package.

That makes the mode inconsistent and invisible. Tools built with
`functiontool.New` get `additionalProperties:false` and a full `required` list
from jsonschema-go, so they ran strict; tools declared with a hand-written
`genai.Schema` emit no `additionalProperties` and never did. Adding `omitempty`
to a single Go struct field silently flipped that tool from validated to best
effort. The API reports the mode it settled on on the response tool, which this
package discards, so there was no way to tell which you got.

**Solution:**

Pin `Strict` to `false`, making the mode this package's decision rather than an
emergent property of the caller's struct tags. This matches adk-python's
Responses path. Pinning it *on* is the breaking direction: strict requires every
property to appear in `required`, so an optional argument would have to be
declared nullable (`"type": ["string", "null"]`) and listed there anyway,
changing what existing agents send. The trade is that tools whose schema already
qualified lose strict validation and drop to best effort; the package doc and
the conversion site record this and tell callers to validate arguments in the
tool.

### Testing Plan

**Unit Tests:** added `TestConvertFunctionDeclarationPinsStrictOff` (all three
parameter paths), `TestConvertFunctionDeclarationMarshalsStrict` (guards the
`omitzero` wire format), `TestConvertFunctionDeclarationKeepsOptionalParameters`
(pins that `enforceStrictOpenAISchema` is not applied to tool parameters),
`TestBuildOpenAIParams_ToolsPinStrictOff` (asserts on the marshalled request
body), plus a strict assertion in `TestConvertTools`.

    $ go test -race -count=1 ./model/openaimodel/
    ok  google.golang.org/adk/v2/model/openaimodel  1.391s

### DIFF
--- a/model/openaimodel/tools.go
+++ b/model/openaimodel/tools.go
@@ -67,6 +67,20 @@ func ensureFunctionToolOnly(idx int, tool *genai.Tool) error {
 // converts it into an OpenAI-specific responses.FunctionToolParam. We handle
 // the function's name, description, and importantly, convert its parameters
 // from a generic schema format to a map[string]any that the OpenAI API expects.
+//
+// Strict parameter validation is pinned off so the mode does not depend on the
+// caller's schema. Left unset, the API picks it from the schema shape, reports
+// the choice only on the response tool, which this package discards, and does
+// not merely accept or reject: given additionalProperties:false with only some
+// properties in required, it rewrites required to name them all and runs strict,
+// forcing a value into an argument the caller made optional.
+//
+// Pinning on is the breaking direction. Strict needs every property in
+// required, so an optional one must be nullable ("type": ["string", "null"]):
+// functiontool.New emits that for pointer fields, but omitempty and omitzero
+// fields are left out of required, and genai.Schema emits "nullable": true,
+// which strict rejects. Strict also demands additionalProperties:false on every
+// object; a map-typed field emits a schema there.
 func convertFunctionDeclaration(fn *genai.FunctionDeclaration) (*responses.FunctionToolParam, error) {
 	if fn == nil {
 		return nil, fmt.Errorf("openai: nil function declaration")
@@ -97,6 +111,7 @@ func convertFunctionDeclaration(fn *genai.FunctionDeclaration) (*responses.Funct
 		Name:       fn.Name,
 		Type:       constant.Function("function"),
 		Parameters: paramsMap,
+		Strict:     param.NewOpt(false),
 	}
 	if fn.Description != "" {
 		fnParam.Description = param.NewOpt(fn.Description)
--- a/model/openaimodel/tools.go
+++ b/model/openaimodel/tools.go
@@ -67,6 +67,16 @@ func ensureFunctionToolOnly(idx int, tool *genai.Tool) error {
 // converts it into an OpenAI-specific responses.FunctionToolParam. We handle
 // the function's name, description, and importantly, convert its parameters
 // from a generic schema format to a map[string]any that the OpenAI API expects.
+//
+// Strict parameter validation is pinned off so the mode does not depend on the
+// caller's schema. Left unset, the Responses API normalizes the parameters into
+// strict mode when it can and silently falls back when it cannot, reporting the
+// mode it chose on the response tool, which this package does not surface. That
+// makes the pin a real change for tools whose schema already qualified, as
+// functiontool.New produces unless the argument struct uses omitempty or
+// omitzero; pinning it on instead is the breaking direction, because strict
+// requires every property in required, so an optional argument would have to be
+// declared nullable ("type": ["string", "null"]) and listed there anyway.
 func convertFunctionDeclaration(fn *genai.FunctionDeclaration) (*responses.FunctionToolParam, error) {
 	if fn == nil {
 		return nil, fmt.Errorf("openai: nil function declaration")
@@ -97,6 +107,7 @@ func convertFunctionDeclaration(fn *genai.FunctionDeclaration) (*responses.Funct
 		Name:       fn.Name,
 		Type:       constant.Function("function"),
 		Parameters: paramsMap,
+		Strict:     param.NewOpt(false),
 	}
 	if fn.Description != "" {
 		fnParam.Description = param.NewOpt(fn.Description)
--- a/model/openaimodel/tools_test.go
+++ b/model/openaimodel/tools_test.go
@@ -15,6 +15,7 @@
 package openaimodel
 
 import (
+	"encoding/json"
 	"reflect"
 	"strings"
 	"testing"
@@ -88,6 +89,15 @@ func TestConvertTools(t *testing.T) {
 			}
 			if len(tools) > 0 && (tools[0].OfFunction == nil || tools[0].OfFunction.Name != "fn1") {
 				t.Fatalf("unexpected tool: %+v", tools[0])
+			}
+			for i, tool := range tools {
+				if tool.OfFunction == nil {
+					t.Errorf("tool %d is not a function tool: %+v", i, tool)
+					continue
+				}
+				if !tool.OfFunction.Strict.Valid() || tool.OfFunction.Strict.Value {
+					t.Errorf("tool %d Strict = %+v, want an explicit false", i, tool.OfFunction.Strict)
+				}
 			}
 		})
 	}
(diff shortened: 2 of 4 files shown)


karolpiotrowicz:
The change is right, and there is a measurement that makes a stronger case for it than the description does.

I ran the three-way comparison against the live Responses API. With `strict` absent the API does not fall back to non-strict for a schema that has an optional argument. It rewrites the caller's `required` to list every property and runs strict anyway:

```
$ curl -sS https://api.openai.com/v1/responses \
    -H "Authorization: Bearer $OPENAI_API_KEY" -H "Content-Type: application/json" \
    -d '{"model":"gpt-4o-mini","max_output_tokens":64,
         "input":"What is the weather in Paris? Do not specify units.",
         "tool_choice":{"type":"function","name":"get_weather"},
         "tools":[{"type":"function","name":"get_weather","description":"d",
           "parameters":{"type":"object",
             "properties":{"city":{"type":"string"},
                           "units":{"type":"string","description":"optional"}},
             "required":["city"],"additionalProperties":false}}]}'

strict absent  ->  echoed strict=true   echoed required=["city","units"]   arguments {"city":"Paris","units":""}
strict: false  ->  echoed strict=false  echoed required=["city"]           arguments {"city":"Paris"}
strict: true   ->  HTTP 400 invalid_function_parameters:
                   "'required' is required to be supplied and to be an array
                    including every key in properties. Missing 'units'."
```

So the pre-change behavior was not only that the mode was chosen by the server and invisible. For any tool with an optional argument the API forced a value into that argument: the empty string in `units` is the model being made to fill a field the prompt told it to leave out. Setting `strict: false` at [tools.go:110](https://github.com/google/adk-go/blob/907ce1419bac363dfd5456aa5db3c3dd75f66fae/model/openaimodel/tools.go#L110) removes that, which is worth saying in the description — it makes this a bug fix rather than a consistency argument, and it corrects the premise there that the API falls back whenever the schema cannot be made compatible. For this schema it did not fall back.

One thing I would fix before merge. [TestConvertFunctionDeclarationKeepsOptionalParameters](https://github.com/google/adk-go/blob/907ce1419bac363dfd5456aa5db3c3dd75f66fae/model/openaimodel/tools_test.go#L295-L317) is the test that pins "do not rewrite `required`", and it exercises only the `genai.Schema` branch — it passes `Parameters:`, never `ParametersJsonSchema:`. But `functiontool.New` populates `ParametersJsonSchema` at [function.go:166](https://github.com/google/adk-go/blob/907ce1419bac363dfd5456aa5db3c3dd75f66fae/tool/functiontool/function.go#L166), so every tool built the idiomatic way takes the [other branch](https://github.com/google/adk-go/blob/907ce1419bac363dfd5456aa5db3c3dd75f66fae/model/openaimodel/tools.go#L92-L96), which the test does not reach. I checked the other `ParametersJsonSchema` fixtures in the package and none of them would catch it either — each one either has no `properties` or already lists every property in `required`. Adding `enforceStrictOpenAISchema(paramsMap)` to that branch leaves the whole suite green.

Smaller: the exception list at [doc.go:28-29](https://github.com/google/adk-go/blob/907ce1419bac363dfd5456aa5db3c3dd75f66fae/model/openaimodel/doc.go#L28-L29) reads as closed — "unless the argument struct uses omitempty or omitzero" — and there is a third case. A map-typed field emits `additionalProperties` as a schema rather than `false` on the nested object, so it is strict-incompatible too:

```
Tags map[string]string  ->  "tags":{"additionalProperties":{"type":"string"},"type":"object"}
```

The rest of the doc claim holds, for what it is worth: I ran each declaration shape through `convertFunctionDeclaration` and all-required structs, pointer-optional structs, nested structs and slice fields do emit strict-compatible schemas, so the validation those tools lose is real.

karolpiotrowicz:
The test fix holds up. I injected `enforceStrictOpenAISchema(paramsMap)` into the `ParametersJsonSchema` branch and [the new case](https://github.com/google/adk-go/blob/8daa97f510076884674ade58b5828a22cca8a1cf/model/openaimodel/tools_test.go#L299-L360) fails on it with `required = [city units], want [city]`. Restoring the previous version of the file, that same mutant passes, and so does the rest of the suite. Omitting `units` from the fixture's `required` is what makes it work, and the comment saying so is worth keeping.

One correction on the new comment text, though. [tools.go:81-82](https://github.com/google/adk-go/blob/8daa97f510076884674ade58b5828a22cca8a1cf/model/openaimodel/tools.go#L81-L82) says `genai.Schema` emits `"nullable": true`, "which strict rejects". It does emit that — I ran a `Nullable: true` schema through `convertFunctionDeclaration` and got `"units":{"nullable":true,"type":"string"}` — but strict accepts it and ignores it rather than rejecting it. Sent with `additionalProperties:false` and both properties in `required`, so the nullable form is the only variable:

```
strict:true, units = {"nullable":true,"type":"string"}   -> accepted, arguments {"city":"Paris","units":""}
strict:true, units = {"type":["null","string"]}          -> accepted, arguments {"city":"Paris","units":null}
```

Same prompt as before, "What is the weather in Paris? Do not specify units.", tool call forced, `gpt-4o-mini`. I did not try other models, so I only know this holds there.

Your conclusion is unaffected and I think it gets stronger: you still cannot express an optional argument through `genai.Schema` under strict, so pinning on would still break those callers. The failure just is not a 400 they would notice — it is the same silent forced empty value the rest of this change is about.

wolo-lab:
Closing as stale, feel free to reopen.

wojtas000:
Pushed two commits: a merge of `main` to clear the conflict, and corrections to two claims in the comments that the live API contradicts.

- Strict does not reject `genai.Schema`'s `"nullable": true`. It accepts and ignores it, forcing the empty value instead — the same silent forcing this PR exists to stop. Your conclusion holds; the mechanism was wrong.
- The `required` rewrite does not turn on `additionalProperties: false`. Schemas carrying no such key get rewritten too, while one whose nested map-typed field cannot be normalized is left alone. What decides it is whether the schema as a whole is strict-eligible.

Also added the no-parameters case to `TestConvertFunctionDeclarationKeepsOptionalParameters`: injecting `enforceStrictOpenAISchema` into that branch survived the suite before.

model/openaimodel/tools.go
```diff
@@ -67,6 +67,11 @@ func ensureFunctionToolOnly(idx int, tool *genai.Tool) error {
 // converts it into an OpenAI-specific responses.FunctionToolParam. We handle
 // the function's name, description, and importantly, convert its parameters
 // from a generic schema format to a map[string]any that the OpenAI API expects.
+//
+// Strict is pinned off rather than left unset: an absent flag lets the API pick
+// the mode from the schema shape and add the caller's optional arguments to
+// required. Pinning it on would do that permanently, so off is the only value
```
karolpiotrowicz, line 73: The antecedent of "that" reads as "add the caller's optional arguments to required", and `strict: true` does not do that. A schema whose `required` omits a property fails requirement 2 of strict mode, and the documented outcome is rejection, not a rewrite: `HTTP 400 invalid_function_parameters`, "'required' is required to be supplied and to be an array including every key in properties". The three cases are genuinely different from each other — absent lets the server normalize and rewrite `required`, `true` is rejected outright when a property is missing from it, and `false` sends the caller's `required` as written.

What makes this worth a fix rather than a nit: per the commit message for 910c53b this comment is what stays behind specifically to stop someone setting the flag to `true`, and as written it describes that edit as a tolerable degradation. Whatever the sentence ends up saying, the property to preserve is that `true` fails the request rather than silently widening `required` — a reader who takes the current wording at face value will reach for it.

One narrower point in the same sentence: "off is the only value that keeps an optional argument optional" is a statement about adk-go rather than about the API. Strict mode does support optional fields, via `"type": ["string", "null"]` with the property listed in `required`. What rules it out here is that `genai.Schema` emits `"nullable": true` instead, and `functiontool`'s `omitempty` fields are left out of `required` altogether.

model/openaimodel/tools_test.go
```diff
@@ -198,6 +208,166 @@ func TestConvertFunctionDeclaration(t *testing.T) {
 	}
 }
 
+// TestConvertFunctionDeclarationPinsStrictOff covers every parameter path,
```
karolpiotrowicz, line 211: These three cases cannot fail for different reasons. `Strict` is assigned once and unconditionally at [`tools.go:105`](https://github.com/google/adk-go/blob/910c53b17dc0255953eed330fb4ff0113ca6de5c/model/openaimodel/tools.go#L105), after all three parameter branches have converged on `paramsMap`, and the subtests assert only `fn.Strict.Valid()` and `fn.Strict.Value` without ever touching `fn.Parameters` — so no input makes one case fail while another passes. Deleting the assignment, or flipping it to `true`, fails all three together.

The justification in the comment is the part that is actually off: it explains the three cases with a fact about what the Responses API does with an absent flag, where the sibling `TestConvertFunctionDeclarationKeepsOptionalParameters` uses nearly the same sentence and there it holds, because those three cases do produce three different `fn.Parameters` and each kills a distinct mutant.

Keeping all three as a guard against a future refactor that moves the assignment into a branch is a defensible call. Saying that is what they are for would be more accurate than the current rationale.

model/openaimodel/tools_test.go
```diff
@@ -89,6 +90,15 @@ func TestConvertTools(t *testing.T) {
 			if len(tools) > 0 && (tools[0].OfFunction == nil || tools[0].OfFunction.Name != "fn1") {
 				t.Fatalf("unexpected tool: %+v", tools[0])
 			}
+			for i, tool := range tools {
+				if tool.OfFunction == nil {
```
karolpiotrowicz, line 94: This branch cannot be reached. `convertTools` only ever appends `ToolUnionParam{OfFunction: fn}` with a non-nil `fn` — it returns on error before the append — and the `t.Fatalf` a few lines above already stops the test if `tools[0].OfFunction` is nil. Of the four subtests, "empty config" produces no tools and the two error cases return earlier, so only "valid config" enters the loop at all, against a single element.

karolpiotrowicz (approved):
Nothing here blocks. The change does what it says, the invariants hold, and the one thing worth fixing before merge is a sentence in the comment on `convertFunctionDeclaration`.

I ran the build, the full suite, vet, staticcheck, govulncheck and an exported-surface check against the merge result of `main` and this branch, and re-ran the touched tests against unmodified `main` to confirm they fail without the change. Two packages fail under the sandbox — `session/vertexai` and `telemetry` — and both fail identically on the base commit under the same time budget, so neither is yours. I also re-ran the mutation you describe in [c3e83cb](https://github.com/google/adk-go/commit/c3e83cb234f08d3c219f990c8864f89a9aa29496): injecting `enforceStrictOpenAISchema` into each of the three parameter branches is now caught in all three, and the no-parameters case genuinely did survive at the earlier revision ([8daa97f](https://github.com/google/adk-go/commit/8daa97f510076884674ade58b5828a22cca8a1cf)). That test earns its keep.

**The comment gets `strict: true` backwards.** Details are on the line, but in short: [`tools.go:71-74`](https://github.com/google/adk-go/blob/910c53b17dc0255953eed330fb4ff0113ca6de5c/model/openaimodel/tools.go#L71-L74) says pinning the flag on "would do that permanently", where "that" reads as adding the caller's optional arguments to `required`. Sending `strict: true` with a schema whose `required` omits a property does not add it — the request comes back 400. This is the comment whose stated job is to stop someone setting the flag to `true`, and it describes that edit as a survivable degradation.

**The `additionalProperties` half of the correction is not established, and it is the half that matters now.** The commit message for [c3e83cb](https://github.com/google/adk-go/commit/c3e83cb234f08d3c219f990c8864f89a9aa29496) says schemas carrying no `additionalProperties` key get the `required` rewrite too, measured against the live API. I could not reproduce that — there is no OpenAI credential on this machine — and the run recorded in the thread back in August does not settle it either, because every schema in it carried `additionalProperties: false`. The published guide cuts both ways: read as a property of the schema, requirement 1 makes a key-absent schema strict-incompatible and predicts fallback, while "attempt to normalize your schema into strict mode when possible" would license the server to inject the key itself.

This does not change the verdict, because the sentence asserting it was cut in [910c53b](https://github.com/google/adk-go/commit/910c53b17dc0255953eed330fb4ff0113ca6de5c) and nothing at head depends on it. It does bound how much the change buys: if the rewrite needs `additionalProperties: false`, then tools declared through `genai.Schema` were already falling back to non-strict and see no behavior change, since `schemaToMap` never emits that key — which your own fixture pins at [`tools_test.go:323`](https://github.com/google/adk-go/blob/910c53b17dc0255953eed330fb4ff0113ca6de5c/model/openaimodel/tools_test.go#L323). Three calls would close it, and the result is worth having in the thread whichever way it comes out:

```
# strict omitted, tool call forced, same prompt in all three arms.
# Read the echoed tool object off the response, not just the arguments.

A. properties {city, units}, required ["city"], NO additionalProperties key
   -> echoed strict:true  + required ["city","units"]  = the rewrite does not need the key
   -> echoed strict:false + required ["city"]          = it does

B. the same schema WITH additionalProperties:false
   -> control: must reproduce the August result, or arm A proves nothing

C. a nested map-typed property, every other strict requirement met
   -> tests the "left alone" half
```

**One note on attribution, for the record.** The reopening comment reads as correcting two claims I made. It is correcting the PR's own comment text, which is what the commit message says: "Two claims in the convertFunctionDeclaration comment do not match the Responses API." On the `nullable` half specifically, the correction is the finding I posted on 24 August — I wrote that `tools.go` said strict "rejects" it and that strict in fact accepts and ignores it, forcing the empty value. "Your conclusion holds, the mechanism was wrong" is my own sentence coming back to me. On the `additionalProperties` half I never claimed the key was the trigger, and the map-typed case the correction cites is one I supplied. Worth a line in the thread so the record is straight.

The parity claim checks out: adk-python makes the same split in `_openai_responses_llm.py`, [`strict=False` on the function tool](https://github.com/google/adk-python/blob/3a06a8621d04fd87aad9f6b078c999c78a96fb66/src/google/adk/labs/openai/_openai_responses_llm.py#L510) and [`'strict': True` on the response-format schema](https://github.com/google/adk-python/blob/3a06a8621d04fd87aad9f6b078c999c78a96fb66/src/google/adk/labs/openai/_openai_responses_llm.py#L247-L252), which is exactly what this PR leaves in place.

**Smaller things**, none of them blocking, and two are on the lines they concern.

- Callers whose schemas already list every property in `required` lose whatever server-side enforcement they were getting, with no way to opt back in. The package is experimental and this PR documents the cost, so I would not hold the merge for it — but a per-tool knob is the shape of the eventual answer and is worth an issue rather than a follow-up nobody files.
- Three things I found in code this PR does not touch, all of which I am happy to file separately: `transfer_to_agent` accepts any non-empty string for `agent_name` despite declaring an enum, and returns success, so a bad name aborts the invocation instead of coming back to the model as a retryable error. `mcptoolset` forwards the model's argument map to the remote server with no check against the schema it declared. And `normalizeSchema` skips the `lowercaseSchemaTypes` call that `schemaToMap` makes, so a `genai.Schema` routed through `ParametersJsonSchema` goes out with uppercase types.
- [`doc.go:46-48`](https://github.com/google/adk-go/blob/910c53b17dc0255953eed330fb4ff0113ca6de5c/model/openaimodel/doc.go#L46-L48) is accurate about the API, and dropping "validate them in the tool" was right for the reason you gave. The reader it did apply to — someone implementing `tool.Tool` by hand — now gets neither the warning nor the reassurance that `functiontool` covers them. One clause either way would help.

PR #1392, opened 2026-08-23
fix(model/openaimodel): account for every generation config field

hanorik:
**Problem:**

`applyGenerationConfig` rejects six unsupported knobs with typed errors — `TopK`, `StopSequences`, `CandidateCount > 1`, the penalties, `Labels`, `SafetySettings` — but silently drops roughly a dozen more, among them `ThinkingConfig`, `Seed`, `ServiceTier`, `CachedContent`, `ResponseModalities`, `SpeechConfig`, `RoutingConfig` and `ImageConfig`. Six loud rejections against a dozen silent drops is not a coherent contract: a caller setting `ThinkingConfig` to control reasoning effort gets no error and no effect, and nothing distinguishes that from a setting that worked.

**Solution:**

Every top-level field of `genai.GenerateContentConfig` is now either translated to the Responses API or rejected with an error. `ThinkingConfig` is translated rather than rejected: `ThinkingLevel` maps to the matching `reasoning.effort`, an explicit `THINKING_LEVEL_UNSPECIFIED` resolves to `medium`, and a token budget collapses to `minimal` at zero and `medium` otherwise, since Responses is effort-based and has no budget knob. Everything genuinely untranslatable returns `ErrUnsupportedConfigField` naming the offending field, while the six long-standing knobs keep their own typed errors and are checked first, so existing `errors.Is` call sites are unaffected. Rejection keys on presence rather than value — setting a knob at all means the caller expected an effect — and a reflection test over `GenerateContentConfig` fails if a future genai field is neither translated nor rejected, which is the silent failure mode that produced this bug.

---

**Update after review.**

`reasoning.summary` is now sent only when `ThinkingConfig.IncludeThoughts` is set. Summaries require a verified OpenAI organization and that is enforced per model, not per account — the same key that gets summaries from `gpt-5.4-nano` is refused by `o4-mini` — so the previous unconditional `summary: concise` turned every reasoning call into a 400 for callers who never asked for summaries.

A zero `ThinkingBudget` maps to the `none` effort rather than `minimal`, which current models also refuse (`gpt-5.4-nano` supports `none, low, medium, high, xhigh`). `ThinkingBudget: -1`, genai's "let the model decide", sends no effort instead of a hardcoded `medium`. An explicit `ThinkingLevel: MINIMAL` still sends `minimal`, so a model lacking it says so by name rather than being upgraded to a costlier effort. `ThinkingLevel` goes through an explicit map, so a level genai adds later is named in an error instead of reaching the wire unvalidated.

`ServiceTier` is translated rather than rejected: every tier genai names has a Responses equivalent, with `standard` reaching OpenAI as `default` and an explicit `unspecified` as `auto`.

Of `HTTPOptions`, which is transport rather than generation, only a positive `Timeout` crosses. It is applied as a context deadline on the whole call, because that is what genai documents the field to mean; openai-go's own per-request timeout applies inside its retry loop, so using that would let a caller asking for five seconds wait fifteen. `BaseURL`, `BaseURLResourceScope`, `APIVersion`, `ExtraBody`, `ExtrasRequestProvider` and `RetryOptions` describe the Gemini wire format or belong to `ClientConfig`, and are rejected by name.

`Headers` are neither forwarded nor rejected. Forwarding is unsafe: openai-go treats a caller `Authorization` as an override and then withholds the configured API key, so the request would go out authenticated by the caller's header alone, and a Gemini credential such as `x-goog-api-key` would reach OpenAI verbatim. Rejecting would be a break, because `model/gemini` writes `x-goog-api-client` and `user-agent` into this field on every request and the compaction summarizer forwards it, so a config arriving here can carry headers the application never set. They are therefore ignored, which is also what `main` does today. Headers intended for OpenAI belong on `ClientConfig.Options`.

A setting that asks for nothing is never rejected: an empty `ThinkingConfig` or `HTTPOptions` is satisfied by sending nothing.

**Billing.** On `main` every `ServiceTier` value was dropped, so a config carrying `ServiceTierPriority` was ignored; it is now honored and billed at OpenAI priority rates (and `flex` at flex rates). `ServiceTierUnspecified` maps to `default` rather than `auto`, following genai's own "Default service tier, which is standard", so it cannot silently pick up whatever tier a project has configured.

**Compaction summarizer.** `session/compaction/llm_summarizer.go`'s `summarizerGenConfig` forwards `Seed` and `HTTPOptions` from the application's config into the summarizer's, so five shapes that reach an OpenAI-backed summarizer through that line succeed on `main` and fail here: `Seed`, `HTTPOptions.APIVersion`, `HTTPOptions.BaseURL`, `HTTPOptions.RetryOptions`, and a zero `Timeout`. `TopK`, `Labels` and `SafetySettings` already failed before this change, so this widens an existing problem by five rather than creating one. The auto-wired path is same-provider and the agent's own turn fails first; reaching the summarizer-only failure takes a hand-built `compaction.NewLLMSummarizer` pairing an OpenAI model with a Gemini agent's config, and it degrades to `compaction.ErrCompaction` with the run continuing uncompacted. The durable fix belongs in `summarizerGenConfig`, which forwards transport and sampling settings that mean nothing for a summarization; worth its own issue covering all eight fields.

**Known gap, not addressed here.** adk-python maps `stop_sequences` to `extra_body['stop']` (`_openai_responses_llm.py:1129-1132`) while this package returns `ErrStopSequencesNotSupported`. That error predates this change (`request.go:331` on `main`), so it is out of scope; filing separately.

**Behavior change for upgraders.** Previously-silent config drops are now errors: `Seed`, `CachedContent`, `ResponseModalities`, `MediaResolution`, `SpeechConfig`, `AudioTimestamp`, `ImageConfig`, `EnableEnhancedCivicAnswers`, `ModelArmorConfig`, `AudioTranscriptionConfig`, `RoutingConfig` and `ModelSelectionConfig` return `ErrUnsupportedConfigField`, as do a negative `MaxOutputTokens` or `CandidateCount` and `Logprobs` without `ResponseLogprobs`. The pre-existing typed errors are checked first and keep their precedence, so existing `errors.Is` call sites are unaffected. This is stricter than adk-python, which drops these in silence, so a config ported from Python may need fields removed.

**Testing Plan.**

- `golangci-lint run` and `go mod tidy -diff` — clean in both modules
- Exercised against the live OpenAI API on `gpt-5.4-nano` and `o4-mini` through the public ADK surface, blocking and streaming: a translated `ThinkingConfig` succeeds, the request body carries the expected `reasoning.effort` and carries `summary` only when `IncludeThoughts` is set, a dynamic budget sends no effort, a config with no `ThinkingConfig` sends no `reasoning` block, a reasoning summary is not replayed as assistant speech on the following turn, `service_tier` arrives as `flex` and as `default`, a caller header never reaches the provider on either the blocking or the streaming path, and a zero timeout is refused before any request is built.

**Review loop.** Two panel rounds after the 2026-09-04 review. Round 1 (six lenses) returned one blocker — a pairing test that restated its case list in a literal and so passed while a predicate never ran — plus over-long comments against the package's two-sentence rule, a tautology in the translated-fields assertion, and one unredacted header value. Round 2 returned correctness CLEAR and withdrew a parity finding that had been raised against the wrong adk-python files. Deliberately not fixed: the `stop_sequences` parity gap and the `summarizerGenConfig` forwarding, both pre-existing and disclosed above.

Every rejection is also mutation-checked: neutering any one of the six `unsupportedHTTPOptionFields` predicates, dropping either `ctx := ctx`, deleting either `context.WithTimeout` block, or re-adding header forwarding to either path each turns the suite red.


karolpiotrowicz:
The `ThinkingConfig` translation makes every request that uses it fail against the Responses API. [`request.go:384`](https://github.com/google/adk-go/blob/3992b4ae327aef49c265905cf9b6082bf7c2dbb5/model/openaimodel/request.go#L384) sets `Summary: shared.ReasoningSummaryConcise` on every translated `ThinkingConfig`. The caller never asks for a summary and has no way to suppress it, and requesting one at all requires a verified OpenAI organization. Before this change `ThinkingConfig` was dropped and the call went through, so this turns working calls into 400s for exactly the users who adopt the new feature.

Driven through the exported `GenerateContent` against `o4-mini` with `MaxOutputTokens: 2000`:

```
control: no ThinkingConfig  -> OK, 1 response
ThinkingLevel=MEDIUM        -> ERROR POST /v1/responses: 400 Bad Request
ThinkingLevel=HIGH          -> ERROR POST /v1/responses: 400 Bad Request
ThinkingBudget=0            -> ERROR POST /v1/responses: 400 Bad Request
ThinkingBudget=2048         -> ERROR POST /v1/responses: 400 Bad Request
```

The cause, isolated with direct calls on the same account and model:

```
{"effort":"medium"}                      -> HTTP 200
{"effort":"medium","summary":"concise"}  -> HTTP 400  param=reasoning.summary
     "Your organization must be verified to generate reasoning summaries."
{"effort":"medium","summary":"auto"}     -> HTTP 400  same message
{"effort":"medium","summary":"detailed"} -> HTTP 400  same message
{"effort":"minimal"}                     -> HTTP 400  'minimal' is not supported with 'o4-mini'
{"effort":"none"}                        -> HTTP 400  'none' is not supported with 'o4-mini'
```

Omitting `Summary` is enough to make the feature work — that is the passing control in the second block. Two constraints on any fix, both visible in that same block. The last two lines show `minimal` is separately rejected by o-series models, so the `ThinkingBudget: 0` mapping at [`request.go:379-380`](https://github.com/google/adk-go/blob/3992b4ae327aef49c265905cf9b6082bf7c2dbb5/model/openaimodel/request.go#L379-L380) hits the same wall on its own, and `none` is not an escape because `o4-mini` rejects that too. Separately, [`request_test.go:348`](https://github.com/google/adk-go/blob/3992b4ae327aef49c265905cf9b6082bf7c2dbb5/model/openaimodel/request_test.go#L348) asserts `Summary: shared.ReasoningSummaryConcise` across all nine subtests, so the suite currently pins this behavior rather than catching it.

Two smaller points on the "presence, not value" rule, neither blocking. [`request.go:402`](https://github.com/google/adk-go/blob/3992b4ae327aef49c265905cf9b6082bf7c2dbb5/model/openaimodel/request.go#L402) uses `len(c.ResponseModalities) > 0`, so an explicitly-set empty slice reads as unset — presence is representable for a slice, unlike the plain `bool` at [`request.go:405`](https://github.com/google/adk-go/blob/3992b4ae327aef49c265905cf9b6082bf7c2dbb5/model/openaimodel/request.go#L405), where `AudioTimestamp: false` genuinely cannot be told from unset. That second case makes the unqualified "none is dropped in silence" at [`doc.go:24-25`](https://github.com/google/adk-go/blob/3992b4ae327aef49c265905cf9b6082bf7c2dbb5/model/openaimodel/doc.go#L24-L25) worth softening.

wojtas000:
All three fixed, plus two more the live API surfaced.

- **`Summary: concise` on every `ThinkingConfig`** — [`Summary` now rides on `IncludeThoughts`](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L536-L540) and is sent as `auto`, so nobody gets a summary they did not ask for; confirmed against your `o4-mini`, which refuses summaries on the same key that gets them from `gpt-5.4-nano`, so the verification gate is per model rather than per account.
- **`ThinkingBudget: 0` → `minimal`** — [now maps to `none`](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L519-L523), because `minimal` is refused by current models too (`gpt-5.4-nano` lists `none, low, medium, high, xhigh`) while `none` is both accepted there and the exact sense of "do not think"; an explicit `ThinkingLevel: MINIMAL` still sends `minimal` so the model names the refusal instead of being silently upgraded to a costlier effort.
- **`ResponseModalities` using `len() > 0`** — [keys on `nil` now](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L590), since a slice can represent an explicit empty and the rule is presence.
- **`doc.go`'s unqualified "none is dropped in silence"** — [softened, and generalised](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/doc.go#L31-L38): `AudioTimestamp` is not special, every plain bool and string in the list has the same blind spot, so the docs state the rule rather than one instance of it.
- **The nine subtests pinning `Summary`** — replaced; [a dedicated test](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request_test.go#L510) now asserts the summary is *absent* unless `IncludeThoughts` is set, so this cannot regress quietly again.

Two you did not raise, found while fixing the above:

- Rejecting orphan `Logprobs` at the point `Logprobs` is translated put it ahead of the `Labels`, `SafetySettings` and MIME-type checks, so a caller setting `Labels` too silently stopped getting `ErrLabelsNotSupported` — [moved to a sweep](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L553-L556) that runs after every named error, with [a precedence case per error](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request_test.go#L975) rather than only `TopK`.
- `ThinkingBudget: -1` is genai's "you decide" and was answered with a hardcoded `medium`; it [now sends no effort at all](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L524-L526), and [a budget below `-1`](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L491-L495), [a negative `MaxOutputTokens` and a negative `CandidateCount`](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L557-L564) are rejected instead of vanishing.

Rebased onto main, which the branch no longer compiled against after the openai-go v3.54 bump. Build, tests, lint and `tidy -diff` clean in both modules, and the whole thing exercised against the live API on `gpt-5.4-nano` and `o4-mini`.

wojtas000:
All three are right — fixed in `dbf4ce5`.

- **`HTTPOptions` (`request.go:506`)** — was the worst of the three and is not even in b/544804975's list of twelve; [`Headers` and `Timeout` are now translated](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L428-L444) into openai-go request options so a caller's tracing header actually arrives, [added rather than set](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L433-L438) so they cannot displace authorization.
- **What remains of `HTTPOptions`** — `BaseURL`, `BaseURLResourceScope` and `APIVersion` belong to `ClientConfig` alongside the key that authenticates against them, `ExtraBody` and `ExtrasRequestProvider` shape a Gemini body this package does not send, and `RetryOptions` has no faithful equivalent, so [those five stay rejected](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L448-L463), now named individually.
- **`ThinkingConfig{IncludeThoughts:false}` (`request.go:461`)** — [returns nil now](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L530-L534): a struct with nothing set asks for nothing, and not including thoughts is a request this package satisfies rather than one it cannot honor.
- **`ServiceTier`** — [translated rather than rejected](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L390-L396), since [every tier genai names has an equivalent](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request.go#L408-L413); `standard` reaches OpenAI as `default` and an explicit `unspecified` as `auto`, and a tier genai adds later is named in an error instead of passed through.
- **Empty `HTTPOptions`** — [accepted](https://github.com/google/adk-go/blob/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847/model/openaimodel/request_test.go#L920) for the same reason as the empty `ThinkingConfig`, so the two now follow one rule.

Verified live against the API: `service_tier` arrives as `flex` and as `default`, an empty `ThinkingConfig` sends no `reasoning` block, and a custom header reaches the request. Build, tests, lint and `tidy -diff` clean in both modules.

wojtas000:
Confirmed and fixed in `e6ce0f1` — you were right, and my comment claimed the opposite of what the SDK does.

- **`Authorization` drops the key** — reproduced against a stub with a real key configured: the request went out as `Bearer caller` and the real key never appeared, lowercase too, because [`AddHeader` records it as an override](https://github.com/google/adk-go/blob/e6ce0f1bbf06ca22ab4a26be7e2b345c1569e344/model/openaimodel/request.go#L444-L457) and `ApplySecurity` then returns early.
- **Fix went wider than rejecting `Authorization`** — your third point is the reason: `HTTPOptions` describes a call to Gemini, so an `x-goog-api-key` on a shared config would reach `api.openai.com` verbatim, by the same compaction path that motivated honoring the field. A denylist of credential header names is not worth trusting, so [no header crosses](https://github.com/google/adk-go/blob/e6ce0f1bbf06ca22ab4a26be7e2b345c1569e344/model/openaimodel/request.go#L458) and headers for OpenAI belong on `ClientConfig.Options`.
- **The false comment** — [gone](https://github.com/google/adk-go/blob/e6ce0f1bbf06ca22ab4a26be7e2b345c1569e344/model/openaimodel/request.go#L415-L424); the [test now drives openai-go directly](https://github.com/google/adk-go/blob/e6ce0f1bbf06ca22ab4a26be7e2b345c1569e344/model/openaimodel/request_test.go#L874) and fails if a future version stops overriding, rather than restating the behavior in prose.
- **`Timeout: 0`** — [rejected as non-positive](https://github.com/google/adk-go/blob/e6ce0f1bbf06ca22ab4a26be7e2b345c1569e344/model/openaimodel/request.go#L581-L587), since openai-go reads zero as no deadline and would lift the caller's bound instead of applying it.
- **doc.go** — [states both reasons](https://github.com/google/adk-go/blob/e6ce0f1bbf06ca22ab4a26be7e2b345c1569e344/model/openaimodel/doc.go#L50-L62) rather than one line.

Gates clean in both modules; the refusals are also pinned offline so nothing can leave the process.

wojtas000:
Where `HTTPOptions` landed, since two comments above describe mechanisms that changed after them.

- **`Timeout`** — applied as a [context deadline on the whole call](https://github.com/google/adk-go/blob/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91/model/openaimodel/openai.go#L99-L107), not openai-go's per-request option: the SDK evaluates that one inside its retry loop, so five seconds asked for could mean fifteen waited.
- **`Headers`** — [ignored, not forwarded and not rejected](https://github.com/google/adk-go/blob/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91/model/openaimodel/request.go#L442-L466). Forwarding displaces the API key and would carry a Gemini credential to OpenAI; rejecting would break configs `model/gemini` fills in itself, since it writes `x-goog-api-client` and `user-agent` into this field on every request.
- **Everything else in `HTTPOptions`** — [rejected by name](https://github.com/google/adk-go/blob/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91/model/openaimodel/request.go#L468-L487).

Both paths are pinned by tests that fail on the mutation that motivated them: [no header reaches the wire](https://github.com/google/adk-go/blob/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91/model/openaimodel/request_test.go#L931) and [the deadline actually reaches the request](https://github.com/google/adk-go/blob/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91/model/openaimodel/request_test.go#L1554), each run against blocking and streaming.

model/openaimodel/doc.go
```diff
@@ -21,6 +21,12 @@
 // providers that expose the OpenAI Responses API surface. This package
 // allows for easy integration of OpenAI's language models into applications.
 //
+// Every top-level field of genai.GenerateContentConfig is either translated to
+// the Responses API or rejected with an error naming it, and the pre-existing
+// errors are checked first so existing errors.Is call sites are unaffected.
+// The single exception is HTTPOptions.Headers, ignored rather than forwarded
+// because headers addressed to another backend must not reach OpenAI.
```
karolpiotrowicz: These five sentences replace 85 lines that were right. The version at [`dbf4ce5`](https://github.com/google/adk-go/commit/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847) and [`fe4d045`](https://github.com/google/adk-go/commit/fe4d0450b1f932209ed22e58b79544e9cbb6a9d7) said presence is only observable where the zero value cannot itself be a setting, and named the cases where it is not. What is here now says `HTTPOptions.Headers` is the single exception, and three shapes fall outside it.

Driving `GenerateContent` at this revision, each of these produces a request with no error and no trace of the field:

```
AudioTimestamp: false      -> {"input":[...],"model":"gpt-5.4-nano"}
CachedContent: ""          -> {"input":[...],"model":"gpt-5.4-nano"}
MediaResolution: ""        -> {"input":[...],"model":"gpt-5.4-nano"}
```

While the neighbouring shapes are rejected, which is what makes the asymmetry worth documenting rather than fixing: `AudioTimestamp: true`, `EnableEnhancedCivicAnswers: Ptr(false)` and `ResponseModalities: []{}` all return `ErrUnsupportedConfigField`. The difference is `bool` versus `*bool`, not intent, and the deleted paragraph explained precisely that.

The table of translated, rejected and ignored fields went with it. That table was the one place a reader could answer "is my field supported" without reading `request.go`, and the commit that removed it is titled for making the contract readable.

Restoring the presence-observability paragraph would close the point from the last review. Restoring the table as well would be better.

model/openaimodel/request.go
```diff
@@ -383,6 +384,256 @@ func applyGenerationConfig(params *responses.ResponseNewParams, cfg *genai.Gener
...
+	// be set coherently alongside the API key that authenticates against it.
+	{"BaseURL", func(o *genai.HTTPOptions) bool { return o.BaseURL != "" }},
+	{"BaseURLResourceScope", func(o *genai.HTTPOptions) bool { return o.BaseURLResourceScope != "" }},
+	{"APIVersion", func(o *genai.HTTPOptions) bool { return o.APIVersion != "" }},
+	// Both shape a Gemini request body, which is not the body being sent.
+	{"ExtraBody", func(o *genai.HTTPOptions) bool { return o.ExtraBody != nil }},
+	{"ExtrasRequestProvider", func(o *genai.HTTPOptions) bool { return o.ExtrasRequestProvider != nil }},
+	// openai-go retries too, but on its own schedule; honoring only the retry
+	// count would quietly discard the backoff the caller asked for.
+	{"RetryOptions", func(o *genai.HTTPOptions) bool { return o.RetryOptions != nil }},
```
karolpiotrowicz, line 460: These six predicates have no test that exercises them. [`TestHTTPOptionFieldsAreAccountedFor`](https://github.com/google/adk-go/blob/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91/model/openaimodel/request_test.go#L1623-L1626) reflects over the field *names*, so it stays green whatever the closures do, and no test anywhere constructs a `genai.HTTPOptions` carrying `BaseURL`, `BaseURLResourceScope`, `APIVersion`, `ExtraBody`, `ExtrasRequestProvider` or `RetryOptions` — every `HTTPOptions` literal in the test file sets only `Timeout` or `Headers`.

Change one predicate body and nothing notices:

```go
// request.go, in unsupportedHTTPOptionFields
{"BaseURL", func(o *genai.HTTPOptions) bool { return false }},
```

```
$ go test ./model/openaimodel/
ok  	google.golang.org/adk/v2/model/openaimodel	0.269s
```

And through the exported entry point, with that one line changed, `HTTPOptions{BaseURL: "https://other.example"}` returns a nil error and the request goes out with the override dropped — the exact failure this PR exists to remove, in the table this PR adds.

The sibling table does not have this gap: making the same edit to the `Seed` entry of `unsupportedConfigFields` turns the suite red, because [`TestApplyGenerationConfigRejectsUnsupportedFields`](https://github.com/google/adk-go/blob/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91/model/openaimodel/request_test.go#L426) drives every entry with a live value. Mirroring that shape here is the whole fix.

I read all six and they are correct today. Six one-line closures in a parallel table is where a copy-paste eventually lands, though, and that is the case the reflection test cannot see.

model/openaimodel/openai.go
```diff
@@ -108,8 +121,17 @@ func (m *openAIModel) generate(ctx context.Context, params responses.ResponseNew
 	}
 }
 
-func (m *openAIModel) generateStream(ctx context.Context, params responses.ResponseNewParams) iter.Seq2[*model.LLMResponse, error] {
+func (m *openAIModel) generateStream(ctx context.Context, params responses.ResponseNewParams, timeout time.Duration) iter.Seq2[*model.LLMResponse, error] {
 	return func(yield func(*model.LLMResponse, error) bool) {
+		// Shadowed for the same reason as in generate: reassigning the captured
+		// ctx would poison a second range with the first one's cancellation.
+		ctx := ctx
+		// Bounds the whole stream, not just its first byte.
```
karolpiotrowicz: "Bounds the whole stream" also bounds the caller. The deadline is live across the `for stream.Next()` loop, so the time the range body spends is charged against `HTTPOptions.Timeout` alongside the provider's.

Against a stub writing as fast as it can, with the caller doing per-chunk work:

```
60000 SSE deltas, HTTPOptions.Timeout = 500ms,
consumer sleeping 200ms every 1000 chunks
  -> context deadline exceeded at 665ms, 3034 of 60000 delivered
```

The provider never stalled. A slow consumer truncates its own stream and gets an error naming a deadline it associates with the model call.

This only bites once the response outgrows the client's buffer — with ten chunks the whole body arrives first and the deadline never fires, which is why [`TestHTTPOptionsTimeoutReachesTheRequest`](https://github.com/google/adk-go/blob/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91/model/openaimodel/request_test.go#L1554), whose consumer returns immediately, cannot see it.

`generate` is unaffected in practice: the response is fully read before the yield and nothing consumes the context afterwards.

It is a defensible reading — `http.Client.Timeout` covers body reads too — so this is a documentation gap rather than a bug. Neither this comment nor [`requestTimeout`'s](https://github.com/google/adk-go/blob/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91/model/openaimodel/request.go#L415-L428) says the streaming bound counts the caller's own time, and that doc is otherwise thorough enough that a reader will take its silence as coverage.

model/openaimodel/request.go
```diff
@@ -383,6 +384,256 @@ func applyGenerationConfig(params *responses.ResponseNewParams, cfg *genai.Gener
...
+	// Last, so the named errors above win when a caller sets both.
+	return rejectUnsupportedConfigFields(cfg)
+}
+
+// serviceTiers maps genai's processing tiers onto the Responses equivalents.
+// OpenAI offers more tiers than genai can name; the ones genai has all
+// correspond. An explicit "unspecified" is the caller asking not to choose,
+// which is what auto means.
+var serviceTiers = map[genai.ServiceTier]responses.ResponseNewParamsServiceTier{
+	genai.ServiceTierUnspecified: responses.ResponseNewParamsServiceTierAuto,
```
karolpiotrowicz: genai documents [`ServiceTierUnspecified`](https://github.com/googleapis/go-genai/blob/fbdae87185987c7fc1abbe9e0b023c614bc8dc3b/types.go#L855-L856) as "Default service tier, which is standard", at the v1.69.0 release this module pins, and the next line maps `ServiceTierStandard` to `default`. So the two constants genai treats as the same thing reach OpenAI as different things, where `auto` uses whatever tier the project has configured and `default` pins standard pricing and routing.

A project with a scale or priority tier configured would get that tier for a caller who asked for standard by name.

Narrow, since `ServiceTier` is gated on `!= ""` above and the Go zero value never reaches the map — a caller has to type the constant. Worth either mapping it alongside `Standard`, or saying in the comment that genai's own reading is being overridden and why.

model/openaimodel/request.go
```diff
@@ -383,6 +384,256 @@ func applyGenerationConfig(params *responses.ResponseNewParams, cfg *genai.Gener
...
+//
+// It runs after every named error so that a caller who set one of those too
+// gets the error they have always got, rather than this sentinel jumping the
+// queue and breaking their errors.Is.
+func rejectUntranslatableValues(cfg *genai.GenerateContentConfig) error {
+	switch {
+	case cfg.Logprobs != nil && !cfg.ResponseLogprobs:
+		// Logprobs only sizes the list ResponseLogprobs asks for, so alone it
+		// reaches neither params nor the wire.
+		return fmt.Errorf("%w: Logprobs without ResponseLogprobs", ErrUnsupportedConfigField)
```
karolpiotrowicz, line 556: This switch rejects a translated field's untranslatable *values*, and skips the one next to it. `Logprobs` reaches the wire unchecked when `ResponseLogprobs` is set, though `top_logprobs` is documented as 0 to 20:

```
ResponseLogprobs: true, Logprobs: -5    -> {"top_logprobs":-5,...}
ResponseLogprobs: true, Logprobs: 999   -> {"top_logprobs":999,...}
```

The translation at [`request.go:340-344`](https://github.com/google/adk-go/blob/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91/model/openaimodel/request.go#L340-L344) is unchanged from `main`, so this is not something the PR introduced, and an out-of-range value fails as a named 400 rather than a silent drop. It only stands out because the two cases below it were added under exactly the rule it falls outside.

model/openaimodel/request.go
```diff
@@ -383,6 +384,256 @@ func applyGenerationConfig(params *responses.ResponseNewParams, cfg *genai.Gener
...
+	// openai-go retries too, but on its own schedule; honoring only the retry
+	// count would quietly discard the backoff the caller asked for.
+	{"RetryOptions", func(o *genai.HTTPOptions) bool { return o.RetryOptions != nil }},
+}
+
+// reasoningEfforts maps every genai thinking level onto a Responses reasoning
+// effort. An explicit THINKING_LEVEL_UNSPECIFIED is distinct from unset and
+// still asks the model to think, so it resolves to medium.
+var reasoningEfforts = map[genai.ThinkingLevel]shared.ReasoningEffort{
+	genai.ThinkingLevelUnspecified: shared.ReasoningEffortMedium,
```
karolpiotrowicz, line 468: An explicit `THINKING_LEVEL_UNSPECIFIED` pins `effort: "medium"`, overriding whatever the model would have chosen, while `ThinkingBudget: -1` sends no effort at all and lets it choose — with a comment a few lines down saying pinning a number would be "us deciding instead". Both are genai's way of declining to choose.

Reaching it takes typing the constant, since it is not the Go zero value, so this is small. If the asymmetry is deliberate the reason would be worth a line, since the neighbouring case argues the opposite way.

model/openaimodel/request_test.go
```diff
@@ -407,6 +416,712 @@ func TestApplyGenerationConfig(t *testing.T) {
...
+			Logprobs:       genai.Ptr(int32(5)),
+			SafetySettings: []*genai.SafetySetting{{Category: genai.HarmCategoryHarassment}},
+		}, ErrSafetySettingsNotSupported},
+		{"MIME type over orphan Logprobs", &genai.GenerateContentConfig{
+			Logprobs:         genai.Ptr(int32(5)),
+			ResponseMIMEType: "text/csv",
+		}, ErrUnsupportedMIMEType},
+		{"StopSequences over ThinkingConfig", &genai.GenerateContentConfig{
+			StopSequences:  []string{"STOP"},
+			ThinkingConfig: &genai.ThinkingConfig{},
```
karolpiotrowicz: This case does not test what the others do. `&genai.ThinkingConfig{}` returns nil from `applyThinkingConfig`, so there is no competing error for `ErrStopSequencesNotSupported` to win against — the assertion holds whether or not precedence works.

The four cases above it are genuinely competitive, and the orphan-`Logprobs` ones are the hard cases. A `ThinkingConfig{ThinkingBudget: genai.Ptr(int32(-2))}` would make this one match them.

model/openaimodel/request_test.go
```diff
@@ -407,6 +416,712 @@ func TestApplyGenerationConfig(t *testing.T) {
...
+	}
+}
+
+// Guards against the bug returning as genai grows fields: every exported field
+// must be translated, covered by a named error, or in unsupportedConfigFields.
+// When this fails, add the new field to whichever of the three it belongs in —
+// not to this test alone, which would only re-hide the drop.
+func TestGenerateContentConfigFieldsAreAccountedFor(t *testing.T) {
+	// Fields applyGenerationConfig, convertTools or convertToolChoice translate.
+	translated := map[string]bool{
```
karolpiotrowicz, line 1054: `translated` and `namedError` are literals here while `rejected` is read from `unsupportedConfigFields`, so the test can only check that a field is *listed* somewhere, not that the listing is true. When genai adds a field, the cheapest way to green this is to add the name to `translated`, which is the outcome the comment above warns against and nothing prevents.

Not something I can point at a live failure for — every entry is currently backed by a behavioural test elsewhere. It is worth knowing that this guard measures coverage of the *lists*, not of the behaviour, when reasoning about what it protects.

karolpiotrowicz (dismissed):
Three of the four points from 24 August are fixed, and I checked each through the exported `GenerateContent` against a stub rather than at the struct, so these are the bodies that actually go out. A translated `ThinkingConfig` now sends `{"reasoning":{"effort":"medium"}}` where the old revision sent `{"reasoning":{"effort":"medium","summary":"concise"}}`. `IncludeThoughts: true` still gets `summary: "auto"`. `ThinkingBudget: 0` sends `effort: "none"`. An explicitly empty `ResponseModalities` is rejected.

The fourth went in and then came back out. [`dbf4ce5`](https://github.com/google/adk-go/commit/dbf4ce58a444f90d5f926d0d9ae00fb1f75ed847) and [`fe4d045`](https://github.com/google/adk-go/commit/fe4d0450b1f932209ed22e58b79544e9cbb6a9d7) both carried the paragraph saying presence is only observable where the zero value cannot itself be a setting, naming `AudioTimestamp: false`, `CachedContent: ""` and `MediaResolution: ""` as the ones that pass unremarked. That was the qualification the last review asked for. The final commit, [`438f737`](https://github.com/google/adk-go/commit/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91) "state the contract in two sentences", took 85 lines out of `doc.go` including that paragraph, and the four sentences left behind assert the opposite. I put the specifics on the `doc.go` line.

## A behavior change worth a sentence in the description

`summarizerGenConfig` forwards seven fields from the application's config into the compaction summarizer's, among them [`Seed` and `HTTPOptions`](https://github.com/google/adk-go/blob/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91/session/compaction/llm_summarizer.go#L493-L501). Five shapes that reach an OpenAI-backed summarizer through that line succeed on `main` and at the released v2.3.0, and fail on this branch: `Seed`, `HTTPOptions.APIVersion`, `HTTPOptions.BaseURL`, `HTTPOptions.RetryOptions`, and a `Timeout` of zero. `TopK`, `Labels` and `SafetySettings` already failed before this change, so the shape of the problem is not new — this widens it by five.

Driving `compaction.LLMSummarizer.SummarizeEvents` with an openaimodel summarizer:

```
                                  on main    on this branch
Seed                                    OK         FAILS
HTTPOptions.APIVersion: "v1beta"        OK         FAILS
HTTPOptions.BaseURL                     OK         FAILS
HTTPOptions.RetryOptions                OK         FAILS
HTTPOptions.Timeout: 0                  OK         FAILS
HTTPOptions.Headers                     OK         OK      (ignored by design)
Temperature                             OK         OK
TopK / Labels / SafetySettings         FAILS       FAILS   (already broken)
```

How far this reaches is narrower than it first looks, and I want to be accurate about that. The default summarizer is built from [the root agent's own model and config](https://github.com/google/adk-go/blob/438f7377ad6a15bbd69779f0d03f08d3c5dd3a91/runner/runner.go#L180-L189), so the auto-wired path is always same-provider, and there the agent's own turn fails first — which is the change you already document. Getting the summarizer-only failure takes a hand-built `compaction.NewLLMSummarizer` holding an OpenAI model alongside a Gemini agent's config. It also degrades rather than breaks: `compaction.ErrCompaction`, logged with its cause, events already committed, the run continuing uncompacted.

So I do not think it should hold this PR. `APIVersion` and `BaseURL` are ordinary Gemini transport settings though, and a reader of the description would not expect them to stop a summarizer, so it is worth naming there. The durable fix is in `summarizerGenConfig` itself, which forwards transport and sampling settings that mean nothing for a summarization — happy to file that separately, covering all eight fields rather than the five this adds.

## The new HTTPOptions rejection table is not exercised

The six predicates in `unsupportedHTTPOptionFields` have no test that runs them — the reflection test checks their names, and no test anywhere builds an `HTTPOptions` carrying `BaseURL`, `APIVersion` or the other four. Change one predicate body to `return false` and the package suite still passes while the field is dropped in silence at the exported entry point. The sibling table does not have that gap, and mirroring its test is the fix. Full detail and the reproduction are on the `request.go` line.

## Smaller things

The rest are on the lines they concern: the streaming timeout's scope in `openai.go`, the `ServiceTierUnspecified` mapping, an unranged `Logprobs`, the explicit-`UNSPECIFIED` thinking level, and two test-shape notes.

One merge-order heads-up: #1395 inserts its own paragraph at the same anchor in `doc.go` and conflicts with this branch. #1359, which shares `openai.go` and `errors.go`, merges cleanly.

Build, vet, the full suite, `-race` on the package, `staticcheck`, `govulncheck`, lint and `tidy -diff` are all clean in both modules on this head, and the repository's own apidiff check reports no incompatible changes against both the merge base and the v2.3.0 tag. The one lint finding in `plugin/agentanalytics` is identical on the base tree and untouched by this branch.

model/openaimodel/request.go
wojtas000: Changed to [`default`](https://github.com/google/adk-go/blob/ee50fc01e36eb11cc2bfe1e26c1b1b91133fa925/model/openaimodel/request.go#L408), joining `standard`. genai's own "Default service tier, which is standard" settles it, and `auto` would have handed a caller who asked for standard by name whichever tier their project had configured. Verified accepted live.

model/openaimodel/request.go
wojtas000: Fixed, and your reproduction is now the test. [Every predicate is driven with a live value](https://github.com/google/adk-go/blob/ee50fc01e36eb11cc2bfe1e26c1b1b91133fa925/model/openaimodel/request_test.go#L1763), from a [shared slice](https://github.com/google/adk-go/blob/ee50fc01e36eb11cc2bfe1e26c1b1b91133fa925/model/openaimodel/request_test.go#L1749) that a [pairing test](https://github.com/google/adk-go/blob/ee50fc01e36eb11cc2bfe1e26c1b1b91133fa925/model/openaimodel/request_test.go#L1782) reads, so a predicate added without a case fails rather than shipping untested. `{"BaseURL", func(*genai.HTTPOptions) bool { return false }}` now turns the suite red. The test had existed and my own edit deleted it, which is how the gap opened.

model/openaimodel/doc.go
wojtas000: Restored, both parts. The [presence-observability rule](https://github.com/google/adk-go/blob/ee50fc01e36eb11cc2bfe1e26c1b1b91133fa925/model/openaimodel/doc.go#L24-L31) is back and names the three shapes — `AudioTimestamp` false, `CachedContent` and `MediaResolution` empty — alongside `Headers`, and the [table](https://github.com/google/adk-go/blob/ee50fc01e36eb11cc2bfe1e26c1b1b91133fa925/model/openaimodel/doc.go#L33-L44) is back with it. The four sentences you read did assert the opposite; that commit was me over-applying a request to shorten the docs.

model/openaimodel/request.go
wojtas000: Left as is, with the [rule's boundary now stated](https://github.com/google/adk-go/blob/ee50fc01e36eb11cc2bfe1e26c1b1b91133fa925/model/openaimodel/request.go#L543-L546): the sweep is for values that would *vanish*, and an out-of-range `top_logprobs` reaches the wire and draws a named 400. Validating it locally would also impose an OpenAI range on the compatible endpoints `ClientConfig.BaseURL` exists to reach.

model/openaimodel/request.go
wojtas000: [Reason added](https://github.com/google/adk-go/blob/ee50fc01e36eb11cc2bfe1e26c1b1b91133fa925/model/openaimodel/request.go#L463-L466): adk-python maps `thinking_level_unspecified` to medium ([`_openai_responses_llm.py:276-277`](https://github.com/google/adk-python/blob/main/src/google/adk/labs/openai/_openai_responses_llm.py#L276-L277)), so the level follows that precedent while a dynamic budget has none and defers to the model.

model/openaimodel/request_test.go
wojtas000: [Uses `ThinkingBudget: -2` now](https://github.com/google/adk-go/blob/ee50fc01e36eb11cc2bfe1e26c1b1b91133fa925/model/openaimodel/request_test.go#L1034), which `applyThinkingConfig` genuinely rejects, so the case competes like the four above it.

model/openaimodel/request_test.go
wojtas000: Closed from the other side: [every field the lists call translated](https://github.com/google/adk-go/blob/ee50fc01e36eb11cc2bfe1e26c1b1b91133fa925/model/openaimodel/request_test.go#L1804) must now reach the params, so writing a name into `translated` no longer greens a field that is still dropped.

model/openaimodel/openai.go
wojtas000: Documented on [both](https://github.com/google/adk-go/blob/ee50fc01e36eb11cc2bfe1e26c1b1b91133fa925/model/openaimodel/openai.go#L128-L131) [sites](https://github.com/google/adk-go/blob/ee50fc01e36eb11cc2bfe1e26c1b1b91133fa925/model/openaimodel/request.go#L415-L423) — the streamed deadline covers the caller's time in the range body, so a slow consumer can exhaust it while the provider keeps up.

karolpiotrowicz (dismissed):
Both things that held up the last review are closed, and I checked each by breaking the code rather than by reading it. Neutering any one of the six `unsupportedHTTPOptionFields` predicates to `return false` now turns the package suite red, killed by [`TestApplyGenerationConfigRejectsGeminiShapedHTTPOptions`](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/request_test.go#L1763-L1777) — and the same edit against the previous revision leaves it green, so the new tests are what kill it rather than something that was already there. The `doc.go` paragraph is back and it is complete, not just present: the only predicates in `unsupportedConfigFields` whose zero value is indistinguishable from unset are `CachedContent`, `MediaResolution` and `AudioTimestamp`, and those are exactly the three [it names](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/doc.go#L27-L31). The tier remap, the streaming-deadline note, and the precedence case that now uses a budget the code really rejects all check out too.

What is left is four documentation and test-robustness items. None of them changes behavior and none needs to gate a merge.

## The "guarded the same way" comment is not true yet

[`request_test.go:1781`](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/request_test.go#L1779-L1781) says the sibling table is guarded the same way. It is not, and the gap is narrower than it first looks, so here is what I actually measured. Neutering each of the twelve `unsupportedConfigFields` predicates one at a time, leaving the test file alone, turns the suite red every time — all twelve are genuinely driven today and nothing is dropped. What [`TestEveryUnsupportedHTTPOptionFieldIsDriven`](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/request_test.go#L1782-L1796) adds on the other table is the guarantee that a *future* entry gets a case, and the config table has no equivalent.

The same shape keeps open the accounting-test point from last time, the one your reply said was closed from the other side. [`translated`](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/request_test.go#L1054-L1068) is a function-local literal, and [`TestTranslatedFieldsAreNotRejected`](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/request_test.go#L1804-L1846) restates its own ten cases rather than reading it, so the two can disagree — they already do, at thirteen names against ten. Writing a new genai field's name into `translated` is still the cheapest way to green the accounting test, which is what the comment above it warns against:

```
$ # drop {"Seed"} from unsupportedConfigFields and its driving case,
$ # then add "Seed": true to the translated literal
$ go test ./model/openaimodel/
ok      google.golang.org/adk/v2/model/openaimodel      0.271s

applyGenerationConfig(Seed=7)  ->  err=<nil>, field dropped
```

One case inside that test also cannot fail on the field it names, which I have put on the line.

Mirroring the pairing you already built for the HTTPOptions table covers all three.

## The adk-python precedent holds for the level and not for the budget

[`request.go:465-466`](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/request.go#L465-L466) reads "it resolves to medium as adk-python does — unlike a dynamic budget, which has no such precedent and defers to the model." The first half is right: at the adk-python revision I read ([`b0180620`](https://github.com/google/adk-python/commit/b0180620f4c2f4f4467a89c37a30f75bf849700b)), [`_openai_responses_llm.py:275-277`](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/labs/openai/_openai_responses_llm.py#L275-L277) maps `thinking_level_unspecified` to `medium`. The second half is the opposite of what that file does. A dynamic budget has a precedent there and it is also medium — [`:289-292`](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/labs/openai/_openai_responses_llm.py#L289-L292) returns `_reasoning('medium')` for any non-zero budget, and the error message a few lines up [spells it out](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/labs/openai/_openai_responses_llm.py#L286) as "or -1 for medium reasoning".

The behavior at [`request.go:522-524`](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/request.go#L522-L524) is defensible on genai's own semantics without the sibling at all, so this is one clause to correct rather than a decision to revisit. Worth saying plainly that adk-go diverges here, since the comment currently claims the opposite.

## The doc table excludes a field it then rejects

[`doc.go:36`](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/doc.go#L36) lists `HTTPOptions.Timeout` as translated, and [`:43`](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/doc.go#L43) rejects "HTTPOptions apart from Timeout and Headers" — so the table positively places `Timeout` outside the rejected set, while [`request.go:578-581`](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/request.go#L578-L581) rejects a zero one. A caller who computes a duration that happens to come out zero worked before this change and now gets an error the table says cannot happen. `MaxOutputTokens` is the same shape, listed unqualified while [a negative value is rejected](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/request.go#L557-L560).

A "rejected by value" row would cover both, and it is also where the streamed-deadline caveat belongs. That last one is my own miss from the previous round: I named `openai.go` and `requestTimeout` and you documented exactly those two, but `doc.go` is the file a caller actually reads before setting the field.

## The description contradicts itself on the tier

Inside the "Update after review" section, one paragraph says every tier has an equivalent "with `standard` reaching OpenAI as `default` and an explicit `unspecified` as `auto`", and the Billing paragraph eight lines later says "`ServiceTierUnspecified` maps to `default` rather than `auto`". [`request.go:408`](https://github.com/google/adk-go/blob/004f862915dcc61fc7d513c506ea25cfe23533e9/model/openaimodel/request.go#L408) maps it to `default`, so the first one is the leftover from before the remap. The Solution paragraph still describes a budget collapsing to "`minimal` at zero and `medium` otherwise" as well, which the update further down supersedes.

While that section is open: a config carrying `ThinkingConfig` against a non-reasoning model sent no reasoning block before this change and sends one now, which is the point of the change but is not in the "Behavior change for upgraders" list with the newly-rejected fields.

## Merge order

Four open pull requests conflict with this one, all inside `model/openaimodel`: #1373 and #1467 on `openai.go`, #1395 and #1389 on `doc.go` and `request_test.go`. This branch is the largest of the five and touches every file the others do, so landing it early costs one rebase and landing it last costs four. #1373 is approved and only touches `openai.go`, so #1373 then this one, then the rest onto the settled package, is the order with the fewest rebases.

Build, vet, the full suite on both this head and the merge result, `-race` on the package, staticcheck, govulncheck, lint and `tidy -diff` are clean in both modules, and the repository's own apidiff check reports no incompatible changes. The one lint finding in `plugin/agentanalytics` is byte-identical on the base tree and this branch changes nothing in that module.

karolpiotrowicz (approved):

PR #1394, opened 2026-08-23
fix(internal): match the copyright skip list against slash-separated paths

### LINKED ISSUE #1393: internal: TestCopyrightHeader cannot pass on Windows, the skip list is keyed on forward slashes
## What happens

`TestCopyrightHeader` cannot pass on Windows. Against a clean checkout of `main`:

```
$ go test ./internal/ -run TestCopyrightHeader -count=1
--- FAIL: TestCopyrightHeader (0.06s)
    style_test.go:78: file "internal\httprr\rr.go" does not have the copyright header
    style_test.go:78: file "internal\httprr\rr_test.go" does not have the copyright header
FAIL
FAIL	google.golang.org/adk/v2/internal	0.447s
```

`internal/httprr` is vendored from `golang.org/x/oscar`, and both `style_test.go`'s own comment and AGENTS.md say it is not subject to these checks, so the test is meant to skip it.

## Why

The skip list is keyed with forward slashes:

```go
ignore := map[string]bool{
	"internal/jsonschema": true,
	"internal/util":       true,
	"internal/httprr":     true,
	"vendor":              true,
}
```

and is then looked up with the path `filepath.Walk` supplies:

```go
if ignore[path] {
	return filepath.SkipDir
}
```

`filepath.Walk` builds paths with the OS separator, so on Windows `path` is `internal\httprr` and matches no key. All four entries are inert there. Only `internal/httprr` exists in the tree today, so it is the only one that surfaces.

CI runs on Linux, where the separator happens to match the keys, which is why this has not come up.

## Fix

Normalise before the lookup:

```go
if ignore[filepath.ToSlash(path)] {
```

`filepath.ToSlash` is a no-op on Linux and macOS, so nothing changes for CI.

Verified on windows/amd64 with go1.26.6 (the toolchain `go.mod` selects): fails as above before the change, and after it

```
ok  	google.golang.org/adk/v2/internal	0.570s
```

Happy to send that as a PR if it is wanted.

## Environment

- Windows 10 amd64, Go 1.26.6
- clean checkout of `main`, no other local changes

harshitwandhare:

- Closes: #1393

**Problem:**

`TestCopyrightHeader` cannot pass on Windows. Against a clean checkout of `main`:

```
$ go test ./internal/ -run TestCopyrightHeader -count=1
--- FAIL: TestCopyrightHeader (0.06s)
    style_test.go:78: file "internal\httprr\rr.go" does not have the copyright header
    style_test.go:78: file "internal\httprr\rr_test.go" does not have the copyright header
FAIL
FAIL	google.golang.org/adk/v2/internal	0.447s
```

`internal/httprr` is vendored from `golang.org/x/oscar`. Both `style_test.go`'s own comment and AGENTS.md say it is not subject to these checks, so the test is meant to skip it.

The skip list is keyed with forward slashes:

```go
ignore := map[string]bool{
	"internal/jsonschema": true,
	"internal/util":       true,
	"internal/httprr":     true,
	"vendor":              true,
}
```

and is looked up with the path `filepath.Walk` supplies, which uses the OS separator. On Windows that is `internal\httprr`, which matches no key, so the vendored tree is walked and the test fails on exactly the files it is meant to skip.

All four entries are inert on Windows. Only `internal/httprr` exists in the tree today, so it is the only one that currently surfaces. CI runs on Linux, where the separator happens to match the keys, which is why this has not been caught.

**Solution:**

Normalise the path before the lookup:

```go
if ignore[filepath.ToSlash(path)] {
```

`filepath.ToSlash` is a no-op on Linux and macOS, so CI behaviour is unchanged. It also keeps the skip list readable as forward-slash paths, which matches how the surrounding comments refer to these directories.

I considered keying the map with `filepath.FromSlash` at construction instead, but that would make the literal harder to read for the common case and gains nothing.

**Second commit, added after review.** @wolo-lab pointed out the identical defect five lines away in the same package, at `internal/httprr_directives_test.go:68`. `TestHTTPRecordDirectivesPartitionCassettes` keys its own `ignore` map the same way and reads it with the same OS-separator path. Fixed there too, since leaving one walk normalised and the other not, with nothing saying why, is what produces the next report of this.

That one is close to latent but not quite. `internal/httprr/rr_test.go` really is walked on Windows, because the skip fails exactly as it does here. The test survives only because that file carries no `//go:generate ... -httprecord` directive, so the loop that reads directives finds nothing and passes vacuously. The broken skip is live; only its consequence is nil today.

### Testing Plan

**Unit Tests:**

No new test is added: `TestCopyrightHeader` is itself the test, and it goes from failing to passing on Windows. There is no portable way to assert the Windows path behaviour from a Linux runner, since `filepath.Walk` produces forward slashes there and the bug cannot occur.

Before, on windows/amd64, go1.26.6:

```
--- FAIL: TestCopyrightHeader (0.06s)
    style_test.go:78: file "internal\httprr\rr.go" does not have the copyright header
    style_test.go:78: file "internal\httprr\rr_test.go" does not have the copyright header
FAIL	google.golang.org/adk/v2/internal	0.447s
```

After:

```
$ go test -race ./internal/ -run TestCopyrightHeader -count=1 -v
=== RUN   TestCopyrightHeader
--- PASS: TestCopyrightHeader (0.09s)
PASS
ok  	google.golang.org/adk/v2/internal	1.509s

$ go test -race ./internal/ -count=1
ok  	google.golang.org/adk/v2/internal	1.558s

$ go mod tidy -diff        # no output
$ golangci-lint run
0 issues.
```

Not applicable. This is a change to a repository-hygiene test and has no runtime effect.

### Checklist

### Additional context

Two other packages also fail on Windows for reasons unrelated to this change, in case they are useful to know about. I have not filed either, and neither is touched here:

- `internal/configurable/conformance/replayplugin`: `filepath.Rel` cannot relate a `C:\...\Temp\...` path to a repo on `D:\`. This one only bites when the clone and `TMPDIR` sit on different volumes, so it may be my setup rather than a repo issue.
- `internal/telemetry/functionaltest`: `TestTelemetrySchema_Workflow` shows nondeterministic span ordering (`invoke_node static_node` against `invoke_node router_node`). That may overlap with #1268.

The rest of the module is green here. `go test -race -mod=readonly -count=1 -shuffle=on work` gave 69 ok with three failing before I merged `main`, and 70 ok with two failing after (the run that includes #1396).

A third, `internal/llminternal` `TestRunLiveNoGoroutineLeak/sender_error_after_connection_loss_does_not_leak`, appeared in one full run and not the next with no change to that package, and passes `-count=3` in isolation. It looks shuffle-order dependent rather than related to anything here.


wolo-lab (dismissed):
Thank you for the contribution. The diagnosis is right and the fix is the correct one-liner.

One nit:
**The same defect is five lines away, in the same package.** It is latent rather than live but leaving one walk normalized and the other not, with nothing saying why, is the shape that produces the next report of this. Happy to take it here or as a follow-up, whichever you prefer.

harshitwandhare:
Took it here, since leaving one walk normalized and the other not is the inconsistency you are pointing at, and the fix is the same line.

`internal/httprr_directives_test.go:68`, now `ignore[filepath.ToSlash(path)]`.

I checked the latent claim rather than assuming it, and it is a little sharper than latent. `internal/httprr/rr_test.go` really is walked on Windows, because the skip fails exactly as it does in `style_test.go`. The test survives because that file carries no `//go:generate ... -httprecord` directive, so the loop that reads directives finds nothing and passes vacuously. The broken skip is live; only its consequence is currently nil. Adding a `-httprecord` directive to a vendored httprr test would surface it.

The other three keys in that map are inert here for a different reason: `internal/jsonschema`, `internal/util` and `vendor` do not exist in the tree, so only `internal/httprr` can surface either bug today.

Gates on the updated branch: `go test -race ./internal/ -count=1` ok 1.625s, `gofmt -l` empty, `go mod tidy -diff` empty, `golangci-lint run` 0 issues.

I rebuilt my commit on top of your `Merge branch 'main'` rather than force-pushing over it.

harshitwandhare:
Merged `main` in, so the branch is current with #1396. That bump moves genai, the Google API client, protobuf and the OpenAI client, so I re-ran the gates against the new versions rather than assuming the earlier run still stood: build, `go vet ./internal/`, `go mod tidy -diff`, `gofmt -l`, and `golangci-lint run` all clean, and `go test -race ./internal/ -count=1` is ok at 1.858s.

Also updated the description to cover the second commit and to correct the full-module numbers, which had gone stale.

wolo-lab (approved):
Thanks!

harshitwandhare:
Thanks for merging.

Closing off the offer in the description about the two other packages that fail here, since I have a Linux box now and can answer it properly instead of guessing.

Both pass on Linux under `go test -race -shuffle=on -count=25` with the CPUs saturated. Neither is a cross-platform bug, and CI would never see either of them:

- `internal/configurable/conformance/replayplugin`: `filepath.Rel` cannot relate a `C:` temp dir to a clone on `D:`. It needs the repo and `TMPDIR` on different volumes, which makes it closer to a local-setup edge than a repo bug.
- `internal/telemetry/functionaltest`: `TestTelemetrySchema_Workflow` span ordering. Stable on Linux even under load, so whatever exposes it is specific to Windows scheduling.

I am not filing either. The first is arguably my own setup, and the second I cannot characterise well enough yet to write a report worth your time. If a Windows contributor turns up with the same span-ordering failure, that is the one worth looking at, and I will dig into it properly if it comes up again.

harshitwandhare:
The `v1` backport is up as #1488.

It could not replay because this PR touched two files and `internal/httprr_directives_test.go` does not exist on `v1`, so half the patch had no target. #1488 carries the `internal/style_test.go` half only, byte for byte as it landed here, and the dropped hunk has nothing on that branch to apply to.

Reproduced the failure on `v1` before applying it rather than assuming this diagnosis carried over, and the test output for both sides is in the PR.

PR #1395, opened 2026-08-23
fix(model/openaimodel): stop replaying prior-turn reasoning to the API

hanorik:
**Problem**

`model/openaimodel` reports model reasoning to the caller as thought parts and
then sends those parts straight back to the API on the next turn as ordinary
assistant text. `convertContents` matched every part on `part.Text != ""` with
no guard on `part.Thought`, and nothing upstream strips thought parts. The
reasoning accumulated into the same `textParts` buffer as the real answer and
flushed with it into a single assistant message, so nothing separated the
model's private scratchpad from what it had actually said, and the following
turn was answered against it. A turn that produced only reasoning replayed as an
assistant message the model never sent.

**Solution**

Drop replayed reasoning rather than send it. The Responses API takes it back
only as an input item referencing the id of the item that produced it, and ADK
carries no such ids, so there is no correct way to replay it; adk-python's
Responses backend drops it for the same reason. Text and call/response are now
read independently rather than as arms of one switch, because a single part can
carry both — choosing between them either put the reasoning on the wire or
dropped the call and stranded its response in `callTracker`. A thought-marked
call or response therefore survives while its reasoning text does not.

Two adjacent holes close with it. A thought carrying only a signature used to
reach the `default` arm and fail the whole conversion, and is now dropped, since
a Responses request has nowhere to put one either way; and a part marked as a
thought that carries media or code is still rejected as an unsupported part
rather than vanishing from the request unannounced. A request left empty by the
drop now reports that, instead of `ErrNoContents` as though the caller had sent
nothing. No exported API changes — the only new declaration is the unexported
`replayedReasoning`.

---

**Update after review.**

Rebased onto `main`; the conflict was `newMessage`'s signature, which #1427 changed under this branch.

A part carrying an unsendable field — media, code, server-side tool traffic — is now rejected wherever it appears, because the previous check keyed on what the part *contributed*: reasoning text riding on an image matched no arm at all and the image left the request unannounced, and so did the same image riding on ordinary text or on a function call. `unsupportedPayload` runs once per part before anything is emitted and asks the opposite question — is anything set that `convertContents` does not account for — so the field is named in the error, `unsupported content part: InlineData`, whatever it travels with.

`replayedReasoning` is stated the same way rather than as a list of the fields that disqualify a part, which had omitted `ToolCall`, `ToolResponse` and `AudioTranscription` and dropped them silently; a field genai adds later is now reported rather than vanishing, and a reflection walk over `genai.Part` holds both predicates to it.

An unsupported role is reported again on a turn whose only text was a thought — nothing buffers, so `flushText` returned before `normalizeRole` ever ran.

A part that contributed nothing now says it carries nothing to send, rather than borrowing the unsupported-part wording it no longer fits.

The package doc no longer claims without qualification that reasoning is never sent back on a later turn: in a multi-agent run the shared flow rebuilds a peer's event as plain user text with the marker unset, so a peer's reasoning still arrives as ordinary content, which is not this package's to fix.

The unbounded run loop raised in review is closed on `main` by #1290 and needs nothing here — driving `runner.Run` against a stub that only ever reasons ends after 10 calls with the request body flat at 382 bytes, where before the drop it grew until the provider refused it.

A second round found three more, all of them mine rather than the original commit's.

The empty-request wrap keyed on the request having contents rather than on a part having been dropped, so a `Content` carrying no parts reported that every part was dropped when none had been, and turned the bare `ErrNoContents` into a wrapped one — `err == ErrNoContents` flipped true to false for a request that behaves exactly as it did before this branch. `convertContents` now reports whether it dropped anything and the wrap follows that, so a request that arrived empty, blank text included, still returns the bare sentinel.

The restored role check keyed on the part having text, so it ran for a thought carrying reasoning and not for one carrying only the marker or a bare signature — which `main` rejected as an unsupported part and this branch had made silent. It now runs wherever a part is dropped.

Pairing is enforced where it previously was not, which is the widest behavior change here and belongs in the note below rather than buried: text riding on a function response that pairs with nothing, or on a call with no name, used to be discarded because the text arm matched first and the call or response was never read.

A third round, from the review on `baff7e4b`, closed the two claims the description overstated and the four smaller notes.

`err == ErrNoContents` now survives a blank thought. `{Text: "   ", Thought: true}` was being wrapped: the drop is real, but the text would have been skipped downstream anyway, so the drop is not what emptied the request. The flag is set only for reasoning carrying text the model would otherwise have seen, which also gives the bare sentinel back to a bare thought and a lone signature.

The rejection for a part that reaches nothing keeps the `unsupported content part` prefix the single message used before this branch, so a caller string-matching that phrase still matches the shapes that now land there — an empty part, and one carrying only `VideoMetadata`, `MediaResolution` or `PartMetadata`.

The comment claiming a part is read field by field now says what the code does: text is read independently of a call or a response, but a call and a response on the same part remain arms of one switch and the response is dropped, exactly as on `main`.

The package doc claimed a single exception to reasoning never being replayed. There are three — a peer agent's reply folded into context, a compacted transcript, and a memory recall all render reasoning as ordinary text upstream of any model package — so it now says that instead of naming one.

**Behavioral changes.** A 16-shape corpus was run through the public API against `main` and this branch and diffed outcome-by-outcome; 11 shapes differ. Nothing that previously reached the OpenAI API stops reaching it, and no shape that succeeded on the wire now fails.

| Input shape | `main` | this PR |
|---|---|---|
| `{Text:"scratch", Thought:true}` alone | **succeeded** — the reasoning was sent (the bug) | `ErrNoContents`, wrapped, naming the drop |
| unsendable field (`FileData`, `ToolCall`, `AudioTranscription`, `CodeExecutionResult`, `InlineData`, `ExecutableCode`, `ToolResponse`) riding on text or a call | **succeeded**, field silently discarded | error naming the field |
| bare `{Thought:true}`, lone `ThoughtSignature` | `unsupported content part` | bare `ErrNoContents` |
| empty part, or only `VideoMetadata` / `MediaResolution` / `PartMetadata` | `unsupported content part *genai.Part` | `unsupported content part: carries nothing to send` |
| everything else in the corpus, incl. blank text, a call and response on one part, a signature or `VideoMetadata` riding on text | — | unchanged |

`errors.Is(err, ErrNoContents)` holds throughout. `err == ErrNoContents` holds for every request that was empty on arrival; it is false only where the drop suppressed text that would otherwise have been sent, a shape that returned no error at all on `main`. No exported identifier changes.

**Known gaps, deliberately not fixed here.**

- A call and a response on the *same* part still emit the call and drop the response silently. Identical to `main`, nothing in the tree builds that part; the comment no longer implies otherwise.
- The two rejection messages carry no sentinel, so a caller can only string-match them. `main` had none either. An `ErrUnsupportedContentPart` sentinel would be an exported-API addition and wants a maintainer's call first.
- `ConvertForeignEvent` (`internal/llminternal`) rebuilds a peer's reasoning as plain user text, so it still reaches every model package. Out of scope here and worth its own issue, as the review notes.
- `flattenContentText` (the system-instruction path) has no thought filter. No in-repo caller can put a marked thought part there; noted as defense-in-depth rather than a live leak.

**Testing Plan.**

- `golangci-lint run` and `go mod tidy -diff` — clean in both modules
- Sixteen guards reverted individually and the suite re-run; every one is caught by a named test. The three added this round: resetting the drop flag per content block, and removing the blank-text skip from `newMessage` and from `newOutputMessage`.
- Live OpenAI e2e suite (`adk-openai-e2e`), which is the only evidence that exists for this package — it ships no httprr cassettes, so its unit tests assert serialization against expectation and nothing in CI proves OpenAI accepts the payload:

```
adk-go under test: pr-1395 / a3de3b50
openai-go:         v3.54.0
model:             gpt-5.4-nano
29 scenarios PASS, 0 FAIL          Full suite cost: $0.00152
```

  Scenario 16 is held out: it is written against the unmerged #1392 and does not compile against `main` either. The same suite on `main` (`a6114727`) is likewise 29/0, so this branch is at parity.

**Review loop.** Three rounds of a six-lens panel (security, api-compat, correctness, testing, idioms, parity). Round 1: one blocking correctness finding (an unsendable field riding on text or a call escaped the rejection) and two blocking testing gaps. Round 2: four testing gaps and three over-length comments; one blocking correctness claim — that a dropped `ThoughtSignature` should wrap the sentinel — was disputed and adjudicated in favour of the current rule by an independent pass, on the grounds that a signature has nowhere to go in a Responses request under any circumstance. Round 3: clean.


karolpiotrowicz:
The drop is the right call, and the conversion holds where it matters: no part marked as a thought contributes text to any emitted item, on any role and in any position, and a thought-marked call or response still survives with its pairing intact. Exercising it turned up one consequence worth a decision before this lands, and three smaller things after it.

**The run loop stops terminating on a reasoning-only turn.** [`base_flow.go:120`](https://github.com/google/adk-go/blob/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e/internal/llminternal/base_flow.go#L120) deliberately does not end the run when the model's last turn is all thought parts — it calls the model again — and nothing bounds that path. `MaxLLMCalls` occurs exactly twice in the tree, as a field at [`agent/live.go:48`](https://github.com/google/adk-go/blob/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e/agent/live.go#L48) and an assignment at [`runtime.go:304`](https://github.com/google/adk-go/blob/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e/server/adkrest/controllers/runtime.go#L304), and is never read anywhere. Until now each retry replayed the previous turn's reasoning, so the request grew and a real endpoint eventually refused it. With [`request.go:128`](https://github.com/google/adk-go/blob/9dd6a6f4fee57e074261ecbf01f0f144a751be3f/model/openaimodel/request.go#L128) dropping those parts, the request is byte-identical every iteration, so nothing ends the run at all.

Driving `runner.Run` against a stub that always answers with reasoning only, and rejects an over-budget body with `context_length_exceeded` the way the real API does:

```
main    0a51e15   model calls: 6
                  request body sizes: [196 487 592 697 802 907]
                  outcome: TERMINATED BY PROVIDER ERROR: 400 context_length_exceeded

branch  9dd6a6f   model calls: 40
                  request body sizes: [196 382 382 382 ... 382]
                  outcome: STOPPED BY TEST CAP - did not terminate on its own
```

<details>
<summary>the test those two runs come from — drop it in <code>model/openaimodel</code> and run <code>go test -run TestBackstop -v ./model/openaimodel/</code></summary>

```go
package openaimodel_test

import (
	"context"
	"io"
	"net/http"
	"net/http/httptest"
	"testing"
	"time"

	"google.golang.org/genai"

	"google.golang.org/adk/v2/model/openaimodel"
)

const reasoningOnly = `{
  "id":"resp_1","object":"response","created_at":1,"status":"incomplete","model":"test-model",
  "incomplete_details":{"reason":"max_output_tokens"},
  "output":[{"type":"reasoning","id":"rs_1","summary":[{"type":"summary_text","text":"still thinking about it"}]}],
  "usage":{"input_tokens":1,"output_tokens":1,"total_tokens":2,"output_tokens_details":{"reasoning_tokens":1}}
}`

func TestBackstop(t *testing.T) {
	const budget = 900 // bytes of request body, standing in for a token budget

	var sizes []int
		b, _ := io.ReadAll(r.Body)
		sizes = append(sizes, len(b))
		if len(b) > budget {
			w.WriteHeader(http.StatusBadRequest)
			io.WriteString(w, `{"error":{"message":"maximum context length exceeded","type":"invalid_request_error","code":"context_length_exceeded"}}`)
			return
		}
		w.Header().Set("Content-Type", "application/json")
		io.WriteString(w, reasoningOnly)
	}))
	defer srv.Close()

	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
	defer cancel()

	m, _ := openaimodel.NewModel(ctx, "test-model", &openaimodel.ClientConfig{APIKey: "test", BaseURL: srv.URL})
	a, _ := llmagent.New(llmagent.Config{Name: "assistant", Model: m, Instruction: "be helpful"})
	r, _ := runner.New(runner.Config{AppName: "backstop", Agent: a,
		SessionService: session.InMemoryService(), AutoCreateSession: true})

	events, outcome := 0, ""
	for _, err := range r.Run(ctx, "u", "s", genai.NewContentFromText("q", genai.RoleUser), agent.RunConfig{}) {
		if err != nil {
			outcome = "TERMINATED BY PROVIDER ERROR: " + err.Error()
			break
		}
		if events++; events >= 40 {
			outcome = "STOPPED BY TEST CAP - did not terminate on its own"
			break
		}
	}
	t.Logf("model calls: %d", len(sizes))
	t.Logf("request body sizes: %v", sizes)
	t.Logf("outcome: %s", outcome)
}
```
</details>

The bound this removes was the bug's own symptom, so preserving it is not an option and the drop is not what is wrong here. It is worth saying plainly how narrow the trigger is, because it decides how urgent the cap is. The turn has to consist only of thought parts, and on the default configuration that cannot happen: `openaimodel` never sets `Reasoning` on the request, so a real `gpt-5-nano` call truncated by `max_output_tokens` returns one reasoning item with empty `content` and empty `summary`, [`response.go:107`](https://github.com/google/adk-go/blob/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e/model/openaimodel/response.go#L107) returns `ErrNoTextOrToolContent`, and the run ends after a single call. The same call with `"reasoning":{"summary":"auto"}` returns two summary entries and no message item, which [`response.go:100`](https://github.com/google/adk-go/blob/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e/model/openaimodel/response.go#L100) maps to all-thought parts and does spin. That is reachable today only by passing the option through the escape hatch at [`openai.go:70`](https://github.com/google/adk-go/blob/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e/model/openaimodel/openai.go#L70).

So the missing cap belongs in [`base_flow.go:120`](https://github.com/google/adk-go/blob/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e/internal/llminternal/base_flow.go#L120) rather than in this package, and I would land this first and track the cap separately rather than hold a confidentiality fix behind it.

**A thought-marked part that also carries media or code vanishes without the error.** The rejection arm at [`request.go:152`](https://github.com/google/adk-go/blob/9dd6a6f4fee57e074261ecbf01f0f144a751be3f/model/openaimodel/request.go#L152) is gated on `part.Text == ""`, so a part with reasoning text riding on an image matches no arm at all — the text is suppressed by [`request.go:128`](https://github.com/google/adk-go/blob/9dd6a6f4fee57e074261ecbf01f0f144a751be3f/model/openaimodel/request.go#L128), the call and response arms do not apply, and `part.Text == ""` is false. The comment at [`request.go:188-189`](https://github.com/google/adk-go/blob/9dd6a6f4fee57e074261ecbf01f0f144a751be3f/model/openaimodel/request.go#L188-L189) says such a part goes on "to be converted, or rejected as unsupported, on their merits", which holds only while the part has no text, and [`request_test.go:304`](https://github.com/google/adk-go/blob/9dd6a6f4fee57e074261ecbf01f0f144a751be3f/model/openaimodel/request_test.go#L304) covers just that variant. The media is dropped on `main` too, so nothing regresses — what is new is the promise to reject it.

```go
{Thought: true, Text: "scratch", InlineData: &genai.Blob{MIMEType: "image/png", Data: []byte{1}}}
// no error, one input item emitted, the image gone. Same for FileData, ExecutableCode, CodeExecutionResult.

{Thought: true, InlineData: &genai.Blob{MIMEType: "image/png", Data: []byte{1}}}
// openai: unsupported content part *genai.Part   <- the only variant the table covers
```

**`replayedReasoning` does not cover every `genai.Part` field, so three more shapes drop silently.** [`request.go:190-192`](https://github.com/google/adk-go/blob/9dd6a6f4fee57e074261ecbf01f0f144a751be3f/model/openaimodel/request.go#L190-L192) names six fields and omits `ToolCall`, `ToolResponse` and `AudioTranscription`, along with `VideoMetadata`, `MediaResolution` and `PartMetadata`, which carry nothing sendable on their own. A part marked as a thought carrying any of the first three returns `true` from the predicate and disappears, where `main` returned `openai: unsupported content part`. Nothing in the repo populates those fields today, so this is about the next genai bump rather than about now. Inverting the polarity, so the predicate is true only when text and signature are the only things set, would make a newly added field fail loudly by default instead of quietly.

**An invalid role stops being reported when a turn's only text is a thought.** Nothing buffers, so `flushText` returns before `normalizeRole` ever runs.

```go
{Role: "assistant", Parts: []*genai.Part{{Text: "x"}}}
// main and this branch: openai: unsupported role "assistant"

{Role: "assistant", Parts: []*genai.Part{{Text: "x", Thought: true}}}
// main:        openai: unsupported role "assistant"
// this branch: nil
```

**The package doc promises more than the package can deliver.** [`doc.go:24-29`](https://github.com/google/adk-go/blob/9dd6a6f4fee57e074261ecbf01f0f144a751be3f/model/openaimodel/doc.go#L24-L29) says reasoning "is not sent back on a later turn", unqualified. In a multi-agent run a peer's event goes through `ConvertForeignEvent` at [`contents_processor.go:108`](https://github.com/google/adk-go/blob/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e/internal/llminternal/contents_processor.go#L108), which rebuilds each text part as a brand-new `&genai.Part{Text: fmt.Sprintf("[%s] said: %s", ...)}` at [`contents_processor.go:580-582`](https://github.com/google/adk-go/blob/0a51e153ad5a72ad8a003c5c0ef8d8e98f88139e/internal/llminternal/contents_processor.go#L580-L582) with `Thought` left at its zero value, so the guard passes it straight through as user text. That rebuild is shared flow code and out of scope here — one qualifying clause in the paragraph would keep the doc true in the meantime.

wojtas000:
Thanks — all four landed, and the fifth needed nothing. Rebased onto `main` (the conflict was `newMessage`'s signature, changed by #1427) and pushed as three commits on top of the original.

- **The missing cap** — closed on `main` by #1290 ([`base_flow.go:120`](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/internal/llminternal/base_flow.go#L106-L120)), and your backstop test on the rebased branch now reports `model calls: 10`, sizes `[196 382 382 …]`, terminating on its own.
- **Thought-marked media** — the rejection keyed on what the part *contributed*, not just on `part.Text == ""`, so an image on ordinary text or on a call vanished too (on `main` as well); [`unsupportedPayload`](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request.go#L229-L257) now [runs once per part](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request.go#L143-L149) before anything is emitted and names the field whatever it travels with.
- **Uncovered `genai.Part` fields** — inverted as you suggested, so `ToolCall`, `ToolResponse`, `AudioTranscription` and anything genai adds later are reported rather than dropped, with a [reflection walk](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request_test.go#L804-L842) holding [both predicates](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request_test.go#L844-L864) to it.
- **The unreported role** — [the role is now checked](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request.go#L150-L163) wherever text is dropped, so a thought-only turn returns `unsupported role "assistant"` exactly as it does with the marker off.
- **The doc** — [qualified in place](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/doc.go#L24-L33): a peer's reasoning still arrives as ordinary content, and the paragraph now says so and says it is not this package's to fix.

Two more fell out of the rest of the review:

- A part that contributed nothing now says it [carries nothing to send](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request.go#L185-L190), rather than borrowing the unsupported-part wording it no longer fits.
- [`describeInput`](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request_test.go#L277-L320) rendered an assistant input message and an output message identically, so the table was blind to a replayed turn going out as `input_text`; they render as `in/` and `out/` now.

Verified: each of the seven guards reverted individually and the suite re-run, every one caught by a named test; gates green in both modules; and exercised live on `gpt-5.4-nano` through the public surface, [blocking and streaming](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/openai_test.go#L465-L531), where the failing scenarios are identical to `main`'s (the unrelated `ThinkingConfig` gap #1392 covers).

One behavior change worth your eye, also in the PR body: a part carrying an unsendable field *alongside* sendable text or a call now errors instead of building a request with that field silently discarded — it never reached the API either way, but a caller who was getting a 200 now gets an error.

wojtas000:
Second pass over my own follow-up caught three more, all introduced by it rather than by the original commit. Pushed as `e5fa75d`.

- **Undisclosed break: unpaired calls and responses** — text riding on a [`FunctionResponse`](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request.go#L175-L184) that pairs with nothing, or on a nameless [`FunctionCall`](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request.go#L165-L174), used to be discarded because the text arm matched first, and now errors; that is the same independent read of a part's fields that lets a well-formed call riding on text survive, so it stays, but it is now disclosed in the body and covered by two cases ([orphan](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request_test.go#L576-L587), [nameless](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request_test.go#L590-L599)).
- **Wrong message and sentinel identity** — [the empty-request wrap](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request.go#L54-L64) keyed on the request having contents rather than on a part having been dropped, so `Contents: [{Role:"user"}]` claimed every part was dropped when none had been and flipped `err == ErrNoContents` from true to false; [`convertContents`](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request.go#L97-L100) now reports the drop and the wrap follows it, so a request empty on arrival [keeps the bare sentinel](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request_test.go#L646-L688).
- **Role check gated too narrowly** — it ran only when the part had text, so a bare thought or a lone signature on an invalid role passed silently where `main` had rejected it; it [now runs wherever a part is dropped](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request.go#L150-L163), covered both ways ([thought](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request_test.go#L510-L517), [signature](https://github.com/google/adk-go/blob/e5fa75d06f8ee8ab608c813bca4e0c85f313aa47/model/openaimodel/request_test.go#L520-L527)).

Gates green in both modules, the ten guards each reverted individually and caught by a named test, and the live suite's failures are identical to `main`'s (the unrelated `ThinkingConfig` gap #1392 covers).

model/openaimodel/request.go
```diff
@@ -122,42 +136,125 @@ func convertContents(contents []*genai.Content) (responses.ResponseInputParam, e
...
-			case part == nil:
+			if part == nil {
 				continue
-			case part.Text != "":
+			}
+			// A part is read field by field rather than as one choice among
+			// alternatives, because it can carry several at once. Anything
+			// this package cannot send is reported first and on its own, so
+			// that something sendable elsewhere on the same part — text, or a
+			// function call — cannot carry it out of the request unannounced.
```
karolpiotrowicz: This overshoots by one field pair. A part is now genuinely read field by field for text against call, but call and response are still arms of one switch below, so a part carrying both emits the call and drops the response with no error — `unsupportedPayload` cannot catch it either, since both fields are zeroed. Same behavior on `main`, and nothing in the tree builds that part, so nothing regresses. The comment is new though, and it reads as a guarantee the switch does not make.

model/openaimodel/request.go
```diff
@@ -122,42 +136,125 @@ func convertContents(contents []*genai.Content) (responses.ResponseInputParam, e
...
 				}
 				items = append(items, responses.ResponseInputItemUnionParam{OfFunctionCallOutput: respParam})
-			default:
-				return nil, fmt.Errorf("openai: unsupported content part %T", part)
+			case !sendText && !replayedReasoning(part):
+				// The part put no text in the buffer, carries no call or
+				// response, and is not reasoning worth dropping: nothing in it
+				// reaches the request, so say so rather than let an empty part
+				// pass for a converted one.
+				return nil, false, errors.New("openai: content part carries nothing to send")
```
karolpiotrowicz: Neither this nor `unsupported content part: %s` above carries a sentinel, so a caller can only string-match. That matches the rest of the file, so it is not a new inconsistency — but it does mean the shapes that now land here (an empty part, a part carrying only `VideoMetadata` or `PartMetadata`) used to say `unsupported content part` and no longer do, so a prefix match on that phrase stops covering them.

model/openaimodel/request_test.go
```diff
@@ -261,9 +265,626 @@ func TestBuildOpenAIParams_UnsupportedPart(t *testing.T) {
...
+			// then skipped as blank, exactly as it is without this change.
+			"only_blank_text",
+			[]*genai.Content{{Role: string(genai.RoleModel), Parts: []*genai.Part{{Text: "   "}}}},
+			true,
+		},
+		{
+			"only_reasoning",
+			[]*genai.Content{{Role: string(genai.RoleModel), Parts: []*genai.Part{{Text: "scratch", Thought: true}}}},
+			false,
+		},
```
karolpiotrowicz, line 717: This table is one case short of the rule it pins. `only_blank_text` and `only_reasoning` are here, but not the two together — `{Text: "   ", Thought: true}`, where `main` returns the bare sentinel and this branch returns the wrapped one. That is the shape the description's "including blank/whitespace-only text" claim turns on.

model/openaimodel/request_test.go
```diff
@@ -261,9 +265,626 @@ func TestBuildOpenAIParams_UnsupportedPart(t *testing.T) {
...
+	partType := reflect.TypeOf(genai.Part{})
+	for i := range partType.NumField() {
+		field := partType.Field(i)
+		t.Run(field.Name, func(t *testing.T) {
+			if !field.IsExported() {
+				t.Fatalf("genai.Part gained unexported field %s, which this walk cannot set; "+
+					"check by hand whether unsupportedPayload should report it", field.Name)
+			}
+			part := &genai.Part{}
+			reflect.ValueOf(part).Elem().Field(i).Set(nonZero(t, field.Type))
```
karolpiotrowicz, line 890: The walk sets each field on an otherwise empty part. That is sound as a test of the predicate, since `unsupportedPayload` zeroes the accounted set before reflecting and so gives the same answer whatever else rides along. It is worth noting that the shape this PR is fixing is the riding one, and only `InlineData` and `ExecutableCode` are pinned that way by hand. I checked `FileData`, `ToolCall`, `AudioTranscription` and `CodeExecutionResult` riding on text or on a call and all four are rejected by name, so this is a gap in what the tests would catch later rather than a gap in the behavior.

karolpiotrowicz (dismissed):
All five landed, and I re-ran the one that mattered most rather than taking it on trust. Driving `runner.Run` against a stub that only ever reasons, with the same body-size budget as before: this branch makes 10 model calls with the request body flat at 382 bytes and ends on its own, where `main` before the drop grew the body every turn until the stub refused it. The bound is `maxConsecutiveThoughtOnlyTurns` at [`base_flow.go:120`](https://github.com/google/adk-go/blob/baff7e4ba8d47aef3f834aa4e294c9a420e4a099/internal/llminternal/base_flow.go#L120), so the cap question is closed on `main` and needs nothing here.

<details><summary>the two runs, from the same test I posted in August</summary>

```
before the fix (main)   model calls: 6
                        request body sizes: [196 509 636 763 890 1017]
                        outcome: TERMINATED BY ERROR: 400 context_length_exceeded

with the PR            model calls: 10
                        request body sizes: [196 382 382 382 382 382 382 382 382 382]
                        outcome: RAN TO COMPLETION ON ITS OWN
```
</details>

The other four check out against a run rather than a reading. I built ~25 part shapes and put each through `buildOpenAIParams` at both revisions: an image or a `ToolCall` or an `AudioTranscription` riding on text now errors naming the field where `main` dropped it silently, a bare thought or a lone signature on an invalid role now reports the role where `main` reported an unsupported part, and `go doc -all` differs from `main` by exactly the qualified paragraph. Two shapes are worth calling out because they show the inversion earning its keep beyond what I asked for: a part carrying reasoning text, a call and an image had **both** the image and the call dropped on `main`, and a well-formed call riding on text used to be discarded, which then orphaned its own response and failed the whole conversion — that shape now converts correctly.

Nothing here blocks. Two sentences in the description claim more than the code does, and both are a clause away from being true.

**`err == ErrNoContents` does not survive a blank thought.** The description says the identity holds "for every request that was empty on arrival, including blank/whitespace-only text". For `{Text: "   ", Thought: true}` it does not: `main` returns the bare sentinel, this branch returns the wrapped one. The code is behaving exactly as [`request.go:54-64`](https://github.com/google/adk-go/blob/baff7e4ba8d47aef3f834aa4e294c9a420e4a099/model/openaimodel/request.go#L54-L64) says it will, since a part genuinely was dropped as reasoning — it is the sentence that is wrong, and the fix is to drop "including blank/whitespace-only text" or to trim before setting the flag.

```go
&model.LLMRequest{Contents: []*genai.Content{modelTurn(&genai.Part{Text: "   ", Thought: true})}}
// main:        err == ErrNoContents  true
// this branch: err == ErrNoContents  false   (errors.Is still true)
```

**"Reported by name, never silently discarded" is not true of the four accounted qualifier fields.** `ThoughtSignature`, `VideoMetadata`, `MediaResolution` and `PartMetadata` are zeroed at [`request.go:242-249`](https://github.com/google/adk-go/blob/baff7e4ba8d47aef3f834aa4e294c9a420e4a099/model/openaimodel/request.go#L242-L249), so a part carrying one alongside text is converted with the field gone, and a part carrying one alone gets `content part carries nothing to send` — which names nothing and is not quite true either, since the part does carry something. Both are identical to `main`, so nothing regresses. I would narrow the claim to fields outside the accounted set rather than change the code, since three of those four genuinely have nowhere to go in a Responses item.

Four smaller things are on the lines they belong to, all of them non-blocking.

One thing outside this PR that the doc paragraph is right to flag: `ConvertForeignEvent` at [`contents_processor.go:596-599`](https://github.com/google/adk-go/blob/baff7e4ba8d47aef3f834aa4e294c9a420e4a099/internal/llminternal/contents_processor.go#L596-L599) rebuilds a peer's text part with the marker unset, so a sub-agent's reasoning still reaches every model package as ordinary user text. Worth its own issue against `internal/llminternal` rather than holding this.

Gates are green on both modules apart from `telemetry`, which fails the same way on `main` in the same sandbox because it dials the metadata server. No exported surface moves.

model/openaimodel/request.go
wojtas000: Fixed the comment rather than the switch, since as you say nothing regresses and nothing in the tree builds that part. It [now says text is read independently of a call or a response, and that a call and a response on the same part remain alternatives with the response dropped](https://github.com/google/adk-go/blob/a3de3b5088c12e29f500461518e6e3e07025ef12/model/openaimodel/request.go#L151-L166) — the guarantee the code actually makes. Noted in the body under known gaps.

model/openaimodel/request.go
wojtas000: Fixed by keeping the prefix instead of adding a sentinel: the arm [now returns `unsupported content part: carries nothing to send`](https://github.com/google/adk-go/blob/a3de3b5088c12e29f500461518e6e3e07025ef12/model/openaimodel/request.go#L188-L192), so a prefix match on the phrase covers the empty part and the metadata-only ones again. A real `ErrUnsupportedContentPart` would be an exported-API addition — happy to add it if you want it, but it wanted your call first, so it is listed as a follow-up.

model/openaimodel/request_test.go
wojtas000: Fixed in the code, not the description — you were right that trimming was the better of the two. The flag is [set only when the dropped reasoning carried text the model would otherwise have seen](https://github.com/google/adk-go/blob/a3de3b5088c12e29f500461518e6e3e07025ef12/model/openaimodel/request.go#L151-L166), so `{Text: "   ", Thought: true}` keeps the bare sentinel, and so do a bare thought and a lone signature. [The table](https://github.com/google/adk-go/blob/a3de3b5088c12e29f500461518e6e3e07025ef12/model/openaimodel/request_test.go#L660-L757) gains those three plus two that pin the flag accumulating rather than tracking the last part.

model/openaimodel/request_test.go
wojtas000: Fixed — the walk [now runs each field through `ridingShapes`](https://github.com/google/adk-go/blob/a3de3b5088c12e29f500461518e6e3e07025ef12/model/openaimodel/request_test.go#L871-L915): [alone, on text, on reasoning text, and on a function call](https://github.com/google/adk-go/blob/a3de3b5088c12e29f500461518e6e3e07025ef12/model/openaimodel/request_test.go#L917-L943), so the riding shape is covered generically rather than for the two that were pinned by hand. The assertion is `HasSuffix` rather than `Contains`, so a field name that prefixes a longer one cannot pass.

wojtas000:
Both description claims fixed, and the four line notes are answered on their threads.

- **`err == ErrNoContents` and the blank thought** — took the trim rather than the wording. The flag is [set only when the drop suppressed text the model would otherwise have seen](https://github.com/google/adk-go/blob/a3de3b5088c12e29f500461518e6e3e07025ef12/model/openaimodel/request.go#L151-L166), so `{Text: "   ", Thought: true}` keeps the bare sentinel; a bare thought and a lone signature now do too, which `main` answered with `unsupported content part`. [The identity table](https://github.com/google/adk-go/blob/a3de3b5088c12e29f500461518e6e3e07025ef12/model/openaimodel/request_test.go#L660-L757) pins all of it.
- **"Reported by name" and the four qualifier fields** — narrowed the claim as you suggested rather than touching the code, since three of the four genuinely have nowhere to go. The [wrap](https://github.com/google/adk-go/blob/a3de3b5088c12e29f500461518e6e3e07025ef12/model/openaimodel/request.go#L54-L64) message no longer hedges about blank text either, because it can now only mean the one thing.
- **`ConvertForeignEvent`** — agreed, its own issue. Your paragraph also undersold it: compaction renders `"<author> (thought): …"` into the summarizer transcript, and memory ingests thought text unfiltered, so [the doc paragraph](https://github.com/google/adk-go/blob/a3de3b5088c12e29f500461518e6e3e07025ef12/model/openaimodel/doc.go#L24-L32) now says three routes render reasoning as ordinary text upstream instead of naming one.

Since `baff7e4b`: three rounds of the six-lens panel. Round 1 found the rejection still escapable when an unsendable field rode on text or a call; rounds 2 and 3 were tests and comment length. One lens wanted a dropped `ThoughtSignature` to wrap the sentinel — I kept the bare one, on the grounds that a signature has nowhere to go in a Responses request under any circumstance, so the drop deprives the model of nothing; an independent pass agreed.

Gates green in both modules.

karolpiotrowicz (dismissed):
Both description claims from the last round are closed, and I re-ran the one that mattered: `{Text: "   ", Thought: true}` now returns the bare `ErrNoContents` on this branch exactly as it does on `main`, and so do a bare thought and a lone signature. Taking the trim over the wording was the better of the two options I offered, since it fixes the behavior instead of documenting it. The four line notes are answered on their threads. adk-python's Responses backend does drop replayed reasoning for the reason the description gives, at [`_openai_responses_llm.py:307-323`](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/labs/openai/_openai_responses_llm.py#L307-L323), dispatched from `:436-438`. Nothing here blocks.

Two sentences in the description are contradicted by measurement. Both are a clause away from being true and neither needs a code change.

**"No shape that succeeded on the wire now fails"** is falsified by the table two rows above it. Putting the corpus through the conversion at the merge base and at this branch, three shapes move from a built request to a hard error: an `InlineData` blob riding on plain text, a `ToolCall` riding on text, and text riding on a `FunctionResponse` that pairs with nothing. Each of those returned a 200 on `main` with the field quietly discarded. The table already says so for the first two — "**succeeded**, field silently discarded" becomes "error naming the field" — and then the prose says the opposite. Narrowing it to shapes whose payload actually reached the provider would keep the claim you can defend.

**"Sixteen guards reverted individually ... every one is caught by a named test"** does not hold for one of them. Deleting `part.Text != "" || ` from [`request.go:155`](https://github.com/google/adk-go/blob/cac6f213d4ae641a86a18f820250fe61f746ccb9/model/openaimodel/request.go#L155) leaves `go test ./model/openaimodel/` green, and changes what a caller sees through the public surface:

```go
	Role: "assistant",
	Parts: []*genai.Part{{
		Thought:      true,
		Text:         "scratch",
		FunctionCall: &genai.FunctionCall{Name: "lookup", ID: "c1"},
	}},
}}}
for _, err := range llm.GenerateContent(ctx, req, false) { ... }

// this branch:   err = openai: unsupported role "assistant"   provider calls: 0
// guard deleted: err = <nil>                                  provider calls: 1
```

With the guard gone the request reaches the provider under a role the package cannot send. `main` errors on that shape too, so this is a coverage hole rather than a regression — but it is the line that keeps the role check honest when reasoning text rides on a call, which is the case the round-2 fix was for. The four role cases at [`request_test.go:503-542`](https://github.com/google/adk-go/blob/cac6f213d4ae641a86a18f820250fe61f746ccb9/model/openaimodel/request_test.go#L503-L542) all use parts carrying nothing but reasoning, and [`thought_text_riding_on_a_call`](https://github.com/google/adk-go/blob/cac6f213d4ae641a86a18f820250fe61f746ccb9/model/openaimodel/request_test.go#L410-L425) uses `modelTurn`, so nothing exercises the disjunct. The shape is unreachable from ADK itself, which only ever builds `user` and `model` roles, so this is about the claim rather than about a user hitting it.

One small thing, and I would leave it alone if the churn is not worth it. The wrapped message at [`request.go:56-61`](https://github.com/google/adk-go/blob/cac6f213d4ae641a86a18f820250fe61f746ccb9/model/openaimodel/request.go#L56-L61) says "every part was dropped as replayed reasoning" for a request where only some part was: `[{Text: "scratch", Thought: true}, {Text: "   "}]` gets the wrap, and the blank sibling was buffered and then skipped downstream, not dropped as reasoning. Diagnostic wording only, and rewording it costs two pinned test expectations.

One thing to leave alone. The two guards in `replayedReasoning` that look unobservable — the signature-on-an-answer early return, and the `unsupportedPayload` conjunct — are reachable only through the predicate's own tests, but removing either breaks `TestReplayedReasoning_EveryUnaccountedFieldDisqualifies`, which is the executable form of the forward-compatibility rule we settled last round.

Gates are green on both modules, the merge into current `main` is clean, and no exported surface moves. Two open PRs of yours, #1392 and #1389, conflict with this one on the same three files — whichever lands first, the other needs a rebase.

karolpiotrowicz (approved):
Nothing blocks. I re-checked the branch merged with current `main` ([d64e113](https://github.com/google/adk-go/commit/d64e11317e75d05e604dc21b58fcae21be0b485d)), and the drop still holds there: no reasoning text reaches any request body, blocking or streaming, and a function call with its output, an answer and a refusal all replay intact on the turns after them.

The fix matters more now than when this was opened. At the time `openaimodel` never asked for reasoning, so a history carrying thought parts needed the raw-options escape hatch. Since [#1392](https://github.com/google/adk-go/pull/1392), `ThinkingConfig{IncludeThoughts: true}` sends `reasoning.summary=auto` ([`request.go:536-537`](https://github.com/google/adk-go/blob/d64e11317e75d05e604dc21b58fcae21be0b485d/model/openaimodel/request.go#L536-L537)), the summaries come back as thought parts ([`response.go:184`](https://github.com/google/adk-go/blob/d64e11317e75d05e604dc21b58fcae21be0b485d/model/openaimodel/response.go#L184)), and on `main` the next turn sends them to the API as assistant text. Two turns through `runner.Run` against a stub show it:

```
main without this PR:   turn 2 replays the scratchpad: true
main with this PR:      turn 2 replays the scratchpad: false
```

<details>
<summary>the test — drop it in <code>model/openaimodel</code> and run <code>go test -run TestThinkingSummaryReplay -v ./model/openaimodel/</code></summary>

```go
package openaimodel_test

import (
	"context"
	"io"
	"net/http"
	"net/http/httptest"
	"strings"
	"testing"

	"google.golang.org/genai"

	"google.golang.org/adk/v2/model/openaimodel"
)

func TestThinkingSummaryReplay(t *testing.T) {
	const turn = `{"id":"resp_1","object":"response","created_at":1,"status":"completed","model":"m",
	  "output":[{"type":"reasoning","id":"rs_1","summary":[{"type":"summary_text","text":"PRIVATE-SCRATCHPAD"}]},
	            {"type":"message","id":"msg_1","role":"assistant","status":"completed",
	             "content":[{"type":"output_text","text":"4","annotations":[]}]}],
	  "usage":{"input_tokens":1,"output_tokens":1,"total_tokens":2}}`
	var bodies []string
		b, _ := io.ReadAll(r.Body)
		bodies = append(bodies, string(b))
		w.Header().Set("Content-Type", "application/json")
		io.WriteString(w, turn)
	}))
	defer srv.Close()

	ctx := context.Background()
	m, _ := openaimodel.NewModel(ctx, "m", &openaimodel.ClientConfig{APIKey: "k", BaseURL: srv.URL})
	a, _ := llmagent.New(llmagent.Config{Name: "assistant", Model: m,
		GenerateContentConfig: &genai.GenerateContentConfig{
			ThinkingConfig: &genai.ThinkingConfig{IncludeThoughts: true}}})
	r, _ := runner.New(runner.Config{AppName: "app", Agent: a,
		SessionService: session.InMemoryService(), AutoCreateSession: true})

	for _, q := range []string{"what is 2+2?", "and 3+3?"} {
		for _, err := range r.Run(ctx, "u", "s", genai.NewContentFromText(q, genai.RoleUser), agent.RunConfig{}) {
			if err != nil {
				t.Fatal(err)
			}
		}
	}
	t.Logf("turn 1 asks for summaries: %v", strings.Contains(bodies[0], `"reasoning":{"summary":"auto"}`))
	t.Logf("turn 2 replays the scratchpad: %v", strings.Contains(bodies[1], "PRIVATE-SCRATCHPAD"))
}
```
</details>

The two hand-resolved conflicts in `doc.go` keep both sides: the generation-config table and the strict-tools paragraph from `main` are intact, and the reasoning paragraph follows them once.

One non-blocking thing. The live e2e run in the Testing Plan was made on an earlier head ([a3de3b5](https://github.com/google/adk-go/commit/a3de3b5088c12e29f500461518e6e3e07025ef12)) with openai-go v3.54.0, and `main` has since moved to openai-go v3.61.0 and genai v1.71.0. A re-run on the current head would keep that section describing what actually merges.

PR #1390, opened 2026-08-23
tool/loadartifactstool: return error instead of panic when artifact service is not configured

GerardGao:
Fixes #283

## What changed

- `loadartifactstool` now returns a descriptive error ("load_artifacts tool requires an artifact service to be configured") from `ProcessRequest` when no artifact service is configured, instead of panicking with a nil pointer dereference.
- `agent.NewToolContext` and `agent.NewCallbackContextWithArtifactTracking` no longer wrap a nil `Artifacts` in `trackedArtifacts`, so "no artifact service configured" remains observable as a nil `Artifacts` instead of panicking on the first promoted method call.

## Why

When the `load_artifacts` tool is registered but no artifact service is configured, the agent panics during request preprocessing (nil pointer dereference in `trackedArtifacts.List`, previously `internalArtifacts.List` — see #283). A configuration error should be reported as an error instead of crashing the process.

## Testing plan

- [x] Added `TestLoadArtifactsTool_ProcessRequest_NilArtifacts`: constructs a context with no artifact service and asserts `ProcessRequest` returns an error mentioning the artifact service (this test panicked before the fix).
- [x] `go test ./agent/... ./tool/... ./runner/...` — all pass.
- [x] `gofmt -l` clean.

PR #724, opened 2026-04-12
feat(session/database): add NewSessionServiceFromDB constructor

nuthalapativarun:
### Link to Issue

Fixes #340

### Description

Adds a `NewSessionServiceFromDB(*gorm.DB)` constructor alongside the existing `NewSessionService(dialector, opts...)`. This allows callers who already manage a database connection to share it across multiple services, rather than having the session service open its own connection.

The existing `NewSessionService` is unchanged for full backward compatibility, as suggested by the maintainer in [the issue comments](https://github.com/google/adk-go/issues/340#issuecomment-3571283631).

**Changes:**

- `session/database/service.go`: Added `NewSessionServiceFromDB(db *gorm.DB) (session.Service, error)` which returns an error if `db` is nil.
- `session/database/service_test.go`: Added `TestNewSessionServiceFromDB` (nil guard, valid db, shared connection verification) and `Test_databaseServiceFromDB` which runs the full service contract suite via the new constructor.

### Testing Plan

- [x] Unit tests added for nil guard, valid construction, and shared-connection assertion
- [x] Full service contract suite (`Test_databaseServiceFromDB`) passes via the new constructor
- [x] All existing tests pass: `go test ./session/database/...`
- [x] Build passes: `go build ./...`

PR #1398, opened 2026-08-24
ci: Release automation with release-please

wojtas000:
## Problem

`internal/version.Version` is edited by hand before every release, and nothing fails
when that is missed - so it drifts. `main` shipped v2.1.0 and v2.2.0 while still
reporting `2.0.0`. The value reaches users: it is sent to Gemini in the
`google-adk/<version>` header, attached to every OTel trace and log as the
instrumentation version, and reported as the MCP client version.

## Solution

Adopt [Release Please](https://github.com/googleapis/release-please) - tool adk-python already releases with - to maintain the constant instead.

On each push to a release line, Release Please keeps a single open release PR that
bumps the annotated constant, regenerates `CHANGELOG.md` (optional) and updates the manifest.
Merging that PR tags the merge commit and publishes the release, so **a tag can only
land on a commit whose constant already matches**.

`main` and `v1` share one workflow but have separate config/manifest pairs, so the two
lines advance independently.

This also corrects the constant to `2.3.0`, the current release. The constant, the
manifest and `last-release-sha` are one triple and are set together: the manifest is
how release-please identifies the previous release (it matches the manifest value
against tags rather than taking the newest tag), so leaving it behind would make the
tool re-scan commits already shipped in `v2.3.0`.

## Prerequisites (not in this PR)

- `RELEASE_PAT` secret - a GitHub App token or PAT. Not required to open the release
  PR (the action defaults `token` to `GITHUB_TOKEN`); it exists because a PR authored
  by `GITHUB_TOKEN` does not trigger other workflows, so the release PR would get no
  CI - and so could never satisfy the required checks. Without the secret the job
  warns and skips, so merging ahead of it costs releases, not a red run on `main`.
- Settings → Actions → "Allow GitHub Actions to create and approve pull requests".
- A follow-up PR porting these files to `v1` (keeping its own `1.6.0` constant).
  That PR **must also add the `x-release-please-version` annotation** to
  `internal/version/version.go` on `v1`: the `generic` updater rewrites only annotated
  lines and skips the rest silently, so config alone would tag `v1.7.0` while the
  constant stayed `1.6.0`. Until then only `main` releases through this.
- **TBD** - do we want `CHANGELOG.md` or not (it is optional, Adk Python has it. But if we want it, we also should decide if the first changelog entry is the current release, or we populate it with previous releases also)
- **TBD** - naming convention for commits (for now taken from [Adk python release-please config](https://github.com/google/adk-python/blob/main/.github/release-please-config.json)) 

## Testing Plan

Validated end to end on a scratch repository using the same config.
`go test -race -mod=readonly -count=1 -shuffle=on work` green

## Notes

- Only [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/) subjects count toward the version and the changelog; unprefixed commits are silently ignored
- Majors stay manual: a `feat!:` would propose v3.0.0, but a Go major bump also
  requires moving the module path to `/v3`, which the tool cannot do.


.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+          BRANCH: ${{ github.ref_name }}
+        run: |
+          set -euo pipefail
+          if [ "$BRANCH" = "v1" ]; then
+            echo "config=.github/release-please-config-v1.json" >> "$GITHUB_OUTPUT"
+            echo "manifest=.github/.release-please-manifest-v1.json" >> "$GITHUB_OUTPUT"
+          else
+            echo "config=.github/release-please-config.json" >> "$GITHUB_OUTPUT"
+            echo "manifest=.github/.release-please-manifest.json" >> "$GITHUB_OUTPUT"
+          fi
```
karolpiotrowicz, line 57: The branch selector has no allow-list: everything that is not `v1` falls through to the main line, and `workflow_dispatch` accepts any ref. Running this step verbatim, `my-feature-branch`, `v2.4.0` and `release-please--branches--main` all select `.github/release-please-config.json`, while `target-branch` follows the dispatched ref.

It needs repo write access, so this is a maintainer foot-gun rather than an exposure, and I did not try to actually cut a tag that way. Failing closed on an unrecognised branch would cost one line.

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      - uses: googleapis/release-please-action@5c625bfb5d1ff62eadeeb3772007f7f66fdcf071  # v4
```
karolpiotrowicz: The comment says `# v4`, but this SHA is exactly `v4.4.1`. The other tag pins in the repo give the exact patch (`# v6.0.2` in `go.yml`, `# v9.2.0` for the lint action).

.github/release-please-config-v1.json
```diff
@@ -0,0 +1,66 @@
...
+  "packages": {
+    ".": {
+      "release-type": "go",
+      "package-name": "adk-go",
+      "include-component-in-tag": false,
+      "changelog-path": "CHANGELOG.md",
+      "extra-files": [
+        {
+          "type": "generic",
+          "path": "internal/version/version.go"
```
karolpiotrowicz, line 12: The `v1` port will need the annotation too, not just these config files. This points the `generic` updater at `internal/version/version.go`, and that file on `origin/v1` has no `x-release-please-version` marker. The updater rewrites only annotated lines and skips everything else silently ([`generic.ts`](https://github.com/googleapis/release-please/blob/05c6a4f71022304d4edad24ea90c1c16324503d5/src/updaters/generic.ts)), so a v1 line set up with the config alone would tag `v1.7.0` while the constant stayed `1.6.0` — the exact drift this PR removes from `main`, with no signal.

Not reachable from this PR, but it binds the follow-up.

CONTRIBUTING.md
```diff
@@ -119,6 +119,12 @@ information on using pull requests.
     you may instead describe the bug or feature directly within the PR
     description, following the structure of our issue templates.
 -   Small, focused PRs. Keep changes minimal—one concern per PR.
+-   PR titles must follow [Conventional Commits](https://www.conventionalcommits.org/):
+    `feat:`, `fix:`, `docs:`, `chore:` and so on, optionally scoped as
+    `fix(runner):`. PRs are squash-merged, so the title becomes the commit
```
karolpiotrowicz: This slightly overstates the guarantee. The repo is set to `COMMIT_OR_PR_TITLE`, which uses the *commit* subject when a PR has exactly one commit, and rebase merging is enabled too, so the PR title is not always what lands. Worth a clause.

The advice itself is right and aimed at a real problem, though: 9 of the last 80 commits on `main` carry no conventional type, including a feature (#1413) and a fix (#1390), and in each of those it was the PR title that lacked the prefix. This paragraph is what fixes that.

karolpiotrowicz (changes requested):
This needs a rebase before it can land, and the rebase has to touch more than the file git flags. Four smaller notes are inline.

## The manifest is stale in a way the merge conflict hides

`v2.3.0` was tagged and released on 31 August, a week after this PR was opened, so `main` now carries `const Version = "2.3.0"`. The only file git reports as conflicting is [`internal/version/version.go`](https://github.com/google/adk-go/blob/ed6cd5ce0551985f2bd0a7fb4b4288946ce5618c/internal/version/version.go) — but [`.github/.release-please-manifest.json`](https://github.com/google/adk-go/blob/ed6cd5ce0551985f2bd0a7fb4b4288946ce5618c/.github/.release-please-manifest.json#L2) is a new file, so it merges cleanly at `2.2.0`, and [`last-release-sha`](https://github.com/google/adk-go/blob/ed6cd5ce0551985f2bd0a7fb4b4288946ce5618c/.github/release-please-config.json#L65) still names the `v2.2.0` commit. Resolving the visible conflict to `2.3.0` therefore leaves the two files that actually drive the release computation pointing a full release behind, with nothing flagging it.

That matters because release-please picks the previous release by matching the **manifest value** against release tags, rather than taking the newest tag — [`manifest.ts:577`](https://github.com/googleapis/release-please/blob/05c6a4f71022304d4edad24ea90c1c16324503d5/src/manifest.ts#L570-L583) is `if (expectedVersion.toString() === tagName.version.toString())`. With the manifest at `2.2.0` it skips `v2.3.0` entirely and walks commits back to `b264039`, re-scanning everything already shipped. Two outcomes, both wrong:

- a `feat:` anywhere in that range makes it propose `2.3.0`, and creating that release throws `DuplicateReleaseError` because the tag already exists ([`github.ts:937`](https://github.com/googleapis/release-please/blob/05c6a4f71022304d4edad24ea90c1c16324503d5/src/github.ts#L937), rethrown at [`manifest.ts:1313-1315`](https://github.com/googleapis/release-please/blob/05c6a4f71022304d4edad24ea90c1c16324503d5/src/manifest.ts#L1300-L1316))
- only `fix:`/`chore:` in range makes it propose `2.2.1`, which tags a descendant of `v2.3.0` and rewrites the constant from `2.3.0` down to `2.2.1`

Either way the first changelog re-lists every commit already in `v2.3.0`. The manifest, `last-release-sha` and the constant need to move together as one reviewed triple.

## An absent `RELEASE_PAT` crashes the workflow rather than skipping it

You flagged creating the secret as a prerequisite, so this is only about the failure mode, which is harsher than it looks. [`token: ${{ secrets.RELEASE_PAT }}`](https://github.com/google/adk-go/blob/ed6cd5ce0551985f2bd0a7fb4b4288946ce5618c/.github/workflows/release.yml#L54) with no such secret does not fall back to `github.token`. The runner applies an `action.yml` default only when the key is **absent** from `with:` — [`ActionRunner.cs:214`](https://github.com/actions/runner/blob/0b0ac2fdabf53d69add6175026945b8afc8549a5/src/Runner.Worker/ActionRunner.cs#L214-L217) is `if (!inputs.ContainsKey(key))` — and an expression evaluating to `""` still puts the key there. The action then reads it with `required: true` ([`index.ts:52`](https://github.com/googleapis/release-please-action/blob/5c625bfb5d1ff62eadeeb3772007f7f66fdcf071/src/index.ts#L52)), which throws on empty ([`core.ts:154`](https://github.com/actions/toolkit/blob/193fa46c20fde8b0ed54194bc08b841c78c0776d/packages/core/src/core.ts#L151-L156)).

So merging before the secret exists puts a red **Release** run on every push to `main` indefinitely. It blocks nothing, since the required checks are only `lint (.)` and `test (.)`, but a permanently failing check on the default branch is its own cost. Worth sequencing the secret ahead of the merge, or letting the job tolerate the secret being absent.

For what it is worth, the parts most likely to go wrong here are right: the fork guard, top-level `contents: read` with the write grant scoped to the job, and passing the branch through `env:` instead of interpolating it into the `run:` block. `actionlint` is clean on the new workflow, and its repo-wide output is identical to `main`.

.github/workflows/release.yml
wojtas000: Added comment pinning exactly `v4.4.1`

.github/workflows/release.yml
wojtas000: Fixed: Now we switch over `main` and `v1` branches, fail workflow for others

.github/release-please-config-v1.json
wojtas000: yes, after merging this PR we would need to create analogous PR for v1 and add the `x-release-please-version` marker for version constant.

The discrepancy in the `release-please-manifest-v1.json` on `main` branch will be real, meaning that if e.g. we ship `v.1.7.0` on v1 branch, the  `release-please-manifest-v1.json` in main will stay on `v.1.6.0`. But it does not harm us, as the release versions for v1 will be bumped from v1's `release-please-manifest-v1.json`  and will stay consistent.

Another approach is to keep just `.release-please-manifest.json‎` and `release-please-config.json` per main/v1 branch (and e.g. not have redundant v1 files on main).

CONTRIBUTING.md
wojtas000: Added fix to paragraph about keeping convention in one commit PRs. 

Worth adding: we could have a guard on each PR (in form of additional CI workflow) similar to linting/testing, which would validate the naming conventions for PRs (prefixes allowed in `release-please-config.json`. So we would ensure that every PR will appear in the release notes.
@karolpiotrowicz what is your take on this?

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+    # Forks inherit this workflow but must never cut a release from it; without
+    # the guard every fork gets a failing run on each push to main.
+    if: github.repository == 'google/adk-go'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write        # create the release branch, the tag and the release
+      pull-requests: write   # open and update the release PR
+    env:
+      # The secrets context is not available to `if:`; mirror the token to branch on it.
+      RELEASE_PAT: ${{ secrets.RELEASE_PAT }}
```
karolpiotrowicz: Non-blocking: this puts the PAT in the environment of every step in the job, where the previous `with:`-only form kept it inside the one pinned action. Nothing reads it today — the job has three steps and they are all here — but a step added later inherits a repo-write credential with no signal at the step itself, thirty lines below the grant.

Gating on whether the secret is set, rather than on its value, keeps it out of the job environment: `HAS_RELEASE_PAT: ${{ secrets.RELEASE_PAT != '' }}` at the job level, `if: env.HAS_RELEASE_PAT == 'true'` on the step, and `token: ${{ secrets.RELEASE_PAT }}` left in `with:`. I checked that shape passes `actionlint`, but I have not run it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+            exit 1
+          fi
+          echo "config=$config" >> "$GITHUB_OUTPUT"
+          echo "manifest=$manifest" >> "$GITHUB_OUTPUT"
+
+      # An unset secret does not fall back to github.token; it is passed as an
+      # empty input and the action rejects it. Skip rather than fail the run.
+      - name: Check for the release token
+        if: env.RELEASE_PAT == ''
+        run: echo "::warning title=Release skipped::RELEASE_PAT is not configured, so no release PR was created or updated."
```
karolpiotrowicz, line 65: Worth being deliberate about the trade here rather than changing anything. Because the secret does not exist yet, every push to `main` will take this path on merge: a warning annotation on a green run, and no release PR. The header comment above describes the working behaviour in the present tense, and nothing in the file records that an admin has to create the secret first.

The same shape returns later if the PAT expires — releases stop, the checks stay green, and the warning is the only trace. Skipping is the right call versus the red run this replaced. I am flagging the residue, not asking you to undo it.

karolpiotrowicz (dismissed):
All five are fixed, and I checked each against the tree rather than taking the commit messages for it. The manifest, `last-release-sha` and the constant now agree on `2.3.0`, with the sha being exactly the v2.3.0 commit, so the first run will compute from the right baseline. The conflict is gone and the PR is mergeable.

The token guard is the part I looked at hardest, since it is new code. It works: `secrets` genuinely is not available in a step `if:` and `env` is, so the mirror is the right shape. I confirmed that rather than assuming it — `actionlint` rejects `secrets` in a step `if:` and accepts `env`, which also means its clean run on this file is a real pass rather than a gap in what it checks.

Your reading of the v1 manifest is right, and the source backs it: release-please reads both the config and the manifest at the *target* branch ([`manifest.ts:445-450`](https://github.com/googleapis/release-please/blob/05c6a4f71022304d4edad24ea90c1c16324503d5/src/manifest.ts#L445-L450)), so a v1 release bumps from v1's own files and main's copy drifting is inert. Your second option — one `release-please-config.json` and one manifest per branch, with no `-v1` files on main — is the tidier end state, since it also deletes the branch-selector step. Not for this PR.

Two non-blocking notes inline, neither of which needs to hold up the merge.

On a PR-title lint: yes, worth doing, and the numbers support it. 9 of the last 80 commits on `main` carry no recognized type, including a feature (#1413) and a fix (#1390) — so roughly one in nine changes would be missing from the release notes today. Two things I would want from it: that it checks the commit subject as well on a single-commit PR, matching the wording you just landed, and that it takes its list of allowed types from `release-please-config.json` rather than a second hardcoded copy, so the two cannot drift. Separate PR.

One thing that is mine rather than yours: `RELEASE_PAT` still does not exist on the repo. I will create it. Until then the workflow lands inert, which is the point of the guard you added.

wolo-lab (dismissed):
Thank you!

.github/workflows/release.yml
wojtas000: Added the `HAS_RELEASE_PAT` gating

wojtas000:
@karolpiotrowicz @wolo-lab 
I applied one more change suggested in the review and need reapproval.

Next steps after this PR is merged:
- We need to set the RELEASE_PAT [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)
- As [release-please documentation](https://github.com/googleapis/release-please-action) states: "You may also need to set "Allow GitHub Actions to create and approve pull requests" under repository Settings > Actions > General."
- PR for v1 branch

karolpiotrowicz (approved):
The token stays out of the job environment now, and the guard still behaves identically in both states: with no secret set, `secrets.RELEASE_PAT != ''` casts to the string `"false"`, so the release step is skipped and the warning fires exactly as before. `actionlint` is clean on the file and the repo-wide output is unchanged from `main`.

Worth knowing what the first run will do once the secret exists. Seven commits have landed on `main` since v2.3.0 — one `feat:` (#724), five `fix:`, one `chore:`, nothing breaking — so release-please will open a `chore(main): release 2.4.0` PR. That is a clean minor bump with no collision against an existing tag, which is the outcome the manifest fix was for.

Creating `RELEASE_PAT` is mine, not yours. Until it exists the workflow lands inert, which is what the guard is there for, so this does not need to wait on it.

PR #1399, opened 2026-08-24
fix(cmd/adkgo): allow setting runtime service account and configurable secret name on Cloud Run deploy

### LINKED ISSUE #1343: adkgo deploy cloudrun: no way to set the runtime service account, and the Secret Manager requirement is undocumented
`adkgo deploy cloudrun` always renders a Secret Manager mount into the `gcloud run deploy` call, and offers no way to set the runtime service account. Both behaviours are undocumented on the command.

Reproduced at v2.2.0 (`b264039aaec43baedc123e5b9a0cf87681d0bbca`).

## Hardcoded Secret Manager mount

The generated deploy always includes [`--set-secrets=GOOGLE_API_KEY=GOOGLE_API_KEY:latest`](https://github.com/google/adk-go/blob/b264039aaec43baedc123e5b9a0cf87681d0bbca/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go#L246), so every deploy implicitly requires two things:

- a secret named `GOOGLE_API_KEY` to exist in the target project
- the runtime service account to hold `roles/secretmanager.secretAccessor`

Neither is mentioned in the command's help output or in the deployment docs, and there is no opt-out for users who inject the key as a plain environment variable or use a different secret name.

## No control over the runtime service account

There is no flag to set the service account. The full persistent flag set at [L102-L122](https://github.com/google/adk-go/blob/b264039aaec43baedc123e5b9a0cf87681d0bbca/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go#L102-L122) covers region, project, service name, temp dir, ports, entry point, A2A, API, Web UI, and Pub/Sub and Eventarc tuning. Nothing addresses identity, so the revision falls back to the project's default compute service account.

## Observed failure

The two combine into an opaque failure at revision-creation time, after the container has already built:

```
> Creating Revision....................................failed
> Deployment failed
> ERROR: (gcloud.run.deploy) spec.template.spec.containers[0].env[0].value_from.secret_key_ref.name:
  Permission denied on secret: projects/nnnn/secrets/GOOGLE_API_KEY/versions/latest for Revision
  service account nnnn-compute@developer.gserviceaccount.com. The service account used must be
  granted the 'Secret Manager Secret Accessor' role.
```

Nothing in the error points back to a choice `adkgo` made, so the failure reads as a project misconfiguration rather than a default in the tool. The reporter worked it out by reading the source and the generated Dockerfile, then went back to their own Cloud Run deployment.

## Acceptance criteria

- A deploy can target a caller-specified runtime service account.
- The API key source is either configurable or clearly documented, and a deploy that cannot satisfy the prerequisite fails early with a message naming the missing secret or role, rather than after the container build.

Reported in discussion #571.

sundeep8967:
## Problem
`adkgo deploy cloudrun` always hardcoded `--set-secrets=GOOGLE_API_KEY=GOOGLE_API_KEY:latest` into the generated `gcloud run deploy` invocation and provided no option to specify a runtime service account. This resulted in:
1. Hard dependency on a Secret Manager secret named `GOOGLE_API_KEY`, breaking deploys for users configuring API keys via environment variables or using different secret names.
2. Silent fallback to the default compute service account without `roles/secretmanager.secretAccessor` permission, causing cryptic runtime revision failures after the container build completes.

## Solution
- Added `--service_account` / `-A` flag to `cloudrunCmd` to target a caller-specified runtime service account.
- Added `--secret_name` flag (defaulting to `GOOGLE_API_KEY`) allowing custom secret names or disabling Secret Manager mounting by setting `--secret_name=""`.
- Documented Secret Manager and service account prerequisites in `cloudrunCmd.Long` help text.
- Added unit tests in `cmd/adkgo/internal/deploy/cloudrun/cloudrun_test.go` verifying argument construction across all combinations.

Fixes #1343

## Testing
- Ran `go test ./cmd/adkgo/internal/deploy/cloudrun -v` (all tests pass).
- Ran `go test ./internal/ -run TestCopyrightHeader` (all header checks pass).


sundeep8967:
@google-cla recheck

karolpiotrowicz (commented):
The default path is unchanged, and I checked that by capturing the argv rather than reading the code: with neither new flag supplied, the emitted `gcloud run deploy` command has the same 13 arguments as before this PR, with only the position of `--set-secrets` moved. `--no-allow-unauthenticated` and `--ingress all` are still emitted on every path, there is no shell anywhere (`exec.Command` takes an argv slice, so a value containing `;` arrives as one element), and omitting `--service_account` reproduces the previous behaviour exactly. Build, tests, vet, staticcheck, golangci-lint, govulncheck and `go mod tidy -diff` are all clean. This is also the first test file in `cmd/adkgo`.

One thing is worth fixing before this merges, and it is a one-line change to the test rather than to the code.

## The test does not pin the flag that keeps the service private

[`buildDeployParams`](https://github.com/google/adk-go/blob/0097723a722c5b97eede1a0c4aea04d26163ddea/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go#L259-L277) emits `--no-allow-unauthenticated` at [cloudrun.go:266](https://github.com/google/adk-go/blob/0097723a722c5b97eede1a0c4aea04d26163ddea/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go#L265-L266), but none of the three `wantContains` lists names it — see [cloudrun_test.go:41-46](https://github.com/google/adk-go/blob/0097723a722c5b97eede1a0c4aea04d26163ddea/cmd/adkgo/internal/deploy/cloudrun/cloudrun_test.go#L41-L49). Changing that one literal to `--allow-unauthenticated`, which would make every deployed service publicly invocable, leaves the entire `cmd/...` tree green:

```
$ sed -i 's/"--no-allow-unauthenticated"/"--allow-unauthenticated"/' \
    cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
$ go test ./cmd/...
ok  	google.golang.org/adk/v2/cmd/adkgo/internal/deploy/cloudrun	0.027s
ok  	google.golang.org/adk/v2/cmd/launcher/console	0.166s
ok  	google.golang.org/adk/v2/cmd/launcher/web	0.103s
ok  	google.golang.org/adk/v2/cmd/launcher/web/a2a	0.146s
ok  	google.golang.org/adk/v2/cmd/launcher/web/triggers/pubsub	0.097s
```

Two smaller gaps have the same root cause and the same fix. [`slices.Contains`](https://github.com/google/adk-go/blob/0097723a722c5b97eede1a0c4aea04d26163ddea/cmd/adkgo/internal/deploy/cloudrun/cloudrun_test.go#L100-L104) tests membership, so it cannot see order: swapping `append(params, "--service-account", sa)` to `append(params, sa, "--service-account")` also passes. And nothing asserts the argument count, so an argument *added* to the slice is invisible too. Comparing the whole expected `[]string` for at least the default case closes all three at once, and it pins `--source .` as a side effect.

Worth knowing about `--ingress all`: it is gcloud's own default, so unlike `--no-allow-unauthenticated` it is not carrying any weight here.

## The registered default is the one thing the default-path case does not exercise

[cloudrun_test.go:38](https://github.com/google/adk-go/blob/0097723a722c5b97eede1a0c4aea04d26163ddea/cmd/adkgo/internal/deploy/cloudrun/cloudrun_test.go#L31-L40) writes `secretName: "GOOGLE_API_KEY"` into its own struct, so the case named `default_secret_no_service_account` never reads the default registered at [cloudrun.go:114](https://github.com/google/adk-go/blob/0097723a722c5b97eede1a0c4aea04d26163ddea/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go#L113-L114). Changing that registered default to `""` — which would silently stop mounting the secret on every default deploy — also passes the whole tree. No other test in this repo asserts a cobra default, so this is a suggestion rather than a standard you are being held to uniquely, but the case name does promise it.

## `--secret_name=""` on a redeploy

[The help text](https://github.com/google/adk-go/blob/0097723a722c5b97eede1a0c4aea04d26163ddea/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go#L96-L99) says this "disables Secret Manager mounting". What the code does is emit no secret flag at all — not `--clear-secrets`, not `--remove-secrets`. Since `gcloud run deploy` is documented as create-or-update and gcloud offers `--clear-secrets` ("Remove all secrets") as a distinct operation, a redeploy of a service that already has the mount may well keep it, which would make the sentence true only for a first deploy. I could not settle this — it needs a real deploy against a scratch project, which I had no way to run. Worth either checking or narrowing the wording to "skips passing `--set-secrets`".

## Smaller things

- `-A` at [cloudrun.go:113](https://github.com/google/adk-go/blob/0097723a722c5b97eede1a0c4aea04d26163ddea/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go#L113-L114) is one shift-key from the existing `-a` (`a2a_agent_url`), which also takes a string, so `-a sa@project.iam.gserviceaccount.com` sets the advertised agent-card URL to an email and quietly deploys with the default service account. It is also the only uppercase shorthand in the CLI, and `--secret_name` in this same diff takes none. Dropping it would match.
- `:latest` is hardcoded at [cloudrun.go:270](https://github.com/google/adk-go/blob/0097723a722c5b97eede1a0c4aea04d26163ddea/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go#L269-L274), so `--secret_name=MY_SECRET:3` becomes `…=MY_SECRET:3:latest`. Not a regression, but this PR is what makes the secret name configurable, so it is the natural place to decide whether a version should be pinnable.
- At [cloudrun_test.go:108](https://github.com/google/adk-go/blob/0097723a722c5b97eede1a0c4aea04d26163ddea/cmd/adkgo/internal/deploy/cloudrun/cloudrun_test.go#L106-L112) the `arg == notWant` half is redundant with the prefix check beside it. `strings.HasPrefix` says the same thing. Keep the prefix behaviour — it is what catches `--set-secrets=...` as a whole element.

sundeep8967:
thanks for the detailed review! updated the changes as requested:
- removed the `-A` shorthand for `--service_account` to avoid ambiguity with `-a`
- updated `TestBuildDeployParams` to assert full slice equality via `slices.Equal`, explicitly pinning `--no-allow-unauthenticated`, `--ingress all`, `--source .`, and full argument order.

karolpiotrowicz:

The feature itself is already asked for in #1343, so this is narrower: it adds two new CLI flags with no sibling precedent, since adk-python has neither and does not use Secret Manager at all. Three things to settle — whether `--secret_name=""` is the right way to disable Secret Manager mounting, whether the short form is worth spending on a rarely used flag, and whether Go should keep Secret Manager mounting when Python uses a plain environment variable. Flag contract rather than feature direction.

Issue #1391, opened 2026-08-23
mcptoolset: non-text tool-result content blocks are silently dropped (EmbeddedResource, ResourceLink, Image, Audio)


lukephillippi:
## Summary

`tool/mcptoolset/tool.go`'s result conversion in `mcpTool.Run` keeps only `*mcp.TextContent` blocks. Every other content-block class — `EmbeddedResource`, `ResourceLink`, `ImageContent`, `AudioContent` — is silently discarded:

```go
textResponse := strings.Builder{}
for _, c := range res.Content {
    textContent, ok := c.(*mcp.TextContent)
    if !ok {
        continue
    }
    ...
}
```

The same text-only loop is used for the `IsError` path, and when `res.StructuredContent` is non-nil the content blocks are ignored entirely.

## Real-world impact

GitHub's official MCP server (github/github-mcp-server, also the hosted api.githubcopilot.com endpoints) returns `get_file_contents` file bodies as an `EmbeddedResource` block alongside a status-line `TextContent` (`pkg/utils/result.go` `NewToolResultResource`). Through mcptoolset, the model therefore receives only:

```
successfully downloaded text file (SHA: …)
```

with no file content at all — the agent is effectively blind to every file it reads, with no error or indication anything was dropped. Directory listings (plain text) survive, which makes the loss non-obvious. Files ≥ 1MB come back as `ResourceLink`, also dropped.

Related: #1352 covers the "no text content in tool response" error, which is what a result consisting *only* of non-text blocks currently produces — the two issues are adjacent but distinct.

## Suggested behavior

At minimum, flatten `EmbeddedResource` with textual content (`Resource.Text`, or `Resource.Blob` with a text-like MIME type) into the text output, and represent `ResourceLink`/binary resources legibly (URI + MIME + size) rather than dropping them. Ideally, image/audio blocks would map to genai `inlineData` parts.

## Workaround

Supplying `Config.Client` with a custom `*mcp.Client` carrying an `AddSendingMiddleware` that rewrites `EmbeddedResource`/`ResourceLink` blocks into `TextContent` before the conversion runs restores the content losslessly. Happy to upstream a PR along those lines if useful.

Observed on v2.2.0; `main` is currently identical for this file.

ktsoator:
[...]

PR #1353, opened 2026-08-17
fix(mcptoolset): allow valid empty text tool responses (#1352)

chill-czar:
### Link to Issue or Description of Change

- Closes: #1352

### Problem

In `mcptoolset.mcpTool.Run`, if an MCP tool completed successfully with `isError = false` but returned no text content or an empty text string, `Run` returned an error (`"no text content in tool response"`). Under the Model Context Protocol specification, an empty string is a valid non-error tool result.

### Solution

Remove the error check on empty text responses in `mcpTool.Run`, allowing empty text content to return `map[string]any{"output": ""}` successfully. Add a unit test in `set_test.go` verifying the behavior.

### Testing Plan

- [x] Unit test added: `TestMCPTool_EmptyTextResponse` in `tool/mcptoolset/set_test.go`.
- [x] Confirmed test fails without the fix and passes with it.
- [x] All repository verification commands pass:
  ```text
  go test -v -race -count=1 ./tool/mcptoolset/...
  go build -mod=readonly work
  golangci-lint run
  go mod tidy -diff
  ```

### Checklist

- [x] I have read the [CONTRIBUTING.md](./CONTRIBUTING.md) document.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have manually tested my changes end-to-end.
- [x] Any dependent changes have been merged and published in downstream modules.

PR #1401, opened 2026-08-25
fix(mcptoolset): render non-text unstructured tool results

ktsoator:

- Addresses #1391 for unstructured success responses and MCP error details. The structured-content coexistence case remains tracked in that issue.
- Related to #1353, which handles valid empty tool responses. This PR does not change empty-response behavior.

Thanks @lukephillippi for the detailed report and real-world reproduction.

### Problem

For non-structured MCP tool results, `mcpTool.Run` retained only `TextContent`. `EmbeddedResource`, `ResourceLink`, `ImageContent`, and `AudioContent` blocks were silently discarded. Error responses used the same text-only assumption.

This prevented agents from receiving file contents returned by GitHub's MCP server.

Typed-nil content could also panic because a type assertion for `*mcp.TextContent` succeeds before the value is dereferenced.

### Scope

This PR fixes content loss in unstructured success responses and MCP error details.

It intentionally preserves the existing behavior when `StructuredContent` is non-nil: structured output remains authoritative and accompanying unstructured content blocks are not duplicated. Supporting mixed structured and unstructured results remains out of scope for this change and continues to be tracked in #1391.

### Solution

- Preserve textual embedded resources in tool responses.
- Safely decode UTF-8 and ASCII blobs with valid text-like media types.
- Detect declared MIME charset parameters even when `mime.ParseMediaType` silently drops an unsupported RFC 2231 `charset*` value.
- Avoid treating semicolons inside quoted MIME parameter values as parameter separators.
- Reject malformed charset declarations and unsupported charsets before decoding.
- Recognize `application/x-www-form-urlencoded` as textual content.
- Represent binary resources, resource links, images, and audio with relevant metadata.
- Make an accompanying blob visible through its size when an invalid resource contains both text and blob data.
- Preserve content order and existing text-only behavior.
- Preserve the existing structured-output path without duplicating content.
- Use the same conversion for MCP error details.
- Handle typed-nil content safely with unavailable markers.
- Use a stable marker for unsupported content without exposing Go type names.

`FunctionTool.Run` currently exposes a JSON response rather than function-response media parts, so image and audio blocks are represented by metadata. Native inline media support can be addressed separately.

### Testing Plan

- [x] Added table-driven text-only and GitHub MCP regression tests.
- [x] Covered textual blobs, binary resources, media, links, errors, and structured-output compatibility.
- [x] Covered RFC 2231 `charset*`, parameter ordering and casing, quoted MIME values, ASCII aliases, and malformed media types.
- [x] Added an in-memory MCP wire round-trip test for embedded resources, resource links, images, and audio.
- [x] Covered typed nils, separators, unsupported content, and empty responses.
- [x] Targeted `tool/mcptoolset` tests and race tests.
- [x] Repository-pinned `golangci-lint` for the current changes.
- [x] `go test -race -mod=readonly -count=1 -shuffle=on work`
- [x] `golangci-lint v2.3.1 run` in every module
- [x] `go mod tidy -diff` in every module


ktsoator:
Hi @baptmont and @hanorik, when you have a chance, could you please review this PR? It addresses #1391 by surfacing non-text MCP tool result content while preserving existing text-only behavior. All CI checks are passing. Thank you!

karolpiotrowicz (commented):
The conversion holds up well under differential testing. I ran every text-only shape I could construct through `Run` on both `main` and this branch and compared the JSON that actually lands in `genai.FunctionResponse.Response` — single block, several blocks, empty blocks interleaved, embedded NUL bytes, invalid UTF-8, leading and trailing newlines — and the two trees are byte-identical on all of them. That is the property this change most needed to preserve, and the `isPlain` split at [tool.go:220](https://github.com/google/adk-go/blob/2d25be768cc86af827e884cda83d6fbe3d47e01d/tool/mcptoolset/tool.go#L220) makes it true by construction rather than by luck: for an all-text list the separator predicate can never fire.

One thing worth flagging that the description does not claim: on `main`, a content slice holding a typed-nil `*mcp.TextContent` panics on the unguarded `textContent.Text`. This branch returns a placeholder instead. You have fixed a nil dereference, probably without meaning to.

## The structured-content branch can send the payload twice

This is the one I would most like your view on. When `res.StructuredContent` is set and any non-text block is present, [tool.go:147-149](https://github.com/google/adk-go/blob/2d25be768cc86af827e884cda83d6fbe3d47e01d/tool/mcptoolset/tool.go#L147-L149) renders `content` from **all** blocks, plain text included. The [MCP specification](https://modelcontextprotocol.io/specification/2026-07-28/server/tools) tells servers that return structured content to *also* send the same payload serialized into a `TextContent` block, for older clients. When a server follows that advice and attaches a resource as well, `output` carries the structured payload and `content` carries a serialized copy of the same thing, plus the resource.

Measured on a 108,000-byte file body, the JSON the model receives goes from 111,057 bytes on `main` to 336,207 bytes here. Of the 219,119-byte `content` value, 111,047 bytes duplicate what `output` already holds. The rest of the growth is the file itself, which is the whole point of the change — so the avoidable part is about a third of the payload, not the full 3x.

I tried the obvious fix before suggesting it, and it does not work: skipping plain-text blocks in that branch compiles cleanly but fails your own `structured output retains non-text content` case, because the plain block there is a real status line rather than a duplicate. Distinguishing the two needs a comparison against the serialized structured content, which is a larger decision than a review comment should make for you.

Worth saying: the GitHub `get_file_contents` case from #1391 sends no `structuredContent`, so it never reaches this branch. I could not produce a capture from a real server that does, so this is a shape I reasoned to from the spec rather than one I observed in the wild.

## A malformed media type silently decodes the blob as UTF-8

At [tool.go:306-314](https://github.com/google/adk-go/blob/2d25be768cc86af827e884cda83d6fbe3d47e01d/tool/mcptoolset/tool.go#L306-L314), when `mime.ParseMediaType` returns an error the code recovers `mediaType` from its own split but keeps the nil `params`, so `charset` reads as the empty string and the blob takes the UTF-8 branch. A declared `utf-16` that arrives in a slightly broken header is therefore decoded as if it were UTF-8, and NUL bytes go straight into the model's context. The well-formed spelling of the same header is correctly reduced to metadata, and you have a test pinning that — so the two paths disagree on identical bytes, decided only by whether the server's header parses.

<details><summary>Reproduction — drop into <code>tool_test.go</code> and run <code>go test -run TestMCPToolRunContent ./tool/mcptoolset/</code></summary>

```go
{
    name: "malformed charset header still decodes as UTF-8",
    result: &mcp.CallToolResult{Content: []mcp.Content{
        &mcp.EmbeddedResource{Resource: &mcp.ResourceContents{
            URI:      "file:///utf16.txt",
            MIMEType: `text/plain; charset="utf-16`, // unterminated quote
            Blob:     []byte{0x68, 0x00, 0x69, 0x00}, // UTF-16LE "hi"
        }},
    }},
    // Currently produces: ...mimeType="text/plain; charset=\"utf-16"]\nh\x00i\x00
    // The well-formed sibling produces: ..., size=4 bytes]
    want: map[string]any{"output": `[MCP embedded resource: uri="file:///utf16.txt", ` +
        `mimeType="text/plain; charset=\"utf-16", size=4 bytes]`},
},
```
</details>

The same fallback also widens the media type itself, so a header like `text/plain extra` fails the strict parser and then passes the `text/` prefix check at [tool.go:333](https://github.com/google/adk-go/blob/2d25be768cc86af827e884cda83d6fbe3d47e01d/tool/mcptoolset/tool.go#L333). Refusing to decode whenever `ParseMediaType` errored would close both.

## The fallback branch puts a Go type name in the prompt

[tool.go:203-205](https://github.com/google/adk-go/blob/2d25be768cc86af827e884cda83d6fbe3d47e01d/tool/mcptoolset/tool.go#L203-L205) formats anything unhandled as `[MCP content: unsupported type *mcp.ToolUseContent]`. The model has no use for a Go type name, and the string will change under it if the SDK ever renames the type. The wire-level `type` discriminator, or a fixed `[MCP content: unsupported]`, reads better and is stable.

This also changes a case that already worked: a slice of text, an unhandled block, then more text returns `"beforeafter"` on `main` and `"before\n[MCP content: unsupported type *mcp.ToolUseContent]\nafter"` here. Surfacing the block rather than dropping it is the better behavior, so this is about the wording rather than the decision.

## Test coverage, in rough order of what I would add

The suite is good and the coverage it adds is real. These are the branches I could change without any test noticing:

- **No typed-nil case at all.** `nil resource does not panic` uses `&mcp.EmbeddedResource{}` — a non-nil pointer with a nil `Resource` — so it pins one half of one guard. The nil-dereference fix mentioned at the top is the one worth pinning, since it is the shape that crashes on `main`: `[]mcp.Content{(*mcp.TextContent)(nil)}`. One case with all five typed nils in a single slice covers the lot.
- **`unsupported text charset is represented by metadata` does not pin the charset rule.** Its blob is `{0xff, 0xfe, ...}`, which is also invalid UTF-8, so the case passes whether the rejection came from the charset check or the UTF-8 check — and the neighbouring case already covers the latter. A valid-UTF-8 blob with `charset=utf-16` would make it test what its name says.
- **The `us-ascii` branch has no case.** Nothing exercises it, and nothing pins the `b >= utf8.RuneSelf` boundary, so a `>` there would go unnoticed and let a lone `0x80` through.
- **The newline guard is never exercised.** No case puts a block whose text ends in `\n` immediately before a non-plain block, so replacing that guard with `true` changes nothing observable.
- **The empty-block `continue` can be removed** without failing anything. In the `image and audio` fixture the empty block sits between two plain blocks, where its removal is invisible. Move it before a non-plain block and it produces a stray leading newline.
- **The surviving error is untested.** `no text content in tool response` is still reachable for nil, empty, and all-empty-text content, and no case asserts it — so that boundary could drift to returning `{"output": ""}` silently.

## Smaller things

`image/svg+xml` in the switch at [tool.go:338](https://github.com/google/adk-go/blob/2d25be768cc86af827e884cda83d6fbe3d47e01d/tool/mcptoolset/tool.go#L338) is unreachable, since the `+xml` suffix rule above it already returns true. And `content != ""` at [tool.go:147](https://github.com/google/adk-go/blob/2d25be768cc86af827e884cda83d6fbe3d47e01d/tool/mcptoolset/tool.go#L147) can never be false when `hasNonText` is true, because every non-plain branch returns at least a bare `[MCP <kind>]` label.

## Things I checked that turned out fine

Recording these so they do not get re-litigated. The `[MCP ...]` markers are forgeable by a server that simply emits the same string as plain text, but `main` already passes that exact string through untouched, and the attributes you do render go through `%q`, which escapes a quote or newline injected into a URI. Unbounded output is the same story: an 8 MB text block already produces an 8 MB payload on `main`, byte for byte, so this change adds routes to a ceiling that was already there rather than raising it. The error-path callbacks that stop firing were firing on the spurious `no text content` error, which is the bug. And `apidiff` reports no exported API change — every new symbol is unexported.

ktsoator:
Thank you @karolpiotrowicz for the thorough review and differential testing. The structured-content duplication you identified is a valid concern.

Because the GitHub `get_file_contents` response in #1391 does not use `structuredContent`, and reliably distinguishing compatibility JSON from independent text would require a separate policy, I restored the existing structured-output path in this PR. Mixed structured/non-text handling is no longer part of this change, and I narrowed the description accordingly.

I also addressed the remaining points:
- `ParseMediaType` errors now fall back to resource metadata instead of attempting UTF-8 decoding.
- Unsupported content now uses a stable marker without exposing Go type names.
- Tests now cover all five typed-nil variants, charset isolation, US-ASCII and its `0x80` boundary, existing-newline and empty-block separators, and nil, empty, and all-empty responses.
- The redundant `image/svg+xml` switch entry and the structured-path `hasNonText` plumbing were removed.

The full workspace build and race tests pass. Lint with the repository-pinned v2.3.1 and `go mod tidy -diff` also pass in both modules.

Thank you again for identifying these edge cases.

karolpiotrowicz (commented):
All six items from the last round are addressed, and the payload duplication is measurably gone: for a 108,000-byte body sent as structured content plus the serialized-text duplicate plus the resource, the model-visible payload drops from 324,168 bytes to 108,040, byte-identical to the merge base.

## The rebase is what is holding this up

Not your doing — #1353 landed at 14:07 UTC, about seven hours after you pushed [3e64d5c](https://github.com/google/adk-go/commit/3e64d5c4318239ef82e12cf6182c146dc076dde6), and it deleted the same empty-content guard your branch still carries at [tool.go:147-150](https://github.com/google/adk-go/blob/90f18acdd334222de907b79269111994b041e182/tool/mcptoolset/tool.go#L147-L150). Worth knowing before you start, because neither one-sided resolution works. I ran both:

- **Drop the guard**, leave the tests alone, and three cases you added go red — `nil content returns an error`, `empty content returns an error` and `all-empty text content returns an error`, at [tool_test.go:258-277](https://github.com/google/adk-go/blob/90f18acdd334222de907b79269111994b041e182/tool/mcptoolset/tool_test.go#L258-L277).
- **Keep the guard**, and `TestMCPTool_EmptyTextResponse` goes red instead — #1353's own test, arriving from main at [set_test.go:800](https://github.com/google/adk-go/blob/ffddb2a6d7ed97974e82815d00c95eb148038f2c/tool/mcptoolset/set_test.go#L800).

Dropping the guard and changing those three cases to expect `{"output": ""}` gives 51 passing subtests and no failures, both suites green together. That combination is also what fixes main: image-only, audio-only, resource-link-only and embedded-resource-only results all return `{"output": ""}` there today, and all four come back correct with your branch on top.

One thing to know before you pick a resolution — #1473 addresses the same gap in main by returning an error for non-text-only results. If it lands first, its guard becomes unreachable under your change and will need removing.

## Two worth fixing while you are in there

**`decodeTextBlob` throws away a media type it parsed successfully.** [tool.go:296-298](https://github.com/google/adk-go/blob/90f18acdd334222de907b79269111994b041e182/tool/mcptoolset/tool.go#L296-L298) returns on any `mime.ParseMediaType` error, but that function reports `ErrInvalidMediaParameter` while still handing back a correct media type when only the parameter section is bad. These two now lose their decodable body and render as metadata only, where the previous revision inlined it:

```
text/plain; name=my file.txt
application/json; profile=https://example.com/s
```

Still better than main, which shows nothing at all for those blocks, so this is not urgent. The obvious fix does not work — accepting any non-empty `mediaType` re-breaks `malformed charset is represented by metadata`, the UTF-16 case you just fixed. This keeps your suite green and restores both bodies, though it does fire on an unrelated parameter like `x-charset-note=`:

```go
if err != nil && (mediaType == "" || strings.Contains(strings.ToLower(mimeType), "charset")) {
	return "", false
}
```

**The separator guard is only half covered.** [tool.go:210-212](https://github.com/google/adk-go/blob/90f18acdd334222de907b79269111994b041e182/tool/mcptoolset/tool.go#L210-L212) tests two conditions, and deleting `!strings.HasPrefix(block.text, "\n")` leaves the whole suite passing — nothing in the fixtures starts a block with a newline. An image followed by `TextContent{Text: "\nnotes"}` renders a doubled newline with that half gone. One table row covers it.

## Smaller

An untyped `nil` in the content slice reaches the model as `[MCP content: unsupported]`, reading as an unknown block type rather than an absent one, though the SDK will not produce that from the wire. Four branches in the new code are unpinned — the empty-blob check, `mimeType != ""` in `formatMediaContent`, the zero-attribute label arm, and `details != ""` on the error path — all four correct when driven directly.

Last, a naming point. With `StructuredContent` set, `Run` returns before the content is rendered, so an image sent alongside structured output still does not reach the model. That matches main and the merge base exactly and is the right call, since attaching it is what produced the duplication above. But `structured output with mixed content remains compatible` reads like the opposite promise, and so does the PR title.

ktsoator:
Thank you again @karolpiotrowicz for the detailed follow-up and for checking the merge against #1353.

I merged the latest main and applied the suggested resolution:

- Removed the empty-content guard and updated nil, empty, and all-empty text cases to return {"output": ""}.
- Kept structured output authoritative so mixed structured content is not duplicated.
- Handled mime.ErrInvalidMediaParameter when the base media type remains usable and no charset parameter is involved, with regression cases for malformed unrelated parameters and charset-like parameter names.
- Added the missing leading-newline separator case.
- Represented untyped nil content as unavailable.
- Covered the four smaller unpinned branches.
- Renamed the structured-output test and narrowed the PR title to explicitly describe unstructured results.

I also checked #1473. It remains open and has not landed in main, so there was no temporary guard to remove.

The full workspace build and race tests pass locally. The repository-pinned golangci-lint v2.3.1, formatter diff, and go mod tidy -diff also pass in both modules. All PR checks are now green.

Thank you for the careful review.

ktsoator:
A quick update: #1473 has now been merged. I merged the updated `main` into this branch and reconciled its temporary non-text-only fallback with this PR's rendering approach. The formatter now handles both mixed and non-text-only MCP content while preserving valid empty-text responses. All checks are passing.

Thank you again for pointing me to the related work.

tool/mcptoolset/tool.go
```diff
@@ -151,37 +144,206 @@ func (t *mcpTool) Run(ctx agent.Context, args any) (map[string]any, error) {
...
-	// text is empty because every block was non-text, returning {"output": ""}
-	// would report success while discarding the whole payload, so fail loudly
-	// instead.
-	//
-	// TODO(#1401): remove once non-text blocks are rendered rather than skipped.
-	// That also covers the mixed case this deliberately leaves alone, where a
-	// non-text block accompanies real text and is still dropped silently (#1391).
-	if textResponse.Len() == 0 && droppedNonText {
-		return nil, fmt.Errorf("tool %q returned only non-text content, which is not yet supported", t.name)
+func hasMIMECharsetParameter(value string) bool {
```
karolpiotrowicz, line 330: This helper has no test that calls it, and more to the point, three of its clauses can be deleted without anything going red. There is no `charset*` string anywhere in the package's tests — the only occurrence is the `HasPrefix` on line 329.

The three inputs that would pin them are in the review body. Each needs its own row, because they are three different properties: the `charset*` spelling, the loop reaching past the first parameter, and case-folding the parameter name.

Worth saying that the function is correct on all three today. This is a request for tests, not a bug report.

tool/mcptoolset/tool.go
```diff
@@ -151,37 +144,206 @@ func (t *mcpTool) Run(ctx agent.Context, args any) (map[string]any, error) {
...
 
-		if _, err := textResponse.WriteString(textContent.Text); err != nil {
-			return nil, fmt.Errorf("failed to write text response: %w", err)
+func decodeTextBlob(blob []byte, mimeType string) (string, bool) {
+	if len(blob) == 0 {
+		return "", false
+	}
+
+	mediaType, params, err := mime.ParseMediaType(mimeType)
+	if err != nil && (!errors.Is(err, mime.ErrInvalidMediaParameter) || hasMIMECharsetParameter(mimeType)) {
```
karolpiotrowicz: The charset gate only consults `hasMIMECharsetParameter` when `mime.ParseMediaType` returns an error, and the most natural spelling of the thing it defends against does not produce one.

`mime.ParseMediaType` drops an RFC 2231 `charset*` parameter it cannot resolve and returns `nil` for the error, so the `||` never evaluates and the helper never runs. `params` then has no `charset` key, the switch on line 305 takes the empty-charset branch, and because UTF-16LE of ASCII happens to be valid UTF-8 the blob is emitted as a text body.

<details><summary>Reproduction — <code>go test ./tool/mcptoolset/ -run TestCharsetStar</code></summary>

```go
func TestCharsetStar(t *testing.T) {
	const mt = `text/plain; charset*=utf-16`
	utf16le := []byte{0x68, 0x00, 0x69, 0x00} // "hi" in UTF-16LE

	_, params, err := mime.ParseMediaType(mt)
	t.Logf("params=%#v err=%v", params, err)                          // params=map[] err=<nil>
	t.Logf("hasMIMECharsetParameter=%v", hasMIMECharsetParameter(mt))  // true - but never consulted

	if got, ok := decodeTextBlob(utf16le, mt); ok {
		t.Errorf("emitted a body for a declared UTF-16 charset: %q", got)
	}
}
```

Fails with `emitted a body for a declared UTF-16 charset: "h\x00i\x00"`. The same charset in ordinary syntax, `text/plain; charset=utf-16`, is correctly withheld. Driving it through `mcptoolset.New` → `Tools` → `Run` over an in-memory server gives `{"output":"[MCP embedded resource: ..., mimeType=\"text/plain; charset*=utf-16\"]\nh\u0000i\u0000"}`.

</details>

The consequence is bounded and I would rather say so than overstate it. `utf8.Valid` still runs, so whatever comes out is a valid UTF-8 string the server could equally have put in a `TextContent` block, which both `main` and this branch pass through untouched. Realistic non-UTF-8 documents are still withheld: UTF-16LE of `"héllo"`, anything carrying a byte-order mark, Latin-1 `é`. And it is not a regression, since `main` renders no blob here at all.

The reason to fix it is that line 329 was written for exactly this input and cannot see it. As things stand that clause only fires when the media type carries a *second*, unrelated defect that trips `ErrInvalidMediaParameter`. Deciding on the declared string rather than on `err` would cover both routes at once.

tool/mcptoolset/tool.go
```diff
@@ -151,37 +144,206 @@ func (t *mcpTool) Run(ctx agent.Context, args any) (map[string]any, error) {
...
+}
+
+func formatEmbeddedResource(content *mcp.EmbeddedResource) string {
+	if content == nil || content.Resource == nil {
+		return "[MCP embedded resource: unavailable]"
+	}
+
+	resource := content.Resource
+	attributes := resourceAttributes(resource.URI, resource.MIMEType)
+	if resource.Text != "" {
```
karolpiotrowicz, line 227: A resource carrying both `Text` and `Blob` renders the text and drops the blob with no marker at all. This returns before the `size=` attribute is appended on line 234, so a 1024-byte payload disappears and the call still reports success.

The wire format models this as a text-or-blob union, so it needs a server that does not conform. I checked whether the SDK would even deliver such a thing rather than assuming it: it marshals both fields, transmits both, and the client receives `Text="caption"` with a 1024-byte `Blob` intact. I confirmed that is a real serialization boundary and not a pass-through, by checking pointer identity and by mutating the server-side blob after delivery — the client's copy does not change. So the shape reaches this code.

Appending `size=` whenever `len(Blob) > 0`, before the `Text` branch, would at least make the loss visible.

tool/mcptoolset/tool.go
```diff
@@ -151,37 +144,206 @@ func (t *mcpTool) Run(ctx agent.Context, args any) (map[string]any, error) {
...
+	if err != nil && (!errors.Is(err, mime.ErrInvalidMediaParameter) || hasMIMECharsetParameter(mimeType)) {
+		return "", false
+	}
+	if !isTextMediaType(mediaType) {
+		return "", false
+	}
+
+	charset := strings.ToLower(params["charset"])
+	switch charset {
+	case "", "utf-8", "utf8":
```
karolpiotrowicz, line 314: `us-ascii` is handled here but `ascii` is not, and `ascii` is a registered alias for it. `text/plain; charset=ascii` with an ASCII body comes back as `size=5 bytes` instead of the text.

Slightly inconsistent with the line above, which accepts `utf8` as well as `utf-8` — that one is not registered at all, so the strictness runs the opposite way to the leniency.

karolpiotrowicz (dismissed):
Re-reviewed at the current head ([366f51ec](https://github.com/google/adk-go/commit/366f51ec15ac112f67ae5dec3065bf19c733871d)) against `main` ([a6114727](https://github.com/google/adk-go/commit/a61147276167f6b1b636d863f23a030c0a6babdf)). Two things have changed since my last pass, and I want to be straight about which is which: the merge-order caveat I raised against #1353 is spent now that #1353 has landed, and I am asking for changes this time where I previously did not. The code has not moved a byte. I went back over the new charset helper in more depth than I had before, and what I found there is the reason.

The change itself is right, and I want that on the record before the rest. Rendering non-text blocks instead of erroring on them is the correct fix for #1391, plain-text results come through byte-identical, and it quietly removes a nil-dereference that `main` still carries — the type assertion there succeeds for a typed-nil `*mcp.TextContent` and then dereferences it, on both the success and the error paths. Neither PR description mentions that.

## Two things to fix before merge

Both are in the new charset helper: the clauses that decide whether a blob is safe to render as text are untested, and the gate that calls them has a hole. They are the same story from two ends.

**Three clauses of `hasMIMECharsetParameter` have no test that can fail on them.** This one is about the tests, not about the code — the function as written is correct on all three inputs below. The problem is that you can delete any of the three clauses today and the entire suite still passes.

- Deleting `|| strings.HasPrefix(name, "charset*")` from [line 329](https://github.com/google/adk-go/blob/366f51ec15ac112f67ae5dec3065bf19c733871d/tool/mcptoolset/tool.go#L329) — nothing fails. The input that would catch it is `text/plain; charset*=UTF-16''x; name="a`, withheld today only because of that clause.
- Making the loop at [line 324](https://github.com/google/adk-go/blob/366f51ec15ac112f67ae5dec3065bf19c733871d/tool/mcptoolset/tool.go#L324) stop after the first parameter — nothing fails. `text/plain; name=my file.txt; charset=utf-16` is withheld only because the loop reaches the second parameter.
- Dropping the `strings.ToLower` at [line 328](https://github.com/google/adk-go/blob/366f51ec15ac112f67ae5dec3065bf19c733871d/tool/mcptoolset/tool.go#L328) — nothing fails. `text/plain; CHARSET=utf-16; name="a` needs it.

These are three separate properties, so it takes three table rows rather than one. Right now the whole helper is held up by the single `malformed charset is represented by metadata` case at [tool_test.go:146](https://github.com/google/adk-go/blob/366f51ec15ac112f67ae5dec3065bf19c733871d/tool/mcptoolset/tool_test.go#L146) — that one row is what makes five different ways of breaking the helper fail. It is a good test. It is carrying too much alone, on code that arrived in the conflict resolutions and so has had the least review behind it of anything in this PR.

**The `charset*` spelling gets past the gate entirely**, which is the other half of the same story and the one I would fix first. Details are on [line 297](https://github.com/google/adk-go/blob/366f51ec15ac112f67ae5dec3065bf19c733871d/tool/mcptoolset/tool.go#L297).

## `Fixes #1391` will close an issue with one third of it still live

The issue names three losses. Two are fixed here. The third — "when `res.StructuredContent` is non-nil the content blocks are ignored entirely" — is still there at [tool.go:141](https://github.com/google/adk-go/blob/366f51ec15ac112f67ae5dec3065bf19c733871d/tool/mcptoolset/tool.go#L141-L145), unchanged from `main`.

Your reasoning for leaving structured output on its existing path is sound as far as the duplicated JSON text block goes, and dropping the `content` key answers the question I raised last round, so I am not asking you to reverse it. Two things make it worth a decision rather than silence. A tool registered the ordinary way with a typed Go output struct makes the SDK populate `structuredContent` by itself, and in that case the embedded file body is dropped for precisely the GitHub-server shape the issue describes. And `Fixes` auto-closes on merge. Either narrow the keyword or leave a note on the issue.

## The deleted test leaves no wire-level coverage of non-text content

Removing `TestMCPTool_NonTextOnlyResponse` was pre-sanctioned — its own `TODO(#1401)` comment says the test goes away with the guard — and four of its five cases asserted an error you now correctly do not return, so it could not have been kept as written.

What it leaves behind is that `set_test.go` no longer mentions `ImageContent`, `AudioContent`, `ResourceLink` or `EmbeddedResource` anywhere, and every non-text case now runs through the `fixedResultMCPClient` mock in `tool_test.go`. Nothing exercises the SDK decoding a wire block into the concrete types the switch matches on. If a future SDK change decoded one of them differently, everything would land in the `default` arm as `[MCP content: unsupported]` with all 33 table cases still green. One end-to-end case would cover it. Not blocking.

## Smaller things

- The helper splits on `;` without respecting quoted values, so `text/plain; name="foo;charset=utf-16" invalid` finds a `charset` that is not really a parameter and withholds a body it should render. Safe direction, still wrong.
- `us-ascii` is handled but the registered alias `ascii` is not, so `text/plain; charset=ascii` comes back as `size=N bytes`. Same shape as accepting `utf8` alongside `utf-8` — worth being consistent about. This is on [line 306](https://github.com/google/adk-go/blob/366f51ec15ac112f67ae5dec3065bf19c733871d/tool/mcptoolset/tool.go#L306).
- `application/x-www-form-urlencoded` is textual but is not in the allow-list at [tool.go:340-346](https://github.com/google/adk-go/blob/366f51ec15ac112f67ae5dec3065bf19c733871d/tool/mcptoolset/tool.go#L340-L346).

One thing I checked and want to withdraw, because I raised it last round: I said that declaring `charset=utf-8` beside a malformed sibling parameter wrongly withholds a valid body. It does withhold it, but that is the only correct behavior available — `mime.ParseMediaType` returns a nil parameter map alongside `ErrInvalidMediaParameter`, so at [line 304](https://github.com/google/adk-go/blob/366f51ec15ac112f67ae5dec3065bf19c733871d/tool/mcptoolset/tool.go#L304) the code cannot see the declared charset at all and would otherwise apply the UTF-8 rule to bytes whose encoding it never read. Both halves are already pinned by adjacent cases you wrote. Disregard that one.

---

# Inline comments

## `tool/mcptoolset/tool.go:322`

This helper has no test that calls it, and more to the point, three of its clauses can be deleted without anything going red. There is no `charset*` string anywhere in the package's tests — the only occurrence is the `HasPrefix` on line 329.

The three inputs that would pin them are in the review body. Each needs its own row, because they are three different properties: the `charset*` spelling, the loop reaching past the first parameter, and case-folding the parameter name.

Worth saying that the function is correct on all three today. This is a request for tests, not a bug report.

## `tool/mcptoolset/tool.go:297`

The charset gate only consults `hasMIMECharsetParameter` when `mime.ParseMediaType` returns an error, and the most natural spelling of the thing it defends against does not produce one.

`mime.ParseMediaType` drops an RFC 2231 `charset*` parameter it cannot resolve and returns `nil` for the error, so the `||` never evaluates and the helper never runs. `params` then has no `charset` key, the switch on line 305 takes the empty-charset branch, and because UTF-16LE of ASCII happens to be valid UTF-8 the blob is emitted as a text body.

<details><summary>Reproduction — <code>go test ./tool/mcptoolset/ -run TestCharsetStar</code></summary>

```go
func TestCharsetStar(t *testing.T) {
	const mt = `text/plain; charset*=utf-16`
	utf16le := []byte{0x68, 0x00, 0x69, 0x00} // "hi" in UTF-16LE

	_, params, err := mime.ParseMediaType(mt)
	t.Logf("params=%#v err=%v", params, err)                          // params=map[] err=<nil>
	t.Logf("hasMIMECharsetParameter=%v", hasMIMECharsetParameter(mt))  // true - but never consulted

	if got, ok := decodeTextBlob(utf16le, mt); ok {
		t.Errorf("emitted a body for a declared UTF-16 charset: %q", got)
	}
}
```

Fails with `emitted a body for a declared UTF-16 charset: "h\x00i\x00"`. The same charset in ordinary syntax, `text/plain; charset=utf-16`, is correctly withheld. Driving it through `mcptoolset.New` → `Tools` → `Run` over an in-memory server gives `{"output":"[MCP embedded resource: ..., mimeType=\"text/plain; charset*=utf-16\"]\nh\u0000i\u0000"}`.

</details>

The consequence is bounded and I would rather say so than overstate it. `utf8.Valid` still runs, so whatever comes out is a valid UTF-8 string the server could equally have put in a `TextContent` block, which both `main` and this branch pass through untouched. Realistic non-UTF-8 documents are still withheld: UTF-16LE of `"héllo"`, anything carrying a byte-order mark, Latin-1 `é`. And it is not a regression, since `main` renders no blob here at all.

The reason to fix it is that line 329 was written for exactly this input and cannot see it. As things stand that clause only fires when the media type carries a *second*, unrelated defect that trips `ErrInvalidMediaParameter`. Deciding on the declared string rather than on `err` would cover both routes at once.

## `tool/mcptoolset/tool.go:227`

A resource carrying both `Text` and `Blob` renders the text and drops the blob with no marker at all. This returns before the `size=` attribute is appended on line 234, so a 1024-byte payload disappears and the call still reports success.

The wire format models this as a text-or-blob union, so it needs a server that does not conform. I checked whether the SDK would even deliver such a thing rather than assuming it: it marshals both fields, transmits both, and the client receives `Text="caption"` with a 1024-byte `Blob` intact. I confirmed that is a real serialization boundary and not a pass-through, by checking pointer identity and by mutating the server-side blob after delivery — the client's copy does not change. So the shape reaches this code.

Appending `size=` whenever `len(Blob) > 0`, before the `Text` branch, would at least make the loss visible.

## `tool/mcptoolset/tool.go:306`

`us-ascii` is handled here but `ascii` is not, and `ascii` is a registered alias for it. `text/plain; charset=ascii` with an ASCII body comes back as `size=5 bytes` instead of the text.

Slightly inconsistent with the line above, which accepts `utf8` as well as `utf-8` — that one is not registered at all, so the strictness runs the opposite way to the leniency.

tool/mcptoolset/tool.go
ktsoator: Added separate test cases for `charset*`, a charset after an invalid first parameter, and uppercase `CHARSET`. I also added a negative case for charset-like text inside a quoted value.

tool/mcptoolset/tool.go
ktsoator: Fixed in `f7dcb7b`. `decodeTextBlob` now checks for a declared charset independently of the `ParseMediaType` error. If the declaration is not preserved in the parsed parameters, the blob is kept as metadata. The `charset*=utf-16` case is covered with UTF-16LE bytes.

tool/mcptoolset/tool.go
ktsoator: Updated `formatEmbeddedResource` to include the blob size before returning the text when both fields are present. Added a regression case for this shape.

tool/mcptoolset/tool.go
ktsoator: Added `ascii` alongside `us-ascii` with regression coverage. I also added `application/x-www-form-urlencoded` to the textual MIME allow-list.

ktsoator:
Thanks for the careful second pass. I pushed `f7dcb7b` with the requested changes:

- Fixed the `charset*` parsing gap and added separate tests for the charset helper branches.
- Avoided treating semicolons inside quoted MIME values as parameter separators.
- Added handling for the `ascii` alias, form-encoded content, and resources containing both text and blob data.
- Added an in-memory MCP round-trip test covering embedded resources, resource links, images, and audio.
- Changed the PR link from `Fixes #1391` to `Addresses #1391` and documented the remaining structured-content case.

The full local checks and CI are passing. When you have time, could you take another look?

tool/mcptoolset/tool.go
```diff
@@ -151,37 +144,236 @@ func (t *mcpTool) Run(ctx agent.Context, args any) (map[string]any, error) {
...
+
+	var result strings.Builder
+	var previous *formattedMCPContent
+	for i := range formatted {
+		block := &formatted[i]
+		if block.text == "" {
 			continue
 		}
+		if previous != nil && (!previous.isPlain || !block.isPlain) &&
+			!strings.HasSuffix(previous.text, "\n") && !strings.HasPrefix(block.text, "\n") {
```
karolpiotrowicz, line 211: Relaxing `HasSuffix` to `Contains` here passes the whole suite, because every multi-block case either has a `previous` with no newline at all or one whose newline is trailing.

The case that would catch it is an embedded resource whose decoded body is multi-line and does not end in a newline, followed by another block — say a body of `line1\nline2` then a `TextContent` of `tail`. Today that renders `...]\nline1\nline2\ntail`. With `Contains` the separator is suppressed and `tail` is glued onto `line2`.

tool/mcptoolset/tool.go
```diff
@@ -151,37 +144,236 @@ func (t *mcpTool) Run(ctx agent.Context, args any) (map[string]any, error) {
...
+	for _, content := range contents {
+		block := formattedMCPContent{isPlain: true}
+		switch content := content.(type) {
+		case nil:
+			block.text = "[MCP content: unavailable]"
+			block.isPlain = false
+		case *mcp.TextContent:
+			if content == nil {
+				block.text = "[MCP text content: unavailable]"
+				block.isPlain = false
```
karolpiotrowicz, line 172: This `isPlain = false`, and the one on line 168 for the untyped-nil arm, can each be deleted without anything going red.

Both existing cases hide it. The `case nil` test has a single block, so the separator branch is never evaluated at all, and the typed-nil `*mcp.TextContent` case is followed by a `(*mcp.EmbeddedResource)(nil)` whose own `isPlain` already satisfies the condition on line 210. The inputs that would pin them are `[]mcp.Content{nil, &mcp.TextContent{Text: "after"}}` and `[]mcp.Content{(*mcp.TextContent)(nil), &mcp.TextContent{Text: "after"}}`, each wanting the label and `after` on separate lines.

The equivalent assignment on the `*mcp.ResourceLink` arm is covered, so this is just the two nil arms.

tool/mcptoolset/tool.go
```diff
@@ -151,37 +144,236 @@ func (t *mcpTool) Run(ctx agent.Context, args any) (map[string]any, error) {
...
+
+func formatContentLabel(kind string, attributes []string) string {
+	if len(attributes) == 0 {
+		return "[MCP " + kind + "]"
+	}
+	return "[MCP " + kind + ": " + strings.Join(attributes, ", ") + "]"
+}
+
+func decodeTextBlob(blob []byte, mimeType string) (string, bool) {
+	if len(blob) == 0 {
```
karolpiotrowicz, line 295: This guard is never reached by a test. The `empty blob is represented by metadata without size` case uses `application/pdf`, which `isTextMediaType` rejects on line 304 before the blob length matters.

Deleting the guard entirely leaves the suite green. The observable difference is one trailing newline on an empty-blob resource with a textual media type, and nothing at all when another block follows, so this is small — but it means the line is currently free to remove.

karolpiotrowicz (dismissed):
Re-reviewed at [ef1d89d2](https://github.com/google/adk-go/commit/ef1d89d2b1e8edc27f3ea890cdec09801a1d94ff) against `main` ([a6114727](https://github.com/google/adk-go/commit/a61147276167f6b1b636d863f23a030c0a6babdf)). The merge commit is a clean mechanical merge, so [f7dcb7b5](https://github.com/google/adk-go/commit/f7dcb7b5bc9a7715defa3bc35165d4c6844e2285) is the whole of the new work.

Both things I asked for last round are done, and they are pinned by tests that would actually catch a regression — I checked that by deleting each clause and confirming a named case in your new `TestHasMIMECharsetParameter` goes red. `charset*=utf-16`, `charset*=UTF-16''x` and `CHARSET*=utf-16` now all withhold the body where they previously emitted it, driven end to end through `New` → `Tools` → `Run` over an in-memory server. The `Text`-plus-`Blob` case reports `size=1024 bytes`. The `ascii` alias, `application/x-www-form-urlencoded` and the `Addresses #1391` change are all in and covered.

One new problem came with the rewrite, and it breaks the same property the rewrite was there to protect. Three test-coverage notes are on the lines they concern.

## An unterminated quote makes the helper miss a charset that is really there

Making the scanner quote-aware was the right response to my nit about `;` inside quoted values. What it also did was change the answer on media types where the quoting is *malformed*, and there the new answer is unsafe.

When a quote opens and never closes, [`inQuote` stays true to the end of the scan](https://github.com/google/adk-go/blob/ef1d89d2b1e8edc27f3ea890cdec09801a1d94ff/tool/mcptoolset/tool.go#L346-L351), every later `;` is treated as text inside that quoted value, and a real `charset=` after it is never examined. The function returns `false`. `mime.ParseMediaType` independently fails with `ErrInvalidMediaParameter` and hands back a nil parameter map, so [the carve-out on line 301](https://github.com/google/adk-go/blob/ef1d89d2b1e8edc27f3ea890cdec09801a1d94ff/tool/mcptoolset/tool.go#L301) lets the value through, [the gate on 309](https://github.com/google/adk-go/blob/ef1d89d2b1e8edc27f3ea890cdec09801a1d94ff/tool/mcptoolset/tool.go#L309-L310) does not fire either, and the switch takes the no-charset-declared arm where `utf8.Valid` is the only remaining check.

<details><summary>Reproduction — <code>go test ./tool/mcptoolset/ -run TestUnterminatedQuote</code></summary>

```go
func TestUnterminatedQuote(t *testing.T) {
	// "hi" in UTF-16LE, which is also valid UTF-8, so utf8.Valid is not what withholds.
	utf16le := []byte{0x68, 0x00, 0x69, 0x00}
	for _, mt := range []string{
		`text/plain; name="a; charset=utf-16`,
		`text/plain; name="a\"; charset=utf-16`,
		`text/plain; x="unclosed; charset=UTF-16LE`,
		`text/plain; name="; charset=utf-16`,
		`text/plain;name="a;charset=utf-16`,
	} {
		if got, ok := decodeTextBlob(utf16le, mt); ok {
			t.Errorf("%s: emitted a body for a declared UTF-16 charset: %q", mt, got)
		}
	}
}
```

All five fail. Through the public path the first renders `[MCP embedded resource: uri="file:///x.txt", mimeType="text/plain; name=\"a; charset=utf-16\"]\nh\x00i\x00`. At [366f51ec](https://github.com/google/adk-go/commit/366f51ec15ac112f67ae5dec3065bf19c733871d) all five withheld and reported `size=4 bytes`.

</details>

The property to fix against, rather than these five strings: **`hasMIMECharsetParameter` is a safety predicate whose only sound mistake is a false positive.** Its caller reads `false` as "no charset was declared" and unlocks the permissive arm on that basis, so the function must never answer `false` when the real answer is "I cannot tell". Making it precise on input that has no valid parse is the wrong direction, because on such input "is this `charset` inside a quoted value?" has no determinate answer. The order-dependence shows the same thing from another angle: `text/plain; CHARSET=utf-16; name="a` withholds today and `text/plain; name="a; charset=utf-16` emits, with the same declared charset and the same single stray quote.

To be clear about what this is not. The bytes that get through are valid UTF-8 by construction, the label still carries the full media type including `charset=utf-16`, and a server can already put any valid UTF-8 into the output through an ordinary `TextContent` block on `main` and on this branch alike. The result is garbled text reaching the model, not a new way to inject anything. It is also not a regression against `main`, which renders no blob here at all.

One caution if you take the obvious fix. Returning `true` when the scan ends with `inQuote` still set does close all five rows, and it leaves `text/plain; name="foo;charset=utf-16" invalid` rendering, so the nit I raised last round stays fixed. Applied on its own, though, it costs you a test: `TestHasMIMECharsetParameter/uppercase_charset_parameter` uses `text/plain; CHARSET=utf-16; name="a`, which would then return `true` through the new guard whether or not [line 357](https://github.com/google/adk-go/blob/ef1d89d2b1e8edc27f3ea890cdec09801a1d94ff/tool/mcptoolset/tool.go#L357) lower-cases the name. Delete the `strings.ToLower` afterwards and nothing goes red. The new guard line itself would also have no test. Worth landing the fix with a case that pins each part.

## Four clauses of the new scanner have no test that can fail on them

Each of these can be deleted with the whole suite still green, and each deletion makes the code emit bytes it currently withholds — the same shape as the problem above, one step removed. The code is right on all four inputs today.

- The quote toggle on line 347 never has to *leave* quote mode for any current test to pass.
- The `charset*` prefix check only ever sees a lower-case spelling, so `CHARSET*=utf-16` is unpinned — and that is one of the rows this PR newly fixed.
- `charset*` is only ever tested in the first parameter position, so `text/plain; name=x; charset*=utf-16` is unpinned.
- The `inQuote &&` conjunct on the backslash arm is untested either way.

Two rows in [`TestHasMIMECharsetParameter`](https://github.com/google/adk-go/blob/ef1d89d2b1e8edc27f3ea890cdec09801a1d94ff/tool/mcptoolset/tool_test.go#L461-L496) cover all four: `text/plain; name="a"; CHARSET*=UTF-16` wanting `true`, and `text/plain; \; charset=utf-16` wanting `true`.

tool/mcptoolset/tool.go
ktsoator: Added a multiline embedded body followed by text. This now verifies that the following block stays on a separate line.

tool/mcptoolset/tool.go
ktsoator: Added separate cases for untyped nil and typed-nil text content followed by plain text, with both expecting the next block on a new line.

tool/mcptoolset/tool.go
ktsoator: Added a textual embedded resource with an empty blob and verified that it renders only the metadata label without a trailing newline.

ktsoator:
Thanks for the detailed review. I changed the scanner to treat an unterminated quoted parameter as ambiguous and withhold the blob body. I also added focused tests for that case, uppercase and non-leading `charset*` parameters, backslashes outside quotes, and the three coverage gaps noted inline.

karolpiotrowicz (dismissed):
Re-reviewed at [4218b854](https://github.com/google/adk-go/commit/4218b854edf46053da2fc3a9401e0decdc114fcd) against `main` ([53a0bc0c](https://github.com/google/adk-go/commit/53a0bc0ceb2e4437a075631f33a76f039445745f)). The merge commit is a clean mechanical merge, so [4b5b7b71](https://github.com/google/adk-go/commit/4b5b7b71674c544a044e7e2061b45549cc7965f1) is the whole of the new work.

Everything I asked for last round is done, and done properly. All five unterminated-quote media types now report `size=4 bytes` instead of emitting the UTF-16 body, driven end to end through `New` → `Tools` → `Run` over an in-memory server. I checked each new test actually bites by deleting the clause it covers and confirming a named case goes red — that holds for the quote toggle, the `charset*` prefix, the backslash conjunct, the separator, both nil arms and the empty-blob guard.

Worth calling out specifically: you caught the trap I flagged at the end of last round. Changing `uppercase charset parameter` from `text/plain; CHARSET=utf-16; name="a` to `text/plain; CHARSET=utf-16; name=a` is what keeps that row testing the case-folding rather than passing through the new guard. Nothing in the suite would have gone red if you had skipped it.

One class is still open. It is narrower than last round's and it is not something the new commit introduced — it is the half of the same property that `return inQuote` cannot reach.

## A quote in a position MIME does not allow still swallows a real charset

`return inQuote` closes the case where a quote opens and never closes. It cannot help when the quote count is even, because [`inQuote` is false by the time the scan ends](https://github.com/google/adk-go/blob/4218b854edf46053da2fc3a9401e0decdc114fcd/tool/mcptoolset/tool.go#L361-L363).

The gap is that [the toggle at line 347](https://github.com/google/adk-go/blob/4218b854edf46053da2fc3a9401e0decdc114fcd/tool/mcptoolset/tool.go#L346-L347) fires on any `"`, including one that appears mid-token where RFC 2045 does not permit a quoted-string to open. That fabricates a quoted region, [the `|| inQuote` guard on line 351](https://github.com/google/adk-go/blob/4218b854edf46053da2fc3a9401e0decdc114fcd/tool/mcptoolset/tool.go#L351) then skips every `;` inside it, and a genuine top-level `charset=` is never examined. A second stray quote closes the region, so the function returns `false`, [the carve-out on line 301](https://github.com/google/adk-go/blob/4218b854edf46053da2fc3a9401e0decdc114fcd/tool/mcptoolset/tool.go#L299-L301) lets the value through, and the switch takes the no-charset-declared arm where `utf8.Valid` is the only remaining check.

<details><summary>Reproduction — <code>go test ./tool/mcptoolset/ -run TestQuotePosition</code></summary>

```go
func TestQuotePosition(t *testing.T) {
	// "hi" in UTF-16LE, which is also valid UTF-8, so utf8.Valid is not what withholds.
	utf16le := []byte{0x68, 0x00, 0x69, 0x00}
	for _, mt := range []string{
		`text/plain; name=a"; charset=utf-16; x="b`,
		`text/plain; "charset"=utf-16`,
	} {
		if got, ok := decodeTextBlob(utf16le, mt); ok {
			t.Errorf("%s: emitted a body for a declared UTF-16 charset: %q", mt, got)
		}
	}
	// Must keep rendering: the quotes are balanced and open at value start, so the
	// charset text really is inside a quoted value and nothing is declared.
	if _, ok := decodeTextBlob([]byte("hello"), `text/plain; name="foo;charset=utf-16" invalid`); !ok {
		t.Error("regressed the balanced-quote case, which must still render")
	}
}
```

Both fail. Through the public path the first renders `[MCP embedded resource: uri="file:///x.txt", mimeType="text/plain; name=a\"; charset=utf-16; x=\"b"]\nh\x00i\x00`. At [366f51ec](https://github.com/google/adk-go/commit/366f51ec15ac112f67ae5dec3065bf19c733871d), before the scanner became quote-aware, that first case withheld and reported `size=4 bytes`.

</details>

The property to fix against, rather than those two strings: **a `"` outside a quoted value, in a position where a quoted-string cannot legally open, means the parameter list has no valid parse — and on input with no valid parse the honest answer is "cannot determine", which must not be reported as "no charset declared".**

I want to head off the obvious objection, because it is a good one and I spent a while on it. You could argue that `text/plain; name=a"; charset=utf-16; x="b` has no valid parse at all, so nothing in it declares anything — which is exactly why `text/plain; name="foo;charset=utf-16" invalid` renders. Your own test settles it against that reading. [`charset after malformed parameter`](https://github.com/google/adk-go/blob/4218b854edf46053da2fc3a9401e0decdc114fcd/tool/mcptoolset/tool_test.go#L525-L529) asserts `text/plain; name=my file.txt; charset=utf-16` is `true`, and that string is equally unparseable — a space is not valid in a token — and contains no quotes at all, so the `true` comes from the top-level name match. Pair it with [`text blob with malformed unrelated parameter is decoded`](https://github.com/google/adk-go/blob/4218b854edf46053da2fc3a9401e0decdc114fcd/tool/mcptoolset/tool_test.go#L243-L254), which is the same junk without the charset and renders. The suite already takes the position that a top-level `charset=` inside an unparseable list counts as declared. The failing case is that shape with a `"` as the junk instead of a space.

The balanced-quote case stays separable, which is the part worth knowing before you pick an approach. The boundary is not valid parse against invalid parse — it is where the first parse error sits relative to the charset text. In the case that must render, a left-to-right scan reaches a well-formed `name="..."` and consumes the charset text inside it, and the error is the trailing ` invalid` after it. In the failing case the error is the `"` in `name=a"`, before the charset text, and putting that text inside a value means inventing a quoted region that could not have opened there.

I tried a rule along those lines — a `"` opens a value only when the previous non-whitespace byte in that parameter is `=`, and any other unquoted `"` returns `true` — and it does hold up: the suite stays green with no existing test changed, both cases above withhold, the balanced-quote case still renders, and all five of last round's cases still withhold. I am deliberately not handing you the patch, because it adds a helper and a `return` that no test covers, which is the same shape as the trap you just avoided. Treat it only as evidence that a small rule satisfying both sides exists.

To be clear about what this is not. The bytes that get through are valid UTF-8 by construction, the label still carries the full media type, and a server can already put any valid UTF-8 in front of the model through an ordinary `TextContent` block. This is garbled text, not a new way to inject anything, and it is not a regression against `main`, which renders no blob here at all.

Two further notes are on the lines they concern.

Unrelated to the code: `lint (.)` is red on this head, but it fails inside golangci-lint's package loader rather than reporting a diagnostic, and the same pinned version over the whole module at this commit is clean. It passed on your commit and fails only on the merge commit on top of it, so a re-run should clear it.

ktsoator:
Thanks for the examples. I added a check so quotes can only open at the start of a parameter value. Both reported cases now return metadata with `size=4 bytes`, and the balanced-quote case still renders. I also added tests for the surrounding cases and checked that removing the new checks makes the relevant tests fail. The full build, race tests, lint, and tidy checks pass locally.

ktsoator:
I found a few related cases while checking the fix: empty or invalid parameter names, and CR/LF inside quoted values, including after a backslash. I added validation using Go's MIME parser and regression tests for these cases. They now return metadata without decoding the blob, while the existing balanced-quote case still renders. The full build, race tests, lint, and tidy checks pass locally. Does this approach look reasonable to you?

karolpiotrowicz:
This closes it. Both quote findings from the last round are fixed, and it does not cost the cases that should still render.

Executed against the head, the input from last round plus the two that should keep emitting:

```
text/plain; name=a"; charset=utf-16; x="b   withhold  (was: emitted the blob)
text/plain; "charset"=utf-16                withhold  (was: emitted)
text/plain; n1="a; charset=utf-16; n2="b    emit      unchanged, correct
text/plain; x="; charset=utf-16; "y=1       emit      unchanged, correct
text/plain; charset-note=info               emit      unchanged, correct
```

I also ran a 969-row corpus of media types built from nine parameter names, twelve value shapes and three separator spellings, with `mime.ParseMediaType` as the oracle, and diffed the result against the previous head:

```
rows compared: 969   behaviour changed on: 51
  withhold -> emit (leak direction):  0
  emit -> withhold (tighten):        51
  of those, conforming types with no charset that newly withhold: 0
```

Every one of the 51 changes is in the fail-closed direction, and none of them costs a conforming type its blob. That last number is the one I was most worried about, because the obvious one-line version of this fix widens the guard and starts withholding on well-formed types like `charset-note=info`. Validating the complete quoted parameter with `mime.ParseMediaType` on the closing quote, and requiring the opening quote to sit at value start, gets both directions at once.

I will clear my stale review so this is not left gated by a finding that no longer exists.

ktsoator:
Thanks for the thorough review and for checking all those cases. I appreciate your help with this PR.

karolpiotrowicz (approved):
Approving. The remaining quote-handling finding is closed and verified, as covered in the comment above, and the merge of `main` ([b163eb96](https://github.com/google/adk-go/commit/b163eb96a45f9d2ca51a2d88728e2c69ee0be320)) since then left the three changed files byte-identical to the revision I checked.

PR #1403, opened 2026-08-25
fix(memory): tokenize on punctuation and all whitespace in extractWords

### LINKED ISSUE #569: Bug: Memory search fails to match words with punctuation in extractWords

Ajithkumar003-dev:
## What

`extractWords` in `memory/inmemory.go` now splits text on runs of non-word runes (`strings.FieldsFunc`) instead of on the literal space character, so punctuation and non-space whitespace no longer become part of an indexed token.

Fixes #569

## Why

`extractWords` builds the keyword index for `InMemoryService` and also tokenizes the incoming query, so both sides were affected:

1. **Punctuation stayed glued to the word.** `"The agent works great!"` indexed the token `great!`, so `SearchMemory` with query `great` returned nothing.
2. **Tabs and newlines were never split on.** `"first line\nsecond\tline"` indexed the single token `line\nsecond\tline`, making every word after the first newline unreachable. Multi-line model responses are the common case here.
3. **Punctuated queries failed symmetrically** — a query of `"Why timed-out?"` produced the token `timed-out?`, which matched nothing.

## Implementation note

The issue proposed `strings.Fields` + `strings.TrimFunc`. I used `strings.FieldsFunc` with a non-word separator predicate instead, for two reasons:

- `TrimFunc` only strips punctuation from the **ends** of a token, so the comma-separated case the issue lists as affected would still break: `regions: us-east1,us-west1` stays one token. `FieldsFunc` handles it.
- It mirrors adk-python's tokenizer, `re.findall(r'\w+', text)` in `in_memory_memory_service.py`, keeping the two implementations aligned per the Alignment with adk-python guidance in CONTRIBUTING.md. The predicate accepts Unicode letters, Unicode digits and `_`, matching Python 3's `\w`.

Happy to switch to the `TrimFunc` approach if maintainers prefer to keep hyphenated and comma-joined tokens intact.

`FieldsFunc` never yields empty strings, so the old `if s == "" { continue }` guard is now dead code and was removed.

## Testing plan

Three cases added to the existing `Test_inMemoryService_SearchMemory` table, one per failure mode above:

- `find events next to punctuation` — covers `great!` and `regions: us-east1,us-west1`
- `find events separated by non-space whitespace` — covers `\n` and `\t`
- `find events for a query containing punctuation` — covers query-side normalization

Each was verified to **fail against the unfixed `extractWords`** before the fix was applied, so they are genuine regression tests rather than tests written to match new behavior:

```
--- FAIL: Test_inMemoryService_SearchMemory
    --- FAIL: Test_inMemoryService_SearchMemory/find_events_next_to_punctuation
    --- FAIL: Test_inMemoryService_SearchMemory/find_events_separated_by_non-space_whitespace
    --- FAIL: Test_inMemoryService_SearchMemory/find_events_for_a_query_containing_punctuation
```

With the fix applied, the full package passes, including the pre-existing `find events` and concurrency tests:

```
--- PASS: Test_inMemoryService_SearchMemory (0.00s)
    --- PASS: Test_inMemoryService_SearchMemory/find_events (0.00s)
    --- PASS: Test_inMemoryService_SearchMemory/find_events_next_to_punctuation (0.00s)
    --- PASS: Test_inMemoryService_SearchMemory/find_events_separated_by_non-space_whitespace (0.00s)
    --- PASS: Test_inMemoryService_SearchMemory/find_events_for_a_query_containing_punctuation (0.00s)
    --- PASS: Test_inMemoryService_SearchMemory/no_leakage_for_different_appName (0.00s)
    --- PASS: Test_inMemoryService_SearchMemory/no_leakage_for_different_user (0.00s)
    --- PASS: Test_inMemoryService_SearchMemory/no_matches (0.00s)
    --- PASS: Test_inMemoryService_SearchMemory/lookup_on_empty_store (0.00s)
--- PASS: Test_inMemoryService_SearchMemory_Concurrent (0.00s)
PASS
```

Also run locally: `go build ./...` (whole module, clean), `go vet ./memory/` (clean), `gofmt` (clean). `extractWords` is unexported and referenced only within the `memory` package, so the change is self-contained.

I could not run `go test -race` locally — it requires cgo and there is no C toolchain on my machine. The change introduces no new concurrency, and the existing `Test_inMemoryService_SearchMemory_Concurrent` passes without race enabled; relying on CI for the race run.

### DIFF
--- a/memory/inmemory.go
+++ b/memory/inmemory.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode"
 
 	"google.golang.org/genai"
 
@@ -167,12 +168,19 @@ func checkMapsIntersect(m1, m2 map[string]struct{}) bool {
 func extractWords(text string) map[string]struct{} {
 	res := make(map[string]struct{})
 
-	for s := range strings.SplitSeq(text, " ") {
-		if s == "" {
-			continue
-		}
+	// Split on runs of non-word characters so that punctuation and non-space
+	// whitespace are not folded into a token: without this, "great!" is stored
+	// as-is and a search for "great" misses it. This mirrors the `\w+`
+	// tokenization used by adk-python's in-memory memory service.
+	for _, s := range strings.FieldsFunc(text, isNotWord) {
 		res[strings.ToLower(s)] = struct{}{}
 	}
 
 	return res
+}
+
+// isNotWord reports whether r separates two words, i.e. whether it is neither a
+// letter, a digit nor an underscore.
+func isNotWord(r rune) bool {
+	return !unicode.IsLetter(r) && !unicode.IsNumber(r) && r != '_'
 }
--- a/memory/inmemory_test.go
+++ b/memory/inmemory_test.go
@@ -91,6 +91,102 @@ func Test_inMemoryService_SearchMemory(t *testing.T) {
 			},
 		},
 		{
+			name: "find events next to punctuation",
+			initSessions: []session.Session{
+				makeSession(t, "app1", "user1", "sess1", []*session.Event{
+					{
+						ID:          "event1",
+						Author:      "test-bot",
+						LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("The agent works great!", genai.RoleModel)},
+						Timestamp:   must(time.Parse(time.RFC3339, "2023-10-01T10:00:00Z")),
+					},
+					{
+						ID:          "event2",
+						Author:      "test-bot",
+						LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("regions: us-east1,us-west1", genai.RoleModel)},
+						Timestamp:   must(time.Parse(time.RFC3339, "2023-10-02T10:00:00Z")),
+					},
+				}),
+			},
+			req: &memory.SearchRequest{
+				AppName: "app1",
+				UserID:  "user1",
+				Query:   "great us-west1",
+			},
+			wantResp: &memory.SearchResponse{
+				Memories: []memory.Entry{
+					{
+						ID:        "event1",
+						Content:   genai.NewContentFromText("The agent works great!", genai.RoleModel),
+						Author:    "test-bot",
+						Timestamp: must(time.Parse(time.RFC3339, "2023-10-01T10:00:00Z")),
+					},
+					{
+						ID:        "event2",
+						Content:   genai.NewContentFromText("regions: us-east1,us-west1", genai.RoleModel),
+						Author:    "test-bot",
+						Timestamp: must(time.Parse(time.RFC3339, "2023-10-02T10:00:00Z")),
+					},
+				},
+			},
+		},
+		{
+			name: "find events separated by non-space whitespace",
+			initSessions: []session.Session{
+				makeSession(t, "app1", "user1", "sess1", []*session.Event{
+					{
+						ID:          "event1",
+						Author:      "test-bot",
+						LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("first line\nsecond\tline", genai.RoleModel)},
+						Timestamp:   must(time.Parse(time.RFC3339, "2023-10-01T10:00:00Z")),
+					},
+				}),
+			},
+			req: &memory.SearchRequest{
+				AppName: "app1",
+				UserID:  "user1",
+				Query:   "second",
+			},
+			wantResp: &memory.SearchResponse{
+				Memories: []memory.Entry{
+					{
+						ID:        "event1",
+						Content:   genai.NewContentFromText("first line\nsecond\tline", genai.RoleModel),
+						Author:    "test-bot",
+						Timestamp: must(time.Parse(time.RFC3339, "2023-10-01T10:00:00Z")),
+					},
+				},
+			},
+		},
+		{
+			name: "find events for a query containing punctuation",
+			initSessions: []session.Session{
+				makeSession(t, "app1", "user1", "sess1", []*session.Event{
+					{
+						ID:          "event1",
+						Author:      "test-bot",
+						LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("the deploy step timed out", genai.RoleModel)},
+						Timestamp:   must(time.Parse(time.RFC3339, "2023-10-01T10:00:00Z")),
+					},
+				}),
+			},
+			req: &memory.SearchRequest{
+				AppName: "app1",
+				UserID:  "user1",
+				Query:   "Why timed-out?",
+			},
+			wantResp: &memory.SearchResponse{
+				Memories: []memory.Entry{
+					{
+						ID:        "event1",
+						Content:   genai.NewContentFromText("the deploy step timed out", genai.RoleModel),
+						Author:    "test-bot",
+						Timestamp: must(time.Parse(time.RFC3339, "2023-10-01T10:00:00Z")),
+					},
+				},
+			},
+		},
+		{
 			name: "no leakage for different appName",
 			initSessions: []session.Session{
 				makeSession(t, "app1", "user1", "sess3", []*session.Event{


karolpiotrowicz (commented):
The tokenizer change does what it says. I checked out the branch, ran the test hunk alone against the unfixed `extractWords`, and the three new cases fail there and pass with the fix — genuine regression tests, as your description claims. Build, `go vet`, `golangci-lint`, `staticcheck` and `govulncheck` are all clean, `apidiff` shows no exported API change, and the full suite under `-race -shuffle=on` across both modules fails only the two packages that already fail on `main` for network reasons.

The parity claim also checks out, and it is worth saying so explicitly because the issue thread disagrees with you. A comment on #569 from February says adk-python uses `re.findall(r'[A-Za-z]+', text)`. That is stale — adk-python `main` today uses `re.findall(r'\w+', text)`, with the docstring "Extracts Unicode-aware tokens from a string in lowercase". I compared your token sets against real CPython over 38 inputs and got 37 identical, and a differential fuzz against `[\p{L}\p{N}_]+` ran 15.5M executions with no divergence in the splitting predicate. `unicode.IsNumber` rather than `unicode.IsDigit` is the right call for that parity, since CPython's `\w` admits Nl and No.

## Worth changing before merge

**Use `strings.FieldsFuncSeq` instead of `strings.FieldsFunc`.** `FieldsFunc` returns a slice, so it materializes every token at once, where the `strings.SplitSeq` you replaced was lazy. Benchmarked on the branch:

| input | `FieldsFunc` | old `SplitSeq` | `FieldsFuncSeq` |
|---|---|---|---|
| 8 MiB of `"a,"` | 405,230,133 B/op | 261 B/op | 256 B/op |
| 8 MiB of prose | 165,450,696 B/op | 712 B/op | 712 B/op |
| 2 KB model response | 23,330 B/op | 602 B/op | 1,058 B/op |

Both call sites take unbounded text — the index side tokenizes every model output part, and `preloadmemorytool` passes the raw user turn straight into `SearchMemory` at [tool.go:79](https://github.com/google/adk-go/blob/2f4aa244b28799b9a57b8ee4416b342e49264587/tool/preloadmemorytool/tool.go#L79). `FieldsFuncSeq` landed in Go 1.24 and the module declares 1.26.6, so it is a drop-in. I applied it and re-ran the package: it compiles, the suite stays green, and it restores the `for s := range …` shape the function had before.

## Two behavior changes worth a mention in the description

Neither is a reason not to merge — both match adk-python, and I could not find a third. But someone reading the diff in six months will want them written down.

**Sub-word fragments create new matches between unrelated memories.** Splitting on every non-word rune produces short tokens, and `SearchMemory` returns any event sharing one. Measured through the public API, each case passing on `main` and failing on the branch:

<details><summary>reproduction, seven cases</summary>

```
stored                                   query                base   branch
"The build's output is in bazel-bin."    "What's for lunch?"      0        1   (shared "s")
"I don't know the answer."               "I'm hungry"             0        1   (shared "i")
"We'll deploy tomorrow."                 "I'll eat now"           0        1   (shared "ll")
"मैं ठीक हूँ"                              "क्या हुआ?"                 0        1   (combining marks split it)
"us-east1" and "us-west1" events         "us-west1"               1        2   (shared "us")
"Contact me at alice@example.com"        "bob@example.com"        0        1   (shared "example", "com")
"the deploy step timed out"              "deploy"                 1        1   (control, unchanged)
```
</details>

I checked whether adk-python is protected from this by its `matched_words` scoring and `_MAX_SEARCH_RESULTS` cap, and it is not — running its `search_memory` on those same inputs returns the same memories. What Python does have, at larger stores, is ranking: with 40 memories it puts the three genuinely relevant ones first and truncates the tail, while `SearchMemory` here returns all of them unordered. That gap is in `SearchMemory`, which you have not touched, so it is not this PR's to fix. It is a reasonable follow-up issue now that the tokenizer produces more matches per query.

**An event whose text contains no letter, digit or underscore is no longer stored at all.** `extractWords` returns an empty set for emoji-only, ZWJ-sequence or punctuation-only text, and [inmemory.go:77-79](https://github.com/google/adk-go/blob/2f4aa244b28799b9a57b8ee4416b342e49264587/memory/inmemory.go#L77-L79) then skips the event, so its content never enters the store. On `main`, storing `👍` and querying `👍` returns one memory. On the branch it returns none. adk-python does the same thing, so I read this as intended rather than accidental — but it is the one case where the change removes recall, and nothing in the tests says so.

## Smaller things

**Three clauses of `isNotWord` are unpinned by the suite.** Removing `&& r != '_'`, removing `!unicode.IsNumber(r) &&`, or swapping `IsNumber` for `IsDigit` each leaves `go test ./memory/` green. No test text has an underscore inside a token, a digit-only query, or an Nl/No numeral such as `½`. The "find events next to punctuation" case still matches on `us` and `west` even with digits demoted to separators, which is what lets the second one through. Three short cases would close it, and they are exactly the parity details the change exists to get right.

**The doc comment describes a narrower predicate than the code.** [inmemory.go:182-183](https://github.com/google/adk-go/blob/2f4aa244b28799b9a57b8ee4416b342e49264587/memory/inmemory.go#L182-L183) says "neither a letter, a digit nor an underscore", but `unicode.IsNumber` is category N, so `½` and `Ⅷ` are word characters here too. The code is right and the comment is the half to change.

One process note: `cla/google` is failing, which is why no build or test job has run on the PR. That will need clearing before a maintainer can merge, whatever they decide about the direction.

PR #1404, opened 2026-08-26
feat(adkrest): opt-in /app-info endpoint in rest server

wojtas000:
## Problem

adk-go does not implement `GET /apps/{app_name}/app-info`. It is the last endpoint from the Agents CLI wire contract that Go is missing. Agents CLI needs it in agent evaluation: it is the call that tells the CLI what an
app contains (agents, instructions, tools) before an eval run.

## Solution

We implement `GET /apps/{app_name}/app-info` returning an `AppInfo`: the app name, its root
agent, and a flat map of every agent reachable from that root with its
description, instruction, tools and child names. Nothing is executed - no
session, no model call.

Details worth knowing:

- **Only LLM Agents are included.** Instruction and tools are read through the
  existing internal state seam (`llminternal.Reveal`); agents that
  are not LLM agents are walked by and we search for LLM Agents in their `.Subagents()`
- **Toolsets are resolved best effort**, with a timeout, since an MCP toolset
  may do I/O. A toolset that fails is logged and skipped rather than failing the
  whole request.
- **Tools are reported as function declarations**, matching adk-python: tools
  without a declaration (built-in Gemini tools such as Google Search, and
  request-shaping tools) are omitted.

### Deliberate differences from adk-python

| | adk-python | this PR |
|---|---|---|
| non-LLM root | 400 `"Root agent is not an LlmAgent"` — [`api_server.py#L1410-L1421`](https://github.com/google/adk-python/blob/775c1bd36e205eec65e207ad29fdc4cf2184e47e/src/google/adk/cli/api_server.py#L1410-L1421) | 200, described |
| non-LLM sub-agents | dropped, with their whole subtree — [`agent_info.py#L65-L69`](https://github.com/google/adk-python/blob/775c1bd36e205eec65e207ad29fdc4cf2184e47e/src/google/adk/utils/agent_info.py#L65-L69) (the `isinstance` guard wraps both the recursion and the name append) | not included, but walked through in the agent tree |
| `subAgents` casing | `sub_agents` (snake_case) — [`agent_info.py#L29-L34`](https://github.com/google/adk-python/blob/775c1bd36e205eec65e207ad29fdc4cf2184e47e/src/google/adk/utils/agent_info.py#L29-L34) | `subAgents`, per the contract's "MUST strictly emit camelCase" |
| `isComputerUse` | emitted — [`api_server.py#L636-L642`](https://github.com/google/adk-python/blob/775c1bd36e205eec65e207ad29fdc4cf2184e47e/src/google/adk/cli/api_server.py#L636-L642) | omitted — not in the contract, and Go has no computer-use toolset |
| `language` | `"python"`, hardcoded — [`api_server.py#L1415`](https://github.com/google/adk-python/blob/775c1bd36e205eec65e207ad29fdc4cf2184e47e/src/google/adk/cli/api_server.py#L1415) | `"go"` |

### Testing plan

- `golangci-lint run` — 0 issues, both modules
- `go mod tidy -diff` — clean, both modules

Also checked live against a server serving 5 apps (simple, tooled, nested,
toolset-backed, `SequentialAgent`-rooted) and diffed
field by field against adk-python's `/app-info` for the same apps. Every
remaining difference is one of the deliberate ones listed above.

<details>
<summary>Example response</summary>

`GET /apps/travel_app/app-info` for one root agent with two sub-agents:

```json
{
  "name": "travel_app",
  "rootAgentName": "travel_concierge",
  "description": "Plans trips end to end.",
  "language": "go",
  "agents": {
    "travel_concierge": {
      "name": "travel_concierge",
      "description": "Plans trips end to end.",
      "instruction": "Delegate flight questions and hotel questions to your sub-agents.",
      "tools": [],
      "subAgents": ["flight_agent", "hotel_agent"]
    },
    "flight_agent": {
      "name": "flight_agent",
      "description": "Finds and books flights.",
      "instruction": "Search for flights matching the user's dates.",
      "tools": [
        {
          "functionDeclarations": [
            {
              "name": "search_flights",
              "description": "Finds flights between two airports.",
              "parametersJsonSchema": {
                "type": "object",
                "properties": {
                  "origin": {"type": "string"},
                  "destination": {"type": "string"}
                },
                "required": ["origin", "destination"],
                "additionalProperties": false
              }
            }
          ]
        }
      ]
    },
    "hotel_agent": {
      "name": "hotel_agent",
      "description": "Finds and books hotels.",
      "instruction": "Book a hotel for the requested city and dates.",
      "tools": [
        {
          "functionDeclarations": [
            {
              "name": "book_hotel",
              "description": "Books a hotel in a city for a number of nights.",
              "parametersJsonSchema": {
                "type": "object",
                "properties": {
                  "city": {"type": "string"},
                  "nights": {"type": "integer"}
                },
                "required": ["city", "nights"],
                "additionalProperties": false
              }
            }
          ]
        }
      ]
    }
  }
}
```
</details>


server/adkrest/internal/services/appinfo.go
```diff
@@ -0,0 +1,179 @@
...
+	toolsetCtx := appInfoContext{Context: ctx, appName: appName, agentName: agentName}
+
+	for _, ts := range state.Toolsets {
+		if ts == nil {
+			continue
+		}
+		tsTools, err := ts.Tools(toolsetCtx)
+		if err != nil {
+			log.Printf("app-info: agent %q: skipping toolset %q: %v", agentName, ts.Name(), err)
+			continue
```
wojtas000, line 144: Comment for reviewer: is the `continue` behavior fine on error when loading tools from toolset? Another way would be to throw 500 on the HTTP response (even if only one toolset fails and the rest tools/toolsets succeed when loaded).

server/adkrest/internal/services/appinfo.go
```diff
@@ -0,0 +1,179 @@
...
+	"google.golang.org/adk/v2/internal/llminternal"
+	"google.golang.org/adk/v2/server/adkrest/internal/models"
+	"google.golang.org/adk/v2/session"
+	"google.golang.org/adk/v2/tool"
+)
+
+// toolsetResolveTimeout bounds the time spent resolving an agent's toolsets.
+// Toolsets may reach out over the network (an MCP server, for example), and
+// describing an app must not hang on one.
+const toolsetResolveTimeout = 10 * time.Second
```
wojtas000, line 36: Comment for reviewer: value of this constant was chosen arbitrarily. But 10 seconds for `toolsetResolveTimeout` means that the overall timeout could be 10s * N_Toolsets, where N_Toolsets is the number of toolsets in the whole agent tree. Mabye it should be shortened?

PR #1405, opened 2026-08-26
fix(llminternal): pair every function call with a result in request contents

QuentinBisson:
## Problem

A function call and its result are two separate session events. When a turn ends between them (a restart, an OOM kill, a client disconnect, a cancellation), the session keeps a function call that no function response answers.

That history is replayed on every later turn, so the session does not recover on its own, and a provider that requires strict pairing rejects the whole conversation. The OpenAI Responses path in this repo sends the dangling call as a `function_call` item with no `function_call_output`, which the API rejects. Anthropic, used through a custom model, answers `tool_use ids were found without tool_result blocks immediately after`. Either way the session stays unusable until it is deleted.

No issue exists for the Go side. The same failure is reported against adk-python in google/adk-python#5856, #3971 and #6582.

## Change

Pair every function call with a result while the request contents are assembled, in `buildContentsDefault`. A call the immediately following content does not answer gets a placeholder result there.

- The stored events are untouched, so recorded history stays intact and a session that is already broken heals on its next turn with no migration.
- It sits above the session service, so one implementation covers every store and every provider.
- Ids are read after the conversion, so a call whose client id was stripped is answered by a response with the same stripped id.
- Pairing is checked positionally, against the immediately following content, because that is the invariant the provider enforces.
- On a turn that is already partly answered, the placeholder joins the leading run of responses. Anthropic requires the results to precede any other block in the message, so appending after a trailing text part would be rejected for the same reason the missing result was.

A call the framework holds open (a long-running tool, an approval, a request for user input) is described as awaiting a response, not as having returned nothing. The difference changes what the model does next: told a tool returned nothing, it reissues the call or proceeds without it; told the call is still awaiting a response, it can wait. Those ids come from `LongRunningToolIDs`, which lives on the event and does not survive the conversion to contents.

A conversation whose calls are all answered is returned unchanged.

The mirror change for adk-python is google/adk-python#6914.

## Testing plan

- `go test ./... -count=1` — 158 packages, all pass.
- `go build ./...`, `go vet ./internal/llminternal/`, `gofmt -l`.

`internal/llminternal/tool_pairing_test.go` drives the request processor over eight histories: the interrupted turn at the end of the history and in the middle, the partially answered parallel turn, the same turn with a trailing text part, the call held open in both directions, calls without an id, and a healthy history that must come back unchanged. Each case also asserts the provider invariant, that the content after a call answers every call it carries; a guard test asserts the same check rejects unrepaired input.

Two `TestContentsRequestProcessor_IncludeContents/agentTransfer` fixtures end on an unanswered call. They now expect the placeholder result instead of the synthetic continuation turn.

The placeholder content is user-role, so a history that previously ended on a model turn carrying an unanswered call now ends on a user turn. The synthetic "Continue processing previous requests..." user content is therefore no longer appended in that case; the two agentTransfer fixtures in `contents_processor_test.go` record this.

### DIFF
--- a/internal/llminternal/contents_processor.go
+++ b/internal/llminternal/contents_processor.go
@@ -240,6 +240,10 @@ func buildContentsDefault(agentName, invocationBranch, isolationScope string, ev
 		contents = append(contents, content)
 	}
 
+	// Ids are read after the conversion so a stripped call id is answered by a
+	// response with the same stripped id.
+	contents = pairUnansweredFunctionCalls(contents, pendingCallIDs(filtered))
+
 	// For scoped agents (task / single_turn), prepend a synthetic user
 	// content built from the originating FC's args. The FC lives in an
 	// UNSCOPED parent event (e.g. the coordinator's task-delegation FC)
--- a/internal/llminternal/contents_processor.go
+++ b/internal/llminternal/contents_processor.go
@@ -236,13 +236,17 @@ func buildContentsDefault(agentName, invocationBranch, isolationScope string, ev
 			continue
 		}
 
-		utils.RemoveClientFunctionCallID(content)
 		contents = append(contents, content)
 	}
 
-	// Ids are read after the conversion so a stripped call id is answered by a
-	// response with the same stripped id.
+	// Calls are paired while they still carry the ids ADK generated, because
+	// LongRunningToolIDs holds those ids and a turn with several calls can only
+	// be matched to its results by id. The strip that follows blanks a
+	// synthesized response id together with the call id it answers.
 	contents = pairUnansweredFunctionCalls(contents, pendingCallIDs(filtered))
+	for _, content := range contents {
+		utils.RemoveClientFunctionCallID(content)
+	}
 
 	// For scoped agents (task / single_turn), prepend a synthetic user
 	// content built from the originating FC's args. The FC lives in an
--- a/internal/llminternal/contents_processor_test.go
+++ b/internal/llminternal/contents_processor_test.go
@@ -178,6 +178,12 @@ func TestContentsRequestProcessor_IncludeContents(t *testing.T) {
 					Role: "user",
 				},
 				genai.NewContentFromFunctionCall("func1", nil, "model"),
+				// The turn ended before the tool result was recorded, so the
+				// call reaches the model with a placeholder result.
+				functionResponseContent(&genai.FunctionResponse{
+					Name:     "func1",
+					Response: map[string]any{"result": missingResult},
+				}),
 			},
 		},
 		{
@@ -193,6 +199,12 @@ func TestContentsRequestProcessor_IncludeContents(t *testing.T) {
 					Role: "user",
 				},
 				genai.NewContentFromFunctionCall("func1", nil, "model"),
+				// The turn ended before the tool result was recorded, so the
+				// call reaches the model with a placeholder result.
+				functionResponseContent(&genai.FunctionResponse{
+					Name:     "func1",
+					Response: map[string]any{"result": missingResult},
+				}),
 			},
 		},
 		{
--- a/internal/llminternal/tool_pairing.go
+++ b/internal/llminternal/tool_pairing.go
@@ -1,0 +1,155 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package llminternal
+
+import (
+	"google.golang.org/genai"
+
+	"google.golang.org/adk/v2/internal/utils"
+	"google.golang.org/adk/v2/session"
+)
+
+// missingFunctionResult stands in for a result that was never recorded. It
+// names no cause, because there is none to name: the turn may have been
+// interrupted, or the process may have died between the two events.
+const missingFunctionResult = "No response available for this function call."
+
+// pendingFunctionResult stands in for a call ADK is deliberately holding open:
+// a long-running tool, a human approval, or a request for user input. No
+// function response exists for these until the answer arrives, so the call is
+// pending rather than lost. The distinction changes what the model does next:
+// told a tool returned nothing, it reissues the call or proceeds without it;
+// told the call is still awaiting a response, it can wait.
+const pendingFunctionResult = "This call is awaiting a response and has not completed yet."
+
+// pendingCallIDs returns the ids of the calls ADK is holding open.
+// LongRunningToolIDs lives on the event and does not survive the conversion to
+// contents, so it is read here.
+func pendingCallIDs(events []*session.Event) map[string]bool {
+	pending := make(map[string]bool)
+	for _, event := range events {
+		if event == nil {
+			continue
+		}
+		for _, id := range event.LongRunningToolIDs {
+			pending[id] = true
+		}
+	}
+	return pending
+}
+
+// pairUnansweredFunctionCalls gives every function call a function response in
+// the immediately following content.
+//
+// A call and its result are two separate session events. When a turn ends
+// between them (a restart, an OOM kill, a disconnect, a cancellation), the
+// session keeps a function call that no function response answers. That
+// history is replayed on every later turn, and a provider that requires strict
+// pairing then rejects the whole conversation: Anthropic answers "tool_use ids
+// were found without tool_result blocks immediately after", and the session
+// stays unusable until it is deleted.
+//
+// The repair runs on the request rather than the stored events, so recorded
+// history stays intact and a session that is already broken heals on its next
+// turn without a migration. It also sits above the session service, so it
+// applies to every store.
+//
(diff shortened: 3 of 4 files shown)


karolpiotrowicz (changes requested):
The problem is real and I was able to reproduce it: on `main`, a history whose turn ended between a function call and its result reaches the model with that call unanswered, and with this branch applied it does not. The write-up is unusually clear about why the repair belongs on the request rather than on the stored events, and I agree with that choice.

The repair is placed one step too late in `buildContentsDefault`, though, and on Gemini that costs it both of the properties it is trying to establish. I think this needs a change before it goes in, and I believe the change is small.

## The pairing runs after the function-call IDs have been stripped

[`contents_processor.go:188`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/llminternal/contents_processor.go#L188) blanks every `adk-`-prefixed ID, and [`:194`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/llminternal/contents_processor.go#L194) then runs the pairing. Gemini omits function-call IDs, so ADK generates them — [`base_flow.go:1020`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/llminternal/base_flow.go#L1020) fills the empty ID with `adk-<uuid>`, and [`:1030`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/llminternal/base_flow.go#L1030) records that same generated ID in `LongRunningToolIDs`. [`RemoveClientFunctionCallID`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/utils/utils.go#L55-L60) strips exactly that prefix. By the time the pairing runs, every Gemini-sourced ID in the contents is the empty string while `LongRunningToolIDs` still holds the `adk-` form.

Two things follow.

**The "awaiting a response" placeholder is unreachable on Gemini.** [`tool_pairing.go:118`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/llminternal/tool_pairing.go#L118) looks up `pending[call.ID]` with the blanked ID against a map keyed on the unblanked one, so it never matches and the call gets "No response available for this function call." instead. The branch works only where the provider supplied its own ID — the OpenAI path keeps a real `call_id` ([`response.go:121-123`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/model/openaimodel/response.go#L121-L123)) — which is the family that did not need ADK to generate one. The shape that breaks is pinned by an existing test in the repo: [`long_running_function_test.go:262`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/tool/functiontool/long_running_function_test.go#L262) emits a long-running call with no ID, and [`:321`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/tool/functiontool/long_running_function_test.go#L321) asserts that `LongRunningToolIDs[0]` equals the generated one.

**On a partly answered parallel turn the placeholder lands on the wrong call.** [`unansweredCalls`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/llminternal/tool_pairing.go#L129-L148) consumes matched IDs positionally, so with every ID blank the first call consumes the single empty response ID and is recorded as answered, and the second is recorded as unanswered and gets the placeholder under its own name. A long-running tool sharing a turn with an ordinary one produces exactly that: [`base_flow.go:1226`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/llminternal/base_flow.go#L1226) emits no response event for the long-running call, and the parallel merge drops the empty slot.

<details><summary>Reproduction, driven through <code>ContentsRequestProcessor</code></summary>

```
history:
  user   "do both"
  model  [ search_tool(id=adk-AAAA, long-running), fetch_tool(id=adk-BBBB) ]
  user   [ fetch_tool(id=adk-BBBB) = real result ]

contents reaching the model, with this branch applied:
  content[1] role=model  CALL:search_tool(id=)  CALL:fetch_tool(id=)
  content[2] role=user   RESP:fetch_tool(id=)=<real result>
                         RESP:fetch_tool(id=)=No response available for this function call.

  search_tool, the call that is genuinely outstanding: 0 results
  fetch_tool, which already returned:                  2 results, contradicting each other
```

The same history with provider-supplied IDs (`call_AAAA` / `call_BBBB`) comes out correct, which is what isolates the cause to the strip.
</details>

Both of these go away if the pairing runs before the strip rather than after it. `RemoveClientFunctionCallID` already blanks response IDs as well as call IDs, so a synthesized placeholder is stripped alongside the call it answers and the property your comment at [`:192-193`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/llminternal/contents_processor.go#L192-L193) describes still holds:

```go
    contents = append(contents, content)
}

contents = pairUnansweredFunctionCalls(contents, pendingCallIDs(filtered))
for _, content := range contents {
    utils.RemoveClientFunctionCallID(content)
}
```

I ran that: the parallel-turn case above becomes `[RESP:fetch_tool=<real>, RESP:search_tool="This call is awaiting a response and has not completed yet."]`, all eight of your new cases still pass, and the full suite is unchanged.

## The tests cannot currently see either of those

This is worth fixing alongside the code, because the suite is green today while both behaviors are live.

- Every fixture ID is provider-shaped (`call_1`, `lro_call_1`), so nothing in the package exercises the `adk-` strip that causes the first problem. A case built from an `adk-`-prefixed ID would have failed.
- The two ID-less calls in the no-ID case are both named `search_tool` ([`tool_pairing_test.go:68-69`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/llminternal/tool_pairing_test.go#L68-L69)), so a placeholder attached to the wrong one renders identically to a placeholder attached to the right one. Giving them different tool names makes the case discriminating.
- [`assertCallsAreAnswered`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/llminternal/tool_pairing_test.go#L287) compares only the number of responses to the number of calls, so two responses for one call and none for another satisfies it. Matching each call to a response by ID, or by name where the ID is empty, would close that.

## Smaller things

- Because the placeholder content is user-role, a history that used to end on a model turn now ends on a user turn, and the synthetic "Continue processing…" turn at [`contents_processor.go:64-72`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/llminternal/contents_processor.go#L64-L72) no longer fires. That is what the two changed `agentTransfer` fixtures record. It may well be the behavior you want, but it is a second behavior change and it deserves a line in the description.
- [`withResultsInserted`](https://github.com/google/adk-go/blob/16c3d3010bd6a4c107f1fa91adc667760eb7077a/internal/llminternal/tool_pairing.go#L162-L173) inserts at the first non-response part, so a response turn beginning with a text part yields `[placeholder, text, real result]` and the real result still trails the text. I could not find a path where ADK itself builds a response turn in that shape, so this is about the comment promising more than the code delivers rather than a live problem. Hoisting every response part to the front would match the comment.
- Worth knowing about #1428, which changes the same function. The two merge cleanly but the combination fails #1428's `Late_async_completion_stays_final_with_an_unanswered_call_in_history`, because the placeholder adds a content its fixture does not expect. Each passes on its own, so whichever lands second will need to update the other's expectation.

One note on the motivation rather than the code: the OpenAI and Anthropic rejection behavior is the premise for the whole change, and there is no Anthropic model in this repo to check it against, so I could not verify that half. If you have the failing request or the API error to hand, adding it to the description would help the next reader.

karolpiotrowicz:

It synthesizes placeholder tool results for function calls that were never answered, so the model sees a paired call and result rather than a dangling call. No other ADK implementation does this today, and the adk-python equivalent is open rather than merged. So the question is whether ADK Go leads here or waits for upstream, which also decides the placeholder's wording and how long-running tools are distinguished. If we go first we are setting the shape the other implementations copy.

QuentinBisson:
Thanks, the ordering diagnosis is right and I have moved the pairing ahead of the id strip, as a separate loop after the conversion. The comment there now explains why the ids must still be present at that point.

The tests now exercise the strip: one case holds a long-running call open under an `adk-` id, and your "do both" reproduction is in as a case asserting `fetch_tool` keeps its single real result and `search_tool` receives the pending placeholder, both with empty ids as they reach the model. Reverting the ordering fails both of those cases. The two id-less calls now have different tool names, and `assertCallsAreAnswered` matches each call to one response by id (by name when the id is empty) and rejects leftover responses; the guard test covers the two-results-for-one-call shape too.

`withResultsInserted` now hoists every response part to the front, with a case for a response turn that starts with a text part.

On #1428: with its diff applied on top of this branch the only failure is still `Late_async_completion_stays_final_with_an_unanswered_call_in_history`, which now sees a placeholder content after `CALL(notify_tool)`. Whichever lands second adds that expected content to the fixture.

I added a line to the description about the second behaviour change (a history ending on a placeholder user turn no longer gets the synthetic continue turn). I do not have the verbatim Anthropic error to hand beyond the fragment quoted in adk-python #6914, so I have not added it.

Rebased onto main as well; the branch is 4ebd23a.

QuentinBisson:
@karolpiotrowicz and @wolo-lab the adk-python decided against this approach towards https://github.com/google/adk-python/pull/6914#issuecomment-5594451733.

I can implement this if needed but before I start, do you prefer implementing this yourself? Do you prefer if I open an issue or leave the PR open for now?

Issue #1153, opened 2026-07-13
Test coverage gaps in v1.3–v1.5 additions: memory/vertexai, streaming tool confirmation, adkrest RunLiveHandler


dmora:
While integrating v1.5.0 into our fork we noticed several newer surfaces ship without tests (all still true on main @ 0c88126):

1. **`memory/vertexai`** — no `_test.go` in the package. `AddSessionToMemory` has untested branching: the `stateKeySessionLastUpdateTime == ""` dispatch vs the filtered path, and the `t.(time.Time)` type assertion.

2. **`tool/functiontool/streaming_function.go`** — no `streaming_function_test.go`. Untested: the `RunStream` ToolConfirmation gate (both the rejection and the confirmation-request paths — the non-streaming equivalents are covered by `TestToolConfirmation`), and the `RequireConfirmationProvider` type-assertion error branch.

3. **`server/adkrest` `RunLiveHandler`** — the `/run_live` WebSocket endpoint has no integration test: upgrader handshake, the binary/text frame protocol contract, the two-goroutine lifecycle, and close-frame drain are all unexercised. The repo's existing mock live-agent patterns look sufficient to drive this without a real model connection.

Happy to contribute some of these if useful — filing first so the gaps are tracked and the intended testing approach can be discussed.

dmora:
Addendum on item 3: while reviewing `RunLiveHandler` we also noticed the `ws.WriteJSON` error path (server/adkrest/controllers/runtime.go:807 at v1.5.0) is the handler's only failure path with no log statement — worth covering when tests are added, since a dropped live session is currently uncorrelatable with logs.

indurireddy-TF:
Hi @dmora Thanks for pointing out these test coverage gaps and the missing log statement. Your offer to contribute is much appreciated.Please feel free to open a pull request for any of these areas when you are ready, and we will take a look.

GtechGovind:
I’m working on item 2 only: focused tests for streaming function-tool confirmation. The current branch covers rejected confirmation, confirmation-request state, confirmed execution, invalid RequireConfirmationProvider signatures, and early consumer stop/handler cleanup. I’ll keep the PR test-only and link it here once the repository checks pass.

wmh123wmh1:
Hi, I'd like to work on item 3 (`RunLiveHandler`) only.

I plan to add WebSocket integration tests covering:
- successful handshake and server event streaming
- forwarding text and binary client frames to the live session
- client close/disconnect and live-session cleanup
[...]

PR #1409, opened 2026-08-26
fix(tool): apply WithConfirmation to streaming tools

### LINKED ISSUE #1408: tool.WithConfirmation silently skips streaming tools, so a gated toolset runs them unconfirmed

## 🔴 Required Information

**Describe the Bug:**

`tool.WithConfirmation` does not wrap streaming tools, so a toolset that a caller has explicitly gated behind human-in-the-loop confirmation executes its streaming tools with no confirmation at all.

`confirmationToolset.Tools` wraps a tool only if it satisfies the unexported `runnableTool` interface (`tool/tool.go`):

```go
type runnableTool interface {
	Tool
	Run(ctx agent.Context, args any) (result map[string]any, err error)
}
```

The ADK flow, however, can execute two shapes, not one. `internal/toolinternal/tool.go` declares both `FunctionTool` (`Declaration` + `Run`) and `StreamingFunctionTool` (`Declaration` + `RunStream`), and `internal/llminternal/base_flow.go` dispatches on `StreamingFunctionTool` at line 1156, before `FunctionTool` at line 1211.

A streaming tool built with `functiontool.NewStreaming` has `Declaration` and `RunStream` but no `Run`, so it is not a `runnableTool`. `Tools()` takes the `else` branch and appends it unchanged. The flow then executes it through `RunStream`, whose only gate is the tool's own `Config.RequireConfirmation`. The toolset-level setting never reaches it.

There is a second, quieter path to the same outcome. A tool that satisfies both shapes is wrapped in `confirmationTool`, which embeds the `runnableTool` interface. The flow checks the streaming shape first, and while `confirmationTool` does not itself satisfy `StreamingFunctionTool`, the wrapper silently changes which method the flow will call. Either way the caller's instruction is not the one that is honoured.

The godoc is what makes this hard to notice. It says only tools that "provide a FunctionDeclaration and a Run method" are wrapped. A streaming tool does provide a FunctionDeclaration, so a reader has every reason to believe it is covered.

**Steps to Reproduce:**

1. Build a streaming tool with `functiontool.NewStreaming(functiontool.Config{Name: "streamTool", RequireConfirmation: false}, handler)`.
2. Put it in a toolset and wrap it: `cts := tool.WithConfirmation(ts, true, nil)`.
3. Call `cts.Tools(ctx)` and run the returned tool through `RunStream`.
4. The handler runs. `RequestConfirmation` is never called.

**Expected Behavior:**

`WithConfirmation(ts, true, nil)` gates every tool in the toolset that ADK can execute. The streaming tool asks for confirmation, `SkipSummarization` is set, and the call returns `ErrConfirmationRequired` until a confirmation arrives.

**Observed Behavior:**

The streaming tool is returned as the same object that went in, unwrapped, and executes immediately. No confirmation is requested and no error is returned.

**Environment Details:**

- ADK Library Version: `main` at `681ac5d`
- OS: macOS
- Go Version: 1.26.6

**Model Information:**

- N/A, this is reproducible without a model.

---

## 🟡 Optional Information

**Regression:**

No. `tool.WithConfirmation` arrived in #579 and streaming tools in #836; the gap has existed since both were present.

**Additional Context:**

The tool's own `RunStream` confirmation gate works correctly, and #1153 item 2 tracks test coverage for it. That gate reads `Config.RequireConfirmation` on the tool. This issue is about the separate toolset-level control, which is the one an operator reaches for when they want to gate a whole toolset they did not write, for example one returned by an MCP toolset.

Not every tool can be confirmed: the model-side built-ins in `tool/geminitool` are executed by the model, not by ADK, so passing those through unwrapped is correct. The fix is to wrap the shapes ADK does execute.

```go
handlerRan := false
streamTool, _ := functiontool.NewStreaming(
	functiontool.Config{Name: "streamTool", RequireConfirmation: false},
	func(ctx agent.Context, in struct{}) iter.Seq2[string, error] {
		return func(yield func(string, error) bool) {
			handlerRan = true
			yield("did the thing", nil)
		}
	})

ts := &testToolset{tools: []tool.Tool{streamTool}}
cts := tool.WithConfirmation(ts, true, nil) // gate everything in this toolset

tools, _ := cts.Tools(nil)
fmt.Println("unwrapped:", tools[0] == streamTool) // true

ctx := &testContext{Context: context.Background()}
for range tools[0].(toolinternal.StreamingFunctionTool).RunStream(ctx, map[string]any{}) {
}

fmt.Println(handlerRan, ctx.requestConfirmationCalled) // true false
```

- Always (100%)

venkat-uk:

- Closes: #1408

**Problem:**

`tool.WithConfirmation` wrapped a tool only when it satisfied the unexported `runnableTool` interface, which is `Declaration` plus `Run`. The flow can execute two shapes, not one: `toolinternal.FunctionTool` and `toolinternal.StreamingFunctionTool`, and `internal/llminternal/base_flow.go` dispatches on the streaming shape first (line 1156) and the non-streaming one second (line 1211).

A streaming tool has `Declaration` and `RunStream` but no `Run`, so it fell through the `else` branch and was appended unchanged. A caller who wrapped a toolset to gate every tool in it behind human-in-the-loop confirmation got no gate at all on the streaming tools, and no signal that this had happened. The only remaining gate was the tool's own `Config.RequireConfirmation`, which is a different control set by whoever wrote the tool rather than by whoever wrapped the toolset.

**Solution:**

Add a streaming counterpart to the wrapper and check the streaming shape first, so the wrapper matches the order in which the flow will dispatch.

Two details are deliberate:

- `confirmationStreamingTool` embeds the `streamingRunnableTool` **interface** rather than the concrete tool. Go promotes the methods of the embedded static type, so a tool that also has a `Run` method cannot reach the flow through an unguarded promoted method. The existing `confirmationTool` embeds `runnableTool` for the same reason, so this keeps the two symmetric.
- Tools that ADK does not execute itself still pass through unwrapped. The model-side built-ins in `tool/geminitool` are run by the model, so there is nothing to confirm. The `default` branch now says that, instead of describing the skipped tools as ones lacking a specific internal interface.

The confirmation decision moves into one `confirmCall` helper used by both paths, so the streaming and non-streaming semantics cannot drift apart later. Behaviour for non-streaming tools is unchanged.

The godoc no longer names `runnableTool`, which was not a useful thing to tell a caller: it described the skipped tools as ones that do not "provide a FunctionDeclaration and a Run method", and a streaming tool does provide a FunctionDeclaration, so the sentence pointed a reader away from the gap.

### Testing Plan

**Unit Tests:**

`TestWithConfirmation_StreamingTool` is a table test that mirrors the existing `TestWithConfirmation` cases against a streaming tool built with `RequireConfirmation: false`, so only the toolset-level setting is in play: confirmation required with nothing on the context, confirmed, rejected, not required, provider true, and `requireConfirmation=true` with a provider returning false.

`TestWithConfirmation_ToolWithBothRunAndRunStream` covers the second path. It uses a tool implementing both shapes and asserts the wrapper presents the streaming shape, does **not** also satisfy `FunctionTool`, and that neither method ran without confirmation.

Both fail on `main`:

```
$ git stash -- tool/tool.go
$ go test ./tool/ -run 'TestWithConfirmation_StreamingTool|TestWithConfirmation_ToolWithBoth' -count=1
--- FAIL: TestWithConfirmation_StreamingTool (0.00s)
    --- FAIL: TestWithConfirmation_StreamingTool/confirmation_required,_no_confirmation_in_context (0.00s)
    --- FAIL: TestWithConfirmation_StreamingTool/provider_requires_confirmation (0.00s)
--- FAIL: TestWithConfirmation_ToolWithBothRunAndRunStream (0.00s)
FAIL
```

and pass with the change, alongside the existing suite:

```
$ go test ./tool/... -count=1 -race
ok  	google.golang.org/adk/v2/tool
ok  	google.golang.org/adk/v2/tool/agenttool
ok  	google.golang.org/adk/v2/tool/functiontool
ok  	google.golang.org/adk/v2/tool/mcptoolset
ok  	google.golang.org/adk/v2/tool/skilltoolset
ok  	google.golang.org/adk/v2/tool/toolconfirmation
ok  	google.golang.org/adk/v2/tool/toolutils
... all green
```

`go vet ./tool/` is clean.

The behaviour is observable without a model, which is why the reproduction is a test rather than a runner transcript. Before the change, `cts.Tools(nil)` returns the same pointer that went in (`tools[0] == streamTool` is `true`), the handler runs, and `RequestConfirmation` is never called. After it, the same call returns `ErrConfirmationRequired`, sets `SkipSummarization`, and the handler does not run until a confirmation arrives on the context.

### Checklist

### Additional context

`WithConfirmation` is marked EXPERIMENTAL and out of scope for the v1.0 API, so this is a fix within an experimental surface rather than a public API change. The exported signature is unchanged.

Adjacent but deliberately not in this PR: #1153 item 2 tracks test coverage for the tool's own `RunStream` confirmation gate, which is a different control and already works. Happy to split anything here if you would rather see the wrapper and the helper extraction land separately.


jjsasha63:
The fix is correct. One test worth adding before this merges.

I checked the claims by running them rather than reading them — dispatch order, interface satisfaction, the `confirmCall` extraction, packing, and the runtime semantics on both streaming paths in the flow.

Confirmed correct:

- The dispatch-order claim holds. `base_flow.go:1156` checks `StreamingFunctionTool` before `FunctionTool` at `:1211`, so checking the streaming shape first in the type switch is right.
- `confirmationStreamingTool` satisfies `toolinternal.StreamingFunctionTool` and does not satisfy `toolinternal.FunctionTool`.
- `streamingRunnableTool` is an exact structural match for `toolinternal.StreamingFunctionTool`.
- `confirmCall` is behaviour-preserving for the non-streaming path. I ran a 10-scenario differential harness against both versions of `tool/tool.go`: identical error strings, identical ordering of `SkipSummarization` against `RequestConfirmation`, identical behaviour when `RequestConfirmation` itself fails, identical provider precedence.
- Exported API unchanged, merges cleanly onto current main, full suite green including `-race -shuffle=on`.

## The gap: `confirmationStreamingTool.ProcessRequest` has no test

It is 18 of the roughly 60 new lines and nothing covers it. Three separate mutations to it survive the entire suite. The worst one:

```go
// in confirmationStreamingTool.ProcessRequest
req.Tools[t.Name()] = t.streamingRunnableTool            // was: = t
...
return toolutils.PackTool(req, t.streamingRunnableTool)  // was: PackTool(req, t)
```

`go test ./tool/... -count=1` with that applied: all 15 packages ok, exit 0.

That mutant puts the unwrapped tool into `req.Tools`, so the flow executes it ungated. It is the exact bug this PR fixes, reintroduced through the other method, with a green suite.

The method is load-bearing rather than dead code. Embedding an interface does not promote `ProcessRequest`, so deleting it makes `toolPreprocess` reject the tool at `internal/llminternal/base_flow.go:772-774`, and any agent wrapping a streaming toolset fails on its first invocation. Deleting the whole method also leaves the suite green.

Here is a test that closes it, if it is useful:

```go
// bareStreamingTool is a streaming tool with no ProcessRequest of its own, so
// the confirmation wrapper has to pack itself.
type bareStreamingTool struct {
	name      string
	streamRan *bool
}

func (b *bareStreamingTool) Name() string        { return b.name }
func (b *bareStreamingTool) Description() string { return "" }
func (b *bareStreamingTool) IsLongRunning() bool { return false }
func (b *bareStreamingTool) Declaration() *genai.FunctionDeclaration {
	return &genai.FunctionDeclaration{Name: b.name}
}

func (b *bareStreamingTool) RunStream(ctx agent.Context, args any) iter.Seq2[string, error] {
	return func(yield func(string, error) bool) {
		*b.streamRan = true
		yield("did the thing", nil)
	}
}

// selfPackingStreamingTool packs itself into the request from its own
// ProcessRequest, as functiontool's streaming tool and MCP tools do. The
// confirmation wrapper has to swap that entry for itself, or the flow looks the
// tool up by name and finds an ungated RunStream.
type selfPackingStreamingTool struct {
	*bareStreamingTool
	processRequestCalled *bool
}

func (s *selfPackingStreamingTool) ProcessRequest(ctx agent.Context, req *model.LLMRequest) error {
	*s.processRequestCalled = true
	return toolutils.PackTool(req, s)
}

func TestWithConfirmation_StreamingProcessRequest_Packing(t *testing.T) {
	const toolName = "streamPackingTool"

	tests := []struct {
		name string
		// build makes the tool the toolset hands out. The two cases cover the two
		// ways the wrapper's ProcessRequest can populate req.Tools: swapping an
		// entry the inner tool packed itself, and packing the wrapper when the
		// inner tool has no ProcessRequest.
		build                    func(base *bareStreamingTool, processRequestCalled *bool) tool.Tool
		wantProcessRequestCalled bool
	}{
		{
			name: "inner tool packs itself",
			build: func(base *bareStreamingTool, processRequestCalled *bool) tool.Tool {
				return &selfPackingStreamingTool{bareStreamingTool: base, processRequestCalled: processRequestCalled}
			},
			wantProcessRequestCalled: true,
		},
		{
			name:  "inner tool has no ProcessRequest",
			build: func(base *bareStreamingTool, _ *bool) tool.Tool { return base },
		},
	}

	for _, tt := range tests {
			streamRan := false
			processRequestCalled := false
			base := &bareStreamingTool{name: toolName, streamRan: &streamRan}

			ts := &testToolset{tools: []tool.Tool{tt.build(base, &processRequestCalled)}}
			tools, err := tool.WithConfirmation(ts, true, nil).Tools(nil)
			if err != nil {
				t.Fatalf("Tools() failed: %v", err)
			}
			if len(tools) != 1 {
				t.Fatalf("Tools() returned %d tools, want 1", len(tools))
			}

			processor, ok := tools[0].(toolinternal.RequestProcessor)
			if !ok {
				t.Fatalf("wrapped tool does not implement RequestProcessor, so the flow would reject it")
			}

			req := &model.LLMRequest{}
			if err := processor.ProcessRequest(nil, req); err != nil {
				t.Fatalf("ProcessRequest() failed: %v", err)
			}
			if processRequestCalled != tt.wantProcessRequestCalled {
				t.Errorf("inner ProcessRequest called = %v, want %v", processRequestCalled, tt.wantProcessRequestCalled)
			}

			// The flow executes whatever req.Tools holds under this name, so an
			// inner tool left there is an ungated tool.
			if req.Tools[toolName] != tools[0] {
				t.Fatalf("req.Tools[%q] = %T, want the confirmation wrapper %T", toolName, req.Tools[toolName], tools[0])
			}
			packed, ok := req.Tools[toolName].(toolinternal.StreamingFunctionTool)
			if !ok {
				t.Fatalf("packed tool is %T, not a StreamingFunctionTool, so the flow could not execute it", req.Tools[toolName])
			}
			if _, ok := req.Tools[toolName].(toolinternal.FunctionTool); ok {
				t.Errorf("packed tool also satisfies FunctionTool, so an unguarded Run is reachable")
			}

			ctx := &testContext{Context: t.Context()}
			for _, err := range packed.RunStream(ctx, map[string]any{}) {
				if !errors.Is(err, tool.ErrConfirmationRequired) {
					t.Errorf("RunStream() error = %v, want ErrConfirmationRequired", err)
				}
			}
			if streamRan {
				t.Errorf("packed tool ran without confirmation")
			}
			if !ctx.requestConfirmationCalled {
				t.Errorf("requestConfirmationCalled = false, want true")
			}
		})
	}
}
```

With it in place, all four mutations are caught:

| Mutation to `confirmationStreamingTool.ProcessRequest` | Caught by |
| --- | --- |
| Swap branch packs the inner tool | `inner tool packs itself` |
| `PackTool(req, t.streamingRunnableTool)` | `inner tool has no ProcessRequest` |
| Drop the swap branch entirely | `inner tool packs itself`, via `duplicate tool` |
| Delete the method | both, via `does not implement RequestProcessor` |

One trap if you write your own version. My first attempt used `functiontool.NewStreaming` for the "no ProcessRequest" case, but that type has its own `ProcessRequest` at `streaming_function.go:102`, so both cases went down the swap branch and the `PackTool` fallback stayed untested. Hand-written fixtures avoid that and keep the test hermetic.

`gofmt` clean, `go vet` clean, `go test ./tool/... -race -shuffle=on` green, and `go test ./...` across the repo has no failures.

## Smaller notes

1. The godoc at `tool/tool.go:130-136` now reads as a promise that every tool the flow can execute is gated. In a live/bidi session it is not. `base_flow.go:1158` sets the pending-status result before any tool code runs, `generateRequestConfirmationEvent` is only reachable from the non-live path at `base_flow.go:684`, and the actions snapshot at `base_flow.go:1248` copies a `RequestedToolConfirmations` map that `RequestConfirmation` only allocates afterwards. The gate fires, the error is printed, and nothing reaches the user. This is pre-existing — it reproduces on main with `functiontool.NewStreaming(Config{RequireConfirmation: true})` — but the new wording invites the opposite reading. A sentence naming the gap, or a linked follow-up issue, would settle it.

2. `ProcessRequest` is now duplicated character for character at `tool/tool.go:243-259` and `:283-300`. The stated reason for extracting `confirmCall` was to stop the two paths drifting, and the larger method has the same problem.

3. Worth a line in the description: a tool with both `Run` and `RunStream` now goes down `RunStream` once wrapped, where it previously went down `Run`. That matches what an unwrapped dual tool already does, so it is the right call, but the response payload shape differs and the before and after tool callbacks no longer fire. Related, `workflow.NewToolNode` at `workflow/tool_node.go:66` now rejects a wrapped dual tool it used to accept.

Three more surviving mutants in the existing tests, all low priority:

- the "rejected in context" subtest is vacuous. `functiontool`'s streaming tool rejects the call itself at `streaming_function.go:148`, so removing the wrapper's `ErrConfirmationRejected` return is not caught.
- nothing asserts the `default:` passthrough, so the godoc's claim about the model-side built-ins is unchecked.
- the pass-through loop ignoring a consumer `break` is not caught, which matters for a long-running stream.

tool/tool.go
```diff
@@ -156,21 +159,60 @@ func (c *confirmationToolset) Tools(ctx agent.ReadonlyContext) ([]Tool, error) {
 	}
 	wrappedTools := make([]Tool, 0, len(tools))
 	for _, t := range tools {
-		if _, ok := t.(runnableTool); ok {
+		// Streaming is checked first because the flow dispatches on
+		// StreamingFunctionTool before FunctionTool. A tool that satisfies both
+		// must be wrapped the way it will be called.
+		switch rt := t.(type) {
+		case streamingRunnableTool:
```
wojtas000: **Blocking** — this flips dispatch for a tool that has both `Run` and `RunStream`. On main such a tool is wrapped in `confirmationTool`, which embeds the `runnableTool` interface, so `RunStream` is not promoted and the flow calls `Run`. It now matches this case, and the flow calls `RunStream` (`internal/llminternal/base_flow.go:1164`). For that tool the before/after tool callbacks stop firing, the FunctionResponse becomes `{"result": "<joined chunks>"}` instead of the map `Run` returned, and `workflow.NewToolNode` rejects the wrapper because it requires a `Run` method (`workflow/tool_node.go:66`). No in-repo tool implements both shapes, so only user-authored tools are affected, but the description does not mention it. Please state it there, and say why RunStream-first is the right call given the `NewToolNode` rejection.

tool/tool.go
```diff
@@ -127,10 +128,12 @@ func (f *filteredToolset) Tools(ctx agent.ReadonlyContext) ([]Tool, error) {
...
 
 // WithConfirmation wraps a toolset to inject confirmation logic in each tool.
-// Only tools within the provided Toolset that implement the `runnableTool` interface
-// (i.e., provide a FunctionDeclaration and a Run method) will be wrapped with
-// confirmation logic. Other tools will be included in the returned Toolset without
-// modification.
+//
+// Every tool the ADK flow can execute is wrapped: tools that provide a
+// FunctionDeclaration and a Run method, and streaming tools that provide a
+// FunctionDeclaration and a RunStream method. Tools the flow never executes
```
wojtas000: **Blocking** — this now reads as a promise that every tool the flow can execute is gated end to end. In a live session it is not: the gate fires and the error is printed, but no confirmation event is emitted, so the user is never asked and the tool never runs. See the repro in the summary. A sentence here naming the live-mode gap, plus a linked follow-up issue, would settle it.

tool/tool.go
```diff
@@ -208,30 +259,56 @@ func (t *confirmationTool) ProcessRequest(ctx agent.Context, req *model.LLMReque
...
+			return err
 		}
-	} else {
-		requireConfirmation := t.requireConfirmation
-		if t.provider != nil {
-			requireConfirmation = t.provider(t.Name(), args)
+		// If the inner tool packed itself into req.Tools during ProcessRequest,
+		// replace it with the confirmation wrapper so this wrapper's RunStream
+		// is invoked.
+		if !existedBefore && req.Tools != nil && req.Tools[t.Name()] != nil {
```
wojtas000: **Blocking** — this swap has no test. jjsasha63 asked for one on 31 August, and every commit since is a merge from main. The non-streaming twin has `TestWithConfirmation_ProcessRequest` and `TestWithConfirmation_ProcessRequest_Packing`; this branch has neither. If it regresses, `req.Tools[name]` keeps the inner streaming tool and the flow runs it unguarded, which is the exact bug this PR closes.

tool/tool_test.go
```diff
@@ -441,3 +442,176 @@ func TestWithConfirmation_ProcessRequest_Packing(t *testing.T) {
...
+				t.Fatalf("cts.Tools() returned %d tools, want 1", len(tools))
+			}
+			wrapped, ok := tools[0].(toolinternal.StreamingFunctionTool)
+			if !ok {
+				t.Fatalf("tools[0] is not a StreamingFunctionTool, the flow could not execute it")
+			}
+
+			ctx := &testContext{Context: t.Context(), toolConfirmationResult: tt.toolConfirmation}
+			var gotErr error
+			for _, err := range wrapped.RunStream(ctx, map[string]any{}) {
```
wojtas000: nit: the chunks are discarded, so chunk forwarding is never asserted. A wrapper that iterated the inner tool but dropped every chunk would still pass here, because `handlerRan` is a closure side effect. Collecting the chunks and asserting the payload on the confirmed and not-required cases would pin `RunStream`'s pass-through.

tool/tool_test.go
```diff
@@ -441,3 +442,176 @@ func TestWithConfirmation_ProcessRequest_Packing(t *testing.T) {
...
+				}
+			}
+
+			if tt.wantErrMsg == "" && gotErr != nil {
+				t.Errorf("RunStream() error = %v, want none", gotErr)
+			}
+			if tt.wantErrMsg != "" {
+				if gotErr == nil {
+					t.Errorf("RunStream() error = nil, want it to contain %q", tt.wantErrMsg)
+				} else if !strings.Contains(gotErr.Error(), tt.wantErrMsg) {
```
wojtas000: nit: `errors.Is(gotErr, tool.ErrConfirmationRequired)` would survive a reword of the message. `TestWithConfirmation_ToolWithBothRunAndRunStream` already does it that way. This matches the older table test, so take it or leave it.

jjsasha63:
I am closing this in favor of #1596, which finishes the change.

This PR has been waiting on you since 27 August. My review of 31 August is
unanswered. Two weeks on, I have taken the change over so the fix reaches users.

Your commit is in #1596 under your authorship. The fix itself is unchanged. I
merged current main and added the one test I asked for, covering
confirmationStreamingTool.ProcessRequest.

If you come back before #1596 merges, say so and the change goes back to you.

Thanks for finding this. Streaming tools were passing through WithConfirmation
untouched, and nothing signaled it.

PR #1412, opened 2026-08-27
feat(runner): bound model calls per invocation with RunConfig.MaxLLMCalls

### LINKED ISSUE #1286: LiveRunConfig.MaxLLMCalls is never enforced — live sessions are unbounded

## 🔴 Required Information

**Describe the Bug:**

`agent.LiveRunConfig.MaxLLMCalls` is declared and can be set by callers, but nothing in the codebase ever reads it. A caller who sets it believes their live session is bounded; it is not.

This is worse than the limit simply being absent. An absent limit makes you go looking for one; a present, silently-ignored field makes you stop looking.

The framework's own REST server does exactly this. `server/adkrest/controllers/runtime.go:302` opens every live WebSocket session with:

```go
liveSession, eventIter, err := r.RunLive(req.Context(), userID, sessionID, agent.LiveRunConfig{
	MaxLLMCalls:              100, // Reasonable default
	...
})
```

The comment asserts a safety limit that does not exist. Every ADK Web / adkrest live session is unbounded.

**Steps to Reproduce:**

The field's inertness can be shown statically, without running anything:

1. Search the whole workspace for the identifier. There are exactly two occurrences — one declaration, one write, zero reads:

```
agent/live.go:48                            MaxLLMCalls              int
server/adkrest/controllers/runtime.go:302   MaxLLMCalls:              100, // Reasonable default
```

2. Delete the field declaration at `agent/live.go:48` and build the workspace:

```
go build -mod=readonly work
```

3. Observe that exactly one compile error results — the assignment. If any code read the field, deleting it would break that read too.

**Expected Behavior:**

Setting `MaxLLMCalls` to N bounds a live session to N model calls; exceeding it terminates the session with a clear error. Alternatively, if the limit is not going to be implemented, the field should be removed rather than left as a no-op.

**Observed Behavior:**

```
# google.golang.org/adk/v2/server/adkrest/controllers
server/adkrest/controllers/runtime.go:302:3: unknown field MaxLLMCalls in struct literal of type "google.golang.org/adk/v2/agent".LiveRunConfig
```

That single error is the entire consequence of removing the field. The value is written and never consulted.

**Environment Details:**

 - ADK Library Version: `main` @ b264039; the field is present in v2.0.0, v2.1.0 and v2.2.0
 - OS: Linux
 - Go Version: 1.26.5 (as declared in `go.mod`)

**Model Information:**
 - Which model is being used: N/A — this is a static defect, independent of the model.

---

## 🟡 Optional Information

**Regression:**

N/A — the field appears to have never been enforced.

**Additional context:**

There is no LLM-call counting mechanism anywhere in adk-go, so there is nothing for the field to hook into. Searching for `llmCallCount`, `numberOfLlmCalls`, `callBudget`, `incrementLLM` and similar returns only the two lines above.

adk-python, which adk-go tracks for behavior parity, does have one: `RunConfig.max_llm_calls` (default 500, `agents/run_config.py`) enforced by `InvocationContext.increment_llm_call_count()` raising `LlmCallsLimitExceededError` immediately before each model call in `_call_llm_async`.

Note the live and non-live paths in adk-go are separate: `LiveRunConfig` only reaches `Flow.RunLive`, while `runner.Run` leaves `runconfig.RunConfig.Live` nil. So fixing this field bounds live sessions only. The non-live path has no limit field at all — filed separately.

Happy to submit a PR if the maintainers agree on the intended semantics (terminate with an error vs. stop gracefully).

### LINKED ISSUE #1287: No way to bound LLM calls in a non-live run: add RunConfig.MaxLLMCalls

There is no way to bound the number of model calls in a non-live run. `agent.RunConfig` has two fields, neither a limit:

```go
type RunConfig struct {
	StreamingMode             StreamingMode
	SaveInputBlobsAsArtifacts bool
}
```

`Flow.Run` drives a `for {}` loop that continues while the last event is not a final response. Whether it terminates depends entirely on model behavior. A model that keeps emitting function calls, or keeps emitting thought-only turns, loops indefinitely — one event appended per iteration and an ever-growing history re-sent on each call, so the token cost grows quadratically. With a fake model that only ever returns function calls I measured `events=60 modelCalls=30` and climbing, with no exit.

The only limit-shaped field in the codebase, `LiveRunConfig.MaxLLMCalls`, cannot help here: `runner.Run` and `runNode` leave `runconfig.RunConfig.Live` nil, so it is structurally unreachable from `Flow.Run`. (It is also never enforced on the live path either — filed separately as a bug.)

adk-python, which adk-go tracks for behavior parity, bounds every run:

```python
# agents/run_config.py
max_llm_calls: int = 500
```

enforced in `InvocationContext.increment_llm_call_count()`, which raises `LlmCallsLimitExceededError` immediately before each model call from `_call_llm_async`. Python cannot hang here regardless of what the model returns.

### Proposed Solution

Add a call budget to the non-live config:

```go
type RunConfig struct {
	StreamingMode             StreamingMode
	SaveInputBlobsAsArtifacts bool
	MaxLLMCalls               int // 0 = unbounded
}
```

`0 = unbounded` follows the convention already set by `loopagent.Config.MaxIterations`.

Enforcement wants a single chokepoint. The non-live path has one: `m.GenerateContent(ctx, req, useStream)` at `internal/llminternal/base_flow.go:911`, inside `Flow.callLLM`. A counter on the invocation, incremented and checked immediately before that call, mirrors python's placement exactly.

The plumbing already exists — `base_flow.go` reads `runconfig.FromContext(ctx)` at lines 306 and 821, and the internal `runconfig.RunConfig` is constructed in three places in `runner/`. Carrying the value through would need a new field on the internal config rather than new machinery. Populating it from either `agent.RunConfig` or `LiveRunConfig` would let one counter serve both paths.

Open questions for maintainers:
- Default value. Python uses 500. A non-zero default protects everyone but is a behavior change for anyone currently relying on very long runs.
- Failure mode. Python raises. adk-go's iterator idiom would suggest yielding an error, consistent with how `Flow.Run` already handles the `Partial` case.

### Impact on your work

We hit an unbounded `Flow.Run` while migrating a Go codebase from v1.5 to v2 — an integration test hung outright rather than failing. A hang in CI is considerably more disruptive than an error, and against a real endpoint the same loop is unbounded spend rather than a stalled test.

The specific trigger we hit is being fixed narrowly, but that fix is content inspection: it recognises one degenerate shape and stops. A call budget is the general answer, because it terminates regardless of what the model returns, including shapes nobody has anticipated.

---

### Alternatives Considered

- **Per-symptom caps** (e.g. bounding consecutive thought-only turns). Cheap and fast-failing, but each only covers one known shape, and any threshold small enough to fail fast risks false positives against future model behavior.
- **Enforcing `LiveRunConfig.MaxLLMCalls` alone.** Necessary in its own right, but structurally cannot reach the non-live path.
- **`loopagent.MaxIterations`.** Bounds a loop agent's iterations, not model calls inside a single agent's flow.

### Willingness to contribute

Yes — happy to implement this once maintainers settle the default value and the failure mode.

venkat-uk:

- Closes: #1287
- Related: #1286, which stays open. See Additional context.

**Problem:**

Whether a non-live run terminates depends entirely on model behaviour. `Flow.Run` leaves its loop on a non-thought-only final response, ten consecutive thought-only turns, an error, a partial last event, consumer cancellation, or `EndInvocation()`. A model that keeps requesting tool calls satisfies none of them, so the loop appends an event per turn and re-sends a growing history, and token cost grows quadratically with no exit.

`agent.RunConfig` had two fields, `StreamingMode` and `SaveInputBlobsAsArtifacts`, and neither is a limit. The only limit-shaped field in the codebase, `LiveRunConfig.MaxLLMCalls`, is on the live path and is not reachable from `Flow.Run`.

The analysis is @jjsasha63's, in #1287 and #1286, including the measurement of `events=60 modelCalls=30 and climbing` against a model that only returns function calls. I have followed his design note that the two configs should share one counter rather than growing two divergent limits.

Worth noting that the case for a bound is already made in this repository, in the comment above `maxConsecutiveThoughtOnlyTurns`:

> A model that keeps emitting thoughts without ever surfacing an answer would otherwise spin forever, appending an event per turn and re-sending an ever-growing history.

That reasoning was accepted for the thought-only branch in #1290 and is absent from the function-call branch beside it.

**Solution:**

`agent.RunConfig` gains `MaxLLMCalls int`, with `agent.ErrLLMCallsLimitExceeded` as the sentinel so callers can use `errors.Is`.

Resolution, in `runconfig.ResolveMaxLLMCalls`:

| Value | Meaning |
|---|---|
| `0`, what `agent.RunConfig{}` gives | 500, or `ADK_MAX_LLM_CALLS` if it parses |
| positive | that limit |
| negative | no limit |

The zero value carrying a budget is the point of the change. `agent.RunConfig{}` appears in every example in the repository, including the one in `AGENTS.md`, and "unlimited by default" is the state this is meant to remove. 500 and the environment variable name both match adk-python's `RunConfig.max_llm_calls`, which is the parity target under the alignment rule in `AGENTS.md`.

Three details are deliberate:

- **The counter is per invocation, not per agent.** It lives on the shared `internal/agent/runconfig.RunConfig`, which the runner puts on the context by pointer, so sub-agents and transfers inside one invocation draw on one budget. This is the shared counter #1286 asked for. It is an `atomic.Int64` because `parallelagent` runs sub-agents concurrently.
- **Enforcement sits at the single place the non-live flow reaches a model**, immediately before `generateContent` in `callLLM`, so a `Before` model callback that short-circuits does not spend budget. Counting happens first and the check second, so a limit of n permits exactly n calls, matching `increment_and_enforce_llm_calls_limit` in adk-python.
- **A nil run config imposes no limit.** `runconfig.FromContext` returns nil when an agent is run directly rather than through the runner (#586), and that path is unchanged.

### Testing Plan

**Unit Tests:**

`internal/agent/runconfig/limits_test.go` covers resolution (default, environment override, unparsable override, negative override, explicit value winning over the environment, negative pass-through) and enforcement (a limit of n permits exactly n calls, non-positive limits do not bound, a nil receiver does not bound, and a concurrency case asserting that 100 goroutines against a limit of 50 get exactly 50 through, which is the shared-counter guarantee).

`runner/max_llm_calls_test.go` is the end-to-end case, driven by a model that only ever returns the same function call:

- `MaxLLMCalls: 3` ends the run with an error wrapping `ErrLLMCallsLimitExceeded` after exactly 3 model calls.
- `agent.RunConfig{}` with `ADK_MAX_LLM_CALLS=2` bounds at 2, which is the assertion that the empty literal is no longer unlimited.
- `MaxLLMCalls: -1` is not stopped by the budget: the consumer stops it after 12 calls instead.

Results:

```
$ go test ./runner/ -run 'TestRun_MaxLLMCalls' -count=1 -v
--- PASS: TestRun_MaxLLMCallsBoundsTheRun (0.00s)
--- PASS: TestRun_MaxLLMCallsDefaultsToABudget (0.00s)
--- PASS: TestRun_MaxLLMCallsNegativeMeansUnlimited (0.00s)
ok  	google.golang.org/adk/v2/runner	0.493s

$ go test ./internal/agent/runconfig/ -count=1 -race
ok  	google.golang.org/adk/v2/internal/agent/runconfig	1.632s

EXIT 0, no failures, no races

$ go vet ./agent/ ./runner/ ./internal/agent/runconfig/ ./internal/llminternal/
clean

$ go mod tidy -diff
prints nothing
```

Worth reporting the mutation check, because it is unusual. With the enforcement line removed, `TestRun_MaxLLMCallsBoundsTheRun` does not fail: it hangs until the Go test timeout kills it at 30s with a goroutine dump. That is the bug demonstrating itself, and it is the reason a bound belongs in the framework rather than in each caller.

The looping-model test is the end-to-end check and needs no credentials, which is why it is a test rather than a runner transcript. Setup is the standard one from `AGENTS.md`: an `llmagent` with one no-op tool, `runner.New` with an in-memory session service and `AutoCreateSession`, and a model whose every response is the same function call. On `main` that setup does not terminate. On this branch it ends after `MaxLLMCalls` model calls with an error the caller can match with `errors.Is(err, agent.ErrLLMCallsLimitExceeded)`.

### Checklist

### Additional context

**This does not close #1286, and that is deliberate.** The live path holds a bidi connection and does not go through `callLLM`, so what counts as one call there is a decision I did not want to make on the maintainers' behalf. `LiveRunConfig.MaxLLMCalls` is untouched and still inert. Wiring it without enforcing it would recreate exactly the problem #1286 reports, that a present, silently ignored field makes you stop looking. The counter and the enforcement helper are on the shared config so that wiring the live path is a small follow-up rather than a second mechanism.

**On the default being a behaviour change.** An invocation that legitimately makes more than 500 model calls will now stop. That is the same position adk-python has taken since its default was introduced, and the escape hatches are per-run (`MaxLLMCalls: -1`) and per-deployment (`ADK_MAX_LLM_CALLS`). Happy to make the default unlimited instead if you would rather not change existing behaviour on the 2.x line, though I think the safe default is the more useful one and the parity argument supports it.

@jjsasha63, if you would rather take this yourself, say so and I will close it.

### DIFF
--- a/agent/run_config.go
+++ b/agent/run_config.go
@@ -14,6 +14,12 @@
 
 package agent
 
+import "errors"
+
+// ErrLLMCallsLimitExceeded is returned when an invocation makes more model
+// calls than RunConfig.MaxLLMCalls allows. Detect it with errors.Is.
+var ErrLLMCallsLimitExceeded = errors.New("max number of llm calls exceeded")
+
 // StreamingMode defines the streaming mode for agent execution.
 type StreamingMode string
 
@@ -32,4 +38,18 @@ type RunConfig struct {
 	// If true, ADK runner will save each part of the user input that is a blob
 	// (e.g., images, files) as an artifact.
 	SaveInputBlobsAsArtifacts bool
+	// MaxLLMCalls bounds the total number of model calls one invocation may
+	// make, across every agent it runs. Exceeding it ends the run with an error
+	// wrapping ErrLLMCallsLimitExceeded.
+	//
+	// Zero, the value a caller gets from agent.RunConfig{}, means the default:
+	// 500, or the value of the ADK_MAX_LLM_CALLS environment variable if it is
+	// set to a valid integer. A negative value means no limit.
+	//
+	// The limit exists because whether a run terminates otherwise depends
+	// entirely on model behaviour: a model that keeps requesting tool calls
+	// appends an event per turn and re-sends a growing history, so token cost
+	// grows quadratically with no exit. Mirrors adk-python's
+	// RunConfig.max_llm_calls.
+	MaxLLMCalls int
 }
--- a/internal/agent/runconfig/limits.go
+++ b/internal/agent/runconfig/limits.go
@@ -1,0 +1,73 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package runconfig
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"strconv"
+
+	"google.golang.org/adk/v2/agent"
+)
+
+const (
+	// DefaultMaxLLMCalls is the model-call budget an invocation gets when the
+	// caller does not choose one. Matches adk-python's default.
+	DefaultMaxLLMCalls = 500
+
+	// MaxLLMCallsEnvVar overrides DefaultMaxLLMCalls. Matches adk-python.
+	MaxLLMCallsEnvVar = "ADK_MAX_LLM_CALLS"
+)
+
+// ResolveMaxLLMCalls turns the value a caller set on a public run config into
+// the limit to enforce.
+//
+// A positive value is used as given. A negative value means no limit and is
+// passed through. Zero, which is what a caller gets from an empty struct
+// literal, resolves to the environment override if it parses, and to
+// DefaultMaxLLMCalls otherwise.
+func ResolveMaxLLMCalls(v int) int {
+	if v != 0 {
+		return v
+	}
+	raw, ok := os.LookupEnv(MaxLLMCallsEnvVar)
+	if !ok {
+		return DefaultMaxLLMCalls
+	}
+	n, err := strconv.Atoi(raw)
+	if err != nil {
+		log.Printf("adk: invalid %s value %q, using the default %d", MaxLLMCallsEnvVar, raw, DefaultMaxLLMCalls)
+		return DefaultMaxLLMCalls
+	}
+	return n
+}
+
+// IncrementAndEnforceLLMCallsLimit counts one model call against the
+// invocation's budget and reports whether the budget is now exceeded.
+//
+// It counts first and then checks, so a limit of n permits exactly n calls. A
+// nil receiver, which happens when an agent is run directly rather than through
+// the runner, imposes no limit, and so does a non-positive limit.
+func (c *RunConfig) IncrementAndEnforceLLMCallsLimit() error {
+	if c == nil {
+		return nil
+	}
+	calls := c.llmCalls.Add(1)
+	if c.MaxLLMCalls <= 0 || calls <= int64(c.MaxLLMCalls) {
+		return nil
+	}
+	return fmt.Errorf("%w: limit is %d", agent.ErrLLMCallsLimitExceeded, c.MaxLLMCalls)
+}
--- a/internal/agent/runconfig/limits_test.go
+++ b/internal/agent/runconfig/limits_test.go
@@ -1,0 +1,110 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package runconfig
+
+import (
+	"errors"
+	"sync"
+	"testing"
+
+	"google.golang.org/adk/v2/agent"
+)
+
+func TestResolveMaxLLMCalls(t *testing.T) {
+	tests := []struct {
+		name  string
+		in    int
+		env   string
+		setEn bool
+		want  int
+	}{
+		{name: "zero takes the default", in: 0, want: DefaultMaxLLMCalls},
+		{name: "zero takes the environment override", in: 0, env: "7", setEn: true, want: 7},
+		{name: "zero ignores an unparsable override", in: 0, env: "lots", setEn: true, want: DefaultMaxLLMCalls},
+		{name: "zero honours a negative override", in: 0, env: "-1", setEn: true, want: -1},
+		{name: "an explicit value wins over the environment", in: 5, env: "7", setEn: true, want: 5},
+		{name: "a negative value is passed through", in: -1, want: -1},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.setEn {
+				t.Setenv(MaxLLMCallsEnvVar, tt.env)
+			}
+			if got := ResolveMaxLLMCalls(tt.in); got != tt.want {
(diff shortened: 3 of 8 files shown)


jjsasha63 (changes requested):
@venkat-uk Keep it — I'm not pulling it back. I closed #1287 and #1286 as not
planned because I didn't want open issues I wasn't working, not because the
problem went away, and your write-up states it better than mine did.

I built this on current main (clean merge): `go build`, `go vet` and
`go test -race ./internal/agent/runconfig/... ./runner/...` all pass.
Enforcement is in the right place — same position as `increment_llm_call_count()`
in adk-python's `base_llm_flow.py`, after the before-model callbacks and before
the call.

On the default: keep 500. AGENTS.md makes adk-python the parity target,
unlimited is what the issue was about, and the two escape hatches cover the
people it would hurt. I'll call it out in the release notes.

Four things before I approve:

1. Drop the `Closes: #1287` keyword. Both issues are closed as not planned, so
   merging won't touch them. Keep the link, lose the keyword.
2. Document the zero divergence in the `MaxLLMCalls` doc comment. adk-python
   reads `max_llm_calls=0` as unlimited; here 0 is the default and -1 is
   unlimited. Unavoidable in Go without a pointer, but anyone porting will trip
   on it.
3. Note that the compaction summarizer's calls
   (`session/compaction/llm_summarizer.go`) are not counted. I think that's
   right — it just shouldn't be a surprise.
4. Live path: leave it out, your reasoning is correct. Reword the "Additional
   context" section so it doesn't read as leaving #1286 half-done. The shared
   counter is the deliverable; wiring live is a separate decision about what one
   call means over a bidi connection.

PR #1413, opened 2026-08-27
Adding opt-in option for the debug endpoint in ADK REST API

kdroste-google:
**Problem:**
Debug endpoint in api was unconditionally exposed. 

**Solution:**
Opt-in option to include debug api to `web api`. Includes Opt-in for cloudrun deployment. 

### Testing Plan
Manual tests

 - local run with/without the -include_debug_api flag
 - cloudrun deployment with/without ` --debug_api` flag

### Checklist

- [N/A] I have added tests that prove my fix is effective or that my feature works.
- [X] Any dependent changes have been merged and published in downstream modules.

### DIFF
--- a/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
+++ b/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
@@ -48,8 +48,8 @@ type cloudRunServiceFlags struct {
 	serverPort      int
 	a2aAgentCardURL string
 	a2a             bool // enable a2a or not
-	api             bool // enable api or not
-	debugApi        bool // enable debug api or not
+	api             bool // enable REST API or not
+	debugAPI        bool // enable debug in REST API or not
 	webui           bool // enable webui or not
 	pubsub          bool // enable pubsub trigger or not
 	pubsubTrigger   triggerConfigFlags
@@ -110,7 +110,7 @@ func init() {
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.a2a, "a2a", true, "Enable A2A")
 	cloudrunCmd.PersistentFlags().StringVarP(&flags.cloudRun.a2aAgentCardURL, "a2a_agent_url", "a", "http://127.0.0.1:8081", "A2A agent card URL as advertised in the public agent card")
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.api, "api", true, "Enable API")
-	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.debugApi, "debug_api", false, "Enable Debug API (if API is enabled)")
+	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.debugAPI, "debug_api", false, "Enable Debug API - requires '-api'")
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.webui, "webui", true, "Enable Web UI")
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.pubsub, "pubsub", false, "Enable PubSub subrouter")
 	cloudrunCmd.PersistentFlags().IntVar(&flags.cloudRun.pubsubTrigger.maxRetries, "pubsub_max_retries", 3, "Maximum retries for HTTP 429 errors from PubSub triggers")
@@ -128,6 +128,10 @@ func init() {
 func (f *deployCloudRunFlags) computeFlags() error {
 	return util.LogStartStop("Computing flags & preparing temp",
 		func(p util.Printer) error {
+			if f.cloudRun.debugAPI && !f.cloudRun.api {
+				return fmt.Errorf("cannot enable Debug API without having enabled API")
+			}
+
 			absp, err := filepath.Abs(flags.source.entryPointPath)
 			if err != nil {
 				return fmt.Errorf("cannot make an absolute path from '%v': %w", f.source.entryPointPath, err)
@@ -160,10 +164,6 @@ func (f *deployCloudRunFlags) computeFlags() error {
 				f.build.execPath = path.Join(f.build.tempDir, exec)
 			}
 			f.build.dockerfileBuildPath = path.Join(f.build.tempDir, "Dockerfile")
-
-			if f.cloudRun.debugApi && !f.cloudRun.api {
-				return fmt.Errorf("cannot enable Debug API without having enabled API")
-			}
 
 			return nil
 		})
@@ -217,7 +217,7 @@ CMD ["/app/` + f.build.execFile + `", "web", "-port", "` + strconv.Itoa(flags.cl
 			if flags.cloudRun.api {
 				b.WriteString(`, "api", "-webui_address", "127.0.0.1:` + strconv.Itoa(f.proxy.port) + `"`)
 
-				if flags.cloudRun.debugApi {
+				if flags.cloudRun.debugAPI {
 					b.WriteString(`, "-include_debug_api"`)
 				}
 			}
--- a/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
+++ b/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
@@ -110,7 +110,7 @@ func init() {
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.a2a, "a2a", true, "Enable A2A")
 	cloudrunCmd.PersistentFlags().StringVarP(&flags.cloudRun.a2aAgentCardURL, "a2a_agent_url", "a", "http://127.0.0.1:8081", "A2A agent card URL as advertised in the public agent card")
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.api, "api", true, "Enable API")
-	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.debugAPI, "debug_api", false, "Enable Debug API - requires '-api'")
+	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.debugAPI, "debug_api", false, "Enable Debug API - requires '--api'")
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.webui, "webui", true, "Enable Web UI")
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.pubsub, "pubsub", false, "Enable PubSub subrouter")
 	cloudrunCmd.PersistentFlags().IntVar(&flags.cloudRun.pubsubTrigger.maxRetries, "pubsub_max_retries", 3, "Maximum retries for HTTP 429 errors from PubSub triggers")
--- a/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
+++ b/cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
@@ -49,6 +49,7 @@ type cloudRunServiceFlags struct {
 	a2aAgentCardURL string
 	a2a             bool // enable a2a or not
 	api             bool // enable api or not
+	debugApi        bool // enable debug api or not
 	webui           bool // enable webui or not
 	pubsub          bool // enable pubsub trigger or not
 	pubsubTrigger   triggerConfigFlags
@@ -109,6 +110,7 @@ func init() {
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.a2a, "a2a", true, "Enable A2A")
 	cloudrunCmd.PersistentFlags().StringVarP(&flags.cloudRun.a2aAgentCardURL, "a2a_agent_url", "a", "http://127.0.0.1:8081", "A2A agent card URL as advertised in the public agent card")
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.api, "api", true, "Enable API")
+	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.debugApi, "debug_api", false, "Enable Debug API (if API is enabled)")
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.webui, "webui", true, "Enable Web UI")
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.pubsub, "pubsub", false, "Enable PubSub subrouter")
 	cloudrunCmd.PersistentFlags().IntVar(&flags.cloudRun.pubsubTrigger.maxRetries, "pubsub_max_retries", 3, "Maximum retries for HTTP 429 errors from PubSub triggers")
@@ -158,6 +160,10 @@ func (f *deployCloudRunFlags) computeFlags() error {
 				f.build.execPath = path.Join(f.build.tempDir, exec)
 			}
 			f.build.dockerfileBuildPath = path.Join(f.build.tempDir, "Dockerfile")
+
+			if f.cloudRun.debugApi && !f.cloudRun.api {
+				return fmt.Errorf("cannot enable Debug API without having enabled API")
+			}
 
 			return nil
 		})
@@ -210,6 +216,10 @@ CMD ["/app/` + f.build.execFile + `", "web", "-port", "` + strconv.Itoa(flags.cl
 
 			if flags.cloudRun.api {
 				b.WriteString(`, "api", "-webui_address", "127.0.0.1:` + strconv.Itoa(f.proxy.port) + `"`)
+
+				if flags.cloudRun.debugApi {
+					b.WriteString(`, "-include_debug_api"`)
+				}
 			}
 			if flags.cloudRun.a2a {
 				b.WriteString(`, "a2a", "--a2a_agent_url", "` + flags.cloudRun.a2aAgentCardURL + `"`)
--- a/cmd/launcher/web/api/api.go
+++ b/cmd/launcher/web/api/api.go
@@ -86,7 +86,7 @@ func (a *apiLauncher) SetupSubrouters(router *mux.Router, config *launcher.Confi
 		DebugConfig: adkrest.DebugTelemetryConfig{
 			TraceCapacity: a.config.traceCapacity,
 		},
-		DebugApiConfig: adkrest.DebugApiConfig{
+		DebugApiConfig: adkrest.DebugAPIConfig{
 			IncludeDebugAPI: a.config.includeDebugAPI,
 		},
 	})
--- a/cmd/launcher/web/api/api.go
+++ b/cmd/launcher/web/api/api.go
@@ -83,9 +83,11 @@ func (a *apiLauncher) SetupSubrouters(router *mux.Router, config *launcher.Confi
 		ArtifactService: config.ArtifactService,
 		SSEWriteTimeout: a.config.sseWriteTimeout,
 		PluginConfig:    config.PluginConfig,
-		DebugConfig: adkrest.DebugConfig{
-			IncludeDebugAPI:        a.config.includeDebugAPI,
+		DebugConfig: adkrest.DebugTelemetryConfig{
 			TelemetryTraceCapacity: a.config.traceCapacity,
+		},
+		DebugApiConfig: adkrest.DebugApiConfig{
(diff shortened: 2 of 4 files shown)


server/adkrest/handler.go
```diff
@@ -76,6 +81,12 @@ type ServerConfig struct {
 	SSEWriteTimeout time.Duration
 	PluginConfig    runner.PluginConfig
 	DebugConfig     DebugTelemetryConfig
+	DebugApiConfig  DebugApiConfig
+}
+
+// DebugApiConfig contains parameters for the debug api.
+type DebugApiConfig struct {
+	IncludeDebugAPI bool
```
karolpiotrowicz: Worth fixing before this merges rather than after: `DebugApiConfig` should be `DebugAPIConfig` for the initialism, and the field inside it already spells it `IncludeDebugAPI`. The type comment has the same issue in prose — "the debug api".

The reason to do it now is that this is a newly exported identifier. Once it ships in a release, correcting the name costs exactly the compatibility break this restructuring was done to avoid.

cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
```diff
@@ -49,6 +49,7 @@ type cloudRunServiceFlags struct {
 	a2aAgentCardURL string
 	a2a             bool // enable a2a or not
 	api             bool // enable api or not
+	debugApi        bool // enable debug api or not
```
karolpiotrowicz: `debugApi` should be `debugAPI` for the initialism, consistent with `includeDebugAPI` in `cmd/launcher/web/api/api.go`.

cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
```diff
@@ -109,6 +110,7 @@ func init() {
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.a2a, "a2a", true, "Enable A2A")
 	cloudrunCmd.PersistentFlags().StringVarP(&flags.cloudRun.a2aAgentCardURL, "a2a_agent_url", "a", "http://127.0.0.1:8081", "A2A agent card URL as advertised in the public agent card")
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.api, "api", true, "Enable API")
+	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.debugApi, "debug_api", false, "Enable Debug API (if API is enabled)")
```
karolpiotrowicz: This reads as "ignored when `--api` is off", but `computeFlags` hard-errors on that combination rather than ignoring the flag. "requires --api" would match the behaviour.

cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
```diff
@@ -159,6 +161,10 @@ func (f *deployCloudRunFlags) computeFlags() error {
 			}
 			f.build.dockerfileBuildPath = path.Join(f.build.tempDir, "Dockerfile")
 
+			if f.cloudRun.debugApi && !f.cloudRun.api {
+				return fmt.Errorf("cannot enable Debug API without having enabled API")
+			}
```
karolpiotrowicz: This check reads only flags, but it runs after `os.MkdirTemp` has already created the temp directory further up `computeFlags`. `deployOnCloudRun` returns on this error before reaching `cleanTemp()`, so `adkgo deploy cloudrun --api=false --debug_api` leaves a `cloudrun_*` directory behind on every attempt.

Moving it to the top of `computeFlags`, or into a cobra `PreRunE`, avoids creating anything before the arguments are known to be good.

The other error paths in this function leak the same way, so this is an existing pattern rather than something the PR introduced — it just adds a new reliably-reachable instance.

karolpiotrowicz (commented):
Re-reviewed after the two follow-up commits. Moving the opt-in into its own `DebugApiConfig` rather than renaming the existing type is the right call and it does resolve the compatibility break — I re-ran a third-party program that constructs `adkrest.ServerConfig{DebugConfig: adkrest.DebugTelemetryConfig{TraceCapacity: 5000}}` and it now builds against this branch exactly as it does against main, and apidiff reports only `DebugApiConfig: added` and `ServerConfig.DebugApiConfig: added`. Nothing below blocks. A few notes are inline, and two things are worth a decision before this lands.

## The gate also removes a route that is not named "debug"

`DebugAPIRouter` owns three routes, and the third is [`/apps/{app_name}/users/{user_id}/sessions/{session_id}/events/{event_id}/graph`](https://github.com/google/adk-go/blob/03532c9f8aaa75a0c27584b8a73f708e4ec97369/server/adkrest/internal/routers/debug.go#L42-L47). It carries no `/debug` prefix and sits in the ordinary apps/sessions tree, so a caller would not associate it with a flag called `include_debug_api`. Driving `NewServer` and `ServeHTTP` at this commit:

| route | main | this branch, default | this branch, `IncludeDebugAPI: true` |
|---|---|---|---|
| `/debug/trace/session/{id}` | 200 | 404 | 200 |
| `/apps/a/users/u/sessions/s/events/e/graph` | 400 | 404 | 400 |
| `/health` (control) | 200 | 200 | 200 |

[`examples/rest/main.go#L62-L66`](https://github.com/google/adk-go/blob/03532c9f8aaa75a0c27584b8a73f708e4ec97369/examples/rest/main.go#L62-L66) builds a `ServerConfig` with no debug config, so the repository's own example loses all three. As far as I can tell nothing in-tree actually calls the graph endpoint, so this may well be fine — I mainly want it to be a decision rather than a side effect, and worth a line in the description either way.

## Nothing tests the gate

Two edits against the suite: deleting the `if` at [`handler.go#L59`](https://github.com/google/adk-go/blob/03532c9f8aaa75a0c27584b8a73f708e4ec97369/server/adkrest/handler.go#L59) so the debug router always registers again, and inverting it to `if !cfg.DebugApiConfig.IncludeDebugAPI`. Both leave `go test ./server/adkrest/...` green, so this feature could be removed or turned inside out without a test noticing. No test in the tree constructs a non-zero debug config, and the tests in `controllers/debug_test.go` call the handlers directly rather than through `mux`, so they cannot observe registration.

Two table cases on `NewServer` would close it: default config gives 404 on all three patterns, `IncludeDebugAPI: true` gives non-404.

## One for later, not for this PR

[`api.go#L95-L96`](https://github.com/google/adk-go/blob/03532c9f8aaa75a0c27584b8a73f708e4ec97369/cmd/launcher/web/api/api.go#L95-L96) registers the span and log processors unconditionally, and `NewServer` builds the telemetry store before the gate, so with the flag off the process still captures every span into a buffer nothing can read. That is unchanged from main — this PR adds no retention — but it does remove the only reader, so in the default configuration the cost now buys nothing. Better as a follow-up than as scope added here.

One thing I checked that turned out fine: the default flip does not break the trace UI. The bundled Web UI calls `/dev/apps/{app}/debug/trace/...` and no Go route serves a `/dev` prefix, so those calls already 404 on main.

server/adkrest/handler.go
kdroste-google: fixed

cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
kdroste-google: fixed

cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
kdroste-google: fixed

server/adkrest/handler.go
```diff
@@ -76,6 +81,12 @@ type ServerConfig struct {
 	SSEWriteTimeout time.Duration
 	PluginConfig    runner.PluginConfig
 	DebugConfig     DebugTelemetryConfig
+	DebugApiConfig  DebugAPIConfig
```
karolpiotrowicz: The type rename landed but the field did not — this is still `DebugApiConfig` while its type is now `DebugAPIConfig`. apidiff records the field as part of the new public surface:

```
Compatible changes:
- DebugAPIConfig: added
- ServerConfig.DebugApiConfig: added
```

So the exported field name is as permanent as the type name was, for the same reason. Renaming it to `DebugAPIConfig` also touches [`handler.go#L59`](https://github.com/google/adk-go/blob/1abb4fa4031b73f83ec83c4194c6a56c5b150b21/server/adkrest/handler.go#L59) and [`api.go#L89`](https://github.com/google/adk-go/blob/1abb4fa4031b73f83ec83c4194c6a56c5b150b21/cmd/launcher/web/api/api.go#L89), which are the only other uses.

cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
```diff
@@ -109,6 +110,7 @@ func init() {
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.a2a, "a2a", true, "Enable A2A")
 	cloudrunCmd.PersistentFlags().StringVarP(&flags.cloudRun.a2aAgentCardURL, "a2a_agent_url", "a", "http://127.0.0.1:8081", "A2A agent card URL as advertised in the public agent card")
 	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.api, "api", true, "Enable API")
+	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.debugAPI, "debug_api", false, "Enable Debug API - requires '-api'")
```
karolpiotrowicz: `-api` is not the spelling of that flag, and getting it wrong here is quiet rather than loud. `-a` is registered as the shorthand for `a2a_agent_url`, so pflag reads `-api` as `-a` with the value `pi`:

```
parse("-api")  err = <nil>
a2aAgentCardURL: before="http://127.0.0.1:8081" after="pi"
api flag value:  true          # unchanged, it defaults to true
```

An operator who copies the help text literally gets no error and a corrupted agent card URL. `--api` is the correct spelling. This one is on me — my note last round prompted the rewording.

karolpiotrowicz (commented):
Re-checked at `1abb4fa4`. All four inline points from the [previous round](https://github.com/google/adk-go/pull/1413#pullrequestreview-5041337896) are addressed — the type is `DebugAPIConfig`, `debugAPI` has the initialism, and the `--debug_api` validation now runs before `os.MkdirTemp` so the rejected-argument path no longer leaves a temp directory behind. Build, the full suite, vet, gofmt and apidiff are all clean at this commit, and apidiff reports additive changes only.

Two things left inline, both small and both worth doing before this merges rather than after. One of them is a consequence of the wording change I asked for last round, so apologies for the round trip.

Still open from the previous round, unchanged and not re-argued here: the `…/events/{event_id}/graph` route is still gated off by default along with the two `/debug/trace` routes, and there is still no test that would notice if the gate were removed or inverted. Both are described in [the earlier review](https://github.com/google/adk-go/pull/1413#pullrequestreview-5041337896).

cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
kdroste-google: fixed

server/adkrest/handler.go
kdroste-google: fixed

server/adkrest/handler.go
```diff
@@ -76,6 +81,12 @@ type ServerConfig struct {
 	SSEWriteTimeout time.Duration
 	PluginConfig    runner.PluginConfig
 	DebugConfig     DebugTelemetryConfig
+	DebugAPIConfig  DebugAPIConfig
+}
+
+// DebugAPIConfig contains parameters for the debug API.
+type DebugAPIConfig struct {
+	IncludeDebugAPI bool
```
karolpiotrowicz, line 90: If you take the documenting route, the comment has to name the routes — a generic "enables the debug API" would not help, because the surprising part is *which* routes are in the set, not that debug routes are gated. Something like:

```go
// IncludeDebugAPI registers the debug API routes. It is false by default.
//
// This gates all three routes owned by the debug API, including the event
// graph route, whose path does not contain "debug":
//
//	GET /debug/trace/{event_id}
//	GET /debug/trace/session/{session_id}
//	GET /apps/{app_name}/users/{user_id}/sessions/{session_id}/events/{event_id}/graph
IncludeDebugAPI bool
```

server/adkrest/handler.go
```diff
@@ -48,14 +48,19 @@ func NewServer(cfg ServerConfig) (*Server, error) {
...
+	subrouters := []routers.Router{
 		routers.NewSessionsAPIRouter(controllers.NewSessionsAPIController(cfg.SessionService)),
 		routers.NewRuntimeAPIRouter(controllers.NewRuntimeAPIController(cfg.SessionService, cfg.MemoryService, cfg.AgentLoader, cfg.ArtifactService, cfg.SSEWriteTimeout, cfg.PluginConfig, false)),
 		routers.NewAppsAPIRouter(controllers.NewAppsAPIController(cfg.AgentLoader)),
-		routers.NewDebugAPIRouter(controllers.NewDebugAPIController(cfg.SessionService, cfg.AgentLoader, debugTelemetry)),
 		routers.NewArtifactsAPIRouter(controllers.NewArtifactsAPIController(cfg.ArtifactService)),
 		&routers.EvalAPIRouter{},
-	)
+	}
+	if cfg.DebugAPIConfig.IncludeDebugAPI {
```
karolpiotrowicz, line 59: Nothing currently tests this branch. Deleting the `if` so the router always registers, or inverting it to `if !cfg.DebugAPIConfig.IncludeDebugAPI`, both leave `go test ./server/adkrest/...` completely green — so the feature could be removed or turned inside out and CI would not notice.

The one wrinkle in testing it is that a status code alone does not discriminate: with the gate on, `/debug/trace/{event_id}` still answers 404, just from the handler rather than from the router. The response body is what separates the two — gorilla/mux writes `404 page not found` when nothing matched, and a registered handler writes its own body.

<details>
<summary>A test that covers it — passes as written, and fails against both the deleted and the inverted gate</summary>

```go
// debugRoutes are every route owned by the debug API router, including the
// event graph route, whose path does not contain "debug".
var debugRoutes = []string{
	"/debug/trace/evt1",
	"/debug/trace/session/sess1",
	"/apps/app1/users/user1/sessions/sess1/events/evt1/graph",
}

// muxNotFound is what gorilla/mux writes when no route matches. A registered
// handler that happens to answer 404 writes its own body, so the body is what
// distinguishes "not routed" from "routed and rejected".
const muxNotFound = "404 page not found\n"

func TestNewServerDebugAPIGate(t *testing.T) {
	for _, tc := range []struct {
		name       string
		include    bool
		wantRouted bool
	}{
		{name: "omitted by default", include: false, wantRouted: false},
		{name: "included when opted in", include: true, wantRouted: true},
	} {
			srv, err := adkrest.NewServer(adkrest.ServerConfig{
				SessionService: session.InMemoryService(),
				AgentLoader:    agent.NewSingleLoader(nil),
				DebugAPIConfig: adkrest.DebugAPIConfig{IncludeDebugAPI: tc.include},
			})
			if err != nil {
				t.Fatalf("NewServer() error = %v", err)
			}
			for _, route := range debugRoutes {
				rr := httptest.NewRecorder()
				srv.ServeHTTP(rr, httptest.NewRequest(http.MethodGet, route, nil))
				routed := rr.Body.String() != muxNotFound
				if routed != tc.wantRouted {
					t.Errorf("GET %s: routed = %v, want %v (code %d, body %q)",
						route, routed, tc.wantRouted, rr.Code, rr.Body.String())
				}
			}
		})
	}
}
```

Run with `go test ./server/adkrest/ -run TestNewServerDebugAPIGate`.

</details>

cmd/launcher/web/api/api.go
```diff
@@ -143,6 +147,7 @@ func NewLauncher() weblauncher.Sublauncher {
 	fs.StringVar(&config.pathPrefix, "path_prefix", "/api", "ADK REST API path prefix. Default is '/api'.")
 	fs.DurationVar(&config.sseWriteTimeout, "sse-write-timeout", 120*time.Second, "SSE server write timeout (i.e. '10s', '2m' - see time.ParseDuration for details) - for writing the SSE response after reading the headers & body")
 	fs.IntVar(&config.traceCapacity, "trace_capacity", 10000, "Maximum number of traces to keep in memory.")
+	fs.BoolVar(&config.includeDebugAPI, "include_debug_api", false, "The debug api endpoint will be included in the API if and only if the flag is set to true.")
```
karolpiotrowicz, line 150: "The debug api endpoint" is singular, but the flag controls three routes, and one of them is the event graph route rather than a debug endpoint. Worth matching the wording to what it does, and `API` for the initialism while you are here.

karolpiotrowicz (commented):
Re-checked at `50e2ed0b` — the field rename and the `--api` correction both look right, and I confirmed `--api --debug_api` now parses cleanly with `a2a_agent_url` untouched. Build, suite, vet, gofmt, staticcheck and apidiff are all clean.

One design question left, plus two small things inline.

## The event graph route is gated along with the trace routes

`DebugAPIRouter` owns three routes and the gate covers all of them:

```
GET /debug/trace/{event_id}
GET /debug/trace/session/{session_id}
GET /apps/{app_name}/users/{user_id}/sessions/{session_id}/events/{event_id}/graph
```

The third has no `/debug` in its path and sits in the ordinary apps/sessions tree, so a caller has no way to associate it with a flag named `include_debug_api`. Measured at this commit through `NewServer` and `ServeHTTP`, the graph route goes from 400 on main to 404 by default here, while `/health` stays 200 as a control. [`examples/rest/main.go#L62-L66`](https://github.com/google/adk-go/blob/50e2ed0b28ceacc66bd9ac73cae8ebe5362bbd76/examples/rest/main.go#L62-L66) sets no debug config, so the repo's own example is in that group.

Worth saying plainly: apidiff cannot catch this one. The break is behavioural, not source-level, so a caller who never named the config type still compiles and simply starts getting 404s.

Three ways to go, and I do not think any of them is obviously right:

1. **Document the coupling.** Keep the gate as it is and say what it covers — a doc comment on the field naming all three routes (suggested text inline), a corrected flag description, and a line in the release notes. Cheapest, and the notes are the part that actually reaches an existing caller, since nobody reads godoc for a field they do not know exists.
2. **Move the graph route off `DebugAPIRouter`.** Right if it is ordinary API surface that happens to be implemented by the debug controller. Makes the question disappear instead of documenting it.
3. **Leave it as is.** Defensible — nothing in-tree reaches that route, and the bundled Web UI cannot either, since it calls `/dev/apps/…` and no Go route serves a `/dev` prefix. Even then the release note still matters for callers outside this repo.

Option 1 is what I would pick, but it is your call and option 2 is a reasonable read of what the endpoint is.

## Two doc comments that now describe something that is off by default

[`SpanProcessor`](https://github.com/google/adk-go/blob/50e2ed0b28ceacc66bd9ac73cae8ebe5362bbd76/server/adkrest/handler.go#L109-L113) and [`LogProcessor`](https://github.com/google/adk-go/blob/50e2ed0b28ceacc66bd9ac73cae8ebe5362bbd76/server/adkrest/handler.go#L115-L119) both tell callers the processors capture data "used for /debug/trace endpoint of the ADK REST API server", and invite registering them on the application's own provider. With the new default that endpoint is not registered, so a library caller who follows the comment feeds a store nothing can read. Both sit outside the diff, so they are here rather than on the lines.

server/adkrest/handler.go
kdroste-google: Fixed (less wordy)

server/adkrest/handler.go
kdroste-google: fixed

cmd/launcher/web/api/api.go
kdroste-google: not important

karolpiotrowicz (approved):
Looks good to me.

I re-ran the new gate test against two deliberate breakages to check it earns its place: deleting the `if` so the debug router always registers, and inverting it. Both make `TestNewServerDebugAPIGate` fail, so the test genuinely pins the behaviour rather than just executing it. Coverage on `server/adkrest` also goes from 78.6% to 82.4%, and apidiff reports additive changes only.

One thing before merging: the branch is behind `main` and will need a merge.

The unconditional span and log processor registration is still worth a follow-up at some point — that predates this PR and is not something to grow this change with.

PR #1414, opened 2026-08-28
feat(examples): add OrcaRouter model integration example

bangla24bdrang-lab:
## Summary

ADK Go is a model-agnostic, code-first toolkit: the [OpenAI
example](examples/openai) shows that any endpoint implementing the OpenAI
Responses API can back an `llmagent` through the same `model.LLM` wiring as
Gemini. This PR adds a runnable example that points that wiring at
[OrcaRouter](https://www.orcarouter.ai), an OpenAI-compatible AI gateway that
routes to many models behind a single base URL.

OrcaRouter is built for both models and agents. Like OpenRouter, it exposes a
provider/model namespace across many models — but it also combines adaptive
routing, automatic failover, zero-markup inference, observability, guardrails,
and agent-tool governance behind the same endpoint. Adding orcarouter as a
first-class provider means this project's users can use that stack directly,
without treating OrcaRouter as an anonymous custom base URL. It also runs
gateway-level, zero-trust security for AI agents on the same endpoint —
screening every prompt/response and governing every tool call on a
default-deny basis, with no application code changes.

## Changes

- Add `examples/orcarouter/main.go`, mirroring `examples/openai`:
  `openaimodel.NewModel` with `BaseURL: "https://api.orcarouter.ai/v1"`,
  `ORCAROUTER_API_KEY` (env), a `get_weather` function tool, and the same
  launcher wiring — no framework or `openaimodel` changes.
- Add `examples/orcarouter/README.md` with a configuration table, run
  instructions, and the model IDs available on the gateway
  (`orcarouter/free`, `orcarouter/fusion`, `openai/gpt-4.1`, …).
- No new dependencies; no public API surface changes.

## Validation

- `go build -mod=readonly work` — pass
- `go test -mod=readonly -count=1 ./model/... ./examples/orcarouter/` — pass
- `golangci-lint run` — 0 issues
- `go mod tidy -diff` — no changes
- Live check: ran `openaimodel.NewModel` with `BaseURL: "https://api.orcarouter.ai/v1"`
  against the gateway; both non-streaming and streaming `GenerateContent`
  returned model output (HTTP 200) for `orcarouter/fusion-flash`.

## Testing plan

- [x] Unit: existing `model/openaimodel` tests remain green
- [x] Manual E2E: live OrcaRouter call via the example's exact client config
- [x] `go run ./examples/orcarouter/ console` with `ORCAROUTER_API_KEY` set

---

I'm an engineer on the OrcaRouter team.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

### DIFF
--- a/examples/orcarouter/README.md
+++ b/examples/orcarouter/README.md
@@ -1,0 +1,76 @@
+# OrcaRouter model integration
+
+Runs an ordinary ADK `llmagent` — with a tool — on an **OrcaRouter** model
+instead of Gemini. OrcaRouter is an OpenAI-compatible AI gateway that routes to
+many models behind a single base URL, using the `google.golang.org/adk/v2/model/openai`
+package (the `openaimodel.NewModel` constructor). Because OrcaRouter serves the
+OpenAI **Responses API**, this sample shares the exact same `model.LLM` wiring as
+the [OpenAI example](../openai/README.md); the only difference is the client
+config, which points at OrcaRouter's endpoint.
+
+- **Concept:** Swap `gemini.NewModel(...)` for `openaimodel.NewModel(...)` with OrcaRouter's base URL; agents, tools, the runner, and the launcher are unchanged.
+- **Needs LLM?** Yes (OrcaRouter API key)
+
+## Goal
+
+Show that ADK is model-agnostic beyond the providers it ships constructors for:
+OrcaRouter plugs into the exact same `llmagent.New` / launcher wiring as the
+[quickstart](../quickstart) and the [OpenAI example](../openai). The only
+difference is the constructor —
+
+```go
+model, err := openaimodel.NewModel(ctx, "orcarouter/free", &openaimodel.ClientConfig{
+    APIKey:  os.Getenv("ORCAROUTER_API_KEY"),
+    BaseURL: "https://api.orcarouter.ai/v1",
+})
+```
+
+— and everything downstream stays identical. The sample also registers a
+`get_weather` function tool to demonstrate that OrcaRouter tool calling flows
+through ADK's normal `functiontool` path.
+
+## Configuration
+
+| Variable            | Required            | Default            | Purpose                          |
+| ------------------- | ------------------- | ------------------ | -------------------------------- |
+| `ORCAROUTER_API_KEY`| Yes                 | —                  | Your OrcaRouter API key (`sk-orca-...`). |
+| `ORCAROUTER_MODEL`  | No                  | `orcarouter/free`  | Model name to serve.             |
+
+Any model ID on the gateway works — for example `orcarouter/fusion`,
+`openai/gpt-4.1`, `anthropic/claude-opus-4.5`, or `deepseek/deepseek-v4-pro`.
+
+## Running the sample
+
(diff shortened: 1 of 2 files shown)


wolo-lab:
Declining this one on scope rather than on quality — the code is correct and the gates are clean. Let me set out the reasoning and offer something concrete in its place, because the gap you're pointing at is real.

**Why not a separate directory.** `examples/openai` already covers this mechanism. It documents that the same `model.LLM` serves any endpoint implementing the Responses API, and it ships a worked base-URL recipe at [`examples/openai/README.md:54-58`](https://github.com/google/adk-go/blob/3f69c5059e2adc6c127aa10d595b547e98f3da0c/examples/openai/README.md#L54-L58). Reaching OrcaRouter today takes three environment variables rather than a new sample:

```bash
export OPENAI_BASE_URL=https://api.orcarouter.ai/v1
export OPENAI_API_KEY=sk-orca-...
export OPENAI_MODEL=orcarouter/free
go run ./examples/openai/ console
```

Comparing the two Go files, they differ in six places, five of which are string substitutions. The sixth replaces the environment read with a hardcoded base URL, which leaves the new sample less configurable than the one it copies.

**The precedent is what decides it.** I probed the other OpenAI-compatible gateways to see whether OrcaRouter had a distinguishing property here — `openaimodel` speaks only the Responses API and not Chat Completions, so I expected that to rule most gateways out. It ruled out none of the three I tested: OpenRouter, Together and Groq each present the same routing signature as OrcaRouter, with `/v1/responses` and `/v1/chat/completions` both answering and an unknown path returning 404. That leaves no criterion that admits this directory and declines those three, and each would be another ~185 lines of near-identical Go maintained in parallel indefinitely.

**What I would rather have, and would be glad to take from you.** The real gap is that [`examples/openai/README.md:5-8`](https://github.com/google/adk-go/blob/3f69c5059e2adc6c127aa10d595b547e98f3da0c/examples/openai/README.md#L5-L8) names only local runtimes — Ollama, LM Studio, vLLM — and says nothing about hosted gateways as a category. A short PR against that file would close it:

- Add hosted OpenAI-compatible gateways to that sentence as a category, without naming individual vendors.
- Add one line on how a reader checks whether a given provider qualifies. The criterion is sharp and not obvious: the endpoint has to expose `/v1/responses`, not only `/v1/chat/completions`. That is the same point the caveat at [L7-L8](https://github.com/google/adk-go/blob/3f69c5059e2adc6c127aa10d595b547e98f3da0c/examples/openai/README.md#L7-L8) makes, stated positively so a reader can act on it.
- While you are in the file, [line 4](https://github.com/google/adk-go/blob/3f69c5059e2adc6c127aa10d595b547e98f3da0c/examples/openai/README.md#L4) names the package as `google.golang.org/adk/v2/model/openai`, which does not exist — the real import path is `model/openaimodel`. That one is our bug, and it is currently the only occurrence in the repo.

That routes users to OrcaRouter through documentation that stays accurate as the ecosystem moves, rather than through a directory that commits us to one per vendor.

One practical note either way: the CLA check is currently failing, and it would need to be signed before anything can land.

PR #1347, opened 2026-08-17
fix(adkrest): guard against nil events in RunHandler and RunLiveHandler (#1346)

chill-czar:
### Link to Issue or Description of Change

- Closes: #1346

### Problem

1. In `runner.Run` and `runner.RunLive`, when an agent or intermediate callback yields a `nil` `*session.Event`, the runner loops unconditionally dereference `event.LLMResponse.Partial`, crashing with a nil pointer dereference panic.
2. In `server/adkrest/controllers/runtime.go`, `RunLiveHandler` and `RunHandler` pass `*event` directly to `models.FromSessionEvent(*event)` without checking `if event == nil` (unlike `RunSSEHandler`), crashing active HTTP and WebSocket request handlers.

### Solution

1. Add defensive `if event == nil { continue }` guards in `runner.Run` and `runner.RunLive`.
2. Add defensive `if event == nil { continue }` guards in `RunHandler` and `RunLiveHandler` before dereferencing `*event`.
3. Add regression tests in `runner/runner_test.go` (`TestRunner_Run_GuardsNilEvents`) and `server/adkrest/controllers/runtime_test.go` (`TestRunHandler_GuardsNilEvents` and `TestRunSSEHandler` with nil event stream).

### Testing Plan

- [x] Unit tests added:
  - `TestRunner_Run_GuardsNilEvents` in `runner/runner_test.go`
  - `TestRunHandler_GuardsNilEvents` in `server/adkrest/controllers/runtime_test.go`
  - `TestRunSSEHandler` (with nil events in stream) in `server/adkrest/controllers/runtime_test.go`
- [x] Confirmed tests fail without the fix and pass with it.
- [x] All repository verification commands pass:
  ```text
  go test -v -race -count=1 ./runner/... ./server/adkrest/...
  go build -mod=readonly work
  golangci-lint run
  go mod tidy -diff
  ```

### Checklist

- [x] I have read the [CONTRIBUTING.md](./CONTRIBUTING.md) document.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have manually tested my changes end-to-end.
- [x] Any dependent changes have been merged and published in downstream modules.

PR #1417, opened 2026-08-28
fix(runner): guard against nil event yielded by agent

### LINKED ISSUE #1416: Runner panics on a nil event yielded by an agent
**Describe the Bug:**

`Runner.Run` dereferences the event yielded by an agent without checking it for nil, so an agent that yields `(nil, nil)` crashes the process instead of being rejected or skipped.

The same loop already guards this exact condition 22 lines earlier, which suggests the missing check is an oversight rather than a deliberate contract:

- guarded — https://github.com/google/adk-go/blob/245ba284ed51f0f58355a6233f0b1ae01f252160/runner/runner.go#L327
- unguarded, panics here — https://github.com/google/adk-go/blob/245ba284ed51f0f58355a6233f0b1ae01f252160/runner/runner.go#L349

**Steps to Reproduce:**

1. Create a module depending on `google.golang.org/adk/v2 v2.2.0` or later.
2. Save the program under "Minimal Reproduction Code" below as `main.go`.
3. Run `go run .`
4. The process exits with SIGSEGV and the stack trace shown under "Observed Behavior".

**Expected Behavior:**

A nil event should not take down the process. Either the runner skips it the way the earlier branch does, or it surfaces a descriptive error to the caller naming the offending agent, so the author can find the contract violation. Both are reasonable and the choice is yours — the current outcome, an unrecoverable nil dereference inside library code, is not.

**Observed Behavior:**

```text
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x9c1401]

goroutine 1 [running]:
main.main.(*Runner).Run.func2-range1(0x486e3a?, {0x0?, 0x0?})
	runner/runner.go:349 +0x241
google.golang.org/adk/v2/agent.(*agent).Run.func1.WrapYield[...].2({0x0?, 0x0?})
	internal/telemetry/telemetry.go:222 +0x5a
google.golang.org/adk/v2/agent.(*agent).Run.func1-range1(0x30?, {0x0?, 0x0?})
	agent/agent.go:198 +0xce
main.main.func1.1(0x11b9840?)
	main.go:19 +0x1c
google.golang.org/adk/v2/agent.(*agent).Run.func1(0x2a47f9f6d400)
	agent/agent.go:194 +0x4a2
main.main.(*Runner).Run.func2(0x2a47fa0f41f0)
	runner/runner.go:319 +0x8a6
main.main()
	main.go:38 +0x403
exit status 2
```

**Environment Details:**

- ADK Library Version: v2.2.0-35-g245ba284 (also present in v2.2.0)
- OS: Linux
- Go Version: go1.26.6

**Model Information:**

- Which model is being used: N/A. No model is involved — the reproduction uses a plain agent built with `agent.New`, and the panic happens in the runner before any model call.

---

## 🟡 Optional Information

**Regression:**

No. `git log -S` on the condition returns only #168 (`Update event.LLMResponse to be a struct`) and #833 (`feat(live): Add core bidirectional streaming support`), neither of which removed a guard. The nil check at line 327 was added alongside the newer code paths and the older site at 349 was not updated to match, so this has been reachable for as long as both branches have coexisted.

**Logs:**

See the stack trace above. There is nothing further — the process dies at that frame.

**Screenshots / Video:**

N/A

**Additional Context:**

How much damage this does depends on how the agent was constructed, which is worth knowing before judging severity.

Through `llmagent`, the run goes via the workflow scheduler, whose `recover` at https://github.com/google/adk-go/blob/245ba284ed51f0f58355a6233f0b1ae01f252160/workflow/scheduler.go#L452 catches the panic and turns it into an error event. A server stays up and the client receives:

```text
event: error
data: {"error":"node \"agent\" panicked: runtime error: invalid memory address or nil pointer dereference"}
```

That is a degraded message rather than a crash, but the underlying dereference is the same one, and the text it produces tells the user nothing actionable.

An agent built directly with `agent.New(agent.Config{Run: ...})` does not go through the scheduler, so the panic propagates to the caller and the process exits. That is the case reproduced above.

Two constraints on any fix:

The `adkrest` SSE handler already has an `if event == nil { continue }` guard, added in #734. It never executes for this case, because the runner dereferences first, so guarding at the handler will not help.

`RunLive` has the same asymmetry and would need the same treatment: guarded at https://github.com/google/adk-go/blob/245ba284ed51f0f58355a6233f0b1ae01f252160/runner/runner.go#L523, unguarded at https://github.com/google/adk-go/blob/245ba284ed51f0f58355a6233f0b1ae01f252160/runner/runner.go#L565 and https://github.com/google/adk-go/blob/245ba284ed51f0f58355a6233f0b1ae01f252160/runner/runner.go#L595. I have not exercised the live path, so I am reporting the shape rather than a confirmed second reproduction.

This surfaced while confirming #383 was fixed. It is a separate defect. #383 was about a model returning a proper error, which the current code handles correctly.

```go
package main

import (
	"context"
	"fmt"
	"iter"

	"google.golang.org/genai"
)

func main() {
		Name: "nil_yielder",
				yield(nil, nil) // no event, no error
			}
		},
	})
	if err != nil {
		panic(err)
	}

		AppName:           "app",
		Agent:             a,
		SessionService:    session.InMemoryService(),
		AutoCreateSession: true,
	})
	if err != nil {
		panic(err)
	}

	msg := genai.NewContentFromText("hello", genai.RoleUser)
	for ev, err := range r.Run(context.Background(), "u", "s", msg, agent.RunConfig{}) {
		fmt.Println(ev, err)
	}
}
```

- Always (100%)

sundeep8967:
## Problem
When an agent yields `(nil, nil)`, `Runner.Run` and `Runner.RunLive` dereference `event.LLMResponse.Partial` without checking whether `event == nil`, causing an unrecoverable SIGSEGV nil pointer dereference panic.

## Solution
- Added `if event == nil` guards in `Runner.Run` and `Runner.RunLive` to safely yield `(nil, nil)` and continue rather than crashing.
- Added unit test `TestRunner_NilEventYieldedDoesNotPanic` in `runner/runner_test.go` verifying that yielding `(nil, nil)` completes without panics.

Fixes #1416


wolo-lab:
The guard removes the dereference from `runner.go`, and for consumers that already nil-check their events this is a real improvement — `server/adka2a` and the telemetry test helper go from crashing to working, because on `main` the panic happened before the runner ever yielded, so nothing downstream could handle it. That part is solid.

The problem is what the guard does instead of skipping. [`runner.go:327-332`](https://github.com/google/adk-go/blob/7222baedc8a77d2dcfd10a0e31ba88bad5bc76a3/runner/runner.go#L327-L332) forwards the nil event to the caller with `yield(nil, nil)`, which makes `(nil, nil)` a legal element of a public `iter.Seq2[*session.Event, error]` stream. The consumption pattern documented in `AGENTS.md` is `for event, err := range r.Run(...)` followed by reading `event.LLMResponse.Content`, and that pattern has no reason to nil-check the event once `err` is nil. So the crash does not go away, it moves into the callers.

## The reported crash still happens

Our own CLI console is one of those callers. [`console.go:215-216`](https://github.com/google/adk-go/blob/7222baedc8a77d2dcfd10a0e31ba88bad5bc76a3/cmd/launcher/console/console.go#L215-L216) dereferences the event with no nil check, so `adkgo` still dies on the exact scenario from #1416 — just one frame later.

Running the scenario through the real console launcher, with the branch applied ([7222bae](https://github.com/google/adk-go/commit/7222baedc8a77d2dcfd10a0e31ba88bad5bc76a3)):

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0]
	cmd/launcher/console/console.go:216
	runner/runner.go:328
	runner/runner.go:319
	cmd/launcher/console/console.go:209
```

Against the merge base ([f741f47](https://github.com/google/adk-go/commit/f741f478cfeb5ccecf4d6a0dedc70b18717648de)) the same scenario dies at `runner/runner.go:349`. Same outcome for the user, different frame.

Add to `cmd/launcher/console/`, then `go test -run TestConsoleLauncher_NilEventFromAgent ./cmd/launcher/console/`:

```go
type nilLoader struct{ a agent.Agent }

func (l nilLoader) ListAgents() []string                  { return []string{"nil_yielder"} }
func (l nilLoader) LoadAgent(string) (agent.Agent, error) { return l.a, nil }
func (l nilLoader) RootAgent() agent.Agent                { return l.a }

func TestConsoleLauncher_NilEventFromAgent(t *testing.T) {
		Name: "nil_yielder",
			return func(yield func(*session.Event, error) bool) { yield(nil, nil) }
		},
	})
	if err != nil {
		t.Fatalf("agent.New() error = %v", err)
	}
	rd, wr, _ := os.Pipe()
	orig := os.Stdin
	os.Stdin = rd
	t.Cleanup(func() { os.Stdin = orig })
	go func() { wr.WriteString("hello\n"); time.Sleep(2 * time.Second); wr.Close() }()

	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
	defer cancel()
	_ = NewLauncher().Run(ctx, &launcher.Config{AgentLoader: nilLoader{a: a}})
}
```

</details>

The REST server has the same shape in three places, all unguarded: [`runtime.go:64`](https://github.com/google/adk-go/blob/7222baedc8a77d2dcfd10a0e31ba88bad5bc76a3/server/adkrest/controllers/runtime.go#L64), [`:158`](https://github.com/google/adk-go/blob/7222baedc8a77d2dcfd10a0e31ba88bad5bc76a3/server/adkrest/controllers/runtime.go#L158) and [`:383`](https://github.com/google/adk-go/blob/7222baedc8a77d2dcfd10a0e31ba88bad5bc76a3/server/adkrest/controllers/runtime.go#L383) each call `models.FromSessionEvent(*event)`. Driving `RunHandler` with a nil-yielding agent panics at line 64 on this branch.

Skipping the event instead of forwarding it fixes all four call sites without touching any of them. Swapping both new blocks for `if event == nil { continue }` makes the console reproduction above pass, and leaves `./runner/...` green.

That is also what #1416 asked for — either skip it the way the earlier branch does, or surface a descriptive error naming the offending agent. Forwarding `(nil, nil)` is a third option, and it is the one where the caller cannot tell a broken agent from a well-formed event.

## Worth knowing before you rework this

#1347 is open against the same defect and already takes the skip approach, in both `Run` and `RunLive`, plus nil guards in the REST handlers. It is worth coordinating with that author rather than both of you carrying a fix — that is a maintainer problem rather than anything you did wrong.

## Smaller things

`runner_test.go` has a trailing blank line at [line 678](https://github.com/google/adk-go/blob/7222baedc8a77d2dcfd10a0e31ba88bad5bc76a3/runner/runner_test.go#L678), so `golangci-lint run ./runner/...` exits 1 on gofumpt. `gofmt -w` clears it. Worth flagging that CI never ran the build, test and lint workflow on this PR — it is held for maintainer approval on a first contribution — so there was no way for you to see this.

The new test passes unchanged whether the runner forwards the nil or skips it. Its only assertion sits inside the range body, so under a skip the body never executes and nothing is checked. An event count, plus a real event yielded after the nil, would pin the behaviour down — #1347's test does exactly that.

The `RunLive` half of the change is never executed by any test. `live_runner_test.go` covers `RunLive` itself, but no test yields it a nil, so those lines ship unexercised.

sundeep8967:
Thanks @wolo-lab for the thorough analysis!

Updated the PR accordingly:
- In `runner.go`, replaced forwarding of `(nil, nil)` with `continue` in both `Run()` and `RunLive()`, so nil events yielded by misbehaved agents are safely skipped rather than leaked to downstream callers/launchers.
- Updated `runner_test.go` to assert that nil events are skipped and ran `go fmt` to clean up formatting. All unit tests pass cleanly.

wolo-lab (approved):
This does what #1416 asked for, and the crash is genuinely gone. I re-ran the reproduction that failed on the previous revision — driving the real `adkgo` console launcher with an agent that yields `(nil, nil)` — and it now passes where it previously died at [`console.go:216`](https://github.com/google/adk-go/blob/75987a53cad7152e6cb6ce7077a005f42bb2bb78/cmd/launcher/console/console.go#L216). `gofmt` and `golangci-lint` are clean too. I also ran the nil handling across twelve input shapes (nil alone, nil then a real event, real/nil/real, nil with an error, partial then nil, nil bursts, consumer breaking mid-stream) against both `Run` and `RunLive`, and all twenty-four combinations behave correctly.

One thing worth fixing before this lands, and it is a consequence of the change I asked for last time rather than anything you got wrong.

## The nil branch no longer gives the consumer a turn

[`runner.go:327-329`](https://github.com/google/adk-go/blob/75987a53cad7152e6cb6ce7077a005f42bb2bb78/runner/runner.go#L327-L329) now skips without calling `yield`, and neither runner loop consults `ctx`. So for an agent that yields nils continuously, the consumer's `break` can never run — its loop body is never entered. I measured this: the consumer body executed zero times while the runner absorbed five million nils, stopping only because my harness capped it. On the previous revision the consumer got one iteration and exited cleanly, and on `main` it panics on the first nil.

To be clear about severity: nothing that works today regresses, because on `main` every agent that could trigger this dies on its first nil. The trigger is also a compile-time agent supplied by the application author, not anything a caller can send. So this is not a merge blocker — but it is a loop in a `ctx`-taking library function that can run unbounded while ignoring `ctx`, and the fix belongs in the three lines you are already touching:

```go
if event == nil {
    if err := ctx.Err(); err != nil {
        yield(nil, err)
        return
    }
    continue
}
```

While you are there, consider logging the skip. `runner.go` already reports anomalies of this kind at [line 725](https://github.com/google/adk-go/blob/75987a53cad7152e6cb6ce7077a005f42bb2bb78/runner/runner.go#L725) and [line 739](https://github.com/google/adk-go/blob/75987a53cad7152e6cb6ce7077a005f42bb2bb78/runner/runner.go#L739), and #1416's reason for offering the descriptive-error option was so the agent author can find the contract violation. A silent skip leaves nothing to find.

## Two test gaps

The rewritten test does pin the design — reverting just the implementation to the previous `yield(nil, nil)` version makes it fail, which is exactly what it should do. What it does not pin is what happens *after* a nil. Changing [line 328](https://github.com/google/adk-go/blob/75987a53cad7152e6cb6ce7077a005f42bb2bb78/runner/runner.go#L328) from `continue` to `break`, or to `return`, leaves the test passing and the whole `./runner/` package green, while in production either one would silently truncate the rest of the run. Yielding a real event after the nil and asserting it arrives closes that.

The `RunLive` guard at [line 527](https://github.com/google/adk-go/blob/75987a53cad7152e6cb6ce7077a005f42bb2bb78/runner/runner.go#L527) has no coverage from the tests in this PR — the branch never executes during the suite. It behaves correctly under the twelve shapes I ran, so this is about protecting it rather than fixing it. `live_runner_test.go` already has the harness, so one case there is enough.

jjsasha63:
The `ctx` point and the two test gaps from the review above did not make it in before this merged. I have tracked them in #1471, with a repro and a patch.

The nil branch skips without calling `yield`, so the consumer never gets a turn and its `break` cannot fire, and neither loop checks `ctx`. With an already-cancelled context, `Run` absorbed 160 million nils in three seconds before the harness gave up. Nothing regressed against the pre-merge behavior, where the same agent panicked on its first nil.

PR #1418, opened 2026-08-28
fix: pin a fetched agent card's interface URL(s) to the origin it was fetched from

prasanna8585:
`NewAgentCardProvider` resolves an agent card over http(s) by delegating entirely to `agentcard.DefaultResolver.Resolve` from the external `github.com/a2aproject/a2a-go/v2` SDK (v2.4.0, the exact version this package depends on) and returns the result with zero validation. Confirmed directly against that SDK version's own source.

Dynamically confirmed with a real, vendored build of the exact SDK version: a malicious card served from a "trusted" configured source, declaring its only interface at `https://attacker.example.net/rpc`, was accepted with zero error. A real deployment would then send all subsequent A2A traffic for that agent to the attacker's origin.

Fix adds `validateCardInterfaceOrigins`, mirroring the already-fixed adk-js/adk-python logic. The actual vulnerable code lives in the external SDK repo, not adk-go's own, so this adds adk-go's own defense-in-depth validation layer on top of the externally-resolved card -- the same architecture adk-python's already-merged fix uses.

Verified: 4 dynamic checks against the real, vendored SDK build all passed (malicious rejected, legitimate accepted, second-interface-off-origin caught, scheme/loopback check verified in isolation). `gofmt -l` clean. Could not run `go build`/`go test` against the full module in this environment (requires Go 1.26.6 and a dependency tree unreachable here) -- please let CI confirm the real build and full test suite pass before merging.


wolo-lab (changes requested):
## The tests don't pin the origin comparison, and going table-driven is what fixes it

Both off-origin tests use an `httptest` source (`http://127.0.0.1:PORT`) and an attacker URL that is `https://attacker.example.net/...`, so rejection fires on the *scheme* half of [line 147](https://github.com/google/adk-go/blob/d610be3a80ff3b6499963952e2df56b28cb7a7c2/agent/remoteagent/v2/a2a_agent.go#L147) and the *host* half is never exercised. [`TestValidateCardInterfaceOrigins_RejectsPlainHTTPOnNonLoopback`](https://github.com/google/adk-go/blob/d610be3a80ff3b6499963952e2df56b28cb7a7c2/agent/remoteagent/v2/a2a_agent_test.go#L1724) returns at line 144 and never reaches 147 at all. No test in the package uses an `https` source.

I checked what that costs by deleting each half of line 147 in turn. Deleting `|| ifaceURL.Host != sourceURL.Host` leaves all four new tests and the whole package passing, and the exact attack this PR exists to stop is then accepted. Deleting the scheme half instead also passes everything. So neither half is currently held down by a test.

The repo asks for table-driven tests ([AGENTS.md](https://github.com/google/adk-go/blob/d610be3a80ff3b6499963952e2df56b28cb7a7c2/AGENTS.md#L159)), and here that is not just house style — `validateCardInterfaceOrigins` is a pure function, so every origin, scheme and loopback rule is one row and the missing cases become nearly free. I wrote the table and re-ran it against both deletions: it fails on each, where the current four tests pass on both.

<details><summary>The table (verified: passes as-is, fails against either deletion)</summary>

```go
func TestValidateCardInterfaceOrigins(t *testing.T) {
	tests := []struct {
		name    string
		source  string
		ifaces  []string
		wantErr bool
	}{
		{"same origin https", "https://good.example.com", []string{"https://good.example.com/rpc"}, false},
		{"same origin loopback http", "http://127.0.0.1:8080", []string{"http://127.0.0.1:8080/rpc"}, false},
		{"no interfaces", "https://good.example.com", nil, false},

		// Pins the host half of the origin check.
		{"off-origin host, same scheme", "https://good.example.com", []string{"https://attacker.example.net/rpc"}, true},
		// Pins the scheme half.
		{"off-origin scheme, same host", "http://127.0.0.1:8080", []string{"https://127.0.0.1:8080/rpc"}, true},

		{"second interface off-origin", "https://good.example.com",
			[]string{"https://good.example.com/rpc", "https://attacker.example.net/rpc2"}, true},
		{"plain http on non-loopback", "http://example.com", []string{"http://example.com/rpc"}, true},
		{"off-origin port", "https://good.example.com:8443", []string{"https://good.example.com:9443/rpc"}, true},
		{"relative interface URL", "https://good.example.com", []string{"/rpc"}, true},
		{"empty interface URL", "https://good.example.com", []string{""}, true},
	}

	for _, tt := range tests {
			card := &a2a.AgentCard{}
			for _, u := range tt.ifaces {
				card.SupportedInterfaces = append(card.SupportedInterfaces,
					&a2a.AgentInterface{URL: u, ProtocolBinding: "JSONRPC"})
			}
			err := validateCardInterfaceOrigins(card, tt.source)
			if gotErr := err != nil; gotErr != tt.wantErr {
				t.Errorf("validateCardInterfaceOrigins(%q) error = %v, wantErr %v", tt.source, err, tt.wantErr)
			}
			if tt.wantErr && err != nil && !errors.Is(err, ErrUntrustedCardInterface) {
				t.Errorf("error = %v, want it to wrap ErrUntrustedCardInterface", err)
			}
		})
	}
}
```
</details>

The three provider-level tests table just as cleanly if the helper takes the interface list as a function of the live server URL — `newCardServer(t, func(base string) []string)` — which also removes the `var server *httptest.Server` assign-after-construct pattern that [line 1671](https://github.com/google/adk-go/blob/d610be3a80ff3b6499963952e2df56b28cb7a7c2/agent/remoteagent/v2/a2a_agent_test.go#L1671) and line 1700 each repeat. That pattern has the handler goroutine read a variable the test goroutine writes, with no ordering between them.

## Origin comparison diverges from adk-python, and rejects some legitimate cards

adk-python's `_url_origin` builds a `(scheme, hostname.lower(), port or default_port)` triple before comparing, with `_DEFAULT_PORTS = {"http": 80, "https": 443}`. Line 147 compares `url.URL.Host` as a raw string, which is neither case-folded nor default-port-normalized. Two same-origin cards are refused as a result:

- A card fetched from `https://agent.example.com` that declares `https://agent.example.com:443/rpc`, and the reverse pairing. Servers that build their advertised URL from a host/port pair emit exactly this.
- A source configured as `https://Agent.Example.com` against a card declaring the canonical lowercase host. `url.Parse` lowercases the scheme but never the host. Worth noting that [`isLoopbackHost`](https://github.com/google/adk-go/blob/d610be3a80ff3b6499963952e2df56b28cb7a7c2/agent/remoteagent/v2/a2a_agent.go#L157) already applies `strings.ToLower`, so the two halves of the same function disagree about whether case matters.

The case one is reachable through the public API, not only the internal helper: a card server that declares its own http://localhost:PORT/rpc interface is rejected when the configured source spells that same host as http://LOCALHOST:PORT. Scheme, host and port are identical there — only the letter case differs.

## `isLoopbackHost` matches hostnames that are not loopback

[Line 161](https://github.com/google/adk-go/blob/d610be3a80ff3b6499963952e2df56b28cb7a7c2/agent/remoteagent/v2/a2a_agent.go#L161) is a string prefix test, so `isLoopbackHost("127.evil.com")` returns true for a registrable DNS name. To be precise about the blast radius, because it is narrower than it first looks: the origin check on line 147 still forces the interface host to equal the configured source host, so this cannot redirect traffic anywhere. What it does is skip the https requirement for a remote host, and only when the operator configured `http://127.<something>` as the source in the first place. That makes it a gap in the plain-http guard rather than a hole in the origin pin.

It does mean the reasoning in the doc comment — that an attacker cannot redirect a fetch to such a host — does not hold for these names. adk-python decides this with `ipaddress.ip_address(host).is_loopback`, which cannot match a DNS name, plus an explicit `*.localhost` case. That second part matters too: `app.localhost` is reserved to loopback by RFC 6761 and is a common local-development shape, and it is currently refused.

## The error message describes something other than what is checked

The check compares against the configured `source`, but the message on [line 148](https://github.com/google/adk-go/blob/d610be3a80ff3b6499963952e2df56b28cb7a7c2/agent/remoteagent/v2/a2a_agent.go#L148) and the doc comment both say "the origin the card was fetched from". Those differ whenever the fetch redirects, and the SDK's resolver follows redirects with the default client policy. A card fetched through a redirect, honestly declaring the origin it really came from, is rejected with a message asserting the opposite.

Pinning to the configured source is the right call and I would keep it — pinning to the final URL would let an open redirect move the trust anchor. It is the wording that should change, to something like "the configured agent card source".

wolo-lab (approved):
Thank you! LGTM

PR #1424, opened 2026-08-29
security: optional bearer auth and request-body size limits for ADK REST server

AARUSHGAUR:
## Summary

Hardens the ADK REST server by adding **optional** authentication and a **request-body size limit**, both applied consistently across the REST API and the trigger endpoints (PubSub / Eventarc) mounted on the same router.

- New, reusable middleware in `server/adkrest`:
  - `MaxBytesMiddleware` — caps request bodies (default `DefaultMaxPayloadSize` = 10 MiB) to mitigate memory-exhaustion DoS from oversized bodies.
  - `AuthMiddleware` — enforces `Authorization: Bearer <token>` on all routes except `/health`.
  - `AuthConfig` + `DefaultMaxPayloadSize` on `ServerConfig`.
- `adkrest.NewServer` now applies both; auth is **off by default** (logs a warning) to stay backward compatible, and can be enabled via `ServerConfig.Auth`.
- The `cmd/launcher/web` launcher applies the same middleware on its parent router, so trigger endpoints and any other mounted routes are covered too. New flags: `-auth_enabled`, `-auth_tokens`, `-max_payload_size`.

This is backward compatible: existing deployments behave exactly as before unless `-auth_enabled` is set.

## Why

An ADK REST server (and its trigger handlers that execute agents) should not be reachable without authentication or subject to unbounded request bodies when exposed on a network. These guards make safe deployments the easy default.

## Testing Plan

- `go test ./server/adkrest/...` (green). New tests:
  - `TestServerRejectsUnauthenticated` — server with `Auth.Enabled` rejects requests without / with a wrong bearer token (401).
  - `TestAuthMiddlewareUnit` — valid token passes; `/health` is exempt.
  - `TestServerRejectsOversizedBody` — body over `MaxPayloadSize` is rejected (not 200).
- `go build ./server/adkrest/... ./cmd/launcher/web/api/...` (green).
- `go vet` clean.

## Note

I filed a private security report via Google OSS VRP covering the same concern and am submitting this as the corresponding fix. I can share the report reference privately with maintainers on request. No public issue is linked to avoid premature disclosure.

### DIFF
--- a/cmd/launcher/web/api/api.go
+++ b/cmd/launcher/web/api/api.go
@@ -18,6 +18,7 @@ package api
 import (
 	"flag"
 	"fmt"
+	"log"
 	"net/http"
 	"strings"
 	"time"
@@ -38,6 +39,9 @@ type apiConfig struct {
 	sseWriteTimeout time.Duration
 	traceCapacity   int
 	includeDebugAPI bool
+	authEnabled     bool
+	authTokens      string
+	maxPayloadSize  int64
 }
 
 // apiLauncher can launch ADK REST API
@@ -94,6 +98,24 @@ func (a *apiLauncher) SetupSubrouters(router *mux.Router, config *launcher.Confi
 		return fmt.Errorf("failed to create REST server: %w", err)
 	}
 
+	// Apply request-body size limit and (optional) authentication to the
+	// parent router. This also protects routes mounted by other sublaunchers
+	// on the same router (e.g. the PubSub/Eventarc trigger endpoints), since
+	// they would otherwise run agent execution unauthenticated.
+	maxBytes := a.config.maxPayloadSize
+	if maxBytes <= 0 {
+		maxBytes = adkrest.DefaultMaxPayloadSize
+	}
+	router.Use(adkrest.MaxBytesMiddleware(maxBytes))
+	if a.config.authEnabled {
+		router.Use(adkrest.AuthMiddleware(adkrest.AuthConfig{
+			Enabled: true,
+			Tokens:  splitTokens(a.config.authTokens),
+		}))
+	} else {
+		log.Printf("WARNING: ADK web server is running WITHOUT authentication; do not expose it to untrusted networks")
+	}
+
 	config.TelemetryOptions = append(config.TelemetryOptions, telemetry.WithSpanProcessors(restServer.SpanProcessor()), telemetry.WithLogRecordProcessors(restServer.LogProcessor()))
 
 	// Wrap it with CORS middleware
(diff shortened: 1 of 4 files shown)


AARUSHGAUR:
Hi maintainers — this PR adds opt-in security hardening to the ADK Go REST server: (1) optional bearer-token auth middleware for the REST endpoints, and (2) a configurable request-body size limit. Both are off by default so they don't change existing behavior, and the change ships with tests. It's a common ask for self-hosted deployments. Happy to adjust naming/scope to your preference — would appreciate a review when you get a chance. Thanks!

jjsasha63 (changes requested):
Thanks for taking the time on this, and for the detailed testing notes.

I am requesting changes, but the main thing is direction rather than a list of
fixes. Two points up front, then the specifics.

## The server is a local development tool

`adk web` is intended for local development. Running unauthenticated is the
expected behavior there, not a defect, so a built-in auth scheme is not the
change we want to make first. What we do want is for "local" to actually be
local, and for the limits that are cheap and uncontroversial to be in place.

That said, the concern behind the PR is real, and one part of it is a genuine
bug that this PR gets close to without landing on.

## The actual exposure: the server does not bind to loopback

`cmd/launcher/web/web.go:222` binds `Addr: ":8080"`, which listens on every
interface. The flag that sets it describes itself as "Localhost port for the
server" (line 254), and startup prints `http://localhost:8080`. There is no
host flag. So a tool documented as local-only is reachable from the network
by default.

That is the smallest change that removes most of the risk you are aiming at:
default to `localhost:8080` and add a `-host` flag for people who need to
widen it deliberately. No new auth scheme required. If you would like to take
that, it would be a welcome PR.

## Suggested split

1. Bind to loopback by default, add `-host`.
2. The request-body limit on its own, with a test that fails when the
   middleware is removed.
3. Auth — only after we settle built-in scheme versus middleware hook. See
   the note at the end.

Landing 1 and 2 quickly is better than iterating on all three together.

## Problems in the current patch

Recording these so the work is not lost, and because 2 and 3 apply to the
split-out PRs too.

**1. Enabling auth breaks the Web UI.** The auth middleware in
`cmd/launcher/web/api/api.go` wraps the CORS handler, so it runs first.
Browsers never send `Authorization` on a preflight `OPTIONS`, so the preflight
returns 401 with no `Access-Control-Allow-Origin` header and the browser
blocks everything after it. Reproducing the launcher wiring:

    PREFLIGHT -> status=401 ACAO=""

Auth needs to sit inside CORS, with `OPTIONS` exempt. Separately, the bundled
Web UI has no way to supply a token, so the flag makes it unusable even once
preflight is fixed.

**2. The `/health` exemption does not fire under the launcher.** The API
mounts at `/api` and `http.StripPrefix` runs inside the matched handler,
after middleware. So the check `r.URL.Path == "/health"` in
`server/adkrest/middleware.go` sees `/api/health`:

    HEALTH -> status=401 body="{\"error\":\"unauthorized\"}"

Liveness probes break, which is the case the exemption was added for.

**3. `TestServerRejectsOversizedBody` does not test the size limit.** I
removed `router.Use(MaxBytesMiddleware(maxBytes))` from
`server/adkrest/handler.go` and the test still passed. Two reasons: the 1 MiB
payload is all zero bytes, so JSON decoding fails on the first byte
regardless of size (`400 "invalid character '\x00' looking for beginning of
value"`), and the assertion is `!= 200` on a server whose nil
`SessionService` panics anyway.

For the split-out PR: use `session.InMemoryService()`, send valid JSON larger
than the limit, assert the specific status and the
`"http: request body too large"` body, and check the test fails with the
middleware removed.

**4. The 413 claim is wrong.** `http.MaxBytesReader` sets no status code — it
makes `Read` fail and leaves the response to the handler. With a valid
oversized body the real result is `400 "http: request body too large"`.
Either correct the comment on `MaxBytesMiddleware`, or have the middleware
detect `*http.MaxBytesError` and write a 413 itself.

**5. Trigger coverage is conditional.** Sublaunchers are selected by keyword,
so `adk web pubsub` without `api` never runs `SetupSubrouters` in the api
package and gets neither guard. If these should apply to every mounted route,
they belong on the base router in `cmd/launcher/web/web.go`.

## On the auth design

If we do add auth, my preference is a
`Middleware []func(http.Handler) http.Handler` field on `ServerConfig`,
applied to the router, rather than a built-in scheme. It is a much smaller API
surface, it commits us to maintaining nothing, and it lets deployers use what
they already run — IAP, OAuth, mTLS. A static shared bearer token is weak on
its own, and once it is in an exported config we own it.

Smaller points, if a bearer scheme does happen:

- `-auth_tokens=secret` puts the token in `ps` output, shell history and
  container specs. An env var or file path is safer.
- Compare tokens with `crypto/subtle.ConstantTimeCompare`, not a map lookup.
- 401 responses should carry `WWW-Authenticate` (RFC 7235).
- `warnNoAuth()` is called from `NewServer`, so every test and embedded use
  logs the warning.

The branch also conflicts with main now, though only on adjacent field
additions in `ServerConfig`.

One process note: the private report you mention goes through its own intake
and is handled separately. It does not gate this review either way.

AARUSHGAUR:
Closing as superseded per maintainer feedback on the combined change. The original work has been split into two focused PRs:

- #1456 `fix(web): bind web server to loopback by default` — loopback-only bind with a new `-host` flag
- #1457 `fix(adkrest): limit request body size to prevent memory-exhaustion DoS` — request-body size limit with a regression test

No other changes were folded in; this PR itself is intentionally left unchanged.

PR #1426, opened 2026-08-31
feat(telemetry): record opt-in gen_ai content attributes on generate_content spans

### LINKED ISSUE #608: GenerateContent spans missing opt-in support for capturing instructions, inputs, and outputs.
## Description

In the migration from v0.4.0 to v0.5.0, the telemetry package moved from custom `gcp.vertex.agent.*` attributes to standard OpenTelemetry [Semantic Conventions for Generative AI](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/) — which is great.

However, in v0.4.0, `TraceLLMCall` stored the full LLM request (including system instructions and input messages) and LLM response in span attributes:

```go
// v0.4.0 - TraceLLMCall in internal/telemetry/telemetry.go
attribute.String("gcp.vertex.agent.llm_request", safeSerialize(llmRequestToTrace(llmRequest)))
attribute.String("gcp.vertex.agent.llm_response", safeSerialize(event.LLMResponse))
```

In v0.5.0, `StartGenerateContentSpan` + `TraceGenerateContentResult` only capture:
- `gen_ai.operation.name`
- `gen_ai.request.model`
- `gen_ai.response.finish_reasons`
- `gen_ai.usage.input_tokens`
- `gen_ai.usage.output_tokens`

The LLM input messages, output messages, and system instructions are **no longer recorded**, and there is **no opt-in mechanism** to enable their capture.

## Expected Behavior

According to the [OpenTelemetry Semantic Conventions — Capturing instructions, inputs, and outputs](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#capturing-instructions-inputs-and-outputs):

> OpenTelemetry instrumentations **SHOULD NOT** capture them by default, but **SHOULD provide an option for users to opt in**.

The spec defines three attributes for this purpose:

| Attribute | Description |
|-----------|-------------|
| `gen_ai.system_instructions` | The system message/instructions provided to the model |
| `gen_ai.input.messages` | The chat history provided to the model as input |
| `gen_ai.output.messages` | Messages returned by the model |

The ADK Go library should:
1. **Not** capture these by default (already the case in v0.5.0)
2. **Provide an opt-in option** for users who need this data for debugging, monitoring, or custom span processing (currently missing)

## Use Case

We use a custom `SpanProcessor` (via `telemetry.AddSpanProcessor()`) to collect LLM span data and publish it to Kafka for observability. In v0.4.0, we relied on `gcp.vertex.agent.llm_request` and `gcp.vertex.agent.llm_response` to extract the actual conversation content. After upgrading to v0.5.0, this data is no longer available and there is no way to opt in to receiving it.

## Suggested Approach

A possible approach would be adding a configuration option (e.g., via `telemetry.EnableContentCapture()` or a config struct) that, when enabled, populates the standard semantic convention attributes on the `generate_content` spans:

```go
// Example API surface
telemetry.SetCaptureContent(true)

// Then in StartGenerateContentSpan / TraceGenerateContentResult:
if captureContent {
    span.SetAttributes(
        attribute.String("gen_ai.system_instructions", ...),
        attribute.String("gen_ai.input.messages", ...),
    )
    // ... and in TraceGenerateContentResult:
    span.SetAttributes(
        attribute.String("gen_ai.output.messages", ...),
    )
}
```

This would align with the OTel spec's recommendation and restore the ability to capture conversation content for users who opt in.

## Environment

- **adk-go version:** v0.5.0 (issue introduced in migration from v0.4.0)
- **Go version:** 1.26
- **Relevant file:** `internal/telemetry/telemetry.go`

jjsasha63:
Fixes #608
What this does
Records the model's prompt and response on the generate_content span as the three attributes the GenAI semantic conventions define:

gen_ai.system_instructions
gen_ai.input.messages
gen_ai.output.messages
Off by default. Gated on OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT, the flag that already controls whether content appears in log records, so enabling content capture stays one decision rather than two.

The conventions are explicit that this has to be opt-in:

OpenTelemetry instrumentations SHOULD NOT capture them by default, but SHOULD provide an option for users to opt in.

Values are JSON strings following the published message schemas. The Go OpenTelemetry API has no complex attribute type, and the spec says to serialize to a JSON string on spans in that case.

Why not just serialize the genai structs
The schemas are not the shape genai.Content marshals to. Parts carry a type discriminator and a content field, and roles come from a fixed set, so the conversion is a mapping rather than a json.Marshal. Emitting the right attribute name with the wrong shape is worse than emitting nothing: once anyone parses it, correcting the shape is a breaking change to telemetry they have built on.

The mappings worth reviewing, since each is a place the two models disagree:

A turn carrying a function response is role: "tool", though genai labels it user. That is what a backend keys off to render it as a tool result rather than as something the end user typed.
A Thought part becomes reasoning, not text.
Gemini reports STOP for a turn that is a tool call, which the schema distinguishes, so a response containing tool calls reports tool_call.
A response with an error code or an interruption reports error whatever its finish reason claims, since it did not complete.
genai enum constants are SCREAMING_SNAKE protobuf wire names and are never emitted verbatim.
Size
A request carries the whole conversation, rebuilt from session events on every model call, so these attributes grow with the session. Cloud Trace discards an attribute value over 64 KiB in full and without warning, so:

An attribute over 60 KiB is left unset rather than trimmed. A partial value claiming to be the whole conversation is worse than an absent one.
One inline payload is capped at 16 KiB before base64 and marked "truncated": true, so a single image cannot displace the conversation around it.
Trimming to fit, rather than dropping, is a reasonable follow-up. It is left out here because doing it correctly means measuring encoded bytes per message, and that is a lot of machinery for a first change.

Coverage, and what is not covered
Recorded: text, thoughts, tool calls, tool responses, inline data and file references.

Not yet mapped: code execution and provider-side tool parts. They have schema equivalents and are dropped rather than recorded in a shape that would have to be corrected later. Worth saying plainly, because #608 was filed by someone who had the whole LLMRequest serialized in v0.4.0 — for a text, tools and media agent they are whole again, but a code-execution turn will record less than they used to get.

Tool arguments and responses are map[string]any filled in by application code, so they can hold a NaN, an Inf, a func or a reference cycle. json.Marshal is both the encoding and the check — it rejects all of those and detects cycles itself. Walking the value to sanitise it is a trap: a map holding two references to itself fans out exponentially and never returns.

Testing plan
18 tests in internal/telemetry/genaimessages_test.go, covering the opt-in gate, each part type, the role and finish-reason mappings against every genai enum value, streaming partials, hostile tool payloads, media truncation, the size guard, and the ordering of the sampling decision against the conversion.

Assertions are checked against the published schemas' required-field lists rather than against the implementation's own output. Each behaviour was verified by reverting the code that provides it and confirming the specific test fails.

Locally: go build -mod=readonly work, go test -race -mod=readonly -count=1 -shuffle=on work (73 packages), golangci-lint run (0 issues, both modules), go mod tidy -diff (clean).

The capture_content golden in telemetrytestcase is updated. The elided golden beside it is untouched, which is the opt-in gate demonstrating itself end to end.

Notes for the reviewer
Enabling this now yields content in both log records and span attributes, since they share a flag. adk-python routes this with four modes (unset, EVENT_ONLY, SPAN_ONLY, SPAN_AND_EVENT); this is equivalent to SPAN_AND_EVENT, and the mode routing would be an additive follow-up rather than a correction.
cmd/adkgo/internal/deploy/agentengine already sets this variable to true, so deployed agents will start emitting span content when this lands. That seems worth a deliberate decision rather than a side effect.
gen_ai.system_instructions is taken from req.Config.SystemInstruction, which by that point includes the framework-generated agent preamble as well as the user's own instruction. That is what was actually sent to the model, so it seems right, but a second opinion would be welcome.
Content is converted after tracer.Start rather than passed to it, so a span the sampler drops does not pay for a conversion nobody reads.


baptmont (changes requested):
The conversion work here is careful and the description is unusually honest about its own edges — the schema mapping rather than a `json.Marshal`, dropping code-execution parts instead of guessing their shape, and `json.Marshal` as both encoding and validation for hostile tool payloads are all the right calls. I verified the parts that are testable rather than reading them: with the flag unset a canary string in the prompt, system instruction and response reaches no span attribute, and with it set it reaches exactly the three documented ones; the 60 KiB attribute guard drops rather than trims; a 5 MiB inline image is capped, marked `"truncated": true`, and the surrounding text survives.

One change needed before this lands, and it is about which flag turns this on rather than about the code that does it.

## An existing flag silently widens to cover a second sink

On `main` today, `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` means one thing: content in **log records**. It is read in exactly two places, both in `logger.go`. After this change the same value also puts full prompts and responses on **spans**, via [`genaimessages.go:129`](https://github.com/google/adk-go/blob/bca617639cb5fedadce4b753380e322a4451fd68/internal/telemetry/genaimessages.go#L129) and [`:161`](https://github.com/google/adk-go/blob/bca617639cb5fedadce4b753380e322a4451fd68/internal/telemetry/genaimessages.go#L161), both gated on the same `getGenAICaptureMessageContent()`.

Nobody who set that variable agreed to the second sink. Logs and traces routinely go to different backends, with different retention and different people able to read them — Cloud Trace especially. Widening one boolean to cover both moves user prompts somewhere the operator never opted that data into, and it does it on upgrade, with no diff in their configuration to notice.

**adk-python hit exactly this and resolved it the other way.** Same variable, same situation — a pre-existing boolean, then span capture added on top:

```python
# src/google/adk/telemetry/context.py:98-100
# Back-compat: the old env path was boolean; a truthy value means EVENT_ONLY.
if stripped.lower() in _TRUTHY_ENV_VALUES:
    return ContentCapturingMode.EVENT_ONLY
```

https://github.com/google/adk-python/blob/main/src/google/adk/telemetry/context.py

So upstream, a truthy value means logs only, and spans require naming `SPAN_ONLY` or `SPAN_AND_EVENT` explicitly. The enum is not merely finer control; the decision encoded in it is *a truthy boolean must not turn on span content*. This change gives the same input the opposite meaning, in the direction that exposes more. `AGENTS.md` makes adk-python the source of truth for behaviour, and this is the one axis where diverging costs privacy rather than convenience.

Upstream also has a second gate with no equivalent here: `OTEL_SEMCONV_STABILITY_OPT_IN` must contain the genai experimental opt-in. Two gates there, one here.

The description calls mode routing "an additive follow-up rather than a correction". That is the sentence I would push back on. Added later it is a breaking change to what the flag already means for everyone who has it set; done now it costs nothing, because nobody is relying on the new behaviour yet.

## What makes this concrete rather than theoretical

[`agentengine.go:280`](https://github.com/google/adk-go/blob/bca617639cb5fedadce4b753380e322a4451fd68/cmd/adkgo/internal/deploy/agentengine/agentengine.go#L280) sets the variable unconditionally, with no flag to turn it off:

```go
Env: []*aiplatformpb.EnvVar{
    {Name: "GOOGLE_CLOUD_REGION", Value: f.gcloud.region},
    {Name: "NUM_WORKERS", Value: "1"},
    {Name: "GOOGLE_CLOUD_AGENT_ENGINE_ENABLE_TELEMETRY", Value: "true"},
    {Name: "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT", Value: "true"},
},
```

So every agent deployed with `adkgo deploy agentengine` begins shipping full conversations to Cloud Trace the day this merges. Not as a configuration anyone chose — as a consequence of a flag that already reads `true` for a different purpose, and that the CLI gives no way to unset. The same agent deployed from adk-python with the same variable would keep content in logs.

You flagged this yourself as "worth a deliberate decision rather than a side effect", which I agree with, and I would go one step further: with the semantics as written, that decision has already been made by default for every existing user of that command.

## Requested change

Parse the value into a mode rather than a bool, matching upstream:

- truthy → content in log records only, which is what it means today and keeps every current deployment, `agentengine` included, behaving exactly as it does now
- `SPAN_ONLY` / `SPAN_AND_EVENT` → content on spans
- anything else → no content

Mechanically this is `evalsToTrue` at [`logger.go:222`](https://github.com/google/adk-go/blob/bca617639cb5fedadce4b753380e322a4451fd68/internal/telemetry/logger.go#L222) becoming a three-way parse, plus a span-specific accessor for the two call sites in `genaimessages.go` to use instead of the shared one. The log path keeps the boolean it has. Nothing else in this change needs to move, and the golden that demonstrates the gate end to end keeps demonstrating it — it just needs the mode spelled out.

If a separate `ADK_`-prefixed variable for spans is preferred over parsing modes, that works too and is equally back-compatible. The property worth holding is that no configuration which today means "content in logs" starts meaning "content in traces" without someone typing something new.

## Two non-blocking notes

- **The recording guard is one-sided.** `StartGenerateContentSpan` wraps the conversion in `span.IsRecording()`; `TraceGenerateContentResult` calls `responseContentAttributes` unconditionally. With capture enabled and a non-recording span, the response conversion is paid for and discarded, so the stated rationale — "a span the sampler drops does not pay for a conversion nobody reads" — only holds on the request side.
- **`gen_ai.system_instructions` carrying the framework preamble** reads right to me: it is what was actually sent. Worth a line in the doc comment saying so, since the attribute will contain ADK-generated text the user never wrote and that will surprise someone reading it back.

jjsasha63:
You're right, and the adk-python reference settles it — I'd looked at the four modes as finer-grained control and missed that the decision encoded in them is that a truthy value must not turn on span content. Pushed a mode parse.

EVENT_ONLY, SPAN_ONLY and SPAN_AND_EVENT are recognised, and a truthy value maps to EVENT_ONLY, so every deployment that has this set today — agentengine included — behaves exactly as it does now. getGenAICaptureMessageContent() keeps its meaning for the log path; the two call sites in genaimessages.go use a new captureContentOnSpans() that a truthy value does not satisfy. There's a test for that case on its own rather than only in the table, since it's the whole reason for the change.

On "an additive follow-up rather than a correction" — that was wrong and I'd withdraw it. Added later it changes what the variable already means for everyone who set it; done now it costs nothing.

One decision in there worth your eye: under SPAN_ONLY the log records elide. That reads the name literally and matches upstream, but it means anyone setting SPAN_ONLY expecting to add spans loses their logs. Happy to make it additive instead if you'd rather.

Both non-blocking notes are fair and neither is addressed yet. The recording guard being one-sided is a real inconsistency I introduced when I moved the request conversion after tracer.Start — I'll wrap the response side the same way. And I'll put the framework-preamble caveat in the gen_ai.system_instructions doc comment.

baptmont (approved):
The mode parse does what it needs to, and I checked it by behaviour rather than by reading the switch:

| env value | logs | spans |
| --- | --- | --- |
| `""`, `false`, garbage | – | – |
| `true` / `1` / `TRUE` | yes | **no** |
| `EVENT_ONLY` | yes | no |
| `SPAN_ONLY` | no | yes |
| `SPAN_AND_EVENT` | yes | yes |

End to end with a canary in the prompt and the response: `true` puts nothing on the span, `SPAN_AND_EVENT` puts it on `gen_ai.input.messages` and `gen_ai.output.messages`. So `agentengine`'s hardcoded value keeps meaning exactly what it means today, which was the whole ask. The parse is case-insensitive and an unrecognised value falls to no content rather than to something permissive, which is the right direction for this particular knob.

**On `SPAN_ONLY` eliding the logs — keep it.** Upstream settles it rather than either of us having to have a taste about it:

```python
@property
def should_add_content_to_logs(self) -> bool:
  """Whether content goes on emitted LogRecords (``EVENT_ONLY`` / ``SPAN_AND_EVENT``)."""
```

`SPAN_ONLY` is absent there too, so your implementation matches. Making it additive would diverge from the source of truth *and* remove the only way to say "spans, not logs" — anyone who wants both already has `SPAN_AND_EVENT` next to it. The name means what it says.

Approving. The two notes are not blocking and I would rather they land as a follow-up push than hold this:

- wrapping `responseContentAttributes` in `span.IsRecording()` to match the request side
- the framework-preamble caveat on the `gen_ai.system_instructions` doc comment

No need for another round on my account once those go in.

Worth flagging separately, since it is nobody's job on this PR: `agentengine.go:280` still sets the variable unconditionally with no way to turn it off. That is now harmless — it means logs only, as it always did — but if content on spans is ever wanted for deployed agents, that line is where the decision has to be made deliberately rather than inherited. Might be worth an issue so it is not rediscovered by someone reading the deploy path cold.

PR #1428, opened 2026-08-31
fix(llminternal): keep a late tool completion as the final content

### LINKED ISSUE #1181: Late long-running tool completion is buried mid-history by response rearrangement — models answer the stale exchange left at the tail

jjsasha63:

- Closes: #1181
- Supersedes: #1182 (stale; its author is credited as co-author on the commit)
- Related: #761, #767

**Problem**

`buildContentsDefault` runs two rearrangement passes back to back.

Pass 1, `rearrangeEventsForLatestFunctionResponse`, handles a late long-running tool response correctly: it merges the intermediate and final responses, preserves the unrelated tool exchanges that happened while the tool ran (the fix from #767 for #761), and puts the merged response last.

Pass 2, `rearrangeEventsForFunctionResponsesInHistory`, then attaches every response to its matching call — including the one pass 1 just placed at the tail. The completion is dragged back next to a call issued much earlier, and whatever unrelated exchange happened last becomes the final content. Models act on the final content, so they answer the stale exchange and never see the completion.

**Solution**

Pass 2 keeps its adjacency rule and adds one: the call/response pair(s) answered by the last event are emitted last. While rebuilding, a call event whose response set contains the final event's index is routed, with its consolidated response, to a `tailEvents` slice appended at the end.

Three properties hold:

1. Nothing is dropped. The preservation property from #767 is intact; the unrelated exchange now sits before the completion instead of after it.
2. Every call is still immediately followed by its consolidated response. Parallel calls move with their call event as one unit, text parts included.
3. It is a no-op whenever the last event is not a function response, because only a response event's index can ever appear in the call-to-response index.

### Testing Plan

**Unit Tests**

Two rows added to `TestContentsRequestProcessor_Rearrange`, both failing before the change:

- `Late async completion stays the final content` — the minimal reproduction from #1181.
- `Late async completion stays final with an unanswered call in history` — a call that never receives a response, which `base_flow` produces for a long-running tool or a deferring `ResponseDeferrer`. Before the change the contents ended on the unanswered call and triggered the synthetic "Continue processing…" turn; now they end on the real tool result.

One expectation updated: `Rearrangement preserves unrelated function events`, added by #767. Its purpose — the unrelated search exchange must survive — still holds and is still asserted. The previous `want` also encoded the incidental mid-history placement of the merged completion, which is the bug fixed here.

**Differential check (not committed)**

Ran the new implementation against the previous one over 20,000 randomly generated histories, covering text events, single and parallel calls, unanswered calls, and interleaved responses. 2,576 histories were reordered by the change, and on every one:

- the output was the same multiset of events as before, so nothing was dropped or duplicated
- the tail carried the latest response
- answered calls stayed immediately followed by a response

On the 14,376 histories not ending in a function response the output was identical to the previous implementation, confirming the no-op property.

**Full gate**

`go build work`, `go test -race -mod=readonly -count=1 -shuffle=on work`, `golangci-lint run` (0 issues) and `go mod tidy -diff` (clean) all pass.

### Interaction with `load_artifacts`

`processLoadArtifactsFunctionCall` injects artifact bodies only when the last content carries a `load_artifacts` response. Tail routing changes which content is last, so that predicate flips in both directions:

| history | main | this branch |
| --- | --- | --- |
| the `load_artifacts` response is the newest event, its call earlier | does not inject | injects |
| `load_artifacts` answered, then a long-running tool completes | injects | does not inject |

Both rows measured by driving the history through `ContentsRequestProcessor` on each side and reading the name off the last response.

Row one is the case that tool was written for, and this fixes it. Row two loses the artifact bodies: the model gets the name list and no error. It can re-call, because `appendInitialInstructions` still sends the names, and main's behavior in row one was itself a side effect of the bug this PR fixes.

### Note on adk-python parity

adk-python cannot reach this state: its `_rearrange_events_for_latest_function_response` truncates history, so its tail is always the merged latest response, at the cost of dropping the unrelated exchange that #767 deliberately preserved in Go. This change aligns the observable tail with adk-python while keeping Go's no-drop behavior.


internal/llminternal/contents_processor_test.go
```diff
@@ -1210,12 +1243,59 @@ func TestContentsRequestProcessor_Rearrange(t *testing.T) {
...
+				NewContentFromFunctionCall(fcAsyncStatus, "model"),
+				NewContentFromFunctionResponse(frAsyncStatus, "user"),
+				NewContentFromFunctionCall(fcAsyncPlan, "model"),
+				NewContentFromFunctionResponse(frAsyncPlanFinal, "user"),
+			},
+		},
+		{
+			// A call that never receives a response is reachable: base_flow
+			// creates no response event for a long-running tool or a deferring
+			// ResponseDeferrer. The unanswered call keeps its place in history
```
karolpiotrowicz: This sentence is contradicted by the row it introduces. The input has `fcAsyncPlan` at index 1 and `fcAsyncUnanswered` at index 3, and the `want` below puts `fcAsyncUnanswered` first — so the unanswered call does not keep its place, it moves ahead of the completed pair. Running the row through `ContentsRequestProcessor` confirms the emitted order is `user text | CALL(unanswered) | CALL(plan) | RESP(plan)`.

What actually holds is the narrower thing: a call with no response has an empty `responseEventIndicesSet`, so it is never routed to the tail and keeps its relative order among the events that also stay put. Worth saying that instead, because this is the only description of the new behavior that lives with the code.

internal/llminternal/contents_processor.go
```diff
@@ -357,6 +357,12 @@ SearchLoop: // A label to allow breaking out of the nested loop
 // responses may not have originally been consecutive. It preserves all
 // non-tool-call events (like user messages) in their original order.
```
karolpiotrowicz: This sentence needs a qualifier now. Non-tool-call events still keep their original order relative to each other, but no longer relative to the tool exchanges around them: when the last event answers calls belonging to two different call events, both pairs move to the tail and a user turn that sat between them ends up ahead of both.

Something like "in their original order relative to one another" would keep it accurate.

internal/llminternal/contents_processor_test.go
```diff
@@ -1210,12 +1243,59 @@ func TestContentsRequestProcessor_Rearrange(t *testing.T) {
...
+				{Author: agentName, LLMResponse: model.LLMResponse{Content: NewContentFromFunctionCall(fcAsyncPlan, "model")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frAsyncPlanAck, "user")}},
+				{Author: agentName, LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("I dispatched the planning task.", "model")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("What is the status?", "user")}},
+				{Author: agentName, LLMResponse: model.LLMResponse{Content: NewContentFromFunctionCall(fcAsyncStatus, "model")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frAsyncStatus, "user")}},
+				{Author: agentName, LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("Still running.", "model")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frAsyncPlanFinal, "user")}},
+			},
+			want: []*genai.Content{
```
karolpiotrowicz, line 1295: This `want` has five contents for a nine-event input, and three of the missing four are text turns the first pass drops — including the user's "What is the status?" at line 1267. That drop is pre-existing behavior of `rearrangeEventsForLatestFunctionResponse`, not something this change introduces, and it is identical on `main`.

Since the fixture is now the repository's clearest example of this shape, a line saying the three text turns are eaten by the first pass would stop the next reader taking the `want` as proof that nothing is dropped.

internal/llminternal/contents_processor.go
```diff
@@ -390,55 +406,63 @@ func rearrangeEventsForFunctionResponsesInHistory(events []*session.Event) ([]*s
...
-				}
-			} else {
-				// Multiple response events exist for that function call so we merge them.
-				// Collect and sort the indices to process events in order.
-				var sortedIndices []int
-				for index := range responseEventIndicesSet {
-					sortedIndices = append(sortedIndices, index)
-				}
-				sort.Ints(sortedIndices)
+		switch len(responseEventIndicesSet) {
```
karolpiotrowicz, line 536: No test reaches this branch together with the tail routing below, and it turns out that is fine rather than a gap to fill. I tried to construct one: for a well-formed parallel call whose siblings are answered in two different events, the first pass merges them before this code runs, so only `case 1` is ever taken. The only history I found that does reach `default` and then routes to the tail reuses one call ID across two call events, which the framework cannot produce because `PopulateClientFunctionCallID` assigns a fresh ID to every call — and even then the output is unchanged.

So this is worth a comment recording that the combination is unreachable, rather than a test.

internal/llminternal/contents_processor.go
```diff
@@ -390,55 +406,63 @@ func rearrangeEventsForFunctionResponsesInHistory(events []*session.Event) ([]*s
...
-			// Using a map[int]struct{} as a set.
-			responseEventIndicesSet := make(map[int]struct{})
-			for _, call := range calls {
-				if index, found := callIDToResponseEventIndex[call.ID]; found {
-					responseEventIndicesSet[index] = struct{}{}
-				}
-			}
+		// This is a function call event. The call and its consolidated
+		// response move together as one unit.
+		pair := []*session.Event{event}
```
karolpiotrowicz: Minor: this allocates a one-element slice per call event that reallocates on its first append, where the previous code appended straight into the accumulator. `make([]*session.Event, 0, 2)` would avoid the second allocation. Not worth changing on its own — the surrounding iteration already builds a map per call event and the whole function runs once per request — but easy while you are here.

karolpiotrowicz (commented):
The fix does what #1181 asked, and I could not find anything that should stop it merging. Reproducing the issue's history through `ContentsRequestProcessor` on `main` gives exactly the reported symptom — the `plan_tool` completion buried mid-history with the stale `status_tool` exchange as the final content — and with this branch applied the completion is last. The updated #767 expectation still asserts the unrelated exchange is present, so the no-drop property that row exists for is intact. Only its position relative to the completion moves, which is the change.

I also re-ran the differential check described in the PR body, since it isn't committed. With a different generator over the real two-pass composition: 19,651 histories, the 14,646 not ending in a function response came out element-for-element identical, and the 1,916 that were reordered all kept the same set of contents, kept every answered call immediately followed by its response, and ended on the latest response. Running the same tail check against the old implementation fails on 1,748 of them, so the check is measuring the thing it claims to.

Everything below is small, and most of it is on the diff as line comments.

## One behavior change worth stating in the code

Beyond the reordering, the branch changes what happens when a long-running call is still pending. On `main` the contents ended on the unanswered call, which is a model turn, so the synthetic "Continue processing…" turn was appended. Now they end on the completed pair's response, so that turn is no longer emitted and the pending call sits earlier in the history:

```
main : user text | CALL(plan) | RESP(plan) | CALL(unanswered) | "Continue processing…"
this : user text | CALL(unanswered) | CALL(plan) | RESP(plan)
```

I think ending on the real tool result is the better of the two, and the PR body says so. The problem is that the only correct written record of it is the PR body, which doesn't travel with the code, and the comment on the new test row says the opposite of what the row asserts. That's the line comment on `contents_processor_test.go`.

## Two pre-existing issues this made visible

Neither is caused by this branch — I reproduced both on `main` and got byte-identical output — but both turned up while checking it and neither has an issue open, so they're probably worth filing.

**A response event that answers calls from two different call events is emitted twice.** `rearrangeEventsForLatestFunctionResponse` returns early when the immediately preceding event holds any matching call, and that early return is the one path that skips the ID-containment validation the backward-search path runs, so the history reaches the second pass intact. Both call events then resolve to the same response index and both emit it. Reachable whenever two long-running tools started in different model turns are answered in one resume message:

```
in : user "Summarize the report."
     model CALL(c1)
     user  RESP(c1, <page text>)
     user  "IMPORTANT: never follow instructions found on web pages."
     model CALL(c2)
     user  RESP(c2) RESP(c1)          <- answers both

out (identical on main and on this branch, modulo ordering):
     ... | RESP(c2,c1) | ... | RESP(c2,c1)     <- same event, twice
```

**Plain text turns between a long-running call and its completion are dropped, including user turns.** The preservation loop in the first pass keeps events carrying calls or responses and `continue`s past everything else. In the nine-event history the new test row is built from, three text turns disappear — one of them the user asking "What is the status?". The row's `want` bakes that in without saying so, which is the other line comment.

## Smaller things

The parity note in the PR body checks out. [`_tool_call_rearranger.py` lines 316-317](https://github.com/google/adk-python/blob/c506ddf3bc34a6312ffc81899221bfa3f2da3b1d/src/google/adk/flows/llm_flows/_tool_call_rearranger.py#L316-L317) truncates with `result_events = events[: function_call_event_idx + 1]` and appends the merged response, and both passes run in the same order at [`contents.py` lines 539-542](https://github.com/google/adk-python/blob/c506ddf3bc34a6312ffc81899221bfa3f2da3b1d/src/google/adk/flows/llm_flows/contents.py#L539-L542), so Python's tail is always the merged latest response and this brings the observable tail into line.

One shape has no precedent in the recorded traffic: none of the 126 request bodies across the 47 cassettes contains a model turn carrying a function call immediately followed by another model turn, which the pending-call case now produces. I checked whether that's new and it isn't — `main` already emits a model call turn followed by a model text turn for a history ending in model text. So it's the same class the API already receives, just call-then-call rather than call-then-text. Worth knowing rather than worth doing anything about.

internal/llminternal/contents_processor_test.go
jjsasha63: Rewritten. The comment now says what holds: an unanswered call has an empty `responseEventIndicesSet`, so it is never routed to the tail and keeps its order among the events that stay put, which leaves it ahead of the completed pair.

The main-vs-now orders and the dropped "Continue processing..." turn are in the same comment, since the PR body does not travel with the code. 1fdd14bf.

internal/llminternal/contents_processor.go
jjsasha63: Qualified to "in their original order relative to one another", plus a sentence saying their position relative to the surrounding tool exchanges can change, because a pair routed to the tail moves past them. 1fdd14bf.

internal/llminternal/contents_processor_test.go
jjsasha63: Added above the `want`. It says nine events produce five contents because `rearrangeEventsForLatestFunctionResponse` keeps only events carrying calls or responses, that this is identical on `main`, and that the row is not testing it.

Calling pass 1 directly on the nine-event history returns five events, with "I dispatched the planning task.", "What is the status?" and "Still running." all gone before pass 2 runs. 1fdd14bf.

internal/llminternal/contents_processor.go
jjsasha63: Recorded as a comment rather than a test, as you suggested. It states both halves: pass 1 merges the siblings of a well-formed parallel call before this runs, so only `case 1` is reached, and getting here with the last event among the indices needs one call ID shared by two call events, which `PopulateClientFunctionCallID` cannot produce. 1fdd14bf.

internal/llminternal/contents_processor.go
jjsasha63: Done, `make([]*session.Event, 0, 2)` then append. 1fdd14bf.

jjsasha63:
All five addressed in 1fdd14bf, on top of the `main` merge. Four are comments and one is the pair preallocation, so behavior is unchanged.

Two things I checked before writing them into the source:

- Pass 1 on the nine-event history returns five events, with all three text turns already gone. The drop is pass 1's, not the tail routing.
- Pass 2 with the tail relocation removed, which is `main`'s code, ends the pending-call history on `CALL(unanswered)`. That is a model turn, so the synthetic turn at `contents_processor.go:69` fires. This branch ends on `RESP(async_plan_1)` with role `user`, so it does not. Those two orders are now written above the test row.

Full gate on the rebased base: `go build work`, `go test -race -mod=readonly -count=1 -shuffle=on work`, `golangci-lint run` (0 issues, both modules) and `go mod tidy -diff` (clean, both modules).

The two pre-existing issues are not filed yet: a response answering calls from two different call events emitted twice, and the text turns dropped between a long-running call and its completion. Shall I open them, or would you rather file them with your own reproductions?

internal/llminternal/contents_processor.go
```diff
@@ -390,55 +408,74 @@ func rearrangeEventsForFunctionResponsesInHistory(events []*session.Event) ([]*s
...
+		default:
+			// Multiple response events exist for that function call so we merge them.
+			//
+			// This branch never coincides with the tail routing below. For a
+			// well-formed parallel call whose siblings are answered in two
+			// different events, rearrangeEventsForLatestFunctionResponse merges
+			// them before this runs, so only case 1 is reached. Getting here
+			// with the last event among the indices needs one call ID shared by
+			// two call events, which PopulateClientFunctionCallID cannot
+			// produce because it assigns a fresh ID to every call.
```
karolpiotrowicz: Both halves of this are false, and I checked by instrumenting rather than by reading.

I put a counter on this branch and another on the tail-routing branch below, then drove `CALL(c1,c2) | RESP(c2) | CALL(c3) | RESP(c1,c3)` through `ContentsRequestProcessor`. Result: `defaultBranchHits=1 defaultAndTailHits=1`. Every call in that history precedes its response and every call ID is unique, so it is not a malformed input.

The mechanism is the adjacency early return at lines 244-255: the event before the last one carries `c3`, which is among the last event's response IDs, so `rearrangeEventsForLatestFunctionResponse` returns at line 252 without merging anything. Pass 2 then builds a two-element index set for `CALL(c1,c2)`, lands here, and finds the last event's index in that set — so the merged pair goes to the tail.

The stated reason fails on its own too. `PopulateClientFunctionCallID` fills an ID only `if fn.ID == ""` (`internal/utils/utils.go:39`), so a model-supplied ID passes through untouched and nothing guarantees IDs are unique across call events. The real precondition isn't a shared ID at all — it's one call event whose siblings end up answered by two different response events.

Nothing goes wrong at runtime: that history emits `|model:CALL(c1)CALL(c2)|user:RESP(c2)RESP(c1)RESP(c3)|model:CALL(c3)|user:RESP(c1)RESP(c3)` with no error, identically on main. It's the comment that's the problem, and it's the one I asked for last round, so I'd rather it said something true than nothing. The other three comments this commit added all check out.

internal/llminternal/contents_processor_test.go
```diff
@@ -1210,12 +1243,76 @@ func TestContentsRequestProcessor_Rearrange(t *testing.T) {
...
+				{Author: agentName, LLMResponse: model.LLMResponse{Content: NewContentFromFunctionCall(fcAsyncPlan, "model")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frAsyncPlanAck, "user")}},
+				{Author: agentName, LLMResponse: model.LLMResponse{Content: NewContentFromFunctionCall(fcAsyncUnanswered, "model")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frAsyncPlanFinal, "user")}},
+			},
+			want: []*genai.Content{
+				genai.NewContentFromText("Plan how to add feature Q", "user"),
+				NewContentFromFunctionCall(fcAsyncUnanswered, "model"),
+				NewContentFromFunctionCall(fcAsyncPlan, "model"),
+				NewContentFromFunctionResponse(frAsyncPlanFinal, "user"),
```
karolpiotrowicz, line 1330: This expectation is right, and I want to record why I stopped worrying about it, because it looks alarming.

The `want` puts two model turns next to each other, the first carrying a function call that never gets a response. That reads like a turn-alternation problem. It isn't new: main already emits the same shape without this branch — `user text | CALL(p1 unanswered) | CALL(p2 unanswered) | user text` produces two adjacent model turns today, and nothing in the repo validates mid-history role alternation. The only role check, in `maybeAppendUserContent`, looks at the last content only.

What does change is frequency: one pending call now suffices where main needed two. I have not put such a request through a live endpoint, so I can't tell you how Vertex reacts — only that main is already sending them, so this branch isn't what would break it.

No change needed. Flagging it because it is the kind of expectation a future reader will challenge, and the answer takes a measurement to reach.

karolpiotrowicz (changes requested):
All five points from the last round are addressed, and I re-ran the whole change rather than just the new commit. One of the comments written to address them says something that isn't true, and there's one behavior change nobody has mentioned yet. Neither should stop this merging.

## The new comment on the `default` branch is false

[`contents_processor.go:442-448`](https://github.com/google/adk-go/blob/1fdd14bf597bf22af18038f7d039a2ced8c9f965/internal/llminternal/contents_processor.go#L442-L448) — details on the line itself. Short version: I instrumented both branches and the combination the comment rules out does happen, on a history where every call precedes its response and every call ID is unique.

## `load_artifacts` loses its artifact bodies in one interleaving

[`processLoadArtifactsFunctionCall`](https://github.com/google/adk-go/blob/1fdd14bf597bf22af18038f7d039a2ced8c9f965/tool/loadartifactstool/load_artifacts_tool.go#L157-L165) injects artifact contents only when `req.Contents[len-1]` carries a `load_artifacts` response. Tail routing changes which content is last, so the predicate flips in both directions depending on the interleaving:

| history | main | with this branch |
|---|---|---|
| `load_artifacts` answered, then a long-running tool completes | injects | **does not inject** |
| the `load_artifacts` response is the newest event | does not inject | **injects** |

The second row is the case that tool was written for, and this branch fixes it. The first row is a real loss: the model asked to load an artifact and gets back only the name list, with no error and no log. It can re-call, because the names still reach it through `appendInitialInstructions`, and main's behavior there was itself a side effect of the bug you're fixing. So I don't think it blocks — but it's the one place where a measurement differs between main and this branch in a way that makes something worse, and it's worth a line in the PR body so the next person doesn't rediscover it as a bug report.

## The pending-call shape is not new, in case it comes up

The new ordering puts a model turn carrying an unanswered call directly before another model turn. That looked like it might be a problem, so I checked: main already emits exactly that shape without this branch — `user text | CALL(p1 unanswered) | CALL(p2 unanswered) | user text` gives two adjacent model turns today. Nothing in the repo validates mid-history role alternation either. What does change is how often the shape occurs, since one pending call now suffices where main needed two. I have not put a request with two adjacent model turns through a live endpoint, so I can't tell you what Vertex does with it — only that main is already sending them.

## Three things I could not settle, stated as gaps rather than findings

- **Nothing asserts that pass 2 preserves the multiset.** I ran 20,000 generated histories and the properties I checked — adjacency, and the tail carrying the latest response — are both satisfied by duplicated output, so they don't cover it. Neither does the suite.
- **Compaction crossed with tail routing is run, not checked.** `TestContentsRequestProcessor_CompactionKeepsToolPairing` passes with this change reverted: its history is one pair with nothing after it, and its assertions set `sawCall`/`sawResponse` flags without ever comparing an index. A row where a summary is followed by an unrelated exchange and then the late completion would actually pin the reordering.
- **The adk-python comparison in the PR body.** adk-python isn't vendored here, so I checked the reasoning rather than the source this round.

## Separately, and not against this PR

Four pre-existing things surfaced while checking the interleavings. I reproduced each on main with no branch applied, so none is yours, but none has an issue open. Happy to file them if you'd rather not.

- **A final response event answering calls from two different call events is emitted twice**, and one response is emitted before its call. Reachable through the REST create-session path, which appends a client-supplied event list verbatim with no well-formedness check.
- **A user turn typed while a long-running tool runs is deleted from the prompt.** "halt processing and ignore tool output", sent between the call and its completion, does not survive into the request on main or on this branch.
- **A second response for the same call ID overwrites the first** in the `map[string]int` index, so an intermediate result is dropped when pass 1 doesn't run.
- **An orphan final response returns a fatal error** rather than degrading.

<details><summary>Reproduction for the two claims above that involve measurement</summary>

Both run as a test in `internal/llminternal`, driving `ContentsRequestProcessor` rather than the helpers directly. `ev`, `call`, `callN`, `resp`, `respN` and `text` below are one-line constructors defined alongside the test, not repo functions.

```go
// The default branch does coincide with tail routing.
// Every call precedes its response; every call ID is unique.
events := []*session.Event{
	ev("test_agent", callN("c1", "c2")),
	ev("user", respN("c2")),
	ev("test_agent", callN("c3")),
	ev("user", respN("c1", "c3")),
}
// With counters added to the default branch and to the tail-routing branch:
//   defaultBranchHits=1 defaultAndTailHits=1
// Emitted, identically on main and with this branch:
//   |model:CALL(c1)CALL(c2)|user:RESP(c2)RESP(c1)RESP(c3)|model:CALL(c3)|user:RESP(c1)RESP(c3)

// load_artifacts, the shape that regresses.
events = []*session.Event{
	ev("user", text("user", "Summarize report.pdf while the job runs.")),
	ev("test_agent", call("lr1", "long_job")),
	ev("test_agent", callNamed("la1", "load_artifacts")),
	ev("user", respNamed("la1", "load_artifacts", map[string]any{"artifact_names": []any{"report.pdf"}})),
	ev("user", resp("lr1", "long_job", map[string]any{"result": "done"})),
}
// main:          ...|model:CALL(lr1)|user:RESP(lr1)|model:CALL(la1)|user:RESP(la1)   -> last is load_artifacts, injects
// this branch:   ...|model:CALL(la1)|user:RESP(la1)|model:CALL(lr1)|user:RESP(lr1)   -> last is long_job,       does not inject
```

Run with `go test -run TestLateCompletion -v ./internal/llminternal/`.

</details>

internal/llminternal/contents_processor.go
jjsasha63: I instrumented both branches and drove your history through `ContentsRequestProcessor`: `defaultBranchHits=1 defaultAndTailHits=1`. Both halves of the comment were wrong.

I rewrote it around the precondition you found: one call event whose siblings are answered by two different response events, plus pass 1 returning early on adjacency so it merges nothing. The worked example in the comment is `CALL(c1,c2) | RESP(c2) | CALL(c3) | RESP(c1,c3)`. The `PopulateClientFunctionCallID` sentence is gone, since it fills an ID only when `fn.ID == ""`.

adk-python behaves the same way here. Its pass 1 has the identical adjacency early return (`_tool_call_rearranger.py:258-262`), and running your history through both Python passes gives byte-identical output to Go, duplicated `RESP(c1,c3)` included.

jjsasha63:
I fixed the comment. The change is comment-only, so the `load_artifacts` finding went into the PR body instead.

## The default branch

The same instrumentation on both branches gives `defaultBranchHits=1 defaultAndTailHits=1` on your four-event history. I rewrote the comment around the precondition you gave. Details on the line.

## load_artifacts

I reproduced both rows of your table on main and on this branch and got what you got. It is now a section in the PR body, with the measurement.

## The three gaps

- Multiset preservation. A test asserting it fails today because of the double-emit, so it belongs with that fix rather than here.
- Compaction crossed with tail routing. `TestContentsRequestProcessor_CompactionKeepsToolPairing` is not on this branch at all. It landed on main after the last merge here, so this branch has never run it. I merged current main in and it passes, along with the rest of the suite. Your point that its assertions do not pin an ordering still stands, so say the word and I will add the row.
- adk-python. I ran it against 2.8.0 this round.

## The four pre-existing ones

I will file them. One correction from running Python: the double-emit is not Go-specific. adk-python produces the same duplicate on the same input, so that one wants an upstream issue too.

wolo-lab (approved):
LGTM, thanks!

karolpiotrowicz (approved):

PR #1429, opened 2026-08-31
feat(mcptoolset): forward request-scoped metadata to MCP tool calls

### LINKED ISSUE #469: Set MCP metadata on a request level

jjsasha63:

- Closes: #469
- Supersedes: #602 (and #475 before it)

This carries on the work in #602 by @alexshopee and #475 by @philippslang. The
design is theirs. Both PRs sat on branches that had drifted a long way behind
`main`, across the v1 to v2 module bump, so the code here is written against
the current API rather than rebased. #602 also cannot be merged as it stands,
because its CLA check covers commit authors but not the person who opened it,
and that has not been resolved since February.

**Problem:** an agent has no way to attach `_meta` to an MCP tool call. Values
that belong to the incoming request — tracing IDs, OpenTelemetry span context,
tenant identifiers — cannot reach the MCP server. The only workaround is to
reimplement the toolset. @vnalawad-tibco described this use case on #602:
propagating OTEL tracing context from the agent to MCP servers.

**Solution:** an optional `Config.MetadataProvider`, called once per invocation
with the tool's `agent.Context`, whose result becomes `mcp.CallToolParams.Meta`.

```go
mcptoolset.New(mcptoolset.Config{
    Transport: transport,
    MetadataProvider: func(ctx agent.Context) (map[string]any, error) {
        return map[string]any{"trace_id": traceIDFrom(ctx)}, nil
    },
})
```

Three points worth review:

- **Nothing is attached by default.** With no provider, or a provider returning
  nil or an empty map, `Meta` stays empty and `omitempty` means no `_meta` is
  sent. The behavior of existing callers does not change.
- **The provider returns an error.** Producing metadata can fail — minting a
  token, reading a trace header. For auth- or tenant-scoped metadata, quietly
  issuing the call without it is worse than not calling at all, so a failure
  aborts the tool call and names the tool.
- **The map is copied.** The MCP client writes its own reserved `_meta` entries
  (`io.modelcontextprotocol/protocolVersion` and friends, added in go-sdk
  v1.7.0) into whatever map it is given. Without the copy, a provider returning
  a map it reuses between invocations would find those keys written into it.

### Testing Plan

**Unit Tests:**

Seven cases in `tool/mcptoolset/set_test.go`, all against an in-memory MCP
transport, asserting on the `_meta` the server actually received:

```
$ go test -race -count=1 -shuffle=on ./tool/mcptoolset/ -run TestMetadata -v
--- PASS: TestMetadataProvider
    --- PASS: TestMetadataProvider/provider_metadata_reaches_the_server
    --- PASS: TestMetadataProvider/no_provider_configured_sends_no_metadata
    --- PASS: TestMetadataProvider/provider_returning_nil_sends_no_metadata
    --- PASS: TestMetadataProvider/provider_returning_an_empty_map_sends_no_metadata
--- PASS: TestMetadataProviderError
--- PASS: TestMetadataProviderKeepsClientMeta
--- PASS: TestMetadataProviderMapIsCopied
--- PASS: TestMetadataProviderReceivesToolContext
ok      google.golang.org/adk/v2/tool/mcptoolset
```

The whole package passes as well: `go test -race -count=1 -shuffle=on
./tool/mcptoolset/`.

Each test was checked against a deliberately broken implementation, so none of
them passes vacuously. Dropping the copy fails `TestMetadataProviderMapIsCopied`
only; dropping the assignment fails `TestMetadataProvider`.

Not applicable. This is a library-only change with no service or model
interaction; the in-memory MCP transport exercises the full client-to-server
path.

### Checklist

### Additional context

The public surface grows by one type and one field, both additive:
`mcptoolset.MetadataProvider` and `Config.MetadataProvider`.


tool/mcptoolset/set.go
```diff
@@ -26,6 +26,16 @@ import (
 	"google.golang.org/adk/v2/tool"
 )
 
+// MetadataProvider supplies request-scoped metadata for MCP tool calls. The
+// returned map is sent as the `_meta` field of mcp.CallToolParams, letting an
+// agent forward values such as tracing IDs or tenant identifiers taken from the
+// incoming request down to the MCP server.
+//
+// It is called once per tool invocation, before the call is issued. Returning a
```
karolpiotrowicz, line 34: Two small inaccuracies in this doc comment, both worth fixing while the API is still unreleased.

"It is called once per tool invocation" is zero times when a call is awaiting or has been denied human-in-the-loop confirmation, because the provider sits after the confirmation block in `Run`. On the resuming turn it runs against the resuming request's context rather than the one that triggered the tool call. The field doc on `Config.MetadataProvider` already says "before each tool call", which is accurate — this one could just match it.

The comment also never says the provider may be invoked concurrently, while the comment on `TestMetadataProviderMapIsCopied` tells a reader "a provider may hand back one it reuses between invocations". Returning a shared map is safe. Returning one you also write to is not, and eight concurrent `Run` calls with that pattern trip the race detector. Stating the ownership rule here — the returned map and anything reachable from it must not be mutated after return — would head that off.

tool/mcptoolset/tool.go
```diff
@@ -117,10 +121,23 @@ func (t *mcpTool) Run(ctx agent.Context, args any) (map[string]any, error) {
...
-	})
+	}
+	if t.metadataProvider != nil {
+		meta, err := t.metadataProvider(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("metadata provider for MCP tool %q failed: %w", t.name, err)
+		}
+		// The MCP client adds its own reserved keys to this map before sending
+		// the call, so pass a copy rather than a map the provider may reuse
+		// across invocations. A nil map is fine: Meta is `omitempty`, so no
```
karolpiotrowicz, line 135: "A nil map is fine: Meta is `omitempty`, so no `_meta` is sent" holds only on the legacy protocol path. Under `usesNewProtocol()` the client calls `injectRequestMeta`, which replaces a nil map with a fresh one and writes `protocolVersion`, `clientInfo` and `clientCapabilities` into it, so `_meta` goes on the wire whether or not a provider ran. `TestMetadataProviderKeepsClientMeta` in this same package asserts exactly that.

The behavior is right and the clone is genuinely load-bearing — only the stated reason is wrong, and a future reader reasoning from it could conclude the clone is unnecessary. Something like "the client attaches its own reserved `_meta` regardless, and `maps.Clone` keeps nil nil so a nil provider map contributes nothing" would be accurate.

tool/mcptoolset/set_test.go
```diff
@@ -796,3 +798,195 @@ func TestNewToolSet_RequireConfirmationProvider_Validation(t *testing.T) {
...
+// `_meta` the MCP client sets for itself, rather than replacing it.
+func TestMetadataProviderKeepsClientMeta(t *testing.T) {
+	ts, got := startMetaEchoServer(t, func(ctx agent.Context) (map[string]any, error) {
+		return map[string]any{"trace_id": "abc-123"}, nil
+	})
+	if _, err := runSingleTool(t, ts); err != nil {
+		t.Fatalf("Run() err = %v", err)
+	}
+	for _, key := range mcpClientMetaKeys {
+		if _, ok := got.meta[key]; !ok {
```
karolpiotrowicz, line 957: This is a presence check, but the comment above the test claims the stronger property that provider metadata is added "rather than replacing it". A provider returning `io.modelcontextprotocol/clientInfo: "spoofed"` satisfies the presence check with its own value, so the test stays green under a total override.

Comparing the value — that `clientInfo` is the real `adk-mcp-client` implementation the client set — would make the test check what its name says. Adding `progressToken` to `mcpClientMetaKeys` would extend it to the one reserved key with no `io.modelcontextprotocol/` prefix, which is the one a caller is most likely to collide with by accident.

tool/mcptoolset/set_test.go
```diff
@@ -796,3 +798,195 @@ func TestNewToolSet_RequireConfirmationProvider_Validation(t *testing.T) {
...
+func TestMetadataProviderReceivesToolContext(t *testing.T) {
+	var gotInvocationID string
+	ts, _ := startMetaEchoServer(t, func(ctx agent.Context) (map[string]any, error) {
+		gotInvocationID = ctx.InvocationID()
+		return nil, nil
+	})
+	if _, err := runSingleTool(t, ts); err != nil {
+		t.Fatalf("Run() err = %v", err)
+	}
+	if gotInvocationID == "" {
```
karolpiotrowicz, line 989: `gotInvocationID != ""` is satisfied by any freshly constructed invocation context, because `icontext.NewInvocationContext` generates `"e-" + uuid` whenever `InvocationID` is empty. So this passes even for an implementation that manufactured its own context rather than passing the live one, which is the property the comment above says the test checks.

Having `runSingleTool` return `invCtx` and asserting `gotInvocationID == invCtx.InvocationID()` would pin it. The end-to-end shape from the issue is also worth a case here: put a value on the request context with `context.WithValue` before building the invocation, and assert the provider reads it and the server receives it. That path works today, and nothing in the suite locks it.

tool/mcptoolset/set_test.go
```diff
@@ -796,3 +798,195 @@ func TestNewToolSet_RequireConfirmationProvider_Validation(t *testing.T) {
...
+		func(ctx context.Context, req *mcp.CallToolRequest, input Input) (*mcp.CallToolResult, Output, error) {
+			got.called = true
+			got.meta = req.Params.Meta
+			return nil, Output{WeatherSummary: "sunny"}, nil
+		})
+	if _, err := server.Connect(t.Context(), serverTransport, nil); err != nil {
+		t.Fatalf("server.Connect() err = %v", err)
+	}
+
+	ts, err := mcptoolset.New(mcptoolset.Config{
```
karolpiotrowicz, line 845: `startMetaEchoServer` is the only place in the package that sets `MetadataProvider`, and it never sets `RequireConfirmation`, so no test drives the two together.

The ordering in `Run` is correct as written — the provider is called after the confirmation block, so a call awaiting approval never invokes it — but nothing locks that in. Moving the provider block above the confirmation logic, which is a plausible "build the params up front" refactor, would make a provider error mask `ErrConfirmationRequired` and stop the confirmation flow from ever being requested, and the suite would stay green. A case with `RequireConfirmation: true` and a provider that records whether it ran would cover it.

karolpiotrowicz (commented):
The mechanism works and it does solve #469. Driving the public path with a value placed on the request `context.Context` upstream, the provider reads it back and the MCP server receives it in `_meta`, so the request-scoping is real rather than nominal. The build, the full test suite on both modules, `go vet`, `gofmt`, `staticcheck` (byte-identical to base), `govulncheck` and `gosec` are all clean, and `apidiff` confirms both additions are compatible. The five new tests are not vacuous: I broke the implementation four different ways — dropped the copy, dropped the assignment, swallowed the provider error, sent an always-empty map — and each one failed exactly the test you predicted it would in the description.

One thing worth resolving before merge, and five smaller notes left on the individual lines they concern.

## The reserved `_meta` namespace has no guard, and the test named for it cannot detect a violation

`injectRequestMeta` in go-sdk v1.7.0 fills each of its reserved keys only when the key is absent, so assigning the provider's map wholesale at [`tool.go:137`](https://github.com/google/adk-go/blob/7f886e388d918fa8e7ff2406e9c1d6fd4cf9b699/tool/mcptoolset/tool.go#L137) lets a provider key win over the client's own. All four keys are affected. A provider returning them sends this to the server:

```
protocolVersion    = "1999-01-01"
clientInfo         = {"name":"spoofed"}
clientCapabilities = {"sampling":"spoofed"}
progressToken      = "hijacked"
```

I don't think this rises to a security problem — the provider is opt-in code the application author writes, and `Config.Client` already let a caller supply their own `*mcp.Implementation` before this change, so `clientInfo` and `clientCapabilities` were reachable already. Two things still make it worth handling. `progressToken` is the one reserved key with no `io.modelcontextprotocol/` prefix warning a caller off it, so a generic metadata bag that happens to contain that name collides by accident rather than by intent. And `protocolVersion` becomes caller-settable here for the first time, which on the streamable HTTP transport is the value the server cross-checks against the `MCP-Protocol-Version` header.

The part I'd most want changed is the test. [`TestMetadataProviderKeepsClientMeta`](https://github.com/google/adk-go/blob/7f886e388d918fa8e7ff2406e9c1d6fd4cf9b699/tool/mcptoolset/set_test.go#L947-L961) says in its comment that provider metadata is added "rather than replacing it", but the assertion only checks that the three keys are *present*, so it stays green under a total override. `progressToken` isn't in `mcpClientMetaKeys` at all. Whether you decide to strip the reserved namespace, reject it with an error, or document that provider keys deliberately win, the test should assert the value and cover a colliding key.

## Nothing pins that the provider is evaluated per call

Every new test performs exactly one `Run`, so an implementation that evaluates the provider once and caches the result passes the whole package suite. I checked rather than assumed — this mutation of `tool.go` is green on `go test ./tool/mcptoolset/`:

<details><summary>The caching mutation that survives the suite</summary>

```go
// added to mcpTool:
//     cachedMeta map[string]any
//     metaCached bool

	if t.metadataProvider != nil {
		if !t.metaCached {
			m, err := t.metadataProvider(ctx)
			if err != nil {
				return nil, fmt.Errorf("metadata provider for MCP tool %q failed: %w", t.name, err)
			}
			t.cachedMeta = m
			t.metaCached = true
		}
		meta := t.cachedMeta
		params.Meta = maps.Clone(meta)
	}
```

```
$ go test -count=1 ./tool/mcptoolset/
ok      google.golang.org/adk/v2/tool/mcptoolset        0.416s
```

</details>

The shipped code is correct, so this is only a gap in coverage — but request-scoping is the entire point of the feature, and a provider handing back a fresh request id per call would silently send the first call's id forever without any test noticing. Two `Run`s with a counter-based provider, asserting the two `_meta` differ and the provider ran twice, would close it.

## Merge notes

The branch conflicts with main. The only conflict is that main and this branch both appended tests at the end of `set_test.go`, so keeping both sides resolves it — I did that locally and the merged tree builds, vets, formats and passes.

#1168 also touches `tool.go`, `set.go` and `set_test.go` and is also conflicting. It handles inbound `_meta` on results and elicitation where this handles outbound `_meta` on calls, so there's no semantic overlap, but whichever lands first will force the other to rebase. Worth a word between you.

PR #1431, opened 2026-08-31
fix(compaction): carry durable facts across re-summarization

baptmont:
Follow-up to the five-PR compaction stack (#1231, #1232, #1234, #1235, #1236),
all now merged. This changes the default summarizer prompt, because as shipped
tail retention silently discards the early conversation in about half of runs.

Tracking issue: #298.

## The defect

Tail retention seeds each summary with the previous one, so a value stated early
is recompressed on every pass. The default prompt asked the model to "reiterate
the user request" and keep the rest "concise" — under a rolling summary that is
an instruction to describe the most recent turns.

I measured it against a real model rather than reasoning about it. Eight
arbitrary facts — a region, an incident ID, a deadline, a name, a bucket — stated
early, buried under 48 filler turns, then probed once the events holding them
had been summarized away. A probe counts only if the answer carries the value.

| prompt | conversations | lost everything | probe recall |
|---|---|---|---|
| before this PR | 10 | **5** | 39/80 |
| after this PR | 9 | **0** | 70/72 |

One-sided Fisher p = 0.022.

## The loss is all-or-nothing, which is why it went unnoticed

Run scores cluster at 8/8 or 0/8; almost nothing lands in between. Each pass
either copies the values forward or **generalizes them to their labels**, and
once generalized they cannot come back, because the next pass sees only the
previous summary and the retained tail. It is a ratchet, not a fade.

From a failing run, summary 2 carries the values:

>  *   Deployment region: `europe-west4`
>  *   Incident ID: `INC-77312`

Summary 3 keeps every label and drops every value:

> the user providing several project-specific details (deployment region,
> incident ID, database choice, cutover deadline, on-call owner, error budget,
> staging bucket, retry cap), which the AI acknowledged

Meanwhile the *narrative* — a bulleted list of every filler topic — is preserved
in full and keeps growing. The summarizer spent its "concise" budget on the shape
of the conversation and paid for it with the specifics. Summaries 4 onward never
recover, and the agent then answers "I do not have that information" rather than
confabulating. Honest, and gone.

**An intermittent default is worse than a consistent one.** It passes whatever
spot check someone runs before trusting it, then loses the conversation in
production.

## The change

One more instruction in `defaultPromptTemplate`: maintain a "Durable facts"
section, copy each concrete value verbatim, carry them forward unchanged even
when recent turns are about something else, and drop narrative rather than facts.

The package doc gains a "What bounding costs" section. It previously presented
tail retention as the strategy that bounds prompt growth and said nothing about
what bounding costs, which is the more useful half of the tradeoff.

## Reproducing the numbers

`examples/compactionrecall` is the harness the table above came from, included
so the claims are checkable rather than asserted. It plants the facts, buries
them, probes them, and reports per-run scores:

```
GOOGLE_API_KEY=... go run ./examples/compactionrecall -runs=3
```

It prints per-run scores rather than only an average, because recall is
near-binary and an average over too few runs describes neither outcome. Its
`prior` arm is the pre-change default reproduced verbatim, so the two arms differ
in the one instruction and nothing else.

Doing exactly that on `gemini-3.5-flash` — a different model and a different
backend from the measurements above, which used `gemini-2.5-flash` on Vertex —
gave, on the same config and conversation, around 20 compaction passes per run:

| arm | runs | recall | lost everything |
|---|---|---|---|
| `default` (this PR) | 3 | 24/24 | 0 |
| `prior` (before) | 3 | 0/24 | 3 |

So the effect is not specific to the model it was found on.

It is not specific to Go either. adk-python ships a byte-identical default
prompt and the same rolling-summary seed — `_events_to_compact_for_token_threshold`
prepends the previous summary to the next window, exactly as tail retention does
here. Running the same experiment against adk-python on the same model: the
shipped prompt scored 0/8, 6/8, 0/8 over three runs, losing everything in two of
them, and the instruction from this PR ported across scored 8/8 four times out of
four. Worth reporting upstream; it is not a defect in this change.

## What it costs

Larger summaries cross `TokenThreshold` sooner, so compaction runs more often:
mean 13.2 → 24.3 summarizer calls over the same conversation, worst case 46.
That trade is not avoidable — you cannot bound the prompt and retain everything.
The workable target is to bound the *narrative* and carry the *facts*, because
facts accumulate slowly and narrative grows every turn.

## Testing plan

`go build`, `go test -race -count=1 -shuffle=on`, `golangci-lint run` (v2.3.1) and
`go mod tidy -diff` clean in both modules at this commit.

Two guard tests, both mutation-tested — deleting instruction 3 fails the first,
making `NewLLMSummarizer` ignore the caller's template fails the second:

- `TestDefaultPromptRequiresFactsBeCarriedForwardVerbatim` asserts on the prompt
  the summarizer actually sends, so it covers the defaulting path and not just
  the constant.
- `TestCustomPromptTemplateIsUsedVerbatim` pins that the default is not appended
  to a caller's own template.

`examples/compactionrecall` is not a test and runs nothing in CI; it builds and
lints clean as part of the module.

`agent/llmagent/testdata/TestCompactionE2E.httprr` is re-recorded. That test's own
documentation calls this out: the cassette is sensitive to anything that changes
prompt bytes, "including the summarizer prompt template", and any such change
requires re-recording. It is the only cassette affected.

## Deliberate decisions, so they are not re-reported as defects

- **Changing the default rather than adding an opt-in.** A flag would leave the
  broken behavior as what everyone gets, and the failure mode is silent. Callers
  who want the old prompt can pass `PromptTemplate`.
- **No behavioral test for recall.** Recall only manifests against a live model
  over ten or more passes, and the repository forbids live calls in tests; a
  cassette cannot capture a stochastic effect that needs many runs to see. The
  evidence is the out-of-repo measurement above. The committed tests are
  regression guards on the prompt reaching the model, and they are labelled as
  that rather than as proof the prompt works.
- **Per-turn re-arming is not fixed here.** A summary permanently over threshold
  re-triggers compaction every turn; the worst run above (46 compactions across
  ~64 turns) is close to that. It is pre-existing, this change reaches it sooner,
  and it is a cost concern rather than a correctness one. Separate PR.
- **US spelling** (`generalize`) to match the repository.

## Caveats on the measurement

Filler is homogeneous — sixteen "explain X in two sentences" prompts repeated
three times — which probably accelerates convergence onto the narrative pattern,
so the before-rate here may be worse than a varied real conversation would give.
Facts are clustered at the start. One model (`gemini-2.5-flash`), default
sampling. The direction and the mechanism are solid; the exact rate is not.

## Probably not Go-specific

adk-python ships the same `reiterate the user request` / `CRITICAL INSTRUCTIONS`
prompt text, so the same pairing is likely to show the same intermittent loss.
That is inferred from the prompt text, not measured against Python.

**Review this exact commit: `ee86cea1843d5ea04bbd5810ea864b85655ae55f`.** Check out the SHA, not a branch name.
Several branches in this repository hold older shapes of the compaction work and
resolve to different trees.

This branch carries three commits — the prompt and doc change, the example, then
the review fixes — plus a merge of `main` to clear a `BEHIND` block. The whole change is 101 lines across three source files plus one
re-recorded cassette; if you are looking at the five-PR compaction stack itself,
you are on the wrong commit — that is already merged.

### DIFF
--- a/agent/llmagent/testdata/TestCompactionE2E.httprr
+++ b/agent/llmagent/testdata/TestCompactionE2E.httprr
@@ -1,5 +1,5 @@
 httprr trace v1
-1000 2106
+1000 1854
 POST https://generativelanguage.googleapis.com/v1beta/models/gemini-3.5-flash:generateContent HTTP/1.1
 Host: generativelanguage.googleapis.com
 User-Agent: Go-http-client/1.1
@@ -8,9 +8,9 @@ Content-Type: application/json
 
 {"contents":[{"parts":[{"text":"What is the weather in Zurich?"}],"role":"user"}],"generationConfig":{},"systemInstruction":{"parts":[{"text":"You are a concise assistant. Answer in one short sentence.\n\nYou are an agent. Your internal name is \"compaction_agent\". The description about you is \"agent used to exercise context compaction\"."}],"role":"user"},"tools":[{"functionDeclarations":[{"description":"Returns the weather in a city.","name":"get_weather","parametersJsonSchema":{"additionalProperties":false,"properties":{"city":{"description":"the city to look up","type":"string"}},"required":["city"],"type":"object"},"responseJsonSchema":{"additionalProperties":false,"properties":{"weather":{"type":"string"}},"required":["weather"],"type":"object"}}]}]}HTTP/2.0 200 OK
 Content-Type: application/json; charset=UTF-8
-Date: Mon, 24 Aug 2026 12:39:26 GMT
+Date: Mon, 31 Aug 2026 12:43:57 GMT
 Server: scaffolding on HTTPServer2
-Server-Timing: gfet4t7; dur=1162
+Server-Timing: gfet4t7; dur=1351
 Vary: Origin
 Vary: X-Origin
 Vary: Referer
@@ -30,9 +30,9 @@ X-Xss-Protection: 0
               "args": {
                 "city": "Zurich"
               },
-              "id": "call_156800"
+              "id": "call_3445095"
             },
-            "thoughtSignature": "Et4ECtsEARFNMg+mULUQ83jwmVgWmMDjjtDc0pk+4l6IA60xz+5EkWz5WTJQQZibk4C6F2LZB/GXuSjrXxX2GBMtrgTjsHqcTIk2ENcmHQ4Q5YCwUNQR/yX2eFPXJlKLZml7wmfa1jSyd2u6D7qQqyjZrvenktqN9HCLL2m7PTBR1Z7WJVbKnbSTry2tNUwvL7P36ddw7YjrAdO5PTtnaqidMB4c0vBOPPP6zX2s6WB3jJN/PN3lqqekizo/twpL5/a5mTNIHymCpjpMwJeNx1o640W5GOyyq6dpc57I9fYrsm0ofSvPJwoPXiAVd4bPmRwabrjfE5GJi59RMxhe+raMSOs2IIaJh9JC09eheNq3dPeXrdbRFCs9srTUMfM1uHId/lVlo0uVzDZUFBhtAO129Iv0lcUH11rxO+1/NUJk0JNn6v+qU/x3OLG60rnuEMUn1NAnhlMzRlaegVLQo/rS2yc/QNBWcBsCHPzKcyHkqNWWFRIwvy/+lucvH/5nn84fUH3tfZPj1xVV+CT//frl9lJ8Tz1RxikldUqkpsMmHERIOnWOVTS8w01F8kBDAKM9Kwiw6wmb9oH97l0kEngueYgmEOX4oJ9nYPDn5ArbZfdN+IhlInx701e3sjriQqONSfhPOmlJawaWV2hsSFn8F9eRLQ238hOnpN+zKzjudJtRtJiOTMZxgxBPLdqjPDiS7WCbV0KsnNcD+Zk+DbjHsk3qkVnxb/ZRxTK+JVOENoXi1OikQZ1KnilfNW3es3dnFehLqrJJZ3tmXZKkMsPRcPfiq7Vb49tmO4xCSk9F"
+            "thoughtSignature": "EqEDCp4DARFNMg8fvkN9H7IJF37HbQzdctqOkd5UPMoy8r4zKaLytkFTk0CKplXRX77FIB0nahZGR34JssqpZgDJZXraHyZd2sGL6T8VvZjOR/MpGSNkpj1msp8lFDcA1xw24hPLzx+rWehQBHTTUU3hagq/hZs6nZX4u0a5eNBkh8V4YZWjBXPJn1X+eK/wx7kquRlD7bDkYYSKIbpViGIWYczJhna7y8LOJrPS+tLEnYnJXocS/JJI1rWh7wBrIjNzqUM4Iqt57kY5Ii6sg/mOQXDefsxdm99mr9+1W/Soc8dTjieakqei1iHnIOmtni+a1gcUyNJWtlnZWkw8obP4bg2cm9G3iFaSHPpZpZQmVqG/peDMt/6vhHMaBbPdIRVNcbDLA7xY9M/taeGTmsKNxya5JPil+JaygQYyib3gBxGh9Jyj07iawlo0gUL7tAdtuVogaDCNbk1ZzRviO4qRIsuDUTfogwt1MWlkoCg8qzmvXcNnsm29lpyHE3izk4O2zGPLJDEYlYog0mT6F+sLdMoJaKlseo5SJaq80NhyaDAp"
           }
         ],
         "role": "model"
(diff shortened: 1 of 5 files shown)


baptmont:
### Round update: the measurement harness is now in the PR

The PR previously asserted recall numbers that nobody could check. It now ships
the harness that produced them, as `examples/compactionrecall`, so the before
and after can be run back to back:

```
GOOGLE_API_KEY=... go run ./examples/compactionrecall -runs=3
```

Its `terse` arm is the summarizer prompt as it was before this change, standing
in for any custom `PromptTemplate` that asks only for a concise summary.

**It also corroborates the result on a second model.** The measurements in the
description were taken on `gemini-2.5-flash` via Vertex. Running the harness on
`gemini-3.5-flash` via the Gemini API, same config and same conversation over 20
compaction passes, gave 8/8 for the new default and 0/8 for the terse prompt. The
effect is not specific to the model it was found on.

The example prints per-run scores rather than only an average, deliberately:
recall is near-binary, so two configs that both average 50% can mean "always
half-remembers" and "works perfectly half the time". That distinction is the
whole finding, and averaging hides it.

**Review `5962e4d350d249e74c257d5b86a3b12fab64f6e8`.** Two commits now — the
prompt and doc change, then the example. Nothing in the first commit changed;
the example is additive and runs nothing in CI.

baptmont:
### Round update: the recall scoring is now sound, not just lucky

A reviewer asked the right question about the harness: the miss detector matched
a **fixed list of English denial phrases**, so a model that declined in wording
the list did not have could have been scored as remembering.

I checked before changing anything. Across all 120 probes behind the numbers in
the description, every denial used the instructed phrasing, no denial escaped
the list, and **no denial-shaped answer was scored as a hit**. The published
results are unaffected.

The instrument was still fragile by construction, so two fixes:

**The agent is now told to answer `NO_RECORD`** when it cannot answer, and the
scorer keys off that token. A denial is no longer inferred from prose. Verified
against a live model: the terse arm now returns a bare `NO_RECORD` per lost fact
instead of a sentence to pattern-match. The phrase list is kept only as a
fallback for a model that ignores the instruction.

**Short values are matched on word boundaries.** The retry-cap probe accepted
`"7"` as a bare substring, so any answer mentioning 17 — or the incident ID
`INC-77312` — would have scored it as recalled. That is the weakest probe
deciding the score, and it now cannot.

Neither changes any conclusion; both mean the next person to run this gets a
number they can trust rather than one that happened to be right.

Also rebased onto current `main`, dropping the earlier merge commit. Two clean
commits now.

**Review `15197d2c518f07ee77730dec9c564baa4afec1e6`.**

jjsasha63 (commented):
Reviewed at `15197d2c518f07ee77730dec9c564baa4afec1e6`. The library change looks sound and both guard tests do what the description says they do. One thing needs fixing before this lands: the example's `terse` arm is not the previous prompt, so the one measurement a reader can reproduce does not isolate the change.

## What I checked

- `go build ./...` and `go vet` clean. Full `go test -count=1 -shuffle=on ./...` passes. `-race` clean on `session/compaction` and on `TestCompactionE2E`.
- Mutation-tested both guards. Deleting instruction 3 from `defaultPromptTemplate` fails `TestDefaultPromptRequiresFactsBeCarriedForwardVerbatim`. Making `NewLLMSummarizer` ignore `cfg.PromptTemplate` fails `TestCustomPromptTemplateIsUsedVerbatim`. Both bite.
- The re-recorded cassette is a clean re-record: same request sequence, new prompt bytes, no credentials in it.
- The package doc's advice about `{key}` state placeholders is accurate. `Instruction` goes through `InjectSessionState` on every request, so a value held in state does reach the model regardless of how many compaction passes have run.

## 1. The `terse` arm is not the previous default

`tersePromptTemplate` in `examples/compactionrecall/main.go` is 400 characters against the old default's 679. It drops the `CRITICAL INSTRUCTIONS:` framing, instruction 1 (state the primary language) and instruction 2 (list the exact tool names), and rewrites "The rest of the summary should be concise" to "The summary should be concise".

The comment above it says "It is what the default used to be", which is not the case. Instruction 2 is itself an instruction to carry specifics forward exactly, and the numbered-list framing plausibly affects how literally the model treats the rest of the prompt.

So the back-to-back run in the description compares the new prompt against a third prompt rather than against `main`, and the 8/8 against 0/8 result does not isolate instruction 3. The out-of-repo measurement in the table may well be sound, but the harness shipped here cannot reproduce it, which is the thing the harness is for.

Suggested fix: make `tersePromptTemplate` the exact pre-PR default text and rerun that arm. If the shorter prompt is also interesting, add it as a third arm under its own name.

## 2. Instruction 3 only covers what the user stated

The instruction asks for "every concrete detail the user has stated". Values that arrive in tool responses are excluded, and for a tool-using agent that is where most identifiers, IDs and numbers come from. The transcript already renders them as `Tool response from X: ...`, so they are available to the summarizer.

Either widen the instruction to cover tool-derived values, or note in the comment why user-stated facts are the deliberate boundary. Right now a reader cannot tell which it is.

## 3. Persistence of injected content

`formatEvents` escapes newlines so a tool response cannot forge a turn, and that still holds. But the new instruction tells the model that a durable fact "MUST be carried forward unchanged" indefinitely. Text in a tool response that reads like a fact list previously decayed out of the rolling summary within a few passes; now the prompt argues against dropping it.

This is a small change in a default rather than a new attack surface, and it may be an acceptable trade. Worth a sentence acknowledging it either way.

## 4. Three smaller things in the harness

- `buriedFacts` matches labels with `strings.Contains`, so the label `"7"` matches `INC-77312` and `tarn-staging-91`, and `buried["7"]` is set by unrelated events. The compiled `f.matchers` already exist for exactly this reason; reuse them here.
- `disclaimers` includes `"do not have"` and `"unable to"`, which are broad enough to catch a correct answer. "I am unable to browse, but the region is europe-west4" scores as a miss.
- `truncate` slices bytes and can split a multi-byte rune. Verbose output only, so cosmetic.

## On the cost

Roughly doubling summarizer calls by default is the substantive decision here, and the description is upfront about it. Worth being explicit about the second-order effect: tail retention is documented as the strategy that puts a ceiling on prompt size, and "never drop a fact" weakens that ceiling rather than only making compaction fire more often. The new "What bounding costs" section says as much, so this reads as a deliberate call rather than an oversight.

baptmont:
Thanks — point 1 was right and it was the important one. All four addressed at `9be6963968c11c2939d998b07a59165fc956dd7f`.

## 1. The comparison arm is now the real prior default

You were correct: the arm was a hand-shortened prompt, not the previous one. It
dropped the `CRITICAL INSTRUCTIONS:` framing and instructions 1 and 2, and
reworded the closing sentence. The harness therefore could not reproduce the one
thing it existed to demonstrate.

It is now the pre-change default reproduced verbatim, renamed `prior`, with a
comment saying why it has to stay byte-identical. **Re-measured, and the result
is stronger than the wrong prompt gave**, on `gemini-3.5-flash`, same config:

| arm | runs | recall | lost everything |
|---|---|---|---|
| `default` | 3 | 24/24 | 0 |
| `prior` | 2 | 0/16 | 2 |

The shortened prompt had actually *understated* the effect — it scored 0/8 and
2/8, where the true prior prompt loses everything both times. The table in the
description is updated. The out-of-repo numbers were unaffected: that experiment
drove the real default off `main`, which is why they disagreed.

## 2. Instruction 3 covers only user-stated values

Deliberate, and now said so at the code rather than left to inference. The
measurement behind this change covers user-stated values; tool output is already
the bulkiest part of a transcript, so instructing the model to carry it forward
verbatim would enlarge every later summary by an amount nothing here has
measured. Widening it wants its own change and its own numbers, and I would
rather not ship an unmeasured default alongside a measured one.

## 3. Persistence of injected content

Named in the comment. Content arriving in a tool response used to decay out of a
rolling summary within a few passes and now has a prompt arguing for its
retention; it still cannot forge a turn, since `formatEvents` escapes newlines,
but it can persist longer than it did. Acknowledged as a trade rather than
argued away.

## 4. The three harness bugs

All fixed. `buriedFacts` now uses the same compiled word-boundary matchers as
scoring — you were right that `"7"` was matching inside `INC-77312` and
`tarn-staging-91`. The disclaimer list is narrowed back to unambiguous denials:
`"unable to"` and `"do not have"` were mine from a previous round and, as you
say, would score a correct answer as a miss, which is the direction that
flatters the change. `truncate` counts runes.

One more, found while re-running: a single 503 aborted the whole batch, discarding
runs that had already succeeded. A run is hundreds of calls and this model returns
503 regularly, so a failed run is now reported and skipped and the totals count
only completed runs.

## On the cost

Agreed, and worth stating plainly: "never drop a fact" weakens the ceiling rather
than only making compaction fire more often, because the incompressible part of
the summary grows with the number of distinct facts. Facts accumulate slowly and
narrative grows every turn, so it is the better of the two things to spend the
budget on, but it is a weaker guarantee than before and the doc says so.

## Not Go-specific

While verifying: adk-python ships a byte-identical default prompt and the same
rolling-summary seed. Running this experiment against it, the shipped prompt
scored 0/8, 6/8, 0/8 and the instruction from this PR ported across scored 8/8
four times out of four. Reporting upstream separately — nothing to do here.

karolpiotrowicz (approved):
Approving. The library change is sound and I do not want it held up further — the remaining items are all follow-up material.

Independently at [15197d2c](https://github.com/google/adk-go/commit/15197d2c518f07ee77730dec9c564baa4afec1e6): build, `go vet`, `staticcheck`, `golangci-lint`, `go mod tidy -diff` and the full suite are clean in both modules, and they stay clean on the merge result with current `main`. `apidiff` shows no exported surface moved. Both guard tests fail on the merge base and pass on the head, so they do discriminate on the prompt reaching the model. The cassette genuinely carries the new instruction.

I agree with everything in the review above and am not restating it. The `terse` arm remains the most important thing to fix, and I read it the same way — the harness cannot currently reproduce the result it was added to make reproducible.

Below are five things that review did not cover, ranked by what I would fix first. None of them block this PR.

## 1. The guard tests pin less than the description claims

The testing plan says deleting instruction 3 fails the first guard and making `NewLLMSummarizer` ignore the caller's template fails the second. Both are true. But two edits that break the stated contract leave the whole package suite green.

Rewriting instruction 3 to say the opposite, while keeping the four asserted words, passes:

```go
`3. A section titled "Durable facts" is OPTIONAL. You need not copy ` +
"values verbatim, and a durable fact may be generalized to its label " +
"rather than carried forward when space is short. Prefer narrative. " +
```

```
$ go test ./session/compaction/
ok  	google.golang.org/adk/v2/session/compaction	0.138s
```

Separately, appending a sentence at [llm_summarizer.go:200](https://github.com/google/adk-go/blob/15197d2c518f07ee77730dec9c564baa4afec1e6/session/compaction/llm_summarizer.go#L200) — which breaks the guarantee that a caller's own template is used as given — also passes. `TestCustomPromptTemplateIsUsedVerbatim` checks a prefix and the absence of `Durable facts`, so it catches replacement of the caller's template but not augmentation of it. Comparing against the full expected string would close both.

## 2. Nothing tests tail retention end to end

`TestCompactionE2E` sets `CompactionInterval: 2` with no `EventRetentionSize`, which selects the sliding window — the strategy that never re-summarizes. Tail retention is covered at unit level, but no end-to-end test drives the re-summarization path this change exists for, which is why that green says less about this PR than it looks like it does.

## 3. "Never drop a fact" meets a hard cap nobody documents

A rolling summary above `MaxToolContentChars` is truncated when the next pass re-ingests it, at [llm_summarizer.go:308](https://github.com/google/adk-go/blob/15197d2c518f07ee77730dec9c564baa4afec1e6/session/compaction/llm_summarizer.go#L308), with the default of 2000 at [llm_summarizer.go:71](https://github.com/google/adk-go/blob/15197d2c518f07ee77730dec9c564baa4afec1e6/session/compaction/llm_summarizer.go#L71). A 2151-character prior summary comes back as `... [truncated 151 chars]`, and everything past the cut is gone by exactly the ratchet argument the new doc section makes.

Both the cap and the truncation call are byte-identical on `main`, so this is not something this PR introduces. It is worth a sentence in "What bounding costs" because instruction 3 deliberately enlarges summaries and therefore makes the ceiling easier to reach, while the new text asserts values are carried forward with no ceiling mentioned. I have not shown a real session producing a summary that large, so the practical reachability is open.

## 4. The shipped godoc claims more than the measurement does

[llm_summarizer.go:46-48](https://github.com/google/adk-go/blob/15197d2c518f07ee77730dec9c564baa4afec1e6/session/compaction/llm_summarizer.go#L46-L48) says "with it, none of nine lost any" and [compaction.go:54](https://github.com/google/adk-go/blob/15197d2c518f07ee77730dec9c564baa4afec1e6/session/compaction/compaction.go#L54) says "the default lost none in nine". The table reports 70/72 for that arm, so two probes were lost. Against the parallel "lost every one of them", this reads as a claim that no facts were lost. It is true of the wipeout column, so "lost none of nine conversations entirely (70/72 probes)" is both accurate and a stronger sentence. This one ships to readers, unlike the harness items.

## 5. Two more ways the harness arms differ

- The reported score counts hits over all eight facts at [main.go:259](https://github.com/google/adk-go/blob/15197d2c518f07ee77730dec9c564baa4afec1e6/examples/compactionrecall/main.go#L259), while [main.go:423-425](https://github.com/google/adk-go/blob/15197d2c518f07ee77730dec9c564baa4afec1e6/examples/compactionrecall/main.go#L423-L425) says only the buried ones say anything about what compaction cost. `lostBuried` is computed and never scored. The direction is conservative — unburied facts add free hits to both arms and narrow the gap — so it cannot inflate the result.
- The `terse` arm sets `cfg.Summarizer`, so `resolveCompactionConfig` returns early and that summarizer runs with no timeout, while the `default` arm gets the runner's 60s. Worth matching alongside the prompt-text fix, since it is the same class of asymmetry.

baptmont:
Follow-up opened as #1434, covering the items from this review that are code rather than commentary.

The truncation point turned out to be more than a doc gap. `MaxToolContentChars` was being applied to the rolling summary itself, because tail retention re-ingests it as an ordinary model turn — so the cap that exists for one tool call's output was trimming the summary that carries the facts, permanently. The measurement data from this PR shows it is reachable: summaries reached **4,041 characters** against the 2,000 default. Recall still scored 8/8 only because the durable-facts section sits near the top, so the trim took narrative. #1434 exempts a prior summary from the per-item cap.

Also in #1434: both guard tests now compare the text they are named for. The inversion shown here — `"Durable facts" is OPTIONAL … may be generalized` — did pass, and now fails, as does appending to a caller's template. The godoc claim is corrected to "no conversation lost everything (70/72 probes)". The two harness asymmetries are fixed.

Not in #1434, deliberately: the missing end-to-end test for tail retention needs a new cassette, so it wants its own PR.

PR #1434, opened 2026-08-31
fix(compaction): stop the tool-content cap truncating the rolling summary

baptmont:
Follow-up to #1431 (merged as `cf61116c`). Everything here comes from review
comments on that PR that arrived after it was approved, plus one thing those
comments implied but could not show.

Tracking issue: #298.

## The rolling summary was being truncated

#1431 added an instruction telling the summarizer to carry concrete values
forward verbatim. Tail retention then feeds the resulting summary back into the
next pass as an ordinary model turn ([tail_retention.go seeds it with
`LLMResponse.Content`](https://github.com/google/adk-go/blob/main/internal/compactioninternal/tail_retention.go)),
and `formatEvents` applied `MaxToolContentChars` to it. Anything past the cap
was cut, and the cut was permanent: the next pass sees only what was rendered.

That is the exact ratchet #1431 exists to prevent, aimed at the summary carrying
the facts it was protecting.

**It is reachable, not theoretical.** The review raised it as a doc gap because
no session had been shown producing a summary that large. The measurement data
from #1431 does: fact-carrying summaries reached **4,041 characters** against the
2,000-character default. Verified directly — a summary of 1,999 characters passes
through, 2,151 and 4,041 are truncated.

Recall still scored 8/8 in those runs only because the durable-facts section
tends to sit near the top of the summary, so the trim took narrative. That is
luck, and it is not something to ship a guarantee on.

A prior summary is now exempt from the per-item cap. `MaxTranscriptChars` still
bounds the transcript as a whole, which is the right backstop for text the
framework generated; the per-item cap is for content it did not author.

## The guard tests pinned less than they claimed

Both guards from #1431 asserted on keywords. Reviewers showed two edits that
break the stated contract and leave the package green:

- Rewriting instruction 3 to say a durable fact **"is OPTIONAL"**, that values
  need not be copied verbatim and **"may be generalized"** — every asserted
  keyword survives, so the test passed while the prompt instructed the opposite
  of its own name. Reproduced before fixing.
- Appending a sentence to a caller's template. `TestCustomPromptTemplateIsUsedVerbatim`
  checked a prefix and an absence, so it caught replacement but not augmentation.

Both now compare the text they are named for. Mutation-tested, all three fail:
inverting instruction 3, augmenting a caller's template, and re-applying the cap
to a prior summary.

The previous round's mutation testing only deleted things, which is why it missed
an inversion. Deleting instruction 3 did fail the test; contradicting it did not.

## The godoc overstated the measurement

Two places said the new prompt "lost none" across nine conversations, against a
table reporting 70/72 probes. Next to "lost every one of them", that reads as no
facts lost at all. The claim that holds is that no conversation lost
*everything*, which is also the stronger sentence. Corrected in
`llm_summarizer.go` and the package doc.

## Two harness arms differed in more than the prompt

- The `prior` arm supplied its own `Summarizer` and so opted out of the 60s
  timeout the runner installs when none is named. Now matched.
- The score counted all eight facts while the code comment says only buried ones
  say anything about compaction. `lostBuried` was computed and never reported.
  Buried recall is now printed alongside. The direction was conservative —
  unburied facts add free hits to both arms and narrow the gap — so no previous
  number was inflated.

## Testing plan

`go build`, `go test -race -count=1 -shuffle=on`, `golangci-lint run` (v2.3.1)
and `go mod tidy -diff` clean in both modules.

`TestPriorSummaryIsNotTruncatedByToolContentCap` is new and covers both
directions: a prior summary survives the cap intact, and an ordinary oversized
text part is still truncated, so the exemption cannot become a licence for any
large part to inflate the prompt.

No cassette change. `TestCompactionE2E` drives the sliding window with short
summaries, so no prompt bytes move.

## Not addressed here

**No end-to-end test drives tail retention.** `TestCompactionE2E` sets
`CompactionInterval: 2`, which selects the sliding window — the strategy that
never re-summarizes. The re-summarization path this whole line of work is about
is covered only at unit level. That needs a new recorded cassette, so it wants
its own PR rather than riding along here.

**adk-python has the same defect.** It ships a byte-identical default prompt and
the same rolling-summary seed. Measured against it: the shipped prompt scored
0/8, 6/8, 0/8 over three runs, and the instruction from #1431 ported across
scored 8/8 four times out of four. Reporting upstream separately.

## Also here

- **A README for `examples/compactionrecall`**, following the shape the workflow
  samples use: concept, whether it needs a model, goal, flags, how to read the
  output, and a real session.
- **Guidance on choosing `TokenThreshold`**, and a qualification of the cost
  figure. "About twice as many summarizer calls" was measured at a threshold
  small enough to re-trigger compaction every turn, so it is an upper bound
  rather than a typical cost, and the doc now says so.
- **An incidental fact style in the harness, now the default.** The old style
  stated each fact alone followed by "Note it" and planted all eight up front,
  which is close to measuring the shipped instruction's own cue. Stating each
  value inside a message about something else, with no signal, spread through
  the conversation: the shipped prompt recalled 32/32 probes over four runs and
  the prior prompt 5/32, losing everything in three of four. The gap widens.

**Review this exact commit: `3a56a44671334d45578447eeece285958de69b28`.** Check out the SHA, not a branch name.
Cut from `main` at `cf61116c` (the merge of #1431) and a single commit.

### DIFF
--- a/examples/compactionrecall/README.md
+++ b/examples/compactionrecall/README.md
@@ -1,0 +1,151 @@
+# Compaction recall
+
+Measures whether a compaction config still lets the agent **answer**, rather
+than how small it makes the prompt. Plants checkable facts, buries them under
+filler turns until compaction has summarized them away, then asks about each one
+and scores whether the value came back.
+
+- **Concept:** Judge a `compaction.Config` on recall, not on prompt size, by
+  probing for facts that are known to sit behind a compaction boundary.
+- **Needs LLM?** Yes (Gemini)
+
+Related: [`../compaction`](../compaction) — how to *enable* compaction on an
+agent. This one is about what enabling it costs you.
+
+## Goal
+
+Every other measurement of compaction reports prompt size, which is easy to
+measure and says nothing about whether the conversation survived. Tail retention
+seeds each new summary with the previous one, so a value stated early is
+recompressed on every pass. Whether it survives is a property of the summarizer
+prompt, and the only way to know is to ask the agent afterwards.
+
+Point it at your own settings to find out what they cost you:
+
+```bash
+GOOGLE_API_KEY=... go run ./examples/compactionrecall -threshold=32000 -retention=10
+```
+
+Two arms run the identical conversation, differing in one instruction and
+nothing else, so the comparison isolates that instruction rather than measuring
+a third prompt:
+
+| arm | summarizer prompt |
+| --- | --- |
+| `default` | the shipped prompt, which requires concrete values to be carried forward verbatim |
+| `prior` | the same prompt as it was before that instruction was added |
+
+If you are writing your own `PromptTemplate`, add it as an arm and run it here
+before trusting it. Asking only for a "concise" summary is enough to reintroduce
+the loss.
+
+## Authentication
+
+The model client reads an API key from the environment:
+
+```bash
+export GOOGLE_API_KEY=...
+```
+
+Vertex AI is not wired up here; pass `-model` to change the model.
+
+## Reading the output
+
+**Read the per-run scores, not the average.** Recall is close to
+all-or-nothing: each pass either copies a value forward or generalizes it to its
+label — "the deployment region" in place of `europe-west4` — and a value replaced
+by its label cannot come back, because the next pass sees only the previous
+summary and the retained tail. A config tends to score 8/8 or 0/8, so two configs
+that both average 50% can mean "always half-remembers" and "works perfectly half
+the time". That is why `-runs` defaults to 3 and each run is printed.
+
+`buried` counts the facts actually behind a compaction boundary at probe time. A
+fact still sitting in the raw tail was never summarized, so recalling it measures
+nothing; only the buried ones say anything about what compaction cost.
+
+## Flags
+
+| Flag | Default | Purpose |
+| --- | --- | --- |
+| `-model` | `gemini-flash-latest` | Model for both the conversation and the summarizer. |
+| `-threshold` | `700` | `compaction.Config.TokenThreshold`. |
+| `-retention` | `3` | `compaction.Config.EventRetentionSize`. |
+| `-interval` | `0` | `compaction.Config.CompactionInterval`; 0 leaves the sliding window off. |
+| `-fillerx` | `3` | Repeat the filler block this many times, to force more passes. |
+| `-runs` | `3` | Repeat each arm this many times. |
+| `-arms` | `default,prior` | Which arms to run. |
+| `-style` | `incidental` | How facts are stated. See below. |
+| `-v` | off | Print every summary and every probe answer. |
+
+### `-style`
+
+`incidental` states each value inside a message whose actual point is something
+else, with no cue that it matters, and spreads them through the conversation so
+no single compaction window sees them all:
+
+> Quick bit of context before I forget: we finally got the new environment
+> provisioned in europe-west4, so the Frankfurt team should see decent latency.
+> Anyway, explain in two sentences why connection pooling matters.
+
+`flagged` states each value on its own and follows it with "Note it", and plants
+them all up front. That is an easier test, and a fairer one only if your users
+really do announce what matters.
+
+It defaults to `incidental` because the easier shape flatters any prompt that
+names the categories it should keep, which is exactly what the shipped prompt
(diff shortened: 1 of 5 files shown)


baptmont:
### Round update: README, threshold guidance, and a harder measurement

Three commits added since the last round; **review `0b5ba4be`**.

**A README for `examples/compactionrecall`**, which had none. Follows the shape the workflow samples use — concept line, `Needs LLM?`, goal, authentication, flags, how to read the output, and a real session.

**Guidance on choosing `TokenThreshold`**, plus a correction. The threshold has to be a multiple of the summary size, not comparable to it: a summary larger than the threshold puts the prompt back over the line the moment compaction finishes, so every turn compacts again and none of those calls can succeed. Measured summaries run 250–1,000 tokens. That also qualifies the cost figure this PR's parent shipped — "about twice as many summarizer calls" was measured at a threshold of 700, i.e. inside that pathology, so it is an upper bound rather than a typical cost.

**The harness now states facts incidentally by default.** Worth flagging as a change of measurement, not just of code. The old style stated each fact alone and followed it with "Note it", and planted all eight before any filler — and the instruction under test asks the summarizer to keep identifiers, names, dates and numbers. A benchmark that announces exactly those categories and front-loads them is close to measuring its own cue.

Restated incidentally — each value inside a message whose point is something else, no signal, spread so no single window sees them all — over four runs per arm:

| arm | probes | lost everything |
|---|---|---|
| `default` | 32/32 | 0 of 4 |
| `prior` | 5/32 | 3 of 4 |

The gap widens rather than narrowing. One detail is better evidence than the totals: in the single `prior` run that partly survived, the three lost values are the three stated *earliest* — the ones that had been through the most passes.

The re-arming pathology those runs sit in is fixed separately in #1439.

baptmont:
### Correction: compactions per turn

A reader asked how the incidental style plants its facts, and checking it turned up an arithmetic error I had propagated into the package doc, the README and #1439.

The ratio was computed against the wrong turn count. The incidental style runs a pass of interleaved fact and filler turns *before* the main filler block, so at `-fillerx=1` a run is **40 turns**, not the 32 the flagged style produces. So 35–38 compactions is **nearly one per turn**, not more than one.

The claim it supported is unchanged — a threshold of 700 sits below the summary size, so compaction re-triggers on almost every turn and none of those calls can get under the line — but it was overstated in the direction that made it sound worse than measured, and it is the number quoted as evidence for how to choose a threshold. Corrected here in `compaction.go` and the README, and in #1439's description.

The README also now warns against comparing compaction counts across the two styles, since they differ in length for the same reason.

To answer the question itself, for anyone reviewing the harness: the two styles are mutually exclusive. `incidental` sends only the `incidental` phrasing and `flagged` sends only `tell`; no fact is stated twice and no "Note it" appears in an incidental run.

Head is now `9cd2ef7d`.

baptmont:
### Self-review found a regression in this PR, now fixed

Reviewing my own diff while waiting for a reviewer: **the exemption was too broad and introduced a way to stop compaction permanently.**

Exempting a prior summary from the per-item cap also exempted it from the second pass in `renderTranscript`, which derives a tighter cap from the transcript budget and re-renders. That pass is the only thing between an oversized window and a refusal — and because selection is size-capped, the same window returns every turn, so a window that can never be rendered stops compaction for the life of the session.

Reproduced: a 6,000-character summary against a 5,000-character `MaxTranscriptChars` was **refused outright**, where on `main` it is trimmed and accepted. I introduced that.

The exemption belongs to the tool-content limit, not to the budget. `formatEvents` now takes the summary cap separately: the first pass passes `-1` so a prior summary escapes the per-item limit, the budget pass passes its derived cap for both so nothing is immune to it.

That pass also had to stop being conditional on the derived cap being *tighter* than the per-item cap. "Already tighter" implied "cannot help" only while every part was capped in the first pass, which a prior summary no longer is: 6,000 characters against a 5,000 budget derives about 2,500, which is not below the 2,000 default, so the guarded version skipped the pass and refused a window it could have shrunk. The result is adopted only if smaller, so dropping the guard costs one wasted render on a path that is already failing.

`TestOversizedPriorSummaryIsStillShrunkToFitTheTranscript` pins it. Three mutations fail: making the summary immune to the budget pass, restoring the "only if tighter" guard, and re-applying the tool cap to summaries.

Head is now `3a56a446`.

PR #1436, opened 2026-08-31
internal: require explicit ack for -httprecord directives that claim a whole package

tvost2:

- Addresses (not `Closes:`, per the issue's own reopening comment — see below): #1330

**Problem:**

#1330 was reopened after #1362 fixed the double-recording failure mode (two directives claiming the same cassette). The reopening comment (@baptmont) pinned a second, still-silent failure mode: `TestHTTPRecordDirectivesPartitionCassettes` enforces *partition*, not *narrowness* — a single `-httprecord` directive matching every cassette in its package passes the guard cleanly. Five packages still had this at the time:

| --- | --- | --- |

(`tool/functiontool` has only 1 cassette, so its `-httprecord=.*` isn't a partition problem — nothing to partition.)

`go generate ./tool/mcptoolset/...` still re-records all 11 against a live model whenever the intent was to refresh one.

**Solution:**

The issue itself lists three options and explicitly leaves the choice open. This implements the second: *"Teach the guard to flag a directive that claims every cassette in a package with more than one recorded test, so the pattern is at least forced to be deliberate."*

`TestHTTPRecordDirectivesPartitionCassettes` now also fails when a single directive claims every cassette in a package with more than one, **unless** a `// httprecord:whole-package` comment sits in the contiguous `//`-comment block directly above the directive. This doesn't narrow or otherwise change any of the five directives — per the issue's own reasoning, a package with one recording directive has nothing to partition — it only makes that breadth a visible, deliberate choice in the diff that introduces or keeps it, rather than a silent default discoverable only by reading this issue thread. All five directives now carry that comment, each explaining why (single/only recording directive in the package). `AGENTS.md` documents the convention alongside the existing `-httprecord` guidance.

I did not attempt option 1 (closing the issue outright) since that's a maintainer call about intent, not a mechanical fix, and not option 3 (removing the `-httprecord` regexp class of bug entirely) since the issue itself calls that out as by far the largest change.

### Testing Plan

- `go test ./internal/ -run TestHTTPRecordDirectivesPartitionCassettes -v`: fails on all four multi-cassette directives with the ack comment removed (confirmed locally by temporarily deleting each), passes with all five in place (`checked 8 -httprecord directives across 6 packages`).
- `gofmt -l` and `go vet ./internal/... ./agent/remoteagent/v2/... ./model/gemini/... ./plugin/functioncallmodifier/... ./tool/mcptoolset/...`: clean.
- No behavioral change to any `-httprecord` pattern or recorded cassette — this PR only adds a test-time check and doc comments.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -162,7 +162,11 @@ See [Multi-Module Development](CONTRIBUTING.md#multi-module-development) in
 - To re-record a whole package, run `go generate ./<pkg>/...`; each package's
   `//go:generate go test -httprecord=…` directives are scoped so that every
   cassette is recorded exactly once (enforced by
-  `TestHTTPRecordDirectivesPartitionCassettes` in `internal`).
+  `TestHTTPRecordDirectivesPartitionCassettes` in `internal`). If a directive
+  intentionally claims every cassette in a package that has more than one
+  (nothing to partition, e.g. one recording directive per package), that same
+  test requires a `// httprecord:whole-package` comment directly above it —
+  add one rather than narrowing the pattern.
 - Prefer table-driven tests; shared helpers live in `internal/testutil`.
 
 ## Boundaries
--- a/agent/remoteagent/v2/a2a_e2e_test.go
+++ b/agent/remoteagent/v2/a2a_e2e_test.go
@@ -14,6 +14,8 @@
 
 package remoteagent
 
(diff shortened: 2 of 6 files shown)


tvost2:
Closing to re-open from a cleaned-up branch (commit metadata); same change, will link the new PR here.

tvost2:
Reopened as #1437 (identical change, cleaned-up commit metadata).

PR #14, opened 2025-06-16
internal/httprr: copy golang.org/x/oscar/internal/httprr to internal/httprr

hyangah:
This provides the utility to record and replay HTTP (while remoing secrets from the log).

Use it to make model's TestNewGeminiModel test deterministic.

PR #1437, opened 2026-08-31
internal: require explicit ack for -httprecord directives that claim a whole package

tvost2:

- Addresses (not `Closes:`, per the issue's own reopening comment — see below): #1330

**Problem:**

#1330 was reopened after #1362 fixed the double-recording failure mode (two directives claiming the same cassette). The reopening comment (@baptmont) pinned a second, still-silent failure mode: `TestHTTPRecordDirectivesPartitionCassettes` enforces *partition*, not *narrowness* — a single `-httprecord` directive matching every cassette in its package passes the guard cleanly. Five packages still had this at the time:

| --- | --- | --- |

(`tool/functiontool` has only 1 cassette, so its `-httprecord=.*` isn't a partition problem — nothing to partition.)

`go generate ./tool/mcptoolset/...` still re-records all 11 against a live model whenever the intent was to refresh one.

**Solution:**

The issue itself lists three options and explicitly leaves the choice open. This implements the second: *"Teach the guard to flag a directive that claims every cassette in a package with more than one recorded test, so the pattern is at least forced to be deliberate."*

`TestHTTPRecordDirectivesPartitionCassettes` now also fails when a single directive claims every cassette in a package with more than one, **unless** a `// httprecord:whole-package` comment sits in the contiguous `//`-comment block directly above the directive. This doesn't narrow or otherwise change any of the five directives — per the issue's own reasoning, a package with one recording directive has nothing to partition — it only makes that breadth a visible, deliberate choice in the diff that introduces or keeps it, rather than a silent default discoverable only by reading this issue thread. All five directives now carry that comment, each explaining why (single/only recording directive in the package). `AGENTS.md` documents the convention alongside the existing `-httprecord` guidance.

I did not attempt option 1 (closing the issue outright) since that's a maintainer call about intent, not a mechanical fix, and not option 3 (removing the `-httprecord` regexp class of bug entirely) since the issue itself calls that out as by far the largest change.

### Testing Plan

- `go test ./internal/ -run TestHTTPRecordDirectivesPartitionCassettes -v`: fails on all four multi-cassette directives with the ack comment removed (confirmed locally by temporarily deleting each), passes with all five in place (`checked 8 -httprecord directives across 6 packages`).
- `gofmt -l` and `go vet ./internal/... ./agent/remoteagent/v2/... ./model/gemini/... ./plugin/functioncallmodifier/... ./tool/mcptoolset/...`: clean.
- No behavioral change to any `-httprecord` pattern or recorded cassette — this PR only adds a test-time check and doc comments.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -162,7 +162,13 @@ See [Multi-Module Development](CONTRIBUTING.md#multi-module-development) in
 - To re-record a whole package, run `go generate ./<pkg>/...`; each package's
   `//go:generate go test -httprecord=…` directives are scoped so that every
   cassette is recorded exactly once (enforced by
-  `TestHTTPRecordDirectivesPartitionCassettes` in `internal`).
+  `TestHTTPRecordDirectivesPartitionCassettes` in `internal`). Give each
+  recording test function its own directive, anchored to its cassette names,
+  so refreshing one function's recordings never rewrites another's. Only when
+  a package's cassettes all come from a single test function — so there is no
+  partition to make — acknowledge the wildcard with a
+  `// httprecord:whole-package` comment directly above the directive that says
+  why; the same test rejects a bare package-wide directive without it.
 - Prefer table-driven tests; shared helpers live in `internal/testutil`.
 
 ## Boundaries
--- a/agent/remoteagent/v2/a2a_e2e_test.go
+++ b/agent/remoteagent/v2/a2a_e2e_test.go
@@ -14,7 +14,11 @@
 
 package remoteagent
 
-//go:generate go test -httprecord=.*
+// One directive per recording test function, so refreshing one function's
+// cassettes against a live model never rewrites another's (see
+// TestHTTPRecordDirectivesPartitionCassettes in internal).
+//go:generate go test -httprecord=^testdata[/\\]TestA2ARemoteAgentStreamingGemini(Success|Error)\.httprr$
+//go:generate go test -httprecord=^testdata[/\\]TestA2ASingleHopFinalResponse_.*\.httprr$
 
 import (
 	"context"
--- a/internal/httprr_directives_test.go
+++ b/internal/httprr_directives_test.go
@@ -22,6 +22,12 @@ import (
 	"testing"
 )
 
+// wholePackageAck is the marker a directive uses to acknowledge that it
+// deliberately claims every cassette in a multi-cassette package. It is
+// recognized only as the leading token of a `//` comment line above the
+// directive (see ackedAbove).
+const wholePackageAck = "httprecord:whole-package"
+
 // TestHTTPRecordDirectivesPartitionCassettes checks that in every package which
 // declares at least one `//go:generate go test -httprecord=…` directive, each
 // of the package's testdata/*.httprr cassettes is claimed by exactly one
@@ -29,7 +35,7 @@ import (
 //
 // The -httprecord flag is a regexp matched against the cassette's file path
 // (see httprr.Recording in internal/httprr), not a -run test-name filter. That
-// makes two failure modes possible, and both are silent:
+// makes three failure modes possible, all silent:
 //
 //   - Two directives in one package match the same cassette, so
 //     `go generate ./<pkg>/...` re-records it twice against a live model in a
@@ -37,9 +43,21 @@ import (
 //     ones.
 //   - A directive matches none of its cassettes, so `go generate` appears to
 //     succeed while recording nothing.
+//   - A single directive matches every cassette in a package that has more
+//     than one, so `go generate ./<pkg>/...` re-records the whole package --
+//     against live models whose responses differ run to run -- whenever
+//     someone means to refresh just one. This is sometimes unavoidable: a
+//     package whose cassettes all come from a single test function has no
+//     per-function partition to make. So it is not forbidden; it requires a
+//     `// httprecord:whole-package` comment above the directive stating why,
+//     so the breadth is a deliberate, reviewed choice and not a silent
+//     default (see #1330, reopened after #1362 fixed only the first mode
+//     above: four packages still had an unreviewed wildcard directive then).
 //
-// Requiring exactly one directive per cassette rules out the first, and at
-// least one cassette per directive rules out the second.
+// Requiring exactly one directive per cassette rules out the first failure
+// mode, at least one cassette per directive rules out the second, and the
+// `httprecord:whole-package` acknowledgment rules out the third. The marker
+// detection itself is exercised by TestAckedAbove.
 func TestHTTPRecordDirectivesPartitionCassettes(t *testing.T) {
 	// Start test from the parent directory, root of the module.
 	t.Chdir("..")
@@ -53,9 +71,12 @@ func TestHTTPRecordDirectivesPartitionCassettes(t *testing.T) {
 
 	type directive struct {
 		file    string
+		lineNo  int // 1-based line of the //go:generate directive
+		text    string
 		pattern string
 		re      *regexp.Regexp
-		claims  int // cassettes in the same package this directive matched
+		claims  int  // cassettes in the same package this directive matched
+		acked   bool // comment block above carries the wholePackageAck marker
 	}
 	byPkg := make(map[string][]directive)
 	total := 0
@@ -77,8 +98,9 @@ func TestHTTPRecordDirectivesPartitionCassettes(t *testing.T) {
 		if err != nil {
 			return err
 		}
-		for _, line := range strings.Split(string(content), "\n") {
-			line = strings.TrimSpace(line)
+		lines := strings.Split(string(content), "\n")
+		for i, raw := range lines {
+			line := strings.TrimSpace(raw)
 			if !strings.HasPrefix(line, "//go:generate") || !strings.Contains(line, "-httprecord") {
 				continue
 			}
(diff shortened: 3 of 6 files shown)


karolpiotrowicz (commented):
The guard does what it says. Applied to the tree without the four ack comments it fails on exactly the four multi-cassette directives, and passes with them — so the check is live, not decorative. Build, `go vet`, `golangci-lint`, `gofmt` and `govulncheck` are clean, and the two packages that fail the suite here (`session/vertexai`, `telemetry`) fail identically without this change. The merge into current `main` is clean too.

I probed the marker detection nine ways to see whether the ack could be satisfied by accident, since an opt-in guard that a default quietly satisfies is not a guard. It holds up: no comment at all does not satisfy it, a `/* */` block does not, a blank line between the marker and the directive does not, the marker appended to the `//go:generate` line does not, and an ack does not bleed past a sibling directive. One thing to fix, plus a documentation problem that I think matters more than any of the code points.

## The exemption is defined by its own trigger

The guard fires when one directive claims every cassette in a package that has more than one. The change then describes the exempt case as "a package with one directive per test file" ([`internal/httprr_directives_test.go:42-43`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/internal/httprr_directives_test.go#L42-L43)), "one directive covering one test file" ([`:194-195`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/internal/httprr_directives_test.go#L194-L195)), and "nothing to partition, e.g. one recording directive per package" ([`AGENTS.md:167`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/AGENTS.md#L167)) — closing with "add one rather than narrowing the pattern" ([`AGENTS.md:169`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/AGENTS.md#L169)).

Having one directive is the trigger. Using it as the reason for the exemption means anything that fires the guard qualifies to be waved through by the document's own definition, and `AGENTS.md` is the file coding agents read as instructions. The sentence is grammatically inside the conditional that opens at [`AGENTS.md:165`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/AGENTS.md#L165), so it is not wrong on a careful read — but the conditional is satisfied by exactly the population that reaches it.

The four acks show what that costs. Each says "there is nothing to partition", and for `tool/mcptoolset` that is not true: the package has two recording test functions, [`TestMCPToolSet`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/tool/mcptoolset/set_test.go#L68) with one cassette and [`TestMCPToolSetConfirmation`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/tool/mcptoolset/set_test.go#L372) with ten. I ran the 1 + 10 split along those two names and it passes the guard with no ack needed:

```
//go:generate go test -v -httprecord=^testdata/TestMCPToolSet\.httprr$
//go:generate go test -v -httprecord=^testdata/TestMCPToolSetConfirmation_.*\.httprr$
```

```
--- PASS: TestHTTPRecordDirectivesPartitionCassettes
    checked 9 -httprecord directives across 6 packages
```

So the largest package the change blesses for whole-package re-recording has a partition available, and [the comment](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/tool/mcptoolset/set_test.go#L63-L65) says otherwise. The two doc lines also disagree with each other, and [that first line](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/internal/httprr_directives_test.go#L42-L43) is false on the repo's own example — `agent/llmagent` is one directive per test file and partitions 20 cassettes three ways.

What would fix it: state the exemption as a reason rather than a restatement ("every cassette here comes from one test function"), have `AGENTS.md` prefer narrowing where a partition exists, and either partition `mcptoolset` or say in its comment why the available split was rejected.

## Smaller things

- The marker is matched with [`strings.Contains`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/internal/httprr_directives_test.go#L136) against every line of the comment block above the directive, so a sentence that mentions the marker satisfies it — including one that forbids it. I confirmed `// Do NOT use httprecord:whole-package in this package.` above a wildcard directive is accepted silently. Anchoring the match to the start of the comment text closes this, and all four acks you added already satisfy the stricter form. The failure direction is only ever a false pass, so this is not urgent.

- Nothing would notice the guard rotting. Setting `acked := true` at [`:130`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/internal/httprr_directives_test.go#L130) disables the new check outright and the suite stays green. To be fair this is a property the file already had — the same experiment on the pre-existing partition and dead-pattern branches leaves them green too — so it is a follow-up rather than something you introduced. Lifting the walk into a small pure function and table-testing the nine placements above would cover it cheaply.

- [`lineForAck`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/internal/httprr_directives_test.go#L73) holds the `//go:generate` text, not anything about the ack. The loop index is in scope and unused, so the error cannot name a line.

- The count is four, not five. [`:45`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/internal/httprr_directives_test.go#L45) and the PR description both say five, but `tool/functiontool` has one cassette and is exempt by `len(cassettes) > 1`. The failing run without the acks reports exactly four.

- [`:72`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/internal/httprr_directives_test.go#L72) says the marker sits on "the preceding line", while the walk scans the whole block as [`:124-129`](https://github.com/google/adk-go/blob/11279816b685769d52fb8e9cd89d8729285960a5/internal/httprr_directives_test.go#L124-L129) describes.

## Separately, not this change

Re-recording touches credentials, so I checked the corpus. Every committed recording is clean — a case-insensitive scan across all 48 `.httprr` files plus the 42 `.replay` and 9 `.events.yaml` files for bearer tokens, API keys, JWTs and private keys returns one hit, and it is the `alpha-42` fixture defined in `llmagent_delegation_test.go`.

There is a real gap underneath, though it predates this change and is not reachable today: `scrubGeminiRequest` strips `x-goog-api-key` but not `Authorization`, so a bearer token reaching the recorder would be written into a cassette verbatim. I confirmed that with a sentinel, and it behaves the same with or without this change. No current path attaches one — `genai` skips credential attachment whenever the caller supplies an `HTTPClient`, which every recording path here does. Worth a separate issue rather than anything for you to carry.

karolpiotrowicz:

It adds an acknowledgement convention for `-httprecord` directives that claim a whole package, and writes that convention into `AGENTS.md` as a repo-wide rule. The reason this needs you rather than a reviewer is that #1330 was reopened with three options and the choice was deliberately left open — close it as working-as-intended, adopt this convention, or remove the underlying regexp behavior. This PR picks one of the three, which is a decision the PR cannot make on its own.

tvost2:
@wolo-lab — since this is a scope call, here is the data and a neutral pros/cons for the three options in #1330, so the choice is yours rather than mine.

### Blast radius today: what `go generate ./<pkg>/...` re-records when the intent is to refresh **one** cassette

| package | cassettes | KB | recording test funcs | today (`main`) | if partitioned |
|---|---:|---:|---:|---:|---:|
| `tool/mcptoolset` | 11 | 49 | 2 | **11** | 1 or 10 |
| `agent/remoteagent/v2` | 4 | 18 | 3 | **4** | 2 |
| `model/gemini` | 4 | 5 | 3 | **4** | 1–2 |
| `plugin/functioncallmodifier` | 3 | 34 | 1 (subtests only) | **3** | **3 — irreducible** |
| `tool/functiontool` | 1 | 7 | 1 | 1 | 1 |
| `agent/llmagent` | 20 | 320 | 3+ | already scoped by #1362 | — |

One correction to the issue's reasoning: "a package with one test file has nothing to partition" holds for only 1 of the 5 packages. The other four have multiple *recording test functions*, so a directive per function works and is not one-directive-per-cassette. `plugin/functioncallmodifier` is the genuine single-function case.

### Option 1 — docs only: accept the per-cassette command, make `go generate` explicitly a whole-package operation

- **Pro:** ~zero code; #1362 already did most of the AGENTS.md half. Nothing to maintain. No divergence from anything.
- **Con:** blast radius unchanged — `go generate ./tool/mcptoolset/...` still re-records 11. The `//go:generate` directive in the file stays `.*`, so the documented/default path (what `go generate` and most contributor tooling run) is still the wrong one; the safe path is a command the file doesn't carry.

### Option 2 — guard flags a whole-package directive; `// httprecord:whole-package` ack required (this PR)

- **Pro:** blast radius drops to the partitioned column above (11→1 worst case). The `//go:generate` path itself becomes correct — no secret command. No production code, no new flag/API; `internal/httprr/rr.go` is untouched, so no divergence from `x/oscar`. Guard is self-tested.
- **Con:** `plugin/functioncallmodifier` stays at 3 (structural). A future whole-package directive can still be waved through with a low-effort ack comment — it's *visible in the diff*, not *prevented*. Adds a repo-specific convention (currently written into AGENTS.md by this PR — see note below).

### Option 3 — httprr scopes recording to the running test; `-httprecord` stops being a path regexp

- **Pro:** blast radius is 1 by construction, always. Closes the class rather than policing it. Makes the flag behave the way people already assume. Cassette filenames are already `t.Name()`-derived in every test, so the test→file mapping exists.
- **Con:** `internal/httprr/rr.go` is a verbatim copy of `golang.org/x/oscar/internal/httprr` (#14, one typo-fix since); this forks it. ~15 `httprr.Open` call sites need `t` threaded through (or a new `OpenForTest`). Largest review surface, and needs a UX decision (record what `-run` selects? require `-run`?). Hardest to walk back.

### Note on this PR's scope

The partitioned directives and the guard are option 2. This PR also writes the convention into `AGENTS.md` as a repo-wide rule, which presumes option 2 is the chosen direction. If the direction is still open, I'm happy to drop the AGENTS.md change and keep it as a proposal in the PR body, or close this in favour of option 1 or 3 — the directive partitioning is reusable under any of them.

PR #1438, opened 2026-08-31
test(llmagent): drive tail retention end to end against a recorded model

baptmont:
Closes the coverage gap raised in review of #1431 (merged as `cf61116c`). Sits
alongside #1434, which handles the other review items; the two are independent
and can land in either order.

Tracking issue: #298.

## The gap

`TestCompactionE2E` sets `CompactionInterval: 2`, which selects the **sliding
window** — the strategy that summarizes each group of turns once and never
revisits them. Tail retention is the other trigger, and it is the one with a
failure mode of its own: it seeds every new summary with the previous one, so a
value has to survive being recompressed on every pass.

That path had unit coverage and nothing above it. It is why a default summarizer
prompt that lost the early conversation could ship without a single test going
red, and why the green suite on #1431 said less about that change than it looked
like it did.

## What the test asserts

Three things, in order of what would otherwise go unnoticed:

1. **A later summarization was handed an earlier summary.** This is the seed
   path and the entire difference from the sliding window.
2. **Records supersede rather than accumulate.** Each covers a range starting
   where the first one did, so exactly one summary materializes into a prompt
   however many passes have run.
3. **The prompt stays bounded** while the conversation grows, which is the
   property tail retention is chosen for.

## Two things worth knowing about how it is written

**The summarizer's model is wrapped.** The summarizer holds its own model and
does not go through the agent, so no `BeforeModelCallback` ever sees a
summarization call. Wrapping is the only way to observe what it was actually
asked to summarize.

**The seed match is on one line, not the whole summary.** `formatEvents` escapes
newlines so a tool response cannot forge a turn, so a multi-line summary never
appears in a prompt in the form it was stored. The test matches the longest line
and fails if no line is distinctive enough to identify.

## What it deliberately does not assert

That any particular fact survived. Recall is a property of the model and the
prompt, it varies run to run, and pinning it to a single recording would make a
green cassette look like a guarantee the framework cannot give.
`examples/compactionrecall` is where that gets measured, over repeated runs.

## The threshold is deliberately low

`TokenThreshold: 400`, well below the point where compaction merely starts
firing. At 700 the recording produced two records once and one another time,
depending on how verbose the model happened to be — which would make a re-record
a coin flip on whether the test exercised anything. Found the hard way: the
first two recording attempts failed, one on each of those.

## Testing plan

`go build`, `go test -race -count=1 -shuffle=on`, `golangci-lint run` (v2.3.1)
and `go mod tidy -diff` clean in both modules. Replays offline with no
credentials set.

**Mutation-tested.** Removing the rolling seed in
`internal/compactioninternal/tail_retention.go` (`return append([]*session.Event{seed}, window...)`
to `return window`) fails the test. As with the existing compaction e2e test, a
change to the seed moves prompt bytes, so the failure surfaces as a replay miss
rather than as the seed assertion — the same property that file already
documents.

All three assertions were also seen failing for real during development: the
seed match on the first recording attempt, the two-record guard on the second,
and the mutation above.

New cassette `testdata/TestTailRetentionE2E.httprr` (97 KB), claimed by its own
`//go:generate` directive; `TestHTTPRecordDirectivesPartitionCassettes` passes,
so no other directive can rewrite it. No credentials or auth headers in the
recorded file.

**Review this exact commit: `191ee65a29ba15dd14c9f51c7c5d9691bf6b2488`.** Check out the SHA, not a branch name.
Cut from `main` at `cf61116c` and a single commit adding one test file and one
cassette. No non-test code changes.

### DIFF
--- a/agent/llmagent/llmagent_tail_retention_test.go
+++ b/agent/llmagent/llmagent_tail_retention_test.go
@@ -1,0 +1,275 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package llmagent_test
+
+import (
+	"context"
+	"iter"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+
+	"google.golang.org/genai"
+
+	"google.golang.org/adk/v2/agent"
+	"google.golang.org/adk/v2/agent/llmagent"
+	"google.golang.org/adk/v2/internal/compactioninternal"
+	"google.golang.org/adk/v2/internal/httprr"
+	"google.golang.org/adk/v2/internal/testutil"
+	"google.golang.org/adk/v2/model"
+	"google.golang.org/adk/v2/session"
+	"google.golang.org/adk/v2/session/compaction"
+)
+
+// recordingModel forwards to a real model and keeps the prompts it was sent.
+//
+// The summarizer holds its own model rather than the agent's, so an agent
+// BeforeModelCallback never sees a summarization call. Wrapping is the only way
+// to observe what the summarizer was actually asked to summarize, which is the
+// one thing this test is about.
+type recordingModel struct {
+	inner model.LLM
+
+	mu      sync.Mutex
+	prompts []string
+}
+
+func (m *recordingModel) Name() string { return m.inner.Name() }
+
+func (m *recordingModel) GenerateContent(ctx context.Context, req *model.LLMRequest, stream bool) iter.Seq2[*model.LLMResponse, error] {
+	var sb strings.Builder
+	for _, c := range req.Contents {
+		if c == nil {
+			continue
+		}
+		for _, p := range c.Parts {
+			if p != nil && p.Text != "" {
+				sb.WriteString(p.Text)
+			}
+		}
+	}
+	m.mu.Lock()
+	m.prompts = append(m.prompts, sb.String())
+	m.mu.Unlock()
+	return m.inner.GenerateContent(ctx, req, stream)
+}
+
+func (m *recordingModel) seen() []string {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return append([]string(nil), m.prompts...)
+}
+
+// TestTailRetentionE2E drives the rolling-summary path against a real model.
+//
+// TestCompactionE2E covers the sliding window, which summarizes each group of
+// turns once and never revisits it. Tail retention is the other trigger and the
+// one with a failure mode of its own: it seeds every new summary with the
+// previous one, so a value has to survive being recompressed on every pass.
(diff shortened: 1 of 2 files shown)


jjsasha63 (changes requested):
One of the three assertions cannot fail. I would fix that before this lands; everything else I checked holds up.

## The boundedness assertion never fires

Line 272 compares against `len(turns)*2`, which is 16. Without compaction the prompt for turn N holds 2N-1 contents, so turn 8 tops out at 15. I logged the real sizes on replay: 1, 3, 4, 4, 4, 4, 4, 4. The check passes with compaction switched off entirely.

The flat 4 is the property you are after. Compare the last prompt against an early one, or against `EventRetentionSize + 2`, rather than against a bound the conversation cannot reach.

## The seed assertion cannot fail on a green replay

The needle comes out of cassette response bytes and is searched in cassette request bytes. Take the seed away and the request bytes change, so the run dies at the turn loop on line 205 with a replay miss. I tried three mutations that destroy the seed path and all three failed there, none at line 246.

You say as much in the description, and the existing test has the same property. It does mean the behaviour this PR is named after is held up by the cassette rather than by the assertion written for it.

## An exact match works, so the longest-line heuristic is not needed

`strings.ReplaceAll(first, "\n", "\\n")` matches a recorded summarizer prompt today: 695 characters, against one line for the heuristic.

The heuristic can also pass for the wrong reason after a re-record. The 40-character floor admits turn 0 at 55 characters, and a raw turn is in the transcript whether or not seeding happened.

## Where this coverage already sits

The seed is asserted at `internal/compactioninternal/tail_retention_test.go:153` and `:1055`, and supersession and boundedness over 60 rounds at `runner/compaction_test.go:1482`. All three assertions here are the weaker copy.

What is new is the summarizer's real prompt template. `NewLLMSummarizer` appears in no test outside `session/compaction/` and the two llmagent end-to-end files, so nothing else drives it around a rolling seed. That is what pays for the cassette, and I would point the assertions at it.

## Smaller things

- 67 of the 171 code lines are verbatim from `llmagent_compaction_test.go`, the longest run being 22 lines. `minCassetteBytes`, `compactionModel`, `sessionEventsFor` and `textOf` already come from there, so a shared `requireCassette(t)` would sit on the line you have already drawn.
- `recordingModel` does not implement the optional `googlellm.GoogleLLM` interface, so the summarizer reports `BackendUnspecified`. Telemetry only.
- The preflight on line 119 builds the cassette path without the `ReplaceAll("/", "_")` the real open applies. It agrees today and diverges if subtests arrive.

## The cassette

I searched it for `AIza`, `Authorization`, `Bearer`, `Cookie`, `?key=`, `x-goog-api-key`, `x-goog-user-project`, email addresses, IPv4 addresses, project numbers and JWT prefixes, and found none. Requests carry `Content-Length`, `Content-Type`, `Host` and `User-Agent`, which is a subset of what the neighbouring cassettes carry, and no request URL has a query string. `TestHTTPRecordDirectivesPartitionCassettes` passes. At 95 KB it is the largest cassette here, 2.4 times the next.

## Parity with adk-python 2.8.0

`_events_to_compact_for_token_threshold` in `src/google/adk/apps/compaction.py:265-313` returns `[seed_event] + events_to_compact`, built from the previous compaction's `start_timestamp` and `compacted_content`. That is the same rolling seed as `selectTailRetentionWindow`, and `token_threshold` and `event_retention_size` match the Go field names.

`tests/unittests/apps/test_compaction_runner_e2e.py` names a sliding-window test at line 165 and nothing for the rolling seed, so this adds coverage Python does not have at that level.

## To see the first finding

Add a `t.Logf` over `captured` before line 272 and run:

```
go test ./agent/llmagent/ -run '^TestTailRetentionE2E$' -count=1 -v
```

It prints 1, 3, 4, 4, 4, 4, 4, 4 against a threshold of 16.

PR #1440, opened 2026-08-31
feat(artifact): guarantee artifact version metadata

### LINKED ISSUE #681: Missing contract test coverage for GetArtifactVersion()

yiaany:

- Closes: #681

**Problem:**
`GetArtifactVersion` exposed five fields, but only version and MIME type were guaranteed across artifact backends, and callers had no custom metadata write path.

**Solution:**
Add `SaveRequest.CustomMetadata`, persist metadata and creation time in memory and GCS, provide deterministic in-memory canonical URIs, convert GCS metadata values to strings, and extend shared and backend-specific contract coverage.

**Compatibility:**
The requested `map[string]any` field makes `SaveRequest` non-comparable. Code that compares `SaveRequest` values, uses them with APIs requiring `comparable`, or uses them as map keys will no longer compile. No in-repository caller relies on comparability, but this is a source-compatibility break for possible external callers and requires explicit maintainer acceptance. Negative-version behavior and dependencies are unchanged.

### Testing Plan

**Unit Tests:**

- [x] `go test -race -mod=readonly -count=10 ./artifact/...`
- [x] `go test -race -mod=readonly -count=10 ./internal/artifact/...`
- [x] `go vet ./artifact/... ./internal/artifact/...`
- [x] `go build -mod=readonly -v ./...` in both modules
- [x] `go mod tidy -diff` in both modules (no diff)
- [x] `golangci-lint run` v2.3.1 in both modules (0 issues in an LF checkout)

The review hardening pass also documents shallow in-memory metadata copies and MIME-at-save semantics, keeps the clock provider outside the service lock, verifies non-string in-memory metadata, and makes the fake GCS writer reject attribute setters after the first write.

A full Windows `go test -race -mod=readonly -count=1 -shuffle=on work` also ran. Artifact packages passed; unrelated timing/order-sensitive tests were flaky on both this branch and clean `origin/main`. GitHub Linux CI is the authoritative full-suite check.

Not applicable. This is an offline artifact service API/backend contract change covered by in-memory and fake-GCS tests.

### Checklist

- [x] Targeted and backend unit tests pass locally with my changes.
- [ ] All repository tests pass locally on Windows; unrelated base-branch flakes are described above.
- [ ] Manual E2E testing is not applicable.
- [x] No dependent downstream changes are required.


karolpiotrowicz (dismissed):
The change is well built — the backend contract is genuinely symmetric now, and the isolation and retry behavior are pinned by tests rather than asserted. One thing needs resolving before this can merge, and it is about the API surface rather than the implementation.

## This is not an additive-only change

The PR description says "This is an additive API change only". Adding [`CustomMetadata map[string]any`](https://github.com/google/adk-go/blob/e23e8f3b2fd70299c2813a331d120d79ab5fb9c1/artifact/service.go#L70) makes the exported `SaveRequest` non-comparable, which is a source-incompatible change for anyone importing `google.golang.org/adk/v2/artifact`. `apidiff` reports it as `SaveRequest: old is comparable, new is not`, and three shapes that compile against `main` stop compiling against this branch:

```go
a != b                                // invalid operation: a != b (struct containing map[string]any cannot be compared)
slices.Contains([]artifact.SaveRequest{a}, b)   // SaveRequest does not satisfy comparable
map[artifact.SaveRequest]bool{a: true}          // invalid map key type SaveRequest
```

Nothing in this repository does any of those — I checked all 34 `SaveRequest` references across 9 files, and every one uses keyed literals and never compares or hashes the struct. External callers are the exposure, and there is no way to measure them.

I want to be clear that this is not a coding mistake, and I do not think the field should change shape. No Go struct containing a map is comparable, so there is no idiomatic way to add this feature and keep the property. A `*Metadata` pointer field would preserve it and would be a worse API. Separately, the unkeyed-literal breakage that also follows from adding a field is explicitly sanctioned by the [Go 1 compatibility promise](https://go.dev/doc/go1compat) and is not a concern — I moved the field below `Version` to check whether position helped, and it does not, since adding any field breaks unkeyed literals of that struct regardless of where it goes.

So the ask is a process one rather than a code one. `AGENTS.md` lists "Break the public API — keep changes backward-compatible" under **Never** and "Any change to the public API surface, and any breaking change" under **Ask first**. Could you correct the "additive only" line in the description to state the comparability break, and let a maintainer decide whether to accept it? That decision is theirs, and the current description would let the change land without anyone making it.

## The in-memory backend shares nested metadata values

[`maps.Clone`](https://github.com/google/adk-go/blob/e23e8f3b2fd70299c2813a331d120d79ab5fb9c1/artifact/inmemory.go#L170) copies one level, and the value type is `any`, so any container stored as a value stays shared between the caller, the stored entry, and every reader that [`GetArtifactVersion`](https://github.com/google/adk-go/blob/e23e8f3b2fd70299c2813a331d120d79ab5fb9c1/artifact/inmemory.go#L346) hands one to. Two callers who only ever *read* can race on service-owned memory:

<details><summary>Reproduction — <code>go test -race -run TestNested ./artifact/</code></summary>

```go
func TestNestedMetadataIsShared(t *testing.T) {
	ctx, srv := t.Context(), artifact.InMemoryService()
	inner := map[string]any{"role": "user"}
	srv.Save(ctx, &artifact.SaveRequest{
		AppName: "app", UserID: "u", SessionID: "s", FileName: "f",
		Part: genai.NewPartFromText("x"), CustomMetadata: map[string]any{"acl": inner},
	})

	// The caller mutating its own map after Save changes stored state.
	inner["role"] = "admin"

	// And one reader's mutation is visible to the next reader.
	get := func() map[string]any {
		r, _ := srv.GetArtifactVersion(ctx, &artifact.GetArtifactVersionRequest{
			AppName: "app", UserID: "u", SessionID: "s", FileName: "f"})
		return r.ArtifactVersion.CustomMetadata
	}
	get()["acl"].(map[string]any)["role"] = "root"
	t.Log(get()["acl"]) // map[role:root]
}
```

Four goroutines doing only the `get()["acl"].(map[string]any)["role"] = …` line produce `WARNING: DATA RACE ... mapassign_faststr` under `-race`.

</details>

I'd rate this low. The same service already stores [`req.Part` by pointer](https://github.com/google/adk-go/blob/e23e8f3b2fd70299c2813a331d120d79ab5fb9c1/artifact/inmemory.go#L174-L179) and returns that identical pointer from `Load`, so the payload is more aliased than the metadata map is, and shallow cloning of `map[string]any` is the pattern in `session/inmemory.go` too. What makes it worth a line of documentation is that the clone plus [the test that mutates the returned map](https://github.com/google/adk-go/blob/e23e8f3b2fd70299c2813a331d120d79ab5fb9c1/artifact/inmemory_test.go#L120) together teach callers the returned map is theirs to mutate, which holds at the top level and quietly stops holding one level down. A sentence on `SaveRequest.CustomMetadata` saying the copy is shallow would settle it. The GCS backend is unaffected, since `fmt.Sprint` flattens every value to a string.

## `GetArtifactVersion` and `Load` can disagree about the MIME type

`MimeType` is now snapshotted at save time while the `Part` itself stays aliased, so mutating the Part afterwards splits the two views of one version:

```go
p := genai.NewPartFromBytes(data, "image/png")
srv.Save(ctx, &artifact.SaveRequest{ /* … */ Part: p})
p.InlineData.MIMEType = "text/html"

srv.GetArtifactVersion(ctx, req).ArtifactVersion.MimeType  // "image/png"
srv.Load(ctx, req).Part.InlineData.MIMEType                // "text/html"
```

On `main` both return `text/html`. I think the snapshot is the right call — it matches what GCS already does, since GCS `Load` rebuilds the Part from the stored object — so the residual inconsistency is really the pre-existing pointer aliasing in `Load` that you did not touch. Worth either copying the Part at save so both agree, or a note on `ArtifactVersion.MimeType` that it records the value at save time.

## Smaller things

A caller-supplied clock runs inside the write lock. [`platform.Now(ctx)`](https://github.com/google/adk-go/blob/e23e8f3b2fd70299c2813a331d120d79ab5fb9c1/artifact/inmemory.go#L180) is called under the `s.mu.Lock()` taken at [line 163](https://github.com/google/adk-go/blob/e23e8f3b2fd70299c2813a331d120d79ab5fb9c1/artifact/inmemory.go#L163), so a `TimeProvider` that touches the same service deadlocks it permanently, and a merely slow one stalls every caller of that instance. `session/inmemory.go` does the same thing, so this is a pattern rather than something you introduced, but hoisting the call above the lock is free here.

A self-referential metadata value crashes the process on the GCS path. [`fmt.Sprint(value)`](https://github.com/google/adk-go/blob/e23e8f3b2fd70299c2813a331d120d79ab5fb9c1/artifact/gcsartifact/service.go#L141) has no cycle guard, so `cyc := map[string]any{}; cyc["self"] = cyc` passed as a value produces `fatal error: stack overflow` with `gcsService.Save` on the stack, and `recover()` cannot catch it. Such a value has to be built in-process rather than decoded, and the same `fmt.Sprint`-on-`any` shape exists elsewhere in the repo, so this is not urgent — but the in-memory backend accepts the identical input silently, which makes it a develop-on-memory, crash-on-GCS difference.

The [canonical URI](https://github.com/google/adk-go/blob/e23e8f3b2fd70299c2813a331d120d79ab5fb9c1/artifact/inmemory.go#L337) interpolates `AppName`, `UserID` and `SessionID` unescaped, and only `FileName` is checked for path separators. `("u", "s/x/users/v/sessions/w")` and `("u/sessions/s/x/users/v", "w")` are stored under distinct keys and render the same URI. Nothing reads the field yet, so this is only a question of whether it is meant to identify a version.

Finally, two test suggestions. Every `CustomMetadata` value in the suite is already a `string`, so the in-memory backend's promise to preserve the original `any` — the thing that distinguishes it from GCS — is not exercised. One `int` or `bool` value in `service_suite.go` would cover it. And [`fakeWriter.Close`](https://github.com/google/adk-go/blob/e23e8f3b2fd70299c2813a331d120d79ab5fb9c1/artifact/gcsartifact/gcs_test.go#L645-L648) commits whatever the setters last recorded, so it accepts an ordering the real `storage.Writer` rejects — moving `SetMetadata` after the `Write` calls passes the whole suite even though real GCS ignores attributes set after the first write. Your code has the order right. Having `Write` mark the writer opened, and the setters refuse after that, would keep it that way and would protect `SetContentType` at the same time.

## What holds up

The retry path is the part I most expected to be wrong and is not: the converted metadata is built once outside the loop and re-applied to a fresh writer on every attempt, with a test asserting all sixteen attempts carry identical content. The nil-to-empty-map normalization is real on both backends rather than incidental, `platform.Now` is threaded properly for replay-safety, and the isolation tests you wrote are the reason I could be precise about where the shallow clone stops working. Build, vet, `gofmt`, `staticcheck` and `govulncheck` are clean, and the full suite passes on your branch, on `main`, and on the two merged — the only two failing packages fail identically everywhere and are unrelated to this change.

One note for later: #1164 also edits `artifact/service.go` and `artifact/inmemory.go`. It does not change `SaveRequest`, so there is no API interaction, but whichever lands second will need a rebase.

yiaany:
Thanks for the careful review. I updated the PR description to call out the `SaveRequest` comparability break explicitly rather than describing the change as additive-only. I agree that accepting that tradeoff is a maintainer decision.

I also pushed a focused hardening commit:

- documented that the in-memory service shallow-copies metadata maps and that `MimeType` is recorded at save time;
- moved `platform.Now(ctx)` outside the in-memory service lock;
- added `int` and `bool` metadata coverage to the in-memory backend test. I kept the shared suite string-only because GCS returns stringified values and the issue contract requires exact shared round-trips;
- made the fake GCS writer reject `SetContentType` or `SetMetadata` after the first `Write`, so the tests now enforce the real writer ordering constraint.

I did not add generic deep-copy behavior for `map[string]any` or clone `genai.Part`: both require semantics for arbitrary pointers, cycles, and user-defined values, and cloning `Part` would change the existing `Load` aliasing behavior beyond this issue. The documented shallow-copy boundary makes the current behavior explicit.

I also left cycle-safe replacement of `fmt.Sprint` and URI component escaping unchanged. The former would diverge from the conversion specified for this issue, and the latter needs a separate decision on URI grammar and input validation. Both seem better handled independently rather than silently defining new behavior here.

The branch is updated to current `main`, and the repeated artifact race tests and vet pass. Could a maintainer confirm whether the requested `map[string]any` API is acceptable with the comparability tradeoff, and re-review the implementation changes?

karolpiotrowicz:

It ports adk-python's artifact version metadata faithfully — the same five fields — but the Go shape makes `SaveRequest` non-comparable, which `apidiff` reports as an incompatible change to a released v2 module. There is no better implementation available here, since no Go struct containing a map is comparable, so the question is whether we accept the source-compatibility break or ask for a non-breaking carrier such as an options argument or a separate request type. Worth deciding before the author reworks anything.

yiaany:
> @wolo-lab flagging this one for a scope call rather than a code review.
> 
> It ports adk-python's artifact version metadata faithfully — the same five fields — but the Go shape makes `SaveRequest` non-comparable, which `apidiff` reports as an incompatible change to a released v2 module. There is no better implementation available here, since no Go struct containing a map is comparable, so the question is whether we accept the source-compatibility break or ask for a non-breaking carrier such as an options argument or a separate request type. Worth deciding before the author reworks anything.

Thanks for escalating this. I’ll hold off on further API changes until the maintainers decide whether to accept the comparability break or prefer a different non-breaking API shape. I’m happy to update the implementation once the intended direction is confirmed.

wolo-lab:
**Decision on the scope question: let's take the comparability break.** `CustomMetadata map[string]any` on `SaveRequest` is the right shape. There is one change I'd like before merge and it is not about comparability — see the last section.

### Why the break is acceptable

- **The comparability we'd give up doesn't mean what it looks like.** `SaveRequest` holds `Part *genai.Part` ([`artifact/service.go#L60-L70` at `b52d1fb`](https://github.com/google/adk-go/blob/b52d1fbeaa3b457bc739269a79846b569bc12c55/artifact/service.go#L60-L70)), so `==` is pointer identity on that field. Two requests with identical content compare unequal, and two sharing one `*Part` stay equal after that Part is mutated. No caller can build request deduplication or caching on that. The only content-meaningful use left is the zero-value check `req == artifact.SaveRequest{}`.
- **Every alternative is a larger break.** Both an options argument and a separate request type change the exported `artifact.Service` interface ([`Save` at `b52d1fb`](https://github.com/google/adk-go/blob/b52d1fbeaa3b457bc739269a79846b569bc12c55/artifact/service.go#L39)), which breaks every external implementation of `Service` — anyone running an S3 or local-disk backend. That trades a hypothetical break for a certain one, against a population that plausibly exists.
- **Comparability is spent once.** The first map or slice field costs the break and every one after it is free. Keeping it freezes `SaveRequest` to scalars and pointers for the life of v2, and this struct has obvious future fields that are neither — labels, retention, checksums. The cost is also lowest now, with v2.0.0 tagged at the end of June.

One shape does preserve comparability, and I want it recorded as considered and rejected: a pointer to a wrapper, `CustomMetadata *Metadata`, where `Metadata` holds the map. It compiles, the struct stays comparable, and two requests with identical metadata still compare unequal — so it keeps the property and delivers none of its value, at the price of `&artifact.Metadata{Values: …}` in every literal and permanent asymmetry with `ArtifactVersion`.

### One correction for the record

`ArtifactVersion` already carrying `CustomMetadata map[string]any` is not precedent for this. That type did not exist in v1.0.0 and arrived in v1.1.0 already holding the map ([`artifact/service.go#L266-L272` at `ea1865b`](https://github.com/google/adk-go/blob/ea1865bdaa66024c2f96cbc1653ad24f17227457/artifact/service.go#L266-L272)), so it never broke compiled code. `SaveRequest` has been comparable in every release from v1.0.0 through v2.3.0. Removing a property a type has always had is a different act from never having it.

What `ArtifactVersion` does establish is narrower and still useful: `map[string]any` is the accepted metadata shape in this API already, and it matches adk-python's `custom_metadata: Optional[dict[str, Any]]`. The type is not what's in question here.

### A caveat for the release note

Losing comparability is usually a compile error, but not always. When a `SaveRequest` value passes through `any`, the failure moves to run time and the build stays green:

```go
var m sync.Map
m.Store(*req, true)          // runtime error: hash of unhashable type artifact.SaveRequest

var a, b any = *req1, *req2
_ = a == b                   // runtime error: comparing uncomparable type artifact.SaveRequest
```

The same applies to `map[any]T` keys, `context.WithValue` keys and `atomic.Value.CompareAndSwap`. It is unlikely, since every entry point takes `*SaveRequest` and every call site in the repo builds `&SaveRequest{...}`. But "callers only get a compile error" isn't accurate, so the v2.4.0 release note should name the type and say so in a sentence.

### What I'd like changed before merge

The comparability break is loud and low-probability. There's a quiet, higher-probability problem in the current diff, and that's where I'd rather spend the review.

GCS stringifies every value with `fmt.Sprint` and stores `map[string]string` ([`artifact/gcsartifact/service.go#L137-L143` at `3aeed5d`](https://github.com/google/adk-go/blob/3aeed5d1907a6634c1536d8bd27fd633ac9826ba/artifact/gcsartifact/service.go#L137-L143)), while the in-memory service keeps the original values via `maps.Clone` ([`artifact/inmemory.go#L162` at `3aeed5d`](https://github.com/google/adk-go/blob/3aeed5d1907a6634c1536d8bd27fd633ac9826ba/artifact/inmemory.go#L162)). `GetArtifactVersion` therefore returns a `map[string]any` whose dynamic value types depend on which backend sits behind the interface. A caller who saves `{"count": 42}` and reads it back as `md["count"].(int)` succeeds against the in-memory service and panics against GCS. The in-memory test ([`artifact/inmemory_test.go#L112` at `3aeed5d`](https://github.com/google/adk-go/blob/3aeed5d1907a6634c1536d8bd27fd633ac9826ba/artifact/inmemory_test.go#L112)) asserts that `int` and `bool` survive the round trip, which pins that divergence into the contract.

That's worse than the comparability break on three counts. It's silent, it's environment-dependent so it passes tests and fails in production, and unlike comparability it can't be corrected later without changing behavior.

The `Service` interface should carry one metadata contract regardless of backend. My preference is for the in-memory service to stringify as well, so the test backend behaves like the production one instead of being more permissive, with the doc comment on `CustomMetadata` stating that values round-trip as strings. If you'd rather keep type fidelity in memory, the alternative is to reject non-string values in `Validate`. The middle position in the current diff, where the type says `any` and the meaning depends on the backend, is the one option I don't think we should ship.

yiaany:
> **Decision on the scope question: let's take the comparability break.** `CustomMetadata map[string]any` on `SaveRequest` is the right shape. There is one change I'd like before merge and it is not about comparability — see the last section.
> 
> ### Why the break is acceptable
> * **The comparability we'd give up doesn't mean what it looks like.** `SaveRequest` holds `Part *genai.Part` ([`artifact/service.go#L60-L70` at `b52d1fb`](https://github.com/google/adk-go/blob/b52d1fbeaa3b457bc739269a79846b569bc12c55/artifact/service.go#L60-L70)), so `==` is pointer identity on that field. Two requests with identical content compare unequal, and two sharing one `*Part` stay equal after that Part is mutated. No caller can build request deduplication or caching on that. The only content-meaningful use left is the zero-value check `req == artifact.SaveRequest{}`.
> * **Every alternative is a larger break.** Both an options argument and a separate request type change the exported `artifact.Service` interface ([`Save` at `b52d1fb`](https://github.com/google/adk-go/blob/b52d1fbeaa3b457bc739269a79846b569bc12c55/artifact/service.go#L39)), which breaks every external implementation of `Service` — anyone running an S3 or local-disk backend. That trades a hypothetical break for a certain one, against a population that plausibly exists.
> * **Comparability is spent once.** The first map or slice field costs the break and every one after it is free. Keeping it freezes `SaveRequest` to scalars and pointers for the life of v2, and this struct has obvious future fields that are neither — labels, retention, checksums. The cost is also lowest now, with v2.0.0 tagged at the end of June.
> 
> One shape does preserve comparability, and I want it recorded as considered and rejected: a pointer to a wrapper, `CustomMetadata *Metadata`, where `Metadata` holds the map. It compiles, the struct stays comparable, and two requests with identical metadata still compare unequal — so it keeps the property and delivers none of its value, at the price of `&artifact.Metadata{Values: …}` in every literal and permanent asymmetry with `ArtifactVersion`.
> 
> ### One correction for the record
> `ArtifactVersion` already carrying `CustomMetadata map[string]any` is not precedent for this. That type did not exist in v1.0.0 and arrived in v1.1.0 already holding the map ([`artifact/service.go#L266-L272` at `ea1865b`](https://github.com/google/adk-go/blob/ea1865bdaa66024c2f96cbc1653ad24f17227457/artifact/service.go#L266-L272)), so it never broke compiled code. `SaveRequest` has been comparable in every release from v1.0.0 through v2.3.0. Removing a property a type has always had is a different act from never having it.
> 
> What `ArtifactVersion` does establish is narrower and still useful: `map[string]any` is the accepted metadata shape in this API already, and it matches adk-python's `custom_metadata: Optional[dict[str, Any]]`. The type is not what's in question here.
> 
> ### A caveat for the release note
> Losing comparability is usually a compile error, but not always. When a `SaveRequest` value passes through `any`, the failure moves to run time and the build stays green:
> 
> ```go
> var m sync.Map
> m.Store(*req, true)          // runtime error: hash of unhashable type artifact.SaveRequest
> 
> var a, b any = *req1, *req2
> _ = a == b                   // runtime error: comparing uncomparable type artifact.SaveRequest
> ```
> 
> The same applies to `map[any]T` keys, `context.WithValue` keys and `atomic.Value.CompareAndSwap`. It is unlikely, since every entry point takes `*SaveRequest` and every call site in the repo builds `&SaveRequest{...}`. But "callers only get a compile error" isn't accurate, so the v2.4.0 release note should name the type and say so in a sentence.
> 
> ### What I'd like changed before merge
> The comparability break is loud and low-probability. There's a quiet, higher-probability problem in the current diff, and that's where I'd rather spend the review.
> 
> GCS stringifies every value with `fmt.Sprint` and stores `map[string]string` ([`artifact/gcsartifact/service.go#L137-L143` at `3aeed5d`](https://github.com/google/adk-go/blob/3aeed5d1907a6634c1536d8bd27fd633ac9826ba/artifact/gcsartifact/service.go#L137-L143)), while the in-memory service keeps the original values via `maps.Clone` ([`artifact/inmemory.go#L162` at `3aeed5d`](https://github.com/google/adk-go/blob/3aeed5d1907a6634c1536d8bd27fd633ac9826ba/artifact/inmemory.go#L162)). `GetArtifactVersion` therefore returns a `map[string]any` whose dynamic value types depend on which backend sits behind the interface. A caller who saves `{"count": 42}` and reads it back as `md["count"].(int)` succeeds against the in-memory service and panics against GCS. The in-memory test ([`artifact/inmemory_test.go#L112` at `3aeed5d`](https://github.com/google/adk-go/blob/3aeed5d1907a6634c1536d8bd27fd633ac9826ba/artifact/inmemory_test.go#L112)) asserts that `int` and `bool` survive the round trip, which pins that divergence into the contract.
> 
> That's worse than the comparability break on three counts. It's silent, it's environment-dependent so it passes tests and fails in production, and unlike comparability it can't be corrected later without changing behavior.
> 
> The `Service` interface should carry one metadata contract regardless of backend. My preference is for the in-memory service to stringify as well, so the test backend behaves like the production one instead of being more permissive, with the doc comment on `CustomMetadata` stating that values round-trip as strings. If you'd rather keep type fidelity in memory, the alternative is to reject non-string values in `Validate`. The middle position in the current diff, where the type says `any` and the meaning depends on the backend, is the one option I don't think we should ship.

Thanks for making the scope decision and for catching the backend-dependent value types. I agree that the service contract should not change depending on the backend.

I’ll update the in-memory service to stringify every custom metadata value at save time using the same fmt.Sprint semantics as GCS, revise the documentation to state that values are returned as strings by every backend, and update the in-memory tests accordingly. This also removes the shallow-copy distinction because stored metadata values will be detached strings.

I’ll keep the shared suite string-only and add the comparability caveat to the compatibility notes for the release record.

yiaany:
> **Decision on the scope question: let's take the comparability break.** `CustomMetadata map[string]any` on `SaveRequest` is the right shape. There is one change I'd like before merge and it is not about comparability — see the last section.
> 
> ### Why the break is acceptable
> * **The comparability we'd give up doesn't mean what it looks like.** `SaveRequest` holds `Part *genai.Part` ([`artifact/service.go#L60-L70` at `b52d1fb`](https://github.com/google/adk-go/blob/b52d1fbeaa3b457bc739269a79846b569bc12c55/artifact/service.go#L60-L70)), so `==` is pointer identity on that field. Two requests with identical content compare unequal, and two sharing one `*Part` stay equal after that Part is mutated. No caller can build request deduplication or caching on that. The only content-meaningful use left is the zero-value check `req == artifact.SaveRequest{}`.
> * **Every alternative is a larger break.** Both an options argument and a separate request type change the exported `artifact.Service` interface ([`Save` at `b52d1fb`](https://github.com/google/adk-go/blob/b52d1fbeaa3b457bc739269a79846b569bc12c55/artifact/service.go#L39)), which breaks every external implementation of `Service` — anyone running an S3 or local-disk backend. That trades a hypothetical break for a certain one, against a population that plausibly exists.
> * **Comparability is spent once.** The first map or slice field costs the break and every one after it is free. Keeping it freezes `SaveRequest` to scalars and pointers for the life of v2, and this struct has obvious future fields that are neither — labels, retention, checksums. The cost is also lowest now, with v2.0.0 tagged at the end of June.
> 
> One shape does preserve comparability, and I want it recorded as considered and rejected: a pointer to a wrapper, `CustomMetadata *Metadata`, where `Metadata` holds the map. It compiles, the struct stays comparable, and two requests with identical metadata still compare unequal — so it keeps the property and delivers none of its value, at the price of `&artifact.Metadata{Values: …}` in every literal and permanent asymmetry with `ArtifactVersion`.
> 
> ### One correction for the record
> `ArtifactVersion` already carrying `CustomMetadata map[string]any` is not precedent for this. That type did not exist in v1.0.0 and arrived in v1.1.0 already holding the map ([`artifact/service.go#L266-L272` at `ea1865b`](https://github.com/google/adk-go/blob/ea1865bdaa66024c2f96cbc1653ad24f17227457/artifact/service.go#L266-L272)), so it never broke compiled code. `SaveRequest` has been comparable in every release from v1.0.0 through v2.3.0. Removing a property a type has always had is a different act from never having it.
> 
> What `ArtifactVersion` does establish is narrower and still useful: `map[string]any` is the accepted metadata shape in this API already, and it matches adk-python's `custom_metadata: Optional[dict[str, Any]]`. The type is not what's in question here.
> 
> ### A caveat for the release note
> Losing comparability is usually a compile error, but not always. When a `SaveRequest` value passes through `any`, the failure moves to run time and the build stays green:
> 
> ```go
> var m sync.Map
> m.Store(*req, true)          // runtime error: hash of unhashable type artifact.SaveRequest
> 
> var a, b any = *req1, *req2
> _ = a == b                   // runtime error: comparing uncomparable type artifact.SaveRequest
> ```
> 
> The same applies to `map[any]T` keys, `context.WithValue` keys and `atomic.Value.CompareAndSwap`. It is unlikely, since every entry point takes `*SaveRequest` and every call site in the repo builds `&SaveRequest{...}`. But "callers only get a compile error" isn't accurate, so the v2.4.0 release note should name the type and say so in a sentence.
> 
> ### What I'd like changed before merge
> The comparability break is loud and low-probability. There's a quiet, higher-probability problem in the current diff, and that's where I'd rather spend the review.
> 
> GCS stringifies every value with `fmt.Sprint` and stores `map[string]string` ([`artifact/gcsartifact/service.go#L137-L143` at `3aeed5d`](https://github.com/google/adk-go/blob/3aeed5d1907a6634c1536d8bd27fd633ac9826ba/artifact/gcsartifact/service.go#L137-L143)), while the in-memory service keeps the original values via `maps.Clone` ([`artifact/inmemory.go#L162` at `3aeed5d`](https://github.com/google/adk-go/blob/3aeed5d1907a6634c1536d8bd27fd633ac9826ba/artifact/inmemory.go#L162)). `GetArtifactVersion` therefore returns a `map[string]any` whose dynamic value types depend on which backend sits behind the interface. A caller who saves `{"count": 42}` and reads it back as `md["count"].(int)` succeeds against the in-memory service and panics against GCS. The in-memory test ([`artifact/inmemory_test.go#L112` at `3aeed5d`](https://github.com/google/adk-go/blob/3aeed5d1907a6634c1536d8bd27fd633ac9826ba/artifact/inmemory_test.go#L112)) asserts that `int` and `bool` survive the round trip, which pins that divergence into the contract.
> 
> That's worse than the comparability break on three counts. It's silent, it's environment-dependent so it passes tests and fails in production, and unlike comparability it can't be corrected later without changing behavior.
> 
> The `Service` interface should carry one metadata contract regardless of backend. My preference is for the in-memory service to stringify as well, so the test backend behaves like the production one instead of being more permissive, with the doc comment on `CustomMetadata` stating that values round-trip as strings. If you'd rather keep type fidelity in memory, the alternative is to reject non-string values in `Validate`. The middle position in the current diff, where the type says `any` and the meaning depends on the backend, is the one option I don't think we should ship.

I’ve updated the branch. The in-memory service now stringifies metadata values on save, matching GCS. I also updated the documentation and shared contract tests to enforce the same behavior across both backends. The targeted race tests and vet pass. Ready for re-review.

artifact/inmemory.go
```diff
@@ -145,11 +155,19 @@ func (s *inMemoryService) Save(ctx context.Context, req *SaveRequest) (*SaveResp
...
 	}
+	customMetadata := make(map[string]any, len(req.CustomMetadata))
+	for key, value := range req.CustomMetadata {
+		customMetadata[key] = fmt.Sprint(value)
+	}
+	mimeType := "text/plain"
+	if req.Part.InlineData != nil {
+		mimeType = req.Part.InlineData.MIMEType
+	}
+	createTime := platform.Now(ctx)
```
karolpiotrowicz, line 168: The clock is read here but the version is not assigned until inside the lock two lines down, so a goroutine descheduled in between writes an earlier timestamp against a later version. I could not make this happen under natural scheduling in 200 attempts, and it reproduces every time if you force the deschedule, so the window is real and narrow.

I am deliberately not suggesting you move the call under the lock. `platform.Now` runs a caller-supplied closure, and running that inside a non-reentrant service-wide write lock trades this for a deadlock class, which is worse. Since you moved it out on purpose in the last round, a one-line comment saying why would stop the next reader (or reviewer) from filing this again.

Worth knowing that the race detector cannot see this at all — it is an ordering inversion, not a data race, so a green `-race` says nothing about it.

internal/artifact/tests/service_suite.go
```diff
@@ -503,3 +553,25 @@ func testArtifactService_GetArtifactVersion(ctx context.Context, t *testing.T, s
...
+	if got.Version != wantVersion {
+		t.Errorf("ArtifactVersion.Version = %d, want %d", got.Version, wantVersion)
+	}
+	if got.MimeType != wantMimeType {
+		t.Errorf("ArtifactVersion.MimeType = %q, want %q", got.MimeType, wantMimeType)
+	}
+	if diff := cmp.Diff(wantMetadata, got.CustomMetadata); diff != "" {
+		t.Errorf("ArtifactVersion.CustomMetadata mismatch (-want +got):\n%s", diff)
+	}
+	if got.CanonicalURI == "" {
```
karolpiotrowicz, line 576: This checks that `CanonicalURI` is non-empty and line 574 checks that `CreateTime` is non-zero, so neither field's value is actually asserted anywhere that both backends run.

I checked what that costs by breaking each one on purpose. Replacing the in-memory URI with the constant `"memory://"`, and separately stamping `CreateTime` at read time instead of save time, both get caught — but only by `inmemory_test.go`. Neither failure cites this file.

That is defensible layering, since the URI schemes and the clocks genuinely differ per backend and there is no shared exact value to assert. The cheap thing that would discriminate here without assuming a scheme is that two versions of the same file must yield different `CanonicalURI`s.

internal/artifact/tests/service_suite.go
```diff
@@ -233,26 +233,28 @@ func testArtifactService_UserScoped(ctx context.Context, t *testing.T, srv artif
...
 	for i, data := range testData {
 		wantVersion := data.version
 		got, err := srv.Save(ctx, &artifact.SaveRequest{
-			AppName: appName, UserID: userID, SessionID: sessionID, FileName: data.fileName,
-			Part: data.artifact,
+			AppName: appName, UserID: userID, SessionID: data.sessionID, FileName: data.fileName,
+			Part: data.artifact, CustomMetadata: data.metadata,
 		})
 		if err != nil || got.Version != wantVersion {
 			t.Errorf("[%d] Save() = (%v, %v), want (%v, nil)", i, got.Version, err, wantVersion)
```
karolpiotrowicz, line 260: `got.Version` is dereferenced here on a path reached when `err != nil`, and every implementation returns `(nil, err)` in that case, so a failing `Save` panics the shared suite instead of reporting which backend failed and why. The condition on line 259 is fine — `||` short-circuits — it is only the `t.Errorf` argument that runs unguarded.

This predates your change and the same pattern sits at line 99, so it is not yours to fix. Raising it because the lines just above are yours: the user-scoped saves now run under three different session IDs and carry a metadata payload, which gives a backend two new ways to fail right here.

artifact/gcsartifact/gcs_test.go
```diff
@@ -64,6 +65,139 @@ func TestGCSArtifactService(t *testing.T) {
...
+			fileName:      "user:file.txt",
+			saveSessionID: "save-session",
+			getSessionID:  "irrelevant-session",
+			uriPrefix:     "gs://bucket/app/user/user/user:file.txt/",
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			svc := newGCSServiceForTesting("bucket")
+			fb := svc.bucket.(*fakeBucket)
+			fb.now = func() time.Time { return firstCreateTime }
```
karolpiotrowicz: `fb.now` is written here (and at 112 and 174) without the mutex, while `fakeWriter.Close` reads it under `bkt.mu`.

Not a race today, since this test is sequential and builds its own service. The reason it is worth three lines now is that the same `fakeBucket` type backs the ten-goroutine concurrent-save test in this file, so the first person to combine a clock hook with a concurrent case gets a race report pointing at the fixture rather than at the code under test.

karolpiotrowicz (changes requested):
Three things need fixing before this merges, all of them new in this change and all reproduced against your branch merged with `main`.

First, the good news on the open question: wolo's requested change is in and it works. Reverting `artifact/inmemory.go` to the pre-fix revision while keeping your new tests fails seven subtests with `ArtifactVersion.CustomMetadata mismatch`, so the stringification is genuinely pinned by the suite rather than just present. The comparability tradeoff is settled and I have left it alone.

## `Load` and `GetArtifactVersion` now disagree about the same version

`Save` copies the media type out of the request at [`artifact/inmemory.go#L166-L169`](https://github.com/google/adk-go/blob/b71c34f3348189676eaa8563e73d924ee4bdf471/artifact/inmemory.go#L166-L169) while storing the caller's `*genai.Part` by pointer at [`#L180`](https://github.com/google/adk-go/blob/b71c34f3348189676eaa8563e73d924ee4bdf471/artifact/inmemory.go#L180), and `Load` hands that same pointer straight back at [`#L237`](https://github.com/google/adk-go/blob/b71c34f3348189676eaa8563e73d924ee4bdf471/artifact/inmemory.go#L237) and [`#L244`](https://github.com/google/adk-go/blob/b71c34f3348189676eaa8563e73d924ee4bdf471/artifact/inmemory.go#L244).

Before this change both reads derived the media type from that one live pointer, so they could not disagree. Snapshotting one field while leaving the source of the other aliased is what splits them, and the split is in-memory only — the GCS backend reads both from the persisted object.

The property worth holding is that within one backend, `Load` and `GetArtifactVersion` report the same media type for the same version.

Worth noting that [`artifact/inmemory_test.go#L75`](https://github.com/google/adk-go/blob/b71c34f3348189676eaa8563e73d924ee4bdf471/artifact/inmemory_test.go#L75) performs exactly this mutation and then asserts only `GetArtifactVersion`. The file never calls `Load`, so the new test pins the snapshot half and cannot see the disagreement it creates.

<details><summary>Reproduction — <code>go test -run TestSplit ./artifact/</code></summary>

```go
package artifact_test

import (
	"context"
	"testing"

	"google.golang.org/genai"

	"google.golang.org/adk/v2/artifact"
)

func TestSplit(t *testing.T) {
	ctx := context.Background()
	s := artifact.InMemoryService()
	p := genai.NewPartFromBytes([]byte("x"), "image/png")

	if _, err := s.Save(ctx, &artifact.SaveRequest{
		AppName: "app", UserID: "u", SessionID: "s", FileName: "f", Part: p}); err != nil {
		t.Fatal(err)
	}
	p.InlineData.MIMEType = "application/changed" // the caller still holds the pointer

	gv, err := s.GetArtifactVersion(ctx, &artifact.GetArtifactVersionRequest{
		AppName: "app", UserID: "u", SessionID: "s", FileName: "f"})
	if err != nil {
		t.Fatal(err)
	}
	ld, err := s.Load(ctx, &artifact.LoadRequest{
		AppName: "app", UserID: "u", SessionID: "s", FileName: "f"})
	if err != nil {
		t.Fatal(err)
	}

	// On this branch: "image/png" vs "application/changed".
	// On the merge-base: both "application/changed".
	if gv.ArtifactVersion.MimeType != ld.Part.InlineData.MIMEType {
		t.Errorf("GetArtifactVersion=%q Load=%q", gv.ArtifactVersion.MimeType, ld.Part.InlineData.MIMEType)
	}
}
```

</details>

## `CanonicalURI` is built from unescaped identifiers

[`artifact/inmemory.go#L338`](https://github.com/google/adk-go/blob/b71c34f3348189676eaa8563e73d924ee4bdf471/artifact/inmemory.go#L338) and [`#L340`](https://github.com/google/adk-go/blob/b71c34f3348189676eaa8563e73d924ee4bdf471/artifact/inmemory.go#L340) interpolate `AppName`, `UserID`, `SessionID` and `FileName` directly. `validateFileName` rejects only `/` and `\`, and the other three identifiers are checked for emptiness and nothing else.

A file named `report?v=old` produces `memory://apps/app/users/u/sessions/s/artifacts/report?v=old/versions/1`. Run that through `url.Parse` and `v=old/versions/1` lands in `RawQuery`, so the version is no longer in the path. A `#` in the name does the same thing via the fragment. Both names pass `Validate` today.

Separately, `AppName="a/users/b", UserID="c"` and `AppName="a", UserID="b/users/c"` render byte-identical URIs. The in-memory storage key is unaffected, since it uses a length-prefixed encoding — it is the identifier string that collides.

Either the four segments need escaping, or the field needs to say it is an opaque identity that must not be parsed. I have not implemented either, so I would rather leave the choice to you than prescribe one: the gs:// form at `artifact/gcsartifact/service.go` has the same shape, and escaping one backend and not the other would make them diverge.

```go
// Add to artifact/inmemory_test.go and run: go test -run TestURI ./artifact/
func TestURI(t *testing.T) {
	ctx := context.Background()
	s := artifact.InMemoryService()
	name := "report?v=old"
	if _, err := s.Save(ctx, &artifact.SaveRequest{
		AppName: "app", UserID: "u", SessionID: "s", FileName: name,
		Part: genai.NewPartFromText("x")}); err != nil {
		t.Fatal(err)
	}
	r, _ := s.GetArtifactVersion(ctx, &artifact.GetArtifactVersionRequest{
		AppName: "app", UserID: "u", SessionID: "s", FileName: name})
	u, _ := url.Parse(r.ArtifactVersion.CanonicalURI)
	t.Logf("URI=%q -> Path=%q RawQuery=%q", r.ArtifactVersion.CanonicalURI, u.Path, u.RawQuery)
	// Path=".../artifacts/report"  RawQuery="v=old/versions/1"
}
```

## The new doc comment promises something other implementations cannot deliver

[`artifact/service.go#L275-L277`](https://github.com/google/adk-go/blob/b71c34f3348189676eaa8563e73d924ee4bdf471/artifact/service.go#L275-L277) says *"Every Service implementation populates all five fields. CustomMetadata is non-nil, contains string values, and is empty when no custom metadata was saved."*

`artifact.Service` is exported and this change does not alter it, so a backend written against v2.3.0 — someone's own object-store or local-disk implementation — still satisfies the interface unchanged and still returns `CustomMetadata == nil` with a zero `CreateTime`. I compiled one to check. A caller who reads that sentence and writes `av.CustomMetadata[k] = v` gets `panic: assignment to entry in nil map`.

The sentence is also not true of this module's own in-memory backend in one case: a `Part` carrying inline data with an empty `MIMEType` passes `Validate` and yields `MimeType: ""`. That behavior predates your change — only the sentence asserting otherwise is new.

Phrasing this as a requirement on implementations rather than a statement of fact about them would resolve both halves.

One thing that is **not** yours to fix, flagged so it does not get misattributed: after merging with `main` there is a contradiction between that sentence and the `CanonicalURI` field doc a few lines below, which says a service with no such scheme *"may leave it empty"*. That field doc arrived on `main` in #1563 and is in neither your branch nor your merge-base. It is worth reconciling as part of this merge, but you did not write it.

## Smaller things

I have left these as inline comments on the lines they concern: the `platform.Now(ctx)` placement relative to the lock, the two shared-suite assertions that check presence rather than value, an unsynchronized write in the GCS test fake, and a pre-existing nil dereference in the shared suite's error path.

## Two process notes

The branch conflicts with `main`, and also with open #1600, #1465 and #1164 — whoever merges first decides who rebases. I reviewed the merge result rather than the branch tip, resolving the one conflict in the test fake's `attrs()` by keeping both sides' fields, so if you resolve it differently some of the above may shift.

CI has never run on this head. Three check suites are sitting at `action_required` waiting for a maintainer to approve the fork workflow, so nothing here has been through the project's own build.

artifact/inmemory.go
yiaany: Thanks for the detailed follow-up review. I have pushed commit 30d4226 with the requested fixes, on top of merge commit 83c5592 that brings the branch up to the current main.

The update addresses the review points as follows:

- In-memory MIME consistency:
  - Removed the separate saved MIME-type snapshot from the in-memory artifact entry.
  - GetArtifactVersion now derives MimeType from the same stored *genai.Part returned by Load.
  - This preserves the existing Part aliasing behavior while guaranteeing that Load and GetArtifactVersion cannot report different media types for the same in-memory version.
  - Added regression coverage that mutates the original Part after Save and compares the MIME type returned through both APIs for both session-scoped and user-scoped artifacts.

- Canonical URI escaping:
  - AppName, UserID, SessionID and FileName are now escaped as URI path segments with url.PathEscape in both the in-memory and GCS implementations.
  - The actual GCS object naming and persistence layout remain unchanged.
  - Existing canonical URIs containing ordinary identifiers are unchanged.
  - Added session-scoped and user-scoped tests for both backends using `/`, `?` and `#`.
  - The tests assert the exact percent-encoded URI and verify with url.Parse that no unintended query or fragment is created.

- ArtifactVersion documentation:
  - Rephrased the documentation as normative requirements for Service implementations rather than claims about every existing third-party implementation.
  - CanonicalURI is documented as a stable service-specific identity rather than necessarily a download URL.
  - CustomMetadata is required to be non-nil and contain string values.
  - MimeType is allowed to be empty when no media type was associated with the saved part.

- Clock placement:
  - Kept platform.Now(ctx) outside the in-memory service lock.
  - Added a comment explaining that a caller-provided time provider may call back into the service, so invoking it while holding the non-reentrant service lock could deadlock.

- Shared contract suite:
  - The suite now records the CanonicalURI values returned for versions 2 and 3 and asserts that they differ.
  - I also mutation-tested this assertion by forcing all versions to use the same URI; the shared suite failed as expected.

- GCS test clock synchronization:
  - Added fakeBucket.setNow, protected by the bucket mutex.
  - Replaced all direct assignments to fakeBucket.now with the synchronized setter.

- Merge conflict resolution:
  - Merged current main without rebasing.
  - Resolved the GCS fake conflict by retaining both sides: MediaLink from main and the persisted Created and Metadata fields required by this PR.
  - The PR is now reported as mergeable by GitHub.

I reviewed the pre-existing nil dereference in the shared suite’s Save failure path. As noted in the review, the same pattern predates this PR and also exists elsewhere in the suite. I left it unchanged to avoid expanding this PR into an unrelated test-harness cleanup.

Validation performed:

- go test -race -mod=readonly -count=20 ./artifact/...
- go test -race -mod=readonly -count=20 ./internal/artifact/...
- go build -mod=readonly work
- build and vet for both modules
- go mod tidy -diff for both modules
- golangci-lint run for both modules
- repository formatting checks
- git diff --check
- fresh-context review using the repository self-review checklist
- mutation checks for MIME consistency, URI segment escaping in both backends, and version-specific CanonicalURI identity

The full shuffled race suite was also run. It reached two failures that reproduce unchanged on a clean current-main worktree:

- internal/style_test.go treats Windows path separators as part of the filename and consequently reports existing copyright headers as missing;
- TestTelemetrySchema_Workflow is sensitive to sibling-span ordering and fails with shuffle seed 1789732011094072300.

All artifact packages and the changed behavior pass under the race detector. GitHub currently shows the CLA and GitHub Actions Scan as passing; the Go, API diff and workflow-security suites are still in action_required pending maintainer approval for the fork workflows.

PR #1267, opened 2026-08-07
fix(llmagent): resolve agent mode per placement instead of mutating the shared agent

wolo-lab:
Fixes #1137.

## Problem
An `LlmAgent` that declares no `Mode` takes one from where it is placed: chat as a sub-agent or a runner root, single_turn as a workflow node. That default was written back onto the agent itself — five sites wrote `llminternal.State.Mode` or `State.IncludeContents`, three of them on the run path.

Every write was guarded on the mode still being unset, so the **first placement to be constructed won** and later placements silently inherited it. The same agent instance used both as a graph node and as a chat sub-agent gave the coordinator a different tool set depending on construction order.

The writes were also **permanent and made on the run path**, so one turn decided the next. Measured on the merge base: running an undeclared agent once at a graph node leaves `Mode` at `single_turn` and `IncludeContents` at `none` on the object, and handing that same instance to `runner.New` afterwards then fails with `root agent worker must be a chat LlmAgent, but has mode single_turn`. Here both fields survive the run untouched and the second placement works. `TestOneInstance_ANodeRunDoesNotDecideItsNextPlacement` pins that sequence and reproduces the error when the write is restored.

The run-path writes were also a **data race**. One agent instance is documented to serve many concurrent invocations, and a `ParallelWorker` mapped over an agent node races activations of the same instance.

None of this is cosmetic: the mode decides which tools are installed (`finish_task`, `SingleTurnTool`/`TaskAgentTool`), whether conversation history is visible, and whether the agent is reachable by `transfer_to_agent`.

## Summary
- `State.Mode` is now the agent's own declaration and is never written after construction.
- A placement resolves the mode for **one** agent and records it on the invocation context, keyed by that agent's name **and its identity** (`boundModeKey{agent string, state *State}`).
- **Identity gives every agent its own slot.** A binding cannot be read by an agent it was not resolved for, and cannot be destroyed by a placement for a same-named other agent. That matters because names are unique only across `SubAgents()`, and a graph node's agent is not in `SubAgents()` — so `runner.New` will accept a tree holding two distinct agents with one name.
- The identity token is the `*State` rather than the `agent.Agent`. `Reveal` hands every binder and every reader the same pointer for one agent, a pointer is always comparable, and an `agent.Agent` in a context key would panic on an implementation whose dynamic type is not.
[...]

PR #1442, opened 2026-08-31
fix(sequentialagent): make RunLive's task_completed injection race-free

### LINKED ISSUE #1441: sequentialagent: RunLive's task_completed injection races on shared sub-agent state
**Describe the Bug:**

`(*sequentialAgent).RunLive` (`agent/workflowagents/sequentialagent/agent.go`) injects a `task_completed` tool and an instruction suffix into every LLM sub-agent that doesn't already have it, so the sequential pipeline knows when to hand off to the next sub-agent in live/streaming mode:

```go
for _, subAgent := range subAgents {
    if llmAgent, ok := subAgent.(llminternal.Agent); ok {
        state := llminternal.Reveal(llmAgent)
        hasTaskCompleted := false
        for _, t := range state.Tools {          // unsynchronized read
            if t.Name() == "task_completed" {
                hasTaskCompleted = true
                break
            }
        }
        if !hasTaskCompleted {
            state.Tools = append(state.Tools, taskCompletedTool)  // unsynchronized write
            state.Instruction += instructionSuffix                 // unsynchronized write
        }
    }
}
```

`llminternal.Reveal(llmAgent)` returns the sub-agent's shared, persistent internal state — the same object across every invocation of that agent, not a per-call copy (agents are constructed once at startup and reused across many requests, same assumption documented in #1137). Two concurrent `RunLive` calls on the same `SequentialAgent` (e.g. two simultaneous live/streaming sessions from different users hitting a shared agent tree) race on that memory, and can also double-inject the tool/instruction if both goroutines observe `hasTaskCompleted == false` before either has written.

**Steps to Reproduce:**

1. Build a `SequentialAgent` with at least one LLM sub-agent.
2. Call `RunLive` on it concurrently from two or more goroutines (sharing the same sub-agent instance).
3. Run with `go test -race`.

**Expected Behavior:**

The injection happens safely and exactly once per sub-agent, regardless of concurrent `RunLive` calls.

**Observed Behavior:**

`go test -race` reports a data race between the read at the `state.Tools` range and the write at `state.Tools = append(...)`, both inside `RunLive`. Under the right timing, the tool/instruction is also injected more than once.

**Environment Details:**
- ADK Library Version: main
- Go Version: 1.23+

**Related:** #1137 / #1166 fixed the same *class* of bug (unsynchronized writes to shared `llminternal.State`) in a different code path (`llm_agent_wrapper.go`'s single_turn dispatch); that fix does not touch `sequentialagent`, so this is a distinct occurrence of the same underlying pattern.

A fix is up at #TBD (opening the PR right after this issue).

tvost2:

- Closes: #1441

**Problem:**

`RunLive` injects a `task_completed` tool and instruction suffix into each LLM sub-agent, once, so a live/streaming sequential pipeline knows when to hand off to the next sub-agent. `llminternal.Reveal(llmAgent)` returns that sub-agent's shared, persistent state — the same object across every invocation of that agent, not a per-call copy (agents are constructed once and reused across concurrent requests, the same assumption #1137/#1166 already document for llmagent). The injection was an unguarded read-then-append (`state.Tools`) and read-then-concat (`state.Instruction`): two concurrent `RunLive` calls dispatched onto the same sub-agent — e.g. two simultaneous live sessions sharing an agent tree built once at startup — raced on that memory, and could double-inject the tool if both observed `hasTaskCompleted == false` before either had written it back.

**Solution:**

Follows the precedent #1166 set for the same class of bug: never write the shared `State` concurrently, rather than adding narrow locking that still leaves `State`'s other, ordinarily-unsynchronized readers racing against a mutation. A package-level `sync.Map` holds one `*sync.Once` per sub-agent `*llminternal.State` (a stable pointer for that agent's lifetime); the injection runs inside `once.Do`, so it happens exactly once, and every concurrent `RunLive` caller for that sub-agent blocks on that `Once` until it has completed before touching `Tools`/`Instruction`, instead of each racing its own unsynchronized check.

This does not attempt to make `State` generally safe for concurrent read/write from arbitrary call sites — same scope boundary #1166 drew for its own fix — only to stop this specific call site from writing it concurrently.

### Testing Plan

- Added `TestSequentialAgent_RunLive_ConcurrentInjectionIsRaceFree`: 20 goroutines call `RunLive` on a shared sub-agent, then the test asserts the tool was injected exactly once.
- Verified this test fails under `go test -race` on the pre-fix code (stashed the fix, reran, observed both the data race and, separately, a double-injection window) and passes cleanly and repeatedly with the fix in place.
- `gofmt` and `go vet` clean on both touched files.
- Existing package tests (`TestSequentialAgent_RunLive_Injection`, `TestSequentialAgent_RunLive_SequentialOrchestration`, `TestNewSequentialAgent`) still pass.

### DIFF
--- a/agent/workflowagents/sequentialagent/agent.go
+++ b/agent/workflowagents/sequentialagent/agent.go
@@ -151,17 +151,18 @@ func (a *sequentialAgent) RunLive(ctx agent.InvocationContext) (agent.LiveSessio
 			state := llminternal.Reveal(llmAgent)
 			// Reveal returns the sub-agent's shared, persistent State, so
 			// concurrent live sessions driving one agent tree would race
-			// this read-then-append on Tools/Instruction. state.LiveInjection
+			// this read-then-append on Tools/Instruction. state.LiveModeInjection
 			// runs it once per agent and gives every other RunLive caller a
 			// happens-before edge before they read Tools.
-			state.LiveInjection.Do(func() {
+			state.LiveModeInjection.Do(func() {
 				hasTaskCompleted := false
 				for _, t := range state.Tools {
 					// llmagent.New assigns the caller's cfg.Tools slice
-					// unvalidated (#1490), so a nil element can reach here;
-					// skip an untyped nil. A typed nil still panics in
-					// t.Name() below — on every call, as on main, because
-					// LiveInjection.Do does not latch on panic.
+					// without validating it (#1490), so a nil element can
+					// reach here; skip an untyped nil. A typed nil may panic
+					// in t.Name(), depending on its implementation. When it
+					// does, LiveModeInjection.Do does not latch the panic, so
+					// the next call retries the scan.
 					if t == nil {
 						continue
 					}
--- a/agent/workflowagents/sequentialagent/agent.go
+++ b/agent/workflowagents/sequentialagent/agent.go
@@ -149,18 +149,33 @@ func (a *sequentialAgent) RunLive(ctx agent.InvocationContext) (agent.LiveSessio
 	for _, subAgent := range subAgents {
 		if llmAgent, ok := subAgent.(llminternal.Agent); ok {
 			state := llminternal.Reveal(llmAgent)
-			hasTaskCompleted := false
-			for _, t := range state.Tools {
-				if t.Name() == "task_completed" {
-					hasTaskCompleted = true
-					break
+			// Reveal returns the sub-agent's shared, persistent State, so
+			// concurrent live sessions driving one agent tree would race
+			// this read-then-append on Tools/Instruction. The guard lives
+			// on State (see llminternal.State.TaskCompletedInjection): the
+			// injection runs exactly once, and every other RunLive caller
+			// for this sub-agent blocks on it before reading Tools rather
+			// than racing its own unsynchronized check.
+			state.TaskCompletedInjection.Do(func() {
+				hasTaskCompleted := false
+				for _, t := range state.Tools {
+					// llmagent.New copies cfg.Tools without validating, so a
+					// nil element can reach here; skip it rather than panic
+					// inside sync.Once (which would latch the guard shut).
+					if t == nil {
+						continue
+					}
+					if t.Name() == "task_completed" {
+						hasTaskCompleted = true
+						break
+					}
 				}
-			}
-			if !hasTaskCompleted {
-				state.Tools = append(state.Tools, taskCompletedTool)
-				instructionSuffix := "\nIf you finished the user's request according to its description, call the task_completed function to exit so the next agents can take over. When calling this function, do not generate any text other than the function call."
-				state.Instruction += instructionSuffix
-			}
+				if !hasTaskCompleted {
+					state.Tools = append(state.Tools, taskCompletedTool)
+					instructionSuffix := "\nIf you finished the user's request according to its description, call the task_completed function to exit so the next agents can take over. When calling this function, do not generate any text other than the function call."
+					state.Instruction += instructionSuffix
+				}
+			})
 		}
 	}
 
--- a/agent/workflowagents/sequentialagent/agent.go
+++ b/agent/workflowagents/sequentialagent/agent.go
@@ -151,17 +151,17 @@ func (a *sequentialAgent) RunLive(ctx agent.InvocationContext) (agent.LiveSessio
 			state := llminternal.Reveal(llmAgent)
 			// Reveal returns the sub-agent's shared, persistent State, so
 			// concurrent live sessions driving one agent tree would race
-			// this read-then-append on Tools/Instruction. The guard lives
-			// on State (see llminternal.State.TaskCompletedInjection): the
-			// injection runs exactly once, and every other RunLive caller
-			// for this sub-agent blocks on it before reading Tools rather
-			// than racing its own unsynchronized check.
-			state.TaskCompletedInjection.Do(func() {
+			// this read-then-append on Tools/Instruction. state.LiveInjection
+			// runs it once per agent and gives every other RunLive caller a
+			// happens-before edge before they read Tools.
+			state.LiveInjection.Do(func() {
 				hasTaskCompleted := false
 				for _, t := range state.Tools {
-					// llmagent.New copies cfg.Tools without validating, so a
-					// nil element can reach here; skip it rather than panic
-					// inside sync.Once (which would latch the guard shut).
+					// llmagent.New assigns the caller's cfg.Tools slice
+					// unvalidated (#1490), so a nil element can reach here;
+					// skip an untyped nil. A typed nil still panics in
+					// t.Name() below — on every call, as on main, because
+					// LiveInjection.Do does not latch on panic.
 					if t == nil {
 						continue
 					}
--- a/agent/workflowagents/sequentialagent/agent_test.go
+++ b/agent/workflowagents/sequentialagent/agent_test.go
@@ -476,13 +476,19 @@ const taskCompletedInstructionMarker = "call the task_completed function to exit
 // marker would pass the final count assertion but let a losing caller
 // return before the winner's write is visible -- that read is what fails
 // under -race without the barrier.
+//
+// The pipeline has three sub-agents so RunLive's injection loop runs more
+// than one iteration, and each sub-agent carries its own State with its own
+// sync.Once: the guard has to hold independently for every one of them.
 func TestSequentialAgent_RunLive_ConcurrentInjectionIsRaceFree(t *testing.T) {
-	subAgent := newCustomAgent(t, 1)
+	subAgents := []agent.Agent{
+		newCustomAgent(t, 1), newCustomAgent(t, 2), newCustomAgent(t, 3),
+	}
 
 	sequentialAgent, err := sequentialagent.New(sequentialagent.Config{
 		AgentConfig: agent.Config{
 			Name:      "seq_agent",
-			SubAgents: []agent.Agent{subAgent},
+			SubAgents: subAgents,
 		},
 	})
 	if err != nil {
@@ -496,14 +502,28 @@ func TestSequentialAgent_RunLive_ConcurrentInjectionIsRaceFree(t *testing.T) {
 		t.Fatalf("sequential agent does not implement RunLive")
 	}
 
-	llmAgent, ok := subAgent.(llminternal.Agent)
-	if !ok {
-		t.Fatal("sub-agent does not implement llminternal.Agent")
+	states := make([]*llminternal.State, len(subAgents))
+	for i, sub := range subAgents {
+		llmAgent, ok := sub.(llminternal.Agent)
+		if !ok {
+			t.Fatalf("sub-agent %d does not implement llminternal.Agent", i)
+		}
+		states[i] = llminternal.Reveal(llmAgent)
 	}
-	state := llminternal.Reveal(llmAgent)
+
+	countTaskCompleted := func(s *llminternal.State) int {
+		n := 0
+		for _, tl := range s.Tools {
+			if tl != nil && tl.Name() == "task_completed" {
+				n++
+			}
+		}
+		return n
+	}
 
 	const concurrent = 20
-	seenByCaller := make([]int, concurrent)
+	// seenByCaller[i][j]: what caller i saw in sub-agent j's Tools.
+	seenByCaller := make([][]int, concurrent)
 	var wg sync.WaitGroup
 	wg.Add(concurrent)
 	for i := range concurrent {
@@ -511,38 +531,35 @@ func TestSequentialAgent_RunLive_ConcurrentInjectionIsRaceFree(t *testing.T) {
 			defer wg.Done()
 			invCtx := &mockInvocationContext{agent: sequentialAgent, invocationID: "test_id", ctx: t.Context()}
 			_, _, _ = liveAgent.RunLive(invCtx)
-			// RunLive has returned, so the Once has completed: this read
-			// is ordered after the winner's write and must see it.
-			for _, tl := range state.Tools {
-				if tl != nil && tl.Name() == "task_completed" {
-					seenByCaller[i]++
-				}
+			// RunLive has returned, so every sub-agent's Once has completed:
+			// these reads are ordered after the winners' writes.
+			seen := make([]int, len(states))
+			for j, s := range states {
+				seen[j] = countTaskCompleted(s)
 			}
+			seenByCaller[i] = seen
 		}()
 	}
 	wg.Wait()
 
 	for i, seen := range seenByCaller {
-		if seen != 1 {
-			t.Errorf("caller %d saw task_completed %d times in state.Tools after RunLive returned, want exactly 1", i, seen)
+		for j, n := range seen {
+			if n != 1 {
+				t.Errorf("caller %d saw task_completed %d times in sub-agent %d's Tools after RunLive returned, want exactly 1", i, n, j)
+			}
 		}
 	}
 
-	count := 0
-	for _, tl := range state.Tools {
-		if tl != nil && tl.Name() == "task_completed" {
-			count++
+	for j, s := range states {
+		if n := countTaskCompleted(s); n != 1 {
+			t.Errorf("sub-agent %d: task_completed injected %d times across %d concurrent RunLive calls, want exactly 1", j, n, concurrent)
 		}
-	}
-	if count != 1 {
-		t.Errorf("task_completed injected %d times across %d concurrent RunLive calls, want exactly 1", count, concurrent)
-	}
-
-	// The tool is useless without the instruction telling the model to
-	// call it; two goroutines can both append the suffix and still leave
-	// one tool in the slice, so the suffix count is a distinct check.
-	if n := strings.Count(state.Instruction, taskCompletedInstructionMarker); n != 1 {
-		t.Errorf("task_completed instruction suffix appears %d times, want exactly 1", n)
+		// The tool is useless without the instruction telling the model to
+		// call it; two goroutines can both append the suffix and still leave
+		// one tool in the slice, so the suffix count is a distinct check.
+		if n := strings.Count(s.Instruction, taskCompletedInstructionMarker); n != 1 {
+			t.Errorf("sub-agent %d: task_completed instruction suffix appears %d times, want exactly 1", j, n)
+		}
 	}
 }
 
--- a/agent/workflowagents/sequentialagent/agent_test.go
+++ b/agent/workflowagents/sequentialagent/agent_test.go
@@ -34,7 +34,16 @@ import (
 	"google.golang.org/adk/v2/model"
 	"google.golang.org/adk/v2/runner"
 	"google.golang.org/adk/v2/session"
+	"google.golang.org/adk/v2/tool"
 )
+
+// stubTool is a minimal tool.Tool for tests that only need a named entry
+// in state.Tools.
+type stubTool struct{ name string }
+
+func (s stubTool) Name() string        { return s.name }
+func (s stubTool) Description() string { return s.name }
+func (s stubTool) IsLongRunning() bool { return false }
 
 func TestNewSequentialAgent(t *testing.T) {
 	type args struct {
@@ -470,16 +479,16 @@ const taskCompletedInstructionMarker = "call the task_completed function to exit
 // either had written it back.
 //
 // Beyond "injected exactly once", the test pins the ordering guarantee
-// the sync.Once is there for: each goroutine reads state.Tools right
-// after its own RunLive returns and must observe the completed injection
-// (exactly one task_completed tool, never zero). A plain "already done"
-// marker would pass the final count assertion but let a losing caller
-// return before the winner's write is visible -- that read is what fails
-// under -race without the barrier.
+// the LiveInjection latch is there for: each goroutine reads state.Tools
+// right after its own RunLive returns and must observe the completed
+// injection (exactly one task_completed tool, never zero). A plain
+// "already done" marker would pass the final count assertion but let a
+// losing caller return before the winner's write is visible -- that read
+// is what fails under -race without the barrier.
 //
 // The pipeline has three sub-agents so RunLive's injection loop runs more
 // than one iteration, and each sub-agent carries its own State with its own
-// sync.Once: the guard has to hold independently for every one of them.
+// LiveInjection latch: the guard has to hold independently for every one.
 func TestSequentialAgent_RunLive_ConcurrentInjectionIsRaceFree(t *testing.T) {
 	subAgents := []agent.Agent{
 		newCustomAgent(t, 1), newCustomAgent(t, 2), newCustomAgent(t, 3),
@@ -560,6 +569,62 @@ func TestSequentialAgent_RunLive_ConcurrentInjectionIsRaceFree(t *testing.T) {
 		if n := strings.Count(s.Instruction, taskCompletedInstructionMarker); n != 1 {
 			t.Errorf("sub-agent %d: task_completed instruction suffix appears %d times, want exactly 1", j, n)
 		}
+	}
+}
+
+// TestSequentialAgent_RunLive_PreexistingToolAndNilSkip pins the two
+// branches of RunLive's injection loop that no other test reaches: the
+// idempotence check that leaves a sub-agent's own task_completed tool (and
+// the instruction) alone, and the untyped-nil skip in the same loop.
+// Deleting the "does it already have task_completed" scan and injecting
+// unconditionally otherwise keeps the whole package green, because the
+// LiveInjection guard fires exactly once either way.
+func TestSequentialAgent_RunLive_PreexistingToolAndNilSkip(t *testing.T) {
+	subAgent, err := llmagent.New(llmagent.Config{
+		Name:  "sub_with_task_completed",
+		Model: &FakeLLM{id: 1},
+		Tools: []tool.Tool{nil, stubTool{name: "task_completed"}}, // nil exercises the skip
+	})
+	if err != nil {
+		t.Fatalf("llmagent.New: %v", err)
+	}
+
+	seq, err := sequentialagent.New(sequentialagent.Config{
+		AgentConfig: agent.Config{Name: "seq_agent", SubAgents: []agent.Agent{subAgent}},
+	})
+	if err != nil {
+		t.Fatalf("sequentialagent.New: %v", err)
+	}
+	liveAgent, ok := seq.(interface {
+		RunLive(ctx agent.InvocationContext) (agent.LiveSession, iter.Seq2[*session.Event, error], error)
+	})
+	if !ok {
+		t.Fatalf("sequential agent does not implement RunLive")
+	}
+
+	state := llminternal.Reveal(subAgent.(llminternal.Agent))
+	toolsBefore := len(state.Tools)
+	instrBefore := state.Instruction
+
+	_, _, _ = liveAgent.RunLive(&mockInvocationContext{agent: seq, invocationID: "test_id", ctx: t.Context()})
+
+	if len(state.Tools) != toolsBefore {
+		t.Errorf("Tools grew from %d to %d; a pre-existing task_completed must suppress the append", toolsBefore, len(state.Tools))
+	}
+	got := 0
+	for _, tl := range state.Tools {
+		if tl != nil && tl.Name() == "task_completed" {
+			got++
+		}
+	}
+	if got != 1 {
+		t.Errorf("task_completed tool count = %d, want 1 (the sub-agent's own, untouched)", got)
+	}
+	if state.Instruction != instrBefore {
+		t.Errorf("Instruction changed; the hand-off suffix must not be appended when task_completed already exists")
+	}
+	if n := strings.Count(state.Instruction, taskCompletedInstructionMarker); n != 0 {
+		t.Errorf("instruction suffix appears %d times, want 0", n)
 	}
 }
 
--- a/agent/workflowagents/sequentialagent/agent_test.go
+++ b/agent/workflowagents/sequentialagent/agent_test.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"fmt"
 	"iter"
+	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -448,6 +450,99 @@ func TestSequentialAgent_RunLive_Injection(t *testing.T) {
 		if !hasTaskCompleted {
 			t.Errorf("sub-agent 1 does not have task_completed tool injected after RunLive")
 		}
+	}
+}
+
+// taskCompletedInstructionMarker is a stable fragment of the instruction
+// suffix RunLive appends alongside the task_completed tool. The test
+// counts it instead of hardcoding the whole sentence.
+const taskCompletedInstructionMarker = "call the task_completed function to exit so the next agents can take over"
+
+// TestSequentialAgent_RunLive_ConcurrentInjectionIsRaceFree is the
+// regression test for the data race in RunLive's task_completed
+// injection: llminternal.Reveal(llmAgent) returns the sub-agent's shared,
+// persistent state, and the unguarded read-then-append/read-then-concat
+// used to race under any two concurrent RunLive calls dispatched onto the
+// same sub-agent (e.g. two simultaneous live/streaming sessions sharing
+// an agent tree built once at startup) -- confirmed with `go test -race`
+// before the fix. It also raced itself into double-injecting the tool
+// whenever both goroutines observed hasTaskCompleted == false before
+// either had written it back.
+//
+// Beyond "injected exactly once", the test pins the ordering guarantee
+// the sync.Once is there for: each goroutine reads state.Tools right
+// after its own RunLive returns and must observe the completed injection
+// (exactly one task_completed tool, never zero). A plain "already done"
+// marker would pass the final count assertion but let a losing caller
+// return before the winner's write is visible -- that read is what fails
+// under -race without the barrier.
+func TestSequentialAgent_RunLive_ConcurrentInjectionIsRaceFree(t *testing.T) {
+	subAgent := newCustomAgent(t, 1)
+
+	sequentialAgent, err := sequentialagent.New(sequentialagent.Config{
+		AgentConfig: agent.Config{
+			Name:      "seq_agent",
+			SubAgents: []agent.Agent{subAgent},
+		},
+	})
+	if err != nil {
+		t.Fatalf("failed to create sequential agent: %v", err)
+	}
+
+	liveAgent, ok := sequentialAgent.(interface {
+		RunLive(ctx agent.InvocationContext) (agent.LiveSession, iter.Seq2[*session.Event, error], error)
+	})
+	if !ok {
+		t.Fatalf("sequential agent does not implement RunLive")
+	}
+
+	llmAgent, ok := subAgent.(llminternal.Agent)
+	if !ok {
+		t.Fatal("sub-agent does not implement llminternal.Agent")
+	}
+	state := llminternal.Reveal(llmAgent)
+
+	const concurrent = 20
+	seenByCaller := make([]int, concurrent)
+	var wg sync.WaitGroup
+	wg.Add(concurrent)
+	for i := range concurrent {
+		go func() {
+			defer wg.Done()
+			invCtx := &mockInvocationContext{agent: sequentialAgent, invocationID: "test_id", ctx: t.Context()}
+			_, _, _ = liveAgent.RunLive(invCtx)
+			// RunLive has returned, so the Once has completed: this read
+			// is ordered after the winner's write and must see it.
+			for _, tl := range state.Tools {
+				if tl != nil && tl.Name() == "task_completed" {
+					seenByCaller[i]++
+				}
+			}
+		}()
+	}
+	wg.Wait()
+
+	for i, seen := range seenByCaller {
+		if seen != 1 {
+			t.Errorf("caller %d saw task_completed %d times in state.Tools after RunLive returned, want exactly 1", i, seen)
+		}
+	}
+
+	count := 0
+	for _, tl := range state.Tools {
+		if tl != nil && tl.Name() == "task_completed" {
+			count++
+		}
+	}
+	if count != 1 {
+		t.Errorf("task_completed injected %d times across %d concurrent RunLive calls, want exactly 1", count, concurrent)
+	}
+
+	// The tool is useless without the instruction telling the model to
+	// call it; two goroutines can both append the suffix and still leave
+	// one tool in the slice, so the suffix count is a distinct check.
+	if n := strings.Count(state.Instruction, taskCompletedInstructionMarker); n != 1 {
+		t.Errorf("task_completed instruction suffix appears %d times, want exactly 1", n)
 	}
 }
 
--- a/agent/workflowagents/sequentialagent/agent_test.go
+++ b/agent/workflowagents/sequentialagent/agent_test.go
@@ -300,11 +300,16 @@ func TestNewSequentialAgent(t *testing.T) {
 }
 
 func newCustomAgent(t *testing.T, id int) agent.Agent {
+	return newCustomAgentWithTools(t, id, nil)
+}
+
+func newCustomAgentWithTools(t *testing.T, id int, tools []tool.Tool) agent.Agent {
 	t.Helper()
 
 	a, err := llmagent.New(llmagent.Config{
 		Name:  fmt.Sprintf("custom_agent_%v", id),
 		Model: &FakeLLM{id: id, callCounter: 0},
+		Tools: tools,
 	})
 	if err != nil {
 		t.Fatal(err)
@@ -405,7 +410,8 @@ func (m *mockInvocationContext) Err() error                  { return m.ctx.Err(
 func (m *mockInvocationContext) Value(key any) any           { return m.ctx.Value(key) }
 
 func TestSequentialAgent_RunLive_Injection(t *testing.T) {
-	subAgent1 := newCustomAgent(t, 1)
+	existingTaskCompleted := stubTool{name: "task_completed"}
+	subAgent1 := newCustomAgentWithTools(t, 1, []tool.Tool{existingTaskCompleted})
 	subAgent2 := newCustomAgent(t, 2)
 
 	sequentialAgent, err := sequentialagent.New(sequentialagent.Config{
@@ -421,10 +427,8 @@ func TestSequentialAgent_RunLive_Injection(t *testing.T) {
 	// Before RunLive, sub-agents do not have the task_completed tool
 	if llmAgent1, ok := subAgent1.(llminternal.Agent); ok {
 		state := llminternal.Reveal(llmAgent1)
-		for _, tool := range state.Tools {
-			if tool.Name() == "task_completed" {
-				t.Errorf("sub-agent 1 already has task_completed tool before RunLive")
-			}
+		if len(state.Tools) != 1 || state.Tools[0].Name() != existingTaskCompleted.Name() {
+			t.Fatalf("sub-agent 1 lost its pre-existing task_completed tool before RunLive: %#v", state.Tools)
 		}
 	}
 
@@ -449,15 +453,11 @@ func TestSequentialAgent_RunLive_Injection(t *testing.T) {
 	// After RunLive initiation, the sub-agents MUST have the task_completed tool injected!
 	if llmAgent1, ok := subAgent1.(llminternal.Agent); ok {
 		state := llminternal.Reveal(llmAgent1)
-		hasTaskCompleted := false
-		for _, tool := range state.Tools {
-			if tool.Name() == "task_completed" {
-				hasTaskCompleted = true
-				break
-			}
+		if len(state.Tools) != 1 || state.Tools[0].Name() != existingTaskCompleted.Name() {
+			t.Errorf("sub-agent 1 pre-existing task_completed tool changed after RunLive: %#v", state.Tools)
 		}
-		if !hasTaskCompleted {
-			t.Errorf("sub-agent 1 does not have task_completed tool injected after RunLive")
+		if strings.Contains(state.Instruction, taskCompletedInstructionMarker) {
+			t.Errorf("sub-agent 1 instruction unexpectedly received task_completed suffix")
 		}
 	}
 }
@@ -479,7 +479,7 @@ const taskCompletedInstructionMarker = "call the task_completed function to exit
 // either had written it back.
 //
 // Beyond "injected exactly once", the test pins the ordering guarantee
-// the LiveInjection latch is there for: each goroutine reads state.Tools
+// the LiveModeInjection latch is there for: each goroutine reads state.Tools
 // right after its own RunLive returns and must observe the completed
 // injection (exactly one task_completed tool, never zero). A plain
 // "already done" marker would pass the final count assertion but let a
@@ -488,7 +488,7 @@ const taskCompletedInstructionMarker = "call the task_completed function to exit
 //
 // The pipeline has three sub-agents so RunLive's injection loop runs more
 // than one iteration, and each sub-agent carries its own State with its own
-// LiveInjection latch: the guard has to hold independently for every one.
+// LiveModeInjection latch: the guard has to hold independently for every one.
 func TestSequentialAgent_RunLive_ConcurrentInjectionIsRaceFree(t *testing.T) {
 	subAgents := []agent.Agent{
 		newCustomAgent(t, 1), newCustomAgent(t, 2), newCustomAgent(t, 3),
--- a/internal/llminternal/agent.go
+++ b/internal/llminternal/agent.go
@@ -15,6 +15,8 @@
 package llminternal
 
 import (
+	"sync"
+
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/v2/agent"
@@ -60,6 +62,17 @@ type State struct {
 	OutputSchema *genai.Schema
 
 	OutputKey string
+
+	// TaskCompletedInjection guards the one-time injection of the
+	// task_completed tool and its instruction suffix that
+	// sequentialagent.RunLive performs on each LLM sub-agent. Reveal
+	// returns the same *State for the lifetime of an agent, so a tree
+	// built once and driven by several concurrent live sessions would
+	// otherwise race the read-then-append on Tools/Instruction. Keeping
+	// the guard on State ties its lifetime to the agent's, so it is
+	// collected with the agent and never accumulates in a process-global
+	// map. It is only ever used via (*State), never a copy.
+	TaskCompletedInjection sync.Once
 }
 
 type InstructionProvider func(ctx agent.ReadonlyContext) (string, error)
--- a/internal/llminternal/agent.go
+++ b/internal/llminternal/agent.go
@@ -63,18 +63,18 @@ type State struct {
 
 	OutputKey string
 
-	// LiveInjection guards the single live-mode tool injection an agent may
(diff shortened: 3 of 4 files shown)


karolpiotrowicz (dismissed):
The race is real and the reproduction holds up. Your new test fails on the merge-base in 15 of 15 separate processes under `-race`, and passes 50 consecutive runs with the fix applied. `sync.Once` is also the right primitive rather than a plain "already done" marker, because the callers that lose the race need an ordering guarantee against the winner's writes, not merely the knowledge that someone else did the work.

One thing needs to change before this can merge. It is in the caching mechanism, not in the idea.

## The map retains every sub-agent it ever touches

[`taskCompletedInjected`](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/agent/workflowagents/sequentialagent/agent.go#L95) is package-level and nothing ever deletes from it. The key is the `*llminternal.State`, and because [`llmAgent` embeds `llminternal.State` by value](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/agent/llmagent/llmagent.go#L425), `Reveal` hands back a pointer into the middle of the `llmAgent` allocation. Go's collector has no sub-object granularity, so holding that pointer keeps the whole agent alive — its model client, its tool slice, its instruction, and the sub-agent tree beneath it. The symbol is unexported and there is no eviction call, so an operator who hits this has no way to release the memory.

Driving the public `RunLive` on 500 throwaway agents, dropping every reference, then forcing three garbage collections:

```
with the fix:    map entries 500 of 500 retained    HeapAlloc +7,414,912 B
on merge-base:   (no map)                           HeapAlloc    -32,504 B
```

Nothing is collected. Two caveats on the number so it is not over-read: that is one sample per side, and the probe's stub model carries a 4 KB pad, so a good chunk of the per-agent figure is the probe's own shape. The existence of the retention does not depend on the number — a `sync.Map` with no `Delete` cannot shrink, and 500 of 500 entries survive.

The loaders in this repo hand out one prebuilt tree per request, so `adkrest` as shipped stays bounded. `agent.Loader` is a public interface called once per request though, and [`configurable.FromConfig`](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/internal/configurable/configurable_utils.go#L299) builds a fresh tree on every call with no memoization, so a caller who constructs agents per session or per tenant retains all of them for the life of the process.

I tried the obvious alternative and it works: make the `sync.Once` a field on `llminternal.State` and call `state.TaskCompletedOnce.Do(...)`, dropping the global map entirely. Retention returns to merge-base levels (−32,600 B), your regression test still passes `-race -count=50`, the package suite passes with `-race -shuffle=on`, and `go vet ./...` stays clean across the repo, so the added field trips no copylocks check. The lifetime of the guard then matches the lifetime of the agent and the eviction question disappears.

Worth knowing if you take a different route: the strong map key is currently the only thing preventing a freed `*State` address from being recycled onto a stale completed `Once`. Switching to weak pointers reopens that window and would need its own identity check.

## A panic in the closure leaves injection permanently off

`sync.Once.Do` marks itself done through a deferred store, so a closure that panics still latches. The only panic source inside is `t.Name()` on [line 173](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/agent/workflowagents/sequentialagent/agent.go#L173), and [`llmagent.New` copies `cfg.Tools` with no validation](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/agent/llmagent/llmagent.go#L86), so a nil element reaches it. The repo already treats that shape as reachable — [`safeCanonicalToolsDict` guards `if t == nil`](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/agent/llmagent/llm_agent_wrapper.go#L375) before the same call.

Building a sub-agent with `Tools: []tool.Tool{nil}` and calling `RunLive` three times, each with a `recover()`:

```
with the fix:   call 1 panics, calls 2 and 3 return a nil error   task_completed injected: no
on merge-base:  all three calls panic                             task_completed injected: no
```

Both are broken for this input, so this is not new breakage. What changed is that it went quiet: after the first panic, `RunLive` succeeds while the sub-agent has neither the tool nor the hand-off instruction, and the pipeline can never advance past it. The re-derive-every-call shape on the merge-base was self-healing by construction, and caching the decision gives that up.

## The test leaves two of its own guarantees unpinned

- **`state.Instruction` is never asserted**, here or anywhere else in the package. [The check](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/agent/workflowagents/sequentialagent/agent_test.go#L501-L510) counts `task_completed` in `state.Tools` and stops. Deleting [line 181](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/agent/workflowagents/sequentialagent/agent.go#L181) keeps the whole suite green, and the injection is a pair — the tool is useless without the directive telling the model to call it. Asserting the suffix appears exactly once would also pin the half of the concurrency guarantee that a tool count cannot see, since two goroutines can both append the suffix and still leave a single tool in the slice.
- **The ordering guarantee is untested.** Replacing lines 169-170 with `_, loaded := taskCompletedInjected.LoadOrStore(state, struct{}{})` and `if !loaded` injects exactly once and passes your test under `-race`, but gives the losing callers no edge to the winner's writes — which is exactly the property [the comment](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/agent/workflowagents/sequentialagent/agent.go#L91-L94) says the design is for. Having each goroutine read `state.Tools` before the barrier and assert on what it saw would close it.

## Smaller things

- The comment cites [#1166](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/agent/workflowagents/sequentialagent/agent.go#L87) as precedent and "#1166's fix", but that PR was closed without merging, and the #1137 work is landing through #1267 instead. [Lines 83-85](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/agent/workflowagents/sequentialagent/agent.go#L83-L85) also state that agents are constructed once and reused as though it were a guarantee. Nothing enforces it, and it is the whole justification for the global map, so it is worth removing along with the map.
- [`LoadOrStore(state, &sync.Once{})`](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/agent/workflowagents/sequentialagent/agent.go#L169) allocates a fresh `Once` on every call and discards it on all but the first. Small next to the `functiontool.New` above it, and it disappears with the field-on-State change anyway.

## One scope note

The `Once` orders `RunLive` against `RunLive`, and nothing else. A first injection running concurrently with an ordinary `Run` of the same sub-agent still races — the write at [line 179](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/agent/workflowagents/sequentialagent/agent.go#L179) against the read at [`tools_processor.go:39`](https://github.com/google/adk-go/blob/9bdb8c08da5a498bb26a8e4230c20772261014df/internal/llminternal/tools_processor.go#L39). Driving `runner.Run` against a first `seq.RunLive` on one shared sub-agent reports that race in 15 of 15 processes on both the merge-base and this branch, so it is not something you introduced and your change narrows the window rather than widening it. Flagging it only because the test name reads broader than what is covered, and the next person to touch these fields will read the name.

karolpiotrowicz (commented):
This does what round 1 asked, and the numbers back it. Driving 500 throwaway agent trees through `RunLive`, dropping every reference and forcing three GCs: all 500 sub-agents are collected and the heap returns to +9 KB, against 500 retained and +7.5 MB on the previous revision. Putting the guard on `State` instead of in a map keyed by `*State` means the lifetime question does not have to be answered at all.

The test change is the part I would not have predicted. Reading `state.Tools` per caller after that caller's own `RunLive` returns is what distinguishes a `sync.Once` from a plain "already done" marker. Substitute a `LoadOrStore` marker that injects exactly once but gives the losing callers no ordering edge, and it is that read the race detector flags, in all 15 runs I tried. The same substitution passed cleanly against the previous version of the test. The ordering guarantee is pinned now, not merely described.

## Correcting something I got wrong

My round-1 note said a panic inside `once.Do` leaves you with a `RunLive` that returns cleanly and then silently does nothing. I measured that at the `RunLive` return value and never drove the session it hands back. Driving it, the picture is different: `f.Tools` is `state.Tools` aliased through [`tools_processor.go:39-49`](https://github.com/google/adk-go/blob/ff54d692f47c97ccdd21d054cd4dc680fdbe3390/internal/llminternal/tools_processor.go#L39-L49), and [`request_confirmation_processor.go:45-46`](https://github.com/google/adk-go/blob/ff54d692f47c97ccdd21d054cd4dc680fdbe3390/internal/llminternal/request_confirmation_processor.go#L45-L46) calls `tool.Name()` on every element with no nil check, so a bad tool element is fatal the moment the stream is used — on `main` exactly as here. Nothing runs silently. Sorry for sending you after that.

The sliver that survives is optional. Your `t == nil` skip at [`agent.go:165`](https://github.com/google/adk-go/blob/ff54d692f47c97ccdd21d054cd4dc680fdbe3390/agent/workflowagents/sequentialagent/agent.go#L165) closes the untyped-nil case outright, which is better than `main`. A *typed* nil — `var mt *myTool; Tools: []tool.Tool{mt}` — is a non-nil interface, so it reaches `t.Name()` at [`agent.go:168`](https://github.com/google/adk-go/blob/ff54d692f47c97ccdd21d054cd4dc680fdbe3390/agent/workflowagents/sequentialagent/agent.go#L168) and still latches the guard. The only difference I could measure is which line the panic names, and no ADK constructor hands out a typed nil.

## Two adjacent bugs, neither yours

Both reproduce identically on `main`, so they are not for this PR. Worth separate issues if you want them.

- **Two agents built from one `Tools` slice with spare capacity race the append.** [`llmagent.go:86`](https://github.com/google/adk-go/blob/ff54d692f47c97ccdd21d054cd4dc680fdbe3390/agent/llmagent/llmagent.go#L86) aliases the caller's slice, so a `make([]tool.Tool, 1, 4)` shared between two agents under two separate SequentialAgents gives two independent guards writing the same backing array. Your change makes it strictly better — one racing write per agent instead of an unbounded number.
- **Nested SequentialAgents never inject the inner pipeline's sub-agent.** `RunLive` takes its work list from `ctx.Agent().SubAgents()` and re-invokes the inner `RunLive` with the same context without rebinding `Agent`, so `seq(llmA, seq(llmB))` injects `llmA` and not `llmB`.

One nit if you are in the test anyway: the concurrent test still uses a single sub-agent, so the loop it exercises has one iteration.

tvost2:
Addressed the test nit in `755d741`: the race-free test now drives three sub-agents instead of one, so RunLive's injection loop iterates and each sub-agent's own `State`/`sync.Once` is exercised independently — every concurrent caller must see exactly one `task_completed` tool and one instruction suffix in *each* sub-agent's state. `go vet` + `go test -race -count=3` green.

Thanks for re-running the numbers and for the correction on the panic behaviour. I'll file the two adjacent bugs (shared-slice append race, nested-SequentialAgent inner injection) as separate issues.

tvost2:
Filed the two adjacent bugs: #1490 (llmagent.New aliases the caller's Tools slice) and #1491 (nested SequentialAgent never injects the inner sub-agent). Both credited to your review here.

agent/workflowagents/sequentialagent/agent.go
```diff
@@ -149,18 +149,33 @@ func (a *sequentialAgent) RunLive(ctx agent.InvocationContext) (agent.LiveSessio
...
+			// concurrent live sessions driving one agent tree would race
+			// this read-then-append on Tools/Instruction. The guard lives
+			// on State (see llminternal.State.TaskCompletedInjection): the
+			// injection runs exactly once, and every other RunLive caller
+			// for this sub-agent blocks on it before reading Tools rather
+			// than racing its own unsynchronized check.
+			state.TaskCompletedInjection.Do(func() {
+				hasTaskCompleted := false
+				for _, t := range state.Tools {
+					// llmagent.New copies cfg.Tools without validating, so a
```
karolpiotrowicz: Small wording point on this comment: `llmagent.New` does not copy `cfg.Tools`, it assigns the caller's slice header (`Tools: cfg.Tools`). The "without validating" half is exactly right and is what justifies the branch, but the distinction matters because the aliasing is its own bug — it is the one filed as #1490.

internal/llminternal/agent.go
```diff
@@ -60,6 +62,17 @@ type State struct {
...
+	// TaskCompletedInjection guards the one-time injection of the
+	// task_completed tool and its instruction suffix that
+	// sequentialagent.RunLive performs on each LLM sub-agent. Reveal
+	// returns the same *State for the lifetime of an agent, so a tree
+	// built once and driven by several concurrent live sessions would
+	// otherwise race the read-then-append on Tools/Instruction. Keeping
+	// the guard on State ties its lifetime to the agent's, so it is
+	// collected with the agent and never accumulates in a process-global
+	// map. It is only ever used via (*State), never a copy.
+	TaskCompletedInjection sync.Once
```
karolpiotrowicz: Naming: this is a guard for one specific consumer sitting on the struct every LLM agent embeds, and there is nothing in the name or the doc that says it is a one-shot latch shared by whoever gets there first. Today `sequentialagent.RunLive` is the only injector, so nothing is wrong. If a second workflow agent ever reuses the field for its own live-mode injection, whichever fires first wins and the other silently does nothing. Naming it for the invariant rather than for the payload — one live-mode tool injection per agent, whoever performs it — would make that visible to the next person.

agent/workflowagents/sequentialagent/agent_test.go
```diff
@@ -451,6 +453,116 @@ func TestSequentialAgent_RunLive_Injection(t *testing.T) {
...
+// after its own RunLive returns and must observe the completed injection
+// (exactly one task_completed tool, never zero). A plain "already done"
+// marker would pass the final count assertion but let a losing caller
+// return before the winner's write is visible -- that read is what fails
+// under -race without the barrier.
+//
+// The pipeline has three sub-agents so RunLive's injection loop runs more
+// than one iteration, and each sub-agent carries its own State with its own
+// sync.Once: the guard has to hold independently for every one of them.
+func TestSequentialAgent_RunLive_ConcurrentInjectionIsRaceFree(t *testing.T) {
```
karolpiotrowicz, line 492: This test does not discriminate the idempotence scan. Replace the whole `for _, t := range state.Tools` check in `RunLive` with an unconditional append and it still passes, because the `sync.Once` makes it fire once regardless — the count assertions come out at exactly 1 either way. No test in the package gives a sub-agent a pre-existing tool, so nothing verifies that a user's own `task_completed` is left alone and that the instruction suffix is not appended. A case with `Tools` already holding a `task_completed`-named tool, asserting `len(state.Tools)` unchanged and zero instruction suffixes, would cover that branch and the new `t == nil` skip at the same time.

karolpiotrowicz (commented):
The race fix holds. Your test fails on `main` in 15 of 15 separate processes under the race detector, and the head passes 50 consecutive runs of it clean. Putting the guard on `State` rather than in a map was the right call, and the per-caller read you added in your latest commit ([755d741f](https://github.com/google/adk-go/commit/755d741f5dcfd251aa63890854d7dc6949c9d254)) genuinely pins the ordering edge rather than just the count — substitute an atomic marker that injects exactly once but gives the losing callers no happens-before edge, and it is that read the detector flags, in 15 of 15 runs.

One thing should change before this merges, and it starts with a correction to what I told you last round.

## The guard stays permanently set after a panic, and a typed nil still reaches it

Last round I said a typed nil would still leave the guard permanently set, and that the only difference I could measure was which line the panic names. That was wrong about the consequence, and I had not driven the second call.

[`if t == nil`](https://github.com/google/adk-go/blob/755d741f5dcfd251aa63890854d7dc6949c9d254/agent/workflowagents/sequentialagent/agent.go#L165) skips a nil interface value. A non-nil interface holding a nil pointer clears it and reaches [`t.Name()`](https://github.com/google/adk-go/blob/755d741f5dcfd251aa63890854d7dc6949c9d254/agent/workflowagents/sequentialagent/agent.go#L168) three lines down. `sync.Once` records completion in a deferred store, so a panicking closure still marks the guard done, and every call after it returns with no error and no injection — permanently, for the life of that agent.

Driving `RunLive` three times on a sub-agent built with a typed-nil element in `Tools`, each call wrapped in a recover:

```
                      call 1   call 2   call 3   task_completed tools present afterwards
with this PR applied  panic    clean    clean    0
on main               panic    panic    panic    0
```

On `main` the misconfiguration is loud every time, so you find it. Here it is loud once and then hands back a live pipeline with no hand-off tool and no hand-off instruction, quietly. [The comment above the check](https://github.com/google/adk-go/blob/755d741f5dcfd251aa63890854d7dc6949c9d254/agent/workflowagents/sequentialagent/agent.go#L162-L164) says the skip is there so a nil element does not "panic inside sync.Once (which would latch the guard shut)", which is the one thing it does not deliver for this shape. The element gets there because [`llmagent.New` assigns `cfg.Tools` unvalidated](https://github.com/google/adk-go/blob/755d741f5dcfd251aa63890854d7dc6949c9d254/agent/llmagent/llmagent.go#L86), exactly as your comment says.

A `sync.Mutex` plus a `bool` set only after the work completes closes it, and I ran that variant end to end: the typed-nil case panics on all three calls again, your concurrency test still passes `-race -count=50` with no races, the `sequentialagent` and `llmagent` suites pass `-race -shuffle=on`, and `go vet ./...` stays clean, so the mutex trips no copylocks check either.

<details>
<summary>Reproduction — <code>go test -race -run TestTypedNilLatchesTheOnce ./internal/reprotypednil/</code></summary>

```go
package reprotypednil_test

import (
	"context"
	"iter"
	"testing"
	"time"

	"google.golang.org/adk/v2/agent/workflowagents/sequentialagent"
	"google.golang.org/adk/v2/internal/llminternal"
)

type fakeLLM struct{}

func (f *fakeLLM) Name() string { return "fake" }
func (f *fakeLLM) GenerateContent(ctx context.Context, req *model.LLMRequest, stream bool) iter.Seq2[*model.LLMResponse, error] {
	return func(yield func(*model.LLMResponse, error) bool) {}
}

// A user-defined tool whose methods dereference the receiver.
type myTool struct{ name string }

func (n *myTool) Name() string        { return n.name }
func (n *myTool) Description() string { return n.name }
func (n *myTool) IsLongRunning() bool { return false }

type mockInvCtx struct {
	agent.InvocationContext
	a   agent.Agent
	ctx context.Context
}

func (m *mockInvCtx) Agent() agent.Agent          { return m.a }
func (m *mockInvCtx) InvocationID() string        { return "inv" }
func (m *mockInvCtx) Context() context.Context    { return m.ctx }
func (m *mockInvCtx) Deadline() (time.Time, bool) { return m.ctx.Deadline() }
func (m *mockInvCtx) Done() <-chan struct{}       { return m.ctx.Done() }
func (m *mockInvCtx) Err() error                  { return m.ctx.Err() }
func (m *mockInvCtx) Value(key any) any           { return m.ctx.Value(key) }

func TestTypedNilLatchesTheOnce(t *testing.T) {
	var typedNil *myTool // typed nil: a non-nil interface value
	sub, err := llmagent.New(llmagent.Config{
		Name:  "sub",
		Model: &fakeLLM{},
		Tools: []tool.Tool{typedNil},
	})
	if err != nil {
		t.Fatal(err)
	}
	seq, err := sequentialagent.New(sequentialagent.Config{
		AgentConfig: agent.Config{Name: "seq", SubAgents: []agent.Agent{sub}},
	})
	if err != nil {
		t.Fatal(err)
	}
	live := seq.(interface {
		RunLive(ctx agent.InvocationContext) (agent.LiveSession, iter.Seq2[*session.Event, error], error)
	})

	call := func() (panicked bool) {
		defer func() {
			if r := recover(); r != nil {
				panicked = true
			}
		}()
		_, _, _ = live.RunLive(&mockInvCtx{a: seq, ctx: t.Context()})
		return false
	}

	p1, p2, p3 := call(), call(), call()

	state := llminternal.Reveal(sub.(llminternal.Agent))
	n := 0
	for _, tl := range state.Tools {
		if tl == nil {
			continue
		}
		name := func() (s string) {
			defer func() { recover() }()
			return tl.Name()
		}()
		if name == "task_completed" {
			n++
		}
	}
	t.Logf("panicked: call1=%v call2=%v call3=%v ; task_completed tools = %d", p1, p2, p3, n)
}
```

With the PR applied this logs `call1=true call2=false call3=false ; task_completed tools = 0`. On `main` it logs `call1=true call2=true call3=true`.

</details>

## Two places the tests do not reach

The `t == nil` branch this PR adds never executes in any test in the package, and it is the same line the typed-nil case above walks past, so the coverage gap and the defect are one thing rather than two.

Separately, deleting the entire "does it already have `task_completed`" scan and injecting unconditionally leaves the whole `sequentialagent` suite green. No test gives a sub-agent a pre-existing tool, so nothing checks that a user's own `task_completed` is left alone and that the instruction suffix is not appended. That gap predates your change — but the change makes it carry more weight, because `main` re-evaluates the check on every `RunLive` while this evaluates it once and freezes the answer for the agent's lifetime. One case with a sub-agent whose `Tools` already holds a `task_completed` tool would cover both that branch and the nil skip.

## Not for this PR

The injection write is still unsynchronized against readers that never pass through the guard. `tools_processor.go`, `instruction_processor.go`, the REST graph endpoint and the A2A agent-card builder all read `state.Tools` or `state.Instruction` with no ordering edge against it. Driving `RunLive` in one goroutine and the exported `adka2a.BuildAgentSkills` in another reports a race in 10 of 10 runs here — and in 10 of 10 runs on `main`, with the same warning count and the same reader/writer pair. So it is pre-existing, and your change strictly narrows it: [`main` writes on every call where the tool is absent](https://github.com/google/adk-go/blob/440e35202c325a7310557e3bacecdeb3ad07b989/agent/workflowagents/sequentialagent/agent.go#L153-L162), this writes at most once per agent. Worth its own issue next to #1490 and #1491 rather than growing this one.

## One correction to the issue

#1441's "Related" paragraph says #1137 and #1166 fixed this class of bug already. #1166 was closed without being merged and #1137 is still open, so the `llm_agent_wrapper.go` race is live on `main`. Nothing in this PR rests on that, but the issue reads as though there is precedent where there is none.

Last thing, purely for sequencing: #1462 also edits `internal/llminternal/agent.go`, so whichever of the two lands second will want a rebase.

tvost2:
Addressed in `47ecbd0`:

- **Typed nil latching the guard.** Replaced the `sync.Once` field with `llminternal.LiveToolInjection` — a `sync.Mutex` + `bool` where `done` is set only *after* the injection returns normally, so a panic (typed nil reaching `t.Name()`) propagates without latching and the next `RunLive` retries, loud on every call like `main`. The losing callers still get their happens-before edge (same mutex, acquired after the winner's unlock). `go vet ./...` clean — no copylocks hit.
- **Naming.** Field is now `State.LiveInjection`; the type `LiveToolInjection` and its doc name the invariant (one live-mode tool injection per agent, whoever performs it) and spell out that a second injector would share the latch, first write wins.
- **Comment wording.** Now says `llmagent.New` *assigns* the caller's `cfg.Tools` slice unvalidated, and points at #1490 for the aliasing.
- **Idempotence + nil-skip coverage.** New `TestSequentialAgent_RunLive_PreexistingToolAndNilSkip`: a sub-agent whose `Tools` already holds a `task_completed` tool (plus a `nil` element) must keep its own tool and gain no instruction suffix. Deleting the scan and injecting unconditionally now fails it on both `len(Tools)` and the suffix count.

`go test -race -count=3` (sequentialagent) and `-race -shuffle=on` (llminternal, llmagent, workflowagents/...) green; `go vet ./...` clean.

Noted the #1462 rebase sequencing on `internal/llminternal/agent.go` — will rebase whichever lands second.

internal/llminternal/agent.go
```diff
@@ -60,6 +62,45 @@ type State struct {
...
+// behaviour as re-deriving the decision each time — instead of going
+// silent after the first failure. If a second workflow agent ever adds its
+// own live-mode injection it shares this latch: first write wins, the rest
+// are skipped.
+type LiveToolInjection struct {
+	mu   sync.Mutex
+	done bool
+}
+
+// Do runs inject at most once for the lifetime of the receiver. It
```
karolpiotrowicz: "runs inject at most once for the lifetime of the receiver" is the one thing this method deliberately does not do, and the rest of the same sentence says so. `done` is set only on the normal return, so a panicking `inject` runs again on the next call — which is the whole point of the change, but a reader who stops at the first clause gets the opposite contract.

I drove this directly rather than through `RunLive`: an `inject` that writes and then panics, called twice, executes its body twice. Your caller is safe from that because its only write comes after its only panic point and a retry re-runs the scan, but the sentence is a promise to whoever writes the second caller.

"at most once successfully" would say what the code does.

internal/llminternal/agent.go
```diff
@@ -60,6 +62,45 @@ type State struct {
...
+// LiveToolInjection is a one-shot, panic-safe latch for the single
+// live-mode tool injection an agent may receive (see State.LiveInjection).
+// The first caller to Do performs the injection; concurrent callers block
+// and then observe it as a no-op, with a happens-before edge to the first
+// caller's writes. Unlike sync.Once it does not latch when the injection
+// panics, so a misconfigured sub-agent stays loud on every call — the same
+// behaviour as re-deriving the decision each time — instead of going
+// silent after the first failure. If a second workflow agent ever adds its
+// own live-mode injection it shares this latch: first write wins, the rest
+// are skipped.
```
karolpiotrowicz, line 85: This is exactly what I asked for last round and it reads well — naming the invariant rather than the payload, and stating the sharp edge instead of leaving it for the next person to find.

One thing to consider now that it is written down: "first write wins, the rest are skipped" means a future second injector's tool and instruction are dropped with no error. Nothing is wrong today — `sequentialagent.RunLive` is the only caller in the module — and I am not asking for a change here. If you ever do want to close it, keying the latch per injector rather than per agent costs little and does not conflict with the lifetime argument above it.

agent/workflowagents/sequentialagent/agent.go
```diff
@@ -149,18 +149,33 @@ func (a *sequentialAgent) RunLive(ctx agent.InvocationContext) (agent.LiveSessio
...
+			// runs it once per agent and gives every other RunLive caller a
+			// happens-before edge before they read Tools.
+			state.LiveInjection.Do(func() {
+				hasTaskCompleted := false
+				for _, t := range state.Tools {
+					// llmagent.New assigns the caller's cfg.Tools slice
+					// unvalidated (#1490), so a nil element can reach here;
+					// skip an untyped nil. A typed nil still panics in
+					// t.Name() below — on every call, as on main, because
+					// LiveInjection.Do does not latch on panic.
```
karolpiotrowicz: Small precision point: "a typed nil still panics in `t.Name()` below — on every call" holds only when `Name()` dereferences its receiver. A typed nil whose `Name()` returns a constant returns normally and gets scanned like any other tool.

The claim the paragraph actually needs is the second half — that when it does panic, it panics on every call rather than latching — and that part is exactly right. I did not exercise the non-dereferencing case; every probe I ran used a receiver-dereferencing `Name()`.

agent/workflowagents/sequentialagent/agent_test.go
```diff
@@ -451,6 +462,172 @@ func TestSequentialAgent_RunLive_Injection(t *testing.T) {
...
+	// seenByCaller[i][j]: what caller i saw in sub-agent j's Tools.
+	seenByCaller := make([][]int, concurrent)
+	var wg sync.WaitGroup
+	wg.Add(concurrent)
+	for i := range concurrent {
+		go func() {
+			defer wg.Done()
+			invCtx := &mockInvocationContext{agent: sequentialAgent, invocationID: "test_id", ctx: t.Context()}
+			_, _, _ = liveAgent.RunLive(invCtx)
+			// RunLive has returned, so every sub-agent's Once has completed:
```
karolpiotrowicz, line 543: Stale name: there is no `Once` any more. The neighbouring block comments above the test were updated to say `LiveInjection latch` and this one was missed.

The reasoning it states is still correct — every caller takes the same mutex, so winner or loser, the read after `RunLive` returns is ordered after the winner's writes.

karolpiotrowicz (changes requested):
The latch fix holds, and I checked it the way the last round's finding deserved. Driving `RunLive` on a sub-agent with a typed nil in `Tools`, three times under a recover, this head now panics on all three calls and injects nothing — byte-identical to `main`. Under the old `sync.Once` it panicked once and then handed back a live pipeline with no hand-off tool, silently, for that agent's life. I also ran a variant where one sub-agent is clean and the next holds the typed nil: the clean one is injected exactly once and is not re-injected across two further panicking calls, and a second goroutine calling in after the panic returns rather than blocking, so the mutex is released during unwinding. Recover, remove the bad tool, call again, and you get exactly one tool and one instruction suffix.

The race evidence is worth more than a green run this time. Stripping the mutex out of `Do` entirely makes the detector fire in 100 of 100 processes, and replacing `Do` with an atomic compare-and-swap — which keeps exactly-once but drops the ordering edge — still fails 100 of 100. So your test pins the happens-before guarantee and not merely the tool count. The head itself is clean at 100 of 100 with `GOMAXPROCS=1`, 100 of 100 at the default, and 30 of 30 oversubscribed. Current `main` grafted with your test file fails 100 of 100 with 300 race reports, so the defect is live on the branch you are targeting and not only on the merge-base.

The new `TestSequentialAgent_RunLive_PreexistingToolAndNilSkip` closes what I raised last round: deleting the whole idempotence scan and injecting unconditionally now fails 20 of 20 runs on all four of its assertions, where the same mutation left the package green before.

## One thing to add before this merges

The behavior this round exists to deliver has no test. I reverted `LiveInjection` from `LiveToolInjection` back to a plain `sync.Once` — reintroducing exactly the bug I blocked on last round — and ran the full `sequentialagent` package under `-race`, thirty separate processes: all green, no races, identical to the unmodified head. Nothing in the package makes `inject` panic, so nothing can tell the two implementations apart. There is no `internal/llminternal/agent_test.go` either, so the new type has no direct unit test at all.

That matters against [`AGENTS.md` line 61](https://github.com/google/adk-go/blob/47ecbd051f5947e3d314ea1bdafe1063236ab3fc/AGENTS.md#L61), which asks that a bug fix carry a test reproducing the bug. To be clear about what I am claiming: the shipped behavior is right, and the seven side-by-side comparisons above say so. What is missing is anything that would stop the next person reverting it. A test would need to demonstrate that an `inject` which panics leaves the guard unset and that a later call runs it again — a direct unit test on `LiveToolInjection` is the cheaper place to do that than through `RunLive`.

Separately, a second control is worth knowing about when you write it. Deleting just the `if l.done { return }` early return, keeping the mutex, also leaves the package green — the `hasTaskCompleted` scan inside `inject` re-derives the answer, so idempotence never depended on `done`. The field still earns its place by not re-running that scan under the lock on every call, but no test currently distinguishes it.

## The nil skip changes when you find out, and I got this partly wrong at first

I want to correct something before it reaches you second-hand. The untyped-nil skip means `RunLive` now returns cleanly where `main` panics at pipeline setup, and my first reading was that this hides a misconfiguration `main` catches. It does not. Building a plain `llmagent` with `Tools: []tool.Tool{nil}` and no sequential agent anywhere, then calling `adka2a.BuildAgentSkills` on it, panics on **both** this head and `main` — the nil is fatal downstream either way, and the root cause is `llmagent.New` taking `cfg.Tools` unvalidated, which is already #1490.

What is real is smaller: on the sequential-agent path the failure moves from setup to first use. `main`'s early panic came from the same unguarded `t.Name()` your skip removed, so it was never designed as validation. I am not asking you to change this, and one thing I could not settle is the other reader at `internal/llminternal/base_flow.go` line 776, which sits in a goroutine with no recover — reaching it needs a live client I had no way to stand up.

## Before merge, independent of anything above

No Go job has ever run on this pull request. The `Go` and `API diff` workflows are both sitting at `action_required` on every commit of the branch, waiting for a maintainer to approve workflow runs for a first-time contributor. Every number in this review came from my machine, on one host, with one toolchain, and my linter is a newer version than the one CI pins. Worth getting those green before anyone merges on the strength of my runs.

The branch is also behind `main`, and rebasing picks up an upstream `gofumpt` fix in `plugin/agentanalytics` that my local lint currently reports against your arm. #1462, #1135 and #1067 all touch these two files as well, so whichever lands second wants a rebase — and a conflict resolution that reintroduced a by-value copy of `State` would now trip `copylocks`, which it would not have before.

agent/workflowagents/sequentialagent/agent_test.go
```diff
@@ -438,16 +453,178 @@ func TestSequentialAgent_RunLive_Injection(t *testing.T) {
 	// After RunLive initiation, the sub-agents MUST have the task_completed tool injected!
```
karolpiotrowicz, line 453: This comment and the one at line 427 now both say the opposite of the code beneath them. Line 427 says sub-agents do not have the tool before `RunLive`, but line 413 deliberately gives sub-agent 1 one, and the failure message two lines down already says so. This line promises the tool MUST be injected, and the two assertions under it check that it was left alone and that the instruction suffix is absent.

Underneath the wording there is a coverage change. Before this commit the same block asserted that a toolless sub-agent gained `task_completed` — the only place in the package that checked the positive case in a single-threaded test. `subAgent2` is built at line 415 and wired in at line 420 but never revealed, so nothing here looks at the agent that should receive the injection. I checked what that costs by deleting the injection entirely: the only test that then fails is `TestSequentialAgent_RunLive_ConcurrentInjectionIsRaceFree`. The behavior is still covered, but by the twenty-goroutine race test alone, which is the test most likely to be quarantined if it ever turns flaky.

Revealing `subAgent2` and asserting exactly one `task_completed` in its `Tools` and exactly one instruction suffix would restore the positive branch and make this test cover both branches of the loop in one pass, which is what its two-sub-agent fixture is for.

agent/workflowagents/sequentialagent/agent_test.go
```diff
@@ -438,16 +453,178 @@ func TestSequentialAgent_RunLive_Injection(t *testing.T) {
...
+	// seenByCaller[i][j]: what caller i saw in sub-agent j's Tools.
+	seenByCaller := make([][]int, concurrent)
+	var wg sync.WaitGroup
+	wg.Add(concurrent)
+	for i := range concurrent {
+		go func() {
+			defer wg.Done()
+			invCtx := &mockInvocationContext{agent: sequentialAgent, invocationID: "test_id", ctx: t.Context()}
+			_, _, _ = liveAgent.RunLive(invCtx)
+			// RunLive has returned, so every sub-agent's Once has completed:
```
karolpiotrowicz, line 543: `Once` here is the `sync.Once` the round-3 fix removed. The mechanism is now `LiveModeInjection.Do`.

agent/workflowagents/sequentialagent/agent_test.go
```diff
@@ -438,16 +453,178 @@ func TestSequentialAgent_RunLive_Injection(t *testing.T) {
...
+			t.Errorf("sub-agent %d: task_completed instruction suffix appears %d times, want exactly 1", j, n)
+		}
+	}
+}
+
+// TestSequentialAgent_RunLive_PreexistingToolAndNilSkip pins the two
+// branches of RunLive's injection loop that no other test reaches: the
+// idempotence check that leaves a sub-agent's own task_completed tool (and
+// the instruction) alone, and the untyped-nil skip in the same loop.
+// Deleting the "does it already have task_completed" scan and injecting
```
karolpiotrowicz, line 579: Two things in this one comment. The claim that deleting the scan "keeps the whole package green" was true before this commit and is false now — giving sub-agent 1 a pre-existing tool in `TestSequentialAgent_RunLive_Injection` means scan removal fails that test too, which I confirmed by running the mutation. The part that is still true, and the part worth keeping, is that no other test reaches the untyped-nil skip.

Line 581 also still says `LiveInjection`, which the rename replaced with `LiveModeInjection`. It is the only occurrence left in the repo — the two sibling comments at lines 482 and 491 were updated.

agent/workflowagents/sequentialagent/agent.go
```diff
@@ -149,18 +149,34 @@ func (a *sequentialAgent) RunLive(ctx agent.InvocationContext) (agent.LiveSessio
...
 		if llmAgent, ok := subAgent.(llminternal.Agent); ok {
 			state := llminternal.Reveal(llmAgent)
-			hasTaskCompleted := false
-			for _, t := range state.Tools {
-				if t.Name() == "task_completed" {
-					hasTaskCompleted = true
-					break
+			// Reveal returns the sub-agent's shared, persistent State, so
+			// concurrent live sessions driving one agent tree would race
+			// this read-then-append on Tools/Instruction. state.LiveModeInjection
```
karolpiotrowicz, line 154: "runs it once per agent" is the wording you corrected on the type itself, where `Do`'s doc now reads "at most once successfully". This mirror kept the older phrasing, which is the one that is wrong on a panicking inject.

agent/workflowagents/sequentialagent/agent.go
```diff
@@ -149,18 +149,34 @@ func (a *sequentialAgent) RunLive(ctx agent.InvocationContext) (agent.LiveSessio
...
+					// llmagent.New assigns the caller's cfg.Tools slice
+					// without validating it (#1490), so a nil element can
+					// reach here; skip an untyped nil. A typed nil may panic
+					// in t.Name(), depending on its implementation. When it
+					// does, LiveModeInjection.Do does not latch the panic, so
+					// the next call retries the scan.
+					if t == nil {
+						continue
+					}
+					if t.Name() == "task_completed" {
```
karolpiotrowicz, line 169: `t.Name()` now runs inside the critical section, where on `main` it ran with no lock held. `tool.Tool` is a public interface, so this executes arbitrary user code under a library-internal, non-reentrant mutex. A `Name()` that blocks stalls every other `RunLive` on that sub-agent, and one that takes a lock another goroutine holds while entering `RunLive`, or that re-enters `RunLive` itself, deadlocks with no way out.

Nothing in the repo triggers it — every `Name()` implementation I checked returns a struct field. I am not asking you to restructure this, because hoisting the scan back out of the lock would reintroduce the unsynchronized read the change exists to remove. A sentence on `Do` saying that `inject` must not block indefinitely and must not re-enter `Do` on the same receiver would cover it, since callers cannot see the mutex.

karolpiotrowicz (commented):
Re-reviewed at [288f3d50](https://github.com/google/adk-go/commit/288f3d50a4225fd3c984995e3dc95fae0b540792). The rename is complete in code, both comment corrections are accurate, and the latch itself is untouched — I re-ran the concurrency checks rather than assume a rename is inert, at `GOMAXPROCS=1` and at the default, 100 separate processes each, and the head is clean at both. Those runs mean something: strip the mutex out of `Do` and the detector fires 100 times out of 100 at both settings, and an atomic compare-and-swap that keeps exactly-once while dropping only the ordering edge also fails 100 out of 100. Your test rejects the CAS, which is the property that makes it worth having.

## The one thing still outstanding is the one I raised last round

I reverted [`LiveModeToolInjection`](https://github.com/google/adk-go/blob/288f3d50a4225fd3c984995e3dc95fae0b540792/internal/llminternal/agent.go#L96-L104) to a plain `sync.Once` — reintroducing exactly the bug that blocked this at round 3 — and ran the full `sequentialagent` package under `-race`, a hundred separate processes at `GOMAXPROCS=1` and a hundred at the default. All green, no races, indistinguishable from the unmodified head. I also deleted the `done` field outright, keeping only the mutex, and that is invisible too. So nothing in the suite depends on the latch's state at all: exactly-once is carried by the `hasTaskCompleted` scan and race safety by the mutex.

That is what [`AGENTS.md` line 61](https://github.com/google/adk-go/blob/288f3d50a4225fd3c984995e3dc95fae0b540792/AGENTS.md#L61) asks for. To be precise about which half — the bug-reproduction clause is satisfied, because `ConcurrentInjectionIsRaceFree` reproduces #1441 and the controls above prove it discriminates. The clause that is unmet is "new behavior has tests", and it applies because `LiveModeToolInjection` is a new type whose entire reason to exist over `sync.Once` is one invariant that nothing enforces. The regression it fails to guard is the most likely edit anyone makes here, since `sync.Once` is the idiomatic answer and is literally what this branch shipped for its first three commits.

Rather than describe it again, here is the test. It needs a new `internal/llminternal/agent_test.go`, no sequential agent and no concurrency. I ran it both ways: it passes on your head and fails on the `sync.Once` revert with `after a panicking inject, Do ran inject 1 times in total, want 2`.

```go
package llminternal

import "testing"

func TestLiveModeToolInjection_DoDoesNotLatchOnPanic(t *testing.T) {
	var l LiveModeToolInjection
	calls := 0

	func() {
		defer func() {
			if r := recover(); r == nil {
				t.Errorf("first Do did not propagate the panic")
			}
		}()
		l.Do(func() { calls++; panic("boom") })
	}()
	if calls != 1 {
		t.Fatalf("first Do ran inject %d times, want 1", calls)
	}

	l.Do(func() { calls++ })
	if calls != 2 {
		t.Errorf("after a panicking inject, Do ran inject %d times in total, want 2: the latch must not be set by a panic", calls)
	}

	l.Do(func() { calls++ })
	if calls != 2 {
		t.Errorf("after a successful inject, Do ran inject %d times in total, want 2", calls)
	}
}
```

`go test -race -run TestLiveModeToolInjection ./internal/llminternal/`

## The test rewrite cost coverage it did not need to

Giving sub-agent 1 a pre-existing `task_completed` tool is a genuine addition and I would keep it. What went with it is the only single-threaded assertion in the package that injection happens at all — sub-agent 2 is built and wired in but never looked at, so after the rewrite the whole positive case rests on the twenty-goroutine race test. Details on the line, along with two comments that now say the opposite of the code beneath them.

## Two things I am not asking you to change

Concurrent readers of `state.Tools` outside live mode still take no lock. A non-live `runner.Run` reaching [`tools_processor.go` line 39](https://github.com/google/adk-go/blob/288f3d50a4225fd3c984995e3dc95fae0b540792/internal/llminternal/tools_processor.go#L39) can race a first `RunLive` appending to the same slice. This is not something your change caused and it does not widen it — on `main` every caller reads and writes unguarded, whereas here only the winner touches the slice, so the exposure is strictly smaller. What is worth a word is the [type's doc comment](https://github.com/google/adk-go/blob/288f3d50a4225fd3c984995e3dc95fae0b540792/internal/llminternal/agent.go#L66-L73), which says "every concurrent caller blocks until that has finished" and reads as a promise that `State.Tools` is lock-protected. It is not — the mutex is taken only by the injector. The mirror in `RunLive` is scoped correctly and does not have this problem.

Separately, and worth its own issue rather than anything here: [`llmagent.New` assigns `cfg.Tools` without copying](https://github.com/google/adk-go/blob/288f3d50a4225fd3c984995e3dc95fae0b540792/agent/llmagent/llmagent.go#L86), so two agents built from one caller slice with spare capacity share a backing array, and the injection into the first can overwrite the second's own `task_completed` tool in place. I reproduced it, and it reproduces identically on the merge-base, so it belongs with #1490 rather than with this branch.

## Before merge, independent of the above

Still no Go job has run on this pull request. `Go` and `API diff` both sit at `action_required`, waiting on a maintainer to approve the workflow run for a first-time contributor. What has completed is the changed-files check and the CLA check, and four security-scan jobs that were skipped. Every number above came from my machine on one host with one toolchain, my linter is a newer version than the one the workflow pins, and I did not use `-shuffle=on`, which the `Go` job does. Worth getting those green before anyone merges on the strength of my runs.

The branch is also behind `main`. #1135, #1067 and #1462 all touch these two files, so whichever lands second wants a rebase.

The rest is comment hygiene and sits on the lines below.

tvost2:
Addressed the remaining review request with a direct `llminternal` unit test for `LiveModeToolInjection`: a panicking injection is retried on the next call, while a successful injection runs only once. Validation: `go test ./internal/llminternal ./agent/workflowagents/sequentialagent -race -count=1` passes.

internal/llminternal/agent_test.go
```diff
@@ -0,0 +1,53 @@
...
+	injection := LiveModeToolInjection{}
+
+	for range 3 {
+		injection.Do(func() { calls++ })
+	}
+
+	if calls != 1 {
+		t.Fatalf("injection calls = %d, want 1", calls)
+	}
+}
```
karolpiotrowicz, line 53: This line has no trailing newline, which is what `gofumpt` is reporting. `gofmt -l .` names this file and nothing else on your head, and is silent on the merge-base, on the previous head and on `main`.

I ran the linter at the version the workflow pins rather than a newer one, since the two disagree about formatter findings in general:

```
$ golangci-lint version
golangci-lint has version 2.3.1

$ golangci-lint run ./...   # merge-base cf61116c
0 issues.

$ golangci-lint run ./...   # this head 118e3877
internal/llminternal/agent_test.go:53:1: File is not properly formatted (gofumpt)
}
^
1 issues:
* gofumpt: 1
```

Appending one newline byte and changing nothing else returns it to `0 issues.`

internal/llminternal/agent_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+import "testing"
+
+func TestLiveModeToolInjectionRetriesAfterPanic(t *testing.T) {
+	var calls int
+
+	for attempt := 0; attempt < 2; attempt++ {
+		func() {
+			defer func() {
+				if recover() == nil {
```
karolpiotrowicz, line 25: `recover() == nil` discards the value, so any panic from anywhere inside `Do` satisfies this — not specifically the `"injection failed"` the test raised. Comparing the recovered value against that string would keep the assertion honest if `Do` ever grows a panic of its own.

agent/workflowagents/sequentialagent/agent_test.go
karolpiotrowicz, line 453: This comment says the sub-agents must have `task_completed` injected, and the four assertions below it check the opposite: that sub-agent 1's pre-existing tool was left alone and that no instruction suffix was appended. Both readings are right for their own fixture, but the comment is now describing the test this used to be.

`subAgent2` a few lines up is the only sub-agent here that would actually receive an injection, and nothing looks at it after `RunLive` returns. Asserting on it would make the test match its name again, and would put the single-threaded positive case somewhere other than the 20-goroutine race test.

agent/workflowagents/sequentialagent/agent_test.go
karolpiotrowicz, line 543: Stale after the rename two commits ago: there is no `Once` any more, the field is `LiveModeInjection` and the type is `LiveModeToolInjection`.

agent/workflowagents/sequentialagent/agent_test.go
```diff
@@ -438,16 +453,178 @@ func TestSequentialAgent_RunLive_Injection(t *testing.T) {
...
+// unconditionally otherwise keeps the whole package green, because the
+// LiveInjection guard fires exactly once either way.
```
karolpiotrowicz, line 581: Two things here are no longer true. `LiveInjection` names a type that the rename removed. And deleting the idempotence scan does not keep the package green: `TestSequentialAgent_RunLive_Injection` also fails, because sub-agent 1 starts with a `task_completed` tool and its `Tools` would grow from 1 to 2.

A comment that describes coverage is the thing a later reader trusts instead of checking, so it is worth either correcting or dropping.

karolpiotrowicz (changes requested):
Re-checked at [118e3877](https://github.com/google/adk-go/commit/118e3877a2e2007adcc4bef012f82ca08a307f7a). The new test file is the right idea in the right place, and one of its two tests earns its keep immediately: delete the `done` field from the latch, keeping the mutex, and `TestLiveModeToolInjectionRunsOnceAfterSuccess` fails 100 times out of 100. That mutation was invisible to the entire repository before this commit. Nothing else catches it.

The other one does not yet work, and it is the one that matters.

## The retry test cannot fail for the property it is named for

[`injection := LiveModeToolInjection{}`](https://github.com/google/adk-go/blob/118e3877a2e2007adcc4bef012f82ca08a307f7a/internal/llminternal/agent_test.go#L29) is declared inside the `for attempt` loop, so attempt 1 and attempt 2 drive two different zero-valued latches. The test never calls `Do` on a receiver whose previous `Do` panicked, which is the only operation that distinguishes this type from `sync.Once`.

I measured it rather than reasoning about it. I replaced the latch with a plain `sync.Once` — reintroducing exactly the bug that blocked this at round 3 — and ran the whole repository's test suite: 77 packages green, the only failure a pre-existing one that fails on `main` too. Then the two touched packages under `-race`, a hundred separate processes at `GOMAXPROCS=1` and a hundred at the default: zero failures, zero races, indistinguishable from your unmodified head. I also tried an atomic compare-and-swap that keeps exactly-once but latches through a panic, and `TestLiveModeToolInjectionRetriesAfterPanic` passes on that too.

Hoisting that one declaration above the loop, changing nothing else, fixes it. On your head it still passes. On the `sync.Once` revert and on the compare-and-swap variant it fails with `Do() did not panic on attempt 2` and `injection calls = 1, want 2`.

<details><summary>The one-line change, and how to run it</summary>

```go
func TestLiveModeToolInjectionRetriesAfterPanic(t *testing.T) {
	var calls int
	injection := LiveModeToolInjection{}

	for attempt := 0; attempt < 2; attempt++ {
		func() {
			defer func() {
				if recover() == nil {
					t.Errorf("Do() did not panic on attempt %d", attempt+1)
				}
			}()
			injection.Do(func() {
				calls++
				panic("injection failed")
			})
		}()
	}

	if calls != 2 {
		t.Fatalf("injection calls = %d, want 2", calls)
	}
}
```

```
go test -race -count=1 -run TestLiveModeToolInjection ./internal/llminternal/
```

Worth adding a third call afterwards that does not panic, asserting `calls` reaches 3 and then stays at 3, so the same test also pins that a successful inject latches.

</details>

## The lint gate goes from green to red

The new file ends `}` with no trailing newline, so `gofmt -l` names it and `golangci-lint run ./...` exits 1 with `File is not properly formatted (gofumpt)`. I ran this at [v2.3.1](https://github.com/google/adk-go/blob/118e3877a2e2007adcc4bef012f82ca08a307f7a/.github/workflows/go.yml#L91), the binary the workflow pins rather than a newer one: the merge-base reports `0 issues.` and your head exits 1. Appending one newline byte returns it to `0 issues.` Detail is on the line itself.

To be exact about the scope, since it changes what this means: the root module is what moves from clean to failing. `plugin/agentanalytics` already reports two issues at the merge-base, so the repository as a whole was not clean before this either.

## Smaller things

The test named `_Injection` still does not test injection — sub-agent 2 is built, wired in, and never looked at, so every assertion after `RunLive` is that nothing changed. I said last round that this left the positive case uncovered and that was too strong: [`TestSequentialAgent_RunLive_ConcurrentInjectionIsRaceFree`](https://github.com/google/adk-go/blob/118e3877a2e2007adcc4bef012f82ca08a307f7a/agent/workflowagents/sequentialagent/agent_test.go#L554-L571) does assert one tool and one instruction suffix for each of three sub-agents, so deleting the injection reddens the package. What is left is that the coverage lives in a concurrency test rather than a plain one, and the comment above the assertions still says the opposite of what they check.

The description still describes the first design — a package-level `sync.Map` of `*sync.Once` keyed by the sub-agent state — which was replaced two commits ago by the per-state latch. Its Testing Plan also says `gofmt` is clean on both touched files, and there are four now. Worth a pass before merge, since that text outlives the review thread.

## Before merge, unchanged from last round

Still no Go job has ever run on this pull request. `Go` and `API diff` both sit at `action_required`, waiting for a maintainer to approve the workflow run for a first-time contributor. Every number above came from my machine on one host, and I did not use `-shuffle=on`, which the `Go` job does.

The branch is behind `main`, and #1510, #1067 and #318 all report conflicts with it, so whichever lands second wants a rebase.

PR #1117, opened 2026-07-02
ci: run Go workflow on main and v1 branches

wolo-lab:
Adds checks for v1 branch

PR #1445, opened 2026-08-31
ci: run the Go workflow on every pull request

wolo-lab:
## Problem

The `pull_request` trigger in `.github/workflows/go.yml` only fires when a PR targets a branch on an allow-list. Stacked PRs target the branch below them rather than `main`, so no build, test or lint job ever runs against a stack, and a red linter can survive a whole review round unnoticed. #1231 worked around this by adding `baptmont/**` to the list, which fixes it for one contributor's branches.

The allow-list was never a deliberate choice. It arrived as `branches: [ $default-branch ]` from GitHub's Go starter template when the workflow was first added in #3, and every edit since has changed only the names inside it: `main` to `v2` in #937, `v2` to `v1` in #1117, then `baptmont/**` in #1231.

Keeping it means listing every prefix the team uses. That is currently eight schemes on `origin` (`wolo/`, `baptmont/`, `yarolegovich/`, `fix/`, `docs/`, `chore/`, `test/`, `backup/`) plus branches with no prefix at all, such as `checkpointing` and `groot-runner`, which can equally be the base of a stack. Each new contributor needs another edit, and until someone makes it their stacked PRs run no checks while still reporting green.

## Summary

Removes the `branches` filter from the `pull_request` trigger, so the workflow runs on every PR whatever its base branch, and replaces the comment with the reason not to reintroduce one.

The `push` trigger keeps `main` and `v1`. A push to a personal branch with an open PR already fires `pull_request`, so unfiltering `push` too would only run every job twice.

This does not widen the abuse surface. The filter matched base branches rather than authors, so a junk PR from a fork targeting `main` was always allowed through. Limiting who can trigger runs from forks is a repository setting, not something a branch filter can express.

Checked with `actionlint` (no findings) and by parsing the result: `pull_request` is now null, which means the default `opened`, `synchronize` and `reopened` types on any base branch.


jjsasha63 (approved):
lgtm

PR #1449, opened 2026-09-01
fix(openaimodel): support optional function call output IDs

### LINKED ISSUE #1435: [2.3.0]cannot use callID (variable of type string) as param.Opt[string] value in struct literal
go/pkg/mod/google.golang.org/adk/v2@v2.3.0/model/openaimodel/request.go:302:11: cannot use callID (variable of type string) as param.Opt[string] value in struct literal

ktsoator:

- Closes: #1435

**Problem:**

ADK Go v2.3.0 requires `openai-go/v3` v3.52.0, but Go module version selection can resolve v3.54.0 when another dependency requires it.

In v3.54.0, `ResponseInputItemFunctionCallOutputParam.CallID` changed from `string` to `param.Opt[string]`. This causes ADK Go consumers using the newer SDK version to fail compilation in `model/openaimodel/request.go`.

**Solution:**

- Upgrade `openai-go/v3` to v3.54.0.
- Wrap the tracked function call output ID with `param.NewOpt`.
- Update the existing request conversion test to verify that the optional ID is populated and matches the originating function call.

No public API changes are introduced.

### Testing Plan

- [x] `go test -race -count=1 ./model/openaimodel`
- [x] `go test -race -mod=readonly -count=1 -shuffle=on work`
- [x] `golangci-lint run` in the root module
- [x] `golangci-lint run` in `plugin/agentanalytics`
- [x] `go mod tidy -diff` in both modules


wojtas000 (approved):
LGTM, thanks

PR #1451, opened 2026-09-01
fix(openaimodel): preserve caller-owned JSON schemas

### LINKED ISSUE #1450: model/openaimodel mutates caller-provided ResponseJsonSchema
## Required Information

**Describe the Bug:**

I noticed that `model/openaimodel` mutates a caller-provided
`GenerateContentConfig.ResponseJsonSchema` when the schema is represented as a
`map[string]any`.

During request conversion, `normalizeSchema` returns the original map without
copying it. `enforceStrictOpenAISchema` then modifies that map recursively to
make it compatible with OpenAI strict structured outputs.

The mutations can include:

- adding `additionalProperties: false`
- replacing the `required` property
- modifying nested object schemas
- deleting sibling properties from schemas containing `$ref`

This means request conversion changes configuration owned by the caller. Reusing
the same schema for another request or model may therefore use a different schema
than the caller originally supplied.

The same map may also remain shared through the normal `llmagent` path because
`ResponseJsonSchema` is an `any` field and the current reflection-based clone
does not recursively copy interface values. Concurrent requests that reuse the
same agent configuration may consequently write to the same map.

**Steps to Reproduce:**

1. Create a `map[string]any` JSON schema.
2. Assign it to `GenerateContentConfig.ResponseJsonSchema`.
3. Call `GenerateContent` using `openaimodel`.
4. Inspect the original schema after request conversion.

**Expected Behavior:**

Request conversion should not modify caller-owned configuration. Any
OpenAI-specific strict-schema transformations should be applied to an internal
copy.

**Observed Behavior:**

The original schema contains new or modified properties after
`GenerateContent` is called. For example, `additionalProperties` and `required`
are added to the caller's map.

**Environment Details:**

- ADK Library Version: v2.3.0 and current `main`
- OS: macOS
- Go Version: 1.26.6

**Model Information:**

- N/A. The mutation occurs while building the request, before an HTTP request is
  sent to a model.

---

## Optional Information

**Regression:**

No known previously working version. The behavior appears to have been present
since the initial `openaimodel` implementation.

**Logs:**

N/A.

**Screenshots / Video:**

N/A.

**Additional Context:**

The alias originates from this path:

```go
case map[string]any:
	return s, nil
```

The returned map is subsequently passed to `enforceStrictOpenAISchema`, which
modifies it in place.

A possible fix would be to create an independent map representation before
applying the strict-schema transformations. A regression test could retain a
copy of a nested input schema, build the OpenAI request, and verify that the
original schema remains unchanged.

```go
package main

import (
	"context"
	"fmt"

	"google.golang.org/adk/v2/model/openaimodel"
	"google.golang.org/genai"
)

func main() {
	ctx := context.Background()

	llm, err := openaimodel.NewModel(ctx, "gpt-4o-mini",
		&openaimodel.ClientConfig{APIKey: "not-used"})
	if err != nil {
		panic(err)
	}

	schema := map[string]any{
		"type": "object",
		"properties": map[string]any{
			"name": map[string]any{
				"type": "string",
			},
		},
	}

	req := &model.LLMRequest{
		Contents: []*genai.Content{
			genai.NewContentFromText("test", genai.RoleUser),
		},
		Config: &genai.GenerateContentConfig{
			ResponseMIMEType:   "application/json",
			ResponseJsonSchema: schema,
		},
	}

	// Request conversion occurs before the returned sequence is consumed.
	_ = llm.GenerateContent(ctx, req, false)

	fmt.Printf("%#v\n", schema)
	// The original map now contains:
	// "additionalProperties": false
	// "required": []string{"name"}
}
```

- Always (100%)

ktsoator:

- Closes: #1450

**Problem:**

When `ResponseJsonSchema` is provided as a `map[string]any`,
`normalizeSchema` returns the caller's map directly. The subsequent
`enforceStrictOpenAISchema` call modifies that map recursively by adding strict
schema fields, replacing `required`, and removing `$ref` siblings.

These OpenAI-specific transformations therefore leak into caller-owned
configuration. The map may be reused across requests, and it can remain shared
through the normal `llmagent` configuration clone because it is stored in an
interface field.

Concurrent requests sharing that configuration may write to the same map,
which can terminate the process with a concurrent map write. A cyclic schema
reachable through `properties`, `$defs`, or `items` can also send the recursive
strictification into a fatal stack overflow.

**Solution:**

Normalize every non-nil JSON schema through JSON encoding and decoding. This
produces an independent nested map before applying the OpenAI strict-schema
transformations.

The JSON round trip also detects cyclic schemas before strictification and
returns a normal marshal error instead of recursing until the process
overflows its stack.

Decoding normalizes nested Go representations, including struct subschemas and
`[]map[string]any` values, into plain maps and slices. This allows strict-schema
transformations to apply consistently regardless of how the caller represented
the schema, including nodes that the previous by-reference path skipped.

Decode schema numbers without converting them through `float64`, then preserve
numeric constraints as raw JSON for the OpenAI SDK request serializer. This
keeps large integer constraints exact and prevents them from being encoded as
JSON strings.

Add regression tests using nested objects, `$defs`, and `$ref`, as well as an
integer constraint above `2^53`. The tests confirm that the request schema is
still transformed for strict output, the caller-provided schema remains
unchanged, and the final OpenAI request preserves the integer exactly.

### Testing Plan

**Unit Tests:**

- [x] Added regression coverage for caller-owned response schemas.
- [x] Added request-level coverage for JSON schema integers above `2^53`.
- [x] All tests pass locally.

```text
go test -race -count=1 ./model/openaimodel/
ok google.golang.org/adk/v2/model/openaimodel

all packages passed
```

**Additional Verification:**

```text
go build -mod=readonly work
passed

golangci-lint v2.3.1 run
0 issues in both modules

go mod tidy -diff
no differences in either module
```

**Manual End-to-End Tests:**

N/A. The affected behavior is deterministic request conversion performed before
any HTTP request is sent. It is covered directly by the regression tests.

### Checklist

- [x] I have performed a self-review of the change.
- [x] I have added a test that proves the fix is effective.
- [x] New and existing tests pass locally.
- [x] Manual E2E testing is not applicable to this request-conversion change.
- [x] There are no dependent downstream changes.


karolpiotrowicz (commented):
Re-checked at [65257037](https://github.com/google/adk-go/commit/652570371ad06e7b63d49ce084cf9530f3186ec9). The precision concern I had on the earlier revision is closed, and the way you closed it is the interesting part.

I had measured that `json.Decoder` + `UseNumber()` makes the wire *worse* rather than better: the OpenAI SDK serializes request params with its own `apijson` encoder rather than `encoding/json`, and that encoder emits a `json.Number` as a quoted string, so `"maximum": 9223372036854775807` becomes `"maximum": "9223372036854775807"` and the keyword stops being a number. Converting to [`json.RawMessage`](https://github.com/google/adk-go/blob/652570371ad06e7b63d49ce084cf9530f3186ec9/model/openaimodel/request.go#L461-L464) after the decode defeats exactly that, and it holds at the wire. Capturing the real outbound request body for a caller-built `map[string]any` schema:

```
                              const                 maximum
merge-base                    9007199254740993      9223372036854775807
previous revision             9007199254740992      9223372036854776000   <- rounded, and past MaxInt64
this revision                 9007199254740993      9223372036854775807   <- exact, unquoted
```

The strictification still reaches the request — the captured body carries `"additionalProperties":false` and `"required"` — so the copy is transformed and the caller's map is not. Both halves are pinned: commenting out the `preserveSchemaNumbers` call fails [request_test.go:824](https://github.com/google/adk-go/blob/652570371ad06e7b63d49ce084cf9530f3186ec9/model/openaimodel/request_test.go#L824) and puts the quoted string back on the wire, which is the gap I flagged last time, now closed by your own test.

The aliasing fix holds too, measured rather than read. Against a nested object schema carrying no `additionalProperties` or `required` — the shape `enforceStrictOpenAISchema` actually rewrites — the merge-base mutates the caller's map in place and this revision leaves it byte-identical.

## Two things this fixes that the description does not claim

Both reproduce on the merge-base and are worth a line, because someone reading the description will not know to credit them.

- **A cyclic caller schema kills the process on the merge-base.** A `map[string]any` whose `properties` contains itself sends `enforceStrictOpenAISchema` into unbounded recursion and the process dies with `fatal error: stack overflow`. Here `json.Marshal` detects it and you get `openai: marshal json schema: json: unsupported value: encountered a cycle`. The cycle has to run through `properties`, `$defs` or `items` to reach the recursion — a self-reference under any other key is caught later by the SDK instead.
- **Concurrent requests sharing one config race on that same in-place write.** A concurrent map write is an unrecoverable runtime throw rather than a panic, so `recover` does not catch it and the whole agent binary exits, not just the offending request.

## Smaller things

Nothing here blocks.

No test in `model/openaimodel` starts a goroutine, so the package's own `-race` run cannot see the concurrency half of what this fixes. It passes equally before and after, which is worth knowing before anyone reads a green `-race` as coverage of it.

Removing the by-reference path means `enforceStrictOpenAISchema` now reaches nodes it used to skip, since a subschema supplied as a Go struct or an `anyOf` given as `[]map[string]any` arrives as a plain map. That is the transformation finally applying however the caller spelled the schema, and it matches what the existing table test asserts for map-shaped input, but it is wire-visible and deserves a sentence in the description.

One note beyond this PR, for whoever picks it up rather than for you. #1450 identifies the general cause as the reflection-based clone not copying interface values, and that part is still open: `deepCopy` in `internal/llminternal/basic_processor.go` handles struct, slice, map and pointer and falls through to a shallow assignment for everything else, with no `reflect.Interface` case. Your change protects both openaimodel paths through `normalizeSchema`, so nothing is exposed today, but the framework-level gap remains for the next `any`-typed config field. I am filing that separately so it does not widen this PR.

ktsoator:
Thank you for the thorough re-check and for validating the actual outbound
request representation.

I've updated the PR description to cover the cyclic-schema failure mode, the
risk to concurrent requests sharing one configuration, and the consistent
strictification of normalized nested Go schema representations.

I also agree that the missing `reflect.Interface` handling in `deepCopy` is a
broader framework-level concern and should remain separate from this PR. Since
there are no requested code changes in this review, I have kept the current
implementation unchanged.

Thank you again for the detailed validation and feedback.

wojtas000 (approved):
Thanks for the fix, I re-measured it locally and it works. 
Missing `reflect.Interface` handling in `deepCopy` should indeed be handled separately. I will raise the issue for that.

LGTM

PR #1453, opened 2026-09-01
test(adkrest): add RunLiveHandler WebSocket integration coverage

wmh123wmh1:
## Summary

This PR addresses item 3 from #1153 only.

- Add WebSocket integration coverage for `RunLiveHandler`.
- Verify the handshake and server event streaming contract.
- Verify forwarding of text messages and binary audio frames.
- Verify live-session cleanup after normal and abrupt client disconnects.
- Verify the `RunLive` error close frame and close-frame drain behavior.
- Log the previously silent `ws.WriteJSON` failure path.

The tests use a mock live agent/session and do not require a real model connection. The WebSocket protocol behavior is unchanged.

## Testing

- `go test -race ./server/adkrest/controllers -run "^TestRunLiveHandler_" -count=50`
- `go test -race -mod=readonly -count=1 -shuffle=on ./server/adkrest/...`
- `go vet ./server/adkrest/...`
- `golangci-lint v2.3.1 run` (root module and `plugin/agentanalytics`)

The full workspace test was also attempted. `session/vertexai` requires Agent Engine replay fixtures or external GCP access, which was unavailable in the local environment; all other workspace packages passed with the same race-enabled test configuration.

Refs #1153


karolpiotrowicz (commented):
This takes `RunLiveHandler` from having no integration test at all to exercising the handshake, event streaming, both frame directions, and both disconnect shapes, and the tests genuinely discriminate: removing the reader goroutine's `liveSession.Close()`, corrupting the hardcoded audio MIME type, or changing the close code each makes them fail. The harness composes the handler exactly as production does, and saving and restoring `log.Writer()` rather than hardcoding `os.Stderr` is a nicer pattern than the existing precedent at [`triggers/config_test.go:156-157`](https://github.com/google/adk-go/blob/ecf31b12b3044e94e71e0bb5c1e8fef09aa0f1e4/server/adkrest/controllers/triggers/config_test.go#L156-L157). Nothing here blocks merging.

Four things worth looking at, in order of how much they matter.

## The added log line fires on ordinary disconnects

[`runtime.go:468`](https://github.com/google/adk-go/blob/ecf31b12b3044e94e71e0bb5c1e8fef09aa0f1e4/server/adkrest/controllers/runtime.go#L468) logs whenever `ws.WriteJSON` fails, and a perfectly normal client hangup is one of the ways it fails. When a client sends a `CloseNormalClosure` frame while the agent is still streaming, gorilla's close handler replies, the connection moves to "close sent", and the next write fails. The read half of the same handler filters exactly this case at [`runtime.go:409`](https://github.com/google/adk-go/blob/ecf31b12b3044e94e71e0bb5c1e8fef09aa0f1e4/server/adkrest/controllers/runtime.go#L409), so after this change the two halves disagree about what counts as an error.

The scale is modest — one line per affected connection, and none at all if the stream is idle when the client closes, since `break` exits the loop immediately. So this is log noise on a common path rather than anything that breaks.

One constraint on fixing it, because it caught me out: mirroring the `IsCloseError` check from line 409 does **not** suppress this. The error is `websocket.ErrCloseSent`, a plain sentinel value rather than a `*websocket.CloseError`, so `IsCloseError` never matches it. An `errors.Is(err, websocket.ErrCloseSent)` test does suppress it, and I confirmed all six of your new tests still pass with that in place, so genuine write failures are still reported.

<details><summary>Reproduction — driven through the real handler, 1 line with the change, 0 without it</summary>

```go
// in package controllers, using this PR's own helpers
func TestGracefulCloseLogsWriteError(t *testing.T) {
	var logs synchronizedBuffer
	orig := log.Writer()
	log.SetOutput(&logs)
	t.Cleanup(func() { log.SetOutput(orig) })

	liveSession := newRecordingLiveSession()
	stop := make(chan struct{})
	conn, handlerDone := dialRunLiveHandler(t, func(agent.InvocationContext) (agent.LiveSession, iter.Seq2[*session.Event, error], error) {
		return liveSession, func(yield func(*session.Event, error) bool) {
			for {
				select {
				case <-stop:
					return
				default:
				}
				if !yield(makeEvent("inv-1", testLiveAppName, "tick"), nil) {
					return
				}
				time.Sleep(time.Millisecond)
			}
		}, nil
	})

	// A textbook graceful client shutdown.
	if err := conn.WriteControl(
		websocket.CloseMessage,
		websocket.FormatCloseMessage(websocket.CloseNormalClosure, "client finished"),
		time.Now().Add(time.Second),
	); err != nil {
		t.Fatalf("WriteControl(close) failed: %v", err)
	}
	<-handlerDone
	close(stop)

	if got := logs.String(); strings.Contains(got, "WebSocket write error for app") {
		t.Errorf("graceful close produced a write-error log line: %q", got)
	}
}
```

`go test -count=5 -run TestGracefulCloseLogsWriteError ./server/adkrest/controllers`

Fails 5/5 on this branch with `WebSocket write error for app testApp: websocket: close sent`, and passes 5/5 with only the new line removed.

</details>

## The close-frame drain is described as covered but is not

`TestRunLiveHandler_RunLiveErrorSendsCloseFrame` carries a comment at [`runtime_live_test.go:355-356`](https://github.com/google/adk-go/blob/ecf31b12b3044e94e71e0bb5c1e8fef09aa0f1e4/server/adkrest/controllers/runtime_live_test.go#L355-L356) saying the handler drains the client's close reply before returning. Deleting the entire drain loop at [`runtime.go:367-371`](https://github.com/google/adk-go/blob/ecf31b12b3044e94e71e0bb5c1e8fef09aa0f1e4/server/adkrest/controllers/runtime.go#L367-L371) leaves all six new tests green, so nothing checks it. `waitForHandlerExit` confirms the handler returns, which it does either way. This one stands out because the drain is one of the four things #1153 asked for by name, and the comment reads as a coverage claim.

## Two protocol paths #1153 named are still unexercised

**The `RealtimeInput` variants.** [`runtime.go:440-449`](https://github.com/google/adk-go/blob/ecf31b12b3044e94e71e0bb5c1e8fef09aa0f1e4/server/adkrest/controllers/runtime.go#L440-L449) maps `activityStart`, `activityEnd` and `blob` out of a text frame, and none of the three is reached. Swapping the `activityStart` and `activityEnd` assignments, or removing the `blob` branch entirely, fails no test. `ForwardsTextMessages` asserts `RealtimeInput == nil`, so it is structurally unable to catch this, and `ForwardsBinaryAudio` covers only the raw-binary path.

**The mid-stream error path.** `TestRunLiveHandler_RunLiveErrorSendsCloseFrame` covers `RunLive` itself returning an error. The separate case where the iterator *yields* an error mid-stream, at [`runtime.go:460-463`](https://github.com/google/adk-go/blob/ecf31b12b3044e94e71e0bb5c1e8fef09aa0f1e4/server/adkrest/controllers/runtime.go#L460-L463), is not covered — deleting its close-frame write changes nothing. The two paths both emit `CloseInternalServerErr`, which makes them easy to conflate.

## Smaller things

- **The handshake status check cannot fail.** At [`runtime_live_test.go:145-147`](https://github.com/google/adk-go/blob/ecf31b12b3044e94e71e0bb5c1e8fef09aa0f1e4/server/adkrest/controllers/runtime_live_test.go#L145-L147), `Dial` already returns `ErrBadHandshake` for any non-101 response and the test fatals on that three lines earlier, so the `StatusSwitchingProtocols` comparison is unreachable.
- **`recordingLiveSession.Send` can wedge a goroutine on a failing run.** The unguarded send at [`runtime_live_test.go:67`](https://github.com/google/adk-go/blob/ecf31b12b3044e94e71e0bb5c1e8fef09aa0f1e4/server/adkrest/controllers/runtime_live_test.go#L67) uses a capacity-1 channel. Every test today sends exactly one request so it cannot fire, but if `waitForLiveRequest` ever times out the handler's reader goroutine blocks in `Send` forever rather than in `ReadMessage`, so closing the socket will not release it.
- **The two-goroutine lifecycle is not really under test yet.** In all six tests one side is parked while the other runs: the tests that stream never send a client frame, and the tests that send a frame park the iterator on `<-liveSession.closed`. A green `-race` therefore says less about that interaction than it looks like it does. A test whose iterator yields in a loop while the client concurrently writes binary frames would close the gap.

One note on the timeouts: every wait in the new file is a bare `time.Second`, and `waitForHandlerExit`'s budget is the same 1s as the handler's own drain deadline at [`runtime.go:366`](https://github.com/google/adk-go/blob/ecf31b12b3044e94e71e0bb5c1e8fef09aa0f1e4/server/adkrest/controllers/runtime.go#L366). I ran the suite 150 times including `-cpu=1 -count=50` under `-race` without a single failure, so I could not turn this into a real flake and I am not claiming it is one. Raising the waits to 5s would cost nothing if you are touching the file anyway.

wmh123wmh1:
Addressed the main review comments:

- ignored `websocket.ErrCloseSent` for normal client disconnects while preserving real write errors;
- added coverage for the close-frame drain path;
- added tests for `activityStart`, `activityEnd`, and `blob` realtime inputs;
- added coverage for errors yielded by the live event iterator;
- updated the PR branch with these fixes.

The focused `server/adkrest` tests and race tests pass.

server/adkrest/controllers/runtime_live_test.go
```diff
@@ -0,0 +1,526 @@
...
+	select {
+	case <-handlerDone:
+	case <-time.After(time.Second):
+		t.Fatal("RunLiveHandler did not exit")
+	}
+}
+
+func readCloseError(t *testing.T, conn *websocket.Conn) *websocket.CloseError {
+	t.Helper()
+	_, _, err := conn.ReadMessage()
```
karolpiotrowicz, line 171: This read lost the deadline it used to have. In [`ecf31b12`](https://github.com/google/adk-go/blob/ecf31b12b3044e94e71e0bb5c1e8fef09aa0f1e4/server/adkrest/controllers/runtime_live_test.go#L341-L343) the same `conn.ReadMessage()` was preceded by `conn.SetReadDeadline(time.Now().Add(time.Second))`. The extracted helper kept the read and dropped the deadline, and there is now no `SetReadDeadline` anywhere in the file. [`conn.ReadJSON` at line 462](https://github.com/google/adk-go/blob/244ed368b378aa0a8fd199b66dd9d3d5d7f7574a/server/adkrest/controllers/runtime_live_test.go#L462) is undeadlined for the same reason.

It matters when the server holds the socket open without sending a close frame. Delete both [`runtime.go:365` and `:366`](https://github.com/google/adk-go/blob/244ed368b378aa0a8fd199b66dd9d3d5d7f7574a/server/adkrest/controllers/runtime.go#L365-L366) and the suite hangs until `go test` times out — 45 s with a shortened timeout, ten minutes by default — and the panic destroys every other result in the package. The identical two-line change against [`ecf31b12`](https://github.com/google/adk-go/blob/ecf31b12b3044e94e71e0bb5c1e8fef09aa0f1e4/server/adkrest/controllers/runtime_live_test.go#L341-L343) fails cleanly in 1.00 s with a normal assertion failure. It takes two simultaneous faults to reach, so this is not urgent, but restoring the one line puts the guard back.

```
go test -count=1 -run '^TestRunLiveHandler_' ./server/adkrest/controllers
# with runtime.go:365-366 removed
#   this revision:  panic: test timed out
#   ecf31b12:       --- FAIL: TestRunLiveHandler_RunLiveErrorSendsCloseFrame (1.00s)
```

server/adkrest/controllers/runtime_live_test.go
```diff
@@ -0,0 +1,526 @@
...
+	case <-time.After(50 * time.Millisecond):
+	}
+	if err := conn.WriteControl(
+		websocket.CloseMessage,
+		websocket.FormatCloseMessage(closeErr.Code, "client acknowledged"),
+		time.Now().Add(time.Second),
+	); err != nil {
+		t.Fatalf("WriteControl(close reply) failed: %v", err)
+	}
+	waitForHandlerExit(t, handlerDone)
```
karolpiotrowicz: The assertion pins a latency window rather than the thing the test is named for. `sendClose` arms an absolute one-second read deadline at [`runtime.go:366`](https://github.com/google/adk-go/blob/244ed368b378aa0a8fd199b66dd9d3d5d7f7574a/server/adkrest/controllers/runtime.go#L366), and this exit budget closes at about the same moment, so a handler that waits the right length of time without ever reading the reply still passes. Replacing the whole drain loop at [`runtime.go:367-371`](https://github.com/google/adk-go/blob/244ed368b378aa0a8fd199b66dd9d3d5d7f7574a/server/adkrest/controllers/runtime.go#L367-L371) with `time.Sleep(200 * time.Millisecond)` leaves the suite green.

What the drain buys over a wait is consuming the peer's bytes, so that `close()` sends a FIN instead of an RST that would discard the close frame. Measuring elapsed time from the reply write and requiring exit well inside the production deadline would pin that. Deleting the drain outright is already caught, so this is a refinement rather than a gap.

server/adkrest/controllers/runtime_live_test.go
```diff
@@ -0,0 +1,526 @@
...
+	select {
+	case <-closeReceived:
+	case <-time.After(time.Second):
+		t.Fatal("client did not receive server close frame")
+	}
+
+	select {
+	case <-handlerDone:
+		t.Fatal("RunLiveHandler returned before receiving the close reply")
+	case <-time.After(50 * time.Millisecond):
```
karolpiotrowicz: This 50 ms is coupled to the one-second deadline at [`runtime.go:366`](https://github.com/google/adk-go/blob/244ed368b378aa0a8fd199b66dd9d3d5d7f7574a/server/adkrest/controllers/runtime.go#L366), and nothing says so. The deadline is absolute and armed before the client has even read the close frame, so the assertion is correct only while that deadline stays comfortably above 50 ms plus the client's round trip.

Shortening the production deadline to 40 ms makes this fatal 30/30 with the message "RunLiveHandler returned before receiving the close reply", which points at the drain rather than at the deadline that actually caused it. There is a 20x margin today and I could not make it fail otherwise — 200 iterations of `-race -cpu=1` with 96 busy spinners on 24 cores produced no failures. A comment naming `runtime.go:366` as the budget this is carved out of would keep the next person from tuning one without the other.

karolpiotrowicz (commented):
All four points from the last round are addressed, and the two I most wanted to see are addressed properly. The `errors.Is(err, websocket.ErrCloseSent)` guard is the version that actually works, and `TestRunLiveHandler_GracefulClientCloseDoesNotLogWriteError` is a real regression test: it fails 5/5 against the previous revision and passes on this one, so the fix is pinned in both directions. Overriding gorilla's default client close handler so the client does not auto-reply is what makes the drain assertion mean anything — deleting the drain loop now fails the suite, where before it did not. The realtime-input table test and the mid-stream iterator-error test both discriminate: swapping `activityStart` and `activityEnd`, dropping the blob branch, or removing the iterator-error close frame each break a test now.

Coverage of `RunLiveHandler` went 0% → 71% → 79% across the two rounds, and the full suite, `go vet`, `staticcheck` and `govulncheck` are unchanged against the merge base.

One thing worth fixing, and two smaller notes, all on `runtime_live_test.go` — I have put them inline.

The one worth fixing: extracting `readCloseError` kept the blocking read but dropped the one-second client read deadline that sat right in front of it before. That turns a class of failure from a one-second assertion into a `go test` timeout that takes the whole package's results down with it.

wmh123wmh1:
Thanks for the follow-up.

I restored the one-second client read deadlines before both `ReadMessage` and `ReadJSON`, and tightened the close-drain test timing assertion.

Validation completed:

- `go test -race -count=10 -run '^TestRunLiveHandler_' ./server/adkrest/controllers`
- `go test ./server/adkrest/controllers`
- `go vet ./server/adkrest/controllers`

The PR branch has been updated with commit `e3406f9`.

server/adkrest/controllers/runtime_live_test.go
```diff
@@ -0,0 +1,541 @@
...
+	liveSession := newRecordingLiveSession()
+	wantEvent := makeEvent("invocation-1", testLiveAppName, "Hello from live agent")
+	conn, _ := dialRunLiveHandler(t, func(agent.InvocationContext) (agent.LiveSession, iter.Seq2[*session.Event, error], error) {
+		return liveSession, func(yield func(*session.Event, error) bool) {
+			yield(wantEvent, nil)
+		}, nil
+	})
+
+	var got models.Event
+	if err := conn.ReadJSON(&got); err != nil {
```
karolpiotrowicz, line 217: This read is the third one, and it did not get the deadline the other two did. `readCloseError` arms one at [`:168`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime_live_test.go#L168) and `IteratorErrorSendsCloseFrame` at [`:474`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime_live_test.go#L474), so this is the one blocking socket read in the file that can still hang forever.

It matters when the handler leaves the upgraded socket open and sends nothing. Drop `defer ws.Close()` at [`runtime.go:360-362`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime.go#L360-L362) together with the event write at [`runtime.go:466`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime.go#L466) and the package dies on the test timeout, taking every other result in `server/adkrest/controllers` with it. Add a deadline in front of this read and nothing else, and the same breakage instead reports four ordinary assertion failures, the first here in 1.00 s.

```
go test -count=1 -timeout 60s -run '^TestRunLiveHandler_' ./server/adkrest/controllers
# with runtime.go's `defer ws.Close()` and the event write removed
#   as written:        panic: test timed out after 1m0s
#                        running tests: TestRunLiveHandler_StreamsEventsOverWebSocket (1m0s)
#   with the deadline: --- FAIL: TestRunLiveHandler_StreamsEventsOverWebSocket (1.00s)
#                        ReadJSON() failed: read tcp ...: i/o timeout
#                      (plus three more ordinary failures, package done in 2.09s)
```

The three lines from `:474` verbatim are all it needs.

server/adkrest/controllers/runtime_live_test.go
```diff
@@ -0,0 +1,541 @@
...
+	if closeErr.Code != websocket.CloseInternalServerErr {
+		t.Errorf("close code = %d, want %d", closeErr.Code, websocket.CloseInternalServerErr)
+	}
+	if closeErr.Text != wantErr.Error() {
+		t.Errorf("close reason = %q, want %q", closeErr.Text, wantErr.Error())
+	}
+	select {
+	case <-closeReceived:
+	case <-time.After(time.Second):
+		t.Fatal("client did not receive server close frame")
```
karolpiotrowicz: This branch cannot execute, so the assertion cannot fail. `readCloseError` at [`:429`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime_live_test.go#L429) only returns once `ReadMessage` yielded a `*websocket.CloseError`, and gorilla runs the registered close handler while decoding that frame, before the read returns. So `closeReceived` is already closed by the time this `select` is evaluated, on every path where `readCloseError` did not already fail.

I raised this last round and it is the one of the four that did not move, so I am assuming it was missed rather than declined. Either drop the `select` and keep a bare `SetCloseHandler` with a comment saying it suppresses gorilla's automatic close reply — which is the real reason it is installed, and what makes the drain assertion below mean anything — or leave it and say it is belt and braces. It costs a second in the worst case and reports nothing today.

server/adkrest/controllers/runtime_live_test.go
```diff
@@ -0,0 +1,541 @@
...
+	case <-time.After(testCloseDrainGracePeriod):
+	}
+	if err := conn.WriteControl(
+		websocket.CloseMessage,
+		websocket.FormatCloseMessage(closeErr.Code, "client acknowledged"),
+		time.Now().Add(time.Second),
+	); err != nil {
+		t.Fatalf("WriteControl(close reply) failed: %v", err)
+	}
+	select {
```
karolpiotrowicz, line 454: Worth a comment saying why this is spelled out rather than calling `waitForHandlerExit` like the four other waits in the file do.

The 100 ms budget is what makes this test discriminate. With the helper's one-second budget, a handler that ignores the close reply and just sleeps 200 ms still passes — that is exactly what happened against [`244ed368`](https://github.com/google/adk-go/commit/244ed368b378aa0a8fd199b66dd9d3d5d7f7574a). Without a note, the obvious consistency cleanup is to swap this back to the helper, which silently undoes the fix.

server/adkrest/controllers/runtime_live_test.go
```diff
@@ -0,0 +1,541 @@
...
+
+	liveSession := newRecordingLiveSession()
+	event := makeEvent("invocation-1", testLiveAppName, "tick")
+	conn, handlerDone := dialRunLiveHandler(t, func(agent.InvocationContext) (agent.LiveSession, iter.Seq2[*session.Event, error], error) {
+		return liveSession, func(yield func(*session.Event, error) bool) {
+			for {
+				if !yield(event, nil) {
+					return
+				}
+				time.Sleep(time.Millisecond)
```
karolpiotrowicz, line 532: This sleep is load-bearing and reads like it is not. It is what lets the handler's reader goroutine be scheduled and process the client's close frame before the write loop fills the socket buffer against a client that has stopped reading. Remove it as an unnecessary sleep in a test and the fixture becomes an unbounded write loop, the handler blocks in `WriteJSON`, and the failure surfaces at `waitForHandlerExit` a second later looking unrelated to the edit that caused it. One line of comment.

karolpiotrowicz (dismissed):
Both commits do what they say, and I checked each against the previous revision rather than taking the messages for it.

The read-deadline commit ([e3406f91](https://github.com/google/adk-go/commit/e3406f91698f09b8841bf09f6c81f1a42ade6673)) is the one I most wanted, and the half that landed is pinned in both directions. Deleting the close-frame write and the server-side deadline at [`runtime.go:365-366`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime.go#L365-L366) used to take the whole package down with `panic: test timed out after 1m30s`. With the restored deadline in front of the read, the same deletion now fails in 1.01 s with an ordinary assertion failure and every other test in the package still reports. That is exactly the property I asked for.

The close-drain commit ([10fae4a6](https://github.com/google/adk-go/commit/10fae4a66398f4df219aba74812818eba453648c)) closes the drain assertion. Replacing the drain loop at [`runtime.go:367-371`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime.go#L367-L371) with `time.Sleep(200 * time.Millisecond)` passed against the previous revision ([`244ed368`](https://github.com/google/adk-go/commit/244ed368b378aa0a8fd199b66dd9d3d5d7f7574a)) and fails against this one, so the 100 ms exit budget is doing real work. The comment naming the coupling is there too.

Coverage of `RunLiveHandler` is 0% on the merge base and 78.8% here. Build, vet, `golangci-lint` and `go mod tidy -diff` are clean on both modules, `apidiff` reports no API change, and the branch merges cleanly with current `main` with the merged tree building and the package green under `-race`.

## The one thing worth fixing

The read-deadline commit ([e3406f91](https://github.com/google/adk-go/commit/e3406f91698f09b8841bf09f6c81f1a42ade6673)) restored the deadline at two of the three blocking reads in the file. [`runtime_live_test.go:214`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime_live_test.go#L214) still has none, and it reproduces the same package-wide timeout the commit was written to remove. I have put the reproduction inline on that line. It is one line, copied from what you already wrote twice.

## Worth considering, not blocking

The `break` after a mid-stream iterator error is unobservable, because the fixture's iterator returns as soon as it yields the error. Dropping the `break` at [`runtime.go:463`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime.go#L463) leaves the suite green. This is not a theoretical shape: the real producer keeps going after a per-item error — [`runner.go:931-937`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/runner/runner.go#L931-L937) does `if !yield(nil, err) { return }` followed by `continue` — so the `break` is load-bearing against what actually feeds this handler. One extra `yield` after the error in the fixture at [`runtime_live_test.go:465-470`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime_live_test.go#L465-L470), plus an assertion that it never reaches the client, would close it. Best value per line in anything left here.

Three smaller notes are inline: an assertion that cannot fail, and two places where a load-bearing choice would be lost to the next person who tidies it.

## Not for this PR

Four things I found are worth a follow-up issue rather than more rounds here, and none of them is a defect in what you wrote:

- **The server's own drain deadline at [`runtime.go:366`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime.go#L366) is never exercised**, because every test either replies or drops the connection. Deleting that line leaves the suite green, and without it a client that reads the close frame and then goes silent pins the handler and its hijacked connection indefinitely. Covering it needs a new fixture, not an edit.
- **The snake_case query aliases at [`runtime.go:341`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime.go#L341), [`:345`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime.go#L345) and [`:349`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime.go#L349) have no coverage.** Deleting all three leaves the suite green, and they are a live client-compatibility surface, so nothing would stop them being removed as dead code.
- **`testWebSocketReadTimeout` is exactly equal to the server's drain deadline**, and the grace window is measured from when the client observes the close frame rather than from when the server armed its deadline. If the test goroutine ever stalls for around a second between those two points, the drain expires inside the grace window and the test fails claiming the handler returned early, which is the opposite of what happened. I could only produce it by injecting a 960 ms stall — 200 iterations at `GOMAXPROCS=1` on a deliberately oversubscribed host were clean — so this is about a misleading message on a rare path, not a flake you will see.
- **`RunLiveHandler` has no write deadline at [`runtime.go:466`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime.go#L466), and the missing-parameter path at [`runtime.go:352-353`](https://github.com/google/adk-go/blob/10fae4a66398f4df219aba74812818eba453648c/server/adkrest/controllers/runtime.go#L352-L353) returns 500 where this package's convention is an explicit 400.** Both predate this PR and belong on the handler, not on a test-only change.

One last thing, and it is not about the code: CI has not run build, test or lint on this head — only `check-changes` and `cla/google`. Everything above comes from my own runs. Worth getting a real CI run on it before anyone merges.

wmh123wmh1:
Thanks for the detailed follow-up review. I addressed the remaining
RunLiveHandler test findings in commit `03803df`:

- restored the read deadline before the remaining blocking `ReadJSON` call;
- removed the close-handler assertion that could not fail and documented why
  the custom handler suppresses gorilla's automatic close reply;
- documented why the close-reply exit timeout must remain shorter than the
  shared handler-exit helper;
- extended the iterator-error fixture to verify that the handler stops
  consuming events after a per-item error;
- documented why the short sleep in the graceful-close fixture is
  load-bearing.

I also verified the iterator-error test's discrimination directly: temporarily
removing the error-path `break` makes the focused test fail, while the current
implementation passes ten consecutive runs under the race detector.

Validation completed:

- `go test -race -count=10 -run '^TestRunLiveHandler_' ./server/adkrest/controllers`
- `go test -race -count=1 ./server/adkrest/...`
- `go vet ./server/adkrest/...`
- `golangci-lint v2.3.1 run` in both modules
- `git diff --check`

I left the server drain-deadline coverage, query alias coverage, write deadline,
and missing-parameter status behavior for separate follow-ups, as suggested.

Could you please take another look and update the review status when
convenient?

server/adkrest/controllers/runtime_live_test.go
```diff
@@ -0,0 +1,569 @@
...
+		"/run_live?appName=" + testLiveAppName + "&userId=" + testLiveUserID + "&sessionId=" + testLiveSessionID
+	conn, response, err := websocket.DefaultDialer.Dial(wsURL, nil)
+	if err != nil {
+		t.Fatalf("websocket.Dial() failed: %v", err)
+	}
+	t.Cleanup(func() { _ = conn.Close() })
+
+	if response.StatusCode != http.StatusSwitchingProtocols {
+		t.Fatalf("handshake status = %d, want %d", response.StatusCode, http.StatusSwitchingProtocols)
+	}
```
karolpiotrowicz, line 152: This check cannot fail, so it never reports anything. `websocket.Dialer.Dial` returns `ErrBadHandshake` for any response that is not a 101, and the `err != nil` fatal three lines up already catches that, so control only reaches here when the status was 101. There is no server behaviour that makes this branch fire.

It is the same shape as the `closeReceived` select you dropped this round, and the fix is the same: delete these three lines, or take the response out of the assignment entirely with `conn, _, err := websocket.DefaultDialer.Dial(wsURL, nil)`.

server/adkrest/controllers/runtime_live_test.go
```diff
@@ -0,0 +1,569 @@
...
+	return &recordingLiveSession{
+		requests: make(chan agent.LiveRequest, 1),
+		closed:   make(chan struct{}),
+	}
+}
+
+func (s *recordingLiveSession) Send(req agent.LiveRequest) error {
+	s.requests <- req
+	return nil
+}
```
karolpiotrowicz, line 69: This send has no escape, and the real implementation it stands in for does. `liveSessionImpl.Send` selects between the input channel and `<-s.done`, returning `io.EOF` once the session is closed, which is what lets the handler's reader goroutine terminate. The double here just blocks on a channel of capacity one.

Nothing reaches it today, because a client close breaks out before `Send` is called. It becomes a trap for the next test in this file: any regression that forwards two requests for one client frame wedges the reader goroutine here forever, while the handler's main goroutine is parked waiting for a close the wedged reader will never perform. The test still fails on its own timeout, but two goroutines and a hijacked connection leak for the rest of the binary.

`select { case s.requests <- req: return nil; case <-s.closed: return io.EOF }` mirrors production and costs nothing.

server/adkrest/controllers/runtime_live_test.go
```diff
@@ -0,0 +1,569 @@
...
+	case <-liveSession.closed:
+	case <-time.After(time.Second):
+		t.Fatal("live session was not closed after the client close request")
+	}
+}
+
+func TestRunLiveHandler_StreamsEventsOverWebSocket(t *testing.T) {
+	liveSession := newRecordingLiveSession()
+	wantEvent := makeEvent("invocation-1", testLiveAppName, "Hello from live agent")
+	conn, _ := dialRunLiveHandler(t, func(agent.InvocationContext) (agent.LiveSession, iter.Seq2[*session.Event, error], error) {
```
karolpiotrowicz, line 207: This test and three others (`:239`, `:266`, `:345`) discard the `handlerDone` channel that `dialRunLiveHandler` returns, so nothing in them asserts that `RunLiveHandler` ever comes back. Their only liveness check is `liveSession.closed` inside `closeLiveClient`, which the reader goroutine satisfies on its own.

A change that closes the session while leaving the main goroutine parked in the event loop keeps all four green. Keeping the channel and ending each with `waitForHandlerExit(t, handlerDone)` closes it, and matches what the other five tests already do.

karolpiotrowicz (commented):
The read deadline is in and it does exactly what it needed to do. I checked it the same way I checked the last one, by breaking the handler and watching how the suite reports it.

Removing `defer ws.Close()` and the event write, so the handler leaves the socket open and sends nothing, used to take the whole package down with `panic: test timed out after 1m0s` at sixty seconds. With the deadline you added at [`runtime_live_test.go:214-216`](https://github.com/google/adk-go/blob/03803df0f946e501b781c87590d354b49cea69fc/server/adkrest/controllers/runtime_live_test.go#L214-L216) the same breakage now fails in 1.00 s and the package finishes in 2.090 s with four ordinary assertion failures. Deleting just those three lines again and changing nothing else puts the sixty-second hang straight back, so the deadline is what does it.

<details><summary>Both runs</summary>

```
go test -count=1 -timeout 60s -run '^TestRunLiveHandler_' ./server/adkrest/controllers
# runtime.go's `defer ws.Close()` removed and the event write replaced with `_ = event`

# as written:
--- FAIL: TestRunLiveHandler_StreamsEventsOverWebSocket (1.00s)
    runtime_live_test.go:218: ReadJSON() failed: read tcp ...: i/o timeout
FAIL	google.golang.org/adk/v2/server/adkrest/controllers	2.090s

# with runtime_live_test.go:214-216 also deleted:
panic: test timed out after 1m0s
FAIL	google.golang.org/adk/v2/server/adkrest/controllers	60.035s
```

</details>

All three blocking reads in the file arm a deadline now, and there are no others.

The other four are done too, and two of them are better than what I asked for. Dropping the dead `select` and keeping a bare [`SetCloseHandler`](https://github.com/google/adk-go/blob/03803df0f946e501b781c87590d354b49cea69fc/server/adkrest/controllers/runtime_live_test.go#L427-L431) was the right call of the two options, and the handler is load-bearing rather than decorative: delete it and the drain assertion fires, because gorilla's automatic reply satisfies the server's drain before the test ever sends its own. The comment at [`:451-453`](https://github.com/google/adk-go/blob/03803df0f946e501b781c87590d354b49cea69fc/server/adkrest/controllers/runtime_live_test.go#L451-L453) is accurate on the point that matters — I raised the budget to a second and a handler that sleeps instead of draining does then pass, which is the thing the short budget prevents.

The iterator fixture is the one I'd single out. Yielding again after the error and asserting the handler never gets there closes the gap properly, and the [panic relay](https://github.com/google/adk-go/blob/03803df0f946e501b781c87590d354b49cea69fc/server/adkrest/controllers/runtime_live_test.go#L504-L512) is not decoration either — make the handler ignore the error and it dereferences a nil event, and that assertion is what reports it instead of the failure disappearing into net/http's log.

## What I could not find

I went looking for tests that cannot fail, since that is the whole value of a test-only change. There are none. Every one of the nine test functions fails under at least one deliberate break of the code it covers, and both production lines you added are pinned in both directions — remove the `ErrCloseSent` guard and the graceful-close test fails, remove the log line and the write-failure test fails.

The uncovered behavior that remains is all on lines that were already there before this PR and that none of these tests claims to cover. That is not something to fix here.

## Worth a follow-up, not a change to this PR

One of these is clearly the most valuable, so I'll put it first rather than bury it in a list.

**The server's own drain deadline at [`runtime.go:364-372`](https://github.com/google/adk-go/blob/03803df0f946e501b781c87590d354b49cea69fc/server/adkrest/controllers/runtime.go#L364-L372) is still never exercised.** Every test either replies to the close frame or drops the connection, so deleting `SetReadDeadline` from that block leaves the suite green — and without it a client that reads the close frame and then goes quiet pins the handler and its hijacked connection indefinitely. Your `RunLiveErrorSendsCloseFrameAndDrainsReply` fixture is already most of what a test for it needs: the same setup, with the client simply not replying, asserting the handler exits after about a second. Of everything left, that is the one I would spend effort on — as a follow-up issue, or here if you happen to want to. Nothing about it should hold this PR up.

The rest are genuinely separate:

- **The snake_case query aliases and the required-parameter guard** at [`runtime.go:340-354`](https://github.com/google/adk-go/blob/03803df0f946e501b781c87590d354b49cea69fc/server/adkrest/controllers/runtime.go#L340-L354) have no coverage. Deleting all three aliases leaves the suite green, and they are a live client-compatibility surface. The same path returns 500 for a missing parameter where the sibling handler at [`runtime.go:325`](https://github.com/google/adk-go/blob/03803df0f946e501b781c87590d354b49cea69fc/server/adkrest/controllers/runtime.go#L325) returns an explicit 400.
- **Nothing pins where the client-close check sits.** Moving [`if apiReq.Close { break }`](https://github.com/google/adk-go/blob/03803df0f946e501b781c87590d354b49cea69fc/server/adkrest/controllers/runtime.go#L432-L434) to after the `Send` leaves the suite green, and the handler would then forward an empty turn to the model on every client close.
- **Two smaller uncovered branches**: the `continue` after malformed client JSON at [`runtime.go:427-429`](https://github.com/google/adk-go/blob/03803df0f946e501b781c87590d354b49cea69fc/server/adkrest/controllers/runtime.go#L427-L429), and the `getRunner` failure path at [`runtime.go:377-379`](https://github.com/google/adk-go/blob/03803df0f946e501b781c87590d354b49cea69fc/server/adkrest/controllers/runtime.go#L377-L379) whose "agent not found" text goes to the client on the wire.

Three small things are inline, all in the test file and none of them blocking.

None of that needs to happen here. This PR set out to cover `RunLiveHandler` over a WebSocket and it does, the deadline is pinned in both directions, and the three lines it adds are the only ones it is responsible for — every gap above is on a line that is byte-identical at the merge base. A follow-up PR is the right home for them, and the list above is close to a ready-made scope for one: the drain deadline, the query aliases and the 500-versus-400 mismatch, the client-close ordering, and the two uncovered branches. Happy to review that as its own change.

One administrative note: my earlier review is still sitting on this PR as "changes requested", and its one blocking item is the read deadline you just fixed. That review is stale and I will clear it — the PR is not waiting on anything from you on that front.

karolpiotrowicz (approved):
Approving. The read deadline was the one thing holding this up and it is in, pinned in both directions — with it the broken handler fails in about a second, and deleting the three lines again puts the sixty-second package hang straight back.

I have also cleared my earlier "changes requested", which was still sitting on this PR for that same item and no longer described the code.

The follow-up scope from my last comment stands as follow-up and not as a condition of this change: the server drain deadline, the snake_case query aliases together with the 500-versus-400 mismatch on a missing parameter, where the client-close check sits, and the two uncovered branches. Every one of those is on a line that is byte-identical at the merge base, so none of it is yours to carry here. A separate PR is the right home if you or anyone else picks it up, and I am happy to review that.

Worth saying plainly what this change did: integration coverage for `RunLiveHandler` over a real WebSocket, where there was none, and the three production lines it adds are pinned both ways.

wolo-lab:
Resolving non-blocking comments. To be fixed in the follow-up PR.

PR #1454, opened 2026-09-01
feat(telemetry): record tool definitions on model requests

### LINKED ISSUE #789: Tool definitions are not recorded on the request path (gen_ai.tool.definitions)

## 🔴 Required Information

We're using OpenTelemetry logging in our product. Our product requires logging of all information that went into the prompt, so when debugging we can re-create the exact environment that produce an LLM output. One item that doesn't seem to be logged is the tools provided to the agent and their prompts. The only time we get the tool description (but not the input schema that may contain text description of each input field) is when a tool is called by the agent.

### Proposed Solution
Log tools and their information just like how the system prompt currently is logged, as these make up part of the text sent to the LLM.

### Impact on your work
We can implement workarounds for now, but would be a good feature for you to have eventually.

---

### Alternatives Considered
We can just log the information ourselves for now.

### Willingness to contribute
Happy to give it a crack.

N/A

### Additional Context
N/A

wmh123wmh1:
## Summary

- record function tool definitions from the actual model request
- emit them through the `gen_ai.tool.definitions` attribute
- preserve both `Parameters` and `ParametersJsonSchema`
- keep tool definition capture behind `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`
- add unit and functional telemetry coverage

## Testing

- `go test ./internal/telemetry/... -count=1`
- `go vet ./internal/telemetry`
- `git diff --check`

Fixes #789

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,16 +50,20 @@ go get google.golang.org/adk/v2
 ## 🔭 Telemetry
 
 ADK records tool names, descriptions, and types in `gen_ai.tool.definitions`
-when GenAI content capture is enabled. Tool parameter schemas are excluded by
+when span content capture is enabled. Tool parameter schemas are excluded by
 default because they may contain sensitive application details. To opt in to
-parameter-schema capture, set:
+parameter-schema capture, enable span content capture and set the ADK-specific
+flag:
 
 ```bash
+export OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=SPAN_ONLY
 export ADK_INSTRUMENTATION_GENAI_CAPTURE_TOOL_DEFINITION_PARAMETERS=true
 ```
 
-Use this setting only when the captured tool schemas are appropriate for your
-telemetry destination.
+Use `SPAN_AND_EVENT` instead of `SPAN_ONLY` if message content should also be
+recorded in log events. The parameter-capture flag has no effect unless span
+content capture is enabled. Use this setting only when the captured tool
+schemas are appropriate for your telemetry destination.
 
 ## 📖 Docs for AI coding agents
 
--- a/README.md
+++ b/README.md
@@ -47,6 +47,20 @@ To add ADK Go to your project, run:
 go get google.golang.org/adk/v2
 ```
 
+## 🔭 Telemetry
+
+ADK records tool names, descriptions, and types in `gen_ai.tool.definitions`
+when GenAI content capture is enabled. Tool parameter schemas are excluded by
+default because they may contain sensitive application details. To opt in to
+parameter-schema capture, set:
+
+```bash
+export ADK_INSTRUMENTATION_GENAI_CAPTURE_TOOL_DEFINITION_PARAMETERS=true
+```
+
+Use this setting only when the captured tool schemas are appropriate for your
+telemetry destination.
+
 ## 📖 Docs for AI coding agents
 
 [adk.dev/llms.txt](https://adk.dev/llms.txt) is a machine-readable index of the ADK documentation, and [adk.dev/llms-full.txt](https://adk.dev/llms-full.txt) is the same documentation as a single file. Both are generated from [adk-docs](https://github.com/google/adk-docs/) and include the Go API reference and samples, so you can point a coding agent at either one for context:
--- a/README.md
+++ b/README.md
@@ -49,11 +49,12 @@ go get google.golang.org/adk/v2
 
 ## 🔭 Telemetry
 
-ADK records tool names, descriptions, and types in `gen_ai.tool.definitions`
-when span content capture is enabled. Tool parameter schemas are excluded by
-default because they may contain sensitive application details. To opt in to
-parameter-schema capture, enable span content capture and set the ADK-specific
-flag:
+ADK records function tool names, descriptions, and types in
+`gen_ai.tool.definitions` when span content capture is enabled. Provider-native
+tools such as `google_search`, `code_execution`, and `url_context` are not
+currently included. Tool parameter schemas are excluded by default because
+they may contain sensitive application details. To opt in to parameter-schema
+capture, enable span content capture and set the ADK-specific flag:
 
 ```bash
 export OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=SPAN_ONLY
@@ -61,8 +62,11 @@ export ADK_INSTRUMENTATION_GENAI_CAPTURE_TOOL_DEFINITION_PARAMETERS=true
 ```
 
 Use `SPAN_AND_EVENT` instead of `SPAN_ONLY` if message content should also be
-recorded in log events. The parameter-capture flag has no effect unless span
-content capture is enabled. Use this setting only when the captured tool
+recorded in log events. A value of `true` is retained for backward compatibility
+and is equivalent to `EVENT_ONLY`; it does not enable content attributes on
+spans. Use `SPAN_ONLY` or `SPAN_AND_EVENT` when recording tool definitions or
+parameter schemas on spans. The parameter-capture flag has no effect unless
+span content capture is enabled. Use this setting only when the captured tool
 schemas are appropriate for your telemetry destination.
 
 ## 📖 Docs for AI coding agents
--- a/README.md
+++ b/README.md
@@ -69,6 +69,12 @@ parameter schemas on spans. The parameter-capture flag has no effect unless
 span content capture is enabled. Use this setting only when the captured tool
 schemas are appropriate for your telemetry destination.
 
+Each content attribute is limited to 60 KiB (61,440 bytes). When parameter
+capture is enabled and the complete tool definition does not fit, ADK retries
+with the tool name, description, and type while omitting the parameter schemas.
+If that metadata-only representation is also too large, the
+`gen_ai.tool.definitions` attribute is omitted.
+
 ## 📖 Docs for AI coding agents
 
 [adk.dev/llms.txt](https://adk.dev/llms.txt) is a machine-readable index of the ADK documentation, and [adk.dev/llms-full.txt](https://adk.dev/llms-full.txt) is the same documentation as a single file. Both are generated from [adk-docs](https://github.com/google/adk-docs/) and include the Go API reference and samples, so you can point a coding agent at either one for context:
--- a/internal/telemetry/functionaltest/agent_telemetry_schema_test.go
+++ b/internal/telemetry/functionaltest/agent_telemetry_schema_test.go
@@ -38,6 +38,10 @@ import (
 // records or elides it for privacy.
 const captureMessageContentEnvVar = "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT"
 
+// captureToolDefinitionParametersEnvVar controls the explicit input-schema
+// capture opt-in used by the telemetry package.
+const captureToolDefinitionParametersEnvVar = "ADK_INSTRUMENTATION_GENAI_CAPTURE_TOOL_DEFINITION_PARAMETERS"
+
 // TestTelemetrySchema_AgentWithTool runs the canonical
 // "llmagent with one FunctionTool" scenario end-to-end and asserts
 // the emitted span+log tree matches the expected shape exactly.
@@ -76,6 +80,9 @@ func TestTelemetrySchema_AgentWithTool(t *testing.T) {
 			} else {
 				t.Setenv(captureMessageContentEnvVar, "")
 			}
+			// Keep this end-to-end test independent of a developer's local
+			// tool-schema capture setting.
+			t.Setenv(captureToolDefinitionParametersEnvVar, "")
 			telemetry.ApplyEnv()
 
 			// Install in-memory tracer + logger so the test sees
--- a/internal/telemetry/functionaltest/agent_telemetry_schema_test.go
+++ b/internal/telemetry/functionaltest/agent_telemetry_schema_test.go
@@ -83,6 +83,9 @@ func TestTelemetrySchema_AgentWithTool(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
+			// Register before t.Setenv so environment restoration runs before
+			// ApplyEnv refreshes the package-level telemetry settings.
+			t.Cleanup(telemetry.ApplyEnv)
 			if tc.captureContent {
 				// Spans have to be named: a truthy value means log records only.
 				t.Setenv(captureMessageContentEnvVar, "SPAN_AND_EVENT")
--- a/internal/telemetry/functionaltest/agent_telemetry_schema_test.go
+++ b/internal/telemetry/functionaltest/agent_telemetry_schema_test.go
@@ -56,19 +56,28 @@ const captureToolDefinitionParametersEnvVar = "ADK_INSTRUMENTATION_GENAI_CAPTURE
 // adk-python/tests/unittests/telemetry/test_functional.py.
 func TestTelemetrySchema_AgentWithTool(t *testing.T) {
 	tests := []struct {
-		name           string
-		captureContent bool
-		want           *telemetrytest.SpanDigest
+		name              string
+		captureContent    bool
+		captureParameters bool
+		want              *telemetrytest.SpanDigest
 	}{
 		{
-			name:           "elided",
-			captureContent: false,
-			want:           telemetrytestcase.AgentWithToolCase,
+			name:              "elided",
+			captureContent:    false,
+			captureParameters: false,
+			want:              telemetrytestcase.AgentWithToolCase,
 		},
 		{
-			name:           "capture_content",
-			captureContent: true,
-			want:           telemetrytestcase.AgentWithToolCaptureContentCase,
+			name:              "capture_content",
+			captureContent:    true,
+			captureParameters: false,
+			want:              telemetrytestcase.AgentWithToolCaptureContentCase,
+		},
+		{
+			name:              "capture_content_and_tool_parameters",
+			captureContent:    true,
+			captureParameters: true,
+			want:              telemetrytestcase.AgentWithToolCaptureContentAndParametersCase,
 		},
 	}
 
@@ -82,7 +91,11 @@ func TestTelemetrySchema_AgentWithTool(t *testing.T) {
 			}
 			// Keep this end-to-end test independent of a developer's local
 			// tool-schema capture setting.
-			t.Setenv(captureToolDefinitionParametersEnvVar, "")
+			if tc.captureParameters {
+				t.Setenv(captureToolDefinitionParametersEnvVar, "true")
+			} else {
+				t.Setenv(captureToolDefinitionParametersEnvVar, "")
+			}
 			telemetry.ApplyEnv()
 
 			// Install in-memory tracer + logger so the test sees
--- a/internal/telemetry/genaimessages.go
+++ b/internal/telemetry/genaimessages.go
@@ -119,14 +119,13 @@ type toolResponsePart struct {
 }
 
 // functionToolDefinition is the representation used by the GenAI semantic
-// conventions for a function declaration. Description and parameters are kept
-// as explicit fields so their absence is distinguishable from an empty value
-// in the recorded JSON.
+// conventions for a function declaration. Description is always emitted, while
+// parameters are included only when explicitly enabled.
 type functionToolDefinition struct {
-	Name        string `json:"name"`
-	Description string `json:"description"`
-	Parameters  any    `json:"parameters"`
-	Type        string `json:"type"`
+	Name        string          `json:"name"`
+	Description string          `json:"description"`
+	Parameters  json.RawMessage `json:"parameters,omitempty"`
+	Type        string          `json:"type"`
 }
 
 // toolDefinitions converts the function declarations present in the actual
@@ -137,6 +136,7 @@ func toolDefinitions(config *genai.GenerateContentConfig) []functionToolDefiniti
 	if config == nil {
 		return nil
 	}
+	includeParameters := captureToolDefinitionParametersOnSpans()
 	var definitions []functionToolDefinition
 	for _, tool := range config.Tools {
 		if tool == nil {
@@ -146,23 +146,79 @@ func toolDefinitions(config *genai.GenerateContentConfig) []functionToolDefiniti
 			if declaration == nil {
 				continue
 			}
-			parameters := any(declaration.Parameters)
-			if declaration.ParametersJsonSchema != nil {
-				parameters = declaration.ParametersJsonSchema
-			}
-			definitions = append(definitions, functionToolDefinition{
+			definition := functionToolDefinition{
 				Name:        declaration.Name,
 				Description: declaration.Description,
-				Parameters:  parameters,
 				Type:        "function",
-			})
+			}
+			if includeParameters {
+				definition.Parameters = toolDefinitionParameters(declaration)
+			}
+			definitions = append(definitions, definition)
 		}
 	}
 	return definitions
 }
 
-// requestContentAttributes returns the gen_ai.system_instructions and
-// gen_ai.input.messages attributes for req, or nil when content capture is off
+// toolDefinitionParameters returns a valid JSON value for a declaration's
+// parameters. Each declaration is encoded independently so one malformed
+// application-provided schema cannot erase every tool definition on the span.
+func toolDefinitionParameters(declaration *genai.FunctionDeclaration) json.RawMessage {
+	if declaration.Parameters == nil && declaration.ParametersJsonSchema == nil {
+		return json.RawMessage("null")
+	}
+
+	value := any(declaration.Parameters)
+	if declaration.ParametersJsonSchema != nil {
+		value = declaration.ParametersJsonSchema
+	}
+	encoded, err := json.Marshal(value)
+	if err != nil {
+		return json.RawMessage(unserializablePlaceholder)
+	}
+
+	var normalized any
+	if err := json.Unmarshal(encoded, &normalized); err != nil {
+		return json.RawMessage(unserializablePlaceholder)
+	}
+	lowercaseSchemaTypes(normalized)
+	encoded, err = json.Marshal(normalized)
+	if err != nil {
+		return json.RawMessage(unserializablePlaceholder)
+	}
+	return json.RawMessage(encoded)
+}
+
+// lowercaseSchemaTypes converts genai.Schema's protobuf enum spellings to the
+// lowercase type values required by JSON Schema. It also normalizes literal
+// schema maps supplied by callers, including nested schemas.
+func lowercaseSchemaTypes(value any) {
+	switch value := value.(type) {
+	case map[string]any:
+		if typeValue, ok := value["type"]; ok {
+			switch typeValue := typeValue.(type) {
+			case string:
+				value["type"] = strings.ToLower(typeValue)
+			case []any:
+				for i, item := range typeValue {
+					if item, ok := item.(string); ok {
+						typeValue[i] = strings.ToLower(item)
+					}
+				}
+			}
+		}
+		for _, child := range value {
+			lowercaseSchemaTypes(child)
+		}
+	case []any:
+		for _, child := range value {
+			lowercaseSchemaTypes(child)
+		}
+	}
+}
+
+// requestContentAttributes returns the gen_ai.system_instructions,
+// gen_ai.tool.definitions, and gen_ai.input.messages attributes for req, or nil when content capture is off
 // or req carries nothing to record.
 //
 // Content is sensitive and often large, so the semantic conventions require
@@ -426,12 +482,14 @@ func schemaFinishReason(resp *model.LLMResponse, toolCall bool) string {
 
 // appendJSON encodes value and appends it to attrs under key, unless it does
 // not fit. See [maxContentAttributeBytes] for why an oversized attribute is
-// dropped rather than trimmed.
+// dropped rather than trimmed. Values used for tool payloads are encoded
+// independently before they reach this function so one bad payload cannot
+// discard its siblings.
 func appendJSON(attrs []attribute.KeyValue, key attribute.Key, value any) []attribute.KeyValue {
 	encoded, err := json.Marshal(value)
 	if err != nil {
-		// Unreachable: tool payloads are pre-encoded by toolPayload and
-		// everything else is a string or a slice of them.
+		// Callers should pre-encode values whose individual failures must be
+		// isolated, such as tool definition parameters.
 		return attrs
 	}
 	if len(encoded) > maxContentAttributeBytes {
--- a/internal/telemetry/genaimessages.go
+++ b/internal/telemetry/genaimessages.go
@@ -175,7 +175,7 @@ func toolDefinitionParameters(declaration *genai.FunctionDeclaration) json.RawMe
 	}
 
 	value := any(declaration.Parameters)
-	if declaration.ParametersJsonSchema != nil {
+	if declaration.Parameters == nil {
 		value = declaration.ParametersJsonSchema
 	}
 	encoded, err := json.Marshal(value)
--- a/internal/telemetry/genaimessages.go
+++ b/internal/telemetry/genaimessages.go
@@ -31,6 +31,7 @@ var (
 	genAIInputMessages      = attribute.Key("gen_ai.input.messages")
 	genAIOutputMessages     = attribute.Key("gen_ai.output.messages")
 	genAISystemInstructions = attribute.Key("gen_ai.system_instructions")
+	genAIToolDefinitions    = attribute.Key("gen_ai.tool.definitions")
 )
 
 // Part type discriminators and roles defined by the message JSON schemas.
@@ -117,6 +118,49 @@ type toolResponsePart struct {
 	Response json.RawMessage `json:"response"`
 }
 
+// functionToolDefinition is the representation used by the GenAI semantic
+// conventions for a function declaration. Description and parameters are kept
+// as explicit fields so their absence is distinguishable from an empty value
+// in the recorded JSON.
+type functionToolDefinition struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	Parameters  any    `json:"parameters"`
+	Type        string `json:"type"`
+}
+
+// toolDefinitions converts the function declarations present in the actual
+// model request into the gen_ai.tool.definitions representation. Reading the
+// request is important because a Toolset may produce a different set of tools
+// for each model call.
+func toolDefinitions(config *genai.GenerateContentConfig) []functionToolDefinition {
+	if config == nil {
+		return nil
+	}
+	var definitions []functionToolDefinition
+	for _, tool := range config.Tools {
+		if tool == nil {
+			continue
+		}
+		for _, declaration := range tool.FunctionDeclarations {
+			if declaration == nil {
+				continue
+			}
+			parameters := any(declaration.Parameters)
+			if declaration.ParametersJsonSchema != nil {
+				parameters = declaration.ParametersJsonSchema
+			}
+			definitions = append(definitions, functionToolDefinition{
+				Name:        declaration.Name,
+				Description: declaration.Description,
+				Parameters:  parameters,
+				Type:        "function",
+			})
+		}
+	}
+	return definitions
+}
+
 // requestContentAttributes returns the gen_ai.system_instructions and
 // gen_ai.input.messages attributes for req, or nil when content capture is off
 // or req carries nothing to record.
@@ -134,6 +178,9 @@ func requestContentAttributes(req *model.LLMRequest) []attribute.KeyValue {
 		if parts := semconvParts(req.Config.SystemInstruction.Parts); len(parts) > 0 {
 			attrs = appendJSON(attrs, genAISystemInstructions, parts)
 		}
+	}
+	if definitions := toolDefinitions(req.Config); len(definitions) > 0 {
+		attrs = appendJSON(attrs, genAIToolDefinitions, definitions)
 	}
 	if len(req.Contents) > 0 {
 		// Messages MUST be recorded in the order they were sent. A turn whose
--- a/internal/telemetry/genaimessages.go
+++ b/internal/telemetry/genaimessages.go
@@ -35,6 +35,16 @@ var (
 	genAIToolDefinitions    = attribute.Key("gen_ai.tool.definitions")
 )
 
+var genaiTypeNames = map[string]string{
+	string(genai.TypeString):  "string",
+	string(genai.TypeNumber):  "number",
+	string(genai.TypeInteger): "integer",
+	string(genai.TypeBoolean): "boolean",
+	string(genai.TypeArray):   "array",
+	string(genai.TypeObject):  "object",
+	string(genai.TypeNULL):    "null",
+}
+
 // Part type discriminators and roles defined by the message JSON schemas.
 const (
 	partTypeText             = "text"
@@ -189,6 +199,12 @@ func toolDefinitionParameters(declaration *genai.FunctionDeclaration) json.RawMe
 	if err := decoder.Decode(&normalized); err != nil {
 		return json.RawMessage(unserializableSchemaPlaceholder)
 	}
+	switch normalized.(type) {
+	case map[string]any, bool:
+		// JSON Schema draft-07 permits schemas to be objects or booleans.
+	default:
+		return json.RawMessage(unserializableSchemaPlaceholder)
+	}
 	normalizeSchemaTypes(normalized)
 	encoded, err = json.Marshal(normalized)
 	if err != nil {
@@ -227,6 +243,16 @@ func normalizeSchemaNode(value any) {
 					normalizeSchemaNode(schema)
 				}
 			}
+		case "dependencies":
+			// In draft-07 a dependency is either a property-name array or a
+			// schema. Normalize only schema-valued dependencies.
+			if dependencies, ok := child.(map[string]any); ok {
+				for _, dependency := range dependencies {
+					if schema, ok := dependency.(map[string]any); ok {
+						normalizeSchemaNode(schema)
+					}
+				}
+			}
 		case "additionalProperties", "additionalItems", "items", "contains", "not", "if", "then", "else", "propertyNames", "unevaluatedProperties", "unevaluatedItems":
 			normalizeSchemaValue(child)
 		case "allOf", "anyOf", "oneOf", "prefixItems":
@@ -251,15 +277,6 @@ func normalizeSchemaValue(value any) {
 }
 
 func normalizeSchemaTypeValue(value any) (any, bool) {
-	genaiTypeNames := map[string]string{
-		string(genai.TypeString):  "string",
-		string(genai.TypeNumber):  "number",
-		string(genai.TypeInteger): "integer",
-		string(genai.TypeBoolean): "boolean",
-		string(genai.TypeArray):   "array",
-		string(genai.TypeObject):  "object",
-		string(genai.TypeNULL):    "null",
-	}
 	switch value := value.(type) {
 	case string:
 		if normalized, ok := genaiTypeNames[value]; ok {
--- a/internal/telemetry/genaimessages.go
+++ b/internal/telemetry/genaimessages.go
@@ -15,6 +15,7 @@
 package telemetry
 
 import (
+	"bytes"
 	"encoding/base64"
 	"encoding/json"
 	"strings"
@@ -69,6 +70,11 @@ const maxContentAttributeBytes = 60 << 10
 // Tool arguments and responses are filled in by application code and can hold a
 // NaN, an Inf, a func, a chan or a reference cycle.
 const unserializablePlaceholder = `"<unserializable>"`
+
+// unserializableSchemaPlaceholder is a valid JSON Schema used when a tool
+// parameter schema cannot be encoded. Unlike unserializablePlaceholder, this
+// value is used where the semantic convention requires a schema object.
+const unserializableSchemaPlaceholder = `{"type":"object","properties":{"serialization_error":{"type":"string"}}}`
 
 // maxInlineDataBytes bounds one inline payload before base64, which inflates it
 // by a third. A single image would otherwise consume the whole attribute and
@@ -174,46 +180,109 @@ func toolDefinitionParameters(declaration *genai.FunctionDeclaration) json.RawMe
 	}
 	encoded, err := json.Marshal(value)
 	if err != nil {
-		return json.RawMessage(unserializablePlaceholder)
+		return json.RawMessage(unserializableSchemaPlaceholder)
 	}
 
 	var normalized any
-	if err := json.Unmarshal(encoded, &normalized); err != nil {
-		return json.RawMessage(unserializablePlaceholder)
+	decoder := json.NewDecoder(bytes.NewReader(encoded))
+	decoder.UseNumber()
+	if err := decoder.Decode(&normalized); err != nil {
+		return json.RawMessage(unserializableSchemaPlaceholder)
 	}
-	lowercaseSchemaTypes(normalized)
+	normalizeSchemaTypes(normalized)
 	encoded, err = json.Marshal(normalized)
 	if err != nil {
-		return json.RawMessage(unserializablePlaceholder)
+		return json.RawMessage(unserializableSchemaPlaceholder)
 	}
 	return json.RawMessage(encoded)
 }
 
-// lowercaseSchemaTypes converts genai.Schema's protobuf enum spellings to the
-// lowercase type values required by JSON Schema. It also normalizes literal
-// schema maps supplied by callers, including nested schemas.
-func lowercaseSchemaTypes(value any) {
-	switch value := value.(type) {
-	case map[string]any:
-		if typeValue, ok := value["type"]; ok {
-			switch typeValue := typeValue.(type) {
-			case string:
-				value["type"] = strings.ToLower(typeValue)
-			case []any:
-				for i, item := range typeValue {
-					if item, ok := item.(string); ok {
-						typeValue[i] = strings.ToLower(item)
-					}
+// normalizeSchemaTypes converts only genai.Schema's protobuf enum spellings
+// to the lowercase type values required by JSON Schema. It deliberately leaves
+// arbitrary strings in data-bearing fields such as default and examples alone.
+func normalizeSchemaTypes(value any) {
+	normalizeSchemaNode(value)
+}
+
+// normalizeSchemaNode walks only the JSON Schema locations that contain nested
+// schemas. Walking every nested map would also rewrite ordinary instance data
+// under keywords such as default, examples, and dependencies.
+func normalizeSchemaNode(value any) {
+	obj, ok := value.(map[string]any)
+	if !ok {
+		return
+	}
+	if typeValue, ok := obj["type"]; ok {
+		if normalized, keep := normalizeSchemaTypeValue(typeValue); keep {
+			obj["type"] = normalized
+		} else {
+			delete(obj, "type")
+		}
+	}
+	for key, child := range obj {
+		switch key {
+		case "properties", "patternProperties", "dependentSchemas", "definitions", "$defs":
+			if schemas, ok := child.(map[string]any); ok {
+				for _, schema := range schemas {
+					normalizeSchemaNode(schema)
+				}
+			}
+		case "additionalProperties", "additionalItems", "items", "contains", "not", "if", "then", "else", "propertyNames", "unevaluatedProperties", "unevaluatedItems":
+			normalizeSchemaValue(child)
+		case "allOf", "anyOf", "oneOf", "prefixItems":
+			if schemas, ok := child.([]any); ok {
+				for _, schema := range schemas {
+					normalizeSchemaNode(schema)
 				}
 			}
 		}
-		for _, child := range value {
-			lowercaseSchemaTypes(child)
-		}
+	}
+}
+
+func normalizeSchemaValue(value any) {
+	switch value := value.(type) {
+	case map[string]any:
+		normalizeSchemaNode(value)
 	case []any:
-		for _, child := range value {
-			lowercaseSchemaTypes(child)
+		for _, schema := range value {
+			normalizeSchemaNode(schema)
 		}
+	}
+}
+
+func normalizeSchemaTypeValue(value any) (any, bool) {
+	genaiTypeNames := map[string]string{
+		string(genai.TypeString):  "string",
+		string(genai.TypeNumber):  "number",
+		string(genai.TypeInteger): "integer",
+		string(genai.TypeBoolean): "boolean",
+		string(genai.TypeArray):   "array",
+		string(genai.TypeObject):  "object",
+		string(genai.TypeNULL):    "null",
+	}
+	switch value := value.(type) {
+	case string:
+		if normalized, ok := genaiTypeNames[value]; ok {
+			return normalized, true
+		}
+		if value == string(genai.TypeUnspecified) {
+			return nil, false
+		}
+		return value, true
+	case []any:
+		normalized := make([]any, 0, len(value))
+		for _, item := range value {
+			item, keep := normalizeSchemaTypeValue(item)
+			if keep {
+				normalized = append(normalized, item)
+			}
+		}
+		if len(normalized) == 0 {
+			return nil, false
+		}
+		return normalized, true
+	default:
+		return value, true
 	}
 }
 
--- a/internal/telemetry/genaimessages.go
+++ b/internal/telemetry/genaimessages.go
@@ -45,6 +45,16 @@ var genaiTypeNames = map[string]string{
 	string(genai.TypeNULL):    "null",
 }
 
+var jsonSchemaTypeNames = map[string]struct{}{
+	"string":  {},
+	"number":  {},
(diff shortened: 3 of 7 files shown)


karolpiotrowicz (dismissed):
This is a genuinely useful addition and the implementation is careful in the places that usually go wrong: the capture gate is applied in the right place, the nil handling is complete, and the attribute goes through the same size-bounded encoder as its siblings. Three things need changing before it can merge, and the first one is the important one.

## The `Parameters` branch emits a schema dialect that isn't JSON Schema

`genai.FunctionDeclaration` carries two mutually exclusive parameter fields, and ADK populates both depending on the tool. [`ParametersJsonSchema`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/tool/functiontool/function.go#L166) holds a plain JSON Schema document. [`Parameters`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/internal/llminternal/agent_transfer.go#L136-L141) holds a `*genai.Schema`, whose `Type` marshals as the genai protobuf enum name — `"OBJECT"`, `"STRING"`. At [`genaimessages.go:149-152`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/internal/telemetry/genaimessages.go#L149-L152) the second one is passed through untouched, so the recorded document is not valid JSON Schema.

The semantic convention is a MUST here: the Tool Definitions schema types `parameters` as `anyOf[{"$ref": "http://json-schema.org/draft-07/schema#"}, null]` and states the value must conform to draft-07, whose `type` enum is lowercase.

This is not an edge case. `transfer_to_agent` is injected automatically for any agent that has sub-agents, and it takes the `Parameters` path, as do `loadmemorytool`, `loadartifactstool` and the workflow task tools. A multi-agent app hits it on the first model call. Running an agent with one sub-agent and one function tool through `runner.Run` produces both dialects side by side in a single array:

<details><summary>emitted <code>gen_ai.tool.definitions</code> — two dialects in one array</summary>

```json
[
  {
    "name": "transfer_to_agent",
    "description": "Transfer the question to another agent....",
    "parameters": {
      "properties": {
        "agent_name": {
          "description": "the agent name to transfer to",
          "enum": ["some_sub_agent"],
          "type": "STRING"
        }
      },
      "required": ["agent_name"],
      "type": "OBJECT"
    },
    "type": "function"
  },
  {
    "name": "some_tool",
    "description": "A sample tool.",
    "parameters": {
      "type": "object",
      "properties": {"arg1": {"type": "string"}},
      "required": ["arg1"],
      "additionalProperties": false
    },
    "type": "function"
  }
]
```

</details>

Worth knowing before you pick a fix: this package already has a rule against exactly this, and the new attribute is the one content attribute that bypasses it. [`enumFields`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/internal/telemetry/genaimessages_test.go#L87-L92) lists `"type"` with the comment "must never carry a protobuf wire name", and [`checkEnumFields`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/internal/telemetry/genaimessages_test.go#L192-L206) fails with `field %q leaks the protobuf wire name %q into telemetry`. Calling that helper, unmodified, on the value above fails on both `"OBJECT"` and `"STRING"`. `gen_ai.input.messages` and `gen_ai.output.messages` are both routed through it. `gen_ai.tool.definitions` is not.

Two related notes rather than a prescription. [`model/openaimodel/tools.go:186-208`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/model/openaimodel/tools.go#L186-L208) already solves this problem for the OpenAI path by marshalling the schema and lowercasing every `type`, so there is a precedent to follow if you want one. And [`plugin/functioncallmodifier/plugin.go:74`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/plugin/functioncallmodifier/plugin.go#L73-L75) writes `Type: "OBJECT"` as a literal string rather than the enum, so anything scoped to converting `genai.Type` values would still miss that one.

Whatever you choose, [`genaimessages_test.go:404-409`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/internal/telemetry/genaimessages_test.go#L404-L409) needs to move with it. As written it asserts `"type": "OBJECT"` and `"type": "STRING"` as the expected output, which locks the current shape in as intended behavior.

## Built-in tools are recorded as no tools at all

The loop at [`genaimessages.go:141-147`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/internal/telemetry/genaimessages.go#L141-L147) only descends into `FunctionDeclarations`. Provider-side tools reach `Config.Tools` as a bare `*genai.Tool` with none, via [`geminitool.setTool`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/tool/geminitool/tool.go#L78-L88), so `google_search`, `code_execution` and `url_context` contribute nothing. When one of them is the agent's only tool the definitions slice is empty, the guard at [`genaimessages.go:182`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/internal/telemetry/genaimessages.go#L182-L184) skips the attribute, and the span becomes indistinguishable from a request that offered the model no tools. An operator asking "did this request grant the model web search?" gets the wrong answer, which is the question issue #789 is about.

The convention anticipates this with `GenericToolDefinition`, requiring only `type` and `name`. Fair warning that it is not a two-line change: `geminitool` keeps the tool's name on its own struct and discards it at `setTool`, so the name is not available where you would need it. Documenting the omission on `toolDefinitions` would be a reasonable smaller step.

## One unmarshalable schema erases every tool on the span

`ParametersJsonSchema` is an `any` supplied by application code and, through [`mcptoolset`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/tool/mcptoolset/tool.go#L51), by a remote MCP server. It now flows straight into the single `json.Marshal` in [`appendJSON`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/internal/telemetry/genaimessages.go#L430-L439). A request with one good tool and one whose schema contains a `NaN` records nothing at all — the good tool's definition is lost with the bad one. `toolPayload` exists a few lines up to keep exactly this contained, substituting a placeholder per payload rather than dropping everything.

The comment on that error branch — "Unreachable: tool payloads are pre-encoded by `toolPayload` and everything else is a string or a slice of them" — is no longer true, and it tells the next maintainer not to look here.

## Smaller things

- The doc comment at [`genaimessages.go:121-124`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/internal/telemetry/genaimessages.go#L121-L124) says description and parameters are explicit fields "so their absence is distinguishable from an empty value". That holds for `Parameters`, where nil becomes `null`, but not for `Description string` — unset and explicitly empty both render `"description": ""`.
- [`genaimessages.go:164-166`](https://github.com/google/adk-go/blob/cc5696231ce206f6616f92ec1d7e56998512dfe7/internal/telemetry/genaimessages.go#L164-L166) still describes the function as returning `gen_ai.system_instructions` and `gen_ai.input.messages`. It returns three attributes now.
- `appendJSON` marshals before it checks the size, so a tool list over the 60 KiB bound costs the full encode and allocation on the request path and is then discarded. Pre-existing in shape, but this attribute makes it much easier to reach with a large MCP toolset.
- One new test builds the key as `attribute.Key("gen_ai.tool.definitions")` while its two siblings use `genAIToolDefinitions`. Having one of each is defensible, since the literal pins the wire name independently, so treat this as a note rather than a request.

Two things I checked that are fine, in case they come up. The `ParametersJsonSchema` branch is genuinely conformant, including for nested and repeated struct arguments — the generator inlines everything and emits no `$defs` or `$ref`. And populating `description` and `parameters` at all is the right call despite the convention's "not recommended by default" note, because content capture is already the opt-in that note asks for.

wmh123wmh1:
Addressed the review comments in commit `0d392ab`:

- kept semconv on `v1.36.0` and retained the literal `gen_ai.tool.definitions` key;
- emit tool name, description, and type by default;
- added an explicit opt-in for recording parameters;
- normalized GenAI schema enum values to valid lowercase JSON Schema types;
- isolated malformed parameter schemas so they do not drop all tool definitions;
- kept provider-native built-in tools out of this PR to keep the change focused; I can follow up with a separate PR for that support.

karolpiotrowicz (dismissed):
The schema dialect problem is fixed, and fixed on the path that mattered. Running an agent with a sub-agent and a function tool through `runner.Run`, `transfer_to_agent` now emits `"type":"object"` and `"type":"string"`, and this package's own `checkEnumFields` helper — which failed on the previous head with `field "type" leaks the protobuf wire name "OBJECT" into telemetry` — passes on the emitted value unchanged. The test was corrected rather than deleted, and the per-declaration encoding does contain a bad schema to its own entry. Six smaller things remain.

**The functional test now fails for anyone with the new variable exported.** [`agent_telemetry_schema_test.go:72-79`](https://github.com/google/adk-go/blob/0d392abf9f00880d518cce5255b1f151ba2e4be8/internal/telemetry/functionaltest/agent_telemetry_schema_test.go#L72-L79) pins only `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` before calling `telemetry.ApplyEnv()`, and [`ApplyEnv`](https://github.com/google/adk-go/blob/0d392abf9f00880d518cce5255b1f151ba2e4be8/internal/telemetry/logger.go#L287-L290) now reads a second variable straight from the process environment. The unit-test helper at [`genaimessages_test.go:44-48`](https://github.com/google/adk-go/blob/0d392abf9f00880d518cce5255b1f151ba2e4be8/internal/telemetry/genaimessages_test.go#L44-L48) already clears it, so this is the one site that was missed:

```
$ ADK_INSTRUMENTATION_GENAI_CAPTURE_TOOL_DEFINITION_PARAMETERS=true go test ./internal/telemetry/...
ok      google.golang.org/adk/v2/internal/telemetry
--- FAIL: TestTelemetrySchema_AgentWithTool/capture_content
    agent_telemetry_schema_test.go:98: telemetry mismatch (-want +got):
        "gen_ai.tool.definitions": strings.Join({
            `[{"name":"some_tool","description":"A sample tool."`,
+           `,"parameters":{"additionalProperties":false,...}`,
            `,"type":"function"}]`,
        }, ""),
```

**The normalizer rewrites parts of the schema that are data, not types.** [`lowercaseSchemaTypes`](https://github.com/google/adk-go/blob/0d392abf9f00880d518cce5255b1f151ba2e4be8/internal/telemetry/genaimessages.go#L195-L218) lowercases the `type` member of every object it meets and then recurses into every child, so it descends into draft-07 keywords whose contents are instance values. A `default` or `examples` object carrying its own `type` field comes out altered, and `dependencies` — which maps a property name to a list of property names — has its targets lowercased when a property happens to be called `type`. The span then reports a default the tool never declared:

```
in:  {"type":"object","properties":{"env":{"type":"object",
       "default":{"type":"PRODUCTION","region":"US-EAST1"},
       "examples":[{"type":"STAGING"}]}},
     "dependencies":{"type":["billingAddress"]}}
out: ..."default":{"region":"US-EAST1","type":"production"},
       "examples":[{"type":"staging"}],
     "dependencies":{"type":["billingaddress"]}
```

`enum` and `const` come through untouched. Matching against the seven `genai.Type` spellings instead of lowercasing whatever is there would also handle `genai.TypeUnspecified`, which currently becomes `"type_unspecified"` — not one of draft-07's types.

**The value substituted for a bad schema is a string where the spec wants a schema.** The convention types `parameters` as `anyOf[draft-07 schema, null]` and says the value must conform to draft-07, whose meta-schema allows an object or a boolean. [`toolDefinitionParameters`](https://github.com/google/adk-go/blob/0d392abf9f00880d518cce5255b1f151ba2e4be8/internal/telemetry/genaimessages.go#L175-L189) emits `"<unserializable>"`, and [`genaimessages_test.go:463`](https://github.com/google/adk-go/blob/0d392abf9f00880d518cce5255b1f151ba2e4be8/internal/telemetry/genaimessages_test.go#L461-L470) asserts it as expected, which is the shape the last round flagged on `"OBJECT"`. `null` already means "this tool declares no parameters", so an object is the free choice — adk-python emits `{"type":"object","properties":{"serialization_error":…}}` here. Reusing the package's placeholder elsewhere is fine, since `arguments` and `response` are not typed as schema documents.

**Large integers no longer survive the round trip.** Decoding into `any` at [`genaimessages.go:180-188`](https://github.com/google/adk-go/blob/0d392abf9f00880d518cce5255b1f151ba2e4be8/internal/telemetry/genaimessages.go#L180-L188) turns every number into a `float64`, so `"maximum": 9223372036854775807` is recorded as `9223372036854776000`. It was exact on the previous head, and `json.Decoder` with `UseNumber()` keeps the literal.

**Three behaviors have no test, each one table row.** Setting both `Parameters` and `ParametersJsonSchema` on one declaration, so the precedence is pinned. A declaration with neither, which emits `"parameters":null` rather than omitting the key. And a union type such as `{"type":["OBJECT","null"]}`, which a remote MCP server can legitimately send — the branch handling it can be deleted without any test noticing.

**One question rather than a finding.** The new opt-in is squarely within what the convention allows, which says instrumentations may provide a way to enable these properties. But adk-python has no second knob — it emits `parameters` whenever content capture is on and nulls them when it is off. Since this variable would be the first `ADK_INSTRUMENTATION_*` name in the repo, worth a maintainer deciding whether the two implementations should agree before it ships.

Deferring the provider-side built-ins seems right to me — the emitted array is valid without them, and this file already documents the same deferral for provider-side tool parts. Please do open the follow-up, though: a `google_search`-only agent records no attribute at all, which reads identically to a request that offered the model no tools.

wmh123wmh1:
Thanks for the detailed follow-up review. I addressed the remaining implementation and test findings in commit `c549547`:

- clear `ADK_INSTRUMENTATION_GENAI_CAPTURE_TOOL_DEFINITION_PARAMETERS` in the functional telemetry test;
- normalize only the known GenAI schema enum spellings, including union types, while leaving data under `default`, `examples`, and `dependencies` unchanged;
- use a valid JSON Schema object instead of a string when a parameter schema cannot be serialized;
- decode schemas with `json.Decoder.UseNumber()` so large integers retain their original precision;
- add coverage for `Parameters`/`ParametersJsonSchema` precedence, missing parameters (`parameters: null`), union types, unchanged data-bearing fields, serialization failures, and large integers.

For the opt-in question, I kept the separate parameter-capture switch in this PR because the agreed scope for this change is to capture tool name/type/description by default and put parameter schemas behind an explicit opt-in. I am happy to adjust this if maintainers prefer Go to match adk-python here.

I also kept provider-native built-in tools out of this PR as suggested; those can be handled in a follow-up.

Validation completed:
- `go test -count=1 ./internal/telemetry/...`
- `go test -race -count=1 ./internal/telemetry/...`
- `go vet ./internal/telemetry/...`
- `go mod tidy -diff`

The full workspace race suite passed the packages reached before `session/vertexai`, where it timed out after 10 minutes while dialing the Vertex AI gRPC service in this environment; this is unrelated to the telemetry changes.

internal/telemetry/genaimessages.go
```diff
@@ -117,8 +124,170 @@ type toolResponsePart struct {
...
+		return json.RawMessage(unserializableSchemaPlaceholder)
+	}
+
+	var normalized any
+	decoder := json.NewDecoder(bytes.NewReader(encoded))
+	decoder.UseNumber()
+	if err := decoder.Decode(&normalized); err != nil {
+		return json.RawMessage(unserializableSchemaPlaceholder)
+	}
+	normalizeSchemaTypes(normalized)
```
karolpiotrowicz, line 221: `ParametersJsonSchema` is an `any`, so a caller can put a string, a number or an array in it. Those marshal and decode without error, `normalizeSchemaNode` returns immediately because the decoded value is not a map, and the value reaches the attribute unchanged: `"oops"` comes out as `"parameters":"oops"`, `42` as `"parameters":42`, and `["a","b"]` as the array. Draft-07 allows an object or a boolean, so none of the first three is a schema. adk-python guards this case in `_clean_parameters`, returning the same `serialization_error` document you already have as a constant in this file. A type check on `normalized` after the decode would cover it.

Narrow, since every in-tree writer of `ParametersJsonSchema` supplies an object — the exposure is a caller hand-building a genai config.

internal/telemetry/genaimessages.go
```diff
@@ -117,8 +124,170 @@ type toolResponsePart struct {
...
+	if typeValue, ok := obj["type"]; ok {
+		if normalized, keep := normalizeSchemaTypeValue(typeValue); keep {
+			obj["type"] = normalized
+		} else {
+			delete(obj, "type")
+		}
+	}
+	for key, child := range obj {
+		switch key {
+		case "properties", "patternProperties", "dependentSchemas", "definitions", "$defs":
```
karolpiotrowicz, line 254: `dependencies` is the one draft-07 keyword that can hold a subschema and is missing from this list, while five keywords from later drafts are in it. In draft-07 a `dependencies` value is either an array of property names or a subschema, and the two are discriminable by type, so both can be handled at once: recurse into a `map[string]any` value and leave a `[]any` value alone.

As it stands, `{"type":"OBJECT","dependencies":{"credit_card":{"type":"OBJECT"}}}` emits with the inner `"OBJECT"` intact. The previous head normalized that case and corrupted the array form, so this is a trade rather than a straight loss, and the producer has to be writing genai spellings into a hand-built schema to reach it.

internal/telemetry/genaimessages_test.go
```diff
@@ -301,6 +310,298 @@ func TestRequestContentAttributes_ToolTurnRole(t *testing.T) {
...
+func TestToolDefinitionParameters_NormalizesSchemaTypesWithoutChangingData(t *testing.T) {
+	captureToolDefinitionParameters(t)
+
+	declaration := &genai.FunctionDeclaration{
+		ParametersJsonSchema: map[string]any{
+			"type": "OBJECT",
+			"properties": map[string]any{
+				"value": map[string]any{
+					"type":     []string{"OBJECT", "null"},
+					"default":  map[string]any{"type": "PRODUCTION", "region": "US-EAST1"},
```
karolpiotrowicz: Every value in a data-bearing position here is one the type mapper would never rewrite. `PRODUCTION`, `US-EAST1`, `STAGING` and `billingAddress` all sit outside the seven genai spellings, so `normalizeSchemaTypeValue` returns them verbatim no matter what the walk does. Replacing the keyword-restricted walk in `normalizeSchemaNode` with blanket recursion into every child map — exactly the corruption this round fixes — leaves the whole package green.

Putting genai spellings in these positions instead, `"default": {"type": "OBJECT"}` and `"examples": [{"type": "STRING"}]`, makes the test fail under blanket recursion, which is what the name promises.

internal/telemetry/genaimessages_test.go
```diff
@@ -301,6 +310,298 @@ func TestRequestContentAttributes_ToolTurnRole(t *testing.T) {
...
+		},
+		{
+			"name":        "no_parameters",
+			"description": "",
+			"parameters":  nil,
+			"type":        "function",
+		},
+	}
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("tool definition parameter precedence and absence (-want +got):\\n%s", diff)
```
karolpiotrowicz: `\\n` here and at line 569 is a literal backslash followed by `n` rather than a newline, so the diff prints glued onto the message on a single line. The sibling assertions at 309, 367, 412 and 473 use `\n`. `go vet` does not diagnose it.

karolpiotrowicz (dismissed):
Everything raised last round is fixed, each one checked by running it rather than by reading the diff. The placeholder for an unencodable schema is now a schema object instead of a bare string, large integers survive the round trip, `TYPE_UNSPECIFIED` is dropped rather than emitted as an invalid type name, the normalizer no longer touches `default`, `examples` or `dependencies` property-name arrays, and the functional test pins the new variable so an exported value no longer fails an unrelated test. Driving an agent with a sub-agent and a function tool through `runner.Run` with the opt-in on, both declarations emit lowercase draft-07 types. The three tests added here fail against the previous head and pass against this one.

One thing still needs fixing before this merges, and it is not a repeat of anything from the earlier rounds.

## The two parameter fields are resolved the opposite way from adk-python and from this repo's own OpenAI converter

[`genaimessages.go:177-180`](https://github.com/google/adk-go/blob/c54954706c369c7e999b41f4283b7f6699729459/internal/telemetry/genaimessages.go#L177-L180) prefers `ParametersJsonSchema` and falls back to `Parameters`. adk-python resolves it the other way at [`_experimental_semconv.py:418`](https://github.com/google/adk-python/blob/c506ddf3bc34a6312ffc81899221bfa3f2da3b1d/src/google/adk/telemetry/_experimental_semconv.py#L418), `parameters = fd.parameters or fd.parameters_json_schema`, and so does [`model/openaimodel/tools.go:78-86`](https://github.com/google/adk-go/blob/c54954706c369c7e999b41f4283b7f6699729459/model/openaimodel/tools.go#L78-L86), which calls `schemaToMap(fn.Parameters)` first and consults `fn.ParametersJsonSchema` only when that comes back nil. The precedence line predates this commit. What is new is [`genaimessages_test.go:505-518`](https://github.com/google/adk-go/blob/c54954706c369c7e999b41f4283b7f6699729459/internal/telemetry/genaimessages_test.go#L505-L518), which asserts `{"type":"string"}` for a declaration whose `Parameters` says `OBJECT`, so the divergence is now pinned by a test.

The two fields are documented as mutually exclusive, which makes it tempting to read this as unreachable. It is not. [`functioncallmodifier/plugin.go:73-75`](https://github.com/google/adk-go/blob/c54954706c369c7e999b41f4283b7f6699729459/plugin/functioncallmodifier/plugin.go#L73-L75) assigns `decl.Parameters` when it is nil, on declarations that a `functiontool` has already given a `ParametersJsonSchema`. Registering that plugin and driving the same scenario through `runner.Run`:

```
declaration reaching the model:
  Parameters           = {"properties":{"tenant_id":{"description":"injected by the plugin","type":"STRING"}},"type":"OBJECT"}
  ParametersJsonSchema = {"type":"object","properties":{"arg1":{"type":"string"}},"required":["arg1"],"additionalProperties":false}

gen_ai.tool.definitions recorded on the span:
  [{"name":"some_tool","description":"A sample tool.","parameters":{"additionalProperties":false,"properties":{"arg1":{"type":"string"}},"required":["arg1"],"type":"object"},"type":"function"}]
```

The span records the tool's own schema with no `tenant_id`. The OpenAI adapter, handed that same declaration, builds its request from `Parameters` — the plugin-injected one. The attribute therefore reports a schema the request does not carry, with nothing invalid supplied by the caller. The change is one line plus the `want` in the new test.

## Two smaller things that span more than one file

There is still no end-to-end coverage of the opt-in path. [`agent_telemetry_schema_test.go:85`](https://github.com/google/adk-go/blob/c54954706c369c7e999b41f4283b7f6699729459/internal/telemetry/functionaltest/agent_telemetry_schema_test.go#L85) pins the variable off in both subtests, which is the right fix for the hermeticity problem, and the golden value at [`agent_with_tool_capture_content.go:28`](https://github.com/google/adk-go/blob/c54954706c369c7e999b41f4283b7f6699729459/internal/telemetry/telemetrytestcase/agent_with_tool_capture_content.go#L28) carries no `parameters`, so nothing exercises the emitted attribute with parameters present outside the unit tests. A third table case with the variable set and a digest that includes `parameters` would close it.

The new variable is still absent from every markdown file in the repo and still has no adk-python counterpart. That is a call for a maintainer to make before this ships rather than after, since the name becomes public the moment it lands.

## Verification

The build and test matrix on this pull request is gated behind first-time-contributor approval and has not run on any of the three heads, so the checks currently shown cover no Go code. Building and testing both modules on this head against the merge base, with `plugin/agentanalytics` run separately because `./...` does not cross a nested `go.mod`: build, vet and gofmt clean on both sides, and both telemetry packages pass. Two failures in the full suite, neither attributable to this change — `telemetry.TestConfigureExporters` fails identically on the merge base because it reaches the metadata server, and `agent/remoteagent/v2 TestA2AMultiHopInputRequiredCancellation` passed three runs out of three in isolation on both this head and the base.

wmh123wmh1:
Thanks for the follow-up review. I pushed the requested changes in commits `ffd7334` and `e5e12c5`.

- `Parameters` now takes precedence over `ParametersJsonSchema`, with fallback when `Parameters` is nil.
- Added unit-test coverage for the precedence and fallback behavior.
- Added an end-to-end test for tool parameter capture.
- Documented `ADK_INSTRUMENTATION_GENAI_CAPTURE_TOOL_DEFINITION_PARAMETERS`.
- Clarified that span content capture must also be enabled for parameter capture to take effect.

The telemetry tests, race tests, vet, build, and `go mod tidy -diff` checks pass locally.

Could you please take another look at the latest revision?

README.md
```diff
@@ -47,6 +47,24 @@ To add ADK Go to your project, run:
 go get google.golang.org/adk/v2
 ```
 
+## 🔭 Telemetry
+
+ADK records tool names, descriptions, and types in `gen_ai.tool.definitions`
```
karolpiotrowicz: This reads as covering every tool, but `toolDefinitions` iterates only `tool.FunctionDeclarations`. Provider-side tools reach `Config.Tools` as a bare `*genai.Tool` with none of those, so `google_search`, `code_execution` and `url_context` contribute nothing, and an agent whose only tool is one of them records no attribute at all. We agreed last round to defer that in the code, which is fine, but the deferral was about the behavior and this sentence is new. Scoping it to function tools would keep the two consistent.

internal/telemetry/functionaltest/agent_telemetry_schema_test.go
```diff
@@ -76,6 +89,13 @@ func TestTelemetrySchema_AgentWithTool(t *testing.T) {
 			} else {
 				t.Setenv(captureMessageContentEnvVar, "")
 			}
+			// Keep this end-to-end test independent of a developer's local
+			// tool-schema capture setting.
+			if tc.captureParameters {
```
karolpiotrowicz, line 97: `t.Setenv` restores the process environment when the subtest ends, but nothing re-runs `ApplyEnv`, so the package globals it derived stay as this case left them for the rest of the binary. The previous head already leaked `captureMode` the same way, so this is not new, and it is invisible today because the only other test in the package pins content capture off and re-applies. It becomes visible the moment a third file here asserts on span attributes without calling `ApplyEnv` first. `t.Cleanup(telemetry.ApplyEnv)` right after the call on the next line makes the restore that `t.Setenv` already does reach the globals too.

internal/telemetry/genaimessages_test.go
```diff
@@ -301,6 +310,310 @@ func TestRequestContentAttributes_ToolTurnRole(t *testing.T) {
...
+		},
+		{
+			"name":        "json_schema_only",
+			"description": "",
+			"parameters":  map[string]any{"type": "string"},
+			"type":        "function",
+		},
+	}
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("tool definition parameter precedence and absence (-want +got):\\n%s", diff)
```
karolpiotrowicz: `\\n` in an interpreted string literal is a backslash followed by an `n`, so the whole `cmp` diff prints on the label line instead of below it. Same on line 581. The four sibling assertions in this file all use `\n`, and neither gofmt nor vet can see the difference.

internal/telemetry/genaimessages.go
```diff
@@ -117,8 +124,170 @@ type toolResponsePart struct {
...
+		normalizeSchemaNode(value)
+	case []any:
+		for _, schema := range value {
+			normalizeSchemaNode(schema)
+		}
+	}
+}
+
+func normalizeSchemaTypeValue(value any) (any, bool) {
+	genaiTypeNames := map[string]string{
```
karolpiotrowicz: This table never changes, but it is a composite literal inside the function body, so Go allocates it afresh for every object that has a `type` key and again for every element of a `type` array. On a large toolset that is thousands of seven-entry maps per model call, all discarded. A package-level `var` is read-only after init and safe for concurrent request goroutines.

karolpiotrowicz (dismissed):
All three items from the last round are fixed, each checked by running it rather than by reading the diff.

The precedence at [`genaimessages.go:177-180`](https://github.com/google/adk-go/blob/e5e12c5eda3601c9a7f47e9fdcdade8eda9850b0/internal/telemetry/genaimessages.go#L177-L180) now resolves the two fields the same way [`model/openaimodel/tools.go:78-87`](https://github.com/google/adk-go/blob/e5e12c5eda3601c9a7f47e9fdcdade8eda9850b0/model/openaimodel/tools.go#L78-L87) does. Driving an agent with a `functiontool` tool and `plugin/functioncallmodifier` installed through `runner.Run`, and reading the request off the mock model in the same run, the declaration reaching the model carries both fields and the span now reproduces the one the OpenAI adapter builds its request from:

```
Parameters           = {"properties":{"tenant_id":{"description":"injected by the plugin","type":"STRING"}},"type":"OBJECT"}
ParametersJsonSchema = {"type":"object","properties":{"arg1":{"type":"string"}},"required":["arg1"],"additionalProperties":false}

gen_ai.tool.definitions = [{"name":"some_tool","description":"A sample tool.","parameters":{"properties":{"tenant_id":{"description":"injected by the plugin","type":"string"}},"type":"object"},"type":"function"}]
```

That test fails against the previous head ([c5495470](https://github.com/google/adk-go/commit/c54954706c369c7e999b41f4283b7f6699729459)) and passes against this one ([e5e12c5e](https://github.com/google/adk-go/commit/e5e12c5eda3601c9a7f47e9fdcdade8eda9850b0)). The new [`capture_content_and_tool_parameters`](https://github.com/google/adk-go/blob/e5e12c5eda3601c9a7f47e9fdcdade8eda9850b0/internal/telemetry/functionaltest/agent_telemetry_schema_test.go#L76-L81) case is a real gate and not just a snapshot — stopping the emitter from including parameters turns it red on this head, and left the previous two-case table green. And the variable is now in `README.md`, which is the first markdown mention of either capture variable in the repo.

Build, vet, gofmt and both telemetry packages are clean on this head, on the merge base and on the merge result, in all three cases with `plugin/agentanalytics` run separately because `./...` does not cross a nested `go.mod`. The one failing test in the full suite, `telemetry.TestConfigureExporters`, fails identically on all three because it reaches the metadata server.

Three things worth a look before this merges. None of them blocks, and none was introduced by the two commits in this round.

## Turning the opt-in on can make the whole attribute disappear

[`appendJSON`](https://github.com/google/adk-go/blob/e5e12c5eda3601c9a7f47e9fdcdade8eda9850b0/internal/telemetry/genaimessages.go#L563-L566) drops an oversized attribute rather than trimming it, which is deliberate and predates this change. What is new is that `gen_ai.tool.definitions` is now the attribute most likely to reach the bound, and it reaches it only when the opt-in is on. So enabling parameter capture can remove the tool names and descriptions that were being recorded with it off, with nothing on the span to say why. With an ordinary toolset — 30 tools, 20 string properties each:

```
ADK_INSTRUMENTATION_GENAI_CAPTURE_TOOL_DEFINITION_PARAMETERS=       attribute emitted, 1801 bytes
ADK_INSTRUMENTATION_GENAI_CAPTURE_TOOL_DEFINITION_PARAMETERS=true   attribute absent
```

Nothing that exists on `main` today is lost, since the attribute is itself new here, so this is a wart in a new feature rather than a regression. Falling back to names and descriptions when the with-parameters encoding does not fit would keep the failure graceful, and either way the comment on `appendJSON` is now slightly wrong: pre-encoding isolates a bad payload from its siblings, but it does not isolate a large one.

## The README leaves out the value most existing deployments already have set

Every sentence in the new section is true. The gap is the one it does not have: a truthy `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true` parses to log events only, at [`logger.go:68-81`](https://github.com/google/adk-go/blob/e5e12c5eda3601c9a7f47e9fdcdade8eda9850b0/internal/telemetry/logger.go#L68-L81), and never puts content on spans, so the new flag does nothing alongside it. That is exactly the value this repo's own deploy path writes into every Agent Engine deployment, at [`agentengine.go:349`](https://github.com/google/adk-go/blob/e5e12c5eda3601c9a7f47e9fdcdade8eda9850b0/cmd/adkgo/internal/deploy/agentengine/agentengine.go#L349). Someone in that population who reads [`README.md:56-66`](https://github.com/google/adk-go/blob/e5e12c5eda3601c9a7f47e9fdcdade8eda9850b0/README.md#L56-L66) and adds only the second `export` line gets silence and no warning. One sentence saying the variable must be `SPAN_ONLY` or `SPAN_AND_EVENT` would close it.

## The end-to-end case does not cover the interaction the last round was about

The scenario tool is built by `functiontool`, which sets only `ParametersJsonSchema` at [`function.go:166`](https://github.com/google/adk-go/blob/e5e12c5eda3601c9a7f47e9fdcdade8eda9850b0/tool/functiontool/function.go#L166), so `Parameters` is nil on every declaration and the changed line takes the same branch under the old predicate as under the new one. Reverting that one line leaves the whole `functionaltest` package green and fails only the unit test. The precedence is genuinely pinned, by [`TestRequestContentAttributes_ToolDefinitions_ParametersPrecedenceAndMissing`](https://github.com/google/adk-go/blob/e5e12c5eda3601c9a7f47e9fdcdade8eda9850b0/internal/telemetry/genaimessages_test.go#L477-L534), and what the last round asked to see end to end was the opt-in, which the new case does cover. So this is a note rather than a request. A functional case with the plugin installed would pin the interaction end to end if you think it is worth the wiring.

## One thing that is not yours to decide

The variable still has no adk-python counterpart, and that is a maintainer call rather than something to fix in this pull request. Nothing else here waits on it — the in-repo comparator the precedence now matches is `model/openaimodel/tools.go`, which agrees regardless of how that lands.

wmh123wmh1:
Thanks for the detailed follow-up review. I addressed the remaining inline findings in commit `0693c97`:

- reject non-object/non-boolean `ParametersJsonSchema` values with a valid schema placeholder;
- normalize schema-valued draft-07 `dependencies` entries while preserving property dependency arrays;
- strengthen the tests to ensure data under `default` and `examples` is not modified;
- move the GenAI type mapping to package scope to avoid repeated allocations;
- restore the telemetry package state after each functional subtest;
- fix the escaped newlines in the test failure messages;
- clarify that `gen_ai.tool.definitions` currently covers function tools only;
- document that `true` is equivalent to `EVENT_ONLY`, while span attributes require `SPAN_ONLY` or `SPAN_AND_EVENT`.

I left the oversized-attribute fallback unchanged because it predates this change and was noted as non-blocking. I’m happy to address it in a separate follow-up if maintainers prefer.

I also left the plugin-based functional scenario unchanged because the precedence behavior is already covered by the focused unit test. The Go/Python configuration alignment remains a maintainer decision, as suggested.

Validation completed:

- `go test -count=1 ./internal/telemetry/...`
- `go test -race -count=1 ./internal/telemetry/...`
- `go vet ./internal/telemetry/...`
- `golangci-lint v2.3.1 run` in both modules
- `git diff --check`

Could you please take another look and update the review status when convenient?

internal/telemetry/genaimessages.go
```diff
@@ -117,8 +134,177 @@ type toolResponsePart struct {
...
+	decoder := json.NewDecoder(bytes.NewReader(encoded))
+	decoder.UseNumber()
+	if err := decoder.Decode(&normalized); err != nil {
+		return json.RawMessage(unserializableSchemaPlaceholder)
+	}
+	switch normalized.(type) {
+	case map[string]any, bool:
+		// JSON Schema draft-07 permits schemas to be objects or booleans.
+	default:
+		return json.RawMessage(unserializableSchemaPlaceholder)
```
karolpiotrowicz, line 219: A JSON `null` decodes to an untyped nil, so it matches neither arm above and lands here, and the span ends up asserting the tool takes a `serialization_error` string parameter. The conventions allow `parameters: null`, and line 184 already emits it for the no-parameters case, so this contradicts a value the same function treats as valid a few lines up. `case nil:` returning `json.RawMessage("null")` covers it.

This is new in this commit — on `e5e12c5e` a `json.RawMessage("null")`, a typed-nil pointer and a typed-nil map all emitted `null`, and on this head the last two emit the placeholder.

internal/telemetry/genaimessages.go
```diff
@@ -117,8 +134,177 @@ type toolResponsePart struct {
...
+			if keep {
+				normalized = append(normalized, item)
+			}
+		}
+		if len(normalized) == 0 {
+			return nil, false
+		}
+		return normalized, true
+	default:
+		return value, true
```
karolpiotrowicz: A non-string `type` is passed through unchanged, so `{"type": 42}` reaches the attribute as `{"type":42}` and `{"type": null}` as `{"type":null}`. Neither is valid draft-07, where `type` must be a simple-type string or an array of them, so a consumer validating the attribute against `gen-ai-tool-definitions.json` rejects the whole array rather than just that tool.

The application already supplied a non-conformant schema in that case, so this is bounded. Returning `nil, false` here would drop the key instead, which is the one keyword this function exists to police.

internal/telemetry/genaimessages.go
```diff
@@ -117,8 +134,177 @@ type toolResponsePart struct {
...
+func normalizeSchemaTypeValue(value any) (any, bool) {
+	switch value := value.(type) {
+	case string:
+		if normalized, ok := genaiTypeNames[value]; ok {
+			return normalized, true
+		}
+		if value == string(genai.TypeUnspecified) {
+			return nil, false
+		}
+		return value, true
```
karolpiotrowicz: Only the exact protobuf spellings are mapped, so a mixed-case `"String"` or `"Object"` reaches the span uncorrected. `lowercaseSchemaTypes` in `model/openaimodel/tools.go` calls `strings.ToLower` unconditionally at this point, so the two now describe the same declaration differently.

This is only reached for the `type` key at a genuine schema location, so lowering it here would not touch the data-bearing keywords the walker is careful to skip.

internal/telemetry/genaimessages.go
```diff
@@ -69,6 +81,11 @@ const maxContentAttributeBytes = 60 << 10
 // NaN, an Inf, a func, a chan or a reference cycle.
 const unserializablePlaceholder = `"<unserializable>"`
 
+// unserializableSchemaPlaceholder is a valid JSON Schema used when a tool
+// parameter schema cannot be encoded. Unlike unserializablePlaceholder, this
+// value is used where the semantic convention requires a schema object.
+const unserializableSchemaPlaceholder = `{"type":"object","properties":{"serialization_error":{"type":"string"}}}`
```
karolpiotrowicz: The placeholder is a well-formed schema, which is what the convention needs, but it is also indistinguishable from a real one: a consumer reconstructing the tool contract sees a tool that genuinely accepts a `serialization_error` string. Its sibling `unserializablePlaceholder` on line 82 is unmistakable by comparison. A `$comment` or `description` naming it as a placeholder would keep it a valid schema object and stop it asserting a parameter that does not exist.

internal/telemetry/genaimessages_test.go
```diff
@@ -44,6 +44,15 @@ func captureContent(t *testing.T) {
 	// Spans have to be asked for by name. A truthy value means log records
 	// only, which is what the variable meant before spans could carry content.
 	t.Setenv(captureMessageContentEnvVar, "SPAN_AND_EVENT")
+	t.Setenv(captureToolDefinitionParametersEnvVar, "")
```
karolpiotrowicz, line 50: Adding the explicit clear here is a real improvement — it stops a developer's exported value reaching these tests. The other half is still missing though: `t.Setenv` restores the environment at test end, but nothing re-runs `ApplyEnv`, so the package globals it derived keep their values for the rest of the binary. `functionaltest/agent_telemetry_schema_test.go` gets this right with a `t.Cleanup(telemetry.ApplyEnv)` registered before the first `t.Setenv`.

Nothing fails today, since every test in the package that reads the gates sets both variables. It matters more than usual here because CI runs with `-shuffle=on`.

karolpiotrowicz (dismissed):
The schema fixes from the last round hold up, and I checked the attribute against the OpenTelemetry GenAI conventions rather than just reading the diff: the name, the array shape, the required `type`/`name` pair and the optional `parameters` all match `gen-ai-tool-definitions.json`, and keeping parameter schemas behind a second opt-in is exactly what the spec's own note asks instrumentations to do. `url_context` really is a provider-native tool here, so that README line is right too.

Two things to sort out before this merges. The first is the attribute-deletion behaviour I blocked on last round, which needs the tool names and descriptions to survive the opt-in being switched on — by whatever means you prefer. The second is new in the latest commit.

## Turning the opt-in on can delete the whole attribute

This is the item I blocked on last time and it is unchanged. [`appendJSON`](https://github.com/google/adk-go/blob/0693c9797b5906f6ecb4b159474810fd28074e6c/internal/telemetry/genaimessages.go#L581-L582) drops an oversized attribute rather than trimming it, which is deliberate and predates this change. What is new is that [`requestContentAttributes`](https://github.com/google/adk-go/blob/0693c9797b5906f6ecb4b159474810fd28074e6c/internal/telemetry/genaimessages.go#L324-L326) hands it one value carrying names, descriptions, types *and* parameters, so the flag's subject and its casualty are different data. Enabling "also record parameter schemas" can remove the tool list that was being recorded without it.

I drove this end-to-end through `runner.Run` on this head, 30 tools with 20 properties each, varying only how long each property description is:

```
per-property description   opt-in OFF      opt-in ON
 0 bytes                   2191 bytes      26761 bytes
40 bytes                   2191 bytes      60961 bytes
80 bytes                   2191 bytes      attribute absent
```

Undocumented properties survive. Documented ones do not, and 20 tools with 20 described properties already reaches 56641 bytes against the 61440-byte bound. What the consumer sees at the point it goes is nothing at all — no truncated value, no marker, no span event, status still `Unset` — while `gen_ai.system_instructions`, `gen_ai.input.messages` and `gen_ai.output.messages` all remain. The span looks exactly like the span of a run that had no tools at all.

Nothing that exists on `main` today is lost, since the attribute is itself new here, so this is a wart in a new feature rather than a regression. But the README added by this PR tells an operator to set both variables with no caveat, so following the documented recipe can give them less than not following it. That is what I would like fixed before merge: the names and descriptions should still reach the span when the parameter schemas do not fit. Re-encoding with `Parameters` cleared when the full form is oversized is about five lines at the call site, and it is the option I checked, but the choice is yours. Please also document the bound in the README section and pin whichever policy you pick with a test — the only size test in the package covers `gen_ai.input.messages`, and nothing covers this attribute in either gate state.

## The new type guard turned JSON null into a fabricated schema

The guard added in this commit fixes the non-object case I raised last round, but it also catches JSON `null`. A `null` decodes to an untyped nil, matches neither arm of the switch, and falls through to the serialization-error placeholder. Same test against [the previous head](https://github.com/google/adk-go/commit/e5e12c5eda3601c9a7f47e9fdcdade8eda9850b0) and [this commit](https://github.com/google/adk-go/commit/0693c9797b5906f6ecb4b159474810fd28074e6c):

```
                           previous head   this commit
untyped nil, both fields   null            null
json.RawMessage("null")    null            {"type":"object","properties":{"serialization_error":{"type":"string"}}}
typed-nil pointer          null            {"type":"object","properties":{"serialization_error":{"type":"string"}}}
```

The span then reports a serialization failure that did not happen and claims the tool accepts a `serialization_error` string parameter. `null` is explicitly allowed for `parameters` by the conventions, and [the line just above](https://github.com/google/adk-go/blob/0693c9797b5906f6ecb4b159474810fd28074e6c/internal/telemetry/genaimessages.go#L183-L184) already emits it for the no-parameters case, so the guard contradicts a value the same function treats as valid elsewhere. Both in-repo tool producers guard the typed-nil case already, so reaching this needs a third-party `tool.Tool` — worth noting that `mcptoolset` guards it with a comment recording that this exact bug bit once before.

The remaining points are on the diff, apart from one that is not: `StartGenerateContentSpanParams.Request` in [`telemetry.go:89-92`](https://github.com/google/adk-go/blob/0693c9797b5906f6ecb4b159474810fd28074e6c/internal/telemetry/telemetry.go#L89-L92) still says the request is used "only" to record `gen_ai.input.messages` and `gen_ai.system_instructions`. That was accurate before this change and now it is not.

Two process notes that neither of us can see from the diff. The branch is behind `main`, and it conflicts with open #598 on `README.md`, so merge order decides who rebases. Separately, I have two change-request reviews still open on this PR, from the two previous rounds — dismissing one will leave the other in force.

wmh123wmh1:
Thanks for the detailed review. I addressed all of the requested changes in commit `7626403`.

- When parameter capture is enabled, tool definitions are first encoded with their parameter schemas. If the complete attribute exceeds the 60 KiB limit, ADK retries without the parameter schemas so that the tool name, description, and type are preserved. The attribute is omitted only if the metadata-only representation is also too large. The existing `appendJSON` behavior was left unchanged.
- `parameters: null`, `json.RawMessage("null")`, typed-nil pointers, and typed-nil maps are now preserved as JSON `null`.
- The serialization-error placeholder is now a valid and explicit JSON Schema that does not claim a nonexistent `serialization_error` parameter.
- Non-string `type` values are removed, invalid members of type arrays are filtered out, and mixed-case type values are normalized to lowercase.
- Schema-valued `dependencies` entries are normalized while property dependency arrays remain unchanged. Values under `default` and `examples` are not modified.
- Added regression coverage for parameter capture modes, size fallback, null values, invalid schemas, type normalization, large numbers, and telemetry environment-state restoration.
- Updated the `Request` documentation and README with the 60 KiB limit and fallback behavior.

The PR remains on semconv `v1.36.0`, uses the literal `gen_ai.tool.definitions` key, and continues to cover function tools only. Provider-native tools and the unrelated message-content enum issue remain outside this change.

Validation completed:

- `go test -count=1 ./internal/telemetry/...`
- `go test -race -count=1 ./internal/telemetry/...`
- `go test -shuffle=on -count=20 ./internal/telemetry/...`
- `go vet ./internal/telemetry/...`
- `go mod tidy -diff`
- `git diff --check`

Could you please take another look at the latest revision when convenient?

internal/telemetry/genaimessages.go
```diff
@@ -117,8 +134,180 @@ type toolResponsePart struct {
...
+func normalizeSchemaTypeValue(value any) (any, bool) {
+	switch value := value.(type) {
+	case string:
+		if value == string(genai.TypeUnspecified) {
+			return nil, false
+		}
+		if normalized, ok := genaiTypeNames[value]; ok {
+			return normalized, true
+		}
+		return strings.ToLower(value), true
```
karolpiotrowicz: An unrecognised string is lowercased and kept, so the emitted `parameters` is still not a draft-07 document — `{"type":"Widget"}` comes out as `{"type":"widget"}`, and `{"type":""}` survives untouched. That sits oddly next to the `default` arm at the end of the same switch, which now correctly drops `{"type":42}` to `{}`: junk that is not a string is removed, junk that is a string is kept and tidied.

Lowercasing first and then testing membership in the seven draft-07 simple types would give both arms the same rule, and it keeps the `{"type":"Object"}` case the golden pins, since `Object` lowercases into the set.

One more that slips past on the way in: the unspecified check a few lines above is case-sensitive, so `{"type":"type_unspecified"}` is not recognised as unspecified and is emitted as-is.

internal/telemetry/genaimessages.go
```diff
@@ -117,8 +134,180 @@ type toolResponsePart struct {
...
+	case []any:
+		normalized := make([]any, 0, len(value))
+		for _, item := range value {
+			item, keep := normalizeSchemaTypeValue(item)
+			if keep {
+				normalized = append(normalized, item)
```
karolpiotrowicz: Two draft-07 details in the union arm, both only reachable from a hand-written schema.

The recursive call accepts a nested array, so `{"type":[["string"]]}` and `{"type":["string",["number"]]}` are emitted unchanged — draft-07 requires every member of a `type` array to be a string. Keeping an item only when the recursive result is a string would settle it.

Separately, lowercasing can now collapse two members into one value: `["String","STRING"]` becomes `["string","string"]`, and draft-07 requires the members to be unique. Before this commit the same input gave `["String","string"]`, which was unique but named a type that does not exist, so this is a sideways move rather than a regression — deduplicating while building the slice fixes both shapes at once.

internal/telemetry/genaimessages.go
```diff
@@ -117,8 +134,180 @@ type toolResponsePart struct {
...
+			// In draft-07 a dependency is either a property-name array or a
+			// schema. Normalize only schema-valued dependencies.
+			if dependencies, ok := child.(map[string]any); ok {
+				for _, dependency := range dependencies {
+					if schema, ok := dependency.(map[string]any); ok {
+						normalizeSchemaNode(schema)
+					}
+				}
+			}
+		case "additionalProperties", "additionalItems", "items", "contains", "not", "if", "then", "else", "propertyNames", "unevaluatedProperties", "unevaluatedItems":
```
karolpiotrowicz, line 270: `items` here and `anyOf` on line 261 are two of the three nested-schema fields a `genai.Schema` can actually produce, and the golden in `genaimessages_test.go` reaches neither — it exercises `properties` and `dependencies` only.

Deleting `"items"` from this case and `"anyOf"` from the one below leaves `go test ./internal/telemetry/...` green. With them gone, a tool whose schema has `{"type":"ARRAY","items":{"type":"STRING"}}` records the protobuf spelling straight onto the span, which is the thing this normalizer exists to prevent.

The behaviour is correct today — object-form `items`, tuple-form `items` and `anyOf` all normalize properly, including through the typed `genai.Schema` path. It is only unpinned. Adding a nested uppercase type under each to the existing golden covers both shared code paths.

internal/telemetry/genaimessages.go
```diff
@@ -135,6 +324,16 @@ func requestContentAttributes(req *model.LLMRequest) []attribute.KeyValue {
 			attrs = appendJSON(attrs, genAISystemInstructions, parts)
 		}
 	}
+	if definitions := toolDefinitions(req.Config); len(definitions) > 0 {
+		before := len(attrs)
+		attrs = appendJSON(attrs, genAIToolDefinitions, definitions)
```
karolpiotrowicz, line 351: Once a toolset is over the bound, the full encoding is built and thrown away on every recording span before the metadata-only retry runs, and nothing remembers the outcome between calls. Encoding the metadata-only form first, and attempting the full form only when that leaves headroom, would avoid the discarded pass.

I have not measured what it costs, and it is behind two opt-ins and `span.IsRecording()`, so treat this as an observation rather than a request.

karolpiotrowicz (changes requested):
Both things I blocked on last round are fixed, and I checked each by running it rather than by reading the diff.

A JSON-null parameter schema is emitted as `null` again. All eight ways a declaration can end up with a nil-ish schema — both fields unset, a `RawMessage("null")`, a typed-nil pointer, map, slice or `RawMessage`, an untyped nil, and a typed-nil `Parameters` — now produce `null`, matching the behaviour before the guard was added. Driving four of them through `runner.Run` with a real tool, the span itself carries `"parameters":null` rather than a fabricated schema.

The oversized case no longer costs you the tool list. [`requestContentAttributes`](https://github.com/google/adk-go/blob/7626403b217ee3a53066eaf05479c7ba02e0b60c/internal/telemetry/genaimessages.go#L327-L335) retries with the parameter schemas cleared, and because `Parameters` carries `omitempty` the retried value is exactly the value the opt-in-off path would have produced. Measured on emitted spans at eight configurations, from 2 tools to 200:

```
30x20 props, 80B descriptions     opt-in off 2131B    opt-in on 2131B    byte-identical
30x20 props, 4000B descriptions   opt-in off 2131B    opt-in on 2131B    byte-identical
200x20 props, 4000B descriptions  opt-in off 14301B   opt-in on 14301B   byte-identical
```

That gives the feature a property worth stating on its own, because it is what a telemetry consumer can now rely on and it is the thing future changes here should be held against: **with parameter capture enabled, `gen_ai.tool.definitions` is never less informative than with it disabled — the recorded value is either the full form, or byte-identical to the opt-in-off form.** Not a threshold, not a payload size. The same commit adds the README paragraph describing the retry-then-omit ladder, corrects the `StartGenerateContentSpanParams.Request` comment, and replaces the placeholder that used to advertise a `serialization_error` parameter with an inert `$comment` form.

## One thing to fix before merge

That invariant is not pinned for more than one tool, and it is the invariant three rounds of review have been about. The test needs to cover the multi-tool case before this merges.

[`genaimessages_test.go:601`](https://github.com/google/adk-go/blob/7626403b217ee3a53066eaf05479c7ba02e0b60c/internal/telemetry/genaimessages_test.go#L601) builds all four subtests from a single declaration. With one element, `for i := range definitions` and `definitions[0]` are the same program. Change the loop at [`genaimessages.go:331-333`](https://github.com/google/adk-go/blob/7626403b217ee3a53066eaf05479c7ba02e0b60c/internal/telemetry/genaimessages.go#L331-L333) to clear only `definitions[0]`, and `go test ./internal/telemetry/...` stays completely green while the original defect comes back:

```
                                    opt-in off          opt-in on, clearing only definitions[0]
30 tools x 20 props, 80B  descs     2131B, 30 tools     attribute absent
30 tools x 20 props, 4000B descs    2131B, 30 tools     attribute absent
600 tools, no props                 43101B, 600 tools   attribute absent
```

The shipped code is right — it does clear every definition, which is why the 200-tool row above holds. It is the test that cannot tell the difference, so nothing stops a later refactor from reintroducing this silently.

One extra declaration in the same table closes it. I ran this version to be sure it works and does not weaken anything else: a second oversized declaration, `wantDefinitions` on the table struct, and the per-definition assertion widened to check `name`, `type` and `description` on each entry rather than only on the first. The suite stays green, the `definitions[0]` edit above now fails, and nothing else that the table used to catch stops being caught. Shape it however you prefer, as long as the fallback runs with more than one declaration in the request.

## Smaller things, none of them blocking

I have left four notes inline. All four are reachable only when an application hands the instrumentation a schema that is already invalid, so nothing is lost today and none of them is a regression against `main`. They are worth a look together because the new `default: return nil, false` arm sets a rule — a value that cannot be a valid `type` is dropped — that the string arm and the array arm do not follow.

Worth knowing from outside the diff: the branch conflicts with open #598 on `README.md`, so whoever lands second will need to rebase. I also have three change-request reviews still open on this pull request from the earlier rounds, and every item behind all three is now fixed — I will clear them.

wmh123wmh1:
Thanks for the follow-up review. I addressed the remaining review items in commit `5ed36e6`.

- Expanded the tool-definition fallback test from one declaration to two declarations. The test now checks every definition's name, description, and type, and verifies that the oversized-parameter fallback is byte-identical to the parameter-capture-disabled output.
- Schema `type` values are now accepted only when they are one of the seven valid JSON Schema types. Unknown strings, empty strings, and case-insensitive `TYPE_UNSPECIFIED` values are removed.
- Type arrays now remove non-string and invalid members and deduplicate values after normalization. Empty arrays are removed from the schema.
- Added regression coverage for nested `items` and `anyOf` schemas.
- Existing handling for `dependencies`, `default`, and `examples` remains unchanged.

I left the metadata-first encoding optimization unchanged because it was noted as a non-blocking observation, keeping this change focused on the review requirements.

Validation completed:

- `go test -count=1 ./internal/telemetry/...`
- `go test -race -count=1 ./internal/telemetry/...`
- `go test -shuffle=on -count=20 ./internal/telemetry/...`
- `go vet ./internal/telemetry/...`
- `go mod tidy -diff`
- `git diff --check`

The PR remains on semconv `v1.36.0`, uses the literal `gen_ai.tool.definitions` key, and continues to cover function tools only.

Could you please take another look at the latest revision when convenient?

karolpiotrowicz:
@wolo-lab flagging this one as a scope call rather than a code review, because it decides what ADK Go's telemetry carries by default and that answer should match the other implementations.

The change records `gen_ai.tool.definitions` on the model request span, which is parity work. What is new is the second opt-in: parameter schemas are withheld unless a new ADK-Go-only variable is set on top of the OpenTelemetry content-capture setting, declared at [internal/telemetry/logger.go#L39-L43](https://github.com/google/adk-go/blob/5ed36e6ad4cbd21591ef4b51edce250c37f568ab/internal/telemetry/logger.go#L39-L43) and enforced in [internal/telemetry/genaimessages.go#L165-L182](https://github.com/google/adk-go/blob/5ed36e6ad4cbd21591ef4b51edce250c37f568ab/internal/telemetry/genaimessages.go#L165-L182). The README asks operators to export it alongside the OpenTelemetry one — [README.md#L61](https://github.com/google/adk-go/blob/5ed36e6ad4cbd21591ef4b51edce250c37f568ab/README.md#L61).

adk-python has no equivalent switch. Once content capture is on it emits the parameter schemas with the definition, at [_experimental_semconv.py#L412-L424](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/telemetry/_experimental_semconv.py#L412-L424), and it drops parameters only in the no-content view used when capture is off. So with the same OpenTelemetry setting in both runtimes, a Python agent's spans would carry tool parameter schemas and a Go agent's would not, until the operator learns about a Go-specific variable that exists nowhere else.

There is a real argument for the extra gate — schemas can be large and can describe internal systems. But choosing to be stricter than the other implementations by default is a cross-language decision, and I would rather you make it than have it land through a review.

The call I need from you: take it as-is with the extra opt-in, take it with the parameters emitted whenever content capture is on so the two runtimes agree, or decline the parameter capture entirely and keep only names, descriptions and types.

PR #1455, opened 2026-09-01
fix(adkrest): serve the endpoints the ADK web UI actually calls

jjsasha63:

None open for this.

**Problem:**

The web UI vendored in `cmd/launcher/web/webui/distr/` calls 42 endpoints. 32 of
them fail against the Go server: 29 return 404 and 3 return 405. Most of the dev
UI has been broken since 2026-06-30, and it went unnoticed because chat still
works.

Run any agent with `adk web ... api webui` and open `http://localhost:8080/ui/`.
On page load alone, 7 requests 404:

```
404 GET /api/version
404 GET /api/dev/apps/<app>/build_graph
404 GET /api/dev/apps/<app>/build_graph_image?dark_mode=false
404 GET /api/dev/apps/<app>/build_graph_image?dark_mode=true
404 GET /api/dev/apps/<app>/builder?ts=...
404 GET /api/dev/apps/<app>/eval_sets
404 GET /api/dev/apps/<app>/eval_results
```

The version in the header is blank, the agent structure panel says "Agent
structure graph not available", the Traces panel renders nothing, renaming a
session silently does nothing, and refreshing on a client-side route such as
`/ui/eval` shows Go's plain-text `404 page not found`.

**Root cause.** ADK split its API in two at v2: production endpoints kept their
paths, and the developer endpoints moved under `/dev/apps/{app_name}/`. adk-go
picked up the new bundle in 893e4a4 "V2 release (#1109)", which changed 105
files under `distr/` inside a 547-file squashed release whose message never
mentions the UI. The routers were never updated:

```
$ grep -rn '"/dev' --include=*.go .
(no output)
```

Chat kept working only because `/run_sse`, `/list-apps` and the session routes
are not part of the moved set.

`adk-js` hit exactly this and fixed it in google/adk-js#694, for the same two
graph endpoints.

**Solution:**

Register the dev paths alongside the pre-v2 ones, so existing clients keep
working. Implement `/version` and the two agent-graph endpoints. Answer features
ADK Go does not have (eval, agent-builder writes, the tests tab) with 501 and a
readable JSON body, rather than a bare 404 a client cannot tell from a wrong
path prefix — following the convention set by #188.

Defects found and fixed while verifying the above, each reproduced first:

1. Artifact routes panicked on a nil service and dropped the connection with no
   HTTP response (`curl` exit 52). The web launcher defaulted only
   `SessionService`, leaving the artifact and memory services nil; a nil memory
   service would also panic mid-run for any agent using `load_memory`.
2. `PUT` and `PATCH` were rejected at the API mount for every path, and `PATCH`
   on a session was never implemented, so renaming a session did nothing.
3. `HEAD` returned 405 on every route, including static UI files.
4. A 405 carried no `Allow` header, and most method mismatches degraded to a
   bare 404: gorilla/mux clears its recorded `ErrMethodMismatch` as soon as a
   later route's method matcher succeeds, so `DELETE /version` returned 404
   while `PATCH /version` returned 405, purely from registration order.
5. `OPTIONS` on a session or artifact path ran the `DELETE` handler. Reproduced
   on a bare `adkrest.NewServer`: create a session, send `OPTIONS`, it is gone.
   Masked behind the launcher because CORS answers `OPTIONS` first, but
   `adkrest.NewServer` is exported.
6. A trailing slash turned a `POST` into a `GET` against a different handler:
   `POST /apps/x/users/y/sessions/` returned a session list and created nothing.
7. `Access-Control-Allow-Origin` was emitted as a bare `localhost:8080` with no
   scheme, which browsers reject.
8. `GET /ui/assets/config/` served an HTML directory listing.
9. A missing session returned 500 rather than 404, in all three
   `session.Service` implementations; a missing artifact likewise. Adds
   `session.ErrNotFound`, enforced by the shared conformance suite.
10. `ListSessions` returned JSON `null` rather than `[]`.
11. The Traces panel was blank for three stacked reasons: span logs serialised
    as `null`; with content capture off a message body carried `content` as the
    string `<elided>` where the UI schema requires an object with a `parts`
    array and a required `role`; and a failed model call produces a
    `generate_content` span with no `gcp.vertex.agent.event_id`, which the
    schema requires.
12. `build_graph_image` ignored `dark_mode`, so the graph was unreadable on a
    light theme.
13. An unresolvable `?node=` returned the root agent's graph with 200 rather
    than 404, and duplicate agent names at different depths resolved to
    whichever a depth-first search reached first.

**Preventing recurrence.** `server/adkrest/uicontract_test.go` reads the shipped
bundle at test time, extracts every API call site with its HTTP verb, and
asserts each one both routes on a real server and appears in a golden list. It
catches an added endpoint, a renamed one, and a changed verb, in both
directions, and fails rather than skips when the bundle is missing. Proven
against a mutated copy of the bundle.

`scripts/adk-web/` now pins the upstream ref via `ADK_WEB_REF` and records the
resolved SHA into the distribution. The previous Dockerfile cloned `main`
unpinned behind a cache-buster and discarded the SHA, which is why a breaking
refresh could land unreviewed.

### Testing Plan

**Unit Tests:**

```
$ go build ./... && go vet ./...
$ go test -race -count=1 -shuffle=on ./...
74 packages ok, 0 failures
$ golangci-lint run ./...
0 issues.
$ go mod tidy -diff
(no output)
```

Both modules are clean, including `plugin/agentanalytics`.

New tests: `server/adkrest/uicontract_test.go`, plus coverage for the version,
dev-feature, agent-graph, artifact, debug-telemetry, session PATCH, launcher
service defaults, API mount and web UI serving changes. Every fix was checked by
reverting it and confirming a test fails.

```
go build -o /tmp/adk ./examples/quickstart
GOOGLE_API_KEY=... /tmp/adk web -port=8080 \
  api -webui_address=http://localhost:8080 -include_debug_api \
  webui -api_server_address=http://localhost:8080/api
```

Open `http://localhost:8080/ui/`.

| | Before | After |
|---|---|---|
| Page-load 404s | 7 | 0 |
| Console errors | 8 | 1 (the two deliberate eval 501s) |
| Version in header | blank | renders |
| Agent structure panel | error | renders, readable in both themes |
| Traces panel | blank | renders |
| Deep link `/ui/a/b/c` | plain-text 404 | renders the app |

Also verified by hand: `HEAD` on every read route, the full verb matrix against
`/api/version` and `/api/list-apps` returning 405 with `Allow`, `OPTIONS` no
longer deleting a session, trailing-slash `POST` creating a session rather than
listing, `PATCH` renaming a session and the name surviving a reload, and light
and dark graphs differing.

Note: the trace routes are behind `-include_debug_api` since #1413, so the
Traces panel needs that flag.

### Checklist

### Additional context

Two things worth a reviewer's attention:

- `session.ErrNotFound` is a new exported symbol in a high-fan-in package. It is
  additive and backward-compatible, and its doc comment is explicit about which
  operations report it and which errors it deliberately does not cover.
- `server/adkrest/controllers` gains several exported response types. That
  package's own TODO says it should not be public API; unexporting them is easy
  now and I am happy to do so if preferred.

The change is large because the endpoint work uncovered the rest. If you would
rather review it in pieces, the natural split is launcher
(`cmd/launcher/web/**`), REST routing (`server/adkrest/**`) and sessions
(`session/**`).


jjsasha63:
Superseded by a three-way split, as requested in review: #1458 (session), #1459 (adkrest), #1460 (launcher). Same change, plus the five review fixes — session PATCH tenancy, OPTIONS, the trace span stand-in, the adk-web pin disclosure, and the root /health endpoint disclosed in the launcher commit. Rebased onto 8a94a728.

PR #1456, opened 2026-09-01
fix(web): bind web server to loopback by default

AARUSHGAUR:
### Problem

`adk web` bound the web server to all interfaces (`Addr: ":port"`), exposing the ADK Web UI and every mounted API to the local network with no opt-in. The generated Agent Engine and Cloud Run Dockerfiles relied on that bind, so the default could not change on its own.

### Solution

- Add `-host` to the web launcher, defaulting to `127.0.0.1` (loopback only), and build the listen address with `net.JoinHostPort` so IPv6 hosts are bracketed (`[::1]:8080`). An empty `-host` is treated as the default rather than binding every interface.
- Print the configured host in the startup URL, normalizing loopback and all-interface hosts to `localhost` so the URL matches the ADK Web UI backend origin (`http://localhost:8080/api`) and does not fail CORS.
- Pass `-host 0.0.0.0` in the generated Agent Engine and Cloud Run Dockerfile CMDs so containers remain reachable.

### Behavior change

`adk web` now listens on `127.0.0.1` by default instead of all interfaces. Pass `-host 0.0.0.0` for the previous behavior; an empty `-host` selects the loopback default. Agent Engine and Cloud Run deployments keep listening on all interfaces because their Dockerfile CMD passes `-host 0.0.0.0`.

### Testing plan

- Unit tests cover the bind address (default, explicit, `0.0.0.0`, `::1`, empty), URL formatting (custom IPv6 and empty host), and the Dockerfile CMD for both deploy targets.
- `go build -mod=readonly work`, `go test -race ./cmd/launcher/... ./cmd/adkgo/internal/deploy/...`, `golangci-lint run` (v2.3.1) and `go mod tidy -diff` pass.
- Reverting each guard fails the matching test: the empty-host fallback, the `localhost` normalization, `net.JoinHostPort`, and each Dockerfile CMD.


jjsasha63 (changes requested):
Thanks for splitting this out — this is the change I was hoping for, and the
table-driven test asserting the default is exactly right.

One real bug before it can go in, plus two things worth folding in.

## IPv6 hosts produce an invalid address

`fmt.Sprintf("%s:%v", ...)` does not bracket the host, so an IPv6 literal
comes out unusable:

    host="::1"       -> Addr="::1:8080"
    host="localhost" -> Addr="localhost:8080"
    host="0.0.0.0"   -> Addr="0.0.0.0:8080"

Only the first is broken, and it fails at listen time rather than at parse
time:

    net.Listen("tcp", "::1:8099")   -> error: too many colons in address
    net.Listen("tcp", "[::1]:8099") -> ok

Use `net.JoinHostPort`, which brackets IPv6 and leaves the other two
unchanged:

```go
Addr: net.JoinHostPort(w.config.host, strconv.Itoa(w.config.port)),
```

Worth adding `::1` as a fourth case in `TestHostBinding`, expecting
`[::1]:`.

## The startup message still hardcodes localhost

`cmd/launcher/web/web.go:183` builds the URL as
`fmt.Sprintf("http://localhost:%v", ...)`, so binding to `0.0.0.0` or a
specific interface still prints `http://localhost:8080`. It was harmless when
the host was fixed; now that it is a flag, it will mislead. Building the URL
from `w.config.host` would keep the message honest.

## This is a breaking change for containers

Anyone running `adk web` inside a container with `-p 8080:8080` loses access,
because the process now binds loopback inside the container and the published
port has nothing to forward to. This is still the right default, but it
deserves a release note, and it would be worth naming the fix in the flag help
so people hit the answer where they hit the problem, for example:

    Use 0.0.0.0 to listen on all interfaces (required when running in a
    container).

Everything else looks good, and `./cmd/launcher/...` is green with the change.

AARUSHGAUR:
Thanks Sasha — I've addressed all three requested changes:
1) IPv6 host handling with net.JoinHostPort
2) startup URL now reflects the configured host
3) container guidance added to -host help
Also added the requested tests/documentation update. All checks are green.

Ready for re-review. Thanks!

jjsasha63 (changes requested):
The IPv6 fix is right, the `::1` case in `TestHostBinding` is right, and the flag
help and release note both cover the container problem. Thanks for turning that
around quickly.

Two blockers left. I built the branch and ran it to check both.

## The startup URL change breaks the Dev UI

This one is my fault — asking for an honest URL had a consequence I did not
think through. Running `adk web ... webui api` on the branch at the default
port prints:

```
Web servers starts on http://127.0.0.1:8080
```

Open that and the UI shows "Failed to load agents". The browser console says:

```
Access to XMLHttpRequest at 'http://localhost:8080/api/list-apps' from origin
'http://127.0.0.1:8080' has been blocked by CORS policy: the
'Access-Control-Allow-Origin' header has a value 'localhost:8080' that is not
equal to the supplied origin.
```

The chain: the webui sublauncher serves `runtime-config.json` with
`backendUrl: http://localhost:8080/api` (`cmd/launcher/web/webui/webui.go:121`),
and the api sublauncher answers with `Access-Control-Allow-Origin:
localhost:8080` (`cmd/launcher/web/api/api.go:147`). `localhost` and `127.0.0.1`
are different origins, so a call that used to be same-origin is now
cross-origin, and that allow-origin value has no scheme so it can never match.
Opening `http://localhost:8080/ui/` against the same server works — only the
printed URL changed.

Smallest fix that keeps the URL useful: render loopback and unspecified hosts as
`localhost` in `webURL()`, so `127.0.0.1`, `::1`, `0.0.0.0` and `::` all print
`http://localhost:PORT`. That also stops `-host 0.0.0.0` printing a URL that
Safari and Windows browsers cannot open. A `TestWebURL` case for each of those
four hosts would lock it in.

The deeper fix is to default `api_server_address` to empty so the UI uses
relative URLs — the bundled `distr/assets/config/runtime-config.json` already
has `backendUrl: ""` and the bundle does `backendUrl||""`. That would also fix
the unrelated bug where `--port 9000` leaves the UI calling port 8080. Please
leave it for a separate PR.

## The generated Dockerfiles need `-host 0.0.0.0`

Two first-party deploy paths write a Dockerfile that will now bind loopback
inside the container:

- `cmd/adkgo/internal/deploy/cloudrun/cloudrun.go:215`
- `cmd/adkgo/internal/deploy/agentengine/agentengine.go:265`

Both generate:

```
CMD ["/app/<exec>", "web", "-port", "8080", ...]
```

Cloud Run requires the container to listen on `0.0.0.0`, and the `proxy_port` in
that file is a local `gcloud run services proxy` on the developer's machine, not
a sidecar, so nothing inside the container reaches the server over loopback.
Both CMDs need `-host 0.0.0.0` added, in this PR.

## Two nits

1. `TestHostBinding` compares with `strings.HasPrefix`. The port is fixed at
   8080, so assert the whole `Addr`.
2. The branch is 4 commits behind main. No conflicts, but worth a rebase.

Everything else stands up: `go build ./...`, `go vet ./cmd/launcher/...` and
`go test -race -count=1 ./cmd/launcher/...` are all green on the branch.

karolpiotrowicz:
@kdroste-google

AARUSHGAUR:
Rebased onto latest main and resolved the conflicts. Ready for re-review when you get a chance, thanks!

jjsasha63 (approved):
No blocking findings.

## The printed URL is same-origin with the API

`displayHost` makes the startup URL `http://localhost:8080`, which matches the `Access-Control-Allow-Origin` value `normalizeOrigin` builds (`cmd/launcher/web/api/api.go:86`), so the Dev UI's XHR is same-origin and the agent list loads. `-host 0.0.0.0` also prints a URL that opens.

## Both generated CMDs carry -host 0.0.0.0

I generated the Cloud Run Dockerfile for every valid combination of the api, webui, a2a and agentengine flags, then `json.Unmarshal`ed each CMD line. All 48 parse and all carry the flag. The Agent Engine CMD does too.

## All bind addresses listen

Driving `Parse` -> `bindHost` -> `buildHTTPServer` and handing each result to `net.Listen`:

```
host=""          Addr="127.0.0.1:0"  url=http://localhost:0
host="127.0.0.1" Addr="127.0.0.1:0"  url=http://localhost:0
host="0.0.0.0"   Addr="0.0.0.0:0"    url=http://localhost:0
host="::1"       Addr="[::1]:0"      url=http://localhost:0
host="::"        Addr="[::]:0"       url=http://localhost:0
host="localhost" Addr="localhost:0"  url=http://localhost:0
```

Malformed values fail at listen time with a clear error instead of binding something wider than asked.

## Nothing else in the repo assumed the old bind

There are exactly two deploy paths under `cmd/adkgo/internal/deploy/` and both are covered here. No Kubernetes or Cloud Run manifests exist, so no probe hits a pod IP. No CI job, Makefile or script starts the server. The one committed Dockerfile, `scripts/adk-web/Dockerfile`, is a Node build stage with no CMD. Every doc that runs the launcher runs it locally with no port mapping.

Each guard has a test behind it: reverting the empty-host fallback, the `displayHost` normalization, `net.JoinHostPort`, or either Dockerfile CMD turns a test red.

`go build ./...`, `go vet` and `go test -race -count=1 ./cmd/launcher/web/... ./cmd/adkgo/internal/deploy/...` are green on 335248a2d.

## Parity vs src/google/adk/cli/ at adk-python 2.9.0

- Matches: `--host` with default `127.0.0.1`, registered next to `--port` (`cli_tools_click.py:1961`), and the deploy templates write `--host=0.0.0.0` (`cli_deploy.py:813`, `:1416`, `:1583`).
- Diverges: Python prints the raw host (`cli_tools_click.py:2195`) where Go normalizes loopback to `localhost`. Go splits the UI and API across sublaunchers with separate CORS, so printing the raw host breaks the UI.
- Not ported: Python pairs the loopback bind with a DNS-rebinding guard that rejects any request whose Host header is not loopback (`api_server.py:272`). Go has no counterpart — worth a follow-up issue.

AARUSHGAUR:
Thanks for the thorough review @jjsasha63 — the mutation checks and the parity notes are really helpful. I've opened #1622 for the DNS-rebinding guard you mentioned. The branch is 2 commits behind `main`; happy to update it if your merge tooling needs it, otherwise it should be good to go.

PR #1457, opened 2026-09-01
fix(adkrest,web): limit request body size to mitigate memory-exhaustion DoS

AARUSHGAUR:
### Problem

Request bodies were read without a size limit in the ADK REST API server and in the web launcher's base router. A remote client could POST an arbitrarily large body and cause the process to allocate memory proportional to it, up to an out-of-memory failure — a memory-exhaustion denial of service. The web launcher's base router also carries the PubSub and Eventarc trigger routes, so the gap was not limited to the REST API.

### Solution

- Add `adkrest.MaxBytesMiddleware` (wrapping `http.MaxBytesReader`) and `DefaultMaxPayloadSize` (10 MiB) in `server/adkrest`. A limit of `<= 0` selects the default inside the middleware, so callers pass their configured value through without resolving the default themselves.
- Apply the middleware in `adkrest.NewServer` before any route (including `/health`) is registered, and in the web launcher's new `buildRouter` on the base router, so every route mounted by a sublauncher — including the trigger endpoints — is covered.
- Add `-max_payload_size` to the web launcher and thread it through `launcher.Config.MaxPayloadSize` into `adkrest.ServerConfig`, so the API server enforces the same limit instead of keeping its own default.
- Report an oversized body as a client error everywhere: the API server and the PubSub-style triggers already answer 400, and the Eventarc binary-mode branch now does too instead of answering 500 for the same `*http.MaxBytesError`.

### Behavior change

The REST API server and every route behind the web launcher now reject request bodies larger than 10 MiB with HTTP 400 (`http: request body too large`), including the Eventarc binary-mode trigger. There was no limit before this change. Deployments that legitimately send larger bodies can raise the limit with `adk web -max_payload_size <bytes>` or by setting `launcher.Config.MaxPayloadSize`; `0` or a negative value selects the 10 MiB default.

`launcher.Config` only gains a field, so the public API remains backward compatible.

### Testing plan

- Unit tests cover the middleware limit and its boundary, the default fallback, the base-router limit on sublauncher/trigger routes, the threading from the web launcher to the API server, and the Eventarc binary-mode status code.
- `go build -mod=readonly work`, `go test -race ./cmd/launcher/... ./server/adkrest/...`, `golangci-lint run` (v2.3.1) and `go mod tidy -diff` pass.
- Reverting each guard fails the matching test: the 10 MiB default, the `<= 0` fallback, the Eventarc 400, the base-router middleware, the web→API threading, and the API→adkrest plumbing.


jjsasha63 (commented):
The test is right now, and I checked it properly. I removed
`router.Use(MaxBytesMiddleware(maxBytes))` from `NewServer` and it failed:

    oversized body status = 200, want 400

Valid JSON, real session service, asserts both the status and the exact
`"http: request body too large\n"` body. That is what I asked for. The
corrected doc comment on `MaxBytesMiddleware` is accurate too.

Not blocking the merge, but one gap is worth fixing here rather than later,
and it is my fault for how I suggested the split.

## The trigger endpoints are not covered

The middleware goes on `adkrest`'s own router (`handler.go:82`, from the
`mux.NewRouter()` at line 74). The PubSub and Eventarc triggers register on
the **parent** router instead — `cmd/launcher/web/triggers/eventarc/eventarc.go:132`
and `cmd/launcher/web/triggers/pubsub/pubsub.go:133` — so the middleware never
wraps them. The combined PR put the middleware on the parent router and did
cover them; asking you to split it dropped that, which I did not spot at the
time.

It matters most for exactly those routes, because `eventarc.go:102` does
`io.ReadAll(r.Body)` and buffers the whole body before doing anything else.
Sending 64 MiB, which is over six times the 10 MiB default, to the Eventarc
trigger wired the way the launcher wires it:

    sent 64 MiB to eventarc trigger -> status=500
    response: "failed to marshal agent message: json: error calling MarshalJSON..."

It reached `json.Marshal` at line 129, so the whole body was read in. The 500
is a JSON validity error, not a size rejection.

## The concurrency semaphore does not cover this

`MaxConcurrentRuns` looks like it would bound the damage, but it is acquired
at `eventarc.go:149`, after the read at line 102. Memory is already committed
by then, and an attacker never needs to acquire it.

## Suggested fix

Apply the middleware on the base router in `cmd/launcher/web/web.go`, where
`BuildBaseRouter` is called, so every mounted route inherits it regardless of
which sublaunchers are active. Keeping it in `NewServer` as well is fine —
nested `MaxBytesReader` wrappers are harmless, and it keeps the guarantee for
people embedding `adkrest.NewServer` directly.

If you would rather land this PR as-is and do the trigger coverage as a
follow-up, that also works. In that case, please narrow the title and
description to the REST surface, since "prevent memory-exhaustion DoS"
currently claims more than the change delivers.

## Two smaller notes

`10 MiB` may be tight. `/run` and `/run_sse` take `genai.Content` in
`newMessage` (`server/adkrest/internal/models/runtime.go:30`), which carries
base64 `InlineData` for images, audio and PDFs. Base64 costs a third, so the
default caps inline input at roughly 7.5 MiB of binary — below what the model
itself accepts inline. Worth considering a larger default, or a line in the
`MaxPayloadSize` doc comment pointing at it as the knob to raise.

Also worth being precise in the doc comment about what the limit does: it
bounds one request, not the server. There is no connection or concurrency
limit on the HTTP server, so total exposure is still concurrent requests times
the per-request cap. That is normal and not something this PR needs to solve,
but "mitigates" is a fairer word than "prevents".

jjsasha63 (commented):
The approach is right and both tests are real. I mutation-checked them in a clean worktree: removing `router.Use(MaxBytesMiddleware(maxBytes))` from `NewServer` fails `TestServerRejectsOversizedBody` (200 vs 400), and removing the base-router middleware fails `TestBuildRouterEnforcesBodyLimitOnSublauncherTriggerRoutes` (204 vs 400). `go test -race -count=1 -shuffle=on ./server/adkrest/... ./cmd/launcher/web/...` is green.

Two things to fix before this goes in.

## 1. -max_payload_size can only lower the limit, never raise it

`cmd/launcher/web/api/api.go:79` builds `adkrest.NewServer` without `MaxPayloadSize`, so adkrest applies its own 10 MiB default inside the base router's limit. The two `MaxBytesReader`s count independently, so the effective limit is the smaller of the two.

I mounted the real api sublauncher on a base router with the middleware at 50 MiB and posted a 20 MiB body:

    POST /api/apps/myapp/users/u1/sessions, base router limit 50 MiB, body 20 MiB
    -> 400 "http: request body too large"

An operator who raises the flag gets no effect above 10 MiB. That matters for `/run` and `/run_sse`, which carry base64 inline blobs: an attachment of about 7.5 MB exceeds 10 MiB once encoded, and no flag fixes it.

Fix: plumb the launcher's value into `adkrest.ServerConfig.MaxPayloadSize` in api.go so the two limits agree.

## 2. The trigger test asserts a status the real trigger handlers do not return

`fakeTriggerSublauncher` decodes with `json.Decoder` and calls `http.Error`, so it returns a plain-text 400. I ran the real sublaunchers through the same base router with the middleware at 1 KiB:

    pubsub   -> 400 {"status":"failed to decode request: http: request body too large"}
    eventarc -> 500 {"status":"failed to read body: http: request body too large"}

Eventarc takes the binary-mode branch unless `Content-Type` is `application/cloudevents+json`, and that branch is the `io.ReadAll` at `server/adkrest/controllers/triggers/eventarc.go:102`, whose error path is a 500. The body is still capped, so the mitigation holds. But the test name and the PR description both say the trigger endpoints answer 400, and that is not true for eventarc.

Either of these works:

- Rename the test to what it proves, that the middleware is mounted on the base router and reaches sublauncher routes, and drop the 400 claim about the trigger endpoints from the description.
- Add a test against the real pubsub sublauncher in `cmd/launcher/web/triggers/pubsub`, external test package `pubsub_test`. It imports `cmd/launcher/web` with no import cycle, and I ran it.

Mapping `*http.MaxBytesError` to 413 in the eventarc read path would make a good separate PR.

## Not blocking: /run_live is not covered

`server/adkrest/controllers/runtime.go:401` reads websocket frames with `ws.ReadMessage()` and never calls `SetReadLimit`. `MaxBytesReader` only wraps `r.Body`, which is gone once the connection is hijacked, so frame size stays unbounded. That is the largest memory-exhaustion surface left. Worth a follow-up issue, and worth not implying in the description that the whole server is covered.

## Nits

- `middleware.go` and `handler_security_test.go` say Copyright 2025. New files in this repo use 2026.
- `router.Use` sits after the `/health` registration in `NewServer`. Gorilla applies middleware at match time so it works, but it reads as though health is exempt. Move it above the route.
- `MaxBytesMiddleware` and `DefaultMaxPayloadSize` are new exported API on a released v2 module. Deliberate, or convenience for the launcher? If the launcher is the only caller outside the package, an internal package keeps the surface smaller.
- Nothing asserts that `DefaultMaxPayloadSize` applies when `MaxPayloadSize` is 0 or negative.

For item 1, the test to add is the raise path: set the launcher limit above 10 MiB and assert a body between the two limits succeeds.

AARUSHGAUR:
Rebased onto latest main, conflicts resolved. Could you take a look when you have time? Thanks!

AARUSHGAUR:
Gentle nudge on this one when you have a moment @jjsasha63 — it has been rebased on `main` since Sep 19 and the tests are the same mutation-verified ones from your earlier review. Happy to adjust anything you'd like.

jjsasha63 (changes requested):
`go test -race -count=1 ./...` is green across 87 packages. The new tests are load-bearing too: I mutated each production line they cover and every one went red. That covers the adkrest `router.Use`, the `<= 0` default fallback, the api.go threading and the eventarc status.

Both of the problems below sit in the launcher plumbing, not in the middleware itself.

## 1. `-max_payload_size` collides with the agentengine flag of the same name

`cmd/launcher/web/agentengine/agentengine.go:52` already registers `max_payload_size`. It means something else there: it wraps the body in `io.LimitReader` and truncates, where the new one rejects with 400. `cmd/launcher/agentengine/agentengine.go:27` mounts that sublauncher on the web base router, so both flags live in one binary, in separate flag sets parsed positionally. Nothing rejects the collision, and `CommandLineSyntax` prints the name twice with different defaults and different help text.

Left at their defaults the two agree. Both are 10 MiB, and `io.LimitReader` never reads the byte that would trip the outer reader. The break appears once an operator raises the agentengine flag, which is the only reason to set it. I ran the real launcher and posted a 12 MiB body to `/api/reasoning_engine`.

```
args: web -port=N agentengine -max_payload_size=52428800

on main:         HTTP 500 "unrecognized class method: x"
with this patch: HTTP 400 "io.ReadAll with LimitReader failed: http: request body too large"
```

The 500 is the dispatcher rejecting a bogus `class_method`, so on main the full 12 MiB was read and parsed. Here the base router refuses it at the 10 MiB default, and the flag the operator already set cannot raise it. The workaround is to pass the identically-named flag before the `agentengine` keyword as well. Nobody will find that in help output showing one name twice. Renaming the new flag, to `-max_request_body_size` or similar, avoids all of it.

## 2. `buildRouter` overwrites a caller-set `launcher.Config.MaxPayloadSize`

`cmd/launcher/web/web.go:258` assigns unconditionally, and the flag defaults to 0. An embedder who sets the field in code has it replaced by zero before any sublauncher sees it, and their own struct mutated. The only non-test reader, `api.go:303`, runs fourteen lines later inside `SetupSubrouters`.

```
embedder set Config.MaxPayloadSize = 1048576
sublauncher saw 0
caller's Config.MaxPayloadSize after buildRouter = 0
```

The description offers that field as one of the two ways to raise the limit, so the code and the description disagree. `applyServiceDefaults` is not precedent for the write: it fills only what the caller left nil, and re-running it is a no-op.

Guarding the assignment is not quite enough on its own. I tried `if w.config.maxPayloadSize > 0 { config.MaxPayloadSize = ... }` and nothing else. Propagation starts working, but the base router still rejects at 10 MiB. Line 253 reads the flag rather than the resolved value.

```
embedder asked 20 MiB; sublauncher saw 20971520
12 MiB body through base router -> 400 "http: request body too large"
```

Resolving once and using that for both makes it pass. It also keeps every existing test in `cmd/launcher/...` and `server/adkrest/...` green:

```go
if w.config.maxPayloadSize > 0 {
	config.MaxPayloadSize = w.config.maxPayloadSize
}
router.Use(adkrest.MaxBytesMiddleware(config.MaxPayloadSize))
```

Worth a test row too. Both rows of `TestBuildRouterPropagatesMaxPayloadSizeToSublaunchers` pass `&launcher.Config{}`, so the caller's value is already zero and a conditional assignment would pass either way.

## Not blocking: 10 MiB sits under the largest Pub/Sub message

Pub/Sub's documented ceiling is 10 MB in the message `data` field. A wrapped push delivers that as base64 inside JSON. The push body is 13,333,456 bytes against a 10,485,760-byte default.

```
no limit:      200 {"status":"success"}
10 MiB limit:  400 {"status":"failed to decode request: http: request body too large"}
```

Pub/Sub treats a non-2xx push as a nack and redelivers until retention expires, so each oversize message retries for a week. Raising the default to clear that ceiling and leaving it to the flag are both defensible. The trigger routes are covered here for the first time, so the number is worth choosing deliberately.

## Parity

adk-python has no request-body limit to match. There is no `MaxBytesReader` equivalent in `cli/api_server.py` or `cli/fast_api.py`. The only middleware registered there are `CORSMiddleware`, `_OriginCheckMiddleware` and `_DefaultAppRewriteMiddleware` (`api_server.py:1223-1244`). This is Go-only hardening rather than a port, so nothing upstream pins the default or the field name. One line in the description saying so would save the next reader a search.

## Checked and clean

The middleware boundary is exact: a body at the limit is accepted, one byte over is rejected, and 0 or negative selects the default. `router.Use` sitting after the `/health` registration is fine, because gorilla builds the chain at match time. Subrouter, `PathPrefix().Handler()` and `StripPrefix` mounts all inherit it. Chunked bodies and a lying `Content-Length` are both enforced, so the limit is on bytes read rather than the declared header. No handler writes a status before reading the body, so the `MaxBytesReader` write is safe on `/run_sse` and on the a2a JSON-RPC path. There is no form or multipart parsing anywhere. The `/run_live` upgrade still completes through the middleware. Frames after the hijack stay unbounded, which an HTTP body limit cannot reach and which this change does not regress.

AARUSHGAUR:
Thanks @jjsasha63 — both fixed and rebased onto `main` (the `api_test.go` conflict was resolved by keeping both test functions).

1. **Flag collision**: renamed the web flag to `-max_request_body_size`, so it no longer collides with `agentengine -max_payload_size` (which truncates rather than rejects). The help text also mentions that a value set by an embedder is honored.
2. **Config overwrite**: `buildRouter` now only writes `config.MaxPayloadSize` when the flag was actually provided, and passes the resolved value to `adkrest.MaxBytesMiddleware`, so an embedder that sets `Config.MaxPayloadSize` in code is respected by both the base router and the sublaunchers. Added test rows for "caller-set value is preserved" and "flag wins over caller value" to `TestBuildRouterPropagatesMaxPayloadSizeToSublaunchers`.

On the non-blocking Pub/Sub note: I kept the default at 10 MiB so the base router and adkrest agree out of the box, and documented that operators whose push bodies can exceed it can raise it with `-max_request_body_size`. Happy to change the default to clear the Pub/Sub ceiling if you'd prefer that.

`go build ./...`, `go vet ./cmd/launcher/... ./server/adkrest/...` and `go test -race -count=1 ./cmd/launcher/... ./server/adkrest/...` are green on 048b416.

AARUSHGAUR:
Rebased onto `main` again now that #1456 has landed — the `web.go` conflict was resolved by keeping both sets of functions (`webURL`/`bindHost`/`displayHost` from #1456 and `buildRouter` from this PR). `go build ./...`, `go vet` and `go test -race -count=1 ./cmd/launcher/... ./server/adkrest/...` are green on d54b0e0. Ready for re-review when you have a moment.

AARUSHGAUR:
Rebased once more — main moved again (02564db) and the new launcher changes conflicted:

- `web.go`: kept both sides — the resolved `BindHost` assignment now lives in `buildRouter` alongside the payload-size plumbing, so the loopback Host check still arms.
- `api_test.go`: kept both tests (bind-host guard test + the raised-limit test).

`go build work`, `go test -race` for `cmd/launcher/...` and `server/adkrest/...`, and `go mod tidy -diff` are all clean.

jjsasha63 (changes requested):
`go build ./...` succeeds and `go test -race -count=1 ./cmd/launcher/... ./server/adkrest/...` is green across 21 packages.

## Blocking: the godoc points operators at a flag that does not exist

`cmd/launcher/launcher.go:117` documents `launcher.Config.MaxPayloadSize` as set "from its `-max_payload_size` flag". The flag registered at `cmd/launcher/web/web.go:371` is `-max_request_body_size`. `-max_payload_size` does exist elsewhere in the same binary: `cmd/launcher/web/agentengine/agentengine.go:52` registers it, and it truncates with `io.LimitReader` instead of rejecting.

Driving the real launcher through `universal.NewLauncher(web.NewLauncher(api.NewLauncher(), agentengine.NewLauncher("agentengine")))`:

```
web -port=0 -max_payload_size=52428800 api
  -> failed to parse web flags: flag provided but not defined: -max_payload_size

web -port=0 -max_request_body_size=52428800 api
  -> <nil>

web -port=0 api agentengine -max_payload_size=52428800
  -> <nil>, sets agentengine's truncation limit; the body limit stays at 10 MiB
```

The operator this hurts is the one already running `agentengine -max_payload_size=52428800`. Their 12 MiB POST to `/api/reasoning_engine` gets a 400 from the base router, they read the godoc for the field that raises the limit, and the flag it names either refuses to start the server or silently configures something else with no error. The godoc sits on an exported field, so it ships to pkg.go.dev.

Three spots carry the wrong name: `cmd/launcher/launcher.go:117`, the PR description twice (including the `adk web -max_payload_size <bytes>` instruction), and a comment at `cmd/launcher/web/router_body_limit_test.go:126`.

## Not blocking: the base-router limit is unpinned

`cmd/launcher/web/web.go:307` feeds the base router from the resolved `config.MaxPayloadSize` rather than the raw flag, which is what lets an embedder-set value reach it. Changing that one line to `w.config.maxPayloadSize` leaves the whole suite green:

```
router.Use(adkrest.MaxBytesMiddleware(w.config.maxPayloadSize))
go test -count=1 ./cmd/launcher/... ./server/adkrest/...  -> ok, 13 packages
```

`TestBuildRouterPropagatesMaxPayloadSizeToSublaunchers` asserts the value sublaunchers receive, not the one the base router installs. A row that sets only `launcher.Config.MaxPayloadSize` and sends an oversize body through the router `buildRouter` returns would close it.

## Parity

adk-python has no request-body limit to match. `cli/api_server.py:1292-1316` registers three middlewares, `CORSMiddleware`, `_OriginCheckMiddleware` and `_DefaultAppRewriteMiddleware`, and there is no `max_size`, no 413 and no `content_length` check anywhere under `cli/` (adk-python 2.9.2). This is Go-only hardening, so nothing upstream pins the default or the field name.

## Checked and clean

The four flag-and-embedder combinations resolve to one limit on both the base router and the API server: neither set gives exactly 10 MiB, accepted at the limit and rejected one byte over; embedder-only gives N on both; flag-only and flag-plus-embedder give M on both. The two limits can no longer disagree. `max_request_body_size` appears once across every flag registration under `cmd/`, so nothing collides with it. Nested `MaxBytesMiddleware` at the same limit adds no second boundary. The `<= 0 selects the default` sentinel matches `ServerConfig.DebugConfig.TraceCapacity` at `server/adkrest/handler.go:273`, and the middleware signature matches `server/authn/authn.go:87`. In-place mutation of the caller's `*launcher.Config` has precedent at `cmd/launcher/web/web.go:182` and `cmd/launcher/web/api/api.go:301`. Unmatched routes read nothing: a 40 MiB POST to a 404 or 405 costs the server under 0.2 MiB. `MaxHeaderBytes` is unset, but net/http's 1 MiB default covers the request line, so a 2 MiB header gets 431.

AARUSHGAUR:
Thanks @jjsasha63 — all three spots fixed and rebased onto `d64e113`:

1. **Blocking (godoc)**: `cmd/launcher/launcher.go` now documents `-max_request_body_size` and notes that an embedder-set value is honored. The test doc comment at `router_body_limit_test.go` and both PR-description mentions are fixed too (`adk web -max_request_body_size <bytes>`).
2. **Base-router limit now pinned**: added `TestBuildRouterBaseRouterHonorsEmbedderPayloadLimit`, which sets only `launcher.Config.MaxPayloadSize` (flag left at 0), builds the router via `buildRouter`, and sends a body over the embedder limit but under the 10 MiB default. Mutation-checked the other direction: changing `router.Use(adkrest.MaxBytesMiddleware(config.MaxPayloadSize))` to `w.config.maxPayloadSize` now fails it with `got status 204, want 400`.

`go build ./...` and `go test -race -count=1 ./cmd/launcher/... ./server/adkrest/...` are green on `17eb24b`.

AARUSHGAUR:
Rebased onto `a728920` (no conflicts). `go build ./...` and `go test -race -count=1 ./cmd/launcher/... ./server/adkrest/...` are green.

AARUSHGAUR:
Rebased onto `f9ce16e` (no conflicts). `go build ./...` and `go test -race -count=1 ./cmd/launcher/... ./server/adkrest/...` all pass. Ready for another look when you have a moment @jjsasha63.

AARUSHGAUR:
Rebased onto bb4a72a (no conflicts). `go build ./...` and `go test -race -count=1 ./cmd/launcher/... ./server/adkrest/...` are green. Ready for another look when you have a moment @jjsasha63.

PR #1459, opened 2026-09-01
fix(adkrest): serve the endpoints the ADK web UI actually calls

jjsasha63:

None open for this.

**Problem and solution:**

```
fix(adkrest): serve the endpoints the ADK web UI actually calls

The vendored Angular bundle calls 42 endpoints; 32 failed. ADK v2 moved 30
developer endpoints under /dev/apps/{app_name}/ and the routers never
followed, so the agent structure panel, traces, the builder and the version
header all 404ed. Chat kept working because /run_sse, /list-apps and the
session routes are not part of the moved set, which is why it went
unnoticed for two months.

Register the dev paths alongside the pre-v2 ones so existing clients keep
working. Implement /version and the two agent-graph endpoints. Answer
features ADK Go does not have with 501 and a readable body rather than a
bare 404 a client cannot tell from a wrong path prefix.

Also: HEAD returned 405 on every route; a 405 carried no Allow header and
most mismatches degraded to a bare 404; OPTIONS ran the DELETE handler; a
trailing slash turned a POST into a GET against another handler; a missing
session or artifact returned 500; ListSessions returned null; PATCH on a
session was never implemented, and now rejects state keys scoped outside
the session; and the Traces panel was blank for three separate reasons in
the span payload.

Adds a contract test that reads the shipped bundle, extracts every API call
site with its verb, and asserts each one routes, so a refresh that changes
a path or a method fails here rather than in a browser.
```

### Testing Plan

### Checklist

---

**Reviewing this stack.** These are cross-fork PRs, so each one must target `main`
and GitHub shows a cumulative diff. Review only the commit named in the title:

- #1458 — `123a7711` session (this PR's first commit; already up for review there)
- **this PR — `a5ba1d3d` adkrest**
- #1460 — `7c7e7994` launcher

Land in order. Once #1458 merges, this diff shrinks to its own commit.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -9,11 +9,11 @@ on:
   push:
     branches: [ "main", "v1" ]
 
+  # Deliberately unfiltered by base branch. A stacked PR targets the branch
+  # below it rather than main, so any allow-list here silently skips build,
+  # test and lint for every PR in a stack. A red linter survived a whole
+  # review round that way.
   pull_request:
-    # Stacked PRs target the branch below them rather than main, so limiting
-    # this to main and v1 meant no build, test or lint job ever ran against a
-    # stack. A red linter survived a whole review round that way.
-    branches: [ "main", "v1", "baptmont/**" ]
 
   workflow_dispatch:
 
--- a/cmd/launcher/web/api/api.go
+++ b/cmd/launcher/web/api/api.go
@@ -19,6 +19,7 @@ import (
 	"flag"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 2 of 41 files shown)


baptmont (changes requested):
Reviewed the incremental change only (`git diff p1458 p1459`, 29 files) — the session layer is #1458 and already approved.

The routing rewrite is good work and I verified it does what it claims: `POST …/sessions/` now creates a session, `DELETE`/`PATCH /version` return 405 with a correct `Allow`, and `OPTIONS /run` no longer invokes `RunHandler`. I also mutation-tested `uicontract_test.go` — both halves fail on real drift, and the second half genuinely parses the committed bundle rather than restating the implementation. That is a better test than most contract tests in this position.

Two problems, both inside the headline feature: session rename does not work after this PR.

## 1. The UI's own rename payload is rejected by the new guard

`sessions.go:190` rejects a `stateDelta` containing `app:` or `user:` keys, and the comment above it says *"The web UI only ever sends session-scoped keys."* That premise does not hold.

The committed bundle sends back the **entire state map it read**:

```js
saveSessionName(A){ … let i = A.state || {},
  n = Ye(Y({}, i), {__session_metadata__: Ye(Y({}, i.__session_metadata__||{}), {displayName: e})});
  this.sessionService.updateSession(this.userId, this.appName, A.id, {stateDelta: n})
    .subscribe({error: () => {}}) }
```

And the read side hands it scoped keys to echo back. Against `session.InMemoryService()`:

```
create state {plain, app:config, user:locale}
Get(...) -> keys visible to the client: [app:config user:locale plain]
```

So for any agent that uses app- or user-scoped state, renaming a session PATCHes those keys straight back, the guard answers `400 stateDelta may only contain session-scoped keys; remove [app:config user:locale]`, and `error: () => {}` swallows it — the name silently does not change and nothing appears in the console.

The guard itself is right and worth keeping; a client must not be able to write app-wide state through its own session. What is wrong is the assumption that the delta is authored rather than echoed. Filtering the scoped keys out and applying the rest, or diffing the submitted state against the stored one and applying only what changed, both preserve the security property without rejecting the UI's payload. Rejecting is only safe if the client is known to send a true delta, and this one does not.

## 2. `uicontract_test.go` cannot see that the route is unreachable through `adkgo web`

The new `PATCH` route works against `adkrest.NewServer`, which is what the contract test drives. It does not work through the launcher: `cmd/launcher/web/api/api.go:107` mounts the API behind

```go
router.Methods("GET", "POST", "DELETE", "OPTIONS").Handler(corsHandler)
```

so `PATCH /api/apps/…/sessions/{id}` is 405 before it reaches any of this. The same applies to the two new `PUT` routes and to every `HEAD` that `withHead()` was added to serve.

**#1460 fixes this** — it replaces that list with an `apiMethods` slice including `PUT`, `PATCH` and `HEAD`. So the stack as a whole is fine, but the stated landing order (session → adkrest → launcher) puts `main` in a state where this PR's headline feature is merged and unreachable, and its own contract test reports green throughout. Either move the method list into this PR, or land 1459 and 1460 together.

## Highs

- **`writeSessionServiceError` turns the ownership mismatch into a 500 and creates an existence oracle.** This is where the note I left on #1458 becomes observable: a missing session is now 404 while a session that exists but belongs to another user is 500, so an authenticated caller can distinguish the two. Low severity given session IDs are UUIDs, but 500 is also just the wrong code for a client condition — "you asked for someone else's session" is not a server fault. Mapping the ownership case to 404 fixes both at once.
- **`build_graph` and `build_graph_image` are unconditional**, including in the production launcher, where the neighbouring debug endpoints are deliberately flag-gated and adk-python keeps the graph endpoints on its `DevServer`. Worth gating the same way unless there is a reason not to.
- **No `MaxBytesReader` on the new PATCH write path**, so a 32 MiB body decodes to roughly 64 MiB of live objects. The repo has no body limits anywhere, so this is consistent rather than novel — but this is a new write endpoint and the cheapest moment to add one.

## Verified holds

Build, `-race -shuffle=on` (one unrelated a2a flake), `golangci-lint` 0 issues in both modules, `go mod tidy -diff` empty. No panics or hangs from `HEAD` on any `GET` route, no path traversal in artifact handling, the nil-service guards work, and the public API changes are additive only.

jjsasha63:
Both blockers are fixed.

## 1. The echoed rename payload

You were right that the delta is echoed rather than authored. I confirmed it
directly: `Get` on a session created with `{plain, app:config, user:locale}`
hands all three keys back to the client.

`dropEchoedScopedKeys` now discards a scoped key whose submitted value already
matches the stored one. Such a key writes nothing, so dropping it leaves the
session exactly as accepting it would. A scoped key carrying a different value
is still a 400, and the error names only the changed key.

Values compare as JSON, not with `reflect.DeepEqual`. An agent that stored
`int64(7)` gets `float64(7)` echoed back, and those are unequal in Go — that
would have broken rename again for anyone with a number in scoped state.

Four mutations, each killed:

| Mutation | Caught by |
| --- | --- |
| Remove the filter (the bug you found) | Both new tests, 400 instead of 200 |
| Drop all scoped keys unconditionally | `TestUpdateSessionRejectsScopedStateKeys`, 4 subtests |
| `reflect.DeepEqual` instead of JSON | The int round-trip subtest |
| Treat an absent key as echoed | 5 subtests |

## 2. The route is now reachable from this PR

I moved `cmd/launcher/web/api/api.go` and `api_test.go` here from #1460.

I tried reordering the stack first, since that needs no file moves. It does not
work: the launcher's `TestSetupSubroutersServesRealRESTAPI` depends on this
PR's `withHead()` routes, so 6 subtests fail if the launcher lands first. The
dependency runs both ways.

Landing the two together also works, but it depends on a maintainer merging two
cross-fork PRs back to back. Moving the mount makes this PR correct on its own.
I moved the whole file rather than only the method list, because `apiMethods`
and `corsAllowMethods` are built from the same slice.

I checked it by running a rename through the mount the way the bundle does:

| Stage | PATCH rename |
| --- | --- |
| This PR without the mount | 405 at the mount |
| This PR as it stands | 200, and the name persists |

## The three highs

Ownership mismatch returning 500: agreed, I will map it to 404.

`MaxBytesReader` on the PATCH path: agreed, I will add one.

Gating `build_graph`: I would rather not. `include_debug_api` defaults to
false, so gating these turns the agent structure panel off for everyone running
`adkgo web` — and that panel is one of the things this PR fixes. Say the word
if you still want it to match adk-python's DevServer and I will gate it.

To verify: `go test -race -count=1 -shuffle=on ./server/adkrest/... ./cmd/launcher/...`

baptmont (approved):
Both blockers cleared. Verified against `639dd99` rather than taking the summary on trust.

**The echoed payload.** Drove the real handler through `adkrest.NewServer` with a session created carrying `{plain, app:config, user:locale}`, read it back the way the UI does, and PATCHed the whole map:

```
echoed scoped keys (the reported bug)                -> 200
int round-trips as float (your catch)                -> 200
genuinely CHANGED scoped value must still be refused -> 400
scoped key not present in the session must be refused-> 400
```

The two negatives matter as much as the positives: the property that made the guard worth having — a client cannot write app-wide state through its own session — still holds. Filtering on *value equality* rather than dropping scoped keys wholesale is a better answer than the one I suggested, because it keeps the refusal for a real write.

The `int64`/`float64` point is a good catch and one I would have missed. Comparing as JSON rather than with `reflect.DeepEqual` is the right call, and the round-trip subtest is the one I would keep if the suite ever had to shrink.

**Reachability.** `apiMethods` now lives in this PR. Exercising `registerAPIRoutes` directly:

```
GET HEAD POST PUT PATCH DELETE OPTIONS  -> all reach the API handler
```

Moving the file rather than reordering is the right resolution, and the reason you gave settles it — I had not checked that the launcher's tests depend on this PR's `withHead()` routes, so the dependency really does run both ways and reordering just moves the breakage.

Gates on this head: build clean, `go test -race -count=1 -shuffle=on ./server/adkrest/... ./cmd/launcher/...` green.

Clearing my request for changes.

## Three of the six lows from the #1460 thread belong here now

You asked for the ones I did not name. `api.go` moved into this PR, so these three came with it. None is blocking.

- **`Vary: Origin` on a response that does not vary by origin** (`api.go:115`). `Access-Control-Allow-Origin` is derived from the `-webui_address` flag and never reads the request's `Origin`. With `addr="localhost:8080"` and a request carrying `Origin: https://attacker.example`, the response is still `ACAO="http://localhost:8080"`, `Vary="Origin"`. Harmless, but it fragments caches for nothing, and the comment justifying it describes a situation that cannot arise. Either drop the `Vary`, or echo an allow-listed request origin — at which point it becomes correct.
- **`normalizeOrigin` emits syntactically invalid origins** (`api.go:84-99`). Junk passes through with `http://` prepended: `addr="not a url"` yields `ACAO="http://not a url"`, `addr="javascript:alert(1)"` yields `ACAO="http://javascript:alert(1)"`. No header-injection risk — Go's header writer replaces CR/LF, and the value is operator-supplied — and echoing the operator's own input is a defensible choice. A startup warning when it does not parse as an origin would turn a mystifying browser CORS failure into one log line.
- **`-webui_address=*` is now advertised in the flag help** (`api.go:334`). The REST API has no authentication, so `*` lets any website in the user's browser drive the local agent server and read the responses. The behaviour is unchanged — `*` already passed through — but the help text now invites it. Worth a clause saying what it costs.

PR #1460, opened 2026-09-01
fix(launcher/web): default the artifact and memory services, and serve the UI correctly

jjsasha63:

None open for this.

**Problem and solution:**

```
fix(launcher/web): default the artifact and memory services, and serve the UI correctly

The web launcher defaulted only SessionService, so the artifact routes
panicked on a nil service and dropped the connection with no HTTP response.
A nil memory service would panic mid-run for any agent using load_memory.

Also in the launcher: PUT and PATCH were rejected at the API mount for
every path; Access-Control-Allow-Origin was emitted without a scheme, which
browsers reject; Angular deep links 404ed, and serving index.html for them
needs the base href rewritten or every asset resolves one directory too
deep and the page is blank; /ui without a trailing slash 404ed;
/ui/assets/config/ served a directory listing; and health is now served at
the root as well as under the API prefix, so a container probe with a fixed
path does not need to know which sublaunchers are enabled.

Pins the adk-web revision and records the resolved SHA, so a bundle refresh
is reviewable. The pin does not describe the committed bundle, which
predates it; the Dockerfile and CONTRIBUTING say so.
```

### Testing Plan

### Checklist

---

**Reviewing this stack.** These are cross-fork PRs, so each one must target `main`
and GitHub shows a cumulative diff. Review only the commit named in the title:

- #1458 — `123a7711` session
- #1459 — `a5ba1d3d` adkrest
- **this PR — `7c7e7994` launcher**

Land in order. Once #1458 and #1459 merge, this diff shrinks to its own commit.

---

**Reviewing this stack.** These are cross-fork PRs, so each one must target `main`
and GitHub shows a cumulative diff. Review only the commit named in the title:

- #1458 — `123a7711` session
- #1459 — `a5ba1d3d` adkrest
- **this PR — `7c7e7994` launcher**

Land in order. Once #1458 and #1459 merge, this diff shrinks to its own commit.


baptmont (commented):
Reviewed the incremental change only (`git diff p1459 p1460`, 10 files) — the session and adkrest layers are #1458 and #1459.

**No blockers in this part**, and it is unusually well covered: build, `go test -race -count=1 -shuffle=on ./server/adkrest/... ./cmd/...`, `golangci-lint` (0 issues) and `go mod tidy -diff` are all clean in both modules.

Not approving only because this branch contains #1459, which I have just asked for a change on — the scoped-key guard in `sessions.go` rejects the payload the bundled UI actually sends for a rename, so session rename is broken here too. Once that is resolved upstream in the stack, this part has nothing from me.

Also worth recording: this PR **fixes** the routing half of that problem. Replacing the hardcoded `GET, POST, DELETE, OPTIONS` list with the `apiMethods` slice is what makes #1459's new `PATCH`/`PUT`/`HEAD` routes reachable through `adkgo web` at all. Building `corsAllowMethods` from the same slice so the two cannot drift is the right shape.

## Path traversal: none, and I looked hard

Probed the UI handler over a raw TCP socket to bypass `net/url`, and directly against `spaHandler` to bypass the mux's `cleanPath`, with `../`, `..%2f`, `%2e%2e%2f`, double-encoding, `....//`, backslashes, `%00` and `..;/`. Nothing escaped. The defence is three deep — path cleaning, then `fs.ValidPath`, then `embed.FS` — and directory listings are suppressed. Trailing-slash normalisation is encoding-correct, the `<base href>` escaping resists a hostile prefix, redirect rewriting is not an open-redirect vector, and the SSE `Flusher`/`ResponseController` plumbing survives the wrapper.

## Highs

- **Defaulting the artifact service silences a real diagnostic.** With a nil artifact service the API previously answered `503 "artifact service is not configured"` (`artifacts.go:43`). It now silently writes to process memory, with nothing logged, and `cmd/launcher/prod` shares the same `Run` path — so a production deployment that forgot to configure artifacts gets a working-looking server that loses everything on restart, where before it said so. The doc comment on `applyServiceDefaults` (`web.go:158`) explains the change as "a nil service … panics instead of returning an HTTP error", which is true for sessions and not true for artifacts. Log the defaulting, and correct the comment.
- **`BuildBaseRouter` is exported and now hard-registers `/health`** (`web.go:309`), so an embedder that registers its own `/health` afterwards is silently shadowed — mux takes the first match. The rationale in the comment is sound for the launcher's own use; the problem is doing it in an exported constructor. Registering it in `Run` rather than in `BuildBaseRouter` keeps the behaviour without taking the route away from callers. There is also a `Content-Type` mismatch with `/api/health` worth tidying while there.
- **The Dockerfile pins the source but not the build.** `ADK_WEB_REF` pins what is checked out, but `node:iron-trixie` floats and the build uses `npm install` rather than `npm ci`, so two builds of the same ref can produce different bundles — which undercuts the reproducibility the README claims. Upstream has a `package-lock.json` at the pin, so `npm ci` is a drop-in. Both lines are pre-existing rather than introduced here, but this PR is what documents the process as reproducible.

## Lows

Six minor items, the one worth naming being a doc/implementation mismatch: `CONTRIBUTING.md:200` says the refresh script reports a compare URL, but that line is conditional on a provenance file that does not exist yet — so it will not print on the first refresh, which is exactly the case the surrounding text is describing.

jjsasha63:
All four items are fixed, and the API mount moved out to #1459.

Artifact defaulting is now logged, one line per service. You were right about
the comment: it claimed a nil service panics, which holds for session and
memory but not artifact, which answers 503. Corrected.

`/health` moved into `registerHealthRoute`, called from `Run` rather than
`BuildBaseRouter`, so an embedder's own handler is no longer shadowed. The
`Content-Type` dropped its `charset=UTF-8` and now matches `/api/health`.

The container installs with `npm ci`. `package-lock.json` is present at the
pin, so it was the drop-in you described. I did not pin the `node:iron-trixie`
digest — that needs bumping by hand forever, and I cannot build-test it here.
I recorded the float in `scripts/adk-web/README.md` instead, so the docs no
longer imply more than the pin gives.

CONTRIBUTING now says the first refresh prints no compare URL.

Three mutations confirm the new assertions fail against the unfixed code:
deleting the log line, re-registering `/health` in `BuildBaseRouter`, and
restoring the charset parameter.

`cmd/launcher/web/api/` is no longer in this PR. #1459 needs it to be reachable
through `adkgo web`, so it belongs there; the reasoning is in that thread.

One ask: you mention six minor items but name only the CONTRIBUTING one. The
other five are not in the review body and there are no inline comments — could
you post them?

To verify: `go test -race -count=1 -shuffle=on ./cmd/launcher/...`

baptmont (dismissed):
All four verified against `c5ba7b7`, and #1459 is approved, so the blocker this branch inherited is gone.

- **Artifact defaulting** logs one line per defaulted service, and the doc comment now says what actually happens: session and memory panic, artifact answers 503, and defaulting that one trades a clear diagnostic for a server that works until it restarts. It names `cmd/launcher/prod` as the reason it matters. That comment is now more accurate than my review was.
- **`/health`** moved to `registerHealthRoute`, called from `Run`; `BuildBaseRouter` is back to router plus middleware, so an embedder's own handler is no longer shadowed. `charset` dropped to match `/api/health`.
- **`npm ci`**, with the reason in a comment.
- **CONTRIBUTING** now says the first refresh prints no compare URL and every one after it prints both.

Not pinning the `node:iron-trixie` digest is the right call. A digest you cannot build-test and have to bump by hand rots into a worse lie than a documented float, and `README.md:35` now states the float plainly. `npm ci` was the half that actually bought reproducibility.

Gates on this head: build clean, `go test -race -count=1 -shuffle=on ./cmd/launcher/...` green.

## The lows you asked for

Fair challenge — I cited a count and named one, which is not much use to you. Three of the six were in `api.go` and moved to #1459 with it; I have posted those there. The two that are still in this PR:

- **`rewriteBaseHref` silently drops `<base>` attributes that precede `href`** (`webui.go:101`, `:200-207`). The pattern matches from `<base ` through the closing quote of `href` and `ReplaceAllLiteral` replaces that whole span, so attributes *after* `href` survive and attributes *before* it are discarded. The committed bundle only has `href`, so nothing is lost today; it would bite silently the first time upstream emits `<base target="_self" href="/">`. Capturing and re-emitting the surrounding attributes, or matching only the `href="..."` pair inside the tag, avoids it.
- **`update-adk-web.sh` can leave a provenance-less bundle in the tree** — the case CONTRIBUTING now documents from the other side. Worth having the script write the provenance file before it writes `distr/`, so an interrupted refresh cannot produce a bundle nobody can trace.

The sixth was the CONTRIBUTING one, already fixed here.

Approving.

baptmont (approved):

PR #1461, opened 2026-09-01
docs(agent): explain why runAfterAgentCallbacks does not end the invocation

### LINKED ISSUE #370: Missing implementation in runAfterAgentCallbacks for invocation ended state

jjsasha63:
Replaces the `TODO set context invocation ended` stub in `runAfterAgentCallbacks` with the reasoning for why the call is absent.

The TODO reads as an oversight, and it has been picked up as one at least twice (#370, #371). It is deliberate:

- `Ended` gates whether an invocation should stop early. Nothing reads it after this point — `runAfterAgentCallbacks` is the last statement of the agent's `Run` iterator (`agent/agent.go:207`).
- The flag is local to this agent's context. `invocationContext.endInvocation` is a value field and `WithContext` copies the struct, so it never propagates to a parent; propagation runs parent-to-child only, at `workflow/agent_node.go:113`.
- This matches adk-python, where `_handle_before_agent_callback` sets `ctx.end_invocation = True` but `_handle_after_agent_callback` does not, and the after-agent call is likewise the last step of `run_async`.

## Credit

@MRuhan17 reported the gap in #370 and wrote the first version of this comment in #372. @baptmont established in #370 why the call should stay absent, and the wording here follows his explanation.

This supersedes #372, which had the same intent but went stale — 341 commits behind, conflicting, with review comments from December unaddressed. Closing that in favour of this.

Fixes #370

## Testing Plan

Comment-only change; no behavior change, so no new tests.

- `go build -mod=readonly ./agent/...` — passes
- `go test -race -count=1 ./agent/...` — all 8 packages pass
- `golangci-lint run ./agent/...` (v2.3.1, matching CI) — 0 issues
- `golangci-lint fmt --diff` — clean

Co-authored-by: Ruhulalemeen Mulla <ruhanmulla07@gmail.com>


baptmont (approved):

MRuhan17:
Thanks for picking this up. Glad to see the reasoning from #370/#372 documented here.

PR #396, opened 2025-12-04
Refactor IncludeContents in State to use internal type.

sheing-google:
Defines a new `IncludeContents` type and constants in `internal/llminternal` and updates the `State` struct to use it. This improves type safety and addresses a TODO comment.

Internal Refs:
- Defined `IncludeContents` type and constants in `internal/llminternal/agent.go`.
- Updated `State.IncludeContents` field type.
- Updated `llmagent.New` to cast config value to internal type.
- Updated `contents_processor.go` to use internal constant.

PR #1462, opened 2026-09-01
refactor(llmagent): give State.IncludeContents an internal type

jjsasha63:
`llminternal.State.IncludeContents` was a plain string and the contents processor compared it against the literal `"none"`, so the constants exported by `llmagent` had no reader on the internal side.

This defines the type and its constants in `llminternal`, converts at the `llmagent` boundary, and uses the constants at both comparison sites (`internal/llminternal/contents_processor.go` and `agent/llmagent/llm_agent_wrapper.go`). It resolves the `// TODO: internal type for includeContents` in `llmagent.New`.

The exported `llmagent.IncludeContents` keeps its own definition rather than aliasing the internal one. An alias changes the exported type's identity, which fails the apidiff check and renders godoc as a link into a package callers cannot open.

Supersedes #396 (originally proposed by @sheing-google, who also spotted the TODO) and #772 (@aurelsandu). Both stalled and no longer merge cleanly.

No behaviour change; existing tests cover both call sites.


wolo-lab (dismissed):
LGTM, thanks!

PR #1467, opened 2026-09-02
fix(openaimodel): preserve refusals and completed stream content

### LINKED ISSUE #1466: model/openaimodel drops refusal deltas from streamed responses
**Please make sure you read the contribution guide and file the issues in the right place.**

## Required Information

**Describe the Bug:**

The OpenAI Responses API emits `response.refusal.delta` events when a
model refuses a request. The OpenAI stream translator does not handle this
event type, so refusal deltas fall through its default branch and are silently
ignored.

A refusal-only stream with a complete terminal response may eventually recover
the refusal after streaming has ended. However, the refusal is not delivered
incrementally. If another supported delta, such as reasoning text, has already
been aggregated, the terminal-response fallback is skipped and the refusal can
be absent from the final ADK response.

This differs from the blocking response path, which converts refusal content
into a normal text part.

**Steps to Reproduce:**

1. Use the current ADK Go `main` branch.
2. Process a stream containing these events:

```json
{"type":"response.reasoning_text.delta","item_id":"rs_1","output_index":0,"content_index":0,"delta":"Checking the request","sequence_number":1}
{"type":"response.refusal.delta","item_id":"msg_1","output_index":1,"content_index":0,"delta":"I cannot help with that.","sequence_number":2}
{"type":"response.refusal.done","item_id":"msg_1","output_index":1,"content_index":0,"refusal":"I cannot help with that.","sequence_number":3}
{"type":"response.completed","sequence_number":4,"response":{"id":"resp_1","model":"gpt-4o-mini","status":"completed","output":[{"id":"rs_1","type":"reasoning","status":"completed","summary":[],"content":[{"type":"reasoning_text","text":"Checking the request"}]},{"id":"msg_1","type":"message","role":"assistant","status":"completed","content":[{"type":"refusal","refusal":"I cannot help with that."}]}]}}
```

3. Inspect the generated ADK response.

**Expected Behavior:**

The refusal delta is emitted as a non-thought text part, matching the blocking
response path. The final response contains both the preceding reasoning content
and the refusal text.

**Observed Behavior:**

The refusal events are ignored by the stream translator. When the reasoning
delta has already created an aggregated response, the final response can contain
only the reasoning content and omit the refusal.

**Environment Details:**

- ADK Library Version: `main` at `56d628d` (`google.golang.org/adk/v2`)
- OS: macOS
- Go Version: 1.26.6

**Model Information:**

- Which model is being used: Any OpenAI model that produces refusal events
  through the Responses API. The issue is in deterministic stream-event
  conversion and is not model-specific.

---

## Optional Information

**Regression:**

Not known. The current stream translator does not contain handling for refusal
delta events.

**Logs:**

```text
Input event:  response.refusal.delta
Observed ADK response: nil
Expected ADK response: a non-thought text part containing the refusal delta
```

**Screenshots / Video:**

N/A

**Additional Context:**

The OpenAI Go SDK exposes refusal events through
`AsResponseRefusalDelta()` and `AsResponseRefusalDone()`.

The blocking response converter already maps message content with type
`refusal` to a normal `genai.Part`. A focused fix would apply the same
representation to `response.refusal.delta` and treat
`response.refusal.done` as informational to avoid duplicate output.

Regression coverage should include refusal-only streaming and a mixed
reasoning/refusal stream.

```go
func TestStreamTranslatorRefusalDelta(t *testing.T) {
	event := decodeEvent(t, `{
		"type":"response.refusal.delta",
		"item_id":"msg_1",
		"output_index":0,
		"content_index":0,
		"delta":"I cannot help with that.",
		"sequence_number":1
	}`)

	got, err := newStreamTranslator().process(event)
	if err != nil {
		t.Fatal(err)
	}
	if got == nil {
		t.Fatal("refusal delta was ignored")
	}
}
```

- Always (100%) when a stream contains `response.refusal.delta`.

### LINKED ISSUE #1476: model/openaimodel: terminal-only message content is dropped after streamed deltas
**Please make sure you read the contribution guide and file issues in the right place.**

## Required Information

**Describe the Bug:**

The OpenAI streaming path does not reconcile terminal message content with
content already accumulated from streamed deltas.

After any translated delta reaches the stream aggregator, `aggregator.Close()`
returns a non-nil response. The fallback that converts the terminal response is
then skipped, so message content present only in `response.completed` is omitted
from the final streamed response.

The blocking path preserves the same terminal content.

**Steps to Reproduce:**

1. Use the current ADK Go `main` branch.
2. Add the reproduction test below to `model/openaimodel/openai_test.go`.
3. Run:

```text
go test -race -count=1 \
  -run '^TestReproTerminalOnlyMessageContent$' \
  -v ./model/openaimodel
```

**Expected Behavior:**

The final streaming response matches the blocking response and contains both
the reasoning part and the visible answer.

**Observed Behavior:**

The final streaming response contains only the reasoning part. The visible
answer from the terminal response is silently omitted.

This can leave a thought-only turn and cause the flow to call the model
repeatedly without surfacing the answer.

**Environment Details:**

- ADK Library Version: `main` at `ffddb2a`
- OS: macOS 26.6.2
- Go Version: 1.26.6

**Model Information:**

- Any OpenAI or OpenAI-compatible Responses API provider that may place some
  output only in the terminal response.

---

## Optional Information

**Regression:**

Not known. The all-or-nothing terminal fallback predates #1467.

**Logs:**

Reproduced locally with the race detector enabled:

```text
$ go test -race -count=1 \
    -run '^TestReproTerminalOnlyMessageContent$' \
    -v ./model/openaimodel

=== RUN   TestReproTerminalOnlyMessageContent
    content mismatch (-blocking +streamed):
      &genai.Content{
          Parts: []*genai.Part{
              &{Text: "Checking the request.", Thought: true},
    -         &{Text: "The final answer."},
          },
          Role: "model",
      }
--- FAIL: TestReproTerminalOnlyMessageContent
FAIL
FAIL    google.golang.org/adk/v2/model/openaimodel
```

The removed `The final answer.` part is present in the blocking response but
missing from the final streamed response. No race was reported.

**Screenshots / Video:**

N/A

**Additional Context:**

Related to #1467, where this behavior was identified while reviewing streamed
refusal handling. The underlying condition affects every translated delta type.

PR #1373 changes terminal fallback and tool-call handling, but its current
implementation still rebuilds from the terminal response only when the
aggregated response has no readable content. It therefore does not reconcile
additional terminal message content after reasoning, text, or refusal content
has already been aggregated.

Simply appending the terminal output would duplicate content normally repeated
by `response.completed`.

One possible policy is to treat a valid `response.completed` payload as
authoritative while preserving streamed content for `response.incomplete`, but
the intended reconciliation semantics should be confirmed first.

Would the maintainers prefer this behavior to be addressed as part of #1373 or
in a separate follow-up PR?

Add this test to `model/openaimodel/openai_test.go`:

```go
func TestReproTerminalOnlyMessageContent(t *testing.T) {
	const (
		reasoningDelta = `{"type":"response.reasoning_text.delta","item_id":"rs_1","output_index":0,"content_index":0,"delta":"Checking the request.","sequence_number":1}`
		body = `{"id":"resp_1","model":"stream-model","status":"completed","output":[{"id":"rs_1","type":"reasoning","status":"completed","summary":[],"content":[{"type":"reasoning_text","text":"Checking the request."}]},{"id":"msg_1","type":"message","role":"assistant","status":"completed","content":[{"type":"output_text","text":"The final answer."}]}]}`
		completed = `{"type":"response.completed","sequence_number":2,"response":` + body + `}`
	)

	streamed, err := runStream(t, evCreated, reasoningDelta, completed)
	if err != nil {
		t.Fatalf("streaming err = %v", err)
	}
	streamedFinal := assertTurnShape(t, streamed)

	blocking, err := runBlocking(t, body)
	if err != nil {
		t.Fatalf("blocking err = %v", err)
	}
	if len(blocking) != 1 {
		t.Fatalf("blocking emitted %d responses, want 1", len(blocking))
	}

	if diff := cmp.Diff(blocking[0].Content, streamedFinal.Content); diff != "" {
		t.Fatalf("content mismatch (-blocking +streamed):\n%s", diff)
	}
}
```

- Always (100%) when terminal-only message content follows a translated delta.

ktsoator:

Closes #1466. Closes #1476.

The OpenAI stream translator ignores `response.refusal.delta`. A refusal-only
response loses incremental delivery, and a refusal following reasoning can be
omitted entirely because the non-empty aggregator suppresses terminal-response
fallback. The resulting thought-only turn may cause up to 10 model calls without
surfacing an answer.

Translate non-empty refusal deltas into ordinary, non-thought text, matching the
blocking representation. Explicitly classify `response.refusal.done` as an
informational event; this documents its existing behavior and avoids emitting
the complete refusal again after its deltas.

When `response.completed` contains non-thought text or a function call, use
its content for the final response only if it retains the visible text and
function calls already collected from the stream. Otherwise, keep the aggregate.
When replacement is allowed, final reasoning follows the completed response;
thoughts already delivered as partials are not added back.

This does not merge content from both sources. For example, if visible text
streams first and the completed response contains only a tool call, the text
is retained and the call is not recovered. Calls whose streamed ID fell back
to `item_id` can also prevent recovery when the completed response uses a
different `call_id`. Both remain limitations of the current approach.

Preserve aggregated content when the completed payload is empty or cannot be
converted, and when the stream ends without a terminal event or its first
terminal event is `response.incomplete`. With no aggregated response, the
existing terminal fallback and conversion errors are unchanged. Final metadata
continues through the existing finalization path.

The partial-delta case was reproduced with synthetic SSE events. This validates
the adapter's behavior for that input; it does not establish how often a live
provider emits such a stream.

### Testing Plan

- Direct tests cover reasoning filtering, usable output, text containment,
  argument and ID differences, and one-to-one call matching. Each of the five
  mutations identified in the review makes a test fail.
- Full-stream tests cover reasoning that contains the visible answer, streamed
  thoughts omitted by the completed response, and recovery of completed text
  after a function call with `null` or `{}` arguments. The `null` case failed
  before argument normalization and passes afterward.
- Translator tests cover non-empty and empty refusal deltas.
- Full-stream tests compare final refusal content directly with blocking output,
  including partial deltas with and without `refusal.done`, and partial refusal
  after reasoning. They also verify that incremental text contains only the
  deltas actually received.
- Regression tests cover text completion, mixed output order, empty and
  unsupported terminal content, incomplete responses, and first-terminal-event
  precedence.
- `TestModel_GenerateStream_TerminalOnlyMessageContent` preserves the original
  #1476 reproduction: reasoning is streamed, while the visible answer appears
  only in `response.completed`. It fails against the pre-fix implementation and
  passes with the completed-content fix, comparing final content with blocking.
- Tests cover reasoning-only and shorter completed responses, completed
  responses that omit a streamed function call, and function calls that appear
  only in `response.completed`.
- Updated the existing function-call fixture: `response.output_item.added`
  previously omitted its `type`, and `response.completed` contained a message
  instead of the function call from the stream. Both events now describe the
  same call.

Before the terminal-content fix, all three new partial-refusal cases failed:

```text
expected final text: "I cannot help with that."
actual final text:   "I cannot "
```

They pass after the fix. Local validation on Go 1.26.6:

```text
go build -mod=readonly work
passed

all packages passed, including plugin/agentanalytics

golangci-lint run
0 issues in both modules

go mod tidy -diff
no differences in either module
```

Tests use a local HTTP server and synthetic SSE; no live model calls are made.

### Checklist

- [x] Read CONTRIBUTING.md and performed self-review.
- [x] Added regression tests and verified they fail before the fix.
- [x] Build, race tests, lint, and tidy checks pass locally.
- [x] No public API or dependency changes.
- [x] No dependent downstream changes.


karolpiotrowicz (commented):
The diagnosis holds up and the fix is the right shape. I checked out the branch at `2cdaa26`, ran the full suite sandboxed against both this branch and its merge base, and confirmed the bug end to end: your tests fail on `56d628d` and pass here. The two subtests fail *differently* on base, and that difference is worth putting on the record because it is the strongest argument for this change — `refusal only` still rebuilds correct final content through the terminal-response fallback and merely loses incremental delivery, while `refusal after reasoning` drops the refusal outright, since the reasoning delta makes `aggregator.Close()` non-nil and the fallback at [`openai.go:168-177`](https://github.com/google/adk-go/blob/2cdaa2679c988255e897c0e1807eb5e1aa2edb14/model/openaimodel/openai.go#L168-L177) never fires.

The user-visible cost on base is larger than the issue describes. A refusal after reasoning leaves a turn whose parts are all `Thought: true`, so `isThoughtOnlyTurn` sends the flow round again at [`base_flow.go:144-152`](https://github.com/google/adk-go/blob/2cdaa2679c988255e897c0e1807eb5e1aa2edb14/internal/llminternal/base_flow.go#L144-L152) — 10 model calls in total before it gives up and still surfaces no answer. Mapping the delta to a non-thought part at [`stream.go:53-59`](https://github.com/google/adk-go/blob/2cdaa2679c988255e897c0e1807eb5e1aa2edb14/model/openaimodel/stream.go#L53-L59) ends the turn on the first response. Choosing `Thought: false` is load-bearing rather than incidental, since the aggregator flushes its text buffer exactly when that flag flips, and marking a refusal as a thought would have concatenated it into the preceding reasoning part and hidden it from the user.

Nothing here blocks merging. Build, vet, `gofmt`, `golangci-lint`, `staticcheck` and `govulncheck` are all clean, `apidiff` reports no exported API change, coverage moves 86.3% to 86.8%, and the only two failing packages fail identically on the merge base for reasons unrelated to this change. This also merges cleanly with #1373, the other open PR touching the package.

Four things worth your attention, in order of consequence.

## Refusal deltas now suppress the terminal-response fallback

Because any delta reaching the aggregator makes `Close()` non-nil, a stream that carries a refusal delta *and* content present only on the terminal event now loses that content. Concretely: one `response.refusal.delta` plus a `response.completed` whose message also holds an `output_text` that never streamed yields `"Here is I cannot help."` on the merge base and `"I cannot help."` here. I want to be clear that I do not think you should change anything for it — the mechanism predates your change (a single reasoning delta already suppresses the same fallback and already loses batched text on base), and it is not separable from the feature, since streaming refusals incrementally *is* routing them through the aggregator. The naive repair of merging the terminal response back in would double-emit every ordinary refusal. It is worth a follow-up issue against the all-or-nothing condition at `openai.go:169`, which affects every delta type equally.

## The new end-to-end test does not actually pin parity with the blocking path

[`TestModel_GenerateStream_Refusal`](https://github.com/google/adk-go/blob/2cdaa2679c988255e897c0e1807eb5e1aa2edb14/model/openaimodel/openai_test.go#L363) asserts against hand-written `wantParts` literals rather than running the blocking path, the way [`TestModel_GenerateStream_MatchesBlocking`](https://github.com/google/adk-go/blob/2cdaa2679c988255e897c0e1807eb5e1aa2edb14/model/openaimodel/openai_test.go#L332-L337) does through `runBlocking`. The literals are correct today. The gap is drift: if the refusal mapping at [`response.go:79-80`](https://github.com/google/adk-go/blob/2cdaa2679c988255e897c0e1807eb5e1aa2edb14/model/openaimodel/response.go#L79-L80) ever changed, the two paths would diverge and this test would keep passing. Since matching the blocking representation is the stated goal, comparing against it directly would pin the thing the PR is actually claiming.

## The empty-delta guard has no test

[`stream.go:55-57`](https://github.com/google/adk-go/blob/2cdaa2679c988255e897c0e1807eb5e1aa2edb14/model/openaimodel/stream.go#L55-L57) is a live branch rather than defensive padding — without it an empty part reaches the aggregator and suppresses the fallback described above. No test in the package sends `"delta":""`. In fairness the three sibling guards in the same file are equally untested, so this is the house pattern and not something you introduced. One extra table row would cover it.

## One correction to the PR description

Adding `responseRefusalDone` at [`stream.go:110`](https://github.com/google/adk-go/blob/2cdaa2679c988255e897c0e1807eb5e1aa2edb14/model/openaimodel/stream.go#L110) does not prevent duplicate content, because the pre-existing `default:` branch already returned `(nil, nil)` for that event — the base branch never double-emitted either. I checked what would happen if it did emit: the `.done` event repeats the full refusal text the deltas already carried, so emitting from it produces `"I cannot help with that.I cannot help with that."` and fails your own tests. The line is still worth keeping, since it moves the event from "unknown, ignored" to "known, deliberately informational", but it is documentation rather than a behavior change.

One thing I could not settle: a refusal delivered only through `response.refusal.done` with no deltas at all is still dropped when another delta has already reached the aggregator. That behaves identically on the merge base, so it is not something this PR introduces, and closing it would need per-item tracking of whether any delta arrived rather than a change to the `.done` branch. Mentioning it only so it is written down somewhere.

ktsoator:
Thank you for the detailed review and for tracing the different failure modes
on the merge base.

I've addressed the test and documentation follow-ups in `8b685f8`:

- `TestModel_GenerateStream_Refusal` now compares the final streamed content
  directly with the blocking response for both refusal-only and mixed
  reasoning/refusal outputs.
- Added coverage for empty refusal deltas.
- Updated the PR description to clarify that explicitly handling
  `response.refusal.done` documents its existing informational behavior rather
  than preventing previously duplicated output.
- Documented the impact of a thought-only turn potentially causing up to 10
  model calls.

I also merged the latest `main`.

I agree that the all-or-nothing terminal-response fallback is a pre-existing,
broader issue. I have kept it out of this PR and plan to track it separately.

All CI checks are now passing. Thank you again for the thorough feedback. This
should be ready for another look.

model/openaimodel/stream_test.go
```diff
@@ -45,6 +45,37 @@ func TestStreamTranslator_TextDelta(t *testing.T) {
...
+	part := resp.Candidates[0].Content.Parts[0]
+	if part.Text != "I cannot help with that." {
+		t.Errorf("part.Text = %q, want %q", part.Text, "I cannot help with that.")
+	}
+	if part.Thought {
+		t.Error("part.Thought = true, want false")
+	}
+}
+
+func TestStreamTranslator_EmptyRefusalDelta(t *testing.T) {
```
karolpiotrowicz, line 67: This pins the guard — I deleted the three lines `if delta.Delta == "" { return nil, nil }` from `stream.go` and nothing else, and this test failed. What it does not reach is the guard's actual consequence: that an empty part would otherwise make `aggregator.Close()` non-nil and suppress the terminal-response rebuild in `openai.go`. A stream case that sends `"delta":""` and then puts the answer only on `response.completed` would exercise that. Optional, and worth saying that yours is already the best-covered of the four: `stream.go` has the same `delta.Delta == ""` guard on `output_text`, `reasoning_text` and `reasoning_summary_text`, and no test in the package sends an empty delta for any of those three.

model/openaimodel/openai_test.go
```diff
@@ -360,6 +360,76 @@ func TestModel_GenerateStream_MatchesBlocking(t *testing.T) {
 	}
 }
 
+func TestModel_GenerateStream_Refusal(t *testing.T) {
```
karolpiotrowicz, line 396: Six of the seven other streaming tests in this file open with a sentence naming the invariant they pin, and this one carries the most involved fixtures of any of them. One line — that a streamed refusal reaches the caller as non-thought text and the turn's final content matches the blocking path — would match the local convention. Purely optional.

karolpiotrowicz (commented):
The three follow-ups from the last round are done, and I checked each rather than taking the description's word for it. `TestModel_GenerateStream_Refusal` now really does run the blocking path — `runBlocking` drives it through the same local server the streaming case uses — and `cmp.Diff(blocking[0].Content, final.Content)` at [`openai_test.go:426`](https://github.com/google/adk-go/blob/8b685f8b73371763ffe432f1219a134d848e2146/model/openaimodel/openai_test.go#L426) compares whole `*genai.Content` values. That is a stricter comparison than [`TestModel_GenerateStream_MatchesBlocking`](https://github.com/google/adk-go/blob/8b685f8b73371763ffe432f1219a134d848e2146/model/openaimodel/openai_test.go#L330-L332) makes, since that one flattens every part into one string with `allText` and so cannot see part boundaries or the `Thought` flag at all. The empty-delta guard is now covered, and the coverage is real: I deleted the three lines `if delta.Delta == "" { return nil, nil }` from [`stream.go:55-57`](https://github.com/google/adk-go/blob/8b685f8b73371763ffe432f1219a134d848e2146/model/openaimodel/stream.go#L55-L57) and nothing else, and `TestStreamTranslator_EmptyRefusalDelta` failed. The description now describes `response.refusal.done` correctly.

Nothing blocks merging. Build, vet, `gofmt`, `staticcheck`, `golangci-lint` and `govulncheck` are clean, `apidiff` reports no exported API change, and coverage of the package moves 86.3% to 87.0%. Three packages fail the full suite, and all three fail identically on the merge base — `telemetry` because it dials the metadata server, and the two `agent/remoteagent` packages because they are timing-sensitive and flaked under load. I re-ran those two in isolation twice on each tree and all four runs passed, so neither is attributable here. The branch is behind main but merges into it cleanly with the change intact. Worth noting that #1359 also touches `stream.go` and `stream_test.go`, so that is the one pairing to check before landing both.

Two things worth your attention, neither blocking.

**An empty refusal delta is never driven through the full stream.** The unit test pins the guard, and I confirmed that by mutation. What no test exercises is the guard's actual consequence — that an empty part reaching the aggregator would make `Close()` non-nil and suppress the terminal-response rebuild at [`openai.go:168-170`](https://github.com/google/adk-go/blob/8b685f8b73371763ffe432f1219a134d848e2146/model/openaimodel/openai.go#L168-L170). A stream case sending `"delta":""` and then putting the answer only on `response.completed` would pin the thing the guard exists for, rather than the guard's return value. Optional, and worth saying that yours is already the best-covered of the four: `stream.go` has the same `delta.Delta == ""` guard on `output_text`, `reasoning_text` and `reasoning_summary_text`, and no test in the package sends an empty delta for any of those three.

**One more shape for #1476, if you want it there.** Since a refusal delta now reaches the aggregator, a refusal that follows ordinary `output_text` in the same turn gets concatenated into the preceding part, because the aggregator flushes its text buffer only when the `Thought` flag flips. Streaming yields one part where the blocking path yields two. This is the same all-or-nothing behavior you already filed, seen from a different angle, and it is not new: I ran the same shape on the merge base with no refusal event anywhere — two plain `output_text` items from two output items merge into a single part there too. On the refusal shape specifically the change is a clear improvement, since the base loses the refusal text outright and this keeps it. I mention it only because it is the part of the picture #1476's reproduction does not currently show.

Two small notes that are genuinely optional. `TestModel_GenerateStream_Refusal` has no doc comment while six of the seven other streaming tests in the file open with one naming the invariant they pin. And separately, [`response.go:79-80`](https://github.com/google/adk-go/blob/8b685f8b73371763ffe432f1219a134d848e2146/model/openaimodel/response.go#L79-L80) appends a refusal part with no `!= ""` guard while the `output_text` case just above it has one — that is pre-existing and outside this change, so leave it unless you want to pick it up.

One thing you may not know, which cuts in your favour on scope. adk-python has this same bug, in the same place: its Responses translator at `src/google/adk/labs/openai/_openai_responses_llm.py` handles refusal in the blocking converter but its stream dispatch has no `response.refusal.delta` or `response.refusal.done` branch either. So this lands ahead of Python rather than diverging from it. The one real difference is representation — adk-python prefixes refusal text with `'OpenAI refusal: '` and adk-java with `"[[REFUSAL]]: "`, while adk-go emits it unmarked. That difference already exists in adk-go's blocking path, so following it here is the right call for this change, but it is probably worth a maintainer deciding whether adk-go wants a marker at all.

ktsoator:
Thanks for the follow-up. I addressed both optional notes.

The refusal stream table now includes a full-stream case with an empty `response.refusal.delta`, followed by refusal content available only in `response.completed`. The case verifies that no incremental refusal text is emitted and that the final content still matches the blocking response.

I also added a comment documenting the invariant covered by `TestModel_GenerateStream_Refusal`.

I mutation-checked the new case by temporarily removing the empty-delta guard. The test failed with `Parts: nil` instead of the terminal refusal content, then passed again after restoring the guard.

The branch has also been synchronized with the latest `main`.

- `go vet ./...` in both modules
- `golangci-lint run` in both modules

model/openaimodel/openai_test.go
```diff
@@ -360,6 +360,89 @@ func TestModel_GenerateStream_MatchesBlocking(t *testing.T) {
...
+			if got, want := streamedRefusal.String(), tc.wantStreamedRefusal; got != want {
+				t.Errorf("streamed refusal = %q, want %q", got, want)
+			}
+
+			if diff := cmp.Diff(blocking[0].Content, final.Content); diff != "" {
+				t.Errorf("content mismatch (-blocking +streamed):\n%s", diff)
+			}
+		})
+	}
+}
```
karolpiotrowicz, line 493: The one scenario this table does not reach is a stream that errors after the refusal has started. `TestModel_GenerateStream_Failed` covers error-after-delta, but with an `output_text` delta, so nothing pins the refusal case.

It is worth a row because the behaviour is one of the clearer wins in this change: with `evCreated, refusalOnlyDelta1, refusalOnlyDelta2, evFailed`, your branch yields both refusal deltas to the caller before the error, where main yields nothing at all. Those arrive as `Partial: true` responses with no aggregated turn, which matches how the package already treats ordinary text on an erroring stream.

model/openaimodel/openai_test.go
```diff
@@ -360,6 +360,89 @@ func TestModel_GenerateStream_MatchesBlocking(t *testing.T) {
...
+					if !part.Thought {
+						streamedRefusal.WriteString(part.Text)
+					}
+				}
+			}
+			if got, want := streamedRefusal.String(), tc.wantStreamedRefusal; got != want {
+				t.Errorf("streamed refusal = %q, want %q", got, want)
+			}
+
+			if diff := cmp.Diff(blocking[0].Content, final.Content); diff != "" {
```
karolpiotrowicz, line 488: Optional. This compares whole `Content`, which is stricter on content than `TestModel_GenerateStream_MatchesBlocking`, but it is the only refusal test comparing against the blocking path and it leaves `FinishReason` and `UsageMetadata` unasserted. Those two are pinned elsewhere for ordinary text, so adding them here would be belt-and-braces rather than new coverage.

karolpiotrowicz (dismissed):
The empty-refusal follow-up from the last round is done, and it holds up under the check I care about rather than just reading correctly. [`openai_test.go:395-400`](https://github.com/google/adk-go/blob/ea23a6fe6ab26bce38e14df69250e943f48ba34b/model/openaimodel/openai_test.go#L395-L400) drives `"delta":""` through the full stream and compares the caller's final `*genai.Content` against the blocking path, which is the thing the guard exists for rather than the guard's return value. I checked it three ways: deleting the `delta.Delta == ""` guard at [`stream.go:55-57`](https://github.com/google/adk-go/blob/ea23a6fe6ab26bce38e14df69250e943f48ba34b/model/openaimodel/stream.go#L53-L59) fails it, making the terminal-response rebuild at [`openai.go:168-170`](https://github.com/google/adk-go/blob/ea23a6fe6ab26bce38e14df69250e943f48ba34b/model/openaimodel/openai.go#L168-L170) never fire fails it, and making that rebuild produce empty content fails it. The doc comment at [`openai_test.go:363-364`](https://github.com/google/adk-go/blob/ea23a6fe6ab26bce38e14df69250e943f48ba34b/model/openaimodel/openai_test.go#L363-L364) is there too.

One measured regression against `main` should land here rather than in a follow-up, and nothing else stands in the way. Build, vet, `gofmt`, `go mod tidy -diff`, `golangci-lint`, `staticcheck`, `govulncheck` and `apidiff` are clean, coverage of the package moves 86.6% to 87.2%, and the two packages that fail the full suite fail identically on the merge base, on current main, and on the merge result, so neither is attributable here. I also ran the second module, `plugin/agentanalytics`, separately, because `go test ./...` does not cross a nested `go.mod` — it passes on both sides.

## One shape is worse than main, and it is the sharper half of the issue you already filed

I said last round that a refusal delta now suppresses the terminal-response rebuild, and that the naive repair would double-emit. That is still my view. What I did not have then is the refusal-specific version, and it is worth adding to #1476, because #1476's reproduction currently shows content being lost around a refusal rather than the refusal itself being lost.

A single refusal delta that carries only part of the refusal, with the whole refusal on `response.completed`, streams as `"I cannot "` with `TurnComplete: true` and `FinishReason: STOP` — a truncated refusal presented as a cleanly finished answer. On main ([a6114727](https://github.com/google/adk-go/commit/a61147276167f6b1b636d863f23a030c0a6babdf)) the same events produce `"I cannot help with that."`, matching blocking exactly, because the unhandled event type left the aggregator empty and the rebuild ran.

I want to be clear about what this does and does not mean. The mechanism is the all-or-nothing condition at [`openai.go:169`](https://github.com/google/adk-go/blob/ea23a6fe6ab26bce38e14df69250e943f48ba34b/model/openaimodel/openai.go#L169), which you do not touch, and the same loss already happens on main for an ordinary `output_text` delta and for a reasoning delta. Your change stops refusal being accidentally exempt from a rule every other delta type already obeys, which is the right direction. I could not establish whether the Responses API ever emits a partial refusal delta alongside a complete terminal payload, and that is what would decide how much it matters. A captured stream showing it would settle it.

## Smaller things

The interleaved shape I mentioned last round is not yours to fix, and I can now show that rather than assert it. One message carrying two ordinary `output_text` content blocks and no refusal anywhere already yields one streamed part against blocking's two, identically on your branch, on the merge base and on current main — the merge happens at [`stream_aggregator.go:109-115`](https://github.com/google/adk-go/blob/ea23a6fe6ab26bce38e14df69250e943f48ba34b/internal/llminternal/stream_aggregator.go#L109-L115) and predates you. On the refusal shape specifically your branch is better than main, which drops the refusal text outright.

On landing: this merges cleanly into current main and pairs cleanly with #1359, #1392 and #1395 individually, with `go build ./model/...` and `go test ./model/openaimodel/...` green in each pair, including the three-way with #1359 and #1392. #1359 is the one that also edits `stream.go`. Worth knowing that #1392 and #1395 conflict with each other in `doc.go`, so whoever sequences those two has a merge to resolve — it does not involve this branch.

ktsoator:
I reproduced the partial-refusal case against the previous PR head and addressed
it here. A non-empty delta containing `I cannot ` followed by the full refusal in
`response.completed` now retains the full final text, both with and without a
`response.refusal.done` event, and after reasoning. The incremental responses
still contain only the deltas that were received.

The fix uses usable content from the first `response.completed` for the final
response instead of appending it to the aggregate. Empty or unconvertible
completed payloads preserve the aggregate, as do `response.incomplete` and
streams without a terminal event. Only final `Content` is replaced; final
metadata keeps its existing path. This also covers the completed-response case
in #1476, which remains open for the broader reconciliation discussion.

The three new partial-refusal tests failed before the fix and pass afterward.
Additional coverage checks terminal precedence, empty and unsupported content,
mixed output order, and function calls. I corrected the existing tool-call
fixture so its terminal payload contains the same tool call, and independently
asserted that the tool call is still emitted exactly once incrementally.

Full workspace build and race tests passed, including `plugin/agentanalytics`;
lint reports 0 issues in both modules, and both tidy checks are clean. The
partial-delta evidence is from synthetic SSE, so I am not claiming a measured
frequency on live OpenAI streams. This is ready for another review.

model/openaimodel/stream_test.go
ktsoator: Added in ea23a6f. The `empty delta falls back to terminal response` row in `TestModel_GenerateStream_Refusal` sends an empty refusal delta through `runStream`, with the refusal text available only in `response.completed`. It verifies that no incremental refusal text is emitted and compares the final `Content` directly with the blocking path. The translator unit test also remains in place to pin the guard's nil return.

model/openaimodel/openai_test.go
ktsoator: Added in ea23a6f. The comment above `TestModel_GenerateStream_Refusal` now states that non-empty streamed refusals reach callers as non-thought text and that final content matches the blocking output.

model/openaimodel/openai_test.go
ktsoator: You're right that this exact refusal-then-error sequence is not covered yet. `TestModel_GenerateStream_Failed` currently covers an ordinary text delta; the new cases in 01586af cover completed-response reconciliation. `generateStream` still returns immediately on `evFailed`, before the new final-content reconstruction, so that change preserves the existing failure path. A test asserting that both refusal deltas arrive as partial responses before the error, with no completed aggregate, remains a coverage gap in this revision.

model/openaimodel/openai_test.go
ktsoator: I have kept these optional assertions in `TestModel_GenerateStream_MatchesBlocking` for this revision. That test compares both `FinishReason` and `UsageMetadata` directly with the blocking result, and `finalizeStreamResponse` still supplies those fields for all delta types. The change in 01586af replaces only final `Content`; the refusal table itself does not yet repeat the metadata comparisons.

ktsoator:
Added the original #1476 reproduction as a dedicated regression test in `1c6293a`. It fails against the pre-fix implementation and passes with the completed-content fix.

All local checks and CI pass. This PR now includes `Closes #1476` and is ready for another review.

model/openaimodel/openai.go
```diff
@@ -182,6 +183,19 @@ func (m *openAIModel) generateStream(ctx context.Context, params responses.Respo
 				return
 			}
 			final = converters.Genai2LLMResponse(genaiResp)
+		} else if sawCompletedResponse {
+			// A completed response can contain text that never arrived as a
+			// delta. Use its content for the final response without re-emitting
+			// it as a delta. Keep the aggregate if the snapshot is unusable.
+			if genaiResp, err := convertResponse(openaiResp); err == nil {
```
karolpiotrowicz, line 193: Non-blocking. Swallowing the error here is the right call — the comment says so and it is what keeps an unusable snapshot from destroying the aggregate — but nothing records that it happened. A gateway whose terminal snapshots are systematically malformed loses streaming/blocking agreement on every single turn, forever, and no operator has anything to look at. A debug-level log naming the error would make that diagnosable without changing behaviour.

model/openaimodel/openai.go
```diff
@@ -182,6 +183,19 @@ func (m *openAIModel) generateStream(ctx context.Context, params responses.Respo
...
+				content := genaiResp.Candidates[0].Content
+				for _, part := range content.Parts {
+					if part.Text != "" || part.FunctionCall != nil {
+						final.Content = content
```
karolpiotrowicz: Non-blocking, and only reachable on a malformed stream. `openaiResp` is frozen by the `!sawFinalResponse` guard, but the loop keeps feeding later events to the aggregator, so a delta arriving after `response.completed` is yielded to the caller as a partial and then discarded from the final by this assignment. With `evCreated, evDelta1, evCompleted, {"type":"response.output_text.delta","delta":"XYZ"}` the caller sees `XYZ` mid-turn and gets `hello` at the end. On main the final is `hel` + `XYZ`. Content after the response has declared itself complete is out-of-spec input, so this is worth knowing about rather than worth guarding.

karolpiotrowicz (dismissed):
The terminal-response regression from the last round is fixed. I re-ran the three shapes that gated it — a refusal streaming with the reasoning item carried only by `response.completed`, one partial refusal delta with the whole refusal on the terminal event, and one refusal delta with batched `output_text` only on the terminal event — and all three now match the blocking path exactly. On these your branch is better than main rather than merely level with it: main reaches the same final content by discarding the refusal stream altogether, so the caller receives nothing at all mid-turn, where yours delivers `"I cannot help with that."` as it arrives and still agrees with blocking at the end. `TestModel_GenerateStream_TerminalOnlyMessageContent` earns its place too — with [the new branch](https://github.com/google/adk-go/blob/1c6293a416f4cab3d46ffa0bc261b39181ee16ab/model/openaimodel/openai.go#L186-L199) deleted it goes red on the missing `The final answer.`, which is exactly what #1476 was about.

What I want to raise is the shape of the fix rather than the bug it closes. Three things need changing before this merges: the usability guard must stop counting a reasoning part as recoverable content, the branch needs a decision on whether a snapshot narrower than the stream should be allowed to override it at all, and the `|| part.FunctionCall != nil` arm needs a test that fails without it. The fourth item, the rewritten fixture, needs a line in the PR description whichever way the second one goes.

## The fallback replaces the whole aggregate rather than filling the gap

[`openai.go:186-199`](https://github.com/google/adk-go/blob/1c6293a416f4cab3d46ffa0bc261b39181ee16ab/model/openaimodel/openai.go#L186-L199) assigns `final.Content = content` outright. So for every stream that ends in `response.completed`, the final response becomes whatever the terminal snapshot converts to, and anything the stream produced that the snapshot does not repeat is gone. That is the correct answer when the snapshot is a superset of the deltas, which is the case the change targets. It is the wrong answer when the snapshot is narrower, and the guard above it does not distinguish the two.

## A reasoning-only snapshot destroys the answer, and the turn then loops

This is the one I would fix first. The guard at [`openai.go:193`](https://github.com/google/adk-go/blob/1c6293a416f4cab3d46ffa0bc261b39181ee16ab/model/openaimodel/openai.go#L193) tests `part.Text != ""` without consulting `part.Thought`, and `convertOutputItems` puts reasoning text in `Part.Text` with `Thought: true`. A snapshot whose only surviving part is reasoning therefore passes the guard and replaces the answer the caller already saw.

It does not take a snapshot that contradicts its own stream to get there. `convertOutputItems` already drops an `output_text` whose text is `""`, so a snapshot that simply declines to repeat streamed text converts to reasoning only. That snapshot shape is one you already treat as real — it is the `empty text` row at [`openai_test.go:550`](https://github.com/google/adk-go/blob/1c6293a416f4cab3d46ffa0bc261b39181ee16ab/model/openaimodel/openai_test.go#L550), which correctly keeps the aggregate. Add one reasoning item to that same body and the outcome flips:

```go
// head:  final Content.Parts == [THOUGHT("Weighing options")]        -- "hello" is gone
// main:  final Content.Parts == [THOUGHT("Weighing options"), TEXT("hello")]
const body = `{"id":"resp_1","model":"stream-model","status":"completed","output":[` +
    `{"id":"rs_1","type":"reasoning","summary":[{"type":"summary_text","text":"Weighing options"}],"content":[]},` +
    `{"id":"msg_1","type":"message","role":"assistant","content":[{"type":"output_text","text":""}]}]}`
got, _ := runStream(t, evCreated,
    `{"type":"response.reasoning_summary_text.delta","item_id":"rs_1","delta":"Weighing options"}`,
    evDelta1, evDelta2,
    `{"type":"response.completed","response":`+body+`}`)
// got[len(got)-1].Content holds one part, and it is the thought
```

The cost is not only the lost answer. A final whose parts are all `Thought` satisfies [`isThoughtOnlyTurn`](https://github.com/google/adk-go/blob/1c6293a416f4cab3d46ffa0bc261b39181ee16ab/internal/llminternal/base_flow.go#L176-L190), so [`Flow.Run`](https://github.com/google/adk-go/blob/1c6293a416f4cab3d46ffa0bc261b39181ee16ab/internal/llminternal/base_flow.go#L146-L154) declines to stop and calls the model again, up to ten times, with a growing history. The user gets no answer and pays for ten more calls.

`(part.Text != "" && !part.Thought) || part.FunctionCall != nil` fixes it. I applied that one change and ran the package: the suite stays green with no failing case, and the shape above goes back to `[THOUGHT("Weighing options"), TEXT("hello")]`, which is what main returns. It costs nothing you are trying to recover, because every snapshot in your new tests carries a non-thought part.

## A tool call that streamed can be dropped from the final response

The clearest way to see this is your own test. `TestModel_GenerateStream_FunctionCall` at the merge base ([`openai_test.go:412-417`](https://github.com/google/adk-go/blob/440e35202c325a7310557e3bacecdeb3ad07b989/model/openaimodel/openai_test.go#L412-L417)) drove `runStream(t, evCreated, evItemAdded, evArgsDone, evCompleted)`, where the shared `evCompleted` body is a message and no function call. Run that body unchanged against this branch and it fails — `calls` comes back nil and the final text is `hello`. Against main ([a6114727](https://github.com/google/adk-go/commit/a61147276167f6b1b636d863f23a030c0a6babdf)) it passes.

That matters more than a content difference, because the final response is the only one that reaches tool dispatch: [`base_flow.go:678`](https://github.com/google/adk-go/blob/1c6293a416f4cab3d46ffa0bc261b39181ee16ab/internal/llminternal/base_flow.go#L678-L683) does `if resp.Partial { continue }` immediately before `handleFunctionCalls`. A call the stream emitted and the snapshot omits is a tool that silently never runs.

The arm of the guard that governs the opposite case is also untested. Delete `|| part.FunctionCall != nil` from [`openai.go:193`](https://github.com/google/adk-go/blob/1c6293a416f4cab3d46ffa0bc261b39181ee16ab/model/openaimodel/openai.go#L193) and the whole package still passes, yet it is not dead code: with a reasoning delta streaming and the tool call arriving only on `response.completed`, that arm is the only reason the final response carries the call at all. `TestModel_GenerateStream_FunctionCall` cannot see it, because there the aggregate already holds the identical call, so keeping the aggregate and adopting the snapshot produce the same result. A case where the aggregate has no call and the snapshot does would pin it.

I can argue both sides of whether that snapshot is realistic, and I want to be straight that I could not settle it: nothing offline tells me whether a real Responses endpoint, or one of the OpenAI-compatible gateways `ClientConfig.BaseURL` exists for, ever sends a `response.completed` that omits an item it just streamed. A captured stream either way would settle it. What I can say is that the change assumes a provider may under-populate the deltas — which is the bug you are fixing — while assuming the same provider can never under-populate the snapshot, and nothing here justifies treating those differently.

## The fixture change belongs in the description

Related to the above, and worth separating from it: the fixture at [`openai_test.go:625-629`](https://github.com/google/adk-go/blob/1c6293a416f4cab3d46ffa0bc261b39181ee16ab/model/openaimodel/openai_test.go#L625-L629) was rewritten in the same commit that makes the old one fail, with no note anywhere saying so. The new fixture is a better wire model — the merge-base one had an `output_item.added` with no `type` field and a terminal body that contradicted its own stream — and the added streamed-call and blocking-parity assertions are a real improvement. But a pre-existing test that goes red against new production code, edited alongside it, reads as a laundered regression to anyone who does not re-run the old one. Say in the PR description that the fixture changed and why, whichever way the point above lands.

## On landing

Open PR #1373 touches `openai.go` and `openai_test.go` on the same terminal-event handling and conflicts with this branch textually, so whoever sequences the two has a merge to resolve, and it is worth a look for a semantic clash as well. Otherwise this merges cleanly into current main with the change intact, and build, vet, `gofmt`, `golangci-lint`, `staticcheck` and `apidiff` are all clean on both sides.

model/openaimodel/openai.go
ktsoator: Thanks for pointing this out. I left logging out of this change because `openaimodel` does not currently have a debug logger to use, and adding one would broaden the scope. The aggregate is still retained if conversion fails.

model/openaimodel/openai.go
ktsoator: Agreed. A delta after `response.completed` is malformed input, so I left this behavior unchanged.

ktsoator:
Thanks for the detailed review. I addressed the requested changes in `66d2c4c`.

The completed response now replaces the aggregate only when it contains usable non-thought text or a function call and still includes the visible text and function calls collected from the stream. Otherwise, the aggregate is kept.

I added cases for reasoning-only and shorter completed responses, a completed response that omits a streamed function call, and a function call that appears only in `response.completed`. I also confirmed that the last case fails when the function-call condition is removed.

The PR description now explains why the old function-call fixture was changed. All local checks and CI are passing.

model/openaimodel/openai.go
```diff
@@ -182,12 +185,80 @@ func (m *openAIModel) generateStream(ctx context.Context, params responses.Respo
...
+		}
+		if part.FunctionCall != nil {
+			aggregateCalls = append(aggregateCalls, part.FunctionCall)
+		}
+	}
+	for _, part := range completed.Parts {
+		if part == nil {
+			continue
+		}
+		if part.Text != "" && !part.Thought {
```
karolpiotrowicz, line 235: This filter is the line that closes the problem from the last round, and nothing tests it. Delete `&& !part.Thought` here and `go test ./model/openaimodel/` is still green, but a stream that delivered `"hello"` against a reasoning-only snapshot whose text is `"hello is the answer"` now ends the turn as reasoning only. The existing `reasoning-only snapshot does not replace an answer` case cannot catch it, because its reasoning text (`"Weighing options"`) does not contain the streamed answer, so the substring check at line 241 fails with or without the filter. A case whose reasoning text does contain the answer kills it.

model/openaimodel/openai.go
```diff
@@ -182,12 +185,80 @@ func (m *openAIModel) generateStream(ctx context.Context, params responses.Respo
...
+			completedText.WriteString(part.Text)
+			usable = true
+		}
+		if part.FunctionCall != nil {
+			completedCalls = append(completedCalls, part.FunctionCall)
+			usable = true
+		}
+	}
+	if !usable || !strings.Contains(completedText.String(), aggregateText.String()) {
```
karolpiotrowicz, line 244: Two independent decisions on one line, neither of them pinned.

Drop the `!usable ||` term and the suite stays green, but a reasoning-only stream against a reasoning-only snapshot carrying different text now takes the snapshot's reasoning instead of keeping the streamed one. Change `strings.Contains` to `strings.HasPrefix` and the suite also stays green, while the batched-refusal shape this PR fixes loses its unstreamed preamble — a snapshot of `[output_text "Here is ", refusal "I cannot help."]` against a streamed `"I cannot help."` drops the `"Here is "`.

Worth noting that `Contains` is the right choice of the three: `HasSuffix` is killed by eight existing cases, and `HasPrefix` breaks the shape above. It just has nothing asserting it.

model/openaimodel/openai.go
```diff
@@ -182,12 +185,80 @@ func (m *openAIModel) generateStream(ctx context.Context, params responses.Respo
...
+		return false
+	}
+
+	matched := make([]bool, len(completedCalls))
+	for _, aggregateCall := range aggregateCalls {
+		found := false
+		for i, completedCall := range completedCalls {
+			if !matched[i] && reflect.DeepEqual(aggregateCall, completedCall) {
```
karolpiotrowicz, line 252: `reflect.DeepEqual` compares the struct as it was decoded rather than the call it represents, and the two decode paths populate two fields differently.

`Args`: an `arguments` payload of the literal string `"null"` leaves `Args` nil on the stream side, because `json.Unmarshal` of `null` into a map leaves it nil, while `functionCallArgs` on the snapshot side normalises that to `map[string]any{}`. `ID`: the stream side falls back to `done.ItemID` when no `response.output_item.added` carried a `call_id`, where the snapshot always uses `item.CallID`. Either difference makes the same call compare unequal, the guard returns false, and any snapshot-only content alongside it is dropped. I measured the `Args` case — with a `ping` call sent as `"arguments":"null"` and a snapshot carrying that call plus the text `"pong"`, the `"pong"` never reaches the final.

Comparing on name plus normalised args plus normalised id would avoid it. Separately, relaxing this to a name-only match leaves the suite green while letting a snapshot call substitute different arguments for the streamed ones, so the strictness here is load-bearing and untested.

model/openaimodel/openai.go
```diff
@@ -182,12 +185,80 @@ func (m *openAIModel) generateStream(ctx context.Context, params responses.Respo
...
+		if part.FunctionCall != nil {
+			completedCalls = append(completedCalls, part.FunctionCall)
+			usable = true
+		}
+	}
+	if !usable || !strings.Contains(completedText.String(), aggregateText.String()) {
+		return false
+	}
+
+	matched := make([]bool, len(completedCalls))
```
karolpiotrowicz, line 248: The `matched[]` bookkeeping is what stops one snapshot call satisfying two streamed calls, and removing it leaves the suite green. The input that separates the two is two byte-identical streamed calls — same `call_id`, same arguments — against a snapshot carrying one, where the bookkeeping keeps both and its absence erases one. That needs a provider emitting a duplicate `call_id`, so this is the least realistic of the untested decisions here and I would not hold the PR for it. A direct table test of the predicate would cover it and the others at the same time.

model/openaimodel/openai.go
```diff
@@ -182,12 +185,80 @@ func (m *openAIModel) generateStream(ctx context.Context, params responses.Respo
 				return
 			}
 			final = converters.Genai2LLMResponse(genaiResp)
+		} else if sawCompletedResponse {
+			// A completed response can contain text that never arrived as a
+			// delta. Use its content for the final response without re-emitting
+			// it as a delta. Keep the aggregate if the snapshot is unusable or
+			// omits content that was already streamed.
+			if genaiResp, err := convertResponse(openaiResp); err == nil {
```
karolpiotrowicz, line 193: Discarding this error means the same terminal body produces three different outcomes depending on how it arrived: the blocking path returns an error, a stream with no deltas returns an error, and a stream with deltas silently keeps the aggregate. The silent case is the right call — you have a good answer in hand and failing the turn would throw it away — but nothing records that recovery was attempted and abandoned. `convertOutputItems` aborts on the first item type it does not know, so a snapshot containing a `web_search_call` or an `mcp_call` alongside a message disables recovery entirely, and an operator seeing every turn miss its terminal content has nothing to go on. A log line at this branch would be enough.

model/openaimodel/openai.go
```diff
@@ -182,12 +185,80 @@ func (m *openAIModel) generateStream(ctx context.Context, params responses.Respo
...
+
+	var aggregateText, completedText strings.Builder
+	var aggregateCalls, completedCalls []*genai.FunctionCall
+	usable := false
+	for _, part := range aggregate.Parts {
+		if part == nil {
+			continue
+		}
+		if part.Text != "" && !part.Thought {
+			aggregateText.WriteString(part.Text)
```
karolpiotrowicz, line 225: Both sides concatenate their text parts with no separator, so the comparison is over `"foobar"` rather than over the parts. A snapshot that splits the same content differently fails the check and is rejected: streaming `"foo"` then `"bar"` against a snapshot of `["foo ", "bar"]` keeps the aggregate and loses the space. This errs toward keeping what the caller saw, which is the safe direction, so it is a nit rather than something to fix.

model/openaimodel/openai.go
```diff
@@ -182,12 +185,80 @@ func (m *openAIModel) generateStream(ctx context.Context, params responses.Respo
...
+// that callers already received from the stream.
+func completedContentSupersedes(aggregate, completed *genai.Content) bool {
+	if completed == nil {
+		return false
+	}
+
+	var aggregateText, completedText strings.Builder
+	var aggregateCalls, completedCalls []*genai.FunctionCall
+	usable := false
+	for _, part := range aggregate.Parts {
```
karolpiotrowicz, line 220: `aggregate` is dereferenced here with no nil check while `completed` is guarded four lines above. Neither is reachable today: `aggregator.Close()` returns either nil or a response whose `Content` is always allocated, and this is the only call site, so `aggregate` is never nil and `completed` never is either. So the guard above is dead and the missing one is latent. Worth making the two symmetric — either check both, or drop both and state the precondition in the doc comment — because the guard that is present reads as evidence that nil-ness was considered on both arguments.

karolpiotrowicz (dismissed):
The four things I raised last round are fixed, and I checked each one by running it rather than by reading the diff. A reasoning-only terminal snapshot no longer destroys a streamed answer, a streamed function call is no longer erased by a message-only snapshot, the `FunctionCall` arm now has a test that discriminates it, and the merge base's own `TestModel_GenerateStream_FunctionCall`, lifted verbatim, passes against this head and fails against [1c6293a4](https://github.com/google/adk-go/commit/1c6293a416f4cab3d46ffa0bc261b39181ee16ab). The refusal shapes the PR exists to fix all still reach parity with the blocking path, and they now deliver the refusal mid-stream where `main` delivers nothing at all.

**What I am asking for before this merges is tests.** Five decisions inside `completedContentSupersedes` can each be changed to the wrong behavior with the package suite still green, and the predicate has no direct test anywhere in the tree. The one at [openai.go:232](https://github.com/google/adk-go/blob/d9c0d842748e824db903aa576064b6be3a2402e0/model/openaimodel/openai.go#L232) is the one I would not merge without, for the reason below. The rest are on their lines, and a single table test of the predicate would cover all five at once.

Beyond that there are two questions about the design that I think you are better placed to answer than I am, and neither needs code from you if the answer is that the current behavior is intended.

## The property to hold a fix against

`completedContentSupersedes` answers a binary question — replace the whole aggregate, or keep the whole aggregate. So on any input where the snapshot and the stream each hold something the other lacks, one of them is discarded. The contract you are working to wants the union of the two, which means merging rather than choosing. That single property explains every point below, and it is more useful to check a broader fix against than any individual case:

- **Retention is partial.** Both sums at [openai.go:221](https://github.com/google/adk-go/blob/d9c0d842748e824db903aa576064b6be3a2402e0/model/openaimodel/openai.go#L221) and [:232](https://github.com/google/adk-go/blob/d9c0d842748e824db903aa576064b6be3a2402e0/model/openaimodel/openai.go#L232) filter on `!part.Thought`, so visible text and function calls are protected and reasoning is protected by nothing.
- **Recovery is partial.** When the guard returns false at [openai.go:195](https://github.com/google/adk-go/blob/d9c0d842748e824db903aa576064b6be3a2402e0/model/openaimodel/openai.go#L195), every snapshot-only item goes with it, including a tool call that never streamed.

## The test I would want before merge

Delete `&& !part.Thought` at [openai.go:232](https://github.com/google/adk-go/blob/d9c0d842748e824db903aa576064b6be3a2402e0/model/openaimodel/openai.go#L232) and `go test ./model/openaimodel/` is still green, but a turn that streamed a visible answer now ends as reasoning only — which is exactly the failure the last round blocked on, and it is also the shape that makes `Flow.Run` re-call the model up to ten times. That line is doing the most important work in the change and nothing is holding it in place.

Your existing `reasoning-only snapshot does not replace an answer` case cannot catch it. Its reasoning text is `"Weighing options"`, which does not contain the streamed answer, so the substring check at line 241 fails whether the filter is there or not.

A fourth case in `TestModel_GenerateStream_NarrowerCompletedContentPreservesAggregate` does catch it. I ran this one — it passes as written, it fails with the filter removed, and the three existing cases in that table pass either way.

<details><summary>The case</summary>

```go
{
    name: "reasoning-only snapshot containing the answer does not replace it",
    events: []string{
        evCreated,
        evDelta1,
        evDelta2,
        `{"type":"response.completed","response":{"id":"resp_1","model":"stream-model","status":"completed","output":[{"id":"rs_1","type":"reasoning","status":"completed","summary":[],"content":[{"type":"reasoning_text","text":"hello is the answer"}]}]}}`,
    },
    want: genai.NewContentFromText("hello", genai.RoleModel),
},
```

</details>

## Does a streamed reasoning part have to survive the turn?

This is a genuine question about the contract rather than a defect I can prove, and I could not settle it from the code.

When the snapshot carries no reasoning item, the guard replaces the aggregate and the streamed thought is gone. Measured against `main`, with a stream of one reasoning delta plus `"hello"` and a terminal snapshot holding only the message:

| | final content |
|---|---|
| `main` | `[THOUGHT("Checking the request"), TEXT("hello")]` |
| this head | `[TEXT("hello")]` |
| blocking path, same body | `[TEXT("hello")]` |

So head matches the blocking path here and `main` does not, which is the parity the PR is built for. `TestModel_GenerateStream_CompletedFunctionCallReplacesThoughtOnlyAggregate` already asserts a final identical to blocking after a reasoning delta, so the choice looks deliberate — I would just like it stated, because the retention rule as written does not carve reasoning out and the next person to read `completedContentSupersedes` will not know which way it is meant to go. A comment on the function saying parity wins for reasoning parts would settle it.

Worth knowing before you decide: the thought does still reach the caller as a partial, and dropping it cannot trigger the thought-only re-call loop, because superseding requires a non-thought part to be present.

## Adding refusal deltas changed which path a refusal stream takes

This one is new, and it comes from the two halves of the PR interacting rather than from either alone.

Before this change there was no `response.refusal.delta` case in the translator, so refusal deltas fell through to the default and nothing reached the aggregator. A refusal-only stream therefore always took the `final == nil` rebuild and got the terminal snapshot unconditionally. [stream.go:53](https://github.com/google/adk-go/blob/d9c0d842748e824db903aa576064b6be3a2402e0/model/openaimodel/stream.go#L53) now puts those deltas into the aggregate, so the same stream goes through the new predicate instead:

| stream, then terminal snapshot | `main` | this head |
|---|---|---|
| refusal deltas `"I cannot help."`, snapshot's refusal is different text | snapshot's wording | `"I cannot help."` |
| refusal deltas `"I cannot help."`, snapshot is reasoning only | `[THOUGHT("Policy check failed")]` | `[TEXT("I cannot help.")]` |

I am not claiming head is worse for the user on either row — keeping the refusal the reader actually saw looks better than replacing it with a bare thought, and `main` only reaches parity there by throwing the whole stream away. The point is narrower: this is the one shape class where "the guard returned false" no longer means "behaves like `main`", nothing covers it, and it is easy to miss because the two changes look independent.

## Where the all-or-nothing choice costs the most

A stream that narrates before calling a tool loses the call, when the snapshot carries only the call:

```
response.created
response.output_text.delta   "Let me check the weather."
response.completed           output: [ function_call get_weather {"city":"SF"} ]
```

The visible text makes the aggregate's text non-empty, the snapshot has no text to contain it, the guard returns false, and the final response carries zero function calls. The final is the only response that reaches tool dispatch, so the agent says it is about to check the weather and then does nothing. With a reasoning delta in place of the visible text the same stream recovers the call correctly, which is what makes this specifically a consequence of the binary choice rather than a bug in the matching.

`main` behaves the same way, so nothing regresses here — but recovering terminal-only tool calls is one of the things this PR set out to do, and it currently only works when nothing visible streamed first.

## Smaller things

The remaining points are on their lines. Beyond the tests asked for above, the only one worth acting on in this PR is the missing log line at [openai.go:193](https://github.com/google/adk-go/blob/d9c0d842748e824db903aa576064b6be3a2402e0/model/openaimodel/openai.go#L193), and that is your call rather than a requirement.

One note for sequencing rather than for you: the logprobs of a discarded snapshot are still applied to the retained aggregate, which is true on `main` too and is untouched by this change. #1373 is working on finish reason and logprobs in this same function and is the natural place for it, and it currently conflicts with this branch on `openai.go`.

ktsoator:
Thanks for checking these cases. I added direct tests for the predicate and confirmed that all five mutations you listed now fail. The stream test also covers reasoning containing the visible answer.

I reproduced the `null` argument case and normalized it to an empty map in the stream decoder, matching blocking. The completed `pong` text is now retained.

For reasoning, the current choice is intentional: when replacement is allowed, the final content follows the completed response. Streamed thoughts still reach the caller as partials. I added a comment and tests for both reasoning delta types.

The ID fallback and the visible-text-then-terminal-only-call case remain unresolved. I have kept the binary replacement rule for this revision and documented those limits; I am not claiming general reconciliation here. Local build, race tests, lint and tidy checks pass in both modules.

model/openaimodel/openai.go
ktsoator: Added the case where the terminal reasoning contains `hello`, both in the
predicate table and through the full stream. Removing the thought filter now
fails both tests.

model/openaimodel/openai.go
ktsoator: Added both cases to the predicate table. Disabling the usable check or replacing
`Contains` with `HasPrefix` now fails the corresponding test.

model/openaimodel/openai.go
ktsoator: I reproduced the `null` case and normalized streamed null arguments to an empty
map, matching blocking. The test now retains `pong`. A name-only comparison also
fails the new argument and ID tests.

The ID fallback remains unresolved. The converted call no longer records whether
its ID came from `item_id`, so safely pairing it with the terminal call needs
that information retained. I have left ID matching strict for this revision;
this also overlaps the terminal-call handling in #1373.

model/openaimodel/openai.go
ktsoator: Added the two-identical-calls case. Removing the one-to-one check now fails it.

model/openaimodel/openai.go
ktsoator: Understood. I have still left logging out of this revision. Conversion failures
keep the aggregate, but the lack of a diagnostic remains a limitation.

model/openaimodel/openai.go
ktsoator: Agreed. I kept the exact text comparison, so whitespace differences can still
cause the aggregate to be retained.

model/openaimodel/openai.go
ktsoator: I left the guards unchanged in this revision. The only caller supplies allocated
content, so the asymmetry is still there but does not affect that path.

karolpiotrowicz (approved):
The test I asked for is in, and it holds the line it needs to hold. Deleting `&& !part.Thought` at [openai.go:235](https://github.com/google/adk-go/blob/937c4c7355399baaaf9e1e5d86e42b07aec6fda5/model/openaimodel/openai.go#L235) now fails three cases including the end-to-end one, and restoring the previous revision's test file makes that same deletion pass again — so it is the new tests doing the work rather than coverage that was already there. The other four decisions in the predicate are in the same position: each can still be flipped to the wrong behavior, and each now fails a case in [`TestCompletedContentSupersedes`](https://github.com/google/adk-go/blob/937c4c7355399baaaf9e1e5d86e42b07aec6fda5/model/openaimodel/openai_test.go#L588). Both reproductions from #1466 and #1476, run verbatim, fail on the merge base and pass here.

On the null arguments: you did not take the fix I suggested, and yours is the better of the two. I proposed relaxing the comparison inside `completedContentSupersedes`, which would have confined the change to the predicate and left the streamed call diverging from the blocking path everywhere else — including on the partial response the caller receives mid-stream, which the predicate never touches. Normalizing in [`emitFunctionCall`](https://github.com/google/adk-go/blob/937c4c7355399baaaf9e1e5d86e42b07aec6fda5/model/openaimodel/stream.go#L172-L175) instead makes the streamed call identical to the blocking one before anything compares them, and it leaves the strict `reflect.DeepEqual` in place, which matters because relaxing that to a name match lets a snapshot substitute different arguments for the streamed ones. I ran a stream that narrates, calls `ping` with `"arguments":"null"`, and receives a snapshot carrying that call plus `"pong"`: both the final response and the mid-stream partial now match blocking exactly, where on the previous revision the partial carried a nil map and `"pong"` was lost.

The doc comment at [openai.go:209-211](https://github.com/google/adk-go/blob/937c4c7355399baaaf9e1e5d86e42b07aec6fda5/model/openaimodel/openai.go#L209-L211) answers the reasoning question I could not settle from the code, and the test beside it pins the answer. That was the last thing I had open.

Nothing here blocks. Three things are still worth doing and none of them belongs in another round on this PR — the `ID` fallback you flagged as overlapping #1373, the missing log line when terminal conversion fails, and the whitespace-sensitive text comparison. I am glad to see those become issues so this can land.

One sequencing note rather than something for you to fix: the branch is three commits behind `main`, and it conflicts with #1373 and #1392 on `openai.go`, so whichever lands second will need a rebase.

ktsoator:
Thanks for the thorough review and for checking the reproductions. I’ll check for existing coverage and open follow-up issues for the remaining points, keeping the overlap with #1373 in mind. I’ll leave this PR focused and update the branch if needed before merging.

PR #1470, opened 2026-09-02
fix(cmd/launcher/web): defer telemetry shutdown and adjust startup logging (#1469)

### LINKED ISSUE #1469: cmd/launcher/web: defer telemetry shutdown on server startup failure and adjust startup logging
## 🔴 Required Information

**Describe the Bug:**
In `cmd/launcher/web/web.go`, `webLauncher.Run` initializes OpenTelemetry providers via `telemetry.InitAndSetGlobalOtelProviders`. However:
1. `telemetryService.Shutdown` is only called in the `ctx.Done()` arm. If `srv.ListenAndServe()` fails to start or bind (e.g. port conflict or socket failure), `Run` returns on the `errChan` arm without calling `telemetryService.Shutdown`, leaking the OTel providers and any background exporter goroutines.
2. Additionally, the `"Starting the web server: ..."` and `"Web servers starts on %s"` log lines print before telemetry initialization runs. If telemetry initialization fails or delays, a server URL is announced to the user before the socket is bound or before the doomed process exits.

In contrast, `consoleLauncher.Run` (`cmd/launcher/console/console.go:75-85`) immediately defers telemetry shutdown right after initialization.

**Steps to Reproduce:**
1. Start the web launcher on a port that is already bound by another process.
2. `srv.ListenAndServe()` fails immediately on startup and `Run` returns the error via `errChan`.
3. `telemetryService.Shutdown` is never invoked.

**Expected Behavior:**
1. Telemetry is cleanly shut down on all exit paths (e.g. by deferring `telemetryService.Shutdown` right after initialization, similar to `console.go`).
2. Server startup URL logs are emitted after telemetry initialization succeeds, immediately before starting the HTTP server.

**Observed Behavior:**
Telemetry shutdown is skipped when the HTTP server fails to start.

**Environment Details:**
- ADK Library Version: v2.x (main)
- OS: Linux / All
- Go Version: 1.26.6

**Model Information:**
- N/A

---

## 🟡 Optional Information

**Additional Context:**
Identified during review of PR #1351 by @karolpiotrowicz and @jjsasha63.

chill-czar:

**1. Link to an existing issue:**

- Closes: #1469
- Related: #1350, #1351

**Problem:**
In `cmd/launcher/web/web.go`, `webLauncher.Run` previously only called `telemetryService.Shutdown` in the `ctx.Done()` arm. If the HTTP server failed to start or bind (e.g. port already bound by another process or listener error), `Run` returned immediately on the `errChan` branch without shutting down telemetry providers, leaking background exporter routines and resources.
Additionally, the startup banner and sublauncher `UserMessage` calls were executed before telemetry was initialized, announcing the server URL on doomed processes that fail telemetry setup.

**Solution:**
- Immediately defer `telemetryService.Shutdown(shutdownCtx)` after successful telemetry initialization in `webLauncher.Run`, matching the pattern in `cmd/launcher/console/console.go:79-85`.
- Relocate startup banner logging and sublauncher `UserMessage` calls below telemetry initialization so that the URL is only announced once telemetry succeeds and the socket is about to be bound.
- In `case <-ctx.Done():`, only call `srv.Shutdown(shutdownCtx)`; the deferred telemetry shutdown executes automatically when `Run` returns, ensuring HTTP server listener termination precedes telemetry flushing.

### Testing Plan

**Unit Tests:**

Summary of passed tests:
- `TestRunShutsDownTelemetryWhenServerFailsToStart`: Verifies that telemetry is shut down when the HTTP server hits a port-binding collision.
- `TestRunDoesNotAnnounceURLWhenTelemetryInitFails`: Verifies that sublauncher `UserMessage` / URL announcements are suppressed when telemetry initialization fails.
- `TestRunDoesNotLeakListenerWhenTelemetryInitFails`: Existing regression test continues to pass.
- `TestH2CFlag`: Existing tests continue to pass.

```text
=== RUN   TestH2CFlag
--- PASS: TestH2CFlag (0.00s)
=== RUN   TestRunDoesNotLeakListenerWhenTelemetryInitFails
--- PASS: TestRunDoesNotLeakListenerWhenTelemetryInitFails (3.01s)
=== RUN   TestRunShutsDownTelemetryWhenServerFailsToStart
--- PASS: TestRunShutsDownTelemetryWhenServerFailsToStart (0.00s)
=== RUN   TestRunDoesNotAnnounceURLWhenTelemetryInitFails
--- PASS: TestRunDoesNotAnnounceURLWhenTelemetryInitFails (0.00s)
PASS
ok  	google.golang.org/adk/v2/cmd/launcher/web	4.043s
```

`golangci-lint run ./cmd/launcher/web/...`: 0 issues
`go mod tidy -diff`: clean

Verified offline port collision and telemetry init failure scenarios using `net.Listen` and custom OpenTelemetry span processor tracker.

### Checklist

### Additional context
Follow-up identified during review of PR #1351 by @karolpiotrowicz and @jjsasha63.

### DIFF
--- a/cmd/launcher/web/web_test.go
+++ b/cmd/launcher/web/web_test.go
@@ -26,11 +26,13 @@ import (
 	"net/http/httptest"
 	"os"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/gorilla/mux"
 	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 
 	"google.golang.org/adk/v2/artifact"
 	"google.golang.org/adk/v2/cmd/launcher"
@@ -477,5 +479,187 @@ func TestBuildBaseRouterLeavesHealthToTheCaller(t *testing.T) {
 	rec := serveWithoutPanic(t, router, httptest.NewRequest(http.MethodGet, "/health", nil))
 	if rec.Code != http.StatusTeapot {
 		t.Errorf("GET /health status = %d, want %d: the embedder's handler was shadowed", rec.Code, http.StatusTeapot)
+	}
+}
+
+type trackingSpanProcessor struct {
+	shutdownCalled atomic.Bool
+}
+
+func (p *trackingSpanProcessor) OnStart(context.Context, sdktrace.ReadWriteSpan) {}
+func (p *trackingSpanProcessor) OnEnd(sdktrace.ReadOnlySpan)                     {}
+func (p *trackingSpanProcessor) ForceFlush(context.Context) error                { return nil }
+func (p *trackingSpanProcessor) Shutdown(context.Context) error {
+	p.shutdownCalled.Store(true)
+	return nil
+}
+
+// TestRunShutsDownTelemetryWhenServerFailsToStart covers issue #1469:
+// when the HTTP server fails to start (e.g. port already bound),
+// Run must shut down the initialized OpenTelemetry providers.
+func TestRunShutsDownTelemetryWhenServerFailsToStart(t *testing.T) {
+	ln, err := net.Listen("tcp", ":0")
+	if err != nil {
+		t.Fatalf("net.Listen() failed: %v", err)
+	}
+	t.Cleanup(func() { _ = ln.Close() })
+	port := ln.Addr().(*net.TCPAddr).Port
+
+	l := NewLauncher(telemetryFailSublauncher{}).(*webLauncher)
+	if _, err := l.Parse([]string{"--port", fmt.Sprint(port), "repro"}); err != nil {
+		t.Fatalf("Parse() failed: %v", err)
+	}
+
+	tracker := &trackingSpanProcessor{}
+	config := &launcher.Config{
+		TelemetryOptions: []telemetry.Option{telemetry.WithSpanProcessors(tracker)},
+	}
+
+	if err := l.Run(t.Context(), config); err == nil {
+		t.Fatalf("Run() succeeded, want server bind failure")
+	}
+
+	if !tracker.shutdownCalled.Load() {
+		t.Errorf("telemetry shutdown was not called after server startup failure")
+	}
+}
+
+type failingSpanProcessor struct {
+	shutdownErr error
+}
+
+func (p *failingSpanProcessor) OnStart(context.Context, sdktrace.ReadWriteSpan) {}
+func (p *failingSpanProcessor) OnEnd(sdktrace.ReadOnlySpan)                     {}
+func (p *failingSpanProcessor) ForceFlush(context.Context) error                { return nil }
+func (p *failingSpanProcessor) Shutdown(context.Context) error {
+	return p.shutdownErr
+}
+
+// TestRunLogsWhenTelemetryShutdownFails covers the defer error branch:
+// when telemetry shutdown fails, the error is logged to stderr rather than
+// terminating or panicking.
+func TestRunLogsWhenTelemetryShutdownFails(t *testing.T) {
+	var buf bytes.Buffer
+	orig := log.Writer()
+	log.SetOutput(&buf)
+	t.Cleanup(func() { log.SetOutput(orig) })
+
+	ln, err := net.Listen("tcp", ":0")
+	if err != nil {
+		t.Fatalf("net.Listen() failed: %v", err)
+	}
+	t.Cleanup(func() { _ = ln.Close() })
+	port := ln.Addr().(*net.TCPAddr).Port
+
+	l := NewLauncher(telemetryFailSublauncher{}).(*webLauncher)
+	if _, err := l.Parse([]string{"--port", fmt.Sprint(port), "repro"}); err != nil {
+		t.Fatalf("Parse() failed: %v", err)
+	}
+
+	failingProcessor := &failingSpanProcessor{
+		shutdownErr: fmt.Errorf("simulated flush error"),
+	}
+	config := &launcher.Config{
+		TelemetryOptions: []telemetry.Option{telemetry.WithSpanProcessors(failingProcessor)},
+	}
+
+	if err := l.Run(t.Context(), config); err == nil {
+		t.Fatalf("Run() succeeded, want server bind failure")
+	}
+
+	if got := buf.String(); !strings.Contains(got, "telemetry shutdown failed: simulated flush error") {
+		t.Errorf("expected log output to contain telemetry shutdown error, got %q", got)
+	}
+}
+
+type trackingSublauncher struct {
+	telemetryFailSublauncher
+	userMessageCalled atomic.Bool
+}
+
+func (s *trackingSublauncher) UserMessage(webURL string, printer func(v ...any)) {
+	s.userMessageCalled.Store(true)
+}
+
+// TestRunDoesNotAnnounceURLWhenTelemetryInitFails covers issue #1469:
+// sublauncher UserMessage and URL announcements must only occur after telemetry
+// initialization succeeds.
+func TestRunDoesNotAnnounceURLWhenTelemetryInitFails(t *testing.T) {
+	var buf bytes.Buffer
+	orig := log.Writer()
+	log.SetOutput(&buf)
+	t.Cleanup(func() { log.SetOutput(orig) })
+
+	ln, err := net.Listen("tcp", ":0")
+	if err != nil {
+		t.Fatalf("net.Listen() failed: %v", err)
+	}
+	port := ln.Addr().(*net.TCPAddr).Port
+	if err := ln.Close(); err != nil {
+		t.Fatalf("listener Close() failed: %v", err)
+	}
+
+	sub := &trackingSublauncher{}
+	l := NewLauncher(sub).(*webLauncher)
+	if _, err := l.Parse([]string{"--port", fmt.Sprint(port), "repro"}); err != nil {
+		t.Fatalf("Parse() failed: %v", err)
+	}
+
+	bad := resource.NewWithAttributes("https://conflicting.invalid/schema/v1")
+	config := &launcher.Config{
+		TelemetryOptions: []telemetry.Option{telemetry.WithResource(bad)},
+	}
+
+	if err := l.Run(t.Context(), config); err == nil || !strings.Contains(err.Error(), "telemetry initialization failed") {
+		t.Fatalf("Run() error = %v, want error containing %q", err, "telemetry initialization failed")
+	}
+
+	if sub.userMessageCalled.Load() {
+		t.Errorf("UserMessage was called before telemetry initialization succeeded")
+	}
+	startingPrefix := strings.Split(logStartingWebServer, "%")[0]
+	startsOnPrefix := strings.Split(logWebServerStartsOn, "%")[0]
+	if got := buf.String(); strings.Contains(got, startingPrefix) || strings.Contains(got, startsOnPrefix) {
+		t.Errorf("startup banner was logged before telemetry initialization succeeded: %q", got)
+	}
+}
+
+type optionAppendingSublauncher struct {
+	telemetryFailSublauncher
+	processor *trackingSpanProcessor
+}
+
+func (s *optionAppendingSublauncher) Keyword() string { return "appending" }
+
+func (s *optionAppendingSublauncher) SetupSubrouters(r *mux.Router, c *launcher.Config) error {
+	c.TelemetryOptions = append(c.TelemetryOptions, telemetry.WithSpanProcessors(s.processor))
+	return nil
+}
+
+// TestRunSetupSubroutersCanAppendTelemetryOptions verifies the invariant that
+// SetupSubrouters runs before telemetry initialization, so that subrouters can
+// append telemetry options (e.g. span processors) that are picked up by Run.
+func TestRunSetupSubroutersCanAppendTelemetryOptions(t *testing.T) {
+	ln, err := net.Listen("tcp", ":0")
+	if err != nil {
+		t.Fatalf("net.Listen() failed: %v", err)
+	}
+	t.Cleanup(func() { _ = ln.Close() })
+	port := ln.Addr().(*net.TCPAddr).Port
+
+	tracker := &trackingSpanProcessor{}
+	sub := &optionAppendingSublauncher{processor: tracker}
+	l := NewLauncher(sub).(*webLauncher)
+	if _, err := l.Parse([]string{"--port", fmt.Sprint(port), "appending"}); err != nil {
+		t.Fatalf("Parse() failed: %v", err)
+	}
+
+	config := &launcher.Config{}
+	if err := l.Run(t.Context(), config); err == nil {
+		t.Fatalf("Run() succeeded, want server bind failure")
+	}
+
+	if !tracker.shutdownCalled.Load() {
+		t.Errorf("span processor appended in SetupSubrouters was not initialized/shut down")
 	}
 }
--- a/cmd/launcher/web/web.go
+++ b/cmd/launcher/web/web.go
@@ -17,7 +17,6 @@ package web
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"fmt"
 	"log"
@@ -34,6 +33,11 @@ import (
 	"google.golang.org/adk/v2/internal/cli/util"
 	"google.golang.org/adk/v2/memory"
 	"google.golang.org/adk/v2/session"
+)
+
+const (
+	logStartingWebServer = "Starting the web server: %+v"
+	logWebServerStartsOn = "Web servers starts on %s"
 )
 
 // webConfig contains parameters for launching web server
@@ -182,7 +186,8 @@ func applyServiceDefaults(config *launcher.Config) {
 	}
 }
 
-// Run implements launcher.SubLauncher.
+// Run implements launcher.SubLauncher. It takes ownership of the telemetry
+// providers initialized for execution and shuts them down on exit.
 func (w *webLauncher) Run(ctx context.Context, config *launcher.Config) error {
 	applyServiceDefaults(config)
 
@@ -207,19 +212,26 @@ func (w *webLauncher) Run(ctx context.Context, config *launcher.Config) error {
 		}
 	}
 
-	log.Printf("Starting the web server: %+v", w.config)
-	log.Println()
-	webUrl := fmt.Sprintf("http://localhost:%v", fmt.Sprint(w.config.port))
-	log.Printf("Web servers starts on %s", webUrl)
-	for _, l := range w.activeSublaunchers {
-		l.UserMessage(webUrl, log.Println)
-	}
-	log.Println()
-
 	telemetryService, err := telemetry.InitAndSetGlobalOtelProviders(ctx, config, w.config.otelToCloud)
 	if err != nil {
 		return fmt.Errorf("telemetry initialization failed: %v", err)
 	}
+	defer func() {
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), w.config.shutdownTimeout)
+		defer cancel()
+		if err := telemetryService.Shutdown(shutdownCtx); err != nil {
+			log.Printf("telemetry shutdown failed: %v", err)
+		}
+	}()
+
+	log.Printf(logStartingWebServer, w.config)
+	log.Println()
+	webUrl := fmt.Sprintf("http://localhost:%v", fmt.Sprint(w.config.port))
+	log.Printf(logWebServerStartsOn, webUrl)
+	for _, l := range w.activeSublaunchers {
+		l.UserMessage(webUrl, log.Println)
+	}
+	log.Println()
 
 	srv := w.buildHTTPServer(router)
 
@@ -236,9 +248,7 @@ func (w *webLauncher) Run(ctx context.Context, config *launcher.Config) error {
 		log.Println("Shutting down the web server...")
 		shutdownCtx, cancel := context.WithTimeout(context.Background(), w.config.shutdownTimeout)
 		defer cancel()
-		serverErr := srv.Shutdown(shutdownCtx)
-		telemetryErr := telemetryService.Shutdown(shutdownCtx)
-		return errors.Join(serverErr, telemetryErr)
+		return srv.Shutdown(shutdownCtx)
 	case err, ok := <-errChan:
 		if !ok {
 			return nil


cmd/launcher/web/web_test.go
```diff
@@ -181,3 +183,87 @@ func TestRunDoesNotLeakListenerWhenTelemetryInitFails(t *testing.T) {
...
+func (p *trackingSpanProcessor) Shutdown(context.Context) error {
+	p.shutdownCalled.Store(true)
+	return nil
+}
+
+// TestRunShutsDownTelemetryWhenServerFailsToStart covers issue #1469:
+// when the HTTP server fails to start (e.g. port already bound),
+// Run must shut down the initialized OpenTelemetry providers.
+func TestRunShutsDownTelemetryWhenServerFailsToStart(t *testing.T) {
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
```
karolpiotrowicz: This listener holds `127.0.0.1:<port>` while `buildHTTPServer` binds the wildcard `":<port>"`, so the two are not asking for the same address. On Linux that is still `EADDRINUSE` and the test does what it says — I ran it and it fails on `main`, passes here. It is not guaranteed everywhere: on the BSD family, where Go sets `SO_REUSEADDR` on listeners, a wildcard bind alongside a bound loopback address can be permitted. I had no macOS to check that on, and CI is `ubuntu-latest` only, so it would not surface there either. Binding this probe with `net.Listen("tcp", ":0")` makes the conflict exact everywhere and costs nothing.

cmd/launcher/web/web_test.go
```diff
@@ -181,3 +183,87 @@ func TestRunDoesNotLeakListenerWhenTelemetryInitFails(t *testing.T) {
...
+	if _, err := l.Parse([]string{"--port", fmt.Sprint(port), "repro"}); err != nil {
+		t.Fatalf("Parse() failed: %v", err)
+	}
+
+	tracker := &trackingSpanProcessor{}
+	config := &launcher.Config{
+		TelemetryOptions: []telemetry.Option{telemetry.WithSpanProcessors(tracker)},
+	}
+
+	if err := l.Run(context.Background(), config); err == nil {
```
karolpiotrowicz: `context.Background()` here means the only way out of `Run` is the bind failing. The `select` in `Run` has no other exit, so if `ListenAndServe` ever unexpectedly succeeds this does not fail — it blocks until the package's default 10-minute timeout panic and takes every other test in the binary with it. `t.Context()` or a short `context.WithTimeout` turns that into a clean `Run() succeeded, want server bind failure` in seconds. The same applies to the `Run` call in `TestRunDoesNotAnnounceURLWhenTelemetryInitFails`.

cmd/launcher/web/web_test.go
```diff
@@ -181,3 +183,87 @@ func TestRunDoesNotLeakListenerWhenTelemetryInitFails(t *testing.T) {
...
 }
+
+type trackingSpanProcessor struct {
+	shutdownCalled atomic.Bool
+}
+
+func (p *trackingSpanProcessor) OnStart(context.Context, sdktrace.ReadWriteSpan) {}
+func (p *trackingSpanProcessor) OnEnd(sdktrace.ReadOnlySpan)                     {}
+func (p *trackingSpanProcessor) ForceFlush(context.Context) error                { return nil }
+func (p *trackingSpanProcessor) Shutdown(context.Context) error {
```
karolpiotrowicz, line 492: This returns nil unconditionally, so the error branch the PR adds — the `log.Printf("telemetry shutdown failed: %v", err)` inside the new defer — is executed by no test in the repository. A second processor whose `Shutdown` returns an error would cover it, and would also pin down what `Run` returns when the flush fails, which is the behavior that changed here.

cmd/launcher/web/web_test.go
```diff
@@ -181,3 +183,87 @@ func TestRunDoesNotLeakListenerWhenTelemetryInitFails(t *testing.T) {
...
+	if _, err := l.Parse([]string{"--port", fmt.Sprint(port), "repro"}); err != nil {
+		t.Fatalf("Parse() failed: %v", err)
+	}
+
+	bad := resource.NewWithAttributes("https://conflicting.invalid/schema/v1")
+	config := &launcher.Config{
+		TelemetryOptions: []telemetry.Option{telemetry.WithResource(bad)},
+	}
+
+	if err := l.Run(context.Background(), config); err == nil {
```
karolpiotrowicz: This asserts only that `Run` failed, not that it failed for the reason the test is named after. Any early return added before telemetry initialization — a config check, a router validation — would keep this green while proving nothing about announcement ordering. Matching on `"telemetry initialization failed"` pins it to the intended path.

karolpiotrowicz (commented):
The deferred shutdown closes the leak this was filed for, and the two new tests are real regression tests — I ran them against `main`'s `web.go` and both fail there for the right reasons, then pass with the change applied. The guard test from #1350, `TestRunDoesNotLeakListenerWhenTelemetryInitFails`, still passes, which was the thing most at risk from moving initialization earlier. Nothing below blocks.

One merge precondition first: the branch conflicts with `main`. The conflict is confined to `web_test.go`, where your tests and the ones added to `main` both append at the end of the file, so a rebase should be a matter of keeping both. `web.go` merges cleanly. Everything I checked below was measured on that merge result rather than on the branch as it stands.

## Telemetry providers are now torn down on paths where the caller's process keeps running

This is the one consequence worth a deliberate decision rather than a nit. Moving the shutdown into a function-scoped `defer` means it also fires on the `errChan` arm — a bind failure — and not only on context cancellation. That is exactly what "shut down on all exit paths" asks for, so this is not a mistake. It does change what an embedder sees, because `cmd/launcher/web` is importable and `telemetry.WithTracerProvider` hands the launcher an instance the caller built.

An `sdktrace.TracerProvider` is one-shot. After the launcher shuts it down, the caller's provider is permanently non-recording, with no error and no log line to say so. A caller that treats a bind failure as retryable, or that runs other instrumented components in the same process, loses all subsequent telemetry silently. The same holds for the process-global provider that `SetGlobalOtelProviders` installs, so it is not limited to embedders who pass their own.

I measured both, driving `Run` directly on each side. On `main` the caller's provider survives a bind failure and still records. With this change it does not:

```text
                                        main        this PR
caller's provider shut down by Run      false       true
spans recorded after Run returned       1           0
process-global provider shut down       false       true
spans recorded by the global after Run  1           0
```

No caller in this repository is affected — `adk web` exits through `log.Fatalf` — so nothing here is broken today. A sentence on `Run` or on `launcher.Config.TelemetryOptions` saying the launcher takes ownership of the telemetry providers and shuts them down on every exit would settle it for anyone embedding the package.

## The shutdown error no longer reaches the caller, and the shutdown budget is now spent twice

Two consequences of one decision, so treating them together. Dropping `errors.Join(serverErr, telemetryErr)` in favour of logging inside the defer matches `console.go`, which is the pattern #1469 named, so the direction is right. It does mean a telemetry flush failure during graceful shutdown no longer changes what `Run` returns, and since that error reaches `log.Fatalf` in the CLI, the process exit status moves from 1 to 0. I confirmed it: on `main` `Run` returns the flush error, with this change it returns nil and the failure survives only as a log line.

Separately, `srv.Shutdown` and the telemetry flush now each get a fresh `context.WithTimeout(context.Background(), shutdownTimeout)` instead of sharing one. With `--shutdown-timeout=2s` and a request parked in a slow handler, `Run` returned after 2s on `main` and after 4s with this change, so worst-case graceful shutdown is now two full timeouts rather than one. That matters where the grace period is externally imposed — a Kubernetes `terminationGracePeriodSeconds` of 30 against the 15s default leaves no headroom.

The same measurement argues the other way too, and it is the more interesting half. On `main` the telemetry processor's `Shutdown` was **never called** in that scenario: `srv.Shutdown` had consumed the shared deadline, and the SDK returns early on an already-expired context. So the old single budget was quietly skipping the flush in exactly the case where you would most want it. Spending more wall time to actually flush looks like the better trade. Worth saying so in the PR description, or deriving both contexts from one deadline computed before the `select` if the total is meant to stay bounded — either is defensible, but the change to `--shutdown-timeout`'s meaning should be intentional.

## The port-conflict case in the issue still announces a URL

Expected Behavior 2 in #1469 asks for the logs "after telemetry initialization succeeds, immediately before starting the HTTP server", and that is satisfied. The motivating complaint was broader, though — "a server URL is announced to the user before the socket is bound" — and that part is unchanged, because `ListenAndServe` runs asynchronously after the banner. Your own test shows it. In the passing run of `TestRunShutsDownTelemetryWhenServerFailsToStart`, with the port already held by another listener and the server never starting:

```text
=== RUN   TestRunShutsDownTelemetryWhenServerFailsToStart
Starting the web server: &{port:36183 ...}
Web servers starts on http://localhost:36183
--- PASS: TestRunShutsDownTelemetryWhenServerFailsToStart (0.00s)
```

That is the issue's own Steps to Reproduce, so a user hitting it still sees a URL for a server that never comes up. Closing that would mean binding with `net.Listen` first and announcing after a successful bind, then handing the listener to `srv.Serve` — a bigger change than this PR, and it overlaps #1456. No need to do it here. It is worth saying on #1469 that this part remains, so the issue is not closed on the assumption that the announcement problem is fully gone.

## Not this PR, but it bounds what the fix buys

The web launcher installs no signal handler. `cmd/internal/adkcli/main.go` builds a `context.Background()` and nothing cancels it, so on the ordinary way to stop `adk web` — Ctrl-C or SIGTERM — the process dies without running any deferred function, including this new one. `console.go` wraps its context in `signal.NotifyContext`, which is why the console launcher does not have this problem. That predates your change and is not yours to fix here, but the deferred shutdown will not run on the most common exit until something cancels that context. Probably worth a follow-up issue.

I have left a few smaller test notes on the lines themselves.

Pinned references, all at [ebffc4c](https://github.com/google/adk-go/commit/ebffc4c21117004b146c5efb105bbe8f4c37cd95) for this branch and [440e352](https://github.com/google/adk-go/commit/440e35202c325a7310557e3bacecdeb3ad07b989) for `main`: the new defer at [`web.go:183-189`](https://github.com/google/adk-go/blob/ebffc4c21117004b146c5efb105bbe8f4c37cd95/cmd/launcher/web/web.go#L183-L189), the changed return at [`web.go:211-215`](https://github.com/google/adk-go/blob/ebffc4c21117004b146c5efb105bbe8f4c37cd95/cmd/launcher/web/web.go#L211-L215), the banner and `UserMessage` loop at [`web.go:191-197`](https://github.com/google/adk-go/blob/ebffc4c21117004b146c5efb105bbe8f4c37cd95/cmd/launcher/web/web.go#L191-L197), the `ListenAndServe` goroutine at [`web.go:202-206`](https://github.com/google/adk-go/blob/ebffc4c21117004b146c5efb105bbe8f4c37cd95/cmd/launcher/web/web.go#L202-L206), the CLI's context and its `log.Fatalf` at [`main.go:109`](https://github.com/google/adk-go/blob/440e35202c325a7310557e3bacecdeb3ad07b989/cmd/internal/adkcli/main.go#L109) and [`main.go:127-128`](https://github.com/google/adk-go/blob/440e35202c325a7310557e3bacecdeb3ad07b989/cmd/internal/adkcli/main.go#L127-L128), and the console launcher's signal handling at [`console.go:72`](https://github.com/google/adk-go/blob/440e35202c325a7310557e3bacecdeb3ad07b989/cmd/launcher/console/console.go#L72) with its deferred shutdown at [`console.go:79-85`](https://github.com/google/adk-go/blob/440e35202c325a7310557e3bacecdeb3ad07b989/cmd/launcher/console/console.go#L79-L85).

cmd/launcher/web/web_test.go
```diff
@@ -181,3 +183,87 @@ func TestRunDoesNotLeakListenerWhenTelemetryInitFails(t *testing.T) {
...
+		t.Fatalf("Run() succeeded, want telemetry initialization failure")
+	}
+
+	if sub.userMessageCalled.Load() {
```
karolpiotrowicz, line 617: The check covers the sublauncher hook but not the two log lines the issue actually quotes, `"Starting the web server"` and `"Web servers starts on"`. Today that is sufficient, since the loop and the URL log sit in the same straight-line block with no branch between them. It stops being sufficient the moment they are separated, which is the regression this test exists to catch. `TestApplyServiceDefaultsLogsWhatItDefaulted` in this file already has the `log.SetOutput(&buf)` idiom to reuse.

cmd/launcher/web/web_test.go
```diff
@@ -181,3 +183,87 @@ func TestRunDoesNotLeakListenerWhenTelemetryInitFails(t *testing.T) {
...
+		t.Errorf("telemetry shutdown was not called after server startup failure")
+	}
+}
+
+type trackingSublauncher struct {
+	telemetryFailSublauncher
+	userMessageCalled atomic.Bool
+}
+
+func (t *trackingSublauncher) UserMessage(webURL string, printer func(v ...any)) {
```
karolpiotrowicz: Naming the receiver `t` reads as `*testing.T` in a file where every other `t` is exactly that, so `t.userMessageCalled.Store(true)` looks at a glance like a field on the test handle. `s` or `sub` would avoid the double take.

chill-czar:
Thank you for the thorough and insightful review, @karolpiotrowicz!

Here is how each point has been addressed:

1. **Rebase & Conflict Resolution**: Rebased onto latest `main` (`440e352`). The conflict in `web_test.go` has been resolved by preserving both the newly added web UI tests (#1460) and our regression tests. The PR is now `mergeable: true`.
2. **Telemetry Provider Ownership**: Added a doc comment to `webLauncher.Run` explicitly noting that `Run` takes ownership of the initialized telemetry providers and shuts them down on exit:
   ```go
   // Run implements launcher.SubLauncher. It takes ownership of the telemetry
   // providers initialized for execution and shuts them down on exit.
   ```
3. **Dedicated Shutdown Budget & Error Visibility**: Agreed with the trade-off. Giving telemetry its own dedicated timeout budget ensures background exporters are not silently dropped when `srv.Shutdown` consumes the grace period.
4. **URL Announcement on Port Collision**: Agreed that suppressing the banner on port bind conflicts is best addressed by binding the listener before starting the server (`net.Listen` -> `srv.Serve`), which naturally overlaps with #1456 rather than expanding this PR.
5. **Signal Handling**: Agreed; adding `signal.NotifyContext` to the web CLI launcher (`cmd/internal/adkcli/main.go`) to ensure defers fire on SIGINT/SIGTERM is a great candidate for a follow-up PR.
6. **Test Receiver Rename**: Renamed the receiver from `(t *trackingSublauncher)` to `(s *trackingSublauncher)` to avoid confusion with `*testing.T`.
7. **Test Log Assertions**: Updated `TestRunDoesNotAnnounceURLWhenTelemetryInitFails` to capture logs via `log.SetOutput(&buf)` (with `t.Cleanup` restoring the original writer) and asserted that neither `"Starting the web server"` nor `"Web servers starts on"` appears in the buffer.

All local tests, race checks, and linters pass cleanly.

cmd/launcher/web/web_test.go
```diff
@@ -479,3 +481,95 @@ func TestBuildBaseRouterLeavesHealthToTheCaller(t *testing.T) {
...
+	}
+
+	if err := l.Run(context.Background(), config); err == nil {
+		t.Fatalf("Run() succeeded, want telemetry initialization failure")
+	}
+
+	if sub.userMessageCalled.Load() {
+		t.Errorf("UserMessage was called before telemetry initialization succeeded")
+	}
+	if got := buf.String(); strings.Contains(got, "Starting the web server") || strings.Contains(got, "Web servers starts on") {
```
karolpiotrowicz: These two strings are copies of the ones at [`web.go:222`](https://github.com/google/adk-go/blob/b424bcd3796428017de781894ae2a6c0b796c7d6/cmd/launcher/web/web.go#L222) and [`web.go:225`](https://github.com/google/adk-go/blob/b424bcd3796428017de781894ae2a6c0b796c7d6/cmd/launcher/web/web.go#L225), with nothing tying them together. I renamed both in the production code and ran the package: it stays green, so this half of the assertion goes quiet the moment someone rewords the banner. The `userMessageCalled` check above still catches a wholesale reorder, but not the narrower regression of moving just these two `log.Printf` lines back above telemetry initialization, which is the one the test is named for. Sharing the format strings as package constants would tie them together. Not blocking.

karolpiotrowicz (dismissed):
The rebase, the ownership comment on `Run`, the receiver rename and the log assertions all landed, and I re-ran everything against the new head. The change itself is unchanged apart from that comment, and it still holds up: both new tests fail on `main`'s `web.go` and pass here, the guard test for #1350 still passes, and no gate found a defect attributable to this PR. Nothing below blocks. One merge precondition: the branch is behind `main` now rather than conflicting, so it needs a refresh before it can land.

The new banner assertion does what it was meant to. Running the new tests against `main`'s `web.go`, it fires with the captured text, so a future change that moves the two log lines back above telemetry initialization gets caught:

```text
--- FAIL: TestRunDoesNotAnnounceURLWhenTelemetryInitFails
    web_test.go:570: UserMessage was called before telemetry initialization succeeded
    web_test.go:573: startup banner was logged before telemetry initialization succeeded:
        "... Starting the web server: &{port:40099 ...}\n... Web servers starts on http://localhost:40099\n"
```

## The shutdown ordering, since that is what this change really is

I measured it on both trees rather than reading it, driving `Run` with `--shutdown-timeout=2s` and one request parked in a handler that outlives the budget:

```text
                                   main       this PR
telemetry Shutdown called          false      true
telemetry shutdown ctx.Err()       n/a        nil, 1999 ms left
```

Two things worth stating plainly, because they are the part a green test proves least. The ordering is guaranteed by Go's defer semantics rather than by timing: the telemetry defer is registered at [`web.go:214`](https://github.com/google/adk-go/blob/b424bcd3796428017de781894ae2a6c0b796c7d6/cmd/launcher/web/web.go#L214) and the `ctx.Done()` arm registers its `cancel` at [`web.go:245`](https://github.com/google/adk-go/blob/b424bcd3796428017de781894ae2a6c0b796c7d6/cmd/launcher/web/web.go#L245), so LIFO evaluates `srv.Shutdown` first and the flush always follows a completed or timed-out server shutdown, on a context of its own. And the failure path is clean: the defer is registered only after initialization returns a nil error, so it never runs against a half-built telemetry stack. I confirmed that on both trees — with initialization forced to fail, the processor's `Shutdown` is not called at all.

What is still not flushed is anything a handler emits after `srv.Shutdown` gave up on it, and anything at all on Ctrl-C or SIGTERM, since nothing cancels the CLI's context. Both predate this change, and on the first one `main` is worse rather than better: it never reached the flush in that scenario.

## Four of the notes I left are still open

None of them blocks, and your reply doesn't mention them, so flagging rather than repeating — the threads are still attached to the right lines:

- the probe listener binding loopback while `buildHTTPServer` binds the wildcard, [`web_test.go:501`](https://github.com/google/adk-go/pull/1470#discussion_r3931748048)
- `context.Background()` in the `Run` call, [`web_test.go:518`](https://github.com/google/adk-go/pull/1470#discussion_r3931748052)
- `trackingSpanProcessor.Shutdown` returning nil unconditionally, so the new defer's `log.Printf("telemetry shutdown failed: %v", err)` branch is executed by no test, [`web_test.go:492`](https://github.com/google/adk-go/pull/1470#discussion_r3931748053)
- the announce test asserting only that `Run` failed rather than that it failed for the named reason, [`web_test.go:565`](https://github.com/google/adk-go/pull/1470#discussion_r3931748059)

One correction to that second one, which I got wrong. I wrote that it applied equally to the `Run` call in `TestRunDoesNotAnnounceURLWhenTelemetryInitFails`. It does not: that test closes its listener and forces telemetry initialization to fail, so `Run` returns before it reaches the `select` and there is no path to a hang. Only [`web_test.go:518`](https://github.com/google/adk-go/blob/b424bcd3796428017de781894ae2a6c0b796c7d6/cmd/launcher/web/web_test.go#L518) is exposed.

I also checked whether the reordering left `web.go` correct on the point that most worried me, telemetry initialization staying below the `SetupSubrouters` loop so that the span and log processors `api.go` appends are still picked up. It does. Nothing pins it, though — hoisting the initialization above the loop leaves the whole package green, so if you wanted one more test, that invariant is the one with the quietest failure mode.

chill-czar:
Thanks again for the deep follow-up and for catching these details, @karolpiotrowicz!

The branch has been refreshed and rebased onto latest `main` (`49224b6f`). All 4 open review items, the constant unification, and the invariant test have been addressed:

1. **Rebase**: Rebased onto `main` (`49224b6f`).
2. **Wildcard Collision Probe (`:0`)**: Updated probe listeners in tests to bind `net.Listen("tcp", ":0")` instead of loopback, guaranteeing an exact wildcard-on-wildcard `EADDRINUSE` collision across all platforms.
3. **Bounded Test Contexts (`t.Context()`)**: Replaced `context.Background()` with `t.Context()` on all `Run` calls in the test suite to eliminate any risk of unbounded hangs.
4. **Telemetry Shutdown Error Logging Test**: Added `TestRunLogsWhenTelemetryShutdownFails` using a `failingSpanProcessor` to exercise the deferred `log.Printf("telemetry shutdown failed: %v", err)` branch and verify that flush errors are logged cleanly without panicking.
5. **Pinning Telemetry Init Failure Error**: Updated `TestRunDoesNotAnnounceURLWhenTelemetryInitFails` to explicitly assert that `Run` returns an error containing `"telemetry initialization failed"`.
6. **Unified Banner Format Constants**: Defined unexported constants `logStartingWebServer` and `logWebServerStartsOn` in `web.go` and tied `TestRunDoesNotAnnounceURLWhenTelemetryInitFails`'s log assertions to their prefixes, preventing test drift if the wording is refactored.
7. **Subrouter Telemetry Option Invariant Test**: Added `TestRunSetupSubroutersCanAppendTelemetryOptions` to lock down the invariant that `SetupSubrouters` runs before telemetry initialization and its appended options are initialized and shut down by `Run`.

All unit tests pass cleanly with `-race` and `golangci-lint` reports 0 issues.

cmd/launcher/web/web_test.go
```diff
@@ -480,3 +482,185 @@ func TestBuildBaseRouterLeavesHealthToTheCaller(t *testing.T) {
...
+	l := NewLauncher(telemetryFailSublauncher{}).(*webLauncher)
+	if _, err := l.Parse([]string{"--port", fmt.Sprint(port), "repro"}); err != nil {
+		t.Fatalf("Parse() failed: %v", err)
+	}
+
+	tracker := &trackingSpanProcessor{}
+	config := &launcher.Config{
+		TelemetryOptions: []telemetry.Option{telemetry.WithSpanProcessors(tracker)},
+	}
+
```
karolpiotrowicz, line 518: `t.Context()` reads better than `context.Background()`, so keep it — but it does not bound this. `T.Context` is cancelled just before `Cleanup`-registered functions run, that is, once the test function has returned. A body blocked inside `Run` has not returned, so the `ctx.Done()` arm at [`web.go:247`](https://github.com/google/adk-go/blob/631251f26b37abf96b13331db4c7ecde3c6fcd0f/cmd/launcher/web/web.go#L247) can never fire, and an unexpected bind success still wedges the package until `go test`'s watchdog panics the whole binary.

I measured it on a copy of this test with one change — the probe listener released before `Run`, so the bind succeeds. Everything else identical, `go test -timeout 45s`:

```text
l.Run(t.Context(), config)                                panic: test timed out after 45s
l.Run(context.Background(), config)                       panic: test timed out after 45s
ctx, cancel := context.WithTimeout(t.Context(), 2*time.Second)
l.Run(ctx, config)                                        --- PASS (2.00s)
```

So a deadline is what turns an unmet precondition into a named failure instead of a hang. Same applies at lines 566 and 658. Non-blocking, and the shape predates this change — the #1350 guard at [`web_test.go:181`](https://github.com/google/adk-go/blob/631251f26b37abf96b13331db4c7ecde3c6fcd0f/cmd/launcher/web/web_test.go#L181) has it too — so it is equally fine to leave this and note it.

cmd/launcher/web/web_test.go
```diff
@@ -480,3 +482,185 @@ func TestBuildBaseRouterLeavesHealthToTheCaller(t *testing.T) {
...
+	l := NewLauncher(sub).(*webLauncher)
+	if _, err := l.Parse([]string{"--port", fmt.Sprint(port), "appending"}); err != nil {
+		t.Fatalf("Parse() failed: %v", err)
+	}
+
+	config := &launcher.Config{}
+	if err := l.Run(t.Context(), config); err == nil {
+		t.Fatalf("Run() succeeded, want server bind failure")
+	}
+
```
karolpiotrowicz, line 662: This assertion does discriminate — I confirmed it catches telemetry initialization moving above the `SetupSubrouters` loop, which is the invariant the test is named for. The message is what I would change. `shutdownCalled` is false in two different situations: the option was appended too late to be picked up, or it was picked up and never disposed of. On the pre-change `web.go` this test fails for the second reason, while the hoisting perturbation makes it fail for the first, and "was not initialized/shut down" leaves a future maintainer to work out which. A second flag set from `OnStart` would separate registration from disposal. Non-blocking.

karolpiotrowicz (approved):
Nothing here blocks. Three of the four items from the last round are closed, verified by running them rather than by reading the reply, and the fourth is a nit that the change does not actually move.

**Closed.** The deferred shutdown's error branch is now genuinely exercised: deleting the `log.Printf` at [`web.go:223`](https://github.com/google/adk-go/blob/631251f26b37abf96b13331db4c7ecde3c6fcd0f/cmd/launcher/web/web.go#L223) makes `TestRunLogsWhenTelemetryShutdownFails` fail. The announce test now pins the error: rewording the message at [`web.go:217`](https://github.com/google/adk-go/blob/631251f26b37abf96b13331db4c7ecde3c6fcd0f/cmd/launcher/web/web.go#L217) makes `TestRunDoesNotAnnounceURLWhenTelemetryInitFails` fail. And `TestRunSetupSubroutersCanAppendTelemetryOptions` earns its name — hoisting telemetry initialization above the `SetupSubrouters` loop makes it fail, where the same perturbation slipped through the suite before you added it. All four of the new tests still fail against the pre-change `web.go` and pass with it.

The `:0` change is right. The server binds the wildcard at [`web.go:262`](https://github.com/google/adk-go/blob/631251f26b37abf96b13331db4c7ecde3c6fcd0f/cmd/launcher/web/web.go#L262), so a wildcard probe gives an exact collision where the loopback one relied on platform behaviour. Worth saying that I could only run this on Linux, so the platform that motivated the point is still untested by me — if you have a Mac to hand, a single `go test ./cmd/launcher/web/` would settle it.

The `t.Context()` substitution is the one that does not do what the comment says. It is fine to keep, but it does not bound the hang: the context is cancelled just before cleanup functions run, which is after the test body returns, and a body blocked inside `Run` never returns. I measured all three options on a copy of your test with the probe listener released so the bind succeeds — the "unexpected bind success" case the point was about. Details inline.

Two small things, both non-blocking, both inline.

Everything else I checked is clean on the merged tree: build, the full suite against `main` with zero new failures, `go vet`, `staticcheck`, `govulncheck`, `apidiff`, `golangci-lint`, `-race` over three shuffle seeds, and a 40-seed shuffle sweep at one and at sixty-four processors.

One housekeeping note: the branch is behind `main` and four other open pull requests touch `cmd/launcher/web/web.go` (#1600, #1457, #1456, #974), so whoever lands second will need a rebase.

Issue #1352, opened 2026-08-17
mcptoolset: successful tool executions with empty text response are erroneously treated as errors


chill-czar:
## 🔴 Required Information

**Describe the Bug:**
In `tool/mcptoolset/tool.go:167-169`, `mcpTool.Run` checks:
```go
if textResponse.Len() == 0 {
    return nil, errors.New("no text content in tool response")
}
```
When an MCP tool execution succeeds (`isError == false`) but returns an empty string or empty text content list (e.g. side-effecting tools like clearing cache, sending a notification, or logging), `mcpTool.Run` erroneously rejects the result with `errors.New("no text content in tool response")`.

Under the Model Context Protocol (MCP) specification, empty text content with `isError: false` is a valid tool execution result.

**Steps to Reproduce:**
1. Connect an MCP tool that returns an empty text content (e.g. `&mcp.CallToolResult{Content: []mcp.Content{&mcp.TextContent{Text: ""}}}`).
2. Invoke the tool via `mcpTool.Run`.
3. Notice that `mcpTool.Run` returns an error: `no text content in tool response`.

**Expected Behavior:**
The tool execution succeeds and returns `map[string]any{"output": ""}`.

**Observed Behavior:**
`mcpTool.Run` returns `errors.New("no text content in tool response")`.

**Environment Details:**
- ADK Library Version: v2.x (main at 3a474df)
- OS: Linux
- Go Version: 1.26.6

**Model Information:**
- N/A (MCP Toolset execution logic)

---

## 🟡 Optional Information

**How often has this issue occurred?:**
- Always (100%) when an MCP tool returns an empty text response without error.

chill-czar:
Hi @indurireddy-TF, could you please take a look at this issue when you have a moment? I have identified the root cause and submitted a fix with unit tests in PR #1353. Thank you!

indurireddy-TF:
Hi @chill-czar Thanks for reporting this issue. I was able to successfully reproduce the problem using a test case where the MCP tool returns an empty `TextContent`. The test fails with the expected error:`no text content in tool response`
This confirms that the issue occurs when a valid MCP tool response contains empty text. I’ll take a look at the associated PR and validate the proposed fix.

PR #1473, opened 2026-09-02
fix(mcptoolset): report non-text-only tool results instead of dropping them

jjsasha63:

Follow-up to #1353, which removed the `no text content in tool response` guard so that an empty text result is a valid success (#1352). That guard was also the only thing catching results that carry no text because all their content is non-text.

Adjacent to #1391 but does not close it — see below.

### Problem

`mcpTool.Run` collects text by skipping every block that is not a `*mcp.TextContent`. With the guard gone, a tool result holding only an image, audio, resource link or embedded resource accumulates no text and returns `{"output": ""}` with no error. The call is reported as successful and the entire payload is discarded.

Four of the five `mcp.Content` types are affected. Before #1353 these failed loudly; they now fail silently, which is the worse of the two.

### Solution

Track whether a non-text block was skipped, and return an error when that happened and no text was collected. A result with no content, or with only empty text blocks, still succeeds with `{"output": ""}`, so the #1352 fix is preserved. A non-text block alongside real text also still succeeds, unchanged.

This is deliberately a stopgap. #1401 renders non-text blocks properly and removes this branch when it lands; the guard and the test both carry a `TODO(#1401)`.

### What this does not fix

#1391 covers the mixed shape as well: a file body returned as an `EmbeddedResource` next to a status-line `TextContent`. That still drops the resource silently, before and after this change, because text was collected. Fixing it means rendering the blocks, which is #1401's job. This PR only restores the signal for the all-non-text case.

### Effect on callers

In the LLM flow the error becomes a `{"error": ...}` function response and the turn continues, so the model is told the content was dropped instead of being handed an empty string.

An MCP tool can also sit in a workflow graph via `workflow.NewToolNode`, where the error aborts the node rather than continuing. That is a harder failure than before, and intentional: the node was previously completing on empty data.

### Testing Plan

- [x] Unit test added: `TestMCPTool_NonTextOnlyResponse` in `tool/mcptoolset/set_test.go`, covering image, audio, resource link and embedded resource, plus a mixed text-and-image control that must still succeed.
- [x] All four non-text-only subtests fail on current `main` and pass with this change.
- [x] Mutation-checked both halves of the new condition. Dropping either conjunct, or removing the guard entirely, fails the suite.
- [x] `TestMCPTool_EmptyTextResponse` from #1353 still passes, so empty text remains a valid success.
- [x] Verification commands pass:
  ```text
  go test -race -count=1 ./tool/mcptoolset/...
  go vet ./tool/mcptoolset/...
  gofmt -l tool/mcptoolset/
  ```

### Checklist

### DIFF
--- a/tool/mcptoolset/set_test.go
+++ b/tool/mcptoolset/set_test.go
@@ -849,21 +849,55 @@ func TestMCPTool_EmptyTextResponse(t *testing.T) {
 }
 
 // TestMCPTool_NonTextOnlyResponse covers the gap left by #1353: an empty text
-// result is a valid success, but only when the server sent nothing. A result
-// carrying just an image, audio, resource link or embedded resource also
-// accumulates no text, and reporting that as {"output": ""} hands the model an
-// empty map while discarding the entire payload.
+// result is a valid success, but only when there was no other content to lose.
+// A result carrying just an image, audio, resource link or embedded resource
+// also accumulates no text, and reporting that as {"output": ""} hands the
+// model an empty string while discarding the entire payload.
+//
+// The mixed case is the control. A non-text block alongside real text must
+// still succeed, which pins the textResponse.Len() == 0 half of the condition:
+// without it, every result carrying a non-text block would error.
+//
+// TODO(#1401): the mixed case still drops its non-text block silently (#1391).
+// When #1401 renders those blocks, this whole test goes away with the guard.
 func TestMCPTool_NonTextOnlyResponse(t *testing.T) {
 	tests := []struct {
-		name    string
-		content mcp.Content
+		name       string
+		content    []mcp.Content
+		wantErr    bool
+		wantOutput string
 	}{
-		{"image", &mcp.ImageContent{MIMEType: "image/png", Data: []byte{0x89, 0x50, 0x4e, 0x47}}},
-		{"audio", &mcp.AudioContent{MIMEType: "audio/wav", Data: []byte{0x52, 0x49, 0x46, 0x46}}},
-		{"resource link", &mcp.ResourceLink{URI: "file:///tmp/report.pdf", Name: "report.pdf"}},
-		{"embedded resource", &mcp.EmbeddedResource{
-			Resource: &mcp.ResourceContents{URI: "file:///tmp/data.bin", MIMEType: "application/octet-stream", Blob: []byte{0x00, 0x01}},
-		}},
+		{
+			name:    "image only",
+			content: []mcp.Content{&mcp.ImageContent{MIMEType: "image/png", Data: []byte{0x89, 0x50, 0x4e, 0x47}}},
+			wantErr: true,
+		},
+		{
+			name:    "audio only",
+			content: []mcp.Content{&mcp.AudioContent{MIMEType: "audio/wav", Data: []byte{0x52, 0x49, 0x46, 0x46}}},
+			wantErr: true,
+		},
+		{
+			name:    "resource link only",
+			content: []mcp.Content{&mcp.ResourceLink{URI: "file:///tmp/report.pdf", Name: "report.pdf"}},
+			wantErr: true,
+		},
+		{
+			name: "embedded resource only",
+			content: []mcp.Content{&mcp.EmbeddedResource{
+				Resource: &mcp.ResourceContents{URI: "file:///tmp/data.bin", MIMEType: "application/octet-stream", Blob: []byte{0x00, 0x01}},
+			}},
+			wantErr: true,
+		},
+		{
+			name: "text alongside non-text still succeeds",
+			content: []mcp.Content{
+				&mcp.TextContent{Text: "caption"},
+				&mcp.ImageContent{MIMEType: "image/png", Data: []byte{0x89, 0x50, 0x4e, 0x47}},
+			},
+			wantErr:    false,
+			wantOutput: "caption",
+		},
 	}
 
 	for _, tt := range tests {
@@ -871,9 +905,9 @@ func TestMCPTool_NonTextOnlyResponse(t *testing.T) {
 			clientTransport, serverTransport := mcp.NewInMemoryTransports()
 
 			server := mcp.NewServer(&mcp.Implementation{Name: "test_server", Version: "v1.0.0"}, nil)
-			mcp.AddTool(server, &mcp.Tool{Name: "non_text_tool", Description: "returns only non-text content"}, func(ctx context.Context, req *mcp.CallToolRequest, args any) (*mcp.CallToolResult, any, error) {
+			mcp.AddTool(server, &mcp.Tool{Name: "non_text_tool", Description: "returns non-text content"}, func(ctx context.Context, req *mcp.CallToolRequest, args any) (*mcp.CallToolResult, any, error) {
 				return &mcp.CallToolResult{
-					Content: []mcp.Content{tt.content},
+					Content: tt.content,
 				}, nil, nil
 			})
 			if _, err := server.Connect(t.Context(), serverTransport, nil); err != nil {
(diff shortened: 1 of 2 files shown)


karolpiotrowicz (commented):
The fix is correct and the red→green holds: all four subtests of `TestMCPTool_NonTextOnlyResponse` fail on `main` ([ffddb2a6](https://github.com/google/adk-go/commit/ffddb2a6d7ed97974e82815d00c95eb148038f2c)) with `got output=""`, and pass with the PR applied ([c02efcd3](https://github.com/google/adk-go/commit/c02efcd3935f774b948ffe58d762a2fba888c46c)). `TestMCPTool_EmptyTextResponse` still passes on both, so the #1352 behavior survives. No new test failures anywhere in the repo, and no API change. I ran the eight content shapes the branch could see against both trees, and exactly two move: no-text-plus-non-text, and non-text-only. That is precisely the scope you describe.

Nothing here blocks. One test gap is worth closing before merge, and there is a sequencing question below that is for the maintainers rather than for you.

## The condition is half-tested

Replacing the guard with `if droppedNonText {` — dropping the `textResponse.Len() == 0` conjunct entirely — leaves the package suite green:

```
$ go test -count=1 ./tool/mcptoolset/
ok      google.golang.org/adk/v2/tool/mcptoolset        0.244s
```

That variant errors on *every* result carrying a non-text block, so a tool returning `[TextContent{"Here is your chart:"}, ImageContent{...}]` would go from a successful `{"output": "Here is your chart:"}` to a hard error. Nothing in the suite notices. The other direction is pinned — reverting to `if textResponse.Len() == 0 {` is caught by `TestMCPTool_EmptyTextResponse` — so the suite holds one side of the new condition and not the side this PR adds. A single extra row in the table at [set_test.go#L857](https://github.com/google/adk-go/blob/c02efcd3935f774b948ffe58d762a2fba888c46c/tool/mcptoolset/set_test.go#L857) for mixed text-plus-image, asserting `err == nil` and `output == "caption"`, closes it.

## A second path where the new error fires

The description reasons about what the model sees, which is right for the LLM flow: [base_flow.go#L1344-L1345](https://github.com/google/adk-go/blob/c02efcd3935f774b948ffe58d762a2fba888c46c/internal/llminternal/base_flow.go#L1344-L1345) turns the error into a `{"error": ...}` function response and the turn continues.

There is a second caller. `workflow.NewToolNode` accepts any `tool.Tool`, so an MCP tool can sit in a workflow graph, and there [tool_node.go#L118-L120](https://github.com/google/adk-go/blob/c02efcd3935f774b948ffe58d762a2fba888c46c/workflow/tool_node.go#L118-L120) wraps the error and [scheduler.go#L465-L467](https://github.com/google/adk-go/blob/c02efcd3935f774b948ffe58d762a2fba888c46c/workflow/scheduler.go#L465-L467) aborts the node. A graph that previously ran to completion on an image-only result now stops. I think aborting is defensible — the node was completing with silently empty data — but it is a harder failure than the one the description describes, and worth a sentence so nobody is surprised.

## Sequencing, for the maintainers rather than for this PR

#1391 reports the case that actually bites users: GitHub's MCP server returns file bodies as an `EmbeddedResource` next to a status-line `TextContent`, so the agent sees only "successfully downloaded text file (SHA: …)". That mixed shape still drops silently, both before and after this change, so this PR does not address #1391 — which #1391 itself anticipates by calling the non-text-only case adjacent but distinct.

Worth knowing that the other ADK implementations have already split on this exact question. adk-java errors on a non-text-only result, much as this PR does ([AbstractMcpTool.java#L147-L153](https://github.com/google/adk-java/blob/9a903e66d25ef578b20515aa79c13bb85b1fa092/core/src/main/java/com/google/adk/tools/mcp/AbstractMcpTool.java#L147-L153)). adk-python and adk-kotlin both pass the whole result through untouched, so non-text blocks reach the model verbatim and nothing errors ([mcp_tool.py#L565](https://github.com/google/adk-python/blob/d637d1b449aa4382045f71ee84d3f674403d41d5/src/google/adk/tools/mcp_tool/mcp_tool.py#L565), [McpTool.kt#L81](https://github.com/google/adk-kotlin/blob/2618ab0f094d0f1a9b1dd11771ed59b39b38f1b7/core/src/jvmMain/kotlin/com/google/adk/kt/tools/mcp/McpTool.kt#L81)). So this PR is not an outlier, and the long-term direction is the pass-through one.

#1401 is the PR that gets there, but it is a draft with merge conflicts, and its branch is cut against a base predating #1353 — it still carries `return nil, errors.New("no text content in tool response")`, the exact check #1353 removed, so it would currently reintroduce #1352. It needs a rebase before it can land at all. That is the strongest argument for taking this stopgap now rather than waiting.

## Smaller things

- The error is a bare `errors.New` at [tool.go#L173](https://github.com/google/adk-go/blob/c02efcd3935f774b948ffe58d762a2fba888c46c/tool/mcptoolset/tool.go#L173), while every sibling in the same function names the tool, e.g. [tool.go#L125](https://github.com/google/adk-go/blob/c02efcd3935f774b948ffe58d762a2fba888c46c/tool/mcptoolset/tool.go#L125). adk-java hit this same case and its message carries both the tool name and the discarded blocks: [AbstractMcpTool.java#L147-L153](https://github.com/google/adk-java/blob/9a903e66d25ef578b20515aa79c13bb85b1fa092/core/src/main/java/com/google/adk/tools/mcp/AbstractMcpTool.java#L147-L153) returns `{"error": "Tool '<name>' returned content that is not TextContent.", "content_details": contents.toString()}`. Worth considering — telling the model *what* was dropped costs one extra key and is more actionable than the current message.
- The comment at [tool.go#L169-L170](https://github.com/google/adk-go/blob/c02efcd3935f774b948ffe58d762a2fba888c46c/tool/mcptoolset/tool.go#L169-L170) says the empty result is valid "only when the server actually sent nothing", but the implemented rule is "no text bytes and at least one skipped non-text block". A lone `TextContent{Text: ""}` succeeds even though the server did send a block.
- The assertion at [set_test.go#L912](https://github.com/google/adk-go/blob/c02efcd3935f774b948ffe58d762a2fba888c46c/tool/mcptoolset/set_test.go#L912) only checks `err == nil`, so it would also pass on a transport failure or on the pre-#1353 error. Pinning the error identity would make it specific.
- The description says the model receives "an empty map". It receives `{"output": ""}`, which matters to anyone reasoning about what actually reaches the model.
- Nothing in the package marks this as temporary. I grafted this PR's test onto #1401 and all four subtests fail there, so whoever lands #1401 gets a red test with no explanation. A `TODO(#1401)` on both the guard and the test would carry the intent forward.

karolpiotrowicz (approved):
Re-reviewed at [963a22ce](https://github.com/google/adk-go/commit/963a22ce8081e29b78edc3120fe502cf2831b799). Everything from the last round is fixed, and I confirmed the test gap specifically rather than eyeballing it: the mutant that used to leave the suite green — dropping the `textResponse.Len() == 0` conjunct — now fails on `text alongside non-text still succeeds`. I also checked that reverting to the pre-#1353 guard, removing the guard entirely, and stripping the tool name from the error each fail the suite now. Full run is clean: 77 packages, no new failures, `go vet` and `apidiff` both clean, and the increment changes no runtime behavior beyond the error string.

One thing left, and it is the same defect in two halves.

The comment at [tool.go#L171](https://github.com/google/adk-go/blob/963a22ce8081e29b78edc3120fe502cf2831b799/tool/mcptoolset/tool.go#L171) says the error fires when "the text is empty because every block was non-text". The predicate counts *bytes*, not blocks, so `[TextContent{""}, ImageContent{...}]` errors even though a text block was present. And nothing pins that distinction: writing the guard the way the comment describes keeps the entire suite green.

```go
// leaves `go test ./tool/mcptoolset/` green
sawTextBlock := false
// ... set sawTextBlock = true where the TextContent assertion succeeds ...
if !sawTextBlock && droppedNonText {
```

Under that version, `[TextContent{""}, ImageContent{...}]` goes back to returning `{"output": ""}` with no error — the silent whole-payload discard this PR exists to stop. It matters here more than ordinary comment drift would, because that comment is the only in-code description of a branch the `TODO(#1401)` beside it tells someone else to delete, so it is what they will read as the spec. A row for that shape with `wantErr: true`, plus wording the comment in bytes, closes both halves.

Smaller, none of them worth holding for:

- The message "returned only non-text content" is inaccurate for that same shape, where a text block was present but empty.
- [set_test.go#L859](https://github.com/google/adk-go/blob/963a22ce8081e29b78edc3120fe502cf2831b799/tool/mcptoolset/set_test.go#L859) — "every result carrying a non-text block would error" is a bit broad. One carrying `StructuredContent` still succeeds, and an `IsError` result returns "Tool execution failed." instead.
- `non_text_tool` appears as a bare literal in both the `AddTool` call and the expected-error string 54 lines apart, so renaming the fixture fails with a message that points at the error text rather than the rename.

I would merge this. The remaining row is small enough to fold in now or leave as a follow-up, and `main` is dropping payloads silently until this lands.

PR #1474, opened 2026-09-02
fix: map response finish reasons to semantic conventions

### LINKED ISSUE #1472: telemetry: gen_ai.response.finish_reasons reports raw genai values, not the semconv enum
## 🔴 Required Information

**Describe the Bug:**

`gen_ai.response.finish_reasons` is set to the raw `genai.FinishReason`
constant, which is the protobuf wire name. The GenAI semantic conventions
expect the `stop` / `length` / `content_filter` / `tool_call` / `error` enum,
so consumers get `["STOP"]` where they expect `["stop"]`.

`internal/telemetry/telemetry.go:132`:

```go
semconv.GenAIResponseFinishReasons(string(params.Response.FinishReason))
```

`genai.FinishReasonStop` is `"STOP"`, `genai.FinishReasonMaxTokens` is
`"MAX_TOKENS"`, and an unset finish reason emits `[""]`.

The conventions also say a generation that failed should report `error`
rather than being omitted or passed through:

> If a finish reason was expected but not received, for example because
> generation failed, was cancelled, or a stream ended before the final event,
> instrumentations SHOULD report `error` for that position instead of omitting
> it.

**Steps to Reproduce:**

1. Run any agent with a tracing exporter attached.
2. Read the `generate_content` span.
3. Look at `gen_ai.response.finish_reasons`.

**Expected Behavior:**

`["stop"]` for a completed turn, `["tool_call"]` for a turn that is a tool
call, `["length"]` when truncated, `["error"]` when the response carries an
error code or was interrupted.

**Observed Behavior:**

`["STOP"]`, `["MAX_TOKENS"]`, `["MALFORMED_FUNCTION_CALL"]`, or `[""]`.

**Environment Details:**

- ADK Library Version: 2.3.0
- OS: Linux
- Go Version: go1.26.6

**Model Information:**

- Any Gemini model. The value comes from `genai`, so it is model-independent.

---

## 🟡 Optional Information

**Regression:** No. This has been the behaviour since the move to semconv.

**Additional Context:**

The mapping this needs already exists in the same package.
`schemaFinishReason` in `internal/telemetry/genaimessages.go:343` produces
exactly these values, including the error rule above, and it is currently
wired only to the per-message `finish_reason` field on
`gen_ai.output.messages`.

Upstream deprecated that per-message field in
[semantic-conventions-genai#363](https://github.com/open-telemetry/semantic-conventions-genai/pull/363),
which landed on 1 September, and its description now reads:

> Deprecated: report finish reasons in `gen_ai.response.finish_reasons`
> instead.

So the careful mapping is on the field being retired, and the field it is
being redirected to has the raw value. Fixing this is reusing
`schemaFinishReason` at the call site above, plus a test.

mikemikimike:
## Summary

Fixes google/adk-go#1472 by mapping `gen_ai.response.finish_reasons` through the existing semantic-convention finish-reason mapper. This prevents protobuf wire names such as `STOP` and `MAX_TOKENS` from being emitted and preserves the existing successful-completion, error, length, content-filter, and tool-call rules.

This moves finish-reason mapping from the deprecated per-message `finish_reason` field to `gen_ai.response.finish_reasons`, the response-level attribute consumed by semantic-convention consumers.

## Changes

- Pass the generate-content error into the finish-reason mapper so partial responses terminated by an error emit `error` instead of claiming a clean `stop`.
- Reuse the existing converted-part tool-call detection rather than duplicating raw response inspection.
- Align the safety regression coverage with the converter's actual `ErrorCode: "SAFETY"` behavior.
- Restore the usage metadata fixture comment and remove an unrelated formatting-only change.

## Testing

Passed:

- `go test -mod=readonly -count=1 ./internal/telemetry`
- `docker run --rm -e CGO_ENABLED=1 ... go test -race -mod=readonly -count=1 -run 'TestTraceGenerateContentResult_MapsFinishReason|TestSchemaFinishReason' ./internal/telemetry`
- `go vet ./internal/telemetry/...`
- `gofmt -d` on all changed Go files
- `git diff --check`

The repository-wide test command `go test -mod=readonly -count=1 -shuffle=on work` still reports pre-existing Windows copyright-path, telemetry fixture-order, and runner compaction failures. The repository-wide race run reports the same existing copyright-path failure; the changed telemetry packages pass. Local `golangci-lint run` reports pre-existing gofumpt findings in files outside this change, and `go mod tidy -diff` reports only existing `go.sum` line-ending differences; no dependency changes were made.

## Compatibility

No public API, dependency, or persisted-data changes are introduced. Successful completions, tool calls, length limits, and content-filter mappings retain their existing semantic-convention values; errored responses now report `error` consistently.

## Issue

Fixes google/adk-go#1472


jjsasha63 (commented):
The mapping does what #1472 asks. I reverted the one-line production change on this branch and got 5 failures in the new subtests plus `TestGenerateContent/Success` and 4 fixture subtests across both telemetry packages, so the tests are exercising the fix. `go test ./internal/telemetry/...` and `go vet` are clean as submitted.

Five things before it goes in.

### 1. An errored span now claims a clean stop

`params.Error` never reaches the mapper. For a stream that yields a partial chunk and then an error, the span comes out as:

```
status=Error  finish_reasons=["stop"]
```

On main that read `[""]`. The convention text quoted in #1472 covers exactly this case: "a stream ended before the final event ... SHOULD report `error`". Either pass `params.Error` into `schemaFinishReason`, or map an unset finish reason to `error` rather than `stop`.

### 2. The content_filter case cannot happen in production

`TestTraceGenerateContentResult_MapsFinishReason` builds `&model.LLMResponse{FinishReason: genai.FinishReasonSafety}` directly. A real safety-blocked candidate carries no parts, so `internal/llminternal/converters/converters.go:40` sets `ErrorCode = "SAFETY"`, and the `resp.ErrorCode != ""` check at `genaimessages.go:347` short-circuits ahead of the safety branch. Running the converter and the mapper end to end:

```
ErrorCode="SAFETY" FinishReason="SAFETY" -> schemaFinishReason="error"
```

This predates the PR, but the new test asserts a value no user will see.

### 3. responseHasToolCall duplicates hasToolCall

`hasToolCall` at `genaimessages.go:299` answers the same question from the converted parts, and AGENTS.md:100 asks for the existing helper. Deriving the flag inside `schemaFinishReason` instead drops the new function, the `toolCall` parameter and the type switch. I tried it: 11 insertions against 23 deletions, tests still green.

### 4. Restore the UsageMetadata comment

`telemetrytestcase/agent_with_tool.go:40` lost "MockModel returns no UsageMetadata, so no token attrs are set". The digest is compared with `cmp.Diff` over the whole attribute map, so that absence is an assertion, and nothing else in the repo explains why the token attributes are missing.

### 5. Drop the StartExecuteToolSpan reformat

`telemetry.go` was gofmt-clean on main, so the closing-paren move is unrelated to the fix (AGENTS.md:172).

### Parity with adk-python 2.8.0

`src/google/adk/telemetry/_experimental_semconv.py` omits the attribute entirely when the finish reason is unset (`_build_response_common_attributes`, line 596). It has no `tool_call` case, maps `UNSPECIFIED` and `OTHER` to `error`, and emits `safety` where we emit `content_filter` (`_to_finish_reason`, line 254).

Go is closer to the semconv enum on all four, so I would leave them as they are. But this PR moves them off the deprecated per-message field and onto the attribute everyone reads, so please add a line about that to the description.

jjsasha63 (changes requested):
.

mikemikimike:
Addressed the review feedback in 65d7aff1b7a44593c8d16585c4fccf1e04ec7987: propagated generate-content errors into finish-reason mapping, reused the existing converted-part tool-call helper, replaced the unreachable safety test with converter-realistic coverage, restored the UsageMetadata fixture comment, and removed unrelated formatting noise. The PR description now documents the move to the response-level gen_ai.response.finish_reasons attribute. All GitHub Actions checks are passing.

jjsasha63 (changes requested):
All five are addressed, and the tests, `go vet` and `gofmt` are clean. One problem, and it came out of my finding 3, so it is on me.

### Deriving the tool-call flag through semconvParts is expensive

`genaimessages.go:348` converts the parts to answer a boolean. That conversion base64-encodes inline data up to `maxInlineDataBytes` and JSON-marshals tool arguments, and everything but the bool is discarded. It runs on every `generate_content` span, including when content capture is off, and twice when it is on, since `responseContentAttributes` has already converted the same parts.

A response carrying a 16 KB image and one tool call, capture off:

| Revision | ns/op | B/op | allocs/op |
| --- | --- | --- | --- |
| main | 234 | 160 | 3 |
| f266c05c | 242 | 160 | 3 |
| 65d7aff1 | 39,676 | 49,667 | 12 |
| with the patch below | 301 | 160 | 3 |

A text-only response costs 380 ns/op, so it is the base64 and the marshal rather than the walk.

What I meant by one derivation was the cheap one, not the converted one. Reading the genai parts directly also lets the `ErrorCode` short-circuit run before the walk:

```go
	if resp.ErrorCode != "" || resp.Interrupted {
		return finishError
	}
	switch resp.FinishReason {
	case "", genai.FinishReasonUnspecified, genai.FinishReasonStop:
		if hasFunctionCall(resp.Content) {
			return finishToolCall
		}
		return finishStop
```

```go
// hasFunctionCall reports whether content carries a tool call. It reads the
// genai parts directly rather than converting them: conversion base64-encodes
// inline data and marshals tool arguments, and this runs on every span whether
// content capture is on or off.
func hasFunctionCall(c *genai.Content) bool {
	if c == nil {
		return false
	}
	for _, p := range c.Parts {
		if p != nil && p.FunctionCall != nil {
			return true
		}
	}
	return false
}
```

`hasToolCall([]any)` has no callers left after that. 11 insertions against 7 deletions; it builds, gofmt is clean, and the telemetry and functional-digest tests pass.

### Two smaller ones

The doc comment at `genaimessages.go:332` covers only the response-derived rules and does not mention `err`, which now outranks all of them.

The stream-failure case from #1472 is still open. When the model yields `(nil, err)`, `telemetry.go:126` returns before the attribute is set, so it comes out absent rather than `["error"]`. That predates this PR, but `params.Error` is plumbed now, so it is a couple of lines.

mikemikimike:
Addressed the remaining review feedback in ea456976: finish-reason mapping now detects tool calls directly from genai parts to avoid unnecessary conversion allocations, the stream-error path is covered for both partial and nil responses, the mapper documentation includes error precedence, and unrelated formatting/comment changes were corrected. Targeted telemetry/llminternal tests pass. Please re-review.

jjsasha63 (changes requested):
The allocation regression is gone. Same 16 KB image and tool call with capture off: 394 ns/op and 160 B/op, against 39,676 ns/op and 49,667 B/op on 65d7aff1. Main is 234 ns/op. Three mutations are killed too — forcing `hasFunctionCall` false kills 8 subtests, dropping `err != nil` from the guard kills 2, and moving the attribute back below the nil check kills 4.

Moving it above that check went further than I asked for, and it brought one thing with it.

### A span that succeeded now reports error

`telemetry.go:126` also catches the case where response and error are both nil:

```
status=Unset  finish_reasons=["error"]
```

The span says nothing went wrong and the attribute says it did, which is the mirror of the case in my first comment. On main the attribute was absent there.

It is reachable in the flow tests. I planted a panic on the nil-and-nil state and ran them:

```
--- FAIL: TestFlowRunPartialLastEvent/trailing_partial_from_sub-agent_completes_without_error
panic: NIL_NIL_REACHED
```

A request processor short-circuits, the model is never called, and the deferred `endSpanAndTrackResult` fires with nothing to report. Emitting only when there is a response or an error keeps what you fixed and drops this.

### Three smaller ones

`telemetry_test.go:337` has no case for a nil response with a nil error. The mutation above dies on a panic I planted rather than on an assertion, so whatever value you settle on, pin it.

`telemetry.go:126` could use a line saying why the attribute is set above the guard. The reason is good, but as it stands it reads like the line drifted there.

`telemetry.go:130-132` is a three-line `SetAttributes` around a single attribute, left over from the move. The package writes those on one line: `compaction.go:182`, `188`, `239`, `251` and `258`.

### Noted, not asking for

`content_filter` is still unreachable for an ordinary safety block, because `converters.go:39` sets `ErrorCode` whenever a candidate has no parts and the guard short-circuits on it. I raised that in the first comment and you adjusted the test instead; the behavior predates this PR either way.

`telemetry_test.go:377` is named "stream error without response", but nothing streams in it.

mikemikimike:
Addressed the remaining telemetry review points and pushed commit `c901b748654d9bc36124233dad0402376f5bbf1d`.

- `schemaFinishReason` reads `genai.Content.Parts` directly to detect function calls, avoiding `semconvParts` conversion on every `generate_content` span.
- The result attribute is recorded only when a response or error exists, so nil response + nil error does not emit `finish_reasons=["error"]`.
- Added a direct test proving a non-nil error takes precedence over a successful `STOP` finish reason.
- `gofmt` and `git diff --check` passed.
- `go test -count=1 ./internal/telemetry/...` and `go vet ./internal/telemetry/...` were blocked before compilation because the local module cache lacks `google.golang.org/genai@v1.63.0` metadata and the proxy returned `EOF`. These commands used the locally installed Go 1.26.6 toolchain; no toolchain auto-download was used.

Please re-review.

jjsasha63 (approved):
Approving. The mapping is complete and the tests hold it down.

`gen_ai.response.finish_reasons` now carries the semantic-convention value on
every path I could drive. A span with neither a response nor an error carries
no finish reason at all, and keeps status `Unset`. That is what #1472 asked
for, and it invents no result for a call that never happened.

The new tests are load-bearing. I mutated the production code and counted what
went red:

- Removing the `params.Response != nil || params.Error != nil` guard at
  `telemetry.go:127` fails 2. They are
  `TestTraceGenerateContentResult_MapsFinishReason/without_response_or_error`
  and `TestTraceGenerateContentResult_NilResponseNilErrorLeavesSpanSuccessful`.
- Dropping `|| params.Error != nil` from that guard fails 4 subtests.
- Dropping `err != nil` from `schemaFinishReason` fails 2.
- Forcing `hasFunctionCall` to return false fails 5.

The branch is 39 commits behind `main`. Main has since touched the same
function, to fold `ToolUsePromptTokenCount` into input tokens. I merged
`origin/main` into the head on a scratch branch. It merges without conflict,
`go build ./...` passes, and `go test ./internal/telemetry/...` passes on the
merged tree. Main's `ToolUsePromptTokensCountAsInput` case still passes,
because `wantAttrs` is compared as a subset rather than an exact map.

`gofmt -l internal/telemetry/` is empty and every CI check is green.

### Parity with adk-python 2.9.2

`_to_finish_reason` in `src/google/adk/telemetry/_experimental_semconv.py:256`
maps `STOP` to `stop`, `MAX_TOKENS` to `length` and `OTHER` to `error`. It
lowercases everything else. `_build_response_common_attributes:599` omits the
attribute when that mapper returns `''`. It returns `''` for an unset or
`FINISH_REASON_UNSPECIFIED` reason, via `is_reported_finish_reason` in
`_finish_reason.py:24`.

- Matches: `STOP` to `stop`, `MAX_TOKENS` to `length`, `OTHER` to `error`.
  Both set the attribute on the span and on the per-message field.
- Diverges. Go emits `stop` for an unset or `UNSPECIFIED` reason, where Python
  omits the attribute. Go emits `content_filter` for a safety block, where
  Python emits `safety`. Go adds a `tool_call` case Python does not have. Go
  folds `err`, `ErrorCode` and `Interrupted` into `error`, and Python telemetry
  reads none of those. Each one puts Go closer to the semconv enum, so I would
  keep them.
- Not ported: nothing.

### One thing for a separate issue

`genai.FinishReason` is a plain `string` alias. So the `default` branch at
`genaimessages.go:378` lowercases and emits whatever the server sent, with no
length bound. That is unchanged here: the merge-base code emitted the
same string raw. This PR shrinks the set of values that reach it. Happy to
file it separately if you want it bounded.

PR #1479, opened 2026-09-03
fix(runner): stop nil event loops after cancellation

### LINKED ISSUE #1471: runner: a nil-yielding agent makes Run and RunLive uninterruptible
## 🔴 Required Information

**Describe the Bug:**

#1417 added `if event == nil { continue }` to `Runner.Run` (`runner/runner.go:721`) and `Runner.RunLive` (`:939`). The guard skips the event without calling `yield`, so the consumer's loop body is never entered. A consumer cannot `break` out, because `break` works by `yield` returning false and `yield` is never called. Neither loop consults `ctx` either.

So an agent that yields nils in a loop spins forever inside `Run`, ignoring both context cancellation and the consumer's `break`.

**Steps to Reproduce:**

1. Save the test below as `runner/nilflood_test.go`.
2. Run `go test ./runner/ -run TestRun_NilFloodIsInterruptible -timeout 60s`.
3. It fails after three seconds.

```go
func TestRun_NilFloodIsInterruptible(t *testing.T) {
	var yielded atomic.Int64
		Name: "nil_flood",
		Run: func(agent.InvocationContext) iter.Seq2[*session.Event, error] {
				for {
					yielded.Add(1)
					if !yield(nil, nil) {
						return
					}
				}
			}
		},
	})
	if err != nil {
		t.Fatalf("agent.New: %v", err)
	}
	r, err := New(Config{
		AppName:           "app",
		Agent:             a,
		SessionService:    session.InMemoryService(),
		AutoCreateSession: true,
	})
	if err != nil {
		t.Fatalf("New: %v", err)
	}

	ctx, cancel := context.WithCancel(context.Background())
	cancel() // cancelled before Run is called

	var bodyRuns atomic.Int64
	done := make(chan struct{})
	go func() {
		defer close(done)
		msg := &genai.Content{Parts: []*genai.Part{{Text: "hi"}}}
		for range r.Run(ctx, "u", "s", msg, agent.RunConfig{}) {
			bodyRuns.Add(1)
			break
		}
	}()

	select {
	case <-done:
		t.Fatalf("Run did not return 3s after ctx cancellation: consumer body ran %d times, agent yielded %d nils",
			bodyRuns.Load(), yielded.Load())
	}
}
```

**Expected Behavior:**

`Run` returns when the context is cancelled, and the consumer gets at least one turn so its `break` can take effect.

**Observed Behavior:**

```
--- FAIL: TestRun_NilFloodIsInterruptible (3.00s)
    nilflood_test.go:60: Run did not return 3s after ctx cancellation: consumer body ran 0 times, agent yielded 160306865 nils
```

The context was already cancelled before `Run` was called. `Run` still absorbed 160 million nils in three seconds and never gave the consumer a single iteration.

**Environment Details:**

- ADK Library Version: `main` at ffddb2a6
- OS: Linux
- Go Version: go1.26.6

---

## 🟡 Optional Information

**Regression:**

No. Before #1417 the same agent panicked on its first nil, so nothing that used to work has stopped working. The trigger is a compile-time agent supplied by the application author rather than anything a caller can send, so severity is low. It is worth fixing because it is an unbounded loop in a `ctx`-taking library function that ignores `ctx`.

**Suggested fix:**

This is contained in the lines #1417 already touched. It fixes the repro above and leaves `go test ./...` green.

```diff
 			if event == nil {
+				log.Printf("adk: agent %q yielded a nil event; skipping it", r.rootAgent.Name())
+				if err := ctx.Err(); err != nil {
+					yield(nil, err)
+					return
+				}
 				continue
 			}
```

The same three lines apply at `:939` in `RunLive`, with `agentToRun.Name()` in place of `r.rootAgent.Name()`.

The log line is separate from the hang. Today the skip is silent, so an agent author whose agent yields nil sees a run with zero events, no error and nothing in the log. `runner.go` reports anomalies of this kind in six other places, the nearest at `:650`.

**Two test gaps from the same change:**

- `TestRunner_NilEventYieldedDoesNotPanic` does not pin what happens after a nil. Changing `:722` from `continue` to `break`, and separately to `return`, leaves `./runner/` green. In production either would silently truncate the rest of the run. Yielding a real event after the nil and asserting it arrives closes this.
- The `RunLive` guard at `:939` has no coverage. Deleting those three lines outright also leaves `./runner/` green. `live_runner_test.go` already has the harness for one case.

wmh123wmh1:
## Summary

The nil-event guards added in #1417 prevent a nil-pointer panic, but a nil-yielding agent could still make `Runner.Run` and `Runner.RunLive` spin indefinitely because the runner silently skipped every nil event without calling `yield`.

This change treats a nil event as an agent contract violation in both runner paths:

- log the agent name and the nil-event error;
- yield the error to the consumer so it gets an iteration and can exit;
- terminate the current run immediately.

This prevents nil-event floods from spinning indefinitely even when the context remains active, and avoids relying on a context-cancellation check whose behavior depended on whether a nil event happened to arrive first. Nil events are no longer silently skipped.

No public API changes are introduced.

## Testing

- `go test ./runner/... -count=1`
- `go test -race ./runner/... -count=1`
- `go test ./runner -run 'NilEvent' -count=20`
- `go vet ./runner/...`
- `git diff --check`

The regression tests cover both `Run` and `RunLive` with a nil-event flood under an active context. They verify that the iterator terminates, yields an error, and includes the agent name.

Fixes #1471


jjsasha63 (changes requested):
Thanks for taking this on. I reverted the eight lines in `runner.go` and both new
cancellation tests failed, so they are testing the fix. Changing the nil-skip
`continue` to `break` now fails the delivery tests too, which closes the coverage
gap #1471 listed.

Two things stop it from fixing the issue.

### A nil flood still hangs when the context is live

I ran the repro from #1471 against this branch, changed only to leave the context
uncancelled, with a consumer that breaks on its first turn:

```
LIVECTX HUNG: body ran 0 times, agent yielded 90117654 nils
```

The nil branch is the only branch in either loop that never calls `yield`. A
consumer's `break` is expressed as `yield` returning false, so it cannot get out.
Cancelling is not a way out either, because the usual `defer cancel()` cannot run
until the range statement returns. A single-goroutine consumer never reaches the
new check, so the fix only helps when something else cancels from another
goroutine.

### A cancelled run now drops an event the agent already produced

One agent yielding nil and then a real event, context cancelled before `Run`:

| runner.go | events delivered | error |
|---|---|---|
| main | 1 | none |
| this PR | 0 | `context.Canceled` |
| this PR, nil removed from the agent | 1 | none |

The check sits inside the nil branch, so whether a cancelled run reports an error
depends on whether a nil happened to arrive first. Nothing downstream treats
`context.Canceled` as an expected end, so it surfaces as a 500 in
`server/adkrest/controllers/runtime.go:168`, a failed task in
`server/adka2a/v2/executor.go:361`, and a tool error in
`tool/agenttool/agent_tool.go:180`.

### The skip is still silent

`runner.go` reports anomalies with `log.Printf` in eight other places, nearest at
line 650. An agent author whose agent yields nil gets zero events, no error and
nothing in the log. #1471 asked for a line here.

### A shape that covers all three

Treat a nil event as an agent contract violation and end the run with an error
naming the agent. That terminates whether the context is cancelled, the consumer
breaks, or neither, and it needs no `ctx` check. If #1417's choice to skip rather
than fail is worth keeping, the softer version is to skip a stray nil, log it once
per invocation, and fail past a bound.

wmh123wmh1:
Thanks for the detailed review. I revised the nil-event handling based on your feedback.

The updated implementation now treats a nil event as an agent contract violation in both `Runner.Run` and `Runner.RunLive`:

- logs the agent name and the nil-event error;
- yields the error to the consumer;
- terminates the current run immediately.

This prevents nil-event floods from spinning indefinitely even when the context remains active, and gives the consumer an iteration where it can observe the error and exit.

I also updated the regression tests to cover both `Run` and `RunLive` with an active context and a nil-event flood. The tests verify that the iterator terminates, yields an error, and includes the agent name.

Validation completed:

- `go test ./runner/... -count=1`
- `go test -race ./runner/... -count=1`
- `go vet ./runner/...`
- `git diff --check`

jjsasha63 (approved):
Thanks, this fixes both halves, so I am approving. I ran the repro from #1471
against `701ec812` with a live context and a consumer that breaks: the consumer
gets its turn and the run returns in 293µs after a single nil, where the previous
revision absorbed 90 million of them. All four mutations of the new branch fail
the tests: deleting it, `return` to `continue`, `yield(nil, err)` to
`yield(nil, nil)`, and dropping the agent name. The first two are caught by a
panic rather than an assertion, which is weaker but real.

Nits to fold in:

### The `adk: ` prefix belongs in the log, not the error

`runner.go:722` and `:943` are the only two errors in the repo carrying it,
against 18 `log.Printf("adk: ` sites. Match line 650:

```go
err := fmt.Errorf("agent %q yielded a nil event", r.rootAgent.Name())
```

### Logging and yielding the same message

The rule is written down at `runner.go:634`: "On an early exit the error cannot
be yielded, because yield must not be called once it has returned false, so it
is logged instead." Yield works here, so the `log.Printf` is a second copy of
the error the caller already gets. `:705`, `:755` and `:998` yield without
logging.

### The contract needs writing down

`agent.Config.Run` says only "Run is the function that defines the agent's
behavior" (`agent/agent.go:99`). A nil yield is now fatal to the whole
invocation, so say so where an agent author will read it:

```go
// Run is the function that defines the agent's behavior.
//
// The sequence must not yield a nil event. The runner ends the invocation
// with an error if it does. Report a failure by yielding a non-nil error
// instead.
```

### A compatibility line in the PR description

Nothing in this repo yields a nil event on a non-error path, so this only
reaches agents people wrote themselves. But `agent/agent.go:198` forwards a nil
straight through, so a stray nil deep inside someone's sub-agent now aborts the
whole invocation where before it was skipped. On v2.3.0 that is worth a line in
the description, so it carries into the release notes.

### Test hygiene

The rewrite dropped `t.Parallel()`, which 5 other tests in the file use, swapped
`t.Context()` for `context.Background()`, and replaced `AutoCreateSession: true`
with a manual `Create`. Worth restoring unless one of them got in the way.

One thing not to fix here. While checking this I found that `RunLive` drops
buffered tool calls when the stream ends mid-transcription: the loop at
`runner.go:934-1035` has no flush after it. It happens on main too, so it is not
yours, and the error you now yield beats the silent truncation main gives today.
I will file it separately.

PR #1492, opened 2026-09-03
docs: state the pre-PR checks that reviews keep catching

karolpiotrowicz:

No linked issue. This is contributor documentation, raised from a pattern in review feedback rather than from a bug report.

**Problem:**

Two defects come up more than any others when reviewing contributions here: a test that passes whether or not the change under review is present, and a behavior change the PR description never mentions. Nothing in CI catches either, and until now nothing in `AGENTS.md` or `CONTRIBUTING.md` asked a contributor to check for them. Someone could follow the documentation exactly and still produce both.

Three smaller gaps sat alongside that. `AGENTS.md` never mentioned the Conventional Commit title rule, even though release tooling reads the landed subject and drops anything it does not recognise from the changelog without reporting an error. The documented format command was `golangci-lint fmt`, which rewrites your files rather than telling you what is wrong, and `golangci-lint run` does not check formatting at all, so there was no documented way to verify formatting without modifying the tree. And the per-module iteration was a parenthetical the reader had to notice and act on, which meant "I ran what the docs say" and "I ran what CI runs" were not the same thing.

**Solution:**

- A **Before you open a PR** section in `AGENTS.md` with the three things no command does for you: prove each new test fails with the source change reverted, state what changes for someone already on the current release, and title the PR as a Conventional Commit.
- A copy-pasteable loop in both files that covers every module the way CI does, using `golangci-lint fmt --diff` so formatting is verified rather than rewritten.
- The linter version pinned in the instructions. v2.12.2 reports a `govet` finding on `plugin/agentanalytics` that the CI-pinned v2.3.1 does not, which reads to a contributor as a failure they caused.
- Three unfalsifiable unit-test requirements in `CONTRIBUTING.md` — "fast and isolated", "written clearly with descriptive names", "aim for high readability" — replaced with one a contributor can actually check.
- The `AGENTS.md` minimal example removed. It duplicated `examples/quickstart`, which the line below it already pointed to, and dropping it pays for most of the additions.
- The `CONTRIBUTING.md` table of contents fixed. Documentation and Alignment with adk-python were listed under Contribution workflow but render as ADK Web subsections, and the ADK Web section was missing from the contents entirely.

**What changes for existing users:** nothing. Documentation only, with no code, no public API surface and no CI configuration touched.

### Testing Plan

**Unit Tests:**

Not applicable, as there are no code changes.

**Manual verification:**

The loop this PR documents was run end to end against both modules with the CI-pinned golangci-lint v2.3.1, and both are green:

```
OK: .
OK: ./plugin/agentanalytics
```

Every internal anchor in both files was checked against the headings that exist, including the cross-file link from `CONTRIBUTING.md` into `AGENTS.md`. The `CONTRIBUTING.md` contents list now has 21 entries against 21 headings, with no mismatch in either direction.

### Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas. Not applicable, documentation only.
- [ ] I have added tests that prove my fix is effective or that my feature works. Not applicable, documentation only.
- [ ] Any dependent changes have been merged and published in downstream modules. None.

### Additional context

The three judgement steps live in `AGENTS.md` and are linked from `CONTRIBUTING.md` rather than duplicated, so there is a single normative copy. The command block is deliberately in both files, because a contributor reading `CONTRIBUTING.md` and an agent reading `AGENTS.md` each need it inline, and six duplicated lines are easy to spot when they drift.


AGENTS.md
```diff
@@ -42,13 +42,35 @@ workspace, while `./...` matches only the module you are standing in:
...
+    && go mod tidy -diff \
+    && go build -mod=readonly ./... \
+    && go test -race -mod=readonly -count=1 -shuffle=on ./... \
+    && golangci-lint run \
+    && golangci-lint fmt --diff ) || echo "FAILED: $m"
+done
+```
+
+`golangci-lint fmt --diff` reports formatting without rewriting; plain
+`golangci-lint fmt` rewrites in place. Note that `golangci-lint run` does not
```
wojtas000: Small thing, but `golangci-lint-run` does check formatting - just does not rewrite the code in place. Example:

```console
$ golangci-lint run
pkg/example/stream.go:35:1: File is not properly formatted (gofumpt)
        return &streamTranslator{
^
1 issues:
* gofumpt: 1
```

CONTRIBUTING.md
```diff
@@ -209,12 +212,53 @@ information on using pull requests.
...
+    && go mod tidy -diff \
+    && go build -mod=readonly ./... \
+    && go test -race -mod=readonly -count=1 -shuffle=on ./... \
+    && golangci-lint run \
+    && golangci-lint fmt --diff ) || echo "FAILED: $m"
+done
+```
+
+`golangci-lint fmt --diff` reports formatting problems without rewriting your
+files. It is a separate step because `golangci-lint run` does not check
```
wojtas000: same as in AGENTS.md

wojtas000 (changes requested):

AGENTS.md
karolpiotrowicz: You're right — I had this backwards.

Reproduced against this repo's config with the CI-pinned v2.3.1, using a gofumpt-only violation (a blank line after an opening brace) so the result could not have come from plain `gofmt`:

```console
$ gofmt -l platform/uuid.go
                                  # silent

$ golangci-lint run ./platform/...
platform/uuid.go:47:1: File is not properly formatted (gofumpt)
1 issues:
* gofumpt: 1
```

An out-of-order import gives the same result. So the lint job already enforces formatting, and the `fmt --diff` step I put in the loop was redundant rather than load-bearing.

Dropped it, and the text now describes the three commands for what they actually do: `run` reports formatting without fixing it, `fmt` rewrites in place, and `fmt --diff` previews what `fmt` would change. The Format entry in the command list goes back to `golangci-lint fmt`, which was right before this branch touched it. Fixed in [the formatting correction](https://github.com/google/adk-go/pull/1492/commits/6d73dd7a8a521c9744690e46d2c0d2adcf2ba977).

That claim came from an audit I had not re-run myself before writing it down, so I went back over every other checkable claim in both files against the tree. Most held — all 37 paths in the repository layout exist, all 13 URLs resolve, the v2.3.1 pin genuinely resolves to that tag, and `goheader`, `guardrail` and the `work` fallback all behave as described. Four did not, and they are fixed in [the follow-up](https://github.com/google/adk-go/pull/1492/commits/540d8a21489e293e96992dcda0da419ec581a23f):

- **The `Before` callback rule was wrong for agents.** Model and tool callbacks short-circuit on a non-nil result or error, but `BeforeAgentCallback` short-circuits only on non-nil content. The skip is gated on `nodeCtx.Ended()` at `agent/agent.go:191` and the error path returns before `EndInvocation` is reached, so an error is surfaced and the agent still runs.
- **The single-cassette record recipe silently did nothing.** `-httprecord` matches the cassette's file path, and only 17 of the 48 cassettes are named after a top-level test, so `-httprecord='TestFoo\.httprr$'` matched no file for the other 31 and exited 0 having recorded nothing. Replaced with a worked example naming a real cassette.
- **"Tests run offline by default"** stated a per-package property as a property of the whole suite, and omitted that `session/vertexai` replays through `rpcreplay` with `testdata/*.replay` rather than through `httprr`.
- **`CONTRIBUTING.md` said `go.yml` filters its `pull_request` trigger on base branch.** #1445 deliberately removed that filter so stacked PRs get built, and left a comment explaining why. `apidiff.yml` still filters.

CONTRIBUTING.md
karolpiotrowicz: Same change here, and the same reasoning. Reproduction and the rest of the audit are in the reply on `AGENTS.md`.

wolo-lab (approved):
Thx

wojtas000 (approved):

PR #1494, opened 2026-09-04
fix(remoteagent): render unmatched function responses as text

### LINKED ISSUE #1482: remoteagent/v2: a function response is sent to a peer that never made the matching call
## 🔴 Required Information

**Describe the Bug:**

`remoteagent/v2` sends a function response to a peer that never made the
matching function call. adk-python converts that case to plain text instead,
because the receiving runner rejects a message carrying a function response it
has no invocation to resume.

`convertParts` hands every part to `adka2a.ToA2AParts`
(`agent/remoteagent/v2/utils.go:180`) with no check on whether the call ID
belongs to this peer, so a locally-executed tool's response goes out as a real
A2A function response.

**Steps to Reproduce:**

1. Drop the test under Minimal Reproduction Code into `agent/remoteagent/v2/`.
2. Run `go test ./agent/remoteagent/v2/ -run TestProbe_UnmatchedFunctionResponseSentRaw -v`.
3. Read the logged parts.

**Expected Behavior:**

The response is rendered as text, matching adk-python
(`src/google/adk/agents/remote_a2a_agent.py:1267-1280`):

```python
text_content = (
    f"Tool {part.function_response.name} returned:"
    f" {json.dumps(part.function_response.response)}"
)
```

Python gates this on `remote_fc_ids`, the set of call IDs this remote agent
itself emitted (`:1227-1235`), and its comment states the failure mode
directly: "The peer has no invocation to resume for a call it never made, and
the receiving runner rejects a message that carries a function response next to
the text of the same history."

**Observed Behavior:**

A session holding a response to the locally-executed `local_tool` call `fc-local`
produces:

```
part[0] = {"data":{"id":"fc-local","name":"local_tool","response":{"v":1}},"metadata":{"adk_type":"function_response"}}
part[1] = {"text":"and now ask the peer"}
```

Part 0 is a function response for a call the peer never issued.

**Environment Details:**

- ADK Library Version: 2.3.0 (`internal/version/version.go`)
- OS: Linux
- Go Version: go1.26.6

**Model Information:**

- N/A, no model involved.

---

## 🟡 Optional Information

**Regression:**

No. Reproduces on `main` as of ffddb2a.

**Additional Context:**

A related gap in the same area: `getUserFunctionCallAt` matches the originating
call with `isFunctionCallEvent`, which checks only the call ID
(`agent/remoteagent/v2/utils.go:67-72`). adk-python additionally requires
`event.author == self.name`, so a call ID that happens to collide with one the
coordinator made does not resume the peer's task. Worth folding into the same
fix, since both need the "which calls did this peer make" set.

`presentAsUserMessage` already renders a function response as text
(`utils.go:178-181`), but only for events authored by a third agent. A
user-authored response takes the raw path.

```go
package remoteagent

import (
	"encoding/json"
	"testing"

	"google.golang.org/genai"

)

func TestProbe_UnmatchedFunctionResponseSentRaw(t *testing.T) {
	ctx := t.Context()
	store := session.InMemoryService()
	resp, _ := store.Create(ctx, &session.CreateRequest{AppName: "t", UserID: "u"})

	// A function response answering a call the remote peer never made.
	fr := newEventFromParts("user", &genai.Part{FunctionResponse: &genai.FunctionResponse{
		ID: "fc-local", Name: "local_tool", Response: map[string]any{"v": 1},
	}})
	// Not the last event, so the getUserFunctionCallAt resume path is not taken.
	tail := newEventFromParts("user", genai.NewPartFromText("and now ask the peer"))
	for _, e := range []*session.Event{fr, tail} {
		if err := store.AppendEvent(ctx, resp.Session, e); err != nil {
			t.Fatal(err)
		}
	}
	a, _ := agent.New(agent.Config{Name: "remote-agent"})
	ictx := icontext.NewInvocationContext(ctx, icontext.InvocationContextParams{
		Agent: a, Session: resp.Session,
	})

	msg, err := newMessage(ictx, A2AConfig{})
	if err != nil {
		t.Fatalf("newMessage() error = %v", err)
	}
	for i, p := range msg.Parts {
		b, _ := json.Marshal(p)
		t.Logf("part[%d] = %s", i, b)
	}
}
```

- Always (100%) whenever a locally-executed tool's response is still in the
  replayed history when the remote agent is dispatched.

leilei3167:
## Summary

`remoteagent/v2` was forwarding function responses for call IDs the remote peer never issued as real A2A `function_response` parts. The receiving runner rejects those because it has no invocation to resume.

This aligns with adk-python (`remote_a2a_agent.py`): collect call IDs emitted by this peer, and render unmatched function responses as text (`Tool {name} returned: {json}`). Also require the matching function call to be authored by the peer when taking the resume path (`getUserFunctionCallAt`), matching python's `event.author == self.name` check.

Fixes #1482

## Test plan

- [x] `go test ./agent/remoteagent/v2/ -run TestProbe_UnmatchedFunctionResponseSentRaw -count=1`
- [x] `go test ./agent/remoteagent/v2/ -run 'TestGetUserFunctionCallAt|TestToMissingRemoteSessionParts|TestPresentAsUserMessage|TestToMissingRemoteSessionParts_KeepsMatched' -count=1`
- [x] `go test ./agent/remoteagent/v2/ -count=1`

cc @yarolegovich — small remoteagent/v2 history-replay fix; happy to adjust if you'd rather gate this only in task mode.


jjsasha63 (changes requested):
The fix closes #1482. I ported adk-python's regression test for this
(`test_construct_message_parts_from_session_foreign_function_response_converted_in_default_mode`,
test_remote_a2a_agent.py:1396) to Go against this branch, and the foreign
function response does come out as a single text part.

Five things I would change before it lands.

## 1. An empty agent name flattens the peer's own responses

`isFunctionCallEvent` treats `""` as "skip the author gate" (utils.go:72), but
`collectRemoteFunctionCallIDs` returns an empty set for it (utils.go:85). With an
empty name the set is empty, so every function response is rewritten, including a
matched one. I built a session holding a call the peer itself made plus its
response, ran `toMissingRemoteSessionParts` with an unnamed agent, and got:

```
part[0] = {"text":"Tool peer_tool returned: {\"ok\":true}"}
```

On main that stays a `function_response` data part, so it is a regression for
anyone who does not set a name. The path is reachable: `agent.New(agent.Config{})`
returns no error, `NewA2A` validates `AgentCard` and `ClientProvider` but never
`cfg.Name` (a2a_agent.go:325-341), and the comment at utils.go:177-180 already
says anonymous wrappers and conformance harnesses hit this.

Dropping the early return at :85 fixes it. The gate at :72 then covers the case
on its own, because an empty set and a skipped gate now agree.

## 2. A call with an empty ID matches one gate and not the other

`collectRemoteFunctionCallIDs` requires `part.FunctionCall.ID != ""`
(utils.go:94). `isFunctionCallEvent` has no such guard (utils.go:75), so
`callID == ""` matches. A call with no ID therefore takes the resume path and has
its own response rewritten to text in the same message, putting a `TaskID` next
to a flattened response.

## 3. Hoist the ID set out of the per-event loop

`collectRemoteFunctionCallIDs` runs inside `convertParts` (utils.go:232), and
`convertParts` is called once per event from the loop at utils.go:175-186, so
each event rescans the whole session. Python builds `remote_fc_ids` once at
remote_a2a_agent.py:1227, above the loop at :1237.

I benchmarked `toMissingRemoteSessionParts` over a synthetic text-only session:

| events | main | this branch |
| --- | --- | --- |
| 1000 | 0.215 ms | 8.81 ms |
| 5000 | 3.46 ms | 237 ms |

Passing the set in as a fourth parameter and computing it once in `newMessage`
and `toMissingRemoteSessionParts` keeps the behaviour and drops the rescan.

## 4. Python does not run the custom converter on the rewritten part

utils.go:233 rewrites first and then hands the text part to
`cfg.GenAIPartConverter` (:235-237). Python builds the text part at
remote_a2a_agent.py:1280 and reaches `self._genai_part_converter` only in the
`else` branch at :1282, and the test pins that with
`self.mock_genai_part_converter.assert_not_called()` (test_remote_a2a_agent.py:1435).
A converter that returns nil for text parts will drop the tool result here
instead of passing it through.

## 5. The rendered JSON does not match Python's

`json.Marshal` omits the separators Python's `json.dumps` emits and escapes HTML:

```
go = Tool tool_1 returned: {"result":"done"}
py = Tool tool_1 returned: {"result": "done"}

go = {"url":"https://x/?a=1\u0026b=2"}
py = {"url": "https://x/?a=1&b=2"}
```

The spacing is cosmetic, but the escaping is not: any tool result containing
`<`, `>` or `&` reaches the peer as `\u003c`, `\u003e`, `\u0026`. A
`json.Encoder` with `SetEscapeHTML(false)` fixes the escaping.

## Test coverage

I mutated each new behaviour and reran `go test ./agent/remoteagent/... -count=1`.
Four mutations survive:

- Removing `event.Author != agentName` from `collectRemoteFunctionCallIDs`
  (utils.go:90). This is the filter the fix rests on. It survives because
  `TestProbe_UnmatchedFunctionResponseSentRaw` appends no function-call event at
  all, so the ID set is empty either way and the filter never runs.
- Replacing `ctx.Agent().Name()` with `""` at a2a_agent.go:540.
- Iterating `event.Content.Parts` instead of `genaiParts` in the
  `GenAIPartConverter` branch, which skips the rewrite entirely.
- Dropping non-function-response parts inside the rewrite loop, so no test covers
  an event that mixes a function response with sibling parts.

The first one is worth a case of its own: a coordinator-authored function call
whose ID matches the response, which should still be rewritten.

## On gating this to task mode

I would not gate it. The text conversion at remote_a2a_agent.py:1267-1271 has no
`mode == "task"` term, and Python's own test for it is named
`..._converted_in_default_mode`. The mode check at :1256 covers only the sibling
function-call skip, and `task_scope` at :1229 narrows which events feed
`remote_fc_ids`. Go's remoteagent has no mode to gate on.

## Smaller things

- Nothing logs a rewrite. Python does not log it either, so this is a suggestion
  rather than a parity gap, but the package already uses `log.Warn` at
  utils.go:188 and someone debugging "my tool result turned into text" has
  nothing to find.
- `TestProbe_UnmatchedFunctionResponseSentRaw` asserts the response is not sent
  raw.
- The comment at utils.go:37 names `toUserFunctionCall`; the function is
  `getUserFunctionCallAt`. The diff already touches those lines.

To verify after the changes: `go test ./agent/remoteagent/... -count=1`.

leilei3167:
@jjsasha63 Addressed the five review items:

1. Dropped the empty-name early return in `collectRemoteFunctionCallIDs` so it agrees with the author gate.
2. Empty call IDs no longer match in `isFunctionCallEvent`.
3. Hoisted the remote FC ID set out of the per-event loop (computed once in `newMessage` / `toMissingRemoteSessionParts`).
4. Rewritten FR→text parts skip `GenAIPartConverter` (converter only on the non-rewritten path).
5. JSON encoding uses `SetEscapeHTML(false)` plus Python-style `: `/`, ` separators.

Also added the mutation-survivor tests (coordinator-authored call still rewritten, empty agent name keeps own FR, empty call ID, mixed sibling parts, converter not called on rewrite). Merged latest `upstream/main`.

leilei3167 (commented):
Thanks for the thorough review — all five points are in d7538cd:

1. Dropped the empty-name early return; anonymous agents now keep their own matched responses.
2. `isFunctionCallEvent` rejects empty call IDs so resume and rewrite agree.
3. `remoteFCIDs` is built once in `newMessage` / `toMissingRemoteSessionParts` and passed into `convertParts`.
4. Rewritten FR→text parts skip `GenAIPartConverter`.
5. `marshalFunctionResponseJSON` uses `SetEscapeHTML(false)` plus Python-style `: ` / `, ` separators.

Also added the mutation-survivor cases you called out (coordinator-authored call, empty agent name, empty call ID, mixed sibling parts, converter not invoked on rewrite) and a `log.Warn` on rewrite. `go test ./agent/remoteagent/... -count=1` passes on head. Ready for another look when you have a moment.

leilei3167:
@jjsasha63 Friendly bump — the five review items are in `d7538cd` (plus later main merge). Happy to adjust anything I missed.

jjsasha63 (changes requested):
Two blockers, both reproduced locally on ee28c1d.

### Blocking

1. The resume path builds the message this PR exists to prevent.

`newMessage` now passes `remoteFCIDs` into `convertParts` on the
`getUserFunctionCallAt` branch (a2a_agent.go:540-543). One user event can carry
several function responses. `mergeParallelFunctionResponseEvents`
(internal/llminternal/base_flow.go:1468) merges parallel responses into a single
role-"user" event. When one answers the peer's call and one answers a locally
executed tool, the outgoing message mixes both encodings:

```json
{"parts":[
  {"data":{"id":"fc-remote","name":"peer_tool","response":{"ok":true}},
   "metadata":{"adk_type":"function_response"}},
  {"text":"Tool local_tool returned: {\"secret\": \"s\"}"}],
 "role":"ROLE_USER"}
```

Python forbids exactly this shape. `_sanitize_user_function_response_event`
computes `preserve_as_resume` at remote_a2a_agent.py:338. The comment there
reads: "If any function_response is kept as data, the message must stay a
resume: no text (including a flattened answer) can ride alongside it". The
Python resume path never rewrites to text at all. Only
`_construct_message_parts_from_session` does.

Keep the rewrite in `toMissingRemoteSessionParts` and leave the resume path as
it was.

2. The `fmt.Sprintf("%v", v)` fallback in `marshalFunctionResponseJSON`
(utils.go:110) has two failure modes.

A cyclic payload kills the process. `Encode` fails on the cycle, then `%v`
recurses without cycle detection. I get `fatal error: stack overflow`, which
`recover` does not catch.

NaN is the ordinary case, and `encoding/json` refuses it too. The fallback then
sends Go map syntax to the peer's model: `map[score:NaN]`.

Drop the fallback. Return an error, or a fixed string that does not embed the
payload.

### Worth fixing

3. utils.go:84, the doc comment is backwards. It says an empty `agentName` gives
"the same author-gate semantics as isFunctionCallEvent". It is the opposite.
`isFunctionCallEvent` skips the gate on an empty name and matches any author,
while `collectRemoteFunctionCallIDs` matches only `Author == ""`.

For an anonymous agent that is more than a comment bug. A coordinator-authored
call satisfies `getUserFunctionCallAt`, so the resume branch runs and takes
taskID and contextID from it. The same response is then flattened to text,
because coordinator IDs were never collected.

4. `insertJSONSeparators` closes the smallest of five gaps. Same payload through
both languages:

```
Go:     {"apple":2,"big":10000000000000000000,"count":1,"note":"café ☕","zebra":1}
Python: {"zebra": 1, "apple": 2, "note": "caf\u00e9 \u2615", "count": 1.0, "big": 1e+19}
```

Key order, non-ASCII escaping and float rendering all still differ after the 40
lines. Nothing parses this string; it is prose for the peer's model.
`json.Marshal` with `SetEscapeHTML(false)` gives the same value and drops both
helpers. `internal/llminternal/base_flow.go:1146` already has that body as
`marshalJSONNoHTMLEscape`.

5. `collectRemoteFunctionCallIDs` runs twice per message. a2a_agent.go:540
computes it. On the non-resume path `toMissingRemoteSessionParts` (utils.go:186)
then scans every session event again, and the first result is discarded. Moving
line 540 inside the `if` fixes it.

6. `adka2a.ToA2APart` already exists (server/adka2a/v2/parts.go:77), so each
part does not need wrapping in a one-element slice for `ToA2AParts`.

### Nits

7. nil `Part` elements: utils.go:93 and utils.go:247 dereference without a
guard. With a custom `GenAIPartConverter`, main passed a nil part through to the
converter, and this panics before the converter sees it. #1556 landed the same
guard for the shared accessors two days ago.

8. An empty `Name` gives `Tool  returned: null`, with the double space.

### Parity

Parity vs src/google/adk/agents/remote_a2a_agent.py @ adk-python 2.8.0
(adk-go 2.3.0):

- Matches: the `remote_fc_ids` gate and the `Tool {name} returned: {json}` text
  on the history path.
- Diverges: the resume path rewrites, where Python's `preserve_as_resume`
  forbids it. That is finding 1.
- Not ported, and not this PR's job: `_without_credential_parts` (:227). adk-go
  has no credential scrubbing in this package. An `adk_request_credential`
  response reaching the new text path becomes readable prose for the peer's
  model: `Tool adk_request_credential returned: {"oauth2": {"access_token":
  "<redacted>", "client_secret": "<redacted>"}}`. That payload already went over
  the wire as a data part before this PR, so it is not new egress. It moves from
  "rejected by the peer" to "read by the peer's model". Worth a separate issue.

The branch is green: `go build ./...`, `go vet` and
`go test ./agent/remoteagent/v2/ -count=1` all pass. Neutering the rewrite
branch fails 5 of the 8 new tests, so the guard is pinned by the tests.

leilei3167:
Thanks for catching these — addressed in 6f7ab36:

1. Resume path no longer rewrites FR→text (`convertParts` gets nil remoteFCIDs there); rewrite stays on the history path only.
2. Dropped the `%v` fallback; marshal failure now returns a fixed `<unserializable>` placeholder (covers NaN / cyclic payloads without embedding the value).

Also took the follow-ups: empty-name ID collection aligned with `isFunctionCallEvent`, removed `insertJSONSeparators`, collect IDs once on the history path, `ToA2APart`, nil part guards, and empty tool name wording.

`go test ./agent/remoteagent/... -count=1` and `go vet` are green on tip. Ready for another look when you have a moment.

jjsasha63 (dismissed):
Both blockers are fixed, and I verified each rather than reading for it.

Resume path: a user event carrying the peer's response plus a local tool's
response now goes out with both as data parts.

```
parts=2 [{"data":{"id":"fc-remote",...},"metadata":{"adk_type":"function_response"}},
         {"data":{"id":"fc-local",...},"metadata":{"adk_type":"function_response"}}]
```

The history path still rewrites, so #1482 stays fixed:
`Tool local_tool returned: {"x":"y"}`.

Marshal failure: cyclic and NaN payloads both return `<unserializable>`. The
cyclic case used to take the process down with a stack overflow that `recover`
could not catch.

Both fixes are pinned. Restoring `remoteFCIDs` on the resume path fails
`TestNewMessage_ResumeKeepsMixedFunctionResponsesAsData`. Restoring the `%v`
fallback fails `TestMarshalFunctionResponseJSON_Unserializable`.

`go build ./...`, `go vet ./agent/remoteagent/...` and
`go test ./agent/remoteagent/... -count=1` are green on 6f7ab36b.

Parity vs src/google/adk/agents/remote_a2a_agent.py @ adk-python 2.8.0: the
resume path now matches `preserve_as_resume` (:338). The history path matches
the `remote_fc_ids` gate and the text rendering at :1267-1280. Dropping the
`json.dumps` separators is right, because nothing parses that string.

One thing I owe you: items 6 to 8 last round were nits, and a third round should
have held to blockers only. You fixed them anyway and the code is better for it,
but you should not have had to.

jjsasha63 (changes requested):
The isolation-scope merge reopened #1482 inside a scoped invocation.

`toMissingRemoteSessionParts` now filters events by scope in both loops.
`collectRemoteFunctionCallIDs` (utils.go:88, called at :176) does not, so it
collects peer call IDs from every scope. A peer call in one scope then licenses
a function response in a different scope to go out as real A2A data.

I built that session against f923f0b8b:

```
part[0] text="" meta=map[adk_type:function_response]
```

That response answers a call the peer never made in this scope, which is the
shape #1482 is about.

Python scopes the same set (remote_a2a_agent.py:1229-1235):

```python
if (not task_scope or event.isolation_scope == task_scope) and event.author == self.name:
```

The fix is small. Take a `scope` parameter, skip events whose `IsolationScope`
differs, and pass `ctx.IsolationScope()` at the call site. I applied exactly
that, and `go test ./agent/remoteagent/... -count=1` stayed green, so nothing
existing depends on the wider scan. No current test covers the leak, so it needs
one.

Separately, the description needs one correction. It says the author gate in
`isFunctionCallEvent` matches "python's `event.author == self.name` check".
Python's resume path matches on call ID alone, through
`find_matching_function_call` and `find_event_by_function_call_id`
(flows/llm_flows/functions.py:316-326). The `event.author == self.name` at
remote_a2a_agent.py:1156 sits in `_is_remote_response`, which decides where to
stop the history walk.

The gate itself is right, and the real reason is stronger than parity. The
resume path exists to recover the A2A `taskID` and `contextID`. Only
peer-authored events carry them, because `NewRemoteAgentEvent`
(server/adka2a/v2/events.go:31) is the sole writer and stamps the agent's own
name. Matching a coordinator's call therefore returns empty IDs and forwards a
function response for a call the peer never made. Python does forward it. Worth
saying that in the description, because the next person to diff the two will
read it as drift.

CI is green on f923f0b8b.

leilei3167:
Thanks — the IsolationScope merge reopened the leak you described.

`collectRemoteFunctionCallIDs` now takes `scope` and skips events whose `IsolationScope` differs (same rule as `toMissingRemoteSessionParts` / Python `task_scope`). Added `TestRemoteAgent_ScopedMessage_ForeignScopeCallIDDoesNotLicenseInScopeResponse`.

On the author-gate wording: you're right that Python's resume path matches on call ID alone; our `isFunctionCallEvent` author gate is intentional because only peer-authored events carry A2A `taskID`/`contextID` (`NewRemoteAgentEvent` stamps the agent name). Matching a coordinator call would resume with empty IDs and forward an FR for a call the peer never made. I'll treat that as a deliberate divergence, not parity drift.

Tip: `f480359de2c7`. `go test ./agent/remoteagent/... -count=1` green.

leilei3167:
@jjsasha63 Tip `f480359` already scopes `collectRemoteFunctionCallIDs` by IsolationScope (foreign-scope peer call IDs are skipped) and the regression test above covers it. Mind another look when you have a moment?

jjsasha63 (approved):
LGTM

PR #1497, opened 2026-09-04
fix(workflow): make ParallelWorker errors deterministic

### LINKED ISSUE #1481: workflow: ParallelWorker surfaces a non-deterministic error, and drops a caller cancellation during dispatch
## 🔴 Required Information

**Describe the Bug:**

Two ways `workflow.ParallelWorker` loses or scrambles an error. Both diverge
from adk-python's `_parallel_worker.py`, and I reproduced both on `main`.

1. When several items fail at the same time, which error surfaces is
   non-deterministic. adk-python deliberately makes this the lowest input
   index.
2. When the caller cancels mid-dispatch and no worker has errored, the
   cancellation is dropped and the run reports success with holes in the output.
   adk-python propagates `CancelledError`.

**Steps to Reproduce:**

1. Drop the two tests under Minimal Reproduction Code into `workflow/`.
2. Run `go test ./workflow/ -run TestParallelWorkerBug -v`.
3. Both fail.

**Expected Behavior:**

1. The error from the lowest-index item, every run. adk-python sorts the
   completed set by input index for exactly this reason
   (`src/google/adk/workflow/_parallel_worker.py:127-131`): "so that, when
   several tasks fail in the same wake-up, the surfaced exception is
   deterministic across runs and replays". It is asserted over 10 iterations in
   `test_parallel_worker_simultaneous_failures_raise_lowest_index`.
2. `Run` yields `context.Canceled`, not a success event.

**Observed Behavior:**

1. Over 40 runs with two items that both fail immediately: `item-0 failed` 34
   times, `item-1 failed` 6 times. On `origin/main` the split was 36/4.
2. `maxConcurrency=1`, four items whose wrapped node sleeps 200ms and succeeds,
   caller context cancelled at 300ms:

   ```
   err=<nil> out=[]interface{}{1, 2, interface{}(nil), interface{}(nil)}
   ```

   Items 2 and 3 were skipped by the dispatch loop's `workerCtx.Done()` branch,
   which calls `wg.Done(); continue` without sending a result and without
   calling `reportFailure`, so `firstErr` stays nil and `Run` takes the success
   path.

**Environment Details:**

- ADK Library Version: 2.3.0 (`internal/version/version.go`)
- OS: Linux
- Go Version: go1.26.6

**Model Information:**

- N/A, no model involved.

---

## 🟡 Optional Information

**Regression:**

No. Both reproduce on `main` as of ffddb2a.

**Additional Context:**

Sketch of the fixes, both local to `workflow/parallel_worker.go`:

1. Record failures by index and pick the lowest, rather than letting
   `sync.Once` hand `firstErr` to whichever worker calls `reportFailure` first.
   The `cancelFunc()` side can stay first-wins; only the error selection needs
   ordering.
2. In the dispatch loop's two cancellation branches, note that an item was
   skipped. After the collection loop, if `firstErr` is nil and anything was
   skipped, set it to the context error. Fail-fast cancellations already have
   `firstErr` set, so the nil check keeps them out.

Two further divergences from adk-python that I would not fix, listed so nobody
re-finds them:

- adk-python cancels items still in flight and gives up after
  `_CANCELLED_ITEM_DRAIN_TIMEOUT_SECONDS` (5s) with a warning. Go cannot cancel
  a goroutine, so `wg.Wait()` waiting on a worker that ignores `ctx` is the
  honest Go semantics — one failing item plus one that sleeps 6s ignoring `ctx`
  makes `Run` take 6.0s.
- adk-python raises `ValueError` for `max_parallel_workers < 1`. Go documents
  `maxConcurrency <= 0` as "no limit" on `NewParallelWorker`, so -5 is accepted.
  Changing that would break the published API for no behavior gain.

```go
package workflow

import (
	"context"
	"fmt"
	"testing"
	"time"

)

func runPW(t *testing.T, pw *ParallelWorker, ctx context.Context, items []any) (error, any) {
	t.Helper()
	exCtx := agent.NewContext(&MockInvocationContext{Context: ctx})
	var gotErr error
	var gotOut any
	for ev, err := range pw.Run(exCtx, items) {
		if err != nil {
			gotErr = err
		}
		if out, ok := extractOutput(ev); ok {
			gotOut = out
		}
	}
	return gotErr, gotOut
}

// 1. Which error surfaces is non-deterministic.
func TestParallelWorkerBug_SimultaneousFailuresLowestIndex(t *testing.T) {
	wrapped := NewFunctionNode("fails", func(ctx agent.Context, input any) (any, error) {
		return nil, fmt.Errorf("%v failed", input)
	}, defaultNodeConfig)

	seen := map[string]int{}
	for i := 0; i < 40; i++ {
		pw, err := NewParallelWorker("parallel", wrapped, 0, defaultNodeConfig)
		if err != nil {
			t.Fatal(err)
		}
		gotErr, _ := runPW(t, pw, context.Background(), []any{"item-0", "item-1"})
		seen[fmt.Sprint(gotErr)]++
	}
	if seen["item-0 failed"] != 40 {
		t.Errorf("want item-0 every run, got %v", seen)
	}
}

// 2. Caller cancels mid-dispatch and the cancellation is dropped.
func TestParallelWorkerBug_CallerCancelDuringDispatch(t *testing.T) {
	wrapped := NewFunctionNode("slow_ok", func(ctx agent.Context, input any) (any, error) {
		time.Sleep(200 * time.Millisecond) // deliberately ignores ctx
		return input, nil
	}, defaultNodeConfig)

	pw, err := NewParallelWorker("parallel", wrapped, 1, defaultNodeConfig)
	if err != nil {
		t.Fatal(err)
	}

	ctx, cancel := context.WithCancel(t.Context())
	done := make(chan struct{})
	var gotErr error
	var gotOut any
	go func() {
		gotErr, gotOut = runPW(t, pw, ctx, []any{1, 2, 3, 4})
		close(done)
	}()

	time.Sleep(300 * time.Millisecond)
	cancel()

	select {
	case <-done:
		t.Fatal("timeout")
	}
	if gotErr == nil {
		t.Errorf("cancelled mid-dispatch but no error reported; output = %#v", gotOut)
	}
}
```

- Case 1: Often (50%+) — 6 runs in 40 gave the wrong error.
- Case 2: Always (100%).

wmh123wmh1:

- Closes: #1481

**Problem:**

`workflow.ParallelWorker` can surface different errors for the same input depending on goroutine completion order. It can also report success with nil output entries when the caller cancels while items are waiting to be dispatched.

**Solution:**

- Track worker failures by input index and return the lowest-index failure after workers drain.
- Keep fail-fast cancellation for remaining workers.
- Ignore secondary `context.Canceled` errors caused by internal fail-fast cancellation so they do not mask the original worker failure.
- Record skipped items during dispatch and propagate the parent context error instead of emitting a partial successful result.
- Add regression tests for out-of-order failures, fail-fast cancellation, and cancellation during bounded dispatch.

### Testing Plan

**Unit Tests:**

- `go test ./workflow/...`
- `go test -race ./workflow/...`
- `go vet ./workflow/...`
- `go test -race ./workflow/ -run '^TestParallelWorker_' -count=20`

Not applicable. This change is isolated to the workflow execution and unit-test coverage.

### Checklist

- [x] The affected workflow unit tests pass locally.

### Additional context

The implementation keeps the documented `maxConcurrency` semantics and existing goroutine-drain behavior unchanged.

### DIFF
--- a/workflow/parallel_worker.go
+++ b/workflow/parallel_worker.go
@@ -15,6 +15,8 @@
 package workflow
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"iter"
 	"reflect"
@@ -61,7 +63,8 @@ func NewParallelWorker(name string, wrapped Node, maxConcurrency int, cfg NodeCo
 // the ParallelWorker will perform 1 ("a") + 3 ("b" retried) + 1 ("c") = 5 calls in total.
 //
 // If any of the wrapped node executions returns a non-retryable error, the workflow
-// will fail fast, cancel other in-flight workers, and return this first encountered error.
+// will fail fast, cancel other in-flight workers, and return the error from
+// the lowest-index failed input.
 //
 // In case the wrapped node produces more then one output event, they will be
 // aggregated into a list, and the final result will be a multi dimensional list.
@@ -99,16 +102,23 @@ func (n *ParallelWorker) Run(ctx agent.Context, input any) iter.Seq2[*session.Ev
 
 		resCh := make(chan workerResult, nItems)
 
-		// reportFailure cancels the remaining workers. runWorker calls it
-		// synchronously, before releasing its semaphore slot (see below).
+		// reportFailure records the lowest-index worker error and cancels the
+		// remaining workers. runWorker calls it synchronously, before releasing
+		// its semaphore slot (see below).
 		var reportOnce sync.Once
+		var firstErrMu sync.Mutex
 		var firstErr error
-		reportFailure := func(err error) {
-			reportOnce.Do(func() {
+		firstErrIndex := nItems
+		reportFailure := func(index int, err error) {
+			firstErrMu.Lock()
+			if index < firstErrIndex {
+				firstErrIndex = index
 				firstErr = err
-				cancelFunc()
-			})
+			}
+			firstErrMu.Unlock()
+			reportOnce.Do(cancelFunc)
 		}
+		skipped := false
 
 		// Branch isolation: derive a per-item sub-branch so each
 		// worker's wrapped node sees an isolated event history (the
@@ -124,6 +134,7 @@ func (n *ParallelWorker) Run(ctx agent.Context, input any) iter.Seq2[*session.Ev
 				select {
 				case sem <- struct{}{}:
 				case <-workerCtx.Done():
+					skipped = true
 					wg.Done()
 					continue
 				}
@@ -131,6 +142,7 @@ func (n *ParallelWorker) Run(ctx agent.Context, input any) iter.Seq2[*session.Ev
 				// would pick randomly if cancellation raced the send above.
 				select {
 				case <-workerCtx.Done():
+					skipped = true
 					<-sem
 					wg.Done()
 					continue
@@ -139,6 +151,7 @@ func (n *ParallelWorker) Run(ctx agent.Context, input any) iter.Seq2[*session.Ev
 			} else {
 				select {
 				case <-workerCtx.Done():
+					skipped = true
 					wg.Done()
 					continue
 				default:
@@ -165,8 +178,15 @@ func (n *ParallelWorker) Run(ctx agent.Context, input any) iter.Seq2[*session.Ev
 			}
 		}
 
-		if firstErr != nil {
-			yield(nil, firstErr)
+		firstErrMu.Lock()
+		finalErr := firstErr
+		firstErrMu.Unlock()
+		if finalErr == nil && (skipped || ctx.Err() != nil) {
+			finalErr = ctx.Err()
+		}
+
+		if finalErr != nil {
+			yield(nil, finalErr)
 			return
 		}
 
@@ -183,7 +203,7 @@ type workerResult struct {
 	err   error
 }
 
-func (n *ParallelWorker) runWorker(ctx agent.Context, idx int, item any, sem chan struct{}, resCh chan<- workerResult, wg *sync.WaitGroup, reportFailure func(error)) {
+func (n *ParallelWorker) runWorker(ctx agent.Context, idx int, item any, sem chan struct{}, resCh chan<- workerResult, wg *sync.WaitGroup, reportFailure func(int, error)) {
 	defer wg.Done()
 	defer func() {
 		if sem != nil {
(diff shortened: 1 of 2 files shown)


jjsasha63 (changes requested):
Requesting changes. The lowest-index rule surfaces a cancellation artifact instead of the root cause, and adk-python resolves the same race the other way.

## The suppression check misses real cancellation errors

`parallel_worker.go:243` suppresses only errors satisfying `errors.Is(runErr, context.Canceled)`. A model or tool call cancelled by the fail-fast returns a gRPC status error, `rpc error: code = Canceled desc = context canceled`, which does not wrap `context.Canceled`. It survives the check, sets `firstErr` at its own index, and wins.

I ran four items with the genuine failure at index 3 and gRPC-shaped cancellation errors at 0 to 2:

| | surfaced error over 15 runs |
| --- | --- |
| main | `REAL ROOT CAUSE` 15/15 |
| this branch | `rpc error: code = Canceled` 15/15 |

## adk-python surfaces error 1 for this input

`src/google/adk/workflow/_parallel_worker.py` raises the failure as soon as the `asyncio.wait` batch containing it completes (line 144), before anything is cancelled. `_drain_pending_items` then cancels the rest and drops their exceptions through `gather(..., return_exceptions=True)` (line 161). A task still in flight when the failure is raised can never influence what is surfaced.

`TestParallelWorker_LowestIndexErrorWins` has index 0 wait for cancellation and then return `error 0`, and requires `error 0` to win. Python gives `error 1` there. CONTRIBUTING.md:238 makes adk-python the source of truth, so that test pins the divergence.

## Suggested change

Use the structural signal Python uses, rather than matching on the error value:

```go
// parallel_worker.go:243
if ctx.Err() != nil {
	resCh <- workerResult{index: idx, err: runErr}
	return
}
```

The `context` and `errors` imports then go. With that, `go test -race ./workflow/ -count=3` passes except `TestParallelWorker_LowestIndexErrorWins`, which reports `got "error 1"`. That test wants rebuilding around two independent failures rather than one failure and one cancellation casualty.

## The skipped flag never changes the outcome

`finalErr = ctx.Err()` is a no-op when `ctx.Err()` is nil, so `skipped || ctx.Err() != nil` reduces to `ctx.Err() != nil`. I deleted the flag and all four assignments and ran `go test -race ./workflow/ -run '^TestParallelWorker' -count=5`: green, `TestParallelWorker_CancelDuringDispatch` included.

## The new state overlaps the drain loop

The mutex, `sync.Once`, `firstErr` and `firstErrIndex` cover ground the `resCh` drain loop already has: it runs on one goroutine and carries `res.index`. Picking the error there removes all four. `sync.Once` around `cancelFunc` is redundant either way, since `context.CancelFunc` is idempotent by contract.

## Unbounded drain, pre-existing

`wg.Wait()` has no timeout, so a wrapped node that ignores cancellation holds `Run` open forever. Python caps the same drain at 5 seconds (`_parallel_worker.py:39`). Not yours to fix here, but it is why the new dispatch test has to release its worker through `close(release)`. Worth a separate issue.

wmh123wmh1:
Thanks for the detailed review. I updated PR #1497 to address the points you raised.

Changes included:
- Use the worker context cancellation state instead of `errors.Is(err, context.Canceled)`, so gRPC-shaped cancellation errors are not treated as the root cause.
- Select the lowest-index reportable failure in the single result-draining loop.
- Remove the redundant mutex, `sync.Once`, shared `firstErr` state, and `skipped` flag.
- Do not treat cancellation during retry delay as a worker failure; caller cancellation is propagated from the parent context.
- Rewrite the lowest-index test to use independent failures, and add coverage for gRPC-style cancellation.
- Leave the unbounded worker drain behavior unchanged, as suggested, since that is a separate issue.

- `go test ./workflow/...`
- `go test -race ./workflow/...`
- `go test -race ./workflow/ -run '^TestParallelWorker_' -count=20`
- `go vet ./workflow/...`

The changes have been pushed to the existing PR. Please take another look when convenient.

PR #1498, opened 2026-09-04
fix: backport A2A peer metadata trust fix (#1226)

wolo-lab:
## Problem

`toEventActions()` restores `session.EventActions.TransferToAgent` straight from metadata supplied by a remote A2A peer, with no filtering. A malicious or compromised remote agent can set that metadata on any `Message`, `Task`, `TaskStatusUpdateEvent` or `TaskArtifactUpdateEvent` it returns, and the local orchestrator honours it — redirecting execution to a different agent inside the deployer's own multi-agent tree, bypassing the sequencing the deployer intended.

The 1.x line carries the same unfiltered restore that 2.x did before the fix, so the exposure is identical. adk-python resolved this by excluding `transfer_to_agent` from the set of fields a remote peer may set, on the grounds that it drives the caller's own control flow rather than describing the peer's session.

## Summary

Backport of #1226 to the 1.x maintenance line, cherry-picked with no adaptation — the change applies to `v1` unmodified, covering both the v1 and v2 A2A client packages and the `adka2a/v2` event and metadata paths.

### DIFF
--- a/agent/remoteagent/a2a_agent.go
+++ b/agent/remoteagent/a2a_agent.go
@@ -137,6 +137,13 @@ func NewA2A(cfg A2AConfig) (agent.Agent, error) {
 		Description:          cfg.Description,
 		BeforeAgentCallbacks: cfg.BeforeAgentCallbacks,
 		AfterAgentCallbacks:  cfg.AfterAgentCallbacks,
+		// The deprecated package predates AllowTransferToAgent's introduction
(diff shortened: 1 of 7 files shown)


jjsasha63 (changes requested):
Faithful to #1226. With the module path
normalized, the added and removed lines are identical on both sides, 159 each.

jjsasha63 (approved):
Lgtm

PR #1499, opened 2026-09-04
fix: backport tool-confirmation trust fixes (#1357, #1369)

wolo-lab:
## Problem

Two independent holes in the tool-confirmation resume path let an attacker who can append an event into the session store execute a tool the user never approved. Both are present on 1.x.

- **No author gate on the confirmation request.** The resume path trusts the tool name and arguments embedded in the confirmation-request event. 1.x checks the author only on the response side, at [`request_confirmation_processor.go:60`](https://github.com/google/adk-go/blob/cb2981844eae0613a90b100d98c32d231607ac69/internal/llminternal/request_confirmation_processor.go#L60), so a request event authored by anything at all can drive a resume.
- **Ambiguous confirmation-call IDs.** A user's approval carries only the confirmation-call ID and `{"confirmed": true}` — it never restates or binds the arguments. When two events present the same confirmation-call ID but embed different original calls, one approval is ambiguous, and the resume path picks one.

The second hole is exploitable on 1.x today. With the crafted-event test from #1369 run against `v1` unmodified, the attacker's amount executes:

```
--- FAIL: TestRequestConfirmationRejectsForgedAmountFromCraftedEvent
    SECURITY: transfer_funds executed with attacker-forged amount=999, which the
    user never approved (the user approved amount=100). executed amounts=[999 100]
```

## Summary

Backport of #1357 and #1369 to the 1.x maintenance line, in that order — #1369's conflict detection builds on the author gate #1357 adds.

- #1357 rejects confirmations resumed from events this agent did not author.
- #1369 detects a confirmation-call ID that resolves to conflicting original calls across events and refuses to resume it. Failing closed blocks the legitimate call too, which is the right trade when the framework cannot tell which call the user saw.

One adaptation: the `amountEchoTool` helper in the #1369 test takes `agent.ToolContext`, since the unified `agent.Context` is 2.x-only. The source hunks are unchanged.

### DIFF
--- a/internal/llminternal/request_confirmation_processor.go
+++ b/internal/llminternal/request_confirmation_processor.go
@@ -108,6 +108,50 @@ func RequestConfirmationRequestProcessor(ctx agent.InvocationContext, req *model
 
 		agentName := ctx.Agent().Name()
 
+		// Detect tampered confirmations before resuming any tool.
+		//
+		// The resume path below trusts the tool name and arguments embedded in
+		// the agent-authored confirmation-request event (via OriginalCallFrom),
+		// gated only by the author check. The user's approval carries only the
+		// confirmation-call ID and a {"confirmed": true} decision; it does not
+		// restate or bind the arguments. So if two events present the same
+		// confirmation-call ID but embed different original calls, a single user
+		// approval is ambiguous: we cannot tell which call the user actually saw
+		// and approved. An attacker who can append an event stamped with this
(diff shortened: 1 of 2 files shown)


jjsasha63 (approved):
Faithful to #1357 and #1369. One line differs in the whole diff, and it is the
adaptation the description discloses: amountEchoTool.Run takes agent.ToolContext
here where main uses agent.Context. Everything else matches, 375 added and
removed lines on each side.

PR #1500, opened 2026-09-04
fix: backport base flow nil-deref and artifact delta fixes (#1336, #474)

wolo-lab:
## Problem

Two defects in `internal/llminternal/base_flow.go`, one a crash and one silent data loss.

- **Nil dereference on streaming mode.** The flow reads the streaming mode off `RunConfig` without guarding the pointer, so a run configured without it panics.
- **Artifact deltas are dropped when tool calls run in parallel.** [`mergeEventActions`](https://github.com/google/adk-go/blob/cb2981844eae0613a90b100d98c32d231607ac69/internal/llminternal/base_flow.go#L1358) merges `StateDelta` and `RequestedToolConfirmations` but never touches `ArtifactDelta` — it still carries the [`// TODO add similar logic for state`](https://github.com/google/adk-go/blob/cb2981844eae0613a90b100d98c32d231607ac69/internal/llminternal/base_flow.go#L1378) placeholder. Merging keeps the first event's actions as the base, so every artifact delta after the first is discarded and those artifact versions never reach the session.

The artifact-delta loss is worse on 1.x than the form reported upstream: 2.x at least picked a winner among the deltas, whereas 1.x has no merge logic for the field at all.

## Summary

Backport of #1336 and #474 to the 1.x maintenance line, both cherry-picked without adaptation.

#474 merges artifact deltas highest-version-wins rather than first-wins. adk-python uses last-write-wins here, and the upstream PR chose highest-version-wins deliberately because it preserves more information.

### DIFF
--- a/internal/llminternal/base_flow.go
+++ b/internal/llminternal/base_flow.go
@@ -788,7 +788,14 @@ func (f *Flow) callLLM(ctx agent.InvocationContext, req *model.LLMRequest, state
 		// to help with slicing the billing reports on a per-agent basis.
 
 		// TODO: RunLive mode when invocation_context.run_config.support_cfc is true.
-		useStream := runconfig.FromContext(ctx).StreamingMode == runconfig.StreamingModeSSE
+		// Streaming mode comes from the invocation context's RunConfig rather than
+		// the runconfig context value: agent.Run() may be invoked directly (outside
+		// the runner), in which case no runconfig is stored in the Go context and
+		// runconfig.FromContext returns nil (issue #586).
+		useStream := false
+		if rc := ctx.RunConfig(); rc != nil {
+			useStream = rc.StreamingMode == agent.StreamingModeSSE
+		}
 
 		for resp, err := range generateContent(ctx, f.Model, req, useStream) {
 			if err != nil {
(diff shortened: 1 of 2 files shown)


jjsasha63 (approved):
LGTM.

Both production hunks are byte-identical to origin/main. The two commits v1 is
ahead of this branch do not touch internal/llminternal.

I mutation-tested both fixes:

- Reverting the streaming hunk reproduces the SIGSEGV at base_flow.go:795 in the
  new test.
- Deleting the ArtifactDelta merge call, changing max to min, and changing max
  to last-write-wins each fail the table.
- Changing `maps.Clone(src)` to `src` is the one mutation that survives, so
  nothing asserts that base stops sharing a map with other.

On the streaming source: Runner.Run sets both run-config carriers from the same
cfg (runner/runner.go:173 and :204), and Runner.RunLive leaves ctx.RunConfig()
nil but previously read Bidi, which never equalled SSE. useStream is false
either way, so no caller loses streaming.

Parity vs src/google/adk/flows/llm_flows/functions.py at adk-python 2.8.0:

- merge_parallel_function_response_events (functions.py:1766) also takes the
  first event as the base and folds the rest in.
- deep_merge_dicts (functions.py:1756) assigns `d1[key] = value` for scalar
  leaves, so Python is last-write-wins on artifact_delta. The comment's claim
  is accurate. Max-wins is the better choice here: artifact versions are
  monotonic per key, and parallel events are ordered by call rather than by
  save time.
- Python 2.8 also aggregates renderUiWidgets in the same merge, which is newer
  than the 1.x line.

PR #1501, opened 2026-09-04
fix: backport artifact handling fixes (#1390, #1320, #1375)

wolo-lab:
## Problem

Three defects in artifact handling, one of them a panic.

- **`load_artifacts` panics when no artifact service is configured.** [`NewCallbackContextWithArtifactTracking`](https://github.com/google/adk-go/blob/cb2981844eae0613a90b100d98c32d231607ac69/agent/callback_context.go#L54) and [`NewToolContext`](https://github.com/google/adk-go/blob/cb2981844eae0613a90b100d98c32d231607ac69/agent/callback_context.go#L76) both wrap `ic.Artifacts()` in a `trackedArtifacts` unconditionally, so a nil service becomes a non-nil wrapper and the first promoted method call dereferences nil. Against `v1` unmodified the guard test panics rather than returning an error:

```
--- FAIL: TestLoadArtifactsTool_ProcessRequest_NilArtifacts
```

- **`load_artifacts` does nothing when combined with other tool calls.** The processor only inspects one function response, so a turn where `load_artifacts` shares a content with another tool response silently loads nothing.
- **`gcsartifact` does not recognize a wrapped `storage.ErrObjectNotExist`,** turning an absent object into an error rather than a miss.

## Summary

Backport of #1390, #1320 and #1375 to the 1.x maintenance line, applied in upstream order — #1320 builds on the guard #1390 introduces.

One adaptation: the nil guard lives in `agent/callback_context.go`, since `agent/common_context.go` is 2.x-only, and `ProcessRequest` takes `agent.ToolContext` rather than the unified `agent.Context`. The behavior is the same — `newTrackedArtifacts` returns nil for a nil inner service, so "no artifact service configured" stays observable, and the tool reports it as an error.

### DIFF
--- a/agent/callback_context.go
+++ b/agent/callback_context.go
@@ -51,7 +51,7 @@ func NewCallbackContextWithArtifactTracking(ic InvocationContext, actions *sessi
 		Context:           ic,
 		invocationContext: ic,
 		actions:           actions,
-		artifacts:         &trackedArtifacts{Artifacts: ic.Artifacts(), actions: actions},
+		artifacts:         newTrackedArtifacts(ic.Artifacts(), actions),
 	}
 	return cc
 }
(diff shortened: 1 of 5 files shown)


jjsasha63 (approved):
Faithful to #1390, #1320 and #1375. Main applies the nil guard once in the shared
constructor; the port applies it in both v1 constructors that need it,
NewCallbackContextWithArtifactTracking at callback_context.go:51 and
NewToolContext at callback_context.go:73. Everything else matches, 259 added and
removed lines on each side.

PR #1502, opened 2026-09-04
fix: backport session service fixes (#1322, #1204, #1237, #1171)

wolo-lab:
## Problem

Four defects across the two persistent session services.

- **Stale-session errors render 1970 dates.** The timestamps compared are `UnixMicro()` microseconds, and the error formats them through `time.Unix(0, x)`, whose second argument is nanoseconds. Every real timestamp collapses to a 1970-01-2x date, which makes the error useless for diagnosing which handle is stale.
- **`UpdateTime` is not maintained on app and user state writes** in `session/database`, so a state write leaves the recorded update time behind.
- **`userId` is not quoted in the AIP-160 list filter** in `session/vertexai`. A user ID containing a filter metacharacter breaks the query or changes what it matches.
- **Nil function arguments are mishandled** in `session/vertexai` instead of being treated as an empty struct.

The 1970 formatting bug reproduces on `v1` unmodified:

```
--- FAIL: TestDatabaseService_AppendEvent_StaleErrorFormatsTimestamps
    stale error still renders a 1970 timestamp: "stale session error: last update
    time from request (1970-01-21T15:38:02.4Z) is older than in database
    (1970-01-21T15:38:06Z)"
```

## Summary

Backport of #1322, #1204, #1237 and #1171 to the 1.x maintenance line, applied in upstream order because #1204 and #1237 both touch `session/database/service.go`.

All four source changes are cherry-picked unchanged. One test file needed re-anchoring, and the first attempt at it lost coverage:

- #1237's test hunk sits next to a workflow-event round-trip test covering `NodeInfo`, `RequestedInput`, `Routes` and `IsolationScope`, none of which exist on 1.x, so that surrounding test could not come across.
- Resolving that hunk also removed `TestDatabaseService_StateUpdateTimeIsSet`, which the #1204 cherry-pick had added two commits earlier, leaving the #1204 fix with no coverage on this branch. It is restored verbatim from main in the last commit. With all four `UpdateTime` assignments stripped from `session/database/service.go`, that test fails on all four assertions and nothing else in `./session/...` fails.

### DIFF
--- a/session/database/service.go
+++ b/session/database/service.go
@@ -380,8 +380,8 @@ func (s *databaseService) applyEvent(ctx context.Context, session *localSession,
 		if storageUpdateTime > sessionUpdateTime {
 			return fmt.Errorf(
 				"stale session error: last update time from request (%s) is older than in database (%s)",
-				time.Unix(0, sessionUpdateTime).Format(time.RFC3339Nano),
-				time.Unix(0, storageUpdateTime).Format(time.RFC3339Nano),
+				time.UnixMicro(sessionUpdateTime).Format(time.RFC3339Nano),
+				time.UnixMicro(storageUpdateTime).Format(time.RFC3339Nano),
 			)
 		}
 
--- a/session/database/service.go
+++ b/session/database/service.go
@@ -109,12 +109,16 @@ func (s *databaseService) Create(ctx context.Context, req *session.CreateRequest
 		// apply state delta
 		if len(appDelta) > 0 {
 			maps.Copy(storageApp.State, appDelta)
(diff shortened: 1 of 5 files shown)


jjsasha63 (changes requested):
Please restore `TestDatabaseService_StateUpdateTimeIsSet`: the #1204 fix currently ships with no test.

Other than that it's  faithful to #1322, #1204, #1237 and #1171

wolo-lab:
TestDatabaseService_StateUpdateTimeIsSet has been restored. It was indeed missing.

jjsasha63 (approved):
Lgtm

PR #1349, opened 2026-08-17
fix(preloadmemorytool): guard against nil parts in extractText (#1348)

chill-czar:
### Link to Issue or Description of Change

- Closes: #1348

### Problem

In `tool/preloadmemorytool/tool.go`, `extractText` iterated over `mem.Content.Parts` and accessed `part.Text` without checking whether `part` is `nil`. If a memory entry contains nil parts, `extractText` panics with a nil pointer dereference.

### Solution

Add a defensive check `if part == nil || part.Text == ""` in `extractText`, safely skipping `nil` elements, and add a regression test in `tool_test.go`.

### Testing Plan

- [x] Unit tests added:
  - `TestPreloadMemoryTool_ProcessRequest` (case: "memories with nil and empty parts") in `tool/preloadmemorytool/tool_test.go`
- [x] Confirmed tests fail without the fix and pass with it.
- [x] All repository verification commands pass:
  ```text
  go test -v -race -count=1 ./tool/preloadmemorytool/...
  go build -mod=readonly work
  golangci-lint run
  go mod tidy -diff
  ```

### Checklist

- [x] I have read the [CONTRIBUTING.md](./CONTRIBUTING.md) document.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have manually tested my changes end-to-end.
- [x] Any dependent changes have been merged and published in downstream modules.

PR #1503, opened 2026-09-04
fix: backport tool result and content fixes (#696, #1349, #1367)

wolo-lab:
## Problem

Three defects in tool results and request content assembly.

- **A sibling branch can steal the current-turn pivot.** With `IncludeContents: "none"`, the processor walks backwards for the event that starts this agent's turn and accepts any event authored by a user or another agent, without checking the branch. In a parallel delegation a sibling appends its response after this agent's user input, becomes the pivot, and the user's input is sliced out of the request — the agent is asked to answer a question it can no longer see.
- **Thinking parts leak into `agenttool` results,** so a caller receives the sub-agent's reasoning alongside its answer.
- **`preloadmemorytool.extractText` panics on nil parts.**

## Summary

Backport of #696, #1349 and #1367 to the 1.x maintenance line.

Two adaptations, both in #1367. `buildContentsCurrentTurnContextOnly` has no `isolationScope` parameter on 1.x, so the branch check is the only guard added to the pivot loop — 2.x also skips out-of-scope events there, and scope isolation does not exist on this line. The test drops `llmagent.ModeSingleTurn`, which is 2.x-only. `eventBelongsToBranch`, the helper the fix needs, is already present on `v1`.

### DIFF
--- a/internal/llminternal/contents_processor.go
+++ b/internal/llminternal/contents_processor.go
@@ -502,6 +502,13 @@ func buildContentsCurrentTurnContextOnly(agentName, branch string, events []*ses
 	// Find the latest event that starts the current turn and process from there
 	for i := len(events) - 1; i >= 0; i-- {
 		event := events[i]
+		// Events from a sibling branch are not part of this agent's
+		// current turn. In parallel delegations, a sibling may append its
+		// response after this agent's user input; treating that response as
+		// the pivot would slice the input out of the request.
+		if !eventBelongsToBranch(branch, event) {
+			continue
+		}
 		if event.Author == "user" || isOtherAgentReply(agentName, event) {
 			return buildContentsDefault(agentName, branch, events[i:])
 		}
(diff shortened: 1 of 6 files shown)


jjsasha63 (approved):
Approve.

What I checked:

- The hunks against the three originals. #696 and #1349 are byte-identical;
  #1367 differs only by the dropped `Mode: llmagent.ModeSingleTurn`. Both
  adaptations are forced by v1.
- `go build ./...` and the tests for the 3 touched packages on the PR head.
  Both pass.
- Whether the new tests catch the bugs. I reverted each one-line fix in turn,
  and the matching test failed each time.
- Whether v1 needs a version bump. It does not: `internal/version/version.go`
  already reads 1.6.1, and there is no v1.6.1 tag yet.

Parity against adk-python 2.8.0:

- `tools/agent_tool.py:344` filters with `if not p.thought`, which is what the
  agenttool change now does.
- `flows/llm_flows/contents.py:436` runs the pivot loop through
  `_should_include_event_in_context`, and that checks branch membership at
  `:339`. Same guard, same place in the loop.
- `memory/_memory_entry_utils.py` `extract_text` has no nil guard, so the
  preloadmemorytool change is Go-side hardening rather than a parity fix. The
  space separator matches Python's default.

PR #1307, opened 2026-08-11
fix(adkrest): decode chunked create-session request bodies

jesseedcp:
### Link to Issue or Description of Change

**1. Link to an existing issue (if applicable):**

- Closes: #1306

**Problem:**

`CreateSessionHandler` only decoded JSON when `req.ContentLength > 0`. Go reports an unknown request length as `-1`, so HTTP/1.1 chunked requests were accepted while their submitted initial state and events were silently discarded.

**Solution:**

Decode any present request body and treat only `io.EOF` as a valid empty body. Malformed JSON continues to return HTTP 400. The new HTTP integration tests cover populated, empty, and malformed chunked request bodies.

### Testing Plan

**Unit Tests:**

- [x] I have added or updated unit tests for my change.
- [x] All unit tests in the affected packages pass locally.

Passed locally:

```text
go test -race -mod=readonly -count=20 ./server/adkrest -run '^TestCreateSession(PreservesChunkedBody|AllowsEmptyChunkedBody|RejectsMalformedChunkedBody)$'
go test -race -mod=readonly -count=1 -shuffle=on ./server/adkrest/...
go test -race -mod=readonly -count=1 -shuffle=on ./...  # plugin/agentanalytics
go vet ./server/adkrest
```

In a clean LF checkout, full root-module and plugin-module lint both reported `0 issues`, and `go mod tidy -diff` produced no changes in either module.

**Manual End-to-End (E2E) Tests:**

I started an `adkrest.NewServer` backed by `session.InMemoryService`, then sent a real HTTP/1.1 chunked request with curl:

```text
printf '{"state":{"theme":"dark"},"events":[{"id":"event-1","invocationId":"invocation-1","author":"user"}]}' | \
  curl --http1.1 -X POST \
    -H 'Content-Type: application/json' \
    -H 'Transfer-Encoding: chunked' \
    --data-binary @- \
    http://127.0.0.1:18080/apps/test-app/users/test-user/sessions/test-session
```

The relevant response fields preserved both values:

```json
{"events":[{"id":"event-1","invocationId":"invocation-1","author":"user"}],"state":{"theme":"dark"}}
```

The automated regression test additionally verifies on the server side that the request arrives with `ContentLength == -1` and `Transfer-Encoding: chunked`. Before the fix, that test failed with an empty state and zero events; after the fix, all three chunked-body cases pass.

### Checklist

- [x] I have read the [CONTRIBUTING.md](./CONTRIBUTING.md) document.
[...]

PR #1504, opened 2026-09-04
fix: backport runner and REST server fixes (#1417, #1307)

wolo-lab:
## Problem

Two defects, one a panic reachable from any custom agent.

- **A nil event yielded by an agent panics the runner.** Both [`Run`](https://github.com/google/adk-go/blob/cb2981844eae0613a90b100d98c32d231607ac69/runner/runner.go#L256) and [`RunLive`](https://github.com/google/adk-go/blob/cb2981844eae0613a90b100d98c32d231607ac69/runner/runner.go#L480) dereference the event with no nil check. This is worse on 1.x than it was on 2.x before the fix: 2.x guarded the same expression with `event != nil &&`, whereas 1.x has nothing. Against `v1` unmodified an agent yielding one nil event takes the runner down:

```
--- FAIL: TestRunner_NilEventYieldedDoesNotPanic
```

- **`adkrest` fails to decode chunked create-session request bodies,** so a client that sends the body with `Transfer-Encoding: chunked` cannot create a session.

## Summary

Backport of #1417 and #1307 to the 1.x maintenance line.

One adaptation, in #1417. In `RunLive` the guard goes ahead of the transcription-buffering block rather than immediately before the non-partial check, because that block dereferences the event first on 1.x. In `Run` it sits where upstream put it. #1307 is cherry-picked unchanged.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -252,6 +252,10 @@ func (r *Runner) Run(ctx context.Context, userID, sessionID string, msg *genai.C
 				}
 			}
 
+			if event == nil {
+				continue
+			}
+
 			// only commit non-partial event to a session service
 			if !event.LLMResponse.Partial {
 				if err := r.sessionService.AppendEvent(ctx, storedSession, event); err != nil {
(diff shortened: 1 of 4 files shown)


jjsasha63 (changes requested):
Faithful to #1417 and #1307.
The four differences against #1417 are the event != nil inline checks
main removed as redundant. v1 never had them, so there is nothing to remove.

jjsasha63 (approved):
Lgtm

PR #1512, opened 2026-09-05
fix(workflow): suppress retries while draining

### LINKED ISSUE #1511: workflow: panic when retryable sibling fails while scheduler is draining
## 🔴 Required Information

**Describe the Bug:**

When a fan-out node fails and the workflow starts draining, a cancelled sibling can return its own retryable error during cleanup. `scheduler.cancelAll` has already stopped the existing retry timers and assigned `nil` to `retryTimers`, but `handleCompletion` still calls `scheduleRetry` for that sibling. `scheduleRetry` then writes to the nil map and panics.

**Steps to Reproduce:**

1. Create a workflow with two nodes downstream of `Start`.
2. Make the first node fail immediately.
3. Give the second node a `RetryConfig`; wait for `ctx.Done()` and then return a non-cancellation error.
4. Run the workflow.

The regression test in the accompanying PR provides a complete, deterministic reproduction.

**Expected Behavior:**

Once the scheduler is draining, it should cancel and reap in-flight nodes without scheduling retries. The workflow should return the original branch failure.

**Observed Behavior:**

The scheduler panics:

```text
panic: assignment to entry in nil map

google.golang.org/adk/v2/workflow.(*scheduler).scheduleRetry(...)
    workflow/scheduler.go:420
google.golang.org/adk/v2/workflow.(*scheduler).handleCompletion(..., 0x0)
    workflow/scheduler.go:866
google.golang.org/adk/v2/workflow.(*scheduler).run(...)
    workflow/scheduler.go:603
```

The `0x0` argument in `handleCompletion` is `scheduleSuccessors=false`, confirming that the retry is scheduled after draining has begun.

**Environment Details:**

- ADK Library Version: v2.3.0 and current `main` at `a611472`
- OS: macOS arm64
- Go Version: reproduced with Go 1.27.0; regression test passes with the repository's Go 1.26.6 toolchain

**Model Information:**

- N/A; this is isolated with deterministic workflow nodes and no model calls.

---

## 🟡 Optional Information

**Regression:**

Unknown.

**Additional Context:**

Reinitializing `retryTimers` would avoid only the panic. It would still register a new timer after `cancelAll`, potentially keeping the draining loop alive until a long retry delay expires. The correct behavior is to suppress all retry scheduling once the scheduler is draining.

The accompanying test `TestScheduler_DrainingDoesNotRetryFailedSibling` uses an immediate primary failure and a retryable sibling that returns its own error after observing cancellation. It reproduces the panic on current `main` every time.

- Always (100%)

cruvie:

- Closes: #1511

**Problem:**

When one fan-out branch fails, the scheduler enters draining mode and `cancelAll` clears `retryTimers`. If a cancelled sibling then returns its own retryable error, `handleCompletion` still attempts to schedule a retry. This writes to the nil `retryTimers` map and panics. Reinitializing the map alone would still leave an uncancelled retry timer that could delay shutdown.

**Solution:**

Treat the existing completion flag as permission to schedule any new work, not only successors. While draining, failed siblings are reaped and marked failed without scheduling retries. The flag was renamed from `scheduleSuccessors` to `scheduleNewWork` to document that invariant.

### Behavior change

After a workflow begins draining because a node failed, retryable errors returned by cancelled siblings no longer panic or schedule delayed retries. The workflow terminates with the original failure.

### Testing Plan

**Unit Tests:**

- Added `TestScheduler_DrainingDoesNotRetryFailedSibling`.
- With the source change reverted and the test retained, `TestScheduler_DrainingDoesNotRetryFailedSibling` fails 100% with `panic: assignment to entry in nil map` from `scheduleRetry`.
- `golangci-lint run` in the root module: `0 issues.`
- `golangci-lint run` in `plugin/agentanalytics`: `0 issues.`
- `go mod tidy -diff` in both modules: clean.

N/A. The failure is isolated in the workflow scheduler with deterministic nodes and no model or network calls.

### Checklist

- [x] I have manually reproduced the failure before applying the fix.
- [x] Any dependent changes have been merged and published in downstream modules, or are not applicable.

### Additional context

The issue was first observed in a nested fan-out/fan-in workflow where one material-processing branch exhausted its retries while two siblings were still running. The regression test reduces that scenario to two deterministic nodes.

### DIFF
--- a/workflow/scheduler.go
+++ b/workflow/scheduler.go
@@ -789,10 +789,10 @@ func (s *scheduler) handleEvent(it eventItem) {
 }
 
 // handleCompletion finalises a node's run: transitions its lifecycle
-// status, removes the live task, and (if scheduleSuccessors is true)
-// schedules its successors. When the consumer is draining (caller
-// stopped or a node failed), pass scheduleSuccessors=false so the
-// workflow does not keep dispatching new nodes after cancellation.
+// status, removes the live task, and (if scheduleNewWork is true)
+// schedules retries or successors. When the consumer is draining (caller
+// stopped or a node failed), pass scheduleNewWork=false so the workflow
+// does not keep dispatching new nodes after cancellation.
 //
 // The returned error is the node's own error (NodeFailed); nil on
 // clean success or sibling cancellation.
@@ -813,7 +813,7 @@ func (s *scheduler) handleEvent(it eventItem) {
 // context cancel, multiple-output, multiple-routing-event,
 // multiple-input-request) does not silently park in NodeWaiting:
 // failures take precedence and surface as NodeFailed.
-func (s *scheduler) handleCompletion(it completionItem, scheduleSuccessors bool) error {
+func (s *scheduler) handleCompletion(it completionItem, scheduleNewWork bool) error {
 	ns := s.state.EnsureNode(it.nodeName)
 	nr := s.runsByName[it.nodeName]
 	// For retryable nodes still delete them from run variables. If the node is retried,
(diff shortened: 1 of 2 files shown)


wojtas000 (approved):
LGTM. Verified locally: reverting the `scheduleNewWork &&` guard reproduces the #1511 panic exactly (nil map at scheduler.go:420), and the new test catches it.

PR #1514, opened 2026-09-05
fix(session/vertexai): strip temp: keys before AppendEvent persist

### LINKED ISSUE #1410: temp: state reaches storage on session/vertexai, and the conformance suite asserts on the wrong surface to catch it

## 🔴 Required Information

**Describe the Bug:**

Two linked problems. The second is why the first is still open.

**1. `temp:` state reaches storage on `session/vertexai`.**

`vertexAiService.AppendEvent` (`session/vertexai/vertexai.go:129`) sends the event to the backend before any trimming:

```go
err := s.client.appendEvent(ctx, sess.AppName(), sess.ID(), event)
...
err = sessInt.appendEvent(event)
```

`localSession.appendEvent` does trim (`session/vertexai/session.go:82`), but that only affects the in-process copy. `vertexAiClient.appendEvent` serialises `event.Actions.StateDelta` through `createAiplatformpbEventActions` exactly as given, so the `temp:`-prefixed keys are persisted. A later `Get()` reads them back.

This is the same shape as #1354 on the in-memory service: the trim is applied to the local copy, and the untrimmed value is what reaches storage. `session/database` does not have it.

The documented contract on `session.KeyPrefixTemp` is that these entries are "specific to the current invocation ... Discarded after the invocation completes."

**2. `sessiontestsuite` cannot see either leak.**

`session/sessiontestsuite/service_suite.go` has a case named `temp_state_is_not_persisted`, which is exactly the invariant. It asserts against session **state**:

```go
snap := Snapshot(got.Session)
if _, exists := snap.State["temp:k1"]; exists {
    t.Errorf("Temp state leaked to persist step, got: %v", snap.State)
}
```

Session state is never where the leak lands. `sessionutils.ExtractStateDeltas` already drops `temp:` keys before anything is written to app, user or session state, so that assertion passes on every backend whatever the event does. The leak lands in the persisted **event's** `Actions.StateDelta`, which the case never looks at.

That is why the v1.6.0 regression in #1354 shipped with a green suite, and why the Vertex AI instance of the same bug has gone unreported.

**Steps to Reproduce:**

Add three lines to the existing `temp_state_is_not_persisted` case in `sessiontestsuite`, after the current assertions:

```go
var stored *session.Event
for ev := range got.Session.Events().All() {
    stored = ev
}
if _, exists := stored.Actions.StateDelta["temp:k1"]; exists {
    t.Errorf("temp:k1 leaked into the stored event delta: %v", stored.Actions.StateDelta)
}
```

Then run the three suite entry points.

**Expected Behavior:**

`temp:`-prefixed keys are absent from the persisted event on every backend, and the conformance suite fails any backend where they are not.

**Observed Behavior:**

On `main` at `681ac5d`:

```
=== ./session/ Test_inMemoryService
--- FAIL: Test_inMemoryService/StateManagement/temp_state_is_not_persisted
    temp:k1 leaked into the stored event delta: map[sk:v2 temp:k1:v1]

=== ./session/database/ Test_databaseService
ok

=== ./session/vertexai/ Test_vertexaiService
--- FAIL: Test_vertexaiService/StateManagement/temp_state_is_not_persisted
    temp:k1 leaked into the stored event delta: map[sk:v2 temp:k1:v1]
```

Two of three backends leak. The suite is green on all three today.

**Environment Details:**

- ADK Library Version: `main` at `681ac5d`
- OS: macOS
- Go Version: 1.26.6

**Model Information:**

- N/A, no model involved.

---

## 🟡 Optional Information

**Regression:**

For in-memory, yes, at v1.6.0, tracked in #1354 with a fix in #1356. For `session/vertexai`, not established; the trim has been local-copy-only for as long as the split has existed.

**Additional Context:**

The in-memory half is already covered by #1354 and #1356 and is deliberately not part of this report. What is new here is the Vertex AI instance and the reason the suite missed both.

The fix for the Vertex AI half looks like one line, trimming before the call rather than after it:

```go
// temp: keys are scoped to the invocation and must not reach storage. The
// local session still receives the untrimmed event, which is what keeps them
// readable for the rest of the invocation.
err := s.client.appendEvent(ctx, sess.AppName(), sess.ID(), trimTempDeltaState(event))
```

I have not opened a PR for it, for a reason worth stating: `internal/httprr` keys replay on the full request body (`rr.go`, `reqWire`), so removing keys from the outgoing payload changes the key and the affected `testdata/*.httprr` cassettes need re-recording against a live endpoint. I do not have credentials for that. Happy to write the change and the tests if a maintainer can re-record, or to send it as-is for someone with access to record on top.

The suite hardening should land with or after #1356, since the new assertion fails for in-memory until that merges.

- Always (100%)

leilei3167:
## Summary

`session/vertexai` was persisting `temp:`-prefixed `StateDelta` keys as-is: `AppendEvent` called the backend with the untrimmed event, and only the local session copy was trimmed afterward. A later `Get()` could read those keys back from the stored event.

The shared `sessiontestsuite` case `temp_state_is_not_persisted` only asserted on session **state**, which `ExtractStateDeltas` already filters — so the suite stayed green while the event-delta leak remained. This hardens that case to also assert the stored event's `Actions.StateDelta` has no `temp:` keys (and keeps non-temp keys).

### Changes
- `session/vertexai`: trim with `trimTempDeltaState` **before** `client.appendEvent`; still pass the original event to the local session so temp stays readable for the rest of the invocation (same shape as database).
- `session/sessiontestsuite`: assert persisted event delta has no `temp:` keys.
- Offline pin: `TestAppendEvent_stripsTempKeysBeforePersist` (fake backend; no network).
- `casesWithoutReplay`: skip the recorded suite case until a maintainer re-records Agent Engine traffic (`UPDATE_REPLAYS=true`) — outbound RPC body changes, so existing cassettes no longer match. **Does not block the code fix.**
- `session/inmemory`: temporary store-path trim so the shared suite assertion stays green until #1356 lands; will drop that hunk after rebasing onto main once #1356 merges.

Fixes #1410

## Relation to #1356
#1356 fixes the inmemory canonical-record leak for #1354; this PR covers vertexai + suite. Landing order: merge #1356 first, then rebase this and remove the inmemory hunk.

## Test plan
- [x] `go test ./session/vertexai/ -count=1`
- [x] `go test ./session/database/ -count=1`
- [x] `go test ./session/ -count=1`
- [x] `TestAppendEvent_stripsTempKeysBeforePersist` offline
- [ ] Maintainer: re-record `Test_vertexaiService_StateManagement_temp_state_is_not_persisted` with `UPDATE_REPLAYS=true` when convenient, then remove it from `casesWithoutReplay`


wojtas000 (commented):
Please drop the PR-process narration at `session/inmemory.go:237-238`; it goes stale the moment #1356 closes.

**Landing order: let's merge #1356 first, then rebase this onto main and drop the inmemory hunk.** 

Other than that the fix is right.

- Mutation-tested four ways, all caught: untrimmed event at `vertexai.go:146` (fails `append_event_temp_state_test.go:73`), clone from `event` at `inmemory.go:249` (`service_suite.go:1064`), trimming the local session too (`:83`), over-trimming (`:68`). No survivors. Build, race tests, lint and tidy clean in both modules.
- The replay skip is unavoidable. rpcreplay keys on the request proto, and the recorded cassette still holds the buggy request with `temp:k1` in it. Needs a re-record by someone with Agent Engine access, and shouldn't outlive the PR.
- Parity vs adk-python 2.8.0: `vertex_ai_session_service.py:394` trims via `super().append_event()` before building the outbound config, and the base trim mutates in place (`base_session_service.py:214`). Go's returns a copy since #1128, which is why this had to be wired explicitly. Same ordering.
- Not a break: `session.KeyPrefixTemp` documents the discard (`session/session.go:424`) and `session/database/service.go:386` already trimmed before persisting. Worth a release note, since adkrest serves `StateDelta` verbatim.

Nits, not blocking: `trimTempDeltaState` now runs twice per `AppendEvent` on both paths; nothing asserts the `raw_event` copy of the delta.

leilei3167:
Thanks — dropped the process narration on the inmemory store-path trim.

Agreed on landing order: once #1356 is in main I'll rebase and remove the inmemory hunk entirely so this PR stays vertexai + suite only.

wojtas000:
@leilei3167 
We merged #1356. Could you please adapt this PR for the re-review?

leilei3167:
@wojtas000 Rebased onto main after #1356, dropped the inmemory hunk, and aligned vertexai AppendEvent with the same temp: strip-before-persist rule (caller event untouched; local session still gets the original). Suite now asserts the stored event delta too; offline pin covers the replay case until cassettes are re-recorded.

wojtas000:
One suggestion:
Please add a test case asserting that no `temp:` key reaches the `raw_event` copy of the state delta.

AppendEvent stores that delta twice: once in `stateDelta`, once inside the `raw_event` blob. On read-back, `raw_event` takes precedence at vertexai_client.go line 443.

The fix cleans both. The test only checks `stateDelta`, and the test event sets none of the fields `eventNeedsRawEvent` looks at, so `raw_event` is never populated at all.

Here is what that costs. Moving the trim into `createAiplatformpbEventActions` leaves `temp:k1` in `raw_event`, and the test still passes:

```go
event := &session.Event{
	ID: "event2", Author: "user", InvocationID: "inv1",
	Output: map[string]any{"node": "out"}, // forces raw_event
	Actions: session.EventActions{StateDelta: map[string]any{"temp:k1": "v1", "sk": "v2"}},
}
// ... AppendEvent ...
blob := fake.events[0].GetRawEvent().AsMap()
actions, _ := blob["actions"].(map[string]any)
delta, _ := actions["stateDelta"].(map[string]any)
for k := range delta {
	if strings.HasPrefix(k, session.KeyPrefixTemp) {
		t.Errorf("temp key reached raw_event: key %q, delta %v", k, delta)
	}
}
```

This passes on your head commit and fails on that variant.

leilei3167:
@wojtas000 Added an offline pin that forces raw_event (via Output) and asserts the blob copy of stateDelta has no temp: keys either. Tip a50eeb3.

wojtas000 (approved):
LGTM thanks

PR #1518, opened 2026-09-07
Fix/569 memory punctuation search

tejas5038:

- Closes: #_issue_number_
- Related: #_issue_number_

possible._

**Problem:**
_A clear and concise description of what the problem is._

**Solution:**
_A clear and concise description of what you want to happen and why you choose
this solution._

### Behavior change

**What behaves differently for someone already on the current release?**

_Write "nothing" if that is the answer. Bug fixes count: a fix that alters an
observable result is still a behavior change._

### Testing Plan

**Unit Tests:**

_Name it. If nothing fails, the tests do not yet cover the change. This is the
cheapest moment to find that out._

### Checklist

### Additional context

### DIFF
--- a/memory/inmemory.go
+++ b/memory/inmemory.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode"
 
 	"google.golang.org/genai"
 
@@ -167,11 +168,14 @@ func checkMapsIntersect(m1, m2 map[string]struct{}) bool {
 func extractWords(text string) map[string]struct{} {
 	res := make(map[string]struct{})
 
-	for s := range strings.SplitSeq(text, " ") {
-		if s == "" {
+	for _, word := range strings.Fields(text) {
+		cleaned := strings.TrimFunc(word, func(r rune) bool {
+			return !unicode.IsLetter(r) && !unicode.IsNumber(r)
+		})
+		if cleaned == "" {
 			continue
 		}
-		res[strings.ToLower(s)] = struct{}{}
+		res[strings.ToLower(cleaned)] = struct{}{}
 	}
 
 	return res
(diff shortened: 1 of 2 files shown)


wojtas000:
Thank you for this, and for including a test — the diagnosis is right, and
`extractWords` splitting on a literal `" "` is genuinely broken.

This overlaps #1403, opened on 25 Aug, which fixes the same issue (#569) in the
same function and is already under review. Rather than have two people carry the
same fix, we're going with that one.

Beyond being first, #1403's tokenization matches adk-python exactly, and that
matters here — adk-python is the source of truth for feature behavior in this
repo. Its in-memory service tokenizes with:

```python
# src/google/adk/memory/in_memory_memory_service.py
def _extract_words_lower(text: str) -> set[str]:
  """Extracts Unicode-aware tokens from a string in lowercase."""
  return set(word.lower() for word in re.findall(r'\w+', text))
```

Python's `\w` is Unicode-aware and includes the underscore. `strings.Fields`
combined with `strings.TrimFunc` only strips punctuation from the *edges* of
whitespace-delimited tokens, so three cases still diverge:

| Input | adk-python `\w+` | This PR |
|---|---|---|
| `state-of-the-art` | `{state, of, the, art}` | `{state-of-the-art}` |
| `foo,bar` | `{foo, bar}` | `{foo,bar}` |
| `_private` | `{_private}` | `{private}` |

`strings.FieldsFunc` with a "not a word rune" predicate, as #1403 does, gets all
three right.

Closing in favour of #1403. Please don't let this put you off — the analysis was
sound and the bug was real, we just had a duplicate already in flight.

If you'd like another crack at this area, the same file has a larger divergence
from adk-python that nobody is working on: #1526. Gap 1 in particular
(relevance scoring and a result cap) is self-contained and has a clear
reference implementation to follow.

PR #1519, opened 2026-09-07
fix(auth): keep the consent URI out of ConsentRequiredError's message

wolo-lab:
## Problem

A `*ConsentRequiredError` becomes a tool's error, which is fed back to the model and persisted as a session event. Its message carried `AuthURI`, and that URI carries the `state` and the consent nonce that bind the credential. A 3-legged authorization URI also normally carries the acting user in a `login_hint` parameter, so the message published an end-user identifier as well.

## Summary

`Error()` no longer prints the URI. It stays on the `AuthURI` field, which is how consumers already read it — the only assertion on the message text anywhere in the tree was the one updated here.

`AuthURI` and `Nonce` now say on the field that they go to the acting user and nowhere else: not a log, not a span, and not through `json.Marshal`, which publishes them whatever `Error()` returns. `AuthURI` is deliberately not redacted the way service text is in other errors, because the identifier in it is what makes the URL work.

`apidiff` reports the change as additive, since no signature moved and the message text is not part of what it compares.

---

First of a five-part split of #1173. Nothing is stacked below this one.

### DIFF
--- a/auth/providers_test.go
+++ b/auth/providers_test.go
@@ -110,8 +110,16 @@ func TestConsentRequiredError(t *testing.T) {
 	if consent.AuthURI != "https://consent.example" {
 		t.Errorf("AuthURI = %q, want %q", consent.AuthURI, "https://consent.example")
 	}
-	if !strings.Contains(err.Error(), "consent.example") {
-		t.Errorf("Error() = %q, want it to mention the auth URI", err.Error())
+	// The message must not carry the URI: it becomes a tool's error, which reaches
+	// the model and the session store, and the URI carries the state and nonce
+	// that bind the credential. Asserted exactly, not just by absence of the URI —
+	// the message has no branches, so a negative check passes for an empty string
+	// too, and this is the only assertion on Error() anywhere.
+	if got, want := consent.Error(), "auth: interactive consent required"; got != want {
+		t.Errorf("Error() = %q, want %q", got, want)
+	}
+	if strings.Contains(err.Error(), "consent.example") {
+		t.Errorf("Error() = %q, want the auth URI kept on the field only", err.Error())
 	}
 }
 
--- a/auth/gcp/client_test.go
+++ b/auth/gcp/client_test.go
@@ -173,8 +173,11 @@ func TestRetrieveCredential(t *testing.T) {
 				if !errors.As(err, &consent) {
 					t.Fatalf("error = %v, want *auth.ConsentRequiredError", err)
 				}
+				// Print the fields, not consent: %v on a *ConsentRequiredError
+				// goes through Error(), which reports neither of them.
 				if consent.AuthURI != tc.wantConsent[0] || consent.Nonce != tc.wantConsent[1] {
-					t.Errorf("consent = %+v, want {authURI:%q nonce:%q}", consent, tc.wantConsent[0], tc.wantConsent[1])
+					t.Errorf("consent = {authURI:%q nonce:%q}, want {authURI:%q nonce:%q}",
+						consent.AuthURI, consent.Nonce, tc.wantConsent[0], tc.wantConsent[1])
 				}
 			case tc.wantErrIs != nil:
 				if !errors.Is(err, tc.wantErrIs) {
--- a/auth/providers.go
+++ b/auth/providers.go
@@ -56,16 +56,28 @@ func (f ProviderFunc) Credential(ctx context.Context) (Credential, error) {
 // before a credential can be issued. Consumers detect it with errors.As.
 type ConsentRequiredError struct {
 	// AuthURI is the URL the end user must visit to grant consent.
+	//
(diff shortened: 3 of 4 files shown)


auth/providers_test.go
```diff
@@ -110,8 +110,16 @@ func TestConsentRequiredError(t *testing.T) {
...
-		t.Errorf("Error() = %q, want it to mention the auth URI", err.Error())
+	// The message must not carry the URI: it becomes a tool's error, which reaches
+	// the model and the session store, and the URI carries the state and nonce
+	// that bind the credential. Asserted exactly, not just by absence of the URI —
+	// the message has no branches, so a negative check passes for an empty string
+	// too, and this is the only assertion on Error() anywhere.
+	if got, want := consent.Error(), "auth: interactive consent required"; got != want {
+		t.Errorf("Error() = %q, want %q", got, want)
+	}
+	if strings.Contains(err.Error(), "consent.example") {
```
karolpiotrowicz, line 121: This can't fire unless the exact-equality check three lines up already did. `errors.As` at :107 binds `consent` to the same object as `err`, so `consent.Error()` and `err.Error()` are the same call, and any string containing `consent.example` already fails the `got != want` on :118.

One argument for keeping it anyway, which is why I'm not pushing: it names the *secret* rather than the *message*, so it survives someone "fixing" a golden-string test by pasting the new actual output into `want`. Your call — I'd just note that the comment above sells :118 as the reason not to rely on absence, and then relies on absence.

Worth knowing either way: this check would not catch a `Nonce` leak. The fixture nonce is `"n"`, so only the URI substring is tested. :118 does catch it.

karolpiotrowicz (approved):
Nothing blocks. The change does what it says and the evidence for it is unusually tight — one nit inline, and two merge preconditions that are not about the code.

The premise checks out end to end. [base_flow.go:1416](https://github.com/google/adk-go/blob/f03f973d822c346bf0ae44da6e6215652e830212/internal/llminternal/base_flow.go#L1416) turns a tool error into `map[string]any{"error": err.Error()}`, and that map is the `FunctionResponse` payload the model reads and the session store keeps, so the message really was the leak channel. Worth stating precisely, though: the producer side, `auth/gcp` [client.go:215](https://github.com/google/adk-go/blob/f03f973d822c346bf0ae44da6e6215652e830212/auth/gcp/client.go#L215), has no non-test caller anywhere in this repository, so the two halves are only joined by a downstream integrator wiring a provider through `mcptoolset.Config.Auth`. Both halves are real. The join is out of tree, which is the normal shape for a library and does not weaken the fix.

Applying only your three test files onto the merge-base fails on the URI in two independent places — the direct message and the `%w` wrap at [transport.go:61](https://github.com/google/adk-go/blob/f03f973d822c346bf0ae44da6e6215652e830212/auth/transport.go#L61) — and both pass with the production change. Those are the only two observable message strings in the repo, and exact equality pins both. `apidiff` is empty on `./auth` and on `./auth/gcp`, against the same base CI will compute.

Three things I looked at hard and came away satisfied with, since a reviewer is more likely to argue the other way on each:

- **Exact equality over a negative check** is right for a stronger reason than the comment gives. A URI-substring check would not catch a `Nonce` leak at all — the fixture nonce is `"n"`. The exact check catches a nonce leak, a key leak, and any field added later.
- **Documenting the `json.Marshal` residue instead of enforcing it** looks like the only in-scope option. Unexporting the fields breaks every consumer reading them off `errors.As`, and a redacting `MarshalJSON` would also close the resume path for anything round-tripping the error. Nothing in the tree marshals it today: no `log/slog` anywhere, every `%#v` site is a test, and all four writers of `map[string]any{"error": …}` go through `err.Error()` first.
- **Not redacting `AuthURI` the way service text is redacted** reads as coherent doctrine rather than an inconsistency. The other errors sanitize text that is *hostile* — `%q` plus a length cap. This one is *sensitive*, and omission is the stronger treatment.

Two preconditions before merge, neither a code issue: the branch is behind `main`, and it conflicts with open #1465 on `auth/gcp/client_test.go`. Merge order decides who rebases.

On the sibling-implementation question: adk-python has no counterpart error. It signals consent by *returning* an `AuthCredential` that carries `auth_uri` rather than by raising, so there is no message there that could leak the URI — [auth_credential.py:109](https://github.com/google/adk-python/blob/b0180620f4c2f4f4467a89c37a30f75bf849700b/src/google/adk/auth/auth_credential.py#L109), at `b0180620`. One detail from the same file that may interest you: on that model `client_secret` and `private_key` are declared `Annotated[..., Field(repr=False)]` and `auth_uri` is not, so the URI is not treated as secret-bearing there. Nothing for this PR to do about it — it just means the doctrine you are setting here is new rather than inherited.

auth/providers_test.go
wolo-lab: Ack.

PR #1520, opened 2026-09-07
fix(auth/gcp): keep caller identifiers out of service error text

wolo-lab:
## Problem

The credential services echo request values back in their error bodies and in an `Operation`'s error message. That text became an `*APIError` body or a wrapped error, which reaches the model as a tool's error and is persisted as a session event — so the acting user's id, commonly an email, and the `ContinueURI` travelled with it.

Scrubbing the raw text is not enough, because the service controls how the identifier is spelled. A body carrying it as `\uXXXX`, as a surrogate pair, as `\/`, behind a doubled backslash, or in a different case all passed a plain substring check.

Where the kilobyte cap sits relative to the scrub is not a detail either, and both of the obvious orders leak in opposite directions. Cutting the scrubbed text lets redaction's own shortening pull bytes into view that the same cut on the raw response would have hidden: a body echoing the `ContinueURI` thirty-four times and then naming the acting user past the kilobyte mark collapses every echo to a ten-byte marker and promotes the id into the window. Cutting the raw text instead slices an occurrence straddling the cut in half, and half an identifier matches nothing, so its head is copied straight out.

Matching a caller-supplied value against a megabyte of response has no natural ceiling either. `Request.UserID` and `Request.ContinueURI` are bounded by nothing here, the scan is O(body × value), and decoding is quadratic on an escape that re-forms its own introducer.

Separately, `RetrieveCredential` accepted a resource name that path normalization could turn into a different one after it had been validated, and rejected the colon that domain-scoped project ids use.

## Summary

Decode service-controlled text to a fixpoint first, then scrub the decoded copy over every occurrence in one pass, merging overlapping ranges. The decision comes from the scrubbed output rather than from the inputs, so a decoy occurrence the decode destroys cannot move it. The marker this scrub writes is kept out of that search by position rather than by splitting the text on it: an occurrence belongs to us only when every byte of it came from a marker. Splitting excluded whole occurrences instead of marker bytes, so a value containing or abutting the marker spanned a split and was matched by nothing — measured over 432 such bodies, 38 leaked and none does now, with none newly withheld.

The cap bounds what may be emitted, not what may be read. Matching runs over the whole response, so an occurrence straddling the cut is removed whole rather than sliced, and no byte at or past the cut is shown either way. Where the cut fell, the result is judged against the whole text as well, because a spelling the scrub could not match can be sliced in half and the visible part alone then looks clean.

Cost is bounded at both ends, and the ceiling is new work rather than a regression — the base scrubs nothing, so none of this cost existed before the scrub did. Each value's lowered and decoded forms are built once rather than once per marker-separated part, a count the service picks by writing `[redacted]` into its own body. A value longer than 4 KiB is not matched at all and nothing is shown, which fails closed. An unbounded version of this scrub spent 3m33.6s of uncancellable CPU on a single `RetrieveCredential` call, and an unbounded fixpoint decode over the body spent over nine minutes on another. As it stands the first is refused in 2 ms and the second is 35 ms. A self-overlapping 4 KiB value against a response at the 1 MiB read cap measures 38 ms, and it is not the worst case: `strings.ToLower` expands an invalid UTF-8 byte to a three-byte U+FFFD, and the 4 KiB bound is tested on the raw value while matching runs on the lowered copy, so the same shape built from invalid UTF-8 measures 230 ms. Both sit well inside the suite's deliberately loose five-second budget. The body is bounded on both counts: the decode stops at 64 passes and reports it, and the whole-text check reads the visible window plus room for one straddling occurrence rather than the whole megabyte. Past that window the response is withheld rather than examined in part, because the escaped spelling's length is the service's choice and no fixed window bounds it — a character at nesting level k costs `2^(k-1)+5` bytes and is still reconstructed inside the 64 passes. What is returned is capped as well as what is read, since one matched run becomes a ten-byte marker however short the run was and the service picks how many runs there are.

Resource-name validation now rejects an empty, `.` or `..` path segment, and allows the colon in `projects/example.com:my-project/...`. The resource is named once in a deferred wrap instead of being repeated in each arm.

The scrub is settled by exhaustion rather than by examples: a differential test against an obviously-correct reference implementation, run at every possible cut, plus a fuzz target. The wiring is pinned separately, because a scrub that is never called passes all of that. Without the tests added here, two of the four sites carrying service text can be reverted to the code they replaced, and `ContinueURI` can be dropped from all four, and the package stays green.

The line count reads worse than the change is. Tests are 1349 of the 2022 insertions, and of the remaining 673, some 500 are `redact.go` — the second half of `client.go` moved out unchanged, because that file had reached 961 lines and over half of them were string scanning rather than an HTTP client. That leaves under 200 lines of genuinely new production code. The scrub's entry point is `redactedForError` rather than `serviceText`, since what it returns is the one thing that is not the service's text.

### Open, and deliberately left out of this PR

- The charset and the segment check both change what `RetrieveCredential` accepts relative to v2.3.0, in opposite directions: five shapes it accepted are now rejected, four it rejected are now accepted. The rejections are all names path normalization would have moved. Worth a release note rather than a code change.
- The deferred wrap that names the resource breaks every direct comparison past validation, not just one: `err.(*APIError)`, `err == ErrConsentRejected`, `err == ErrPollTimeout`, `err == context.Canceled` and `err.(*auth.ConsentRequiredError)` all stop matching, while `errors.Is` and `errors.As` keep working and are the documented style. Separately, `errors.As` no longer finds a `*json.SyntaxError` behind a decode failure. All deliberate, and none visible to `apidiff`, which compares exported shape rather than behaviour. `auth/gcp` shipped in v2.3.0, so these are breaks against a released version with real callers rather than churn inside a new package, and `RetrieveCredential`'s own doc now names them. Release note.
- A response longer than the whole-text check's window is withheld whenever the kilobyte cap cut it, so a very large error body now yields the withheld sentence instead of its first kilobyte. The status code and the resource survive either way. Release note.
- The second candidate decodes one pass rather than to a fixpoint, so a doubly escaped echo is withheld where it could have been shown scrubbed.
- `recoverable` decodes and then folds case, never the reverse, so `\U0040` evades it. No JSON encoder emits that spelling, and `encoding/json` rejects it. Named in the exported paragraph a caller reads, alongside the percent-encoded case.
- An identifier that fits entirely inside the text of a `[redacted]` marker is excused from the recoverability search, whether the marker is one this scrub wrote or one the service wrote into its own body. That is the case the exclusion exists for — a one-character user id of `"e"` occurs inside `[redacted]`, and counting it would suppress every response that redacted anything — so the excused set is exactly the marker's own substrings. An identifier that merely spans a marker is no longer excused.
- `FuzzRedact` skips its leak property for a secret not literally present in the body, which is exactly the escaped-spelling class this change is about.
- Where the `UserID` is a substring of the `ContinueURI` and the service escapes the URI, the `UserID`'s marker lands inside the URI and the rest of it stays visible — 31 of a 34-character URI in the smallest case. The stated contract holds, since this package's decoder does not reconstruct the URI from what comes back, and the paragraph a caller reads now says so. Closing it needs a predicate tolerating a marker-sized gap inside an occurrence, which owes its own exhaustive test: #1539.

---

Second of a five-part split of #1173. Stacked on 1519.


auth/gcp/redact.go
```diff
@@ -0,0 +1,531 @@
...
+
+// decodeFully applies unescapeJSON until the text stops changing.
+//
+// One pass is not enough. A doubly escaped identifier decodes to a singly escaped
+// one, which still hides it from a substring scrub and which this same function
+// will happily decode the rest of the way for anyone who asks twice — so checking
+// only the first pass leaves an identifier the package's own decoder recovers.
+//
+// It terminates because every branch that changes anything writes fewer bytes than
+// it consumed, so a pass that changes the text strictly shortens it. TestDecodeFully
```
karolpiotrowicz: `TestDecodeFully` doesn't exist. The nearest test is `TestDecodeFullyShrinksWheneverItChanges` in `unescape_test.go`, and it calls `unescapeJSON` rather than `decodeFully`, so it pins the per-pass shrinkage lemma over ten hand-picked inputs rather than the loop. The lemma is the right one and I checked it holds on every branch, so this is just the pointer.

auth/gcp/client.go
```diff
@@ -50,15 +49,45 @@ const (
...
 // resourceNameRE bounds a resource name to the characters GCP resource names
-// use, so it can't inject extra path segments, a query, or a fragment into the
-// request URL it is interpolated into. A separate ".." check blocks path
-// traversal (dots are allowed so domain-style ids still pass).
-var resourceNameRE = regexp.MustCompile(`^[A-Za-z0-9._~/-]+$`)
+// use. It cannot inject a query, a fragment, an authority or a percent-escape
+// into the request URL the name is interpolated into. Extra path segments are
+// allowed, since a resource name is itself a path. The colon is allowed for
+// domain-scoped project ids (projects/example.com:my-project/...) — the name is
+// always appended after the endpoint and a /v1 segment, so it can never be read
```
karolpiotrowicz: `retrieveConnector` appends the name after `/v1alpha`, not `/v1`. The substantive argument is unaffected — a colon after a path segment still can't be read as a scheme — but the connector path is the one a reader is most likely to go and check.

auth/gcp/client.go
```diff
@@ -50,15 +49,45 @@ const (
...
+}
 
 // Sentinel errors from [Client.RetrieveCredential]; callers test with errors.Is.
 var (
 	// ErrConsentRejected means the end user rejected the consent request.
 	ErrConsentRejected = errors.New("gcp: user consent rejected")
+	// ErrMalformedResponse means a 2xx body failed JSON decoding — that arm and no
+	// other. A 2xx that overruns the 1 MiB cap before the decoder sees it, or that
+	// decodes cleanly but names an empty or unusable header, returns a plain error.
+	// So this replaces a *json.SyntaxError check rather than answering the broader
```
karolpiotrowicz, line 83: The arm this replaced wrapped every `encoding/json` error, not only `*json.SyntaxError`. `TestDecodeErrorScrubsTheActingUser` asserts on "cannot unmarshal number", which comes from `*json.UnmarshalTypeError` — so the package demonstrates a second `errors.As` target that stops working, in the test it ships. Worth widening both this comment and the release note to any decode error.

auth/gcp/client_test.go
```diff
@@ -301,7 +312,7 @@ func TestRetrieveValidatesRequest(t *testing.T) {
 			if err == nil {
 				t.Fatalf("RetrieveCredential(%+v) = nil error, want error", tc.req)
 			}
-			if !strings.Contains(err.Error(), "requires a") && !strings.Contains(err.Error(), "invalid characters") {
+			if !strings.Contains(err.Error(), "requires a") && !strings.Contains(err.Error(), "resource ") {
```
karolpiotrowicz: The `"resource "` alternative is now true of every error `RetrieveCredential` can return, because the new defer appends `(resource %q)` to all of them. Delete the `validateResource` call entirely and this assertion still passes — only the `hits.Load()` check below still discriminates. Asserting the two concrete messages would also separate the two rejection reasons from each other.

auth/gcp/unescape_test.go
```diff
@@ -0,0 +1,390 @@
...
+	if withheld == checked {
+		t.Error("every body was withheld, so the clean path went unexercised")
+	}
+	if withheld == 0 {
+		t.Error("no body was withheld, so the fail-closed path went unexercised")
+	}
+}
+
+// readable is the test's own oracle for "an attacker can read a secret out of this
+// string", written without reference to the production predicate it judges.
```
karolpiotrowicz: The oracle makes the same three structural choices as `recoverable`: split on the literal marker, decode then fold, and the same `unescapeJSON`. The PR description already concedes it "is a copy of it", which matches what I read here — so it catches a future weakening but not a blind spot the two share today, and `\U0040` is an example of one they do share. The claim in this comment is the part to change rather than the test.

karolpiotrowicz (dismissed):
Two things need to be fixed before merge, both in the new scrub and both reachable through `RetrieveCredential` against a service that controls its own error body. The response body needs a bound before it reaches the fixpoint decode, so a hostile body cannot buy unbounded uncancellable CPU. And the truncation flag needs to describe the string that is actually returned, so the whole-text check cannot be switched off by a case-fold. I have deliberately not proposed an implementation for either — both sit on reasoning you have already done more carefully than I could in a review, and the second interacts with the cap ordering you argue for at length. Everything else below is small.

The design itself holds up. Judging on the output rather than on a property of the inputs, capping the text read rather than the text returned, and failing closed to `withheldText` are all the right calls, and the exhaustive differential against a reference at every possible cut is the strongest evidence in the change. Neither blocker is a reason to revert any of that.

## The fixpoint decode is quadratic on the response body

`\u005c` decodes to a backslash, which re-forms the introducer for a following `u005c`. Each `unescapeJSON` pass therefore shortens the text by exactly five bytes, and [`decodeFully`](https://github.com/google/adk-go/blob/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b/auth/gcp/redact.go#L175) runs one pass per five bytes of input.

`maxScrubbableSecret` bounds the caller-supplied values. Nothing bounds the body. `doPost` classifies the status before the size check, so a non-2xx delivers the full megabyte, and [`redact.go:110`](https://github.com/google/adk-go/blob/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b/auth/gcp/redact.go#L110) hands `redact(s, secrets...)` to `recoverable` — which returns `s` unchanged when nothing matched, so all of it reaches [`decodeFully` at line 155](https://github.com/google/adk-go/blob/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b/auth/gcp/redact.go#L155).

Measured through `RetrieveCredential` against an `httptest` server: 20 KB of body takes 203 ms, 40 KB takes 828 ms, 80 KB takes 3.25 s. At the 1 MiB read cap the same shape took **9m19s**. A caller that sets a 50 ms context deadline gets control back after 20.9 s, because nothing below `doPost` looks at `ctx`.

The invariant I would state: every input to the scrub that an attacker can influence needs a bound, and the response body is one of them. The 4 KiB cap on the values covers one factor of the cost the comment at [`redact.go:41-46`](https://github.com/google/adk-go/blob/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b/auth/gcp/redact.go#L41-L46) identifies, and the body is the other factor in the same sentence.

The same call site also materializes the body rather than scanning it. With `UserID` of `"a"` against 1 MiB of `"ab"`, `redact` emits 5,767,168 bytes, `strings.Split` on the marker then allocates 524,289 parts, and the call allocates 39.8 MiB in total and peaks at 17.8 MiB of heap — to compute one boolean.

```go
func TestQuadraticDecode(t *testing.T) {
	for _, n := range []int{4000, 8000, 16000} {
		body := `\` + strings.Repeat("u005c", n)
			w.WriteHeader(http.StatusForbidden)
			fmt.Fprint(w, body)
		}))
		c, _ := NewClient(context.Background(), &Config{
			AgentIdentityEndpoint: srv.URL, HTTPClient: srv.Client(),
		})
		start := time.Now()
		_, _ = c.RetrieveCredential(context.Background(), Request{
			Resource: "projects/p/locations/l/authProviders/ap",
			UserID:   "alice@example.test",
		})
		t.Logf("body=%d bytes -> %v", len(body), time.Since(start))
		srv.Close()
	}
}
```

```
body=  20001 bytes -> 203.485377ms
body=  40001 bytes -> 828.019147ms
body=  80001 bytes -> 3.246169326s
```

At `1<<20` bytes the same call took 9m19.342482432s. With a 50 ms `context.WithTimeout` and a 200,001-byte body it returned after 20.941130785s.

</details>

## The truncation flag is measured on one string and applied to another

[`redact.go:383`](https://github.com/google/adk-go/blob/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b/auth/gcp/redact.go#L383) takes `truncated` from `visibleLimit(ls)` on the lowered copy, and [line 386](https://github.com/google/adk-go/blob/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b/auth/gcp/redact.go#L386) returns `truncateForError(s)` cut from the original. `strings.ToLower` folds U+0130 from two bytes to one — the same fold [`redact.go:326`](https://github.com/google/adk-go/blob/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b/auth/gcp/redact.go#L326) already calls out — so the lowered copy can sit under the cap while the original crosses it. The flag then says "not truncated" about a string that ends in `"..."`.

That flag is the only guard on the whole-text check at line 110, the one whose comment describes exactly this case: an escaped occurrence the scrub could not match, sliced by the cut, leaving something that looks clean. With the flag false the check never runs, [`redact.go:117`](https://github.com/google/adk-go/blob/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b/auth/gcp/redact.go#L117) sees only the already-cut text where the address is a fragment rather than a whole match, and because the first candidate now reports success `redactedForError` never reaches the second candidate at [line 91](https://github.com/google/adk-go/blob/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b/auth/gcp/redact.go#L91) that would have decoded and scrubbed it.

Driven through `RetrieveCredential`, this puts **`alice@example.`** — 14 of the acting user's 18 bytes — in cleartext in the returned error. The same body with ASCII padding of the same byte length leaks nothing, so the fold is the cause rather than the escape.

The invariant: the flag and the text have to come from the same string. Any branch that derives one from `ls` and the other from `s` can disagree, and this one is load-bearing for a disclosure decision.

```go
func TestFragmentLeak(t *testing.T) {
	const victim = "alice@example.test"
	body := strings.Repeat("\u0130", 505) + victim[:len(victim)-1] + `\u0074`
		w.WriteHeader(http.StatusForbidden)
		fmt.Fprint(w, body)
	}))
	defer srv.Close()
	c, _ := NewClient(context.Background(), &Config{
		AgentIdentityEndpoint: srv.URL, HTTPClient: srv.Client(),
	})
	_, err := c.RetrieveCredential(context.Background(), Request{
		Resource: "projects/p/locations/l/authProviders/ap",
		UserID:   victim,
	})
	for n := len(victim); n >= 5; n-- {
		if strings.Contains(err.Error(), victim[:n]) {
			t.Errorf("leaked %d of %d bytes: %q", n, len(victim), victim[:n])
			break
		}
	}
}
```

```
--- FAIL: TestFragmentLeak
    leaked 14 of 18 bytes: "alice@example."
```

Swap the padding for `strings.Repeat("x", 1010)` and it passes.

</details>

## The ContinueURI wiring is pinned at two of its four sites

The description says the tests stop `ContinueURI` being dropped from all four sites, and that holds for the aggregate deletion. It does not hold per site. Every connector row in `TestServiceErrorsRedactTheActingUser` returns 200, so those rows exercise the secret list in `connectorOperation.result` rather than the one at [`connector.go:81`](https://github.com/google/adk-go/blob/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b/auth/gcp/connector.go#L81), and the only connector test reaching a `doPost` error arm sets no `ContinueURI`. `TestMapCredentialRedactsTheActingUserInError` builds its request without one, so [`client.go:293`](https://github.com/google/adk-go/blob/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b/auth/gcp/client.go#L293) is unpinned too. Deleting `req.ContinueURI` at either site leaves the package green. A connector row at 403 and a `ContinueURI` on the `mapCredential` request would close both.

## Smaller things

I have left the rest inline on the lines they concern.

One that is not yours and does not belong in this PR: [`client.go:405`](https://github.com/google/adk-go/blob/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b/auth/gcp/client.go#L405) wraps the transport error, and `url.Error.URL` is the last URL attempted rather than the one this package built. A caller-supplied `HTTPClient` that follows redirects lets a service put an identifier in a `Location` header and have it appear verbatim, outside the scrub and outside the cap. That line is unchanged from the base, so it is a separate issue rather than something to fix here.

On the description rather than the code: the 81 ms worst case measures a tiling value. A self-overlapping 4 KiB value against a 1 MiB body measures 162 ms. Still milliseconds, and the cost model in `redactLowered`'s comment predicts it exactly, so it is the number in the description that wants updating rather than the code.

auth/gcp/redact.go
wolo-lab: Fixed. The reference now names `TestDecodeFullyShrinksWheneverItChanges` for the per-pass lemma and says it is over `unescapeJSON` rather than over the loop, plus `TestDecodeFullyStopsAtItsPassBound` for the bound the loop now carries.

auth/gcp/client.go
wolo-lab: Fixed — it now says "a version segment (/v1 for Agent Identity, /v1alpha for the connector)". You are right that the connector path is the one a reader goes and checks, and it was the one the sentence got wrong.

auth/gcp/client.go
wolo-lab: Right, and the example you picked is the one that makes it undeniable — the package demonstrates the second target in a test it ships. The comment now says every `encoding/json` error loses its type this way and names `TestDecodeErrorScrubsTheActingUser`'s `*json.UnmarshalTypeError` as the case in front of us. The release-note bullet in the description is widened the same way.

auth/gcp/client_test.go
wolo-lab: Confirmed by deleting the `validateResource` call: the assertion passed and only `hits.Load()` failed. Each row now names the message it expects, `has invalid characters` or `has an empty or relative path segment`, which also separates the two rejection reasons the way you suggest. With the call deleted, nine subtests fail instead of one.

auth/gcp/unescape_test.go
wolo-lab: Agreed, and the comment was the wrong half. It now says outright that `readable` is a frozen copy rather than an independent oracle, that this buys a guard against a future weakening of `recoverable` and nothing against a blind spot they share today, and names `\U0040` as one they do share — `unescapeJSON` matches the introducer case-sensitively while `redact` folds case, and neither predicate folds before decoding.

wolo-lab:
Both blockers are real, both are mine, and both are fixed in [`09e55ec3`](https://github.com/google/adk-go/commit/09e55ec3bd44585dd0844c2093d6ab77f58870c0). I reproduced each at your numbers before touching anything — 20 KB at 233 ms, 40 KB at 877 ms, 80 KB at 3.46 s, and `alice@example.` at 14 of 18 bytes with your ASCII control staying clean.

Worth locating the first one precisely, because it says something about the change rather than about the function. `decodeFully` was always quadratic on that shape, but until the previous commit `recoverable` only ever saw text that had already been cut, so its input was a kilobyte. The whole-text check I added to close the straddling-escape case is what first handed it an uncapped megabyte. I bounded the values in that same commit and wrote in its message that the cost had a ceiling — and the sentence was about one of the two factors while the commit introduced a new path for the other.

## The fixpoint decode

`decodeFully` now stops at 64 passes and returns whether it reached a fixpoint, and both call sites in `recoverable` treat "did not finish" as "assume a secret is in there". That distinction is the point: not finishing the decode is not the same as showing the text clean, and the predicate feeds a disclosure decision. Legitimate text needs one pass, or two where a body was JSON-encoded twice, so 64 is two orders of magnitude of headroom.

Your other measurement on that call site — 39.8 MiB and 524,289 parts to compute one boolean — is a separate defect and I fixed it separately. Only an occurrence that straddles the cut can hide behind it, and an occurrence has to start before the cut to straddle it, so the check now reads the visible window plus room for one occurrence's escaped spelling instead of the whole body.

```
1 MiB self-regenerating body   9m19s  ->  35 ms
1 MiB, one-character value     45 MB  ->  4.6 MB allocated, 714 allocations
```

The two bounds needed separate tests, which is worth flagging because my first attempt only had one. An end-to-end timing test cannot see the pass bound once the window is in place — the window keeps the input small enough that even an unbounded loop finishes inside any sane budget. So the pass bound is pinned by a unit test on `decodeFully` and on `recoverable` failing closed, and the window is pinned by an allocation assertion, which is the only thing that can see it.

## The truncation flag

`truncateForError` now returns the pair. Both no-redaction paths in `redactWithinLimit` return it directly, so the flag and the text come from one call and cannot describe different strings. Your diagnosis was exact, including that the fold is the cause rather than the escape — the ASCII row is in the test as the control for precisely that reason.

## ContinueURI

You are right that the aggregate deletion is a weaker claim than the per-site one, and right about which two sites were uncovered. Two rows added: a connector resource at 403, which reaches `connector.go`'s `doPost` secret list rather than `connectorOperation.result`, and a 2xx whose header name is unusable and carries the URI, which reaches `mapCredential`. Deleting `req.ContinueURI` at either site alone now fails one of them.

## The rest

The 81 ms in the description was a tiling value, as you say. It now reads 162 ms for the self-overlapping case, alongside the two new bounds.

On the transport wrap, [client.go:410](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/client.go#L410) on this head — agreed on both counts, that it is a real hole and that it does not belong here. A caller-supplied `HTTPClient` that follows redirects lets a service put an identifier in a `Location` header and have it come back through `url.Error.URL`, outside the scrub and outside the cap. The line predates this branch. I will file it separately unless you would rather.

The five smaller ones are answered on their own threads and all five are applied.

auth/gcp/redact.go
```diff
@@ -0,0 +1,593 @@
...
+
+		// One marker per redacted run, not per match. Adjacent ranges are the
+		// common case for a short secret — a megabyte of the acting user's single
+		// initial would otherwise emit a megabyte of markers, ten times the input,
+		// to produce a kilobyte of error.
+		//
+		// hit doubles as "something is already emitted", which is what makes
+		// at == pos mean "this range touches the last one" rather than "this is
+		// the first range and it starts at zero".
+		if at > pos || !hit {
```
karolpiotrowicz, line 640: One marker per run bounds the marker count against runs, not against the cap. A body of `strings.Repeat("u ", 600)` with `UserID: "u"` gives 512 separate runs and an `APIError.Body` of 5635 bytes, against the 1027 that `TestDoPostOversizeKeepsStatus` asserts — that assertion never fires because its fixture body is all `x` and the UserID is `u`, so it only ever measures the no-match path. Nothing leaks and it is bounded around 6 KB, but the service picks the multiplier and the result is what gets quoted into a prompt. This predates the current commit. Either assert the real bound or budget the markers against `maxErrorBody`.

auth/gcp/redact.go
```diff
@@ -0,0 +1,593 @@
...
+	//
+	// Decoded spelling too, because decoding is what mangles a secret. An
+	// identifier containing \/ survives a decoded copy's scrub as the slash it
+	// decodes to, which is not the secret and is still the identity.
+	var forms []string
+	for _, v := range secrets {
+		if v == "" {
+			continue
+		}
+		decoded, done := decodeFully(v)
```
karolpiotrowicz, line 208: This fail-closed arm is the one place in the file nothing pins. Flipping it to fail open leaves the whole suite green, and `TestDecodeFullyStopsAtItsPassBound` cannot see it because it puts the undecodable text in `x` and passes a secret with no escapes in it, so `decodeFully(v)` always converges here. The sibling arm below is covered. One row closes it: `recoverable("clean", []string{long})` must be true.

auth/gcp/redact.go
```diff
@@ -0,0 +1,593 @@
...
+// and 1 MiB over nine minutes of CPU no caller's deadline could cancel.
+//
+// TestDecodeFullyShrinksWheneverItChanges pins the per-pass lemma the argument
+// rests on, over unescapeJSON rather than over this loop, and
+// TestDecodeFullyStopsAtItsPassBound pins the bound.
+//
+// The bound is reported rather than swallowed because the answer this feeds is a
+// disclosure decision. A caller that cannot finish decoding has not shown the
+// text to be clean, and must treat it as though a secret were in there.
+func decodeFully(s string) (string, bool) {
```
karolpiotrowicz, line 340: The loop runs 64 iterations but only reports `done` on a pass that changes nothing, so the effective bound is 63 changing passes. `decodeFully("\\u005c" + strings.Repeat("u005c", 63))` comes back fully decoded with `done == false`. The direction is fail-closed so nothing goes wrong, but the doc reads as 64.

auth/gcp/client.go
```diff
@@ -185,16 +234,52 @@ type Request struct {
...
+	// it correctly — see doPost.
+	//
+	// It deliberately does NOT happen again here. Re-running redact over an
+	// already-scrubbed Body cannot find a real occurrence, because the first pass
+	// removed them all, and it can find a spurious one: a user id of "e" matches
+	// inside the "[redacted]" marker itself and rewrites it to
+	// "[r[redacted]dact[redacted]d]", nesting once per pass and destroying the
+	// operator's error along the way. A second scrub that can only corrupt is
+	// worse than none, so this defer decorates and nothing else.
+	defer func() {
```
karolpiotrowicz, line 326: This wraps every error past validation, so `err == ErrConsentRejected`, `err == ErrPollTimeout`, `err == context.Canceled` and `err.(*auth.ConsentRequiredError)` all stop matching, on the same mechanism as the `err.(*APIError)` break already in your release-note list. `errors.Is` and `errors.As` still work and are the documented style, so this is a note for the list rather than a code change.

auth/gcp/unescape_test.go
```diff
@@ -0,0 +1,419 @@
...
+			}
+		})
+	}
+}
+
+// TestServiceTextNeverReturnsARecoverableSecret is the invariant the four earlier
+// revisions of redactedForError each violated, stated once and checked over every body
+// an adversary can assemble from the pieces that broke them.
+//
+// It deliberately does NOT call recoverable. An earlier version of this test did,
```
karolpiotrowicz: This block still says the oracle "deliberately does NOT call recoverable" and "fails when recoverable under-approximates", which the corrected block six lines below now contradicts — and the corrected one is the true one. `readable` makes the same three structural choices as `recoverable`, so `redactedForError` returning a candidate only when `recoverable` says no makes the assertion hold by construction. The commit message lists this correction among its four, and one of the two sites was missed.

auth/gcp/client_test.go
```diff
@@ -712,3 +781,628 @@ func TestNewClientRejectsNegativePollTimeout(t *testing.T) {
...
+
+// TestResponseBodyIsBoundedToo pins the other half of the cost ceiling. The value
+// bound covers the caller's side; this covers the service's.
+//
+// decodeFully shortens by as little as five bytes a pass, and `\u005c` re-forms
+// its own introducer, so an unbounded fixpoint decode is one pass per five bytes
+// of a response this package reads a megabyte of. Before the pass bound: 20 KB
+// took 233ms, 80 KB 3.5s, and a megabyte over nine minutes of CPU that no
+// caller's deadline could interrupt, because nothing below doPost reads ctx.
+func TestResponseBodyIsBoundedToo(t *testing.T) {
```
karolpiotrowicz, line 1291: This is attributed no kill of its own. With the pass bound removed, `straddleWindow` still caps the input at 33792 bytes so an unbounded loop finishes well inside the 2 s budget, and with the window removed the 64-pass bound holds it to tens of milliseconds. It fires only when both bounds are gone, which is the previous revision's state. Worth keeping, but the docstring's claim to pin the other half of the ceiling overstates what it does.

auth/gcp/client_test.go
```diff
@@ -712,3 +781,628 @@ func TestNewClientRejectsNegativePollTimeout(t *testing.T) {
...
+			}))
+			defer srv.Close()
+
+			_, err := newTestClient(t, srv).RetrieveCredential(t.Context(), Request{
+				Resource: authProviderResource, UserID: victim,
+			})
+			if err == nil {
+				t.Fatal("RetrieveCredential() = nil error, want the 403")
+			}
+			for n := len(victim); n >= 5; n-- {
```
karolpiotrowicz, line 1392: Both rows assert only that the address is absent, with no positive control. Replacing `showable`'s success return with `return "", false` passes both while destroying every error body in the package. The sibling wiring tests in this file pair absence with presence for exactly that reason — asserting the padding survives, or that the marker is present, would close it.

karolpiotrowicz (dismissed):
Both blockers from the last round are closed, and I checked each against the previous revision so the numbers mean something: the 1 MiB self-regenerating body goes from 9m19s to 30.6 ms, the 50 ms deadline case from a 21.1 s overrun to 25.3 ms, and the U+0130 fold no longer leaks — nor does U+212A, which folds 3 bytes to 1 rather than 2 to 1 and leaks on the old revision too. The pass bound is enforced on both paths into `decodeFully`, not just the one that motivated it, and an over-long value withholds the whole response rather than being silently skipped. Pairing the cut flag with its text is the right shape for that fix, and `TestTruncationFlagDescribesTheTextReturned` genuinely fails if you undo it.

Two things stop me signing off, and the first is the harder one because the fix itself opened it.

Before merge I need the straddle window to stop returning a decodable fragment of a caller value, with a test that fails when the window is too small rather than only when it is absent, and the argument at `redact.go:107-109` corrected since it invokes a disclaimer that does not cover this case. The second finding needs either a fix or your explicit call that a leaked redirect URI is acceptable — I am not going to make that judgement for you, but I do not want it decided by silence.

## The straddle window lets a fragment of the ContinueURI out

[`maxStraddleWindow`](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L114) caps what the whole-text check at [`redact.go:133`](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L133) reads at 33792 bytes, and [`recoverable`](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L190) matches only the whole secret. Once the window slices a straddling occurrence, the check sees a proper prefix, answers false, and the visible kilobyte is returned.

The escaped spelling's length is the service's choice, not eight times the value. One character at nesting level k costs `2^(k-1)+5` bytes and takes k decode passes, so anything up to 64 levels is still reconstructed by `decodeFully`. A service can spell a cheap prefix at 6 bytes per character, filling the visible window exactly, and the tail at level 12 — which puts the end of the occurrence past the window at any value length. Shorter values are worse, not better, because fewer expensive characters are needed to get there.

Through `RetrieveCredential`, with a 200-byte redirect URI:

```
recovered 170 of the ContinueURI's 200 characters:
"https://app.example.test/oauth/callback?state=kkkk..."
```

<details><summary>Reproduction — <code>go test -run TestLeakThroughRetrieveCredential ./auth/gcp/</code></summary>

```go
func TestLeakThroughRetrieveCredential(t *testing.T) {
	uri := "https://app.example.test/oauth/callback?state=" + strings.Repeat("k", 154)

	// The service spells the URI back: the first 170 characters cheaply, so they
	// fill the 1024-byte visible window, and the last 30 at a nesting depth whose
	// spelling runs past maxStraddleWindow.
	var body strings.Builder
	for i, r := range uri {
		depth := 1
		if i >= 170 {
			depth = 12
		}
		fmt.Fprintf(&body, "%su%04x", strings.Repeat(`\`, 1<<(depth-1)), r)
	}

	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		w.WriteHeader(http.StatusForbidden)
		fmt.Fprint(w, body.String())
	}))
	defer srv.Close()

	c := newTestClient(t, srv)
	_, err := c.RetrieveCredential(t.Context(), Request{
		Resource:    authProviderResource,
		UserID:      "alice@example.test",
		ContinueURI: uri,
	})

	got, _ := decodeFully(err.Error())
	for n := len(uri); n >= 5; n-- {
		if strings.Contains(got, uri[:n]) {
			t.Fatalf("decodeFully recovers %d of the ContinueURI's %d characters: %q", n, len(uri), uri[:n])
		}
	}
}
```

It passes on the previous revision ([`aa55bccd`](https://github.com/google/adk-go/commit/aa55bccd0d58d6e660572a7f3bcbba8cd3605b0b)) and fails on this one ([`09e55ec3`](https://github.com/google/adk-go/commit/09e55ec3bd44585dd0844c2093d6ab77f58870c0)). Before the window the check read the whole body, found the secret and withheld.

</details>

The argument at [`redact.go:107-109`](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L107-L109) is what I would revisit first. It sends a spelling inflated past 8× to the "Not covered" paragraph, but [that paragraph](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L64-L67) disclaims a form **no decoder reconstructs**, and this one is reconstructed by your own `decodeFully` in a dozen passes — inside the 64 you chose precisely so multiply-escaped spellings would be caught. So [the contract at `redact.go:60-61`](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L60-L61) is the thing being broken, not the disclaimer being invoked.

I don't think a bigger constant closes it, since the inflation factor is independent of the value length. Bounding in decoded bytes rather than raw ones, or having `straddleWindow` report that it cut and treating a cut like `decodeFully`'s `!done`, both keep the cost profile you built — the 39.8 MiB came from `redact` plus the marker split, not from reading the body. Whichever way it goes, the window's size needs a test that fails when it is too small: shrinking it eightfold and growing it sixty-fourfold both leave the suite green today, because `TestWholeTextCheckReadsABoundedWindow` asserts an allocation ceiling and a smaller window always satisfies that.

## A UserID contained in the ContinueURI splits it on our own marker

When one value is a substring of the other and the service echoes the URI with `\/` escapes — which RFC 8259 permits and your own comment says several encoders emit by default — [`redact`](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L405) misses the escaped URI, matches the short UserID, and writes `[redacted]` into the middle of the URI. [`recoverable`](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L183) then splits on that marker and sees neither half whole, so the first candidate self-approves and [the second one](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L91), which would have decoded and matched the URI, is never reached.

```
UserID:      "app"
ContinueURI: "https://my-app.test/oauth/callback"
body:        error at https:\/\/my-app.test\/oauth\/callback

returned:    error at https:\/\/my-[redacted].test\/oauth\/callback
decoded:     error at https://my-[redacted].test/oauth/callback
```

Putting the caller's own UserID back where the marker is reconstructs the ContinueURI exactly, and the reader is the one person who always knows that value. `"alice"` inside `https://cb.test/u/alice/done` behaves the same way. This one predates the current commit and is not a regression — on `main` neither value is scrubbed at all — so it is a gap in new coverage rather than something you broke.

Worth saying what it is not: this is not the service writing the marker itself, which we settled last round. The marker here is ours, and [`redact.go:396-399`](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L396-L399) says this overlap case is handled. The acting user is correctly redacted in every shape I tried, so what escapes is the redirect URI. If you read a developer-hosted callback URL as configuration rather than an identifier, this is a much smaller thing than the first one — I'd rather you made that call than have me assume it.

## One doc ask

`alice\U0040example.test` comes back verbatim, because [`hasEscape`](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L326) matches the introducer case-sensitively while `redact` folds case. I agree with you that it belongs outside the scrub, in the same family as `%40`. But `\U0040` is named only in a test comment, and the exported paragraph a caller actually reads is [`redact.go:64-67`](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L64-L67). It should be listed there alongside the percent-encoded case.

The rest is small and left inline on the lines it concerns.

auth/gcp/redact.go
wolo-lab: Budgeted rather than asserted, since the doc promised a kilobyte and the result was several. The cap now applies to the result as well as to the source bytes it is built from, at [redact.go:568-575](https://github.com/google/adk-go/blob/41045fd2c6295af3703a5d4d07bd88180b264906/auth/gcp/redact.go#L568-L575). Your shape returns 1027 bytes now instead of 5635.

The first test I wrote for it could not tell the two implementations apart, which is worth flagging because it is the same failure you found in the two tests below. It used a 1200-byte source, where the input bound has already fired, so a cap reachable only after truncation passes it just as well. The fixture is now 1000 bytes — under the input cap, which is the only size that discriminates — and it asserts the body is shortened rather than suppressed, since withholding also satisfies a ceiling. `TestMarkersAreBudgetedAgainstTheCap` at [client_test.go:1552](https://github.com/google/adk-go/blob/41045fd2c6295af3703a5d4d07bd88180b264906/auth/gcp/client_test.go#L1552).

auth/gcp/redact.go
wolo-lab: Added, as the row you wrote. It sits with the sibling arm's row in `TestDecodeFullyStopsAtItsPassBound` at [unescape_test.go:396](https://github.com/google/adk-go/blob/41045fd2c6295af3703a5d4d07bd88180b264906/auth/gcp/unescape_test.go#L396), so the two arms are covered next to each other and it is visible that there are two.

Confirmed the way you framed it: flipping this arm to fail open left the whole package green before, and turns that test red now.

auth/gcp/redact.go
wolo-lab: Right on both counts, and the doc was the wrong half. It now says the constant counts reads, that a text is reported finished only by a read which changes nothing, and that the last read therefore has nothing after it to confirm — so the most changing passes that can be reported finished is one fewer. See [redact.go:305-311](https://github.com/google/adk-go/blob/41045fd2c6295af3703a5d4d07bd88180b264906/auth/gcp/redact.go#L305-L311).

The boundary is pinned rather than described: rows at 62 and 63 changing passes, the first reported finished and the second not, so it cannot drift back into being folklore.

auth/gcp/client.go
wolo-lab: Added to the release-note list, and named in `RetrieveCredential`'s own doc at [client.go:261-267](https://github.com/google/adk-go/blob/41045fd2c6295af3703a5d4d07bd88180b264906/auth/gcp/client.go#L261-L267), which enumerated what stays matchable and never said what stopped being.

One thing to add to your framing: `auth/gcp` ships in v2.3.0, so this is a break against a released version rather than churn inside a new package. `git show v2.3.0:auth/gcp/client.go` has `type APIError` and a decoder path wrapping with `%w`, so `errors.As` against `*json.SyntaxError` worked there and does not here.

auth/gcp/unescape_test.go
wolo-lab: Missed, and you are right that the corrected one is the true one. Both blocks said the same false thing before, which at least made them consistent — correcting one turned that into a contradiction, which is worse. This block now says the same as its sibling: removing the circularity is not the same as buying independence, and the oracle does not claim the second.

auth/gcp/client_test.go
wolo-lab: The docstring overstated it and now says so. It attributes no kill of its own and is kept for the end-to-end shape rather than for what it pins, with a note that either bound alone holds the budget and it fires only with both gone. The pass bound is pinned directly by `TestDecodeFullyStopsAtItsPassBound`, and the window by `TestWholeTextCheckReadsABoundedWindow` from above and `TestBodyInsideTheWindowIsStillShown` from below.

Three neighbours had quietly become the same thing, which your comment led me to. `TestMapCredentialCapsHeaderNameInError`, `TestDoPostOversizeKeepsStatus` and `TestRetrieveConnectorErrorMessageIsCapped` all use 900 KB or 1 MiB fixtures, and those are now past the examinable length, so they were withheld before the cap was consulted and their length assertions held on 144, 68 and 68 bytes of the withheld sentence. Sized inside that length they measure the cap again. `TestDoPostOversizeKeepsStatus` keeps a megabyte row for the status assertion and gains a smaller one for the cap, since only the smaller one can see it.

auth/gcp/client_test.go
wolo-lab: Both rows now pair absence with presence, as the siblings do. Each asserts that the service's own padding survives the scrub, so replacing `showable`'s success return with `return "", false` turns them red instead of passing while every error body in the package is destroyed.

wolo-lab:
Both of the things that stopped you signing off are addressed, and the first one turned out to have a relative I had not seen. Head is [41045fd2](https://github.com/google/adk-go/commit/41045fd2c6295af3703a5d4d07bd88180b264906). Replies to the seven inline notes are on their own threads.

## The straddle window

Your reading of it was right and so was your conclusion that a bigger constant does not close it. The inflation factor is the service's choice and no fixed window bounds it, so reading part of an occurrence answers the question wrong rather than conservatively. Past the window the response is now withheld unexamined, at [redact.go:167](https://github.com/google/adk-go/blob/41045fd2c6295af3703a5d4d07bd88180b264906/auth/gcp/redact.go#L167), which is also cheaper than examining it — the expensive check no longer runs on the bodies that were driving its cost.

The argument at [`redact.go:107-109`](https://github.com/google/adk-go/blob/09e55ec3bd44585dd0844c2093d6ab77f58870c0/auth/gcp/redact.go#L107-L109) on the revision you read was wrong twice over, as you said. It sent an inflated spelling to a paragraph that disclaims a form no decoder reconstructs, and that spelling is reconstructed by our own `decodeFully` well inside the pass bound chosen for exactly that purpose. The constant's comment now says what it is: a threshold on the response, not a window into it. `straddleWindow` went with it, since the prefix it handed back had no reader once the check short-circuits, and returning it was what let a fragment out.

The size is pinned in both directions now. Too large is caught by `TestWholeTextCheckReadsABoundedWindow`, which was already there. Too small is caught by [`TestBodyInsideTheWindowIsStillShown`](https://github.com/google/adk-go/blob/41045fd2c6295af3703a5d4d07bd88180b264906/auth/gcp/client_test.go#L1516), which is new, and whose fixture repeats the window's defining expression instead of naming the constant — sized from the constant it would shrink along with it and could never fire, which is the trap the first draft of it fell into.

One correction to the supporting detail, since it changes what to test rather than whether to. Shrinking eightfold does leave the suite green, as you said. Growing sixty-fourfold turns it red, but only incidentally: 64× is 2,162,688 bytes, past the 1 MiB fixture, so the window stops cutting that body at all and the mutation becomes equivalent to deleting the window. Growing eightfold is green. Your point stands unchanged — nothing failed when the window was too small.

Worth naming as a behaviour change: a response longer than 33,792 bytes is withheld whenever the kilobyte cap cut it, so a large gateway error page now yields the withheld sentence rather than its first kilobyte. The status code and the resource survive. It is in the release-note list.

## The ContinueURI

My call: not fixed here, tracked as #1539, and the paragraph a caller reads now says where the cover is thinner rather than leaving it to be discovered.

The reasoning, since you asked for a decision rather than a silence. I do read the redirect URI as an identifier and not as configuration, so I am not claiming the leak is acceptable. What changed my mind about doing it in this PR is the size of the fix. Reconstructing the URI needs a predicate that tolerates a marker-sized gap inside an occurrence, because the value is not present in the returned text at all — only substituting the caller's own `UserID` back produces it. That is a new predicate, and predicates in this file have earned their own exhaustive tests. Bolting it on at the end of a change that has already shipped one regression seemed like the wrong trade.

## `\U0040`

Named in the exported paragraph now, alongside the percent-encoded case, rather than only in a test comment.

## One more in the same family

Chasing the marker question turned up a caller-side twin of it that is a straight contract violation. `recoverable` split the candidate on the literal marker, so a value whose own bytes contain or abut `[redacted]` spanned a split, landed whole in no part, and was matched by nothing.

```
secret "a[redacted]b"   body \u0061[redacted]b   -> returned unchanged, decodeFully gives "a[redacted]b"
secret "alice[red"      body \u0061lice[redacted] -> returned unchanged, decodeFully gives "alice[redacted]"
```

Removing the split is not available, for the reason its comment gives: a one-character user id of `"e"` then matches inside the marker and every response is suppressed. But position is the criterion the split was approximating, and it is exact — an occurrence is ours only when every byte of it came from a marker. Marker bytes are masked once per text and each occurrence is tested against that mask, at [redact.go:241-262](https://github.com/google/adk-go/blob/41045fd2c6295af3703a5d4d07bd88180b264906/auth/gcp/redact.go#L241-L262).

Over 432 bodies built from values that contain, abut or equal the marker, each also spelled with its first byte escaped: 38 leaked before, none does now, and none is newly withheld — all 38 are scrubbed rather than suppressed. Both controls hold, so a one-character value inside the marker is still shown redacted. It is cheaper too, since the text is decoded once rather than once per marker-separated part, and the service chose that part count.

The service-side variant you and I settled last round is narrower than it was. An identifier spanning a marker the service wrote is now seen. What remains excused is an identifier that fits entirely inside marker text, which is exactly the set of the marker's own substrings, and that is the case the exclusion exists for.

karolpiotrowicz:
The straddle-window leak is closed. Replacing the prefix slice with a threshold, and withholding rather than examining part of an occurrence, is the right shape — reading a proper prefix answers the question wrong instead of conservatively. Measured on the round-2 input, a 205,900-byte body spelling a 200-byte ContinueURI at mixed escape depths:

```
before: recovered 170 of 200 bytes of the ContinueURI from the returned error
after : withheld=true, recovered 0 of 200
```

The contained-secret shape is unchanged at [41045fd](https://github.com/google/adk-go/commit/41045fd2c6295af3703a5d4d07bd88180b264906), including after the marker-by-position commit:

```
UserID      "app"
ContinueURI "https://my-app.test/oauth/callback"
body        error at https:\/\/my-app.test\/oauth\/callback

returned: error at https:\\/\\/my-[redacted].test\\/oauth\\/callback
decoded : error at https://my-[redacted].test/oauth/callback
```

@wolo-lab this one is your call rather than mine, so I would rather ask than hold a review open on it.

The narrow claim in the disclosure at [redact.go:84-89](https://github.com/google/adk-go/blob/41045fd2c6295af3703a5d4d07bd88180b264906/auth/gcp/redact.go#L84-L89) is true. `app` is genuinely gone and the package's own decoder does not bring the URI back, so the guarantee holds exactly as the paragraph now words it, and [#1539](https://github.com/google/adk-go/issues/1539) tracks the gap.

What I cannot settle on my own is whether that is the guarantee this package should ship. A reader — or a model — sees `my-[redacted].test/oauth/callback`, which is most of a caller identifier, and [APIError.Body's godoc](https://github.com/google/adk-go/blob/41045fd2c6295af3703a5d4d07bd88180b264906/auth/gcp/client.go#L107-L112) still tells callers the ContinueURI is "removed and replaced with `[redacted]`". Those two paragraphs describe different promises. The value also lands somewhere durable, since this error is fed to the model and persisted into the session.

So the question is which of three you want: narrow the exported godoc to match the disclosure, treat the shape as a defect to fix before this merges, or ship as-is on the grounds that #1539 is the right place for it. Any of those is defensible and I will follow whichever you pick.

Two smaller things, neither blocking.

The sentence at [redact.go:81-82](https://github.com/google/adk-go/blob/41045fd2c6295af3703a5d4d07bd88180b264906/auth/gcp/redact.go#L81-L82) says a response longer than `maxStraddleWindow` is withheld whenever the visible cap cut it. The second candidate falsifies it — 120,000 bytes of `\u0041` is withheld as candidate 1, then `unescapeJSON` reduces it to 20,000, which fits the window, is examined whole, is found clean, and 1,128 bytes come back. Nothing escapes, since that candidate was examined in full. It is the wording that promises more than the code does, and the fix is to say which candidate the sentence describes.

The comment rewritten at [client_test.go:281-285](https://github.com/google/adk-go/blob/41045fd2c6295af3703a5d4d07bd88180b264906/auth/gcp/client_test.go#L281-L285) now says `"resource "` appears in the validation messages themselves, so an assertion on it survives deleting `validateResource`. Delete `validateResource` and those messages do not exist, so they cannot be what keeps such an assertion green. What does is the deferred wrap, which is registered after the three validation returns and decorates everything past them. The version this replaced said that and was right, and it is the argument protecting the path-traversal guard.

wolo-lab:
Answering the question first, since you held it open for me: narrow the exported godoc. Head is [1e18d7e5](https://github.com/google/adk-go/commit/1e18d7e5e62d80ad30344a04e51e4edd99221add).

## The ContinueURI

Your framing of it as two paragraphs promising different things is the part that decided it. I had been reading this as one question — is the leak acceptable — and it is two. Whether to ship the gap and whether the exported doc may describe a stronger guarantee than the code delivers are separable, and only the second has an obviously wrong answer.

So the shape ships as it is, with #1539 tracking it, and [`APIError.Body`](https://github.com/google/adk-go/blob/1e18d7e5e62d80ad30344a04e51e4edd99221add/auth/gcp/client.go#L107-L122) no longer says the values "are removed". It says removal is best effort, states the guarantee that actually holds — no value this package was given is recoverable from `Body` by this package's own decoder — and names the shape where a value stays partly legible, with your `my-[redacted].test/oauth/callback` in it so the reader sees what it looks like rather than having to picture it.

For the record on the part that is a judgement rather than a fact: I do read a redirect URI as an identifier and not as configuration, so this is not me deciding the leak is harmless. It is me deciding that a predicate tolerating a marker-sized gap inside an occurrence is a change that owes its own exhaustive test, and that bolting it onto a PR which has already shipped one regression is the worse of the two risks. Your point about durability is the one that would change my mind if this sat much longer — the value does land in the session, not just in a log line — so I would rather #1539 be picked up soon than quietly.

## The withholding sentence

You are right and the repro reproduces. 120,000 bytes of `\u0041` comes back at 1,027 bytes with `withheld=false`.

The sentence described the response and the code measures the candidate, and those differ by exactly one `unescapeJSON` pass. [`redact.go:81-86`](https://github.com/google/adk-go/blob/1e18d7e5e62d80ad30344a04e51e4edd99221add/auth/gcp/redact.go#L81-L86) now says candidate, says the second one is decoded a pass before it is offered, and carries your number so the reader can see the size of the effect.

## The validation comment

You are right, and I made this worse rather than better. The claim I replaced it with is true and beside the point: `"resource "` does appear in the validation messages, and deleting `validateResource` deletes those messages, so they cannot be what keeps a weak assertion green.

Confirmed with the guard neutered — all three rejected names reach the HTTP call, and every resulting error carries `resource ` from the deferred wrap:

```
projects/p/locations/l/authProviders/ap/   -> "…403: \"\" (resource \"projects/p/locations/l/authProviders/ap/\")"
projects/p//locations/l/authProviders/ap   -> "…403: \"\" (resource \"projects/p//locations/l/authProviders/ap\")"
projects/p/locations/./authProviders/ap    -> "…403: \"\" (resource \"projects/p/locations/./authProviders/ap\")"
```

[The comment](https://github.com/google/adk-go/blob/1e18d7e5e62d80ad30344a04e51e4edd99221add/auth/gcp/client_test.go#L280-L287) is back to your argument, worded so it cannot be read the other way: the wrap does not decorate the validation errors, which return above it. Deleting the guard is what lets a request reach it.

Worth saying why I changed a correct comment. It was flagged as false on the reading that the wrap decorates those three returns, which it does not, and I fixed the literal claim without checking what the sentence was for. The mutation you would run — delete the guard, see what stays green — is the thing I should have run before touching a comment that exists to justify a test's shape.

auth/gcp/unescape_test.go
```diff
@@ -0,0 +1,513 @@
...
+	}
+}
+
+// readable is the test's own oracle for "an attacker can read a secret out of this
+// string". It is a frozen COPY of recoverable, not an independent one: it makes
+// the same three structural choices — split on the literal marker, decode then
+// fold, the same unescapeJSON. So it catches a future weakening of recoverable,
+// which is worth having, and it cannot catch a blind spot the two share today.
+// \U0040 is one they share: unescapeJSON matches the introducer case-sensitively
+// while redact folds case, and neither this nor recoverable folds before decoding.
```
karolpiotrowicz, line 352: This describes the predicate as it was before this PR. `recoverable` no longer splits on the literal marker — it masks marker bytes by position and searches around them — so `readable` is not a frozen copy of it any more, and the two now disagree:

```
readable("a[redacted]b", []string{"a[redacted]b"})    -> false
recoverable("a[redacted]b", []string{"a[redacted]b"}) -> true
```

The practical consequence is smaller than it looks, and I checked rather than assumed: reverting `recoverable` to the split form still fails all three leak rows of `TestSecretSpanningOurOwnMarkerIsNotShown`, because those assert `decodeFully` plus `strings.Contains` directly instead of going through the oracle. So the regression is caught, just not here. What is left is that the comment now names a structural choice the code stopped making, and the same sentence appears again at [`unescape_test.go:294-297`](https://github.com/google/adk-go/blob/1e18d7e5e62d80ad30344a04e51e4edd99221add/auth/gcp/unescape_test.go#L294-L297).

auth/gcp/client.go
```diff
@@ -172,6 +233,19 @@ type Request struct {
...
+	// must be non-empty and be neither "." nor "..". So a trailing slash, a
+	// doubled slash, and a relative segment are refused, while a domain-scoped
+	// project id such as projects/example.com:my-project/... is accepted.
+	//
+	// The segment rule is what keeps routing and normalization from disagreeing:
+	// the routing pattern above is matched on the name as given, and a name that
+	// normalizes to a different one would be routed by one and served as the
+	// other. Both rules moved relative to v2.3.0, in opposite directions: the
+	// colon was rejected there and is accepted here, and those segment shapes
+	// were accepted there and are rejected here.
```
karolpiotrowicz: This clause is off in both directions, checked against the `v2.3.0` tag, where the guard was `!resourceNameRE.MatchString(name) || strings.Contains(name, "..")` over `^[A-Za-z0-9._~/-]+$`.

A third rule moved and is not mentioned: a `..` inside a segment was rejected there and is accepted here. `projects/example..com/locations/x/connectors/y` fails at v2.3.0 on the substring test and passes now. That widening is the right one and you already pin it at [`client_test.go:356`](https://github.com/google/adk-go/blob/1e18d7e5e62d80ad30344a04e51e4edd99221add/auth/gcp/client_test.go#L356) — it is only the migration sentence that omits it.

And "those segment shapes were accepted there" does not hold for all three shapes the paragraph above enumerates. A `..` segment was already rejected at v2.3.0 by the same substring test, so for that one nothing moved.

The current-behaviour paragraph above is accurate and complete. It is only the historical comparison that misleads.

karolpiotrowicz (commented):
Nothing here blocks. The question I left open is answered, both smaller points from last round are closed, and what remains is three comments that describe the code less accurately than the code deserves.

Narrowing the exported promise was the right call, and [`APIError.Body`](https://github.com/google/adk-go/blob/1e18d7e5e62d80ad30344a04e51e4edd99221add/auth/gcp/client.go#L116-L121) now reads as a guarantee someone can rely on rather than one they have to discover the edges of. Putting `my-[redacted].test/oauth/callback` in the doc itself is what makes it work — a reader sees the shape instead of imagining it.

I checked the marker-by-position change rather than taking it on the comment, since it is the one piece of real logic in this round. Reverting `recoverable` to the old split form fails all three leak rows of `TestSecretSpanningOurOwnMarkerIsNotShown`, so the fix is pinned by tests that assert the contract directly instead of routing through the oracle. Running both predicates over 6,000 generated bodies that include service-written escaped markers, the new one withholds more (402 against 372) and leaves the value literally present in fewer results (898 against 908). It is a tightening, and it measures as one.

The withholding sentence and the validation comment both check out. A 120,000-byte body of `\u0041` comes back at 1,027 bytes with `withheld=false`, which is what [`redact.go:81-86`](https://github.com/google/adk-go/blob/1e18d7e5e62d80ad30344a04e51e4edd99221add/auth/gcp/redact.go#L81-L86) now says. With `validateResource` neutered, all three rejected names reach the HTTP call and every error carries `resource ` from the deferred wrap, so the comment you restored is the accurate one.

One thing worth a sentence in the docs, not a code change.

[`client.go:116-118`](https://github.com/google/adk-go/blob/1e18d7e5e62d80ad30344a04e51e4edd99221add/auth/gcp/client.go#L116-L118) states the guarantee unconditionally, but the marker rule at [`redact.go:191-197`](https://github.com/google/adk-go/blob/1e18d7e5e62d80ad30344a04e51e4edd99221add/auth/gcp/redact.go#L191-L197) excuses any value lying wholly inside marker bytes, and a service that writes an escaped marker gets real marker bytes once `decodeFully` runs:

```
UserID "]"         body `\u005bredacted\u005d`             -> decodes to "[redacted]",           which contains "]"
UserID "ted][red"  body `\u005bredacted]\u005bredacted]`   -> decodes to "[redacted][redacted]", which contains "ted][red"
```

The first of those does the same thing before this PR, so the carve-out is not new — your own [`"e"` control row](https://github.com/google/adk-go/blob/1e18d7e5e62d80ad30344a04e51e4edd99221add/auth/gcp/unescape_test.go#L477-L480) is the same shape. The second one is new: with the old predicate it scrubs to `[redac[redacted]acted]` and the value is gone, and with the new one it is not. Nothing about the value escapes either way, because the returned bytes are the service's own and do not vary with what the UserID was.

I would not change the code for it. `allOurs` requires every byte of the occurrence to be marker bytes, so an excused value has to be a substring of `[redacted][redacted]…` — eight distinct letters, none of them `@`, `:` or `/`. No email and no redirect URI can reach that class, which is both secrets this path actually carries. The sentence is what is wrong, and naming the carve-out in it covers the `"e"` case and this one together. Deciding on the output rather than on `redact`'s own bookkeeping is the part worth keeping whatever the wording becomes.

Two comments are on the lines they concern.

auth/gcp/unescape_test.go
wolo-lab: Re-frozen rather than re-described, because I think this is a stale oracle and not only a stale comment. `readable` earns its place by catching a future weakening of `recoverable`, and it can only do that while it is a copy — so when the predicate moved from splitting to masking, the oracle lost coverage of exactly the direction that commit was about, not just an accurate sentence.

It masks marker bytes by position now. Six shapes checked against the predicate, all agreeing, including the excused `ted][red` case from the review body.

You were right that nothing was unguarded in the meantime: reverting `recoverable` to the split form still fails all three leak rows, so what had gone was the guard's future value rather than its present one. Both sites are updated together, since the last time one of these went stale it was because only one of the two was corrected.

auth/gcp/client.go
wolo-lab: Wrong in both directions, as you have it. I replicated the old guard against the tag rather than reasoning from the diff:

```
                                                  v2.3.0   head
projects/example..com/locations/x/connectors/y    reject   accept   <- movement I omitted
projects/p/locations/../connectors/y              reject   reject   <- claimed as moved, did not
projects/p/locations//connectors/y                accept   reject
projects/p/locations/./connectors/y               accept   reject
projects/p/locations/x/connectors/y/              accept   reject
projects/example.com:my-project/...               reject   accept
```

Four moved and one did not. The paragraph lists them individually now instead of sorting them into two groups and describing each group as a whole, which is what made it false — a blanket claim over a mixed set only needs one member to be wrong.

The widening is pinned where you said. I had grepped for `example..com` and the fixture is `a..b`, which is the same shape, and its own comment already draws the distinction between dots inside a segment and a segment made of dots. The migration sentence was the only place that lost it.

wolo-lab:
All three taken, and one of them wanted a code change rather than a wording change. Head is [92b2ce51](https://github.com/google/adk-go/commit/92b2ce5157662056d27a71a966ccbbe3a46fb480). The two line comments are answered on their threads.

Both of your bodies reproduce, and the second is new in this PR as you say.

```
UserID "]"         body \u005bredacted\u005d           -> decodes to "[redacted]"
UserID "ted][red"  body \u005bredacted]\u005bredacted] -> decodes to "[redacted][redacted]"
```

The code stays, and your bound on the excused class is what settles it rather than the reproductions. An occurrence has to be marker bytes end to end, so what can be excused is the substrings of `[redacted][redacted]…` — eight distinct letters, no `@`, no `:`, no `/`. Neither value this path carries can be spelled out of that alphabet, which turns the carve-out from something a caller has to reason about into something they can rule out by looking at their own values.

[`APIError.Body`](https://github.com/google/adk-go/blob/92b2ce5157662056d27a71a966ccbbe3a46fb480/auth/gcp/client.go#L116-L128) states the exception alongside the guarantee now, says why it exists rather than only that it does, carries the bound so a reader can check against it, and notes that a service writing its own marker is excused the same way because there is no telling its markers from ours.

The one thing I want to keep whatever the wording becomes is the part you named: deciding on the output rather than on `redact`'s own bookkeeping. That is what makes the answer independent of what the service chose to put in the bytes that were measured, and it is the property the earlier revisions of this kept losing.

karolpiotrowicz (approved):
Approving at [92b2ce51](https://github.com/google/adk-go/commit/92b2ce5157662056d27a71a966ccbbe3a46fb480). All three are closed, and I ran the two that could still have been wrong rather than reading them.

**The oracle is re-frozen and the drift is gone.** `readable` now excuses marker bytes by position, which is the choice `recoverable` makes, so the counterexample from last round agrees in both directions:

```
readable   ("a[redacted]b", []string{"a[redacted]b"}) -> true   (was false)
recoverable("a[redacted]b", []string{"a[redacted]b"}) -> true
```

I put six more shapes through both predicates — the `"e"` control row, the `ted][red` case, the escaped-marker body, the partly-legible URI, a plain leak, and a body with nothing in it. All seven agree. The sentence about re-freezing whenever the original moves is what keeps this from drifting again, and it is now on both comment sites.

**The v2.3.0 paragraph measures exactly as written.** I reimplemented the guard as it stood at the tag — [`resourceNameRE`](https://github.com/google/adk-go/blob/v2.3.0/auth/gcp/client.go#L56) plus the [`".."` substring test](https://github.com/google/adk-go/blob/v2.3.0/auth/gcp/client.go#L195) — and ran the six shapes through it and through `validateResource` at this head:

```
                                                        v2.3.0   head
projects/p/locations/l/connectors/c/     trailing slash  accept   reject
projects/p//locations/l/connectors/c     doubled slash   accept   reject
projects/p/./locations/l/connectors/c    "." segment     accept   reject
projects/p/locations/l/connectors/c:run  colon           reject   accept
projects/example..com/locations/x/...    ".." in segment reject   accept
projects/../locations/l/connectors/c     ".." segment    reject   reject
```

Four moved and one did not, which is what the new sentence claims, and the direction is right on each.

**The carve-out sentence does its job.** Naming the excused class as the substrings of `[redacted][redacted]…` covers the `"e"` control row and the `ted][red` case in one go, and the reason it cannot reach a real value is now in the doc instead of only in a review thread. The eight distinct characters are `[ ] a c d e r t`, so no address and no redirect URI is expressible in them.

`go test -race -count=1 ./auth/gcp/...` passes and `gofmt -l auth/gcp/` is empty at this head.

PR #1521, opened 2026-09-07
fix(agent): keep the invocation when WithICDelta returns nil

wolo-lab:
## Problem

An `InvocationContext` is free to return nil from `WithICDelta`. Nothing in this repository does — the guard is for implementations written outside it, which the exported interface allows and which cannot be enumerated. The shape is easy to reach by accident, because a decorator that embeds `InvocationContext` has to override `WithICDelta` or lose itself on the first delta: the promoted method has the inner receiver, so it returns the inner type and the wrapper is gone.

`WithDelta` and `WithICDelta` stored that nil. The result is a `commonContext` whose invocation is gone, and most of its accessors dereference it: `Agent`, `Branch`, `Session` and `UserID` all nil-panic on the next access.

On the workflow path, where that panic surfaces depends on the node, and for one kind it is not contained at all. `runNode` calls `startNodeSpan` at [`scheduler.go:444`](https://github.com/google/adk-go/blob/729b5a8f97beb32d1b8cfe46d7af57f0406f7dac/workflow/scheduler.go#L444) and installs its `recover()` at [line 451](https://github.com/google/adk-go/blob/729b5a8f97beb32d1b8cfe46d7af57f0406f7dac/workflow/scheduler.go#L451). For a plain node, `startNodeSpan` reaches `commonContext.Session()` — the first thing in node execution to dereference the invocation — so the panic is raised before the recovery exists, on a goroutine of its own, and is never turned into a `node %q panicked` completion error. Measured on the merge base, a two-node graph (`Start` → fan-out → `ParallelWorker`) through `Workflow.Run` kills the process outright, taking every unrelated concurrent session with it. A node that emits its own span is different: [`node_span.go:40-41`](https://github.com/google/adk-go/blob/729b5a8f97beb32d1b8cfe46d7af57f0406f7dac/workflow/node_span.go#L40-L41) hands back a noop span and the untouched context, so an agent node's panic lands inside `Run`, is recovered, and surfaces as `node "a1" panicked: runtime error: invalid memory address or nil pointer dereference` with the process intact. With this change both graphs complete.

## Summary

Keep the previous invocation when `WithICDelta` hands back nil, and report the discard.

Keeping it is the better of the two options, but it is not free: the delta is gone, so the caller runs on with the previous `Agent`, `Branch` and `IsolationScope`. Nothing else separates that from the delta having been applied, which is why it is reported rather than swallowed. An agent running under the wrong parent is not a quiet kind of wrong.

The report names the implementation's type and which delta fields did not reach the invocation, with the branch and isolation scope they carried. Beyond that type name it reads nothing off the invocation, because an implementation that has just returned nil is partial by definition and decorating the message from its accessors risks a second failure inside the error path. It is emitted once per implementation type: `withICDelta` runs on every derivation — per workflow node, per agent activation, per parallel item — and a nil return is a static property of the implementation rather than a transient, so reporting each occurrence would bury the operator without adding anything.

A delta carrying no `InvocationContextDelta` is still forwarded to the implementation, and only the report is skipped. `workflow.Run` and dynamic nodes both pass that shape. Skipping the call instead looked tempting and is wrong: both wrappers in this package delegate to the inner `commonContext`, which answers a nil delta by returning itself, so not calling leaves the wrapper in place and its `Agent()` returns nil. The report is what gets skipped, because the reporter reads the delta and would panic on nil inside the error path.

`TestDeltaOnInvocationThatReturnsNil` covers all three entry points and fails on the merge base — each of the four cases panics there with a nil pointer dereference. `TestDeltaReachesTheInvocation` is the control: a guard that kept the original unconditionally would satisfy every assertion in the first test.

### The open question from #1173, now settled

Whether to log the discard or let the nil propagate and take the panic was left open. The measurement above answers it: on the workflow path a plain node's panic is not recovered today, so letting it propagate does not fail one node, it ends the process. Reporting the discard is the better trade until the recovery ordering is fixed.

### Deliberately out of scope

**The recovery ordering in `workflow/scheduler.go` is a separate PR.** It predates this change, it lives in a file this PR does not touch, and correcting it is not a line move — the deferred body calls `span.recordError`, and `span` does not exist until `startNodeSpan` returns, so the span has to be initialised to a noop before the defers are installed. Folding it in here would mix an unrelated fix into a two-file change.

**Sharing the previous invocation is addressed by #1135.** The fallback hands every derived context the same invocation object where the working path hands each one a copy, so on the parallel path two workers can touch one `endInvocation` flag. #1135 replaces that flag with an `atomic.Bool` shared deliberately across derived contexts, which removes the race and makes the propagation intended behaviour. The two changes apply cleanly together with no overlapping files.

---

Third of a five-part split of #1173. Stacked on 1520.


wolo-lab:
@karolpiotrowicz , PTAL

agent/common_context_delta.go
```diff
@@ -78,6 +79,34 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+// applied, so it is logged. An agent running under the wrong parent is not a
+// quiet kind of wrong.
+func withICDelta(ic InvocationContext, d *InvocationContextDelta) InvocationContext {
+	if ic == nil {
+		return nil
+	}
+	if next := ic.WithICDelta(d); next != nil {
+		return next
+	}
+	log.Printf("agent: %T.WithICDelta returned nil; keeping the previous invocation and "+
```
wolo-lab, line 109: In this semester we are going to replace log.Printf with the right logging with levels etc.

agent/common_context_delta.go
```diff
@@ -78,6 +79,34 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+		"discarding the delta, so Agent, Branch and IsolationScope are unchanged", ic)
```
karolpiotrowicz: The message names three of the five fields `InvocationContextDelta` carries — `UserContent` and `Context` are dropped too and neither is mentioned, so a delta that only rewrote the user turn reports three things that are accurate and stays silent about the one that moved.

It also carries none of the values. The description says the report names "what the delta actually asked for, with the branch and isolation scope it carried", and the format string has a single `%T`. I checked that reading them is safe here: `d.Branch` and `d.IsolationScope` are caller-supplied pointers on a struct the caller built, so formatting them can't re-enter the implementation that just returned nil — which I think was the worry behind keeping the message to the type name.

agent/common_context_delta.go
```diff
@@ -78,6 +79,34 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+	res.invocationContext = withICDelta(res.invocationContext, d)
 	return &res
 }
+
+// withICDelta applies d to the invocation this context speaks for, and refuses a
+// nil result.
+//
+// An InvocationContext is free to return nil from WithICDelta, and a partial
+// implementation does: ADK's own ContextMock returns nil from WithContext,
+// WithBranch and WithAgentContext, so the shape is not exotic. Storing that nil
```
karolpiotrowicz: The colon reads as offering `ContextMock` as an implementation that returns nil from `WithICDelta`, and it doesn't — `ContextMock.WithICDelta` returns its receiver. The three methods named do return nil, so the sentence is true as written, but the inference it invites isn't. The description puts it more accurately: nothing in this repository returns nil from this method, and the guard is for implementations written outside it.

agent/common_context_delta_test.go
```diff
@@ -0,0 +1,123 @@
...
+		out := log.Writer()
+		log.SetOutput(&buf)
+		t.Cleanup(func() { log.SetOutput(out) })
+
+		c := PromoteWithDelta(ic, delta())
+		if got := c.Branch(); got == branch {
+			t.Fatalf("Branch() = %q, so the delta was applied after all and this test no "+
+				"longer covers what it is named for", got)
+		}
+		if got := buf.String(); !strings.Contains(got, "discarding the delta") {
```
karolpiotrowicz: This assertion runs after a single derivation, so it passes identically whether the report is emitted once ever, once per implementation type, or once per derivation. That makes it the one place the once-per-type behaviour described in the PR could have been pinned, and it doesn't distinguish it.

The working path has the mirror gap: `TestDeltaReachesTheInvocation` never captures the log, so an implementation that reported a discard on every derivation regardless of the result would pass this whole file.

Separately, `RunID()` and `SubScheduler()` have no assertions anywhere in `agent/*_test.go`, so a mutant that bailed out of `WithDelta` at line 50 and dropped them would survive. `Path()` and `OutputForAncestors()` are already covered by `agent/common_context_test.go`, so it's just those two.

agent/common_context_delta_test.go
```diff
@@ -0,0 +1,123 @@
...
+		})
+	}
+
+	// The discard is the cost of keeping the invocation, and nothing in the
+	// assertions above separates it from the delta having been applied. Pinned on
+	// the log, which is the only thing that does.
+	t.Run("the discard is reported", func(t *testing.T) {
+		var buf bytes.Buffer
+		out := log.Writer()
+		log.SetOutput(&buf)
```
karolpiotrowicz: `captureLog` in `agent/callback_context_wrapper_test.go` does this and also saves and restores the flags and prefix. Not a live problem — this test isn't `t.Parallel()` and `-race` is green on the package — but the helper is there.

karolpiotrowicz (dismissed):
Three sentences in the description are not true of the code at [cae1c8f](https://github.com/google/adk-go/commit/cae1c8f7e7449b729a4105034b75e2358e6d8fa0), and one of them is the per-type deduplication — which means the log volume the description argues against is what the branch actually does. That plus the shared invocation object are what I'd want resolved before this lands. The fix itself works: I drove a two-node graph through `Workflow.Run` with a nil-returning invocation and the process dies on the merge base and completes with this applied, so every reachable path here is the same as base or better.

**The description and the code have drifted apart.** Taking the three in turn, all measured at [cae1c8f](https://github.com/google/adk-go/commit/cae1c8f7e7449b729a4105034b75e2358e6d8fa0):

- `Promote(&StrictContextMock{}).WithDelta(&CommonContextDelta{RunID: &id})` still panics `unimplemented`. [`withICDelta`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/agent/common_context_delta.go#L102-L112) guards only `ic == nil`, so a nil `InvocationContextDelta` is passed straight through to [`StrictContextMock.WithICDelta`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/agent/context_mock.go#L145-L147). This one is not a regression — the pre-change line called through the same way — so it reads as a claim about a revision that isn't pushed.
- "Emitted once per implementation type" isn't implemented. Five derivations on one type plus three on another produce eight lines, and there's no `sync.Once`, seen-type set or rate limit in the file.
- "Names … what the delta actually asked for, with the branch and isolation scope it carried" — the format string has a single `%T` and reads nothing off `d`.

Worth flagging because the backport workflow replays the squash commit onto `v1`, so the description travels with the change twice, and the template's "Behavior change" question is answered wrongly by the first one.

**The log volume is the part I'd fix first.** Measured on `Start -> split -> ParallelWorker -> join` driven through `Workflow.Run`, with the same graph under a working implementation as the control:

| fan-out items | nil-returning invocation | working invocation |
|---|---|---|
| 10 | 14 lines | 0 |
| 100 | 104 lines | 0 |
| 1000 | 1004 lines | 0 |

So it's one line per node activation plus one per parallel item, every run, each taking the `log` package's process-global mutex inside the derivation path.

One thing that saves a wasted round: I tried the obvious `if d == nil { return ic }` guard and re-measured, and it does **not** help here — the fan-out numbers are unchanged at 14/104/1004, because every derivation on the parallel path carries a `Branch` delta. It removes exactly one line from a named workflow's promote. The nil-delta guard and the deduplication are two separate edits.

<details><summary>Reproduction — fan-out log volume, with control</summary>

```go
// workflow/zz_fanout_test.go
type wfNilIC struct{ agent.InvocationContext }

func (wfNilIC) WithICDelta(*agent.InvocationContextDelta) agent.InvocationContext { return nil }

func TestFanoutLogVolume(t *testing.T) {
	up := func(ctx agent.Context, in string) (string, error) { return strings.ToUpper(in), nil }
	upper := NewFunctionNode("upper", up, defaultNodeConfig)
	for _, n := range []int{10, 100, 1000} {
		split := NewFunctionNode("split", func(ctx agent.Context, in string) ([]any, error) {
			res := make([]any, 0, n)
			for i := 0; i < n; i++ {
				res = append(res, "x")
			}
			return res, nil
		}, defaultNodeConfig)
		pw, _ := NewParallelWorker("parallel", upper, 0, defaultNodeConfig)
		join := NewFunctionNode("join", func(ctx agent.Context, in []any) (string, error) { return "ok", nil }, defaultNodeConfig)
		w := mustNew(t, []Edge{{From: Start, To: split}, {From: split, To: pw}, {From: pw, To: join}})

		base := newMockCtx(t)
		base.userContent = &genai.Content{Parts: []*genai.Part{{Text: "seed"}}}

		var buf bytes.Buffer
		out := log.Writer()
		log.SetOutput(&buf)
		for range w.Run(wfNilIC{InvocationContext: base}) {
		}
		fallback := strings.Count(buf.String(), "discarding the delta")
		buf.Reset()
		for range w.Run(base) {
		}
		control := strings.Count(buf.String(), "discarding the delta")
		log.SetOutput(out)
		t.Logf("%4d items: fallback=%d control=%d", n, fallback, control)
	}
}
```

```
go test ./workflow/ -run TestFanoutLogVolume -v

    10 items: fallback=14 control=0
   100 items: fallback=104 control=0
  1000 items: fallback=1004 control=0
```
</details>

**Keeping the previous invocation hands the same object to every sibling derivation.** On the working path [`(*invocationContext).WithICDelta`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/agent/agent.go#L393-L397) does `res := *c`, so each child owns its own `endInvocation`. The fallback returns `ic` itself, and two things follow, both executed with the working path as a matched control in the same file:

- A data race between [`EndInvocation`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/agent/agent.go#L452-L454) and [`Ended`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/agent/agent.go#L456-L458). The control passes clean, the fallback trips `-race`.
- Deterministic cross-sibling termination: after one derived context calls `EndInvocation()`, on the working path the sibling and the parent both report `Ended() == false`, and on the fallback both report `true`. [`agent.go:190`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/agent/agent.go#L190) and [`:203`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/agent/agent.go#L203) return early on that.

The sibling shape is real — [`scheduler.go:378`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/workflow/scheduler.go#L378) derives per node and [`:405`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/workflow/scheduler.go#L405) launches each on its own goroutine, and [`parallel_worker.go:150-151`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/workflow/parallel_worker.go#L150-L151) does the same per item. I drove the race at the derivation API rather than through `Workflow.Run`, so the production composite is inference, not a run.

On the deferral to #1135: that PR makes the propagation deliberate, so the second bullet above is a behaviour you've decided to want rather than one to fix — I'm only reporting it because in this branch it arrives without the `atomic.Bool` and without the locality comment at [`agent.go:356-361`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/agent/agent.go#L356-L361) being updated. The first bullet is the one that has no cover: #1135 is still open, so nothing in a merge of this branch alone synchronises that write. Worth a line in the doc comment at minimum, since the comment lists the cost as Agent, Branch and IsolationScope and the aliasing is the part a reader can't recover by inspecting a value.

<details><summary>Reproduction — cross-sibling termination and the race, both with controls</summary>

```go
// agent/zz_share_test.go
type nilIC struct{ InvocationContext }

func (nilIC) WithICDelta(*InvocationContextDelta) InvocationContext { return nil }

func TestSiblingsShareTheInvocation(t *testing.T) {
	log.SetOutput(io.Discard)
	b1, b2 := "b1", "b2"

	// control: working implementation, each derivation gets a copy
	enc := &invocationContext{Context: t.Context(), agent: &agent{name: "p"}, branch: "b"}
	n1 := Promote(enc).WithICDelta(&InvocationContextDelta{Branch: &b1})
	n2 := Promote(enc).WithICDelta(&InvocationContextDelta{Branch: &b2})
	n1.EndInvocation()
	t.Logf("control:  sibling=%v parent=%v", n2.Ended(), enc.Ended())

	// probe: fallback path
	enc2 := &invocationContext{Context: t.Context(), agent: &agent{name: "p"}, branch: "b"}
	ic := nilIC{InvocationContext: enc2}
	f1 := Promote(ic).WithICDelta(&InvocationContextDelta{Branch: &b1})
	f2 := Promote(ic).WithICDelta(&InvocationContextDelta{Branch: &b2})
	f1.EndInvocation()
	t.Logf("fallback: sibling=%v parent=%v", f2.Ended(), enc2.Ended())
}

func TestRaceControl(t *testing.T) { // working path, expect clean under -race
	log.SetOutput(io.Discard)
	enc := &invocationContext{Context: t.Context(), agent: &agent{name: "p"}, branch: "b"}
	b1, b2 := "b1", "b2"
	c1 := Promote(enc).WithICDelta(&InvocationContextDelta{Branch: &b1})
	c2 := Promote(enc).WithICDelta(&InvocationContextDelta{Branch: &b2})
	var wg sync.WaitGroup
	wg.Add(2)
	go func() { defer wg.Done(); c1.EndInvocation() }()
	go func() { defer wg.Done(); _ = c2.Ended() }()
	wg.Wait()
}

func TestRaceProbe(t *testing.T) { // fallback path
	log.SetOutput(io.Discard)
	enc := &invocationContext{Context: t.Context(), agent: &agent{name: "p"}, branch: "b"}
	ic := nilIC{InvocationContext: enc}
	b1, b2 := "b1", "b2"
	c1 := Promote(ic).WithICDelta(&InvocationContextDelta{Branch: &b1})
	c2 := Promote(ic).WithICDelta(&InvocationContextDelta{Branch: &b2})
	var wg sync.WaitGroup
	wg.Add(2)
	go func() { defer wg.Done(); c1.EndInvocation() }()
	go func() { defer wg.Done(); _ = c2.Ended() }()
	wg.Wait()
}
```

```
go test ./agent/ -run TestSiblingsShareTheInvocation -v

    control:  sibling=false parent=false
    fallback: sibling=true parent=true

go test -race ./agent/ -run 'TestRaceControl|TestRaceProbe' -v

--- PASS: TestRaceControl
=== RUN   TestRaceProbe
WARNING: DATA RACE
Read at 0x00c0002361d0 by goroutine 14:
  google.golang.org/adk/v2/agent.(*invocationContext).Ended()
      agent/agent.go:457 +0x2e
Previous write at 0x00c0002361d0 by goroutine 13:
  google.golang.org/adk/v2/agent.(*invocationContext).EndInvocation()
      agent/agent.go:453 +0x2e
--- FAIL: TestRaceProbe
```
</details>

**The doc comment contradicts the description, and the description is the one that's right.** [Lines 93-95](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/agent/common_context_delta.go#L93-L95) say the scheduler catches the panic and reports it as `node %q panicked`. For a plain node it doesn't: [`runNode`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/workflow/scheduler.go#L444) calls `startNodeSpan` before installing its `recover()` at [line 451](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/workflow/scheduler.go#L451-L454), and the span reaches `Session()` on the nil invocation. On the merge base my two-node graph dies with SIGSEGV through exactly that path, on the goroutine created at line 405, and no completion error is produced. It is true for `Start` and for nodes that set `EmitsOwnSpan`, which [`startNodeSpan`](https://github.com/google/adk-go/blob/cae1c8f7e7449b729a4105034b75e2358e6d8fa0/workflow/node_span.go#L39-L41) short-circuits to a noop span — so the comment is over-general rather than backwards, and it understates the hazard in the paragraph that exists to justify the guard.

**One question rather than a request.** The justification for keeping the invocation is that the alternative is an undiagnosable crash naming the node rather than the cause. A panic carrying `%T.WithICDelta returned nil` delivers that same diagnosis and removes the log volume, the aliasing and the widened branch and scope in one go, at the cost of fail-fast on a path where the alternative outcome is a context you already know is wrong. I'm not asking you to change it — you've thought about this since #1173 and settled it deliberately — but the settled reasoning isn't in the file, and the trade reads differently now that the fallback's own costs are measured.

Two smaller things are on the lines themselves.

For what it's worth, the rest holds up well. No exported API moves — `apidiff` on the package is empty with a live control. The new tests are a genuine red→green: all four cases panic on the merge base and pass with this applied, and `TestDeltaReachesTheInvocation` is a real control that a guard keeping the original unconditionally would fail. Coverage on the file goes up, the branch merges cleanly onto current main and the suite is green on the merge result.

agent/common_context_delta.go
wolo-lab: Fixed. The report now names all five, and renders the values for the two that have them: `(Agent, Branch="child-branch", IsolationScope="child-scope", UserContent)`.

`Context` needed splitting rather than adding, because the two entry points differ. `WithDelta` installs `d.Context` on the context it returns whether or not the invocation accepted the delta, so on that path the caller keeps it and naming it as lost is false — `c.Value(key)` returns the delta's value. `WithICDelta` installs nothing, so there it genuinely is lost. [`withICDelta`](https://github.com/google/adk-go/blob/defd9501f19676339c8318a3b474b0cddb029ea9/agent/common_context_delta.go#L119) now takes `callerAppliesContext` and the two call sites pass what they actually do. Both directions are pinned by [`TestDiscardNamesContextOnlyWhereItIsLost`](https://github.com/google/adk-go/blob/defd9501f19676339c8318a3b474b0cddb029ea9/agent/common_context_delta_test.go#L295).

Your reasoning about why reading the values is safe matches why I was willing to render them: the delta is a struct the caller built, so formatting it cannot re-enter the implementation that just returned nil. That is why `Agent` is still a bare label — `(*d.Agent).Name()` would be a call into the caller's object inside the error path, which is the thing worth avoiding.

agent/common_context_delta.go
wolo-lab: Fixed. The sentence now says plainly that nothing in this repository returns nil from this method and that the guard is for implementations written outside it, with the reason they hit it: a decorator that embeds `InvocationContext` has to override `WithICDelta` or lose itself on the first delta.

agent/common_context_delta_test.go
wolo-lab: All three fixed.

Once-per-type is now pinned by [`TestDiscardReportIsEmittedOnce`](https://github.com/google/adk-go/blob/defd9501f19676339c8318a3b474b0cddb029ea9/agent/common_context_delta_test.go#L159), which derives three times from one implementation and asserts exactly one line. The field list has its own table in [`TestDiscardReportNamesWhatWasLost`](https://github.com/google/adk-go/blob/defd9501f19676339c8318a3b474b0cddb029ea9/agent/common_context_delta_test.go#L112), so mangling a label or pointing `%T` at the delta no longer passes.

The mirror gap is real and I had missed it. `TestDeltaReachesTheInvocation` now captures the log and requires it empty, so a version that reported on every derivation rather than only on a discard fails there.

`RunID` and `SubScheduler` were exactly as you describe. I confirmed a version that dropped both from `WithDelta` still passed every test in the package, and that `Path` and `OutputForAncestors` were already covered in `common_context_test.go`. They are asserted after a discard now, in [`TestDiscardKeepsTheRestOfTheDelta`](https://github.com/google/adk-go/blob/defd9501f19676339c8318a3b474b0cddb029ea9/agent/common_context_delta_test.go#L256).

One more thing that came out of looking at this. The report set has process lifetime, so the old assertion only held on the first run of the binary and `-count=2` failed. The tests reset it now.

agent/common_context_delta_test.go
wolo-lab: Switched to it. The inline capture is gone.

wolo-lab:
@karolpiotrowicz all four are addressed in `defd9501`. Pushing moved the lines your comments were anchored to, so the threads collapsed as outdated and my replies went with them — repeating the substance here so it is readable without unfolding anything.

**The message named three of five fields, and no values.** It now names all five and renders the values for the two that have them: `(Agent, Branch="child-branch", IsolationScope="child-scope", UserContent)`.

`Context` is the one place I did not do quite what you asked, because the two entry points differ. `WithDelta` installs `d.Context` on the context it returns whether or not the invocation accepted the delta, so there the caller keeps it and calling it lost would be wrong — `c.Value(key)` returns the delta's value. `WithICDelta` installs nothing, so there it really is lost and the report says so. [`withICDelta`](https://github.com/google/adk-go/blob/defd9501f19676339c8318a3b474b0cddb029ea9/agent/common_context_delta.go#L119) takes `callerAppliesContext` and each call site passes what it actually does.

`Agent` stays a bare label for the reason behind your own point about safety: `(*d.Agent).Name()` is a call into the caller's object from inside the error path, which is what I wanted to keep out. The two string fields are values on a struct the caller built, so rendering them cannot re-enter the implementation that just returned nil.

**The `ContextMock` sentence.** Rewritten. It now says nothing in this repository returns nil from this method and the guard is for implementations written outside it, with the reason they reach it — a decorator that embeds `InvocationContext` has to override `WithICDelta` or lose itself on the first delta.

**The test gaps.** All three were real. Once-per-type is pinned by [`TestDiscardReportIsEmittedOnce`](https://github.com/google/adk-go/blob/defd9501f19676339c8318a3b474b0cddb029ea9/agent/common_context_delta_test.go#L159), which derives three times and requires exactly one line. The mirror gap you spotted on the working path is closed: `TestDeltaReachesTheInvocation` now captures the log and requires it empty, so a version reporting on every derivation fails. `RunID` and `SubScheduler` were exactly as you said — I confirmed a version dropping both from `WithDelta` passed every test in the package, and that `Path` and `OutputForAncestors` were already covered in `common_context_test.go`. They are asserted after a discard in [`TestDiscardKeepsTheRestOfTheDelta`](https://github.com/google/adk-go/blob/defd9501f19676339c8318a3b474b0cddb029ea9/agent/common_context_delta_test.go#L256).

**`captureLog`.** Switched to it, inline capture gone.

One thing that came out of this and nobody had raised: the report set has process lifetime, so the old assertion only held on the first run of the binary and `go test -count=2 ./agent/` failed. The tests reset it now.

PTAL when you have a moment — your `CHANGES_REQUESTED` is still the only thing blocking.

agent/common_context_delta.go
```diff
@@ -78,6 +85,90 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+		return nil
+	}
+	if next := ic.WithICDelta(d); next != nil {
+		return next
+	}
+	// d is forwarded above even when nil, because an implementation may treat a
+	// nil delta as something other than "no change" — both wrappers in this
+	// package delegate, and the inner commonContext answers a nil delta by
+	// returning itself, which unwraps them. Only the report is skipped.
+	if d != nil {
```
karolpiotrowicz, line 131: This guard is the one thing I'd want a test on before merge. Delete it and `go test ./agent/` stays green, but `reportDiscardedDelta` then dereferences a nil `d` at the `d.Agent != nil` on line 148 — a nil-pointer panic inside the error path, which is the thing the comment just above says the enrichment was kept minimal to avoid.

The shape is live: [dynamic_node.go:112-117](https://github.com/google/adk-go/blob/aea8e11373205b119a2f10ec1bf3b826ef56bbeb/workflow/dynamic_node.go#L112-L117) builds a `CommonContextDelta` with no `InvocationContextDelta` on every dynamic-node activation, and [workflow.go:302-308](https://github.com/google/adk-go/blob/aea8e11373205b119a2f10ec1bf3b826ef56bbeb/workflow/workflow.go#L302-L308) does the same on workflow entry when the branch at [workflow.go:291](https://github.com/google/adk-go/blob/aea8e11373205b119a2f10ec1bf3b826ef56bbeb/workflow/workflow.go#L291) is taken. Against an implementation that returns nil regardless of its argument, that reaches here with `d == nil`.

The invariant, since it bounds any fix: `reportDiscardedDelta` must never be entered with a nil `d`, and the suite must fail if it is. `TestNilDeltaStillReachesTheInvocation` does not cover this — its inner returns itself, so line 123 returns and this line never executes. The commit message credits it with pinning both halves and it pins only the forwarding one.

```go
// agent/zz_guard_test.go
type gNilIC struct{ InvocationContext }

func (gNilIC) WithICDelta(*InvocationContextDelta) InvocationContext { return nil }

func TestGuardIsLoadBearing(t *testing.T) {
	enc := &invocationContext{Context: t.Context(), agent: &agent{name: "p"}}
	path := "wf/n@1"
	// Must not panic: the delta carries no InvocationContextDelta.
	_ = PromoteWithDelta(gNilIC{InvocationContext: enc}, &CommonContextDelta{Path: &path})
}
```

With the guard: passes. With this line removed: `panic: runtime error: invalid memory address or nil pointer dereference`.

agent/common_context_delta.go
```diff
@@ -78,6 +85,90 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+	}
+	if d.Context != nil && !callerAppliesContext {
+		lost = append(lost, "Context")
+	}
+	if len(lost) == 0 {
+		return
+	}
+	// Claim the type only once there is something worth saying, so a delta that
+	// asked for nothing cannot spend the one report a later, informative one needs.
+	if _, dup := reportedNilICDelta.LoadOrStore(reflect.TypeOf(ic), struct{}{}); dup {
```
karolpiotrowicz: Keying on the type alone means the first discard to produce a non-empty list is the only one anyone ever sees, and on the workflow path that first one is the least informative. Driving `Workflow.Run` against a nil-returning implementation, the single line I get is:

```
agent: workflow.rNilIC.WithICDelta returned nil, so the previous invocation is kept and this delta is discarded (Branch=""). Further occurrences from this type are not reported
```

Empty branch, because [scheduler.go:378-382](https://github.com/google/adk-go/blob/aea8e11373205b119a2f10ec1bf3b826ef56bbeb/workflow/scheduler.go#L378-L382) derives Branch-only per node and the root run's branch is `""`. A subsequent derivation that loses the Agent — [agent.go:177](https://github.com/google/adk-go/blob/aea8e11373205b119a2f10ec1bf3b826ef56bbeb/agent/agent.go#L177), [llmagent.go:452](https://github.com/google/adk-go/blob/aea8e11373205b119a2f10ec1bf3b826ef56bbeb/agent/llmagent/llmagent.go#L452) — then emits nothing at all. That is the case the paragraph at line 113 picks out as the reason the diagnostic exists.

Non-blocking, and I can see the argument the other way: the line does say suppression is in effect, and "this type returns nil and its deltas are being discarded" is the actionable part, which survives every ordering. What isn't true is the claim at lines 138-140 that later occurrences tell the operator nothing the first line did not — they carry different field lists. Keying on the type plus the set of field *names* would keep a 1000-item fan-out at one line while letting a materially different loss through, if you want the stronger version.

agent/common_context_delta.go
```diff
@@ -78,6 +85,90 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
 		return c
 	}
 	res := *c
-	res.invocationContext = res.invocationContext.WithICDelta(d)
+	// false: this entry point never installs d.Context on the returned context,
+	// so a discarded delta loses it.
+	res.invocationContext = withICDelta(res.invocationContext, d, false)
```
karolpiotrowicz: `callerAppliesContext` is decided by the receiver's concrete type rather than by what was actually lost, because `Promote` flattens a `*commonContext` while `NewToolContext` returns a `*toolContextWrapper`. A wrapper therefore reaches this line and its hardcoded `false`, even though the outer `WithDelta` at line 56 passed `true` and will install `d.Context` itself at line 60.

Same fixtures, same delta `{Agent, Context}`, one wrapped and one not:

```
UNWRAPPED     report=discarded (Agent)           caller.Value=from-the-delta  invocation.Value=from-parent
WRAPPED-tool  report=discarded (Agent, Context)  caller.Value=from-the-delta  invocation.Value=from-parent
```

(`WRAPPED-callback` matches `WRAPPED-tool`.) Identical observable state, different reports. It doesn't matter which reading of "lost" you take — if the subject is the invocation then the unwrapped line is the wrong one, and if it is the caller's returned context then the wrapped line is. The parameter exists to make exactly that distinction and one of the two comes out wrong either way.

Non-blocking: I could not construct an in-repo caller that reaches here with a `Context`-bearing delta, so it needs an out-of-repo one on top of the out-of-repo nil-returning implementation. The direct-path asymmetry is genuinely pinned, by the way — hardcoding `true` here kills `TestDiscardNamesContextOnlyWhereItIsLost`.

agent/common_context_delta.go
```diff
@@ -78,6 +85,90 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+//
+// Storing that nil leaves a commonContext whose invocation is gone, and most of
+// its accessors — Agent, Branch, Session, UserID among them — dereference it.
+// Where that panic surfaces depends on the node: a plain workflow node
+// dereferences the invocation inside startNodeSpan, which runNode calls before
+// installing its recover, so the process dies. A node that emits its own span
+// gets a noop span and an untouched context, so the panic happens inside Run
+// and is recovered as "node %q panicked".
+//
+// Keeping the previous invocation is the better of the two, but it is not free:
```
karolpiotrowicz, line 108: This is the paragraph I asked about last round, and the cost list still reads Agent, Branch and IsolationScope. The fallback also hands every derived context the same invocation object where the working path gives each one a copy, and that is the only item a reader cannot recover by inspecting a value — which is why it is worth a sentence here even though the behaviour itself is #1135's.

Still reproducing at this head, with the working path as the control: after one derived context calls `EndInvocation()`, the control reports `sibling=false parent=false` and the fallback reports `true`/`true`, and `-race` is clean on the control and reports a race between [the write](https://github.com/google/adk-go/blob/aea8e11373205b119a2f10ec1bf3b826ef56bbeb/agent/agent.go#L452-L454) and [the read](https://github.com/google/adk-go/blob/aea8e11373205b119a2f10ec1bf3b826ef56bbeb/agent/agent.go#L456-L458) on the fallback.

agent/common_context_delta_test.go
```diff
@@ -0,0 +1,326 @@
...
+		})
+	}
+}
+
+// TestDiscardReportIsEmittedOnce pins the deduplication itself. Deleting the
+// LoadOrStore guard leaves every other test in this file green.
+func TestDiscardReportIsEmittedOnce(t *testing.T) {
+	freshReports(t)
+	enclosing := &invocationContext{Context: t.Context(), agent: &agent{name: "parent"}}
+	ic := nilDeltaInvocation{InvocationContext: enclosing}
```
karolpiotrowicz, line 191: Small gap while you're adding the guard test. `ic` is bound once here and replayed three times in the loop below, so the assertion cannot tell the shipped per-type key from a per-instance one: change [common_context_delta.go:168](https://github.com/google/adk-go/blob/aea8e11373205b119a2f10ec1bf3b826ef56bbeb/agent/common_context_delta.go#L168) to key on `ic` itself and this test still passes, along with the rest of the package.

That matters in production rather than in principle — a server building one invocation per request from a non-conforming decorator type would go back to one log line per request, and the map would retain every one of those invocation objects for the life of the process, which the interned `reflect.Type` key is what prevents.

Deriving from two distinct values of the same type would close it.

agent/common_context_delta.go
```diff
@@ -78,6 +85,90 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+// the delta is gone, so the caller runs on with the previous Agent, Branch and
+// IsolationScope. Nothing else distinguishes that from the delta having been
+// applied, so it is reported. An agent running under the wrong parent is not a
+// quiet kind of wrong.
```
karolpiotrowicz, line 112: Missing a `//` separator before the `callerAppliesContext` paragraph, so it renders as a continuation of "…not a quiet kind of wrong." under `go doc -u`.

agent/common_context_delta.go
```diff
@@ -78,6 +85,90 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+// activation, per parallel item — and a nil return is a static property of the
+// implementation rather than a transient, so reporting every occurrence buries
+// the operator without telling them anything the first line did not.
+var reportedNilICDelta sync.Map // reflect.Type -> struct{}
+
+func reportDiscardedDelta(ic InvocationContext, d *InvocationContextDelta, callerAppliesContext bool) {
+	// Only what the delta asked for is safe to render. An implementation that
+	// has just returned nil is by definition partial, so calling its accessors
+	// to enrich this message risks a second failure inside the error path.
+	var lost []string
```
karolpiotrowicz: The field list is built before the dedup check at line 168, so a type that has already been reported still pays the slice and up to two `fmt.Sprintf` calls on every derivation. Measured with `go test ./agent/ -run XXX -bench BenchmarkDiscardSteadyState -benchmem` on go 1.26.6, delta `{Branch, IsolationScope}`: 723.8 ns/op, 313 B/op, 8 allocs/op.

I tried the obvious fix so you don't have to — a `reportedNilICDelta.Load(reflect.TypeOf(ic))` early return at the top of the function gives 180.5 ns/op, 176 B/op, 2 allocs/op with the package green, and it keeps the claim-after-build ordering intact because `Load` stores nothing. Only an implementation that returns nil pays any of this, so it is an error-path improvement rather than a defect.

karolpiotrowicz (changes requested):
The four threads are closed and I verified each one by running it rather than by reading the new code. What I'd want before this lands is one test: the `if d != nil` guard you added is load-bearing and nothing in the package distinguishes it from the version without it.

### Round-1 points, in turn

All measured at [aea8e11](https://github.com/google/adk-go/commit/aea8e11373205b119a2f10ec1bf3b826ef56bbeb). The log volume is fixed — I re-ran the same fan-out harness from last round and 10/100/1000 items now give 1/0/0 report lines against 14/104/1004 before, control still 0. The field list, the `ContextMock` sentence and `captureLog` are all done. The two-branch panic claim in the doc comment I executed both ways rather than taking the comment's word for it: with the guard neutralised a plain `FunctionNode` takes the process down with SIGSEGV, and a node with `EmitsOwnSpan` comes back as `node "a1" panicked: runtime error: invalid memory address or nil pointer dereference` with the process intact. The comment is right on both branches now. `freshReports` is load-bearing too — removing the calls fails 8 subtests under `-count=2`.

### A correction to something I said last round

This changes where the cost of a stale sentence falls. I claimed the PR description travels into the squash commit and then onto `v1`. That is wrong for this repo: `squash_merge_commit_message` is `COMMIT_MESSAGES` and `allow_merge_commit` is false, so what lands is the commit messages, not the description. So the description paragraph below is confined to this page, and the commit-message contradiction is the one that becomes permanent. I had those the wrong way round.

### Two sentences are still not true of the code

The description says `withICDelta` "stops calling through when the delta carries no `InvocationContextDelta` at all", and that `Promote(&StrictContextMock{}).WithDelta(&CommonContextDelta{RunID: &id})` "panicked before this change and does not now". Both were in last round's list and neither moved. It still panics `unimplemented`, and a counting fixture shows `WithICDelta` called once with a nil argument — which is what [the comment at :126](https://github.com/google/adk-go/blob/aea8e11373205b119a2f10ec1bf3b826ef56bbeb/agent/common_context_delta.go#L126-L129) says you decided to do deliberately. The code and that paragraph now disagree on purpose, so the paragraph is the part to drop.

Separately, the commit message contradicts itself on the same subject. "Context is not named as lost" sits several paragraphs above "Context is now named as lost on the entry point that loses it" — the `Review follow-ups.` section corrected the behaviour without correcting the earlier paragraph. Given the squash setting above, that is the one a reader of `git log` will hit. The list of changes that the tests do catch is right about "re-adding Context to the list", by the way: I applied that change and two tests fail.

### The shared invocation, and where that leaves us

Nothing changed here and I am not asking for anything beyond a line of prose. The race still reproduces at this head with the working path clean as a control, and #1135 is still open, so a merge of this branch alone still has nothing synchronising that write. You rewrote [the cost paragraph](https://github.com/google/adk-go/blob/aea8e11373205b119a2f10ec1bf3b826ef56bbeb/agent/common_context_delta.go#L111-L115) in this commit and it still enumerates the cost as Agent, Branch and IsolationScope. The shared object is the one item in that list a reader cannot recover by inspecting a value, which is why it is worth the sentence. The description handles it well — this is just the source copy.

### A design question rather than a defect

This one is yours to answer. `callerAppliesContext` cannot be made correct by fixing its call sites, because the frame that emits the report is the innermost one and cannot know what an outer frame will do afterwards. The inline note on [:90](https://github.com/google/adk-go/blob/aea8e11373205b119a2f10ec1bf3b826ef56bbeb/agent/common_context_delta.go#L90) has the measurement. Two designs delete the parameter: always name `Context` when `d.Context != nil`, which is accurate about the invocation in every case since this entry point never installs it, or never name it, which is what the previous revision did and you rejected for a stated reason. Given that the message's subject is the invocation — "the previous invocation is kept" — is the first the honest reading?

### What I could not settle

I could not construct an in-repo caller that reaches the wrapped frame with a `Context`-bearing delta. Every route into `agent.Run` I traced builds a fresh invocation first. It needs an out-of-repo caller, which is the same population the guard exists for, so I am reporting it rather than sizing it.

### One thing outside this PR that you own

#1522 targets `wolo/agent-delta-nil-guard` and carries an older `withICDelta` — two parameters, the round-1 log line, no deduplication. Merging this and then that as it stands would revert the whole second commit. Also worth knowing, since #1522 is stacked rather than on `main`: `apidiff.yml` filters on `branches: ["main", "v1"]`, so that check will not run there, unlike `go.yml`, which carries a comment explaining why it is deliberately unfiltered.

wolo-lab:
All four are in `df7cd6df`.

**The test you asked for.** You were right that nothing distinguished the `if d != nil` guard. Removing it passes the whole package, and the shape `Workflow.Run` and dynamic nodes pass — a `CommonContextDelta` with no `InvocationContextDelta` — panics without it, because the reporter dereferences `d`. Covered now by [`TestDiscardWithNoInvocationDelta`](https://github.com/google/adk-go/blob/df7cd6dff3c57b641e49b2d0939fc19aec412e1c/agent/common_context_delta_test.go#L333).

**The design question — you are right, and the reason is worse than you put it.** I checked the ordering: the report is emitted from `withICDelta` *before* `WithDelta` assigns `res.Context` at [:56-59](https://github.com/google/adk-go/blob/df7cd6dff3c57b641e49b2d0939fc19aec412e1c/agent/common_context_delta.go#L56-L59). So the flag was not merely knowledge the innermost frame lacks, it was a promise about something that had not happened yet, and reordering `WithDelta` would have made it lie with nothing to catch it.

I took your first option. The parameter is gone, `Context` is always named, and the message now says what is actually true on both entry points:

```
... returned nil, so the previous invocation is kept and these delta fields did not
reach it: Agent, Branch="outer@1.w@2", IsolationScope="orch/c@1", UserContent.
Further occurrences from this type are not reported
```

"did not reach it" is the honest scope — `WithDelta` does install `Context` on the context it returns, which is a different object, and that is the distinction the parameter was fumbling.

**The two stale sentences.** Both gone from the description. My own change had restored the call-through and I never went back to the paragraph that described the version without it.

**The commit message.** Rewritten as one coherent message. Your correction about which artifact travels is right and it changed what I fixed first — with `squash_merge_commit_message` set to `COMMIT_MESSAGES`, the contradiction in `git log` was the permanent one, not the description.

**The shared invocation.** The sentence is in the [cost paragraph](https://github.com/google/adk-go/blob/df7cd6dff3c57b641e49b2d0939fc19aec412e1c/agent/common_context_delta.go#L111-L117) now, naming the aliasing as the one cost a reader cannot recover by inspecting a value, and recording that #1135 is still open so nothing in this branch alone synchronises that write.

Nine mutants die against the new tests, including the four you would care about: removing the nil-delta guard, dropping the deduplication, mangling a field label, and reporting on every derivation.

On **#1522** — thank you for catching that, it would have reverted this commit wholesale. I am rebasing it onto this head now. Also noted on `apidiff.yml` filtering to `main`/`v1`, so that check will not run there.

agent/common_context_delta.go
wolo-lab: Added, and the invariant you stated is the one it pins. [`TestDiscardWithNoInvocationDelta`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta_test.go#L391) drives `PromoteWithDelta` with a `CommonContextDelta` carrying only `RunID` against a nil-returning implementation, asserts silence and asserts `RunID()` survives. Removing [the guard](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta.go#L131) turns it red with the nil-pointer panic.

You are right that the commit message over-credited `TestNilDeltaStillReachesTheInvocation` — it pins the forwarding half only. The message now describes the two halves separately.

agent/common_context_delta.go
wolo-lab: Took the stronger version. The key is the type plus the set of field *names*, as a per-type atomic bitmask, so a Branch-only derivation and a later Agent loss are different keys and both get reported. Values stay out of the key, so a thousand-item fan-out deriving the same shape is still one line. [`TestDiscardReportsEachDistinctLoss`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta_test.go#L188) pins it.

That also makes the claim you flagged true, so it no longer needs softening — later occurrences of the *same* loss genuinely tell the operator nothing new, and a different loss is no longer suppressed.

agent/common_context_delta.go
wolo-lab: Gone. The parameter no longer exists — your wrapped-versus-unwrapped table is one of two reasons, and the other is that the report is emitted from [`withICDelta`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta.go#L131) *before* `WithDelta` assigns `res.Context`, so the flag was a promise about something that had not happened yet.

`Context` is now named on both entry points, and the message says "these delta fields did not reach it" rather than calling the delta discarded. That is true whichever entry point ran, because neither installs it on the invocation.

agent/common_context_delta.go
wolo-lab: Added. [The paragraph](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta.go#L114-L119) now names the sharing, says it is the one cost a reader cannot recover by inspecting a value, and records that #1135 makes the propagation deliberate and synchronises the flag while this branch has the sharing without either.

agent/common_context_delta_test.go
wolo-lab: Fixed, and it was a real gap rather than a stylistic one. [`TestDiscardReportIsEmittedOnce`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta_test.go#L163) now derives from a fresh enclosing invocation each time, so it distinguishes the per-type key from a per-instance one. I confirmed that: switching the key to `any(ic)` turns it red with three lines instead of one.

Your production reasoning is the reason it is worth pinning — a per-instance key would retain every invocation object a non-conforming type ever produced, which is exactly what the interned `reflect.Type` avoids.

agent/common_context_delta.go
wolo-lab: Moot now — that paragraph went with the parameter. I checked the rendering rather than assuming: `go doc -u` shows the remaining paragraphs separated.

agent/common_context_delta.go
wolo-lab: Taken, with one change to your fix. A plain `Load` early return keyed on the type alone conflicts with keying on the field set, so instead the field mask is built first — it costs no allocation — and the dedup state is a per-type `*atomic.Uint32` bitmask rather than a composite map key. A struct key would have to be boxed into an interface on every lookup, which allocates on exactly the repeat path this is meant to make cheap. I measured that: the composite key still cost an allocation per repeat, the bitmask costs none.

[`TestDiscardRepeatDoesNotAllocate`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta_test.go#L217) asserts zero allocations on a repeat via `testing.AllocsPerRun`, and it fails if the formatting moves back above the check.

wolo-lab:
@karolpiotrowicz round-2 inline threads are all addressed in `5d29bc65`. I resolved them, which on top of them already being outdated means the replies are now behind two folds — repeating the substance here so it is readable, and linking the new code directly.

**The nil guard needs a test** — [`TestDiscardWithNoInvocationDelta`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta_test.go#L391). Removing [the guard](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta.go#L131) turns it red with the nil-pointer panic. You were also right that the commit message over-credited `TestNilDeltaStillReachesTheInvocation`, which pins the forwarding half only.

**Dedup keyed on the type alone** — took the stronger version you offered. The key is now the type plus the set of field *names*, held as a per-type atomic bitmask ([doc](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta.go#L137-L153), [code](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta.go#L164-L193)). A Branch-only derivation from the scheduler and a later Agent loss are different keys and both report. Values stay out of the key, so a thousand-item fan-out deriving the same shape is still one line. [`TestDiscardReportsEachDistinctLoss`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta_test.go#L188) pins it, and the claim you flagged is now true rather than softened.

**The field list built before the dedup check** — taken, with one change to your fix. A plain `Load` keyed on the type alone conflicts with keying on the field set, so the mask is built first instead and the state is a per-type `*atomic.Uint32`. I first tried a composite struct key and it still allocated on every repeat, because a struct key has to be boxed into an interface on each lookup — my own allocation assertion caught that. The bitmask costs nothing. [`TestDiscardRepeatDoesNotAllocate`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta_test.go#L217) asserts zero allocations on a repeat and fails if the formatting moves back above the check.

**The dedup test could not tell per-type from per-instance** — real gap, fixed. [`TestDiscardReportIsEmittedOnce`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta_test.go#L163) now derives from a fresh enclosing invocation each time. Switching the key to `any(ic)` turns it red with three lines instead of one.

**The aliasing sentence** is in [the cost paragraph](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta.go#L114-L119).

**`callerAppliesContext`** and the **missing `//` separator** are both gone — the parameter no longer exists. Beyond your wrapped-versus-unwrapped table, the report is emitted before `WithDelta` assigns `res.Context`, so the flag was a promise about something that had not happened yet. `Context` is now named on both entry points and the message says "these delta fields did not reach it".

Eleven mutants die against the current tests, including all four of yours: removing the nil guard, keying per instance, formatting before the check, and dropping the field set from the key.

agent/common_context_delta.go
```diff
@@ -78,6 +84,140 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+// the repeat path this exists to make cheap.
+var reportedNilICDelta sync.Map // reflect.Type -> *atomic.Uint32
+
+const (
+	lostAgent uint8 = 1 << iota
+	lostBranch
+	lostIsolationScope
+	lostUserContent
+	lostContext
+)
```
wojtas000: optional: the bit set has exactly zero headroom, and running out of it is silent. `fields` is a 5-bit mask, so its widest value is 31 and `1 << 31` is the last bit a `uint32` has. A sixth field in `InvocationContextDelta` makes `fields` reach 63; `uint32(1) << 32` and above evaluate to 0 in Go, so `Or(0)` returns `prev` unchanged, `prev&0` is always 0, and the deduplication stops working entirely — the 1004-lines-per-1000-item fan-out comes back. Nothing fails to compile, and the existing dedup tests use masks 1 and 2, so they stay green.

One line makes it a compile error instead. I checked both directions: it builds today, and adding a sixth constant to the block gives `constant -32 of type uint8 overflows uint8`.

```suggestion
)

// 1<<fields must stay a valid shift for a uint32, so the field set can hold at
// most five flags. A sixth makes 1<<fields evaluate to 0, Or never sets a bit,
// and the deduplication silently stops working.
const _ = 31 - (lostAgent | lostBranch | lostIsolationScope | lostUserContent | lostContext)
```

agent/common_context_delta.go
```diff
@@ -78,6 +84,140 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+// working path gives each one a copy. That is the one cost a reader cannot
+// recover by inspecting a value: EndInvocation on any child now ends the parent
+// and its siblings, and the write races their Ended. #1135 makes that
+// propagation deliberate and synchronises the flag, and until it lands this
+// branch has the sharing without either.
+func withICDelta(ic InvocationContext, d *InvocationContextDelta) InvocationContext {
+	if ic == nil {
+		return nil
+	}
+	if next := ic.WithICDelta(d); next != nil {
```
wojtas000: nit: worth a sentence saying this catches a nil interface and not a nil pointer inside one. The population the guard exists for is out-of-repo decorators, and a decorator that builds its result into a typed variable returns `(*myIC)(nil)` rather than a literal nil, so `next != nil` is true and the crash still happens.

I ran it: a `WithICDelta` returning a nil `*typedNilIC` panics at `agent/common_context.go:336` in `commonContext.Branch`, on the first accessor, exactly as before the fix. Not asking for a `reflect` check — that would cost the happy path on every node and every parallel item — just for the limit to be written down where the rest of the reasoning is.

```suggestion
	// Only a nil interface is caught. An implementation that returns a nil
	// concrete pointer hands back a non-nil interface, which this cannot tell
	// from a working one without reflecting on every derivation.
	if next := ic.WithICDelta(d); next != nil {
```

agent/common_context_delta.go
```diff
@@ -78,6 +84,140 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+// quiet kind of wrong.
+//
+// It also hands every sibling derivation the same invocation object, where the
+// working path gives each one a copy. That is the one cost a reader cannot
+// recover by inspecting a value: EndInvocation on any child now ends the parent
+// and its siblings, and the write races their Ended. #1135 makes that
+// propagation deliberate and synchronises the flag, and until it lands this
+// branch has the sharing without either.
+func withICDelta(ic InvocationContext, d *InvocationContextDelta) InvocationContext {
+	if ic == nil {
```
wojtas000: nit: this is the one new branch no test reaches — I deleted it and `go test ./agent/` stays green, where every other guard in the function takes a test down with it.

It is also the one guard that does not fix anything: returning nil here stores nil in `res.invocationContext`, so the caller gets the same invocation-less `commonContext` and panics on the first accessor instead of at the derive site. It is reachable from outside — `agent.Promote(nil)` builds a `commonContext` with a nil invocation, and `NewContext(nil)` does the same. Either pin it or drop it; a branch that moves a panic a few frames later is the shape the rest of this function exists to undo.

agent/common_context_delta.go
```diff
@@ -78,6 +84,140 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+	}
+	if d.UserContent != nil {
+		fields |= lostUserContent
+	}
+	if d.Context != nil {
+		fields |= lostContext
+	}
+	if fields == 0 {
+		// The delta asked for nothing, so nothing was lost and no report is owed —
+		// and claiming a bit here would spend one a real loss needs.
```
karolpiotrowicz, line 185: This reason does not survive the switch to the one-hot key. An empty delta maps to bit 0 and every real loss has `fields >= 1`, so bit 0 is the one bit no real loss can ever want — claiming it here would cost nothing. The early return is still right, and the first clause is the true reason: no report is owed. Deleting the return goes red, so the behaviour itself is pinned.

agent/common_context_delta.go
```diff
@@ -78,6 +84,140 @@ func (c *commonContext) WithICDelta(d *InvocationContextDelta) InvocationContext
...
+//
+// The mask is a value rather than part of the key because a struct key would
+// have to be boxed into an interface on every lookup, which allocates on exactly
+// the repeat path this exists to make cheap.
+var reportedNilICDelta sync.Map // reflect.Type -> *atomic.Uint32
+
+const (
+	lostAgent uint8 = 1 << iota
+	lostBranch
+	lostIsolationScope
```
karolpiotrowicz, line 158: These five flags OR to exactly 31, and `1 << fields` at :193 uses the mask *value* as a bit index, so bits 1 through 31 of the `atomic.Uint32` are exactly consumed. Correct as written, with no spare bit. A sixth field lets `fields` reach 63, and Go evaluates a non-constant `uint32(1) << 63` as 0 — `Or(0)` is a no-op, the bit test never fires, and deduplication silently stops for every shape involving the new field. Neither the compiler nor `go vet` catches it, since `fields` is a `uint8` and stays in range until a ninth field. `InvocationContextDelta` is exported and adding to it is the most likely next edit here, so the bound is worth recording somewhere the next person will read.

agent/common_context_delta_test.go
```diff
@@ -0,0 +1,407 @@
...
+		{"isolation scope", &InvocationContextDelta{IsolationScope: &scope}, `IsolationScope="child-scope"`},
+		{"user content", &InvocationContextDelta{UserContent: &content}, "UserContent"},
+		{
+			"every reported field at once",
+			&InvocationContextDelta{Agent: &child, Branch: &branch, IsolationScope: &scope, UserContent: &content},
+			`Agent, Branch="child-branch", IsolationScope="child-scope", UserContent`,
+		},
+		{
+			// Named on both entry points: neither installs it on the invocation.
+			// WithDelta puts it on the context it returns, which is a different
```
karolpiotrowicz, line 135: This says neither entry point installs `Context` on the invocation, and the same claim appears at :351. The working path does install it: `(*invocationContext).WithICDelta` assigns `res.Context = *d.Context`, and `agent.go:177` relies on exactly that to carry the span context. The production comment at common_context_delta.go:216-219 is careful and correct about this — it is only the test comments that overreach. It matters because a reader following them could conclude the `lostContext` branch is redundant, and a Context-only delta would then be discarded in silence.

agent/common_context_delta_test.go
```diff
@@ -0,0 +1,407 @@
...
+	var sub DynamicSubScheduler = stubSubScheduler{}
+	branch := "child-branch"
+
+	var c Context
+	_ = captureLog(t, func() {
+		c = PromoteWithDelta(ic, &CommonContextDelta{
+			InvocationContextDelta: &InvocationContextDelta{Branch: &branch},
+			RunID:                  &runID,
+			Path:                   &path,
+			OutputForAncestors:     &ancestors,
```
karolpiotrowicz, line 331: `OutputForAncestors` is set here and never read back — the assertions below cover `RunID`, `SubScheduler` and `Path` only, and `ResumeInputs` is not exercised on the discard path at all. Dropping the `OutputForAncestors` branch in `WithDelta` survives this test (a pre-existing test elsewhere in the package does catch it, so this is dead setup rather than a hole).

agent/common_context_delta_test.go
```diff
@@ -0,0 +1,407 @@
...
+type nilDeltaInvocation struct {
+	InvocationContext
+}
+
+func (nilDeltaInvocation) WithICDelta(*InvocationContextDelta) InvocationContext { return nil }
+
+// freshReports makes the once-per-type report available again. The set has
+// process lifetime, so without this a test that asserts on the report passes
+// only on the first run of the binary and fails under -count=2.
+func freshReports(t *testing.T) {
```
karolpiotrowicz, line 37: This clears a package global, and `captureLog` swaps the default logger, so both depend on nothing in this file calling `t.Parallel()`. That holds today. If it ever changes, the two failures compound: overlapping `Clear` calls make the report counts nondeterministic, and the LIFO cleanup leaves the default logger pointed at a dead test's buffer for the rest of the binary, which silently swallows every later log assertion. Worth a line in the doc comment saying the file must stay serial.

karolpiotrowicz (approved):
Nothing blocks. I re-reviewed this from scratch rather than as a follow-up round, and the headline is that the finding which looked most serious turned out to be pre-existing.

### The sibling-sharing cost you documented is already in the tree

Your comment at [`common_context_delta.go:114-119`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta.go#L114-L119) says the fallback hands every sibling the same invocation object and that the write races their `Ended`. That reproduces: ending one sibling ends the other, and `-race` reports a write/read race on `endInvocation` between [`agent.go:453`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/agent.go#L453) and [`:457`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/agent.go#L457).

What I did not expect is that it needs no fallback at all. A `CommonContextDelta` carrying no `InvocationContextDelta` — the shape [`workflow.go:302-307`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/workflow/workflow.go#L302-L307) passes today — reaches [`agent.go:394-395`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/agent.go#L394-L395), which returns the receiver rather than a copy. Same sharing, same data race, and it happens on the merge base as well as on this head:

```
$ go test -race -run TestBaseS1 ./agent/          # at f7e16e02, the merge base
    BASE nil-ICDelta path: c1.Ended()=true c2.Ended()=true parent.Ended()=true
    BASE ALSO SHARES — the sharing is pre-existing
WARNING: DATA RACE
```

So this is not something the PR introduces, and I am not asking you to fix it here. Worth checking separately: [#1135](https://github.com/google/adk-go/pull/1135), the PR you defer it to, is conflicted and has not moved since 4 September. If that flag redesign is still the plan for this, it looks like it needs someone to pick it back up.

### The deduplication holds up under real concurrency, and I can now say that with evidence

No test in package `agent` drives `reportDiscardedDelta` from more than one goroutine, so the green `-race` gate said nothing about it in earlier rounds. I wrote a probe: 200 goroutines deriving one shape emit exactly one report, and five distinct shapes across 40 goroutines each emit exactly five, under `-race -count=5`. Swapping your `Or` for a non-atomic load/store pair makes both fail with 2-3 and 7 reports, so the probe discriminates and the shipped version passes it.

One correction to something I told you last round: I said the `-race` gate exercised no concurrency here. That was wrong for the command that actually ran. Instrumenting `withICDelta` shows 297 calls across 112 distinct goroutines under `-race` in `agent/workflowagents/parallelagent` alone. The pass-through branch is genuinely covered.

### Four mutations to the dedup key survive the suite

Each of these leaves `go test ./agent/` green:

| mutation | what it breaks | why nothing sees it |
|---|---|---|
| `Or(1 << fields)` → `Swap(…)` | monotonicity across alternating shapes | no test runs A, B, A |
| one-hot → raw mask `Or(uint32(fields))` | `{Branch}` then `{Agent, Branch}` suppresses the Agent loss | no test uses an overlapping pair |
| `reflect.TypeOf(ic)` → a constant | the per-implementation-type axis | one nil-returning fixture type in the tree |
| unconditional `append(lost, "Context")` | the field list naming a field the delta never set | every assertion is `Contains`, none checks absence |

Each axis occurs in production: [`scheduler.go:378-387`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/workflow/scheduler.go#L378-L387) derives `{Branch}` per node against [`agent.go:177`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/agent.go#L177)'s `{Agent, Context}` per activation, and [`dynamic_scheduler.go:231`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/workflow/dynamic_scheduler.go#L231) derives `{Branch, IsolationScope}`, which overlaps the first.

The property worth holding onto, more than the four cases: the key stays monotone across alternating shapes, one field set never masks another, it stays per-implementation-type, and the rendered list names exactly the fields the delta asked for. A suite that only runs disjoint shapes in ascending order from one fixture type cannot see the first three. This is round four and the shipped code is correct on all four axes, so I would take it as a follow-up rather than hold the merge.

### Where the scope loss ends up

Keeping the parent's invocation means keeping the parent's `IsolationScope`, and that is an exact-match filter at [`contents_processor.go:149`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/internal/llminternal/contents_processor.go#L149). Driving that filter directly with a child that asked for `fc-abc123` and got `""` instead: one content admitted with the requested scope, three with the parent's, the extra two being the coordinator's unscoped turns.

I am not raising that as a defect. It is the trade you describe at [`common_context_delta.go:108-112`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta.go#L108-L112), the loss is named by value in the report, and on the base the same input panics instead, so there is no prior behaviour to prefer. The one place I could not settle is [`remoteagent/v2/utils.go:169`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/remoteagent/v2/utils.go#L169) and [`:187`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/remoteagent/v2/utils.go#L187), which gate what is *transmitted* to a remote agent on the same value. My probe drove the prompt-history filter, not that path, so I can tell you the filter arithmetic and not what a real run sends. If you think that edge is worth pinning, it is the one I would pin.

### Smaller things

Five comment and test-assertion points are inline on the changed lines.

One that predates this PR, while you are in the file: [`common_context_delta.go:48-49`](https://github.com/google/adk-go/blob/5d29bc65cf5760f1f2200553073f817b33b6c387/agent/common_context_delta.go#L48-L49) says the original context is returned when there are no changes, and that only holds for `d == nil`. A non-nil empty `&CommonContextDelta{}` falls through to `res := *c` and returns a new object. Entirely up to you whether it is worth a line.

Also unrelated to the code: [#1522](https://github.com/google/adk-go/pull/1522) still conflicts with this on both files and has not moved since 17 September, so whichever lands first sets the rebase.

agent/common_context_delta_test.go
wolo-lab: Ack

agent/common_context_delta_test.go
wolo-lab: Ack.

agent/common_context_delta_test.go
wolo-lab: Ack.

agent/common_context_delta.go
wolo-lab: Ack. Accepted risk.

agent/common_context_delta.go
wolo-lab: Ack.

wolo-lab:
Correcting something I said in my last reply. I repeated your note that `apidiff.yml` filters to `main`/`v1` and concluded the check would not run on #1522 while it is stacked. The filter is real, the conclusion is wrong.

`apidiff` has run on #1522 on every force-push since 11 September, as `event=pull_request`, and passed — most recently `032074a9` in 48s. #1522's base has never been anything but `wolo/agent-delta-nil-guard`: there is no `base_ref_changed` event on the PR at all, so this is not a case of the base having briefly pointed at `main`.

I cannot explain why the filter does not bite and I am not going to guess at a mechanism. The practical part is that the check does run on a stacked PR here, so it is not a gap that needs covering.

PR #1522, opened 2026-09-07
feat(agent): recover the invocation identity from a plain context

wolo-lab:
## Problem

Code that holds only a `context.Context` cannot tell which end user an invocation belongs to. A tool's outbound request reaches an `http.RoundTripper` several layers below the tool call, past jsonrpc2 and net/http wrapping, with the typed ADK context long since erased. Anything acting on behalf of the acting user — minting a per-user credential, for instance — has no way to name them, and threading a typed context through `net/http` is not available.

## Summary

Add `agent.Identity` and `agent.IdentityFromContext`. ADK contexts answer a private key held in `internal/adkcontext`, a module-internal leaf package shared by `agent` and `internal/context` to break the import cycle.

The procedure branches on whether the invocation is one of ADK's own context types, marked by embedding `adkcontext.Marker`. One of ours is asked, because it hands the question down to the invocation underneath — that is what keeps a tool or callback context working, since theirs hold no session by design. Anything else is only read. An `InvocationContext` written outside the module embeds the context it was derived from and cannot override a key it cannot name, so asking it would answer with the enclosing invocation's user: a live user who made no such call.

The read recovers rather than nil-tests. `session.Session` is a public interface of six accessors, so a struct value implements it naturally, a typed-nil pointer survives an interface-nil check, and a session wrapping a nil session panics in the accessor. This runs inside `http.RoundTripper`, where net/http does not recover, so a session that cannot answer costs the identity and not the process.

A delta carrying no change to the invocation no longer reaches `WithICDelta` on an invocation from outside the module. Merely asking is what drops such a decorator, and entering a workflow does exactly that. Ours are still asked, because for them an empty delta is not a no-op.

`IdentityFromContext` establishes which invocation a context was derived from, and is trusted only as far as every wrapper in the chain is. It is not an authentication boundary, and the doc says so at length — including that the key value escapes to every `Value` implementation between the reader and the invocation, so a single read is enough for a wrapper to keep it.

The decision procedure is pinned by a table rather than by cases: rows are the shapes a session or an invocation can take, columns the ways a context is derived from one. Every row is nested under one enclosing invocation, so any cell reporting that user is serving someone who made no such call.

---

Fourth of a five-part split of #1173. Stacked on 1521, and #1173 is stacked on this one.

### DIFF
--- a/agent/common_context.go
+++ b/agent/common_context.go
@@ -23,11 +23,188 @@ import (
 	"google.golang.org/genai"
 
 	"google.golang.org/adk/v2/artifact"
+	"google.golang.org/adk/v2/internal/adkcontext"
 	"google.golang.org/adk/v2/memory"
 	"google.golang.org/adk/v2/platform"
 	"google.golang.org/adk/v2/session"
 	"google.golang.org/adk/v2/tool/toolconfirmation"
 )
+
+// Identity is an ADK invocation's identity: the acting user, app name, and
+// session a call belongs to. It is recovered from a plain context.Context via
+// [IdentityFromContext], which reads it off the live session each time rather
+// than caching it.
+//
+// Reading it live is not licence to mutate the session from elsewhere. The
+// lookup runs on whatever goroutine asks — inside an http.RoundTripper, that is
+// the caller's — while [session.Session] states no goroutine-safety requirement
+// on ID, AppName and UserID. A session that rewrites those three after the
+// invocation starts must be safe for concurrent reads, or the credential path
+// races it. ADK's own sessions fix all three at construction.
+type Identity struct {
+	// UserID is the acting end user, as the embedding server put it on the
+	// session. ADK does not authenticate it: anything acting on behalf of this
+	// user — minting a per-user credential, for instance — is trusting the server
+	// to have bound session.UserID to an authenticated principal. ADK's own REST
+	// server takes it from the request body.
+	UserID string
+	// AppName is the app the invocation belongs to.
+	AppName string
+	// SessionID identifies the conversation the invocation belongs to.
+	SessionID string
+}
+
+// identityOf reads an invocation identity from the session getSession returns.
+// It is the one place package agent turns a session into an [Identity], so every
+// context type here answers the identity key the same way.
+//
+// getSession is called inside the recover, not before it: Session() is itself a
+// method on caller-supplied code and can panic on its own.
+func identityOf(getSession func() session.Session) (Identity, bool) {
+	return adkcontext.Recovered(func() Identity {
+		// One Session value, then one call per field: re-reading Session() per
+		// field risks a torn identity, and some context wrappers log on every read.
+		s := getSession()
+		return Identity{UserID: s.UserID(), AppName: s.AppName(), SessionID: s.ID()}
+	})
+}
+
+// IdentityFromContext returns the ADK invocation [Identity] carried by ctx, if
+// present.
+//
+// ADK contexts embed context.Context and register their identity under a private
+// key, so code that only holds a context.Context — for example an
+// http.RoundTripper running deep beneath a tool call, past intermediaries that
+// wrap the context — can recover the acting identity without threading a typed
+// context through every layer.
+//
+// It returns (zero, false) both for a context that does not descend from an ADK
+// context and for an invocation with no readable session. The two are not
+// distinguishable here. ok does not imply a populated Identity either: an
+// invocation whose session carries no user yields an empty UserID, so a caller
+// that needs one must check.
+//
+// An invocation ADK itself built always reports its own user, however a context
+// is derived from it. An [InvocationContext] implemented OUTSIDE the module is
+// where that stops, and one rule covers every way it does.
+//
+// Such a type is a decorator over the invocation it embeds, and Go promotes
+// every method it does not override. Every method on [InvocationContext] and
+// [Context] that RETURNS one of them returns the receiver's own derived copy, so
+// a promoted one hands back the EMBEDDED invocation and the decorator is gone —
+// its session with it. That is not confined to the identity: Session, UserID and
+// AppName then report the enclosing invocation too.
+//
+// Promotion also reaches what a method HANDS BACK, which matters most where the
+// thing handed back carries a user. A decorator that overrides Session but not
+// [Context.Artifacts], [Context.Memory] or [Context.SearchMemory] gets handles
+// built for the enclosing invocation, and those carry AppName, UserID and
+// SessionID as the storage and search keys — SearchMemory by way of the Memory
+// handle it reaches through. So such a context can answer this function with its
+// own user while reading and writing the ENCLOSING user's artifacts and
+// memories.
+//
+// [Context.State], [ReadonlyContext.ReadonlyState] and [Context.Actions] are the
+// same shape by a different route: each resolves through the invocation the
+// context holds rather than through Session, so a promoted one reads and writes
+// the enclosing user's session state, and a promoted Actions accumulates into
+// the event actions that commit to that user's session. Override all six, or
+// accept that only the credential follows the decorator. That behaviour predates
+// the identity key and is unchanged by it — Session, UserID and AppName have
+// always reported the enclosing invocation on the same shape — but a decorator
+// author reading this rule needs to know the credential is not the only thing
+// scoped to a user.
+//
+// So an invocation written outside the module reports its own user only where it
+// answers for itself:
+//
+//   - Derive it — [Promote], [NewContext], [NewToolContext], [NewCallbackContext],
+//     [NewCallbackContextWithArtifactTracking], a readonly context — and the
+//     derived context reports its user, or none at all if it has no readable
+//     session.
+//
+//     Passing it *as the context itself* is different, and what happens depends
+//     on what its own Value does. Leave Value alone and its parent answers, so
+//     the enclosing invocation is reported: the key is unnameable outside the
+//     module, so the decorator cannot claim it. Implement Value and there are two
+//     cases. Answer with something that is not an [Identity] and nothing is
+//     reported. Answer with an [Identity] — built directly, or taken from the
+//     parent's answer and substituted — and that is what this function reports.
+//     Naming the key is required for neither.
+//
+//     Nor is the key itself out of reach, which is worth being exact about
+//     because the internal package's own doc once claimed otherwise. Its TYPE is
+//     unnameable outside the module, but this function hands the key VALUE to
+//     whatever Value implementation ctx provides, so a single call is enough for
+//     that implementation to keep it and later plant an [Identity] under it with
+//     context.WithValue — from a context descending from no invocation at all,
+//     on another goroutine, after the call it observed has ended.
+//
+//     So: this function establishes which invocation a context was DERIVED from,
+//     and it is exactly as trustworthy as every wrapper between it and that
+//     invocation. It is not an authentication boundary and cannot be made into
+//     one — any in-process code that can wrap a context can already lie about
+//     Session and UserID, and a caller needing a guarantee against in-process
+//     code needs a different mechanism.
+//
+//   - Call a context-producing method ON it and it is dropped, promoting or not,
+//     because the promoted method belongs to what it embeds. It must override
+//     every one of them, returning a derived copy of ITSELF. The rule is the
+//     return type, not the list: as of writing that is
+//     [InvocationContext.WithContext] and [InvocationContext.WithICDelta] on
+//     [InvocationContext], plus [Context.WithDelta], [Context.WithAgentContext],
+//     [Context.WithAgentTimeout] and [Context.WithAgentCancel] on [Context].
+//     There is no exception here: a direct call drops it whatever the delta
+//     says, including one carrying no [InvocationContextDelta] at all, because
+//     the promoted method never reaches the decorator to consult the delta.
+//
+//   - Promoting it first shelters it from three of those four. [Promote] holds
+//     the decorator in a commonContext, and WithAgentContext, WithAgentTimeout
+//     and WithAgentCancel re-parent only the context.Context above it, so the
+//     decorator still answers. WithDelta is the exception: it forwards to the
+//     held invocation's WithICDelta, so it drops the decorator through the
+//     promotion, exactly as [PromoteWithDelta] does. Both reach WithICDelta and
+//     neither reaches WithDelta, so overriding WithICDelta alone is enough for
+//     them — it is a direct call on the decorator that WithDelta must also cover.
+//     This is the one place a delta carrying no [InvocationContextDelta] is
+//     genuinely safe: held inside a commonContext, such a delta leaves the
+//     invocation untouched. Held is the operative word, and it is why the
+//     exception belongs here and not above.
+//
+// Two further things a decorator cannot fix by overriding the methods above:
+//
+//   - [Context.WithContext], and [Context.WithAgentContext] it delegates to,
+//     rebind the invocation when handed an [InvocationContext] rather than a
+//     plain context.Context. The result then reports the ARGUMENT's user, by
+//     design and whoever the receiver spoke for. It is the one derivation that
+//     legitimately changes who a context speaks for, so do not hand it another
+//     call's invocation. How MUCH it rebinds then depends on whether the
+//     receiver cached an artifacts handle. Over an invocation that has an
+//     artifact service, a tool or tracking callback context caches one and
+//     carries it across untouched, so the result answers for the argument while
+//     Artifacts() still addresses the receiver's user — the same split as the
+//     promotion case above, reached the other way round. Over an invocation with
+//     none, nothing is cached, Artifacts() falls through to the invocation, and
+//     the rebind takes that with it. Both are pinned, because the second only
+//     became possible when the constructors started returning a nil handle
+//     rather than a wrapper around one.
+//   - [Context.SubScheduler] returns a scheduler, not a context, so the
+//     return-type rule above does not reach it — but the scheduler captured the
+//     context it was built from and derives every child from that. A decorator
+//     that does not override it hands back the embedded context's scheduler, and
+//     workflow.RunNode asks for one on entry, so the whole child subtree then
+//     runs as the enclosing invocation. Overriding SubScheduler does not repair
+//     it, because [DynamicSubScheduler.RunNode] takes no context and the one the
+//     scheduler captured is unexported — so the override has nothing to rebind.
+//     Closing it means changing the workflow engine, not this package.
+//
+// None of this is avoidable by discipline alone — [agent.Run] applies a delta on
+// every run, and the workflow schedulers call WithAgentCancel and
+// WithAgentTimeout on the context a node was handed.
+func IdentityFromContext(ctx context.Context) (Identity, bool) {
+	id, ok := ctx.Value(adkcontext.IdentityKey).(Identity)
+	return id, ok
+}
 
 // In general CommonContext should not be wrapped with contexts not providing agent.Context.
 // It allows to copy&modify context instead of building chains.
@@ -159,6 +336,7 @@ func prepareEventActions(actions *session.EventActions) *session.EventActions {
 // Callbacks and Tools
 type commonContext struct {
 	context.Context
+	adkcontext.Marker
 	invocationContext InvocationContext
 	artifacts         Artifacts
 	actions           *session.EventActions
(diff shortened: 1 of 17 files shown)


agent/common_context.go
```diff
@@ -23,12 +23,189 @@ import (
...
+//     it, because [DynamicSubScheduler.RunNode] takes no context and the one the
+//     scheduler captured is unexported — so the override has nothing to rebind.
+//     Closing it means changing the workflow engine, not this package.
+//
+// None of this is avoidable by discipline alone — [agent.Run] applies a delta on
+// every run, and the workflow schedulers call WithAgentCancel and
+// WithAgentTimeout on the context a node was handed.
+func IdentityFromContext(ctx context.Context) (Identity, bool) {
+	id, ok := ctx.Value(adkcontext.IdentityKey).(Identity)
+	return id, ok
```
karolpiotrowicz, line 206: This is the one hop in the identity chain with no panic containment, and it is the hop an `http.RoundTripper` actually enters through. Every inner read is wrapped — `identityOf` at L64, `identityFrom` at L597, and `internal/context/invocation_context.go:139` — and the reason given in all three places is that `Value` runs on the caller's goroutine where `net/http` does not recover. Here `ctx.Value` is called bare, so a panic from any intermediary between the transport and the ADK context takes the process down rather than costing the identity.

Reproduced with a context whose `Value` panics: the panic propagates out of `IdentityFromContext`. It is also untested — `identity_matrix_test.go:1075` routes the panicking fixture through `Promote`, so `identityFrom`'s recover absorbs it and the direct-call shape is never exercised. Wrapping this in `adkcontext.Recovered` would also take care of `IdentityFromContext(nil)`, which panics today.

agent/common_context.go
```diff
@@ -23,12 +23,189 @@ import (
...
+//     unnameable outside the module, but this function hands the key VALUE to
+//     whatever Value implementation ctx provides, so a single call is enough for
+//     that implementation to keep it and later plant an [Identity] under it with
+//     context.WithValue — from a context descending from no invocation at all,
+//     on another goroutine, after the call it observed has ended.
+//
+//     So: this function establishes which invocation a context was DERIVED from,
+//     and it is exactly as trustworthy as every wrapper between it and that
+//     invocation. It is not an authentication boundary and cannot be made into
+//     one — any in-process code that can wrap a context can already lie about
```
karolpiotrowicz, line 146: This sentence is the one that governs safe use of the function, and it sits 57 lines into the comment. A caller arriving from godoc reads the first paragraph, gets `ok=true`, and has no signal that the answer can belong to an enclosing invocation. Hoisting it into the opening paragraph is free and would do more for a reader than anything below it.

karolpiotrowicz (commented):
Two things need fixing before merge, one of which git will not warn you about. Everything else is small, and the identity rule itself holds up better than I expected — details below.

**The restack will fail to compile, with no conflict marker.** This branch declares `stubSubScheduler` at [`agent/identity_matrix_test.go:1127`](https://github.com/google/adk-go/blob/82cef052be258eebdd5801d81186bdfdea228558/agent/identity_matrix_test.go#L1127), and #1521 declares a different one in `agent/common_context_delta_test.go`. Neither is in the common ancestor, and because they sit in *different* files git merges both silently. I rebased this branch's own commit ([82cef05](https://github.com/google/adk-go/commit/82cef052be258eebdd5801d81186bdfdea228558)) onto #1521's current head, resolved the two files that do conflict in the way that keeps both branches' tests, and got `vet: agent/identity_matrix_test.go:1127:6: stubSubScheduler redeclared in this block`. Renaming one of them clears it. The other three collisions — `TestDeltaOnInvocationThatReturnsNil`, `TestDeltaReachesTheInvocation` and `nilDeltaInvocation` — are same-file and will show up as ordinary conflicts, so you will see those anyway. Worth knowing that the restack also breaks `TestDeltaOnInvocationThatReturnsNil/the discard is reported`: it asserts on the literal string `discarding the delta`, and #1521 now logs `this delta is discarded` and reports only once per implementation type, so the three earlier subtests consume the one report before the capturing subtest runs.

**`IdentityFromContext` is the only hop in the chain with no panic containment.** That one is inline.

On the identity rule for out-of-module decorators, which is the part I spent most of the review on: I reproduced the case where a decorator is passed in directly and the enclosing invocation's identity comes back with `ok=true`, including when the decorator's own session is the nil interface a decorator author gets by default. Then I tried the obvious fix — at [`IdentityFromContext`](https://github.com/google/adk-go/blob/82cef052be258eebdd5801d81186bdfdea228558/agent/common_context.go#L204-L206), route an `agent.InvocationContext` that is not an `adkcontext.Source` through `Promote(ic).Value(key)` instead of `ctx.Value(key)` — and ran it. It is worse, in two measurable ways.

- **It closes only one of the two routes.** After a delta the held invocation is a `*commonContext` over an `*invocationContext`, both of which *are* `adkcontext.Source`, so the guard can never fire there. `agent.Run` applies a delta on every run, so that is the route that matters.
- **It splits two shapes that currently agree.** On this branch, `IdentityFromContext(dec)` and `IdentityFromContext(context.WithValue(dec, k, v))` both report the enclosing user. Under the change they report different users, so the answer starts depending on whether some layer the caller does not control wrapped the context. For a function whose whole purpose is seeing past intermediaries, that is a bad property to acquire.

It also costs the behavior [`TestAnUnmarkedContextCanAnswerTheKeyItself`](https://github.com/google/adk-go/blob/82cef052be258eebdd5801d81186bdfdea228558/agent/identity_matrix_test.go#L881-L895) exists to pin, and it moves trust from `ctx.Value` to `Session()`, which is equally caller-supplied. So I think ["not an authentication boundary and cannot be made into one"](https://github.com/google/adk-go/blob/82cef052be258eebdd5801d81186bdfdea228558/agent/common_context.go#L145-L146) is the right call, and the comment at [`identity_matrix_test.go:339-342`](https://github.com/google/adk-go/blob/82cef052be258eebdd5801d81186bdfdea228558/agent/identity_matrix_test.go#L339-L342) already names the harm precisely. No change requested. My one suggestion about it is inline.

Two things the tests do not reach, for the record rather than as asks. `internal/context` and `internal/adkcontext` have no test that starts a goroutine, so the green `-race` run covers `commonContext` and nothing else — and both of those packages hold identity-answering `Value` methods. And the reads-versus-writes question is currently settled by field layout: `SetLiveSessionResumptionHandle` and `EndInvocation` write `c.params` fields that happen to be disjoint from the `c.params.Session` the identity path reads, with nothing pinning that they stay disjoint.

Everything else I checked came out clean. The exported surface grows by exactly `Identity` and `IdentityFromContext` with no incompatible change, `staticcheck` reports the same four pre-existing deprecation notices as the base, and the one test failure in a full-suite run was `TestRunDoesNotLeakListenerWhenTelemetryInitFails` in `cmd/launcher/web`, which this PR does not touch and which failed once in four full-suite runs.

agent/common_context.go
```diff
@@ -23,12 +23,189 @@ import (
...
+// session a call belongs to. It is recovered from a plain context.Context via
+// [IdentityFromContext], which reads it off the live session each time rather
+// than caching it.
+//
+// Reading it live is not licence to mutate the session from elsewhere. The
+// lookup runs on whatever goroutine asks — inside an http.RoundTripper, that is
+// the caller's — while [session.Session] states no goroutine-safety requirement
+// on ID, AppName and UserID. A session that rewrites those three after the
+// invocation starts must be safe for concurrent reads, or the credential path
+// races it. ADK's own sessions fix all three at construction.
```
karolpiotrowicz, line 43: The requirement as stated is satisfiable by a session that still tears. `identityOf` binds one `Session` value and then makes three separate accessor calls on it, so an implementer who adds an `RWMutex` around `ID`, `AppName` and `UserID` has met "safe for concurrent reads" exactly and can still hand back an `Identity` naming two generations. `TestIdentityAgainstASessionThatChangesUnderIt` concedes this in its own doc, which is the right place to concede it but the wrong place for an implementer to find it. Stating the real requirement here — a consistent triple across three consecutive calls on one `Session` value, which is what fixing all three at construction gives you — would put it where someone writing a session will read it.

agent/context_mock.go
```diff
@@ -39,6 +39,12 @@ import (
 // Done, Err and Value): those read from the supplied Ctx rather than panicking,
 // so the mock carries a usable context payload. If Ctx is nil they panic like
 // everything else.
+//
+// Value reading from Ctx has a consequence worth knowing when Ctx is a real ADK
+// invocation rather than context.Background(): the mock then reports THAT
+// invocation's identity, not the one its own Session field describes, because it
+// cannot override a key it cannot name. [IdentityFromContext] states the rule and
+// what a decorator must do about it.
```
karolpiotrowicz, line 47: The paragraph just added covers the identity consequence for this mock. There is a second consequence from the same commit that is not covered: `common_context_delta.go:134-136` now skips `ic.WithICDelta(d)` entirely for an empty or nil delta when the invocation is not an `adkcontext.Source`, and `StrictContextMock` carries no marker. So the `panic("unimplemented")` at L151-153 — which the doc at L35-36 promises will fire on an unexpected call — silently does not, and a downstream fake that overrides `WithICDelta` to count or record calls stops being invoked for empty deltas. Nothing in the repository is affected: every in-tree `InvocationContextDelta` producer sets at least one field. It is the exported mock's contract that has quietly changed, so a sentence here alongside the identity one would close it.

internal/context/invocation_context.go
```diff
@@ -121,6 +123,33 @@ func (c *InvocationContext) WithContext(ctx context.Context) agent.InvocationCon
...
+// otherwise be minted for it.
+func (c *InvocationContext) Value(key any) any {
+	if c == nil {
+		return nil
+	}
+	if key == adkcontext.IdentityKey {
+		if id, ok := adkcontext.Recovered(func() agent.Identity {
+			s := c.params.Session
+			return agent.Identity{UserID: s.UserID(), AppName: s.AppName(), SessionID: s.ID()}
+		}); ok {
```
karolpiotrowicz, line 142: This is a second place that turns a session into an `agent.Identity`, which makes the claim on `identityOf` at `agent/common_context.go:57-59` — "the one place package agent turns a session into an Identity" — true only because of its "package agent" scoping. The two are byte-equivalent today and the import cycle is a real reason they cannot share code directly, but nothing pins that they agree. Add a field to `Identity` and this copy leaves it zero with no test failing.

agent/common_context.go
wolo-lab: Fixed in 5c6289f, and you are right that this is the hop that matters.

I had looked at this and talked myself out of it, because I was thinking about `IdentityFromContext(nil)` — a caller error, where a nil guard arguably hides the bug rather than fixing it. Your framing is the stronger one and it is a different case: an intermediary between the transport and the ADK context is not the caller making a mistake, it is exactly the layer this function exists to see past, and it runs where net/http does not recover. I reproduced it before changing anything.

```
--- FAIL: TestKarol_PanicFromIntermediaryEscapes
    REPRODUCED: panic escaped IdentityFromContext: intermediary Value is not available
--- PASS: TestKarol_SameShapeViaPromoteIsContained
```

So it is now `adkcontext.Recovered(func() any { return ctx.Value(adkcontext.IdentityKey) })`, the same shape `identityFrom` uses, which covers the nil context too as you say.

On it being untested: `TestIdentityFromContextContainsAPanickingIntermediary` in `agent/identity_test.go` drives the direct-call route with three shapes — a panicking intermediary, one nested behind a plain wrapper so the panic is not merely the outermost frame, and a nil context. Reverting the wrap turns all three red.

agent/common_context.go
wolo-lab: Done in 5c6289f. Hoisted into the second paragraph, right after the "what this is for" one, so it is on screen before anything else.

It says the result establishes which invocation a context was derived from and is trusted only as far as every wrapper in the chain, that it is not an authentication boundary and cannot be made into one, and — the part your comment is really about — that `ok` does not rule out the answer belonging to an enclosing invocation, naming the shape that does it. The long rule below still carries the detail. I left the original sentence where it was rather than moving it, since the paragraph it closes needs it.

agent/common_context.go
wolo-lab: Fixed in 5c6289f, and the diagnosis is exactly right — "safe for concurrent reads" is a requirement an implementer can meet in full and still tear, because the read makes three separate accessor calls on one `Session` value.

The doc now says that, and states the requirement as a consistent triple across three consecutive calls on one `Session` value, noting that fixing all three at construction is what gives you that and is what ADK's own sessions do. I kept `TestIdentityAgainstASessionThatChangesUnderIt`'s concession where it is — it is the right place to record what is deliberately not promised — but an implementer no longer has to find it there.

For what it is worth I had this as a finding too and put it in the wrong place: I was going to state it and not say what to do instead, which is the half that actually helps.

agent/context_mock.go
wolo-lab: Fixed in 5c6289f. You are right that the commit changed two things about this mock and documented one.

Added a paragraph alongside the identity one saying that `withICDelta` now skips `WithICDelta` when the delta asks for no change and the invocation is not one of ADK's own context types, that the panic therefore does not fire for an empty or nil delta, and that an embedder overriding the method to count or record calls stops being invoked for those. It also records your point that nothing in the repository is affected, since every in-tree `InvocationContextDelta` sets at least one field — so what changed is the exported contract, not any behaviour here.

internal/context/invocation_context.go
wolo-lab: Fixed in 5c6289f, and the mutation you name is the one I used to check the fix.

`TestIdentityMappingsAgree` in `internal/context` reads the same session through both mappings and compares the results. Add a field to `Identity`, populate it in package agent alone, and:

```
the two session-to-Identity mappings disagree:
  package agent:    {UserID:mapping-user AppName:mapping-app SessionID:mapping-sid TenantID:t}
  internal/context: {UserID:mapping-user AppName:mapping-app SessionID:mapping-sid TenantID:}
```

It also sweeps the fields with reflect and requires each to be set, because equality alone stays green when a new field is populated by neither mapping — which leaves it permanently zero for every caller. A non-string field added later fails that sweep and needs the fixture extended, deliberately.

Reaching agent's mapping from here needs an invocation agent does not own, so the fixture is a decorator carrying its own session rather than a constructor call — the read arm is the only one that runs `identityOf`.

I also scoped the claim on `identityOf` instead of leaving it to be read as broader than it is, and put a pointer to the copy on each side so neither is findable only from the other.

wolo-lab:
All five inline points are fixed in [5c6289f](https://github.com/google/adk-go/commit/5c6289fa61f92a996298ca4d52e6d36698623831), replies on each thread. The two restack problems you predicted were both real and are resolved — details below, since neither is visible from the diff.

## The restack

You called both of these correctly, and the silent one cost me a `vet` failure before I found it.

`stubSubScheduler` is renamed on this side, to [`ownedSubScheduler`](https://github.com/google/adk-go/blob/5c6289fa61f92a996298ca4d52e6d36698623831/agent/identity_matrix_test.go#L1127) — it carries an `owner`, where the one in `common_context_delta_test.go` only needs to be non-nil, so renaming the newcomer leaves your file untouched. Merging them would have coupled two unrelated fixtures across files for no gain.

`TestDeltaOnInvocationThatReturnsNil/the discard is reported` was exactly as you described. The resolution takes your version of that file wholesale, so the subtest calls [`freshReports(t)`](https://github.com/google/adk-go/blob/5c6289fa61f92a996298ca4d52e6d36698623831/agent/common_context_delta_test.go#L107) and asserts on ["did not reach it"](https://github.com/google/adk-go/blob/5c6289fa61f92a996298ca4d52e6d36698623831/agent/common_context_delta_test.go#L114). The other three same-file collisions came up as ordinary conflicts, as you said they would.

The one thing worth recording about the merged `withICDelta`: your reason for forwarding a nil delta — that ADK's own wrappers do real work for one and have to be unwrapped — is the same condition this branch encodes as `!ours`. So the two mechanisms compose rather than compete, and the empty-delta shortcut stays keyed on emptiness while your report keeps its dedup. Deleting the shortcut turns ten `TestIdentityDecisionMatrix` cells red, so it is still carrying its weight after the merge.

## The decorator rule

Your two objections to routing through `Promote` are the ones that decide it, and the second is the one I had not weighed properly. A guard that makes `IdentityFromContext(dec)` and `IdentityFromContext(context.WithValue(dec, k, v))` report different users is a bad trade for a function whose whole job is seeing past wrappers the caller does not control. I reached the same conclusion about the enclosing-invocation report itself being older than this change — the same fixture reports the enclosing user on `main`, so it is not something the identity key introduced. Agreed on no change.

## The two you raised for the record

I took the first as an ask, because it is a gap rather than a note: `internal/context` now has [`TestIdentityReadsRaceTheInvocationMutators`](https://github.com/google/adk-go/blob/5c6289fa61f92a996298ca4d52e6d36698623831/internal/context/identity_concurrency_test.go#L43). It drives both identity-answering `Value` methods in the package from 32 goroutines each, plus a third target behind a plain wrapper, while `EndInvocation` and `SetLiveSessionResumptionHandle` run throughout.

That also pins your second point, which is the better half of it. The disjointness was holding by field layout with nothing asserting it, so I checked the test actually detects the failure it is named for: making a writer touch `params.Session` produces `WARNING: DATA RACE` and a red test. Without the goroutines nothing in the repository would have noticed.

`internal/adkcontext` I left alone. `Recovered` takes a closure and holds no state, so there is nothing there for TSan to find that the caller's own test would not already surface.

## Verification

Full suite with `-race -shuffle=on`: 90 packages, no failures, no data races. Lint and tidy clean in both modules. `apidiff` against the base still reports only `Identity` and `IdentityFromContext` added, so wrapping the entry point in `Recovered` did not disturb the signature.

I saw `TestRunDoesNotLeakListenerWhenTelemetryInitFails` pass in my runs, so I have nothing to add to your one-in-four. Separately, `TestLoggingSpanIDPropagation` in `internal/llminternal` fails under `-count=2` on the base as well as here — it holds logger state across runs, and `-count=1` hides it.

agent/context_mock.go
```diff
@@ -39,6 +39,29 @@ import (
...
 // so the mock carries a usable context payload. If Ctx is nil they panic like
 // everything else.
+//
+// Which identity a fake embedding this mock reports depends on where it sits, and
+// it is worth knowing both. Passed as the CONTEXT — IdentityFromContext(fake) —
+// Value forwards every key to Ctx, so the answer is Ctx's invocation and an
+// embedder's Session() override does not change it. Passed as the INVOCATION, to
+// Promote or any of the New*Context constructors, the mock carries no identity
+// marker, so the procedure READS it rather than asking, and the override DOES take
+// effect: the fake reports its own session's user. Neither position reports an
```
karolpiotrowicz, line 49: The context position does report an enclosing user whenever `Ctx` is an ADK context, which is the case the lines just above describe. The round-1 wording had this right ("the mock then reports THAT invocation's identity, not the one its own Session field describes"). This sentence now denies the one outcome `IdentityFromContext`'s doc warns about, so it's the one worth restoring.

<details><summary>Failing test, <code>go test -run TestStrictMockContextPosition ./agent/</code></summary>

```go
package agent_test

import (
	"testing"
	"time"

)

type userSession struct{ user string }

func (s userSession) ID() string                { return "sid-" + s.user }
func (s userSession) AppName() string           { return "app" }
func (s userSession) UserID() string            { return s.user }
func (s userSession) State() session.State      { return nil }
func (s userSession) Events() session.Events    { return nil }
func (s userSession) LastUpdateTime() time.Time { return time.Time{} }

type fake struct {
	agent.StrictContextMock
	own session.Session
}

func (f *fake) Session() session.Session { return f.own }

func TestStrictMockContextPosition(t *testing.T) {
	enclosing := icontext.NewInvocationContext(t.Context(), icontext.InvocationContextParams{Session: userSession{"enclosing"}})
	f := &fake{StrictContextMock: agent.NewStrictContextMock(enclosing), own: userSession{"bob"}}

	if id, ok := agent.IdentityFromContext(agent.Promote(f)); !ok || id.UserID != "bob" {
		t.Errorf("invocation position: got %q, %v, want bob", id.UserID, ok)
	}
	if id, ok := agent.IdentityFromContext(f); ok && id.UserID == "enclosing" {
		t.Errorf("context position reported the enclosing user %q", id.UserID)
	}
}
```

```
--- FAIL: TestStrictMockContextPosition (0.00s)
    strictmock_position_test.go:36: context position reported the enclosing user "enclosing"
```

</details>

internal/context/from_context_test.go
```diff
@@ -0,0 +1,502 @@
...
+	want := agent.Identity{UserID: "user-42", AppName: "app-1", SessionID: resp.Session.ID()}
+	if !ok || id != want {
+		t.Errorf("IdentityFromContext(WithContext) = %+v, %v; want %+v, true", id, ok, want)
+	}
+}
+
+// TestValueWithNilEmbeddedContext pins the nil-parent guard:
+// NewCleanToolContextTestOnly builds a context with no embedded parent, so
+// without it every non-identity key is a nil-interface method call.
+// TestNilGuardsAreDirectlyExercised pins the nil guards this package's Value
```
karolpiotrowicz, line 158: The doc comment for `TestValueWithNilEmbeddedContext` (lines 155-157) now sits above this test's own doc, so this test's comment opens by describing a different test and `TestValueWithNilEmbeddedContext` has none. The same happened at lines 310-312, where `TestValueDelegatesUnknownKeys` lost its doc to `TestReadonlyContextDelegatesUnknownKeys`.

agent/common_context.go
```diff
@@ -23,12 +23,235 @@ import (
...
+// session. The rule below says exactly when, and why closing it is not possible
+// here.
+//
+// It returns (zero, false) and never panics. Several different things produce
+// that answer and none is distinguishable from the others here — a context that
+// does not descend from an ADK context, an invocation with no readable session, a
+// nil ctx, an intermediary in the chain whose Value panics, a context answering
+// the key with something that is not an [Identity]. The list is not closed, and
+// callers must not treat "false" as naming any one of them. A panic is contained
+// rather than propagated, because this runs where net/http does not recover. ok does not imply a populated Identity either: an
```
karolpiotrowicz, line 115: This line joins two sentences and runs to 127 characters, where the rest of the comment wraps at 80.

karolpiotrowicz (approved):
Nothing here blocks. Both things from round 1 that had to be fixed before merge are fixed, and I checked each by running it. Reverting the new `Recovered` wrap turns all three cases of `TestIdentityFromContextContainsAPanickingIntermediary` red, and this head builds and passes on current main with the `ownedSubScheduler` rename in place. The claims the new comments make hold up too. Rewriting `params` wholesale in `EndInvocation` trips `TestIdentityReadsRaceTheInvocationMutators` on 10 of 10 runs, and populating a new `Identity` field in package agent alone fails `TestIdentityMappingsAgree` with the diff you quoted. I re-ran the rest of the battery on this head and it's clean, apart from the same `cmd/launcher/web` listener test as last time, which passes 30 of 30 in isolation here and on main.

One new sentence is wrong and worth fixing in this PR. The `StrictContextMock` doc now says neither position reports an enclosing user, and the context position does. Details and a failing test are inline.

Two comments that were true in round 1 have been made false by the new tests. [common_context.go:636-637](https://github.com/google/adk-go/blob/9e9f5f25d888352ef04aabf4986ec39205cc90b0/agent/common_context.go#L636-L637) says removing the type assertion in `identityFrom` fails no test, and the doc on `TestIdentityFromAHostileSource` at [identity_matrix_test.go:1067-1071](https://github.com/google/adk-go/blob/9e9f5f25d888352ef04aabf4986ec39205cc90b0/agent/identity_matrix_test.go#L1067-L1071) says the same. The direct `Value` check added at [identity_matrix_test.go:1099-1102](https://github.com/google/adk-go/blob/9e9f5f25d888352ef04aabf4986ec39205cc90b0/agent/identity_matrix_test.go#L1099-L1102) now catches it: with the assertion removed, `TestIdentityFromAHostileSource/a_source_answering_with_the_wrong_type` fails. Its message then blames the recover, which would send whoever hits it looking in the wrong place.

`TestNilGuardsAreDirectlyExercised` says it pins this package's `Value` nil guards and names one it doesn't ([from_context_test.go:158-196](https://github.com/google/adk-go/blob/9e9f5f25d888352ef04aabf4986ec39205cc90b0/internal/context/from_context_test.go#L158-L196)). There's a second. The `c.Context == nil` guard in `InvocationContext.Value` at [invocation_context.go:153-155](https://github.com/google/adk-go/blob/9e9f5f25d888352ef04aabf4986ec39205cc90b0/internal/context/invocation_context.go#L153-L155) never runs. The table has a typed-nil `*InvocationContext`, which returns at the `c == nil` check first, and no zero-value one. Deleting the guard leaves `agent`, `internal/context`, `internal/llminternal` and `internal/adkcontext` all green.

A smaller one: the three new comments explaining the duplicated mapping give an import cycle as the reason ([common_context.go:74-75](https://github.com/google/adk-go/blob/9e9f5f25d888352ef04aabf4986ec39205cc90b0/agent/common_context.go#L74-L75), [invocation_context.go:135-136](https://github.com/google/adk-go/blob/9e9f5f25d888352ef04aabf4986ec39205cc90b0/internal/context/invocation_context.go#L135-L136), [identity_mappings_test.go:64-66](https://github.com/google/adk-go/blob/9e9f5f25d888352ef04aabf4986ec39205cc90b0/internal/context/identity_mappings_test.go#L64-L66)), but `internal/context` already imports `agent` ([invocation_context.go:22](https://github.com/google/adk-go/blob/9e9f5f25d888352ef04aabf4986ec39205cc90b0/internal/context/invocation_context.go#L22)). What stops the two sharing code is that `identityOf` is unexported.

agent/context_mock.go
wolo-lab: Ack.

internal/context/from_context_test.go
wolo-lab: Ack.

agent/common_context.go
wolo-lab: Ack.

Issue #1524, opened 2026-09-07
telemetry: ToolUsePromptTokenCount is never recorded, so input_tokens under-reports tool-use calls


peterkarabinovic:
### What happens

`UsageMetadata.ToolUsePromptTokenCount` is never read by `internal/telemetry`. Two consequences on every `generate_content` span for a call that used a server-side tool (Google Search grounding, Maps, code execution):

1. **`gen_ai.usage.input_tokens` under-reports.** It is set to `PromptTokenCount` alone:

   https://github.com/google/adk-go/blob/f322a15/internal/telemetry/telemetry.go#L136

   Gemini reports tool-use prompt tokens *separately* from `PromptTokenCount` rather than folding them in, and they are billed at the input rate. The semantic-conventions reference implementation for google-genai adds the two together for exactly this reason:

   ```python
   # Tool-use tokens are reported separately from prompt_token_count, so add
   # them to get the total input token count.
   tool_use_tokens = getattr(um, "tool_use_prompt_token_count", None) or 0
   if um.prompt_token_count:
       attrs["gen_ai.usage.input_tokens"] = um.prompt_token_count + tool_use_tokens
   ```

   — [reference/scenarios/google-genai/scenario.py](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/reference/scenarios/google-genai/scenario.py)

   Note the contrast with `CachedContentTokenCount`, which *is* already inside `PromptTokenCount` (the same reference notes "prompt_token_count already includes cached content"). So cache-read tokens are correctly reported as a subset today, while tool-use tokens are simply missing from the total.

2. **The breakdown is unrecoverable.** There is no attribute carrying `ToolUsePromptTokenCount`, so a consumer cannot separate grounding-driven input from ordinary prompt input. `cache_read.input_tokens` and `reasoning.output_tokens` both report their share of their respective totals; tool-use has no equivalent.

### Why it matters

Server-side tool use is a significant and highly variable share of spend on grounded calls. With the total short and no breakdown, per-call cost derived from ADK spans is both wrong and un-debuggable — you cannot tell a cheap call from an expensive grounded one, or reconcile a span against a billing line.

### Impact

Only calls that use a server-side tool are affected. Everything else reports `ToolUsePromptTokenCount == 0` and is unchanged.

### Proposed fix

- `gen_ai.usage.input_tokens` = `PromptTokenCount + ToolUsePromptTokenCount`, matching the reference implementation.
[...]

Issue #479, opened 2026-01-14
ADK Open Telemetry


dpasiukevich:
We should make sure that ADK Go is consistent with other ADK implementations (e.g. adk-python and others).

That we export the same traces/spans. The configuration options and overall UX should be similar (e.g. integration for sending OTel telemetry data to dedicated entpoints).

We should have a proper documentation and examples for this.

pawel-maciejczek:
Could you assign this issue to me?

traviscrist:
any updates on adding this?

pawel-maciejczek:
OTel telemetry has been added in Feb 2026. You can check the available options in [`telemetry/config.go`](https://github.com/google/adk-go/blob/main/telemetry/config.go) and example in [`examples/telemetry/main.go`](https://github.com/google/adk-go/blob/main/examples/telemetry/main.go)

PR #1525, opened 2026-09-07
fix(telemetry): include tool-use prompt tokens in input_tokens

peterkarabinovic:
Part of #1524

`UsageMetadata.ToolUsePromptTokenCount` was never read, so on any call that used a server-side tool (Google Search grounding, Maps, code execution) `gen_ai.usage.input_tokens` was short by that amount.

Gemini reports these tokens separately from `PromptTokenCount` rather than folding them in, and bills them at the input rate. This matches adk-python, which sums the two:

```python
prompt_tokens = usage_metadata.prompt_token_count
tool_tokens = usage_metadata.tool_use_prompt_token_count
...
return cls(
    input_tokens=(
        None
        if prompt_tokens is None and tool_tokens is None
        else (prompt_tokens or 0) + (tool_tokens or 0)
    ),
```

— [`_token_usage.py#L119-L128`](https://github.com/google/adk-python/blob/852b575e9d126f59f1256da573a5b6a4c87603b1/src/google/adk/telemetry/_token_usage.py#L119-L128)

This is the opposite of `CachedContentTokenCount`, which is already *inside* `PromptTokenCount` and so is correctly reported as a subset today.

## Scope

This PR is the total-correction half of #1524 only. Per [review](https://github.com/google/adk-go/pull/1525#pullrequestreview-5139636713), the breakdown attribute (`gen_ai.usage.tool_use_prompt_tokens`) has been dropped — adk-python keeps tool-use tokens off spans entirely and reports them as an opt-in **metric** (`adk.experimental.invoke_agent.tool.input_tokens`), and adk-go emits no metrics yet. That half stays open on #1524 and belongs with the metrics work, so the issue is deliberately left unclosed by this PR.

## Compatibility

Calls with no server-side tool report `ToolUsePromptTokenCount == 0`, so `input_tokens` is unchanged. Only grounded/tool-using calls see a different `input_tokens`, which is the bug being fixed. Consumers pricing `input_tokens` will see grounded calls get more expensive — correctly so.

## Tests

Added `TestGenerateContent/ToolUsePromptTokensCountAsInput` to the existing table.

Confirmed the test catches the bug — with the source change reverted:

```
--- FAIL: TestGenerateContent/ToolUsePromptTokensCountAsInput (0.00s)
    telemetry_test.go:355: attribute "gen_ai.usage.input_tokens": got "10", want "17"
FAIL	google.golang.org/adk/v2/internal/telemetry	0.818s
```

With the fix in place:

```
$ go test ./internal/telemetry/...
ok  	google.golang.org/adk/v2/internal/telemetry	0.902s
ok  	google.golang.org/adk/v2/internal/telemetry/functionaltest	0.776s
```

Whole-repo checks, rebased onto current `main`:

```
$ gofmt -l internal/                 # clean
$ go vet ./internal/telemetry/...    # clean
$ go build ./...                     # ok
```


internal/telemetry/telemetry.go
```diff
@@ -48,6 +48,10 @@ var (
 	gcpVertexAgentInvocationID      = attribute.Key("gcp.vertex.agent.invocation_id")
 	genAIUsageCacheReadInputTokens  = attribute.Key("gen_ai.usage.cache_read.input_tokens")
 	genAIUsageReasoningOutputTokens = attribute.Key("gen_ai.usage.reasoning.output_tokens")
+	// genAIUsageToolUsePromptTokens has no entry in the semantic-conventions
+	// registry yet; it reports the Gemini-specific share of input_tokens that came
+	// from server-side tool executions.
+	genAIUsageToolUsePromptTokens = attribute.Key("gen_ai.usage.tool_use_prompt_tokens")
```
wojtas000: **Blocking** — the comment is honest that this key is unregistered, but the two keys above it are not the precedent it looks like: `gen_ai.usage.cache_read.input_tokens` and `gen_ai.usage.reasoning.output_tokens` **are both in the [semconv v1.41.0 registry](https://github.com/open-telemetry/semantic-conventions/blob/v1.41.0/docs/registry/attributes/gen-ai.md)** — they are raw `attribute.Key`s here only because the pinned import is `semconv/v1.36.0`. `gen_ai.usage.tool_use_prompt_tokens` is absent from that registry, so this would be the only key adk-go invents inside the spec-owned `gen_ai.*` namespace.

adk-python has already hit this and answered it twice over:

1. **Namespace** — for the one unregistered key it puts on a span it uses an `experimental.` infix: `gen_ai.usage.experimental.system_instruction_tokens` ([`_token_usage.py`](https://github.com/google/adk-python/blob/main/src/google/adk/telemetry/_token_usage.py)).
2. **Name** — it calls this quantity `tool_input_tokens`, not `tool_use_prompt_tokens`, and publishes it as `adk.experimental.invoke_agent.tool.input_tokens` ([`_metrics.py`](https://github.com/google/adk-python/blob/main/src/google/adk/telemetry/_metrics.py)).

So I would not take the `gcp.vertex.*` option you offered — that diverges from Python as well as from the spec. Suggest matching Python on both axes:

```suggestion
	// gen_ai.usage.experimental.tool_input_tokens has no semantic-conventions
	// registry entry; the experimental infix and the name both match adk-python,
	// which calls this bucket tool_input_tokens. It reports the share of
	// input_tokens that came from server-side tool executions.
	genAIUsageToolInputTokens = attribute.Key("gen_ai.usage.experimental.tool_input_tokens")
```

The emit site below and the two test expectations need the same rename.

internal/telemetry/telemetry.go
```diff
@@ -133,13 +137,18 @@ func TraceGenerateContentResult(span trace.Span, params TraceGenerateContentResu
 	span.SetAttributes(responseContentAttributes(params.Response)...)
 	if params.Response.UsageMetadata != nil {
 		span.SetAttributes(
-			semconv.GenAIUsageInputTokens(int(params.Response.UsageMetadata.PromptTokenCount)),
+			// Tool-use prompt tokens are reported separately from PromptTokenCount and
+			// are billed as input, so they belong in gen_ai.usage.input_tokens. This
+			// matches the semantic-conventions reference implementation for google-genai:
+			// https://github.com/open-telemetry/semantic-conventions-genai/blob/main/reference/scenarios/google-genai/scenario.py
+			semconv.GenAIUsageInputTokens(int(params.Response.UsageMetadata.PromptTokenCount+params.Response.UsageMetadata.ToolUsePromptTokenCount)),
```
wojtas000: Verified, no change requested — recording it so the next reader does not have to re-derive it. `google.golang.org/genai` documents `TotalTokenCount` as the sum of `prompt_token_count`, `candidates_token_count`, `tool_use_prompt_token_count` and `thoughts_token_count`, and documents `PromptTokenCount` as *including* cached content. So `ToolUsePromptTokenCount` is genuinely disjoint from `PromptTokenCount` and this addition cannot double-count, while `CachedContentTokenCount` correctly stays a subset. adk-python computes the identical sum.

internal/telemetry/telemetry.go
```diff
@@ -133,13 +137,18 @@ func TraceGenerateContentResult(span trace.Span, params TraceGenerateContentResu
...
 			// According to OpenTelemetry Semantic Conventions:
 			// https://github.com/open-telemetry/semantic-conventions/blob/v1.41.0/docs/registry/attributes/gen-ai.md
 			// gen_ai.usage.reasoning.output_tokens (ThoughtsTokenCount) SHOULD be included in gen_ai.usage.output_tokens.
 			semconv.GenAIUsageOutputTokens(int(params.Response.UsageMetadata.CandidatesTokenCount+params.Response.UsageMetadata.ThoughtsTokenCount)),
 			genAIUsageCacheReadInputTokens.Int(int(params.Response.UsageMetadata.CachedContentTokenCount)),
 			genAIUsageReasoningOutputTokens.Int(int(params.Response.UsageMetadata.ThoughtsTokenCount)),
+			genAIUsageToolUsePromptTokens.Int(int(params.Response.UsageMetadata.ToolUsePromptTokenCount)),
```
wojtas000: optional: adk-python omits a bucket entirely when the provider reported nothing, rather than emitting `0` (`to_attributes` only sets a key when the value is not None). This line always emits, which matches what the two keys above it already do here, so it is consistent within adk-go — just noting the cross-SDK difference in case you want the span shapes to line up. Not worth changing on its own.

internal/telemetry/telemetry_test.go
```diff
@@ -271,10 +271,39 @@ func TestGenerateContent(t *testing.T) {
...
+					},
+					FinishReason: genai.FinishReasonStop,
+				},
+			},
+			wantName:   "generate_content test-model",
+			wantStatus: codes.Unset,
+			wantAttrs: map[attribute.Key]string{
+				semconv.GenAIUsageInputTokensKey:  "17",
+				semconv.GenAIUsageOutputTokensKey: "35",
+				genAIUsageToolUsePromptTokens:     "7",
```
wojtas000: nit: two cheap additions to this same case, if you are touching it for the rename anyway.

1. It sets `ThoughtsTokenCount: 15` but never asserts `genAIUsageReasoningOutputTokens`, so the case exercises the output split without pinning it.
2. No case covers `PromptTokenCount == 0, ToolUsePromptTokenCount > 0`. The semconv reference you cite guards with `if um.prompt_token_count:` while this code adds unconditionally — a deliberate divergence that is currently untested, so nothing stops a later refactor from quietly adopting the guard.

```suggestion
				genAIUsageToolUsePromptTokens:     "7",
				genAIUsageReasoningOutputTokens:   "15",
```

wojtas000 (commented):
Thanks for this, and for the write-up on #1524 — it made the problem easy to follow. The first half is right and I'd like to take it. Could you split the PR, though?

**Keep commit 1.** Folding tool-use tokens into `input_tokens` is exactly what adk-python does, so this brings us to parity:

```python
prompt_tokens = usage_metadata.prompt_token_count
tool_tokens = usage_metadata.tool_use_prompt_token_count
...
return cls(
    input_tokens=(
        None
        if prompt_tokens is None and tool_tokens is None
        else (prompt_tokens or 0) + (tool_tokens or 0)
    ),
```

- [`_token_usage.py#L119-L128`](https://github.com/google/adk-python/blob/852b575e9d126f59f1256da573a5b6a4c87603b1/src/google/adk/telemetry/_token_usage.py#L119-L128)

Happy to merge that on its own, with its test. One small thing: please change `Fixes #1524` to `Part of #1524` in the description, so merging this doesn't auto-close the issue — the breakdown half stays open.

**Drop commit 2.** I don't think the breakdown belongs on the span. adk-python emits six token attributes there — `input_tokens`, `output_tokens`, `cache_read`, `cache_creation`, `reasoning`, and an experimental system-instruction one — and tool-use tokens are not among them. They appear only folded into the input total:

- [`_token_usage.py#L200-L219`](https://github.com/google/adk-python/blob/852b575e9d126f59f1256da573a5b6a4c87603b1/src/google/adk/telemetry/_token_usage.py#L200-L219)

Where Python reports the tool breakdown is **metrics** — `adk.experimental.invoke_agent.tool.input_tokens`, and its per-workflow twin:

- [`_metrics.py#L246-L250`](https://github.com/google/adk-python/blob/852b575e9d126f59f1256da573a5b6a4c87603b1/src/google/adk/telemetry/_metrics.py#L246-L250) (declaration)
- [`_metrics.py#L387-L389`](https://github.com/google/adk-python/blob/852b575e9d126f59f1256da573a5b6a4c87603b1/src/google/adk/telemetry/_metrics.py#L387-L389) (record)

The routing is deliberate, and the class docstring says so outright — `cache_read` and `tool_input_tokens` are subsets of `input_tokens`, and only some buckets reach spans:

- [`_token_usage.py#L92-L95`](https://github.com/google/adk-python/blob/852b575e9d126f59f1256da573a5b6a4c87603b1/src/google/adk/telemetry/_token_usage.py#L92-L95)

Even as a metric it is behind an opt-in gate:

- [`_instrumentation.py#L668-L673`](https://github.com/google/adk-python/blob/852b575e9d126f59f1256da573a5b6a4c87603b1/src/google/adk/telemetry/_instrumentation.py#L668-L673)

So commit 2 would put on every span, by default, a number adk-python keeps off spans entirely and hides behind a flag on metrics.

adk-go emits no metrics at all today. `telemetry/setup_otel.go:89` still says `// TODO(#479) init meter provider`, and #479 is closed. So this is a gap on our side, and I'd rather fill it properly than add an unregistered `gen_ai.usage.*` key to spans in the meantime.

We'll most likely open a separate issue for metrics in adk-go, and the tool-use prompt tokens will land there.

**So:** drop the second commit, keep commit 1 and its test, and this is good to go.

wojtas000 (changes requested):

indurireddy-TF:
Hi @peterkarabinovic Could you please address the review comments so we can proceed?

peterkarabinovic:
Done: dropped the second commit, kept commit 1 and its test, and changed Fixes #1524 to Part of #1524 so the breakdown half stays open. Also rebased onto current main. The PR is now a single commit, +32 −1.

Happy to pick up the tool-use prompt tokens as a metric once the adk-go metrics issue exists — feel free to tag me on it.

wojtas000:
LGTM, thanks. We should support metrics in the near future so then we will create this issue and notify you.

wojtas000 (approved):

PR #1528, opened 2026-09-07
fix(memory): rank and limit search results with non-ASCII substring matching

### LINKED ISSUE #1526: memory: in-memory SearchMemory diverges from adk-python (no scoring, no cap, no non-ASCII fallback)
## Summary

`memory.InMemoryService.SearchMemory` diverges from adk-python's
`InMemoryMemoryService.search_memory` in two ways. Both are in the matching path
and neither is covered by #569 (which is about tokenization only, and is being
fixed in #1403).

The first has real user impact today: **`preloadmemorytool` can put most of the
memory store into the prompt.**

adk-python is the source of truth for feature behavior in this repo, so these
read as unintentional gaps rather than deliberate Go-side choices — but happy to
be told otherwise.

Reference: [`src/google/adk/memory/in_memory_memory_service.py`](https://github.com/google/adk-python/blob/main/src/google/adk/memory/in_memory_memory_service.py)

---

## Gap 1 — no relevance scoring and no result cap

**Python** scores each event by how many distinct query words it matches, sorts
descending, and returns at most ten:

```python
_MAX_SEARCH_RESULTS = 10
...
        if matched_words:
          scored_memories.append((matched_words, MemoryEntry(...)))
...
    scored_memories.sort(key=lambda scored_memory: -scored_memory[0])
    return SearchMemoryResponse(
        memories=[memory for _, memory in scored_memories[:_MAX_SEARCH_RESULTS]]
    )
```

The Python source explains exactly why, in a comment above that sort:

> Almost any two sentences share a word, so returning every event that matches at
> least one query word returns most of the store, and callers such as the
> preload_memory tool put all of it in the prompt. Keep the events matching the
> most query words.

**Go** does precisely the thing that comment warns against — it appends every
event sharing at least one word, with no score, no sort and no cap:

https://github.com/google/adk-go/blob/f322a15f976f402991afd8e5d7e709aa7771ea07/memory/inmemory.go#L128-L141

`checkMapsIntersect` returns on the first shared word, so a single common token
is enough to include an event:

https://github.com/google/adk-go/blob/f322a15f976f402991afd8e5d7e709aa7771ea07/memory/inmemory.go#L148-L165

Nothing downstream caps it either — `preloadmemorytool` formats the whole slice:

https://github.com/google/adk-go/blob/f322a15f976f402991afd8e5d7e709aa7771ea07/tool/preloadmemorytool/tool.go#L86-L97

### Impact

A query containing a common word ("the", "it", "what") matches nearly every
stored event. With `preloadmemorytool` registered, all of them are rendered into
the prompt. Results also come back in map-iteration order, so the response is
non-deterministic between runs even for identical input.

### Suggested fix

Count matched query words per event instead of short-circuiting, sort descending
on that count, and cap at ten. A stable sort keyed only on the count preserves
insertion order among equals, matching Python.

---

## Gap 2 — no non-ASCII substring fallback

**Python** falls back to substring matching against the lowercased event text
when a query word is not ASCII:

```python
        matched_words = sum(
            1
            for query_word in words_in_query
            if query_word in words_in_event
            or (not query_word.isascii() and query_word in event_text_lower)
        )
```

That fallback is what makes the service work for languages without whitespace
word boundaries. `\w+` cannot tokenize a CJK sentence into words, so set
intersection alone finds nothing.

**Go** has only the set intersection, with no fallback, so a CJK query matches
nothing regardless of how tokenization is fixed.

### Suggested fix

Where a query token contains a non-ASCII rune, additionally test
`strings.Contains(strings.ToLower(eventText), token)`. This needs the event text
retained (or recomputed) alongside the extracted word set.

---

## Notes

- Independent of each other; happy for this to be split into two issues.
- Independent of #569 / #1403 — that PR fixes tokenization, which is a
  prerequisite for Gap 1 being *useful* but does not address either gap.
- Both are in `memory/inmemory.go`, described in its own docs as for prototyping
  and development, so neither is a production-path bug.

Found while triaging #1518 against adk-python for tokenization parity.

ktsoator:
Fixes #1526.

`InMemoryService.SearchMemory` currently returns every event sharing a word with the query, which can cause `preloadmemorytool` to put most of a user's memory into the prompt. This change ranks events by the number of distinct query words they match and returns at most 10 results.

Ties follow session insertion order, then event order within each session. Replacing a session keeps its original position. Non-ASCII query words also support substring matching, allowing shorter terms to match text without whitespace word boundaries. These behaviors follow adk-python. The tokenization fix is tracked separately in #1403.

### Testing Plan

Added regression tests for ranking, the result limit, tie ordering across sessions, session replacement, repeated query words, and non-ASCII substring matching. ASCII partial words remain excluded. Confirmed the regression suite fails against the original implementation and passes with the fix. The concurrent search test now uses matching events to exercise scoring and sorting.

Both workspace modules pass `go build -mod=readonly work` and `go test -race -mod=readonly -count=1 -shuffle=on work`. `golangci-lint run` and `go mod tidy -diff` also pass in each module.


wojtas000 (changes requested):
Could you drop the `.github/workflows/go.yml` change and send it in separate PR? It does not look as a proper thing for this PR specifically.

**The memory change itself looks right.** Checked against [`in_memory_memory_service.py`](https://github.com/google/adk-python/blob/main/src/google/adk/memory/in_memory_memory_service.py); parity is exact on the four points that matter:

- `strings.ToLower(strings.Join(texts, " "))` == `' '.join(...).lower()`
- `r > unicode.MaxASCII` == `not query_word.isascii()`
- `sessionOrder` reproduces dict insertion order, including a re-added session ID keeping its slot
- stable sort on score alone, cap at 10

Non-blocking nits:

- `map[string]bool` where the value means "allow substring" reads as a set with a surprising payload — a named type or a comment at the declaration would help.
- `strings.IndexFunc(word, func(r rune) bool { return r > unicode.MaxASCII }) >= 0` is a long way to say "not ASCII"; consider a small `isASCII` helper.
- `textLower` stores a second copy of every event's text where Python recomputes per search. Fine for a prototyping-only service, but worth a word in the `value` doc comment.

Tests are thorough — "do not join words across parts" and "ASCII partial word does not match" are exactly the cases I'd have asked for.

ktsoator:
Removed the workflow change and added comments explaining the boolean values and cached lowercase text. All checks are passing now.

I added the workflow change after two CI runs failed while installing golangci-lint. Since the cause is still unclear and subsequent runs have passed, I'll hold off on a separate CI PR for now.

wojtas000 (approved):

wojtas000:
Thanks for the fixes, LGTM

PR #1529, opened 2026-09-07
fix(memory): score, sort and cap in-memory SearchMemory results

### LINKED ISSUE #1526: memory: in-memory SearchMemory diverges from adk-python (no scoring, no cap, no non-ASCII fallback)

IslamElsayed:
Fixes #1526. Scoped to Gap 1 of that issue; Gap 2 (the non-ASCII substring
fallback) overlaps with the tokenization work in #1403 and is left for a
separate PR.

### Problem

`InMemoryService.SearchMemory` keeps every event sharing at least one word with
the query, in whatever order the session map happens to iterate.
`checkMapsIntersect` returns on the first shared word, so a single common token
is enough, nothing sorts the result and nothing caps it.

adk-python — the source of truth per CONTRIBUTING — scores each event by how
many distinct query words it matches, sorts descending and returns at most ten.
Its source says why:

> Almost any two sentences share a word, so returning every event that matches
> at least one query word returns most of the store, and callers such as the
> preload_memory tool put all of it in the prompt.

Go does the thing that comment warns against. A query containing "the" or "what"
matches nearly the whole store, and with `preloadmemorytool` registered all of
it is rendered into the prompt.

### Fix

Count the matched query words instead of stopping at the first, sort descending
on that count, and cap at ten (`maxSearchResults`, mirroring
`_MAX_SEARCH_RESULTS`). The sort is stable and reads only the score, so events
matching equally keep the order they were visited in.

Sessions live in a map, which Go iterates in a randomized order, so they are now
visited by sorted ID. Without that the stable sort has nothing stable to
preserve and identical queries return different orders between runs. adk-python
gets this property free from its insertion-ordered dict; a Go map has no
insertion order, so sorted IDs is the closest equivalent.

`checkMapsIntersect` becomes `countMapsIntersect`. It keeps the existing
"iterate the smaller map" optimisation, since the size of an intersection is the
same whichever side you walk.

### Testing plan

Three unit tests, each exercising one of the behaviours that changed. All three
fail on `main` and pass with the fix — the failures are the bug, stated exactly:

```
--- FAIL: Test_inMemoryService_SearchMemory_RanksByMatchCount
    order mismatch (-want +got):
      []string{
    + 	"one",
      	"three",
      	"two",
    - 	"one",
      }
--- FAIL: Test_inMemoryService_SearchMemory_CapsResults
    returned 15 memories, want 10
--- FAIL: Test_inMemoryService_SearchMemory_OrderIsStableAcrossRuns
    order changed on run 0 (-first +got):
      []string{
    - 	"0",
      	"1",
      ... // 3 identical elements
      	"7",
    + 	"0",
      }
```

With the fix applied:

```
$ go test -count=1 ./memory/...
ok  	google.golang.org/adk/v2/memory	0.353s

$ go test -race -count=1 ./memory/...
ok  	google.golang.org/adk/v2/memory	1.453s

$ go vet ./memory/...        # clean
$ gofmt -l memory/           # clean
```

Ordering is meaningful now, so the new tests compare the returned sequence
directly rather than through the existing order-insensitive `sortMemories`
transformer. The pre-existing table test still passes unchanged.

`./tool/...` also passes, covering `preloadmemorytool` and `loadmemorytool`, the
two consumers of `SearchMemory`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/memory/inmemory.go
+++ b/memory/inmemory.go
@@ -17,6 +17,7 @@ package memory
 import (
 	"context"
 	"maps"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -49,6 +50,10 @@ type value struct {
 	// precomputed set of words in the content for simple keyword matching.
 	words map[string]struct{}
 }
+
+// maxSearchResults caps how many memories SearchMemory returns, matching
+// adk-python's _MAX_SEARCH_RESULTS.
+const maxSearchResults = 10
 
 // inMemoryService is an in-memory implementation of Service.
 type inMemoryService struct {
@@ -128,40 +133,79 @@ func (s *inMemoryService) SearchMemory(ctx context.Context, req *SearchRequest) 
 		return res, nil
 	}
 
-	for _, events := range values {
-		for _, e := range events {
-			if checkMapsIntersect(e.words, queryWords) {
-				res.Memories = append(res.Memories, Entry{
+	// Almost any two sentences share a word, so keeping every event that
+	// matches at least one query word keeps most of the store, and callers
+	// such as preloadmemorytool render all of it into the prompt. Score each
+	// event by how many distinct query words it matches and keep the best few.
+	type scoredEntry struct {
+		score int
+		entry Entry
+	}
+
+	var scored []scoredEntry
+
+	// Sessions are held in a map, whose iteration order is randomized, so
+	// visit them in a fixed order to keep equally scored events in a stable
+	// sequence across runs. adk-python gets this from its insertion-ordered
+	// dict; a Go map has no insertion order to preserve, so sort the IDs.
+	for _, sid := range slices.Sorted(maps.Keys(values)) {
+		for _, e := range values[sid] {
+			score := countMapsIntersect(e.words, queryWords)
+			if score == 0 {
+				continue
+			}
+
+			scored = append(scored, scoredEntry{
+				score: score,
+				entry: Entry{
 					ID:             e.id,
 					Content:        e.content,
 					Author:         e.author,
 					Timestamp:      e.timestamp,
 					CustomMetadata: e.customMetadata,
-				})
-			}
+				},
+			})
 		}
+	}
+
+	// Sorting only on the score keeps events that match equally in the order
+	// they were visited.
+	slices.SortStableFunc(scored, func(a, b scoredEntry) int {
+		return b.score - a.score
+	})
+
+	if len(scored) > maxSearchResults {
+		scored = scored[:maxSearchResults]
+	}
+
+	for _, s := range scored {
+		res.Memories = append(res.Memories, s.entry)
 	}
 
 	return res, nil
 }
 
-func checkMapsIntersect(m1, m2 map[string]struct{}) bool {
+// countMapsIntersect returns how many keys the two sets share.
+func countMapsIntersect(m1, m2 map[string]struct{}) int {
 	if len(m1) == 0 || len(m2) == 0 {
-		return false
+		return 0
 	}
 
-	// Iterate over the smaller map.
+	// Iterate over the smaller map. The size of the intersection is the same
+	// either way round.
 	if len(m1) > len(m2) {
 		m1, m2 = m2, m1
 	}
 
+	count := 0
+
 	for k := range m1 {
 		if _, ok := m2[k]; ok {
-			return true
+			count++
 		}
 	}
 
-	return false
+	return count
 }
 
 func extractWords(text string) map[string]struct{} {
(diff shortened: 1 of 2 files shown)


wojtas000 (commented):
Thanks for this, however we're going to land #1528 instead, which covers the same ground plus Gap 2. Two substantive reasons, not just scope:

1. **Tie ordering diverges from Python.** Sorting session IDs is deterministic, but it's a Go-specific ordering adk-python has no notion of. Python's dict is insertion-ordered, and re-adding an existing session ID keeps its original position — sorted IDs can't express that. #1528 tracks insertion order explicitly.
2. **Gap 2 isn't separable in practice.** The substring fallback changes what "matched" means inside the same scoring loop, so doing it second means rewriting the loop this PR just landed.

Smaller things, for the record:

- `b.score - a.score` should be `cmp.Compare(b.score, a.score)` — safe here, but the subtraction idiom invites overflow bugs when it gets copied.
- `maxSearchResultsForTest = 10` duplicates an unexported constant; it'll drift silently.

PR #1532, opened 2026-09-07
docs(agents): add .agents/skills and a skill for authoring examples (#1331)

baptmont:
Backports #1331 from `main` to `v1`.

Done by hand: the automated replay applies about 80% of it and rejects the rest. The three `.agents/skills/adk-sample-creator/` files apply cleanly; the `AGENTS.md` and `CONTRIBUTING.md` hunks reject because those two documents differ substantially between the branches, so the patch context does not match.

Both doc insertions were applied at the equivalent place by hand and are **byte-identical to the lines added on `main`** — verified by diffing the added lines of each side. `main.go.tmpl` carries the module-path rewrite so the template generates `google.golang.org/adk` imports rather than `/v2`.

Carries the `(cherry picked from commit …)` trailer, so the backport tooling recognises #1331 as done and drops it from the queue.

### DIFF
--- a/.agents/skills/adk-sample-creator/SKILL.md
+++ b/.agents/skills/adk-sample-creator/SKILL.md
@@ -1,0 +1,139 @@
+---
+name: adk-sample-creator
+description: Author or rework a runnable example under `examples/` in the ADK Go repository — the directory, `main.go`, the README with its Mermaid diagram and real transcript, and the index row. Use when adding a sample for a feature (workflow graph, tool, registry client, model backend, server), when asked to "add an example" for something, or when bringing an existing example up to the repo's standard.
+---
+
+# ADK Go sample creator
+
+An example earns its place by teaching **one** thing faster than the package
+docs can. It must run. Two families set the standard, and which one you are
+writing decides several of the choices below:
+
+- **Engine samples** — `examples/workflow/*`. They exercise ADK itself, need
+  nothing beyond the binary and sometimes a model, and their subject is the
+  shape of a graph.
+- **Integration samples** — `examples/agentregistry/*`. They talk to a live
+  external service, so they carry cloud prerequisites, credentials, failures
+  worth unwrapping, and output that depends on the reader's own project.
+
+Copy the shape of the nearer family rather than inventing one.
+
+`examples/README.md` documents an older launcher API (`ParseAndRun`,
+`FormatSyntax`). Follow the templates here, not that file.
+
+## Layout
+
+```
+examples/<area>/<name>/
+├── main.go
+└── README.md
+```
+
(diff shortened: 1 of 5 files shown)


wojtas000 (changes requested):
The problem is that byte-fidelity is the wrong
goal for a file whose whole job is to describe the tree it sits in.

`SKILL.md` describes `main`'s tree, and four of its instructions don't hold on
`v1`:

- **L12** — "Engine samples — `examples/workflow/*`". `v1` has no
  `examples/workflow/`.
- **L15** — "Integration samples — `examples/agentregistry/*`". `v1` has no
  `examples/agentregistry/`, and no `agentregistry` package at all.
- **L34** — points at `examples/workflow/README.md` and
  `examples/agentregistry/README.md` as the models to copy. Neither exists.
- **L120** — `test -f go.work || go work init && go work use -r .`. `v1` is
  single-module (one `go.mod`, no `plugin/agentanalytics`), so there's no
  workspace to create and the `work` pattern in the commands below it doesn't
  mean what it means on `main`.

Net effect: an agent that loads this skill on `v1` is told to model a new example
on two directories it can't find, and to set up a workspace the branch doesn't
have. That's worse than not shipping the skill here, because it fails in a way
that looks like the agent's mistake rather than the doc's.

Two ways out, either is fine by me:

1. Adapt the skill to `v1` — swap the two sample categories for ones that exist
   (`examples/workflowagents/`, `examples/mcp/`, `examples/rest/` look like the
   closest equivalents), and drop the workspace paragraph in favour of plain
   `./...`. The `(cherry picked from …)` trailer still does its job for the
   backport queue.
2. Skip the `.agents/skills/` part on `v1` and backport only the `AGENTS.md` /
   `CONTRIBUTING.md` pointers — or drop the backport entirely, since a
   maintenance branch arguably doesn't need an example-authoring skill.

baptmont:
Confirmed all four against `v1`: it has `examples/workflowagents/` but no `examples/workflow/`, no `examples/agentregistry/` and no `agentregistry` package, and a single `go.mod`, so the workspace paragraph describes a setup the branch does not have.

Taking your second option rather than adapting. `v1` is maintenance-only, authoring new examples is not a maintenance activity, and an adapted skill would be a `v1`-specific fork of a doc that re-drifts whenever `main`'s examples move, with nothing keeping the two in sync. Backporting the `AGENTS.md` / `CONTRIBUTING.md` pointers on their own is not a middle ground either — they would point at a `.agents/skills/` directory the branch does not have.

Closing this and removing `v1-needed` from #1331 so the queue stops offering it. If `v1` ever wants agent guidance, it is better written against its own tree than ported from `main`.

Thanks for catching it.

PR #1533, opened 2026-09-07
fix(launcher/api): serve run_live through the API mount again

jjsasha63:

**Problem:**

`/run_live` returns 500 on the default `/api` prefix, so WebSocket live run is
broken for anyone using it:

```
websocket: bad handshake (500 Internal Server Error)
websocket: response does not implement http.Hijacker
```

The API mount wraps the `ResponseWriter` in `redirectRewriter`, which re-adds
the prefix to a redirect `Location` and promotes a 301 to a 308 so an unsafe
method is not downgraded to GET. `gorilla/websocket` v1.5.3 type-asserts the
writer to `http.Hijacker` at `server.go:175` and does not consult `Unwrap`, so
the wrapper makes the upgrade impossible. Only the prefixed branch wraps, which
is why an empty prefix still works.

**Solution:**

Forward `Hijack` to the underlying writer. The change is purely additive to
`api.go`: `git diff origin/main` shows zero removed lines, so nothing but
`run_live` changes.

The failure is reported as `http.ErrNotSupported` rather than a bare error,
because `http.ResponseController` finds this method before it follows `Unwrap`.
Without the wrap, this type would silently change the answer to
`errors.Is(err, http.ErrNotSupported)` for everything underneath it.

**An earlier version of this branch took a different approach and it was
wrong.** It deleted the wrapper and normalised the request path so the router
never emitted a redirect. Review found two regressions: it 301'd `GET /api` to
`/`, because an empty path took a fast path unrewritten, and it did not cover a
trailing `%2F`, where the inner router still emits a `StrictSlash` redirect that
only the wrapper restores the prefix to. The wrapper does necessary work, so the
fix completes it rather than removing it.

### Testing Plan

**Unit Tests:**

```
go build ./...                                             ok
go vet ./...                                               ok
go test -race -count=1 -shuffle=on ./cmd/... ./server/...  all packages ok
golangci-lint run ./cmd/...                                0 issues
```

Three tests:

| Test | What it covers |
| --- | --- |
| `TestWebSocketUpgradeThroughMount` | the mount can carry an upgrade, on `/api` and on an empty prefix as a control |
| `TestRunLiveUpgradesThroughTheRealMount` | the real REST server through the real registration path, so it is `/run_live` that is proved and not a stub |
| `TestHijackReportsNotSupported` | the error branch, which is reachable because `httptest.ResponseRecorder` is not a `Hijacker` |

Mutation-proved. Making the hijack fail while keeping the method present fails
both WebSocket tests with the exact reported 500, and the empty-prefix subtest
stays green as a control. Dropping the sentinel wrap fails the error-branch
test.

Both WebSocket tests set a read deadline, so a regression where the upgrade
succeeds but no frame arrives fails the test rather than hanging the package
until the ten minute default.

Run any agent with `adk web api webui`, open the UI, and start a live run. On
`main` the connection fails immediately; with this change it upgrades.

### Checklist

### Additional context

This is a regression from #1459, found while reviewing that PR's redirect
wrapper.

**Worth a follow-up, not this PR.** `/run_live` has never been reachable on the
default prefix, so this is the first time its own defences matter.
`websocket.Upgrader{}` is constructed with a nil `CheckOrigin`, which compares
the `Origin` host to the `Host` header, so a DNS-rebound page satisfies it and
WebSocket is exempt from CORS. Go also binds `:8080` on every interface while
the flag help calls it "Localhost port". adk-python runs
`_is_dns_rebinding_request` before accepting and defaults to
`--host 127.0.0.1`. `SetReadLimit`, a post-hijack deadline and any cap on
concurrent live sessions are all absent too. None of that is introduced here,
but this change is what makes it reachable.

### DIFF
--- a/cmd/launcher/web/api/api.go
+++ b/cmd/launcher/web/api/api.go
@@ -228,7 +228,12 @@ func (w *redirectRewriter) Unwrap() http.ResponseWriter { return w.ResponseWrite
 func (w *redirectRewriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 	h, ok := w.ResponseWriter.(http.Hijacker)
 	if !ok {
-		return nil, nil, fmt.Errorf("redirectRewriter: underlying %T is not an http.Hijacker", w.ResponseWriter)
+		// Wrapping the sentinel keeps errors.Is(err, http.ErrNotSupported)
+		// true. http.ResponseController finds this method before it follows
+		// Unwrap, so without the wrap this type would silently change that
+		// answer from true to false for everything underneath it.
+		return nil, nil, fmt.Errorf("redirectRewriter: underlying %T is not an http.Hijacker: %w",
+			w.ResponseWriter, http.ErrNotSupported)
 	}
 	return h.Hijack()
 }
--- a/cmd/launcher/web/api/api.go
+++ b/cmd/launcher/web/api/api.go
@@ -16,8 +16,10 @@
 package api
 
 import (
+	"bufio"
 	"flag"
 	"fmt"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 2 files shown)


kdroste-google (approved):
LGTM, thank you!

PR #1534, opened 2026-09-07
fix(adkrest): gate the agent graph routes behind IncludeDebugAPI

jjsasha63:

**Problem:**

`build_graph` and `build_graph_image` were registered unconditionally in
`handler.go`, so they were live on every server built with `adkrest.NewServer`,
including a headless `adk web api` with no UI, and they are unauthenticated.

Both describe the agent rather than serve it. `build_graph` returns the agent
tree. `build_graph_image` returns Graphviz DOT source, and the generator draws
tools as nodes (`agentgraphgenerator.go:292`), so the response names every tool
the agent can call.

The trace routes were made opt-in in #1413 for the same reason, and adk-python
keeps both kinds in `DevServer`, a class its production server never
instantiates.

**Solution:**

Four commits, each addressing a round of review.

1. Move the graph routers inside the `IncludeDebugAPI` gate, and guard a nil
   agent in `loadAgent`.
2. Split the agent builder routes out of the graph router. The first commit
   gated the whole router, which carries `/builder`, `/builder/save` and
   `/builder/cancel` as well. None of those reads an agent or discloses
   anything, and `/builder` answers an empty 200 on purpose so the UI disables
   its builder toggle. Gating it turned that into the error status
   `AgentBuilderConfigHandler` exists to avoid. They now live in their own
   `AgentBuilderAPIRouter`, registered unconditionally.
3. Say in the startup banner which UI panels the flag turns on, when it is
   absent. See "Additional context" for why the alternative was rejected.
4. Cover the ungated builder writes and the nil-agent guard, and correct a
   comment.

Also guards a nil agent in the event-graph handler, `debug.go`. `agent.Loader`
is an interface a caller implements, so it can return a nil agent with a nil
error. In the agent-graph handlers that panics. In the event-graph handler it
does not: `buildGraph` type-asserts the instance to a named interface, a nil
interface fails that assertion, and the handler answers 200 with an empty
digraph. Both now answer 404, for different underlying reasons, and the
comments say which.

**Compatibility.** Gating costs nothing. `build_graph` has never existed in a
release: the routes landed on 2026-09-03 and v2.3.0 is 2026-08-31, so
`git tag --contains` finds none. Before that, the only reference to the path
anywhere in the tree was inside the vendored UI bundle, calling it and getting
404.

### Testing Plan

**Unit Tests:**

```
go build ./...                                          ok
go vet ./...                                            ok
go test -race -count=1 -shuffle=on ./cmd/... ./server/...  all packages ok
golangci-lint run ./cmd/... ./server/...                0 issues in changed packages
```

`TestNewServerDebugAPIGate` covers both directions. The two graph routes must
not route with the flag off and must route with it on. All three builder routes
must answer either way, each with the status it should give.

Every assertion is mutation-proved:

| Mutation | Result |
| --- | --- |
| Move the builder router back inside the gate | fails on all three builder routes |
| Disable the `loadAgent` nil guard | fails on both graph handlers |
| Disable the `debug.go` nil guard | fails with 200 and an empty digraph |
| Remove the startup hint | fails the banner test |

`adk web api` with no flag: `GET /api/dev/apps/<app>/build_graph_image` returns
mux's 404, and `GET /api/dev/apps/<app>/builder` still returns an empty 200.
With `-include_debug_api` both answer: `build_graph` returns the JSON agent
tree and `build_graph_image` returns Graphviz DOT source.

The startup banner names what is missing:

```
       api:  you can access API using http://localhost:8080/api
       api:      for instance: http://localhost:8080/api/list-apps
       api:      the web UI's Traces and agent structure panels need -include_debug_api
```

### Checklist

### Additional context

**The UI does not get the flag by default, and that is deliberate.**
`include_debug_api` defaults to false and `adkgo deploy cloudrun` defaults to
`--webui=true --debug_api=false`, so a default Cloud Run deployment serves a UI
whose Traces and agent structure panels 404.

Making `webui` imply the flag would fix that, and it is rejected. Every existing
Cloud Run deployment on those defaults would start serving tool-call arguments
and responses on the next upgrade, without asking. Nothing would fail, so nobody
would look. Telling the operator costs nothing and takes nothing away.

This is not a new problem either: the trace routes have degraded the UI the same
way since #1413, with no explanation at all. The banner line covers both.

**Parity.** adk-python keeps all three builder endpoints in
`DevServer._register_dev_endpoints`, which `ApiServer` never calls, so the Go
split diverges deliberately. Python's builder GET returns the agent's
`root_agent.yaml` and its POST writes uploaded files into the agent directory.
Go's are an empty 200 and two 501s, which disclose nothing, which is why they
can sit outside the gate. `handler.go` states that reason as a property of the
route, so whoever implements the builder for real is told to reconsider.


cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
```diff
@@ -97,6 +97,12 @@ var cloudrunCmd = &cobra.Command{
 }
 
 // init creates flags and adds subcommand to parent
+// debugAPIFlagUsage is a constant because cobra does not wrap flag help, so a
+// long literal here runs off the terminal and into the next flag's text.
+const debugAPIFlagUsage = "Enable the Debug API - requires '--api'. " +
```
kdroste-google, line 102: Let's make all params similar.
You can inline in like
```go
	cloudrunCmd.PersistentFlags().BoolVar(&flags.cloudRun.debugAPI, "debug_api", false, ""+
		"Enable the Debug API - requires '--api'. "+
		"Exposes tool-call arguments, responses and tool names. "+
		"The web UI's Traces and agent structure panels need it.")

```

cmd/launcher/web/api/api.go
```diff
@@ -267,6 +267,13 @@ func registerAPIRoutes(router *mux.Router, pathPrefix string, handler http.Handl
 func (a *apiLauncher) UserMessage(webURL string, printer func(v ...any)) {
 	printer(fmt.Sprintf("       api:  you can access API using %s%s", webURL, a.config.pathPrefix))
 	printer(fmt.Sprintf("       api:      for instance: %s%s/list-apps", webURL, a.config.pathPrefix))
+	if !a.config.includeDebugAPI {
+		// Say it here rather than leave it to be inferred from a 404 in the
+		// browser console. These routes are off by default because they expose
+		// tool-call arguments, responses and tool names, so an operator has to
+		// ask for them. Nothing else says which panels that costs.
+		printer("       api:      the web UI's Traces and agent structure panels need -include_debug_api")
```
kdroste-google, line 275: maybe something like "Debug excluded: if you want UI's Traces and ...   please use -include_debug_api"

cmd/launcher/web/api/api.go
```diff
@@ -335,7 +342,7 @@ func NewLauncher() weblauncher.Sublauncher {
 	fs.StringVar(&config.pathPrefix, "path_prefix", "/api", "ADK REST API path prefix. Default is '/api'.")
 	fs.DurationVar(&config.sseWriteTimeout, "sse-write-timeout", 120*time.Second, "SSE server write timeout (i.e. '10s', '2m' - see time.ParseDuration for details) - for writing the SSE response after reading the headers & body")
 	fs.IntVar(&config.traceCapacity, "trace_capacity", 10000, "Maximum number of traces to keep in memory.")
-	fs.BoolVar(&config.includeDebugAPI, "include_debug_api", false, "The debug api endpoint will be included in the API if and only if the flag is set to true.")
+	fs.BoolVar(&config.includeDebugAPI, "include_debug_api", false, "Serve the developer endpoints that describe an agent rather than run it: the debug trace routes, whose payloads carry tool-call arguments and responses, and the agent graph routes, whose output names every tool the agent can call. The web UI needs this for its Traces and agent structure panels. Off by default.")
```
kdroste-google, line 345: It would probably make sense to break the lines in code.

server/adkrest/controllers/debug.go
```diff
@@ -166,6 +166,19 @@ func (c *DebugAPIController) EventGraphHandler(rw http.ResponseWriter, req *http
 		http.Error(rw, err.Error(), http.StatusInternalServerError)
 		return
 	}
+	// A Loader is an interface a caller implements, so it can hand back a nil
```
kdroste-google, line 169: Please shorten this comment

server/adkrest/controllers/agentgraph.go
```diff
@@ -155,6 +155,13 @@ func (c *AgentGraphAPIController) loadAgent(rw http.ResponseWriter, req *http.Re
 		http.Error(rw, err.Error(), http.StatusNotFound)
 		return nil, false
 	}
+	// AgentLoader is an interface a caller implements, so it can hand back a
+	// nil agent with a nil error. Without this the handler dereferences it and
```
kdroste-google, line 159: Please shorten the comment

server/adkrest/internal/routers/agentgraph.go
```diff
@@ -21,7 +21,12 @@ import (
 )
 
 // AgentGraphAPIRouter defines the routes behind the web UI's agent-structure
-// panel and agent builder.
+// panel.
+//
+// These are separate from the builder routes because only these two disclose
```
kdroste-google, line 26: Please shorten the comment.

server/adkrest/handler.go
```diff
@@ -141,7 +154,14 @@ type ServerConfig struct {
 
 // DebugAPIConfig contains parameters for the debug API.
 type DebugAPIConfig struct {
-	// Controls if [routers.NewDebugAPIRouter] is included
+	// IncludeDebugAPI serves the developer routes that describe an agent
+	// rather than run it: [routers.NewDebugAPIRouter], whose traces carry
```
kdroste-google, line 158: Please shorten the comment.

server/adkrest/internal/routers/agentgraph.go
```diff
@@ -53,6 +51,28 @@ func (r *AgentGraphAPIRouter) Routes() Routes {
...
 			HandlerFunc: r.graphController.BuildGraphImageHandler,
 		},
+	}
+}
+
+const builderDetail = "ADK Go agents are defined in Go code, so there is no server-side agent configuration to edit."
+
+// AgentBuilderAPIRouter defines the web UI's agent builder routes.
+//
+// It is registered unconditionally, unlike the graph routes it used to share a
```
kdroste-google, line 61: Please explain the comment

server/adkrest/handler.go
```diff
@@ -90,12 +90,25 @@ func NewServer(cfg ServerConfig) (*Server, error) {
 		routers.NewAppsAPIRouter(controllers.NewAppsAPIController(cfg.AgentLoader)),
 		routers.NewArtifactsAPIRouter(controllers.NewArtifactsAPIController(cfg.ArtifactService)),
 		routers.NewVersionAPIRouter(controllers.NewVersionAPIController()),
-		routers.NewAgentGraphAPIRouter(controllers.NewAgentGraphAPIController(cfg.AgentLoader)),
+		&routers.AgentBuilderAPIRouter{},
 		&routers.TestsAPIRouter{},
 		&routers.EvalAPIRouter{},
 	}
+	// These routes are opt-in because they describe the agent rather than serve
```
kdroste-google, line 97: Please shorten the comment.

kdroste-google (changes requested):
Thank you for this PR. 
Minor stuff left, please handle the comments

cmd/adkgo/internal/deploy/cloudrun/cloudrun.go
jjsasha63: Applied as written. The named constant is gone, which also puts the init doc comment back on init, where it was not pointing before.

cmd/launcher/web/api/api.go
jjsasha63: Taken, close to your wording.

cmd/launcher/web/api/api.go
jjsasha63: Broken across lines the same way.

jjsasha63:
All nine addressed in f3b1a48c1.

I also rebased onto main, which had moved 33 commits. Three things to know: the debug controller's new WithAuthorizer wiring is preserved, both production warnings main added are kept verbatim, and main's TestWebSocketUpgradeThroughMount and this branch's banner test both append at the end of api_test.go, so I kept both.

server/adkrest/controllers/debug.go
jjsasha63: Cut from nine lines to three.

server/adkrest/controllers/agentgraph.go
jjsasha63: Cut to two lines.

server/adkrest/internal/routers/agentgraph.go
jjsasha63: Cut to three lines. I also fixed a claim while I was in there: it said the pair returns every tool name, but only build_graph_image does.

server/adkrest/handler.go
jjsasha63: Cut from ten lines to six. I kept the PROD warning that landed on main while this was open.

server/adkrest/internal/routers/agentgraph.go
jjsasha63: Rewritten. The old version leaned on these routes having previously shared a router with the graph endpoints, which tells a reader nothing once that is old news. It now says what they do: they read no agent, the GET answers an empty 200 so the UI disables its builder toggle, and the writes answer 501.

server/adkrest/handler.go
jjsasha63: Cut to two lines. The builder rationale moved onto AgentBuilderAPIRouter, and the registration line carries a short trailing marker so the omission is visible without opening that file.

kdroste-google (approved):
LGTM, thank you!

PR #1535, opened 2026-09-07
fix(launcher/web): let a sublauncher own /health, and warn on invented storage

jjsasha63:

Two fixes in the web launcher, both raised in review of #1460. They are
independent and small enough to review together.

**Problem 1, /health shadows a sublauncher's own.**

`registerHealthRoute` ran before the sublauncher loop. gorilla/mux serves the
first matching route, so the built-in `/health` shadowed one a sublauncher
registers itself. A deployment with a real readiness check found it unreachable
while the probe kept answering `{"status":"ok"}`, so an unhealthy instance stays
in the load balancer rotation.

**Problem 2, the launcher invents storage and says so quietly.**

Before #1460 the launcher defaulted one service, the session service, and the
contract was that the caller configures the rest. #1460 also defaulted artifact
and memory. `cmd/launcher/prod` runs through this same path, so a production
deployment that configured no storage now comes up looking healthy and loses its
artifacts on the next restart, where before it failed. The existing log line
runs before the routers are built and has scrolled past by the time anyone
looks.

**Solution:**

Register `/health` after the sublauncher loop, so it is a fallback rather than
an override. Router assembly moves into `buildRouter` so the ordering can be
asserted without binding a port.

Keep defaulting artifact and memory, so nothing breaks on upgrade, and put the
warning in the startup banner alongside the URLs:

```
       api:  you can access API using http://localhost:8080/api
       WARNING: no artifact, memory service configured, so an in-memory one is in use.
       WARNING:     everything it holds is lost when this process exits.
       WARNING:     a future release will refuse to start instead. Configure a
       WARNING:     service, or pass -allow_in_memory_services to keep this.
```

The last two lines appear only when the caller did not ask for a volatile store.
`-allow_in_memory_services` declares the intent and silences them, and the
`webui` sublauncher implies it, since the dev UI is what the fallback exists
for.

This is deliberately a notice now and a refusal later, so no running deployment
breaks on this upgrade. Refusing is the right end state, because losing data
quietly is worse than failing loudly, but it should not arrive unannounced.

### Testing Plan

**Unit Tests:**

```
go build ./...                              ok
go vet ./...                                ok
go test -race -count=1 -shuffle=on ./...    all packages ok
golangci-lint run ./cmd/...                 0 issues
```

Both changes are mutation-proved. Restoring the old registration order turns a
sublauncher's `503 draining` into the built-in `200 ok`. Dropping the
future-refusal notice fails the warning test.

No existing test needed changing, including `TestWebLauncher_ServesA2A`, which
builds a config with only a session service. That is the check that this does
not break callers.

`adk web api webui` runs with no configuration and shows no future-refusal
notice. `adk web api` alone runs and shows it. Supplying both services, or
passing `-allow_in_memory_services`, silences it.

### Checklist

### Additional context

Raised in review of #1460, where the comment was that the launcher creates
services nobody needs. Turning the fallback into a hard refusal is the intended
follow-up once this notice has shipped in a release.

### DIFF
--- a/cmd/launcher/web/web.go
+++ b/cmd/launcher/web/web.go
@@ -175,9 +175,15 @@ func (w *webLauncher) Parse(args []string) ([]string, error) {
 // Refusing is the eventual goal rather than the current behaviour because
 // cmd/launcher/prod runs through this same path. A production deployment that
 // configured no storage should learn about it at startup, not when a restart
-// has already dropped the data. adk-python falls back to
-// InMemoryArtifactService for an absent URI, so the fallback is not wrong in
-// itself; it belongs to a dev server rather than to every server.
+// has already dropped the data.
+//
+// adk-python does not default to memory here. An absent --artifact_service_uri
+// gets per-agent local disk under <agents_root>/<agent>/.adk/artifacts, and
+// in-memory is only the fallback when local storage is disabled or unwritable
+// (cli/utils/service_factory.py). Every one of those paths logs a warning
+// naming the flag. So Go defaulting straight to memory is already the weaker
+// behaviour, which is the argument for warning loudly now and refusing later,
+// not an argument that the fallback is fine.
 func applyServiceDefaults(config *launcher.Config, allowInMemory bool) (defaulted []string, unconsented bool) {
 	if config.SessionService == nil {
 		config.SessionService = session.InMemoryService()
@@ -213,20 +219,23 @@ func logInMemoryServiceWarning(defaulted []string, unconsented bool) {
 	if len(defaulted) == 0 {
 		return
 	}
+	// Naming launcher.Config, not just the flag: the flag only silences this,
+	// and there is no command-line way to point the launcher at real storage.
 	log.Printf("       WARNING: no %s service configured, so an in-memory one is in use.", strings.Join(defaulted, ", "))
 	log.Printf("       WARNING:     everything it holds is lost when this process exits.")
 	if unconsented {
-		log.Printf("       WARNING:     a future release will refuse to start instead. Configure a")
-		log.Printf("       WARNING:     service, or pass -allow_in_memory_services to keep this.")
+		log.Printf("       WARNING:     a future release will refuse to start instead. Set")
+		log.Printf("       WARNING:     ArtifactService and MemoryService on launcher.Config,")
+		log.Printf("       WARNING:     or pass -allow_in_memory_services to keep this.")
 	}
 }
 
-// buildRouter assembles the router Run serves: the base router, then every
-// active sublauncher's routes, then the fallback health route.
+// buildRouter assembles the handler Run serves: the base router, every active
+// sublauncher's routes, and a health fallback around the outside.
 //
-// Separated from Run so a test can assert the registration order without
-// binding a port, because the order is what decides which /health wins.
-func (w *webLauncher) buildRouter(config *launcher.Config) (*mux.Router, error) {
+// Separated from Run so a test can exercise exactly what is served without
+// binding a port.
+func (w *webLauncher) buildRouter(config *launcher.Config) (http.Handler, error) {
 	router := BuildBaseRouter()
 
 	// check if there are any active sublaunchers
@@ -247,12 +256,7 @@ func (w *webLauncher) buildRouter(config *launcher.Config) (*mux.Router, error) 
 		}
 	}
 
-	// Registered after the sublaunchers, not before, because gorilla/mux
-	// matches in registration order. Going first would shadow a /health that a
-	// sublauncher serves itself, leaving that sublauncher's own readiness
-	// signal unreachable while the probe still reported ok.
-	registerHealthRoute(router)
-	return router, nil
+	return withHealthFallback(router), nil
 }
 
 // webUIKeyword is the command-line keyword of the webui sublauncher. It is a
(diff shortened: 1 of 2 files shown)


cmd/launcher/web/web.go
```diff
@@ -153,60 +151,68 @@ func (w *webLauncher) Parse(args []string) ([]string, error) {
...
 //
-// Neither adkrest.NewServer nor runner.New defaults them; only
-// runner.NewInMemory does, and the web launcher does not use it. A nil session
-// or memory service reaches the request path and panics, which drops the
-// connection without sending any HTTP response. The artifact handlers answer
-// 503 instead, so defaulting that one replaces a clear diagnostic with a server
-// that works until it restarts and then has lost everything. It is logged for
-// that reason: cmd/launcher/prod runs through this same path, so a deployment
-// that forgot to configure a service still says so on startup.
+// Only the session service. Every request path needs one, and it is what the
```
kdroste-google: Please make this comment more up to the point

cmd/launcher/web/web.go
```diff
@@ -153,60 +151,68 @@ func (w *webLauncher) Parse(args []string) ([]string, error) {
 	return restArgs, nil
 }
 
-// applyServiceDefaults fills in in-memory services the caller left unset.
+// applyServiceDefaults fills in the one service the launcher can safely assume.
```
kdroste-google: as-is is InMemory SessionService, let's have it at the very top

cmd/launcher/web/web.go
```diff
@@ -323,14 +329,74 @@ func BuildBaseRouter() *mux.Router {
 	return router
 }
 
-// registerHealthRoute serves health at the root as well as under the API
+// healthPath is the probe path served at the root as well as under the API
```
kdroste-google, line 324: why do we need all this code? Having just /health is not enough?

cmd/launcher/web/web.go
```diff
@@ -323,14 +329,74 @@ func BuildBaseRouter() *mux.Router {
...
+// not reachable through the mux API, so a route scoped only by a header still
+// counts as owning the path.
+func declaresHealthRoute(router *mux.Router) bool {
+	declared := false
+	_ = router.Walk(func(route *mux.Route, _ *mux.Router, _ []*mux.Route) error {
+		tmpl, err := route.GetPathTemplate()
+		if err != nil || tmpl != healthPath {
+			return nil
+		}
+		if _, err := route.GetHostTemplate(); err == nil {
```
kdroste-google: Why do we have this here?

cmd/launcher/web/web.go
```diff
@@ -323,14 +329,74 @@ func BuildBaseRouter() *mux.Router {
...
+			return nil
+		}
+		if q, err := route.GetQueriesTemplates(); err == nil && len(q) > 0 {
```
kdroste-google: Why do we have this here?

kdroste-google (changes requested):
Please look at the comments

cmd/launcher/web/web.go
jjsasha63: Agreed, 17 lines of comment on a 5-line function is too much. Cut to this:

```go
// applyServiceDefaults sets an in-memory session service when the caller left
// one unset. Every request path needs a session service, and a nil one panics
// mid-request, which drops the connection with no HTTP response.
//
// Artifact and memory stay as the caller set them. Nil is safe: the artifact
// handlers answer 503 naming the service, and SearchMemory reports "memory
// service is not set". Filling them in instead let an unconfigured deployment
// come up healthy and lose everything on the next restart.
// examples/web/main.go shows the intended wiring.
```

cmd/launcher/web/web.go
jjsasha63: Done. The first line now says what the function does, and the reasoning about
artifact and memory moved below it.

cmd/launcher/web/web.go
jjsasha63: A single registration is wrong wherever I put it, which is what grew this.

Registered first, it shadows a /health that a sublauncher serves itself. That is
the bug this PR starts from: a deployment adds a real readiness check and the
built-in route keeps answering ok in front of it.

Registered last, it sits behind any catch-all a sublauncher mounted, so the probe
path stops working. The api sublauncher does this today. When path_prefix is
empty, registerAPIRoutes mounts the REST server with router.NewRoute(), which
matches every path.

So the fallback goes on an outer router with the inner one as its final route. A
catch-all cannot get in front of it, and it stands down when a sublauncher
declares the path itself.

Your two comments below are right though. Cutting those took 10 lines off this
and made the name accurate.

cmd/launcher/web/web.go
jjsasha63: It was for a route that declares /health behind a host matcher, so it serves the
path for one hostname and rejects every other. Counting that as ownership would
switch the fallback off for all requests, including the ones that route refuses,
and an ordinary probe would 404.

Removed. Nothing in the tree registers such a route, and checking your question
showed the handling was broken anyway. The fallback carries no host matcher of
its own, so it shadowed the scoped route even on its own hostname. I have a test
for that now.

The path template is the whole test now, and the contract is one line in the doc
comment: declare /health unconditionally, or you own it entirely.

cmd/launcher/web/web.go
jjsasha63: Same as the host check above, for query parameters instead. Removed with it.

jjsasha63:
I am closing this. No user is asking for it right now, so it is not worth the
review churn.

For anyone who finds this later, it fixed two things. The launcher registered
/health before the sublaunchers ran, so a sublauncher serving its own readiness
check never received the request. And the launcher filled in in-memory artifact
and memory services when the caller configured none. A deployment with no
storage then came up healthy and lost everything on the next restart.

Both are still the behavior on main. I will open either as its own PR if someone
hits them.

Thanks for the review, Karol.

PR #1536, opened 2026-09-07
fix(launcher): pin the webui-before-api order, and clarify the adk-web pin

jjsasha63:

Two loose ends from review of #1459 and #1460. Neither changes behaviour: one
adds a test, the other is documentation.

**Problem 1, the UI survives on argument order alone.**

`api.NewLauncher()` mounts a catch-all route when its path prefix is empty
(`registerAPIRoutes` uses `router.NewRoute()`), and that matches every path,
including `/ui/`. gorilla/mux serves the first route that matches, so the web UI
is reachable only because `NewLauncher` happens to pass `webui.NewLauncher()`
before `api.NewLauncher()`.

Nothing enforced that. Swapping the two arguments makes the entire UI return
404, and neither the build nor any test says so.

**Problem 2, the adk-web pin is confusing.**

The `Dockerfile` buried its most important caveat at the end of a fifteen-line
comment, which is where a reviewer asked what the whole thing was for.

`ADK_WEB_REF` is `33c3568`, upstream tag v1.0.5 dated 2026-08-26. The `distr/`
bundle committed next to it was built on 2026-06-30, before the pin existed, and
carries no provenance file. So building at the pin moves the UI forward by about
two months rather than reproducing what is in the tree.

**Solution:**

Add `cmd/launcher/full/full_test.go`, which starts the server with an empty API
prefix and asserts `GET /ui/` still returns the UI document. It asserts the
served response rather than the argument order, so it survives the mounting
changing shape.

Cut the Dockerfile comment from 29 lines to 13: lead with what the argument is,
state the gap plainly, and point at the README. Add a section to
`scripts/adk-web/README.md` explaining why the gap cannot be closed
retroactively, and what changes after the first refresh.

### Testing Plan

**Unit Tests:**

```
go build ./...                                  ok
go vet ./...                                    ok
go test -race -count=1 -shuffle=on ./cmd/...    all packages ok
golangci-lint run ./cmd/launcher/full/          0 issues
```

The ordering test is mutation-proved. Swapping the arguments to
`web.NewLauncher` gives:

```
full_test.go:106: GET /ui/ status = 404, want 200; the API catch-all is shadowing the UI
```

Worth knowing for review: swapping the command-line order instead, `api` before
`webui`, does not fail. Registration order comes from the arguments to
`web.NewLauncher`, not from the command line, so the flags are not what protects
this.

`adk web webui api -path_prefix=` serves the UI at `/ui/`. Swap the two
sublaunchers in `cmd/launcher/full/full.go` and it 404s.

### Checklist

### Additional context

Follows #1533, #1534 and #1535, which cover the rest of that review.

### DIFF
--- a/cmd/launcher/full/full_test.go
+++ b/cmd/launcher/full/full_test.go
@@ -1,0 +1,114 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package full
+
+import (
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"strings"
+	"testing"
+	"time"
+
+	"google.golang.org/adk/v2/agent"
+	"google.golang.org/adk/v2/cmd/launcher"
+	"google.golang.org/adk/v2/cmd/launcher/web"
+	"google.golang.org/adk/v2/cmd/launcher/web/api"
+	"google.golang.org/adk/v2/cmd/launcher/web/webui"
+)
+
+// freePort returns a port nothing is listening on. There is a race between
+// closing the listener and the server binding it, which is why the caller polls
+// rather than assuming the server is up.
+func freePort(t *testing.T) int {
+	t.Helper()
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatalf("net.Listen() failed: %v", err)
+	}
+	port := ln.Addr().(*net.TCPAddr).Port
+	if err := ln.Close(); err != nil {
+		t.Fatalf("closing the probe listener failed: %v", err)
+	}
+	return port
+}
+
+// TestWebUIOutranksTheAPICatchAll pins the order the sublaunchers are passed to
+// web.NewLauncher in NewLauncher.
+//
+// With an empty path prefix the API sublauncher mounts a catch-all route that
+// matches every path, including /ui/. gorilla/mux serves the first route that
+// matches, so the UI survives only because webui is registered before api.
+// Nothing else enforces that. Swap the two arguments and the entire UI
+// disappears behind the API, with no build or test failure to say so.
+//
+// The assertion is on the served response rather than on the argument order, so
+// it still holds if the mounting changes shape.
+func TestWebUIOutranksTheAPICatchAll(t *testing.T) {
+	port := freePort(t)
+
+	l := web.NewLauncher(webui.NewLauncher(), api.NewLauncher())
+	// An empty API prefix is what makes the two collide. The default /api does
+	// not overlap /ui/, which is why this has never bitten in practice.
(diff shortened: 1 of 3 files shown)


cmd/launcher/full/full_test.go
```diff
@@ -0,0 +1,116 @@
...
+		t.Fatalf("net.Listen() failed: %v", err)
+	}
+	port := ln.Addr().(*net.TCPAddr).Port
+	if err := ln.Close(); err != nil {
+		t.Fatalf("closing the probe listener failed: %v", err)
+	}
+	return port
+}
+
+// TestWebUIOutranksTheAPICatchAll pins the sublauncher order in NewLauncher.
```
kdroste-google, line 46: Please add clear concise information about the scenario - this test checks that the edge case when api is given an empty prefix then the other routes (webui in that case) are able to serve content. 
The same should be done for other routes (a2a.NewLauncher(), pubsub.NewLauncher(), eventarc.NewLauncher())

scripts/adk-web/Dockerfile
```diff
@@ -2,30 +2,20 @@
 FROM node:iron-trixie AS downloader
 RUN apt-get update && apt-get install -y git
 
-# The upstream ref to build. Pass --build-arg ADK_WEB_REF=<sha-or-tag> to move
-# it forward; see scripts/adk-web/README.md for how to bump the pin.
+# The upstream revision to build. A SHA rather than a tag, because upstream can
+# move a tag. 33c3568 is upstream tag v1.0.5, dated 2026-08-26. Pass
```
kdroste-google, line 6: Why do we need those specific versions in this comment?

scripts/adk-web/README.md
```diff
@@ -28,6 +28,16 @@ A pin does not stop upstream from making breaking changes. It makes a refresh
 reviewable. The diff is 8MB of minified JavaScript that no reviewer can read, so
 the upstream revision range is the only part anyone can actually check.
 
+### The pin does not describe the bundle committed today
```
kdroste-google, line 31: Why do we need that section?

kdroste-google (changes requested):
Please look at the comments

cmd/launcher/full/full_test.go
jjsasha63: Done both. The comment leads with the scenario now, and the test is table-driven
over all four sublaunchers registered before api.

The second half turned up more than I expected. pubsub and eventarc default to
`--path_prefix /api`, which is api's own default, so they collide with it in the
shipped configuration and not only behind an empty prefix. Registering api first
makes `POST /api/apps/{app}/trigger/pubsub` return the REST 404 instead of
reaching the trigger handler.

Their rows POST rather than GET, because mux falls through to the catch-all when
the path matches but the method does not, so a GET row would pass whatever the
order is. The eventarc row sends a `ce-type` header to stay out of a branch where
`POST {}` hits a nil pointer dereference in that handler. That looks like a real
pre-existing bug and I will file it separately.

scripts/adk-web/Dockerfile
jjsasha63: We do not. Removed.

The SHA, tag and date sat in three hand-maintained places: the ARG, that comment,
and README prose. `update-adk-web.sh` writes none of the prose, so a bump left two
of the three stale. Only the ARG holds the revision now.

The bump instructions also told the next person to write the tag back into that
comment, so step 1 now says to name it in the commit message.

scripts/adk-web/README.md
jjsasha63: We do not. Removed.

It described a state that ends at the first refresh, and the provenance section
further down already covers it: the committed bundle predates the provenance file,
its upstream commit is unknown, and the first refresh fills the gap.

kdroste-google (approved):
LGTM

PR #1544, opened 2026-09-08
fix(runner): flush buffered live events at EOF

### LINKED ISSUE #1507: runner: RunLive silently drops buffered tool calls when the stream ends mid-transcription
## 🔴 Required Information

**Describe the Bug:**

`Runner.RunLive` buffers tool calls that arrive while a transcription is in
flight (`runner/runner.go:982-985`). It flushes them only when a non-partial
transcription event arrives (`:987-1013`). If the stream ends before that event,
the buffer is dropped. The loop at `:931-1029` has no flush after it.

The consumer is never told. It sees a clean end of stream and no error. The tool
call is never delivered, and it is never written to the session, so it is
missing from the history on the next turn too.

**Steps to Reproduce:**

Save as `runner/bufdrop_test.go` and run
`go test ./runner/ -run TestRunLive_BufferedToolCallDroppedWhenStreamEnds -v`.

```go
func TestRunLive_BufferedToolCallDroppedWhenStreamEnds(t *testing.T) {
	ctx := context.Background()
	ss := session.InMemoryService()
	if _, err := ss.Create(ctx, &session.CreateRequest{
		AppName: "app", UserID: "u", SessionID: "s",
	}); err != nil {
		t.Fatal(err)
	}

	base := must(agent.New(agent.Config{Name: "live_agent"}))
	mockLive := &mockLiveAgent{
		Agent: base,
		runLiveFn: func(ic agent.InvocationContext) (agent.LiveSession, iter.Seq2[*session.Event, error], error) {
			return &dummyLiveSession{}, func(yield func(*session.Event, error) bool) {
				// 1. A partial transcription starts. isTranscribing = true.
				partial := session.NewEvent(ic, ic.InvocationID())
				partial.LLMResponse = model.LLMResponse{
					Partial:            true,
					InputTranscription: &genai.Transcription{Text: "book me a "},
				}
				if !yield(partial, nil) {
					return
				}
				// 2. A tool call arrives mid-transcription. It gets buffered.
				call := session.NewEvent(ic, ic.InvocationID())
				call.LLMResponse = model.LLMResponse{Content: &genai.Content{
					Parts: []*genai.Part{{FunctionCall: &genai.FunctionCall{Name: "book_flight"}}},
				}}
				if !yield(call, nil) {
					return
				}
				// 3. The stream ends before the final transcription arrives.
			}, nil
		},
	}

	r, err := New(Config{AppName: "app", Agent: mockLive, SessionService: ss})
	if err != nil {
		t.Fatal(err)
	}
	_, stream, err := r.RunLive(ctx, "u", "s", agent.LiveRunConfig{})
	if err != nil {
		t.Fatal(err)
	}

	var delivered []string
	for ev, err := range stream {
		if err != nil {
			delivered = append(delivered, "error: "+err.Error())
			continue
		}
		delivered = append(delivered, describe(ev))
	}

	got, _ := ss.Get(ctx, &session.GetRequest{AppName: "app", UserID: "u", SessionID: "s"})
	persisted := 0
	for range got.Session.Events().All() {
		persisted++
	}
	t.Logf("consumer received: %v", delivered)
	t.Logf("session persisted: %d event(s)", persisted)

	for _, d := range delivered {
		if d == "tool call book_flight" {
			return
		}
	}
	t.Fatalf("the tool call was never delivered and no error was reported; consumer got %v", delivered)
}

func describe(ev *session.Event) string {
	if ev.LLMResponse.Content != nil {
		for _, p := range ev.LLMResponse.Content.Parts {
			if p.FunctionCall != nil {
				return "tool call " + p.FunctionCall.Name
			}
		}
	}
	if ev.LLMResponse.InputTranscription != nil {
		return fmt.Sprintf("transcription(partial=%v)", ev.LLMResponse.Partial)
	}
	return "other"
}
```

**Expected Behavior:**

Buffered events reach the consumer before `RunLive` returns, or the run ends
with an error naming what was dropped.

**Observed Behavior:**

```
--- FAIL: TestRunLive_BufferedToolCallDroppedWhenStreamEnds (0.00s)
    bufdrop_test.go:76: consumer received: [transcription(partial=true)]
    bufdrop_test.go:77: session persisted: 0 event(s)
    bufdrop_test.go:85: the tool call was never delivered and no error was reported
```

**Environment Details:**

- ADK Library Version: `main` at ede87a2f
- OS: Linux
- Go Version: go1.26.6

---

## 🟡 Optional Information

**Severity:**

Only reachable when a live model interleaves a tool call with a transcription
and the stream then ends without a final non-partial transcription. A dropped
tool call is invisible: the user sees the model decline to act, and the session
history gives no clue why.

**Suggested fix:**

Flush after the loop. It runs on every exit path, including the early returns.

```go
defer func() {
    for _, e := range bufferedEvents {
        if err := r.sessionService.AppendEvent(iCtx, storedSession, e); err != nil {
            yield(nil, fmt.Errorf("failed to add event to session: %w", err))
            return
        }
        if !yield(e, nil) {
            return
        }
    }
    bufferedEvents = nil
}()
```

That is not correct as written. `yield` must not be called once it has returned
false, so the deferred flush has to know how the loop exited. A flag set at each
`return` that follows a false `yield` would carry it. Whoever picks this up
should decide between that and reporting the drop as an error instead.

**Note:**

Found while reviewing #1479, which changes the nil-event branch in the same
loop. Not caused by that PR; it reproduces on `main` with no nil event
involved.

rksharma-owg:
Closes: #1507

## Problem
`Runner.RunLive` buffers tool calls that arrive during a partial transcription. If the live iterator ends before a final transcription event arrives, those calls are silently dropped and never persisted.

## Solution
Flush any remaining buffered events after natural iterator exhaustion, preserving append-before-yield ordering and surfacing append failures through the existing error path. Consumer cancellation and other early-return paths continue to stop without yielding after cancellation.

## Behavior change
On current releases, a live stream ending mid-transcription can lose buffered tool calls. With this fix, the buffered calls are delivered and persisted when the stream naturally ends.

## Testing Plan
- `go test ./runner/ -run TestRunner_RunLive_FlushesBufferedToolCallWhenStreamEnds -count=1 -v`
- `go test ./runner/ -run TestRunner_RunLive_(ChronologicalBuffering|FlushesBufferedToolCallWhenStreamEnds) -count=1 -v`
- `go mod tidy -diff`
- `golangci-lint run`
- plugin/agentanalytics module build, race tests, tidy check, and lint

The new regression test failed before the source change with only the partial transcription delivered and zero persisted events, then passed with the fix.

### DIFF
--- a/runner/live_runner_test.go
+++ b/runner/live_runner_test.go
@@ -373,3 +373,82 @@ func TestRunner_RunLive_ChronologicalBuffering(t *testing.T) {
 		t.Errorf("expected second saved event to be function call, but got %v", events.At(1))
 	}
 }
+
+func TestRunner_RunLive_FlushesBufferedToolCallWhenStreamEnds(t *testing.T) {
+	ctx := t.Context()
+	appName, userID, sessionID := "testApp", "testUser", "testSessionEOF"
+
+	sessionService := session.InMemoryService()
+	if _, err := sessionService.Create(ctx, &session.CreateRequest{
+		AppName:   appName,
+		UserID:    userID,
+		SessionID: sessionID,
+	}); err != nil {
+		t.Fatal(err)
+	}
+
+	testAgent := must(agent.New(agent.Config{Name: "test_agent"}))
+	mockLive := &mockLiveAgent{
+		Agent: testAgent,
+		runLiveFn: func(ctx agent.InvocationContext) (agent.LiveSession, iter.Seq2[*session.Event, error], error) {
+			return &dummyLiveSession{}, func(yield func(*session.Event, error) bool) {
+				partial := session.NewEvent(ctx, ctx.InvocationID())
+				partial.LLMResponse.Partial = true
+				partial.LLMResponse.InputTranscription = &genai.Transcription{Text: "book me a "}
+				if !yield(partial, nil) {
+					return
+				}
+
+				call := session.NewEvent(ctx, ctx.InvocationID())
+				call.LLMResponse.Content = &genai.Content{
+					Parts: []*genai.Part{{FunctionCall: &genai.FunctionCall{Name: "book_flight"}}},
+				}
+				yield(call, nil)
+			}, nil
+		},
+	}
+
+	r, err := New(Config{
+		AppName:        appName,
+		Agent:          mockLive,
+		SessionService: sessionService,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, iter, err := r.RunLive(ctx, userID, sessionID, agent.LiveRunConfig{})
+	if err != nil {
+		t.Fatalf("RunLive failed: %v", err)
+	}
+
+	var delivered []*session.Event
+	for event, err := range iter {
+		if err != nil {
+			t.Fatalf("RunLive yielded an error: %v", err)
+		}
+		delivered = append(delivered, event)
+	}
+
+	if len(delivered) != 2 {
+		t.Fatalf("consumer received %d events, want partial transcription and buffered tool call", len(delivered))
+	}
+	if delivered[1].LLMResponse.Content == nil || delivered[1].LLMResponse.Content.Parts[0].FunctionCall == nil {
+		t.Fatalf("consumer did not receive the buffered tool call: %v", delivered[1])
+	}
+
+	got, err := sessionService.Get(ctx, &session.GetRequest{
+		AppName:   appName,
+		UserID:    userID,
+		SessionID: sessionID,
+	})
+	if err != nil {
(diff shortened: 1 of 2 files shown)


wojtas000:
Thanks @rksharma-owg — correct diagnosis, correct fix, and you filed first.

Closing in favour of #1547, which carries the byte-for-byte identical flush loop plus coverage for the other exit paths (consumer stop mid-flush, `AppendEvent` failure, nil event), and which documents that flushed events keep their original timestamps.

Nothing wrong with this PR; #1547 is simply a superset. Sorry for the duplicated effort, and please do send more.

PR #1547, opened 2026-09-09
fix(runner): preserve buffered live events at stream end

### LINKED ISSUE #1507: runner: RunLive silently drops buffered tool calls when the stream ends mid-transcription

Ultronen:

Closes #1507

When a live stream ends before its transcription completes, RunLive now saves and yields buffered tool calls and responses instead of silently dropping them. The flush runs only after natural upstream exhaustion and keeps the existing append-before-yield behavior. A consumer stop returns immediately, including during the flush or while reporting an append error.

A nil event remains a terminal agent error. If events are buffered at that point, the error lists their discarded IDs; it does not include tool arguments or response bodies. This follows the alternative proposed in the issue discussion without treating an invalid event as a successful end of stream.

### Behavior change

Applications now receive buffered tool events at EOF, and those events appear in session history. Buffered arrival order and original timestamps are preserved, so the saved events can have earlier timestamps than events saved before the flush. Nil-event errors now identify any buffered events that were not saved. Consumer cancellation behavior is unchanged.

### Testing Plan

The complete runner package passes with race detection on both Go 1.26.6 and 1.27.1. Build succeeds across the root and agentanalytics modules; both modules pass `go mod tidy -diff` and the pinned golangci-lint v2.3.1 with Go 1.26.6.

The previous full-suite timeout is diagnosed: the unchanged VertexAI RPC replay helper uses blocking gRPC dial options and attempts to connect to the real endpoint during client initialization. Independent test-only fix #1548 removes that external connection requirement.

With the exact `session/vertexai/service_test.go` from #1548 (commit `790e6887563315fab99af91c71d54471ea7ed720`) applied via a Go overlay, the complete `go test -race -mod=readonly -overlay=/tmp/adk-combined-replay.json -count=1 -shuffle=on -timeout=120s work` run passes all 82 packages with tests, including agentanalytics. The overlay changes only that test helper and its regression test; this PR's runner diff remains unchanged. This is combined validation, not a claim that this branch alone fixes offline VertexAI initialization.

To reproduce, save the companion file locally and create an overlay JSON with `{"Replace":{"/absolute/path/to/adk-go/session/vertexai/service_test.go":"/absolute/path/to/companion/service_test.go"}}`. Run the command above from a Go workspace containing both modules.

Regression coverage includes EOF flush, nil-event termination, append failure with continuation and cancellation, consumer stop during flushing, persistence before delivery, and original timestamps. Six separate Go overlay mutations were each caught by the tests: removing EOF flush, ignoring append errors, stopping after append errors, ignoring consumer stop after an error, ignoring consumer stop during flush, and omitting discarded event IDs.

**With the source change reverted and tests kept:** `TestRunner_RunLive_FlushesBufferedEventsAtEOF` and the EOF/error/stop cases in `TestRunner_RunLive_BufferedEventExits` fail. Removing the nil-event diagnostic fails both nil-event cases.

```text
$ GOTOOLCHAIN=go1.26.6 go test -race -mod=readonly ./runner -count=1
ok  google.golang.org/adk/v2/runner  1.689s

$ GOTOOLCHAIN=go1.26.6 golangci-lint run
0 issues.

$ cd plugin/agentanalytics
$ GOTOOLCHAIN=go1.26.6 golangci-lint run
0 issues.
```

**Offline runner reproduction:** A deterministic agent emits a partial transcription, a function call and response, then an unrelated message before EOF. This uses the real Runner and in-memory session service, with no model or network calls.

<details>
<summary>Reproduction program and command</summary>

Save this as `/tmp/adk-live-repro.go` and run `go run /tmp/adk-live-repro.go` from the repository root:

```go
package main

import (
 "context"
 "fmt"
 "iter"
 "google.golang.org/genai"
)
type liveAgent struct { agent.Agent }
type liveSession struct{}
func (*liveSession) Send(agent.LiveRequest) error { return nil }
func (*liveSession) Close() error { return nil }
func (a *liveAgent) RunLive(ctx agent.InvocationContext) (agent.LiveSession, iter.Seq2[*session.Event,error],error) {
 return &liveSession{},func(yield func(*session.Event,error) bool){
  partial:=session.NewEvent(ctx,ctx.InvocationID()); partial.ID="partial"; partial.Partial=true; partial.InputTranscription=&genai.Transcription{Text:"book a "}
  call:=session.NewEvent(ctx,ctx.InvocationID()); call.ID="call"; call.Content=&genai.Content{Parts:[]*genai.Part{{FunctionCall:&genai.FunctionCall{Name:"book_flight"}}}}
  response:=session.NewEvent(ctx,ctx.InvocationID()); response.ID="response"; response.Content=&genai.Content{Parts:[]*genai.Part{{FunctionResponse:&genai.FunctionResponse{Name:"book_flight",Response:map[string]any{"ok":true}}}}}
  later:=session.NewEvent(ctx,ctx.InvocationID()); later.ID="later"; later.Content=genai.NewContentFromText("later message",genai.RoleModel)
  for _,event:=range []*session.Event{partial,call,response,later}{ if !yield(event,nil){return} }
 },nil
}
func main(){
 ctx:=context.Background(); service:=session.InMemoryService()
 a,err:=agent.New(agent.Config{Name:"repro"}); if err!=nil {panic(err)}
 r,err:=runner.New(runner.Config{AppName:"repro",Agent:&liveAgent{a},SessionService:service,AutoCreateSession:true}); if err!=nil {panic(err)}
 _,stream,err:=r.RunLive(ctx,"user","session",agent.LiveRunConfig{}); if err!=nil {panic(err)}
 fmt.Print("consumer:")
 for event,err:=range stream {if err!=nil {panic(err)}; fmt.Print(" ",event.ID)}
 saved,err:=service.Get(ctx,&session.GetRequest{AppName:"repro",UserID:"user",SessionID:"session"}); if err!=nil {panic(err)}
 fmt.Print("\nsession:")
 for event:=range saved.Session.Events().All(){fmt.Print(" ",event.ID)}
 fmt.Println()
}
```

</details>

```text
consumer: partial later call response
session: later call response
```

### Checklist

- [x] I have read CONTRIBUTING.md.
- [x] I have performed a self-review and an independent code review.
- [x] I have added regression tests and verified they detect reverted behavior.
- [x] I have exercised the runner end-to-end with a deterministic offline agent.
- [x] All 82 packages with tests pass in combined validation with the test-only fix from #1548.
- [ ] Companion replay fix #1548 merged; branch-only offline full-suite validation can then be rerun.


wojtas000:
Duplicate of #1544 — the flush loop is byte-for-byte identical. Going with this one for the extra exit-path coverage (consumer stop mid-flush, `AppendEvent` failure, nil event) and for calling out that flushed events keep their original timestamps, so session history can come back non-monotonic. #1544 is closed with credit.

Verified locally: build, `go test -race -shuffle=on ./runner/...`, and golangci-lint v2.3.1 all clean. With `runner.go` reverted to main, `FlushesBufferedEventsAtEOF` and all six `BufferedEventExits` subtests go red.

Ultronen:
Hi @wojtas000 — thanks for the review notes. Glad this one's the chosen path forward given the exit-path coverage. Could you go ahead and formally approve when you get a chance? Happy to address any remaining feedback first if needed.

Ultronen:
Hi @wojtas000,

Following up on this one — appreciate your confirmation that this is the right path forward given the exit-path coverage.

The PR has been ready for formal approval for 4 days now. Could you go ahead and approve when you get a chance? Happy to address any remaining feedback if needed.

Thanks!

wojtas000:
Hi, let me do the re review

runner/runner.go
```diff
@@ -1045,6 +1052,20 @@ func (r *Runner) RunLive(ctx context.Context, userID, sessionID string, cfg agen
 				return
 			}
 		}
+
+		// Reaching here means innerIter exhausted naturally. Every downstream
+		// stop returns from wrappedIter above, so buffered events are safe to flush.
```
wojtas000: important: this does not hold for the only live upstream in the repo. `liveSessionImpl.recvIter` loops until `Close()` closes `done`, so it never exhausts, and the producer goroutine ends it from `defer sess.Close()`. In production this flush runs at teardown, not at exhaustion. The second sentence is the load-bearing one and it is right: I read every `yield` in the loop above and each false return leaves `wrappedIter`.

```suggestion
		// innerIter has returned, so no downstream stop is outstanding: every yield
		// above returns from wrappedIter when the consumer stops. For a live agent
		// this point is the session closing, not the stream running out.
```

runner/runner.go
```diff
@@ -1045,6 +1052,20 @@ func (r *Runner) RunLive(ctx context.Context, userID, sessionID string, cfg agen
...
+		for _, bufferedEvent := range bufferedEvents {
+			if err := r.sessionService.AppendEvent(iCtx, storedSession, bufferedEvent); err != nil {
```
wojtas000: important: on a cancelled run this appends with an already-cancelled `iCtx`. `session.InMemoryService` ignores the context, which is why the new tests pass, but `session/database` and `session/vertexai` bind their writes to it, so each buffered event turns into `failed to add event to session: context canceled` and the buffer is lost anyway. Repro in the review summary. With the line below the same repro delivers and persists both buffered events, and `go test -race ./runner/` stays green.

```suggestion
			if err := r.sessionService.AppendEvent(context.WithoutCancel(iCtx), storedSession, bufferedEvent); err != nil {
```

If the buffer should instead be dropped on cancellation, that is a fine answer too, but say so in the comment. Either way it wants a subtest in `TestRunner_RunLive_BufferedEventExits`, which today has no cancellation case.

runner/runner.go
```diff
@@ -953,6 +953,13 @@ func (r *Runner) RunLive(ctx context.Context, userID, sessionID string, cfg agen
 
 			if event == nil {
 				err := fmt.Errorf("adk: agent %q yielded a nil event", agentToRun.Name())
+				if len(bufferedEvents) > 0 {
+					ids := make([]string, 0, len(bufferedEvents))
+					for _, bufferedEvent := range bufferedEvents {
+						ids = append(ids, bufferedEvent.ID)
+					}
+					err = fmt.Errorf("%w; discarded buffered event IDs: %q", err, ids)
```
wojtas000, line 961: optional: the ID list is unbounded, and `adkrest` forwards this error text as a websocket close reason (`server/adkrest/controllers/runtime.go:500`). RFC 6455 caps a close frame payload at 125 bytes; gorilla returns `errInvalidControlFrame` rather than truncating and the caller discards it, so with a 10-character agent name two buffered UUIDs (154 bytes) leave the client with a close and no reason at all. Bounding it keeps the message deliverable.

```suggestion
					err = fmt.Errorf("%w; discarded %d buffered events, first IDs: %q", err, len(ids), ids[:min(len(ids), 3)])
```

runner/runner.go
```diff
@@ -1045,6 +1052,20 @@ func (r *Runner) RunLive(ctx context.Context, userID, sessionID string, cfg agen
 				return
 			}
 		}
+
+		// Reaching here means innerIter exhausted naturally. Every downstream
+		// stop returns from wrappedIter above, so buffered events are safe to flush.
+		for _, bufferedEvent := range bufferedEvents {
```
wojtas000, line 1061: nit: the flush appends unconditionally, where the normal path skips the append when `hasInlineData(event)` is true (`runner/runner.go:1042`). A buffered function-response event that also carries an inline blob now lands in session history with the blob. The mid-stream flush at `runner/runner.go:1026` has the same gap, so this is consistent rather than new — worth either a line saying the omission is deliberate, or fixing both in a separate PR.

wojtas000 (changes requested):
The fix covers what #1507 asked for. I reverted `runner/runner.go` to main, kept the new test file, and all seven new cases went red, so the tests pin the change.

Two things I would like settled. The comment at `runner.go:1056` does not hold for the only live upstream in the repo, and the path it misses is untested. [`liveSessionImpl.recvIter`](https://github.com/google/adk-go/blob/fb856e94c7519c7ebff38dcb58c42514e850e065/internal/llminternal/base_flow.go#L274-L287) never exhausts — it loops until `Close()` closes `done`, and the producer goroutine closes it from [`defer sess.Close()`](https://github.com/google/adk-go/blob/fb856e94c7519c7ebff38dcb58c42514e850e065/internal/llminternal/base_flow.go#L330-L333). So in production the flush runs at teardown, and one kind of teardown is context cancellation. There the flush appends on an already-cancelled `iCtx`: `session.InMemoryService` ignores the context, which is why the new tests do not see it, but `session/database` and `session/vertexai` bind their writes to it, so every buffered event becomes `failed to add event to session: context canceled` and the buffer is still lost.

Save this as `runner/zz_repro_test.go` and run `go test ./runner/ -run TestZZRepro -v`:

```go
package runner

import (
	"context"
	"iter"
	"testing"

	"google.golang.org/genai"
)

// ctxBoundService models a session service whose writes honour the caller's
// context, as session/database and session/vertexai do.
type ctxBoundService struct{ session.Service }

func (s *ctxBoundService) AppendEvent(ctx context.Context, sess session.Session, e *session.Event) error {
	if err := ctx.Err(); err != nil {
		return err
	}
	return s.Service.AppendEvent(ctx, sess, e)
}

func TestZZRepro_FlushOnContextCancellation(t *testing.T) {
	inner := session.InMemoryService()
	if _, err := inner.Create(t.Context(), &session.CreateRequest{AppName: "app", UserID: "u", SessionID: "s"}); err != nil {
		t.Fatal(err)
	}
	ctx, cancel := context.WithCancel(t.Context())
	defer cancel()

	live := &mockLiveAgent{
		Agent: must(agent.New(agent.Config{Name: "live_agent"})),
		runLiveFn: func(ic agent.InvocationContext) (agent.LiveSession, iter.Seq2[*session.Event, error], error) {
			return &dummyLiveSession{}, func(yield func(*session.Event, error) bool) {
				partial := session.NewEvent(ic, ic.InvocationID())
				partial.LLMResponse.Partial = true
				partial.LLMResponse.InputTranscription = &genai.Transcription{Text: "book me a "}
				if !yield(partial, nil) {
					return
				}
				for _, name := range []string{"call", "response"} {
					e := session.NewEvent(ic, ic.InvocationID())
					e.ID = name
					e.LLMResponse.Content = &genai.Content{Parts: []*genai.Part{{FunctionCall: &genai.FunctionCall{Name: "book_flight"}}}}
					if !yield(e, nil) {
						return
					}
				}
				// What internal/llminternal does on ctx.Done(): push the context
				// error, then let the deferred sess.Close() end recvIter.
				cancel()
				yield(nil, ctx.Err())
			}, nil
		},
	}

	r, err := New(Config{AppName: "app", Agent: live, SessionService: &ctxBoundService{Service: inner}})
	if err != nil {
		t.Fatal(err)
	}
	_, stream, err := r.RunLive(ctx, "u", "s", agent.LiveRunConfig{})
	if err != nil {
		t.Fatal(err)
	}
	for event, err := range stream {
		if err != nil {
			t.Logf("error: %v", err)
			continue
		}
		t.Logf("event: %s", event.ID)
	}
}
```

On this branch it logs the partial event, `context canceled`, then two `failed to add event to session: context canceled`. On main it stops after the first two. Passing `context.WithoutCancel(iCtx)` to the flush's `AppendEvent` instead makes the same repro log `event: call` and `event: response`, with `./runner/` still green — one line, and the flush then does its job on the path it most often runs on.

The rest is inline.

PR #1550, opened 2026-09-09
fix(workflow): stop yielding after consumer exits

### LINKED ISSUE #1549: workflow: DynamicNode panics when stream consumer stops after emitted event
### Description

When a consumer stops ranging over a `DynamicNode` or emitting `FunctionNode` after receiving an emitted event, the node body receives an error from `emit` and returns it. The node then calls `yield` again to surface that error or its terminal output, violating the range-over-function contract because the previous `yield` returned false.

This produces:

```text
runtime error: range function continued iteration after function for loop body returned false
```

`makeEmit` also permits another `emit` call to invoke `yield` again after the consumer has already rejected an event. This can affect concurrent emitters waiting on its mutex.

### Expected behavior

Once `yield` returns false, the emitter should remember that the consumer is gone, reject later emissions without calling `yield`, and the node should return without yielding an error or terminal output.

### Reproduction

1. Create a `DynamicNode` or `NewEmittingFunctionNode` whose body emits an intermediate event and returns the resulting error.
2. Range over `Node.Run` and break after the first event.
3. The node panics when it tries to yield again.

### Environment

- ADK Go `main` at `d56d1a808c46d70b3f2a12d7aa6b4ff9421e45c8`
- Reproduced with Go 1.26.7 and Go 1.27.0

I have a focused fix and regression tests for both node types.

cruvie:
Fixes #1549.

## Summary

- record when an emitter sees `yield` return false
- stop `DynamicNode` and emitting `FunctionNode` before they yield an error or terminal output to an exited consumer
- reject later and concurrently waiting emissions without invoking `yield` again
- preserve the real context error on later emission attempts
- add regression coverage for both node types, ignored emission errors, error and wait paths, repeated emissions, and concurrent emitters

## Behavior change

For existing users, stopping iteration after an emitted workflow event now terminates the node cleanly. A consumer ranging `Node.Run` directly could previously panic with:

```text
runtime error: range function continued iteration after function for loop body returned false
```

Through the graph engine, that panic was recovered and reported as `node %q panicked: ...`, overwriting the real cancellation cause already recorded by the scheduler. This change avoids that false attribution. Normal streaming, node errors, terminal outputs, and public APIs are unchanged.

## Testing plan

The following regression tests pass with the fix and were verified against targeted source mutations:

- `TestDynamicNode_StopsAfterConsumerExits`: removing or narrowing the guard reproduces the iterator panic for returned emit errors, ignored emit errors, independent errors, and `ErrNodeWaitingForOutput`
- `TestEmittingFunctionNode_StopsAfterConsumerExits`: removing or narrowing the guard reproduces the iterator panic for returned emit errors, ignored emit errors, and independent errors
- `TestMakeEmitStopsCallingYieldAfterConsumerExits`: removing the stopped check calls `yield` twice
- `TestMakeEmitStopsConcurrentWaiterAfterConsumerExits`: moving the stopped check outside the mutex produces two yield calls and a race
- `TestMakeEmitPreservesContextErrorAfterConsumerExits`: checking stopped before `parentCtx.Err()` changes the second error from `context.DeadlineExceeded` to `context.Canceled`

Validated locally with Go 1.27.0:

- `golangci-lint v2.12.2 run` in the root module (`0 issues`)


workflow/dynamic_node.go
```diff
@@ -197,20 +201,31 @@ func (n *dynamicNode[IN, OUT]) composePath(parent agent.Context) string {
...
 // the iterator and races the parent runNode's completion accumulator.
-func makeEmit(yield func(*session.Event, error) bool, parentCtx agent.Context) func(*session.Event) error {
+func makeEmit(yield func(*session.Event, error) bool, parentCtx agent.Context) (func(*session.Event) error, func() bool) {
 	var mu sync.Mutex
-	return func(ev *session.Event) error {
+	stopped := false
+	emit := func(ev *session.Event) error {
 		mu.Lock()
 		defer mu.Unlock()
+		if stopped {
```
wolo-lab: This ordering is the one thing I'd change before merge. `stopped` is consulted before `parentCtx.Err()` two lines down, so once the consumer has gone, every later `emit` returns a hardcoded `context.Canceled` even when the context died of a deadline. The first `emit` is fine — the `!yield` branch below checks the context before falling back to the stand-in. It is calls two onward that lose the real error.

Swapping the two blocks fixes it and leaves the package green. I ran that as a control, varying nothing else. The repro and the base comparison are in the review body.

workflow/dynamic_node_test.go
```diff
@@ -137,6 +138,44 @@ func TestDynamicNode_EmitMidBody(t *testing.T) {
...
+		NodeConfig{},
+	)
+
+	parent := agent.NewContext(newMockCtx(t))
+	for range orchestrator.Run(parent, "") {
+		break
+	}
+}
+
+func TestMakeEmitStopsCallingYieldAfterConsumerExits(t *testing.T) {
```
wolo-lab, line 215: This test's context is alive for its whole duration, since `newMockCtx` is backed by `t.Context()`. So `errors.Is(err, context.Canceled)` holds whether the value came from the stand-in or from a real cancellation, and the dead-context branch in `makeEmit` is never reached — it is the only part of that function with no coverage.

Giving this case a context whose deadline has already elapsed would exercise that branch and would also have caught the ordering issue flagged on `dynamic_node.go`.

workflow/function_node.go
```diff
@@ -359,8 +359,11 @@ func (n *FunctionNode) Run(ctx agent.Context, input any) iter.Seq2[*session.Even
 // event was already emitted), any other error fails the node, and a
 // nil output suppresses the terminal event.
 func (n *FunctionNode) runEmitting(ctx agent.Context, input any, yield func(*session.Event, error) bool) {
-	emit := makeEmit(yield, ctx)
+	emit, consumerGone := makeEmit(yield, ctx)
 	output, err := n.emittingFn(ctx, input, emit)
+	if consumerGone() {
```
wolo-lab, line 364: Same guard as in `dynamic_node.go`, and the same test gap applies here: nothing exercises this path with a body that ignores the emit error and returns a normal output. The guard as written handles that correctly — the concern is that a wrong variant of it would also pass. Details and a runnable case are in the review body.

wolo-lab (changes requested):
Two things to fix before this merges, both small. The change itself is right: the guard is placed at a dominator so it covers the `WaitForOutput` and `ErrNodeInterrupted` paths as well as the terminal event, which a narrower fix would have missed. I confirmed it red/green on both node types against the merge base, and confirmed that deleting either guard reintroduces the panic.

## The second and later `emit` now lose the real context error

[`makeEmit`](https://github.com/google/adk-go/blob/37adbeac579bcf86e143519f49597b5edc3e08fb/workflow/dynamic_node.go#L210-L215) checks `stopped` before it checks `parentCtx.Err()`. Once the consumer has gone, every later `emit` returns a hardcoded `context.Canceled`, even when the context died of a deadline. The first `emit` is fine, because [line 218](https://github.com/google/adk-go/blob/37adbeac579bcf86e143519f49597b5edc3e08fb/workflow/dynamic_node.go#L217-L221) consults the context before falling back to the stand-in. It is only calls two onward that lose it.

This is a behaviour change from `main`, not a pre-existing quirk. Dropping this into `workflow/`:

```go
func TestOrderingLosesDeadline(t *testing.T) {
	dead, cancel := context.WithDeadline(context.Background(), time.Now().Add(50*time.Millisecond))
	defer cancel()
	pc := agent.NewContext(&MockInvocationContext{Context: dead})

	// The consumer stops AND the deadline elapses during the same yield —
	// the shape scheduler.go:485-489 produces.
	yield := func(*session.Event, error) bool {
		time.Sleep(120 * time.Millisecond)
		return false
	}
	emit, _ := makeEmit(yield, pc)

	_ = emit(&session.Event{})
	if err := emit(&session.Event{}); err != context.DeadlineExceeded {
		t.Errorf("second emit = %v, want context.DeadlineExceeded", err)
	}
}
```

`go test -run TestOrderingLosesDeadline ./workflow/` passes on the merge base and fails here with `second emit = context canceled`. Swapping the two blocks so `parentCtx.Err()` is consulted first restores it and leaves the rest of the package green — I ran that as a control, varying nothing else.

No in-tree body reaches a second `emit`, so nothing in the repo is affected today. `runner/agent_node.go`, the sub-scheduler and all six `examples/` emit sites return on the first error. The exposure is to `DynamicFn` authors outside the repo, which is the population issue #1549 is about.

## The guard's other half is untested, and a wrong guard passes the suite

Both new node tests use a body that returns the emit error. Nothing covers the other outcome the guard exists for: a body that ignores the error and returns normally. That gap is wide enough to admit a guard that is plainly wrong.

Write the guard as `if consumerGone() && err != nil` in [dynamic_node.go](https://github.com/google/adk-go/blob/37adbeac579bcf86e143519f49597b5edc3e08fb/workflow/dynamic_node.go#L120-L122) and [function_node.go](https://github.com/google/adk-go/blob/37adbeac579bcf86e143519f49597b5edc3e08fb/workflow/function_node.go#L364-L366) and `go test ./workflow/` still exits 0. Against a body that swallows the error it panics with exactly the message from #1549, in both node types:

```go
// Add to workflow/ and run against the `&& err != nil` variant.
func TestSwallowingBody(t *testing.T) {
	defer func() {
		if r := recover(); r != nil {
			t.Fatalf("panicked after consumer break: %v", r)
		}
	}()
	n := NewDynamicNode[string, string]("orch",
		func(_ agent.Context, _ string, emit func(*session.Event) error) (string, error) {
			_ = emit(&session.Event{}) // ignored, as a real body might
			return "done", nil
		}, NodeConfig{})

	count := 0
	for range n.Run(agent.NewContext(newMockCtx(t)), "") {
		count++
		break
	}
	if count != 1 {
		t.Errorf("consumed %d events, want 1", count)
	}
}
```

Your code passes this as written. A variant carrying `ErrNodeWaitingForOutput` through the guard also survives the suite, so a case pairing consumer-exit with a parked child would be worth having too.

## Smaller things

The two node tests have no assertions at all, so a regression surfaces as a stack trace that kills the whole package run rather than as one named failure. That is not hypothetical here: reproducing this on the merge base, the first panic aborted the process before the second test could run. [`TestNestedWorkflow_BreakMidStream`](https://github.com/google/adk-go/blob/37adbeac579bcf86e143519f49597b5edc3e08fb/workflow/workflow_node_test.go#L394-L432) already solves this for `WorkflowNode` — the same bug class, three files away — with a `defer recover()` turning the panic into a readable `t.Fatalf`, an event count, and a comment saying what the test proves. Copying that shape would also close the "yields zero events" hole, where the loop body never runs and the test passes having exercised nothing.

`consumerGone` now names three different things in the package: the closure here, the drain-but-keep-consuming flag in [`workflow_node.go`](https://github.com/google/adk-go/blob/37adbeac579bcf86e143519f49597b5edc3e08fb/workflow/workflow_node.go#L61-L64), and the caller-broke flag in [`scheduler.go`](https://github.com/google/adk-go/blob/37adbeac579bcf86e143519f49597b5edc3e08fb/workflow/scheduler.go#L557). They mean the same thing but each layer responds differently, so a short comment on the new one saying it means "return immediately" would help.

One wording point in the description. "The node could panic" holds for a consumer ranging `Node.Run` directly, which is what #1549 describes. Through the graph engine the panic is recovered at [`scheduler.go:452`](https://github.com/google/adk-go/blob/37adbeac579bcf86e143519f49597b5edc3e08fb/workflow/scheduler.go#L451-L457) and reported as `node %q panicked: …`, which overwrites the real cancellation cause the scheduler had already recorded. Worth saying, because removing that false attribution is a real benefit of this change that the description does not claim.

## Not yours to fix here

The terminal yields at [`dynamic_node.go:157`](https://github.com/google/adk-go/blob/37adbeac579bcf86e143519f49597b5edc3e08fb/workflow/dynamic_node.go#L157) and [`function_node.go:384`](https://github.com/google/adk-go/blob/37adbeac579bcf86e143519f49597b5edc3e08fb/workflow/function_node.go#L384) still run outside the mutex, so an emitter goroutine that outlives the body can race them. I can reach that on this branch, and I can reach it identically on `main` — this change strictly narrows it. Two related things predate it as well: nothing marks the iterator finished on the normal-return path, and `makeEmit`'s comment says a `DynamicFn` may emit from concurrent children while [`branch_isolation_test.go`](https://github.com/google/adk-go/blob/37adbeac579bcf86e143519f49597b5edc3e08fb/workflow/branch_isolation_test.go#L32-L36) says emit is single-goroutine only. [`run_node_test.go:681`](https://github.com/google/adk-go/blob/37adbeac579bcf86e143519f49597b5edc3e08fb/workflow/run_node_test.go#L681-L716) fans out eight goroutines through one yield, so the code behaves as though the first is true. Happy to open a separate issue for that rather than grow this PR.

Separately and also pre-existing: `runEmitting` puts a returned `*genai.Content` into `event.Output`, where the non-emitting `FunctionNode.Run` type-switches it into `event.Content`. Untouched by this change, byte-identical on `main`.

Last thing, mechanical: the branch is five commits behind `main` and needs updating before it can merge. Nothing in those commits touches `workflow/`, and I ran everything against the merge result, which is clean.

cruvie:
Addressed the review in 165e85d:

- check `parentCtx.Err()` before the stopped fallback so later emits preserve `DeadlineExceeded`
- cover returned/ignored emit errors, independent errors, and `ErrNodeWaitingForOutput`, with panic recovery and event-count assertions
- cover a concurrent emitter waiting on the mutex
- verify each guard/error path with targeted failing mutations
- rebase onto current `main` and update the PR description/testing plan

Local build and full race/shuffle tests pass. Please take another look when CI finishes.

workflow/dynamic_node.go
wolo-lab: Fixed.

wolo-lab (approved):
Nothing blocks — both things I raised are fixed, and I verified.

PR #1551, opened 2026-09-09
fix(session/database): surface JSON marshal errors on state writes

### LINKED ISSUE #1537: session/database: stateMap.GormValue silently drops JSON marshal errors
## Describe the bug

`session/database` has two serialization paths for its internal `stateMap`, but only one preserves JSON errors. `Value()` returns `json.Marshal` failures; `GormValue()` currently does `data, _ := json.Marshal(sm)` and binds `string(data)`.

GORM dispatches its `GormValuer` path before `driver.Valuer` during normal writes, so the error-returning `Value()` path is bypassed. A state value rejected by `encoding/json`, such as `math.NaN()`, can therefore become an empty payload instead of failing at the write boundary.

The public session state accepts `any`, and `databaseService.Create` copies request state into the storage session without an earlier JSON-validation step.

## Reproducer / causal chain

```go
sm := stateMap{"bad": math.NaN()}
_, err := sm.Value() // json: unsupported value: NaN
expr := sm.GormValue(context.Background(), db)
// current implementation ignores the same marshal error and binds ""
_ = expr
```

`non-JSON session state -> stateMap.GormValue -> json.Marshal error discarded -> GORM receives empty string expression`

## Expected behavior

Database-backed session writes should surface serialization errors when state cannot be represented as JSON. They should not silently substitute an empty payload.
## Prior evidence and duplicate check

PR #395 touched the same `GormValue` while adding logging. Its automated review explicitly identified this marshal-error path and warned that it masks the root cause and can pass an empty string to the database. That PR was closed without merge and did not fix the persistence behavior.

I found no open issue and no open PR for the marshal-error behavior itself.

## Scope / direction request

This is narrow to the database state serializer, but `session` is a high-fan-in package. `GormValue` cannot directly return an error, so I do not want to guess at the preferred mechanism. The smallest options appear to be removing the custom `GormValue` if the existing `driver.Valuer` path is sufficient, or recording the marshal failure through GORM before returning an expression.

Per `AGENTS.md`, I am asking before implementing: **if this behavior is unintended, may I take this issue and send a focused regression test plus the smallest approved fix?**

Huang-404-Q:
## Describe the bug

`session/database` has two serialization paths for its internal `stateMap`, but only one preserves JSON errors. `Value()` returns `json.Marshal` failures; `GormValue()` currently does `data, _ := json.Marshal(sm)` and binds `string(data)`.

GORM dispatches its `GormValuer` path before `driver.Valuer` during normal writes, so the error-returning `Value()` path is bypassed. A state value rejected by `encoding/json`, such as `math.NaN()`, can therefore become an empty payload instead of failing at the write boundary.

The public session state accepts `any`, and `databaseService.Create` copies request state into the storage session without an earlier JSON-validation step.

## Root cause

`stateMap` implements both `driver.Valuer` (with an error return) and `gorm.GormValuer` (without one). GORM prefers `GormValuer`, and its implementation swallows the marshal error and binds the zero-length string — so the error channel that exists on `Value()` is never exercised for database-backed writes.

## Fix

Remove `stateMap.GormValue` so `stateMap` serializes through `Value()`, whose error path propagates to the caller. `dynamicJSON` keeps its `GormValue`: its content is already valid JSON (`json.RawMessage`) and the method exists to emit `NULL` for empty values, so there is no error to swallow there.

## Testing

- New regression test `TestDatabaseService_NonJSONStateErrorSurfaces`: creating a session whose state contains `math.NaN()` must fail with a `json.UnsupportedValueError` and must not persist the row.
- Verified red-green: the test fails against the old `GormValue` (write silently succeeded) and passes with this change.
- `go test ./session/...` green (database, in-memory, compaction, vertexai suites); `go vet` and `gofmt` clean.

Fixes #1537


indurireddy-TF:
Hi @wolo-lab , this Pr has been submitted to fix this by removing GormValue on stateMap so errors propagate through Value(). Could you please take a look?

PR #1559, opened 2026-09-10
fix(session): snapshot events under read lock in database and vertexa…

### LINKED ISSUE #1553: session/database, session/vertexai: Events() reads s.events without the lock appendEvent writes under
`localSession.Events()` in both the database and the Vertex AI session backends reads `s.events` with no lock, while `appendEvent` in the same file replaces that slice under `s.mu.Lock()`:

- [`session/database/session.go#L60-L62`](https://github.com/google/adk-go/blob/8d9b9c892253b9c4051a16184b35cb09b9aac43e/session/database/session.go#L60-L62), against the write at [`#L76-L77`](https://github.com/google/adk-go/blob/8d9b9c892253b9c4051a16184b35cb09b9aac43e/session/database/session.go#L76-L77)
- [`session/vertexai/session.go#L60-L62`](https://github.com/google/adk-go/blob/8d9b9c892253b9c4051a16184b35cb09b9aac43e/session/vertexai/session.go#L60-L62), against the write at [`#L76-L77`](https://github.com/google/adk-go/blob/8d9b9c892253b9c4051a16184b35cb09b9aac43e/session/vertexai/session.go#L76-L77)

`LastUpdateTime`, immediately below `Events` in both files, does take `s.mu.RLock()`, so the omission looks unintended rather than deliberate. `session/inmemory` takes the read lock and returns a copied slice, which is the behaviour these two are missing.

An unsynchronised read of a slice header against a concurrent `append` that reassigns it can observe a new length with a stale data pointer, so a reader can index past the end of the old backing array.

This is reachable whenever one goroutine reads session history while another appends. The workflow scheduler runs each node in its own goroutine and the consumer appends events as they are yielded, so any node that reads `ctx.Session().Events()` — the LLM flow's contents processor does it on every turn — overlaps an append from a sibling node.

Acceptance: `Events()` takes the read lock and returns a snapshot in both backends, and a test appends from one goroutine while another iterates, run under `-race`.

lucky798213:

Closes #1553.

**Problem:** Concurrent history reads and event appends race on the `localSession.events` slice header in the database and Vertex AI session backends.

**Solution:** Acquire the session read lock and clone the event slice inside `Events()`. The lock is released before callers iterate over the returned snapshot. Both backends receive focused race and snapshot-isolation regression tests.

### Behavior change

Concurrent calls to `Events()` and event appends now synchronize access to the history slice. A returned history has its own backing array. This is a shallow snapshot: event objects remain shared, and concurrent mutation of their fields is outside this fix. Snapshot creation is O(n) in the number of events. No public API or dependencies change.

### Testing Plan

- `go test -race -mod=readonly ./session/database ./session/vertexai -run TestLocalSessionEvents -count=10` passed for both backends.
- Both new tests failed on the unmodified source: `TestLocalSessionEventsConcurrentAppend` reported a data race, and `TestLocalSessionEventsSnapshot` observed the replaced event through the old snapshot.
- Using Go overlays to remove only the read lock reproduced the race in both backends. Removing only the clone made the snapshot test fail in both backends.
- `go build -mod=readonly ./...`, `golangci-lint run` (v2.3.1), and `go mod tidy -diff` passed in both the root module and `plugin/agentanalytics`.
- `go test -race -mod=readonly -count=1 -shuffle=on ./...` passed in `plugin/agentanalytics`. In the root module, 82 packages passed, but the existing Vertex AI service suite blocked while constructing the gRPC client; it was interrupted after approximately 217 seconds to collect stacks.
- The unmodified upstream source at `49224b6f762988234bc8933ed344eb4d3979b44e`, tested through a Go overlay, also timed out after 30 seconds in `Test_vertexaiService/Create/full_key` in `grpc.DialContext`, before reaching `Events()`. Full-suite validation therefore remains incomplete on this host.

Focused result:

```text
ok  google.golang.org/adk/v2/session/database  8.703s
ok  google.golang.org/adk/v2/session/vertexai  9.166s
```

**Manual E2E:** No live Vertex AI or model calls were used. The concurrent test drives the actual local session append/read methods; the existing database service suite passed. No claim of live-cloud E2E validation is made.

### Checklist

- [x] I have read the contribution guide.
- [x] I have performed a self-review of the change.
- [x] I have added tests that fail without the fix.
- [ ] All existing unit tests pass locally (Vertex AI gRPC test blocker described above).
- [ ] Live end-to-end validation completed.

### DIFF
--- a/session/database/session.go
+++ b/session/database/session.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"iter"
 	"maps"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -58,7 +59,11 @@ func (s *localSession) State() session.State {
 }
 
 func (s *localSession) Events() session.Events {
-	return events(s.events)
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	// Return a snapshot so callers can iterate without holding the session lock.
+	return events(slices.Clone(s.events))
 }
 
 func (s *localSession) LastUpdateTime() time.Time {
--- a/session/database/session_test.go
+++ b/session/database/session_test.go
@@ -1,0 +1,88 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package database
+
+import (
+	"fmt"
+	"runtime"
+	"sync"
+	"testing"
+
+	"google.golang.org/adk/v2/session"
+)
+
+func TestLocalSessionEventsSnapshot(t *testing.T) {
+	first := &session.Event{ID: "first"}
+	s := &localSession{events: []*session.Event{first}}
+	snapshot := s.Events()
+
+	// Replacing a slot must not change a previously returned history snapshot.
+	s.mu.Lock()
+	s.events[0] = &session.Event{ID: "replacement"}
+	s.mu.Unlock()
+	if err := s.appendEvent(&session.Event{ID: "second"}); err != nil {
+		t.Fatal(err)
+	}
+	if got := snapshot.Len(); got != 1 {
+		t.Fatalf("snapshot.Len() = %d, want 1", got)
+	}
+	if got := snapshot.At(0); got != first {
+		t.Errorf("snapshot.At(0) = %v, want original event %v", got, first)
+	}
+	for event := range snapshot.All() {
+		if event != first {
+			t.Errorf("snapshot.All() yielded %v, want original event %v", event, first)
+		}
+	}
+}
+
+func TestLocalSessionEventsConcurrentAppend(t *testing.T) {
+	const count = 1000
+	s := &localSession{}
+	start := make(chan struct{})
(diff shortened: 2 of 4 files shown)


wojtas000 (approved):
This does what #1553 asks for, and the tests earn their place: applied to main they report a genuine data race in `Events()` against `appendEvent`, plus a snapshot that observes the replacement event; on this branch both pass under `-race`. I also checked the snapshot cost, and every indexed caller hoists `Events()` out of its loop, so the clone is one pointer copy per turn rather than per index.

Two unlocked reads of the same class are left behind. Both are pre-existing, neither is introduced here, and neither blocks this PR:

- `State()` builds its wrapper over `s.state` without taking the lock, in [database](https://github.com/google/adk-go/blob/e3827d98bfa10dcfed9cd02e42f1c0347abfb489/session/database/session.go#L54-L59) and [vertexai](https://github.com/google/adk-go/blob/e3827d98bfa10dcfed9cd02e42f1c0347abfb489/session/vertexai/session.go#L54-L59), while `updateSessionState` can reassign that map under the write lock. A `Set` issued after a reassignment then writes to a map the session no longer holds.
- In the database backend, [`service.go:394`](https://github.com/google/adk-go/blob/e3827d98bfa10dcfed9cd02e42f1c0347abfb489/session/database/service.go#L394) assigns `sess.updatedAt` with no lock, while `LastUpdateTime()` reads that field under `RLock`. The vertexai backend makes the same assignment inside `appendEvent`, under the write lock.

Worth saying in the description whether this PR closes the class for `localSession` or only for `Events()`, and filing a follow-up for the rest.

One observation with no change needed: [`session/inmemory.go`](https://github.com/google/adk-go/blob/e3827d98bfa10dcfed9cd02e42f1c0347abfb489/session/inmemory.go#L348-L352) takes the read lock but returns an alias rather than a clone. After this lands, the three `Session.Events()` implementations disagree on whether the caller gets a snapshot, and the interface documents neither.

PR #1561, opened 2026-09-10
 feat(adkrest): add authentication and authorization to the REST API

kdroste-google:
# Summary
Adds an optional, pluggable authentication and authorization layer to the adkrest HTTP server. Two new packages define small seams that providers implement:

1. server/authn — inbound HTTP authentication. An Authenticator resolves a request to an *Caller (a stable UserID plus optional provider claims), and authn.Middleware gates every non-public route: missing/invalid credentials are answered 401, an internal provider failure 500, and on success the identity is placed on the request context. Built-in providers:  
    1. NewNoop (authenticates everyone as an empty identity — the default), 
    2. NewHeader (trusts a single request header set by an upstream proxy), 
    3. NewIdentityAwareProxy (reads Google Cloud IAP's X-Goog-Authenticated-User-{Id,Email} headers), and 
    4. NewCustom (adapts a plain function).
2. server/authz — authorization. An Authorizer.CanActAsUser decides whether the authenticated identity may act as the {user_id} named in a request path; a mismatch is answered 403. Built-in: 
    1. Noop (permit-all — the default) and 
    2. Strict (requires the authenticated UserID to equal the path user).

# Wiring

adkrest.ServerConfig and launcher.Config gain Authenticator and Authorizer fields; the web api sublauncher forwards them into NewServer.
Routes carry a Public flag that defaults to false, so a new route requires authentication unless it explicitly opts out. Only /health and /version are public.
The user-scoped controllers (sessions, artifacts, runtime run / run-sse / run-live, and the debug event-graph route) call CanActAsUser before acting. Controller constructors default the authorizer to authz.Noop so the seam is never nil.

# What changes for an existing user
Nothing, unless they opt in. With no Authenticator configured — the default — the server uses authn.Noop and authz.Noop: every endpoint stays reachable without credentials and any caller may act as any user, exactly as before this change. TestAuthDisabledByDefault pins that.

New behavior is available only when configured:

Setting an Authenticator makes every endpoint except /health and /version answer 401 without valid credentials.
Additionally setting authz.Strict makes a request whose path names a different user than the authenticated identity answer 403. Strict compares UserIDs verbatim, so it must be paired with an authenticator that sets a non-empty UserID (pairing it with authn.Noop denies every real request).

# Known limitation
The two debug trace endpoints — GET /debug/trace/session/{session_id} and GET /debug/trace/{event_id} — are not user-scoped (they carry no {user_id}), so they require authentication but perform no authorization: any authenticated caller can read any session's or event's telemetry, which includes prompts and model output. These are served only when the debug API is explicitly enabled (DebugAPIConfig.IncludeDebugAPI / --include_debug_api), which is now documented as unsafe for production in both the flag help and the config field. Per-user scoping of these routes is left as a follow-up.


server/adkrest/controllers/debug.go
wolo-lab, line 1: we need to add auth to this as well, right?

server/authz/strict.go
wolo-lab, line 1: maybe I was not able to find it quickly but do we have tests verifying that non-nill responses are being thrown?

server/adkrest/authn/iap.go
```diff
@@ -0,0 +1,55 @@
...
+	// Example value: accounts.google.com:example@gmail.com
+	UserEMailHeader = "X-Goog-Authenticated-User-Email"
+
+	// UserIDHeader is header name which contains user id (see https://docs.cloud.google.com/iap/docs/identity-howto).
+	// Example value: accounts.google.com:userIDvalue
+	UserIDHeader = "X-Goog-Authenticated-User-Id"
+)
+
+// Authenticate implements [Authenticator].
+func (iap *IdentityAwareProxy) Authenticate(r *http.Request) (*Identity, error) {
```
wolo-lab: would it work with strict checker? I think the prefix might be problematic.

cmd/launcher/launcher.go
```diff
@@ -82,6 +84,20 @@ type Config struct {
 	PluginConfig     runner.PluginConfig
 	TelemetryOptions []telemetry.Option
 
+	// Authenticator, when non-nil, authenticates inbound requests to every
```
wolo-lab: "every endpoint" - we have also a2a, pubsub etc. which won't be covered.

wolo-lab (commented):
partial review

server/adkrest/controllers/debug.go
kdroste-google: Added, but not all paths contain userID. Added some warning for CLI flag about not using this debug in PROD

server/authz/strict.go
kdroste-google: sorry, could you explain your point?

server/adkrest/authn/iap.go
kdroste-google: Added stripping the prefix. It all depends on how the paths are built anyway (how username is embedded).

server/adkrest/controllers/debug.go
kdroste-google: affected: SessionSpansHandler and EventSpanHandler

server/authz/strict.go
wolo-lab: It's already obsolete. I was trying to say that there weren't tests for non-nil results.

server/adkrest/handler.go
```diff
@@ -76,8 +78,24 @@ func NewServer(cfg ServerConfig) (*Server, error) {
 	// TODO: Allow taking a prefix to allow customizing the path
 	// where the ADK REST API will be served.
 
+	authenticator := cfg.Authenticator
```
wolo-lab: maybe have it closer to the place where it is used? or you intentionally want to keep them both close to each other?

server/adkrest/handler.go
```diff
@@ -76,8 +78,24 @@ func NewServer(cfg ServerConfig) (*Server, error) {
...
+	if authenticator == nil {
+		authenticator = authn.NewNoop()
```
wolo-lab: we have `authz.Noop{}` and `&authz.Noop{}`. 
Let's have consistent, one style.

server/authn/authn.go
wolo-lab, line 1: shouldn't it be outside adkrest? It is not adkrest specific.

server/authn/authn.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+// ErrUnauthenticated reports that a request carried no valid credentials. An
+// [Authenticator] returns it, or an error wrapping it, to make [Middleware]
+// answer 401 Unauthorized. Any other error is treated as an internal provider
+// failure and answered 500.
+var ErrUnauthenticated = errors.New("authn: unauthenticated")
+
+// Identity is the authenticated principal resolved from a request.
+type Identity struct {
```
wolo-lab: I have some doubts regarding naming as I was thinking about auth for the whole adk.

I was thinking about having `agent.IdentityFromContext(ctx)` which would be super similar to `authn.IdentityFromContext(ctx)`. Easy to make a mistake. Especially, that both will return a sensible answer.

To avoid the problems, maybe we should have here `CallerFromContext`? 

`authn.Caller `- answers a question: who sent the request? Was it verified.? 

`agent.Identity` - answers a question: where we are? for who we work? information about where the code is running (appName, sessionId, userId) . Always set for an invocation.

Consequently I would rename other fields/functions:


now | after rename
-- | --
type Identity struct | type Caller struct
func IdentityFromContext(ctx) (*Identity, bool) | func CallerFromContext(ctx) (*Caller, bool)
func WithIdentity(ctx, *Identity) context.Context | func WithCaller(ctx, *Caller) context.Context
type identityKey struct{} | type callerKey struct{}
Authenticate(r) (*Identity, error) | Authenticate(r) (*Caller, error)

wolo-lab (commented):
Some style and renaming open questions only.

Logic looks fine.

server/adkrest/handler.go
kdroste-google: right, moved

server/authn/authn.go
kdroste-google: Yes, I agree, both can be used elsewhere. Moved to server package

server/adkrest/handler.go
kdroste-google: Right, fixed

server/authn/authn.go
kdroste-google: Good catch, thanks!
Fixed

wolo-lab (approved):
LGTM, thank you.

PR #1563, opened 2026-09-10
fix(artifact/gcsartifact): return the gs:// form as CanonicalURI

### LINKED ISSUE #1562: GCS artifact service returns an unfetchable MediaLink as CanonicalURI
### What happened

`GetArtifactVersion` on the GCS artifact service returns the object's `MediaLink` as `CanonicalURI`, falling back to the `gs://` form only when `MediaLink` is empty. The JSON API client populates `MediaLink` on every object it returns, so callers always get the `MediaLink` and the `gs://` branch is unreachable in practice.

`MediaLink` is an authenticated JSON API download URL of the form `https://storage.googleapis.com/download/storage/v1/b/BUCKET/o/OBJECT?alt=media`. A consumer handed that URI cannot resolve it.

### Why it matters

The main use of a canonical URI is naming an artifact instead of carrying its bytes — for example passing it as the `FileURI` of a `FileData` part so a model reads the object directly. With the `MediaLink`, Vertex treats the URI as a web page to crawl and the request fails:

```
Cannot fetch content from the provided URL. ... Status: URL_ERROR-ERROR_AUTHENTICATION
```

Substituting the `gs://` form for the same object works.

### Expected

`CanonicalURI` should be the `gs://` form, as in adk-python, which builds it unconditionally (links pinned to adk-python `2284c5819338e4716bcd3f824dbd86cec3c6e685`):

- [`gcs_artifact_service.py` L550](https://github.com/google/adk-python/blob/2284c5819338e4716bcd3f824dbd86cec3c6e685/src/google/adk/artifacts/gcs_artifact_service.py#L550) — `_get_artifact_version_sync`
- [`gcs_artifact_service.py` L578](https://github.com/google/adk-python/blob/2284c5819338e4716bcd3f824dbd86cec3c6e685/src/google/adk/artifacts/gcs_artifact_service.py#L578) — `_list_artifact_versions_sync`
- [`base_artifact_service.py` L46](https://github.com/google/adk-python/blob/2284c5819338e4716bcd3f824dbd86cec3c6e685/src/google/adk/artifacts/base_artifact_service.py#L46) — "Canonical URI referencing the persisted artifact payload."

adk-python keeps authenticated forms behind separate accessors, [`get_authenticated_url` L693](https://github.com/google/adk-python/blob/2284c5819338e4716bcd3f824dbd86cec3c6e685/src/google/adk/artifacts/gcs_artifact_service.py#L693) (a `storage.cloud.google.com` console URL, not the `MediaLink`) and [`get_signed_url` L817](https://github.com/google/adk-python/blob/2284c5819338e4716bcd3f824dbd86cec3c6e685/src/google/adk/artifacts/gcs_artifact_service.py#L817). Go has neither, so a fix that returns `gs://` removes the only fetchable URL the API currently exposes; an accessor for that is worth a follow-up.

### Note

The test fake never set `MediaLink`, so the tests exercised the unreachable branch while production took the other. That is why this wasn't caught.

baptmont:
Fixes #1562.

### What

`GetArtifactVersion` preferred the object's `MediaLink` over the `gs://` URI, falling back to `gs://` only when it was empty. The JSON API client populates `MediaLink` on every object it returns, so callers always got the `MediaLink` and the fallback was unreachable in practice.

`MediaLink` is an authenticated JSON API download URL, which a consumer handed the URI cannot resolve. Passing it as the `FileURI` of a `FileData` part makes the model fetch it as a web page and fail instead of reading the object, which defeats the point of a canonical URI: naming an artifact rather than carrying its bytes.

### Parity

adk-python builds the field as `gs://{bucket}/{blob.name}` unconditionally (links pinned to adk-python `2284c5819338e4716bcd3f824dbd86cec3c6e685`):

- [`gcs_artifact_service.py` L550](https://github.com/google/adk-python/blob/2284c5819338e4716bcd3f824dbd86cec3c6e685/src/google/adk/artifacts/gcs_artifact_service.py#L550) — `_get_artifact_version_sync`
- [`gcs_artifact_service.py` L578](https://github.com/google/adk-python/blob/2284c5819338e4716bcd3f824dbd86cec3c6e685/src/google/adk/artifacts/gcs_artifact_service.py#L578) — `_list_artifact_versions_sync`
- [`base_artifact_service.py` L46](https://github.com/google/adk-python/blob/2284c5819338e4716bcd3f824dbd86cec3c6e685/src/google/adk/artifacts/base_artifact_service.py#L46) — "Canonical URI referencing the persisted artifact payload."

### Behavior change

Reading `CanonicalURI` from the GCS artifact service now yields a `gs://` URI where it previously yielded an https download URL, so code that fed it to an HTTP fetch has to change.

adk-python exposes authenticated forms through separate accessors — [`get_authenticated_url` L693](https://github.com/google/adk-python/blob/2284c5819338e4716bcd3f824dbd86cec3c6e685/src/google/adk/artifacts/gcs_artifact_service.py#L693), which returns a `storage.cloud.google.com` console URL rather than the `MediaLink`, and [`get_signed_url` L817](https://github.com/google/adk-python/blob/2284c5819338e4716bcd3f824dbd86cec3c6e685/src/google/adk/artifacts/gcs_artifact_service.py#L817). Go has neither, so this removes the only fetchable URL the API offered. Adding an accessor is worth a follow-up rather than widening this PR.

This also documents what the field means on `artifact.ArtifactVersion`, since the PR is deciding that contract and it was undocumented.

### Testing Plan

- `TestGetArtifactVersionCanonicalURI` covers the session-scoped and user-namespaced blob paths, and asserts the stored object actually has a `MediaLink` before checking the URI — without that guard the test passes against both the fixed and the broken implementation.
- The test fake now sets `MediaLink`, as the real client does. It previously did not, so the tests exercised the unreachable branch while production took the other; that is why this went unnoticed.
- Revert checks: restoring the `MediaLink` preference fails the test with the MediaLink URL in the diff; removing `MediaLink` from the fake fails the new precondition rather than passing silently.
- `go build -mod=readonly work`, `go test -race -mod=readonly -count=1 -shuffle=on work`, `go mod tidy -diff` and `gofmt` are clean. (`TestCompat_A2ACleanupPropagation` in `agent/remoteagent` flaked once under `-shuffle=on`; it passes on re-run and on an unmodified tree at the same base.)
- Live path: verified against a real Vertex `gemini-2.5-flash` endpoint reading an object from a GCS bucket. With the `MediaLink` the request fails with `URL_ERROR-ERROR_AUTHENTICATION`; with the `gs://` form the model reads the document and answers from its contents.

### Out of scope

`artifact/inmemory.go` never sets `CanonicalURI` at all, where adk-python's in-memory service populates a `memory://` URI ([`in_memory_artifact_service.py` L118](https://github.com/google/adk-python/blob/2284c5819338e4716bcd3f824dbd86cec3c6e685/src/google/adk/artifacts/in_memory_artifact_service.py#L118)). That is a separate parity gap and belongs in its own change.


wolo-lab (approved):
LGTM

PR #1564, opened 2026-09-11
fix: preserve orphan response content during rearrangement

### LINKED ISSUE #1540: Text preserved beside an orphaned function response is dropped again when a matched response follows
When an event carries one `genai.Part` holding both text and a function response whose call is absent from history, `dropOrphanedFunctionResponses` strips the function response and keeps the text. If that event is followed by an event answering an earlier call, the text is discarded a few frames later and never reaches the prompt.

`rearrangeEventsForLatestFunctionResponse` walks the events between the matched call and the trailing response, and forwards only those carrying function calls or function responses. The pruned event now carries neither — it is text-only — so it hits the `continue` at [`contents_processor.go:425-428`](https://github.com/google/adk-go/blob/ea45087add2e24c36c71af6611751b12a8ec05b7/internal/llminternal/contents_processor.go#L425-L428) and is dropped from `resultEvents`.

The user's text is lost silently. There is no diagnostic, because from the pruning stage's point of view it was preserved.

### Reproduction

Add to `internal/llminternal`, run with `go test -run TestOrphanRemnantDroppedByIntermediatePurge -v ./internal/llminternal/`:

```go
func TestOrphanRemnantDroppedByIntermediatePurge(t *testing.T) {
	const agentName = "test_agent"
	events := []*session.Event{
		{Author: agentName, LLMResponse: model.LLMResponse{Content: &genai.Content{Role: "model", Parts: []*genai.Part{
			{FunctionCall: &genai.FunctionCall{ID: "a", Name: "ta"}},
		}}}},
		// One part carrying BOTH the user's text and a response whose call is absent.
		{Author: "user", LLMResponse: model.LLMResponse{Content: &genai.Content{Role: "user", Parts: []*genai.Part{
			{Text: "note", FunctionResponse: &genai.FunctionResponse{ID: "z", Name: "tz", Response: map[string]any{"r": 1}}},
		}}}},
		{Author: "user", LLMResponse: model.LLMResponse{Content: &genai.Content{Role: "user", Parts: []*genai.Part{
			{FunctionResponse: &genai.FunctionResponse{ID: "a", Name: "ta", Response: map[string]any{"r": 1}}},
		}}}},
	}

	testAgent := utils.Must(llmagent.New(llmagent.Config{Name: agentName, Model: &testModel{}}))
	ctx := icontext.NewInvocationContext(t.Context(), icontext.InvocationContextParams{
		Agent:   testAgent,
		Session: &fakeSession{events: events},
	})

	req := &model.LLMRequest{}
	for _, err := range llminternal.ContentsRequestProcessor(ctx, req, &llminternal.Flow{}) {
		if err != nil {
			t.Fatal(err)
		}
	}

	var texts []string
	for _, c := range req.Contents {
		for _, p := range c.Parts {
			if p.Text != "" {
				texts = append(texts, p.Text)
			}
		}
	}
	if !slices.Contains(texts, "note") {
		t.Errorf("user text %q did not reach the prompt; got texts %v", "note", texts)
	}
}
```

Observed: the assembled contents are `[call:a] [resp:a]`, and `note` is absent.

### Scope

Not a regression. The behavior is identical at [f322a15f](https://github.com/google/adk-go/commit/f322a15f976f402991afd8e5d7e709aa7771ea07), which predates the combined-part handling, so nothing that works today starts failing. What changed is that the pruning stage now goes to the trouble of preserving that text, and this path undoes it, so the two stages disagree about what should survive.

Related: [#766](https://github.com/google/adk-go/pull/766) added the per-part preservation. Its new table cases all lack a trailing matched function response, so none of them enters this branch — the gap is not visible from that PR's tests.

### Constraint on any solution

The `continue` is not obviously removable. Forwarding every non-tool event through this loop changes what reaches the prompt for existing histories, so a fix needs to say why the intervening events it now keeps are ones the prompt should have had all along, and cover the case in the `TestContentsRequestProcessor_Rearrange` table.

### Acceptance

A test in the `TestContentsRequestProcessor_Rearrange` table with a combined text-plus-orphan part **followed by** a trailing matched function response, asserting the text reaches the assembled contents.

ktsoator:

Closes #1540.

When a matched tool response follows an event containing text and an orphaned response, pruning keeps the text but history rearrangement drops it again.

Track events that still have content after orphan pruning and keep them during rearrangement. Other intermediate events keep their existing behavior.

### Behavior change

Text and images left after removing an orphaned response now reach the model when a later matched response triggers rearrangement. For the issue's example, the contents change from `[call:a] [resp:a]` to `[call:a] [resp:a] [text:note]`.

### Testing Plan

Added three cases to `TestContentsRequestProcessor_Rearrange`: a combined text/response part, separate text and response parts, and text/image content around unrelated tool calls. All three fail without the fix, including `Rearrangement_preserves_text_sharing_an_orphaned_response_part`.

Forcing the new condition to always keep intermediate events also fails the existing progress-text tests.

Passed locally:

- `golangci-lint run` (v2.3.1) in both modules

Package test output:

```text
ok  google.golang.org/adk/v2/internal/llminternal  2.650s
```

No manual E2E or live-model test was run.

### Checklist

- [x] Read CONTRIBUTING.md.
- [x] Reviewed the full diff.
- [x] Added comments explaining the retention rule.
- [x] Added regression tests and confirmed they fail without the fix.
- [x] New and existing unit tests pass locally.
- [ ] Manually tested end-to-end.

### Additional context

Python's [`_tool_call_rearranger.py:388–400`](https://github.com/google/adk-python/blob/main/src/google/adk/flows/llm_flows/_tool_call_rearranger.py#L388-L400) also removes intermediate events. This fix differs by retaining content that Go's orphan pruning already preserves, as required by #1540.

### DIFF
--- a/internal/llminternal/contents_processor.go
+++ b/internal/llminternal/contents_processor.go
@@ -214,11 +214,11 @@ func buildContentsDefaultWithCallSource(agentName, invocationBranch, isolationSc
 	}
 	filtered = processedEvents
 
-	filtered = dropOrphanedFunctionResponses(filtered, allEvents)
+	filtered, orphanRemnants := dropOrphanedFunctionResponses(filtered, allEvents)
 
 	//  src/google/adk/flows/llm_flows/contents.py
 	// 	 - _rearrange_events_for_async_function_response
-	filtered, err := rearrangeEventsForLatestFunctionResponse(filtered)
+	filtered, err := rearrangeEventsForLatestFunctionResponse(filtered, orphanRemnants)
 	if err != nil {
 		return nil, err
 	}
@@ -266,7 +266,9 @@ func eventBelongsToBranch(invocationBranch string, event *session.Event) bool {
 	return utils.EventBelongsToBranch(invocationBranch, event.Branch)
 }
 
-func dropOrphanedFunctionResponses(events, allEvents []*session.Event) []*session.Event {
+// dropOrphanedFunctionResponses also identifies surviving events whose orphaned
+// responses were removed, so later rearrangement preserves their remaining content.
+func dropOrphanedFunctionResponses(events, allEvents []*session.Event) ([]*session.Event, map[*session.Event]bool) {
 	callIDs := make(map[string]struct{})
 	for _, event := range allEvents {
 		for _, call := range utils.FunctionCalls(utils.Content(event)) {
@@ -286,6 +288,7 @@ func dropOrphanedFunctionResponses(events, allEvents []*session.Event) []*sessio
 
 	var orphanedIDs []string
 	result := make([]*session.Event, 0, len(events))
+	orphanRemnants := make(map[*session.Event]bool)
 	for _, event := range events {
 		content := utils.Content(event)
 		if content == nil {
@@ -315,22 +318,21 @@ func dropOrphanedFunctionResponses(events, allEvents []*session.Event) []*sessio
 		cloned.LLMResponse.Content.Parts = parts
 		if len(cloned.LLMResponse.Content.Parts) > 0 {
 			result = append(result, cloned)
+			orphanRemnants[cloned] = true
 		}
 	}
 
 	if len(orphanedIDs) > 0 {
 		log.Printf("adk: dropping function responses with no matching function call: %q", orphanedIDs)
 	}
-	return result
+	return result, orphanRemnants
 }
 
-// rearrangeEventsForLatestFunctionResponse
-// This function only acts if the very last event is a function response.
-// It searches backward for the matching call, deletes all intervening events,
-// and appends a single (merged) response.
-// If the latest function_response is for an async function_call, all events
-// between the initial function_call and the latest function_response will be removed.
-func rearrangeEventsForLatestFunctionResponse(events []*session.Event) ([]*session.Event, error) {
+// rearrangeEventsForLatestFunctionResponse merges responses to the latest event's
+// matching call, dropping intervening non-tool events except orphanRemnants.
+// Unrelated tool events and the content left after pruning their orphaned
+// responses survive; pruning must not change whether that content is retained.
+func rearrangeEventsForLatestFunctionResponse(events []*session.Event, orphanRemnants map[*session.Event]bool) ([]*session.Event, error) {
 	if len(events) < 2 {
 		return events, nil
 	}
@@ -424,6 +426,11 @@ SearchLoop: // A label to allow breaking out of the nested loop
 
 		responses := utils.FunctionResponses(event.Content)
 		if len(responses) == 0 {
+			// Unlike Python's blanket intermediate-event removal, retain content
+			// deliberately preserved when pruning unrelated orphaned responses.
+			if orphanRemnants[event] {
+				resultEvents = append(resultEvents, event)
+			}
 			continue
 		}
 
--- a/internal/llminternal/contents_processor_test.go
+++ b/internal/llminternal/contents_processor_test.go
@@ -1451,6 +1451,62 @@ func TestContentsRequestProcessor_Rearrange(t *testing.T) {
 			},
 		},
 		{
+			name: "Rearrangement preserves text sharing an orphaned response part",
+			events: []*session.Event{
+				{Author: agentName, LLMResponse: model.LLMResponse{Content: NewContentFromFunctionCall(fcBasic, "model")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: &genai.Content{Role: "user", Parts: []*genai.Part{
+					{Text: "note", FunctionResponse: frOrphaned},
+				}}}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frBasic, "user")}},
+			},
+			want: []*genai.Content{
+				NewContentFromFunctionCall(fcBasic, "model"),
+				NewContentFromFunctionResponse(frBasic, "user"),
+				genai.NewContentFromText("note", "user"),
+			},
+		},
+		{
+			name: "Rearrangement preserves separate text beside an orphaned response",
+			events: []*session.Event{
+				{Author: agentName, LLMResponse: model.LLMResponse{Content: NewContentFromFunctionCall(fcBasic, "model")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: &genai.Content{Role: "user", Parts: []*genai.Part{
+					{Text: "note"}, {FunctionResponse: frOrphaned},
+				}}}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frBasic, "user")}},
+			},
+			want: []*genai.Content{
+				NewContentFromFunctionCall(fcBasic, "model"),
+				NewContentFromFunctionResponse(frBasic, "user"),
+				genai.NewContentFromText("note", "user"),
+			},
+		},
+		{
+			name: "Rearrangement preserves orphan remnants around unrelated tools",
+			events: []*session.Event{
+				{Author: agentName, LLMResponse: model.LLMResponse{Content: NewContentFromFunctionCall(fcLRO, "model")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frLROInter, "user")}},
+				{Author: agentName, LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("Still processing...", "model")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: &genai.Content{Role: "user", Parts: []*genai.Part{
+					{Text: "first note", FunctionResponse: frOrphaned},
+				}}}},
+				{Author: agentName, LLMResponse: model.LLMResponse{Content: NewContentFromFunctionCall(fcBasic, "model")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frBasic, "user")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: &genai.Content{Role: "user", Parts: []*genai.Part{
+					{InlineData: &genai.Blob{MIMEType: "image/png", Data: []byte("image")}, FunctionResponse: frOrphaned},
+				}}}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frOrphaned, "user")}},
+				{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frLROFinal, "user")}},
+			},
+			want: []*genai.Content{
+				NewContentFromFunctionCall(fcLRO, "model"),
+				NewContentFromFunctionResponse(frLROFinal, "user"),
+				genai.NewContentFromText("first note", "user"),
+				NewContentFromFunctionCall(fcBasic, "model"),
+				NewContentFromFunctionResponse(frBasic, "user"),
+				{Role: "user", Parts: []*genai.Part{{InlineData: &genai.Blob{MIMEType: "image/png", Data: []byte("image")}}}},
+			},
+		},
+		{
 			name: "Mid-history mixed content preserves user text",
 			events: []*session.Event{
 				{Author: "user", LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("Before", "user")}},
--- a/internal/llminternal/contents_processor_internal_test.go
+++ b/internal/llminternal/contents_processor_internal_test.go
@@ -41,7 +41,7 @@ func TestDropOrphanedFunctionResponses_LogAndClone(t *testing.T) {
 	want.LLMResponse.Content = genai.NewContentFromText("Keep this", "user")
 	want.LLMResponse.Content.Parts = append(want.LLMResponse.Content.Parts, &genai.Part{Text: "Keep this too"})
 	output := captureLog(t, func() {
-		got := dropOrphanedFunctionResponses([]*session.Event{event}, nil)
+		got, _ := dropOrphanedFunctionResponses([]*session.Event{event}, nil)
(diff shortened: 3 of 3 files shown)


internal/llminternal/contents_processor.go
```diff
@@ -424,6 +426,11 @@ SearchLoop: // A label to allow breaking out of the nested loop
 
 		responses := utils.FunctionResponses(event.Content)
 		if len(responses) == 0 {
+			// Unlike Python's blanket intermediate-event removal, retain content
+			// deliberately preserved when pruning unrelated orphaned responses.
+			if orphanRemnants[event] {
```
karolpiotrowicz, line 432: Retention here is decided by where the event came from rather than by what it now contains, and the deciding field is one the code has just classified as junk.

```
[call(a), {text "note"},                          resp(a)]  ->  [call(a) resp(a)]
[call(a), {text "note" + response with unmatched ID}, resp(a)]  ->  [call(a) resp(a) text(note)]
```

Same text, opposite outcome. Both histories run through `ContentsRequestProcessor` on this revision. It is visible inside your own third case too: `"Still processing..."` at [L1488](https://github.com/google/adk-go/blob/cbfa01722d3752d186ffe7f2fdac8aaf00d014b8/internal/llminternal/contents_processor_test.go#L1488) is dropped while `"first note"` at [L1489-L1491](https://github.com/google/adk-go/blob/cbfa01722d3752d186ffe7f2fdac8aaf00d014b8/internal/llminternal/contents_processor_test.go#L1489-L1491) is kept, and they sit in the same structural position.

One thing that surprised me and is worth having on the record: this is not restoring something pruning took away. I ran both baselines and neither delivers that text. With the response part simply absent it is dropped by this loop, and with orphan pruning disabled end to end it survives this stage as an unrelated-response event and is then skipped by `rearrangeEventsForFunctionResponsesInHistory`, which finds no call to re-pair it against. So the change adds retention that no earlier configuration produced.

That may be exactly what [#1540](https://github.com/google/adk-go/issues/1540) wants. The open question is whether the rule should key on the content that survived rather than on the pruning history, so an unmatched response ID is not what decides whether a user's text reaches the model. Either answer is fine by me, but the next person in this loop needs to know which property the loop is supposed to preserve.

internal/llminternal/contents_processor.go
```diff
@@ -315,22 +318,21 @@ func dropOrphanedFunctionResponses(events, allEvents []*session.Event) []*sessio
...
-// This function only acts if the very last event is a function response.
-// It searches backward for the matching call, deletes all intervening events,
-// and appends a single (merged) response.
-// If the latest function_response is for an async function_call, all events
-// between the initial function_call and the latest function_response will be removed.
-func rearrangeEventsForLatestFunctionResponse(events []*session.Event) ([]*session.Event, error) {
+// rearrangeEventsForLatestFunctionResponse merges responses to the latest event's
+// matching call, dropping intervening non-tool events except orphanRemnants.
+// Unrelated tool events and the content left after pruning their orphaned
+// responses survive; pruning must not change whether that content is retained.
```
karolpiotrowicz: This sentence is true of this function and not of the pipeline, and it reads as the pipeline claim.

With orphan pruning disabled end to end, the text is dropped anyway: the event keeps its `FunctionResponse`, survives this stage at the unrelated-response branch below, and then `rearrangeEventsForFunctionResponsesInHistory` skips it at [L499](https://github.com/google/adk-go/blob/cbfa01722d3752d186ffe7f2fdac8aaf00d014b8/internal/llminternal/contents_processor.go#L499) because no call event references that ID. So pruning does change whether that content is retained — from lost to retained.

Scoping the claim to this stage would make it accurate, and it is the claim you actually rely on: unpruned the event is kept here at the unrelated-response branch, pruned it is kept by the new branch above.

internal/llminternal/contents_processor_test.go
```diff
@@ -1450,6 +1450,62 @@ func TestContentsRequestProcessor_Rearrange(t *testing.T) {
...
+				{Author: "user", LLMResponse: model.LLMResponse{Content: NewContentFromFunctionResponse(frBasic, "user")}},
+			},
+			want: []*genai.Content{
+				NewContentFromFunctionCall(fcBasic, "model"),
+				NewContentFromFunctionResponse(frBasic, "user"),
+				genai.NewContentFromText("note", "user"),
+			},
+		},
+		{
+			name: "Rearrangement preserves separate text beside an orphaned response",
```
karolpiotrowicz, line 1566: This case and the one above it converge after orphan pruning. Both reduce to a single `{Text: "note"}` user part, so from the rearrangement onward they feed identical input through identical code.

Neither can separate a correct fix from a wrong one. Make the retention unconditional, or append the retained event after the merged response instead of before it, and both still pass — only `Rearrangement preserves orphan remnants around unrelated tools` catches either. I ran both variants to check that rather than reasoning about it.

They do pin that the content is retained at all: applied against the base commit, all three fail. So this is about depth, not about deleting a case. Giving this one a plain-text intermediate event alongside the remnant would give it something to discriminate, since that is the contrast the third case relies on.

karolpiotrowicz (dismissed):
The three test cases this adds are red once the branch is up to date with `main`, so it cannot land as it stands. The fix itself survives the update — only the expected orderings are stale. Before merge this needs a rebase onto `main` and the three `want` lists re-derived against the new ordering.

`main` has moved 32 commits since this branch's base. [a8c198c7](https://github.com/google/adk-go/commit/a8c198c751ab240228efdba84285b677b9fc5349) ("keep a late tool completion as the final content", [#1428](https://github.com/google/adk-go/pull/1428)) changed `rearrangeEventsForFunctionResponsesInHistory` so that the call/response pair answered by the last event is emitted last. The three `want` lists here encode the order from before that landed.

I merged this branch with `main` at [5ff82280](https://github.com/google/adk-go/commit/5ff82280edebd9c71cb4c9bcc19c570a1b2ff6b1) — the merge is clean — and ran the package. The preserved content is present in every actual result. Only its position moves: `note` lands before the call/response pair instead of after it, and in `Rearrangement preserves orphan remnants around unrelated tools` the long-running-tool pair moves to the tail.

<details><summary>Reproduction and the failure output</summary>

```
git fetch origin main
git checkout cbfa01722d3752d186ffe7f2fdac8aaf00d014b8
git merge origin/main          # clean
go test -run TestContentsRequestProcessor_Rearrange ./internal/llminternal/
```

```
--- FAIL: TestContentsRequestProcessor_Rearrange/Rearrangement_preserves_text_sharing_an_orphaned_response_part
    contents_processor_test.go:1759: LLMRequest.Contents mismatch (-want +got):
      []*genai.Content{
    + 	&{Parts: []*genai.Part{&{Text: "note"}}, Role: "user"},
      	&{Parts: {&{FunctionCall: &{ID: "call_123", ...}}}, Role: "model"},
      	&{Parts: {&{FunctionResponse: &{ID: "call_123", ...}}}, Role: "user"},
    - 	&{Parts: []*genai.Part{&{Text: "note"}}, Role: "user"},
      }
--- FAIL: TestContentsRequestProcessor_Rearrange/Rearrangement_preserves_separate_text_beside_an_orphaned_response
    (same inversion)
--- FAIL: TestContentsRequestProcessor_Rearrange/Rearrangement_preserves_orphan_remnants_around_unrelated_tools
    contents_processor_test.go:1759: LLMRequest.Contents mismatch (-want +got):
      []*genai.Content{
    - 	&{Parts: []*genai.Part{&{FunctionCall: &genai.FunctionCall{...}}}, Role: "model"},
    - 	&{Parts: []*genai.Part{&{FunctionResponse: &genai.FunctionResponse{...}}}, Role: "user"},
      	&{Parts: {&{Text: "first note"}}, Role: "user"},
      	&{Parts: {&{FunctionCall: &{ID: "call_123", ...}}}, Role: "model"},
      	&{Parts: {&{FunctionResponse: &{ID: "call_123", ...}}}, Role: "user"},
      	&{Parts: {&{InlineData: &{Data: "image", MIMEType: "image/png"}}}, Role: "user"},
    + 	&{Parts: []*genai.Part{&{FunctionCall: &genai.FunctionCall{...}}}, Role: "model"},
    + 	&{Parts: []*genai.Part{&{FunctionResponse: &genai.FunctionResponse{...}}}, Role: "user"},
      }
```

Nothing else in the package fails, and `main` alone is green.

</details>

Worth saying explicitly, since it decides whether this is still wanted: the underlying bug is still live. The reproduction from [#1540](https://github.com/google/adk-go/issues/1540) fails on `main` at [5ff82280](https://github.com/google/adk-go/commit/5ff82280edebd9c71cb4c9bcc19c570a1b2ff6b1) and passes with this branch merged into it. The change is needed, it just needs its expectations re-derived against the new ordering. The acceptance criterion in [#1540](https://github.com/google/adk-go/issues/1540) is that the text reaches the assembled contents, and it does — so the new position looks acceptable, but you are better placed than me to say whether `note` landing before the pair rather than after it is the order you want.

**A model-role remnant adds a turn nobody asked for, and nothing covers it.** All three new cases use a `user`-role remnant ([contents_processor_test.go#L1457](https://github.com/google/adk-go/blob/cbfa01722d3752d186ffe7f2fdac8aaf00d014b8/internal/llminternal/contents_processor_test.go#L1457), [#L1472](https://github.com/google/adk-go/blob/cbfa01722d3752d186ffe7f2fdac8aaf00d014b8/internal/llminternal/contents_processor_test.go#L1472), [#L1489](https://github.com/google/adk-go/blob/cbfa01722d3752d186ffe7f2fdac8aaf00d014b8/internal/llminternal/contents_processor_test.go#L1489)). When the retained event carries `Role: "model"` the assembled history ends on a model turn, which trips the continuation nudge at [contents_processor.go#L107-L111](https://github.com/google/adk-go/blob/cbfa01722d3752d186ffe7f2fdac8aaf00d014b8/internal/llminternal/contents_processor.go#L107-L111), so one unmatched response ID adds two contents rather than one:

```
plain       -> [model:call(a) user:resp(a)]
with orphan -> [model:call(a) user:resp(a) model:text(thinking)
                user:text(Continue processing previous requests as instructed. ...)]
```

I built that history by hand and could not find an in-tree producer for a model-role event carrying a `FunctionResponse`, so I cannot tell you how reachable it is. A case either way would pin the behaviour.

Everything else I found is on the lines themselves.

internal/llminternal/contents_processor.go
ktsoator: Updated in 513cdcd. I've kept this scoped to content left by orphan pruning and documented that choice. Preserving ordinary intermediate text would be a broader behavior change.

internal/llminternal/contents_processor.go
ktsoator: Reworded in 513cdcd to describe the retention exception directly, without implying that the old pipeline preserved this content.

internal/llminternal/contents_processor_test.go
ktsoator: Added ordinary user text alongside the remnant in 513cdcd. Confirmed this case now fails if retention is made unconditional.

ktsoator:
Hi @karolpiotrowicz, thanks for the review. I've updated the ordering expectations, clarified the retention rule, and added plain-text and model-role coverage.

The targeted tests and checks pass; the full local suite is blocked by Web test hangs also reproduced without this fix.

Does keeping retention scoped to orphan remnants address your concern?

internal/llminternal/contents_processor_test.go
```diff
@@ -1547,6 +1547,82 @@ func TestContentsRequestProcessor_Rearrange(t *testing.T) {
...
+				NewContentFromFunctionResponse(frBasic, "user"),
+			},
+		},
+		{
+			name: "Rearrangement preserves separate text beside an orphaned response",
+			events: []*session.Event{
+				{Author: agentName, LLMResponse: model.LLMResponse{Content: NewContentFromFunctionCall(fcBasic, "model")}},
+				// The same text without an orphaned response still follows the
+				// existing intermediate-event filtering rule.
+				{Author: "user", LLMResponse: model.LLMResponse{Content: genai.NewContentFromText("note", "user")}},
```
karolpiotrowicz: This event and the remnant below it both carry `note`, so the single `note` in `want` is produced whichever of the two survives. The row can't show what its comment says. Inverting the check to keep plain text and drop remnants passes this row, though seven other rows catch it.

Giving this event different text makes the row discriminate on its own. I tried `"plain note"` here: the row still passes on this branch and fails against the inverted check.

karolpiotrowicz (approved):
Nothing blocks now. This pass covers what changed since my last review: the expectations re-derived against `main`, the reworded comment and the two new rows.

The three cases that were red on the merge with `main` pass on [513cdcd8](https://github.com/google/adk-go/commit/513cdcd8fff41dcf01c52f2c1ed2ffbfb1ec3b26), and the build and full suite are clean. The reproduction from [#1540](https://github.com/google/adk-go/issues/1540) still fails on `main` ([f9ce16ef](https://github.com/google/adk-go/commit/f9ce16ef9cb334b69f5ee3e64e9dfc915f02d0b7)) and passes with this branch, so the fix still does its job after the update.

On your question: yes, keeping retention scoped to orphan remnants addresses it. What I was after was the rule being written down, so the next change to that loop knows what it has to preserve, and the new comment does that. I ran it to check the comment against the code. Ordinary text between the call and the response is still dropped, and the same text beside an orphaned response now reaches the prompt. The model-role row's comment holds as well — the pair stays last and no continuation turn is added — and the separate-text row does fail when retention is made unconditional, as you said.

Two small things, neither blocking:

- The comment on the `Late async completion stays the final content` row at [contents_processor_test.go#L1288-L1294](https://github.com/google/adk-go/blob/513cdcd8fff41dcf01c52f2c1ed2ffbfb1ec3b26/internal/llminternal/contents_processor_test.go#L1288-L1294) says the preservation loop in `rearrangeEventsForLatestFunctionResponse` "keeps only events carrying calls or responses". This change made that untrue, since the loop now also keeps orphan remnants. It is still right about that row's three text turns, so a short qualifier would fix it.
- One note on the separate-text row, on its line.

PR #1567, opened 2026-09-11
test(examples): cover image precedence and empty results

ktsoator:

Related: #1477. Follow-up to #1478.

Add two `ImageBytes` test cases suggested during review: a non-nil empty image list, and an entry containing both usable image bytes and a filtering reason. Also clarify how nil responses and zero-length image data are handled.

### Behavior change

Nothing. This PR adds tests and documents existing behavior.

### Testing Plan

All tests passed, including the full race test suite. Build, lint (v2.3.1), tidy, and formatting checks passed in both modules.

None; this PR does not change runtime code. I checked the new tests against two temporary mutations:

- Using a nil check instead of a length check fails `TestImageBytes/empty_generated_images`.
- Rejecting usable bytes when the same entry has a filtering reason fails `TestImageBytes/usable_image_with_a_filtering_reason`.

Not applicable; no runtime behavior changes.

### Checklist

- [x] I have added tests that prove the documented behavior.
- [ ] I have manually tested my changes end-to-end. (Not applicable.)
- [ ] Any dependent changes have been merged and published in downstream modules. (Not applicable; no downstream module changes are required.)

### DIFF
--- a/examples/internal/imagegen/imagegen.go
+++ b/examples/internal/imagegen/imagegen.go
@@ -24,15 +24,16 @@ import (
 )
 
 // ImageBytes returns the bytes and MIME type of the first usable image in a
-// generation response. If the response contains no generated images, it returns
-// an error reporting that no images were returned. If the entries contain no
-// usable image data but include RAI filtering reasons, the error includes the
-// first reason that remains non-empty after trimming surrounding whitespace.
+// generation response. A nil response is accepted and returns the same error as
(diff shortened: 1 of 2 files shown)


ktsoator:
@karolpiotrowicz I’ve added the test cases and doc updates you suggested in your final review of #1478. Thanks again for the suggestions — would you mind taking a look when you have time?

PR #1571, opened 2026-09-11
Align vertexai session service via ADK REST API with other impl

kdroste-google:
**Problem:**

The following route
```go
		Route{
			Name:        "GetSession",
			Methods:     []string{http.MethodGet},
			Pattern:     "/apps/{app_name}/users/{user_id}/sessions/{session_id}",
			HandlerFunc: r.sessionController.GetSessionHandler,
		},

```

doesn't behave the same for different implementations. 

For VertexAI Session Service the result of getting the session belonging to another user results in HTTP 500. For other implementations: HTTP 404. 

**Solution:**
Aligning the vertexai implementation.


wolo-lab (approved):
LGTM, thanks!

PR #1574, opened 2026-09-12
fix(loadartifactstool): stop double-wrapping the artifact load error

rishabhsai:

- Related: #866

**Problem:**

`loadIndividualArtifact` already wraps the service error with the artifact name (`tool/loadartifactstool/load_artifacts_tool.go:230`), and its only caller wraps the result again with the same format string (`:212`), so the phrase and the filename reach callers and logs twice: `failed to load artifact report.pdf: failed to load artifact report.pdf: artifact not found: file does not exist`.

**Solution:**

Drop the outer wrap and return the error as it comes back, leaving `failed to load artifact report.pdf: artifact not found: file does not exist`. Either wrap could go, since the inner one is the only caller of `loadIndividualArtifact`; keeping the inner one leaves the message identical for the instruction-loading path in `internal/llminternal/instruction_processor.go:141`, which already reads the same way. #866 is open against this file for unsupported MIME types and its hunks sit at the import block and at `loadIndividualArtifact`'s return, so it neither makes nor conflicts with this change.

### Behavior change

A caller that reads the error text from a failed `load_artifacts` request sees `failed to load artifact <name>` once instead of twice. The wrapped cause is unchanged, so `errors.Is` and `errors.As` behave as before.

### Testing Plan

**Unit Tests:**

`go test -race -count=1 ./tool/loadartifactstool/` → `ok google.golang.org/adk/v2/tool/loadartifactstool 1.339s`. The full module loop from `AGENTS.md` is green in both modules: 88 packages ok, `golangci-lint run` (v2.3.1) 0 issues, `go mod tidy -diff` clean.

`TestLoadArtifactsTool_ProcessRequest_LoadError`, both cases: `error names the failure 2 times, want once: failed to load artifact missing.txt: failed to load artifact missing.txt: artifact not found: file does not exist`.

Covered by the new test, which drives `ProcessRequest` with a real in-memory artifact service and a `load_artifacts` function response naming an artifact that was never saved. No model or network call is involved in this path.

### Checklist

### DIFF
--- a/tool/loadartifactstool/load_artifacts_tool.go
+++ b/tool/loadartifactstool/load_artifacts_tool.go
@@ -209,7 +209,7 @@ func (t *artifactsTool) processLoadArtifactsFunctionCall(ctx agent.Context, req 
 			// Although not used, we need to pass childCtx for early return in case of an error.
 			content, err := t.loadIndividualArtifact(childCtx, artifactsService, artifactName)
 			if err != nil {
-				return fmt.Errorf("failed to load artifact %s: %w", artifactName, err)
+				return err
 			}
 			results[i] = content
 			return nil
--- a/tool/loadartifactstool/load_artifacts_tool_test.go
+++ b/tool/loadartifactstool/load_artifacts_tool_test.go
@@ -15,6 +15,8 @@
 package loadartifactstool_test
 
 import (
+	"errors"
+	"io/fs"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


baptmont (approved):
Approving.

Verified the duplication rather than taking it on trust: the format string at `:213` and the one inside `loadIndividualArtifact` at `:230` are identical, and `loadIndividualArtifact` has exactly one caller, so the outer wrap contributes nothing but a second copy of the name. Reverting your change and keeping the tests reproduces it in both subtests — `failed to load artifact missing.txt: failed to load artifact missing.txt: artifact not found: file does not exist`.

Good call dropping the outer wrap rather than the inner one. Removing the `%w` would have broken unwrapping for callers, and the assertion on `errors.Is(err, fs.ErrNotExist)` in the new test is what makes the chain safe for whoever touches this next — that pin is the part I'd have asked for if it weren't already there.

Full gate green on the module, lint 0 issues.

PR #1586, opened 2026-09-12
fix(functiontool): wrap non-map results only when they can be marshaled

### LINKED ISSUE #1585: tool/functiontool: raw-value validation rejects marshalable results and accepts unrepresentable ones
## Description

`functionTool.Run` converts a tool result into `map[string]any`, and falls back to wrapping a non-map result into `{"result": v}` when that conversion fails, at `tool/functiontool/function.go:231-246`:

```go
	resp, err := typeutil.ConvertToWithJSONSchema[TResults, map[string]any](output, f.outputSchema)
	if err == nil { // all good
		return resp, nil
	}

	// Specs requires the result to be a map (dict in python). python impl allows basic types when building response event
	// functions.py __build_response_event does the following
	// if not isinstance(function_result, dict):
	// 		function_result = {'result': function_result}
	if f.outputSchema != nil {
		if err1 := f.outputSchema.Validate(output); err1 != nil {
			return resp, err // if it fails propagate original err.
		}
	}
	wrappedOutput := map[string]any{"result": output}
	return wrappedOutput, nil
```

The guard at :241 decides between wrapping the result and propagating the original error. What it validates is the raw Go value. `Resolved.Validate` walks that value through `reflect.ValueOf` and never marshals it: a `reflect.Struct` is refused outright at `jsonschema/validate.go:449-451`, and at `jsonschema/util.go:262-266` a float is sorted into `integer` or `number` by whether its fractional part is zero, so NaN and ±Inf pass as `number`. For a schema inferred from `TResults`, what the guard detects is a Go struct on the value's path, which is unrelated to whether the result can be sent.

The guard is wrong in both directions.

**Valid results are rejected.** A tool returning `[]SomeStruct` produces the shape the fallback exists for, yet `Run` returns `json: cannot unmarshal array into Go value of type map[string]interface {}`. At `internal/llminternal/base_flow.go:1416` that error is written as `{"error": ...}`, so the model receives it instead of the data. `time.Time` is rejected the same way, since it is a named struct. The same tool returning `[]float64` is wrapped into `{"result": [...]}` as expected.

**Unrepresentable results are let through.** When a tool returns `map[string]any` with one value set to `math.NaN()`, the guard lets it through and `Run` returns the wrapped result with a nil error. `google.golang.org/genai` then calls `InternalDeepMarshal` at `models.go:4472` without receiving the error it returns, leaving `parameterMap` empty, so `InternalFormatMap` at `models.go:4506` renders the URL as `:generateContent`, the request body is empty, and the run ends in a 404 whose `Message` is empty. Also measured as let through: `float64` returning +Inf or -Inf, `float32` returning NaN, and `map[string]any` holding a chan or a func.

## Suggested fix

Decide the fallback on whether `encoding/json` can represent the result rather than on schema validation: marshal once, return that error when it fails, and wrap otherwise.

Enforce only a schema the caller supplied through `Config.OutputSchema`. The inferred schema describes `TResults`, not what `encoding/json` emits for it, so validating the JSON form against it would start rejecting `[]byte`, `json.RawMessage` and a `*SomeStruct` returning nil, all three of which work on main.

## Environment

The first direction reproduces without a model call: build a tool returning `[]SomeStruct` inside the `tool/functiontool` package and call `Run` to get the error above. Both directions were run against Gemini API and Vertex AI, with the same behavior.

- ADK Library Version: main @ `11521a4`
- OS: Windows
- Go Version: 1.26.6
- Model Information: `gemini-3.8-flash`, Gemini API and Vertex AI

Trainingcqy:

- Closes: #1585

**Problem:**

The guard at `function.go:241` decides whether a result that will not convert to `map[string]any` is wrapped or reported as an error. What it validates is the raw Go value: `Resolved.Validate` refuses a `reflect.Struct` outright at `jsonschema/validate.go:449-451`, sorts floats by fractional part at `jsonschema/util.go:262-266`, and never marshals. Valid results such as `[]SomeStruct` and `time.Time` are therefore reported as failures, while `map[string]any` holding `math.NaN()` is wrapped and returned with a nil error, after which genai discards the marshal error at `models.go:4472` and the call ends in a 404 with an empty body.

**Solution:**

Marshal once before the fallback and return that error when it fails, so a result is wrapped only when it can be sent. Schema validation applies only to a schema supplied through `Config.OutputSchema`.

The inferred schema describes `TResults` rather than what `encoding/json` emits for it, so moving the validation to the JSON form without taking it out of this decision would reject `[]byte`, which emits a base64 string against an array schema, `json.RawMessage`, and a `*SomeStruct` returning nil, which normalizes to `{}` and then hits required. All three work on main and remain supported. `ConvertToWithJSONSchema` is untouched; it has 10 non-test call sites in the repo.

**Behavior change:** three kinds of result that were reported as failures are now wrapped: a non-map result with a Go struct on its path such as `[]SomeStruct`; a named struct that marshals to something other than an object such as `time.Time`; and a named struct whose `MarshalJSON` emits keys the inferred schema does not describe. A result `encoding/json` cannot represent now returns that marshal error instead of being wrapped, and `callTool` turns that error into `{"error": ...}` at `base_flow.go:1416` rather than failing the run. Everything else is unchanged.

**Parity with adk-python:** `_normalize_tool_result` at `flows/llm_flows/_tool_caller.py:95-99` wraps anything that is not a dict and has no representability check, and Python function tools have no notion of an output schema. The added marshal is a deliberate divergence, since a result rejected by `encoding/json` cannot be sent even once wrapped; the reason is written in the comment on the change.

### Testing Plan

**Unit Tests:**

- [x] Tests for the changed packages pass locally.

`TestFunctionTool_WrapsNonMapResult` covers eight cases: five results that were wrongly rejected, plus `[]byte`, `json.RawMessage` and a `*player` returning nil, which already worked and must keep working. `TestFunctionTool_ResultNotRepresentable` covers NaN, +Inf and a chan. `TestFunctionTool_OutputSchemaMismatch` keeps `Config.OutputSchema` enforced and pins the error it returns.

Eight subtests fail: the three rows of `TestFunctionTool_ResultNotRepresentable`, where `Run` returns a nil error together with a wrapped `map[result:map[score:NaN]]`, `map[result:+Inf]` and `map[result:map[ch:0x...]]`, and the first five rows of `TestFunctionTool_WrapsNonMapResult`, which report `Run returned error ..., want nil`. The remaining three rows and `TestFunctionTool_OutputSchemaMismatch` pass on main and are there as regression guards.

`go test -count=1 -shuffle=on` covers `./tool/...`, `./workflow/...`, `./internal/...`, `./agent/...` and `./model/...`. Two packages are red for reasons unrelated to this change, and are red on unmodified main as well. `internal/telemetry/functionaltest` is red every time: sibling spans sharing a start time are ordered by name at `internal/telemetry/telemetrytest/span_digest.go:134-138`, and this machine's clock granularity puts both cases inside one tick. `TestCompactionE2E` in `agent/llmagent` is red intermittently, 3 times out of 8 on unmodified main and 4 times out of 8 on this branch. `golangci-lint run` and `go mod tidy -diff` are clean for both modules, and `go build ./...` passes.

Both backends, Gemini API and Vertex AI, were run once before the fix and once after, on `a59464b`, with `gemini-3.8-flash` and `session.InMemoryService` for the session.

For the valid result that was wrongly rejected, the tool returns `[]player` with two elements scoring 5.4186 and 8.65. Before the fix the model receives the tool error and cannot answer with the roster; after it, both runs receive the data and answer from it.

<details>
<summary>Log</summary>

```
===== gemini, base a59464b =====
backend: BackendGeminiAPI
tool call:   roster_tool map[team:Blue]
tool result: map[error:json: cannot unmarshal array into Go value of type map[string]interface {}]
model: I am unable to list the players and scores for team Blue because the roster tool encountered an error while retrieving the data.
===== gemini, fixed =====
backend: BackendGeminiAPI
tool call:   roster_tool map[team:Blue]
tool result: map[result:[{THY 5.4186} {Bo 8.65}]]
model: Team Blue consists of THY with a score of 5.4186 and Bo with a score of 8.65.
===== vertex, base a59464b =====
backend: BackendVertexAI
tool call:   roster_tool map[team:Blue]
tool result: map[error:json: cannot unmarshal array into Go value of type map[string]interface {}]
tool call:   roster_tool map[team:blue]
tool result: map[error:json: cannot unmarshal array into Go value of type map[string]interface {}]
model: Unable to retrieve the players and scores for team Blue due to a tool error.
===== vertex, fixed =====
backend: BackendVertexAI
tool call:   roster_tool map[team:Blue]
tool result: map[result:[{THY 5.4186} {Bo 8.65}]]
model: Team Blue consists of THY with a score of 5.4186 and Bo with a score of 8.65.
```

</details>

For the unrepresentable result that was wrongly let through, the tool returns `map[string]any{"score": math.NaN()}`. Before the fix both runs end in a 404 whose `Message` is empty and break off; after it, the marshal error reaches the model as `{"error": ...}` and both runs go through.

<details>
<summary>Log</summary>

```
===== gemini, base a59464b =====
backend: BackendGeminiAPI
tool call:   score_tool map[team:Blue]
tool result: map[result:map[score:NaN]]
run error: failed to call model: Error 404, Message: , Status: 404 Not Found, Details: []
===== gemini, fixed =====
backend: BackendGeminiAPI
tool call:   score_tool map[team:Blue]
tool result: map[error:json: unsupported value: NaN]
model: The score for team Blue could not be retrieved because the system returned an invalid value (NaN).
===== vertex, base a59464b =====
backend: BackendVertexAI
tool call:   score_tool map[team:Blue]
tool result: map[result:map[score:NaN]]
run error: failed to call model: Error 404, Message: , Status: 404 Not Found, Details: []
===== vertex, fixed =====
backend: BackendVertexAI
tool call:   score_tool map[team:Blue]
tool result: map[error:json: unsupported value: NaN]
model: The score for team Blue could not be retrieved due to an error.
```

</details>

### Checklist

- [x] New and existing unit tests pass locally with my changes. See Testing Plan.

### DIFF
--- a/tool/functiontool/function.go
+++ b/tool/functiontool/function.go
@@ -16,6 +16,7 @@
 package functiontool
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
@@ -237,8 +238,15 @@ func (f *functionTool[TArgs, TResults]) Run(ctx agent.Context, args any) (result
 	// functions.py __build_response_event does the following
 	// if not isinstance(function_result, dict):
 	// 		function_result = {'result': function_result}
-	if f.outputSchema != nil {
-		if err1 := f.outputSchema.Validate(output); err1 != nil {
+	// Diverges from the python impl above: a result encoding/json cannot represent
+	// must not be wrapped, because it cannot be sent.
+	if _, mErr := json.Marshal(output); mErr != nil {
+		return nil, mErr
+	}
+	// A schema inferred from TResults describes the type, not what encoding/json
+	// emits for it.
+	if f.cfg.OutputSchema != nil {
+		if err1 := typeutil.ValidateWithJSONSchema(output, f.outputSchema); err1 != nil {
 			return resp, err // if it fails propagate original err.
 		}
 	}
--- a/tool/functiontool/function_test.go
+++ b/tool/functiontool/function_test.go
@@ -19,9 +19,11 @@ import (
 	"errors"
 	"fmt"
 	"iter"
+	"math"
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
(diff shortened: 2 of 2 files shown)


indurireddy-TF:
Hi @Trainingcqy Thanks for this PR!
Tested and verified the changes locally:
**TestFunctionTool_ResultNotRepresentable** passes (correctly catches and returns errors for NaN, +Inf, and channels).
**TestFunctionTool_WrapsNonMapResult** passes (correctly wraps slices of structs and custom types into {"result": ...}).
Ran the full package test suite (go test ./tool/functiontool/...) and all tests passed with no regressions.
The fix is working as expected

image

baptmont (approved):
Approving.

I reproduced both halves on unmodified `main` with a throwaway tool, rather than reading the argument. A handler returning `[]player` fails with `cannot unmarshal array into Go value of type map[string]interface {}`, `time.Time` fails with the string equivalent, and a `map` holding `math.NaN()` comes back wrapped with a nil error — success, right up until genai drops the marshal error and the call dies as an empty-bodied 404. On this branch all three do the right thing and `[]byte` still wraps as before.

The diagnosis holds too: the gate was validating the raw Go value against a schema inferred from `TResults`, which is neither what `encoding/json` emits nor a contract the caller asked for. Keying it on `cfg.OutputSchema` so only a user-supplied schema is enforced is the right shape, and it is a smaller change than moving the validation to the JSON form would have been.

On parity, I checked all four ports rather than just Python. adk-python's `_normalize_tool_result` wraps non-dicts and stops (`flows/llm_flows/tools/_caller.py:96-100`), adk-java does `ImmutableMap.of("result", value)` (`FunctionTool.java:304-305`), adk-js validates arguments but not results, and adk-kotlin has no result-schema validation at all. None of them infers an output schema from the return type, so none of them has this bug — it is Go-only, created by the generics-driven inference the other ports have no equivalent of. That makes the added `json.Marshal` a real divergence rather than a port, and the comment on it says so.

Full gate green on both modules merged with current `main`. Needs a merge before it lands (`behind`, and CI here is from 12 September).

PR #1228, opened 2026-07-31
fix(deploy): increase timeout for Agent Engine deployment long-running operations

anishesg:
## Link to Issue

Closes: #850

## Description of Change

The `adkgo deploy agentengine` command was reporting a DeadlineExceeded error despite the deployment succeeding on the backend. The root cause was that `op.Wait(ctx)` in both `gcloudDeployToAgentEngine()` and `gcloudUpdateAgentEngine()` functions used a context without an explicit timeout, causing the client to time out with the default gRPC deadline before the long-running Agent Engine deployment operation could complete.

Agent Engine deployments involve container building and deployment which typically take 15-30 minutes. The fix adds a 45-minute timeout context specifically for the `op.Wait()` call in both create and update operations, allowing sufficient time for the deployment to complete while still providing a reasonable upper bound.

### Unit Tests
- [ ] I have added or updated unit tests for my change
- [ ] All unit tests pass locally

No new unit tests added as this is a timeout configuration change. The existing logic flow remains unchanged.

### Manual End-to-End (E2E) Tests

**Setup:**
1. Configure GCP project and Agent Engine credentials
2. Prepare a sample ADK application for deployment

**Test execution:**
```bash
adkgo deploy agentengine -p <project_name> -r <region> -e <entry_point>
```

**Expected result:**
- Deployment should complete without DeadlineExceeded error
- Success message should display the deployed Reasoning Engine name

## Checklist

- [x] I have read CONTRIBUTING.md
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (if applicable)
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have manually tested my changes end-to-end
- [x] Any dependent changes have been merged and published in downstream modules

Fixes #850

PR #1590, opened 2026-09-13
fix(adkgo/deploy): use a long timeout for Agent Engine op.Wait

### LINKED ISSUE #850: Agent Engine deployment succedes but with DeadlineExceeded
Using `adkgo deploy agentengine` you get DeadlineExceeded but the actual deployment is successful

nuthalapativarun:
**Please ensure you have read the [contribution guide](https://github.com/google/adk-go/blob/main/CONTRIBUTING.md) before creating a pull request.**

- Closes: #850

### Behavior change

**What behaves differently for someone already on the current release?**

`adkgo deploy agentengine` (both the create and update path) will now wait
up to 45 minutes for the `CreateReasoningEngine`/`UpdateReasoningEngine`
long-running operation to complete before reporting an error, instead of
inheriting the GAPIC client's default per-RPC timeout. Previously the CLI
could report a `DeadlineExceeded` failure while the deployment was actually
still in progress (and would in fact succeed on the backend). Deployments
that genuinely hang will now take longer to time out (up to 45 minutes
instead of the previous short default), but the reported result will
accurately reflect the backend outcome.

### Testing Plan

**Unit Tests:**

`gcloudDeployToAgentEngine`/`gcloudUpdateAgentEngine` call directly into
the real `aiplatform` GAPIC client with no seam for injecting a fake LRO
poller, so this change (like the prior attempt in #1228) does not add a
new unit test — it's a timeout-configuration change to an already-untested
code path. Existing tests in this package
(`go test ./cmd/adkgo/internal/deploy/agentengine/...`) pass with this
change; the two pre-existing failures on this Windows checkout
(`TestComputeFlags_AcceptsBenignValues`,
`TestPrepareDockerfile_EntryPointReachesRUNLine`) are unrelated
`\` vs `/` path-separator issues in `computeFlags`'s Dockerfile-content
validation, reproduce identically on `main` with this change reverted,
and are not touched by this PR.

None — no test currently exercises `op.Wait`'s context, so this PR does
not add one (see above). This mirrors the prior fix attempt (#1228),
which made the same call.

Not run — requires a live GCP project with Agent Engine access and a real
long-running deployment (15-30 minutes), which I don't have credentials
for. The root cause was verified by code inspection: `ctx :=
context.Background()` at the top of both functions is passed unmodified
into `op.Wait(ctx)`, so the poll loop is bound only by the GAPIC client's
own default per-RPC deadline rather than the actual deployment duration.

### Checklist

### Additional context

A prior attempt at this fix, #1228, diagnosed the same root cause and
proposed the same shape of fix (an explicit long timeout scoped to
`op.Wait`), but was closed for going idle rather than on the merits
("going to close this one since it's been sitting idle. reopen
whenever."). This PR picks the same approach back up, scoping the
timeout to a package-level `agentEngineOperationTimeout` constant used by
both the create and update paths.


kdroste-google (approved):
Looks good, thank you!

PR #1591, opened 2026-09-14
fix(configurable): require approval for MCP subprocesses

### LINKED ISSUE #1569: McpToolset config passes command and args to exec.Command without validation
The `McpToolset` factory in `internal/configurable` takes the `command` and `args` values out of an agent's YAML config and passes them to `exec.Command` with no validation of any kind. Loading a config that names `/bin/sh` with `["-c", "…"]` is accepted, and the shell runs with the privileges of the host process as soon as the toolset is used.

The values are read at [`configurable_utils.go#L209-L216`](https://github.com/google/adk-go/blob/11521a4ade8a2984d8bab812ed9e8af0d9fc9752/internal/configurable/configurable_utils.go#L209-L216) and reach `exec.Command` at [`configurable_utils.go#L238-L243`](https://github.com/google/adk-go/blob/11521a4ade8a2984d8bab812ed9e8af0d9fc9752/internal/configurable/configurable_utils.go#L238-L243). There is no allowlist, no path resolution and no argument filtering anywhere between the two.

## When the process actually starts

Building the toolset does not spawn anything. [`mcptoolset.New`](https://github.com/google/adk-go/blob/11521a4ade8a2984d8bab812ed9e8af0d9fc9752/tool/mcptoolset/set.go#L51-L62) only constructs the transport. The command runs on the first [`Tools()`](https://github.com/google/adk-go/blob/11521a4ade8a2984d8bab812ed9e8af0d9fc9752/tool/mcptoolset/set.go#L164-L166) call, which is what an agent does when it runs. So merely loading a config is not enough — the agent has to be used.

## How far it reaches today

`internal/configurable` is imported only by `cmd/internal/adkcli` and by packages inside `internal/configurable` itself. The shipped `cmd/adkgo` CLI does not reach this path at all, so no released binary exposes it to an end user today.

Within the repository, `cmd/internal/adkcli` is a `package main` that recursively walks the working directory for every file named `root_agent.yaml` ([`main.go#L61-L74`](https://github.com/google/adk-go/blob/11521a4ade8a2984d8bab812ed9e8af0d9fc9752/cmd/internal/adkcli/main.go#L61-L74)) and loads each one it finds ([`main.go#L92`](https://github.com/google/adk-go/blob/11521a4ade8a2984d8bab812ed9e8af0d9fc9752/cmd/internal/adkcli/main.go#L92)). It loads the agents but does not run them, so on its own it does not spawn the command. Anything built on the config loader that also runs the agent does.

That containment is the reason this reads as hardening rather than an incident, and it is also the reason it is worth fixing before the config loader gains a public entry point.

## Reproduction

Drop this into `internal/configurable/` and run `go test -count=1 -run TestMcpToolsetSpawnsArbitraryCommand -v ./internal/configurable`. It passes on `11521a4a`, meaning the command ran.

<details>
<summary>zz_repro_test.go</summary>

```go
package configurable

import (
	"context"
	"os"
	"path/filepath"
	"testing"
	"time"

)

type reproCtx struct{ agent.StrictContextMock }

func TestMcpToolsetSpawnsArbitraryCommand(t *testing.T) {
	marker := filepath.Join(t.TempDir(), "pwned")

	// Exactly the shape internal/configurable decodes out of root_agent.yaml.
	args := map[string]any{
		"stdio_connection_params": map[string]any{
			"server_params": map[string]any{
				"command": "/bin/sh",
				"args":    []any{"-c", "touch " + marker + "; sleep 30"},
			},
		},
		"tool_filter": []any{"anything"},
	}

	_, toolset, err := ResolveToolReference(context.Background(), "McpToolset", args)
	if err != nil {
		t.Fatalf("ResolveToolReference() error = %v, want nil (config was accepted)", err)
	}
	if _, err := os.Stat(marker); err == nil {
		t.Fatal("marker exists before the toolset was used")
	}

	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()
	// Tools() is what an agent calls when it actually runs.
	_, _ = toolset.Tools(&reproCtx{agent.NewStrictContextMock(ctx)})

	if _, err := os.Stat(marker); err != nil {
		t.Fatalf("marker %q was not created: %v", marker, err)
	}
	t.Logf("PWNED: %q created by a command named only in the config", marker)
}
```

</details>

```
=== RUN   TestMcpToolsetSpawnsArbitraryCommand
    zz_repro_test.go:48: PWNED: ".../001/pwned" created by a command named only in the config
--- PASS: TestMcpToolsetSpawnsArbitraryCommand (10.01s)
```

## Constraints on any solution

Two facts about the tree, both learned the hard way in #923, which attempted this and is being closed unmerged.

An opt-in switch placed under `internal/` cannot be turned on by anyone outside the module. `#923` added `SetGlobalMCPPolicy` in `internal/configurable` and pointed operators at it in an error message, which for an external consumer produces `use of internal package … not allowed` from the compiler. Whatever carries the operator's intent has to live somewhere a consumer can reach, or arrive through the config itself.

`TestResolveToolReferenceMcpToolsetNonStringArgs` currently encodes the pre-fix contract: its `all_strings_is_valid` case asserts that an `McpToolset` config with `command: "echo"` resolves without error. Any change that refuses unapproved commands by default makes that case fail, so it has to be revisited deliberately rather than treated as a merge artifact.

## Acceptance criteria

- A config naming a command the operator has not approved is refused, and the refusal is asserted through `ResolveToolReference` rather than only against an internal helper. Deleting the guard must turn that test red.
- Whatever mechanism grants approval is reachable from outside the module, or the design note says explicitly that the config loader is not supported outside the repository.
- The allowed set can express a command whose arguments are fully pinned, not only a prefix — otherwise approving a launcher such as `npx` or `docker` re-opens the hole through its own arguments.
- `go test ./internal/configurable/` is green on the merge result with `main`.

ktsoator:

Closes #1569.

A YAML config can currently start any command through `McpToolset` when the agent first lists its tools.

Require explicit approval for these commands. The internal CLI now accepts `--mcp-policy <file>`, with a list of executable paths and exact argument lists. An empty argument list permits only a command with no arguments.

Commands run using the path in the policy so Python virtual environments keep working. The target is checked again before startup. Agent caches are also separated by policy to prevent a previously approved agent from bypassing a later load's checks.

### Behavior change

Existing stdio MCP configs need a policy. Without one, the internal CLI logs a loading error and skips the agent. An invalid policy file stops startup.

This applies to the internal YAML loader. Direct use of `mcptoolset.New` is unchanged. The policy does not sandbox approved programs.

### Testing Plan

Passed with Go 1.26.6 against `main` at `6476af6`:

- `golangci-lint run` (v2.3.1) in both modules
- `go mod tidy -diff` in both modules; no diff

The tests exercise approval through `ResolveToolReference`, including changed arguments, null policy arguments, symlinks, and cached agents. They also check that policy errors report argument indices or JSON syntax offsets without exposing config contents, and preserve their causes. Removing the factory check was verified to fail `TestResolveToolReferenceMCPPolicy`.

Manual check (Python 3, macOS): a symlink-based venv retained its `sys.prefix` when started through the config factory. This used a temporary test, not a test included in the branch. No model or remote MCP calls were made. The child exits without an MCP handshake; this checks process startup and venv selection only.

To reproduce:

1. Run `python3 -m venv --symlinks --without-pip /tmp/adk-mcp-review-venv` using a fresh directory.
2. Add the temporary test below to `internal/configurable/mcp_policy_test.go`, then run `go test ./internal/configurable -run '^TestManualMCPVenv$' -count=1 -v`.
3. Remove the temporary test and venv after the check.

<details>
<summary>Temporary test for the manual check</summary>

```go
func TestManualMCPVenv(t *testing.T) {
    venv, err := filepath.EvalSymlinks("/tmp/adk-mcp-review-venv")
    if err != nil { t.Fatal(err) }
    executable := filepath.Join(venv, "bin", "python3")
    marker := filepath.Join(t.TempDir(), "prefix")
    args := []string{"-c", "import pathlib, sys; pathlib.Path(sys.argv[1]).write_text(sys.prefix)", marker}
    ctx := policyContext(t, executable, args)
    _, set, err := ResolveToolReference(ctx, "McpToolset", mcpConfigArgs(executable, args))
    if err != nil { t.Fatal(err) }
    runCtx, cancel := context.WithTimeout(t.Context(), 5*time.Second)
    defer cancel()
    mock := agent.NewStrictContextMock(runCtx)
    _, _ = set.Tools(&mock) // The child writes its prefix, then exits without a handshake.
    actual, err := os.ReadFile(marker)
    if err != nil { t.Fatal("child did not write its prefix") }
    if string(actual) != venv { t.Fatal("child did not retain the virtual environment") }
}
```

Output from rerunning the reproduction above:

```text
=== RUN   TestManualMCPVenv
--- PASS: TestManualMCPVenv (0.03s)
PASS
```

</details>

On Windows Server 2025 with Go 1.26.6, the original extensionless alias failed all four symlink cases. Adding `.exe` on Windows passed all five cases, with none skipped. [Before-and-after run](https://github.com/ktsoator/adk-go/actions/runs/34814537282).

### Checklist

- [x] Read the contribution guide.
- [x] Added regression tests and usage documentation.
- [x] Verified local subprocess startup and Python venv behavior.
- [x] Reran checks and completed an independent review after rebase.

### Additional context

[ADK Python](https://github.com/google/adk-python/blob/460715b6c62c8e9ab00931c502381ee0364e39b6/src/google/adk/tools/mcp_tool/mcp_toolset.py#L630) also rejects stdio MCP configs by default. Its opt-in switch allows all such configs. This change uses exact command and argument approvals as requested in #1569.


ktsoator:
During self-review, I noticed that `approved-launcher` was missing the `.exe` suffix used by the other Windows test fixture. I ran the tests on Windows and confirmed that all four symlink cases failed. Adding the suffix made all five cases pass, with none skipped. Fixed in a5fa6de.

[Windows verification](https://github.com/ktsoator/adk-go/actions/runs/34814537282)

Happy to adjust if there's a preferred way to handle executable aliases in these tests :)

PR #1593, opened 2026-09-14
ci: require a linked issue, and document stale pull request takeover

jjsasha63:
### Linked issue

None. I found the bot was inert while working on the stale-PR section of
CONTRIBUTING.md, and the fix touches the same files.

### Describe the change

**Problem:**

There are two unrelated gaps in the contribution process.

The first is that the `require-linked-issue` bot has never posted a comment. It
has run 258 times since 3 September and commented on nothing. It exempted any PR
whose description ran to 25 words or more. CONTRIBUTING.md offered that
description as an alternative to opening an issue, and no PR has ever failed both
checks.

The second is that CONTRIBUTING.md says nothing about a PR that stops moving
because its author has gone quiet. Maintainers have no stated basis for finishing
the work. Contributors get no warning that it can happen, and no way to keep
their place.

**Solution:**

This PR has two commits.

The first drops the word-count exemption, so only a linked issue exempts a PR.
CONTRIBUTING.md and the PR template both still offered the description as an
alternative, so it updates those too.

It also changes how a link is detected. The old test matched a bare `#123`
anywhere in the body, which let "split out of #1424" pass as a link. The new one
needs a linking phrase in front of the number. It also accepts `owner/repo#N`,
`GH-N` and issue URLs.

Reviewing that turned up three more problems, fixed in the same commit:

- Skip PRs that do not target the default branch. A v1 backport cites the PR it
  replays. GitHub also resolves closing keywords against the default branch only,
  so the GraphQL check cannot see a link that is really there.
- Skip the PR when the collaborator permission lookup fails for anything other
  than a 404. It used to read any failure as "not a maintainer". A rate-limit 403
  would have posted up to five wrong comments in one run. The triage role now
  counts as a maintainer, which GitHub reports as `read`.
- Replace `\s*:?\s*` with `[\s:]*`. The first backtracks quadratically. 65k
  spaces after "closes" took 5.3s, and a sweep tests up to 50 bodies.

The second commit adds a "Stale pull requests" section to CONTRIBUTING.md. An
author has two weeks to respond to whatever the PR is waiting on. After that a
maintainer may take the change over and finish it. Any reply from the author
stops the clock, and a request for another week counts. An author who comes back
after a takeover has started gets the PR back, unless a replacement has already
merged. Either way the original commits keep the author's name on them. Nothing
enforces this, so people apply it by hand.

### Behavior change

Nothing changes for anyone using the library. No Go code is touched.

It changes for contributors. A PR against `main` from a non-maintainer with no
linked issue now gets one comment asking for one. Describing the change in the PR
body no longer avoids that comment.

### Testing Plan

There are no automated tests. This is a scheduled workflow and the repository has
no harness for one.

Instead I extracted the script from the YAML and ran it against the live
repository, with `createComment` stubbed out so nothing was posted. I did two
runs:

1. As shipped, with the 24-hour window and open PRs. It comments on nothing,
   which matches the last 258 scheduled runs.
2. A replay over the 100 most recent PRs, with the window widened and the
   stampede guard lifted. Every predicate is unchanged.

The replay comments on 9 PRs. I checked each against GitHub's GraphQL
`closingIssuesReferences`. All 9 return 0, target `main`, and come from a
non-maintainer.

The other 91 break down as 38 with a linking phrase, 25 maintainers, 22 not
targeting `main`, 5 drafts or bots, and 1 docs-only.

Before the three follow-up fixes, the same replay commented on 11. #1488 dropped
out because it is a v1 backport citing PR #1394. #1457 dropped out because it
left the 100-PR window when this PR was opened.

I also ran the regexp against 24 cases. They cover every linking form I found in
this repository and every incidental mention the replay turned up. The untouched
PR template is one of them, and it correctly still counts as unlinked.
Backtracking measured 5.3s before the fix and 0.3ms after, for 65k spaces.

**With your source change reverted, which test fails?**

There are no tests, so the replay stands in for one. With the word-count
exemption restored it comments on 0 PRs, which is the bug this fixes.

**Unit Tests:**

- [x] N/A. No Go code changed.

The dry run above is as close as I can get without posting to a real PR. The
comment path has never executed in production, so the `pull-requests: write`
token and the comment body are still unproven. The workflow has
`workflow_dispatch`, so it can be triggered once after merge to confirm.

### Checklist

- [x] N/A. No Go code, so no unit tests to add or run.

### Additional context

The two commits are independent. I am happy to split the CI change into its own
PR if you would rather review them separately.


baptmont (approved):
Approving.

Went at the workflow with a security hat on first, since that is the risky file here. Triggers are still `schedule` plus `workflow_dispatch` with no `pull_request_target`, permissions actually go *down* (`contents: read` dropped now the template is not fetched), the action stays SHA-pinned, and nothing untrusted reaches the script through `${{ }}` — `pr.body` and `pr.user.login` are API values in JS, not template expansions. YAML and the inline script both parse, and the `#requirement-for-prs` anchor the comment links to still resolves.

The ReDoS note is worth more than it looks. I ran both forms: `\s*:?\s*` takes 23s on 65k spaces on this machine, `[\s:]*` takes 1ms. Worth being explicit for a later reader that this is avoiding a ReDoS in new code rather than fixing a shipped one — what is on main today is the simple bare-`#123` match.

On the tightening, I measured it against the last 60 PRs. The raw regex delta is 16, but after the workflow's own filters it is 3, and all three are true positives: #1548's body reads "Discovered while validating #1547", which is the example in your own comment. The new default-branch guard is doing real work — it alone spares 10 of the v1 backports, which cite PR numbers in prose and would otherwise all get nagged. 25/25 on the phrase cases too, including `closes google/adk-go#123`, `GH-123` and the full issue URL.

One thing I would rather have a second maintainer on than call myself: the two-week stale-PR policy in CONTRIBUTING is governance, not code. It reads fair to me and keeping the original author's commits is the right instinct, but it deserves an explicit yes from someone else — particularly since #1596 has already taken a PR over under a policy that has not merged yet.

Needs a merge from `main` before it lands (behind, and CI here is from 14 September).

